PR #8331, opened 2026-06-22
feat(teams): gate bulk invite behind Scale plan and rate-limit invites

Dhruwang:
## Why

We saw a spam incident where the team bulk-invite flow was abused to send a large volume of invitations. This locks high-volume invites behind a paid plan and adds rate limits so the vector is bounded even if the UI is bypassed.

### 1. Bulk invite is now a Scale-plan feature (Formbricks Cloud)
- New `bulkInviteUsersAction` server action:
  - Requires organization **owner/manager**.
  - On cloud, enforces the org plan is in `{scale, custom}` — otherwise `OperationNotAllowedError`.
  - Caps a single import at **500 invitees** and validates roles for the whole batch up front.
  - Skips individual failures (e.g. duplicate invite) instead of aborting the whole batch, returning a per-email success list.
- The CSV bulk tab now routes to this action; **individual invites are unchanged**.
- Self-hosted keeps bulk invite available (no Stripe plan to gate on).

### 2. Per-organization invite rate limits (defense-in-depth)
- `inviteMember`: **30/hour per org** applied to `inviteUserAction` — bounds the looped single-invite path, which is the same underlying abuse vector.
- `bulkInviteMembers`: **10 imports/hour per org** applied to the bulk action.

### 3. UI
- When the plan does not allow bulk invite, the bulk tab renders the shared **`UpgradePrompt`** component (Upgrade plan / Request trial license + Learn more) instead of the uploader — consistent with other gated features.

## Notes / decisions
- `custom` (enterprise) plans are allowed in addition to `scale`.
- Rate-limit caps (`30/hr`, `10/hr`) are a starting point and easy to tune in `rate-limit-configs.ts`.

## Verification
- `pnpm turbo run build --filter=@formbricks/web` — production build + typecheck pass.
- Teams test suite green (`46/46`).


jobenjada (dismissed):
confirmed that the UX is as discussed.

Didn't test the rate limting in the single invite though

BhagyaAmarasinghe (commented):
.

Dhruwang:
Thanks for the review 

"Bulk import is limited by import count, not recipient count.
The new bulk action allows 500 invitees per import and 5 imports/day, so an entitled org can still send up to 2,500 invite emails/day. It does not consume the inviteMember 20/day limiter inside the loop. That may be acceptable for Scale customers, but it is not a hard abuse cap." is planned for next ticket

BhagyaAmarasinghe (approved):
Great then, LGTM!

Dhruwang:
Screenshot 2026-06-22 at 8 29 14 PM


Rate limit in action

PR #8337, opened 2026-06-23
feat(auth): add Better Auth foundation (additive, pre-cutover) [ENG-1054]

xernobyl:
## What does this PR do?

**PR 1 of 4 — the base of the ENG-1054 Better Auth migration stack** (base: `epic/better-auth-migration`). Adds the entire Better Auth foundation **additively**: NextAuth still owns sessions on this branch, so nothing is cut over here and existing flows are unaffected.

- `betterAuth()` instance (`modules/auth/lib/auth.ts`) — hardened config, bcrypt password verify (reuses the existing crypto), node-redis secondary storage, `accountLinking.enabled:false`, `twoFactor` + `nextCookies` plugins, cuid2 id generation.
- Providers (enterprise-gated): Google/GitHub via `socialProviders`; Azure AD / OIDC / SAML-over-BoxyHQ via `genericOAuth`; JIT provisioning + verify-before-link SSO recovery re-expressed through hooks.
- Account-deletion config (asymmetric: password for credential users, email-link for SSO), observability parity (audit + Sentry), and the `"token"` email sign-in port (email verification + the `/sso-recovery/sign-in` endpoint).
- Schema foundation + idempotent data migrations (credential / SSO / 2FA backfill) so existing users keep working without re-registration.
- A net-new integration-test harness driving the real BA handler against a real Postgres.

Subsequent PRs in the stack: client UI swaps → cutover flip → NextAuth decommission. The whole stack merges into `epic/better-auth-migration` (= current main).

ENG-1054.

## How should this be tested?

- `pnpm test` (unit) and `pnpm test:integration` (real Postgres) green.
- `tsc --noEmit -p tsconfig.json` clean.
- Additive only — manually confirm existing NextAuth sign-in/up/2FA/SSO still work on this branch.

### Required
- [ ] Ran `pnpm build` — CI runs the production build
- [x] Merged the latest changes from main (the `epic/better-auth-migration` base is current main)


apps/web/app/(app)/layout.tsx
```diff
@@ -10,13 +9,13 @@ import {
 } from "@/lib/constants";
 import { getUser } from "@/lib/user/service";
 import { NextAuthProvider } from "@/modules/auth/components/next-auth-provider";
-import { authOptions } from "@/modules/auth/lib/authOptions";
+import { getSession } from "@/modules/auth/lib/session";
 import { ClientLogout } from "@/modules/ui/components/client-logout";
 import { NoMobileOverlay } from "@/modules/ui/components/no-mobile-overlay";
 import { ToasterClient } from "@/modules/ui/components/toaster-client";
 
 const AppLayout = async ({ children }: { children: React.ReactNode }) => {
```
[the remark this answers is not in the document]
xernobyl: Done in 304641b44 — `AppLayout` props are now wrapped in `Readonly<>` per the repo TSX convention.

apps/web/modules/auth/layout.tsx
```diff
@@ -1,13 +1,12 @@
-import { getServerSession } from "next-auth";
 import { redirect } from "next/navigation";
 import { Toaster } from "react-hot-toast";
 import { getIsFreshInstance } from "@/lib/instance/service";
-import { authOptions } from "@/modules/auth/lib/authOptions";
+import { getSession } from "@/modules/auth/lib/session";
 import { getIsMultiOrgEnabled } from "@/modules/ee/license-check/lib/utils";
 
 export const AuthLayout = async ({ children }: { children: React.ReactNode }) => {
```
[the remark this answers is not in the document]
xernobyl: Done in 304641b44 — `AuthLayout` props wrapped in `Readonly<>`.

apps/web/modules/setup/(fresh-instance)/layout.tsx
```diff
@@ -1,10 +1,9 @@
-import { getServerSession } from "next-auth";
 import { notFound, redirect } from "next/navigation";
 import { getHasNoOrganizations, getIsFreshInstance } from "@/lib/instance/service";
-import { authOptions } from "@/modules/auth/lib/authOptions";
+import { getSession } from "@/modules/auth/lib/session";
 
 export const FreshInstanceLayout = async ({ children }: { children: React.ReactNode }) => {
```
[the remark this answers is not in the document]
xernobyl: Done in 304641b44 — `FreshInstanceLayout` props wrapped in `Readonly<>`.

apps/web/lib/env.ts
```diff
@@ -283,6 +283,10 @@ const parsedEnv = createEnv({
     MAIL_FROM: z.email().optional(),
     NEXTAUTH_URL: z.url().optional(),
     NEXTAUTH_SECRET: z.string().optional(),
+    // Better Auth (ENG-1054). Optional during the additive migration; BA requires a strong
+    // (>=32 char) secret in production and throws if unset there.
+    BETTER_AUTH_SECRET: z.string().optional(),
+    BETTER_AUTH_URL: z.url().optional(),
```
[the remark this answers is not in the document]
xernobyl: Done in 12bc7a427 — `BETTER_AUTH_SECRET` is now `z.string().min(32).optional()`, so a weak secret fails env validation when one is set.

apps/web/modules/auth/lib/cutover/reencode-two-factor.integration.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    .map((c) => c.split(";")[0])
+    .join("; ");
+
+beforeEach(async () => {
+  await resetDb();
+});
+
+describe("2FA secret re-encode (real Postgres)", () => {
+  test("an existing NextAuth-era 2FA user verifies their current authenticator code via BA after re-encode", async () => {
+    authenticator.options = { digits: 6, step: 30 };
```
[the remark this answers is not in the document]
xernobyl: Done in 42ea72f47 — `authenticator.options` is captured once and restored in `afterEach`, so the test no longer leaks global TOTP state across tests.

apps/web/integration/global-setup.ts
```diff
@@ -0,0 +1,56 @@
...
+const REDIS_CONTAINER = process.env.TEST_REDIS_CONTAINER ?? "formbricks-valkey-1";
+const REDIS_TEST_DB = process.env.TEST_REDIS_DB ?? "15";
+
+const sh = (cmd: string): string => execSync(cmd, { stdio: "pipe" }).toString();
+
+export default function setup(): void {
+  const adminPsql = (sql: string) =>
+    sh(`docker exec ${CONTAINER} psql -U postgres -q -c ${JSON.stringify(sql)}`);
+  const testPsql = (sql: string) =>
+    sh(`docker exec ${CONTAINER} psql -U postgres -q -d ${TEST_DB} -c ${JSON.stringify(sql)}`);
```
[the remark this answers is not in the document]
xernobyl: Done in 53a139164 — the harness `TEST_*` overrides are validated against a safe pattern (alphanumerics + `_.-` for names, digits for the Redis DB index) before any shell interpolation; a hostile value aborts the harness loudly instead of being interpolated into `docker exec`.

apps/web/modules/ee/sso/lib/sso-provisioning.ts
```diff
@@ -0,0 +1,179 @@
...
+    return { action: "reject", reason: "missing_default_team_id" };
+  }
+
+  // When not skipping invites, require a valid invite token whose email matches the user's.
+  if (!SKIP_INVITE_FOR_SSO) {
+    if (!callbackUrl) return { action: "reject", reason: "missing_callback_url" };
+    try {
+      const url = new URL(callbackUrl);
+      const inviteToken = url.searchParams.get("token") || "";
+      const source = url.searchParams.get("source") || "";
```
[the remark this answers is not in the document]
xernobyl: Done in b39dd9d13 — switched to `new URL(callbackUrl, WEBAPP_URL)` so a root-relative invite callback (e.g. `/auth/signup?token=…`) parses instead of throwing `invite_token_validation_error`.

apps/web/modules/ee/sso/lib/better-auth-providers.ts
```diff
@@ -0,0 +1,158 @@
...
+                // Capture for verify-before-link recovery; name parity with the OIDC mapping.
+                captureSsoIdentity({ email: profile.email, providerAccountId: profile.sub });
+                return {
+                  email: profile.email,
+                  name:
+                    profile.name ||
+                    [profile.given_name, profile.family_name].filter(Boolean).join(" ") ||
+                    profile.preferred_username,
+                };
+              },
```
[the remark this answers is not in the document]
xernobyl: Done in 2ae22e88f — enabled `requireIssuerValidation` for Azure, gated on a configured `AZUREAD_TENANT_ID`. A single-tenant discovery issuer is stable and matches the token's `iss`; multi-tenant `common` tokens carry the caller's home-tenant issuer, which a strict check would reject — so it stays off there. That matches the caveat you flagged.

apps/web/modules/ee/sso/lib/better-auth-recovery-signin.integration.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+    const token = createToken(user.id, { purpose: "sso_recovery" });
+    const query = { token, callbackUrl: WEBAPP_URL };
+
+    await auth.api.ssoRecoverySignIn({ query, asResponse: true });
+    await auth.api.ssoRecoverySignIn({ query, asResponse: true });
+
+    // not single-use: the same token signs in twice (documents the known replay-until-expiry parity)
+    expect(await prisma.session.count({ where: { userId: user.id } })).toBe(2);
+  });
```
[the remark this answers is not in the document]
xernobyl: Verified, but leaving as-is for this PR (tracked for the SSO/auth security-hardening follow-up). The replayable-until-expiry behavior is intentional like-for-like parity with the legacy `token` provider's sso_recovery path. Single-use is a genuine hardening but a behavior change beyond the migration's scope, and non-trivial here: the recovery token is a stateless signed JWT, so single-use needs a consumed-token store/denylist. Abuse is already bounded by the 1-day expiry + the `/sso-recovery/*` rate limit, and the token only establishes a session for the already-verified user — the account LINK still goes through verify-before-link completion.

apps/web/modules/ee/sso/lib/better-auth-recovery-signin.ts
```diff
@@ -0,0 +1,74 @@
...
+          const user = await ctx.context.internalAdapter.findUserById(id);
+          if (!user) {
+            throw new Error("User not found");
+          }
+          const session = await ctx.context.internalAdapter.createSession(user.id, false);
+          if (!session) {
+            throw new Error("Failed to create recovery session");
+          }
+          await setSessionCookie(ctx, { session, user });
+          destination = getValidatedCallbackUrl(callbackUrl, WEBAPP_URL) ?? WEBAPP_URL;
```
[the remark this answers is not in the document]
xernobyl: The open-redirect angle is already covered: the destination is run through `getValidatedCallbackUrl(callbackUrl, WEBAPP_URL)` before redirecting, so a missing/invalid callback falls back to `WEBAPP_URL`, never an attacker URL. Establishing a session for the JWT-verified user is the intended recovery-sign-in behavior (parity with the legacy `token` provider). The deeper ask — scoping the token strictly to the completion route so it cannot act as a general sign-in — is the same recovery-flow redesign as the single-use-token thread above; tracking both together in the hardening follow-up.

packages/database/migration/20260619120000_eng_1054_credential_account_backfill/migration.ts
```diff
@@ -0,0 +1,99 @@
...
+        ORDER BY "id" ASC
+        LIMIT ${CREDENTIAL_USER_BATCH_SIZE}
+      `;
+    }
+    return tx.$queryRaw<TCredentialUserRow[]>`
+      SELECT "id", "password" FROM "User"
+      WHERE "password" IS NOT NULL
+      ORDER BY "id" ASC
+      LIMIT ${CREDENTIAL_USER_BATCH_SIZE}
+    `;
```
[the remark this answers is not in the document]
xernobyl: Intentionally NOT filtering on `isActive` here. Under Better Auth a `credential` Account row is required to sign in, so skipping deactivated users would leave a later-REACTIVATED user with no credential account and unable to log in — a regression. The backfill should mint the account for every user with a password (their bcrypt hash is preserved verbatim). Blocking deactivated users is the auth layer's responsibility, not the backfill's — and enforcing `isActive` at sign-in is tracked separately as a known gap. So the correct fix is the sign-in gate, not omitting the account.

packages/database/schema.prisma
```diff
@@ -866,6 +877,20 @@ model VerificationToken {
...
+  id          String  @id @default(cuid())
+  secret      String
+  backupCodes String
+  userId      String
+  user        User    @relation(fields: [userId], references: [id], onDelete: Cascade)
+  verified    Boolean @default(true)
+
+  @@index([userId])
+}
+
```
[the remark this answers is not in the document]
xernobyl: Agree a unique row per user is the right end state — holding it for the hardening follow-up rather than this PR. A DB unique index is only safe once we confirm Better Auth's `twoFactor` plugin keeps exactly one row per user across disable→re-enable (an insert-on-re-enable would break under the constraint). The TOCTOU is also only the one-time cutover re-encode (run once, idempotent), not a runtime path. Clean fix once verified: `@@unique([userId])` + an upsert in the re-encode.

xernobyl:
Heads-up @Dhruwang @BhagyaAmarasinghe @pandeymangg — pushed SonarCloud code-smell fixes since you were assigned. No behavior changes; tsc + all affected unit tests green.

- `secondary-storage.ts` — positive branch first (drops the negated condition)
- `better-auth-hooks.ts` — optional chain for the callback-path guard
- `reencode-two-factor.ts` — keyset cursor via `.at(-1)` with a type-safe guard (no `!`)
- `sso-provisioning.ts` — extracted the invite-token validation into a helper to bring `gateSsoProvisioning` under the cognitive-complexity limit. Parity logic unchanged (pinned by `sso-handlers.test.ts`).
- `email/index.tsx` — converted the resend-link `TODO` into a tracked post-cutover note (the dedicated BA resend endpoint is a real cutover item, in the runbook).
- `gen-boolean-client.mjs` — the Account-model patch is now backtracking-free. Worth a look: the original `[\s\S]*?` **and** my first pass `[^}]*?` are both ~O(n²) — `[^}]` matches newlines too, so it overlaps the trailing `\n\s*type` and backtracks (measured ~85ms at n=8000 for both). It now locates the block by index and uses `[ \t]` (never `\n`) for intra-line whitespace, so the match can't run across lines (0.13ms at n=64000). Verified behavior-identical on the real schema.

packages/database/schema.prisma
```diff
@@ -866,6 +877,20 @@ model VerificationToken {
...
+/// at rest by Better Auth using BETTER_AUTH_SECRET.
+model TwoFactor {
+  id          String  @id @default(cuid())
+  secret      String
+  backupCodes String
+  userId      String
+  user        User    @relation(fields: [userId], references: [id], onDelete: Cascade)
+  verified    Boolean @default(true)
+
+  @@index([userId])
```
Dhruwang: Can we replace this with `@@unique([userId])`,  just to make sure that one user does not end up with two entries ?

packages/database/schema.prisma
```diff
@@ -906,6 +931,7 @@ model User {
   memberships               Membership[]
   accounts                  Account[]
   sessions                  Session[]
+  twoFactors                TwoFactor[]
```
Dhruwang: // after adding unique → can become one-to-one
  model User      { twoFactor  TwoFactor? }   // single, optional
  model TwoFactor { @@unique([userId]) }

packages/database/migration/20260619120000_eng_1054_credential_account_backfill/migration.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  while (hasMoreUsers) {
+    const users = await fetchBatch(lastProcessedUserId);
+    if (users.length === 0) {
+      hasMoreUsers = false;
+      continue;
+    }
+    stats.scanned += users.length;
+
+    for (const user of users) {
```
Dhruwang: PERF: writes aren't batched like reads are — one INSERT per user. Collapse to a single multi-row INSERT … VALUES … ON CONFLICT per batch (~1000× fewer statements, shorter lock window)

apps/web/modules/ee/sso/lib/better-auth-providers.ts
```diff
@@ -0,0 +1,163 @@
...
+              // BoxyHQ SAML bridge — points at the existing local Jackson endpoints (unchanged).
+              providerId: "saml",
+              clientId: "dummy",
+              clientSecret: "dummy",
+              authorizationUrl: `${WEBAPP_URL}/api/auth/saml/authorize`,
+              tokenUrl: `${WEBAPP_URL}/api/auth/saml/token`,
+              userInfoUrl: `${WEBAPP_URL}/api/auth/saml/userinfo`,
+              scopes: [],
+              pkce: true,
+              authorizationUrlParams: { provider: "saml" },
```
Dhruwang, line 149: Existing NextAuth SAML sends tenant/product to BoxyHQ /authorize; the BA generic-OAuth config sends only {provider:'saml'}. 🤔

apps/web/app/api/v1/management/me/lib/utils.ts
```diff
@@ -2,14 +2,15 @@ import { NextApiRequest, NextApiResponse } from "next";
 import type { Session } from "next-auth";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/modules/auth/lib/authOptions";
+import { getSession } from "@/modules/auth/lib/session";
 
 export const getSessionUser = async (req?: NextApiRequest, res?: NextApiResponse) => {
   // check for session (browser usage)
   let session: Session | null;
   if (req && res) {
     session = await getServerSession(req, res, authOptions);
```
Dhruwang, line 11: The req && res branch is dead — the sole caller (me/route.ts:162) is the App Router, with no arguments. Only tests are passing. To simplify, pass req/res and drop the params, allowing this to collapse into a single getSession() call. The raw getServerSession(req,res,authOptions) is a session read outside the session.ts choke point, which won't flip at Phase-7 cutover. I think we can remove it. 🤔

apps/web/modules/ee/sso/lib/sso-provisioning.ts
```diff
@@ -0,0 +1,189 @@
...
+          );
+        });
+        assigned = true;
+      } catch (error) {
+        // The user + account are already committed by Better Auth; never throw here (it would not
+        // roll them back and would break sign-in). On the final attempt, log an error for alerting:
+        // there is no automatic retry on later sign-ins, so a sustained failure needs manual
+        // reconciliation (the writes are idempotent, so an operational retry is safe).
+        if (attempt === MAX_ATTEMPTS) {
+          logger.error(error, "SSO provisioning: failed to assign new SSO user to its organization");
```
Dhruwang, line 175: This `logger.error` is the sole indication that a user has been orphaned (committed by the Business Analyst, but without an organization). The comment above confirms there's no automatic retry upon subsequent sign-ins. Comparing new vs legacy implementations: `sso-handlers.ts` at line 368 performs user+membership within a single transaction (atomic — orphaning is impossible); this runs in `user.create.after`, so the user is already committed by the time the handler mounts. The issue is dormant until the Business Analyst handler mounts, so it's not a merge blocker. However, before Phase 7, it's recommended to implement self-healing on existing-user SSO sign-ins (re-run the idempotent assignment if no membership) or a durable retry job. At a minimum, wire an alert on this log line so orphaned users aren't left unnoticed.

packages/database/migration/20260619120000_eng_1054_credential_account_backfill/migration.ts
```diff
@@ -0,0 +1,99 @@
...
+        stats.inserted += 1;
+      } else {
+        stats.skippedExisting += 1;
+      }
+    }
+
+    lastProcessedUserId = users[users.length - 1].id;
+  }
+
+  console.log(
```
Dhruwang: We can also swap this with `logger.info`.

packages/database/schema.prisma
```diff
@@ -854,6 +861,10 @@ model Session {
   userId       String
   expires      DateTime
 
+  // Better Auth (ENG-1054): BA maps token←sessionToken, expiresAt←expires via session.fields.
```
Dhruwang: Is this comment relevant here?

Dhruwang (changes requested):
Hey Tiago, the PR looks good with solid testing coverage. I left a few minor comments. Could you please take a look? :)

packages/database/schema.prisma
xernobyl: Done in `51a770cdb` — added `@@unique([userId])` on `TwoFactor` and made `User`↔`TwoFactor` one-to-one. Verified safe: Better Auth's `enableTwoFactor` already does `deleteMany({userId})` → `create` (one row per user), so the constraint matches its intent, and the re-encode now uses `findUnique`. The foundation migration's index is now `UNIQUE` (the table is new + empty in this unreleased PR).

packages/database/schema.prisma
xernobyl: Done in `51a770cdb` — `twoFactors TwoFactor[]` → `twoFactor TwoFactor?`, enabled by the `@@unique([userId])` above. Nothing referenced the old `twoFactors` relation accessor.

packages/database/migration/20260619120000_eng_1054_credential_account_backfill/migration.ts
xernobyl: Done in `4462ddb4b` — one multi-row `INSERT … VALUES … ON CONFLICT DO NOTHING` per batch via `Prisma.join`; the affected-row count is `inserted`, the remainder counted as existing. Same idempotency, far fewer statements + a shorter lock window.

apps/web/modules/ee/sso/lib/better-auth-providers.ts
xernobyl: Checked — this is at parity, not a regression. The current NextAuth SAML provider (`providers.ts`) also sends only `{ scope, response_type, provider: 'saml' }` with `clientId: 'dummy'` — no `tenant`/`product` — and `/api/auth/saml/authorize` forwards the query params verbatim to Jackson's `oauthController.authorize`. So the BA config hands Jackson the same params as the production-proven NextAuth one. It's on the cutover SAML dry-run checklist (real IdP) as belt-and-suspenders. Happy to add `tenant`/`product` to `authorizationUrlParams` if you'd prefer them explicit, but that'd be a change vs. today rather than parity.

apps/web/app/api/v1/management/me/lib/utils.ts
xernobyl: Removed on the decommission PR (#8336), which migrates `me` to the session DAL — there `getSessionUser` is just `const session = await getSession(); return session?.user;` (no `getServerSession`, no `req/res`), so the read flips at the Phase-7 cutover via the `session.ts` choke point. Left as-is here on additive, where NextAuth still owns sessions (the `getServerSession` path is intentional transitional code).

apps/web/modules/ee/sso/lib/sso-provisioning.ts
xernobyl: Agreed — not a merge blocker (the `databaseHooks` don't mount until the Phase-7 cutover). And you're right about the semantics: the legacy flow was atomic (user+membership in one txn), while the BA flow writes membership post-commit in `user.create.after`, so a failure logs + retries once rather than rolling back. Tracking self-healing on existing-user SSO sign-in (re-run the idempotent assignment) as a pre-cutover follow-up in the cutover runbook.

packages/database/migration/20260619120000_eng_1054_credential_account_backfill/migration.ts
xernobyl: Done in `4462ddb4b` — `console.log` → `logger.info` (matching the other migrations on `@formbricks/logger`).

packages/database/schema.prisma
xernobyl: Done in `51a770cdb` — reworded so it covers both the mapping and the columns it sits above: `token←sessionToken, expiresAt←expires (mapped via session.fields); ipAddress + userAgent below are new BA columns.`

Dhruwang (approved):
Thank you for the PR, looks good 🚀

apps/web/modules/auth/lib/auth.ts
```diff
@@ -0,0 +1,238 @@
...
+  user: {
+    additionalFields: {
+      identityProvider: { type: "string", required: false, input: false },
+      identityProviderAccountId: { type: "string", required: false, input: false },
+    },
+    // Account deletion (design doc §14): native Better Auth deleteUser with Formbricks' pre/post
+    // cleanup (sole-owner-org guard + org/invite removal, then Brevo + audit). Confirmation friction
+    // is asymmetric and wired at the edges in Phase 6 — password for credential users, email-link for
+    // SSO — so `sendDeleteAccountVerification` is intentionally NOT set here (it would email everyone).
+    deleteUser: accountDeletionConfig,
```
BhagyaAmarasinghe, line 172: This enables Better Auth's native `deleteUser` flow while intentionally leaving `sendDeleteAccountVerification` unset, so the raw `POST /api/auth/delete-user` endpoint can delete a fresh-session user without the Formbricks confirmation step. In Better Auth 1.6.18, the password is only verified when `body.password` is present; otherwise, with `freshAge: DAY_IN_SECONDS`, an empty-body request falls through to `beforeDelete` and deletes the user. That bypasses the intended policy here: credential users must confirm with password, and SSO users must confirm via the email-link callback. The new tests cover the password path and callback-token path, but not the empty-body fresh-session path. Please add a server-side guard so `/delete-user` rejects requests without a valid password/token (or otherwise forces SSO through the callback flow), plus an integration test proving `{}` does not delete a fresh-session user.

BhagyaAmarasinghe (changes requested):

apps/web/modules/auth/lib/auth.ts
xernobyl: Good catch — confirmed against `better-auth@1.6.18`: `POST /delete-user` verifies the password only when `body.password` is present; with no password it falls through on session freshness alone (the `freshAge` window, 1 day here), so an empty-body request from a recent session would delete the account with no confirmation.

Fixed in 641df3869 with `accountDeletionGuardPlugin` — a path-scoped `before` hook on `/delete-user` that rejects the request unless a password is supplied. Credential users send their password (Better Auth still verifies it); SSO users have none and stay on the `GET /delete-user/callback?token=…` email-link flow, which the matcher deliberately doesn't touch. `freshAge` stays in place as defense-in-depth. It's a plugin so it composes with the existing top-level `hooks.before` (the SSO license gate) instead of clobbering it.

Added an integration test that drives the real BA handler against Postgres: an empty-body `{}` delete from a fresh session is rejected, and neither the user nor its session row is removed.

Note the `[...all]` route is still unmounted on this branch, so the endpoint only goes live at the flip — where the decommission branch carries the equivalent server-side guard (`requireDeletionConfirmationBeforeHandler`).

BhagyaAmarasinghe (approved):

apps/web/modules/ee/sso/lib/sso-provisioning.ts
xernobyl: Filed this as **ENG-1601** so it's tracked durably beyond the runbook: https://linear.app/formbricks/issue/ENG-1601/sso-self-heal-orphaned-users-re-run-idempotent-membership-assignment

It captures the self-heal-on-next-sign-in approach (preferred — re-run the idempotent `provisionSsoUserMemberships`), the durable-retry-job alternative, and the at-minimum alert on the `sso-provisioning.ts:175` log line, plus acceptance criteria. Flagged as a gating item for the Phase-7 cutover (the `databaseHooks` don't mount until then, so it's dormant on this PR). Leaving this thread open as the in-PR marker.

PR #8338, opened 2026-06-23
feat(auth): migrate client auth UI to Better Auth [ENG-1054]

xernobyl:
## What does this PR do?

**PR 2 of 4 in the ENG-1054 Better Auth migration stack** (base: `feat/ENG-1054_better-auth-migration`). Swaps the **client-side** auth UI from `next-auth/react` to the Better Auth client. These changes are cutover-coupled (they go live with the flip in PR 3), so they're inert until `[...all]` is mounted.

- `useSignOut` → `authClient.signOut()` (keeps the NextAuth navigate/`{url}` contract; best-effort logout).
- Social buttons → `authClient.signIn.social`; Azure/OIDC/SAML → `authClient.signIn.oauth2` (providerIds match the server config); `errorCallbackURL` preserves the `?error=` surfacing.
- Login form 2FA → BA `twoFactorRedirect` + `verifyTotp`/`verifyBackupCode`; SAML `{tenant,product}` moved into the server `authorizationUrlParams`.
- Email-change sign-out + `DeleteAccountModal` (credential `deleteUser({password})` vs SSO email-link), `NextAuthProvider` → passthrough.

ENG-1054.

## How should this be tested?

- `tsc --noEmit` clean; ESLint clean.
- Inert on this branch (no `[...all]`); full validation happens with the flip (PR 3) + E2E at cutover.

### Required
- [ ] Ran `pnpm build` — CI runs the production build
- [x] Merged the latest changes from main (via the stack base, ultimately `epic/better-auth-migration` = current main)
### Appreciated
- [ ] UI changes are behind the cutover; screenshots will accompany the flip/E2E validation


xernobyl:
Heads-up @Dhruwang @BhagyaAmarasinghe @pandeymangg — since you were assigned: the SonarCloud foundation fixes from #8337 are merged up here, plus one client-specific fix.

- `use-sign-out.ts` — `window.location` → `globalThis.location` (Sonar portability finding). `use-sign-out.test.tsx` still green (8/8).

No behavior changes. See #8337 for the merged-up foundation fixes.

apps/web/modules/account/lib/better-auth-account-deletion.ts
BhagyaAmarasinghe, line 80: This enables Better Auth’s native `/delete-user` endpoint, but the password/email-link confirmation contract is only enforced by the modal. In Better Auth 1.6.18, `POST /delete-user` can delete without `password` or `token` when the session is still fresh; `auth.ts` sets `freshAge` to `DAY_IN_SECONDS`, so an empty-body `POST /api/auth/delete-user` can bypass both credential password confirmation and the SSO email-link flow.

Please enforce the confirmation server-side before deletion runs, e.g. reject `/delete-user` unless a password is supplied for credential accounts and keep SSO users on the `/delete-user/callback` token flow. Also add an integration test that an empty-body delete request does not delete a fresh-session user.

BhagyaAmarasinghe (changes requested):

apps/web/modules/account/lib/better-auth-account-deletion.ts
xernobyl: Confirmed against `better-auth@1.6.18` — `POST /delete-user` only verifies `body.password` when it's present, otherwise it proceeds on session freshness (`freshAge` = 1 day), so an empty-body request from a recent session deletes without confirmation. `beforeDelete` only receives `(user, request)` and can't see whether a factor was supplied, so the guard has to live in the request pipeline.

Fixed in 829d492b9 (merged up from the additive branch): `accountDeletionGuardPlugin`, a path-scoped `before` hook on `/delete-user` that rejects the request unless a password is supplied. This matches how the modal already drives deletion — credential users call `authClient.deleteUser({ password })` (the submit button stays disabled until the password field is non-empty), and SSO users go through `requestSsoAccountDeletionEmailAction()` → the `GET /delete-user/callback` token flow, which the matcher doesn't match. So a real client always satisfies the guard; it closes the direct-API bypass.

Added an integration test proving an empty-body `{}` delete from a fresh session is rejected and the account + session survive.

BhagyaAmarasinghe (approved):

apps/web/modules/auth/lib/auth.ts
BhagyaAmarasinghe: Cutover forces a one-time re-login, but I think we still need an `isActive` gate before Better Auth creates a new session. The credential account backfill migrates every user with `User.password IS NOT NULL`, including inactive users, and BA email/password sign-in can then create a fresh session from valid credentials without the legacy `authOptions` `user.isActive === false` rejection.

Can we add a `databaseHooks.session.create.before` check here that rejects inactive users before session creation? That should cover credential sign-in and SSO-created sessions. Please also add an integration test for an inactive migrated credential user: sign-in should fail and no `Session` row should be created.

BhagyaAmarasinghe (changes requested):

apps/web/modules/auth/lib/auth.ts
xernobyl: Good catch — confirmed the gap: the BA config had no `isActive` check (only the audit `create.after`), and the credential backfill migrates every `password IS NOT NULL` user including deactivated ones, so a deactivated user could sign in and get a session. The legacy `authOptions` `user.isActive === false` rejection wasn't reproduced anywhere.

Fixed in e43fe3b48 (merged up to client + decommission): a new `rejectInactiveUserOnSessionCreate` wired as `databaseHooks.session.create.before`. It runs on every session creation — the single choke point covering credential, SSO, and the recovery magic-link — and throws (`FORBIDDEN`, same "Your account is currently inactive…" message as authOptions) *before* the `Session` row is written, so a deactivated user neither signs in nor leaves an orphaned session. Session refresh goes through `updateSession` (not `create`), so still-active sessions are unaffected.

Added the integration test you asked for (`auth-email-password.integration.test.ts`): a verified-but-deactivated migrated credential user → `signInEmail` rejects with `/inactive/i` and `prisma.session.count()` stays `0`.

BhagyaAmarasinghe: Credential users can still take the SSO email-link deletion path by calling the server action directly. [actions.ts (line 19)](/tmp/formbricks-pr-8338-5d3a84e4/apps/web/modules/account/components/DeleteAccountModal/actions.ts:19) calls requestSsoAccountDeletionEmail() for any authenticated session, and [better-auth-account-deletion-request.ts (line 29)](/tmp/formbricks-pr-8338-5d3a84e4/apps/web/modules/account/lib/better-auth-account-deletion-request.ts:29) only checks that a session exists before minting a delete-account-* token. The modal only reaches this action when requiresPasswordConfirmation is false, but server actions are still callable directly. A credential user can therefore bypass the intended password-confirmation path and delete via the emailed callback link. Add a server-side account-type check before minting the token, e.g. reject identityProvider === "email" / credential accounts, and add an integration test that a credential user cannot mint the SSO deletion email.

xernobyl: Good catch — fixed in b2b45ae16 (merged up to decommission). `requestSsoAccountDeletionEmail` now resolves the account type and **rejects credential users (`identityProvider === "email"`) server-side before minting any token**, reusing the existing `requiresPasswordConfirmationForAccountDeletion` helper. So even though the server action is directly callable, a credential user can't take the SSO email-link path to sidestep the password confirmation — they get an `AuthorizationError` and no token/email is produced.

Integration tests (real Postgres, verified locally 3/3): a credential user is rejected (no token minted, no email sent, user untouched), and the happy-path test now seeds an SSO user (the only account type that path serves).

apps/web/modules/auth/login/components/login-form.tsx
```diff
@@ -99,44 +94,47 @@ export const LoginForm = ({
...
-            buildVerificationRequestedPath({
-              token: emailTokenActionResponse.data,
-              callbackUrl: resolvedCallbackUrl,
-            })
-          );
-        } else {
-          const errorMessage = getFormattedErrorMessage(emailTokenActionResponse);
-          toast.error(errorMessage);
+      // Step 1 — email + password.
+      const { data: signInData, error } = await authClient.signIn.email({
```
BhagyaAmarasinghe, line 114: Preserve the callback URL when resending verification emails

The page already resolves `resolvedCallbackUrl`, but this Better Auth sign-in call does not pass it to `authClient.signIn.email()`. Better Auth uses `body.callbackURL` when `sendOnSignIn` resends a verification email for an unverified user; without it, the verification link falls back to `/`.

That regresses invite/deep-link flows: an unverified user signing in from an invite can verify successfully but lands on `/` instead of the original invite/callback destination. Please pass `callbackURL: resolvedCallbackUrl` here and add a regression test for unverified sign-in from an invite/callback URL.

BhagyaAmarasinghe (changes requested):

BhagyaAmarasinghe:
@xernobyl Also please fix the coderabbit issues :)

apps/web/modules/auth/login/components/login-form.tsx
xernobyl: Good catch — fixed in c17235371 (merged to decommission). `authClient.signIn.email` now passes `callbackURL: resolvedCallbackUrl`, so Better Auth's `sendOnSignIn` verification resend builds its link around the intended destination instead of falling back to `/`.

The end-to-end regression (unverified sign-in from an invite → verify → land on the original destination) needs the `[...all]` route mounted, so it belongs with the cutover Playwright suite on the decommission PR (#8336) where the BA sign-in flow is actually live — flagged there as a cutover-validation item rather than a (routeless) client-branch unit test.

apps/web/modules/account/lib/better-auth-account-deletion-request.integration.test.ts
```diff
@@ -83,4 +86,15 @@ describe("requestSsoAccountDeletionEmail (real Postgres)", () => {
...
+
+  test("rejects a credential (email-identity) user — they must confirm with their password", async () => {
+    const email = "credrequest@example.com";
+    const userId = await createVerifiedUser(email, "Passw0rd!"); // identityProvider defaults to "email"
+    getSessionMock.mockResolvedValue({ user: { id: userId, email } });
+
+    // A directly-called server action must not let a credential user take the SSO email-link path.
+    await expect(requestSsoAccountDeletionEmail()).rejects.toThrow(/password confirmation/i);
+    expect(sendDeleteAccountConfirmationEmailMock).not.toHaveBeenCalled(); // rejected before minting/emailing
+    expect(await prisma.user.findUnique({ where: { id: userId } })).not.toBeNull(); // user untouched
```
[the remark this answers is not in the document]
xernobyl: Done in 582fed5b1 — the credential-rejection test now spies on `ctx.internalAdapter.createVerificationValue` and asserts it is never called, proving no `delete-account-*` token is minted (not just that no email is sent). That covers exactly the "refactor mints before throwing → live delete token" case you flagged.

apps/web/modules/auth/components/next-auth-provider.tsx
```diff
@@ -1,15 +1,15 @@
...
-  const refetchInterval = Math.min(Math.max(Math.floor(sessionMaxAge / 3), 60), 300);
-
-  return <SessionProvider refetchInterval={refetchInterval}>{children}</SessionProvider>;
+/**
+ * Better Auth's `useSession` (createAuthClient) fetches on demand and needs no React provider, so this
+ * is now a passthrough. Kept (with its props) so the layout import stays stable until the Phase 7
+ * cutover cleanup removes the wrapper. (ENG-1054)
+ */
+export const NextAuthProvider = ({ children }: NextAuthProviderProps) => {
+  return <>{children}</>;
```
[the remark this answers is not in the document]
xernobyl: Applied on the client branch (582fed5b1). Note this passthrough is removed entirely on the decommission PR (#8336) as part of the NextAuth teardown, so it's gone post-cutover regardless.

apps/web/modules/auth/login/components/login-form.tsx
```diff
@@ -99,44 +94,47 @@ export const LoginForm = ({
...
+
+      if (error) {
+        // Better Auth (re)sends its own verification email for an unverified address, so we only
+        // point the user at their inbox instead of minting a verification token ourselves.
+        if (error.code === "EMAIL_NOT_VERIFIED") {
+          toast.error(t("auth.login.please_verify_your_email_to_continue"));
+          return;
         }
+        toast.error(error.message ?? t("common.something_went_wrong"));
         return;
```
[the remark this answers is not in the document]
xernobyl: Leaving this as-is for now. The one case worth special-handling — `EMAIL_NOT_VERIFIED` — is already mapped to a localized key; the remaining branch falls back to Better Auth's message or the localized `common.something_went_wrong`. Mapping the full set of BA auth error codes to dedicated keys means adding strings across all nine locale files, which is an i18n task beyond this like-for-like cutover (the legacy flow surfaced raw messages here too). Better as a focused follow-up than bundled into the migration PR.

apps/web/modules/ee/sso/components/saml-button.tsx
```diff
@@ -1,24 +1,22 @@
...
 interface SamlButtonProps {
   returnToUrl?: string;
   lastUsed?: boolean;
-  samlTenant: string;
-  samlProduct: string;
   source: "signin" | "signup";
 }
 
-export const SamlButton = ({ returnToUrl, lastUsed, samlTenant, samlProduct, source }: SamlButtonProps) => {
+export const SamlButton = ({ returnToUrl, lastUsed, source }: SamlButtonProps) => {
```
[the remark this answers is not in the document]
xernobyl: Done in 582fed5b1 — `SamlButton` props wrapped in `Readonly<>`.

apps/web/playwright/utils/helper.ts
```diff
@@ -329,22 +329,10 @@ export const login = async (page: Page, email: string, password: string): Promis
...
-    json: "true",
-    csrfToken,
-  };
-
-  return page.context().request.post("/api/auth/callback/credentials", {
-    data,
+  // Better Auth sign-in (replaces the NextAuth csrf + credentials-callback flow). The signed session
+  // cookie set on the response is shared with the browser through Playwright's request/context jar.
+  return page.context().request.post("/api/auth/sign-in/email", {
+    data: { email, password },
```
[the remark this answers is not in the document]
xernobyl: By design for the stacked cutover. On this client branch the Better Auth `[...all]` route isn't mounted yet (NextAuth's `[...nextauth]` still owns `/api/auth/*`); `apiLogin` targets the post-flip BA endpoint. The route is mounted on the decommission/flip PR (#8336), where the Playwright suite exercises the real `/api/auth/sign-in/email` flow end-to-end. So the helper is intentionally cutover-coupled — validated at #8336, not on the client branch in isolation.

BhagyaAmarasinghe (approved):

PR #8347, opened 2026-06-24
refactor(navigation): dedupe org/workspace switcher into useSwitcherData

Dhruwang:
## Why

Stacked on #8342. While reviewing the settings route-scoping work we found the organization / workspace **switcher** logic — "load on dropdown open → sort by name → surface a load error" — independently re-implemented in five components. That's five copies to keep in sync, and they had already drifted (different error handling, retry, and init-state behavior).

## What changed

Extracted a single `useSwitcherData` hook and pointed every switcher at it.

- **New** `apps/web/modules/settings/hooks/use-switcher-data.ts` — owns `items / isLoading / error / hasLoaded`, loads once (guarded against duplicate/looping loads), sorts by name, and exposes `load()` / `retry()`. An optional `onError` callback lets the breadcrumbs keep their Sentry + logger reporting.
- Migrated all five consumers — each keeps its own render markup and navigation target; only the data concern is shared:
  - `MainNavigation`
  - `settings-navigation` (the standalone settings shell)
  - `landing-sidebar`
  - `organization-breadcrumb`
  - `workspace-breadcrumb`
- Unit test for the hook (load+sort, load-once guard, error fallback, retry).

No behavior change intended.

## Verification

- `tsc` clean on all touched files
- `vitest`: hook test + settings/auth/nav component suites green (244 tests)

## Follow-up (not in this PR)

Discussed during review, intentionally deferred to a **third PR**:

1. **No-workspace owner lockout.** `/organizations/[id]/landing` has no "create workspace" CTA, and the page hardcodes `isOwnerOrManager={false}` / `organizationWorkspacesLimit={0}` on its switcher — so an org owner with zero workspaces is stuck on a warning screen with no way forward. Fix: add a create-workspace CTA and pass the real membership role + workspace limit.
2. **Share switcher building blocks, not the whole shell.** `landing-sidebar` still carries its own copy of the user dropdown and the org-switcher *render* markup. Point it at the shared `UserDropdown` and extract a presentational `OrganizationSwitcher` so the landing sidebar, breadcrumbs, and settings sidebar share the markup too — without merging onboarding and settings into one over-branched component.


apps/web/app/(app)/workspaces/[workspaceId]/components/SettingsSidebarContent.tsx
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/components/SettingsSidebarContent.tsx]: pls fix this sonar warning

itsjavi:
the CI is failing @Dhruwang please take a look

itsjavi (changes requested):

apps/web/app/(app)/(onboarding)/organizations/[organizationId]/landing/components/landing-sidebar.tsx
```diff
@@ -116,53 +81,16 @@ export const LandingSidebar = ({ user, organization, publicDomain }: Readonly<La
...
-            )}
+            <SwitcherDropdownBody
+              type="organization"
+              isLoading={organizationSwitcher.isLoading}
+              error={organizationSwitcher.error}
+              onRetry={organizationSwitcher.retry}
+              items={organizationSwitcher.items}
+              selectedId={organization.id}
+              onSelect={handleOrganizationChange}
+              showSettings={false}
```
Comment (itsjavi) [inline apps/web/app/(app)/(onboarding)/organizations/[organizationId]/landing/components/landing-sidebar.tsx]: **must fix:** this drops the Settings item the landing sidebar had on main (it navigated to `/organizations/{id}/settings/general`), and UserDropdown only covers account settings, so users stuck on the landing page now have no way into org settings at all. The shared body's default settingsPath for "organization" is exactly the old target, so just removing `showSettings={false}` restores parity.

apps/web/modules/settings/components/switcher-dropdown-body.tsx
```diff
@@ -0,0 +1,123 @@
...
+  onSelect,
+  children,
+  showSettings = true,
+}: Readonly<SwitcherDropdownBodyProps>) => {
+  const { t } = useTranslation();
+  const router = useRouter();
+  const [, startTransition] = useTransition();
+  const { icon: Icon, settingsPath } = SWITCHER_CONFIG[type];
+  const headerLabel =
+    type === "organization" ? t("common.choose_organization") : t("common.choose_workspace");
```
itsjavi, line 66: **nit:** consolidating on `choose_*` changes the visible header in MainNavigation and the landing sidebar (they said "Change organization/workspace" before), and the collapsed-mode aria-labels in MainNavigation still use `change_*`, so the accessible name and the visible header now disagree. Fine to standardize on one wording, but let's align the aria-labels too (and note it in the PR body since it's technically a copy change).

apps/web/modules/settings/hooks/use-switcher-data.ts
```diff
@@ -0,0 +1,91 @@
...
+          setItems(sorted);
+          hasDataRef.current = sorted.length > 0;
+        } else {
+          const message = (result ? getFormattedErrorMessage(result) : "") || fallbackError;
+          setError(message);
+          hasErrorRef.current = true;
+          onErrorRef.current?.(message);
+        }
+      } catch (err) {
+        console.error(err);
```
itsjavi: **nit:** since the hook already exposes `onError` for reporting (and the breadcrumbs wire it to logger + Sentry), this bare `console.error` feels redundant; either route it through `@formbricks/logger` or drop it and let `onError` own the reporting. Sonar also tends to complain about console statements.

apps/web/app/(app)/workspaces/[workspaceId]/components/workspace-breadcrumb.tsx
```diff
@@ -152,7 +136,7 @@ export const WorkspaceBreadcrumb = ({
   return (
     <BreadcrumbItem isActive={isWorkspaceDropdownOpen}>
       <DropdownMenu onOpenChange={setIsWorkspaceDropdownOpen}>
-        <DropdownMenuTrigger className="flex cursor-pointer items-center gap-1 outline-hidden" asChild>
+        <DropdownMenuTrigger className="flex cursor-pointer items-center gap-1 outline-none" asChild>
```
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/components/workspace-breadcrumb.tsx]: **nit:** this flips `outline-hidden` to `outline-none`, which in Tailwind v4 are different utilities: `outline-none` is a real `outline: none` and drops the forced-colors-safe transparent outline that `outline-hidden` keeps. The org breadcrumb trigger right next to this still uses `outline-hidden`, so unless this was deliberate I'd revert to keep the two consistent.

itsjavi (changes requested):
Nice cleanup overall; the ref-based guard on the duplicate fetch and the sanitized error path look good, and lint/typecheck/unit tests all pass locally on the branch.

Found one regression though (Settings entry gone from the landing sidebar dropdown), plus a few small things; details inline.

One non-blocking thought: `renderSwitcherContent` in SettingsSidebarContent is now a third copy of the same spinner/error/retry/list markup that SwitcherDropdownBody owns. I know the PR body defers presentational unification, so no action here, just worth folding into that follow-up.

apps/web/app/(app)/(onboarding)/organizations/[organizationId]/landing/components/landing-sidebar.tsx
Comment (Dhruwang) [reply apps/web/app/(app)/(onboarding)/organizations/[organizationId]/landing/components/landing-sidebar.tsx]: Nice catch !

itsjavi (approved):

PR #8348, opened 2026-06-24
feat(workflows): v3 read API for workflow runs (list + detail)

Dhruwang:
## Summary

Adds the production read API for **workflow runs** in `@formbricks/workflows` + the v3 Next routes, ahead of the runs-list UI (ENG-1230). Two endpoints:

- `GET /api/v3/workflows/runs` — cursor-paginated list of run **summaries** for a workspace, newest-first.
- `GET /api/v3/workflows/runs/{runId}` — full run **resource** (trigger payload, run data, idempotency/retry fields) with its step logs ordered by sequence.

## Endpoints

**List** — `GET /api/v3/workflows/runs`
- Required: `workspaceId`. Optional: `workflowId`, `responseId`, `filter[status][in]=…`, `filter[isDryRun][eq]=true|false`, `limit` (1–250), `cursor`.
- Returns `{ data: WorkflowRunSummary[], meta: { limit, nextCursor } }`, ordered `createdAt desc, id desc`.

**Detail** — `GET /api/v3/workflows/runs/{runId}`
- Returns the full `WorkflowRunResource` including `triggerPayload`, `data`, `logs[]` (ordered by `sequence`), `idempotencyKey`, `nextAttemptAt`, `lastErrorAt`.

## Architecture

Mirrors the existing v3 workflows endpoints:

- **Route** (`apps/web/app/api/v3/workflows/runs/**`) — thin Next adapter via `withV3ApiWrapper`, builds the API context, delegates.
- **Handlers** (`@formbricks/workflows`) — `listRuns` / `getRun`: parse query, authorize, call the service, validate output against the Zod contracts, shape the response.
- **Service** — `listWorkflowRuns` / `getWorkflowRun`, talking to the injected `prisma` through the narrow `WorkflowsDb` port. The package keeps no dependency on `@formbricks/database`.
- **Cursor** — opaque base64url keyset cursor over `(createdAt, id)`; over-fetch by one to compute `nextCursor`. Malformed cursors are rejected as `400`.

## Authorization & safety

- The detail endpoint loads a run by id, then authorizes against **that run's workspace**, returning **403 (never 404)** for unknown or cross-workspace runs so existence is not leaked — matching the by-id workflow handlers.
- The list endpoint is scoped to the authorized `workspaceId`; a workspace the caller can't access is denied with 403.
- `triggerType` is kept as `string` in the port (the column is a free-form `String`, not an enum) and narrowed at the response boundary by `validateOutput`; documented inline so it isn't "tightened" in a way that breaks the structural Prisma wiring.

## Testing

- **Unit** — cursor encode/decode (incl. tamper/forgery → 400), serializers (summary/resource/log, null + non-null timestamp round-trips), service (filters, keyset where-clause, page-size boundary, malformed cursor), handlers (filters incl. `isDryRun=false`, empty page, limit bounds, cross-workspace 403, malformed cursor → 400), and the web api-client.
- **End-to-end (HTTP)** — verified the auth gate, every filter and combination, cursor pagination across pages (no overlap), the detail payload + ordered logs, cross-workspace and unknown-run **403**, and malformed-cursor / out-of-range-limit **400**s against a seeded local workspace.

## Out of scope

- **Run creation** is intentionally not here. Real runs are written by the workflow engine when the trigger fires; the manual/dry-run path is `POST /api/v3/workflows/{workflowId}/test` (ENG-1229), which lives under the workflow resource, not the `/runs` collection.
- Runs-list UI (next, ENG-1230).

### DIFF
--- a/packages/workflows/src/services/workflows.service.ts
+++ b/packages/workflows/src/services/workflows.service.ts
@@ -338,6 +338,9 @@ export const createWorkflowsService = ({ prisma }: { prisma: WorkflowsDb }): Wor
       where: buildRunListWhere(input, cursor),
       orderBy: [{ createdAt: "desc" }, { id: "desc" }],
       take: input.limit + 1,
+      // List summaries never use these large JSON columns; skip them so PG doesn't read/ship
+      // up to 100 big blobs per page. The detail `findUnique` still loads them.
+      omit: { triggerPayload: true, data: true },
     });
 
     const hasMore = rows.length > input.limit;
--- a/packages/workflows/src/services/workflows.service.ts
+++ b/packages/workflows/src/services/workflows.service.ts
@@ -1,4 +1,9 @@
-import type { TCreateWorkflowInput, TListWorkflowsInput, TWorkflowSortBy } from "../contracts";
+import type {
+  TCreateWorkflowInput,
+  TListWorkflowRunsInput,
+  TListWorkflowsInput,
+  TWorkflowSortBy,
+} from "../contracts";
 import { WorkflowConflictError, isUniqueConstraintViolation } from "../errors";
 import {
   type TWorkflowListCursor,
@@ -6,6 +11,12 @@ import {
   decodeWorkflowListCursor,
   encodeWorkflowListCursor,
 } from "../handlers/cursor";
+import {
+  type TWorkflowRunListCursor,
+  buildNextWorkflowRunListCursor,
+  decodeWorkflowRunListCursor,
+  encodeWorkflowRunListCursor,
+} from "../handlers/runs-cursor";
 import type { TWorkflowStatus } from "../types/common";
 import type { TWorkflowDefinition, TWorkflowExecutableDefinition } from "../types/document";
 import type {
@@ -13,6 +24,9 @@ import type {
   WorkflowDelegate,
   WorkflowOrderByInput,
   WorkflowRowWithLastRun,
+  WorkflowRunRow,
+  WorkflowRunWhereInput,
+  WorkflowRunWithLogsRow,
   WorkflowWhereInput,
   WorkflowsDb,
 } from "./ports";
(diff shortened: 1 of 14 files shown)


Dhruwang:
Addressed the CodeRabbit review — checked each against the current commit:

**1. `workflows.service.test.ts` — forbidden non-null assertions (Major)** ✅ Fixed in bde794ba2.
The `!` was already gone (replaced with `String(page.nextCursor)`), but that stringifies `null`→`"null"`. Replaced both cursor assertions with an explicit null-narrow + throw, as suggested — real type narrowing, no footgun, clear failure message. Lint clean, 86 tests pass.

**2. `parse-list-query.ts` — reject unknown query params (Minor)** ⏭️ Skipping — working as intended.
The lenient handling is deliberate and documented in the code: unrecognized params are ignored to match the **v3 survey list** convention so third-party callers can append their own params. The comments state this explicitly (the finding read them as promising a 400, which is the opposite of what they say). A bad *value* on a recognized key still throws a ZodError → 400; only unknown *keys* are ignored, by design.

**3. `workflows.handlers.test.ts` — exercise `responseId` (Minor)** ⏭️ Already covered.
The forwarding test already sends `responseId=${responseId}` in the request and asserts it in the `expect(service.listWorkflowRuns)` object (alongside workspaceId/workflowId/statusIn/isDryRun). The finding appears to have reviewed an earlier commit.

packages/workflows/src/services/workflows.service.ts
```diff
@@ -288,4 +330,30 @@ export const createWorkflowsService = ({ prisma }: { prisma: WorkflowsDb }): Wor
...
     return updateWorkflowRow(prisma.workflow, { workflowId, workspaceId }, { status: "disabled" });
   },
+
+  async listWorkflowRuns(input) {
+    const cursor = input.cursor ? decodeWorkflowRunListCursor(input.cursor) : null;
+
+    const rows = await prisma.workflowRun.findMany({
+      where: buildRunListWhere(input, cursor),
+      orderBy: [{ createdAt: "desc" }, { id: "desc" }],
+      take: input.limit + 1,
```
xernobyl, line 340: **[perf] List query over-fetches the large JSON columns.**

`findMany` has no `select`, so it pulls every column for each row — including `triggerPayload` and `data` (the full trigger snapshot + all step results, which can be large) — only for `toWorkflowRunSummary` to discard them and emit scalar fields. With `limit` up to 100, that reads up to 100 large JSON blobs from PG, ships them over the wire, and allocates them in memory per page for no benefit.

Suggest adding a `select` (or `omit: { triggerPayload: true, data: true }`) here and threading a `select`/`omit` through the `WorkflowRunDelegate.findMany` port type. The detail `findUnique` legitimately needs those columns; the list does not.

xernobyl (approved):
Thank you for the changes! LGTM!

PR #8349, opened 2026-06-24
feat: i18n - add canonical BCP-47 language map + normalizer (ENG-1067)

pandeymangg:
## What does this PR do?

Relates to **ENG-1067** (language-tag standardization) — this is the **foundation PR** in a phased series, and is intentionally **non-functional on its own**: it adds the single source of truth that the later PRs (catalog/picker, write-validation, surveys-runtime resolver, data migration) will all import, so they can't drift on what "canonical" means.

Adds `packages/i18n-utils/src/canonical.ts`:

- **`LANGUAGE_CANONICAL_MAP`** — `code → canonical tag` for every known input language in the system right now.
- **`CANONICAL_LANGUAGE_CODES`** — the deduped list of canonical tags (204), for allowlists/picker.
- **`normalizeLanguageCode(input)`** — resolves any incoming code (legacy bare code, mixed case, `_` separators, a URL `?lang=` value, an SDK-persisted value) to its canonical tag; returns `null` when unresolvable. Static-map first with a CLDR likely-subtags fallback (`Intl.Locale().maximize()`) for anything unrecognized.

**The standard**: one canonical tag per language, **always region-tagged** — `lang-REGION`, or `lang-Script-REGION` for script languages (Chinese). The default region for a bare/legacy code is the CLDR likely-subtags region. Examples: `de→de-DE`, `pt→pt-BR`, `ar→ar-EG`, `zh-Hans`/`zh-CN→zh-Hans-CN`, `zh-Hant`/`zh-TW→zh-Hant-TW`, `zh-HK→zh-Hant-HK`.

**Not in this PR**: updating the dropdown/catalog to use these codes, write-side validation, the surveys-runtime fallback chain, and the data migration. Nothing imports this module yet, so there is **no behavior change**.


xernobyl (approved):
Thank you for the PR. Looks great, maybe address the coderabbitai comment, if you see fit. I don't think it adds much.

PR #8352, opened 2026-06-24
feat: canonical language picker + write validation

pandeymangg:
Since we already have the normalization process for the language codes from #8349, this PR further adds the backend change when creating a language so any newly created language has a normalized code in the db. Also wires up the normalization changes in the dropdown picker on the workspace settings' language page so that the UI and backend are wired up together to create new language rows with normalized codes.

https://github.com/user-attachments/assets/482261d3-6eef-4fa7-91de-ced4d6e85a16


xernobyl (approved):
LGTM

PR #8355, opened 2026-06-25
fix: api/v3 - canonicalize survey language codes + legacy inbound back-compat (ENG-1067)

pandeymangg:
## What does this PR do?

Relates to **ENG-1067** (language-tag standardization) — part of the phased series, building on the foundation `canonical.ts` already merged into this epic.

Brings the **v3 Surveys API** language handling in line with the single canonical BCP-47 source, so v3 can no longer disagree with the picker, the DB, and the upcoming data migration. Two changes, both in `app/api/v3/surveys/language.ts`:

1. **Canonicalization sourced from the shared map.** The hardcoded `V3_LEGACY_LANGUAGE_CODE_MAP` — which carried v3's own, drifting opinions (`ar → ar-SA`, `no → nb-NO`, and no `pt`/`zh`) — is now **derived from `LANGUAGE_CANONICAL_MAP`** for a curated set of bare codes. Net effect: `ar → ar-EG`, `pt → pt-BR` (added), `no → no-NO`, and the values can never drift from the shared source again.

2. **Inbound back-compat on write.** `normalizeV3SurveyWriteLanguageCode` now also matches a request against the full CLDR-canonical form of the survey's configured codes. So once the data migration flips a stored code to its canonical tag (`gu → gu-IN`), a client still sending the legacy `gu` keeps resolving to the configured `gu-IN` instead of being rejected.

**Deliberately conservative:** codes outside the curated set (e.g. `vi`, `gu`, script-only `zh-Hans`) are **preserved as-stored** on read, so an unrelated patch never silently migrates a survey's language (which would orphan its content keys). The ENG-1067 data migration canonicalizes those remaining codes in one controlled pass.

No change to: the tag/locale normalizers, the read resolver and its ambiguity detection, alias/base-language matching, or PATCH replacement semantics.

## How should this be tested?

```bash
cd apps/web && pnpm vitest run app/api/v3/surveys
```

- `normalizeV3SurveyLanguageIdentifier`: `de → de-DE`, `pt → pt-BR`, `ar → ar-EG`; un-mapped legacy codes (`vi`/`gu`) return null so callers preserve them as-stored.
- `normalizeV3SurveyWriteLanguageCode`: legacy `gu` resolves to a configured `gu-IN` (post-migration inbound), while strict locale validation otherwise still rejects unknown bare codes.
- `resolveV3SurveyLanguageCode`: ambiguity (`en` with `en-US`+`en-GB`), alias, base-language, and disabled-language reads are unchanged.

### Required

- [ ] Ran `pnpm build` — _scoped `pnpm vitest run app/api/v3/surveys` (274 pass), ESLint, and `tsc --noEmit` pass locally; full build deferred to CI._
- [ ] Merged the latest changes from main onto my branch — _N/A: stacked on `epic/language-codes-stabilization` by design._
- [x] My changes don't cause any responsiveness issues — _API-only change, no UI._
- [ ] First PR at Formbricks? CLA — _N/A._

### Appreciated

- [ ] If a UI change was made: screenshots — _N/A, no UI change._
- [ ] Updated the Formbricks Docs if changes were necessary — _N/A; the supported-languages docs land with the catalog/docs PR._

### DIFF
--- a/apps/web/app/api/v3/surveys/language.test.ts
+++ b/apps/web/app/api/v3/surveys/language.test.ts
@@ -57,8 +57,9 @@ describe("normalizeV3SurveyLanguageIdentifier", () => {
     expect(normalizeV3SurveyLanguageIdentifier(input)).toBe(expected);
   });
 
-  test("does not guess ambiguous legacy identifiers", () => {
-    expect(normalizeV3SurveyLanguageIdentifier("pt")).toBeNull();
+  test("resolves bare codes to their CLDR default region (ENG-1067 canonical)", () => {
+    expect(normalizeV3SurveyLanguageIdentifier("pt")).toBe("pt-BR");
+    expect(normalizeV3SurveyLanguageIdentifier("ar")).toBe("ar-EG");
   });
 });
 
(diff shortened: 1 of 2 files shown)


xernobyl (approved):
LGTM

PR #8363, opened 2026-06-25
fix: harden response queue IndexedDB handling

### LINKED ISSUE #8029: [Bug] Fire-and-forget IndexedDB calls in ResponseQueue can silently lose responses
## Problem

`packages/surveys/src/lib/response-queue.ts` has three async calls dispatched as fire-and-forget without `.catch()` handlers. Each fails silently if IndexedDB rejects (quota exceeded, private-browsing quirks, transaction aborted, schema mismatch). Severity differs by site.

### Site 1 — line 120, `addPendingResponse` (highest severity)

```ts
if (this.config.persistOffline && this.config.surveyId) {
  void addPendingResponse({...}).then((dbId) => {
    if (dbId > 0) this.pendingDbIds.set(responseUpdate, dbId);
    this.processQueue();
  });
}
```

If `addPendingResponse` rejects: the `.then` callback never runs, `processQueue()` is never called, the response sits in the in-memory queue with no path forward, and the rejection leaks as unhandled. **The user sees no error but their response never gets sent.** This defeats the whole point of `persistOffline`.

### Site 2 — line 164, `syncPersistedResponses`

```ts
if (pendingCount > 1) {
  void this.syncPersistedResponses();
  return { success: false };
}
```

If sync throws: silent. The next `processQueue` call will re-attempt sync, so data isn't permanently lost — but every intermittent sync failure is invisible to telemetry, making debugging impossible.

### Site 3 — line 177, `removePendingResponse`

```ts
if (result.success) {
  const dbId = this.pendingDbIds.get(responseUpdate);
  if (dbId !== undefined) {
    void removePendingResponse(dbId);
    this.pendingDbIds.delete(responseUpdate);   // ← runs synchronously, unconditionally
  }
  ...
}
```

`this.pendingDbIds.delete(...)` runs synchronously. If the DB delete fails: in-memory state says "done," IndexedDB still has the row, next sync re-reads it and re-POSTs the same response. Depending on server-side dedup, either creates duplicates or leaves stale rows that fail to clear on every sync.

## Severity ranking

| Site | Failure mode | Severity |
| -- | -- | -- |
| 1 | Silent response loss when persistOffline on | High |
| 3 | Duplicate POSTs or stale-row buildup | Medium |
| 2 | Lost telemetry on sync failures | Low |

## Fix the issue proposes vs. what's actually needed

Adding `.catch()` is necessary but not sufficient.

**Site 1** needs a `.catch()` that ALSO calls `processQueue()` — otherwise IndexedDB failure silently disables the entire submission path. Treat `persistOffline` as best-effort:

```ts
void addPendingResponse({...}).then((dbId) => {
  if (dbId > 0) this.pendingDbIds.set(responseUpdate, dbId);
  this.processQueue();
}).catch((err) => {
  logger.error({ err, surveyId: this.config.surveyId }, "Failed to persist pending response to IndexedDB");
  this.processQueue();   // ← fall through so the response isn't deadlocked
});
```

**Site 3** should defer `this.pendingDbIds.delete(responseUpdate)` until *after* the DB delete resolves, so in-memory and persistent state stay consistent. Otherwise just adding `.catch()` still leaves the duplicate-submission risk.

**Site 2** is fine with a plain `.catch(err => logger.error(...))`.

## External fix branch

[https://github.com/mirzasaikatahmmed/formbricks/tree/fix/surveys-validation-indexeddb-styles-test](<https://github.com/mirzasaikatahmmed/formbricks/tree/fix/surveys-validation-indexeddb-styles-test>)

BhagyaAmarasinghe:
Fixes: ENG-953

## What changed

- Hardened `ResponseQueue` background persistence/sync calls so IndexedDB failures are logged and do not strand the in-memory response queue.
- Await IndexedDB cleanup before dropping queue-to-DB mappings, and track already-processed DB rows whose cleanup failed so later sync/count paths do not resend them.
- Added regression coverage for rejected persistence, count failures, background sync rejection, delayed/failed cleanup, cleanup retry, and sync read/delete failures.

## Root cause

`ResponseQueue` treated several IndexedDB operations as fire-and-forget work. If persistence or cleanup rejected, responses could stop progressing silently or be re-read from IndexedDB and submitted again later.

Fixes #8029

## Validation

- `pnpm exec prettier --write packages/surveys/src/lib/response-queue.ts packages/surveys/src/lib/response.queue.test.ts`
- `cd packages/surveys && NODE_OPTIONS=--max-old-space-size=8192 pnpm exec vitest run src/lib/response.queue.test.ts --pool=forks --maxWorkers=1 --reporter=dot`
- `pnpm --filter @formbricks/surveys test -- src/lib/response.queue.test.ts`
- `pnpm --filter @formbricks/surveys typecheck`
- `rm -rf packages/surveys/dist apps/web/public/js/surveys.* node_modules/.cache/turbo && pnpm build --filter=@formbricks/surveys... --force`

### DIFF
--- a/packages/surveys/src/components/general/survey.test.tsx
+++ b/packages/surveys/src/components/general/survey.test.tsx
@@ -32,8 +32,11 @@ vi.mock("@/lib/api-client", () => ({
 vi.mock("@/lib/offline-storage", () => ({
   addPendingResponse: offlineStorageMocks.addPendingResponse,
   countPendingResponses: offlineStorageMocks.countPendingResponses,
+  countPendingResponsesStrict: offlineStorageMocks.countPendingResponses,
   getPendingResponses: offlineStorageMocks.getPendingResponses,
+  getPendingResponsesStrict: offlineStorageMocks.getPendingResponses,
   removePendingResponse: offlineStorageMocks.removePendingResponse,
+  removePendingResponseStrict: offlineStorageMocks.removePendingResponse,
   clearSurveyProgress: offlineStorageMocks.clearSurveyProgress,
   getSurveyProgress: offlineStorageMocks.getSurveyProgress,
   patchSurveyProgressSnapshot: offlineStorageMocks.patchSurveyProgressSnapshot,
--- a/packages/surveys/src/lib/offline-storage.ts
+++ b/packages/surveys/src/lib/offline-storage.ts
@@ -130,54 +130,66 @@ export const addPendingResponse = async (entry: Omit<PendingResponseEntry, "id">
   }
 };
 
+export const getPendingResponsesStrict = async (surveyId: string): Promise<PendingResponseEntry[]> => {
+  const db = await openDb();
+  return new Promise((resolve, reject) => {
+    const tx = db.transaction(STORE_PENDING_RESPONSES, "readonly");
+    const store = tx.objectStore(STORE_PENDING_RESPONSES);
+    const index = store.index("surveyId");
+    const request = index.getAll(surveyId);
+
+    request.onsuccess = () => {
+      const results = (request.result as PendingResponseEntry[]).sort((a, b) => a.createdAt - b.createdAt);
+      resolve(results);
+    };
+    request.onerror = () => reject(request.error ?? new Error("IndexedDB request failed"));
+  });
+};
+
 export const getPendingResponses = async (surveyId: string): Promise<PendingResponseEntry[]> => {
   try {
-    const db = await openDb();
-    return new Promise((resolve, reject) => {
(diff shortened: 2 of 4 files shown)


xernobyl (dismissed):
LGTM

xernobyl (changes requested):
There seems to be some duplicated function reported by sonarqube. Can you have a look and see if it can be refactored or is not an issue? Other than that looks good :)

BhagyaAmarasinghe:
> There seems to be some duplicated function reported by sonarqube. Can you have a look and see if it can be refactored or is not an issue? Other than that looks good :)

Correct, fixed

xernobyl (approved):
Thank you for the changes. LGTM :)

PR #8364, opened 2026-06-25
fix: place Traefik labels on Formbricks service

BhagyaAmarasinghe:
## What changed

- Replaced the broad one-click installer awk range with `add_formbricks_traefik_labels`, which only targets the exact `formbricks` service.
- Inserted Traefik labels before the web service `environment:` block so labels no longer land on `formbricks-migrate`.
- Added focused regression coverage for HTTPS/HSTS, HTTP fallback labels, and a missing insertion-point failure.

## Why

Discussion #8340 reported that one-click installs could route Traefik to the migration container instead of the web app. The reproduced root cause was the old awk range starting on the migration service image line and inserting labels at the migration service `<<: *environment` anchor.

## Self-review

- No breaking issues found after review.
- One code-quality issue was fixed before publishing: the new installer regression tests were moved out of the RustFS-specific test file into a dedicated `formbricks-script.test.ts`.
- Live one-click installs still depend on the `stable` tag moving during a latest non-prerelease release.

## Validation

- `bash -n docker/formbricks.sh`
- `git diff --check HEAD~1..HEAD`
- `pnpm --filter @formbricks/storage test`
- `pnpm --filter @formbricks/storage lint`
- `pnpm exec prettier --check packages/storage/src/formbricks-script.test.ts`
- Temp compose transform plus `docker compose config`, including verification that `formbricks-migrate` has no Traefik labels

### DIFF
--- a/docker/__tests__/formbricks-script.test.ts
+++ b/docker/__tests__/formbricks-script.test.ts
@@ -5,10 +5,8 @@ import { join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { afterEach, describe, expect, test } from "vitest";
 
-const formbricksScriptPath = fileURLToPath(new URL("../../../docker/formbricks.sh", import.meta.url));
-const dockerComposeTemplatePath = fileURLToPath(
-  new URL("../../../docker/docker-compose.yml", import.meta.url)
-);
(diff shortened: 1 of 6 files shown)


xernobyl (dismissed):
LGTM

pandeymangg (approved):

PR #8365, opened 2026-06-25
fix(surveys): WCAG AA accessibility for the survey package [ENG-1345]

itsjavi:
## What does this PR do?

Brings the survey rendering package up to WCAG 2.1 / 2.2 **AA** by resolving the eight accessibility violations tracked under the "WCAG for survey package" epic. Each subticket is its own atomic commit on this branch.

Fixes ENG-1346, ENG-1347, ENG-1348, ENG-1349, ENG-1350, ENG-1351, ENG-1352, ENG-1353
Part of ENG-1345

| Area | What changed |
| --- | --- |
| ENG-1349 Visible focus | One global black + bright-yellow double `:focus-visible` ring scoped to `#fbjs`, theme-agnostic with a forced-colors fallback. Lives in the `components` layer with `!important` so it uniformly overrides the per-component Tailwind `focus-visible:ring-*` utilities (for important declarations an earlier layer wins). |
| ENG-1347 Colour contrast | The "Required" label and rating scale legends defaulted to a brand-tinted color at 0.6 opacity (failing 4.5:1). They now default to a fixed muted neutral (`#64748b`) at full opacity, in both the styling defaults and the CSS token. |
| ENG-1353 Landmarks & headings | The survey body is wrapped in `<main>`, the logo in `<header>`, and the legal nav stays a `<footer>` outside `<main>`, so they resolve as banner / main / contentinfo. Welcome, Ending and Cal card titles are now real headings with a paragraph description instead of two `<label>` elements that pointed at no control. |
| ENG-1346 Keyboard order | "Powered by Formbricks" and the legal footer links no longer carry `tabIndex={-1}`, so keyboard users reach them in natural reading order now that the footer is a landmark. |
| ENG-1351 Aria-label | The auto-close wrapper was a generic `<div>` with an `aria-label` plus click/key handlers (invalid on a generic role, not announced). It is now a plain container that detects the first interaction via bubbled event listeners on a ref. Removed the now-unused `common.auto_close_wrapper` key from all 19 locales and the i18n lockfile. |
| ENG-1352 Page language | `<html lang>` was stuck at the app UI locale. The survey now reports its active language via an `onLanguageChange` callback and the link-survey client keeps `document.documentElement` `lang`/`dir` in sync (resolving `default` to the configured tag, e.g. `en-AU`, and flipping RTL languages to `dir="rtl"`). Scoped to link surveys; the embedded JS widget passes no callback, so host pages are untouched. |
| ENG-1348 Select input | Single-select, multi-select and matrix options were Radix buttons (`role=radio/checkbox`) nested in a `<label htmlFor>` pointing at a button, with `aria-required` on the button and the title labelling nothing. Rebuilt on native `<input type=radio|checkbox>`, each with a unique `id` and a single `<label for>`, grouped in a `<fieldset>` whose `<legend>` carries the headline. Single-select uses `role="radiogroup"` so `aria-required` is valid; matrix cells are native radios named per row with `aria-labelledby` to row + column headers; the element `<form>` gets an `aria-label`. Dropdown, Other, None, deselect, disabled, error and `dir` behavior preserved. |
| ENG-1350 Widget role | Rating items were a `<label tabindex="0">` (a tabbable element with no valid widget role). The native radio `<input>` is now the single focusable widget (arrow/space navigation via the shared radio-group name), wrapped in a `role="radiogroup"` fieldset/legend. Ranking already used native `<button>` widgets, so it needed no change. |

Net change: 49 files. No new committed Playwright specs (see testing notes).

### Post-review refinements

Three follow-ups applied after the first review pass (commit `714364c`):

- **Select "Other" input (ENG-1348):** the free-text field now renders outside the option `<label>` (the bordered box is a plain container; the label wraps only the radio/checkbox row). A label may own a single labelable control, and a click in the label area around the field could previously forward to the radio/checkbox and clear the "Other" selection.
- **Rating focus ring (ENG-1349 / ENG-1350), UX-affecting:** for star and smiley scales the focusable label is now sized to its icon (inside a `flex-1` wrapper), so the focus ring hugs the option as in the design instead of spanning the whole column. Side effect: the click target for star/smiley options is the visible icon (up to 74px) rather than the full column cell. The number scale is unchanged, and the icons themselves look identical (they were already capped at 74px).
- **Fieldset sizing (ENG-1353):** the preflight resets `fieldset { min-width: 0 }` so grouped questions, now wrapped in a `<fieldset>`, can shrink and the matrix horizontal scroll keeps working on narrow viewports.

## How should this be tested?

Automated gate that already passes locally (and mirrors CI):

- `pnpm --filter @formbricks/survey-ui typecheck && pnpm --filter @formbricks/survey-ui test && pnpm --filter @formbricks/survey-ui lint` (60 tests, 0 lint errors)
- `pnpm --filter @formbricks/surveys typecheck && pnpm --filter @formbricks/surveys test && pnpm --filter @formbricks/surveys lint` (603 tests)
- `pnpm --filter @formbricks/web typecheck` and the styling + survey-link unit tests (142 tests)
- `pnpm --filter @formbricks/survey-ui build && pnpm --filter @formbricks/surveys build` (the surveys build regenerates `apps/web/public/js/surveys.umd.cjs`, which is what link surveys load)

Manual / e2e accessibility pass (run on the side, against a published kitchen-sink survey that includes single-select, multi-select, rating, ranking, matrix, plus Welcome and Ending cards, with a second language such as Hebrew enabled):

- axe-core walk with `@axe-core/playwright` across desktop / mobile / tablet, dark mode and forced-colors: expect 0 violations for the rules in scope.
- Keyboard: Tab through a card and confirm the order is lang switch -> options -> Next -> Back -> branding -> footer, and every control shows the focus ring in light and dark themes.
- Screen reader / a11y tree: native radio/checkbox inputs, `fieldset`/`legend` grouping, `main`/`header`/`footer` landmarks, real headings on cards, and no `aria-label` on a generic `<div>`.
- Switch the survey to Hebrew and confirm `document.documentElement` shows `lang="he"` and `dir="rtl"`, and that an embedded survey leaves the host page's `<html>` untouched.

### Required

- [x] Ran `pnpm build` (built `@formbricks/survey-ui` and `@formbricks/surveys`, including the UMD bundle)
- [x] Checked for warnings, there are none (0 lint errors; only pre-existing warnings in untouched files)
- [x] Merged the latest changes from main onto my branch with `git pull origin main` (branch is 2 commits behind `main`; the new commits touch disjoint files, so it is conflict-free, will rebase before merge)
- [ ] My changes don't cause any responsiveness issues (changes are semantic / ARIA plus a focus ring; responsive spot-check is part of the manual pass above)
- [ ] First PR at Formbricks? (returning contributor)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots (to be added with the manual accessibility pass)
- [ ] Updated the Formbricks Docs if changes were necessary (no doc changes needed)


apps/web/lib/styling/constants.ts
```diff
@@ -43,7 +47,7 @@ export const getSuggestedColors = (brandColor: string = DEFAULT_BRAND_COLOR) =>
     // Headlines & Descriptions
     "elementHeadlineColor.light": questionColor,
     "elementDescriptionColor.light": questionColor,
-    "elementUpperLabelColor.light": questionColor,
+    "elementUpperLabelColor.light": mutedLabelColor,
```
Dhruwang: Screenshot 2026-06-29 at 4 30 45 PM


Any specific reason for making this a constant because now this labels are not configurable.

packages/survey-ui/src/components/elements/rating.tsx
Dhruwang, line 1: https://github.com/user-attachments/assets/fed51716-6735-4618-9f38-84e09bf9dbc0

After applying all the changes in this PR, I've encountered an issue where I can no longer use the Tab key to navigate between the different options in the rating question. As you can see, when I press Tab for the first time, the first rating option is selected, but on the next Tab press, the focus shifts to the Next button instead.

packages/survey-ui/src/components/elements/multi-select.tsx
Dhruwang, line 1: https://github.com/user-attachments/assets/c0c37cbd-43b2-44cc-9422-4af2c31bf02b

Is this the expected UI? I mean the focus UI.

packages/survey-ui/src/components/elements/matrix.tsx
Dhruwang, line 1: Screenshot 2026-06-29 at 4 51 00 PM


The labels are no longer colored

packages/survey-ui/src/components/general/element-header.tsx
```diff
@@ -97,6 +67,22 @@ function ElementHeader({
...
         </Label>
       ) : null}
+    </>
+  );
+
+  if (isLegend) {
+    // The public props are div-shaped; drop the div-typed ref when rendering as a <legend>.
+    const { ref: _ref, ...legendProps } = props;
+    return (
+      <legend className={cn("w-full space-y-2", className)} {...legendProps}>
```
Dhruwang: ` <legend>` is a caption slot for text only; the PR stuffed boxes + an image into it, which is invalid HTML → unpredictable layout and a messy screen-reader group name. Fix = put only the question text in the legend, keep images/boxes outside it.

Dhruwang (changes requested):
Hey Javi, thanks for the PR 🚀 

Mostly looks good ! Just left a few comments.

Can you please fix the sonar issues ?

Also, the aria labels are currently using headline text directly, which is in HTML format. This is unrelated to this specific PR, but since we're addressing accessibility issues here, it might be worth considering fixing it as well.

Screenshot 2026-06-29 at 5 07 11 PM

apps/web/lib/styling/constants.ts
itsjavi: Good catch — reverted in ef10ed2b1. `elementUpperLabelColor` is brand-derived again (`questionColor`), so it stays configurable and cohesive with the rest of the palette. The real WCAG 1.4.3 failure was the old `0.6` opacity diluting the color into the card, not the hue itself — so I kept opacity at `1` (the actual fix) and dropped the hardcoded muted constant. The "Required" and scale labels are colored again.

packages/survey-ui/src/components/elements/matrix.tsx
itsjavi: Same root cause as the `constants.ts` comment: the matrix row/column labels resolve `--fb-label-color`, which `styles.ts` maps from `elementUpperLabelColor` — and that had been pinned to the muted neutral. Restored to the brand-derived color in ef10ed2b1, so they are colored again.

packages/survey-ui/src/components/general/element-header.tsx
itsjavi: Agreed — fixed in ef10ed2b1. I dropped the `<legend>` entirely for grouped questions (rating / matrix / single- / multi-select). The headline now stays a normal block (`ElementHeader` renders a `<div>` as before), and the surrounding `<fieldset>` takes its accessible name from it via `aria-labelledby={`${inputId}-headline`}`. So no media/boxes are nested in a `<legend>`, the HTML is valid, and the group name is just the question text.

packages/survey-ui/src/components/elements/rating.tsx
itsjavi: This is the intended radio-group keyboard model (WAI-ARIA): Tab moves focus *into* the group — onto the selected option, or the first one when nothing is selected yet — and the next Tab leaves the group. You move between the options with the **Arrow keys** (←/→), which also select. That is the standard, expected pattern for a single-choice scale and what screen-reader users navigate by; the old behavior (every option its own tab stop) is the anti-pattern this epic was fixing. The options share a radio-group `name`, so Arrow-key navigation across 1–5 works here — could you confirm the arrows move the selection on your end?

packages/survey-ui/src/components/elements/multi-select.tsx
itsjavi: Yes, that is the intended focus indicator. For WCAG 2.4.7 the survey uses one uniform, high-visibility focus ring for every focusable control — a black + bright-yellow double ring — so it stays visible on any background and in dark mode without per-component overrides. On a dropdown option it hugs the full row, which is why it looks heavier there. It is functionally the deliberate single focus style; if the design team wants it toned down specifically for menu items we can scope a variant, but I left it consistent for now.

Dhruwang (dismissed):
Thanks for the PR looks good 🚀

Dhruwang (approved):

PR #8357, opened 2026-06-25
chore: align runtime locale bundles to canonical BCP-47 + render back-compat (ENG-1067)

pandeymangg:
## What does this PR do?

Relates to **ENG-1067** (language-tag standardization) — the **rendering back-compat** step, building on the foundation `canonical.ts` already in this epic. Makes the survey runtime render correctly for **both legacy and canonical** language codes, and aligns the runtime translation bundles with the canonical BCP-47 source (the runtime was the last bare-ISO holdout). This is a **prerequisite for the data migration**.

Changes:

1. **Runtime bundles renamed to canonical** — `de.json → de-DE.json`, `pt.json → pt-BR.json`, … `zh-Hans.json → zh-Hans-CN.json` (all 19, each `= normalizeLanguageCode(bare)`, via `git mv`). The runtime is no longer the bare-ISO outlier; bundle names now match the catalog/DB/v3.

2. **`i18n.config.ts`** — canonical imports / `resources` keys / `supportedLngs`, plus a **script-preserving fallback** (`fallbackLng` function): a non-default or legacy variant resolves to its language's default bundle (`de-AT`/`de` → `de-DE`, `ar-SA` → `ar-EG`, `pt-PT` → `pt-BR`), then English. **Script is preserved** — `zh-Hant-TW` resolves to a Traditional bundle (none shipped yet) → English, **never** the Simplified `zh-Hans-CN`.

3. **Renderer `getI18nLanguage`** — resolves the requested code to the survey's **stored** code by canonical equivalence, so it works for legacy↔canonical, a stale SDK cache, and bare→region. The returned stored code lines up with the survey's i18n **content** keys; i18next maps it to a bundle via the fallback above.

4. **Link `getLanguageCode`** — matches `?lang=` against the survey's languages by canonical equivalence, so an old shared link (`?lang=pt`) keeps resolving to a migrated survey (`pt-BR`), and a canonical link resolves a not-yet-migrated survey.

**Why:** a survey configured with a canonical code (`de-DE`, from the new picker) had no matching runtime bundle → the renderer's UI strings fell back to English. Now `de-DE` → `de-DE` bundle (German UI) and a legacy `de` → `de-DE` via fallback. Survey **content** (author translations) is unaffected — only the renderer's built-in UI strings come from the bundle.

## How should this be tested?

**Automated**
```bash
cd packages/surveys && pnpm vitest run
```

**Manual (link survey — no SDK needed)**
1. Create a multi-language survey (English + German) and publish it.
[...]

PR #8367, opened 2026-06-26
chore: language codes data migration

pandeymangg:
## What does this PR do?

Part of **ENG-1067** (standardize survey, app, and translation language tags on BCP-47).

This is the **data migration** that canonicalizes the language codes already stored in the database to their canonical BCP-47 form (e.g. `de` → `de-DE`, `pt` → `pt-BR`, `zh-Hans` → `zh-Hans-CN`), using the shared `normalizeLanguageCode` / `LANGUAGE_CANONICAL_MAP` foundation from #8349. It runs through the existing data-migration framework (one transaction, tracked in `DataMigration`, run-once + idempotent).

### What it migrates (in order, all in one transaction)

1. **`Language.code` rows** — relabel each code to canonical. When a bare + region row collide on the same canonical within a workspace (`@@unique([workspaceId, code])`), one row survives (preferring a row already at the canonical code, else the oldest) and the rest are absorbed: `SurveyLanguage` links are repointed/deduped (composite PK `@@id([languageId, surveyId])`), `default`/`enabled` flags carried over, aliases preserved, absorbed rows deleted.
2. **Survey content i18n keys** (multi-language surveys only) — recursively rewrite every `i18nString`'s non-`default` language keys across `welcomeCard`, `blocks`, `endings`, `metadata`, `surveyClosedMessage`, `questions`. Keys that collapse to the same canonical (e.g. `de` + `de-DE`) are merged (non-empty / canonical-key value wins).
3. **`Response.language`** — remap distinct codes (skips `NULL`, the `"default"` sentinel, and already-canonical values).
4. **`Response.contactAttributes.language`** (snapshot JSON) — same remap via `jsonb_set`, so the snapshot stays consistent with `Response.language` and the contact attribute.
5. **Contact `language` attribute** (`ContactAttribute` rows whose key is `language`) — index-friendly batched remap (resolves the `language` attribute keys up front, then `attributeKeyId = ANY(...) AND value = ...` to hit the `[attributeKeyId, value]` index).

### Design notes

- **Idempotent & safe to re-run** — already-canonical values are skipped; unparseable/junk codes are left untouched (never dropped) and logged at the end.
- **Atomic** — the whole `run` executes in one transaction; any error rolls back with no partial state.
- **`"default"` i18n sentinel preserved** everywhere.
- **No migration-specific backup step** — disaster recovery relies on the existing DB backup policy.
- The merge planner was dry-run against a production `Language` dump: **8 collision groups** resolve as expected with **0 alias-loss** cases.
- Pure logic (canonicalization, recursive key rewrite, merge planning) lives in `utils.ts` and is unit-tested; `migration.ts` only orchestrates the SQL.

### Dependencies / sequencing

- Builds on the canonical foundation from #8349.
- Code is independent, but the migration should be **run** after the picker/v3/runtime PRs (#8352, #8355, #8357) are in the epic, so that post-migration reads and writes are all canonical (legacy inbound is still accepted via the back-compat in those PRs).

## How should this be tested?

- **Unit tests:** `pnpm --filter @formbricks/database exec vitest run migration/20260625104904_canonicalize_language_codes/utils.test.ts` — covers `toCanonical`, the recursive i18n key rewrite (incl. the `de` + `de-DE` merge fixture), language-row merge planning (collision survivor selection, alias copy, per-workspace scoping), and `SurveyLanguage` repoint/dedupe.
- **Build:** `pnpm --filter @formbricks/database build` — confirms the migration bundles correctly with `@formbricks/i18n-utils` inlined (so it resolves at runtime in both `tsx` and built `node`).
- **Dry-run:** run the merge planner against a prod `Language` dump and assert the expected merges resolve with 0 orphans.

> [!NOTE]
> Step 5 (contact `language` attribute) is the heaviest part on instances with large `ContactAttribute` tables. It is index-friendly and logs per-code progress, but because the framework wraps the whole migration in a single transaction, on very large tables this is a long-running transaction — worth scheduling for an off-peak window. Splitting the high-volume backfill out into a batched out-of-transaction script is a possible follow-up if needed.

### Required

- [x] Ran `pnpm build`
- [ ] Merged the latest changes from the epic onto my branch

### Appreciated


packages/database/migration/20260625104904_canonicalize_language_codes/migration.ts
```diff
@@ -0,0 +1,340 @@
...
+      }
+
+      if (setClauses.length > 0) {
+        surveyUpdates.push({ id: survey.id, setClauses, params });
+      }
+    }
+
+    for (let i = 0; i < surveyUpdates.length; i += SURVEY_BATCH_SIZE) {
+      const batch = surveyUpdates.slice(i, i + SURVEY_BATCH_SIZE);
+      await Promise.all(
```
xernobyl: **Step 2 issues writes concurrently on the interactive-transaction client — make it sequential.**

`tx` here is a Prisma interactive-transaction client (a single connection, via the pg adapter in `migration-runner`). Prisma doesn't support running queries concurrently on one `tx`: depending on the adapter it either serializes them — so `Promise.all` buys nothing over a single connection — or throws and aborts the migration mid-run.

The other four steps all use a sequential `for…await`; this is the only one using `Promise.all`. CI won't surface it either: a test DB rarely has ≥2 multi-language surveys, so `batch` is ≤1 and the concurrency path never runs — but it will on real data.

Suggest matching the other steps:

```ts
for (const update of batch) {
  await tx.$executeRawUnsafe(
    `UPDATE "Survey" SET ${update.setClauses.join(", ")} WHERE id = $${(update.params.length + 1).toString()}`,
    ...update.params,
    update.id
  );
}
```

There's no parallelism to lose on a single connection anyway, and it keeps this step consistent with the other four.

xernobyl (approved):
LGTM

PR #8368, opened 2026-06-26
fix: harden server-side validation on survey response submission

xernobyl:
## What does this PR do?

Security report [ENG-1083](https://linear.app/formbricks/issue/ENG-1083) (via security@) flagged that the survey response submission API relies on client-side validation. Most of it no longer reproduces — the response path now runs real server-side validation — but two gaps were genuine, and this hardens them at the right layers. (Full per-claim breakdown is on the Linear ticket.)

**ttc bounds** — `ttc` (per-question time-to-complete) was an unbounded `z.number()`, so negative or absurdly large values went straight into storage and the summed `_total`, skewing analytics. It's now clamped per-element into `[0, 24h]` at the schema boundary, on the untrusted input schemas only. I went with clamping rather than rejecting on purpose: ttc is noisy telemetry, and I'd rather sanitize a bad timing number than drop an otherwise-valid response over it. The stored/returned schema stays unbounded so historical data still parses, and the authenticated management-create endpoint keeps its own unbounded schema (trusted imports shouldn't be silently clamped).

**Choice membership** — single/multiple-choice questions *without* an "Other" option accepted any arbitrary value (e.g. submitting a made-up value for a Yes/No question). Validation only length-checked unknown values as potential "Other" text, never as membership. Now, for a choice element with no "Other" option, the value must match a configured choice (by id or localized label) or it's rejected. Elements that *do* expose an "Other" option still accept free text, so legacy "Other"-without-sentinel submissions are unaffected. The check lives in the shared validation evaluator, so the management response endpoints get it too (defense in depth).

**Deliberately not fixed (claims 1/2/5 — required-on-finish):** the report also asked the server to reject `finished: true` when required questions are unanswered. I'm leaving that out on purpose — a response legitimately finishes incomplete via quota early-termination, branching/jump-to-ending logic, and redirect endings, so a server gate would false-reject real completed responses (i.e. data loss, which is worse than the low-severity integrity gain). Completeness is enforced client-side; server-side, "finished" is driven by quotas/logic/endings (`endingId` records why). Rationale + evidence are on the ticket.

**Scope note:** membership covers MultipleChoice single/multi (the reported case). PictureSelection is the same class but has a different choice shape (id + imageUrl, no label) — left as a possible follow-up.

## How should this be tested?

- **Choice membership** — `pnpm --filter @formbricks/surveys test src/lib/validation/evaluator.test.ts`: a no-"Other" choice rejects an unconfigured value (`invalidOption`), accepts a configured value by id or label, an "Other"-enabled element still accepts free text, and a malformed element with no `choices` doesn't throw.
- **ttc** — `cd apps/web && pnpm test "responses/types/response.test" validated-response-update-input`: out-of-range ttc is clamped into `[0, 24h]` on both the POST (v2) and PUT (v1) client boundaries, in-range values are untouched.
- **Manual** — against `POST /api/v2/client/{workspaceId}/responses`:
  - `data: { <singleChoiceQuestionId>: "NOT_AN_OPTION" }` for a Yes/No question → `400` "Validation failed".
  - `ttc: { q1: -5, q2: 9000000000000 }` → accepted, stored as `{ q1: 0, q2: 86400000 }`.

### Required

- [ ] Ran `pnpm build` (package builds + targeted vitest + eslint run locally; full app build left to CI)
- [x] Merged the latest changes from main onto my branch (rebased onto `main`)
- [ ] First PR at Formbricks?

### Appreciated

### DIFF
--- a/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/validated-response-update-input.test.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/validated-response-update-input.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test } from "vitest";
+import { MAX_RESPONSE_TTC } from "@formbricks/types/responses";
 import { getValidatedResponseUpdateInput } from "./validated-response-update-input";
 
 describe("getValidatedResponseUpdateInput", () => {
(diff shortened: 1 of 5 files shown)


Dhruwang (approved):
Thank you for the changes, looks good.

PR #8371, opened 2026-06-26
feat(workflows): enqueue workflow runs on completed responses [ENG-1226]

xernobyl:
## What does this PR do?

Wires the **producer half** of the workflows Runner into the response pipeline, so completing a survey response actually starts the configured workflows. The foundation already exists on `epic/workflows-v1` (DB models, the `response.completed` trigger, enable→`WorkflowVersion` snapshots, the runs table UI) but nothing connected a completed response to its workflows — this is that connection.

Ticket: https://linear.app/formbricks/issue/ENG-1226

> Stacked on `epic/workflows-v1` (not `main`).

Three layers, one per commit:

1. **`packages/jobs` — the `workflow-run.process` background job.** Mirrors the existing `response-pipeline.process` wiring: `JOB_NAMES.workflowRun`, a Zod-validated `ZWorkflowRunJobData`, an `enqueueWorkflowRunJob` producer, and a default handler that is a **throwing stub** (logs + throws "override not registered"). Real execution is registered by ENG-1228 as an apps/web override; until then the stub failing loudly is the intended behavior.
2. **`apps/web/modules/workflows/lib/runner/` (new) — match, dispatch, orchestrate.**
   - `match-workflows.ts` — pure, no I/O: given a completed response (`surveyId` + `endingId`) and candidate published versions, returns which workflows fire. Trigger-type guard + `surveyId` match + ending-card filter (empty = any). Exhaustively table-tested.
   - `dispatch.ts` — a backend-agnostic `DispatchWorkflowRun` **port** plus the one BullMQ adapter (`dispatchWorkflowRunViaJobs`). This is the *only* file that imports `@formbricks/jobs`.
   - `enqueue-response-completed-runs.ts` — loads enabled workflows for the workspace, matches by **published** `WorkflowVersion` (not the draft), and creates a `queued` `WorkflowRun` bound to that version, then dispatches.
3. **Response pipeline hook** (`process-response-pipeline-job.ts`) — on `responseFinished` + an ending card reached, calls the orchestrator alongside the other side-effects.

**Design decisions worth a look:**
- **Backend-agnostic by isolation, not abstraction.** `WorkflowRun` is the durable source of truth; the queue is just a hand-off. Swapping BullMQ for something else (River/SQS/DB-poller) means replacing the adapter + `packages/jobs` wiring — the matcher, run-creation, and pipeline hook don't change.
- **Idempotent.** The run is created with `idempotencyKey = responseId` (`@@unique([workflowId, idempotencyKey])`); on P2002 we look up the existing run and re-dispatch. The job uses a deterministic `jobId = runId`, so a pipeline replay produces **0 new runs and 0 duplicate jobs**, and a "created-but-not-dispatched" run gets re-dispatched.
- **Failure-isolated.** Per-workflow try/catch (one bad workflow doesn't skip the rest) inside an outer try/catch (the runner never fails the response pipeline job or its other side-effects).
- **`attempts: 1`** for the job (not the default 3) — run-level retry is row-modelled and belongs to ENG-1228, not BullMQ.

**Out of scope (explicitly):** run *execution* — worker, actions, emails, run logs. That's **ENG-1228**, which this PR blocks. Also out of scope: dry-runs, triggers other than `response.completed`, and a reconciler for orphaned `queued` runs (the `status`/`nextAttemptAt` indexes already exist for a future sweeper).

## How should this be tested?

Automated (what I ran, all green):

```bash
# rebuild deps so apps/web picks up the new job types
pnpm --filter @formbricks/database build   # prisma generate
pnpm --filter @formbricks/jobs build
pnpm --filter @formbricks/workflows build

pnpm --filter @formbricks/jobs run test typecheck lint
# the ~17 new apps/web tests:
cd apps/web && npx vitest run modules/workflows/lib/runner modules/response-pipeline/lib/process-response-pipeline-job.test.ts
```

Manual local smoke (no worker yet):
1. Create a survey, attach a workflow, and **enable** it (this snapshots a published `WorkflowVersion`).
2. Complete a response on that survey through to a thank-you/ending card.
3. Expect a `queued` `WorkflowRun` with `workflowVersionId` set and a `workflow-run.process` job enqueued.
4. Replay the response-pipeline job → **no duplicate run, no duplicate job** (idempotency).
5. The job itself will throw "override not registered" — expected until ENG-1228 registers the executor.

Edge behavior covered by tests: not-finished / no-ending-card → 0 runs; ending-card filter (empty = any, specific list hits/misses); fires from the **published** version even after the draft is edited; enabled-but-no-published-version → log + skip; draft/disabled/archived/other-workspace/non-matching-survey → never fire; fan-out (N matches → N runs/jobs); malformed published definition → per-workflow skip, others continue.

### Required

- [x] Ran `pnpm build` (affected packages: database, jobs, workflows, web typecheck)
- [ ] Merged the latest changes from main onto my branch — N/A, this is stacked on `epic/workflows-v1`
- [x] My changes don't cause any responsiveness issues (backend-only, no UI)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots — N/A, backend-only
- [ ] Updated the Formbricks Docs — docs are tracked separately in the M5 milestone (ENG-1232)

### DIFF
--- a/apps/web/modules/workflows/lib/runner/enqueue-response-completed-runs.ts
+++ b/apps/web/modules/workflows/lib/runner/enqueue-response-completed-runs.ts
@@ -1,0 +1,134 @@
+import { prisma } from "@formbricks/database";
+import { Prisma } from "@formbricks/database/prisma";
+import { PrismaErrorType } from "@formbricks/database/types/error";
+import { logger } from "@formbricks/logger";
+import { ZWorkflowTriggerRunPayload } from "@formbricks/workflows";
+import { type DispatchWorkflowRun } from "./dispatch";
+import { type WorkflowMatchCandidate, matchWorkflowsForResponse } from "./match-workflows";
+
+/** The subset of a completed response the runner needs. Structurally satisfied by the response-pipeline payload. */
+interface RunnerResponse {
+  id: string;
+  surveyId: string;
+  finished: boolean;
+  endingId?: string | null;
+  updatedAt: Date;
+  data: Record<string, unknown>;
+}
+
+interface EnqueueResponseCompletedWorkflowRunsInput {
+  response: RunnerResponse;
+  workspaceId: string;
+  dispatch: DispatchWorkflowRun;
+  logContext?: Record<string, unknown>;
+}
+
+const isUniqueConstraintError = (error: unknown): boolean =>
+  error instanceof Prisma.PrismaClientKnownRequestError &&
+  error.code === PrismaErrorType.UniqueConstraintViolation;
+
+/**
+ * Producer half of the workflow runner. On a completed response that reached an ending card, find the
+ * enabled workflows whose current published version targets this survey/ending, persist one `queued`
+ * `WorkflowRun` per match (bound to that published version), and hand each to the injected dispatcher.
+ *
+ * Idempotent: `idempotencyKey = responseId` + `@@unique([workflowId, idempotencyKey])` plus a
+ * deterministic dispatch `jobId` mean a replayed `responseFinished` creates no duplicate runs or jobs.
+ * Each match is isolated so one workflow's failure never blocks the others. The caller wraps the whole
+ * call so a runner failure never affects the response pipeline.
+ */
+export const enqueueResponseCompletedWorkflowRuns = async ({
+  response,
+  workspaceId,
+  dispatch,
+  logContext,
+}: EnqueueResponseCompletedWorkflowRunsInput): Promise<void> => {
+  // Only completed responses that reached an ending card enqueue runs.
+  if (!response.finished || !response.endingId) {
+    return;
+  }
+  const endingId = response.endingId;
+
+  const workflows = await prisma.workflow.findMany({
+    where: { workspaceId, status: "enabled" },
+    select: {
+      id: true,
+      versions: { orderBy: { version: "desc" }, take: 1, select: { id: true, definition: true } },
+    },
+  });
+
+  const candidates: WorkflowMatchCandidate[] = [];
+  for (const workflow of workflows) {
+    const publishedVersion = workflow.versions[0];
+    if (!publishedVersion) {
+      // Enable always publishes a version, so this is a data-integrity guard: never enqueue an
+      // unrunnable run for an enabled workflow that somehow has no published version.
+      logger.warn(
+        { ...logContext, workflowId: workflow.id, workspaceId },
+        "Enabled workflow has no published version; skipping workflow run enqueue"
+      );
+      continue;
+    }
+    candidates.push({
+      workflowId: workflow.id,
+      publishedVersionId: publishedVersion.id,
+      definition: publishedVersion.definition,
+    });
+  }
+
+  const matches = matchWorkflowsForResponse(candidates, { surveyId: response.surveyId, endingId });
+  if (matches.length === 0) {
+    return;
+  }
+
+  // One trigger payload for every matched workflow. `triggeredAt` is derived from `response.updatedAt`
+  // (not wall-clock) so a pipeline retry rebuilds a byte-identical payload.
+  const triggerPayload = ZWorkflowTriggerRunPayload.parse({
+    type: "response.completed",
+    workspaceId,
+    surveyId: response.surveyId,
+    responseId: response.id,
+    endingCardId: endingId,
+    data: response.data,
+    triggeredAt: response.updatedAt.toISOString(),
+  });
+
+  for (const match of matches) {
+    try {
+      const run = await prisma.workflowRun.create({
+        data: {
+          workflowId: match.workflowId,
+          workspaceId,
+          workflowVersionId: match.publishedVersionId,
+          status: "queued",
+          triggerType: "response.completed",
+          surveyId: response.surveyId,
+          responseId: response.id,
+          isDryRun: false,
+          attempt: 0,
+          idempotencyKey: response.id,
+          triggerPayload,
+        },
+        select: { id: true },
+      });
+      await dispatch({ workflowRunId: run.id, workflowId: match.workflowId, workspaceId });
+    } catch (error) {
+      if (isUniqueConstraintError(error)) {
+        // The run already exists from an earlier pipeline pass. Re-dispatch (idempotent via jobId) in
+        // case that pass created the run but failed to enqueue — no duplicate run, no duplicate job.
+        const existing = await prisma.workflowRun.findUnique({
+          where: { workflowId_idempotencyKey: { workflowId: match.workflowId, idempotencyKey: response.id } },
+          select: { id: true },
+        });
+        if (existing) {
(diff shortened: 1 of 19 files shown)


xernobyl:
Thanks for the review @coderabbitai — all four addressed in 5490b1430:

**1. Pool exhaustion swallowed by the enqueue hook — fixed.** Transient DB pool exhaustion (P2024 / connection-pool timeout) now propagates from both the pipeline hook and the per-workflow enqueue loop, so the job retries instead of silently dropping the completed-response trigger — same contract as the outer pipeline catch. Pulled the predicate into a shared `@/lib/jobs/pool-exhaustion` util so the runner can reuse it without importing the pipeline.

**2. Test locked in silent success on a non-recoverable P2002 — fixed.** A unique violation with no findable existing run is a contradictory state, so it's now surfaced via `logger.error` and the test asserts it.

**3. Dispatch failure can orphan a queued run — by design for this scope, transient case now covered.** The `WorkflowRun` row is the durable record (transactional-outbox), so transient failures now retry (see #1). The residual non-transient dispatch-orphan is meant to be reconciled by a sweeper over `queued` runs — the `status` / `nextAttemptAt` indexes already exist for it — which is explicitly out of scope for this ticket (ENG-1226) and tracked separately. We deliberately don't rethrow dispatch errors here, since that would abort sibling workflows and re-run the earlier `responseFinished` side-effects (webhooks/follow-ups) on retry.

**4. Don't register the throwing processor — keeping it, intentionally.** The registration is required for the producer to enqueue and the worker to resolve a handler; the throwing default is the intended fail-loud for "no executor yet" (silently dropping a queued run would be worse). This is a pre-prod epic branch — production rollout is gated by release sequencing (ships alongside the executor in ENG-1228) / a feature flag, not by removing the registration.

@coderabbitai resolve

apps/web/modules/workflows/lib/runner/match-workflows.ts
```diff
@@ -0,0 +1,42 @@
...
+ * the trigger targets the response's survey and its ending filter passes (empty `endingCardIds` means
+ * "any ending", otherwise the reached `endingId` must be listed). No I/O, so it's exhaustively
+ * table-testable. Matching reads the published snapshot, never the mutable draft.
+ */
+export const matchWorkflowsForResponse = (
+  candidates: WorkflowMatchCandidate[],
+  { surveyId, endingId }: ResponseMatchInput
+): WorkflowMatch[] =>
+  candidates
+    .filter(({ definition }) => {
```
itsjavi, line 61: Matcher isn't failure-safe when a definition is malformed:

`matchWorkflowsForResponse` reads `definition.trigger.triggerType` and
`trigger.config.*` unguarded, across all candidates in one `.filter()`.
`definition` is unvalidated DB JSON (only a Prisma `///` brand, never
parsed on read), so one malformed published version throws a TypeError.

Why that's an issue: the matcher runs *before* the per-workflow try/catch in
`createAndDispatchWorkflowRun`, so the throw bubbles to the pipeline hook,
is swallowed as non-pool-exhaustion, and drops *every* workflow for that
response — breaking the "one bad workflow never blocks the rest" guarantee.

Fix (either):
  - a -  guard each candidate — skip + warn when trigger/config is missing
  - b - Zod-parse the published definition per candidate in
      `loadEnabledWorkflowCandidates`, skip the ones that fail

pls also add the malformed-definition test the PR description claims
exists — `match-workflows.test.ts` currently only covers well-typed defs.

apps/web/modules/workflows/lib/runner/enqueue-response-completed-runs.ts
```diff
@@ -0,0 +1,193 @@
...
+        surveyId: response.surveyId,
+        responseId: response.id,
+        isDryRun: false,
+        attempt: 0,
+        idempotencyKey: response.id,
+        triggerPayload,
+      },
+      select: { id: true },
+    });
+    await dispatch({ workflowRunId: run.id, workflowId: match.workflowId, workspaceId });
```
itsjavi: A non-transient dispatch failure orphans a `queued` run.

`create()` and `dispatch()` share one try block. If `dispatch()` fails with
anything that isn't a unique-constraint or pool-exhaustion error (e.g. a Redis
connection error — the realistic dispatch failure), it hits the final
`logger.error` and is swallowed.

The fallout: the `WorkflowRun` row is already persisted as `status: "queued"`
with no backing job. Idempotent-replay recovery only fires if the pipeline job
retries, but a swallowed error never propagates — the pipeline job succeeds and
never retries. The run sits orphaned until the (deferred) reconciler exists.

Fix (either):
-  (a) rethrow dispatch failures after logging — create is idempotent on replay
      via the @@unique key, so a pipeline retry safely re-dispatches the run
-  (b) split create vs dispatch so a dispatch error is classified retryable
      and propagated

At minimum: confirm the orphan-until-reconciler window is an accepted risk for
this epic, not an oversight.

itsjavi (changes requested):
thanks for the PR works well, I only left a couple of small remarks

apps/web/modules/workflows/lib/runner/match-workflows.ts
xernobyl: Addressed in bc3f76259 — went with option (a). `matchWorkflowsForResponse` now reads the trigger through a total `readResponseCompletedTriggerConfig` that treats the branded-but-unvalidated DB JSON as untrusted: it returns `null` instead of throwing, so a malformed published version skips its own workflow rather than aborting matching for every workflow on the response. `loadEnabledWorkflowCandidates` warns + skips published versions with no usable `response.completed` trigger. Added the malformed-definition tests in `match-workflows.test.ts` (no trigger / missing config / malformed-does-not-block-a-valid-one) plus a loader skip test.

apps/web/modules/workflows/lib/runner/enqueue-response-completed-runs.ts
xernobyl: Addressed in bc3f76259 — split persist from dispatch. A dispatch failure *after* the run is persisted now logs a distinct, alertable orphan ("Workflow run persisted but dispatch failed; queued run is orphaned until the reconciler re-dispatches it") with the run id, instead of the generic swallow.

And to confirm directly: yes, the orphan-until-reconciler window is a deliberate, accepted trade-off, not an oversight. I did **not** rethrow-to-retry, because the runner runs after the pipeline's other `responseFinished` side-effects (webhooks/follow-ups/notifications) — failing the job to retry would re-run those and risk duplicate sends. The durable `queued` row + the `status`/`nextAttemptAt` indexes are the outbox; a reconciler that re-dispatches orphaned runs is the proper fix and the right separate follow-up (out of scope for this ticket). Pool exhaustion during create still rethrows (retryable). Net: the window is bounded and now observable.

itsjavi (approved):
thanks, lgtm!

PR #8377, opened 2026-06-29
feat(unify-feedback): display translated feedback record text

Dhruwang:
## What does this PR do?

Displays Hub-translated feedback text in **Unify Feedback**. Wherever a feedback record's text is shown, the translated text (`value_text_translated`) for the configured target language is preferred, with a clean fallback to the original `value_text` when no translation exists yet. The original text stays accessible for transparency.

Linear: **ENG-1253** — https://linear.app/formbricks/issue/ENG-1253/display-translated-feedback-records-in-formbricks

Screenshot 2026-06-29 at 11 39 22 AM

### Changes
- **`resolveFeedbackDisplayText`** helper — translated-preferred, original fallback; guards empty/whitespace translations and translations identical to the original (so a no-op translation shows no affordance).
- **Feedback records table** — renders the translated value with a language badge; the original is revealed on hover/focus, and the badge is keyboard-focusable with an accessible label.
- **Record drawer** — read-only "Translated text" block beside the editable original (translation is Hub-derived, never edited here).
- **Taxonomy preview card** — same translated preference.
- **Type bridge** — augmented the Hub `FeedbackRecordData` read type with the optional, read-only `value_text_translated` / `translation_lang_key` fields. The released `@formbricks/hub` SDK predates these fields; this augmentation is removed once the SDK is regenerated to include them.
- **i18n** — added translation copy under `workspace.unify` across all locales.

### Dependency
Requires the Hub release that includes the translated-text enrichment (ENG-1255) for the fields to be populated. Until that ships, the UI gracefully shows the original text.

## How should this be tested?

- **Unit:** `pnpm --filter web test modules/ee/unify-feedback/lib/utils.test.ts` — covers translated-preferred, empty/whitespace fallback, identical-translation suppression, missing-text, and partial cases.
- **UI:** with a Hub that returns `value_text_translated` for a record, open **Unify Feedback → Feedback records**:
  - Translated rows show the translated text + a language badge; hover/focus the badge to see the original.
  - Open a translated row → the drawer shows a read-only **Translated text** block under the editable original.
  - Records without a translation render the original unchanged (no badge).

### Required

- [ ] Ran `pnpm build`

### Appreciated


xernobyl (approved):
LGTM

apps/web/modules/ee/unify-feedback/components/feedback-records-table.tsx
```diff
@@ -498,8 +506,10 @@ const FeedbackRecordRow = ({
   onSelectChange,
   onClick,
 }: Readonly<FeedbackRecordRowProps>) => {
-  const value = formatValue(record, t, locale);
+  const { text, isTranslated, original, langKey } = resolveFeedbackDisplayText(record);
+  const value = formatValue(record, text, t, locale);
   const isLongValue = value.length > 60;
+  const translatedLangLabel = langKey ? (getLanguageLabel(langKey, locale) ?? langKey) : null;
```
xernobyl, line 512: **[P2] The language badge will render the raw locale (e.g. `en-US`) instead of the language name in production.**

`getLanguageLabel` matches only ISO-639 **2-letter** codes (`iso639Languages.find((l) => l.code === languageCode)` → `en`, `de`), but the Hub stores `translation_lang_key` as canonical **BCP-47** (`en-US`, `de-DE`). So `getLanguageLabel("en-US", …)` returns `undefined` and we fall through to the raw key — here, and at the same call in `feedback-record-form-drawer.tsx` and `topics-subtopics-preview.tsx`.

The unit tests only use `"en"` (a 2-letter code that *does* match), so they stay green while the real data path shows `en-US`.

Suggest passing the primary subtag and keeping the raw fallback:

```ts
const translatedLangLabel = langKey
  ? (getLanguageLabel(langKey.split("-")[0], locale) ?? langKey)
  : null;
```

…and adding an `"en-US"` case to the test so it's locked. (Same one-line change at the other two call sites.)

Dhruwang: Screenshot 2026-06-29 at 3 34 46 PM


Hey Tiago, I believe this won't be an issue for us since the ISO639Languages array also includes language codes like en-US. The array name might cause confusion as it mentions ISO 639, but it does contain non-two-letter codes as well.

PR #7685, opened 2026-04-07
fix: strip @layer properties block to prevent host page CSS pollution

mariusbolik:
## Problem

`surveys.umd.cjs` injects a `<style id="formbricks__css">` tag into `document.head`. That tag contains two sections:

1. **Correctly scoped** — Tailwind preflight reset under `#fbjs *` selectors. ✅
2. **Globally polluting** — A `@layer properties` block with bare `*, :before, :after, ::backdrop` selectors. ❌

```css
/* This part leaks out of #fbjs and resets ALL elements on the host page */
@layer properties {
  @supports ((-webkit-hyphens: none) and (not (margin-trim: inline))) or
            ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
    *, :before, :after, ::backdrop {
      --tw-shadow: 0 0 #0000;
      --tw-ring-shadow: 0 0 #0000;
      --tw-ring-offset-width: 0px;
      --tw-border-style: solid;
      /* ~25 more --tw-* properties... */
    }
  }
}
```

**Why it can't be scoped:** CSS `@layer` at-rules are globally scoped by the CSS specification — they cannot be confined by a surrounding selector. This is a spec constraint, not a browser quirk.

**Effect:** Any host site using Tailwind CSS v4 will have its `--tw-*` custom properties reset to `initial` / zero on every element the moment a Formbricks survey initializes. This breaks `shadow-*`, `ring-*`, `rotate-*`, `translate-*`, and other utilities across the entire page — not just inside the survey widget.

Reported in: formbricks/js#46

## Fix

Add a small PostCSS plugin in `packages/surveys/postcss.config.cjs` that removes any `@layer properties { ... }` block from the compiled CSS output:

```js
const stripLayerProperties = () => ({
  postcssPlugin: "postcss-strip-layer-properties",
  AtRule: {
    layer: (atRule) => {
      if (atRule.params === "properties") atRule.remove();
    },
  },
});
stripLayerProperties.postcss = true;
```

The `@property` declarations already emitted earlier in the same stylesheet serve the same browser-compatibility purpose for all supporting browsers. Removing `@layer properties` does not affect survey rendering in any browser.

## Testing

- Build `packages/surveys` and confirm `@layer properties` is absent from `dist/index.umd.cjs`.
- Load the SDK on a Tailwind v4 page, trigger a survey, and verify that host page shadows/rings/transforms are unaffected.

PR #8378, opened 2026-06-29
fix(survey-ui): scope Tailwind v4 global CSS to #fbjs to stop host-page pollution

Dhruwang:
## TL;DR (plain words)

When a Formbricks survey loads on a site, it adds CSS to style the survey popup. That CSS is supposed to touch **only the survey** — but Tailwind v4 emits a few constructs that are **global by nature** and ignore any prefix:
- a `*` rule that resets a hidden variable on **every** element
- `:root` theme colors/spacing
- `@property` declarations

The widget bundle (`@formbricks/surveys`) already strips/scopes these since #7685 / #7805. **This PR fixes the artifact that was still shipping them — `@formbricks/survey-ui`'s standalone stylesheet (`dist/survey-ui.css`, the `./styles` npm export) — and hardens both packages against the leak ever coming back.**

> **Correction after review (thanks @itsjavi):** an earlier version of this description claimed the served widget bundle was leaking. Verified against fresh builds of `main`: the UMD (`surveys.umd.cjs`) was **already clean** — the `?inline` import of survey-ui styles goes through the surveys package's PostCSS chain, which runs the scoping plugins. The leak lives only in the standalone `dist/survey-ui.css` artifact. The customer-facing half of ENG-1333 was fixed by #7805 (Apr 22); a report reproducing after that date would point at a different vector (most likely a cached pre-#7805 SDK).

## Problem (detail)

Tailwind v4 emits three constructs that are **global by spec** and cannot be confined by Tailwind's `important: "#fbjs"` selector scoping:

- `@layer properties { *, :before, :after, ::backdrop { --tw-*: ... } }` — resets every `--tw-*` custom property on **every element of the host page**
- `@layer theme { :root, :host { --color-*, --spacing, ... } }` — overrides the host's own theme variables
- `@property --tw-* { inherits: false; ... }` — breaks `--tw-*` inheritance on the host and rejects host values via `syntax`

On any host app using Tailwind, these break shadows, rings, transforms, gradients, and theme colors. Reported in ENG-1333 (and previously formbricks/js#46); fixed for the widget bundle in #7685 / #7805.

`@formbricks/survey-ui` was built with only `@tailwindcss/postcss`, so its **standalone stylesheet** (`dist/survey-ui.css`, exported as `@formbricks/survey-ui/styles`) still contained all three. Nothing inside this monorepo serves that file to end users — the widget imports the styles through its own (scoped) build — but any direct npm consumer of the `./styles` export received leaking CSS, and the duplicated plugin logic meant the two builds could silently drift apart.

## Fix

- Extract the three scoping plugins (`stripLayerProperties`, `scopeLayerTheme`, `replaceAtPropertyWithScoped`) into `packages/vite-plugins/postcss-scope-fbjs.cjs`, **shared by both packages** so the logic can't drift again.
- `survey-ui` now compiles Tailwind via **PostCSS** (not `@tailwindcss/vite`) so the scoping plugins run *after* Tailwind emits its CSS.
- Regression test (`packages/survey-ui/src/styles/scope-fbjs.test.ts`) asserting no global `@layer properties` / `:root` / `:host` / `@property --tw-*` survive scoping.
- **New (review follow-up): post-build guard on the real artifact.** The plugins match Tailwind's *current* emission shapes and the unit tests use a frozen fixture, so a Tailwind upgrade could reintroduce the leak with tests green. `packages/survey-ui/scripts/assert-scoped-css.mjs` now runs after every `vite build` and fails the build if `dist/survey-ui.css` contains `@layer properties`, `@property --tw-*`, or any selector reaching `:root`/`:host` without a `#fbjs` scope (loose-matched, so future shapes like `:root:where(...)` are caught too).

## Verification

All measured on **fresh builds** (clean dist, `--force`):

| artifact | main | this branch |
|---|---|---|
| `packages/surveys/dist/index.umd.cjs` (served widget) | ✅ clean — 0 `@layer properties`, 0 `@property --tw-`, 0 bare `:root`/`:host`, 21 `#fbjs`-scoped rules | ✅ clean (unchanged) |
| `packages/survey-ui/dist/survey-ui.css` (npm `./styles` export) | ❌ 1 `@layer properties`, 71 `@property --tw-*`, bare `:root`/`:host` | ✅ **0 / 0 / 0** |

- `survey-ui` test suite: 65/65 pass (incl. 5 new).
- Post-build guard: passes on this branch; negative-tested by appending `@layer properties`, `@property --tw-foo`, `:root:where(.x)` and `:host` to the built CSS — all four flagged, build fails.
- Survey rendering unaffected — all `--tw-*` defaults preserved, scoped to `#fbjs`.

## Reproduce locally

**See the leak in the built file, main vs this branch:**
```bash
# broken (main):
git checkout main
pnpm --filter @formbricks/survey-ui build
grep -c '@layer properties' packages/survey-ui/dist/survey-ui.css   # 1  <- leak
grep -c '@property --tw-'   packages/survey-ui/dist/survey-ui.css   # 71 <- leak

# fixed (this branch):
git checkout dhruwang/eng-1333-scope-survey-ui-css-to-fbjs
pnpm --filter @formbricks/survey-ui build
grep -c '@layer properties' packages/survey-ui/dist/survey-ui.css   # 0
grep -c '@property --tw-'   packages/survey-ui/dist/survey-ui.css   # 0
```

**See what the leak does in a browser** — save as `repro.html`, open, click the button:
```html
<!doctype html><html><head>
<style>
  .box{ --tw-ring-shadow:0 0 0 6px green; box-shadow:var(--tw-ring-shadow);
        width:120px;height:120px;margin:40px }
</style></head><body>
  <div class="box"></div>
  <button onclick="leak()">inject survey-ui's global CSS</button>
  <script>function leak(){
    const s=document.createElement('style');
    s.textContent='@layer properties{*,:before,:after,::backdrop{--tw-ring-shadow:0 0 #0000}}';
    document.head.appendChild(s);   // green ring vanishes -- host element broke
  }</script>
</body></html>
```
Green ring before click → gone after. That one line is exactly the kind of rule `dist/survey-ui.css` shipped to `./styles` consumers. After the fix, that block never reaches the page (stripped), and the survey's own version is scoped to `#fbjs`.

## ENG-1333 status

The customer-facing widget leak was fixed by #7805 (2026-04-22). This PR closes the remaining unscoped artifact (`./styles` npm export) and adds two layers of drift protection (shared plugins + post-build guard). Before closing the ticket, worth confirming the original report predates #7805 — if a customer reproduced it later, the vector is elsewhere (likely a cached pre-#7805 SDK) and needs its own follow-up.

### DIFF
--- a/packages/survey-ui/postcss.config.mjs
+++ b/packages/survey-ui/postcss.config.mjs
@@ -1,5 +1,14 @@
+import tailwindcss from "@tailwindcss/postcss";
+// Same three #fbjs-scoping plugins used by @formbricks/surveys. Tailwind v4
+// emits `@layer properties`, `@layer theme { :root, :host }` and
+// `@property --tw-*` globally (by spec), which leak into and break host-page
+// styles when the survey widget injects this bundle. These plugins confine
+// them to #fbjs. They MUST run after Tailwind has compiled, so the Tailwind
+// PostCSS plugin is processed here (not via @tailwindcss/vite) to guarantee
+// ordering. See packages/vite-plugins/postcss-scope-fbjs.cjs and
+// https://github.com/formbricks/js/issues/46.
+import scopeFbjs from "../vite-plugins/postcss-scope-fbjs.cjs";
+
 export default {
-  plugins: {
-    '@tailwindcss/postcss': {},
-  },
+  plugins: [tailwindcss(), ...scopeFbjs.scopeFbjsPlugins()],
 };
--- a/packages/vite-plugins/postcss-scope-fbjs.cjs
+++ b/packages/vite-plugins/postcss-scope-fbjs.cjs
@@ -108,7 +108,16 @@ const replaceAtPropertyWithScoped = () => {
       for (const [name, value] of collected) {
         rule.append(postcss.decl({ prop: name, value }));
       }
-      root.append(rule);
+      // Emit the fallback inside the low-priority `theme` layer rather than as an
+      // unlayered author rule. `@property` initial values are defaults that any
+      // real declaration overrides; an unlayered rule, by contrast, wins over
+      // every Tailwind `@layer` (theme/base/components/utilities) and would reset
+      // `--tw-*` set by the survey's own utilities. `theme` is the lowest layer
+      // (see `@layer theme, base, components, utilities` in the globals), so
+      // utilities still win and the fallback only applies when nothing sets it.
+      const layer = postcss.atRule({ name: "layer", params: "theme" });
+      layer.append(rule);
+      root.append(layer);
     },
   };
 };
--- a/packages/vite-plugins/postcss-scope-fbjs.cjs
+++ b/packages/vite-plugins/postcss-scope-fbjs.cjs
@@ -1,0 +1,130 @@
+// Shared PostCSS plugins that confine Tailwind v4's globally-scoped emissions
+// to the Formbricks survey container (#fbjs), preventing host-page CSS
+// pollution. Consumed by both @formbricks/surveys and @formbricks/survey-ui so
+// the scoping logic cannot drift between the two CSS bundles that ship together
+// in the injected <style> element (see packages/surveys/src/lib/styles.ts).
+//
+// Tailwind v4 emits three constructs that are global by spec and therefore
+// cannot be confined by Tailwind's `important: "#fbjs"` selector scoping:
+//   1. @layer properties { *, :before, :after, ::backdrop { --tw-*: ... } }
+//   2. @layer theme      { :root, :host { --color-*, --spacing, ... } }
+//   3. @property --tw-*  { inherits: false; syntax: ...; initial-value: ... }
+// Each leaks into the host page and breaks host Tailwind utilities (shadows,
+// rings, transforms, gradients, theme variables).
+//
+// See: https://github.com/formbricks/js/issues/46
+
+// Strips the `@layer properties { ... }` block that Tailwind v4 emits as a
+// browser-compatibility fallback for `@property` declarations.
+//
+// Problem: CSS `@layer` at-rules are globally scoped by spec — they cannot be
+// confined by a surrounding selector. Even though all other Formbricks survey
+// styles are correctly scoped to `#fbjs`, the `@layer properties` block
+// contains a bare `*, :before, :after, ::backdrop` selector that resets all
+// `--tw-*` CSS custom properties on every element of the host page. This
+// breaks shadows, rings, transforms, and other Tailwind utilities on any site
+// that uses Tailwind v4 alongside the Formbricks SDK.
+//
+// The `@property` declarations already present in the same stylesheet cover
+// the same browser-compatibility need for all supporting browsers, so removing
+// `@layer properties` does not affect survey rendering.
+const stripLayerProperties = () => {
+  return {
+    postcssPlugin: "postcss-strip-layer-properties",
+    AtRule: {
+      layer: (atRule) => {
+        if (atRule.params === "properties") {
+          atRule.remove();
+        }
+      },
+    },
+  };
+};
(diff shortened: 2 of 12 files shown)


packages/survey-ui/postcss.config.mjs
```diff
@@ -1,5 +1,14 @@
...
+// Same three #fbjs-scoping plugins used by @formbricks/surveys. Tailwind v4
+// emits `@layer properties`, `@layer theme { :root, :host }` and
+// `@property --tw-*` globally (by spec), which leak into and break host-page
+// styles when the survey widget injects this bundle. These plugins confine
+// them to #fbjs. They MUST run after Tailwind has compiled, so the Tailwind
+// PostCSS plugin is processed here (not via @tailwindcss/vite) to guarantee
+// ordering. See packages/vite-plugins/postcss-scope-fbjs.cjs and
+// https://github.com/formbricks/js/issues/46.
+import scopeFbjs from "../vite-plugins/postcss-scope-fbjs.cjs";
+
```
itsjavi, line 11: I rebuilt surveys.umd.cjs from a clean dist at both the merge-base and current main: it already contains no `@layer properties`, no `@property --tw-` and no bare `:root`/`:host` (the `?inline` import of @formbricks/survey-ui/styles goes through the surveys package's PostCSS config, which already ran the three scoping plugins). 

Rendering a survey with the pre-PR UMD on a Tailwind-style host page produces zero host breakage. So the "before: 1 / 71" numbers in the verification table appear to come from dist/survey-ui.css, not the UMD, and "Cloud users get the fix on redeploy" likely doesn't apply , the served bundle was already scoped. 

Worth doing before closing: 

- (a) fix the verification section so it describes the artifact that actually changes (the standalone survey-ui.css, i.e. direct npm consumers of @formbricks/survey-ui/styles plus drift protection) 

- (b) re-verify the original ENG-1333 report against a freshly built main UMD - if it still reproduces there, the leak has a different vector and this PR alone won't close it. 

The shared-plugin extraction and the raw-artifact fix are still worth merging either way.

packages/vite-plugins/postcss-scope-fbjs.cjs
```diff
@@ -0,0 +1,139 @@
...
+// descendants still pick them up.
+const scopeLayerTheme = () => {
+  return {
+    postcssPlugin: "postcss-scope-layer-theme",
+    AtRule: {
+      layer: (atRule) => {
+        if (atRule.params !== "theme") return;
+        atRule.walkRules((rule) => {
+          rule.selectors = rule.selectors.map((sel) => {
+            if (sel === ":root" || sel === ":host") return "#fbjs";
```
itsjavi, line 63: regression test only covers a frozen fixture, so a Tailwind output-shape change would silently reintroduce the leak.

These plugins match Tailwind's current emission shapes exactly (`params === "properties"`, `sel === ":root" || sel === ":host"`), and the vitest suite feeds them a frozen fixture - so a future Tailwind upgrade that changes the emitted selector/layer shape (e.g. `:root:where(...)`, functional `:host(...)`) would reintroduce the host-page leak with all tests green. 

Consider adding a guard against the real artifact: either a test that runs the actual Tailwind compile of globals.css through the chain, or a post-build assertion on dist/survey-ui.css (grep for `@layer properties`, `@property --tw-`, and bare `:root`/`:host`) so a dependency bump can't silently regress this.

itsjavi (changes requested):
Verified locally with a before/after kitchen-sink host page: the code is correct and dist/survey-ui.css is fully scoped after this PR, but note the served surveys.umd.cjs was already clean on main, so please double-check ENG-1333 actually reproduces against a fresh main build and adjust the verification claims; details in my two inline comments.

Here are my visual tests:

Host page clean design:

Image

Before this PR:

Image

After:

Image

itsjavi (approved):

PR #8379, opened 2026-06-29
fix: drops responses from the response queue if they encounter a 400 error

pandeymangg:
Fixes https://linear.app/formbricks/issue/ENG-1319/survey-client-surfaces-400-response-is-already-finished-as-generic

## Bug
Client retried a permanent 400 "already finished" rejection like a network error and never dropped it from the queue, so users got a fake "servers unreachable" screen with an infinite Retry loop.

## Fix
Detect terminal 4xx errors, stop retrying, drop the dead item from the queue, and show an accurate "this response has already been completed" message.


packages/surveys/src/lib/response-queue.ts
```diff
@@ -205,8 +205,45 @@ export class ResponseQueue {
...
+   * createResponse), and 429 (rate limited — transient).
+   */
+  private isTerminalClientError(error?: ApiErrorResponse): boolean {
+    if (!error) return false;
+    return (
+      error.status >= 400 &&
+      error.status < 500 &&
+      error.status !== 401 &&
+      error.status !== 404 &&
+      error.status !== 429
```
xernobyl: [P3] This drops *all* 4xx except 401/404/429, which is broader than ENG-1319 (the 400 "already finished" loop). Worth noting that 408 (Request Timeout) and 425 (Too Early) are semantically retryable but would now be dropped without retry. Near-zero practical impact for these endpoints, so flagging only so the broadening is a conscious decision rather than an accident.

packages/surveys/src/lib/response-queue.ts
```diff
@@ -205,8 +205,45 @@ export class ResponseQueue {
...
+   * (e.g. a second tab / email link scanner finished the same single-use response).
+   * Server returns 400 "Response is already finished" or 409 on a stale update.
+   */
+  private isResponseAlreadyCompletedError(error?: ApiErrorResponse): boolean {
+    if (!error) return false;
+    return (
+      error.status === 409 ||
+      (error.status === 400 &&
+        typeof error.message === "string" &&
+        error.message.includes("already finished"))
```
xernobyl, line 246: [P3] This couples the surveys client to the exact English server string. It matches today (`put-response-handler.ts` returns `badRequestResponse("Response is already finished")`, hardcoded, not localized), so it works. But if that server message is ever reworded or i18n'd, detection silently degrades to the generic `ResponseSendingError` path with no test guarding the contract. Consider keying off a stable marker (an error code in `details`) long-term. Not blocking.

packages/surveys/src/lib/response-queue.ts
pandeymangg: Good catch — added 408 and 425 to the retryable exclusion set so only genuinely permanent 4xx get dropped. Pulled the list into a named `RETRYABLE_CLIENT_ERROR_STATUSES` set so the intent is explicit rather than a chain of `!==`.

packages/surveys/src/lib/response-queue.ts
pandeymangg: Done — server now sets a stable details.code: "response_already_finished" marker and the client keys off that. The 409 + message checks remain as fallbacks for older servers.

xernobyl (approved):
LGTM

PR #8381, opened 2026-06-29
ci(auth): run the Better Auth integration suite in CI

xernobyl:
## What does this PR do?

Wires the **Better Auth integration suite** (`apps/web/integration/*` + the `*.integration.test.ts` files) into CI. That harness is the only runtime validation of the Better Auth ↔ Formbricks boundary — it drives the real Better Auth handler against a real Postgres + Redis (no DB mocks), covering sign-up/in/out, email verification + password reset, 2FA, SSO provisioning, account deletion, the forward-auth proxy session read, and the recent security guards (password-less delete-user + inactive-user sign-in). Until now **no workflow ran it**: `pnpm test` mocks `@formbricks/database` and the e2e job is Playwright, so these ~35 tests only ran when someone ran them locally.

Changes:
- **New `.github/workflows/integration-tests.yml`** — stands up Postgres (pgvector) + Valkey services, provisions a fresh `formbricks_ba_test` (migrate + the two Better-Auth-shape ALTERs: `email_verified` → boolean, `Account.type` nullable), and runs `pnpm test:integration`. Services / Node / pnpm / `.env` setup mirror `e2e.yml`.
- **`global-setup.ts` — a `TEST_DB_PROVISIONED` escape hatch.** The harness normally provisions the test DB via `docker exec <container> pg_dump|psql`, which can't run against GitHub Actions service containers (reached over `localhost`, not `docker exec`-able by name). When `TEST_DB_PROVISIONED=1`, global-setup trusts the workflow's out-of-band provisioning and skips the docker clone (per-test isolation still comes from `resetDb`). **Local dev behavior is unchanged** when the flag is unset.
- **`fix(test)`: the 2FA integration test was already broken on the epic** — it queried the old plural `twoFactors` relation and asserted `.toHaveLength(1)`, but `51a770cdb` made `TwoFactor` one-per-user (`User.twoFactor TwoFactor?`, `@@unique(userId)`). It went unnoticed precisely because the suite wasn't in CI. Fixed to the singular relation. (This is the first thing the new CI gate caught.)

Base: `epic/better-auth-migration` (the harness lives on the epic, not `main`, so the workflow has to land where it can run).

## How should this be tested?

Verified locally against the dev Postgres + Valkey, both provisioning paths:

- **Escape-hatch / CI path (exact):** fresh `formbricks_ba_test` → migrated from scratch (confirmed `TwoFactor_userId_key` unique index) → the two ALTERs (`email_verified` confirmed `boolean`) → `TEST_DB_PROVISIONED=1 pnpm test:integration` → **35/35 pass**.
- **Local docker path (flag unset):** `pnpm test:integration` → global-setup's docker clone still runs → **35/35 pass** (no regression to local dev).
- Authoritative check: the new **Integration Tests** job on this PR.

### Required

- [x] Commented on my code in hard-to-understand bits (the escape hatch + the workflow's provisioning steps)
- [x] Ran `pnpm build` (the `@formbricks/database` package, for the migration runner)
- [ ] Merged the latest changes from main — N/A: this targets `epic/better-auth-migration` and is branched off its current head

### Appreciated

- [ ] If a UI change was made — N/A
- [ ] Updated the Formbricks Docs — N/A

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -65,6 +65,12 @@ jobs:
         run: pnpm install --frozen-lockfile --config.platform=linux --config.architecture=x64
         shell: bash
 
+      # Build the workspace packages the tests import (@formbricks/logger, cache, types, email, …) so
+      # vite can resolve their package.json `exports` → dist. Deps only (`^...`), not the Next app.
+      - name: Build workspace package dependencies
+        run: pnpm build --filter=@formbricks/web^...
+        shell: bash
+
       - name: Create .env
         run: pnpm dev:setup
         shell: bash
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,103 @@
+name: Integration Tests
+
+# Runs the Better Auth integration harness (apps/web/integration + the *.integration.test.ts files)
+# against a real Postgres + Valkey — the only runtime validation of the BA↔Formbricks boundary
+# (no DB mocks). The unit suite (`pnpm test`) mocks @formbricks/database, so these would otherwise
+# only run locally. Services + .env setup mirror e2e.yml.
+
+on:
+  pull_request:
+    paths:
+      - "apps/web/**"
+      - "packages/database/**"
+      - "packages/cache/**"
+      - ".github/workflows/integration-tests.yml"
+  workflow_dispatch:
+
+env:
+  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+  TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+
+permissions:
+  contents: read
+
+jobs:
+  integration:
+    name: Better Auth integration tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    services:
+      postgres:
+        image: pgvector/pgvector@sha256:9ae02a756ba16a2d69dd78058e25915e36e189bb36ddf01ceae86390d7ed786a
+        env:
+          POSTGRES_DB: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd="pg_isready -U postgres"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=5
+      valkey:
+        image: valkey/valkey@sha256:12ba4f45a7c3e1d0f076acd616cb230834e75a77e8516dde382720af32832d6d
+        ports:
+          - 6379:6379
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/dangerous-git-checkout
+
+      - name: Setup Node.js 22.x
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 22.x
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile --config.platform=linux --config.architecture=x64
+        shell: bash
+
+      - name: Create .env
+        run: pnpm dev:setup
+        shell: bash
+
+      - name: Point .env at the CI services + test database
+        run: |
+          sed -i "s|REDIS_URL=.*|REDIS_URL=redis://localhost:6379|" .env
+          sed -i "s|DATABASE_URL=.*|DATABASE_URL=postgresql://postgres:postgres@localhost:5432/formbricks_ba_test?schema=public|" .env
+          sed -i "s/ENTERPRISE_LICENSE_KEY=.*/ENTERPRISE_LICENSE_KEY=${{ secrets.ENTERPRISE_LICENSE_KEY }}/" .env
+        shell: bash
+
+      - name: Create the test database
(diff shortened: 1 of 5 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+        ports:
+          - 6379:6379
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./.github/actions/dangerous-git-checkout
```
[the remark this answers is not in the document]
xernobyl: Done in 94de6eb9c — added `persist-credentials: false` to the initial checkout. Confirmed `dangerous-git-checkout` runs its own `actions/checkout` (with its own auth) and nothing else in the job needs git auth, so dropping the persisted token is safe.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Shape the schema for Better Auth (emailVerified boolean, Account.type nullable)
+        run: |
+          psql "postgresql://postgres:postgres@localhost:5432/formbricks_ba_test" -v ON_ERROR_STOP=1 <<'SQL'
+          ALTER TABLE "User" ALTER COLUMN email_verified DROP DEFAULT;
+          ALTER TABLE "User" ALTER COLUMN email_verified TYPE boolean USING (email_verified IS NOT NULL);
+          ALTER TABLE "User" ALTER COLUMN email_verified SET DEFAULT false;
+          ALTER TABLE "User" ALTER COLUMN email_verified SET NOT NULL;
+          ALTER TABLE "Account" ALTER COLUMN type DROP NOT NULL;
+          SQL
```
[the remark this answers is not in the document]
xernobyl: Good catch on the template-injection / special-char breakage. Rather than work around it, I removed the `ENTERPRISE_LICENSE_KEY` line entirely (94de6eb9c): the harness drives the enterprise-gated flows (e.g. SSO provisioning) by calling the functions directly, and I verified the full suite is green with the key empty — so the secret isn't needed and is no longer interpolated into the workflow at all.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Create .env
+        run: pnpm dev:setup
+        shell: bash
+
+      - name: Point .env at the CI services + test database
+        run: |
+          sed -i "s|REDIS_URL=.*|REDIS_URL=redis://localhost:6379|" .env
+          sed -i "s|DATABASE_URL=.*|DATABASE_URL=postgresql://postgres:postgres@localhost:5432/formbricks_ba_test?schema=public|" .env
+          sed -i "s/ENTERPRISE_LICENSE_KEY=.*/ENTERPRISE_LICENSE_KEY=${{ secrets.ENTERPRISE_LICENSE_KEY }}/" .env
+        shell: bash
```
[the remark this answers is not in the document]
xernobyl: Fixed in 94de6eb9c — extracted the five ALTERs into `apps/web/integration/ba-test-schema-shape.sql` as the single source of truth. `global-setup.ts` reads + applies that file (local), and the workflow applies the same file via `psql -f` (CI), so the two provisioning paths can't drift.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,115 @@
+name: Integration Tests
+
+# Runs the Better Auth integration harness (apps/web/integration + the *.integration.test.ts files)
+# against a real Postgres + Valkey — the only runtime validation of the BA↔Formbricks boundary
+# (no DB mocks). The unit suite (`pnpm test`) mocks @formbricks/database, so these would otherwise
+# only run locally. Services + .env setup mirror e2e.yml.
+
+on:
```
BhagyaAmarasinghe, line 8: The new Better Auth integration suite is a standalone pull_request workflow, not part of the required PR Check Summary, and it also has no merge_group trigger. The live branch ruleset: required checks are PR Check Summary, CLA, SonarQube, build/e2e/lint/unit, but not Better Auth integration tests; merge queue is enabled. So this suite can fail without blocking a merge, and it will not validate merge-queue refs. Wire it into .github/workflows/pr.yml and include it in the summary needs, or update the ruleset and add merge_group.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,115 @@
...
+  pull_request:
+    # apps/web holds the auth code under test; packages/** because the harness transitively imports
+    # several workspace packages (@formbricks/logger, cache, types, email, database) whose changes
+    # can affect the BA boundary.
+    paths:
```
BhagyaAmarasinghe, line 15: The path filter misses files that can change this job’s behavior, especially pnpm-lock.yaml, root package.json, pnpm-workspace.yaml, and .github/actions/dangerous-git-checkout/action.yml. A lockfile-only Better Auth/Prisma transitive update or checkout action change can skip the exact suite meant to catch BA integration regressions.

BhagyaAmarasinghe (changes requested):

.github/workflows/integration-tests.yml
xernobyl: Added a `merge_group` trigger in 46d8cc83d so the suite runs in the merge queue alongside `pr.yml` / `docker-build-validation.yml` — that's what lets it gate once it's a required check (merge_group ignores `paths`, so it always runs in the queue). Promoting it to a *required* check itself is a ruleset change on your side — it's intentionally not required yet (this PR adds the capability, per the plan). Happy to coordinate whenever you want to flip it on.

.github/workflows/integration-tests.yml
xernobyl: Good catch — widened the `pull_request` path filter in 46d8cc83d to include `pnpm-lock.yaml`, root `package.json`, `pnpm-workspace.yaml`, and `.github/actions/dangerous-git-checkout/action.yml`, so a lockfile-only Better Auth/Prisma transitive bump or a checkout-action change now triggers the job instead of silently skipping it. (merge_group still covers the queue path regardless.)

BhagyaAmarasinghe (approved):

PR #8382, opened 2026-06-29
fix: fixes the environment and responses api for legacy sdks breaking

pandeymangg:
## What does this PR do?

Part of **ENG-1067** (standardize survey/app/translation language tags on BCP-47).

After language codes are canonicalized to region-tagged BCP-47 (`de` → `de-DE`, `hi` → `hi-IN`), SDK clients can still send **legacy, non-canonical codes** — most importantly anonymous clients, whose language lives only in local storage and never round-trips through the server (so it's never re-canonicalized), and older SDK versions. This PR makes the **survey renderer** and **responses API** resilient to those codes so nothing breaks and the data stays clean.

> Context: a survey's content i18n keys and `Language.code` are now canonical (`hi-IN`), but the SDK-facing environment back-compat still exposes the bare legacy code (`hi`) so older clients keep matching. That mismatch (`languages` has `hi`, content is keyed `hi-IN`) is what this PR handles end-to-end.

### Changes

**Survey renderer (`packages/surveys`)**
- `index.ts` — resolve the survey **content** language canonically before rendering (`getI18nLanguage`, preserving the `default` sentinel). A legacy/mismatched code (`hi`) now renders the correct `hi-IN` content (and responses record `hi-IN`) instead of silently falling back to the default language.
- `recall.ts` — `parseRecallInformation` now reads via `getLocalizedValue` (which falls back to the `default` key) instead of indexing `headline[languageCode]` / `subheader[languageCode]` directly. Fixes `Uncaught TypeError: Cannot read properties of undefined (reading 'includes')` when the active code isn't a content key.

**Responses API (v1 client, v2 client, v2 management)**
- Canonicalize `Response.language` **on write** (`hi` → `hi-IN`), preserving `default` and unresolvable values. Stale SDK caches can submit a legacy code at any time; normalizing server-side keeps the `Response` table canonical regardless of what the client sends.

**Language switcher (`packages/surveys`)**
- `language-switch.tsx` — dedupe the dropdown by canonical code, so the bare legacy aliases the environment back-compat adds (e.g. `hi` next to `hi-IN`) don't appear as duplicate options. The canonical entry is kept, so selecting it sets the correct content language.
- `language-display-name.ts` — fall back from the exact code → language+script → bare language for the native display name, so region-tagged canonical codes without their own entry render in-script: `hi-IN` → "हिन्दी", `zh-Hans-CN` → "简体中文", `da-DK` → "Dansk" (previously showed the raw code, e.g. "Hi-IN").

## How should this be tested?

- **Renderer / recall:** `pnpm --filter @formbricks/surveys exec vitest run src/lib/recall.test.ts` — incl. a regression for a language code that isn't a content key (no longer throws).
- **Labels:** `pnpm --filter @formbricks/surveys exec vitest run src/lib/language-display-name.test.ts` — region-tagged fallback (`hi-IN` → "हिन्दी", `zh-Hans-CN` → "简体中文").
- **Responses:** `pnpm --filter @formbricks/web exec vitest run app/api/v1/lib/utils.test.ts` — `language` canonicalized on write; `default`/unresolvable preserved.
- **Manual:** with an SDK still holding a legacy code (e.g. `hi`) against a migrated survey, track an action: the survey renders (in `hi-IN`), no recall crash, the switcher shows a single in-script entry, and the stored `Response.language` is `hi-IN`.

## Notes

- Complements the SDK-facing back-compat that exposes legacy language codes (environment "send-both" + user-state de-canonicalization) shipped separately; this PR is the runtime/data-integrity half.
- No DB or schema changes here.

### Required

- [x] Ran `pnpm build`
- [ ] Merged the latest changes from the epic onto my branch

### Appreciated

### DIFF
--- a/apps/web/app/api/client/[workspaceId]/responses/lib/response.ts
+++ b/apps/web/app/api/client/[workspaceId]/responses/lib/response.ts
@@ -4,6 +4,7 @@ import { Prisma } from "@formbricks/database/prisma";
 import type { TContactAttributes } from "@formbricks/types/contact-attribute";
 import type { TResponse } from "@formbricks/types/responses";
 import type { TTag } from "@formbricks/types/tags";
+import { normalizeResponseLanguage } from "@/lib/response/utils";
 import { evaluateResponseQuotas } from "@/modules/ee/quotas/lib/evaluation-service";
 
 type TQuotaEvaluationResponseInput = {
@@ -31,6 +32,9 @@ export const createResponseWithQuotaEvaluation = async <TInput extends TQuotaEva
   responseInput: TInput,
   createResponse: (responseInput: TInput, tx: Prisma.TransactionClient) => Promise<TResponse>
 ) => {
+  // Canonicalize once so quota evaluation uses the same code persisted on the response (createResponse
+  // canonicalizes the stored value via the same helper). Keeps a request internally consistent.
+  const canonicalLanguage = normalizeResponseLanguage(responseInput.language) ?? undefined;
   return await prisma.$transaction(async (tx: Prisma.TransactionClient) => {
     const response = await createResponse(responseInput, tx);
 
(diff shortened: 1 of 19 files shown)


xernobyl (approved):
LGTM

PR #8390, opened 2026-06-30
fix: Canonicalization of legacy language codes

pandeymangg:
## What does this PR do?

**ENG-1067** — standardizes survey, app, and translation language tags on a single canonical **BCP-47** format across the whole stack (catalog, app UI, surveys runtime, APIs, and the database).

Before this, language codes were inconsistent across layers: the survey catalog mixed bare and region-tagged codes, the app UI used region-tagged codes, and the surveys runtime used bare ISO codes. The same language could be stored as `de`, `de-DE`, or `de_DE` depending on where it originated, which broke language matching, content lookup, and analytics.

This is the **epic → main** roll-up of the work landed across the sub-PRs below. **Legacy codes remain accepted everywhere** (inbound back-compat), so this is not a breaking change for existing integrations or already-deployed SDKs.

### Canonical format ("Rule B")

- Every code is **region-tagged** BCP-47: `lang-REGION` (and `lang-Script-REGION` for Chinese) — e.g. `de` → `de-DE`, `pt` → `pt-BR`, `zh-Hans` → `zh-Hans-CN`.
- The default region comes from **CLDR likely-subtags** (`Intl.Locale(code).maximize()`), Product-approved (e.g. `ar` → `ar-EG`, `pt` → `pt-BR`).
- **All ~200 languages are kept** — nothing is removed from the picker.
- Non-default regions are preserved as-is (`de-AT` stays `de-AT`).

## What's included (by layer)

1. **Foundation — `@formbricks/i18n-utils`** (#8349): `LANGUAGE_CANONICAL_MAP` (catalog ∪ observed production codes) + `normalizeLanguageCode()` (exact → case-insensitive → CLDR fallback). Single source of truth used everywhere else.
2. **App write paths** (#8352): canonical language picker + `createLanguage` normalizes & validates the code on write, so new languages are always stored canonical.
3. **v3 API** (#8355): canonicalizes survey language codes on read/write while still accepting legacy inbound codes (curated back-compat map; un-mapped codes preserved).
4. **Runtime / renderer** (#8357): survey locale bundles renamed to their canonical CLDR-default tag (`de.json` → `de-DE.json`, … `zh-Hans.json` → `zh-Hans-CN.json`), with a script-preserving fallback chain in the renderer's i18n config.
5. **Data migration** (#8367): canonicalizes everything already stored — `Language.code` rows (with merge handling for bare+region collisions), survey content i18n keys, `Response.language`, the `Response.contactAttributes` language snapshot, and the contact `language` attribute. Idempotent, atomic (single transaction), batched, and non-destructive (junk codes left untouched + logged, aliases preserved).
6. **SDK back-compat for legacy clients** (#8382): keeps already-deployed SDKs working after canonicalization (see below).

## SDK / mobile back-compat (the tricky part)

SDKs (web js-core + native via WebView) can still hold **legacy** codes — especially anonymous clients, whose language lives only in local storage and never round-trips through the server. To keep them working *without* forcing an SDK update:

- **Environment API "send-both":** alongside each canonical survey language we expose a duplicate entry under its bare legacy code (`de-DE` + `de`), so an SDK matching by exact code still finds its language. Multi-language surveys only; canonical entry kept first.
- **User-state "don't-heal":** the contact `language` returned to the SDK is de-canonicalized to the bare legacy form, so older clients (notably old React Native builds with a case-sensitive matcher) keep matching. The DB stays canonical.
- **Renderer resilience:** the survey renderer resolves the content language canonically (so a legacy `hi` renders the canonical `hi-IN` content), and recall parsing falls back via `getLocalizedValue` instead of indexing content by the raw code — fixes a `Cannot read properties of undefined` crash on mismatched codes.
- **Language switcher:** de-dupes by canonical (drops the back-compat aliases) and shows native-script labels for region-tagged codes (`hi-IN` → "हिन्दी").
- **`Response.language` normalize-on-write:** create + update, client + management paths all canonicalize the language on write, so a stale legacy code submitted by any SDK never re-pollutes the table. Quota evaluation uses the same canonical value.

The renderer (`surveys.umd.cjs`) is **served by the server, not bundled in the SDK**, so old SDKs automatically pick up the fixed renderer once their cache refreshes (see Rollout).

## How should this be tested?

- **Unit:** `pnpm --filter @formbricks/i18n-utils test`, `pnpm --filter @formbricks/surveys test`, and the response/renderer/i18n suites in `apps/web` (`normalizeLanguageCode`, recall hardening, native-name fallback, response canonicalize-on-write, environment send-both, language-switch dedup).
- **Migration:** dry-run the merge planner against a production `Language` dump (validated: 8 collision workspaces resolve as expected, **0 alias-loss**); `pnpm --filter @formbricks/database build` confirms the migration bundles with `@formbricks/i18n-utils` inlined.
- **e2e:** the management responses Playwright spec asserts `Response.language` is canonicalized on create + update.
- **Manual / SDK:** with an SDK holding a legacy code (`hi`) against a migrated survey — survey shows, renders `hi-IN` content, no crash, switcher shows a single native entry, and the stored `Response.language` is `hi-IN`. (Test a multi-step survey so a response update/PUT happens too.)

## Rollout / sequencing

1. Deploy the code (canonical writes + renderer + SDK back-compat) **first** — legacy reads/writes are still accepted.
2. Run the **data migration** (framework data migration, runs once in a transaction). Recovery relies on the existing DB backup policy — merges are lossy, so there is no down-migration.
3. **Purge the CDN copy of `/js/surveys.umd.cjs`** on deploy so clients pick up the fixed renderer (it's served with `s-maxage=30d`; the client `max-age` is 1h). Otherwise a stale cached renderer against migrated data could briefly mis-render for legacy codes.

## Notes

- **No breaking change:** legacy inbound codes are accepted across v3, the client APIs, and the renderer; existing SDKs keep working unchanged.
- The React Native SDK's case-sensitive language matcher fix ships in the RN repo separately; the server back-compat (send-both + don't-heal) covers un-updated RN builds in the meantime.

### Required

- [x] Ran `pnpm build`
- [x] Merged the latest changes from main onto my branch

### Appreciated


apps/web/lib/i18n/utils.ts
```diff
@@ -239,3 +239,26 @@ export const appLanguages = [
...
+ * so the two line up and exact matching still works. The DB stays canonical throughout; the renderer
+ * canonicalizes whatever it receives, so content lookup is unaffected.
+ *
+ * Both call sites MUST use this single helper so the codes they emit stay aligned.
+ *
+ * Remove once SDK clients holding pre-canonicalization codes have drained.
+ *
+ * @returns the bare legacy code (`de-DE` → `de`), or `null` when the code is already bare.
+ */
+export const toLegacyLanguageCode = (code: string): string | null => {
```
xernobyl: **`toLegacyLanguageCode` collapses script languages to a code no legacy client holds — breaks the Chinese back-compat.**

This strips to the bare language (`code.split("-")[0]`), so `zh-Hans-CN` → `zh` and `zh-Hant-TW` → `zh`. But the catalog never had a bare `zh` — Chinese was always `zh-Hans` / `zh-Hant` / `zh-CN` / `zh-TW` / `zh-HK`. So the legacy alias this emits (in the environment "send-both" and the user-state echo) matches **nothing** an already-deployed Chinese client holds:

- A client holding `zh-Hans` won't match the emitted `zh` (exact match: `zh-Hans` ≠ `zh`) → its Chinese survey is hidden, which is exactly what this back-compat exists to prevent.
- A single `zh` alias also can't distinguish Simplified vs Traditional, and `normalizeLanguageCode("zh")` → `zh-Hans-CN`, so Traditional clients would land on Simplified content.

It works for non-script languages only because their legacy form *was* bare (`de-DE` → `de` matches old `de` clients).

Fix: preserve the script subtag — strip only the region. This matches the `[locale.language, locale.script]` pattern already used in `i18n.config.ts`'s `resolveFallbackBundles`:

```ts
export const toLegacyLanguageCode = (code: string): string | null => {
  try {
    const locale = new Intl.Locale(code);
    // legacy form = language (+ script for script langs like Chinese), region dropped
    const legacy = [locale.language, locale.script].filter(Boolean).join("-");
    return legacy.toLowerCase() === code.toLowerCase() ? null : legacy;
  } catch {
    const bare = code.split("-")[0].toLowerCase();
    return bare === code.toLowerCase() ? null : bare;
  }
};
```

So `zh-Hans-CN` → `zh-Hans`, `zh-Hant-TW` → `zh-Hant`, `de-DE` → `de`, `de` → `null`.

(`zh-Hans` is the dominant legacy spelling; clients holding `zh-CN`/`zh-TW` would need those emitted too if we want full coverage, but `zh-Hans`/`zh-Hant` is strictly better than bare `zh`, which covers nobody.)

pandeymangg: replaced this with a reverse map (canonical -> legacy codes)

xernobyl (commented):
Re-review (head `b7b78ad`) 🙏 — the correctness and back-compat engineering here is strong. My earlier `toLegacyLanguageCode` finding is resolved (the `LEGACY_CODES_BY_CANONICAL` reverse map + defensive copy — nice), CodeRabbit's threads are addressed, and I verified the canonical map is idempotent/collision-free and round-trips for the codes that matter (`zh`→`zh-Hans-CN`, `pt`/`pt-BR`, `tl`/`fil-PH`, `en`/`en-US`), the JSON i18n rewrite preserves the `default` sentinel + block/ending order, and write-side canonicalization is consistent across v1/v2/v3/client.

The one thing I'd treat as **gating before this runs against a large production DB** is the migration transaction/lock model (P1 inline). Two P3s inline, plus one more note:

**[P3] v3 write-validation regex rejects the new canonical tags.** `V3_SURVEY_LOCALE_CODE_REGEX` in `apps/web/app/api/v3/surveys/language.ts` requires a 2-letter base + 2-letter region, so it rejects `fil-PH`, `eo-001`, etc. Those are picker-supported and accepted by `createLanguage`, so *creating* such a survey language via the v3 API fails (`normalizeV3SurveyWriteLanguageCode` → null). Not a regression (the regex is unchanged from `main`) and reads/existing languages are unaffected — but this PR makes `tl`→`fil-PH` the canonical target, so the interaction is newly reachable.

_(FYI: the failing "Better Auth integration tests" check is unrelated CI noise from a workflow that landed on `main`; #8402 fixed it — a fresh push here will clear it, a re-run won't.)_

packages/database/migration/20260625104904_canonicalize_language_codes/migration.ts
```diff
@@ -0,0 +1,360 @@
...
+      logger.info(
+        `Survey content progress: ${Math.min(i + SURVEY_BATCH_SIZE, surveyUpdates.length).toString()}/${surveyUpdates.length.toString()}`
+      );
+    }
+    stats.surveysContentUpdated = surveyUpdates.length;
+
+    // ---------------------------------------------------------------------------------------------
+    // 3. Response.language (skip NULL, the "default" sentinel, and already-canonical values)
+    // ---------------------------------------------------------------------------------------------
+    logger.info("Canonicalizing Response.language...");
```
xernobyl: **[P1] The whole migration runs in one ~30-min transaction, and this step full-scans the unindexed `Response` table.**

`migration-runner.ts` wraps `migration.run` in a single `prisma.$transaction(…, { timeout: 30m })` and every step uses that one `tx`. `Response.language` has no index (only `createdAt` / `surveyId+createdAt` / `contactId+createdAt`), so `SELECT DISTINCT language FROM "Response"` here + the follow-up `UPDATE … WHERE language = …` are full sequential scans, and the ContactAttribute step adds a second scan — all inside the one transaction, at container startup during a live deploy.

On a large `Response` table this can exceed the 30-min budget → the **entire** transaction rolls back (all four steps lost) and the next deploy re-runs from scratch, never converging; meanwhile it holds row locks long enough to stall response ingestion.

Suggest running the big-table steps (Response, ContactAttribute) as short, id/keyset-batched, **committed-per-chunk** statements *outside* the mega-transaction (the runner already exposes an autocommit `prisma` handle) so progress commits and locks release promptly. The current array-chunking doesn't break up the transaction or the per-statement lock footprint.

packages/database/migration/20260625104904_canonicalize_language_codes/migration.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    // Snapshot multi-language survey IDs BEFORE Step 1 mutates SurveyLanguage. A survey that links two
+    // codes collapsing to the same canonical (e.g. `de` + `de-DE`) gets deduped down to a single link by
+    // the merge below; querying `COUNT(*) > 1` afterwards would then skip it and leave its legacy content
+    // key un-rewritten. Pre-merge `COUNT(*) > 1` is a superset of the post-merge set (a merge only ever
+    // reduces a survey's link count) and single-language surveys only hold the never-rewritten `default`
+    // key — so this is exactly the set Step 2 must process.
+    const multiLanguageSurveyIds = (
+      await tx.$queryRaw<{ surveyId: string }[]>`
+        SELECT "surveyId" FROM "SurveyLanguage" GROUP BY "surveyId" HAVING COUNT(*) > 1
```
xernobyl, line 55: **[P2] Skipping single-language surveys relies on an unenforced invariant.**

This assumes single-language surveys only carry the never-rewritten `default` key. But a survey that once had ≥2 languages, accumulated a `de`/`pt`/`zh` content key, then dropped back to one language will still hold that stale legacy key — and it's skipped here, so its keys stay un-canonicalized in the DB. The runtime renderer canonicalizes on read so display still resolves, but the DB won't be fully canonical as intended. Worth validating the invariant against prod data, or widening the candidate set.

packages/database/migration/20260625104904_canonicalize_language_codes/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+  return delta !== 0 ? delta : a.id.localeCompare(b.id);
+};
+
+/**
+ * Plan how every `Language` row should change. Rows are grouped per workspace by their canonical code:
+ * a singleton group is a plain relabel; a group of 2+ is a collision (e.g. bare `de` + region `de-DE`)
+ * where one row survives (preferring a row already at the canonical code, else the oldest) and the rest
+ * are absorbed. The `@@unique([workspaceId, code])` constraint is why collisions can't just be relabelled.
+ */
+export const planLanguageMerges = (languages: LanguageRow[]): LanguageMergePlan => {
```
xernobyl, line 131: **[P3] The planning logic here is the trickiest, most data-mangling-prone code in the PR and has no unit tests.**

`planLanguageMerges` / `planSurveyLanguageMoves` / `rewriteI18nKeys` / `pickI18nValue` handle collision survivorship, `SurveyLanguage` composite-PK dedup + flag promotion, and order-preserving block/ending rewrites — exactly where a silent data bug would hide. `canonical.ts` / `language.ts` are well covered; these pure functions deserve direct tests given they migrate production data.

xernobyl:
**`updateLanguage` write path still bypasses canonicalization — the last hole in the "`Language.code` stays canonical regardless of caller" invariant.**

`createLanguage` was just hardened (normalize + `CANONICAL_LANGUAGE_CODES` allowlist, `apps/web/lib/language/service.ts:66-71`), but the update path has neither guard:

- `ZUpdateLanguageAction.languageInput` is `ZLanguageInput` (carries `code`) — `apps/web/modules/survey/multi-language-surveys/lib/actions.ts`.
- `updateLanguageAction` passes it straight to `updateLanguage(…, parsedInput.languageInput)`.
- `updateLanguage` does `data: { ...languageInput, updatedAt: new Date() }` (`apps/web/lib/language/service.ts`), spreading the raw `code`. The `validateInputs(…, ZLanguageUpdate)` call's parsed result is ignored, so nothing strips it.

So `updateLanguageAction({ workspaceId, languageId, languageInput: { code: "anything", alias } })` writes an arbitrary, non-canonical code to `Language.code` — exactly what the create-side hardening now prevents. Normal UI flow is fine (it sends the existing code), but the invariant no longer holds "regardless of caller," and create/update are now inconsistent.

Fix (either):
1. Mirror `createLanguage`: normalize + `CANONICAL_LANGUAGE_CODES.includes(...)` check, and write the canonical `code`; or
2. Cleaner (the code shouldn't change on update): switch `ZUpdateLanguageAction.languageInput` to the existing alias-only `ZLanguageUpdate` so `code` is stripped before the service.

packages/surveys/src/components/general/language-switch.tsx
```diff
@@ -45,6 +46,22 @@ export function LanguageSwitch({
...
+    if (!surveyLanguage.enabled) continue;
+    const code = surveyLanguage.language.code;
+    const canonical = normalizeLanguageCode(code) ?? code;
+    const existing = languagesByCanonical.get(canonical);
+    if (!existing || code === canonical) {
+      languagesByCanonical.set(canonical, surveyLanguage);
+    }
+  }
+  const visibleLanguages = [...languagesByCanonical.values()];
+
```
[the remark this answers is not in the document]
xernobyl: Verified against the current head (`8bbb5d6a`): still unaddressed. The dedup Map was added (it keeps the canonical row), but the comparison this thread is about now lives at **`language-switch.tsx:73`** and is still a raw string compare:

```ts
const calculatedLanguageCode = languageCode === defaultLanguageCode ? "default" : languageCode;
```

For a survey with a bare+region duplicate whose default is the legacy row (`hi` default + `hi-IN`), the dedup surfaces the `hi-IN` option while `defaultLanguageCode` stays `hi`, so selecting it yields `"hi-IN"` instead of `"default"` → the default language resolves the wrong i18n key. Narrow/transient (only unmigrated duplicate surveys, in the window before the data migration runs), but real and shipping to `main`. The suggested canonical-to-canonical comparison still applies.

xernobyl (dismissed):
Approving. I've reviewed this across the full series and re-verified the current head (`e7b58642`).

**Verified fixed:**
- `updateLanguage` write path — now alias-only (`ZUpdateLanguageAction` → `ZLanguageUpdate`, service writes `alias` explicitly, with a regression test asserting `code` is never persisted). Closes the last hole in "`Language.code` stays canonical regardless of caller."
- Data migration — multi-language survey IDs snapshotted before the merge mutates `SurveyLanguage`; `blocks`/`endings` rebuilt with `array_agg(... ORDER BY ord) WITH ORDINALITY`; sequential writes on `tx`; high-volume tables moved to idempotent committed chunks.
- SDK back-compat — `toLegacyLanguageCodes` reverse map emits every legacy alias (fixes the Chinese `zh-Hans`/`zh-Hant` case); env send-both + user-state echo aligned.
- v3 resolver — PATCH preserves the stored code, CREATE canonicalizes; wider locale regex.
- `createLanguage` rejects non-catalog (CLDR-only) codes; `resolveFallbackBundles` now has colocated tests.

**One non-blocking follow-up (recommend a fast-follow, not gating this):**
- `language-switch.tsx` still compares the selected code against the raw `defaultLanguageCode` (line ~73). For a survey with a bare+region duplicate whose default is the legacy row, the deduped picker can resolve the default to the wrong i18n key. Narrow and transient — only unmigrated duplicate surveys in the window before the migration runs, and it degrades to a fallback render, not data loss. A canonical-to-canonical comparison fixes it.

**Process reminders (not blockers):**
- CodeRabbit's incremental reviews are disabled, so its open threads are anchored to pre-rework lines and are largely stale/addressed — reviewed manually instead. A fresh full CodeRabbit pass wouldn't hurt.
- Run the data migration in an off-peak window and re-run on timeout — it's idempotent and reads are back-compat, so a partial run converges on retry.

xernobyl (approved):
Approving. Reviewed the epic→main roll-up at head b610f5f and it's good to go.

**Scope** — all 92 files are ENG-1067 (i18n-utils foundation, app write paths, v1/v2/v3 canonicalization, runtime/renderer bundles, the data migration, response canonicalize-on-write, SDK back-compat, locales). Nothing unrelated in the roll-up. This is the aggregate of the already-reviewed sub-PRs (#8349–#8420) plus the two clean main→epic syncs (#8425/#8427); all 18 review threads are resolved.

**Migration** (the one irreversible, production-run artifact) — sound two-tier design: bounded catalog + survey-content steps run inside the runner transaction; the high-volume `Response`/`ContactAttribute` tables run in committed 5000-row chunks outside it. I confirmed the hot-table steps are genuinely idempotent and resumable (walked by PK, `<> canon` skips already-canonical rows, converges on retry), and non-destructive (junk left + logged, aliases preserved) — consistent with the dump validation (8 collision workspaces, 0 alias-loss).

**CI** — all gates green on this exact head, including the load-bearing ones: E2E (15m40s, incl. the Response.language canonicalize-on-write spec), Docker Build (16m41s), Unit, Build, SonarQube, Better Auth, Linters, Translations.

Two non-blocking follow-ups for whenever:
- The PR body says the migration is "atomic (single transaction)"; it deliberately isn't for the two hot tables (correct choice — they're chunked + idempotent). Worth tightening the wording since the rollout note leans on DB backups for recovery.
- Don't forget the rollout step 3: purge the CDN copy of `/js/surveys.umd.cjs` on deploy so legacy clients pick up the fixed renderer.

PR #8392, opened 2026-06-30
fix(ai): surface Vertex 429 quota errors instead of failing silently

Dhruwang:
## What

AI translation (and AI survey generation) silently did nothing when Vertex AI returned **429 "Resource exhausted"**. This surfaces the failure to the user with a clear, retryable message and makes quota exhaustion visible to ops.

Linear: [ENG-1502](https://linear.app/formbricks/issue/ENG-1502/ai-translation-silently-fails-on-vertex-ai-429-quota-exhaustion-in) · Sentry: FORMBRICKS-14M

## Why

`translateFields` → `generateOrganizationAIObject` calls Vertex through the AI SDK. Under load Vertex returns 429; after the SDK exhausts its built-in retries it rethrows a raw `APICallError` / `RetryError`. That error name isn't in `EXPECTED_ERROR_NAMES`, so the action client masked it as the generic server error (+ Sentry) and the translation modal's result branch showed nothing — a **silent no-op**. There was no signal to the user and only Sentry noise for ops.

## How

- **`packages/ai` — `classifyAIProviderError`** (new): detects a provider 429 across both a direct `APICallError` and a retry-exhausted `RetryError` (unwraps the latest underlying `APICallError`). Returns `{ isQuotaExhausted, isRetryable, statusCode, retryAfterSeconds }`, parsing `retry-after` as either delay-seconds or an RFC-7231 HTTP-date. The `ai` SDK error types stay encapsulated in the package that owns the dependency.
- **`apps/web/lib/ai/service.ts`**: both `generateOrganizationAIText` / `…Object` classify the caught error and, on a quota 429, throw `TooManyRequestsError(AI_ERROR_CODES.QUOTA_EXCEEDED, retryAfterSeconds)`. `TooManyRequestsError` is an expected error, so the action client returns its message to the client as `serverError` **without** capturing to Sentry. The error log now includes `statusCode` / `isQuotaExhausted` / `isRetryable` for ops.
- **`manage-translations-modal.tsx`**: maps the quota code to a localized "temporarily unavailable — try again shortly" message; the catch/fallback now shows a generic message instead of leaking a raw error code.
- **`v3 surveys/generate` route**: maps `TooManyRequestsError` → `problemTooManyRequests` (HTTP 429) instead of a generic 502.
- **i18n**: `ai_translation_quota_exceeded` added to `en-US` and propagated to all locales.

### Backoff (requirement #3)

Relies on the AI SDK's built-in retries and surfaces an explicit "try again shortly" message rather than hand-rolling extra backoff inside a user-blocking, 45s-timeout action. No silent triple-429-then-throw: the user now gets clear feedback.

## Ops follow-up (not code)

Confirm whether this is a genuine Vertex quota ceiling for prod traffic and raise the quota / add monitoring. The new `statusCode: 429` / `isQuotaExhausted` logs make this greppable.

## Testing

- `packages/ai`: 10 unit tests for `classifyAIProviderError` (direct 429, non-429, malformed + HTTP-date retry-after, RetryError unwrap, abort/errorNotRetryable reasons, absent header, non-SDK error).
- `apps/web/lib/ai/service.test.ts`: 429 from text + object generation → `TooManyRequestsError`; non-quota errors rethrown unchanged; log shape updated.
- All suites green; typecheck + lint clean.

## Acceptance criteria

- [x] Vertex 429 shows the user a clear, non-silent message
- [x] Non-quota failures still handled gracefully (distinct from quota)
- [x] Quota exhaustion surfaced for ops (structured logs) instead of Sentry-only noise
- [x] No immediate triple-429-then-throw — explicit retry guidance


apps/web/lib/ai/service.ts
```diff
@@ -117,15 +130,22 @@ export const generateOrganizationAIObject = async <T = unknown>({
...
         isInstanceConfigured: aiConfig.isInstanceConfigured,
         errorCode: error instanceof AIConfigurationError ? error.code : undefined,
+        statusCode: providerError?.statusCode,
+        isQuotaExhausted: providerError?.isQuotaExhausted,
+        isRetryable: providerError?.isRetryable,
         err: error,
       },
       "Failed to generate organization AI object"
     );
+    if (providerError?.isQuotaExhausted) {
```
BhagyaAmarasinghe, line 146: `generateOrganizationAIObject` is shared beyond translation and v3 survey generation. When AI chart generation hits provider quota, this now throws `TooManyRequestsError("ai_quota_exceeded")`; the safe action client returns that as `serverError`, and `getTranslatedAIChartError` falls through to returning the raw code. The chart UI will `toast ai_quota_exceeded` instead of a user-facing translated message. Add the same quota-code mapping for AI chart errors, or centralize AI error-code translation for all AI consumers.

BhagyaAmarasinghe (changes requested):
Had one small comment. Also there seems to be merge conflicts

apps/web/lib/ai/service.ts
Dhruwang: Good catch — confirmed the chart toast would show the raw `ai_quota_exceeded` code. Fixed in 0419798f2: `getTranslatedAIChartError` now maps `ai_quota_exceeded` to a translated message (`workspace.analysis.charts.ai_quota_exceeded`, added to all 15 locales), with a test. Kept the mapping client-side like the translations modal since `@/lib/ai/service` is server-only; the code value is documented as needing to stay in sync. Centralizing AI error-code translation for all consumers is worth doing but broader than this PR.

BhagyaAmarasinghe (approved):

PR #8396, opened 2026-07-01
chore(experiments): Remove trial banner button and create new A/B exp…

harshsbhat:
## What & why

Screenshot 2026-07-01 at 9 51 35 AM

Iterates on the trial-banner navigation A/B experiment. Both experiment
variants now behave as a single clickable banner (no separate CTA button)
and emit a PostHog event on click so we can measure billing-page intent
consistently across the control and treatment groups.

Rolls the experiment to a fresh flag (`v2`) so we start from clean data.

## Changes

- **New experiment flag**: `a-b_navigation_rich-trial-banner` →
  `a-b_navigation_rich-trial-banner-v2`. Treatment is now the `test`
  variant (previously `new-trial-banner`).
- **Treatment banner (`TrialBannerNew`)**: the whole card is now a link to
  billing. Removed the "Go to Billing" button — the banner itself is the CTA.
- **Both variants** fire `main_nav_go_to_billing_clicked` on click, so the
  CTA event is tracked identically for control (`TrialAlert`) and treatment.

## PostHog setup (done separately)

- Create multivariate flag `a-b_navigation_rich-trial-banner-v2` with
  variants `control` and `test` (treatment key **must** be exactly `test`).

## Notes

- Styling is unchanged except two small tweaks tied to the behavior change:
  a hover state on the now-clickable treatment card, and removing the bottom
  margin that previously spaced the deleted button.
- Both variants emit the **same** event name with no variant property;
  splitting relies on PostHog's flag/experiment association. Can add a
  `variant` property if we want the event to distinguish groups on its own.
- Banner visibility is unchanged: Formbricks Cloud only, active trial,
  expanded sidebar.
- The i18n key `workspace.settings.billing.go_to_billing` is now unused
  (was the removed button label).


apps/web/locales/pt-PT.json
```diff
@@ -3667,7 +3666,7 @@
       "csv_columns": "Colunas CSV",
       "csv_data_preview": "Pré-visualização de dados",
       "csv_empty_column_headers": "O CSV contém cabeçalhos de coluna vazios. Todas as colunas devem ter um nome.",
-      "csv_file_too_large": "O ficheiro CSV é demasiado grande. O tamanho máximo é 2MB.",
+      "csv_file_too_large": "O ficheiro CSV é demasiado grande. O tamanho máximo é 15MB.",
```
jobenjada: are these changes related? 🤔

jobenjada (changes requested):

apps/web/locales/pt-PT.json
harshsbhat: nope, removed and resolved merge conflicts.

jobenjada (approved):

PR #8398, opened 2026-07-01
chore(auth): capture page/UTM/source attribution on signup & login

harshsbhat:
## What & why

Marketing has no visibility into where sign-ups and logins originate. Today the
auth pages (`/auth/login`, `/auth/signup`) drop any incoming query params, and the
PostHog `user_signed_up` / `user_signed_in` events only carry `auth_provider`,
`email_domain`, and `signup_source`.

This PR captures marketing attribution from the auth-page URL and attaches it to
those events, for **all providers** (credentials + SSO), so acquisition can be
analyzed in PostHog.

Example: landing on `/auth/login?page=home&utm_source=twitter` and signing up/in
now stamps `page` and `utm_*` onto the event.

## Captured params (fixed whitelist)

`page`, `utm_source`, `utm_medium`, `utm_campaign`, `utm_term`, `utm_content`,
`initial_source`. Anything else is ignored.

## How it works

1. **First-touch cookie** — a small `"use client"` `AttributionTracker` mounted in the
   auth layout reads the whitelisted params from the URL on load and writes them to a
   short-lived `fb_attribution` cookie (only if not already set). `SameSite=Lax`
   (+ `Secure` on HTTPS) so it survives the OAuth redirect; 30-minute lifetime so it
   can't linger across users/sessions.
2. **Server-side read at capture time** — the cookie is read via `next/headers`
   `cookies()` and spread into the existing PostHog captures:
   - `user_signed_up` (credentials) — `signup/actions.ts`
   - `user_signed_up` (SSO) — `ee/sso/lib/sso-handlers.ts`
   - `user_signed_in` (credentials + SSO login) — `sign-in-tracking.ts`
   The cookie is cleared once consumed.
3. **Link forwarding** — the login↔signup links forward the params so `?page=home`
   stays in the address bar when toggling between the two pages.

## Privacy

`fb_attribution` stores **only** the whitelist above — non-personal, non-user-unique
marketing values (source / campaign / landing page). No email, id, or device
fingerprint, so it's a functional attribution cookie rather than user tracking.

## Files

| File | Change |
|------|--------|
| `modules/auth/lib/attribution.ts` | **new** — whitelist, cookie parse/serialize + query-suffix helpers |
| `modules/auth/components/attribution-tracker.tsx` | **new** — client first-touch cookie writer |
| `modules/auth/layout.tsx` | mount `<AttributionTracker />` |
| `modules/auth/signup/actions.ts` | attach + clear attribution on `user_signed_up` |
| `modules/ee/sso/lib/sso-handlers.ts` | attach attribution on SSO `user_signed_up` |
| `modules/auth/lib/sign-in-tracking.ts` | read/clear in request scope; attach on `user_signed_in` |
| `modules/auth/login/components/login-form.tsx` | forward params on → signup link |
| `modules/auth/signup/components/signup-form.tsx` | forward params on → login link |
| `modules/auth/lib/attribution.test.ts` | **new** — unit tests for the helpers |

## Notable design decisions (from self-review)

- **Attribution never drops the core event.** The cookie is read in the request scope
  (not the detached `void captureSignIn` promise), and any failure degrades to no
  attribution — so a `cookies()` error can't silently drop `user_signed_in`.
- **No double-decode.** Next already percent-decodes cookie values; the helper parses
  the value as-is so values containing literal `%XX` sequences aren't corrupted.
- **Bounded staleness.** 30-minute lifetime + clear-on-consume prevents one user's
  attribution bleeding onto another user's events on a shared browser.
- **Oversized cookies skipped.** Values that would exceed the ~4KB browser limit
  (long multibyte UTMs) are not written rather than silently dropped.
- **Trusted props win.** Attribution is spread *before* server-computed props, so a
  crafted URL param can't override e.g. `signup_source`.

## Testing

- Unit tests for `pickAttributionParams`, `getAttributionPropertiesFromCookies`
  (incl. malformed JSON + percent-preservation), `buildAttributionQuerySuffix`.
- `pnpm typecheck` clean.

**Manual QA checklist:**
- [ ] `/auth/login?page=home&utm_source=twitter` → click "Sign up" → params persist in URL
- [ ] `fb_attribution` cookie set (`SameSite=Lax`, `Secure` on HTTPS)
- [ ] First-touch: second visit with different params does not overwrite
- [ ] `user_signed_up` (credentials + SSO) carries the props in PostHog
- [ ] `user_signed_in` (credentials + SSO) carries the props in PostHog
- [ ] Cookie cleared after signup/login


jobenjada:
@harshsbhat pls resolve merge conflicts & address code rabbit feedback

assigned @xernobyl to review afterwards since it touches auth code

apps/web/locales/de-DE.json
```diff
@@ -2299,8 +2299,8 @@
       "billing": {
         "billing_interval_toggle": "Abrechnungsintervall",
         "cancel_at_period_end": "Am Ende des Zeitraums kündigen",
-        "confirm_upgrade_body": "Du bist dabei, auf den {plan}-Plan für {amount} {period} zu upgraden. Eine anteilige Gebühr für den Rest deines aktuellen Abrechnungszeitraums wird sofort berechnet, und alle anfallenden Steuern werden bei der Zahlung berechnet.",
-        "confirm_upgrade_body_with_charge": "Du bist dabei, auf den {plan}-Tarif ({period}) zu upgraden. Dir werden jetzt {chargeNow} für den Rest deines aktuellen Abrechnungszeitraums berechnet, wobei alle anfallenden Steuern bei der Zahlung kalkuliert werden.",
```
jobenjada: @harshsbhat seems like it has unrelated code changes, pls always check on Github after pushing to removte.

Moved to draft until this and other comments are resolved

jobenjada (dismissed):

xernobyl (approved):
Reviewed with a focus on the auth paths. LGTM.

Reconciled the existing bot review against the current head:
- The two CodeRabbit "clear the cookie" findings are addressed — the sign-in helper and signup action both delete `fb_attribution` whenever the cookie is present (independent of parse success), and the SSO signup path clears it in the same request via `finalizeSuccessfulSignIn` in the session after-hook.
- The remaining CodeRabbit note (write attribution before an SSO button can redirect) is a genuine but low-probability edge case — the `useEffect` runs within a frame of mount; fine to leave.

Verified the load-bearing assumption: Next 16 decodes cookie values once (`parseCookie` → `decodeURIComponent`), so reading the cookie as plain JSON without a second decode is correct and attribution lands on the events end-to-end.

Design is careful — trusted props win over URL-supplied ones (spread order), fixed whitelist, values trimmed/capped, TTL + clear-on-consume backstops.

Nit (non-blocking): the PR description's "Captured params" list omits `entry_page`, which the whitelist actually includes — worth syncing the description.

PR #8399, opened 2026-07-01
fix: canonicalize v3 survey language read/write + simplify resolver (ENG-1067)

pandeymangg:
## What does this PR do?

Part of the **ENG-1067** language-code canonicalization epic (targets the `epic/language-codes-stabilization` branch). It makes the **v3 survey APIs** consistently canonical for language codes, fixes a migration-triggered read-contract break, and simplifies the v3 language module.

### Read — `GET /api/v3/surveys/{id}?lang=…`

The `?lang=` resolver is now **canonical-aware**: a legacy / script / region selector resolves to the matching migrated survey language by canonical equivalence. For example, against a survey whose language was migrated to `zh-Hans-CN`, `?lang=zh-Hans` and `?lang=zh-CN` now resolve to it instead of returning **HTTP 400**. An exact stored-code match still always wins, so there is no broadened ambiguity. (The write path already did this — the read resolver was the asymmetric miss that the data migration would otherwise have turned into a 400.)

### Write — create / patch

- **Create** canonicalizes a new language on write, so newly stored codes are always canonical (`zh-CN` → `zh-Hans-CN`) while legitimate non-default regions are preserved (`de-AT` stays `de-AT`, `en-GB` stays `en-GB`).
- **Patch** matches the request against the survey's existing languages and writes the **stored code as-is** — never rewriting it, even when the client sends the canonical/script-complete form — so a survey's existing content i18n keys are never orphaned.

### Quota evaluation

The v1/management response-create path now passes the **canonical** language to quota evaluation (it was passing the raw request code). This matches every other response path and keeps language-scoped quotas consistent with the value actually persisted on the response.

### Simplification / refactor

Removed the bespoke v3 curated legacy map (`V3_LEGACY_LANGUAGE_CODE_MAP` / `V3_CANONICALIZABLE_BARE_CODES`) and `normalizeV3SurveyLanguageIdentifier`, unifying all matching and normalization on the shared `normalizeLanguageCode`. Read/resolve now return the survey's **stored code as-is** (already canonical post-migration); canonicalization happens only on write and for input matching. Net **−29 lines** with behavior preserved for all post-migration data (verified OLD-vs-NEW across the resolver, write, serializer and generate paths).

## How should this be tested?

- `pnpm --filter=@formbricks/web test app/api/v3/surveys` — **275 tests pass** (resolver, write-validation, serializers, patch-prepare, AI-generate). Typecheck and lint are clean.
- **Read back-compat:** `GET /api/v3/surveys/{id}?lang=zh-Hans` (also `zh-CN`, `de`, `hi`) against a migrated multi-language survey → resolves to the canonical survey language and returns its content (previously **400** for the script/region Chinese selectors).
- **Write:** `POST`/`PATCH` a survey language `zh-CN` → stored as `zh-Hans-CN`; patching a survey that already stores a code preserves that stored code (no orphaned content keys).
- **Quota:** submit a response via the v1/management responses API with a legacy language code → language-scoped quotas evaluate against the canonical code.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/app/api/v1/management/responses/lib/response.ts
+++ b/apps/web/app/api/v1/management/responses/lib/response.ts
@@ -11,7 +11,7 @@ import { TTag } from "@formbricks/types/tags";
 import { buildPrismaResponseData } from "@/app/api/v1/lib/utils";
 import { RESPONSES_PER_PAGE } from "@/lib/constants";
 import { getResponseContact } from "@/lib/response/service";
-import { calculateTtcTotal, normalizeResponseLanguage } from "@/lib/response/utils";
+import { calculateTtcTotal } from "@/lib/response/utils";
 import { getSurvey } from "@/lib/survey/service";
 import { getOrganizationIdFromWorkspaceId } from "@/lib/utils/helper";
 import { validateInputs } from "@/lib/utils/validate";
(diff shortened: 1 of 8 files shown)


xernobyl (changes requested):
Address the coderabbit commetns. Other than that looks good :)

xernobyl (approved):
Thank you for the changes. LGTM.

PR #8402, opened 2026-07-01
ci(auth): skip Better Auth integration suite when the harness is absent

xernobyl:
## What does this PR do?

Stops the **Better Auth integration-test** workflow from red-failing on branches that predate the Better Auth cutover (#8384).

Since #8384 landed `integration-tests.yml` on `main`, GitHub resolves the workflow definition from the base branch, so the job now triggers for *every* PR whose paths match — including branches that forked before the cutover. `dangerous-git-checkout` checks out the PR **head**, which on those branches contains neither the workflow nor `apps/web/integration/ba-test-schema-shape.sql`, so the "Shape the schema" step fails with:

```
psql: error: apps/web/integration/ba-test-schema-shape.sql: No such file or directory
```

Seen e.g. on `epic/language-codes-stabilization`. It's noise (the check isn't required, so it doesn't block merges), but it nags every stale PR that touches the path filter.

**Fix:** a `Detect Better Auth integration harness` step checks whether the harness is present on the checked-out ref, and the harness-dependent steps are gated on `steps.harness.outputs.present == 'true'`. Branches without the harness now **skip cleanly (green)** instead of hard-failing on the missing file — and don't burn ~5 min of setup either. Branches that have the harness run the suite unchanged.

Workflow-only change; no product code touched.

## How should this be tested?

- This PR's own `Integration Tests` run (this branch **has** the harness) must run the full suite and pass — proving the guard doesn't skip when the files are present.
- After merge, re-run CI on a pre-cutover branch (e.g. `epic/language-codes-stabilization`): the job should now **skip** the harness steps and go green, with the `::notice::` explaining why.

## Checklist
- [x] Filled out the "How to test" section
- [x] Self-reviewed

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -72,25 +72,46 @@ jobs:
           persist-credentials: false # no later step needs git auth (dangerous-git-checkout re-auths)
       - uses: ./.github/actions/dangerous-git-checkout
 
+      # Branches that predate the Better Auth cutover (#8384) don't contain the integration harness,
+      # yet this workflow — now on main — still triggers for their PRs (the definition is resolved from
+      # the base branch) and dangerous-git-checkout checks out the PR *head*. Detect whether the harness
+      # is actually present on the checked-out ref so the steps below can skip instead of hard-failing on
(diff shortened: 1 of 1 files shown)


pandeymangg (approved):
LGTM! 🚀

PR #8403, opened 2026-07-01
feat(workflows): enable workflow-run retries and reconcile orphaned runs [ENG-1228][ENG-1608]

xernobyl:
## What does this PR do?

Closes [ENG-1608](https://linear.app/formbricks/issue/ENG-1608) and enables the workflow-run retry policy paired with [ENG-1228](https://linear.app/formbricks/issue/ENG-1228). **Stacked on `epic/workflows-v1`** (base), like the other ENG-12xx workflow tickets. This combines the two workflow-run resilience changes that were briefly split as #8403 + #8404 (the latter is folded in here and closed) — they touch the same `packages/jobs/queue` surface and only make sense together.

Two independent recovery paths for the runner:

### 1. Enable BullMQ retries for `workflow-run.process` (ENG-1228)

Drops the `attempts: 1` override on `enqueueWorkflowRunJob` so the job inherits the shared retry policy — `attempts: 3` + exponential backoff — from the queue's `defaultJobOptions`, same as every other job.

With `attempts: 1`, `maxAttempts` was always 1, so the executor's retry/resume machinery and pool-exhaustion handling never actually fired (every attempt was "final") — a transient failure (SMTP blip, DB pool exhaustion, stalled redelivery) killed the run on the first try. Now that `send_email` execution is idempotent per step (claim-before-send + `@@unique([runId, stepId])`, #8395), a BullMQ retry resumes without re-sending. The deterministic `jobId` (the run id) keeps a re-enqueue idempotent — including the reconciler below re-dispatching an orphan.

> ✅ **Merge ordering (satisfied):** #8395's per-step send idempotency has landed in `epic/workflows-v1` (this branch is merged up to include it), so enabling retries here is safe — a retry resumes without re-sending. (The ordering mattered because retries must never be enabled without that idempotency, or a retry could double-send an email.)

### 2. Reconcile orphaned `queued` runs (ENG-1608)

Adds a recurring `workflow-run.reconcile` background job that recovers **producer-side orphans**. The ENG-1226 producer persists a `queued` `WorkflowRun` and *then* dispatches its BullMQ job — deliberately non-transactional, with both halves idempotent. If the process dies or the Redis enqueue fails after the row is committed, the run is stranded `queued` with no backing job and the workflow silently never fires. (`createAndDispatchWorkflowRun` already logs exactly this: *"…queued run is orphaned until the reconciler re-dispatches it."*)

The reconciler sweeps every 3 min for `queued` runs older than a 2-min grace window and re-dispatches each through the **existing** dispatch port (`dispatchWorkflowRunViaJobs` → `enqueueWorkflowRunJob(input, { jobId: run.id })`). The deterministic `jobId` makes re-dispatch a no-op if a job still exists, and a genuine recovery if it was lost.

Design notes:
- **Mirrors the `survey-scheduling.reconcile` template** — same `packages/jobs` plumbing (job name, Zod data, throw-only processor stub, definition, upsert/remove recurring helpers, producer method) and the same `instrumentation-jobs` wiring + worker/queue gating.
- **No in-tick re-query loop.** Survey-scheduling can loop because it mutates status as it goes; re-dispatch does *not* change a run's status synchronously (the worker does, later), so re-querying would re-select the same rows. Instead: one bounded batch (`take: 250`) per tick, oldest-first, index-backed by `@@index([status, createdAt])`; a larger backlog drains over subsequent ticks (harmless — re-dispatch is idempotent).
- **Scoped to real runs** (`isDryRun: false`); `running` / `completed` / `failed` / `canceled` are never touched — execution-failure retry is the executor's concern (now handled by path 1), not the reconciler's.
- **Bounded**: a run still `queued` past a 24h ceiling is marked `failed` via a status-guarded `updateMany` (`where: { id, status: "queued" }`, so a concurrent `queued → running` claim is never clobbered), surfacing the stuck run instead of re-dispatching forever.
- **Recovery boundary** (documented in the reconciler docstring): the deterministic `jobId` makes re-dispatch a no-op while a job of that id still exists in *any* state — including a terminally `failed` one (kept by `removeOnFail` for 7d, longer than the 24h ceiling). So re-dispatch genuinely recovers only runs whose job was never created (or has since been evicted); a run dispatched but whose job failed while the row stayed `queued` is cleaned up by the ceiling instead.
- The orchestration takes an injected `dispatch` + clock, so it's unit-tested without Redis/BullMQ.

**Scope boundary:** `queued` orphans only. A run stuck **`running`** after a mid-execution crash (step row left `running`, so every redelivery bails cleanly and the run never completes) is a distinct failure mode and is **out of scope** here — it needs a `running`-sweep that marks the run `failed` and orphaned `running` step rows `skipped` (never `failed`, which would re-arm a re-send under the at-most-once guarantee). Tracked as a follow-up to extend this reconciler.

## How should this be tested?

Unit-tested — no Redis needed, the orchestration injects a fake dispatch + fixed clock:

```bash
pnpm --filter @formbricks/jobs run test                       # 62 pass
cd apps/web && pnpm exec vitest run \
  modules/workflows/lib/runner instrumentation-jobs.test.ts   # 41 pass
```

- `queue.test.ts` asserts `enqueueWorkflowRunJob` no longer overrides `attempts` (so the job inherits the shared `attempts: 3` + backoff policy), plus the new recurring reconcile upsert/remove/producer helpers.
- `reconcile-orphaned-runs.test.ts` covers: re-dispatches queued orphans past the grace window; query scoping (`status: "queued"`, `isDryRun: false`, `createdAt < now − grace`, oldest-first, bounded); the 24h ceiling → `failed` (and the status-guard race → not counted); a per-run dispatch failure isolated (sweep continues); empty sweep.
- `process-workflow-run-reconcile-job.test.ts` + `instrumentation-jobs.test.ts` assert the handler wiring, the log-only-when-acted behaviour, and the recurring schedule registration.

Manual (optional, needs Redis + a worker): enable a workflow on a survey, complete a response, then before the worker picks it up flip its `WorkflowRun` back to `queued` (or block the enqueue) → within ~3–5 min the reconciler re-dispatches it and it runs. For retries: make a `send_email` step throw a transient error once → the job retries and resumes from the claimed-step log instead of failing permanently.

Verified locally: `@formbricks/jobs` 62 tests + apps/web runner/instrumentation 41 tests pass, `@formbricks/web` typecheck 0 errors, ESLint clean on touched files.

### Required

- [x] Commented on my code in hard-to-understand bits (the no-in-tick-loop rationale, the status guard, the retry-idempotency pairing, the recovery boundary)
- [x] Ran the build (jobs build + affected typecheck)
- [x] Branched off the current base (`epic/workflows-v1`)
- [x] My changes don't cause any responsiveness issues (backend job, no UI)

### Appreciated

- [ ] If a UI change was made: screenshots — N/A, backend job
- [ ] Updated the Formbricks Docs — N/A (internal; the self-host worker docs are tracked separately under the epic)


packages/jobs/src/processors/workflow-run-reconcile.ts
```diff
@@ -0,0 +1,20 @@
...
+      queueName: context.queueName,
+      scope: data.scope,
+    },
+    "BullMQ workflow run reconcile processor override is not registered"
+  );
+
+  throw new Error(
+    `BullMQ workflow run reconcile processor override missing for job ${context.jobId} (${data.scope})`
+  );
+};
```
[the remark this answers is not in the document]
xernobyl: Verified — this is a build-order artifact, not a real issue. `logger.error` only comes back as an `any`/`error` type because the analysis typechecked before `@formbricks/logger` was built; every processor stub in this package (`survey-scheduling.ts`, `response-pipeline.ts`, …) makes the identical `logger.error({ … }, "…")` call. In CI — which builds workspace deps before typecheck/lint — the declarations resolve: the jobs typecheck + lint checks are green here and SonarCloud's gate passed. Nothing to change in this handler. Resolving.

apps/web/modules/workflows/lib/runner/process-workflow-run-reconcile-job.ts
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  const result = await reconcileOrphanedWorkflowRuns({
+    dispatch: dispatchWorkflowRunViaJobs,
+    now: new Date(),
+    logContext,
+  });
+
+  if (result.redispatched > 0 || result.agedOutFailed > 0) {
+    logger.info({ ...logContext, ...result }, "Workflow run reconciliation re-dispatched orphaned runs");
```
itsjavi: nit: when the sweep only hits the ceiling (`redispatched: 0, agedOutFailed > 0`) this still logs "re-dispatched orphaned runs", which is precisely the case where nothing was re-dispatched. There is the per-run warn, but anyone grepping for the ceiling path will land on a misleading summary line. Something neutral like "reconciliation acted on orphaned runs" (or branching the message) would fix it.

apps/web/modules/workflows/lib/runner/reconcile-orphaned-runs.ts
```diff
@@ -0,0 +1,116 @@
...
+ * `jobId = run.id`): a no-op if a job still exists, a genuine recovery if it was lost.
+ *
+ * - **Scope:** only real runs (`isDryRun: false`); `running` / `completed` / `failed` / `canceled` are
+ *   never touched (execution-failure retry is the executor's concern, not the reconciler's).
+ * - **Ceiling:** a run still `queued` past `WORKFLOW_RUN_ORPHAN_MAX_AGE_MS` is treated as permanently
+ *   un-dispatchable and marked `failed` (bounds the loop; surfaces the stuck run) instead of being
+ *   re-dispatched forever.
+ * - **Re-dispatch vs. ceiling:** the deterministic `jobId` makes re-dispatch a no-op while a job of
+ *   that id still exists in *any* state — including a terminally `failed` one (kept by `removeOnFail`
+ *   for 7d, longer than the 24h ceiling). So re-dispatch genuinely recovers only runs whose job was
```
itsjavi: Tiny precision issue in this note: `removeOnFail` is age 7d but also `count: 5000` (packages/jobs/src/constants.ts), so under a failure backlog a failed job can get count-evicted well inside the 24h window, and then re-dispatch actually re-runs it instead of leaving it to the ceiling. That is safe (step idempotency plus the terminal-status guard cover it; I checked the re-add-on-failed-job semantics against a real Redis and it no-ops while the job is retained), but since this docstring is the load-bearing explanation of the boundary it should probably not read as absolute. nit, doc-only.

itsjavi (commented):
Pulled the branch and ran this locally before reviewing. Lint, typecheck and the unit suites are green (`@formbricks/jobs` 62 pass; the apps/web runner + instrumentation tests are at 78 now after the epic merge, not the 41 the description mentions, all passing). I also verified the behaviour for real against local Redis + Postgres: the job now inherits `attempts: 3` + exponential backoff from the shared policy, re-enqueueing the same jobId while a job exists is a no-op (including when the job is sitting in `failed`, which confirms the recovery-boundary note), and a live reconciler sweep with seeded rows did exactly what it should: re-dispatched a 10-min orphan, force-failed a 25h one with error + finishedAt, and left fresh / dry-run / `running` rows alone.

One heads-up, not caused by this PR: with `REDIS_URL` set, 2 tests in `jobs-integration.test.ts` fail against Valkey (`waitUntilFinished` resolves `null` vs `undefined`). Reproduced identically on `epic/workflows-v1`, so pre-existing.

No blockers from my side, just two nits inline (a log message and a docstring precision thing).

apps/web/modules/workflows/lib/runner/process-workflow-run-reconcile-job.ts
xernobyl: Good catch — fixed in bb87699. The summary line is now neutral ("…acted on orphaned runs") so a ceiling-only sweep no longer reads as a re-dispatch; the `redispatched`/`agedOutFailed` counts in the payload say which path ran. Kept it a single info line since the per-run `warn` already flags each aged-out run.

apps/web/modules/workflows/lib/runner/reconcile-orphaned-runs.ts
xernobyl: You're right, thanks — the 7d framing ignored the `count: 5000` bound. Reworded the docstring in bb87699: retention is age 7d *and* count 5000, so under a failure backlog a failed job can be count-evicted inside the 24h ceiling and a later re-dispatch re-runs it — safe via per-step idempotency + the terminal-status guard — so it now reads "job gone → recover, job retained → no-op, ceiling as backstop" instead of an absolute 7d-vs-24h rule.

itsjavi (commented):
Took a second, deeper pass, this time on whether the reconciler is the right design or just a patch, plus re-verified the new head.

**Re-verification on bb87699:** both wording fixes look good, thanks for the quick turnaround. `@formbricks/jobs` 62 pass, runner + instrumentation 78 pass, apps/web typecheck and lint clean. The delta is comment/log-string only so I didn't redo the live Redis/Postgres run; everything I verified live on the previous head applies byte-for-byte.

**On the architecture:** I went in expecting to argue for a transactional outbox and came out convinced this is the right shape for this codebase.

- The dual-write isn't an accident you're papering over; it's forced by the call site. `enqueueResponseCompletedWorkflowRuns` runs inside the response-pipeline job between other non-checkpointed side effects, and a dispatch failure is deliberately swallowed because rethrowing would replay the siblings (follow-ups, notification emails, webhooks). So the producer can't lean on its parent job's retries and needs its own durable record plus a recovery loop.
- The `WorkflowRun` row already is a domain outbox in all but name: idempotency unique constraint, `@@index([status, createdAt])`, "row is the source of truth" in the dispatch docstring. This PR just supplies the missing relay leg. The textbook alternatives all lose here: enqueue-first leaves no durable record when the crash goes the other way (the executor drops on a missing row, so the workflow silently never fires and there's nothing to reconcile from); Prisma has no afterCommit and hand-rolling one still doesn't survive process death between commit and hook; queue-as-truth contradicts the portability goal and finite Redis retention. A generic outbox table would be speculative when exactly one flow crosses this boundary today.
- Retry enablement: claim-before-send resolves the crash window the safe way (a crash between SMTP handoff and the succeeded-commit leaves a `running` row, every redelivery bails on it, so the mail is potentially lost but never duplicated). That's real at-most-once for the crash case, which is what makes dropping `attempts: 1` sound. The residual window (ambiguous SMTP failure recorded as `failed`, re-sent on retry) exists in any claim-based scheme and the deterministic Message-ID already mitigates it.

**The one thing I'd push on** (should-fix at the epic level, not in this PR): the deferred `running`-sweep isn't symmetrical polish. A run stuck `running` after a mid-execution crash is never recovered by BullMQ stalled redelivery either; on redelivery the claim succeeds, the resume path bails on the orphaned `running` step row, and the job completes without finalizing, so the run is stuck forever with nothing watching it. Per crash that's arguably the more likely stranding than the ms-wide commit-to-enqueue gap this PR covers. The follow-up deserves real priority rather than a backlog entry, and its semantics differ in the way you already noted (orphaned `running` steps must become `skipped`, never `failed`, or the sweep re-arms a re-send). Until it lands, the run UI can accumulate permanently-`running` runs that nothing surfaces.

Two non-blocking debt notes for the record: there's no dispatched-at marker on the run row, so recovery correctness leans on BullMQ's jobId-dedup semantics plus the tuned grace window rather than on a DB fact (fine now; revisit if the queue backend ever changes), and the per-recurring-job boilerplate in `packages/jobs` (8 files touched here for one concept) will deserve a generic helper around the third recurring job.

Verdict: right problem, right solution at this scale. No blockers from my side; I'd just like the `running`-sweep follow-up actually scheduled before the epic ships.

xernobyl:
Thanks for the deep pass, Javi — appreciate you stress-testing the outbox-vs-this-shape question rather than just the diff. 🙏

On the one thing you'd push on — the `running`-sweep — good news, it's already out of the backlog: filed as [ENG-1656](https://linear.app/formbricks/issue/ENG-1656) and implemented in #8433, stacked right on this branch (Dhru's reviewing it now), so it lands with this stack ahead of the epic. It does exactly what you flagged — force-fails the stale `running` run (status-guarded so a live owner still wins) and flips its orphaned `running` step rows to `skipped`, never `failed`, so it can't re-arm a re-send. Staleness keys off `run.updatedAt` (the executor bumps it on non-final failures but not between steps, so it's the real "last activity" signal), 1h threshold.

Both debt notes are fair and worth capturing as follow-ups:
- **No `dispatchedAt` marker** — agreed, recovery leans on BullMQ's jobId-dedup + the grace window rather than a DB fact. Fine while we're on BullMQ; the thing to revisit if we ever swap the backend.
- **Recurring-job boilerplate** — also agreed, a generic helper is the right call around the third recurring job (we're at two: survey-scheduling + workflow-run.reconcile).

I'll get both tracked. Thanks again!

itsjavi (approved):

PR #8404, opened 2026-07-01
fix(workflows): enable BullMQ retries for workflow-run.process [ENG-1228]

xernobyl:
## What does this PR do?

Pairs with #8395 ([ENG-1228](https://linear.app/formbricks/issue/ENG-1228)), stacked on `epic/workflows-v1`. Now that `send_email` execution is idempotent per step (claim-before-send + `@@unique([runId, stepId])`, in #8395), this drops the `attempts: 1` override on `enqueueWorkflowRunJob` so `workflow-run.process` inherits the shared retry policy — `attempts: 3` + exponential backoff from the queue's `defaultJobOptions`, same as every other job.

**Why:** with `attempts: 1`, `maxAttempts` was always 1, so the executor's retry/resume machinery and its pool-exhaustion handling never actually fired (every attempt was "final") — a transient failure (SMTP blip, DB pool exhaustion, stalled redelivery) killed the run on the first try and could strand it. With retries enabled *and* the per-step idempotency from #8395, a transient failure now retries and resumes without re-sending. The deterministic `jobId` (the run id) keeps a re-enqueue idempotent (e.g. the ENG-1608 reconciler re-dispatching an orphan).

**⚠️ Merge ordering:** must merge **with or after #8395** — retries must never be enabled without the per-step send idempotency, or a retry could double-send. (#8395 is approved.)

## How should this be tested?

```bash
pnpm --filter @formbricks/jobs run test   # 59 pass
```

One-line behaviour change (drop the `attempts: 1` override). The `workflow-run` enqueue test now asserts **no** per-job `attempts` override, so the job inherits the shared `attempts: 3` + backoff policy — which itself is covered by the existing "creates the shared queue with the expected defaults" test. End-to-end (worker running): a `send_email` step that throws a transient error now retries up to 3× and resumes from the claimed-step log, instead of failing permanently on attempt 1.

### Required

- [x] Commented on my code in hard-to-understand bits (the retry-policy inheritance + the #8395 pairing)
- [x] Ran the jobs test + typecheck + lint
- [x] Removed all `console.logs` (none)
- [x] Branched off the current base (`epic/workflows-v1`)
- [x] My changes don't cause any responsiveness issues (backend job)

### Appreciated

- [ ] If a UI change was made: screenshots — N/A
- [ ] Updated the Formbricks Docs — N/A

### DIFF
--- a/packages/jobs/src/queue.test.ts
+++ b/packages/jobs/src/queue.test.ts
@@ -181,15 +181,16 @@ describe("@formbricks/jobs queue helpers", () => {
     expect(mockQueueAdd).toHaveBeenCalledWith(JOB_NAMES.surveyScheduling, surveySchedulingJobData, undefined);
   });
 
-  test("enqueues the workflow run job with attempts:1 and a deterministic jobId", async () => {
+  test("enqueues the workflow run job with a deterministic jobId and the shared retry policy", async () => {
     const mockJob = { id: "job-workflow-run-1" };
     mockQueueAdd.mockResolvedValue(mockJob);
 
     const job = await enqueueWorkflowRunJob(workflowRunJobData, { jobId: workflowRunJobData.workflowRunId });
 
     expect(job).toBe(mockJob);
+    // No per-job attempts override: the job inherits attempts + backoff from the queue's
+    // defaultJobOptions (retries are safe now that execution is idempotent per step — ENG-1228).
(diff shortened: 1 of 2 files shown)


xernobyl:
Folding this into #8403 — the retry-enablement and the orphan reconciler touch the same `packages/jobs/queue` surface and only make sense together, so they're now a single PR. The exact commit from here (`6677dbd`, drop the `attempts: 1` override) is cherry-picked onto #8403, which now carries both changes. Closing in favor of #8403.

PR #8407, opened 2026-07-02
fix: resolves feedback on the epic for language codes migration

pandeymangg:
## What does this PR do?

Addresses @xernobyl's re-review feedback on the ENG-1067 language-codes epic (#8390). Targets the `epic/language-codes-stabilization` branch. Three items:

### [P1 — gating] Migration transaction / lock model

Previously the whole data migration ran in one ~30-minute transaction, and the `Response` step did a single big `UPDATE` against the unindexed `Response.language` column. On a large `Response` table that risked (a) exceeding the timeout → the **entire** transaction rolls back → the deploy never converges, and (b) holding row locks long enough to stall response ingestion.

The two high-volume steps now run **outside** the mega-transaction, via the runner's autocommit `prisma` handle, in **committed-per-chunk** batches:

- **`Response`** (`language` + the `contactAttributes` `language` snapshot) is walked by **primary key** in fixed chunks (`RESPONSE_BATCH_SIZE`), each chunk committed. The remap set comes from the shared canonical map instead of a `SELECT DISTINCT` full-table scan.
- **`ContactAttribute`** updates each legacy value in **index-backed** (`[attributeKeyId, value]`) chunks that commit per batch, so a code held by millions of contacts never sits under one long lock.

Locks release per chunk (no ingestion stall), and committed progress survives a timeout — the migration is idempotent, so a retry simply resumes. The small, bounded steps (Language-row merges, survey content) stay inside the transaction.

*Trade-offs (intentional):* whole-migration atomicity is replaced by commit-as-you-go + idempotent resume; a code outside the canonical map is left as-is and still resolves at read time (the renderer canonicalizes on read).

### [P3] v3 write-validation regex rejected canonical tags

`V3_SURVEY_LOCALE_CODE_REGEX` only allowed a 2-letter base + 2-letter region, so it rejected canonical catalog codes with a 3-letter base or a UN M.49 numeric region (`fil-PH`, `bho-IN`, `eo-001`, `ia-001`, `io-001`, `vo-001`) — codes the picker and `createLanguage` accept — meaning *creating* such a survey language via the v3 API failed. Widened the pattern to accept them; verified it now matches **all 204** canonical catalog codes (was rejecting 6). Reads/existing languages are unaffected.

### [P3] Unit tests for the migration planning logic

Added `utils.test.ts` — **20 characterization tests** for the trickiest, most data-mangling-prone pure functions (`toCanonical`, `rewriteI18nKeys`, `planLanguageMerges`, `planSurveyLanguageMoves`): collision survivorship, `SurveyLanguage` composite-PK dedup + flag promotion, order-preserving block/ending rewrites, `default` sentinel preservation, and unresolved-code handling.

_(P2 — single-language surveys skipped in the content step — was intentionally not changed: the renderer canonicalizes on read so display is unaffected; the candidate set can be widened later if prod data shows stale legacy keys.)_

## How should this be tested?

- `pnpm --filter=@formbricks/web test app/api/v3/surveys/language.test.ts` → **45 pass** (incl. new `fil-PH` / `bho-IN` / `eo-001` write-locale cases).
- `cd packages/database && npx vitest run migration/20260625104904_canonicalize_language_codes/utils.test.ts` → **20 pass**.
- `packages/database` typecheck + eslint clean; `apps/web` v3 lint clean.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/app/api/v3/surveys/language.ts
+++ b/apps/web/app/api/v3/surveys/language.ts
@@ -23,7 +23,11 @@ type TResolveV3SurveyLanguageCodeResult =
 type TParseV3SurveyLanguageQueryResult = { ok: true; languages: string[] } | { ok: false; message: string };
 
 const V3_SURVEY_LANGUAGE_TAG_REGEX = /^[a-z]{2}(?:-[A-Z]{2}|-[A-Z][a-z]{3}(?:-[A-Z]{2})?)$/;
-const V3_SURVEY_LOCALE_CODE_REGEX = /^[a-z]{2}(?:-[A-Z][a-z]{3})?-[A-Z]{2}$/;
+// A region-qualified write locale: 2-3 letter base language, optional 4-letter script, and a region that
+// is either a 2-letter country or a 3-digit UN M.49 area. The 3-letter base + numeric region forms matter
+// because canonical catalog codes include them (e.g. `fil-PH`, `bho-IN`, `eo-001`, `vo-001`); the old
+// 2-letter-only pattern rejected those on write even though the picker/`createLanguage` accept them.
+const V3_SURVEY_LOCALE_CODE_REGEX = /^[a-z]{2,3}(?:-[A-Z][a-z]{3})?-(?:[A-Z]{2}|\d{3})$/;
 
 // Validate + case/separator-normalize a fully region/script-qualified BCP-47 tag. This only recognises a
 // valid tag — it does NOT canonicalize or complete it (`zh-Hans` stays `zh-Hans`, `zh-CN` stays `zh-CN`).
--- a/apps/web/app/api/v3/surveys/language.ts
+++ b/apps/web/app/api/v3/surveys/language.ts
@@ -22,37 +22,39 @@ type TResolveV3SurveyLanguageCodeResult =
 
 type TParseV3SurveyLanguageQueryResult = { ok: true; languages: string[] } | { ok: false; message: string };
 
-// A valid BCP-47 tag: 2-3 letter base, then either a region (2-letter country or 3-digit UN M.49 area) or a
-// 4-letter script optionally followed by a region. Widened alongside the write locale regex below so the
-// six 3-letter-base / numeric-region catalog codes (e.g. `fil-PH`, `eo-001`) are recognised on READ too —
-// otherwise the resolver mis-reports "not a valid locale code" (invalid) instead of "not configured"
-// (unknown) when such a code is queried on a survey that doesn't have it.
-const V3_SURVEY_LANGUAGE_TAG_REGEX =
-  /^[a-z]{2,3}(?:-(?:[A-Z]{2}|\d{3})|-[A-Z][a-z]{3}(?:-(?:[A-Z]{2}|\d{3}))?)$/;
-// A region-qualified write locale: 2-3 letter base language, optional 4-letter script, and a region that is
+// A region-qualified locale: 2-3 letter base language, optional 4-letter script, and a region that is
 // either a 2-letter country or a 3-digit UN M.49 area. The 3-letter base + numeric region forms matter
 // because canonical catalog codes include them (e.g. `fil-PH`, `bho-IN`, `eo-001`, `vo-001`); the old
 // 2-letter-only pattern rejected those on write even though the picker/`createLanguage` accept them.
 const V3_SURVEY_LOCALE_CODE_REGEX = /^[a-z]{2,3}(?:-[A-Z][a-z]{3})?-(?:[A-Z]{2}|\d{3})$/;
+// A script-only tag with no region (e.g. `zh-Hans`). Kept as its own small regex so the tag check below is
+// just "region-qualified OR script-only" — two simple patterns instead of one nested mega-regex.
+const V3_SURVEY_SCRIPT_ONLY_REGEX = /^[a-z]{2,3}-[A-Z][a-z]{3}$/;
 
-// Validate + case/separator-normalize a fully region/script-qualified BCP-47 tag. This only recognises a
-// valid tag — it does NOT canonicalize or complete it (`zh-Hans` stays `zh-Hans`, `zh-CN` stays `zh-CN`).
-// Canonicalization is `normalizeLanguageCode`'s job; this is used for tag recognition and query dedup.
-// Returns null for bare/underspecified/invalid input.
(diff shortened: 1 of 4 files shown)


apps/web/app/api/v3/surveys/language.ts
```diff
@@ -23,7 +23,11 @@ type TResolveV3SurveyLanguageCodeResult =
 type TParseV3SurveyLanguageQueryResult = { ok: true; languages: string[] } | { ok: false; message: string };
 
 const V3_SURVEY_LANGUAGE_TAG_REGEX = /^[a-z]{2}(?:-[A-Z]{2}|-[A-Z][a-z]{3}(?:-[A-Z]{2})?)$/;
```
xernobyl: [P3] The read-side tag regex wasn't widened along with the write one, so the six newly-writable codes report a misleading error when they're *not* configured on a survey:

- `?lang=fil-PH` → `invalid` / "Language 'fil-PH' is not a valid locale code or configured language alias"
- `?lang=pt-PT` (for comparison) → `unknown` / "Language 'pt-PT' is not configured for this survey" + `normalizedCode`

Configured `fil-PH` surveys resolve fine on read (resolver step 3 falls back to `normalizeLanguageCode` with the raw value — I verified `FIL_ph` → `fil-PH` works), so this is error-classification only. But it's a bit odd that you can now create `fil-PH` via the API and then be told it "is not a valid locale code" when querying a survey that doesn't have it.

Same family as the write fix:

```ts
const V3_SURVEY_LANGUAGE_TAG_REGEX = /^[a-z]{2,3}(?:-(?:[A-Z]{2}|\d{3})|-[A-Z][a-z]{3}(?:-(?:[A-Z]{2}|\d{3}))?)$/;
```

plus a couple of test rows for the unknown-vs-invalid split. Could fold in CodeRabbit's negative-case suggestion while at it.

packages/database/migration/20260625104904_canonicalize_language_codes/migration.ts
```diff
@@ -220,76 +228,87 @@ export const canonicalizeLanguageCodes: MigrationScript = {
...
-    const responseCodes = await tx.$queryRaw<{ language: string }[]>`
-      SELECT DISTINCT language FROM "Response"
-      WHERE language IS NOT NULL AND language <> 'default'
-    `;
-    const responsePairs = buildCanonicalPairs(
-      responseCodes.map((r) => r.language),
-      stats
-    );
+    logger.info("Canonicalizing Response.language + contactAttributes snapshot...");
+    const responsePairs = buildRemapPairsFromMap();
```
xernobyl, line 248: [P3] Trade-off worth making explicit: matching `r.language` by exact string against the map keys means case/underscore variants (`EN`, `de_DE`, `pt-br`) no longer remap — the old `SELECT DISTINCT` → `toCanonical` path caught those, and step 4 still does for ContactAttribute (it discovers observed values). So post-migration, `Response.language` can keep variant forms of *known* codes, not just exotic unknown ones.

I don't think it's user-visible today (responses canonicalize on write since ENG-1067, and summary matching is read-tolerant), so fine to accept as-is — but if we want to close it cheaply:

- match on `lower(replace(r.language, '_', '-')) = data.old` with lowercased map keys (they're unique lowercase — `LANGUAGE_CANONICAL_MAP_BY_LOWERCASE` relies on that), and
- add `AND r.language <> data.canon` so already-canonical rows don't get rewritten as dead tuples.

Either way, maybe add a line to the step-3 comment noting variants are intentionally left to read-time normalization.

packages/database/migration/20260625104904_canonicalize_language_codes/migration.ts
```diff
@@ -21,7 +26,10 @@ export const canonicalizeLanguageCodes: MigrationScript = {
   type: "data",
   id: "wu5owuszcu0ge717a5vodm0p",
   name: "20260625104904_canonicalize_language_codes",
-  run: async ({ tx }) => {
+  // `tx` is the runner's mega-transaction (used for the small, bounded catalog + survey-content steps).
+  // `prisma` is the autocommit handle used for the two high-volume tables so their work commits in small
+  // chunks — see steps 3 & 4 for why (lock footprint + timeout/convergence).
+  run: async ({ prisma, tx }) => {
```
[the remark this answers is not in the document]
xernobyl: Verified this against the runner — the transaction-scope concern is real: `run()` executes inside `prisma.$transaction(..., 30min)`, so steps 3–4 burn the same budget and steps 1–2's locks (Language / SurveyLanguage / Survey content rows) are held until `run()` returns.

Two corrections to the framing though:

- Response ingestion is **not** stalled: there's no `responseCount` column/trigger on `Survey`, and an insert's `KEY SHARE` lock on the Survey FK row is compatible with the content updates' `FOR NO KEY UPDATE` locks. What blocks is survey *editing* during the migration — which was already true before this PR.
- The timeout failure mode converges: committed chunks survive, and retries re-walk mostly read-only, so each attempt is faster. Worst case is a crash-looping deploy that converges after a few attempts, not a stuck migration.

So I wouldn't gate the PR on this given the body explicitly accepts the trade-off — but let's file the runner opt-out (or a split into a bounded-tx migration + a chunked autocommit one) as a follow-up rather than resolving this thread as-is.

xernobyl (approved):
Rereviewed at caa1a2a0 — all prior feedback addressed:

- Read-side tag regex widened alongside the write regex, so the newly-writable codes (fil-PH, bho-IN, eo-001, …) now report "not configured" (unknown) instead of a misleading "invalid" — covered by new tests.
- Response/contactAttributes remap is case/underscore-insensitive again (lower(replace(...)) vs lowercased map keys), with the `<> canon` guard avoiding dead-tuple rewrites. Verified there are no lowercase-key collisions in LANGUAGE_CANONICAL_MAP, so the remap is unambiguous.
- Confirmed against the runner that the two high-volume tables commit per chunk via the autocommit `prisma` handle; step-4's batch loop provably terminates (pairs only include codes where canon ≠ old).

No blocking findings. One non-blocking note: the documented transaction-scope limitation is real — steps 1–2's locks are held for the whole run() and everything shares the 30-min budget. Fine to merge as the documented follow-up; let's make sure a ticket tracks the runner opt-out / big-table decoupling so it doesn't get lost.

(CodeRabbit's negative-boundary regex tests remain an optional nice-to-have.)

PR #8415, opened 2026-07-03
fix: fixes update language code issue

pandeymangg:
## What does this PR do?

Two fixes on the ENG-1067 language-canonicalization epic (targets `epic/language-codes-stabilization`).

### 1. `Language.code` is now immutable on update — closes the last hole in the "code stays canonical regardless of caller" invariant

`createLanguage` was hardened earlier (normalize + `CANONICAL_LANGUAGE_CODES` allowlist), but the update path had neither guard: `updateLanguage` did `data: { ...languageInput, updatedAt }`, spreading a raw `code`, and `ZUpdateLanguageAction.languageInput` was `ZLanguageInput` (which carries `code`). So `updateLanguageAction({ …, languageInput: { code: "anything", alias } })` could write an arbitrary, non-canonical code to `Language.code` — exactly what the create-side hardening prevents. Normal UI flow was unaffected (it re-sends the existing code), but the invariant didn't hold "regardless of caller".

Fixed defense-in-depth:

- **Service (`updateLanguage`)**: writes only `alias` (no longer spreads `languageInput`), so `Language.code` is immutable regardless of caller. Changing a language means delete + create, which enforces the canonical catalog.
- **Action (`updateLanguageAction`)**: `languageInput` schema switched to the alias-only `ZLanguageUpdate`, so any `code` a caller sends is stripped at the boundary (the authorization check schema was updated to match).
- Added a service test locking the invariant (a `code` snuck into the runtime object is ignored).

Verified every `Language.code` write path afterwards: `createLanguage` (canonical + catalog check), `updateLanguage` (alias-only now), and the v3 `ensureV3WorkspaceLanguages` upsert (`update: {}` never touches an existing code; `create` uses a `normalizeV3SurveyWriteLanguageCode`-validated canonical code). So no API can change an existing code or write an arbitrary one.

### 2. Picker label "Identifier (ISO)" → "Identifier (BCP-47)"

The stored codes are now BCP-47 language tags (`de-DE`, `zh-Hans-CN`, `fil-PH`, `eo-001`), not plain ISO codes: BCP-47 is an IETF standard that composes ISO 639 / 15924 / 3166 subtags, and numeric regions like `001` are UN M.49, not ISO 3166 — so "(ISO)" is inaccurate. Relabeled the `identifier` field across all 15 locales (ASCII parens for most, CJK fullwidth parens preserved for `ja-JP` / `zh-Hans-CN` / `zh-Hant-TW`).

## How should this be tested?

- `pnpm --filter=@formbricks/web test lib/language` → **13 pass** (incl. the new "never writes `code`" invariant test).
- UI → workspace languages: the middle column now reads **"Identifier (BCP-47)"**.
- Edit a language's **alias** via the UI → works; the **code** cannot be changed.

### Required

- [x] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/lib/language/service.ts
+++ b/apps/web/lib/language/service.ts
@@ -152,9 +152,14 @@ export const updateLanguage = async (
     validateInputs([languageId, ZId], [languageInput, ZLanguageUpdate], [workspaceId, ZId]);
     const workspace = await getWorkspace(workspaceId);
     if (!workspace) throw new ResourceNotFoundError("Workspace not found", workspaceId);
+    // Only the alias is mutable on update — the `code` is immutable, so `Language.code` stays canonical
+    // regardless of caller (createLanguage enforces the canonical catalog; a code change means delete +
+    // create). Write `alias` explicitly rather than spreading `languageInput`: a caller can pass a `code`
+    // in the runtime object even though the declared type is alias-only, and spreading it would persist an
+    // arbitrary, non-canonical code — the exact hole the create-side hardening closed.
     const prismaLanguage = await prisma.language.update({
       where: { id: languageId },
-      data: { ...languageInput, updatedAt: new Date() },
+      data: { alias: languageInput.alias, updatedAt: new Date() },
       select: { ...languageSelect, surveyLanguages: { select: { surveyId: true } } },
     });
 
--- a/apps/web/lib/language/tests/language.test.ts
+++ b/apps/web/lib/language/tests/language.test.ts
@@ -110,6 +110,24 @@ describe("updateLanguage", () => {
     expect(result).toEqual(mockUpdatedLanguage);
   });
 
+  test("never writes `code` — only alias is mutable (invariant regardless of caller)", async () => {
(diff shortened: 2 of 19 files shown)


xernobyl (approved):
LGTM — clean, well-scoped defensive fix. Makes a language's `code` immutable on update (alias-only), with sensible defense in depth:

- `service.ts` writes `{ alias }` explicitly instead of spreading `languageInput` — necessary because `validateInputs` only asserts (doesn't reassign the Zod-stripped value), so an injected `code` would otherwise survive into the spread.
- action + authz schema switched to `ZLanguageUpdate`, stripping `code` at the boundary.
- No UX regression: the code input is always disabled and the language select is disabled for existing languages, so `code` was never user-editable on update anyway.
- Nice regression test that sneaks a `code` in and asserts it's never persisted.
- `(ISO)` → `(BCP-47)` label is the correct term for `en-US`-style tags and is consistent across all 16 locales.

Sonar gate + unit/lint/translation checks all green.

PR #8420, opened 2026-07-06
fix: preserve language aliases through canonicalization (contact-language, editor-lock, merge) (ENG-1067)

pandeymangg:
## What does this PR do?

Fixes language-**alias** back-compat gaps in the ENG-1067 canonicalization epic (targets `epic/language-codes-stabilization`). The migration canonicalizes a language's `code` but leaves its user-set `alias` untouched, which surfaced three alias edge cases. Verified impact against a production `Language` dump (3,184 rows).

### ① Contact language set to a survey-language ALIAS no longer degrades to the default language

An identified contact whose `language` is a human-word alias (the documented `setLanguage("<alias>")`) regressed: the user-state echo `normalizeLanguageCode("deutsch") → null → undefined`, so the SDK adopted `undefined` and rendered the **default** language instead of the aliased one. (87 free-word aliases exist in the dump, so the feature is in real use.)

Fixed without weakening the "drop junk" hardening — we distinguish a real alias from junk by checking the workspace's configured languages:
- New `isWorkspaceLanguageIdentifier(workspaceId, value)` — on the *non-canonical branch only* (so the common language-code path pays no extra query), checks the value against the workspace `Language` rows (code OR alias, case-insensitive).
- Write: a value that matches a configured alias is **stored verbatim**; genuine junk is still dropped + surfaced via `invalid_language_ignored`.
- Echo: a non-canonical stored value (now guaranteed a verified alias) is **echoed verbatim**, so the SDK matches the survey's alias again.

### ② Canonicalization producing `code == alias` no longer blocks the Languages editor

If a bare-code language had an alias equal to its own region tag (e.g. code `de`, alias `"de-DE"`), the migration relabels the code to `de-DE`, making `code == alias`. The workspace Languages editor's `validateLanguages` rejects any alias equal to a code, which blocked **every** save in that editor (even edits to unrelated languages). Measured: **19 rows across 15 workspaces** in the dump.

Fix: after the relabel/merge step, the migration nulls self-referential aliases:
```sql
UPDATE "Language" SET alias = NULL WHERE alias IS NOT NULL AND lower(trim(alias)) = lower(code)
```
A self-referential alias only duplicates the code (adds no matchable value), so clearing it is safe. The condition mirrors `validateLanguages` exactly (lower + trim, no underscore normalization), so only rows that would actually trigger the lock are touched (underscore variants like `en_us` are left alone). Cross-row collisions were checked in the dump: **0** occurrences.

### ③ Merge no longer silently drops an absorbed alias

When two rows collapse to one canonical (e.g. `de` + `de-DE`) and both carry a distinct alias, only one can survive (a row has a single `alias` field). `planLanguageMerges` now computes `droppedAliases`, and the migration logs a warning so an operator can re-add a lost alias if a link/SDK relied on it. (Dump has **0** such cases today — purely future-defensive; no data-preserving fix is possible with one alias slot.)

## How should this be tested?

- `pnpm --filter=@formbricks/web test .../user/lib/update-user.test.ts` → **11 pass** (incl. alias-stored-verbatim + junk-still-dropped).
- `cd packages/database && npx vitest run migration/20260625104904_canonicalize_language_codes/utils.test.ts` → **22 pass** (incl. `droppedAliases`).
- `apps/web` + `packages/database` typecheck & eslint clean.
- Manual: `setLanguage("<alias>")` on an identified user → survey renders in the aliased language; a workspace whose migrated code equals an old alias can save in Workspace > Languages.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/update-user.test.ts
+++ b/apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/update-user.test.ts
@@ -182,6 +182,24 @@ describe("updateUser", () => {
     expect(result.messages).toBeUndefined();
   });
 
+  test("matches a configured alias case-insensitively and preserves the caller's casing", async () => {
+    vi.mocked(prisma.contact.findFirst).mockResolvedValue(mockContactData as any);
+    // Configured alias is lowercase "deutsch"; the SDK sends "DEUTSCH". Matching is case-insensitive
+    // everywhere (SDK/renderer/v3), so it must be kept — and stored verbatim (not lowercased to the config
+    // value), since downstream matchers lowercase both sides anyway.
+    vi.mocked(prisma.language.findMany).mockResolvedValue([{ code: "de-DE", alias: "deutsch" }] as any);
+    const newAttributes = { email: "new@example.com", language: "DEUTSCH" };
+
+    const result = await updateUser(mockWorkspaceId, mockUserId, "desktop", newAttributes);
+
+    expect(updateAttributes).toHaveBeenCalledWith(mockContactId, mockUserId, mockWorkspaceId, {
+      email: "new@example.com",
+      language: "DEUTSCH",
+    });
+    expect(result.state.data?.language).toBe("DEUTSCH");
+    expect(result.messages).toBeUndefined();
+  });
+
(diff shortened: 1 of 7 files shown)


xernobyl (approved):
thank you for the PR. LGTM.

PR #8423, opened 2026-07-06
feat(i18n): validate i18n lockfile synchronization in scan-translations

Dhruwang:
### Description

This PR introduces offline validation for the `i18n.lock` file directly within the `scan-translations` script. This prevents situations where developers commit new translation changes but forget to regenerate the `i18n.lock` file via `pnpm i18n`.

### Changes

* **`scan-translations` Script Updates**:
  * Implemented an offline `validateLockfile` check.
  * Added `object-hash` as a dependency to `@formbricks/i18n-utils` to compute the MD5 checksums of translation values using Lingo.dev's hashing algorithm.
  * Checks if every key/value in the default locale file is present and matches the hash in `i18n.lock`.
  * Outputs detailed error messages pointing to missing, stale/out-of-sync, or extra checksum keys.
  * Fails the validation run if any discrepancies are found, preventing out-of-sync commits in CI.
* **Dependencies**: Added `object-hash` dependency to `packages/i18n-utils`.


packages/i18n-utils/src/scan-translations.ts
```diff
@@ -442,8 +458,120 @@ function displayResults(results: ScanResults, packageName: string, defaultLocale
...
+  }
+
+  for (const key of Object.keys(lockChecksums)) {
+    if (sourceKeysWithValues[key] === undefined) {
+      extra.push(key);
+    }
+  }
+
+  return { missing, outOfSync, extra };
+}
```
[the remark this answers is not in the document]
Dhruwang: Thanks for the feedback! 

1. Removed the unused `packageName` parameter.
2. Updated the key presence check from `=== undefined` to use `!(key in sourceKeysWithValues)` to align with TypeScript/ESLint rules.
3. Regarding the hashing method: `objectHash.MD5` is actually correct here. The Lingo.dev CLI generates lockfile checksums using the `object-hash` library which prepends type information tags (e.g., `string:` prefix) before hashing. Using standard MD5 would result in a mismatch, whereas using `objectHash.MD5` matches the lockfile's checksums perfectly for all 3,808 keys.

packages/i18n-utils/src/scan-translations.ts
```diff
@@ -442,8 +458,135 @@ function displayResults(results: ScanResults, packageName: string, defaultLocale
...
+
+  recurse(translations);
+  return flattened;
+}
+
+interface ObjectHash {
+  MD5: (val: unknown) => string;
+}
+
+async function validateLockfile(
```
xernobyl: **[P2] The new validation logic ships with no tests, and the CI gate never ran on this PR.**

`Validate Translation Keys` skipped on every push here (the path filter excludes `packages/i18n-utils/**`), and there are no unit tests for `validateLockfile` / `parseLockfile` / the flattening / the hashing. I verified it by hand and it's correct today — but the whole thing hinges on `objectHash.MD5` matching Lingo.dev's serialization, which is exactly the kind of invariant that silently breaks on a dependency bump.

Could you add a few cases to `scan-translations.test.ts`?
- a known value → known object-hash checksum (locks in the object-hash-vs-plain-md5 distinction);
- one fixture each that trips `missing`, `outOfSync`, and `extra`;
- a `parseLockfile` case for the 4-space/32-hex line format.

packages/i18n-utils/src/scan-translations.ts
```diff
@@ -442,8 +458,135 @@ function displayResults(results: ScanResults, packageName: string, defaultLocale
...
+  const lockfileContent = await fs.promises.readFile(lockfilePath, "utf-8");
+  const lockChecksums = parseLockfile(lockfileContent);
+  const sourceKeysWithValues = await loadDefaultLocaleWithValues(localesDir, defaultLocale);
+
+  const missing: string[] = [];
+  const outOfSync: string[] = [];
+  const extra: string[] = [];
+
+  for (const [key, val] of Object.entries(sourceKeysWithValues)) {
+    const expected = (objectHash as unknown as ObjectHash).MD5(val);
```
xernobyl: **[P3, trivial] Redundant cast.** Now that `@types/object-hash` is installed, `objectHash.MD5` is properly typed — `(objectHash as unknown as ObjectHash).MD5(val)` throws away the types you just added. This can be simplified to `objectHash.MD5(val)` and the local `ObjectHash` interface (line 547) deleted.

xernobyl (commented):
Re-reviewed after the latest pushes. The build break is fully resolved (typings + `ScanResults` + unused param), lint is green, and I independently re-verified the hashing is correct — `objectHash.MD5` reproduces the lockfile checksums (plain md5 does not), and running the full validation against the current tree is clean (web 3808/3808, surveys 79/79, zero discrepancies). So the logic is sound and the gate would pass green.

Two things still open before this is a CI check I'd fully trust, plus a trivial cleanup (inline). Neither is a correctness bug — flagging for judgment.

**Path-filter gap (can't inline — file isn't in this diff):** `.github/workflows/translation-check.yml` only triggers `scan-translations` on `apps/web/**`, `packages/surveys/**`, `packages/email/**` and locale JSON. It doesn't include `packages/i18n-utils/**` or the `**/i18n.lock` files — which is why `Validate Translation Keys` **skipped on every push of this PR**, so the new validator has never actually executed in CI here. Consider adding `packages/i18n-utils/**` and `**/i18n.lock` to the `translations` filter so the validator runs when the validator or a lockfile changes.

xernobyl (approved):
Thank you for the quick fixes. LGTM.

PR #8437, opened 2026-07-07
fix: normalize signup email token creation

BhagyaAmarasinghe:
Fixes: ENG-1548

## What changed

- Normalize signup email token creation so mixed-case signup input is looked up with the canonical lowercase email.
- Mint legacy email tokens from the stored `user.email` value instead of the raw form input.
- Stop the signup form from routing to success/verification pages when token creation fails.
- Render an invalid-token state on the no-verification success page instead of throwing `jwt malformed`.
- Add unit coverage for `createEmailTokenAction` and update the signup Playwright happy path to use mixed-case email.

## Root cause

Signup creation lowercased the email before storing the user, but the client then requested an email token with the original mixed-case form value. The token action performed a case-sensitive lookup, failed to find the stored user, and the UI still navigated with `token=undefined`. The success page then tried to decode that value as a JWT and threw `jwt malformed`.

## Validation

- `pnpm exec vitest run modules/auth/actions.test.ts modules/auth/signup/actions.test.ts`
- `git diff --check`

Playwright signup was not run locally because the config expects an already-running app at `localhost:3000`, and no server was listening there during validation.

### DIFF
--- a/apps/web/modules/auth/actions.test.ts
+++ b/apps/web/modules/auth/actions.test.ts
@@ -1,0 +1,49 @@
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { createEmailToken } from "@/lib/jwt";
+import { getUserByEmail } from "@/lib/user/service";
+import { createEmailTokenAction } from "./actions";
+
+vi.mock("@/lib/jwt", () => ({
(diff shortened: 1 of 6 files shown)


xernobyl (approved):
LGTM

PR #8438, opened 2026-07-07
fix: prevent manager privilege escalation via v2 organization-users API [ENG-1518]

pandeymangg:
## What does this PR do?

Fixes a privilege-escalation issue in the v2 organization-users management API (**ENG-1518**, reported privately via `SECURITY.md`).

`POST`/`PATCH /api/v2/organizations/{orgId}/users` wrote the `role` from the request body straight onto the membership after only a write-access check — nothing verified that the caller was allowed to grant that role. On self-hosted, a manager can mint a write-scoped org API key, so a manager could:

- create a new **owner** (`POST` with `"role": "owner"`), or
- promote an existing member up to owner/manager (`PATCH`),

escalating themselves (or anyone) to full owner. The settings/session path already blocked this in `ee/role-management/actions.ts`; the API path had no equivalent clamp.

While closing that, the related gap where a **non-owner could demote an existing owner** is also fixed — on both the API path and the session/settings action (the UI already disabled it client-side, but the server didn't enforce it).

Self-hosted only — the endpoint bails on Cloud via `IS_FORMBRICKS_CLOUD`.

### Changes

Org API keys carry no role of their own, so authorization is anchored to the key's **creator**:

- `getApiKeyCreatorRole` — resolves the creating user's current org role; returns `null` for legacy keys with no `createdBy`, deleted users, or creators who are no longer members (callers fail safe).
- `getMembershipRoleByEmail` — resolves the target user's current role.
- `canAssignOrganizationRole(assignerRole, targetRole, targetCurrentRole?)` — an owner may assign any role; a non-owner may not change an existing owner's role; a manager may only assign `member`; anyone else (including an unresolved creator) may not assign a role at all.
- `canModifyOrganizationMember(assignerRole, targetCurrentRole)` — only an owner may modify an existing owner's membership at all. This guards every field a `PATCH` can touch (active state, email, teams), not just the role, so a manager-scoped key can't deactivate or lock out an owner via a non-role field.
- `canManageOrganizationUsers(assignerRole)` — enforces the org's `USER_MANAGEMENT_MINIMUM_ROLE` floor (via the shared `getUserManagementAccess`) on the API path, matching the settings/session path. An install that restricts user management to owners only (or disables it) is no longer bypassed by a manager-created key; an unresolved creator never clears the floor.
- `route.ts` — both `POST` and `PATCH` first enforce the user-management floor, then apply the role clamp. `PATCH` additionally blocks a non-owner from modifying an existing owner (any field). Violations return `403`.
- `ee/role-management/actions.ts` — `updateMembershipAction` now rejects a non-owner changing an owner's role, matching the API path and the existing UI behavior, and reuses the already-fetched target membership for the audit-log `oldObject` (drops a redundant read).

### ⚠️ Upgrade note (self-hosted)

Because org API keys carry no role, the assignable role is anchored to the key's **creator**. If that creator can't be resolved — a **legacy key created before `createdBy` was recorded**, a key whose **creator was deleted**, or a creator who is **no longer a member** — the fail-safe denies role assignment, so `POST` (role is required) and any role-changing `PATCH` on such a key return `403`. Non-role `PATCH`es on non-owner targets are unaffected. Self-hosted installs whose user-provisioning automation runs on a legacy/orphaned key should **regenerate the key** (so it has a resolvable creator with a sufficient role) after upgrading. Worth a line in the release notes / upgrade guide.

## How should this be tested?

Self-hosted instance, in an org where the API key was created by a **manager** (the key creator's role is what gets enforced). With `x-api-key` set to a manager-created, write-scoped key:

1. `GET  /api/v2/organizations/{orgId}/users` → `200`
2. `POST` `{"name":"...","email":"m@example.com","role":"member"}` → `201` (manager can add members)
3. `POST` `{"name":"...","email":"e@example.com","role":"owner"}` → `403` (escalation blocked)
4. `POST` `{"name":"...","email":"e2@example.com","role":"manager"}` → `403` (manager can only assign `member`)
5. `PATCH` `{"email":"<an existing owner>","role":"member"}` → `403` (owner demotion blocked)
6. `PATCH` `{"email":"<an existing owner>","isActive":false}` → `403` (non-owner can't deactivate an owner, even with no role change)

An **owner**-created key still assigns any role and can modify any member, so there's no regression for legitimate use. On `main` (pre-fix), step 3 returns `201` and the new user comes back with `"role":"owner"`.

Unit coverage added in `apps/web/modules/api/v2/organizations/[organizationId]/users/lib/tests/utils.test.ts` (role-clamp matrix + creator/target resolution, incl. legacy/deleted-creator fail-safe).

### Required

- [x] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
+++ b/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
@@ -1,6 +1,8 @@
 import { prisma } from "@formbricks/database";
 import { OrganizationRole, Prisma } from "@formbricks/database/prisma";
 import { TOrganizationRole } from "@formbricks/types/memberships";
+import { USER_MANAGEMENT_MINIMUM_ROLE } from "@/lib/constants";
+import { getUserManagementAccess } from "@/lib/membership/utils";
 import { buildCommonFilterQuery, pickCommonFilter } from "@/modules/api/v2/management/lib/utils";
 import { TGetUsersFilter } from "@/modules/api/v2/organizations/[organizationId]/users/types/users";
 
@@ -125,4 +127,16 @@ export const canModifyOrganizationMember = (
 ): boolean => {
   if (assignerRole === OrganizationRole.owner) return true;
   return targetCurrentRole !== OrganizationRole.owner;
+};
+
+/**
+ * Whether the acting user (the API key creator) clears the org's user-management floor. The v2 API
+ * anchors this to the key creator's role so the `USER_MANAGEMENT_MINIMUM_ROLE` policy the
+ * settings/session path enforces (via `getUserManagementAccess`) is applied consistently here — an
+ * install that restricts user management to owners only, or disables it, isn't silently bypassed by
+ * a manager-created key. An unresolved creator (null) never clears the floor.
+ */
+export const canManageOrganizationUsers = (assignerRole: TOrganizationRole | null): boolean => {
+  if (!assignerRole) return false;
+  return getUserManagementAccess(assignerRole, USER_MANAGEMENT_MINIMUM_ROLE);
 };
--- a/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
+++ b/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
@@ -112,3 +112,17 @@ export const canAssignOrganizationRole = (
   if (assignerRole === OrganizationRole.manager) return targetRole === OrganizationRole.member;
   return false;
 };
+
+/**
+ * Whether the acting user (the API key creator) may modify the target membership at all. Only an
+ * owner may act on an existing owner: this guards not just the role, but every other membership
+ * field an update can touch (active state, email, teams). Without it a manager-scoped key could,
+ * for example, deactivate or lock out an owner even though it can't change their role.
+ */
+export const canModifyOrganizationMember = (
+  assignerRole: TOrganizationRole | null,
+  targetCurrentRole: TOrganizationRole | null
+): boolean => {
+  if (assignerRole === OrganizationRole.owner) return true;
+  return targetCurrentRole !== OrganizationRole.owner;
+};
--- a/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
+++ b/apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
@@ -1,4 +1,6 @@
-import { Prisma } from "@formbricks/database/prisma";
+import { prisma } from "@formbricks/database";
+import { OrganizationRole, Prisma } from "@formbricks/database/prisma";
+import { TOrganizationRole } from "@formbricks/types/memberships";
 import { buildCommonFilterQuery, pickCommonFilter } from "@/modules/api/v2/management/lib/utils";
 import { TGetUsersFilter } from "@/modules/api/v2/organizations/[organizationId]/users/types/users";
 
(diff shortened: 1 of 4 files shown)


apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
```diff
@@ -40,3 +42,73 @@ export const getUsersQuery = (organizationId: string, params?: TGetUsersFilter)
...
+export const getApiKeyCreatorRole = async (
+  apiKeyId: string,
+  organizationId: string
+): Promise<TOrganizationRole | null> => {
+  const apiKey = await prisma.apiKey.findUnique({
+    where: { id: apiKeyId },
+    select: { createdBy: true },
+  });
+
+  if (!apiKey?.createdBy) return null;
```
Comment (xernobyl) [inline apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts:65]: **[P3 — intended fail-safe, but please document the blast radius]**

Any org key that cannot resolve a creator role returns `null` here — legacy keys with no `createdBy`, a **deleted** creator, or a creator who is **no longer a member**. Because `role` is required on `POST`, `canAssignOrganizationRole(null, …)` then denies **every** `POST` (not only owner-role requests), and likewise any `PATCH` that sets a role.

This is the correct security default, but the impact is broader than "escalation blocked": self-hosted installs whose user-provisioning automation runs on a legacy/orphaned key will have **all** user creation break with `403` after upgrade. Worth an explicit line in the release notes / upgrade guide.

apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts
```diff
@@ -146,6 +166,27 @@ export const PATCH = async (request: Request, props: { params: Promise<{ organiz
         );
       }
 
+      // Only a role change needs clamping; when no role is supplied the membership role is left
+      // untouched. Anchor the assignable role to the API key creator's role so a manager cannot
+      // promote a target above the member role — or demote an existing owner — through the
+      // management API.
+      if (body.role) {
```
Comment (xernobyl) [inline apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts]: **[P3 — related gap, likely a follow-up rather than a blocker]**

The target-owner guard only fires inside `if (body.role)`, but `updateUser` also mutates `name`, `email`, `isActive`, and `teams`. So a manager-created key can still `PATCH` an **owner** to `isActive: false` (disable them), change their `email`, or strip their teams — the same "manager acts on a superior" threat class this PR closes, just via non-role fields.

This is **pre-existing** and outside ENG-1518's stated scope, so not a blocker here. But given the PR frames the goal as managers not being able to act on owners, consider a follow-up that applies a `targetCurrentRole === owner` check to any mutation by a non-owner, not just role changes.

Comment (pandeymangg) [reply apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts]: Good catch — addressed in ddf175c. Added `canModifyOrganizationMember(assignerRole, targetCurrentRole)` and the PATCH handler now runs it on every request (not just when `body.role` is set), so a non-owner key can't touch an existing owner's membership at all — role, `isActive`, `email`, or `teams`. The role clamp still applies on top when a role is supplied. Tests + upgrade note updated.

apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts
Comment (pandeymangg) [reply apps/web/modules/api/v2/organizations/[organizationId]/users/lib/utils.ts:65]: Agreed, keeping the fail-safe (deny when the creator role can't be resolved) and documenting the blast radius. Added an explicit "Upgrade note (self-hosted)" to the PR description: legacy keys without `createdBy`, keys whose creator was deleted, or a creator no longer in the org will get 403 on POST / role-changing PATCH after upgrade, and such installs should regenerate the key. Happy to also add this to the release notes / upgrade guide.

apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts
```diff
@@ -86,6 +92,21 @@ export const POST = async (request: Request, props: { params: Promise<{ organiza
         );
       }
 
+      // Org API keys carry no role of their own, so clamp the assignable role to the role of the
+      // user who created the key. This mirrors the settings/session-path guard and stops a manager
+      // from escalating a user (or themselves) to owner through the management API.
+      const assignerRole = await getApiKeyCreatorRole(authentication.apiKeyId, authentication.organizationId);
```
Comment (xernobyl) [inline apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts:100]: **[P2/P3 — authorization-consistency gap, likely a follow-up rather than a blocker]**

The creator-role clamp closes the escalation-to-owner path, which is the important fix here. But the v2 API path has no equivalent to the session path's `USER_MANAGEMENT_MINIMUM_ROLE` floor.

That constant is env-configurable (`owner` / `manager` / `disabled`, default `manager`) and the session/settings action enforces it via `getUserManagementAccess()` in `ee/role-management/actions.ts`. The API handlers only apply the creator-role clamp, so an install that has deliberately restricted user management to **owners only** (or `disabled` it) still lets a *manager-created* key create members and modify non-owner members through `POST`/`PATCH`.

No owner-escalation, so not a blocker — but it's a real policy bypass for self-hosters who set `USER_MANAGEMENT_MINIMUM_ROLE=owner` and assume it's enforced everywhere. Cheap fix: gate both handlers on `getUserManagementAccess(assignerRole, USER_MANAGEMENT_MINIMUM_ROLE)` before the clamp.

apps/web/modules/ee/role-management/actions.ts
```diff
@@ -130,6 +130,14 @@ export const updateMembershipAction = authenticatedActionClient.inputSchema(ZUpd
       throw new OperationNotAllowedError("Managers can only assign users to the member role");
     }
 
+    const targetMembership = await getMembershipByUserIdOrganizationId(
```
xernobyl, line 133: **[P3 — minor, DRY/perf]**

`targetMembership` is fetched here, but a few lines below `ctx.auditLoggingCtx.oldObject` re-fetches the identical `(userId, organizationId)` record — a redundant DB round trip on the same request path. Reuse this value: `ctx.auditLoggingCtx.oldObject = targetMembership;`

(Same thing CodeRabbit flagged; still open.)

apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts
Comment (pandeymangg) [reply apps/web/modules/api/v2/organizations/[organizationId]/users/route.ts:100]: Good call — fixed. Added `canManageOrganizationUsers(assignerRole)` which runs `getUserManagementAccess(assignerRole, USER_MANAGEMENT_MINIMUM_ROLE)`, and both POST and PATCH now gate on it before the role clamp. So an install with `USER_MANAGEMENT_MINIMUM_ROLE=owner` (or `disabled`) is enforced on the API path too, not just the session path. An unresolved creator (null) never clears the floor.

apps/web/modules/ee/role-management/actions.ts
pandeymangg: Done — `ctx.auditLoggingCtx.oldObject = targetMembership;` now reuses the value fetched a few lines up instead of re-querying.

xernobyl (approved):
Approving. The privilege-escalation fix is sound — creator-anchored role clamp, owner-target protection across all mutable fields, org-match enforced, and the unresolved-creator fail-safe all hold. The follow-up commit also closes the `USER_MANAGEMENT_MINIMUM_ROLE` parity gap (the API now honors an owner-only/disabled floor) and reuses `targetMembership` for the audit `oldObject`. Tests pass locally (23/23).

Non-blocking: a couple of handler-level integration tests (e.g. manager-created key → POST owner → 403) would lock in the route wiring, but the authz logic is well covered by the unit tests.

PR #8439, opened 2026-07-07
fix: validate userId length before tearing down existing user

### LINKED ISSUE #8288: bug: invalid replacement userId tears down current SDK user state
### Issue Summary

```text
bug: invalid replacement userId tears down current SDK user state
```

```md
## Issue

`setUserId` currently tears down the existing SDK user state before validating the replacement `userId` length.

This means an invalid oversized replacement `userId` can clear the current valid user even though the new value is rejected.

## Reproduction

Current order in `packages/js-core/src/lib/user/user.ts`:

```ts
if (currentUserId) {
  tearDown();
}

const MAX_USER_ID_LENGTH = 255;
if (userId.length > MAX_USER_ID_LENGTH) {
  return okVoid();
}
```

Example flow:

```ts
await formbricks.setUserId("existing-user");
await formbricks.setUserId("a".repeat(256));
```

## Current behavior

- Existing user state is torn down.
- The oversized replacement `userId` is rejected.
- No new user is set.
- The SDK loses the previously valid user state.

## Expected behavior

The replacement `userId` should be validated before tearing down the current user.

If the new `userId` is invalid, the SDK should reject it without clearing the existing valid user state.

## Minimal fix

Move the max-length validation before the `tearDown()` call.

I verified this with a focused regression test:

```bash
vitest run src/lib/user/tests/user.test.ts
```

Result after fix:

```text
1 file passed, 7 tests passed
```

Focused ESLint also passes for:

```bash
eslint src/lib/user/user.ts src/lib/user/tests/user.test.ts
```

Branch:
https://github.com/bharathkumar39293/formbricks/tree/fix/sdk-setuserid-validate-before-teardown

Since PR creation is restricted, I’m happy to open a PR if access is enabled, or a maintainer can cherry-pick the commit if preferred.
```

### Expected Behavior

_No response_

_No response_

### Your Environment

None

pandeymangg:
## What does this PR do?

`setUserId` in `packages/js-core` validated the replacement `userId` length **after** tearing down the existing user state. So calling `setUserId` with an oversized (> 255 chars) userId while a valid user was already set would tear the current user down and *then* reject the new value — leaving the SDK with no user at all.

This PR moves the `MAX_USER_ID_LENGTH` validation to the start of `setUserId`, before any state is mutated. An invalid replacement `userId` is now rejected up front without clearing the existing valid user. Behavior for valid userIds is unchanged.

Fixes #8288

## How should this be tested?

Added a regression test in `packages/js-core/src/lib/user/tests/user.test.ts` for the exact scenario: an existing user is set, then `setUserId` is called with a 256-character userId — `tearDown` must not be called and no updates are queued.

- `cd packages/js-core && CI=true pnpm test -- --run src/lib/user/tests/user.test.ts` — all user tests pass
- Verified the new test fails against the pre-fix source (at `expect(tearDown).not.toHaveBeenCalled()`) and passes with the fix
- `pnpm lint` on the touched files is clean

### Required

- [x] Ran `pnpm build`

### Appreciated

### DIFF
--- a/packages/js-core/src/lib/user/tests/user.test.ts
+++ b/packages/js-core/src/lib/user/tests/user.test.ts
@@ -185,6 +185,41 @@ describe("user.ts", () => {
       expect(mockUpdateQueue.updateUserId).not.toHaveBeenCalled();
       expect(mockUpdateQueue.processUpdates).not.toHaveBeenCalled();
     });
+
+    test("does not tear down the existing user when the replacement userId is too long", async () => {
+      const mockConfig = {
+        get: vi.fn().mockReturnValue({
+          user: {
+            data: {
+              userId: "existing-user",
+            },
+          },
+        }),
+      };
+
+      const mockLogger = {
+        debug: vi.fn(),
+        error: vi.fn(),
+      };
+
+      const mockUpdateQueue = {
+        updateUserId: vi.fn(),
+        processUpdates: vi.fn(),
+      };
+
+      getInstanceConfigMock.mockReturnValue(mockConfig as unknown as Config);
+      getInstanceLoggerMock.mockReturnValue(mockLogger as unknown as Logger);
+      getInstanceUpdateQueueMock.mockReturnValue(mockUpdateQueue as unknown as UpdateQueue);
+
+      const longId = "a".repeat(256);
+      const result = await setUserId(longId);
+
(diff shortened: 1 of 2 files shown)


Dhruwang (approved):

PR #8441, opened 2026-07-07
fix: multi-language responses split select options into per-language categories (ENG-1566)

Dhruwang:
## What

Selected choice values were ingested into Hub as labels localized to the response language, so a bilingual survey split one option into a category per language in analytics — e.g. grouping by Gender produced 4 bars (`Male`, `Female`, `ذكر`, `أنثى`) instead of 2.

Fixes [ENG-1566](https://linear.app/formbricks/issue/ENG-1566/bug-multi-language-responses-split-gender-into-4-categories-instead-of).

## How

`transformResponseToFeedbackRecords` now maps each selected label back to its choice (same `findChoiceByLabel` mechanism the matrix/ranking expansion already uses for `field_label`) and stores the **default-language label** as the canonical `value_text`:

- single select / multi select answers (per array entry)
- matrix column values (row labels were already canonicalized)

Values that match no choice label pass through unchanged — this covers "other" free-text answers and choice labels edited since submission. Ranking is unaffected (`value_number` only); picture selection stores stable ids.

## Why ingestion-side (vs grouping by a value ID)

- Hub has no value/option ID column, and one would be NULL for every non-formbricks source (CSV import, integrations).
- Canonical labels live in survey definitions, not Hub, so Cube could not display a label after grouping by ID without a cross-system join.
- Normalizing `value_text` fixes charts, filters, and the value pick-list at once with zero Cube/UI changes.

## Not in this PR

- **Backfill of existing Hub rows** — historical records keep their localized labels and stay split until backfilled; needs a separate script/follow-up.
- Editing a choice label still means new records get the new canonical label while old rows keep the old one (pre-existing behavior, unchanged).

## Testing

- New `transform.test.ts` cases: bilingual single select, multi select, other/free-text passthrough, default-language no-op, matrix column canonicalization.
- `vitest run lib/feedback-source`: 131/131 green; typecheck + eslint clean on touched files.


xernobyl (commented):
Reviewed as a follow-up to the Hub `value_id` work (formbricks/hub#104). CI and the Sonar gate are green and the approach is solid — match each choice back to its element and attach the stable id, label stored as submitted. A few notes inline; the main one is the `value_id` strip in `service.ts` describing an enable flag that doesn't exist. Nothing blocks correctness at runtime. CodeRabbit's `element.choices` narrowing note is valid — replied on that thread rather than duplicating it here.

apps/web/modules/hub/service.ts
```diff
@@ -37,6 +37,16 @@ export type HubFeedbackRecordResult = {
   error: HubError | null;
 };
 
+/**
+ * The deployed Hub rejects unknown request fields, and value_id ships with ENG-1671.
+ * Ingestion always computes it (ENG-1673); send it only when the Hub deployment
+ * supports it (HUB_VALUE_ID_ENABLED=1), otherwise strip it so creates keep working.
+ */
+const toHubCreatePayload = (input: FeedbackRecordCreateParams): FeedbackRecordCreateParams => {
```
xernobyl: **[P2]** The comment says `value_id` is sent only when the Hub supports it (`HUB_VALUE_ID_ENABLED=1`), but there's no such env var (`env.ts` only gained a blank line) and this strips it unconditionally — with a test locking the strip. So there's no way to actually turn it on without a code change.

Stripping is correct *today*: Hub's create decoder uses `DisallowUnknownFields()`, so a pre-`value_id` Hub returns `400` on the extra field. But enabling it later then means editing code + the test, not flipping a flag. Suggest either wiring the real gate (read `HUB_VALUE_ID_ENABLED` in `env.ts`, strip conditionally) so it can be enabled once a Hub with `value_id` is deployed, or dropping the flag wording and linking the follow-up so the rollout step is explicit.

apps/web/lib/feedback-source/transform.ts
```diff
@@ -33,6 +33,37 @@ const findChoiceByLabel = <T extends { id: string; label: TSurveyElementChoice["
...
+  valueId?: string;
+}
+
+/**
+ * Selected choice values arrive as labels localized to the response language. The label
+ * is stored as submitted; cross-language consolidation relies on value_id (ENG-1673),
+ * the stable id of the matched choice. Values that match no choice (an "other" free-text
+ * answer, or a choice label edited since submission) carry no id.
+ */
+const normalizeChoiceValue = (
```
xernobyl, line 52: **[P3]** `normalizeChoiceValue` returns no `valueId` for array values, and `convertValueToHubFields` joins multi-select into one localized `value_text` — so multi-select choice answers don't get a `value_id` and stay subject to the same per-language split this PR fixes for single-select. ENG-1566's case (gender) is single-select, so this may be intentional scope, but it'd be good to make it an explicit follow-up ticket rather than only a test asserting the omission.

apps/web/modules/hub/types.ts
```diff
@@ -1,6 +1,10 @@
 import type FormbricksHub from "@formbricks/hub";
 
-export type FeedbackRecordCreateParams = FormbricksHub.FeedbackRecordCreateParams;
+// value_id (ENG-1671/ENG-1673): stable id of the selected option in the source system (a survey
```
xernobyl, line 3: **[P3/nit]** This comment is truncated mid-sentence (`…in the source system (a survey`) with a stray blank line before the export — looks like an editing slip. Worth completing or removing.

.env.example
```diff
@@ -66,7 +66,10 @@ HUB_API_URL=http://localhost:8080
 HUB_DATABASE_URL=postgresql://postgres:postgres@postgres:5432/hub?sslmode=disable
 # Hub image tag used by docker-compose.dev.yml (hub + hub-migrate). Leave unset to use the
 # pinned default in the compose file; override here when testing a specific Hub release.
-# HUB_IMAGE_TAG=0.3.0
+# HUB_IMAGE_TAG=0.8.0
+
+# Send the stable option id (value_id) with choice-answer feedback records. Requires Hub >= 0.8.0
```
xernobyl: **[nit]** This note names no actual variable and has two trailing blank lines. It ties into the `toHubCreatePayload` flag comment — if you wire a real `HUB_VALUE_ID_ENABLED`, document it here; otherwise reword so it doesn't imply a config knob that doesn't exist.

apps/web/lib/feedback-source/transform.ts
```diff
@@ -243,7 +276,15 @@ export function transformResponseToFeedbackRecords(
...
-    const valueFields = convertValueToHubFields(value, mapping.hubFieldType);
+
+    const isChoiceElement =
+      element?.type === TSurveyElementTypeEnum.MultipleChoiceSingle ||
+      element?.type === TSurveyElementTypeEnum.MultipleChoiceMulti;
+    const normalizedValue = isChoiceElement
+      ? normalizeChoiceValue(element.choices, value, baseFields.language ?? "default")
+      : value;
+
+    const valueFields = convertValueToHubFields(normalizedValue, mapping.hubFieldType);
```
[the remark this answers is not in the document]
xernobyl: Confirmed — `isChoiceElement` is a separate const, so it doesn't narrow `element` (`TSurveyElement | undefined`, and `choices` isn't on every union member). Runtime is safe (the ternary guards it), but it is a real type error under strict `tsc` — it only slips CI because the strict `typecheck` task isn't one of this PR's checks. `const isChoiceElement = element && (element.type === … || element.type === …)` narrows it in the true branch.

apps/web/lib/feedback-source/transform.ts
Dhruwang: Good call — filed ENG-1702 to track it. Multi-select needs the record to expand one-per-option (like matrix/ranking) so each carries its own `value_id`; a single joined record can't hold multiple ids, so it's a structural change rather than a tweak here. Out of scope for this PR since ENG-1566 is single-select (gender), but now tracked beyond the test assertion.

apps/web/modules/hub/service.ts
Dhruwang: Dropped the flag wording — Hub latest supports `value_id`, so we now send it through unconditionally (the SDK forwards unknown fields untouched). Removed the strip in `service.ts` and updated the test to assert `value_id` is forwarded.

apps/web/modules/hub/types.ts
Dhruwang: Fixed — completed the comment explaining what `value_id` is and why it's bridged as an optional param.

.env.example
Dhruwang: Removed the note — no flag anymore, and cleaned up the stray blank line in `env.ts` too.

apps/web/lib/feedback-source/transform.ts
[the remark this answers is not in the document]
Dhruwang: Fixed with `const isChoiceElement = element && (element.type === ... || element.type === ...)` — narrows `element` in the truthy branch, typechecks clean under strict `tsc`.

xernobyl (approved):
Re-reviewed after the review-feedback fixes and the `main` merge. All prior findings are addressed:

- `value_id` is now sent through to the Hub (strip + phantom `HUB_VALUE_ID_ENABLED` flag removed). Safe because `value_id` shipped in Hub 0.8.0 and this PR pins the deployment to 0.8.0 (docker-compose + Helm digest), so the receiving Hub accepts the field.
- `element.choices` is now type-narrowed via `element && (…)`.
- Truncated `types.ts` comment completed; `.env.example` note removed.
- Multi-select `value_id` gap tracked as ENG-1673/ENG-1702 follow-up.

Verified the `main` merge preserved the value_id logic (transform core unchanged; service/types resolutions correct). Full CI is green (Unit, Build, Linters, E2E, SonarQube, Translation Keys, Docker). LGTM 🚀

PR #8442, opened 2026-07-07
fix(link-surveys): make footer links & suggested colors WCAG AA compliant

jobenjada:
## What & why

The legal footer links of a link survey (**Imprint · Privacy · Terms · Report survey**) were hard-coded to `text-slate-500` (#64748B) and sit directly on the survey's page background. On the **default** background — a light-blue tint (`mixColor("#1e40af", "#fff", 0.855)` ≈ `#DEE3F3`) — that is only **~3.7:1**, below the WCAG AA minimum of 4.5:1 for this text. Any darker or more saturated background made it worse, with no way to fix it.

This PR makes the footer links WCAG AA compliant by default by adapting their color to the chosen background, and adds a styling field so creators can override it.

## How it behaves

| Survey background | Footer links |
|---|---|
| Solid color (incl. the default) | Color auto-picked to clear 4.5:1 against the (brightness-adjusted) background — subtle `slate-700`/`slate-100`, escalating to black/white only for mid-tones where the subtle tone can't reach AA |
| Image / animation / upload | Links render on a near-white backdrop (`rgba(255,255,255,0.9)`) with dark text, guaranteeing contrast against a known surface |
| Any | A new optional **Footer link color** styling field overrides the auto value; leaving it empty keeps the automatic, background-aware default |

## Changes

- **`packages/types/styling.ts`** — new optional `linkColor` on `ZBaseStyling`. Nullish, so unset = auto-adjust; no DB migration (styling is JSON).
- **`apps/web/lib/utils/colors.ts`** — `getRelativeLuminance()` + `getContrastRatio()` (real WCAG math; a naive light/dark switch can't guarantee 4.5:1 for mid-tones).
- **`apps/web/lib/styling/footer-link-color.ts`** (new) — `getFooterLinkStyle(styling)` resolves the color/backdrop from `styling.background`. Unit-tested in `footer-link-color.test.ts`.
- **`link-survey-wrapper.tsx` / `legal-footer.tsx`** — compute once and apply via inline style to both footer render sites (cardless in-flow + card-based). Explicit `linkColor` override wins.
- **`background-styling-card/index.tsx`** — a `Footer link color` `ColorField`. This card is shared, so the field appears in **both** the survey styling editor and the workspace theme editor.
- **`apps/web/locales/en-US.json`** — label + description keys (other locales fall back to English).

## Verification

- ✅ `footer-link-color.test.ts` (8 cases) + existing `constants.test.ts` — 9/9 pass. Every returned color/surface pair asserts `≥ 4.5:1` (default, dark, mid-tone `#808080`, brightness, and image/animation/upload backgrounds).
- ✅ eslint on all changed files; `@formbricks/types` typecheck.
- ℹ️ The full `pnpm --filter @formbricks/web typecheck` surfaces only pre-existing *missing generated module* errors (Prisma client not generated in a fresh checkout) — none are in the files this PR touches. CI runs it against a built environment.

### Manual test suggestions
Open a link survey with Imprint/Privacy/Terms set (Report survey shows on Cloud) and check footer contrast on: the default background, a light solid, a dark solid, and an image background (backdrop appears). Then in the survey styling editor → **Background** card, set **Footer link color** and confirm it overrides; clear it to return to auto.

Resolves ENG-1071


jobenjada:
@itsjavi if it can go into 5.2 great, if not we can ship with 5.3 :)

itsjavi (commented):
Reviewed locally (checkout, full check suite, e2e, and visual before/after comparison).

**Checks**: lint, typecheck, unit tests (surveys 50/50, web 28/28), prettier and scan-translations all pass. E2E: `survey-styling.spec.ts` 4/4 and the link-survey create & submit flow 1/1 pass.

**Visual verification** (measured from rendered pixels, before -> after):
- Footer links on the default background: `#62748E` at 3.7:1 -> `#334155` at 7.2:1
- Footer links on a dark background (`#0f172a`): ~2.9:1 -> slate-100 at ~15:1
- Footer on an animation background: raw slate-500 over the video -> white backdrop pill
- Ghost Back button with a light brand (`#fde047`): raw yellow on the white card, unreadable -> contrast-adjusted dark gold
- Suggested question color for brand `#fbbf24`: `#A37C17` at 3.9:1 -> `#826312` at 5.6:1

The consolidation of the color helpers is behavior-preserving and the new logic is well tested. A few remarks inline; the only behavioral one is the interplay between the explicit link color override and the forced backdrop.

apps/web/modules/survey/link/components/link-survey-wrapper.tsx
```diff
@@ -61,6 +62,11 @@ export const LinkSurveyWrapper = ({
     }
   };
   const styling = determineStyling();
+  // Footer legal links sit on the survey background; resolve an AA-compliant color (and an
+  // optional backdrop for non-solid backgrounds). An explicit linkColor override wins.
+  const footerLinkStyle = getFooterLinkStyle(styling);
+  // `||` (not `??`) so an empty-string override falls back to the auto-adjusted color.
+  const footerLinkColor = styling.linkColor?.light || footerLinkStyle.textColor;
```
itsjavi: when `linkColor` is set explicitly, `backdropColor` is still rendered for image/animation/upload backgrounds, so a light override sits on the near-white backdrop and becomes unreadable. Reproduced locally: animation background + `linkColor: #ffffff` renders white links on `rgba(255,255,255,0.9)`. Consider skipping the backdrop when an explicit override exists (the user takes ownership of contrast), or at least mentioning the backdrop in the field description.

apps/web/modules/survey/editor/components/animated-survey-bg.tsx
```diff
@@ -74,27 +76,30 @@ export const AnimatedSurveyBg = ({ handleBgChange, background }: AnimatedSurveyB
...
+          const isSelected = animation === value;
           return (
             <button
               type="button"
               key={key}
               onMouseEnter={() => debouncedManagePlayback(index, "play")}
               onMouseLeave={() => debouncedManagePlayback(index, "pause")}
               onClick={() => handleBg(value)}
-              className="relative cursor-pointer overflow-hidden rounded-lg">
+              className={cn(
```
itsjavi, line 93: selection is now conveyed only visually (border + check icon), and the tile button has no accessible name (video-only content). The old nested checkbox was odd, but it at least exposed checked state to screen readers. Please add `aria-pressed={isSelected}` and an `aria-label` (e.g. derived from the animation index) to the button.

packages/types/colors.ts
```diff
@@ -0,0 +1,122 @@
...
+    getContrastRatio(darkText, surface) >= getContrastRatio(lightText, surface) ? darkText : lightText;
+  if (getContrastRatio(soft, surface) >= minRatio) return soft;
+
+  return getContrastRatio("#000000", surface) >= getContrastRatio("#ffffff", surface) ? "#000000" : "#ffffff";
+};
+
+// Nudges `preferred` toward black (on light surfaces) or white (on dark surfaces) just far
+// enough to clear `minRatio` against `surface`, preserving as much of the original hue as
+// possible. Falls back to the pure pole, which always clears AA on a near-white/near-black surface.
+export const ensureReadable = (
```
itsjavi, line 127: `ensureReadable` calls `isLight(surface)`, which throws for 4/8-digit hex values that `ZColor` accepts and the management API can persist (e.g. `cardBackgroundColor.light = "#ffffff80"` plus a low-contrast `buttonBgColor` makes `addCustomThemeToDom` throw and the survey fails to render). Since the helpers now live in one shared place, make `isLight` (or `ensureReadable`) tolerate or normalize 4- and 8-digit hex instead of throwing; the same latent throw already exists for `brandColor` at `styles.ts:99`, so hardening here fixes all call sites at once.

packages/types/styling.ts
```diff
@@ -61,6 +61,17 @@ export const ZBaseStyling = z.object({
   brandColor: ZStylingColor.nullish(),
   accentBgColor: ZStylingColor.nullish(),
   accentBgColorSelected: ZStylingColor.nullish(),
+  // Color of the link-survey footer legal links (imprint/privacy/terms/report survey).
+  // When unset, the color is auto-adjusted at render time for AA contrast with the background.
+  // The color-picker field always emits an object, so a cleared value arrives as
+  // `{ light: "" }`; allow the empty string (treated as "auto-adjust" at render time via a
+  // truthy fallback) so the optional field validates instead of failing ZColor.
+  linkColor: z
```
itsjavi: **should-fix**: `linkColor` reads as "color of links in the survey" but it only affects the legal footer of link surveys. Once merged this is persisted styling JSON and public API surface, so renaming later needs a migration; suggest `footerLinkColor` now while it is free. Also consider transforming the cleared `""` to `undefined` on save instead of widening the schema with `ZColor.or(z.literal(""))`, which no other color field allows.

packages/types/colors.ts
```diff
@@ -0,0 +1,122 @@
+// Shared color utilities used by both the web app and the survey renderer. Kept in
```
itsjavi, line 1: **nit**: this is 122 lines of runtime WCAG math in a package named for types. There is precedent (`error-handlers.ts`, `getLinkSurveyCardMaxWidth`) and `@formbricks/types` is the only shared dependency between `apps/web` and `packages/surveys`, so it is fine for now; but let's treat it as the exception rather than the pattern. If more shared runtime utils appear, extracting a small `packages/utils` package (like `i18n-utils`) would be the cleaner home.

apps/web/modules/survey/link/components/link-survey-wrapper.tsx
jobenjada: Fixed in 557d8cdb1 — when an explicit footer link color is set, the near-white backdrop is now skipped entirely (for media backgrounds too). The override means the user takes ownership of contrast, so their color renders directly on the background.

apps/web/modules/survey/editor/components/animated-survey-bg.tsx
jobenjada: Added in 557d8cdb1 — the tile button now has `aria-pressed={isSelected}` and a translated `aria-label` derived from the animation index ("Animated background {n}", new i18n key in all locales).

packages/types/colors.ts
jobenjada: Hardened in 557d8cdb1 — added a shared `normalizeHex` helper that accepts every hex form `ZColor` allows (#RGB/#RGBA/#RRGGBB/#RRGGBBAA, alpha stripped) and used it in both `isLight` and `hexToRGBA`, so `ensureReadable`/luminance/contrast all tolerate 4/8-digit values — including the latent `brandColor` throw at styles.ts:99. Truly invalid strings still throw. Covered by new unit tests (incl. the `cardBackgroundColor: "#ffffff80"` scenario).

packages/types/styling.ts
jobenjada: Both done in 557d8cdb1 — renamed to `footerLinkColor` while it's still free, and reverted the schema to a strict `ZStylingColor.nullish()`. The background styling card now transforms a cleared picker to `undefined` at the form layer instead of persisting `{ light: "" }`, so no other color field's contract changes.

packages/types/colors.ts
jobenjada: Agreed — added a note to the header comment in `packages/types/colors.ts` in 557d8cdb1 stating this is a pragmatic exception, not a pattern, and that further shared runtime utils should go into a dedicated `packages/utils` package (like `i18n-utils`).

itsjavi (approved):
thanks LGTM. I also checked that it doesn't reintroduce this issue  https://linear.app/formbricks/issue/ENG-839/back-button-is-not-visible-on-dark-card-background-and-cannot-be

PR #8443, opened 2026-07-07
fix(surveys): standalone Conditional Logic section + unified jump-to dropdown (ENG-1328)

jobenjada:
## Summary

image

Three survey-editor improvements in one PR.

### 1. Conditional Logic gets its own collapsible section

Previously the whole Conditional Logic UI (headline, rules, and "Add logic") lived *inside* the "Block settings" collapsible, which also auto-opened whenever a block had logic — conflating logic with the Back/Next button-label settings.

Now:
- The Conditional Logic section is extracted into its **own always-visible section, above "Block settings"**.
- The **"Conditional Logic" headline becomes a collapsible trigger** once a block has ≥1 rule — **collapsed by default**, showing a **rule-count** badge when collapsed.
- **"Add logic" stays visible** outside the collapsible and **auto-expands** the section when clicked.
- **"Block settings" now holds only the Back/Next button labels** and **defaults to closed** (the logic-based auto-open was removed).

### 2. ENG-1328 — consistent "Jump to block" / fallback dropdowns

The "Jump to block" action picker (`InputCombobox`, capped at `max-h-[400px]`) and the "All other answers will continue to…" fallback (shadcn `Select`, capped only by viewport height) used different components with different height behavior, so on surveys with many blocks the fallback grew to fill the screen and was hard to use.

Now:
- The fallback picker uses the **same `InputCombobox`** as "Jump to block" (fixed max-height + scroll), so both are consistent and bounded.
- The block/target picker gains a **search box**, so choosing among 30+ blocks is easy.

Resolves ENG-1328.

## Testing

- ✅ ESLint (0 issues) and full-project `tsc --noEmit` (0 errors) on the changed files.
- ✅ Unit suites `blocks.test.ts` + `surveyLogic/utils.test.ts` (83 passed).
- ✅ Playwright `helper.ts` logic flows analyzed as compatible — no fallback interaction; "Add logic" auto-expands the section; block cards collapse one-at-a-time so `.first()` still resolves. No helper changes needed.
]


Dhruwang (approved):

PR #8450, opened 2026-07-07
feat(unify-feedback): rework Topics & Subtopics on React Query + /api/v3

jobenjada:
## Summary

Reworks the Unify Feedback **Topics & Subtopics** view. The old page was a single 1,329-line client component that eager-loaded everything, hard-gated on *any* unembedded record, self-polled with `setTimeout`, and rendered a miller-column tree. This replaces the presentation layer with a **TanStack Query + `/api/v3`** architecture (per the `build-frontend` conventions) behind a new two-pane design: a nested keyword tree on the left, a feedback-records pane on the right.

All Hub access stays funnelled through the `@formbricks/hub` **SDK** via the server-only `modules/hub/service.ts` — the new routes never call the Hub API directly.

## What changed

**Server — new v3 routes** (`app/api/v3/unify-feedback/taxonomy/*`, session-auth), each proxying the existing Hub service fns:
- `GET fields`, `GET state`, `GET runs/[runId]`, `POST runs`, `GET nodes/[nodeId]/records`, `PATCH`/`DELETE nodes/[nodeId]`.
- `lib/access.ts` (`requireUnifyDirectoryAccess`) reproduces the legacy actions' guard exactly — org owner/manager **or** workspace-team permission, **plus** the `feedbackDirectories` entitlement and directory-belongs-to-workspace checks that v3 workspace auth doesn't do on its own (no access widening).
- The legacy `next-safe-action` actions + the 1,212-line preview component are deleted.

**Client** — scoped `QueryClientProvider` → container → controls + gate + two-pane display, driven by TanStack Query hooks over a thin `lib/api-client.ts`:
- **750 gate** kept; **embedding gate softened** — hard-block only when >50 records are pending embedding, otherwise the UI renders with an inline progress banner (RQ `refetchInterval`, replacing the 1s `setTimeout`).
- **Nested collapsible tree**; clicking a topic selects it (loads its records) *and* expands it. **View/Edit** toggle lives in the Taxonomy card header; edit mode supports inline **rename** + **remove** (confirm dialog, optimistic + rollback). **Add is intentionally absent** (Hub has no create-node endpoint).
- **Records pane** shows a capped ~100-record sample per topic. Each card: consolidated-language text (+ original in tooltip), and a unified `Label: value` meta row — **Source** (truncated, full name in tooltip), **Type**, **Sentiment** (real), **Emotion** (placeholder), user, date.
- **Record-count placeholder** (`XXX` in a light-gray box) on every tree row and in the records header, as the single seam to swap once the Hub exposes a per-node total.

## ⚠️ Required Hub-side follow-ups (not in this PR)

These are needed to complete the feature; the UI is built to accept them with minimal change:
1. **Per-node record total** — `nodes.listRecords` returns a capped sample with no `total`, and `feedbackRecords.list` can't be scoped to a node/cluster. Add a `total` to `nodes.listRecords` (works for any node) **or** a `cluster_id`/`node_id` filter on `feedbackRecords.list` (enables full leaf drill-in + count). Until then the count renders as an `XXX` placeholder.
2. **Enrichment population** — `sentiment`/`sentiment_score` and `value_text_translated`/`translation_lang_key` exist in the SDK type but are empty until Hub enrichment runs; **`emotion` doesn't exist** in the model at all (rendered as a placeholder).
3. **Create-node endpoint** — required to add "add topic/subtopic" to edit mode.

## i18n

New keys are in `en-US` and the 14 other locales carry **English placeholders** (`i18n:validate` is green, lockfile rebuilt). Run `pnpm i18n:web:generate` with `LINGODOTDEV_API_KEY` to produce real translations before release.

## Verification

- ✅ `turbo run typecheck --filter=@formbricks/web` — clean.
- ✅ `pnpm test` — 47 unit tests (route access-parity + error mapping, gate logic, query cache, api-client, optimistic-remove + rollback, fields hook).
- ✅ `pnpm i18n:validate` — green.
- ✅ `pnpm lint` — clean on touched files.
- ✅  live end-to-end pass (needs a running app + external Hub + a ≥750-record dataset).


pandeymangg:
@coderabbitai review

pandeymangg (approved):

Issue #3344, opened 2024-10-09
[BUG] Broken Redirects on `Developer Docs` -> `Contributing` Page


naineel1209:
### Issue Summary

Several broken redirect links were identified on the `Developer Docs` page. These links currently point to incorrect paths, causing users to get 404 error.

### Steps to Reproduce

1. **Developer Docs -> Overview Page Redirect**:
   On the page `https://formbricks.com/docs/developer-docs/overview`, under the section `Contributing to Formbricks`, the redirect link `/developer-docs/contributing` is broken.

2. **Developer Docs -> Contributing Section Redirects**:
   On the page `https://formbricks.com/docs/developer-docs/contributing/get-started`, under the section `Setup Dev Environments`, the redirects for Gitpod and GitHub Codespaces are broken.

### Expected behavior

1. **Developer Docs -> Overview Page Redirect**:
    `/developer-docs/contributing` should point to `/developer-docs/contributing/get-started`.

2. **Developer Docs -> Contributing Section Redirects**:
   - For Gitpod: The link `/docs/developer-docs/contributing/gitpod` should be updated to `/developer-docs/contributing/gitpod`.
   - For GitHub Codespaces: The link `/docs/developer-docs/contributing/codespaces` should be updated to `/developer-docs/contributing/codespaces`.

### Other information

These URL issues occur because the `basePath` is set to `/docs`, and the current links unnecessarily include `/docs` in the redirect path, leading to broken routes.

### Screenshots

1. For Overview Page:
![image](https://github.com/user-attachments/assets/68890631-ca70-4151-baa9-0b9459e2a2d1)

2. For Getting Started Page:
![image](https://github.com/user-attachments/assets/6ab9e6fa-fa15-495f-afcc-cd59a374284f)

### Environment

- [X] Formbricks Cloud (app.formbricks.com)
- [ ] Self-hosted Formbricks

### Desktop (please complete the following information)

```markdown
- OS: Windows
- Browser: Brave
```

jobenjada:
/award 150 thanks for spotting these!

Who is up to fix them for another 150?

adityadeshlahre:
> /award 150 thanks for spotting these!
> 
> Who is up to fix them for another 150?

he solved this himself here 
- #3343

PR #8451, opened 2026-07-07
feat: block personal email sign-ups on Formbricks Cloud

xernobyl:
## What does this PR do?

Restricts new account creation on Formbricks Cloud to company email addresses. Personal/free (gmail, yahoo, outlook…) and disposable/burner domains are blocked at sign-up time, on both the email/password and SSO flows. Self-hosted is completely unaffected, and existing accounts keep working — the check only runs where a *new* user row would be created.

Fixes ENG-1582

**How it works**

- A single server-only predicate, `isSignupEmailDomainBlocked(email, hasValidMatchingInvite)` ([`signup-email-domain.ts`](apps/web/modules/auth/lib/signup-email-domain.ts)), owns the policy: it returns `false` unless `IS_FORMBRICKS_CLOUD`, the domain is on the blocklist, and (by default) there's no valid matching invite. Both auth flows call it, so the logic lives in one place.
- **Email/password** — checked in `createUserAction` before `signUpUserSafely`, so a blocked address never creates a user. The sign-up form shows it as a field-level error under the email input.
- **SSO** — the check sits at the top of `gateSsoProvisioning`, *above* the multi-org/fresh-instance bypass (Cloud is multi-org, so the bypass would otherwise wave these through). It runs in Better Auth's `user.create.before`, so a reject rolls back inside the user+account transaction — no orphan row. A new `hooks.after` handler then rewrites Better Auth's create-failure redirect to `/auth/signup`, where the form toasts the same message (mirrors the existing `ssoRecoveryAfterHandler` pattern).
- **Invites are exempt by default** (an existing workspace vouched for the invitee), gated on a validated invite token whose email matches the address — not merely "came through the invite page". The new `SIGNUP_DOMAIN_CHECK_ON_INVITES` env flag (default off) enforces the check on invites too, as an ops kill-switch against invite-farming.

**Blocklist source** — the union of a small **curated** list of popular free providers ([`personal-email-domains.ts`](apps/web/modules/auth/lib/personal-email-domains.ts) — gmail, outlook, yahoo, icloud, proton, gmx, yandex, qq… ~25 domains) and the `disposable-email-domains` package. We start curated and expand the free list as we go; its comment points at the full ~12.9k-domain `free-email-domains` list to vendor from if we ever want exhaustive coverage (not a dependency — it re-publishes daily and ships a `postinstall`). Free providers are head-heavy (a handful covers the vast majority of real sign-ups), so curating them is cheap; disposable/burner domains are long-tail and churn constantly, so `disposable-email-domains` stays comprehensive (stable, no install scripts). Nothing leaves our infra at sign-up time — no third-party validation API.

**Things to know when reviewing**

- **Minor behavior change**: an *existing* personal-email user who re-submits the sign-up form now sees the block error instead of today's enumeration-safe pretend-success. It stays enumeration-safe (the decision is domain-only and reveals nothing about whether an account exists), and their login is untouched — so "existing users can still log in" holds where it matters. The block deliberately runs *before* the existing-user check: gating on whether the account already exists would make the sign-up response differ for registered vs. unregistered personal-domain emails, reintroducing an account-enumeration oracle. (Raised by CodeRabbit; resolved on the review thread.)
- **Translations**: the 14 non-English locales were translated by hand (the lingo key in my local `.env` is invalid, so I couldn't run `pnpm i18n`). They're register-matched per locale; happy to re-canonicalize with a `pnpm i18n` run if you'd prefer.
- The ticket describes the old NextAuth `signIn` callback; `main` is on Better Auth now, so the SSO integration is via the hook set instead. Observable behavior is the same.

## How should this be tested?

**Automated**

```
pnpm --filter @formbricks/web test modules/auth/lib/signup-email-domain.test.ts \
  modules/auth/lib/after-auth-hooks.test.ts modules/auth/signup/actions.test.ts \
  modules/auth/signup/lib/invite.test.ts modules/ee/sso/lib/sso-provisioning.test.ts \
  modules/ee/sso/lib/better-auth-hooks.test.ts modules/ee/sso/lib/sso-request-context.test.ts
pnpm i18n:validate
```

(116 unit tests cover the predicate, the Cloud gate, the invite exemption + kill-switch, the credentials action, the shared invite-match helper, the SSO gate ordering above the multi-org bypass, the after-hook ordering, and the redirect after-hook.)

**Manual, on a Cloud staging env (`IS_FORMBRICKS_CLOUD=1`)**

- Email/password sign-up with `@gmail.com` → field-level error, **no user created**.
- **Google** SSO with a gmail account → lands back on `/auth/signup` with the error toast, **no user created**.
- **GitHub** SSO with a gmail account → same. (This is the one path only exercisable end-to-end on staging — the redirect mechanism is verified against Better Auth's callback source + the shipped `ssoRecoveryAfter` analogy, but the live OAuth round-trip is worth confirming.)
- Invite flow (credentials *and* SSO) using a gmail address whose invite email matches → **succeeds** (exemption).
- Set `SIGNUP_DOMAIN_CHECK_ON_INVITES=1` → the same invite is now **blocked**.
- Company-domain sign-up (both paths) → proceeds as today.
- Existing gmail user → can still **log in**.

**Self-hosted** — with `IS_FORMBRICKS_CLOUD` unset, every path above is bypassed (a unit test pins this; one manual sign-up confirms).

### Required

- [x] Ran `pnpm build` (green in CI: Build Formbricks-web)
- [x] Checked for warnings, there are none (Linters green in CI)

### Appreciated


xernobyl:
### Follow-up note: Better Auth plugin / simplification options (researched, not adopting in this PR)

Checked whether a Better Auth plugin could replace the hand-rolled pieces here.

**Blocklist — no plugin fits, and our approach is the recommended one.** The only real candidate, [`better-auth-harmony`](https://github.com/GeKorm/better-auth-harmony), blocks *disposable* domains only (Mailchecker) and deliberately leaves free providers (gmail/yahoo/…) allowed — which is our primary target — and its docs cover email/password only, not OAuth. Better Auth's own docs recommend exactly the `databaseHooks.user.create.before` + `APIError` pattern we use ([hooks](https://better-auth.com/docs/concepts/hooks), [discussion #3344](https://github.com/better-auth/better-auth/discussions/3344)), so we're on the intended path. Keeping the curated list + hook.

**SSO — no plugin, but there's a possible simplification for later.** Better Auth redirects OAuth-callback hook rejections to `errorCallbackURL` with the thrown error code. So instead of the current "return `false` → stash the reason in `sso-request-context` → read it in the after-hook", we could throw an `APIError` carrying our code and let it land in the redirect URL directly, dropping the request-context stash. Two reasons I didn't do it here:

1. The after-hook is still needed regardless — all SSO buttons use `errorCallbackURL: "/auth/login"`, and we can't repoint it at `/auth/signup` without hijacking the `account_not_linked` login-recovery redirect.
2. Whether a thrown `APIError`'s code actually propagates into the OAuth error redirect (vs BA's generic `unable_to_create_user`) is undocumented and the shipped dist is minified — I couldn't confirm it by reading. The current return-`false` + stash path is the one I verified (it mirrors the already-shipped `ssoRecoveryAfter`).

Net: leaving the verified approach in place. If we want to trim the `sso-request-context` stash down the line, it should go behind an integration test that proves the thrown-code-in-redirect behaviour first.

apps/web/modules/auth/signup/actions.ts
```diff
@@ -247,6 +268,16 @@ export const createUserAction = actionClient.inputSchema(ZCreateUserAction).acti
...
+    // Formbricks Cloud only: reject personal/free/disposable email domains before any user is created.
+    // Invited users are exempt unless SIGNUP_DOMAIN_CHECK_ON_INVITES is enabled.
+    if (
+      await isSignupEmailDomainBlocked(parsedInput.email, () =>
+        hasValidMatchingInvite(parsedInput.email, parsedInput.inviteToken)
+      )
+    ) {
+      throw new InvalidInputError(SIGNUP_EMAIL_DOMAIN_BLOCKED_ERROR_CODE);
+    }
+
```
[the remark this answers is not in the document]
xernobyl: Looked at this closely — keeping the current ordering (block before the existing-user check) on purpose, because the suggested `!existingUser &&` guard would reintroduce account enumeration.

The block is applied to every personal-domain address regardless of whether an account exists, so the signup response is identical for a registered vs. unregistered gmail (both get the domain error). Adding the `!existingUser` guard splits those: a *registered* personal-domain email would fall through to `signUpUserSafely`'s enumeration-safe "already existed → success", while an *unregistered* one gets the domain error — which is an existence oracle for any personal-domain address. Enumeration-safety is a deliberate existing property here (that's what the "already exists → success" handling in `signUpUserSafely` is for), so I'd rather not regress it for the sign-up path.

The "existing accounts continue to work" guarantee is about **login**, which this doesn't touch — the check lives only in the sign-up action, so an existing personal-domain user still logs in normally. The one observable change is that if they re-submit the *sign-up* form they'll get "use your company email" instead of the pretend-success; that's the safer trade-off and it's called out in the PR description.

On the suggested regression test — `actions.test.ts` already covers it: the "personal email domain block" suite's `beforeEach` mocks `getUserByEmail` to return an existing user, and the block test still asserts the rejection with no `signUpEmail` call, i.e. an existing personal-domain account is (intentionally) still blocked on re-submit.

apps/web/modules/auth/signup/actions.ts
```diff
@@ -247,6 +268,16 @@ export const createUserAction = actionClient.inputSchema(ZCreateUserAction).acti
     await applyIPRateLimit(rateLimitConfigs.auth.signup);
     await verifyTurnstileIfConfigured(parsedInput.turnstileToken);
 
+    // Formbricks Cloud only: reject personal/free/disposable email domains before any user is created.
+    // Invited users are exempt unless SIGNUP_DOMAIN_CHECK_ON_INVITES is enabled.
+    if (
+      await isSignupEmailDomainBlocked(parsedInput.email, () =>
```
Dhruwang: **P1 (blocker): this is the *only* place the email/password domain block runs, but it isn't the account-creation chokepoint.**

Better Auth's native `POST /api/auth/sign-up/email` is live (`emailAndPassword.enabled=true`; only `/token` is in `disabledPaths`) and `databaseHooks.user.create.before` early-returns for credential signups (`if (!identityProvider) return`). So a scripted POST straight to that endpoint with `@gmail.com`/`@mailsac.com` creates a Cloud `User` row, never touching this action. Confirmed by manual test. `requireEmailVerification` + `trustedOrigins` raise the bar but a server-side POST forging `Origin` defeats both.

Suggest enforcing the check at the chokepoint too: add a `user.create.before` branch keyed on `providerId === "credential"` that calls `isSignupEmailDomainBlocked` and returns `false` (rollback-safe, keeps it enumeration-safe). That closes personal + disposable for both flows in one place.

apps/web/modules/auth/signup/actions.ts
```diff
@@ -69,6 +74,22 @@ async function verifyTurnstileIfConfigured(turnstileToken: string | undefined):
   }
 }
 
+/**
+ * Invite-exemption check for the personal-email sign-up block: resolves true only when `inviteToken`
+ * is a valid, non-expired invite whose email matches the address being registered. Validated here,
+ * before any user is created, rather than relying on the post-creation handleInviteAcceptance step.
+ */
+async function hasValidMatchingInvite(email: string, inviteToken: string | undefined): Promise<boolean> {
```
Dhruwang: **P2:** the invite-validation trio (verify token -> email match -> `getIsValidInviteToken`) is duplicated here and inside `validateSsoInviteToken` in `sso-provisioning.ts`, and the two already diverge on case handling (see the SSO comment). Suggest extracting a shared `isValidMatchingInvite(inviteToken, email): Promise<boolean>` into `auth/signup/lib/invite.ts`; this action calls it directly, and the SSO side delegates after URL-parsing the token.

apps/web/modules/auth/signup/actions.ts
```diff
@@ -69,6 +74,22 @@ async function verifyTurnstileIfConfigured(turnstileToken: string | undefined):
...
+  if (!inviteToken) return false;
+  try {
+    const { inviteId, email: invitedEmail } = verifyInviteToken(inviteToken);
+    if (invitedEmail.toLowerCase() !== email.toLowerCase()) return false;
+    return await getIsValidInviteToken(inviteId);
+  } catch {
```
Dhruwang: **P3:** this bare `catch` (and the equivalents in `validateSsoInviteToken` and `setSsoSignupRejectReason`) all fail in the safe direction, but swallow silently. A transient DB timeout on `getIsValidInviteToken` here is indistinguishable from a genuinely invalid token. Add a `logger.warn` before returning so it's observable in prod.

apps/web/modules/ee/sso/lib/sso-provisioning.ts
```diff
@@ -71,6 +73,19 @@ export const gateSsoProvisioning = async ({
...
   callbackUrl: string;
 }): Promise<TSsoProvisioningDecision> => {
+  // Formbricks Cloud only: block SSO sign-ups from personal/free/disposable email domains, before any
+  // org resolution. Placed above the multi-org / fresh-instance bypass below (Cloud is multi-org, so
+  // the bypass would otherwise let these through). Invited users — a valid token whose email matches —
+  // are exempt unless SIGNUP_DOMAIN_CHECK_ON_INVITES is set; the exemption reuses the same invite
+  // validation as the gate below and runs lazily, only when the domain is actually blocked.
+  const isDomainBlocked = await isSignupEmailDomainBlocked(
+    email,
+    async () => (await validateSsoInviteToken(email, callbackUrl)) === null
```
Dhruwang, line 91: **P2 (two things on this line):**

1. **Case-sensitivity divergence.** `validateSsoInviteToken` compares `inviteEmail !== email` raw, whereas the email/password path lowercases both sides (`actions.ts` `hasValidMatchingInvite`). A gmail invitee whose IdP asserts a differently-cased address is exempted on credentials but hard-blocked on SSO. Normalize both sides.
2. **Readability.** `(await validateSsoInviteToken(...)) === null` as the "has valid invite" predicate is a double-negative that needs out-of-band knowledge of the null-on-success convention. A boolean-returning helper (from the shared extraction suggested on `actions.ts`) would read directly.

Both are resolved by the shared `isValidMatchingInvite` helper.

apps/web/modules/auth/lib/auth.ts
```diff
@@ -244,6 +245,9 @@ export const auth = betterAuth({
     after: createAuthMiddleware(async (ctx) => {
       await ssoRecoveryAfterHandler(ctx);
       await auditFailedAuthAfter(ctx);
+      // Personal-email SSO rejection → redirect to /auth/signup with a toast. Runs last so the
+      // failed-auth audit above still records the attempt; only fires on our stashed reject reason.
+      await blockedSignupDomainRedirectAfterHandler(ctx);
```
Dhruwang: **P2 (test gap):** this ordering is load-bearing. `blockedSignupDomainRedirectAfterHandler` throws (`ctx.redirect`), so if it were composed before `auditFailedAuthAfter`, the failed-auth audit would be skipped on every personal-email SSO rejection. Only a prose comment guards it today — please pin it with a test asserting the audit runs before the redirect throws.

apps/web/modules/auth/lib/signup-email-domain.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Union of two sources:
+ *  - Free/personal providers (gmail, yahoo, outlook, …): a small curated list in
+ *    ./personal-email-domains.ts that we expand as we go (its comment points at the full ~12.9k-domain
+ *    `free-email-domains` list to vendor from if we ever want exhaustive coverage).
+ *  - Disposable/burner providers (mailinator, 10minutemail, …): the `disposable-email-domains` npm
+ *    package — kept comprehensive because burner domains are long-tail and churn constantly.
+ *
+ * Server-only: the disposable list alone is ~121k domains and must never be pulled into the client bundle.
+ */
+const blockedEmailDomains = new Set<string>(
```
Dhruwang, line 18: **P2 (edge case):** the normalization does `trim().toLowerCase()` but doesn't strip a trailing dot, so `gmail.com.` misses the Set. Harmless on the credentials path (zod `z.email()` rejects it upstream) but the SSO email is *not* zod-validated before reaching the gate. Low real-world risk (IdPs won't assert `gmail.com.`) — add a test, and consider stripping a trailing dot here for defense in depth.

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -82,6 +83,19 @@ export const SignupForm = ({
 
   const turnstile = useTurnstile();
 
+  // An SSO sign-up rejected for a personal email domain redirects back here with ?error=<code>.
+  // Match the known code exactly (never echo the raw param) and show it at most once — the ref guards
+  // against a re-toast on locale change and React strict mode's double effect invocation in dev.
+  const oauthError = searchParams?.get("error");
+  const hasShownDomainErrorToast = useRef(false);
```
Dhruwang: **P3:** this guard latches `true` for the component's lifetime, so a second genuinely-distinct SSO domain rejection in the same mounted session is silently swallowed. Key it on the last-shown `oauthError` value rather than a permanent boolean — still dedupes the strict-mode double-invoke and locale re-renders, but re-notifies on a fresh rejection.

apps/web/modules/auth/lib/signup-email-domain.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+  });
+
+  test("is case- and whitespace-insensitive", () => {
+    expect(isBlockedEmailDomain("Test@Gmail.Com")).toBe(true);
+    expect(isBlockedEmailDomain("  TEST@YAHOO.COM  ")).toBe(true);
+    expect(isBlockedEmailDomain("carol@STRIPE.com")).toBe(false);
+  });
+
+  test("matches on the domain after the last @", () => {
```
Dhruwang, line 78: **P3 (coverage):** nice to pin two more behaviors adjacent to this test: subdomain-of-a-blocked-domain (`isBlockedEmailDomain("user@sub.gmail.com") === false`, documenting the intentional exact-match semantic) and plus-addressing (`isBlockedEmailDomain("user+tag@gmail.com") === true`, a common evasion pattern that currently works correctly via `lastIndexOf("@")`).

Dhruwang (commented):
**Review — 1 blocker, then ready-with-fixes.**

Well-tested and thoroughly documented. One issue should block merge (inline on `actions.ts`); the rest are should-fix / polish. Verified non-issues: enumeration ordering (matches the resolved CodeRabbit thread), the `=== null` invite check's correctness, AsyncLocalStorage request isolation, and after-hook hijack safety.

**P2 — not inline-able (file has no diff hunk): `.env.example`**
The new `SIGNUP_DOMAIN_CHECK_ON_INVITES` kill-switch isn't documented. Please add a commented, default-off entry beside `AUTH_SKIP_INVITE_FOR_SSO` (~line 323) with the Cloud-only note, so operators can discover it.

Inline comments follow.

Dhruwang (commented):
The PR looks good overall, but I noticed two UI/UX issues that should be addressed.

1. While temporary email sign-ups are blocked, the flow doesn't clearly communicate this to the user. A user can enter a temporary email and password, and the system even sends a verification email. However, after clicking the verification link, nothing happens. There should be clear validation or messaging earlier in the flow to let users know that temporary email addresses are not supported.

2. The same issue exists for user invitations. Although temporary email addresses are blocked, users can still send invites to them. It would be better to validate the email before sending the invite and display a clear message explaining that temporary email addresses cannot be used for invitations. This would avoid confusion and provide a better user experience.

Screenshot 2026-07-09 at 4 26 47 PM

apps/web/modules/auth/signup/actions.ts
xernobyl: Fixed in a7b6eaf0d — enforced at the chokepoint. The domain block now also runs in `user.create.before`; `createUserAction` marks a request-scoped flag (`signup-request-context`) once its policy passes, so the normal action path is not double-blocked, while a direct POST to the native `/sign-up/email` (no flag) is re-checked and blocked there. Added hook tests for both paths.

apps/web/modules/auth/signup/actions.ts
xernobyl: Done — extracted `resolveInviteMatch(inviteToken, email)` into `auth/signup/lib/invite.ts`; both `createUserAction` and `validateSsoInviteToken` use it now (the SSO wrapper maps its discriminated result to the granular reject reasons).

apps/web/modules/auth/signup/actions.ts
xernobyl: Done — the shared `resolveInviteMatch` now logs a `logger.warn` on verification failure instead of swallowing it, so a transient failure is observable.

apps/web/modules/ee/sso/lib/sso-provisioning.ts
xernobyl: Fixed — `resolveInviteMatch` normalizes both sides with `trim().toLowerCase()`, so the credentials and SSO paths no longer diverge on case. `validateSsoInviteToken` delegates to it and maps the outcome to its granular reasons.

apps/web/modules/auth/lib/auth.ts
xernobyl: Done — extracted the `hooks.after` chain into `runAfterAuthHooks` (after-auth-hooks.ts) with a test asserting the failed-auth audit runs before the redirect handler throws.

apps/web/modules/auth/lib/signup-email-domain.ts
xernobyl: Done — normalization now strips a trailing FQDN dot (`gmail.com.` maps to `gmail.com`); added a test.

apps/web/modules/auth/signup/components/signup-form.tsx
xernobyl: Done — the toast now dedupes on the last-shown error value rather than a permanent latch, so a fresh distinct rejection re-notifies while strict-mode/locale re-renders do not.

apps/web/modules/auth/lib/signup-email-domain.test.ts
xernobyl: Done — added subdomain (exact-match), plus-addressing, and trailing-dot cases here, plus `resolveInviteMatch` outcome coverage in invite.test.ts.

xernobyl:
Thanks for the review @Dhruwang 🙏

Addressed all eight comments — replies inline on each thread. Since your pass I also pushed a small polish commit on top (`6ee6aaf69`): a defensive guard on the domain predicate so an unexpected caller degrades to "allow" instead of throwing, one more free provider (`126.com`, sibling of the listed `163.com`), and a doc-accuracy fix on the `runAfterAuthHooks` ordering comment (the failed-auth audit is `/sign-in/email`-only, so the ordering is future-proofing rather than currently load-bearing).

116 unit tests + lint/format green locally. Ready for another look whenever you have a sec 🙂

Dhruwang (commented):
Hey Tiago, thanks for the changes – they look good! I did notice one potential issue, though. When I'm on a self-hosted instance and Formbricks Cloud is set to zero, I'm still unable to create an account with a blocked invalid email address. Ideally, we should only block this for Formbricks Cloud users. I can confirm that I receive the verification email, but when I try to verify that email, I'm just taken back to the login screen instead of being actually logged in. Could you please check that?

xernobyl:
@Dhruwang this one turned out to be pre-existing from the Better Auth migration (ENG-1054), not caused by this PR — `autoSignInAfterVerification` is set to `false`, so after verifying your email you land on the login screen instead of being auto-logged-in. I reproduced it on self-hosted: the account is created fine (the domain block correctly doesn't fire there), and logging in after verifying works — so if you sign in after the verify step, you're in. ✅

I'll fix the auto-login-after-verify UX on a separate PR to keep this one focused on the domain block — tracked in ENG-1746. Thanks for catching it! 🙏

jobenjada:
I merged the other one with auto-login, so after rebase this one should have the expected UX, right?

xernobyl:
Yes, doesn't affect this one, interface should work as normal. They're independent features.

Dhruwang (approved):

PR #8454, opened 2026-07-08
feat(editor): add bullet and numbered list buttons to the toolbar

jobenjada:
## What does this PR do?

The Rich Text (Lexical) editor already renders **ordered and unordered lists** — e.g. when content containing lists is pasted in — but there was no toolbar control to create them directly. The only existing UI path was the block-type dropdown, which every caller hides via `excludedToolbarItems={["blockType"]}`.

This PR exposes the missing controls by adding two dedicated toolbar buttons:

- **Bulleted list** (unordered) — `List` icon
- **Numbered list** (ordered) — `ListOrdered` icon

They're wired to the editor's **already-existing** `formatBulletList` / `formatNumberedList` handlers (the same ones the hidden dropdown used), toggle a list on/off, and show an active state when the caret is inside a list. No changes to the underlying list rendering pipeline were needed — `ListPlugin` / `ListNode` were already configured, which is why paste already worked.

The buttons render by default alongside bold/italic/underline and remain hideable per-caller via `excludedToolbarItems` (keys `"bulletList"` / `"orderedList"`), consistent with the existing toolbar-item pattern.

**ENG-1612**

### Notes for reviewers

- **`disableLists` is intentionally left untouched.** That prop only ever filtered the markdown auto-shortcut transformers (`- ` / `1. `); it never gated paste or `ListPlugin`. The new buttons are an explicit control and are gated only by `excludedToolbarItems`, matching how the block-type list UI was already controlled.
- **i18n:** the two tooltip keys (`workspace.surveys.edit.bulleted_list` / `numbered_list`) were added to all 15 web-app locales plus `i18n.lock` (MD5s verified against `object-hash`). `pnpm scan-translations` passes with 0 missing / 0 unused / all locales complete. Translations can be re-canonicalized via lingo.dev if preferred.

## Additional tweaks (follow-up)

A few small survey-editor/rendering tweaks were bundled into this branch:

- **Survey-list "Change status" reorder** — moved the `Change status` submenu to position 3 in the survey-list dropdown (now **Edit → Duplicate → Change status → Copy to… → Preview → Copy link → Delete**), so the most-used actions sit at the top. `survey-dropdown-menu.tsx`.
- **Stronger bold vs non-bold contrast in rendered surveys** — bold rich text (`<strong>`/`<b>`) inside a question headline or description was visually indistinguishable from normal text. Root cause: the headline/description font-weight rules (`#fbjs .label-headline *`, injected by `styles.ts` and defined in `survey-ui/globals.css`) target *every* descendant, flattening bold spans to the base weight (600). Fixes:
  - Default headline weight lowered one step: `elementHeadlineFontWeight` **600 → 500** (`STYLE_DEFAULTS` + the CSS-var fallback). Existing surveys with a customized weight keep their value.
  - Added a higher-specificity `!important` rule so `<strong>`/`<b>` inside `.label-headline` / `.label-description` / `.text-subheading` / `.htmlbody` always render at a true **700**, overriding the flattening rules — added to both the rendered-survey CSS (`surveys/global.css`) and the editor/storybook CSS (`survey-ui/globals.css`).
  - `preflight.css` `strong`/`b`: `bolder` → `bold` for a predictable fixed weight.
  - Result (verified via headless computed styles): headline **non-bold 500 / bold 700**, description **non-bold 400 / bold 700**.
- **Copy tweak** — "Changes **will** lead to inconsistencies" → "Changes **may** lead to inconsistencies" (`caution_text`), updated across all 15 locales + `i18n.lock`.

## How should this be tested?

1. Open any Rich Text editor in the app — e.g. a survey element's content field (headline/description), the **Manage translations** rich-text inputs, or a survey **follow-up email body**.
2. Confirm two new toolbar buttons appear after **Underline**: bulleted-list and numbered-list icons (with tooltips).
3. Put the cursor on a line (or select several), click **Bulleted list** → the line(s) become a `<ul>`; click again → toggles back to a paragraph. The button is highlighted while the caret is inside the list.
4. Repeat with **Numbered list** → `<ol>`.
5. Confirm pasting content that contains lists still renders lists as before (unchanged).
6. **Bold contrast:** in a survey preview, bold some text inside a question headline → it should now read clearly heavier (700) than the surrounding text (500).
7. **Status menu:** open the ⋮ menu on a survey in the survey list → `Change status` is the third item.

## Preview

image

image


itsjavi (approved):
LGTM but this reveals a wider issue in the rich text editor: we cannot use tabs (focus is lost on tab), therefore we cannot create nested lists.

Is this intended? If we should fix it can you create a follow up Linear ticket?

jobenjada:
@itsjavi created this follow up: https://linear.app/formbricks/issue/ENG-1747/fix-tabs-in-rich-text-editor

PR #8457, opened 2026-07-08
fix(surveys): return 400 instead of 500 for invalid survey image URLs [ENG-1329]

pandeymangg:
## What does this PR do?

Fixes **ENG-1329** / Sentry **FORMBRICKS-146**.

Creating a survey with a Picture Selection question whose choice `imageUrl` fails `isValidImageFile` threw a plain `Error` that propagated uncaught and exited as an HTTP **500**, paging Sentry — when it's really invalid user input that should be a clean **400**.

### Root cause

- `validateMediaAndPrepareBlocks` (`apps/web/lib/survey/utils.ts`) threw a plain `Error`.
- `createSurvey`'s catch only converts Prisma errors; the `POST /api/v1/management/surveys` catch only mapped `DatabaseError`, so everything else re-threw → `withV1ApiWrapper` returned **500** → `reportApiError` fires Sentry only on `status >= 500`.
- The legacy `questions` path had the same problem: `checkForInvalidImagesInQuestions` already threw `InvalidInputError`, but the POST catch never mapped it either → also a 500.
- `PUT /[surveyId]` was already returning 400 (it throws `InvalidInputError`, mapped by `handleErrorResponse`).

### The fix

1. **Media validators now emit `InvalidInputError` in the `Result`** (`validateChoiceImage`, `validateElement`, `checkForInvalidMediaInBlocks`) instead of a plain `Error`. Both throw boundaries (`validateMediaAndPrepareBlocks` for create, `updateSurveyInternal` for update) simply forward `validation.error` — single source of truth for the error type.
2. **POST handler catch maps `InvalidInputError` → `badRequestResponse` (400)**, fixing both the blocks and legacy-questions paths.
3. Because the **v3 create path** already maps `InvalidInputError` → 400, throwing the correct type fixes v3 too (it previously fell through to its own 500 branch). One change, all paths consistent.
4. Side benefit: an invalid **video URL** in a block is now also a proper `InvalidInputError` (was a plain `Error` — a latent version of the same 500 bug).

### Allowlist decision (required by the ticket)

The picture-choice renderer uses a plain `` tag (`packages/survey-ui` `PictureSelect`), **not** `next/image` — so the browser, not the allowlist, is the real display constraint.

- **Widened** the image allowlist from `png, jpeg, jpg, webp, heic` to also include **`gif`, `avif`, `bmp`** (all decode natively in ``). Exported as `IMAGE_FILE_EXTENSIONS`.
- **`svg` deliberately excluded** — unsanitized SVGs are an XSS vector; only allow if a sanitizer is added.
- **Extension-less external/CDN URLs** (Unsplash, signed URLs, `?format=auto`) are **still rejected** — we can't verify content-type synchronously — but the 400 message is now actionable: names the offending choice/question/block, lists the supported formats, and tells the user to host the image on Formbricks or use a direct image URL.
- Note: the editor's upload widgets still restrict *uploads* to `png/jpeg/jpg/webp/heic` (a separate file-upload security surface); the server-side URL validator is intentionally a superset. Left untouched.

### Automated

- `apps/web/lib/survey/utils.test.ts` — new `validateMediaAndPrepareBlocks` suite asserts it throws `InvalidInputError` (not a plain `Error`) and that the message names the offending choice/question/block + supported formats.
- `apps/web/modules/storage/utils.test.ts` — `isValidImageFile` now covers `gif`/`avif`/`bmp` (accepted) and `svg` (rejected).
- Verified locally: unit tests (166 pass), `tsc --noEmit`, and ESLint all green.

### Manual (curl against a running instance, API key with write access)

```bash
export FB_URL="http://localhost:3000"; export FB_KEY="<api-key>"; export FB_ENV="<environmentId-or-workspaceId>"

# 1) Unparseable/extension-less choice image -> expect HTTP 400 (was 500, no Sentry)
curl -i -X POST "$FB_URL/api/v1/management/surveys" \
  -H "x-api-key: $FB_KEY" -H "Content-Type: application/json" \
  -d '{"environmentId":"'"$FB_ENV"'","name":"ENG-1329 repro","type":"link",
    "questions":[{"id":"q1","type":"pictureSelection","headline":{"default":"Design Vote"},
    "required":true,"allowMulti":false,
    "choices":[{"id":"c1","imageUrl":"https://images.unsplash.com/photo-12345"},
               {"id":"c2","imageUrl":"https://example.com/valid.png"}]}]}'

# 2) .gif now accepted (200), .svg still rejected (400)
```

Expected: 400 with `"Invalid image URL in choice 1 of question 1 of block \"Design Vote\". The URL must end in a supported image extension (png, jpeg, jpg, webp, heic, gif, avif, bmp). ..."`. Valid images still create the survey unchanged.

### Required

- [ ] Ran `pnpm build`

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (no UI changes)

### DIFF
--- a/apps/web/lib/survey/utils.ts
+++ b/apps/web/lib/survey/utils.ts
@@ -11,7 +11,16 @@ import {
 } from "@formbricks/types/surveys/elements";
 import { TSurvey, TSurveyQuestion, TSurveyQuestionTypeEnum } from "@formbricks/types/surveys/types";
 import { isValidVideoUrl } from "@/lib/utils/video-upload";
-import { isValidImageFile } from "@/modules/storage/utils";
+import { IMAGE_FILE_EXTENSIONS, isValidImageFile } from "@/modules/storage/utils";
+
+/**
+ * Actionable hint appended to invalid-image messages. Names the supported formats and explains why
+ * extension-less external/CDN URLs (e.g. Unsplash, signed URLs, `?format=auto`) are rejected, since
+ * validation keys off the URL's file extension.
+ */
+const INVALID_IMAGE_HINT = `The URL must end in a supported image extension (${IMAGE_FILE_EXTENSIONS.join(
+  ", "
+)}). For CDN or extension-less links, upload the image to Formbricks or use a direct image URL.`;
 
 export const transformPrismaSurvey = <T extends TSurvey | TJsWorkspaceStateSurvey>(surveyPrisma: any): T => {
   let segment: TSegment | null = null;
@@ -70,18 +79,18 @@ export const checkForInvalidImagesInQuestions = (questions: TSurveyQuestion[]) =
  * @param choiceIdx - Index of the choice for error reporting
  * @param elementIdx - Index of the element for error reporting
  * @param blockName - Block name for error reporting
- * @returns Result with void data on success or Error on failure
+ * @returns Result with void data on success or InvalidInputError on failure
  */
 const validateChoiceImage = (
   choice: TSurveyPictureChoice,
   choiceIdx: number,
   elementIdx: number,
   blockName: string
-): Result<void, Error> => {
+): Result<void, InvalidInputError> => {
   if (choice.imageUrl && !isValidImageFile(choice.imageUrl)) {
     return err(
-      new Error(
-        `Invalid image URL in choice ${choiceIdx + 1} of question ${elementIdx + 1} of block "${blockName}"`
+      new InvalidInputError(
+        `Invalid image URL in choice ${choiceIdx + 1} of question ${elementIdx + 1} of block "${blockName}". ${INVALID_IMAGE_HINT}`
       )
     );
   }
@@ -94,13 +103,13 @@ const validateChoiceImage = (
  * @param element - Element with choices to validate
  * @param elementIdx - Index of the element for error reporting
  * @param blockName - Block name for error reporting
- * @returns Result with void data on success or Error on failure
+ * @returns Result with void data on success or InvalidInputError on failure
  */
 const validatePictureSelectionChoiceImages = (
   element: TSurveyElement,
   elementIdx: number,
   blockName: string
-): Result<void, Error> => {
+): Result<void, InvalidInputError> => {
   // Only validate choices for picture selection elements
   if (element.type !== TSurveyElementTypeEnum.PictureSelection) {
     return ok(undefined);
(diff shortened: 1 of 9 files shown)


apps/web/lib/survey/utils.ts
```diff
@@ -11,7 +11,16 @@ import {
...
 import { isValidVideoUrl } from "@/lib/utils/video-upload";
-import { isValidImageFile } from "@/modules/storage/utils";
+import { IMAGE_FILE_EXTENSIONS, isValidImageFile } from "@/modules/storage/utils";
+
+/**
+ * Actionable hint appended to invalid-image messages. Names the supported formats and explains why
+ * extension-less external/CDN URLs (e.g. Unsplash, signed URLs, `?format=auto`) are rejected, since
+ * validation keys off the URL's file extension.
+ */
+const INVALID_IMAGE_HINT = `The URL must end in a supported image extension (${IMAGE_FILE_EXTENSIONS.join(
```
Dhruwang, line 22: The legacy `checkForInvalidImagesInQuestions` path (line 54) doesn't append this hint, so those errors stay terse. Consider reusing `INVALID_IMAGE_HINT` there too for consistent messages.

apps/web/modules/storage/utils.ts
```diff
@@ -322,15 +322,24 @@ export const validateClientFileUploads = ({
...
 
+/**
+ * Image file extensions accepted for survey media (question/element/choice imageUrl).
+ *
+ * These are all rendered client-side via a plain `<img src>` tag (see
+ * `packages/survey-ui` PictureSelect), so any raster format the browser can decode is safe to
+ * allow. `svg` is deliberately excluded: SVGs can embed scripts and we do not sanitize them, so
+ * allowing them would be an XSS vector.
+ */
+export const IMAGE_FILE_EXTENSIONS = ["png", "jpeg", "jpg", "webp", "heic", "gif", "avif", "bmp"] as const;
```
Dhruwang: `gif/avif/bmp` here diverge from the upload allowlist (`ALLOWED_FILE_EXTENSIONS_TUPLE`) and the editor picker, which reject them. The 400 hint says "upload to Formbricks" in these formats, but the upload then fails. Either widen the upload list too, or reword the hint to "externally-hosted URL only". Prefer deriving both from one shared const so they can't drift.

Dhruwang (approved):

PR #8459, opened 2026-07-08
feat(surveys): generate 20 AI example responses instead of 10

jobenjada:
## What

Bumps `EXAMPLE_RESPONSE_COUNT` from **10 → 20** in `example-responses.ts`, so the **Generate example responses** action seeds a survey with a richer synthetic dataset for the summary dashboard.

## Why

10 rows leaves the analysis dashboard looking sparse (thin charts, few open-text quotes). Doubling to 20 gives a fuller, more representative preview of what a real response set looks like, at a modest extra LLM cost for the open-text answers.

## Changes

- `EXAMPLE_RESPONSE_COUNT = 20` (single constant; `EXAMPLE_IMPRESSION_ONLY_COUNT` unchanged at 6).
- Updated the deterministic-output unit tests to the new count:
  - drop-off split → **16 finished / 4 dropped** (`ceil(20 * 0.2)`)
  - lumpy single-select distribution → `[8, 6, 2, 2, 2]`
  - open-text batching now produces **4 chunks** (still ≤ 20 requested answers/chunk); the mid-batch failure path is unchanged
  - first-finished response maps to fixture index 4 (AU address / Jordan Lee contact)

## Note for reviewers — synthetic dashboard rates shift

Displays = responses + 6 impression-only, so the summary metrics move:

| Metric | Before (10) | After (20) |
| --- | --- | --- |
| Starts (`responses / displays`) | 10/16 ≈ **63%** | 20/26 ≈ **77%** |
| Completed (`finished / displays`) | 8/16 ≈ **50%** | 16/26 ≈ **62%** |

The `EXAMPLE_IMPRESSION_ONLY_COUNT` comment says the completion rate "lands around 62%" — after this change the dashboard's **Completed** card actually reads ~62%. Left the impression count (6) and the comment as-is. If we'd rather hold the *starts* rate near 62%, bumping impressions to ~12 would do it — happy to add that if preferred.

## Testing

`vitest` for `example-responses.test.ts` — 19/19 passing.

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/example-responses.test.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/example-responses.test.ts
@@ -216,8 +216,8 @@ describe("generateExampleResponseDataset", () => {
     expect(result.responses).toHaveLength(EXAMPLE_RESPONSE_COUNT);
     expect(result.displays).toHaveLength(EXAMPLE_IMPRESSION_ONLY_COUNT);
     expect(result.tagName).toBe(EXAMPLE_AI_GENERATED_TAG_NAME);
-    expect(Object.values(roleCounts)).toContain(4);
-    expect(Object.values(roleCounts)).not.toEqual([2, 2, 2, 2, 2]);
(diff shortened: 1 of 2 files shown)


Dhruwang (approved):
Thank you for the PR. Looks good. 🚀.

PR #8464, opened 2026-07-09
feat(charts): language-stable option grouping via value_id + canonical value_text [ENG-1673]

Dhruwang:
## What

Choice answers were ingested with the label in the response language, so grouping/filtering a choice question split one option into a category per language (Gender → `Male`, `Female`, `ذكر`, `أنثى`). This PR makes choice grouping **language-stable** and adds **stable option identity** end-to-end:

- **`value_text`** is stored as the **canonical default-language label** at ingestion, so grouping by **Value (Text)** consolidates an option across languages with no cross-system join.
- **`value_id`** (the stable survey choice id) is stored per option, so grouping by **Value (Option)** survives label edits and language changes; ids are resolved back to human labels for display via a server-built `optionLabels` map (from the survey's choices).

Both single- and multi-select now store **one record per selected option** with its own `value_id`, so the two choice types behave identically.

> **History (for reviewers who saw earlier states):** this PR iterated a few times — id-grouping with query-time resolution → canonical `value_text` only → the current design, which keeps canonical `value_text` **and** restores per-option `value_id` grouping with label resolution (hence the branch name `option-id-charts`). The user's Group By choice (Value Text vs Value Option) is now always respected — no silent dimension rewriting in either direction.

## Changes

**Ingestion** (`apps/web/lib/feedback-source/transform.ts`)
- Single-select & matrix: store the choice/column **default-language label** as `value_text`, plus its `value_id`.
- **Multi-select: split into one record per selected option**, each with its own `value_id` (previously one joined, per-language string). Split records share `field_group_id` (the question) and `submission_id` (the respondent) — same pattern as matrix/ranking.

**Charts**
- `resolveOptionGrouping` (extracted to `charts/lib/option-grouping.ts` so the query action and the dashboard page share it) attaches a `value_id → default-label` map for resolved choice questions and **no longer rewrites the grouping dimension**. Removed the multi-select `splitMultiSelectRows` shim — unnecessary now that ingestion emits per-option rows.
- **Dashboard charts** now resolve option labels the same way the builder does (saved widgets previously rendered raw `value_id`s).
- **Filter / Group By field list reordered by relevance** — **Question** is now the first option and the default filter field; then answer values → AI enrichment → source → time/language → identifiers & Hub meta.
- Kept from earlier work on the branch: searchable field combobox, Source ID value pick-list, dropdown height cap, only continuous measures are filterable, emotions pick-list, and duplicate-a-chart on dashboards.
- **Cube schema**: `valueId` and `fieldId` dimensions are present.

**Unify Feedback — record drawer is now view-only**
- Removed manual create/edit of feedback records: the **"Add feedback record"** button + create form, inline editing, and save/discard/metadata editing. The drawer is a read-only viewer (all fields disabled; footer = **Delete + Close**). CSV import stays (toolbar dropdown relabeled "Import feedback").
- The **Value ID** field is always shown (even when empty), paired 50/50 with Value (Text).
- Removed the now-dead `createFeedbackRecordAction` / `updateFeedbackRecordAction` and their schemas/types + `getCreateDefaults`.

**i18n**
- Renamed the drawer label **"Option ID" → "Value ID"**; removed 12 keys orphaned by the drawer cleanup. `pnpm i18n:validate` is green (0 unused / 0 missing, lockfile in sync).

## How to test

1. Multi-language survey with a single- and a multi-select question, connected to a Hub directory; submit responses in several languages.
2. Chart with measure **Count**, dimension **Value (Text)** → one category per option regardless of response language, labelled in the default language.
3. Switch the dimension to **Value (Option)** → it stays on Value (Option) (no auto-swap back to Text), and slices are labelled via the option map. Same for a multi-select question.
4. Open the same chart on a **dashboard** → labels render (not raw `value_id`s).
5. Open a feedback record → drawer is read-only (Delete + Close only); there is no "Add feedback record" button in the table.

## Notes / follow-ups

- **Per-option `value_id` applies to newly-ingested data.** Not released yet, so no backfill is planned; existing (pre-split) multi-select rows aren't migrated.
- `effectiveQuery` returned by `executeQueryAction` is now always identical to the input query (nothing rewrites); the plumbing is left in place and can be simplified in a follow-up.
- The multi-select split supersedes the earlier per-language handling (ENG-1702).
- If we ever want single-select's old auto-consolidation back, it should be an explicit opt-in, not a silent rewrite.


jobenjada:
image


Hey @Dhruwang —  I pushed a round of changes on top of your ENG-1673 branch. Wanted to walk you through them since a couple shift the approach you started with. Happy to adjust any of this.

## TL;DR on the direction change

Your branch made single-select grouping language-stable by canonicalizing `value_text` and **rewriting the Group By dimension** server-side (`valueText → valueId` for single-select, `valueId → valueText` + row-splitting for multi-select). That kinda worked, but it silently overrode the user's Group By choice — pick "Value (Text)" and it popped to "Value (Option)", and the reverse for multi-select. It also only handled single-select cleanly because multi-select stored a joined string with no per-option id.

We flipped the model: **store a stable per-option `value_id` at ingestion for *both* single- and multi-select**, then let the user group by whatever dimension they pick — no rewrite. Labels are still resolved for display via an `optionLabels` map. Net result: the same language-stable grouping, but explicit and without the surprise swaps.

## What changed

**1. Ingestion — per-option `value_id` for multi-select** (`lib/feedback-source/transform.ts`)
`MultipleChoiceMulti` answers now expand into one record per selected option (like matrix/ranking already do), each carrying its own `value_id`, sharing `field_group_id` (the question) and `submission_id` (the respondent). Previously multi-select collapsed into one joined `value_text` (`"A, B"`) with no per-option id.

**2. Dropped the dimension rewrite + split shim** (`charts/lib/option-grouping.ts`, `charts/actions.ts`, dashboard)
`resolveOptionGrouping` no longer rewrites the grouping dimension in either direction. Single- and multi-select are now handled identically: keep the user's chosen dimension, and just build the `value_id → label` map so a Value (Option) grouping still renders human labels. Deleted `charts/lib/multi-select-split.ts` (+ test) — with per-option `value_id` rows it's no longer needed. This fixes both the single- and multi-select "auto-pop" bugs.

**3. Dashboard label lookup** (`dashboards/pages/dashboard-detail-page.tsx` + client/widget-data)
Saved charts on the dashboard were rendering raw `value_id`s (e.g. `c-male`) because `executeWidgetQuery` never called the option-grouping resolver. It now resolves `optionLabels` the same way the builder does and threads them into `ChartRenderer`. (I extracted `resolveOptionGrouping` into `charts/lib/option-grouping.ts` so both the action and the dashboard page can share it.)

**4. Filter / Group By field order** (`analysis/lib/schema-definition.ts`)
Reordered `FEEDBACK_FIELDS.dimensions` by relevance: **Question first** (it's now also the default filter field), then answer values → AI enrichment → source → time/language → identifiers & Hub meta. Pure array reorder, no logic change; the Group By picker benefits too.

**5. Feedback record drawer is now view-only** (`unify-feedback/*`) — product call from Johannes
Removed manual create/edit of feedback records entirely: the "Add feedback record" button + create form, inline editing, and save/discard/metadata-editing. The drawer is a read-only viewer now (all fields disabled; footer = Delete + Close). CSV import stays (dropdown relabeled "Import feedback"). Also removed the now-dead `createFeedbackRecordAction` / `updateFeedbackRecordAction` + their schemas/types and `getCreateDefaults`. The **Value ID** field is always shown (even when empty), paired 50/50 with Value (Text).

**6. i18n** — renamed the drawer label **"Option ID" → "Value ID"**, and removed 12 keys orphaned by the drawer cleanup. `pnpm i18n:validate` is green (0 unused / 0 missing, lockfile in sync).

## Notes / things to know

- **No backfill needed** since this isn't released yet. Existing multi-select rows are still the old joined format, but we're not migrating them — new ingestion produces the per-option rows.
- `effectiveQuery` returned from `executeQueryAction` is now always identical to the input query (nothing rewrites), so that feedback path is inert. I left the plumbing in place to avoid touching several hooks; easy follow-up to remove if we want.
- The PR title still says "canonical value_text" — the canonicalization is still there for `value_text`, but grouping stability now comes from `value_id`, so you may want to retitle.

Full web test suite (6323 tests) and `pnpm i18n:validate` pass; local build run before pushing. Let me know if you'd rather keep the auto-consolidation behavior for single-select — easy to reinstate as an explicit opt-in rather than a silent rewrite.

Dhruwang:
@jobenjada Thank you for the changes, looks good 🚀

pandeymangg (approved):

PR #8466, opened 2026-07-09
refactor(docker): move migrations to a dedicated image and slim the Prisma CLI out of the web runtime [ENG-1153]

xernobyl:
## What does this PR do?

**Moves Formbricks database migrations out of the web runtime image entirely** ([ENG-1153](https://linear.app/formbricks/issue/ENG-1153/move-formbricks-migrations-out-of-the-web-runtime-image)).

Until now the production web image had to carry the Prisma CLI plus the `@prisma/config` loader chain (c12, effect, valibot, jiti, … ~30 modules / ~48 MB) purely to run `prisma migrate deploy` at container startup — surface the running app never uses. This PR relocates migration execution to a dedicated image and slims the web image down to a pure runtime, in one change (no intermediate state where the web image still ships the CLI).

**1. Dedicated migration image**
- New `migrations` Dockerfile stage → `ghcr.io/formbricks/formbricks-migrate`. Reuses the web build's traced `node_modules` and overlays the Prisma CLI + `@prisma/config` closure, then runs `apply-migrations.js` (schema `migrate deploy` + the in-process data-migration runner). Short-lived, one-shot, unprivileged; a build-time `prisma -v` smoke test fails the image build if the CLI closure is incomplete.
- CI publishes it alongside the web image at the **same version tag** (community + experimental), via a new optional `target` input on the `build-and-push-docker` action.

**2. Slim web image (the CLI is gone)**
- Removed the Prisma CLI, `prisma.config.mjs`, and the whole `@prisma/config` chain from the `runner` stage (~48 MB / ~30 modules). Kept `@prisma/client`, the pg driver, pino, zod, cuid2, uuid, `@noble/hashes` — verified **zero** app/runtime imports of any removed package. The chain now lives solely in the migration image (no cross-stage duplication).

**3. Startup no longer migrates — it verifies and fails fast**
- `next-start.sh` no longer runs `prisma migrate deploy`. Instead it runs a new **read-only migration-status check** (`packages/database` → `check-migrations.js`) that diffs the shipped `dist/migration` directory against `_prisma_migrations` (applied) and `DataMigration` (`status='applied'`), and **exits non-zero with a clear message if anything is pending** — so the app never serves against an empty or half-migrated schema. The check matches by directory name (verified every data migration's exported `name` equals its dir), needs only `SELECT`, and deliberately does *not* import the data-migration modules, so it stays robust in the slim image. SAML DB setup (Prisma client, no CLI) is unchanged.
- **`SKIP_STARTUP_MIGRATION` is retired** (removed from `next-start.sh`, the Helm deployment env, and the compose web service) — it no longer means anything now that the web image can't migrate.

**4. Orchestration**
- **Compose**: the `formbricks-migrate` one-shot uses the dedicated image; `formbricks` and `formbricks-migrate` share `FORMBRICKS_IMAGE_REF` (mirroring `HUB_IMAGE_REF`) so they can't drift; and `formbricks` now `depends_on` the migrate service with `service_completed_successfully` (the gate `hub-migrate` already uses).
- **Helm**: the migration Job runs the dedicated image via a `formbricks.migrationImage` helper (digest-first). `migration.image.repository` is now **required** — the helper `fail`s rendering with guidance rather than silently falling back to the CLI-less app image. The `migration.enabled` comment warns that disabling the Job means you must apply migrations yourself.
- **Docs**: `docker/README.md` documents the final flow (run `formbricks-migrate` first; the web container verifies and fails fast), single-container usage, version pinning, and rollback guidance.

**Behaviour change for self-hosters (please note):** a web container run standalone no longer migrates on its own — the dedicated migration step must run first, otherwise the web container exits at startup with a clear message. Compose and Helm wire this up automatically; the README documents the `docker run` flow and the upgrade note.

## How should this be tested?

Locally verified (in the worktree): `@formbricks/database` **builds** (both guard scripts emit to `dist/scripts/`), **typechecks clean** (`tsc --noEmit`, exit 0), **lints clean**, and the pure `diffPendingMigrations` logic passes case checks (all-applied → none pending; partial → correct remainder; empty DB → all pending). `docker compose config` resolves and `FORMBRICKS_IMAGE_REF` substitutes for both services. The image builds themselves run in CI (the monorepo image build OOMs locally).

Reviewer checks:

**Slim web image:**
```bash
docker build -f apps/web/Dockerfile --target runner -t formbricks-web .
docker run --rm formbricks-web sh -c 'test ! -e node_modules/prisma && echo "no prisma CLI ✓"'
# empty/unmigrated DB → the container exits non-zero with a clear "run formbricks-migrate first" message
docker run --rm -e DATABASE_URL="postgresql://user:pass@host:5432/empty" formbricks-web
```

**Migration image + full flow (Compose):**
```bash
cd docker && docker compose up
# formbricks-migrate runs to completion; formbricks then passes its migration check and serves
FORMBRICKS_IMAGE_REF="@sha256:<digest>" docker compose config | grep 'image: ghcr.io/formbricks/formbricks'  # both pinned to same ref
```

**Helm:**
```bash
helm template t charts/formbricks | grep -A1 'name: migration'          # dedicated migration image
helm template t charts/formbricks --set migration.image.digest=sha256:… # Job pins the digest
helm template t charts/formbricks --set migration.image.repository=""     # fails rendering with guidance (intended)
```

**CI to watch:** `Validate Docker Build` / release build (both the slim `runner` and the `migrations` stage), unit + typecheck.

### Required

- [x] Commented on my code in hard-to-understand bits (migration image stage, the startup guard's name-based/import-free rationale, the Helm fail-closed helper)
- [x] Ran `pnpm build` (`@formbricks/database` build + typecheck + lint locally)
- [x] Removed all `console.logs` (none added; the guard uses `@formbricks/logger`)
- [x] Merged the latest changes from main onto my branch

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots — N/A, no UI changes
- [x] Updated the Formbricks Docs if changes were necessary (`docker/README.md`)


xernobyl:
Closing as **deferred, not abandoned**.

This is a breaking change for self-hosters: once the web image drops the Prisma CLI it can no longer self-migrate, so bare `docker run`, plain (non-Argo) `helm install`, and custom docker-compose setups must run the dedicated `formbricks-migrate` step first. We decided not to ship a cosmetic-but-breaking change in a routine release.

The implementation here is complete and was CI-green (dedicated migrate image + CI publish, slimmed web runner, read-only startup migration guard, Helm/compose wiring, docs). It stands as the proven blueprint for doing this properly later as a **versioned release with an upgrade guide** (ideally a deprecation cycle). Tracked on ENG-1153.

Branch `refactor/ENG-1153_dedicated-migration-image` is preserved — reopen or re-branch from it when we pick this up.

PR #8468, opened 2026-07-09
fix: enforce workspace scoping when generating personal survey links [ENG-1635]

pandeymangg:
## What does this PR do?

Hardens personal-link generation for Contacts (EE) so the segment/survey involved must belong to the **same workspace** as the resource the caller was authorized against. Previously the authorization check and the data being read/linked could refer to **different workspaces**, so a caller authorized for one workspace could reference a resource ID belonging to another.

Two server-action paths are affected:

- **`generatePersonalLinksAction`** (summary → `generatePersonalLinks`): authorized the caller against the survey's workspace, but then read contacts from a caller-supplied `segmentId` whose workspace was never verified. It now asserts `survey.workspaceId === segment.workspaceId` **before any contact is read**, and rejects otherwise.
- **`generatePersonalSurveyLinkAction`**: authorized the caller against the contact's workspace, but accepted an unchecked `surveyId`. It now asserts the survey shares the contact's workspace before minting a link.

Both mirror the guard already enforced by the v2 management API (`.../contact-links/segments/[segmentId]`), which rejects the same workspace mismatch. The check sits at the operation entry (not inside the shared `getContactSurveyLink` helper) to avoid adding a per-contact query on the bulk path.

Resolves ENG-1635.

## How should this be tested?

**Automated** — `apps/web/modules/ee/contacts/lib/contacts.test.ts`, added regression coverage for `generatePersonalLinks`:
- segment in a different workspace than the survey → throws `ValidationError`, and **no** contact query / link generation runs (no foreign data touched)
- non-existent survey → returns `null`

```
pnpm --filter=@formbricks/web exec vitest run modules/ee/contacts/lib/contacts.test.ts
# 41 passed
```

**Manual** (needs EE + Contacts, two workspaces A/B):
- From workspace A, run the personal-links export with A's survey + a `segmentId` from workspace B → rejected, zero contacts returned.
- From workspace A, call the single personal-link action with A's contact + a `surveyId` from workspace B → rejected.

### Required

- [ ] Ran `pnpm build`
- [x] Checked for warnings, there are none (lint clean on changed files)
- [x] My changes don't cause any responsiveness issues (no UI change)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (N/A — no UI change)
- [ ] Updated the Formbricks Docs if changes were necessary (N/A)


Dhruwang (changes requested):
Thanks for the PR, it looks good! I just had one suggestion regarding the translated error messages. Currently, we display the toast in English regardless of the user's language settings. Could we potentially send the error codes from the actions and then use those codes to display the translated error messages instead?

Dhruwang (approved):

PR #8471, opened 2026-07-09
fix: preserve current workspace when navigating to org settings (ENG-1700)

Dhruwang:
## Problem

Navigating from a secondary workspace into any Organization settings page snapped the settings shell back to the first workspace in the org.

Organization settings live at a workspace-agnostic route (`/organizations/[organizationId]/settings/**`) — there is no `workspaceId` in the URL. The shell rebuilt its "current workspace" from `workspaces[0]`, and that list was queried with no `orderBy`, so the pick was arbitrary (in practice the oldest workspace). Result: the sidebar Workspace section, back button, and `WorkspaceContext` all flipped to the wrong workspace.

## Solution

Remember the active workspace in a `formbricks-workspace-id` cookie and resolve the shell's current workspace from it.

- **`proxy.ts`** sets the cookie from the `/workspaces/[workspaceId]` path segment on every matching request.
- **`getSettingsLayoutData`** reads the cookie via `cookies()` and uses it **only if** it belongs to the user's accessible workspaces; otherwise it falls back to the first workspace.
- **`workspace.ts`** now orders the workspace query by `createdAt asc`, so the fallback is deterministic instead of arbitrary DB order.

### Why a cookie (not the URL or localStorage)

- The settings **shell is a server component** (rendered by the layout). Next.js layouts do **not** receive `searchParams`, so a `?workspaceId=` query param can't drive it — and it wouldn't survive refresh/deep-link anyway.
- The app already tracks the last workspace in **localStorage**, but that's browser-only and unreadable during server render. A cookie is the same "last active workspace" signal, readable server-side with no flash of the wrong workspace.

## Testing

- `navigation-data.test.ts`: added cases for cookie-in-accessible-list (uses it) and cookie-not-accessible (falls back to first). 6/6 green.
- `proxy.test.ts`: added cases asserting the cookie is set on `/workspaces/[id]` paths and not on other paths.
- `workspace.test.ts`: updated `findMany` expectations for the new `orderBy`.
- `pnpm build --filter=@formbricks/web` green; typecheck + lint clean on touched files.

Fixes ENG-1700.

### DIFF
--- a/apps/web/lib/localStorage.ts
+++ b/apps/web/lib/localStorage.ts
@@ -2,3 +2,8 @@ export const FORMBRICKS_SURVEYS_FILTERS_KEY_LS = "formbricks-surveys-filters";
 export const FORMBRICKS_ENVIRONMENT_ID_LS = "formbricks-environment-id";
 export const FORMBRICKS_WORKSPACE_ID_LS = "formbricks-workspace-id";
 export const FORMBRICKS_LOGGED_IN_WITH_LS = "formbricks-logged-in-with";
+
+// Server-readable mirror of the "last active workspace". The proxy sets this from the
+// /workspaces/[workspaceId] path so server components (e.g. the workspace-agnostic org-settings
+// shell) can resolve the current workspace during render — localStorage is browser-only.
+export const FORMBRICKS_WORKSPACE_ID_COOKIE = "formbricks-workspace-id";
--- a/apps/web/app/(app)/workspaces/[workspaceId]/lib/workspace.test.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/lib/workspace.test.ts
@@ -47,6 +47,7 @@ describe("Workspace", () => {
           id: true,
           name: true,
         },
+        orderBy: { createdAt: "asc" },
       });
       expect(result).toEqual(mockWorkspaces);
     });
@@ -77,6 +78,7 @@ describe("Workspace", () => {
           id: true,
           name: true,
         },
+        orderBy: { createdAt: "asc" },
       });
       expect(result).toEqual(mockWorkspaces);
     });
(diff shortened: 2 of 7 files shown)


jobenjada:
forwarding the review to @itsjavi as it's beyond my understanding if this is the right way to solve this :)

apps/web/lib/localStorage.ts
```diff
@@ -2,3 +2,8 @@ export const FORMBRICKS_SURVEYS_FILTERS_KEY_LS = "formbricks-surveys-filters";
 export const FORMBRICKS_ENVIRONMENT_ID_LS = "formbricks-environment-id";
 export const FORMBRICKS_WORKSPACE_ID_LS = "formbricks-workspace-id";
 export const FORMBRICKS_LOGGED_IN_WITH_LS = "formbricks-logged-in-with";
+
+// Server-readable mirror of the "last active workspace". The proxy sets this from the
+// /workspaces/[workspaceId] path so server components (e.g. the workspace-agnostic org-settings
+// shell) can resolve the current workspace during render — localStorage is browser-only.
+export const FORMBRICKS_WORKSPACE_ID_COOKIE = "formbricks-workspace-id";
```
itsjavi, line 9: what would happen if this workspace is deleted? do we show a 404, do cleanup this entry?

itsjavi (changes requested):
The cookie approach + sorting is the right call, but it also reveals a UX design flaw: The workspace selector from organization settings pages is not needed here, and we should have 2 different shells instead of an unified one for workspaces/orgs.

If it can be easily doable here, would be good, or in a follow up PR.
 
image

jobenjada:
yeah that's a good point! we should get some insight from PostHog on how many people use it vs. the switch in the main nav, its a bit redundant and we're giving away space

itsjavi (approved):
LGTM for the scope of the fix
I created this follow up: https://linear.app/formbricks/issue/ENG-1705/investigate-removing-the-workspace-selector-from-organization-settings

PR #8475, opened 2026-07-09
chore(billing): emit paid-subscription lifecycle events to PostHog

harshsbhat:
## What

Emits three PostHog analytics events when an organization's **paid** Stripe
subscription changes state, so growth/sales can track conversions, churn, and
plan changes (and pipe them into Attio as deal signals):

| Event | Fires when | Key properties |
|-------|------------|----------------|
| `subscription_started` | A subscription first becomes **active on a paid plan** (trial/incomplete → active) | `plan`, `interval`, `organization_id` |
| `subscription_canceled` | An active paid subscription **ends** (canceled, deleted, or downgraded to hobby) | `plan` (churned-from), `interval`, `organization_id` |
| `subscription_updated` | An active paid subscription **switches tier** (Pro ↔ Scale, either direction) | `previous_plan`, `plan`, `interval`, `organization_id` |

All events are keyed off the **organization owner** (ties to a person in
PostHog/Attio) and attach the `organization` group for company attribution.

## Why

Today a paid conversion produces **no analytics signal** — the Stripe webhook only
reconciles billing state in the DB, and the existing `free_trial_started` /
`reverse_trial_started` events fire at *trial* creation, before any payment. There
was no way to see "became a paying customer", "churned", or "upgraded" in PostHog.

## How

Detection lives in `syncOrganizationBillingFromStripe` — the single function that
already computes and persists subscription state on every Stripe webhook (and on
read-through syncs). It compares the freshly computed state to the
previously-persisted snapshot:

- A subscription only reaches `active` **after Stripe collects payment** (a failed
  charge lands in `incomplete`/`past_due`), so *paid + active* means the org is
  actually paying.
- Recovering from a dunning state (`past_due`/`unpaid`/`paused` → `active`) is **not**
  a fresh start.
- Dropping to `hobby` / no subscription is a cancel; a **trial lapsing unpaid never
  fires** (it was never paid-active).

This needs **no new DB column/migration** (it reuses the already-persisted
`subscriptionStatus`/`plan`) and **no webhook changes** — the webhook already calls
this function. Analytics is emitted *after* the billing write and is fire-and-forget
(internal try/catch), so it can never delay or break billing sync.

## Notes / trade-offs

- Emits **once per transition**: the new state is persisted before the event, so the
  next sync sees it as "previous" and won't repeat.
- Works from any trigger (webhook or read-through sync), so a missed webhook is still
  caught on the next load.
- The extra owner lookup runs **only** on an actual transition, not every webhook.
- Cloud-only (`syncOrganizationBillingFromStripe` returns early when not Formbricks Cloud).
- **Known edge:** the stateless comparison could theoretically double-fire if two
  syncs for the same transition interleave before either persists. It's bounded by
  the existing per-event dedup (`shouldSkipStripeSyncForEvent`) and is low-probability.

## Testing

- Existing billing suites pass (`organization-billing`, `stripe-webhook` — 52 tests).
- Updated two responses-API test mocks to include `POSTHOG_KEY` (the new import pulls
  the PostHog server module into their graph); both suites pass.
- Typecheck + lint clean.
- Manually verified against Stripe test-mode transitions (trial → active, cancel,
  Pro ↔ Scale).


apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -1267,6 +1271,84 @@ const resolvePendingPlanChange = async (subscription: Stripe.Subscription | null
...
+  switchedPaidPlan: boolean;
+};
+
+const resolveSubscriptionLifecycleTransition = (
+  existingStripeSnapshot: TOrganizationBilling["stripe"],
+  subscription: Stripe.Subscription | null,
+  subscriptionStatus: TOrganizationStripeSubscriptionStatus | null,
+  cloudPlan: TCloudStripePlan
+): TSubscriptionLifecycleTransition => {
+  const wasPaidActive =
```
xernobyl, line 1286: **[P2] Dunning (involuntary) churn never emits `subscription_canceled`.**

`wasPaidActive` requires the *previous* status to be exactly `"active"`, but `resolveCurrentSubscription` keeps `past_due`/`unpaid`/`paused` subs and `resolveCloudPlanFromSubscription` still resolves them to `pro`/`scale`. So the common failed-payment churn path drops the event:

1. `active`+`pro` persisted.
2. Payment fails → `customer.subscription.updated` → sync persists `{status: past_due, plan: pro}` (no event, correct).
3. Dunning exhausts → `canceled` → sync runs with `existing.status === "past_due"` ⇒ `wasPaidActive = false` ⇒ `canceledPaidSubscription = false`. **No event ever fires.**

Involuntary churn is the dominant SaaS churn mode and this silently misses it whenever `past_due` was persisted first. Voluntary cancel (stays `active` until period end) is fine.

*Fix direction:* for cancel detection treat the org as previously-paying when `isPaidCloudPlan(existing.plan)` and `existing.status ∈ {active, past_due, unpaid, paused}` — but keep `active`-only for `startedPaidSubscription`, since `trialing` is also in `ACTIVE_SUBSCRIPTION_STATUSES` and a reverse-trial sits at `{plan: pro, status: trialing}`; widening "started" would suppress the primary conversion event.

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -1267,6 +1271,84 @@ const resolvePendingPlanChange = async (subscription: Stripe.Subscription | null
...
+  transition: TSubscriptionLifecycleTransition;
+}): Promise<void> => {
+  const { organizationId, existingStripeSnapshot, cloudPlan, billingInterval, transition } = input;
+  const { startedPaidSubscription, canceledPaidSubscription, switchedPaidPlan } = transition;
+
+  if (!startedPaidSubscription && !canceledPaidSubscription && !switchedPaidPlan) {
+    return;
+  }
+
+  const owner = await getOrganizationOwner(organizationId);
```
xernobyl: **[P3] "fire-and-forget (internal try/catch)" isn't quite accurate.** This `getOrganizationOwner()` call is `await`ed and unguarded — `capturePostHogEvent` guards itself, but a DB blip here rejects the awaited `emitSubscriptionLifecycleEvent`. Both callers catch (webhook → 500+retry, read-through → logged), so billing state is safe, **but the event is lost permanently**: the retried sync sees the already-persisted new state and detects no transition. Wrap the emit body in try/catch (matching the stated intent) so a lookup failure can't drop the event. Minor: it's also one awaited DB round-trip per transition, so not truly fire-and-forget.

xernobyl (commented):
Nice, clean approach — reusing the persisted snapshot with no migration and keying off the owner with the `organization` group is exactly right, and the PostHog usage matches the repo conventions (snake_case events/props, `distinctId = user.id`, no PII, `flushAt: 1` + shutdown flush). Two correctness notes inline, plus one here.

**[P2] No test coverage for the transition matrix.** The diff only adds `POSTHOG_KEY` to two unrelated response-API mocks; nothing exercises `subscription_started/canceled/updated`. Given the dunning edge below, this logic really wants a table-driven test through `syncOrganizationBillingFromStripe` (start / voluntary-cancel / dunning-cancel / Pro↔Scale / dunning-recovery / trial-lapse) in the existing `organization-billing` or `stripe-webhook` suite.

Minor / non-blocking: interval-only changes (monthly↔yearly, same tier) emit nothing since `switchedPaidPlan` keys off `plan` only — fine if out of scope, but it's a real expansion signal worth a follow-up.

xernobyl (approved):
LGTM. The dunning-churn gap, the unguarded emit, and the missing test coverage are all resolved — the `wasPaidRecoverable` split correctly fires `subscription_canceled` on involuntary churn while keeping trial-lapse silent, the emit is now best-effort (try/catch + log, can't reject the sync), and the `test.each` matrix covers the full transition table including the dunning edge. CI is green.

One optional non-blocker: `getOrganizationOwner` uses `findFirst` without an `orderBy`, so on a multi-owner org the owner picked for the `distinctId` is arbitrary. In practice this rarely matters, so fine to leave as-is or clean up in a follow-up.

PR #8479, opened 2026-07-09
refactor(workflows): live autosaving editor with guided validation [ENG-1234]

itsjavi:
## What does this PR do?

Reworks the workflow editor's saving and validation model around a single live document, polishes the header lifecycle UX, and fixes three bugs found along the way (one of them in the runner). Part of ENG-1234.

### Saving: one live document instead of per-panel Save buttons

- **Node config edits write straight into the editor state** (Jotai + Immer), so canvas cards, summaries, and validation react per keystroke. The per-node sidebar Save button is gone; that was the original intent of the editor-state design, and the intermediate "save each piece" flow fought it.
- **Silent autosave** persists any dirty draft 2s after the user stops editing, flushes on unmount and before lifecycle transitions (enable aborts if the flush fails, so it never publishes a definition other than the one on screen), and never retries an unchanged failing draft (no error-toast loops). Dirty tracking compares against a snapshot of what was last **sent**, so server-side normalization can't cause save loops and mid-flight edits stay dirty.
- **Save-state visibility, mirroring the survey editor**: an "Auto-save on" pill that flashes green "Changes saved" after every successful save, a brand-colored dot on the Save button while there is something unsaved (the button doubles as a manual "persist now"), and a native `beforeunload` confirm on refresh/close with unsaved edits (SPA navigation is covered by the unmount flush).

### Validation: know the workflow's state before going live

- **Live whole-workflow validity** (`workflowValidityAtom`) mirrors the exact server-side enable/test pre-flight (`ZWorkflowExecutableDefinition` + survey binding + name), computed on every edit.
- **Guided next step instead of "Workflow has issues"**: an unfinished draft isn't broken, so the header now shows the one concrete next step in build order — "Add a trigger to get started" → "Complete the trigger configuration" → "Add an action step" → "Complete the action configuration" → "Name your workflow" — as quiet slate guidance on drafts and a red warning only for previously-live workflows that can no longer run. The generic copy remains only as a structural fallback.
- **The executable schema now rejects incomplete send_email actions** (empty to/subject/body). Before, the dry-run reported an unconfigured email action as "valid and ready to run" and enable happily published it (it would then fail at send time).
- **Dry-run works on drafts** — the point of testing is checking the setup *before* going live. Only archived workflows are rejected.

### Header lifecycle UX

- **Status dropdown replaces the enabled/disabled toggle** (same shape as the surveys list "New survey" menu): the button reads the current status, the menu holds the transitions available from it — Disable/Archive while enabled, Enable/Archive on drafts and disabled (Enable gated on live readiness), Unarchive/Delete once archived. Delete confirms and returns to the list.
- **Settings sidebar and canvas cog removed**: the name is editable in-place in the page title, lifecycle actions live in the dropdown, and the description had no remaining surface. The inspector now only exists while a node's config is open; clicking empty canvas dismisses it (the Back arrow was also removed earlier — node config and workflow settings were siblings, not a hierarchy).
- **Node deep-linking**: the inspected node is mirrored into `?node=` via the native `history.replaceState` (shallow — no navigation or reload; same pattern as the survey summary page), so reloading reopens the same node with its config panel and canvas selection.

### Fixes

- **Runner: surveys without ending cards never triggered workflows.** The enqueue gate required `response.endingId`, so an "all endings" trigger bound to a survey with no ending cards silently never fired. Completed responses now enqueue regardless; explicit ending selections still require reaching one of the chosen cards.
- **Respondent email removed from the response notification subject** — subjects leak into mailbox lists and push previews; replying still reaches the respondent via `replyTo`.
- **Bound-survey readiness re-syncs after hydration** (the guided hint could skip the "complete the trigger" step on a fresh trigger).

## How should this be tested?

Manually verified end-to-end against the local stack (MailHog):

- Fresh workflow: guided hint progression, autosave pill + "Changes saved" flash, dirty dot, disabled Save when clean, Draft dropdown with Enable gated until ready.
- Dry-run on an incomplete draft → problems dialog (trigger/entryNodeId, incomplete email); on a complete draft → `ok: true`.
- `?node=` deep link: click node → param appears (no reload); hard reload → same node config reopens; closing the panel clears the param and preserves other params.
- `beforeunload`: blocks with unsaved edits, silent when clean.
- Dropdown lifecycle: enable (flushes dirty name first), archive (confirm), unarchive, delete (confirm → back to list).
- Runner e2e with three surveys: **no ending card** (all-endings trigger), **one ending card** (specific selection), **multiple ending cards** (all selected). Each completed response produced a workflow run and an email in MailHog containing the authored body plus the attached response data block.

Automated: `packages/workflows` 210 tests, `apps/web/modules/workflows` 300 tests, `modules/email` 32 tests, typecheck, lint, and `pnpm i18n:validate` all green.

### Required

- [x] Ran `pnpm build`
- [ ] Merged the latest changes from main onto my branch with `git pull origin main` (branch targets `refactor/ENG-1234-workflows-polish-ui`)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (verified locally against MailHog; happy to attach recordings on request)


jobenjada (changes requested):
Overall much better, love the auto-save!

Things i noticed:

If workflow is enabled, the Body isn't disabled:

image

-----

UX confusion: Does "Test" run the workflow or does it check validitiy?

image

----

## Error Type
Runtime TypeError

## Error Message
Cannot assign to read only property 'width' of object '#<Object>'


    at WorkflowCanvasContent.useCallback[handleNodesChange] (modules/workflows/components/canvas/workflow-canvas.tsx:142:65)
    at Object.write (modules/workflows/state/editor.ts:286:58)
    at WorkflowCanvasContent.useCallback[handleNodesChange] (modules/workflows/components/canvas/workflow-canvas.tsx:142:18)

## Code Frame
  140 |
  141 |   const handleNodesChange: OnNodesChange<Node<TWorkflowNodeData>> = useCallback(
> 142 |     (changes) => setFlowNodes((currentNodes) => applyNodeChanges(changes, currentNodes)),
      |                                                                 ^
  143 |     [setFlowNodes]
  144 |   );
  145 |

Next.js version: 16.2.6 (Turbopack)


-----

### Loading layout for "Runs" 

I'm not sure if I really saw it but it looked like the loading skeleton on /Runs looked like the canvas, before it showed the table loading. Can you double check if the layouts are nested correctly? I see that the routes are nested so maybe we need to move the loading sekelton from the parent layout into the edit view?

---

Nit: the loading skeleton of the Workflows list looks different than the one from Survey list (even tho its the same list). Pls pick the one you like more and align :) 

----

Nit: This took a while to load:

image

----

Nit: Are we caching the workflows list? Looks like its being loaded on every page load but would be great to implement the caching best practices we've talked about 🤓. Maybe we can have a loading skeleton for the "Run" count, if such granularity makes sense.

----

### Rest looks great!!

itsjavi:
@jobenjada  changes:

- Test renamed to validate with new icon
- changed placement of the hint of what to do next -> inside canvas
- removed superfluous save button
- email body not editable when the rest isn't
- Auto-save on hint changes to "Read-only" when workflow cannot be edited
- Better slide animation timings

API call caching would need more investigation since it falls into performance rather than UI changes

PR #8484, opened 2026-07-10
chore: Add email confirmed posthog event

harshsbhat:
## Description

Adds a server-side PostHog event, `user_email_confirmed`, that fires when a user confirms their email address for the first time. The event is captured inside the Better Auth `afterEmailVerification` hook (`createBrevoCustomerAfterEmailVerification`), which Better Auth guarantees fires exactly once per user, alongside the existing Brevo CRM side effect.

## Motivation and Context

We had no analytics signal for email confirmation, leaving a gap in the signup → activation funnel. This event lets us measure how many signups verify their email and how confirmation correlates with downstream activation.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Changes

- Capture `user_email_confirmed` in `createBrevoCustomerAfterEmailVerification` with `user.id` as the `distinctId` and `{ email }` as an event property.
- Import `capturePostHogEvent` from `@/lib/posthog` (already `server-only`, no-ops when PostHog is not configured).

## How Has This Been Tested?

- [ ] Manual QA: signed up a new user, confirmed the email via the verification link, and verified the `user_email_confirmed` event appears in PostHog with the correct `distinctId` and `email`.
- [ ] Confirmed the event does not re-fire on subsequent logins (Better Auth blocks re-verification).

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published

## Notes

- Scope is first-time signup confirmation only. The email-*change* flow (`modules/auth/verify-email-change/actions.ts`) is not covered by this event.

### DIFF
--- a/apps/web/modules/auth/lib/better-auth-email-verification.ts
+++ b/apps/web/modules/auth/lib/better-auth-email-verification.ts
@@ -1,5 +1,6 @@
 import "server-only";
 import { logger } from "@formbricks/logger";
+import { capturePostHogEvent } from "@/lib/posthog";
 import { createBrevoCustomer } from "@/modules/auth/lib/brevo";
 
 /**
@@ -14,6 +15,8 @@ export const createBrevoCustomerAfterEmailVerification = async (user: {
   id: string;
   email: string;
 }): Promise<void> => {
+  capturePostHogEvent(user.id, "user_email_confirmed", { email: user.email });
+
   void createBrevoCustomer({ id: user.id, email: user.email }).catch((err) =>
     logger.error(err, "Failed to create Brevo customer after email verification")
   );
--- a/apps/web/modules/auth/lib/better-auth-email-verification.ts
+++ b/apps/web/modules/auth/lib/better-auth-email-verification.ts
@@ -15,7 +15,7 @@ export const createBrevoCustomerAfterEmailVerification = async (user: {
   id: string;
   email: string;
 }): Promise<void> => {
-  capturePostHogEvent(user.id, "user_email_confirmed", { email: user.email });
+  capturePostHogEvent(user.id, "user_email_confirmed");
 
   void createBrevoCustomer({ id: user.id, email: user.email }).catch((err) =>
     logger.error(err, "Failed to create Brevo customer after email verification")
--- a/apps/web/modules/auth/lib/auth-email-verification.integration.test.ts
+++ b/apps/web/modules/auth/lib/auth-email-verification.integration.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { prisma } from "@formbricks/database";
 import { resetDb } from "@/integration/reset-db";
+import { capturePostHogEvent } from "@/lib/posthog";
 import { auth } from "@/modules/auth/lib/auth";
 import * as brevo from "@/modules/auth/lib/brevo";
 import {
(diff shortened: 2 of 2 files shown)


xernobyl (commented):
The mechanics here are right — `distinctId = user.id` merges cleanly with the `user_signed_up`/`user_signed_in` person, `capturePostHogEvent` is a safe sync no-op without `POSTHOG_KEY`, and the hook's documented once-per-user guarantee (ENG-1054) makes this fire exactly once. Two things to fix before merge — one inline on the `email` property, and the test gap below.

---

### [P2] Extend the existing integration test

`auth-email-verification.integration.test.ts` already drives this exact hook end-to-end and asserts the sibling side effect — but it wasn't updated for the new event, so the "added tests" box is unchecked and the emit is unverified. The harness is already there; mock the PostHog boundary the same way `brevo` is mocked and add one assertion.

At the top, alongside the existing `brevo` mock:

```ts
import { capturePostHogEvent } from "@/lib/posthog";

vi.mock("@/lib/posthog", () => ({
  capturePostHogEvent: vi.fn(),
}));
```

Then in the "consuming it flips emailVerified" test, next to the existing `createBrevoCustomer` assertion:

```ts
expect(capturePostHogEvent).toHaveBeenCalledWith(verifiedUser?.id, "user_email_confirmed");
```

And since idempotency is a stated guarantee, assert it after the replay `verifyEmail` call — the already-verified branch returns `{ status: true, user: null }`, so the hook must NOT fire again:

```ts
expect(capturePostHogEvent).toHaveBeenCalledTimes(1);
```

That both proves the event fires and locks in the once-per-user contract. (Mocking the analytics boundary is consistent with how `createBrevoCustomer` is already mocked in this file.)

apps/web/modules/auth/lib/better-auth-email-verification.ts
```diff
@@ -14,6 +15,8 @@ export const createBrevoCustomerAfterEmailVerification = async (user: {
   id: string;
   email: string;
 }): Promise<void> => {
+  capturePostHogEvent(user.id, "user_email_confirmed", { email: user.email });
```
xernobyl: **[P2] Drop the raw `email` property — it breaks the repo's own PII convention.**

Every other person-scoped event here deliberately avoids raw email and captures only the domain when a signal is needed:

- `user_signed_up` (credentials): `email_domain: user.email.split("@")[1]` — `modules/auth/signup/actions.ts`
- `user_signed_up` (SSO): `email_domain: email.split("@")[1]` — `modules/ee/sso/lib/sso-provisioning.ts`
- `user_signed_in`: no email at all

Raw email is sent only to Brevo (the CRM, where it belongs). Sending `{ email }` on the PostHog event is a regression on three counts: (1) it duplicates PII into every event payload, widening the GDPR/deletion surface and exposing raw email in live event feeds; (2) it's redundant — the event is already tied to the person via `distinctId = user.id`, so PostHog resolves the person's email from the profile; (3) it's inconsistent with the three sites above.

**Recommended fix — drop it** (the `distinctId` already carries person attribution; `properties` is optional on `capturePostHogEvent`):

```ts
capturePostHogEvent(user.id, "user_email_confirmed");
```

**Alternative** — if you want a segmentation dimension on the event, mirror the signup events and send the domain, not the address:

```ts
capturePostHogEvent(user.id, "user_email_confirmed", {
  email_domain: user.email.split("@")[1] ?? null,
});
```

Either keeps the event PII-clean and consistent with the rest of the codebase. If you take the drop option, update the test assertion accordingly (`toHaveBeenCalledWith(verifiedUser?.id, "user_email_confirmed")`).

xernobyl (approved):
LGTM. Both points are resolved: the raw `email` property is dropped so the event stays PII-clean with person attribution via `distinctId = user.id` (consistent with the signup/sign-in events), and the existing real-Postgres integration test now asserts `user_email_confirmed` fires with the right distinctId and — via `toHaveBeenCalledTimes(1)` after the replay — locks in the once-per-user contract. Unit + Better Auth integration tests green.

PR #8490, opened 2026-07-10
docs(api-keys): clarify workspace vs organization access scopes

jobenjada:
## Problem

"Organization Access" in the API key create dialog reads like "access to the whole organization, including all workspaces." It isn't: it only gates organization-administration endpoints (members, teams) and never grants workspace data access. Users enable Organization Access, skip workspaces, and then hit errors calling the Management API.

## Change

**Create-key modal** (`add-api-key-modal.tsx`)
- Add a description under **Workspace Access** explaining it controls workspace data (surveys, responses, …) and that each workspace must be added — there is no "all workspaces" option.
- Add an empty-state hint when no workspace is selected.
- Rewrite the **Organization Access** description to say it covers organization administration only and does not grant workspace data access.
- On **Formbricks Cloud**, show an inline info note under Organization Access that member/user management via the API isn't available (threaded via a new `isFormbricksCloud` prop: `page` → `ApiKeyList` → `EditAPIKeys` → `AddApiKeyModal`). The section stays visible because team & workspace-team management still work on Cloud.

**Docs** (`generate-key.mdx`)
- Add an access-scope table (Workspace vs Organization → which endpoints).
- State explicitly that Organization Access is not an "all workspaces" wildcard.
- Note scopes are immutable after creation (only the label can change).
- Note the Cloud user-management limitation.

**i18n**
- New keys added to `en-US.json`; the other 14 locales + `i18n.lock` regenerated via lingo.

## Companion PR

Bug fix that makes organization-only keys actually reach the org endpoints: #8489


xernobyl (approved):
Thank you for the PR. Looks good to ship.

PR #8495, opened 2026-07-12
chore(auth): identify user in PostHog on sign-up

harshsbhat:
## Problem

Signed-up users show up as **unidentified** in PostHog. The only `identify` happens client-side, but it degrades to a bare property `$set` (the browser SDK already uses the user id as its `distinct_id`, so no `$identify` event fires and `is_identified` never flips).

## Solution

Identify the person **server-side at sign-up**, so they're marked identified with email/name in PostHog regardless of what the client does.

- Add `identifyPostHogPerson()` helper wrapping `posthog-node`'s `identify()` (sets person properties **and** fires `$identify`).
- Call it in the sign-up action.

## Notes

- Sends full email/name to PostHog on all deployments (Cloud is on EU hosting + DPA). A follow-up could gate the PII behind `IS_FORMBRICKS_CLOUD` for self-hosted.
- Sign-in path has the same issue and is not covered here.

## Testing

- `pnpm test` — updated the `@/lib/posthog` mock in the sign-up action tests; suite passes.


jobenjada:
I let @pandeymangg do the review but we might want to gate the identify call to only happen when the corresponding env vars are set

Dhruwang (commented):
Clean fix, matches the existing helpers 👍

SSO new users have the same issue. apps/web/modules/ee/sso/lib/sso-provisioning.ts:186 fires user_signed_up but no identify — SSO signups would still be unidentified. Can we add identifyPostHogPerson(userId, { email, name }) there too?

Dhruwang (dismissed):

pandeymangg (approved):

PR #8498, opened 2026-07-13
chore(pnpm): activate supply-chain release cooldown

mattinannt:
## Problem

`pnpm-workspace.yaml` declares a `minimumReleaseAgeExclude` for `@formbricks/hub` but never sets `minimumReleaseAge` itself. On pnpm 11 a **default 1-day cooldown** applies, but the effective policy was implicit, undocumented, and weaker than intended for a security-conscious repo — and nothing pinned it against upstream default changes.

*(Correction from review: the original description claimed the cooldown was fully inactive — that was true for pnpm 10 but not for pnpm 11, which defaults `minimumReleaseAge` to 1440 minutes. The PR now frames this accurately.)*

## Solution

- `minimumReleaseAge: 4320` (3 days) set explicitly — raising the implicit 1-day default; most npm supply-chain compromises are detected and yanked within 1–2 days.
- `trustLockfile: false` pinned explicitly (pnpm's current default): `pnpm install` **re-verifies entries already in the lockfile** against the cooldown, so a poisoned or bot-written lockfile bump younger than 3 days fails install until it ages or is excluded. Pinning it documents this as a deliberate choice and protects against a future upstream default flip.
- Comment block documents the real pnpm 11 semantics, including the operational consequence: dependency-bump PRs whose lockfile entries are younger than 3 days will fail CI install until the packages age (that is the feature working). The existing `@formbricks/hub` name-based exclusion (first-party, published by our own CI) is unchanged.

## Validation

- `pnpm install` ✅ and `pnpm install --frozen-lockfile` ✅ (exit 0) **with re-verification active** against the current lockfile
- `pnpm lint` ✅ exit 0
- `pnpm test` ✅ `Tasks: 21 successful, 21 total`

Fixes ENG-1675

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -86,10 +86,16 @@ enablePrePostScripts: true
 legacyPeerDeps: true
 nodeLinker: hoisted
 saveExact: true
-# Supply-chain release cooldown: refuse to resolve package versions published less than
+# Supply-chain release cooldown: refuse to install package versions published less than
 # 3 days (4320 minutes) ago, so compromised releases are usually detected/yanked before we
-# can pull them. Existing lockfile entries are unaffected; only new resolutions are gated.
+# can pull them. pnpm 11 already defaults to 1 day; we raise it to 3 days and pin it
+# explicitly. Note: with trustLockfile disabled (below), `pnpm install` re-verifies entries
+# already in the lockfile against this policy too — a lockfile bump younger than 3 days
+# (e.g. written by a bot) fails install until it ages or is excluded below. Intended.
 minimumReleaseAge: 4320
+# Keep the install-time re-verification of lockfile entries active (this is pnpm's default;
+# pinned explicitly so a future default flip can't silently weaken the cooldown).
+trustLockfile: false
 minimumReleaseAgeExclude:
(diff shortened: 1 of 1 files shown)


mattinannt:
**Review finding verified against pnpm 11 docs — valid, fixed in 88afe55.**

Two corrections landed:
1. `minimumReleaseAge` already defaults to **1440 min (1 day) since pnpm v11**, so the cooldown was not fully inactive as originally claimed (`pnpm config get` returning `undefined` only reflects the *configured* value, not the effective default). This PR now explicitly raises it to 3 days.
2. The comment claiming *"existing lockfile entries are unaffected"* was wrong: pnpm 11 re-applies `minimumReleaseAge` to loaded-lockfile entries during install unless `trustLockfile: true`. The behavior **is** intended here (it is what catches poisoned/bot-written lockfile bumps), so per the suggestion it is now configured explicitly: `trustLockfile: false`, with a comment explaining the operational consequence for fresh dependency-bump PRs.

Verified: normal and `--frozen-lockfile` installs pass with re-verification active against the current lockfile; lint + full test suite green.

xernobyl (approved):
Thank you for the PR. LGTM.

PR #8502, opened 2026-07-13
chore(vitest): remove dead environmentMatchGlobs, adopt projects layout

mattinannt:
## Problem

`environmentMatchGlobs` was **removed in Vitest 4**, but is still configured in `apps/web/vite.config.mts` and `packages/surveys/vite.config.mts` — silently ignored. Component tests only pass because each file carries a `// @vitest-environment` pragma (the surveys config even references `jsdom`, which isn't installed there — proof it never ran under v4). The next `.test.tsx` added without a pragma runs in a node environment and fails with confusing `document is not defined` errors. The root `vitest.workspace.ts` also used the deprecated workspace-file approach.

## Solution

- **apps/web**: `unit` (node) + `components` (jsdom) projects. `*.test.tsx` get jsdom automatically — pragma removed from `use-sign-out.test.tsx` to prove it.
- **packages/surveys**: `surveys-unit` (node) + `surveys-components` (happy-dom) projects, mirroring the effective pragma reality. Pragmas removed from both `.test.tsx` files; the handful of DOM-dependent `.ts` tests keep their per-file pragma.
- **jsdom** added as an explicit `apps/web` devDependency — it was a phantom dep resolved only via hoisting (ENG-1681's sibling problem).
- **Root `vitest.workspace.ts` deleted, deliberately without a root replacement.** I verified empirically that a root `projects` list referencing package configs **silently discards their nested project definitions** — the jsdom test ran in node from the root (`environment 0ms`) while passing per-package. A root config would therefore be a trap that mis-runs web/surveys tests. Nothing uses root-level vitest: CI and dev run per package via `turbo run test`, and IDE extensions auto-discover per-package configs (where the projects work correctly). Deviation from the issue's AC noted here for review.

## Validation

- Per-package: `packages/surveys` → 26 files / 650 tests ✅; `apps/web` → 503 files ✅ (identical counts to before — no tests lost by the include/exclude split)
- De-pragma'd jsdom/happy-dom tests pass via project env selection ✅
- `pnpm install` ✅ exit 0 · `pnpm lint` ✅ exit 0 · `pnpm test` ✅ `Tasks: 21 successful, 21 total — Time: 42.101s`

Fixes ENG-1680

### DIFF
--- a/apps/web/modules/auth/hooks/use-sign-out.test.tsx
+++ b/apps/web/modules/auth/hooks/use-sign-out.test.tsx
@@ -1,4 +1,3 @@
-// @vitest-environment jsdom
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import { FORMBRICKS_ENVIRONMENT_ID_LS, FORMBRICKS_WORKSPACE_ID_LS } from "@/lib/localStorage";
 import { logSignOutAction } from "@/modules/auth/actions/sign-out";
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -160,6 +160,7 @@
     "@vitest/coverage-v8": "4.1.6",
     "cross-env": "10.1.0",
     "dotenv": "17.3.1",
+    "jsdom": "29.1.1",
     "postcss": "8.5.14",
     "resize-observer-polyfill": "1.5.1",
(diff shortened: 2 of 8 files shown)


pandeymangg (approved):
LGTM

PR #8506, opened 2026-07-13
fix(surveys): enforce link-survey PIN on the client response API [ENG-1579]

Dhruwang:
## Problem

PIN protection for link surveys was enforced only in the link-survey UI (the `PinScreen` gate). The public client response-submission endpoints did not validate `survey.pin` or require any server-verifiable proof that the client completed PIN authentication. A response could therefore be submitted to a PIN-protected survey without the PIN — the PIN acted as a client-side UI restriction only.

Reported via the security policy (ENG-1579). Class: CWE-602 (client-side enforcement of server-side security).

## Solution

Establish server-side authorization for PIN-protected link surveys, mirroring the existing single-use-link token pattern:

- On successful `validateSurveyPinAction`, mint a short-lived signed token (JWT HS256, 1h TTL) bound to the survey id and a `link_survey_pin` purpose.
- Thread the token from the PIN screen through to the response payload (in the request **body**, never the URL).
- Require and verify the token in **both** the v1 and v2 client response endpoints whenever `survey.pin` is set. Missing, invalid, expired, or wrong-survey tokens are rejected with `403`.

Surveys with no PIN are unaffected (the guard only runs when `survey.pin` is set).

## Files

- `modules/survey/link/lib/pin-token.ts` — mint/verify helpers (+ unit tests).
- `modules/survey/link/actions.ts` — return the token on successful PIN validation.
- `modules/survey/link/components/*` — thread the token to the survey component.
- `packages/surveys` (survey-state, response-queue, survey) + `packages/types` — carry `pinAuthToken` in the response input.
- `app/api/v1/.../responses/route.ts` and `app/api/v2/.../responses/lib/utils.ts` — server-side enforcement.

## Testing

- New unit tests for the token helper: valid → accepted; wrong survey / tampered / missing / wrong purpose / expired → rejected.
- `pnpm test` for the v1 + v2 response routes, link actions, and surveys state/queue suites — all passing (99 tests across the touched areas).
- Rebuilt the `@formbricks/surveys` UMD bundle (client change).

## Notes

- `@formbricks/surveys` is a pre-compiled bundle — CI rebuilds it; reviewers testing locally should rebuild + hard-refresh.
- Suggested regression coverage (per the report): v1/v2 submission without a token, with an invalid/expired token, with a token issued for another survey, and a successful submission after valid PIN entry.


pandeymangg (approved):

icarito-labo:
This seems to have landed in Formbricks production today, and it's a breaking change.  It broke our integration.
How can we know in advance to prepare for breaking changes in the API?

PR #8507, opened 2026-07-13
fix: prevent infinite re-render loop when authenticated actions refresh the route

pandeymangg:
## What does this PR do?

Fixes an infinite re-render / refetch loop on the **survey summary page** (and, from the same root cause, the **Google Sheets integration** settings), where the client fired a server action + a full route refresh several times per second — pegging the dev server and spamming the network tab.

### Root cause

Every authenticated server action goes through `authenticatedActionClient`, whose middleware calls Better Auth `getSession()`. On the durable-read path this **re-sets the session cookie**, and `nextCookies()` writes it via `cookies().set()` _during the action_. In Next.js, `cookies().set()` inside a Server Action invalidates the router cache, so Next issues a follow-up **route refresh** that re-renders the server page.

That refresh hands the client component **fresh object references** for its server props (`survey`, `initialSurveySummary`, `googleSheetIntegration`). When a `useEffect` that calls the action also lists those references in its dependency array (directly or via a `useCallback`), React — which compares deps **by reference, not by value** — sees "deps changed" and re-runs the effect → another action → another refresh → **loop**.

```
effect → action → session cookie set → route refresh → new prop references
   → React thinks deps changed → effect re-runs → action → … 🔁
```

This trigger is **systemic** (any authenticated action can refresh its route), but it only bites components that _fetch inside an effect keyed on a server-derived object_. An app-wide scan found just the two touched here; most actions are event-handler-driven (a single, harmless refresh) or key their effects on primitives.

### Changes

- **`SummaryPage.tsx`** — the fetch effect now dedupes by the actual **filter values** (a serialized key kept in a ref) instead of re-firing on prop-reference churn from a route refresh. Also fixed a dead guard (`hasNoFilters` was always `false` because the default date range always sets `to: today`), so the server-provided initial summary is reused when no filters are active. The dedupe key is rolled back on fetch failure so re-selecting the same filter still retries.
- **`GoogleSheetWrapper.tsx`** — dropped the server-derived `googleSheetIntegration` object from `validateConnection`'s deps (now `[workspaceId, isConnected]`, both stable), mirroring the already-safe `SlackWrapper` pattern. The `!isConnected` guard already implies an integration exists, and the action only needs `workspaceId`.

### Follow-up (intentionally not in this PR)

The durable fix lives in the auth layer — avoid re-setting the session cookie on pure reads in `action-client` / `session.ts` — which would eliminate this whole class of loop app-wide. Left out here because it touches authentication globally and deserves its own reviewed change.

## How should this be tested?

**Survey summary page**
- Open a survey's **Summary** tab → Network tab should be quiet (one summary fetch; no repeating `POST` / `GET .../summary` pairs).
- Apply a date filter (e.g. **Last 7 days**) → exactly one refetch, no loop.
- Switch ranges, then **Clear all** → shows the unfiltered summary, no loop.
- Click the **Refresh** button → still refetches the current view.

**Google Sheets integration** (Workspace → Settings → Integrations → Google Sheets, with a connected account)
- Open the page → the connection is validated once; no repeating `validateGoogleSheetsConnection` calls.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/google-sheets/components/GoogleSheetWrapper.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/google-sheets/components/GoogleSheetWrapper.tsx
@@ -41,13 +41,19 @@ export const GoogleSheetWrapper = ({
     (TIntegrationGoogleSheetsConfigData & { index: number }) | null
   >(null);
 
+  // Depend only on stable values (workspaceId is a primitive; isConnected is local state). The
+  // server-derived `googleSheetIntegration` object must NOT be a dependency: every authenticated
+  // server action re-sets the Better Auth session cookie, which refreshes the route and hands this
+  // component a new `googleSheetIntegration` reference — depending on it would re-fire the action in
+  // an infinite loop. `isConnected` already implies an integration exists, and the action only needs
+  // `workspaceId`, so the object isn't needed here.
   const validateConnection = useCallback(async () => {
-    if (!isConnected || !googleSheetIntegration) return;
+    if (!isConnected) return;
     const response = await validateGoogleSheetsConnectionAction({ workspaceId });
     if (response?.serverError === GOOGLE_SHEET_INTEGRATION_INVALID_GRANT) {
       setShowReconnectButton(true);
     }
-  }, [workspaceId, isConnected, googleSheetIntegration]);
+  }, [workspaceId, isConnected]);
 
   useEffect(() => {
     validateConnection();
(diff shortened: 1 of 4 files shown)


Dhruwang (commented):
ResponseFilter.tsx:89-111 — same anti-pattern, NOT in this PR:

```
useEffect(() => {
  const handleInitialData = async () => {
    if (isOpen) {
      const surveyFilterData = await getSurveyFilterDataAction({ surveyId: survey.id });
      ...
    }
  };
  handleInitialData();
}, [isOpen, setSelectedOptions, survey]);   // <- survey is server-derived object
```

Changes looks good. Can you confirm if we need to also fix it here?

Dhruwang (approved):

PR #8509, opened 2026-07-13
fix: let MCP clients request surveys:write and correct validate_survey scope

xernobyl:
## What does this PR do?

Fixes a scope mismatch found during 5.2 staging QA: connecting Claude (or any MCP client) to `/api/mcp` over OAuth only ever granted `surveys:read`, so every write tool (`create_survey`, `patch_survey`, `delete_survey`) failed with `403 "OAuth token does not include the required MCP scope"` and the consent screen offered only "Read surveys". Two separate root causes, fixed here:

**1. Write was never offered at consent.** The RFC 9728 protected-resource metadata already listed `surveys:write`, but the 401 `WWW-Authenticate` challenge advertised only `surveys:read` and `surveys:write` was not a default DCR scope. Clients that build their request from the challenge (rather than the PRM) therefore registered/authorized read-only and could never reach the write tools. Fixed with two changes that have to move together (change one without the other and you get `invalid_scope`, same failure mode as the earlier `offline_access` bug):

- The 401 challenge now advertises `surveys:read surveys:write` (built from `MCP_RESOURCE_SCOPES`, not hard-coded).
- `surveys:write` is now a **default** DCR scope, so clients are registered for it and the provider validates the write authorize request against their registration.

The minimum scope required to authenticate stays `surveys:read`. Granting the write scope is safe: actual write access is still enforced downstream by the user's workspace permission in the v3 layer (`requireV3WorkspaceAccess`), so a user without `write`/`manage` still gets a 403 there even with the scope.

**2. `validate_survey` was mis-scoped.** It is annotated `readOnlyHint` and described as validating "without writing survey changes", but it required `surveys:write` for create/patch payloads (the only two operations its schema allows), so a read-scoped client couldn't even dry-run a validation. It now requires only `surveys:read`. This makes the code match what the docs already state (overview.mdx lists `validate_survey` under `surveys:read`).

No schema/migration/env changes.

## How should this be tested?

Automated (from `apps/web`):

```
pnpm test -- modules/mcp modules/auth/lib app/.well-known app/api/mcp
```

242 tests pass, including a new DCR test asserting a default (no-scope) registration grants read + write, and updated challenge-scope assertions.

Manual / live (verified on a dev server off this branch):

1. `curl -X POST http://localhost:3000/api/mcp -i` → `WWW-Authenticate` advertises `scope="surveys:read surveys:write"`.
2. Register a client with no scope:
   `curl -X POST http://localhost:3000/api/auth/oauth2/register -H 'content-type: application/json' -d '{"client_name":"t","redirect_uris":["http://localhost:9999/cb"],"grant_types":["authorization_code","refresh_token"],"response_types":["code"],"token_endpoint_auth_method":"none"}'` → returned `scope` includes `surveys:write`.
3. Connect via `npx -y mcp-remote http://localhost:3000/api/mcp` (or Claude Code `/mcp`) → the consent screen shows **both** "Read surveys" and "Write surveys" → `create_survey` succeeds for a user with workspace write permission, and still 403s (at the v3 layer) for a user without it.
4. `validate_survey` with a read-only token now succeeds instead of 403.

### Required

- [ ] Ran `pnpm build`

### Appreciated

- [x] Updated the Formbricks Docs if changes were necessary (docs already described this behavior; code now matches)

### DIFF
--- a/apps/web/app/api/mcp/route.test.ts
+++ b/apps/web/app/api/mcp/route.test.ts
@@ -43,6 +43,7 @@ vi.mock("@formbricks/database", () => ({
 }));
 
 vi.mock("@/modules/auth/lib/oauth-urls", () => ({
+  MCP_RESOURCE_SCOPES: ["surveys:read", "surveys:write"],
   getAuthIssuerUrl: () => "http://localhost/api/auth",
   getMcpOrigin: () => "http://localhost",
   getMcpProtectedResourceMetadataUrl: () => "http://localhost/.well-known/oauth-protected-resource/api/mcp",
@@ -160,7 +161,7 @@ describe("POST /api/mcp", () => {
     expect(response.status).toBe(401);
     expect(response.headers.get("Content-Type")).toBe("application/problem+json");
     expect(response.headers.get("WWW-Authenticate")).toBe(
-      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read"'
+      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read surveys:write"'
     );
     expect(applyIPRateLimit).toHaveBeenCalled();
   });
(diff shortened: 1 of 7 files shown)


BhagyaAmarasinghe (approved):
LGTM

PR #8510, opened 2026-07-13
fix: let MCP clients request surveys:write and correct validate_survey scope (5.2 backport)

xernobyl:
## What does this PR do?

Backport of #8509 to `release/5.2` — 5.2 is the first release that ships MCP OAuth (ENG-1055), so the QA fix has to ride along.

Fixes a scope mismatch found in 5.2 staging QA: connecting an MCP client to `/api/mcp` over OAuth only ever granted `surveys:read`, so every write tool (`create_survey`, `patch_survey`, `delete_survey`) failed with `403 "OAuth token does not include the required MCP scope"` and the consent screen offered only "Read surveys". Two root causes:

1. **Write was never offered at consent.** The 401 `WWW-Authenticate` challenge advertised only `surveys:read` and `surveys:write` wasn't a default DCR scope, so clients registered/authorized read-only and could never reach the write tools. Now the challenge advertises `surveys:read surveys:write` (built from `MCP_RESOURCE_SCOPES`) **and** `surveys:write` is a default DCR scope. The minimum scope to authenticate stays `surveys:read`. Safe because real write access is still enforced downstream by the user's workspace permission in the v3 layer.
2. **`validate_survey` was mis-scoped** — annotated `readOnlyHint` / "without writing survey changes" but required `surveys:write`; now requires only `surveys:read`.

**Source-only backport:** the test changes from #8509 are intentionally omitted per our backport convention. Note this means the existing MCP tests on `release/5.2` still assert the pre-fix behavior (challenge scope, validate-requires-write) and will be red until re-cut — expected for a backport; merge accordingly.

## How should this be tested?

Verified on `main` (#8509) via unit tests (242 passing) and a live `mcp-remote` run against a dev server, and confirmed locally by @xernobyl: the consent screen shows **both "Read surveys" and "Write surveys"**, and `create_survey` / `validate_survey` succeed for a user with workspace write permission (and still 403 at the v3 layer without it).

On 5.2:
1. `curl -X POST <origin>/api/mcp -i` → `WWW-Authenticate` advertises `scope="surveys:read surveys:write"`.
2. Register a client with no scope → returned `scope` includes `surveys:write`.
3. Connect via `mcp-remote` → consent offers write → write tools work (subject to workspace permission).

### Required

- [ ] Ran `pnpm build`

### DIFF
--- a/apps/web/app/api/mcp/route.test.ts
+++ b/apps/web/app/api/mcp/route.test.ts
@@ -43,6 +43,7 @@ vi.mock("@formbricks/database", () => ({
 }));
 
 vi.mock("@/modules/auth/lib/oauth-urls", () => ({
+  MCP_RESOURCE_SCOPES: ["surveys:read", "surveys:write"],
   getAuthIssuerUrl: () => "http://localhost/api/auth",
   getMcpOrigin: () => "http://localhost",
   getMcpProtectedResourceMetadataUrl: () => "http://localhost/.well-known/oauth-protected-resource/api/mcp",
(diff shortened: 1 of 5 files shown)


Dhruwang (approved):
Thank you for the PR, looks good 🚀

PR #8512, opened 2026-07-13
fix(database): seed Better Auth credential accounts so seed users can log in

itsjavi:
## What does this PR do?

Fixes email/password login for seeded users on fresh installs.

`packages/database/src/seed.ts` was missed in the NextAuth → Better Auth migration (ENG-1054): it still wrote the bcrypt hash to `User.password` only. Better Auth verifies credential sign-in against a `credential` `Account` row (`provider = "credential"`, `providerAccountId = <userId>`, password hash on `Account.password`), so logging in with any seed user failed with `Credential account not found`.

The ENG-1054 backfill migration (`20260619120000_eng_1054_credential_account_backfill`) doesn't cover this case: on a fresh install migrations run *before* the seed, so it runs against an empty `User` table and is then marked as applied.

Changes:

- The seed now upserts a `credential` Account row per seed user (same shape as the backfill migration), so Better Auth login works. Re-running the seed on an already-seeded broken DB also repairs it, since the Account upsert runs even when the user already exists.
- The three copy-pasted user upserts are folded into a single loop over a seed-user list.

Note: `User.email_verified_at` being `NULL` on seeded users is expected — it's a rollback-safety temp column from the ENG-1054 cutover migration, not read by anything; the real field is the `email_verified` boolean, which the seed sets to `true`.

## How should this be tested?

- On a fresh DB: `pnpm db:migrate:dev`, then `pnpm db:seed` (in `packages/database`).
- Verify three `Account` rows exist with `provider = 'credential'`, `providerAccountId` = the seed user ids, and a bcrypt hash in `password`.
- Start the app (`pnpm dev`) and log in at `/auth/login` with `admin@formbricks.com` / the password from `packages/database/src/seed/constants.ts` — you should land on the Surveys dashboard. Same for the manager and member users.

Verified locally: re-ran the seed against an existing broken DB, confirmed the Account rows via Prisma, and logged in through the browser as the admin user successfully.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/packages/database/src/seed.ts
+++ b/packages/database/src/seed.ts
@@ -382,62 +382,64 @@ async function main(): Promise<void> {
   // Users
   const passwordHash = await hashPassword(SEED_CREDENTIALS.ADMIN.password, 10);
 
-  await prisma.user.upsert({
-    where: { id: SEED_IDS.USER_ADMIN },
-    update: {},
-    create: {
+  const seedUsers = [
+    {
       id: SEED_IDS.USER_ADMIN,
       name: "Admin User",
       email: SEED_CREDENTIALS.ADMIN.email,
-      password: passwordHash,
-      emailVerified: true,
-      memberships: {
-        create: {
-          organizationId: organization.id,
-          role: "owner",
-          accepted: true,
-        },
-      },
+      role: "owner",
     },
-  });
-
-  await prisma.user.upsert({
-    where: { id: SEED_IDS.USER_MANAGER },
-    update: {},
-    create: {
+    {
       id: SEED_IDS.USER_MANAGER,
       name: "Manager User",
       email: SEED_CREDENTIALS.MANAGER.email,
-      password: passwordHash,
-      emailVerified: true,
-      memberships: {
-        create: {
-          organizationId: organization.id,
-          role: "manager",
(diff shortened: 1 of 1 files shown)


xernobyl (approved):
LGTM ✅ — nice catch on the seed drifting from the Better Auth contract.

The fix is spot on: the `credential` Account row shape (provider/providerAccountId = user id, bcrypt hash copied verbatim onto `Account.password`) matches the ENG-1054 backfill migration exactly, so it's the same thing the cutover did for existing users. Doing it as an upsert is a nice touch — it repairs already-seeded broken DBs on a re-run, not just fresh installs. And folding the three copy-pasted upserts into a loop is a good cleanup.

One small non-blocking follow-up (P3), not for this PR: the note about `User.email_verified_at` being an expected-NULL column is correct — but I'd argue that's the actual smell. It's a rollback-safety temp column we added during the cutover (`20260620000000_eng_1054_...`), it's no longer in the Prisma model, and nothing reads it — so it's just orphaned in the DB. Rather than documenting a dead field, I think we should drop it in a small follow-up migration so we stop accumulating unused columns. Happy to file that separately so it doesn't hold this up.

PR #8513, opened 2026-07-13
fix(web): stop infinite server-action refresh loop on survey analysis pages

itsjavi:
## What does this PR do?

Fixes an infinite POST/refresh loop on the survey summary page (`/workspaces/[workspaceId]/surveys/[surveyId]/summary`): the page endlessly alternates server-action POSTs and RSC refresh GETs (visible in the network tab), hammering the server and constantly re-rendering.

### Root cause

Two bugs compound:

1. **The proxy sets the `formbricks-workspace-id` cookie on every `/workspaces/*` request** — including server-action POSTs. When a cookie is set on a server-action response, Next.js marks the action as revalidated (`x-action-revalidated: 1`) and the client router refetches the page's RSC payload.
2. **`SummaryPage`'s "skip fetch when no filters" guard never matches.** The filter context's default date range is `{ from: undefined, to: getTodayDate() }`, and the guard checked `!dateRange.from && !dateRange.to` — always false. Meanwhile `getFormattedFilters` only applies a date filter when *both* `from` and `to` are set, so the pristine state actually produces empty criteria. Result: the summary effect always fetched on mount even though the page already server-renders `initialSurveySummary`.

The loop: mount effect fires `getSurveySummaryAction` → the middleware cookie write flags the action as revalidated → the router refetches the RSC page → new `survey` / `initialSurveySummary` object identities re-trigger the effect → another POST → another refresh → forever.

### Fix

- `apps/web/proxy.ts`: only set the workspace cookie when its value actually changes (first visit or workspace switch). No more cookie writes on every action POST → no spurious revalidation/refresh.
- `SummaryPage.tsx`: derive "no filters" from the actual `getFormattedFilters` output (empty criteria object) instead of the never-true date check, and restore `initialSurveySummary` when filters are cleared. This removes the redundant mount fetch entirely and keeps the page loop-proof even if some future middleware writes cookies again.
- `apps/web/proxy.test.ts`: cover the new cookie behavior (no re-set on same value, update on workspace switch).

## How should this be tested?

- Log in, open a survey summary page, and watch the network tab: no repeating POST + `?_rsc=` GET pairs; on mount there is no `getSurveySummaryAction` call at all (server-rendered data is used).
- Click the Refresh icon: exactly one `getSurveySummaryAction` POST, no cascade.
- Apply a date filter (e.g. "Last 7 days"): exactly one POST with `createdAt` criteria, no cascade. Reset back to "All time": summary returns to the unfiltered data.
- Repeat requests to `/workspaces/*` no longer carry a `set-cookie: formbricks-workspace-id=...` header (only the first visit / workspace switch does); the org-settings shell still resolves the active workspace.
- `pnpm exec vitest run proxy.test.ts` in `apps/web` (8 tests pass).

Verified locally in the browser (network trace before/after) plus lint and `pnpm typecheck` on `apps/web`.

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/SummaryPage.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/SummaryPage.tsx
@@ -160,12 +160,13 @@ export const SummaryPage = ({
 
   // Only fetch data when filters change or when there's no initial data
   useEffect(() => {
-    // If we have initial data and no filters are applied, don't fetch
-    const hasNoFilters =
-      (!selectedFilter || Object.keys(selectedFilter).length === 0 || selectedFilter.filter?.length === 0) &&
-      (!dateRange || (!dateRange.from && !dateRange.to));
+    // If we have initial data and no filters are applied, don't fetch. "No filters" must mirror
+    // getFormattedFilters: the default date range ({ from: undefined, to: today }) produces empty
+    // criteria, so checking dateRange fields directly would treat the pristine state as filtered.
+    const hasNoFilters = Object.keys(getFormattedFilters(survey, selectedFilter, dateRange)).length === 0;
 
     if (initialSurveySummary && hasNoFilters) {
+      setSurveySummary(initialSurveySummary);
       setIsLoading(false);
       return;
     }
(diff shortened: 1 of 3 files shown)


Dhruwang (dismissed):

Dhruwang (approved):
Thank you for the PR, looks good. 🚀.

PR #8530, opened 2026-07-15
fix(account): redirect SSO account deletion to the post-deletion survey (ENG-1780)

Dhruwang:
## Problem

SSO (email-link) account deletion always redirected to app home `/`, ignoring the post-deletion redirect the credential delete path applies. The link in `better-auth-account-deletion-request.ts` hardcoded `callbackURL=/`.

Credential deletes already branch the redirect in `DeleteAccountModal` (survey on Formbricks Cloud, `/auth/login` otherwise). The SSO path skipped that branch — a regression from the Better Auth migration (ENG-1054).

## Solution

Set `callbackURL` conditionally in the SSO deletion request, matching the credential path:

```ts
const callbackURL = IS_FORMBRICKS_CLOUD ? FORMBRICKS_CLOUD_ACCOUNT_DELETION_SURVEY_URL : "/auth/login";
```

The survey is served from the app's own origin, so on Formbricks Cloud it satisfies Better Auth's same-origin `trustedOrigins` check on `delete-user/callback`. The value is URL-encoded.

## QA note

Better Auth rejects **cross-origin** `callbackURL`s (`INVALID_CALLBACK_URL`). This only bites in a local build with the Cloud flag forced on (`WEBAPP_URL=localhost` ≠ survey origin). In production Cloud the origins match. To verify the survey redirect end-to-end, test where `WEBAPP_URL` shares the survey's origin.

## Tests

Updated `better-auth-account-deletion-request.integration.test.ts` to assert the self-hosted callback (`/auth/login`, since the Cloud flag is off under test).

Closes ENG-1780

### DIFF
--- a/apps/web/modules/account/lib/better-auth-account-deletion-request.integration.test.ts
+++ b/apps/web/modules/account/lib/better-auth-account-deletion-request.integration.test.ts
@@ -65,7 +65,8 @@ describe("requestSsoAccountDeletionEmail (real Postgres)", () => {
     const mailArgs = sendDeleteAccountConfirmationEmailMock.mock.calls[0][0];
     expect(mailArgs.email).toBe(email);
     expect(mailArgs.deleteLink).toContain("/api/auth/delete-user/callback?token=");
-    expect(mailArgs.deleteLink).toContain("callbackURL=/");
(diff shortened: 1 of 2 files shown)


itsjavi (approved):
thanks, it worked 🚢

PR #8534, opened 2026-07-15
fix(billing): refresh plan state after a successful upgrade (ENG-1790)

Dhruwang:
## Problem

After a successful upgrade, the billing page kept showing the previous plan until a manual reload. The charge and plan change are applied correctly server-side — this is a client-refresh gap on the post-checkout return.

Two return paths were affected:

1. **Finalize path** (`pricing-table.tsx`): the success toast fired unconditionally (ignoring whether `waitForBillingPlanAction` actually observed the new plan), and `router.replace()` to the bare billing URL could serve a stale prefetched RSC even after `router.refresh()`.
2. **First-subscription path** (`ConfirmationPage.tsx`): the *Back to billing overview* button was a `next/link`, so it served the billing page's prefetched (pre-upgrade) RSC — no refresh.

## Solution

**Finalize path**
- Gate the success toast on the plan `waitForBillingPlanAction` actually observed. If Stripe hasn't reflected it within the poll window, show the existing `upgrade_checkout_pending` message instead of a false "You're now on the … plan."
- Strip the one-time checkout query params with `window.history.replaceState` (no router navigation → no stale prefetched RSC), then `router.refresh()` to refetch the billing route with the synced plan.

**First-subscription path**
- Use a full-document navigation (`<a>`) for the back button so the billing page re-renders fresh with the upgraded plan.

No new i18n keys. No server/action changes.

## Testing

No `.tsx` unit tests (per repo convention — React covered by E2E). Manual QA on staging: create org → upgrade to Pro via Stripe checkout on both paths → billing reflects Pro without a manual reload.

Closes ENG-1790

### DIFF
--- a/apps/web/app/(app)/billing-confirmation/components/ConfirmationPage.tsx
+++ b/apps/web/app/(app)/billing-confirmation/components/ConfirmationPage.tsx
@@ -2,15 +2,23 @@
 
 import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
+import { waitForBillingPlanAction } from "@/modules/ee/billing/actions";
 import { Button } from "@/modules/ui/components/button";
 import { Confetti } from "@/modules/ui/components/confetti";
 
 const BILLING_CONFIRMATION_ORGANIZATION_ID_KEY = "billingConfirmationOrganizationId";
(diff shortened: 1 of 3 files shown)


pandeymangg (approved):
LGTM 🚀

PR #8536, opened 2026-07-15
fix(api): reject response updates on a closed survey (ENG-1654)

Dhruwang:
## Problem

The client `PUT /api/v1/client/{workspaceId}/responses/{responseId}` endpoint only checked whether the response was already finished — never the survey's status. A partial response created while a survey was open could be updated and marked `finished: true` after the survey closed. The POST endpoint already rejects submissions on a closed survey (403); PUT didn't.

Impact: late finishes on closed surveys, quota consumed after close, analytics with post-close timestamps.

## Fix

In `validateUpdateRequest`, reject with 403 when `survey.status !== "inProgress"`, mirroring the POST endpoint's check and message. The v2 client PUT re-exports this handler, so both v1 and v2 client PUT are covered.

## Scope — client vs management endpoints

Audited all response create/update endpoints:

| Endpoint | Survey-status gate |
| --- | --- |
| v1 client POST | already present |
| v1 client PUT | **fixed here** |
| v2 client POST | already present (`checkSurveyValidity`) |
| v2 client PUT | **fixed here** (re-exports the v1 handler) |
| v1/v2 management POST + PUT | intentionally not gated — see below |

The **management** (API-key) endpoints do **not** gate on survey status, and this PR deliberately leaves that unchanged. They're a different trust model: authenticated survey owners / their integrations, with legitimate reasons to write to a closed survey (imports, backfills, corrections). The client check exists precisely because respondents are untrusted. Restricting management writes on closed surveys would be a separate product decision, not part of this bug. Flagging for the team — happy to open a follow-up if we want to change that.

## Tests

- Unit: added a case asserting 403 `"Survey is not accepting submissions"` when the survey is closed; gave the base survey mock `status: "inProgress"`. `pnpm test put-response-handler` → 21 passed.
- Manual API (local dev): reproduced the exact issue flow —
  - POST partial (survey `inProgress`) → 200
  - close survey (`status = 'completed'`) → PUT finalize → **403** (was 200 before the fix)
  - reopen survey → PUT → 200 (happy path intact)

Closes ENG-1654

### DIFF
--- a/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/put-response-handler.test.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/put-response-handler.test.ts
@@ -108,6 +108,7 @@ const getBaseSurvey = () =>
   ({
     id: surveyId,
     workspaceId,
+    status: "inProgress",
     blocks: [],
     questions: [],
   }) as const;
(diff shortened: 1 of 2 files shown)


xernobyl (approved):
LGTM

PR #8537, opened 2026-07-15
fix(api): allow displayLimit recontact + independent displayPercentage in v3 surveys

pandeymangg:
## What does this PR do?

Fixes the v3 Surveys API app-survey `distribution` validation, which wrongly coupled `displayOption: "displaySome"` to `displayPercentage`. Found during QA 5.2 §7 (ENG-1757).

Linear: ENG-1794

**Before:**
- `displaySome` ("show a limited number of times") **required** `displayPercentage`, so the natural payload for it — `{"displayOption":"displaySome","displayLimit":3}` — returned **400** (`missing_required_field` on `distribution.displayPercentage`). You couldn't configure the "show at most N times" recontact option via the API.
- `displayPercentage` was **rejected** (`unsupported_field`) on every other option (`displayOnce`/`displayMultiple`/`respondMultiple`).

This contradicted the source of truth: in `ZSurvey` (`packages/types/surveys/types.ts`) and the js-core runtime (`packages/js-core/src/lib/common/utils.ts`), `displaySome` is capped by **`displayLimit`**, and `displayPercentage` is an **independent** throttle applied regardless of `displayOption`. The editor (`recontact-options-card.tsx`) reflects this too.

**Change** — `apps/web/app/api/v3/surveys/schemas.ts` → `addAppDistributionIssues`:
- Removed the `displaySome ⇄ displayPercentage` coupling. `displayPercentage` (0.01–100 or null) is now accepted with **any** `displayOption`.
- Added a guard: `displayOption: "displaySome"` now requires `displayLimit >= 1`, mirroring the editor's "show survey maximum of N times" input (so a "limited" survey can't be created without a real limit).
- Link surveys still reject `distribution`/`targeting`; all per-field constraints (enum, ranges) are unchanged.

## How should this be tested?

Unit tests (77 pass, incl. new cases for the fixed + guarded behavior):

```
pnpm --filter=@formbricks/web exec vitest run app/api/v3/surveys/schemas.test.ts app/api/v3/surveys/distribution.test.ts app/api/v3/surveys/patch.test.ts app/api/v3/surveys/create.test.ts
```

Manual (`POST /api/v3/surveys`, `type: "app"`), verified on localhost:

| body `distribution` | before | after |
| --- | --- | --- |
| `{"displayOption":"displaySome","displayLimit":3}` | 400 | **201** |
| `{"displayOption":"displayOnce","displayPercentage":50}` | 400 | **201** |
| `{"displayOption":"displaySome","displayLimit":3,"displayPercentage":75}` | 400 | **201** |
| `{"displayOption":"displaySome"}` (no limit) | 400 | **400** (`missing_required_field` on `distribution.displayLimit`) |
| `{"displayOption":"displayOnce"}` on a `link` survey | 400 | **400** (unchanged) |
| `{"displayOption":"displayOnce","displayPercentage":150}` | 400 | **400** (range, unchanged) |

### Required

- [ ] Ran `pnpm build`
- [x] My changes don't cause any responsiveness issues (API-only, no UI change)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (N/A — API-only)

### DIFF
--- a/apps/web/app/api/v3/surveys/schemas.ts
+++ b/apps/web/app/api/v3/surveys/schemas.ts
@@ -1118,9 +1118,11 @@ const ZV3SurveyDistribution = z
 const ZV3SurveyTargeting = z.object({ filters: ZSegmentFilters }).strict();
 
 /**
- * Body-level validation for app-survey distribution/targeting. `distribution`/`targeting` are
- * rejected for explicit `link` surveys; for `app` (or when the type is unknown, e.g. the generic
- * patch schema) the displayOption ↔ displayPercentage coupling is enforced.
+ * Body-level validation for app-survey distribution/targeting: `distribution`/`targeting` are
+ * app-only, so reject them on explicit `link` surveys. Within `distribution`, `displaySome` ("show a
+ * limited number of times") must carry a `displayLimit` >= 1 (mirrors the editor's required "show
+ * maximum of N times" input). `displayPercentage` is an INDEPENDENT throttle valid with ANY
+ * `displayOption` — never coupled to `displaySome` — matching `ZSurvey` and the js-core runtime.
  */
 function addAppDistributionIssues(
   survey: {
@@ -1150,26 +1152,20 @@ function addAppDistributionIssues(
     return;
   }
 
+  // "displaySome" is capped by displayLimit, so a "limited" survey must specify a real limit (>= 1),
+  // mirroring the editor. displayPercentage stays independent and is validated by the field schema.
   const distribution = survey.distribution;
-  if (!distribution) {
-    return;
-  }
-
-  if (distribution.displayOption === "displaySome") {
-    if (distribution.displayPercentage === null || distribution.displayPercentage === undefined) {
-      ctx.addIssue({
-        code: "custom",
-        message: "displayPercentage is required when displayOption is 'displaySome'",
-        params: { code: "missing_required_field" },
-        path: ["distribution", "displayPercentage"],
-      });
-    }
-  } else if (distribution.displayPercentage !== null && distribution.displayPercentage !== undefined) {
+  if (
+    distribution?.displayOption === "displaySome" &&
+    (distribution.displayLimit === null ||
+      distribution.displayLimit === undefined ||
+      distribution.displayLimit < 1)
+  ) {
     ctx.addIssue({
       code: "custom",
-      message: "displayPercentage is only allowed when displayOption is 'displaySome'",
-      params: { code: "unsupported_field" },
-      path: ["distribution", "displayPercentage"],
+      message: "displayLimit must be at least 1 when displayOption is 'displaySome'",
+      params: { code: "missing_required_field" },
+      path: ["distribution", "displayLimit"],
     });
   }
 }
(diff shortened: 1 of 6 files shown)


apps/web/app/api/v3/surveys/schemas.ts
```diff
@@ -1150,26 +1152,20 @@ function addAppDistributionIssues(
...
+      distribution.displayLimit === undefined ||
+      distribution.displayLimit < 1)
+  ) {
     ctx.addIssue({
       code: "custom",
-      message: "displayPercentage is only allowed when displayOption is 'displaySome'",
-      params: { code: "unsupported_field" },
-      path: ["distribution", "displayPercentage"],
+      message: "displayLimit must be at least 1 when displayOption is 'displaySome'",
+      params: { code: "missing_required_field" },
```
xernobyl, line 1167: [P3, optional] When `displayLimit: 0` is explicitly sent with `displaySome`, the field is *present* but out of range, yet the error carries `code: missing_required_field` — semantically it's a range violation. Minor (the message text is clear and the field-level schema does permit `0` via `nonnegative()`), so fine to leave for consistency with the "needs a limit" framing; flagging only in case you'd rather split the genuinely-missing case from the zero case.

xernobyl (commented):
Reviewed the code + ran a security pass. The fix itself is correct — removing the `displaySome ⇄ displayPercentage` coupling and requiring `displayLimit >= 1` for `displaySome` matches all three sources of truth (`ZSurvey` fields are independent, js-core gates `displaySome` on `displayLimit`, and the editor's "show maximum of N times" input is `min="1"`). Tests 77/77 green.

**[P2] The OpenAPI v3 contract wasn't updated and now misdocuments this behavior.** `docs/api-v3-reference/` still describes the old coupling (these files aren't in this diff, hence the review comment rather than inline):

- `src/components/schemas/SurveyDistribution.yml` — `displayOption` description says *"`displaySome` shows it to `displayPercentage` of contacts"* (wrong — it's capped by `displayLimit`), and `displayPercentage` says *"Required when `displayOption` is `displaySome`; must be null/omitted otherwise"* (now false).
- Same file's canonical **`example`** is `displayOption: displaySome` + `displayPercentage: 50` + `displayLimit: null` with a comment restating the old rule — under the new validation that exact payload returns **400** (`displayLimit must be at least 1`). The documented example is now an invalid request.
- Error-code prose still stating the old rule: `openapi.yml` (~L419, ~L1015) and the source files `src/paths/api_v3_surveys.yml` and `api_v3_surveys_{surveyId}.yml`.

Suggest fixing the `src/` files and re-bundling (`pnpm api:v3:*`), with the example changed to e.g. `displaySome` + `displayLimit: 3` (plus optional independent `displayPercentage`).

**Security:** no findings. Single-surface change (input validation only) — gitleaks clean, no DB/log/Sentry/outbound/secret/error-shaping surface touched. The relaxation only widens two already-bounded fields (`displayPercentage` 0.01–100, `displayLimit` int≥0) on an app-only path; link surveys still reject `distribution` wholesale, so nothing new is exploitable.

xernobyl (approved):
Re-reviewed at `11010b311` — the OpenAPI contract drift is fully resolved across all four files (SurveyDistribution schema + example, both 400 error blocks), the example is now a payload that actually validates, and `pnpm api:v3:check`/`api:v3:lint` pass with the CI bundle-freshness gate green. Code is unchanged and correct. The remaining `missing_required_field`-code note on the `displayLimit: 0` case is optional and not merge-blocking. 🚀

PR #8540, opened 2026-07-15
fix(charts): scale rating chart Y-axis to the question's scale

itsjavi:
## What does this PR do?

Contributes to ENG-1796

Rating-scale charts (Rating/CSAT/CES/NPS averages) currently get a data-driven "nice" Y-axis, so the axis top lands on an arbitrary value instead of the question's scale. QA on ENG-1754 (TC2) observed:

| Chart | Value | Axis before | Axis after |
| --- | --- | --- | --- |
| Rating: Average, 1–5 question | 3.33 | 0–4 | **0–5** |
| Rating: Average, 1–10 question | 6.33 | 0–8 | **0–7** (see decision below — the true 1–10 scale is not stored) |
| CSAT: Average | 2.4 | 0–3 | **0–5** |
| NPS: Average | 6.33 | 0–8 | **0–10** |

Changes:

- `schema-definition.ts`: new optional `axisMaxCandidates` on `MeasureDefinition` — the candidate axis maxima for measures answered on a bounded scale: `ratingAverage: [5, 7, 10]`, `cesAverage: [5, 7]`, `csatAverage: [5]`, `npsAverage: [10]` — plus a `getMeasureAxisMaxCandidates` lookup.
- `y-axis-scale.ts` (new): the Y-axis domain/tick logic extracted from `cartesian-chart.tsx` into a pure `.ts` module (so it can be unit-tested per our testing rules). When **every** series in a chart is a fixed-scale measure, the axis pins to `[0, smallest candidate ≥ each series' max]` with integer ticks that end exactly on the scale max (0–5 → step 1, 0–7 → step 1, 0–10 → step 2). Everything else keeps the existing nice scaling byte-for-byte, and pinning defensively falls back to nice scaling if data is negative or exceeds the largest known scale (never clips real data).
- `cartesian-chart.tsx`: now imports `computeYAxis` from the new module; no behavioral change of its own.

Non-goals / unchanged: counts, NPS: Score (−100..100), Sentiment: Average (−1..1, needs a pinned *negative* min so it's a different mechanism — left as data-driven, which QA TC7 showed renders −1..1 correctly already), pie/big-number (no axis), and mixed charts (e.g. Rating Average + Responses) which keep nice scaling so a 5-cap never squashes a count series.

## Decision needed (@dhruwang)

**Where does the question's scale come from?** I investigated and it is currently **not stored anywhere the charts can reach**:

- Ingestion (`apps/web/lib/feedback-source/transform.ts`) writes rating answers as bare `value_number`; the survey element's `range` (3/4/5/6/7/10 in `@formbricks/types`) is dropped. `metadata` is only populated for matrix/ranking.
- The Cube schema (`FeedbackRecords.js`) consequently has no scale dimension/measure.

**Chosen approach (heuristic, chart-side only):** pin the axis max to the smallest *standard* scale (5, 7, 10) that contains the series max. Monotone, never clips, and fixes the reported cases without touching ingestion, Cube (both schema copies stay identical), or needing a backfill.

Known limitation: a 1–10 question whose average sits in (5, 7] renders a 0–7 axis (the ENG-1796 staging fixture with avg 6.33 shows 0–7, not the ticket's ideal 0–10), and low-scoring 1–10 questions render 0–5. The heuristic cannot distinguish these without the real scale.

**Alternatives rejected:**

1. **Store the scale at ingestion** (e.g. `metadata.scale_max` from the element `range`) **+ a Cube `MAX(scale)` measure + query/chart plumbing** — the *correct* long-term fix, but it spans transform.ts, both byte-identical Cube schema copies, query building, and needs a backfill for all existing records (staging/prod data would still misrender). Happy to do as a follow-up if you prefer; the chart-side candidate mechanism here would then consume the real scale instead of guessing.
2. **Drop 7 from the rating candidates (`[5, 10]`)** so the staging 6.33 case shows 0–10 — but Formbricks rating questions support a 1–7 range, which would then render on a misleading 0–10 axis. Trivial one-line change if you'd rather match the ticket examples exactly.
3. **Pin only `ratingAverage`** — CSAT (fixed 1–5) and NPS average (fixed 0–10) have *known* scales, so pinning them is strictly correct rather than heuristic; leaving them data-driven would keep the same QA smell on their charts. Included them deliberately; shout if you want the scope narrower.

## How should this be tested?

1. Workspace → Analysis → create a chart: measure **Rating: Average**, filter to a 1–5 rating question (e.g. via Field ID), bar chart. With an average of ~3.33 the Y-axis must top out at **5** with ticks 0,1,2,3,4,5 (before: 4).
2. Same with a 1–10 question averaging > 7: axis tops at **10** with ticks 0,2,4,6,8,10 (an average ≤ 7 pins to 7 — see decision above).
3. **CSAT: Average** chart: axis always 0–5. **NPS: Average**: axis always 0–10.
4. Mixed chart (Rating: Average + Responses): unchanged nice scaling driven by the count.
5. Counts/`NPS: Score`/`Sentiment: Average` charts: axis behavior unchanged.
6. Unit tests: `cd apps/web && pnpm exec vitest run modules/ee/analysis/charts/lib/y-axis-scale.test.ts` (19 tests: pinning per measure, multi-measure, fallbacks, edge cases).

### Required

- [ ] Ran `pnpm build`

### Appreciated

Verification run: `vitest run` on `y-axis-scale.test.ts` + `schema-definition.test.ts` → 56 passed; `eslint` + `prettier --check` clean on all 4 touched files; `tsc --noEmit` introduces no new errors (the 3 pre-existing `schema-definition.test.ts` `group` errors exist on main too). No full build / dev server run (scoped verification only).

Screenshot 2026-07-15 at 20 19 43
Screenshot 2026-07-15 at 20 19 59
Screenshot 2026-07-15 at 20 20 18

### DIFF
--- a/apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
+++ b/apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
@@ -3,6 +3,7 @@
 import { type ElementType, type ReactNode } from "react";
 import { CartesianGrid, XAxis, YAxis } from "recharts";
 import { formatXAxisTick } from "@/modules/ee/analysis/charts/lib/chart-utils";
+import { computeYAxis } from "@/modules/ee/analysis/charts/lib/y-axis-scale";
 import type { TChartDataRow } from "@/modules/ee/analysis/types/analysis";
 import type { ChartConfig } from "@/modules/ui/components/chart";
 import { ChartContainer, ChartLegend, ChartLegendContent, ChartTooltip } from "@/modules/ui/components/chart";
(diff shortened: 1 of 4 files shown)


Dhruwang (approved):

PR #8542, opened 2026-07-15
fix(charts): remove leading gap on category bar charts

itsjavi:
## What does this PR do?

Contributes to ENG-1796

Bar charts built from measures only (e.g. the 6 "Sentiment: …" or "Emotion: …" count measures with no grouping) rendered with a large empty gap before the first bar — the bars looked centered instead of starting at the left, as flagged in release QA (ENG-1754).

**Root cause:** a measure-only query (no dimension and no time grouping) returns a single result row with one column per measure. `ChartRenderer` rendered that row as N `<Bar>` series, so Recharts saw a single category band with N grouped bars and centered the band in the plot area — the "gap" is the empty left half of the lone category slot, and it grows with chart width.

**Fix:** for the measure-only case, pivot the measures into categories (`pivotMeasuresToCategories` in `chart-utils.ts`, colocated unit tests added). Each measure becomes its own category row (`{ measure, value, fill, tooltipLabel }`), rendered as a single-series bar chart, so the bars fill the x-axis from the left like any standard category bar chart. As a side effect this also resolves the PM's secondary asks from the ticket:

- Each bar is labelled by its (translated) measure name on the x-axis.
- Each bar shows its value on top via `LabelList` — zero-count measures display a visible "0".
- The tooltip is shared per category column, so hovering the slot of a zero-height bar still shows e.g. "Very positive — 0" (no ghost divs needed).
- Sentiment are not consistently sorted in the charts (see the "before" and "after" screenshots)

Charts **with** a dimension (e.g. Responses grouped by Sentiment) and multi-measure charts **with** a dimension (grouped series + legend) go through the untouched code path and are unchanged. Line/area/pie/big-number are unchanged.

Support changes:

- `CartesianChart`: new optional `tooltipHideLabel` prop to decouple the tooltip header from `hasCategoryAxis` (the pivoted chart has a real category axis but hides the redundant header, since each tooltip row carries the measure label).
- `PolishedChartTooltip`: a row-level `tooltipLabel` on the payload now takes precedence over deriving the label from the Cube column key (needed because the pivoted series' dataKey is the generic `value`).

> Note: I could not run the app in this environment — before/after screenshots need to be added by a human reviewer.

## How should this be tested?

1. Analysis → Charts → create a **bar** chart.
2. Add the 6 **Sentiment: …** count measures ("Sentiment: Very negative" … "Sentiment: Very positive", "Sentiment: Mixed"). Add **no** grouping (no dimension, no time grouping).
   - Expected: 6 bars starting at the **left** edge of the plot, one per measure, x-axis ticks showing the measure names, value labels on top of each bar (including "0" for empty ones), and hovering a zero-count column shows a tooltip like "Very positive — 0".
3. Repeat with the 6 **Emotion: …** count measures — same expectations.
4. Regression — chart **with** a dimension: bar chart with "Responses" grouped by **Sentiment**. Expected: unchanged rendering (per-category coloured bars, enum ordering, "Not enriched" gray bucket, value labels on top).
5. Regression — multi-measure **with** a dimension (e.g. two count measures grouped by Source Name): unchanged grouped bars with legend.
6. Single measure with no grouping (e.g. just "Responses") as a bar chart: now renders one left-aligned bar labelled with the measure name instead of a lone centered bar.

Verification run locally (scoped):

- `pnpm exec vitest run modules/ee/analysis` → 22 files, 284 tests passed (includes 5 new `pivotMeasuresToCategories` tests).
- `pnpm exec eslint` on all 5 touched files → clean.
- `pnpm typecheck` → no errors in touched files (remaining errors are pre-existing unbuilt-workspace-package resolution noise in a fresh worktree).
- `pnpm exec prettier --check` on touched files → clean.

### Required

- [ ] Ran `pnpm build`

### Appreciated

before:
Screenshot 2026-07-15 at 18 53 20

after:
image

### DIFF
--- a/apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
+++ b/apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
@@ -26,6 +26,10 @@ export interface CartesianChartProps {
   /** False for measure-only charts with no real category: hides the meaningless x-axis tick and
    * tooltip header (which would otherwise show the fallback measure value, e.g. a stray "1"). */
   hasCategoryAxis?: boolean;
+  /** Overrides whether the tooltip header is hidden (defaults to `!hasCategoryAxis`). Pivoted
+   * measure charts keep their category axis but hide the header, since each tooltip row already
+   * carries the measure label and a header would just repeat it. */
+  tooltipHideLabel?: boolean;
 }
 
 const TARGET_TICK_COUNT = 5;
@@ -112,6 +116,7 @@ export function CartesianChart({
   zeroBaseline = false,
   xAxisTickFormatter,
   hasCategoryAxis = true,
+  tooltipHideLabel,
 }: Readonly<CartesianChartProps>) {
   const yScale = computeYAxis(data, dataKeys, zeroBaseline);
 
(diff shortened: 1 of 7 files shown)


Dhruwang (approved):

PR #8547, opened 2026-07-15
fix(api): reject publishing an app survey with zero triggers

pandeymangg:
## What does this PR do?

Fixes an app survey being publishable (moved to a non-draft status like `inProgress`) with **zero triggers** through the API. Such a survey shows as "running" but can never be displayed — it has no action to trigger it. The survey editor forbids this, but the rule was enforced **client-side only**. Found during QA 5.2 §7 (ENG-1757).

Linear: ENG-1792

Enforces the invariant server-side: **an `app` survey in a non-draft status must have at least one trigger.**

> Note on placement: the ticket suggested enforcing this once in `updateSurveyInternal`. That alone is **insufficient** — the v3 Surveys API create/patch do their own `createSurvey` / `prisma.survey.update` and don't route through `updateSurveyInternal` (the exact repro is a v3 PATCH). So the check is applied at the three real write choke points via a shared predicate:

- **`isAppSurveyMissingTriggersToPublish(type, status, triggers)`** — shared helper in `apps/web/lib/survey/utils.ts`.
- **`createSurvey`** (all create paths incl. v3) → throws `InvalidInputError` → **400** (v3-create maps `InvalidInputError` to 400 via `mapV3SurveyCreateError`).
- **`updateSurveyInternal`** (v1/v2 management update + the app's `updateSurveyAction`) → `InvalidInputError` → **400** (gated by the existing `!skipValidation`, matching the other invariant checks).
- **`executeV3SurveyPatch`** (v3 PATCH) → `V3SurveyReferenceValidationError` → **422**. Effective triggers = the patch's `distribution.triggers` when present (top-level replacement), else the survey's current triggers.

Draft app surveys and link surveys are unaffected.

## How should this be tested?

Unit tests (133 pass; new: helper unit test, `createSurvey` reject test, v3-patch reject test):

```
pnpm --filter=@formbricks/web exec vitest run app/api/v3/surveys/patch.test.ts lib/survey/service.test.ts lib/survey/utils.test.ts app/api/v3/surveys/create.test.ts
```

Manual (v3 API, verified on localhost):

| step | before | after |
| --- | --- | --- |
| create app survey, no triggers (`status:"draft"`) | 201 | **201** |
| `PATCH /api/v3/surveys/{id}` `{"status":"inProgress"}` (no triggers) | 200 | **422** (`invalid_params: [{name:"triggers", reason:"An app survey must have at least one trigger…"}]`) |
| create app survey `status:"inProgress"` + no triggers | 201 | **400** |
| publish an app survey **with** ≥1 trigger | 200 | **200** (unchanged — guard only fires on empty triggers) |

### Required

- [ ] Ran `pnpm build`
- [x] My changes don't cause any responsiveness issues (API/service-only, no UI change)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (N/A — no UI change)

### DIFF
--- a/apps/web/app/api/v3/surveys/patch.test.ts
+++ b/apps/web/app/api/v3/surveys/patch.test.ts
@@ -425,6 +425,29 @@ describe("patchV3Survey", () => {
     );
   });
 
+  test("rejects setting a non-draft status on an app survey with no triggers", async () => {
+    await expect(
+      executeV3SurveyPatch({
+        currentSurvey: { ...currentSurvey, type: "app", triggers: [] } as TSurvey,
+        document: {
+          name: currentSurvey.name,
+          status: "inProgress",
+          metadata: currentSurvey.metadata,
+          defaultLanguage: "en-US",
+          languages: [{ code: "en-US", enabled: true }],
+          welcomeCard: currentSurvey.welcomeCard,
+          blocks: currentSurvey.blocks,
+          endings: currentSurvey.endings,
+          hiddenFields: currentSurvey.hiddenFields,
+          variables: currentSurvey.variables,
+        } as unknown as Parameters<typeof executeV3SurveyPatch>[0]["document"],
+        languageRequests: [{ code: "en-US", default: true, enabled: true }],
+        requestId: "req_1",
+      })
+    ).rejects.toBeInstanceOf(V3SurveyReferenceValidationError);
+    expect(prisma.survey.update).not.toHaveBeenCalled();
+  });
+
   test("reconciles due survey schedules without refetching when no schedule transition persisted", async () => {
     vi.mocked(isSurveySchedulingDue).mockReturnValue(true);
     vi.mocked(reconcileDueSurveySchedules).mockResolvedValue({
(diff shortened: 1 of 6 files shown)


xernobyl (approved):
Reviewed and smoke-tested. Logic is correct and the placement is right — I traced every survey write path (create, updateSurveyInternal for v1/v2 mgmt + editor + status dropdown, executeV3SurveyPatch) and confirmed the three choke points cover the surface; duplicate always creates as draft and response-pipeline auto-complete only acts on already-running surveys, so no gaps.

Verified locally at the PR head:
- 4 touched suites: 133/133 pass
- Full survey surface (lib/survey, app/api/v3/surveys, modules/survey/editor): 1030/1030 pass, no adjacent regressions
- Typecheck: 0 errors in the changed source files

Also confirmed `document.status`/`effectiveTriggers` can't be undefined in the v3 patch (base document is seeded from the current survey and only merges defined patch fields), so the guard can't false-positive on a status-less patch.

One non-blocking nit already acknowledged in the description: the same logical error returns 400 on create/update but 422 on v3 PATCH. It's consistent with how each path maps its other validation errors, so no change needed.

LGTM 🚀

PR #8551, opened 2026-07-15
fix(docker): validate environment before migrations [QA 5.2]

BhagyaAmarasinghe:
## Problem

Docker startup and the default Docker Compose migration service could run database migrations before the application environment schema was loaded. Invalid AI configuration could therefore mutate the database before the container failed.

## Changes

- bundle the existing web environment schema into a self-contained production validator
- run validation before Docker startup migrations, SAML setup, and server startup
- run the same validation before the default Compose migration service
- preserve `SKIP_STARTUP_MIGRATION=true` and the Helm migration flow
- add CI coverage for invalid providers, missing models, malformed credentials, invalid provider URLs, secret-safe errors, Compose migrations, and exactly-once startup work

## Validation

- focused environment tests: 38 passed
- ESLint, Prettier, shell syntax, workflow YAML parsing, and Compose rendering
- web type-check and full `@formbricks/web...` production build
- production Docker image build
- local fail-fast tests with an unreachable database
- local Docker Compose migration validation
- local end-to-end startup with fresh PostgreSQL and Valkey containers:
  - database migrations completed once
  - SAML setup ran once
  - Next.js started
  - `/health` returned 200
- verified `SKIP_STARTUP_MIGRATION=true` still skips startup migrations

[ENG-1806](https://linear.app/formbricks/issue/ENG-1806/52-blocker-validate-docker-environment-variables-before-running)


pandeymangg (approved):

xernobyl:
One scope note worth confirming, not a blocker: the new validate-before-migrate guard covers the Docker single-container startup (`next-start.sh`) and the Compose `formbricks-migrate` service, but the Helm migration job runs `node packages/database/dist/scripts/apply-migrations.js` directly without `validate-env.mjs` (`charts/formbricks/templates/migration-job.yaml` L52-54). So Helm deployments still migrate without the env check.

Looks intentional given the PR says it preserves the Helm flow — just flagging since ENG-1806 is framed as a migration-safety blocker and Helm is the prod path for a lot of self-hosters. Happy to leave it out of scope if that's the call. 🙂

PR #8566, opened 2026-07-16
refactor(surveys): WCAG-compliant keyboard and focus model for survey elements

itsjavi:
## What does this PR do?

Makes the survey player's keyboard interaction model and focus handling WCAG-compliant, following the WAI-ARIA APG radio-group guidance. Covers ENG-1779 findings 1, 2, 3, 6, 9 and the `label-content-name-mismatch` part of finding 10 (the screen-reader-semantics and visual-polish findings ship in the sibling PRs).

### Keyboard model — selection decoupled from focus

Native radios select on arrow-key focus moves; combined with auto-progress this submitted the card while a keyboard user was still browsing (options past the first arrow press were unreachable). A shared `useRovingRadioGroup` hook now gives every radio-semantics element one Tab stop, arrow keys that move focus **without** selecting (RTL-aware, wrapping, Home/End), and Space/Enter to select. Applied to:

- **NPS** — also rebuilt from 11 tabbable labels + a hand-rolled key handler into a native radio group.
- **Rating** (number / star / smiley) — CSAT & CES render through Rating.
- **Single-select list.**
- **Single picture selection** — was a Radix radio group (which also selects on arrow moves); now native radios with the roving hook, keeping the selected border + corner dot and the uniform focus ring.

Checkbox groups (multi-select list, multi picture) intentionally keep native per-checkbox Tab + Space — there is no APG arrow-key pattern for checkbox groups, and overriding platform semantics would hurt screen-reader users.

### Dropdown search focus trap

Radix's menu key handling ignores keydowns from the embedded search input, so keyboard users were trapped there in **both** select dropdown variants (only Esc escaped). ArrowDown/ArrowUp now move from the search into the options, ArrowUp on the first option returns to the search, and typing while on the options edits the query instead of triggering the menu typeahead.

### Focus management

- After Next / Back / auto-progress, the new card focuses its first interactive control (inputs are named by the question headline, so context is announced — this also stops VoiceOver re-announcing "Survey dialog" on every card).
- Ending cards without a button focus the card container so the thank-you is read.
- Failed validation focuses the first invalid control (previously it only scrolled).
- The embedded-widget guard is unchanged: initial render only autofocuses where it did before; card-to-card moves are user-initiated (safe under WCAG 3.2.x).

### Focus indicator

- Scale cells (NPS / rating) sit flush together and against the card's scroll edge, so the default outer ring was clipped (WCAG 2.4.11). They now paint a **single inset ring** and raise the focused cell; the negative outline-offset keeps it inset under forced-colors too.
- Dropdown options: the menu's `overflow-hidden` clipped the ring's sides — they now use the same inset ring.
- `tabindex="-1"` script-focus targets (e.g. the ending-card container) no longer draw a ring — they aren't operable controls.
- Fixed a **black ring flashing on every focus**: Tailwind v4's `transition-colors` animates `outline-color`, and the focus rule's transparent outline was transitioning from the default black. Ring-carrier elements now rest at `outline-color: transparent`, making the transition a no-op.

### Accessible names

The select dropdown triggers were named via `aria-label={headline}` — a raw HTML string for rich-text headlines, never containing the visible button text (axe `label-content-name-mismatch`, WCAG 2.5.3, the only violations that survived forced-colors). They are now named via `aria-labelledby` → rendered headline + visible value. Picture-selection groups gained matching `aria-labelledby` names (the old `htmlFor` pointed at an id no control carried).

### Consciously not changed (rationale in the ticket)

- Tab order stays inputs → Submit → Back — WCAG 2.4.3 wants a *meaningful* order, and primary-action-first is one.
- The "longer auto-progress debounce" idea is superseded: selection is now always an explicit act, so the existing 350 ms delay is fine.

### Tests

- New self-seeding Playwright spec `survey-keyboard.spec.ts` (4 tests): arrow-browse never selects or auto-progresses; Space/Enter select and auto-progress exactly once; the dropdown search is escapable into the options in both variants; focus lands on the new card after navigation and on the first invalid control after a failed submit; picture-selection roving.
- Unit tests for `useRovingRadioGroup` (roving Tab stop, arrow wrap, RTL inversion, Home/End, Enter-select, native Space passthrough, blur reset, ref unregistration).

Fixes ENG-1779

## How should this be tested?

- `pnpm --filter @formbricks/survey-ui test` — includes the `useRovingRadioGroup` unit suite.
- `npx playwright test survey-keyboard` — the keyboard-interaction e2e suite.
- `npx playwright test survey-accessibility` — the existing axe gate still passes; an axe scan of a dropdown-variant card now reports zero WCAG A/AA violations (previously `label-content-name-mismatch`).
- Manually, on any link survey: Tab into an NPS / rating / single-select / picture card → arrows move the ring **without** selecting → Space/Enter selects (and auto-progresses if enabled). Open a select dropdown with >3 options → ArrowDown from the search reaches the options, ArrowUp returns. Submit a required field empty → focus lands in the offending input.
- VoiceOver spot-check: card-to-card navigation reads the next question instead of re-announcing the survey dialog; the focused rating/NPS ring shows no dark flash.

> NOTE: multi-selects as lists (list of checkboxes) don't use the roving focus since they don't belong to a group like radios or selects. It is expected that exactly one item is focused on every TAB key stroke, and arrow buttons won't change that focus.

### Required

- [x] Ran `pnpm build`

### Appreciated

https://github.com/user-attachments/assets/362ce52b-647b-433a-b530-bec0fdcc65c7


jobenjada (approved):
works well, good job :)

PR #8570, opened 2026-07-16
fix(unify): disable active write controls for read-only members on Feedback Data

pandeymangg:
## What does this PR do?

Read-only (viewer) members could still see and interact with write-only controls on the **Feedback Data** (Unify) pages. The underlying server actions already reject the writes, but the UI looked active and either silently no-op'd or surfaced an error. This PR hides/disables that UI so viewers get a consistent read-only experience.

**Changes**

- **Feedback records table** — hide the per-row and select-all checkboxes (and therefore the bulk-delete toolbar) for read-only members. Selection only existed to drive bulk delete, which viewers can't perform. Clicking a row still opens the record in the read-only drawer. (`col`/`th`/`td` gated on `canWrite`, `colSpan` adjusted, `canWrite` threaded into the row.)
- **Edit source modal – CSV mapping** — the field-mapping dropdowns stayed active for viewers because they render `InputCombobox`, which is a `div[role="combobox"]` and therefore **not** disabled by the surrounding `<fieldset disabled>`. Added an optional `disabled` prop to `InputCombobox` (forces the menu closed, `pointer-events-none`, removes it from the tab order) and threaded it through `MappingUI → FormTargetField`, so the mapping dropdowns and the fixed-value edit button are inert in read-only mode.

The new `InputCombobox` `disabled` prop is additive and defaults to `false`, so its other usages are unaffected.

The rest of the Feedback Data surface (record drawer, source row menu, add-source button, survey suggestions, taxonomy edit controls) was already correctly gated and is unchanged.

## How should this be tested?

As a **read-only / viewer** member of a workspace that has a feedback directory, at least one **CSV** feedback source, and some feedback records:

- **Feedback Data → Feedback records**: the row checkboxes and the header select-all checkbox are no longer shown, so there is no way to reach the bulk-delete toolbar. Add / import controls remain hidden. Clicking a row opens the record drawer in read-only mode (all fields disabled, no Save/Delete).
- **Feedback Data → Sources**: click a CSV source to open the edit dialog. Every field-mapping dropdown is disabled and non-clickable; only **Close** is available.

As an **Owner / Manager / read-write** member, verify all of the above still work fully: row selection, bulk delete, editing CSV mappings, and saving.

### Required

- [x] Ran `pnpm build`
- [x] Checked for warnings, there are none (lint + typecheck clean on changed files)

### Appreciated

### DIFF
--- a/apps/web/modules/ee/unify-feedback/components/feedback-records-table.tsx
+++ b/apps/web/modules/ee/unify-feedback/components/feedback-records-table.tsx
@@ -85,6 +85,7 @@ interface FeedbackRecordRowProps {
   contactId?: string;
   locale: string;
   t: TFunction;
+  canWrite: boolean;
   isSelected: boolean;
   onSelectChange: (checked: boolean) => void;
   onClick: () => void;
@@ -422,7 +423,7 @@ export const FeedbackRecordsTable = ({
           <div className="overflow-x-auto">
             <table className="w-full min-w-[1040px] table-fixed">
               <colgroup>
-                <col className="w-10" />
+                {canWrite && <col className="w-10" />}
                 <col className="w-40" />
                 <col className="w-40" />
                 <col className="w-40" />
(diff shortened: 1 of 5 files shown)


Dhruwang (approved):

PR #8574, opened 2026-07-17
fix(sso): normalize IdP display name & accept common name punctuation (ENG-1743)

pandeymangg:
## What does this PR do?

**Linear:** ENG-1743 · **Sentry:** FORMBRICKS-131

An SSO identity-provider display name containing punctuation outside a narrow allowlist — a period (`J. Smith`), ampersand (`Smith & Co`), slash, etc. — failed `ZUserName` validation (`ValidationError: Invalid name format`). On the legacy NextAuth path this threw inside the user-provisioning transaction, aborting the SSO sign-in and surfacing as a 500-class error in Sentry. The email-derived fallback name was already sanitized before validation, but the **provider-supplied** name was passed through raw — that inconsistency is the defect.

### Where the fix lives (and why not `sso-handlers.ts`)

The Sentry events come from release `5.1.x`, which still runs the **NextAuth** path (`sso-handlers.ts` → `createUser` → `ZUserName`). The NextAuth → Better Auth migration (ENG-1054, #8384, first shipped in `5.2.0-rc`) removed that route, so `sso-handlers.ts` is now **dead code** on `main`. The live path is Better Auth's `ssoDatabaseHooks`, which writes the name via Better Auth's Prisma adapter — it no longer throws, but it persisted the **raw** IdP name, which would later break a profile save. So the fix is applied to the live hook, not the dead file.

### Changes

**Fix**
- `packages/types/user.ts`: `ZUserName` now also accepts `.` and `&`, so ordinary names like `J. Smith` / `Smith & Co` validate. The allowed-character set is declared once (`USER_NAME_ALLOWED_CHARS`) and drives **both** the validator and a new `normalizeUserName()` helper — so the sanitizer can never drift from the validator, and every normalized string is guaranteed to satisfy `ZUserName`.
- `better-auth-hooks.ts`: the SSO `user.create.before` hook now **normalizes** the IdP-supplied name (`normalizeUserName(user.name) || <email fallback>`) instead of storing it raw. External IdP names are untrusted input at the auth boundary — we normalize to fit the model rather than fail the login (the user can't fix what their IdP sends).
- `deriveNameFromEmail` now delegates to `normalizeUserName` (single source of truth for the name allowlist). Email-derived output is unchanged: `john.doe@…` → `john doe`.

**Cleanup**
- Removed the now-dead `sso-handlers.ts` + `sso-handlers.test.ts` (the NextAuth handler; zero runtime callers post-migration — its live equivalents `gateSsoProvisioning` / the Better Auth hooks keep their own dedicated tests).
- Fixed stale comments in `sso-provisioning.ts` that referenced the deleted test file.

No data migration needed — affected sign-ins never wrote a user row (the transaction aborted).

## How should this be tested?

**Automated (unit):**

```bash
cd apps/web && pnpm test modules/ee/sso/lib
```

Covers: IdP names with `.` / `&` / `/` / emoji → normalized to a valid form (no throw); a name that normalizes to empty → falls back to the email local-part; existing provisioning-gate behavior unchanged.

**Manual:**

1. Configure an SSO provider (OIDC / SAML / Azure / Google).
2. Sign in as a user whose IdP display name contains a period or ampersand (e.g. `J. Smith`).
3. Sign-in **succeeds** and the user is provisioned with a normalized, valid name; opening **Profile settings** and saving does **not** error.

_Verified locally: `pnpm test modules/ee/sso/lib` (132 pass) and `eslint` on the changed files (clean). Full `pnpm build` / `git pull origin main` not run in this session._

### Required

- [ ] Ran `pnpm build`
- [x] Checked for warnings (lint clean on changed files)
- [ ] First PR at Formbricks? (N/A)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (N/A — no UI changes)
- [ ] Updated the Formbricks Docs if changes were necessary (N/A)


xernobyl:
One cleanup miss: `apps/web/modules/ee/sso/lib/tests/__mock__/sso-handlers.mock.ts` is now orphaned — its only consumer was `tests/sso-handlers.test.ts`, which this PR deletes, and nothing else imports it (`git grep sso-handlers.mock` at head returns nothing). Since the stated goal here is removing the dead NextAuth handler + its test, the mock should go with them.

xernobyl (changes requested):
apps/web/modules/ee/sso/lib/tests/__mock__/sso-handlers.mock.ts is now orphaned — its only consumer was tests/sso-handlers.test.ts, which this PR deletes, and nothing else imports it

apps/web/modules/ee/sso/lib/better-auth-hooks.ts
```diff
@@ -124,7 +126,11 @@ export const ssoDatabaseHooks: NonNullable<BetterAuthOptions["databaseHooks"]> =
             // picture): transformInput drops undefined fields that have no schema default, so this
             // prevents a prisma.user.create validation error on SSO sign-up.
             image: undefined,
-            ...(user.name ? {} : { name: deriveNameFromEmail(user.email) }),
+            // Normalize the IdP-supplied display name to a form ZUserName accepts (ENG-1743): external
+            // provider names are untrusted and may carry punctuation ("J. Smith", "Smith & Co") that
+            // would otherwise persist raw and later break a profile save. Fall back to the email
+            // local-part when the IdP gave no name, or normalization leaves nothing (e.g. only emoji).
+            name: (user.name && normalizeUserName(user.name)) || deriveNameFromEmail(user.email),
```
xernobyl: **[P3] Name can still resolve to `""` for degenerate input.**

If `normalizeUserName(user.name)` and `deriveNameFromEmail(user.email)` both collapse to `""`, the persisted name violates `ZUserName.min(1)`. Verified:
- `name:"🎉🎉"` + `email:"___@example.com"` → `""`
- `name:"🎉🎉"` + `email:"😀@example.com"` → `""`

The write still succeeds (Better Auth doesn't enforce `ZUserName` on its Prisma adapter), but it reintroduces the exact ENG-1743 failure — the user's first Profile-settings save throws _"Name should be at least 1 character long"_. Narrow (needs a degenerate email local-part, e.g. a service/machine account) so low severity, but a constant final fallback closes it and makes the "every normalized string satisfies ZUserName" guarantee literal:

```suggestion
            name: (user.name && normalizeUserName(user.name)) || deriveNameFromEmail(user.email) || "User",
```

xernobyl (commented):
**[P3/nit] Stale reference to the now-deleted file** — `apps/web/modules/ee/sso/lib/better-auth-hooks.ts:180`

The `ssoLicenseGateBeforeHandler` doc comment still cites `sso-handlers.ts:492-523`, which this PR deletes — so it's now a dangling reference with meaningless line numbers. You already cleaned up the equivalent citations in `sso-provisioning.ts`; worth dropping the `(sso-handlers.ts:492-523)` part here too.

_(Left in the review body rather than inline because line 180 is outside the diff.)_

xernobyl (approved):
Both prior findings resolved in 4de2da7 — the empty-name path now falls back to a constant (`|| "User"`) with a dedicated test, and the dangling `sso-handlers.ts` citations are gone (including a second one in sso-provisioning.ts). Clean from my side. 👍

PR #8575, opened 2026-07-17
fix(auth): read the correct field from the OAuth consent response (MCP)

xernobyl:
## What does this PR do?

Fixes the OAuth consent page always showing the toast **"OAuth consent failed. Please try again."** even when consent succeeds and the MCP client connects (Formbricks-as-OAuth-provider, ENG-1055). Found during 5.2 QA.

**Root cause.** Better Auth's `oauth2.consent` doesn't return `{ redirect_uri }`. On accept it delegates to the authorize handler, which for a fetch/JSON request returns `{ redirect: true, url: "<callback?code=…>" }` (`handleRedirect` in `@better-auth/oauth-provider`). The plugin's OpenAPI schema documents `redirect_uri`, but the runtime field is `url`. The component read `response.data.redirect_uri`, which is always `undefined`, so it fired the error toast and skipped its own `window.location` redirect. The connection still completed only because the consent row is persisted server-side and the client's `/authorize` retry then finds consent granted (the "works after 2–3s" symptom).

**Fix.** Extract `resolveConsentRedirectUrl()` (reads `url`, falls back to `redirect_uri` for the documented/legacy shape) and use it in `OAuthConsentActions`. This also fixes the **deny** path, which returns the same `{ redirect, url }` shape with `error=access_denied` and previously toasted instead of redirecting the error back to the client.

**Security.** The URL feeds `window.location.href`, so the helper rejects `javascript:`/`data:` schemes (XSS guard for the location sink). It intentionally does **not** restrict to http/https — native MCP clients use loopback (`http://127.0.0.1:PORT`) and custom app schemes, and the oauth-provider already validated the target `redirect_uri` against the registered client at `/authorize` (BA `SafeUrlSchema`). No server/schema/env changes.

## How should this be tested?

Automated (from `apps/web`):
```
pnpm test -- "app/(app)/account/authorize/lib/consent-redirect.test.ts"
```
Covers the real accept shape (`{ redirect, url }`), the deny/error `url`, loopback + custom-scheme targets, the `redirect_uri` fallback, `url` precedence, null/empty/non-object inputs, and `javascript:`/`data:` rejection.

Manual (the real repro):
1. Connect an MCP client (e.g. Claude Code `/mcp`) through the Formbricks OAuth flow → reach `/account/authorize` → click **Allow** → lands back at the client authenticated, **no error toast**, redirect is immediate.
2. Retry and click **Deny** → redirected back with `error=access_denied`, no stuck page/toast.

### Required

- [ ] Ran `pnpm build`

### DIFF
--- a/apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
+++ b/apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
@@ -7,19 +7,26 @@
  * the runtime field is `url`. Deny returns the same shape with an `error=access_denied` URL. We read
  * `url` first, and keep `redirect_uri` as a defensive fallback for the documented/legacy shape.
  *
- * Security: the returned string is fed to `window.location.href`, so we reject `javascript:`/`data:`
- * URLs (never valid OAuth redirect targets; the XSS vector for a location sink). We do NOT restrict to
- * http/https — native MCP clients use loopback (`http://127.0.0.1:PORT`) and custom app schemes, and the
- * oauth-provider already validated the target `redirect_uri` against the registered client at
- * `/authorize`. The value here always originates from that pre-validated callback.
+ * Security: the returned string is fed to `window.location.href`, so we reject XSS-capable schemes.
+ * We parse with the `URL` constructor and deny the dangerous protocols rather than allowlisting — native
+ * MCP clients use arbitrary custom/loopback schemes (`cursor://`, `vscode://`, `http://127.0.0.1:PORT`),
+ * so the safe set can't be enumerated, and the oauth-provider already validated the target `redirect_uri`
+ * against the registered client at `/authorize`. Parsing (vs a regex) also normalizes obfuscated schemes
+ * like `java\tscript:` and rejects malformed / protocol-relative values.
  */
-const DANGEROUS_SCHEME = /^\s*(?:javascript|data|vbscript):/i;
+const DANGEROUS_PROTOCOLS = new Set(["javascript:", "data:", "vbscript:", "file:", "blob:"]);
 
 const asRedirectString = (value: unknown): string | null => {
   if (typeof value !== "string") return null;
   const trimmed = value.trim();
   if (trimmed.length === 0) return null;
-  if (DANGEROUS_SCHEME.test(trimmed)) return null;
+  let protocol: string;
+  try {
+    protocol = new URL(trimmed).protocol;
+  } catch {
+    return null; // not a well-formed absolute URL
+  }
+  if (DANGEROUS_PROTOCOLS.has(protocol.toLowerCase())) return null;
   return trimmed;
 };
 
--- a/apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
+++ b/apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
@@ -1,0 +1,30 @@
+/**
+ * Extract the post-consent redirect URL from a Better Auth `oauth2.consent` response.
+ *
+ * On accept, `@better-auth/oauth-provider` delegates to its authorize handler, which for a fetch/JSON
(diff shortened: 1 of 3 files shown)


apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+ * the runtime field is `url`. Deny returns the same shape with an `error=access_denied` URL. We read
+ * `url` first, and keep `redirect_uri` as a defensive fallback for the documented/legacy shape.
+ *
+ * Security: the returned string is fed to `window.location.href`, so we reject `javascript:`/`data:`
+ * URLs (never valid OAuth redirect targets; the XSS vector for a location sink). We do NOT restrict to
+ * http/https — native MCP clients use loopback (`http://127.0.0.1:PORT`) and custom app schemes, and the
+ * oauth-provider already validated the target `redirect_uri` against the registered client at
+ * `/authorize`. The value here always originates from that pre-validated callback.
+ */
+const DANGEROUS_SCHEME = /^\s*(?:javascript|data|vbscript):/i;
```
itsjavi: I think we should add `file:` to the blocklist as well, it shouldn't be a valid OAuth callback. However, I'd go further and use an **allowlist of schemes** instead:

```typescript
// Only allow http, https, and custom schemes (used by native MCP clients like Cursor, VSCode)
const VALID_SCHEME = /^(?:https?:|[a-z][a-z0-9+\-.]*:)/i;

const asRedirectString = (value: unknown): string | null => {
  if (typeof value !== "string") return null;
  const trimmed = value.trim();
  if (trimmed.length === 0) return null;
  if (!VALID_SCHEME.test(trimmed)) return null;
  return trimmed;
};
```

This blocks `javascript:`, `data:`, `file:`, `vbscript:`, etc. in one go and is more defensible against future XSS vectors. Allowlists are harder to bypass than blocklists.

Make sure the test still covers MCP client schemes like `cursor://` and loopback URLs, they should all match.

apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const asRedirectString = (value: unknown): string | null => {
```
itsjavi, line 19: should we have a more general util for safe redirect urls?

itsjavi (changes requested):
thanks, I left some remarks to harden security

apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
xernobyl: Good call — pushed 6f2596df8. Went a step further than a scheme regex: it now parses with the `URL` constructor and denies dangerous protocols (`javascript:`, `data:`, `vbscript:`, `file:`, `blob:`) — `file:` included per your suggestion.

Two reasons for parse-over-regex:
1. An allowlist like `^(?:https?:|[a-z][a-z0-9+\-.]*:)` still matches `javascript:` (it is a valid-looking scheme), so it would not actually block the thing we care about.
2. URL parsing also normalizes obfuscated schemes like `java\tscript:` and rejects malformed / protocol-relative values.

`cursor://` and loopback still pass — covered in the tests.

apps/web/app/(app)/account/authorize/lib/consent-redirect.ts
xernobyl: Looked for one — the closest is `getValidatedCallbackUrl` (`lib/utils/url.ts`), but that validates *internal* same-origin callback URLs against `WEBAPP_URL`, so it would reject the external / custom-scheme OAuth callbacks we need here. Different semantics, and this is the only caller today, so I kept it local. Happy to promote the scheme check into `lib/utils/url.ts` as a shared helper if/when a second consumer appears.

itsjavi (approved):

PR #8577, opened 2026-07-17
feat(mcp): add a list_workspaces tool (ENG-1776)

xernobyl:
## What does this PR do?

Fixes ENG-1776 — adds a `list_workspaces` tool to the MCP server.

Every MCP survey tool (`list_surveys`, `create_survey`, …) requires a `workspaceId`, but a client had no way to discover which workspaces the connected user can access. This adds a `list_workspaces` tool that returns the workspaces the authenticated principal is a member of, so a client can bootstrap the `workspaceId` it then passes to the survey tools.

**How it works:**
- New v3 lib operation `listV3Workspaces`:
  - **Session user (OAuth):** aggregates the workspaces across the orgs they belong to — org owners/managers see all of an org's workspaces; `member`-role users see only team-scoped ones (enforced by the existing `getUserWorkspaces`). Scoped strictly to the user's own orgs via `getOrganizationsByUserId`.
  - **API key:** returns only the workspaces in its `workspacePermissions`.
  - Returns a minimal `{ id, name, organizationId }` DTO in the standard v3 list shape (never the raw `Workspace` entity). No `workspaceId` input → no IDOR surface.
- New `list_workspaces` MCP tool (read-only), gated on the existing **`surveys:read`** scope (no new OAuth scope, so oauth-provider/PRM/DCR/consent/docs are untouched).
- Extracted the shared `guardMcpScopes` helper so the surveys and workspaces tools reuse one scope gate.

Scope kept deliberately small: **no** public `GET /api/v3/workspaces` REST route/OpenAPI, **no** new scope, **no** schema change.

## How should this be tested?

Automated (from `apps/web`):
```
pnpm test -- app/api/v3/workspaces/lib/operations.test.ts modules/mcp/tools/workspaces.test.ts modules/mcp/tools/surveys.test.ts
```
Covers: session-user aggregation + dedupe across orgs returning the minimal DTO; API-key path returning only its permitted workspaces; unauthenticated → 401; empty (no orgs); and the tool's `surveys:read` scope gate (insufficient-scope error skips the operation). The surveys tool test is included to confirm the shared-guard refactor is behavior-preserving.

Manual (MCP): connect a client (Claude Code `/mcp`) → call `list_workspaces` → returns the caller's workspaces; a multi-org user sees all their workspaces, a `member`-role user sees only team-scoped ones.

### Required

- [ ] Ran `pnpm build`

### DIFF
--- a/apps/web/app/api/v3/workspaces/lib/operations.ts
+++ b/apps/web/app/api/v3/workspaces/lib/operations.ts
@@ -88,6 +88,10 @@ export async function listV3Workspaces({
         .map(serializeV3WorkspaceListItem);
     }
 
+    // Stable, deterministic order (the underlying queries have no ORDER BY) so the tool output — and
+    // which items survive the cap — don't vary between calls. Name first, id as a tiebreaker.
+    items.sort((a, b) => a.name.localeCompare(b.name) || a.id.localeCompare(b.id));
+
     const capped = items.slice(0, MAX_WORKSPACES);
     return successListResponse(
       capped,
--- a/apps/web/app/api/v3/workspaces/lib/operations.ts
+++ b/apps/web/app/api/v3/workspaces/lib/operations.ts
@@ -1,0 +1,104 @@
+import "server-only";
+import { logger } from "@formbricks/logger";
+import type { TAuthenticationApiKey } from "@formbricks/types/auth";
+import { DatabaseError } from "@formbricks/types/errors";
+import { problemInternalError, problemUnauthorized, successListResponse } from "@/app/api/v3/lib/response";
+import type { TV3Authentication } from "@/app/api/v3/lib/types";
+import { getOrganizationsByUserId } from "@/lib/organization/service";
+import { getUserWorkspaces, getWorkspace } from "@/lib/workspace/service";
+
+type TListV3WorkspacesParams = {
+  authentication: TV3Authentication;
+  requestId: string;
+  instance: string;
+};
+
+/** Minimal DTO — never return the raw Workspace entity (no config/env/styling internals). */
+type TV3WorkspaceListItem = {
+  id: string;
+  name: string;
+  organizationId: string;
+};
+
+// Defensive cap: a user's accessible workspaces are naturally small, but never stream an unbounded list.
+const MAX_WORKSPACES = 500;
+
+const serializeV3WorkspaceListItem = (workspace: {
+  id: string;
+  name: string;
+  organizationId: string;
+}): TV3WorkspaceListItem => ({
+  id: workspace.id,
+  name: workspace.name,
+  organizationId: workspace.organizationId,
+});
+
+/**
+ * List the workspaces the authenticated principal can access.
+ *
+ * - Session user: every workspace across the orgs they're a member of (org owners/managers see all of
+ *   an org's workspaces; `member`-role users see only team-scoped ones — enforced by `getUserWorkspaces`).
+ * - API key: only the workspaces in its `workspacePermissions`.
+ *
+ * There is no `workspaceId` input, so there is no IDOR surface — the result is always derived from the
+ * caller's own live membership / key grants, never from client-supplied ids.
+ */
+export async function listV3Workspaces({
+  authentication,
+  requestId,
+  instance,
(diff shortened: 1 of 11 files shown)


apps/web/app/api/v3/workspaces/lib/operations.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Session user's accessible workspaces: every workspace across the orgs they're a member of. Scoping is
+ * enforced by the reused services — `getOrganizationsByUserId` limits to the user's own orgs, and
+ * `getUserWorkspaces` returns all of an org's workspaces for owners/managers but only team-scoped ones
+ * for `member`-role users.
+ */
+async function fetchSessionWorkspaces(userId: string): Promise<TV3WorkspaceListItem[]> {
+  const organizations = await getOrganizationsByUserId(userId);
+  const workspacesPerOrg = await Promise.all(
+    organizations.map((organization) => getUserWorkspaces(userId, organization.id))
```
BhagyaAmarasinghe, line 41: Billing-only users receive workspace metadata they cannot access. [fetchSessionWorkspaces](https://github.com/formbricks/formbricks/blob/a05640cceec61a014877ec86144389b939402ddf/apps/web/app/api/v3/workspaces/lib/operations.ts#L41-L46) calls getUserWorkspaces, which only restricts the member role. A billing member therefore receives every workspace ID/name in the organization, despite billing users being excluded from product data and failing the survey tools’ authorization. Explicitly return no workspaces for billing roles or use a query that mirrors v3 workspace authorization.

apps/web/app/api/v3/workspaces/lib/operations.ts
```diff
@@ -0,0 +1,110 @@
...
+      return problemUnauthorized(requestId, "Not authenticated", instance);
+    }
+
+    // Dedupe by id (defensive) + a stable, deterministic order — the underlying queries have no ORDER BY,
+    // so without this the output and which items survive the cap could vary between calls.
+    const deduped = Array.from(new Map(items.map((item) => [item.id, item])).values()).sort(
+      (a, b) => a.name.localeCompare(b.name) || a.id.localeCompare(b.id)
+    );
+
+    const capped = deduped.slice(0, MAX_WORKSPACES);
```
BhagyaAmarasinghe: Workspaces after item 500 are permanently undiscoverable. [The result is sliced to 500](https://github.com/formbricks/formbricks/blob/a05640cceec61a014877ec86144389b939402ddf/apps/web/app/api/v3/workspaces/lib/operations.ts#L98-L102), but returns nextCursor: null and totalCount: 500. Since the tool accepts no pagination input, a multi-org or unlimited-plan user cannot retrieve the remaining workspaces. The cap also occurs after fetching everything, so it does not bound database work. Add pagination or remove the truncation.

BhagyaAmarasinghe (changes requested):

apps/web/app/api/v3/workspaces/lib/operations.ts
xernobyl: Good catch — this was a pre-existing, app-wide gap that the new tool would have newly exposed via the API. `getUserWorkspaces` (and its sibling `getUserWorkspacesByOrganizationIds`) only team-scoped the `member` role, so `billing` fell through to the unscoped "all workspaces" branch.

Fixed at the root in 7aaf4793c: both functions now scope **every** non-owner/manager role (member, billing, …) to their team-accessible workspaces, mirroring the per-workspace v3 authorization (`requireV3WorkspaceAccess`: org owner/manager OR workspace-team membership). Added a billing-role regression test. Since the UI callers mock `getUserWorkspaces`, the only test touched is `service.test.ts`.

apps/web/app/api/v3/workspaces/lib/operations.ts
xernobyl: Agreed — removed the cap in 7aaf4793c. The truncation only hid workspaces (no cursor input to page past it) without bounding DB work, since it ran after fetching everything. A principal’s accessible workspaces are bounded by their memberships (small), so we now return them all with an accurate `totalCount` and `nextCursor: null`.

BhagyaAmarasinghe (approved):
LGTM

PR #8588, opened 2026-07-20
feat(segments): interaction-based segment filter (ENG-1275)

Dhruwang:
## What & why

Customers want to target (or exclude) contacts based on how they have interacted with surveys — e.g. "people who have **not seen** the NPS survey in the last month" or "people who **completed** any survey in the last 2 weeks" — so they can track how metrics like NPS evolve over time and avoid re-surveying the same people.

This PR adds a **Survey interaction** segment filter.

Closes ENG-1275.

## Filter shape

A single filter row reads as:

> `Users who` **[interaction]** **[survey scope]** `within` **[N]** **[period]**

- **Interaction** (operator): `have completed`, `have not completed`, `have seen`, `have not seen`, `have started responding to`
- **Survey scope**: `any survey` (default) or `specific surveys` (searchable multi-select)
- **Window**: integer `1–999` (default `1`) + `day(s)` / `week(s)` / `month(s)` (default `month`)

## Approach

The ticket sketch labelled this a "user attribute", with a note to verify. It is **not** modelled as an attribute — the data already lives on the `Contact.displays` (survey shown) and `Contact.responses` (`finished` = completed) relations. Forcing it into the attribute system would require write-time syncing and go stale.

Instead it is a new `surveyInteraction` filter **root type**, following the existing non-attribute roots (`device`, `segment`). This means:

- **No DB migration** — required indexes already exist (`Display@@index([contactId, createdAt])`, `Response@@index([contactId, createdAt])`).
- Two evaluation paths share **one operator spec** (`.../segments/lib/filter/survey-interaction.ts`) so they can never drift:
  - **Dashboard / segment preview** — a workspace-scoped Prisma relation query via the existing `segmentFilterToPrismaQuery` pipeline.
  - **Contact-sync hot path** (`getPersonSegmentIds`) — evaluated **in memory** against the contact's already-loaded `displays`/`responses`, so an interaction-only segment costs **zero extra queries** at survey-display time (see Performance).

Screenshot 2026-07-21 at 12 27 52 PM

### Semantics
- `have seen` → a `Display` row exists within the window; `have started responding to` → a `Response` row exists; `have completed` → a `Response` row with `finished: true`.
- Negative operators (`have not seen`, `have not completed`) wrap the positive clause in `NOT`, i.e. "no matching interaction **within the window**" (interactions older than the window are intentionally ignored — this is what enables re-targeting).
- `any survey` omits the `surveyId` constraint; `specific surveys` constrains to `surveyId in [...]`.

## Performance

Segment membership is evaluated on **every contact sync**, so at scale (e.g. 100k DAU × several syncs/day) the cost is `syncs × segments`. To keep interaction filters from adding to that multiplier, they are evaluated in memory on the hot path.

**How it works.** The contact-sync endpoint already loads the contact's `displays`/`responses` in one query (`getContactWithFullData`). That select was widened by two columns (`createdAt`, `finished`) so `getPersonSegmentIds` can evaluate interaction-only segments directly against the in-memory arrays and skip the per-segment DB check. Segments that mix in other filter types fall back to the existing Prisma path unchanged. Parity between the two paths is guaranteed by the shared operator spec.

**Impact** (illustrative: 500k syncs/day, 10 segments):

| | membership `findFirst` / sync | / day |
|---|---|---|
| Per-segment DB check (before) | 10 | 5,000,000 |
| Interaction-only segments in memory | interaction-only → 0 | up to 5,000,000 fewer |

Each interaction-only segment removes `syncs × 1` queries/day (≈500k here) and shortens the sync transaction (less connection-pool pressure). Cost is unchanged for non-interaction segments.

A further optimization — pruning membership checks to only segments attached to live app surveys — is deliberately deferred to **ENG-1932** (no correctness impact; introduces a cache-invalidation edge case handled there).

## Changes

**Types** (`packages/types/segment.ts`)
- `SURVEY_INTERACTION_OPERATORS`, `SURVEY_INTERACTION_TIME_UNITS`, `ZSegmentSurveyInteractionFilter` (+ structured value with `min(1).max(999).int()` and a "specific requires ≥1 survey" refinement), added to the filter/value unions and `ALL_OPERATORS`.

**Evaluation**
- Shared spec (`.../segments/lib/filter/survey-interaction.ts`): operator semantics table + window helper + the in-memory evaluator (`evaluateSurveyInteractionFilterInMemory`, `tryEvaluateSurveyInteractionSegmentInMemory`) and the connector combiner (`combineFilterResults`, now shared with `evaluateSegment`).
- `prisma-query.ts`: `buildSurveyInteractionFilterWhereClause` refactored to consume the shared spec.
- Hot path (`.../user/lib/segments.ts`, `update-user.ts`): interaction-only segments evaluated in memory; contact select widened by `createdAt` + `finished`.

**Editor UI** (`.../segments/components/`)
- New `SurveyInteractionFilter` row (reuses the shared `Select`, `Input`, `MultiSelect`, connector and context-menu building blocks — no new primitives).
- Survey picker shows the **name as the primary label** with the id as secondary text (search matches both).
- The survey(s) the segment itself gates are **excluded** from the picker (in the survey editor that's the survey being edited — self-referential targeting made no sense).
- Leaf filter components now declare precise props (`TBaseFilterProps` + per-leaf extras) instead of the shared kitchen-sink type — clears the unused-prop code smells.
- `add-filter-modal`: "Survey interaction" entry in the All tab (searchable), created pre-filled with a valid default.
- Accessibility: visually-hidden labels via `aria-label` per [WCAG ARIA14](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA14) — `survey interaction`, `surveys`, `number`, `period`.

**Server / tenancy**
- `getSurveyRefsForWorkspace` + `getSurveysForSegmentFilterAction` power the picker (self-fetched, no prop-drilling).
- Write-time guard `assertSurveyInteractionSurveyIds` in create/update segment actions rejects survey ids outside the workspace.
- v3 public surveys targeting API (`app/api/v3/surveys/targeting.ts`) validates referenced survey ids (unknown/foreign → `422 invalid_reference`).

**i18n**: en-US keys added; all locales regenerated + validated.

## How to test

Prereq: enterprise **Contacts** entitlement enabled (advanced targeting). Identify a few contacts via the SDK, then create displays/responses for them.

### Happy path (UI)
1. Workspace → **Segments** → **Create segment**.
2. **Add filter** → **Survey interaction**. Row appears pre-filled (`have seen`, `any survey`, `within 1 month`).
3. Try each interaction operator; switch scope to **specific surveys** and pick surveys via the search box.
4. Save. Reopen the segment — configuration persists.

### Data correctness
Seed one contact with: a `Display` for survey A, an unfinished `Response` for survey B, a finished `Response` for survey C (all recent). Then verify each operator returns the expected membership (segment preview / SDK user sync):
- `have seen` + specific [A] → matches; `have not seen` + [A] → excludes.
- `have started responding to` + [B] → matches (unfinished still counts).
- `have completed` + [C] → matches; + [B] → does **not** match (B is unfinished).
- `have not completed` + [C] → excludes.

### Edge cases to check
- **Window boundary**: set an interaction to just outside the window (e.g. a display 40 days ago with `within 1 month`) → `have seen` should **not** match; `have not seen` **should**. Move it inside the window → flips.
- **Period units + pluralization**: `1 month` shows singular; `2 months` plural. Units are English-only per spec (flagged below).
- **Number bounds**: `0` and values `>999` are rejected/clamped; non-integers rejected.
- **Empty specific surveys**: switching to `specific surveys` with nothing selected shows the inline error **"Select at least one survey"** and blocks save; error clears on first selection.
- **`any survey` vs specific**: `any survey` matches interactions with **any** survey in the workspace.
- **Cross-workspace / deleted survey**: a `specific` filter referencing a survey outside the workspace is rejected on save (and on the v3 API with `422 invalid_reference`). A survey deleted after the segment is saved simply matches nothing at runtime (no error).
- **Nesting & connectors**: place the filter inside AND/OR groups and nested groups — evaluation composes correctly.
- **View-only**: in read-only contexts every control is disabled.
- **Accessibility**: each control is reachable/announced via its `aria-label`.

### Automated
- Unit: `pnpm test` — `segment-schema`, `prisma-query`, `helper`, `utils`, and v3 `targeting` suites extended (all operators × scope, window math, negation, refinements, workspace-id validation), plus a new `survey-interaction` spec covering the in-memory evaluator, connector combine, and DB-fallback classification, and hot-path tests asserting interaction-only segments issue no DB query. All passing.
- E2E: `apps/web/playwright/segments.spec.ts` — create-with-interaction-filter flow + empty-survey validation + add-filter search (runs in CI where the Contacts entitlement is present).
- `pnpm lint`, `tsc`, `pnpm i18n:validate` — clean.

## Notes for reviewers (decisions to confirm)
1. **Pre-filled vs Figma "Select…" empty state** — the schema requires a valid operator, so the filter is created pre-filled with a valid default (matching every other filter type) rather than an empty intermediate state that autosave/validation would reject.
2. **Survey picker prioritizes name over id** — the option/chip label is the survey **name**, with the id as secondary text (search still matches both). This supersedes the ticket's "display survey IDs in the field" wording, which was confusing when names are what users recognize.
3. **Pluralization is standard English** (singular only for exactly `1`), not the ticket's literal "1, 21, 31 → singular". English-only per the ticket.

## Out of scope
- No JS SDK / `@formbricks/surveys` changes (evaluation is fully server-side).
- No cleanup job for survey ids left dangling by a later survey deletion (runtime is already safe).


apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/segments.ts
```diff
@@ -66,6 +79,14 @@ export const getPersonSegmentIds = async (
...
       }
 
+      const inMemoryResult = tryEvaluateSurveyInteractionSegmentInMemory(filters, interactionData, now);
+      if (inMemoryResult !== null) {
+        if (inMemoryResult) {
+          inMemoryMatchIds.push(segment.id);
+        }
+        continue;
+      }
+
```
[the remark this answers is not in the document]
Comment (Dhruwang) [reply apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/segments.ts]: Fixed in dff979bdb. Wrapped the in-memory evaluation in try/catch: on an unexpected throw it now logs a warn and leaves the result `null`, so the segment falls through to the Prisma path instead of bubbling to the outer catch. This mirrors the DB path's per-segment isolation — a single bad segment no longer drops the whole membership set (including already-computed `alwaysMatchIds`).

apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/update-user.ts
```diff
@@ -31,9 +31,12 @@ const getContactWithFullData = async (workspaceId: string, userId: string) => {
...
         },
       },
-      // Include user state data in the same query
+      // Include user state data in the same query. `createdAt` and `finished` are also consumed by the
+      // in-memory survey-interaction segment evaluator (see getPersonSegmentIds), which relies on this
+      // load being complete — do NOT add a `take` limit here without also bounding it by the widest
+      // interaction window, or the in-memory and Prisma evaluators would diverge.
       responses: {
-        select: { surveyId: true },
+        select: { surveyId: true, createdAt: true, finished: true },
```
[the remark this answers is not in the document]
Comment (Dhruwang) [reply apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/update-user.ts]: Fixed in dff979bdb. Extracted a shared `contactStateSelect` constant (`satisfies Prisma.ContactSelect`) consumed by both `getContactWithFullData` and `createContact`, so the responses/displays shape the in-memory evaluator depends on lives in one place and can no longer drift.

apps/web/modules/ee/contacts/segments/components/segment-filter.tsx
```diff
@@ -57,31 +64,39 @@ import {
...
   segment: TSegment;
-  segments: TSegment[];
-  contactAttributeKeys: TContactAttributeKey[];
   setSegment: (segment: TSegment) => void;
-  handleAddFilterBelow: (resourceId: string, filter: TBaseFilter) => void;
   onCreateGroup: (filterId: string) => void;
   onDeleteFilter: (filterId: string) => void;
   onMoveFilter: (filterId: string, direction: "up" | "down") => void;
   viewOnly?: boolean;
 }
```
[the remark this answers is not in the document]
Dhruwang: Fixed in dff979bdb. Marked the filter component props as `Readonly<>` — the `TBaseFilterProps`-based leaves (Attribute/Person/Segment/Device/SurveyInteraction) and the dispatcher `SegmentFilter`.

apps/web/modules/ee/contacts/segments/components/segment-filter.tsx
```diff
@@ -850,6 +867,220 @@ function DeviceFilter({
...
+    let active = true;
+    getSurveysForSegmentFilterAction({ workspaceId: segment.workspaceId }).then((result) => {
+      if (active && result?.data) {
+        setSurveys(result.data);
+      }
+    });
+    return () => {
+      active = false;
+    };
+  }, [segment.workspaceId]);
```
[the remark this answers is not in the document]
Dhruwang: Leaving as-is (intentional tradeoff). The self-fetch avoids prop-drilling the survey list through the four SegmentEditor render sites, and `reactCache` dedups the actual DB read. Duplicate network calls only occur when one segment holds multiple survey-interaction filters, which is rare; if that becomes common we can lift the fetch to the dispatcher/context. Not worth the added coupling now.

jobenjada (dismissed):
Thanks Dhru, love the quick turnaround and UI looks & feels great from the get go 💪

I was a bit confused about how this feature works and how it interacts with the Recontact & Cooldown options, so I wrote Docs for it and updated / clarified the docs for Recontact Options & Cooldown Period. It got a bit out of hand, so its a different PR.

----

E2E tests would be great but kinda hard to do, AI suggests:

```
My take: a table-driven integration test over the resolution seam (getPersonSegmentIds + filterSurveys) is the high-value, low-flake option — a matrix of displayOption × recontactDays × workspace cooldown × segment membership × interaction operator/window asserting shown/not-shown. A real Playwright browser test that boots the JS widget and identifies a user would be the gold standard but is expensive and flaky, so I'd reserve it for one happy-path smoke test at most. This is worth raising with the engineer — the interplay is exactly the kind of thing that regresses silently.
```

What do you think?

----

The 30-mins TTL for segment sync becomes a bit of an issue with this new setting. What do you think about refetching the updated user state on response complete? Could avoid drift between the client state and the server state.

Any concerns?

-----

I resolved some naming inconsistencies that confused AI and pushed them back up to this PR, pls have a look.

----

I didn't test this feature extensively, only for the day setting and one of the options -> please make sure that all options work for all time periods if you havent done that already ✅

Dhruwang:
Thanks Johannes! 🙏

**Tests** — went with your table-driven idea: added a matrix test over `filterSurveys` covering `displayOption × recontactDays × cooldown × lastDisplayAt × segment membership × anonymous` (14 cases). That's the low-flake layer where regressions hide. Skipped full Playwright for now; can add one happy-path smoke later.

**TTL drift** — fixed it: on **display *and* response**, identified users now refetch → `filterSurveys` re-runs immediately. Both matter (display flips `have seen`, response flips `completed`). Plain refetch, fires after the event is persisted. Longer term we could eval pure-interaction segments client-side to drop the round-trip — follow-up.

**Naming cleanup** — reviewed, 👍

**All options × periods** — covered by unit + parity tests (in-memory vs Prisma identical across every operator/period); I'll still do a manual pass to be safe.

jobenjada:
Thanks @Dhruwang works for me. However, please consider server /db load on refetching user state on each survey display. Might be fine, but pls get a second opinion :)

From UX point its ready to go from my side 🚀

Here are the docs changes: https://github.com/formbricks/formbricks/pull/8611

apps/web/modules/ee/contacts/segments/components/survey-interaction-filter.tsx
```diff
@@ -0,0 +1,247 @@
...
+  // which TypeScript does not treat as a plain (spreadable) object without this annotation.
+  const value: TSegmentSurveyInteractionFilterValue = resource.value;
+  const [surveys, setSurveys] = useState<{ id: string; name: string; status: string }[]>([]);
+
+  // Load the workspace's surveys once so the "specific surveys" picker is ready the moment the user
+  // switches scope. Self-fetching here avoids prop-drilling surveys through the four SegmentEditor
+  // render sites.
+  useEffect(() => {
+    let active = true;
+    getSurveysForSegmentFilterAction({ workspaceId: segment.workspaceId }).then((result) => {
```
pandeymangg: the action ideally should not throw any errors but it would be better to be safe here and a .catch block as well 🙏 
or maybe even use the async await pattern with try...catch

apps/web/app/api/v3/surveys/targeting.ts
```diff
@@ -274,6 +288,22 @@ export async function assertV3SurveyTargetingFilterReferences(
     }
   }
 
+  const surveyRefs = references.filter((ref) => ref.kind === "survey");
+  if (surveyRefs.length > 0) {
+    // Scope to the workspace's own surveys so a survey-interaction filter cannot reference another
+    // workspace's survey.
+    const surveys = await getSurveyRefsForWorkspace(workspaceId);
```
pandeymangg: this does fetch the first 1k surveys which were updated recently, but in a very large workspace with lots of surveys, this does leave some valid surveys if they are outside the 1k bracket, its prolly a very edge case but its technically possible

pandeymangg (changes requested):
works well! Please take a look at the comments! 🙏 
Apart from this, the issue with the user client api getting called too many times is real and can hurt the BE performance improvements that we did with this api, I think we should atleast filter out the workspaces that have a segment with a survey interaction filter and only call the user api again for those, we can gate this using the workspace api (/environment api call from the client) that can calculate this 🙏

apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts
```diff
@@ -206,13 +206,37 @@ export const getWorkspaceStateData = async (workspaceId: string): Promise<Worksp
...
+      const workspaceSegments = await prisma.segment.findMany({
+        where: { workspaceId },
+        select: { id: true, filters: true },
+      });
+      segmentFiltersById = new Map(
+        workspaceSegments.map((segment) => [segment.id, segment.filters as unknown as TBaseFilters])
+      );
+    }
+
+    const { refreshBySurveyId, hasAny: hasSurveyInteractionSegments } = buildSurveyInteractionRefreshMap(
```
Comment (pandeymangg) [inline apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts]: I think we don't need `hasSurveyInteractionSegments`, this feature is not yet released so there is no need for a back compat approach here

packages/js-core/src/lib/survey/widget.ts
```diff
@@ -22,6 +22,43 @@ export const setIsSurveyRunning = (value: boolean): void => {
...
+  userId: string | null,
+  survey: TWorkspaceStateSurvey,
+  source: TInteractionSource
+): void => {
+  if (!userId) return;
+
+  const config = Config.getInstance();
+  const shouldRefresh = survey.interactionRefresh
+    ? survey.interactionRefresh[source]
+    : (config.get().workspace.data.hasSurveyInteractionSegments ?? false);
```
pandeymangg: I think we should also remove the `hasSurveyInteractionSegments` from the config state as well, it does nothing anymore now. The comments here are also misleading about backwards compatibility, the feature has not been released yet 🙏

packages/types/segment.ts
```diff
@@ -356,6 +434,117 @@ export const ZJsWorkspaceStateSegment = z.object({
 });
 export type TJsWorkspaceStateSegment = z.infer<typeof ZJsWorkspaceStateSegment>;
 
+/**
+ * Recursively scans a segment's filter tree for any `surveyInteraction` leaf. Used server-side to
+ * expose a single `hasSurveyInteractionSegments` flag in the workspace state so the SDK only pays for
+ * a post-interaction `/user` segment refresh when interaction-based targeting is actually in play.
+ */
+export const segmentFiltersContainSurveyInteraction = (filters: TBaseFilters): boolean => {
```
pandeymangg: not needed anymore 🙏

packages/types/segment.ts
```diff
@@ -112,18 +134,42 @@ export const ZRelativeDateValue = z.object({
...
+// The base object is kept separate from the refined schema so the exported TYPE is inferred from a
+// plain ZodObject (spreadable in consumers); the refinement only adds runtime validation and does not
+// change the output shape.
+const ZSegmentSurveyInteractionFilterValueBase = z.object({
+  surveyScope: z.enum(["any", "specific"]),
+  // Bounded to cap the payload from an untrusted client (no unbounded array). 100 surveys is far
+  // beyond any realistic single-filter selection. Id validity/ownership is enforced at write time by
+  // assertSurveyInteractionSurveyIds (checks the ids exist in the workspace), which is stronger than a
+  // format check, so we do not additionally constrain the string shape here.
+  surveyIds: z.array(z.string()).max(100),
```
pandeymangg: I think it should be ZId instead of z.string() 🙏

pandeymangg (changes requested):
works well! Please take a look at the comments! 🙏 
we should remove all the references to the `hasSurveyInteractionSegments` as the field and that concept is not used anymore, the comments mentioning that this is related to backwards compatibility should also be removed

pandeymangg (dismissed):

pandeymangg (approved):
LGTM 🚀

PR #8591, opened 2026-07-20
fix: enforce workspace-level access in getWorkspaceAuth [ENG-1769]

xernobyl:
## What does this PR do?

`getWorkspaceAuth` (the workspace-scoped auth helper behind ~35 pages/routes) only asserted org membership and left the real workspace-access gate to the `[workspaceId]` route layout. That made it a footgun to reuse: any page/route calling it outside that layout would admit an org member with no `WorkspaceTeam` grant and show them write UI. On top of that, `isReadOnly = isMember && hasReadAccess` failed open — a team-less member (no resolved permission) came back as `isReadOnly = false`, i.e. mislabeled as a writer.

This isn't a live exploit today (every current caller sits under the layout that already blocks team-less members), but it's a latent gap flagged by the internal security audit (I-34c) — and it's already being worked around, since `getWorkflowsRouteAuth` re-derives the missing check by hand.

Changes:

- Make `getWorkspaceAuth` self-contained: it now calls `hasUserWorkspaceAccess` itself and throws `AuthorizationError` when the caller has no access to the workspace (the same check its siblings `getWorkspaceLayoutData` / `workspaceIdLayoutChecks` already use). Runs in parallel with the existing permission lookup, so no added latency.
- Make `isReadOnly` fail safe: derive it from the *absence* of write/manage access (`isMember && !hasReadWriteAccess && !hasManageAccess`), so a member with no resolved permission is the most restricted, never a writer.
- Fix a stale doc comment (it referenced a `getEnvironmentAuth` that no longer exists) and add unit coverage (there was none for this file).

Notes:

- No behavior change for anyone who can currently reach these pages (owner/manager/billing, or a member with a grant) — only team-less members are newly rejected, and they were already blocked upstream by the layout.
- Billing-role access is intentionally unchanged here; that's tracked separately as I-30 ([ENG-1763](https://linear.app/formbricks/issue/ENG-1763)).
- Targets `main`; it flows into `epic/workflows-v1` via the regular main merge (the function is identical on both today).

Linear: https://linear.app/formbricks/issue/ENG-1769

## How should this be tested?

Unit tests (from `apps/web`):

- `pnpm vitest run modules/workspaces/lib/utils.test.ts` — covers the new access gate (team-less member → `AuthorizationError`, and asserts the check is called with the right `userId`/`workspaceId`) plus the full `isReadOnly` matrix (read → read-only; readWrite/manage → not; no resolved permission → read-only; owner → not).
- `pnpm vitest run modules/workspaces modules/workflows/lib/auth.test.ts redirect-billing-role` — the existing suites that consume/mock `getWorkspaceAuth` still pass.

Behavioral check (optional): as an org member with no team grant on a given workspace, hitting a page backed by `getWorkspaceAuth` should return "Not authorized" instead of rendering write UI. Owner/manager/billing and members with a grant are unaffected.

`pnpm lint` and the changed-file typecheck are clean. (Skipped a full `pnpm build` for a two-file auth change.)

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/modules/workspaces/lib/utils.test.ts
+++ b/apps/web/modules/workspaces/lib/utils.test.ts
@@ -1,0 +1,145 @@
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { AuthenticationError, AuthorizationError, ResourceNotFoundError } from "@formbricks/types/errors";
+import type { TMembership, TOrganizationRole } from "@formbricks/types/memberships";
+import type { TOrganization } from "@formbricks/types/organizations";
+import type { TWorkspace } from "@formbricks/types/workspace";
+import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
+import { getOrganization } from "@/lib/organization/service";
+import { hasUserWorkspaceAccess } from "@/lib/workspace/auth";
+import { getWorkspace } from "@/lib/workspace/service";
+import { getSession } from "@/modules/auth/lib/session";
+import { getWorkspacePermissionByUserId } from "@/modules/ee/teams/lib/roles";
+import { getWorkspaceAuth } from "./utils";
+
+vi.mock("@/lib/membership/service", () => ({
+  getMembershipByUserIdOrganizationId: vi.fn(),
+}));
+vi.mock("@/lib/organization/service", () => ({
+  getOrganization: vi.fn(),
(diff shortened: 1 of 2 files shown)


Dhruwang (changes requested):
Looks good. Can you please resolve the merge conflicts on the PR?

Dhruwang (approved):

PR #8592, opened 2026-07-21
fix(auth): block billing-role members from workspace product data [ENG-1763]

pandeymangg:
## What does this PR do?

Closes **ENG-1763** (I-30, security — High).

**Problem.** The two workspace-access helpers disagree on the `billing` role: `hasUserWorkspaceAccess` returns `true`, while `hasUserWorkspaceAccessForAction` returns `false`. The workspace layout gates on the permissive one, and product pages (contacts, survey summary/responses, dashboards, unify) authorize via `getWorkspaceAuth`, which only checks that a membership exists — it never re-checks `isBilling`. As a result a finance user invited as `billing` (intended: billing/enterprise screens only) could direct-navigate to `/workspaces/{id}/contacts`, a dashboard, or a survey summary and read customer emails/attributes and feedback aggregates. Because `isReadOnly = isMember && hasReadAccess` is `false` for billing, they were also treated as non-read-only, so write CTAs rendered too.

**Fix.** Add the billing guard at the single choke point every workspace product page already flows through — `getWorkspaceAuth`. A billing-role member is now redirected to their org billing/enterprise home (`getBillingFallbackPath`), exactly as the settings pages and the workspace root page already do. This closes every exposed product surface at once (contacts, dashboards, charts, unify, survey summary/responses) and realigns `getWorkspaceAuth` with `hasUserWorkspaceAccessForAction`, which already denies billing. Existing inline billing guards on individual pages remain correct as defense in depth.

Chosen over the alternative (tightening `hasUserWorkspaceAccess` itself) because that helper also gates the workspace *layout*, so denying billing there would turn a clean redirect into an `AuthorizationError` error page.

### Out of scope (noted for follow-up)
The integration API routes (`notion`/`airtable`/`slack`/`google-sheet`) also use the permissive `hasUserWorkspaceAccess` and would let a billing user hit integration setup. That's a privilege issue rather than the PII/product-data leak this ticket covers — likely a separate follow-up.

## How should this be tested?

Automated: `apps/web/playwright/billing-role-access.spec.ts` (new) — both tests pass locally.

1. **Billing member is redirected** — a billing-role member of an org navigates directly to `/workspaces/{id}/contacts` and `/workspaces/{id}/dashboards`; asserts they land on `/organizations/{id}/settings/{billing|enterprise}` and never on the product route.
2. **Owner keeps access** (control) — an owner navigates to `/contacts` and stays there, so the fix doesn't blanket-redirect legitimate roles.

Manual:
- Invite a user as `billing`, then as that user open `/workspaces/{id}/contacts`, a dashboard, and a survey summary → each should redirect to the billing/enterprise home.
- Repeat as an `owner`/`manager`/`member` → product pages load as before.

```
pnpm --filter=@formbricks/web exec playwright test billing-role-access
```

### Required

- [ ] Ran `pnpm build`

### Appreciated

### DIFF
--- a/apps/web/modules/workspaces/lib/utils.test.ts
+++ b/apps/web/modules/workspaces/lib/utils.test.ts
@@ -1,0 +1,87 @@
+import { redirect } from "next/navigation";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import type { TMembership, TOrganizationRole } from "@formbricks/types/memberships";
+import { getBillingFallbackPath } from "@/lib/membership/navigation";
+import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
+import { getOrganization } from "@/lib/organization/service";
+import { getWorkspace } from "@/lib/workspace/service";
+import { getSession } from "@/modules/auth/lib/session";
+import { getWorkspacePermissionByUserId } from "@/modules/ee/teams/lib/roles";
+import { getWorkspaceAuth } from "./utils";
+
+const mocks = vi.hoisted(() => ({ isFormbricksCloud: false }));
+
+// Real getAccessFlags is used on purpose so the tests exercise the actual role -> isBilling mapping
+// that the redirect branch keys off of. Everything else getWorkspaceAuth touches is stubbed.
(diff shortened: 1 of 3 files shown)


Dhruwang (approved):
Thank you for the PR looks good 🚀

PR #8594, opened 2026-07-21
refactor: merge main into epic/workflows-v1

itsjavi:
## What does this PR do?

Merges the latest `main` into `epic/workflows-v1` to keep the epic branch up to date (follow-up to #8525).

- Merged `main` at 0fcdf3be4 (39 commits since the last sync), on top of the epic tip 834078105.
- **Conflicts resolved** (2 files, both in the MCP module — main added the `list_workspaces` tool via #8577 while the epic registers the workflow tools):
  - `apps/web/modules/mcp/server.ts` — kept both registrations (`registerWorkflowTools` + `registerWorkspaceTools`).
  - `apps/web/app/api/mcp/route.test.ts` — `tools/list` expectation now contains the union of both tool sets, ordered to match registration order (survey → workflow → workspace tools).

## Verification

- `pnpm test` ✅ (25/25 tasks; `@formbricks/web` ran fresh: 551 files / 6,696 tests passed)
- Forced fresh test runs for the epic-owned packages ✅ (`workflows` 209, `jobs` 62, `database` 35, `audit-logs` 5)
- `pnpm run scan-translations` ✅ (all locales complete)
- OpenAPI v3 spec: `bundle.mjs --lint` and `--check` ✅ (committed bundle in sync)
- `pnpm install --frozen-lockfile` ✅ (lockfile consistent)

**Note:** `turbo typecheck` has 2 pre-existing failures in `packages/database` (`vite.config.ts` PluginOption type, `zod/workflows.ts` importing model types from `@prisma/client`). Both reproduce identically on the epic tip before this merge and are untouched here; CI does not run typecheck.

PR #8600, opened 2026-07-21
fix: prevent feedback-source ingestion crash on malformed choice elements (ENG-1939)

xernobyl:
## What & why

Feedback-source ingestion crashed with `TypeError: Cannot read properties of undefined (reading 'some')` when a choice element reached `normalizeElementValue` without a `choices` array. Elements come straight from the stored `survey.blocks` JSON, which isn't re-validated on this path — so the array can be missing at runtime even though the Zod schema requires it (`.min(2)`) and TypeScript therefore treats it as always present. Introduced by the `value_id` work in ENG-1673 (#8464), spotted by CodeRabbit while reviewing the main→epic sync (#8594), pre-existing on `main`.

Five small commits — the first four stop the crash, the last closes the hole that lets the bad data exist:

1. **`normalizeElementValue` guard** — fall back to `[]` like the two siblings in the same file that already guarded (`normalizeChoiceValue`, `expandMultiChoiceToRecords`). This is the actual bug.
2. **Contain per-response transform failures in the historical import** — the live pipeline already isolates each response in a `try/catch`, but `processBatch` flat-mapped the transform across a whole batch unguarded, so one throwing response aborted the entire import and lost every remaining batch. Now it logs (`surveyId`, `responseId`, message) and skips that one response.
3. **Hoist `getErrorMessage`** into `./utils` — commit 2 would otherwise have left `import.ts` and `pipeline-handler.ts` holding identical copies.
4. **Same guard for ranking (`element.choices`) and matrix (`element.rows`)** — both passed the array straight into `findChoiceByLabel`, which calls `.find()` on it: the same crash one function over. Matrix `columns` already flows through the guarded `normalizeChoiceValue`, so only `rows` needed it.
5. **Only allow skipping survey validation while the survey is a draft** — the root cause of how invalid blocks get stored. `ZSurveyDraft` types blocks as `z.array(z.record(z.string(), z.unknown()))` and never applies the element schemas, then `updateSurveyInternal` runs with `skipValidation=true`. Nothing checked that the *persisted* survey was a draft — only the payload's own `status`, which the caller supplies. So an authenticated editor could push structurally invalid blocks onto a **live** survey (one that already has responses, which is exactly what makes the ingestion crash reachable) and silently revert it to draft, stopping it collecting responses. Now gated on `currentSurvey.status`, placed after the ENG-1749 tenant guards so a cross-workspace attempt still reports the authorization failure first, and before any write.

Net effect: all element-array accesses in `transform.ts` are consistently guarded, any residual throw is contained on **both** ingestion paths (one skipped response + a log, never a fatal batch abort), and the write path that produced the invalid data is closed.

Scope note: (1) and (2) are what the ticket asks for (it lists the import guard as optional hardening); (3) is cleanup of duplication that (2) introduced; (4) is the same bug class in adjacent functions; (5) is outside `feedback-source/` but without it we'd only ever be catching bad data, never preventing it. Happy to split (4) or (5) out if a reviewer prefers.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1939/bug-normalizeelementvalue-crashes-on-a-choice-element-with-no-choices

## How this was tested

- `pnpm exec vitest run lib/survey/ modules/survey/editor/ lib/feedback-source/` → ✅ **705 passed** (26 files), after merging `main`.
- `pnpm exec eslint` on all 9 changed files → ✅ clean (including under main's new ESLint 9 flat config).
- **Every guard proven to fail without its fix.** For the ingestion guards I swapped in `origin/main`'s pre-fix `transform.ts` and re-ran: each case reproduced the real crash — `reading 'some'` at `normalizeElementValue` for choice single/multi, `reading 'find'` at `findChoiceByLabel` for ranking and matrix — and the import case rejected instead of completing. For the draft guard I deleted the check and re-ran: all three `inProgress`/`paused`/`completed` cases failed. Restoring each fix turned them green.
- To rule out the tests only "working" because of an `as unknown as TSurvey` cast, I ran a throwaway repro that builds the survey via `JSON.parse` of a raw blocks string (the shape Prisma actually returns for a `Json` column): **5/5 crash on `origin/main`, 5/5 pass on this branch.** So the defect is in the data path, not an artifact of test typing.
- Verified containment is complete: `transformResponseToFeedbackRecords` has exactly two callers (`pipeline-handler.ts`, `import.ts`) and both are now guarded; the crash sites are private to `transform.ts`, so the CSV import path never reaches them.
- Verified the draft guard affects no real flow: the editor only takes the draft path for drafts (`survey-menu-bar.tsx` routes on `status === "draft"`, and autosave is draft-only), while pause/complete go through the strict `updateSurveyAction` and the v3 `PATCH`. Six existing tests had to move because they modelled combinations production never performs — two service tests reused an `inProgress` fixture for what their own comments call a draft save, and four scheduling tests passed `skipValidation=true` for live-survey transitions that production runs through the validated path. Both were corrected to match real behavior rather than relaxing the guard; the four scheduling tests pass unchanged under strict validation, confirming their fixtures were always valid.

## QA / Test Plan

**How to test**

Note for the tester: the ingestion crash needs a survey whose stored `blocks` JSON is missing a `choices`/`rows` array. The product UI won't create that shape, so that part is **not reproducible by clicking** — it's covered by unit tests. What's worth verifying by hand is that normal ingestion and the survey editor did not regress.

- [ ] With an enterprise license (Unify Feedback / feedback directories enabled), go to Unify Feedback → Sources and create a feedback source for a survey that has a **multiple-choice (single)**, a **multiple-choice (multi)**, a **ranking** and a **matrix** question → the source is created without error.
- [ ] In the create dialog, choose to import historical responses → the import finishes and reports a non-zero success count (no error toast, no partial abort).
- [ ] In the Hub feedback directory, confirm records exist for each question type: choice answers show the option label, ranking shows one record per ranked item, matrix shows one record per answered row.
- [ ] Submit a **new** response to that survey (live path) → within a few seconds new records appear in the feedback directory for that submission.
- [ ] Edge case — answer a choice question via its **"Other"** free-text field → a record is stored with the typed text (it groups under the stable "other" option rather than as its own bucket).
- [ ] Edge case — answer only *some* questions and submit → only the answered questions produce records; the import/pipeline reports no failures.

Survey editor regression checks (these exercise commit 5 — none should behave differently than today):

- [ ] Create a new survey, add a multiple-choice question but **don't** fill in the options, leave it a few seconds → the draft autosaves with no error (incomplete drafts must still save).
- [ ] Reload the editor → the half-finished question is still there.
- [ ] Complete the question and publish → publishes normally.
- [ ] On a **published** survey: pause it, then resume it, then mark it completed → each status change succeeds with no error toast, and the status badge updates.
- [ ] Edit a **published** survey (change a headline) and save → saves normally.

**Preconditions / test data**

- Enterprise license with feedback directories enabled (the page shows an upgrade / request-trial prompt otherwise), plus a reachable Hub instance and a configured feedback directory.
- A survey with at least one each of: multiple-choice single, multiple-choice multi, ranking, matrix — and a handful of existing responses so the historical import has something to move.
- For the editor checks: one draft survey and one already-published survey.

**Risks & regressions**

- Nearby behavior to re-check: `value_id` / canonical `value_text` on choice answers (ENG-1673) and the per-option split for multi-selects (ENG-1702) — the guarded code sits directly in that logic. Charts that group choice answers cross-language are the visible consumer.
- Commit 5 changes an existing endpoint's behavior: a draft-shaped update aimed at a non-draft survey now returns an error instead of silently succeeding. No shipped client does that, but it's the thing to watch — the symptom would be an error toast when saving or changing the status of a **published** survey, which is why the editor checks above are worth running.
- Historical-import counters: a response whose transform now throws is tallied under **`skipped`** rather than `failures` (the cause is logged with its `responseId`). Deliberate — it keeps the run alive — but worth knowing if a tester compares counts against the response total.
- Malformed elements now yield *no* records for that element instead of crashing, which is silent. That matches how the code already treats unmatched ranking/matrix labels, so it isn't new data loss, but it does mean bad survey data won't announce itself here.

- none


xernobyl:
@coderabbitai review

Dhruwang (approved):

PR #8601, opened 2026-07-22
fix(scheduling): drop NEXT_PUBLIC_ prefix from survey scheduling env vars (ENG-1665)

Dhruwang:
## Problem

The env vars controlling the survey publish/close scheduling execution time were prefixed with `NEXT_PUBLIC_`:

- `NEXT_PUBLIC_SURVEY_SCHEDULING_TIME_ZONE`
- `NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_HOUR`
- `NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_MINUTE`

Next.js inlines `NEXT_PUBLIC_*` vars into the bundle **at build time**. This means self-hosting users running our prebuilt Docker image cannot configure these values at runtime — they are baked into the image. (Reported by @mattinannt on Slack.)

Closes ENG-1665.

## Solution

Move the three vars to **server-only** env vars (no `NEXT_PUBLIC_` prefix), read from `process.env` at runtime:

- `SURVEY_SCHEDULING_TIME_ZONE`
- `SURVEY_SCHEDULING_LOCAL_HOUR`
- `SURVEY_SCHEDULING_LOCAL_MINUTE`

The scheduling feature is functionally server-side (the daily reconciliation cron in `instrumentation-jobs.ts` and normalization in `survey-scheduling.ts`), so those paths now pick up runtime config for free.

The survey editor is a client tree and also needs these values — for the schedule label ("Survey will be published at `HH:MM` in the `<tz>` timezone") and for calendar date math. Since server-only env vars are `undefined` in the browser bundle, the resolved config is now **threaded from the server** down to the client:

- `page.tsx` (server) reads `SURVEY_SCHEDULING_CONFIG` and passes it as a `surveySchedulingConfig` prop
- `SurveyEditor` → `SettingsView` → `ResponseOptionsCard` forward the prop (same pattern as the existing `locale` prop)
- `date-utils.ts` is refactored from module-level constants into `createSurveySchedulingDateUtils(config)`, so the exact same timezone/hour/minute logic runs on both the server (reconciliation) and the client (editor) with runtime-configured values
- Client-safe helpers (`TSurveySchedulingConfig`, `getSurveySchedulingTimeLabel`) live in a new `config.ts` that does not touch `process.env`

This keeps the editor label and calendar boundaries correct for any runtime override, instead of falling back to the build-time default.

## Changes

- `apps/web/lib/env.ts` — move the three vars from the `client` block to the `server` block; drop the prefix; update `runtimeEnv` and the validation message
- `apps/web/modules/survey/scheduling/lib/constants.ts` — read non-prefixed vars from `process.env`; export `SURVEY_SCHEDULING_CONFIG`
- `apps/web/modules/survey/scheduling/lib/config.ts` — **new** client-safe config type + label helper
- `apps/web/modules/survey/scheduling/lib/date-utils.ts` — `createSurveySchedulingDateUtils(config)` factory; `toDateOnlySelection`/`isDateDue` stay pure
- `apps/web/modules/survey/scheduling/lib/survey-scheduling.ts` — instantiate the utils with `SURVEY_SCHEDULING_CONFIG`
- `apps/web/modules/survey/editor/{page,components/survey-editor,components/settings-view,components/response-options-card}` — thread `surveySchedulingConfig` prop
- `turbo.json` — move the vars out of `build.env` (no longer build inputs) into `passThroughEnv`
- `.env.example` — document them as server-side runtime vars

## Testing

- Updated `env.test.ts` and `date-utils.test.ts`; added a test asserting a runtime-configured timezone/hour/minute is honored
- `pnpm vitest run lib/env.test.ts modules/survey/scheduling` → **48 passed**
- `eslint` and `prettier` clean on all changed files
- Type-checked the touched files (no new errors)

## Notes for reviewers

- No breaking change for Formbricks Cloud (defaults unchanged: `Europe/Berlin`, `00:00`).
- **Self-hosters who set the old `NEXT_PUBLIC_*` vars must rename them** to the non-prefixed names. Called out in `.env.example`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,11 +44,12 @@ LOG_LEVEL=info
 # Number of concurrent jobs each BullMQ worker can process.
 # BULLMQ_WORKER_CONCURRENCY=1
 
-# Survey publish/close scheduling is configured with public build-time env vars because the editor UI
-# also needs to render the selected execution time and timezone.
-# NEXT_PUBLIC_SURVEY_SCHEDULING_TIME_ZONE=Europe/Berlin
-# NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_HOUR=0
-# NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_MINUTE=0
+# Survey publish/close scheduling execution time. These are server-side runtime env vars, so
+# self-hosted Docker deployments can change them without rebuilding the image. The editor UI
+# receives the configured values from the server to render the selected execution time and timezone.
+# SURVEY_SCHEDULING_TIME_ZONE=Europe/Berlin
+# SURVEY_SCHEDULING_LOCAL_HOUR=0
+# SURVEY_SCHEDULING_LOCAL_MINUTE=0
 
 ##############
 #  DATABASE  #
--- a/apps/web/lib/env.test.ts
+++ b/apps/web/lib/env.test.ts
@@ -301,35 +301,35 @@ describe("env", () => {
 
   test("uses the default survey scheduling configuration when env vars are not set", async () => {
     setTestEnv({
-      NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_HOUR: undefined,
-      NEXT_PUBLIC_SURVEY_SCHEDULING_LOCAL_MINUTE: undefined,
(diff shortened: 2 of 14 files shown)


xernobyl (approved):
LGTM

PR #8609, opened 2026-07-22
feat: soft-delete (archive) surveys (ENG-1042)

Dhruwang:
## Context

Users (and regulators in some jurisdictions) need an "undo" for deleting surveys. Instead of an irreversible hard delete, this adds **archive / restore** with a 30-day retention window before permanent deletion.

- Linear: [ENG-1042](https://linear.app/formbricks/issue/ENG-1042/design-soft-delete-surveys)
- Figma: node `8315-2926` in Formbricks-UI

Archived surveys are hidden from the list by default, surfaced via an "Archived" status filter, restorable, permanently deletable, and auto-purged after 30 days.

### Data model
- New nullable `Survey.archivedAt` + `@@index([archivedAt])` (migration `20260722000000_add_survey_archived_at`).
- **Not** a new `SurveyStatus` enum value — the enum is baked into zod types, the v3 document schemas, SDK sync, and response intake; adding a value would ripple everywhere. `archivedAt` is an orthogonal soft-delete marker.

### Archive semantics
- **Archive** → set `archivedAt = now`, clear `publishOn`, and if the survey was `inProgress` move it to `paused`. Pausing stops response/display intake immediately (intake gates on `status !== "inProgress"`); clearing `publishOn` stops the scheduling cron from ever auto-publishing an archived (or later restored) survey. Idempotent.
- **Restore** → clear `archivedAt`; status is preserved (a formerly in-progress survey returns paused, user re-publishes manually).
- **Delete forever** → existing hard `deleteSurvey` (cascades + private-segment cleanup).

### Server / API (v3)
- Dedicated sub-routes `POST /api/v3/surveys/:id/archive` and `.../restore` (not a PATCH-body flag — archive has a status side-effect that doesn't belong in the document pipeline). Standard `withV3ApiWrapper` + `getAuthorizedV3Survey({ access: "readWrite" })` + audit log (`archived` / `restored` actions added to `ZAuditAction`).
- **List filtering** goes through one choke point, `buildWhereClause`:
  - default → `archivedAt: null` (archived hidden everywhere it's used: list, relevance sort, counts)
  - `includeArchived` + statuses → `OR(active-of-status, any-archived)`
  - `includeArchived` only → archived only
- `"archived"` is a **pseudo-status** in the list filter: the client sends it in `filter[status][in]`, the query parser splits it into `includeArchived`; it is never a real status server-side.
- `hasArchived` meta on the first list page drives show/hide of the "Archived" filter option.

### Read-only enforcement (defense in depth)
- **v3 PATCH** rejects edits to an archived survey (422).
- The **shared update service** (`updateSurveyInternal`, behind `updateSurveyAction` / editor / summary status dropdown) also rejects writes to archived surveys — so every write path is covered, not just the v3 API. `archivedAt` is stripped from update payloads; only archive/restore own that field.
- **SDK / triggers**: the environment-state query and the contact link-survey query exclude archived (`archivedAt: null`). Archiving already forces a survey out of `inProgress` (the SDK filter), so an archived app survey can't be synced or triggered by a code action — the guard makes it explicit.
- **Scheduling job**: excludes archived from auto publish/close candidates.

### 30-day purge (BullMQ)
- **No per-survey delayed job.** A single global recurring cron `daily-survey-archive-purge` (daily at 01:30, registered on worker boot in `instrumentation-jobs.ts` next to survey scheduling) scans `archivedAt < now − 30d`, batches of 100, hard-deletes each with per-survey error isolation, a no-progress guard against infinite loops, and a system-actor audit entry.
- This keeps restore free (nothing to cancel) and makes the retention window a single constant. A survey is deleted on the first daily run after it crosses 30 days (≈ up to 24h after the exact mark).

### Client / UI
- New TanStack Query hooks `useArchiveSurvey` / `useRestoreSurvey` (optimistic removal + rollback + list invalidation), following the existing `useDeleteSurvey` pattern.
- **Survey list**: per-card dropdown shows "Archive" (last item, destructive) with a soft confirm for draft/paused/completed and a **hard** "Stop and archive" confirm for in-progress. Archived cards show only **Restore** and **Delete forever**. Archived status chip (archive icon in the same round badge as other statuses). Status filter gains a conditional "Archived" option (hidden when the workspace has none). Empty-state guarded so a workspace with only archived surveys can still reach the filter; stale `"archived"` filter is auto-dropped when none remain.
- **Summary page**: status control renders a disabled "Archived" select (reads as a natural status); Share, Edit, Reset, Generate-examples, and Configure-alerts are hidden while archived (read-only). Refresh / Preview remain.

### i18n
- New strings added to `en-US.json` only; all locales + lockfile regenerated via Lingo (`pnpm i18n`).

## Out of scope / decisions
- **Public v1 management API left unchanged** — archived surveys remain visible there to avoid a breaking change for existing API consumers (product decision).
- The purge cron is **global-scope** (single runner); BullMQ scheduler-id dedup keeps it safe under multiple worker replicas.

## Testing
- **Unit (vitest):** `buildWhereClause` (default hide / archived-only / OR), archive & restore services (in-progress→paused + publishOn cleared, idempotency, P2025 mapping), query parsing (`archived` → `includeArchived`), v3 operations (archive/restore auth + audit, `hasArchived` meta, PATCH-while-archived 422), archive/restore hooks (optimistic + rollback), purge job (cutoff boundary, per-survey error isolation, no-progress abort), scheduling + SDK/contacts exclusion.
- **E2E (Playwright):** `survey-archive.spec.ts` — full journey: archive in-progress (hard dialog) → hidden → filter → archived card actions → restore → re-archive (soft dialog) → delete forever.
- `pnpm typecheck` clean; `pnpm i18n` validation passes; migration applied via `pnpm db:migrate:dev`.

## Migrations / env
- One additive migration (nullable column + index) — safe, no backfill.
- No new env vars (purge cadence/retention are constants; TZ reuses the existing survey-scheduling TZ).


packages/types/surveys/types.ts
```diff
@@ -950,6 +950,10 @@ export const ZSurveyBase = z.object({
     .date()
     .nullish()
     .transform((value) => value ?? null),
+  // Soft-delete marker (ENG-1042). Optional (no transform) so it stays non-required on the inferred
+  // TSurvey type — existing survey literals/payloads that omit it keep compiling and validating.
+  // Reads populate it from the DB so archived state is available to the editor/summary surfaces.
+  archivedAt: z.coerce.date().nullish(),
```
itsjavi, line 956: **Blocker — `archivedAt` is mass-assignable on create, and the purge cron then deletes the survey.**

`archivedAt` landed on `ZSurveyBase`, but neither `ZSurveyCreateInput` (:3840) nor `ZSurveyCreateInputWithWorkspaceId` (:3888) omits it. It survives `.partial().omit()`, rides `restSurveyBody` through `createSurvey` (`apps/web/lib/survey/service.ts:739`), and lands in `tx.survey.create`.

**Why this matters:** `POST /api/v1/management/surveys` with `{"archivedAt": "2020-01-01T00:00:00Z"}` creates a survey that is invisible in every in-app list and is hard-deleted by the next purge run — no `archiveSurvey` call, no `archived` audit event, no 30-day window. Add `"status": "inProgress"` and it also collects responses in the meantime, since intake only gates on status. This is reachable by accident too: `selectSurvey` now returns `archivedAt` (`service.ts:59`), so the common "GET a survey, POST it back to clone it" integration silently produces an archived, purge-eligible copy.

**Possible fix:** add `archivedAt: true` to the `.omit({…})` list in both create schemas, mirroring what `updateSurveyInternal` already does deliberately ("owned exclusively by the archive/restore flows", `service.ts:320`). Worth a unit test asserting a create payload carrying `archivedAt` doesn't persist it.

apps/web/lib/survey/service.ts
```diff
@@ -259,6 +262,8 @@ export const getSurveyCount = reactCache(async (workspaceId: string): Promise<nu
     const surveyCount = await prisma.survey.count({
       where: {
         workspaceId,
+        // Archived surveys are hidden by default across the app.
+        archivedAt: null,
```
itsjavi: **Blocker — this makes archiving your last survey re-trigger the onboarding flow on every login.**

This `getSurveyCount` has exactly one consumer: the onboarding gate (`app/(app)/(onboarding)/lib/redirect-if-onboarding-complete.ts:8,26`). `app/page.tsx:56` redirects any owner/manager whose onboarding workspace reports 0 surveys to `/organizations/{org}/workspaces/new/plan` (Cloud) or `…/new/survey` (self-hosted), and `redirectIfOnboardingComplete` then keeps them there.

**Why this matters:** login pushes `"/"`, so once the last survey in that workspace is archived, every subsequent login lands the user in "create your first survey" — a full-screen page with no app navigation and therefore no route to the Archived filter — while the archived survey counts down to permanent deletion. Pre-PR this count was unfiltered, so the state was unreachable.

**Possible fix:** keep the onboarding count archive-inclusive — either leave this query unfiltered, or add an explicit `includeArchived` option and pass it from the onboarding gate. Same spirit as the `hasArchived` guard you added to the surveys-list empty states.

apps/web/modules/survey/lib/surveys.ts
```diff
@@ -49,3 +71,83 @@ export const deleteSurvey = async (surveyId: string) => {
...
+
+      if (survey.archivedAt) {
+        return survey;
+      }
+
+      return await tx.survey.update({
+        where: { id: surveyId },
+        data: {
+          archivedAt: new Date(),
+          publishOn: null,
```
itsjavi: **Blocker — archive irreversibly destroys a scheduled survey's publish date, and the dialog tells the user the opposite.**

`archiveSurvey` writes `publishOn: null`; `restoreSurvey` (:131) only clears `archivedAt`. A "Scheduled" survey is exactly `status === "paused" && publishOn !== null` (`survey-card.tsx:42`) — and it takes the *soft* confirm path, whose copy reads "You can restore it anytime within the next 30 days."

**Why this matters:** archive → restore returns a plain paused survey with the publish date gone for good, so a survey scheduled to launch next month silently becomes one that never launches. That's unrecoverable user configuration lost by an operation the UI sells as reversible. `closeOn` is left untouched, so restore can also produce `paused` + `publishOn: null` + a past `closeOn` — the combination `normalizeSurveyScheduling` deliberately normalizes away. Related: an in-progress survey comes back paused, and neither the dialog nor the "Survey restored" toast says it must be re-activated.

**Possible fix:** drop `publishOn: null` from the archive update — it's redundant, since this same PR added `archivedAt: null` to both the scheduling candidate scan and the conditional `updateMany` (`scheduling/lib/survey-scheduling.ts:178`, `:204`), so an archived survey can't auto-publish either way. Note the unit test "should archive a paused survey without changing its status" currently asserts the lossy behaviour. If the clearing is deliberate, then snapshot and restore it, clear `closeOn` for consistency, and say so in the dialog copy.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/SurveyAnalysisCTA.tsx
```diff
@@ -232,20 +232,20 @@ export const SurveyAnalysisCTA = ({
...
       disabled: isGeneratingExamples || aiUnavailableReason !== null || responseCount > 0,
-      isVisible: !isReadOnly,
+      isVisible: !isReadOnly && !survey.archivedAt,
     },
     {
       icon: ListRestart,
       tooltip: t("workspace.surveys.summary.reset_survey"),
       onClick: () => setIsResetModalOpen(true),
-      isVisible: !isReadOnly,
+      isVisible: !isReadOnly && !survey.archivedAt,
```
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/SurveyAnalysisCTA.tsx:248]: **Blocker — Reset and Generate-examples are hidden here but not blocked server-side.**

`resetSurveyAction` (`../../actions.ts:83`) has no `archivedAt` check and calls `deleteResponsesAndDisplaysForSurvey`. `generateExampleResponsesAction` (`:135`) is the same shape in reverse — it writes responses, displays and a tag into an archived survey.

**Why this matters:** a stale tab or a direct action call permanently deletes every response and display of an archived survey — precisely the data the 30-day window exists to preserve, so restore hands back an empty survey. It's not an authz bypass (both require `readWrite`), but `generateExampleResponsesAction` already re-checks `responseCount` server-side with the comment "we re-check here so a stale tab can't insert noise into a live survey", which is exactly the rule being skipped here.

Everything else this PR hides *is* enforced server-side (`updateSurveyInternal:302`, the v3 PATCH 422, `renderSurvey` `notFound()`, the SDK sync exclusion) — these two are the only gaps I found.

**Possible fix:** reject both with an `OperationNotAllowedError` when `survey.archivedAt` is set; both actions already have the survey loaded, so it's a two-line guard each. The `generateExampleResponsesAction` one belongs in its existing stale-tab guard.

apps/web/modules/survey/list/components/survey-dropdown-menu.tsx
```diff
@@ -193,7 +235,38 @@ export const SurveyDropDownMenu = ({
         </DropdownMenuTrigger>
         <DropdownMenuContent className="inline-block w-auto min-w-max">
           <DropdownMenuGroup>
-            {canManageSurvey && (
+            {isArchived && canManageSurvey && (
+              <DropdownMenuItem>
+                <button
+                  type="button"
+                  data-testid="restore-survey"
```
itsjavi: **Blocker — archived cards are a keyboard dead-end: Restore and Delete forever are mouse-only.**

All three new items (`restore-survey`, `delete-survey-forever`, `archive-survey`) put the handler on a `<button>` nested inside `DropdownMenuItem`. Radix prevents `Tab` inside menu content and dispatches Enter/Space as `event.currentTarget.click()` on the `div[role="menuitem"]`, not on the nested button — so arrow-to-item + Enter closes the menu and does nothing.

**Why this matters:** the pattern is pre-existing for Edit/Duplicate/Delete, but archived cards expose *only* Restore and Delete forever. So a keyboard-only or screen-reader user who archives a survey has no way to get it back — the survey then purges in 30 days.

**Possible fix:** move the handlers onto the item, the way `copy-to-workspace` in this same file already does (:327). `DropdownMenuItem` takes an `icon` prop, so it's mechanical:

```tsx
<DropdownMenuItem
  data-testid="restore-survey"
  icon={<ArchiveRestoreIcon className="size-4" />}
  onSelect={(e) => { e.preventDefault(); void handleRestoreSurvey(); }}>
  {t("workspace.surveys.restore")}
</DropdownMenuItem>
```

Same for the other two (the red styling moves to the item's `className`). Worth a keyboard assertion in `survey-archive.spec.ts` — there's precedent in `survey-keyboard.spec.ts`.

apps/web/app/api/v3/surveys/[surveyId]/archive/route.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from "zod";
+import { withV3ApiWrapper } from "@/app/api/v3/lib/api-wrapper";
+import { archiveV3Survey } from "../../lib/operations";
+import { ZV3EmptyQuery } from "../../schemas";
+
+const surveyParamsSchema = z.object({
+  surveyId: z.cuid2(),
+});
+
+export const POST = withV3ApiWrapper({
```
Comment (itsjavi) [inline apps/web/app/api/v3/surveys/[surveyId]/archive/route.ts:10]: **Blocker — two new public `/v3` endpoints ship with no OpenAPI entry, and the published spec now contradicts the handler.**

`docs/api-v3-reference/src/` has zero changes in this PR. Missing: path files for `POST /api/v3/surveys/{surveyId}/archive` and `…/restore` (including the `{ id, status, archivedAt }` ack shape built in `lib/operations.ts`), `archivedAt` on `SurveyResource.yml` and `SurveyListItem.yml`, `meta.hasArchived` on the list response, and the new PATCH 422 reason.

**Why this matters:** it's worse than an omission on one point — `paths/api_v3_surveys.yml` documents `filter[status][in]` as `enum: [draft, inProgress, paused, completed]`, so the published contract states `archived` is a 400 while the handler accepts it. And `pnpm api:v3:check` only compares the committed bundle against `src/`, so with `src/` untouched CI stays green and the drift is invisible. The `api-v3-refining` guidance is explicit that anything under `/v3` must have a public spec entry, otherwise it belongs in the internal group.

**Possible fix:** add the two path files, `archivedAt` to both schemas, `hasArchived` to the list meta, `archived` to the `filter[status][in]` enum with a note that it's a pseudo-status, and the PATCH 422 reason — then `pnpm api:v3:bundle`.

itsjavi (changes requested):
Really solid design work — `archivedAt` as an orthogonal marker instead of a `SurveyStatus` value is the right call, `buildWhereClause` as a single choke point is the right shape, and the restore-vs-purge race is genuinely closed (the `SELECT … FOR UPDATE` + in-transaction re-check in `deleteSurvey`, not just a resolved thread). I also verified that tenancy holds throughout, the migration is safe, the cutoff arithmetic never purges early, and the `useDeleteSurvey` refactor is a pure move with no behaviour change.

What I'd like fixed before merge is the blast radius. Six blockers below, each as an inline remark. Two of them (1 and 4) cause silent, permanent loss of customer data; one (2) strands the user in onboarding.

A short recap of the six:

1. `archivedAt` is mass-assignable on create → born-archived surveys that the purge cron deletes.
2. `getSurveyCount` now excludes archived → archiving your last survey re-triggers onboarding on every login.
3. Archive nulls `publishOn` and restore never puts it back, while the dialog promises a full restore.
4. `resetSurveyAction` / `generateExampleResponsesAction` are hidden but not blocked for archived surveys.
5. Restore / Delete forever are mouse-only, so archived cards are a keyboard dead-end.
6. Two new public `/v3` endpoints ship with no OpenAPI entry, and the published spec now contradicts the handler.

Separately, on the required checklist: the description has no `## QA / Test Plan` section. `## Testing` lists which automated tests exist rather than how a human reproduces the behaviour, and all four required subsections (How to test / Preconditions / Risks & regressions / Migrations · env · cutover) plus the `## Checklist` are missing. Per AGENTS.md that section is the source of truth for release QA, so right now there's nothing to generate a release checklist from for a feature that can permanently delete customer data. Scenarios worth covering, none of which is exercised today: archive from each status and what restore returns; an archived survey over `/s/{id}` and `/c/{jwt}` including the OG preview; an archived app survey vs the SDK (workspace state, code-action trigger, the 60s Redis window); how to force a purge run and what the audit log should contain; a read-only member on an archived summary page; a workspace whose only surveys are archived; and archiving the last survey in your first workspace then logging out and back in.

Happy to re-review quickly once these are addressed.

Dhruwang:
Hey @itsjavi , thank you for the review. I've addressed all your feedback. Could you please recheck?

apps/web/modules/survey/list/components/survey-dropdown-menu.tsx
```diff
@@ -193,7 +235,31 @@ export const SurveyDropDownMenu = ({
...
+                onSelect={(e) => {
+                  e.preventDefault();
+                  void handleRestoreSurvey();
+                }}>
+                {t("workspace.surveys.restore")}
+              </DropdownMenuItem>
+            )}
+            {isArchived && canManageSurvey && (
+              <DropdownMenuItem
+                data-testid="delete-survey-forever"
```
itsjavi, line 251: why are we keeping a "Delete" action for the not archived surveys and then we have "Delete forever" which is identical to the other Delete action? In both it's a "delete forever"

Dhruwang: Yes I think it makes sense to have the delete as well, will update it 👍

itsjavi (approved):
thanks for the changes, looks good

PR #8611, opened 2026-07-22
docs(surveys): refine in-prod targeting docs

jobenjada:
Refines the docs pages around attribute-based targeting, adds page for interaction-based segments


Dhruwang (dismissed):
Thanks for the PR! 🚀

Dhruwang (approved):

PR #8613, opened 2026-07-23
feat: add canonical AuthZed schema deployment

BhagyaAmarasinghe:
## Problem

[ENG-1933](https://linear.app/formbricks/issue/ENG-1933/deploy-and-verify-canonical-spicedb-schema-across-environments) needs an explicit and verifiable way to deploy the checked-in canonical SpiceDB schema. Schema writes must not happen during Formbricks startup, health checks, database migrations, or Helm reconciliation.

Depends on the AuthZed client foundation and canonical schema already merged into `epic/authzed` through ENG-1708 and ENG-1710.

## Solution

- extend the server-only Formbricks AuthZed facade with aggregate `DiffSchema` and explicit `WriteSchema` operations
- compare schema semantics at full consistency and report only digests plus aggregate difference kinds
- add `pnpm authzed:schema check` and guarded `apply` commands
- require the reviewed current digest before replacing any non-empty remote schema
- read back and semantically verify every successful write
- keep schema contents, object identifiers, tokens, metadata, and raw SDK errors out of CLI output and logs
- document Docker and Helm invocation, backup, rollback, and expand/backfill/contract requirements
- extend isolated Docker validation for empty install, idempotency, deliberate drift, recovery, permissions, outage handling, and persistence

## Operational contract

- `check`: read-only; exit `0` when matched and `2` on drift
- `apply`: exit `0` only after verified convergence; an already matching schema returns `unchanged`
- failures: exit `1` with a stable `authzed_*` code
- no HTTP endpoint or browser page
- no startup, readiness, liveness, migration, or Helm hook integration

## Validation

- `pnpm --dir apps/web test lib/authzed` — 80 tests passed
- `pnpm --filter @formbricks/web typecheck` — passed on Node 22
- `pnpm authzed:validate` — 27 relationships, 131 assertions, 2 expected relations
- `bash docker/authzed-compose-contract.sh` — passed
- `pnpm authzed:smoke` — passed, including full-consistency drift detection and persistence after SpiceDB recreation
- focused ESLint, Prettier, shell syntax, and diff checks — passed

## Deliberate exclusions

Relationship projection/backfill, permission enforcement, startup schema writes, public health routes, and new authorization capabilities remain out of scope.

## Sandbox deployment — 2026-07-23

Deployed commit `58ed567a7` to `core-eks/authzed-sandbox` through the public TLS endpoint and verified it directly against each replica.

- backed up the pre-existing two-definition POC schema and its one disposable relationship to local mode-0600 files
- pre-apply semantic check exited `2` with 9 aggregate differences; no schema contents or identifiers were emitted
- deleted exactly the backed-up POC relationship, then verified zero `document` relationships remained
- applied using the reviewed pre-apply digest as the concurrency guard; post-write diff was zero
- public TLS check: `matched`, zero differences
- both existing replicas checked directly over port-forwarded plaintext gRPC: `matched`, zero differences
- recycled one of two replicas and checked the replacement directly: `matched`, zero differences; 2/2 replicas ready
- Formbricks `/health` and `/api/v2/health`: HTTP 200
- retained command-output and SpiceDB-log scans found neither the token nor canonical schema text

The deployed canonical source digest is `sha256:84517c430f065fc2d122545fd650fd1c5314cab0d28cc2cfafd8e0d9c1d4e406`. The private rollback backup remains outside the repository.

### DIFF
--- a/.github/workflows/docker-build-validation.yml
+++ b/.github/workflows/docker-build-validation.yml
@@ -49,6 +49,9 @@ jobs:
       - name: Build web workspace dependencies
         run: pnpm build --filter=@formbricks/web^...
 
+      - name: Test AuthZed client and schema lifecycle
+        run: pnpm --dir apps/web test lib/authzed
+
       - name: Validate Compose contracts
         run: bash docker/authzed-compose-contract.sh
 
--- a/apps/web/lib/authzed/__mocks__/client-dependencies.ts
+++ b/apps/web/lib/authzed/__mocks__/client-dependencies.ts
@@ -3,8 +3,10 @@ import { vi } from "vitest";
 export const sdkMocks = {
   close: vi.fn(),
   deadlineInterceptor: vi.fn((timeoutMs: number) => ({ timeoutMs })),
+  diffSchema: vi.fn(),
   newClient: vi.fn(),
   readSchema: vi.fn(),
+  writeSchema: vi.fn(),
 };
 
 export const configMocks = {
--- a/apps/web/lib/authzed/client.test.ts
+++ b/apps/web/lib/authzed/client.test.ts
@@ -9,8 +9,10 @@ describe("AuthZed client facade", () => {
     closeAuthzedClient();
     sdkMocks.close.mockReset();
     sdkMocks.deadlineInterceptor.mockClear();
+    sdkMocks.diffSchema.mockReset();
     sdkMocks.newClient.mockReset();
     sdkMocks.readSchema.mockReset();
+    sdkMocks.writeSchema.mockReset();
     configMocks.isAuthzedEnabled.mockReset();
     retryMocks.execute.mockClear();
     envMock.AUTHZED_CONSISTENCY = undefined;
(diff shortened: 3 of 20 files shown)


BhagyaAmarasinghe:
Sandbox deployment is complete. The PR description now contains sanitized rollout evidence: guarded digest apply, zero semantic differences through public TLS and each replica, persistence after a one-replica recycle, 2/2 ready replicas, healthy Formbricks endpoints, and passing secret/schema-log scans.

xernobyl (approved):
Nice work — this is clean, and the apply path is properly fail-closed (digest gate → write → semantic read-back verify). I also traced the output path and the redaction holds: only digests, counts, and status are ever printed, never schema text or tokens. Approving. 🚀

A few notes to save you some triage on the CodeRabbit thread:

- **CR #1 (`client.ts`, `promises.diffSchema`)** — false alarm, safe to dismiss. `@authzed/authzed-node` 1.6.1 does expose it (`ZedPromiseClientInterface` includes the schema service), and you're using the right proto fields (`comparisonSchema`, `diffs[]`).
- **CR #6 (smoke-test no-leak)** — lower severity than tagged, I think. The Formbricks-CLI outputs that actually must not leak are captured with `2>&1` and scanned in `application_outputs`; the uncaptured write it points at is the raw `zed schema write` drift *setup*, not the code under test. Worth capturing that one too for completeness, but the guarantee is genuinely exercised.
- **CR #2** (wrap `closeClient` in the `finally` so a cleanup throw can't mask the exit code) and **CR #3** (use `AUTHZED_ERROR_CODES.*` instead of the hardcoded `"authzed_invalid_request"`/`"authzed_internal"` strings) are both worth taking — cheap and real.

One thing not caught elsewhere: the digest precondition is best-effort, not atomic. `apply` reads the remote digest, then calls `WriteSchema`, and SpiceDB has no compare-and-swap on schema writes — so a concurrent writer landing in that window would be clobbered unreviewed. Totally fine for an operator-run, backup-first deploy, but might be worth a one-line note in the README that `apply` assumes no concurrent schema writer.

PR #129, opened 2022-11-03
Arsy

arslene-kabinda:
Responsive of sourcing detail page

PR #8617, opened 2026-07-23
feat(workflows): meter workflow runs like responses [ENG-1936]

Dhruwang:
## What does this PR do?

Part of **[ENG-1936](https://linear.app/formbricks/issue/ENG-1936)** — Workflows is priced **metered like Responses** (Scale plan, 1000 runs/mo included, per-run price ~30% above responses), not a simple on/off flag.

This PR adds the **app-side metering plumbing** for workflow runs, mirroring the responses pattern end-to-end. The boolean entitlement (`getIsWorkflowsEnabled`) stays as the *availability* gate (Scale plan / EE license); this adds *usage billing* on top.

### Decisions (confirmed)
- **Meter on run creation** (when the `WorkflowRun` row is persisted), excluding dry runs — exact parity with response metering.
- **No hard cap** — Stripe bills overage beyond the included volume; the limit is display-only (parity with responses, which has no enforcement).
- **Self-hosted = flag-only** — not metered, gated solely by the boolean `workflows` license feature (ee repo PR #129). Metering is Cloud/Stripe-only.

### Changes (all mirror the responses template)
- `metering.ts` — `recordWorkflowRunCreatedMeterEvent`: posts a `workflow_run_created` Stripe meter event (`identifier: workflow_run_created:<runId>`, `value: "1"`). Cloud-only, never throws (failures are logged, run creation is never blocked), deduped on the stable run id.
- `enqueue-response-completed-runs.ts` — emits the meter event when a run is persisted (before dispatch; persisted = billable), **new runs only** — a unique-constraint replay re-dispatches without re-metering — timestamped from the run's persisted `createdAt`. Meter promises are collected and settled after the dispatch loop (review): no extra round-trip inside the serialized dispatch, but the job doesn't finish until every billed run is reported, so a worker exit can't drop an in-flight event. `stripeCustomerId` threaded in from the pipeline.
- `process-response-pipeline-job.ts` — passes `organization.billing.stripeCustomerId` into the workflow runner.
- `organizations.ts` — `limits.monthly.workflowRuns` (nullable, defaults null so existing billing rows still parse).
- `organization-billing.ts` — resolves `workflowRuns` from the `workflow-runs-included-<n>` Stripe entitlement (`unlimited` → null). An absent entitlement resolves to **null, not the previous value** (review): the key only exists on plans with workflows, so preserving would keep a stale included volume forever after a downgrade.
- entitlements `types.ts` / `cloud-provider.ts` / `self-hosted-provider.ts` — surface `monthlyWorkflowRuns` on the context.

### Tests
`pnpm --filter=@formbricks/web test metering organization-billing cloud-provider self-hosted-provider enqueue-response-completed-runs` → **102 passing**. Covers the new meter event (all branches), the meter call site in the enqueue spec (new run metered exactly once with the run id + persisted `createdAt`; unique-constraint replay never metered), the `workflow-runs-included-` entitlement → `limits.monthly.workflowRuns` including the non-sticky downgrade, and the updated context shape.

## Out of scope (remaining ENG-1936 work)
Metering only bills once the Stripe side exists:
1. **Stripe setup** — create the `workflow_run_created` meter + a metered price on the Scale product (metadata mirroring responses: `formbricks_price_kind: "workflow_runs"`), and a `workflow-runs-included-1000` entitlement feature on Scale.
2. **Catalog provisioning** — extend `stripe-billing-catalog.ts` (`getPriceKind` / `getCatalogItemsForPlan`) to attach the metered workflow-runs line item to the subscription (currently binary base/responses).
3. **Billing UI** — present workflows as metered in `plan-comparison.tsx` (`comparison_row_workflows`, ~L199) and `pricing-table.tsx` (~L591) instead of a boolean checkmark, with overage copy.

## QA / Test Plan

**How to test**

- [ ] Unit suite: `pnpm --filter=@formbricks/web test metering organization-billing cloud-provider self-hosted-provider enqueue-response-completed-runs` → 102 passing.
- [ ] Cloud org on Scale (once the Stripe meter/entitlement from "Out of scope" exists, test mode): submit and finish a response on a survey targeted by an enabled workflow → Stripe Dashboard → the customer's `workflow_run_created` meter shows one event with identifier `workflow_run_created:<runId>` timestamped at the run's creation.
- [ ] Retry/replay the same response through the response pipeline → the existing run is re-dispatched but **no** second meter event appears (replays are never re-metered).
- [ ] Trigger a dry run of a workflow → no meter event.
- [ ] Self-hosted (`IS_FORMBRICKS_CLOUD` unset/false) or an org without a Stripe customer → runs are created and dispatched normally, zero meter events, no errors logged.
- [ ] Simulate a Stripe failure during metering (e.g. invalid key in dev) → log warning `Failed to record Stripe meter event for workflow_run_created`; run creation and dispatch unaffected.
- [ ] Downgrade an org off Scale (workflow-runs entitlement removed in Stripe) → after the next billing sync (Stripe webhook), `organizationBilling.limits.monthly.workflowRuns` is `null`, not the stale `1000`.

**Preconditions / test data**

- Formbricks Cloud env (`IS_FORMBRICKS_CLOUD=true`) with a Stripe test-mode customer on the org; the Stripe-side meter, metered price, and `workflow-runs-included-1000` entitlement are follow-up work — until they land, only the unit suite and the "no meter event" paths are verifiable end-to-end.
- An enabled workflow whose published version has a `response.completed` trigger targeting the test survey.

**Risks & regressions**

- Response pipeline job: the workflow enqueue now settles meter promises before returning — re-check that response side-effects (webhooks, follow-ups, notifications) and job latency are unchanged; meter calls overlap the dispatch loop and the helper never throws.
- Billing sync: only `workflowRuns` switched to null-on-absent; `workspaces`/`responses` must still preserve the previous value on a missing entitlement (guards against bad reads).
- Double billing: the unique-constraint replay path must never meter (pinned by unit test — Stripe's identifier dedup only covers ~24h).

- none (no DB migrations, no env changes; Stripe meter/price/entitlement setup tracked in "Out of scope")

## How to test
- Unit suite above.
- Manual (Cloud, once Stripe setup lands): complete a response that triggers a workflow → a `workflow_run_created` meter event appears on the customer in Stripe; usage accrues against the workflow-runs meter; billing charges overage beyond 1000/mo. Dry runs emit nothing. Self-hosted: no meter events.

Draft — opening for review of the metering approach + the remaining Stripe/UI plan before wiring the catalog.

### DIFF
--- a/apps/web/modules/ee/billing/lib/organization-billing.ts
+++ b/apps/web/modules/ee/billing/lib/organization-billing.ts
@@ -52,6 +52,9 @@ export const getDefaultOrganizationBilling = (): TOrganizationBilling => ({
     workspaces: IS_FORMBRICKS_CLOUD ? 1 : 3,
     monthly: {
       responses: IS_FORMBRICKS_CLOUD ? 250 : 1500,
+      // No included workflow runs by default — the Scale entitlement grants the volume, and
+      // self-hosted gates workflows by the boolean license feature rather than metering.
+      workflowRuns: null,
     },
   },
   stripeCustomerId: null,
@@ -1221,6 +1224,10 @@ const resolveEntitlementDrivenLimits = (
 ) => {
   const workspaceLimitFromEntitlements = parseEntitlementLimit(featureLookupKeys, "workspace-limit-");
   const responsesIncludedFromEntitlements = parseEntitlementLimit(featureLookupKeys, "responses-included-");
+  const workflowRunsIncludedFromEntitlements = parseEntitlementLimit(
+    featureLookupKeys,
+    "workflow-runs-included-"
+  );
 
   const workspacesLimit =
     workspaceLimitFromEntitlements === undefined
@@ -1246,10 +1253,18 @@ const resolveEntitlementDrivenLimits = (
     );
   }
 
+  // Absent workflow-runs entitlement preserves the previous value (or null). No warning: unlike
+  // responses, workflow metering is opt-in per plan, so most subscriptions legitimately have none.
+  const workflowRunsIncludedLimit =
+    workflowRunsIncludedFromEntitlements === undefined
+      ? (previousLimits?.monthly?.workflowRuns ?? null)
+      : workflowRunsIncludedFromEntitlements;
+
   return {
     workspaces: workspacesLimit,
     monthly: {
       responses: responsesIncludedLimit,
+      workflowRuns: workflowRunsIncludedLimit,
     },
   };
 };
--- a/apps/web/modules/ee/billing/lib/organization-billing.ts
+++ b/apps/web/modules/ee/billing/lib/organization-billing.ts
@@ -1253,12 +1253,12 @@ const resolveEntitlementDrivenLimits = (
     );
   }
 
-  // Absent workflow-runs entitlement preserves the previous value (or null). No warning: unlike
-  // responses, workflow metering is opt-in per plan, so most subscriptions legitimately have none.
-  const workflowRunsIncludedLimit =
-    workflowRunsIncludedFromEntitlements === undefined
-      ? (previousLimits?.monthly?.workflowRuns ?? null)
-      : workflowRunsIncludedFromEntitlements;
+  // Absent workflow-runs entitlement resolves to null, NOT the previous value: unlike workspaces/
+  // responses (present on every plan, so absence signals a bad read worth preserving against), this
+  // entitlement only exists on plans with workflows — absence is the normal state, and preserving
+  // would keep a stale included volume forever after a downgrade. A transient bad read self-heals
+  // on the next sync; `unlimited` still parses to null upstream.
+  const workflowRunsIncludedLimit = workflowRunsIncludedFromEntitlements ?? null;
 
   return {
     workspaces: workspacesLimit,
--- a/apps/web/modules/ee/workflows/lib/runner/enqueue-response-completed-runs.ts
+++ b/apps/web/modules/ee/workflows/lib/runner/enqueue-response-completed-runs.ts
@@ -4,6 +4,7 @@ import { PrismaErrorType } from "@formbricks/database/types/error";
 import { logger } from "@formbricks/logger";
 import { type TWorkflowTriggerRunPayload, ZWorkflowTriggerRunPayload } from "@formbricks/workflows";
 import { isDatabasePoolExhaustionError } from "@/lib/jobs/pool-exhaustion";
+import { recordWorkflowRunCreatedMeterEvent } from "@/modules/ee/billing/lib/metering";
 import { getIsWorkflowsEnabled } from "@/modules/ee/license-check/lib/utils";
 import { type DispatchWorkflowRun } from "./dispatch";
 import { markWorkflowRunDispatched } from "./mark-dispatched";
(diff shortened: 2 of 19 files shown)


apps/web/modules/ee/workflows/lib/runner/enqueue-response-completed-runs.ts
```diff
@@ -160,21 +167,43 @@ const createAndDispatchWorkflowRun = async ({
...
+  // (ENG-1936). A replayed enqueue that re-found an existing run is skipped so it can't double-bill
+  // past Stripe's ~24h identifier-dedup window. These runs are always real (isDryRun: false); dry
+  // runs are metered nowhere. Timestamped from the run's own createdAt so usage lands in the correct
+  // billing period (response.updatedAt could be stale/older than Stripe's 35-day accept window).
+  //
+  // Cloud-only and fire-and-forget: the helper swallows its own errors, so metering never blocks or
+  // fails run creation. Deliberately NOT awaited — awaiting would add a Stripe round-trip to (and, on
+  // a Stripe hang, stall) the serialized per-match dispatch loop below. The helper owns its errors,
+  // so voiding cannot surface an unhandled rejection.
+  if (isNew) {
```
itsjavi, line 203: this part isn't covered by any test: enqueue-response-completed-runs.test.ts doesn't reference the metering module, and the metering.test.ts branches only exercise the helper, not this call site.

Please pin it in the enqueue spec: mock @/modules/ee/billing/lib/metering and assert:
- (1) a newly created run calls recordWorkflowRunCreatedMeterEvent exactly once with the run id and its persisted createdAt
- (2) the unique-constraint replay path (create throws, findUnique returns the existing run) does not call it. 

Both prisma paths are already mocked in that spec, so this should be a small addition.

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -1246,10 +1253,18 @@ const resolveEntitlementDrivenLimits = (
     );
   }
 
+  // Absent workflow-runs entitlement preserves the previous value (or null). No warning: unlike
+  // responses, workflow metering is opt-in per plan, so most subscriptions legitimately have none.
+  const workflowRunsIncludedLimit =
```
itsjavi: Preserving the previous value when the workflow-runs entitlement is absent makes downgrades sticky: the entitlement only exists on Scale, so after a Scale -> Pro/Hobby downgrade every future sync keeps limits.monthly.workflowRuns = 1000 forever: nothing ever overwrites it. 

For workspaces/responses the preserve-on-absent fallback guards against a bad entitlements read because every plan carries those keys; that protection doesn't apply here since absence is the normal state. 

Suggestion:

```
const workflowRunsIncludedLimit = workflowRunsIncludedFromEntitlements ?? null;
```

A transient bad read still self-heals on the next sync, and unlimited (null) is preserved correctly. Impact today is none (no consumers of the field yet), but the planned billing UI would render the stale included volume on a plan without workflows.

apps/web/modules/ee/workflows/lib/runner/enqueue-response-completed-runs.ts
```diff
@@ -160,21 +167,43 @@ const createAndDispatchWorkflowRun = async ({
...
+    void recordWorkflowRunCreatedMeterEvent({
```
itsjavi: Optional: consider awaiting this like the response meter event (process-response-pipeline-job.ts awaits recordResponseCreatedMeterEvent in this same job). 

The helper never throws, so awaiting can't fail run creation, and the Stripe-hang concern in the comment applies equally to the already-awaited response call. 

The difference with void is that an in-flight request is dropped if the worker exits right after the job completes; and because isNew is then already consumed, that run is permanently unbilled. 

If the extra round-trip inside the serialized dispatch loop is the worry, collect the promises and await them once after the loop.

itsjavi (changes requested):
thank you, I left some remarks

itsjavi (approved):

PR #8618, opened 2026-07-23
fix: hard-gate contacts page PII reads on the contacts entitlement (#ENG-1768)

mattinannt:
## What does this PR do?

`ContactsPage` computed `isContactsEnabled` only to toggle the upsell UI (via `ContactsPageLayout`), but unconditionally called `getContactAttributeKeys`/`getContacts` and passed the results down to `ContactDataView`.

This meant an organization whose `contacts` entitlement had lapsed — but which still had historical contacts in the database — could still have its stored contact PII read from the database on every page render, even though the UI itself was hidden behind an upsell prompt.

This PR hard-gates the page-level data reads on the entitlement itself, not just the CTA/UI: `getContactAttributeKeys` and `getContacts` are now only called when `isContactsEnabled` is `true`; otherwise the page passes empty arrays through (which are never rendered anyway, since `ContactsPageLayout` already swaps in the upgrade prompt when the entitlement is disabled).

Fixes [ENG-1768](https://linear.app/formbricks/issue/ENG-1768/i-34b-contacts-page-readsrenders-pii-even-when-the-contacts)

## How should this be tested?

- As an org without the `contacts` entitlement (or with a lapsed entitlement) that still has contacts stored in the DB, visit `/workspaces/[workspaceId]/contacts`: the upgrade prompt should show, and no contact PII should be fetched from the database (previously fetched unconditionally).
- As an org with the `contacts` entitlement enabled, visit the same page and confirm contacts/attributes still load and render normally (no behavior change).
- `pnpm test`, `pnpm lint`, and `pnpm build` all pass.

### Required

- [x] Ran `pnpm build`

### Appreciated

---
_Generated by [Claude Code](https://claude.ai/code/session_013u4Qvtj7AqFAciz8i3yQVd)_

### DIFF
--- a/apps/web/modules/ee/contacts/actions.ts
+++ b/apps/web/modules/ee/contacts/actions.ts
@@ -4,7 +4,7 @@ import { z } from "zod";
 import { prisma } from "@formbricks/database";
 import { ZId } from "@formbricks/types/common";
 import { ZContactAttributesInput } from "@formbricks/types/contact-attribute";
-import { ResourceNotFoundError } from "@formbricks/types/errors";
+import { OperationNotAllowedError, ResourceNotFoundError } from "@formbricks/types/errors";
 import { capturePostHogEvent } from "@/lib/posthog";
 import { authenticatedActionClient } from "@/lib/utils/action-client";
 import { checkAuthorizationUpdated } from "@/lib/utils/action-client/action-client-middleware";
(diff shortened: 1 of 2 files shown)


Dhruwang (changes requested):
This looks good but I found one more place that we can do the same. Can you check on that once? 😊 

getContactsAction returns getContacts(...) → attributes.value = the same emails/names/phones the page now hides. ContactDataView calls it client-side for pagination + search (contact-data-view.tsx:76,121).

mattinannt:
Good catch — fixed in 7a4018d. `getContactsAction` (used by `ContactDataView` for client-side pagination/search at `contact-data-view.tsx:76,121`) now also checks `getIsContactsEnabled` and throws `OperationNotAllowedError` if the entitlement is disabled, matching the pattern already used in `segments/actions.ts`. Previously it only checked workspace access, so it could still be invoked directly to read contact PII even with the page-level gate in place.

All checks (lint, unit tests, build) pass locally with the change.


---

Dhruwang (approved):
Thank you for the changes, looks good 🚀

PR #8619, opened 2026-07-23
perf(e2e): small CI cleanup + measured findings on the real bottlenecks

mattinannt:
## What does this PR do?

Speeds up the Playwright E2E CI job **a little**, and — more valuably — **measures in live CI where the ~16 min actually goes**, ruling out several plausible-but-wrong optimizations with data. Scope was deliberately trimmed as measurements came in: everything that didn't actually help was reverted/removed.

### Kept (small, safe wins)
- **pnpm store cache** via `setup-node` (`cache: "pnpm"`) — install ~30 s.
- **Playwright browser cache** (`~/.cache/ms-playwright`) — skips the ~100 MB Chromium download on re-runs.
- **Parallel browser install** — `playwright install --with-deps` runs concurrently with `pnpm build` instead of after it.
- **Removed inert `TURBO_TOKEN`/`TURBO_TEAM` env** from the e2e job (no Turbo remote-cache backend is configured, so they did nothing).

### Tried and removed/reverted (measured, no benefit)
| Change | Measurement | Outcome |
| --- | --- | --- |
| e2e build cache (restore `.next`/`.turbo` then rebuild) | build ~4 min cold == warm | removed |
| Cache Next.js `.next/cache` | warm build 4m17s vs cold 3m46s | removed |
| `--workers=20` (Azure service) | 6m43s vs 5m47s @ 10 | reverted to 10 |
| Remove `slowMo: 150` from `survey.spec.ts` | 3 tests flaked (DOM-detach races) | reverted (slowMo is load-bearing) |

### Measured breakdown of the E2E job (~14m52s after this PR)
- **Build ~4 min** — not cacheable here without a **Turbo remote-cache backend**.
- **Prisma migrate ~3 min** — slow due to Postgres checkpoint I/O on the runner (being addressed separately).
- **Playwright tests ~5m47s** — on the Azure Playwright service at `--workers=10`.

The job is **setup-bound, not test-bound**, so test sharding would not help. The real levers are all outside a CI-config change: a Turbo cache backend (build), the Prisma migration fix, and aligning the Azure Playwright workspace region with the GitHub runners (tests + flakiness).

## How should this be tested?

- CI is the test: the E2E job is green and ~14m52s (vs ~16 min baseline); the test phase is unchanged at 5m47s.
- Locally: `pnpm test:e2e` with the app on `:3000` still passes (no behavioral changes to specs; `slowMo` is unchanged from `main`).

### Required

### Notes for reviewers

- No product/runtime code changes — CI workflow (`e2e.yml`), `playwright.config.ts` (`workers` for local only), and the `test-e2e:azure` script.
- Follow-up cleanup (separate PR): the same inert `TURBO_TOKEN`/`TURBO_TEAM` refs still exist in `build-web.yml`, `integration-tests.yml`, `docker-build-validation.yml`, and the `cache-build-web` action.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -12,10 +12,6 @@ on:
       # Add other secrets if necessary
   workflow_dispatch:
 
-# NOTE: no TURBO_TOKEN/TURBO_TEAM env — Turbo remote cache is not configured
-# (no backend), so these were inert. Re-add them if a Turbo cache backend is
-# set up later.
-
 permissions:
   contents: read
   actions: read
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -125,13 +125,22 @@ jobs:
             minio/mc@sha256:95b5f3f7969a5c5a9f3a700ba72d5c84172819e13385aaf916e237cf111ab868 \
             /tmp/rustfs-init.sh
 
-      # NOTE: no local build cache here. Without Turbo remote cache (not
-      # configured — no Vercel plan), restoring .next/.turbo does not let
-      # `pnpm build` short-circuit; it recompiles cold either way, so caching
-      # the build only added save/restore overhead. Measured cold==warm at
-      # ~4min. If a build cache is wanted later, reuse build-web.yml's artifact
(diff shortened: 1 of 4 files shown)


xernobyl (approved):
LGTM

PR #8621, opened 2026-07-23
fix(mcp): enforce workflows:write scope on MCP workflow write tools

xernobyl:
## What does this PR do?

Security fix from the workflows-epic QA ([ENG-1967](https://linear.app/formbricks/issue/ENG-1967)). **Stacked on `epic/workflows-v1`** (the workflow MCP tools don't exist on `main`), so please review/merge against that base.

The MCP workflow write tools mutated with no MCP-layer OAuth scope check — `workflows.ts` never called `guardMcpScopes`, so all 8 mutations (`create`/`patch`/`duplicate`/`delete`/`enable`/`disable`/`archive`/`unarchive`) leaned only on the downstream v3 workspace-permission check. A read-scoped OAuth token held by a write-capable user could therefore still write via MCP — the token exceeded its granted scope. This is the **token-scope** boundary (defense-in-depth), not user-privilege escalation: workspace authz still blocks a read-only *user*. Survey write tools already did this correctly (`surveys:write`); this brings workflow tools to parity and makes the guarantee structural.

What changed:

- **`workflows:read` / `workflows:write` are now first-class OAuth scopes** alongside `surveys:*` — advertised in the challenge, the protected-resource metadata (PRM), the DCR default scopes, and the consent screen, with the same 15m step-up expiry on `:write`. Without this, a token could never actually carry a workflow scope. `surveys:*` is unchanged. (Migration note: OAuth tokens issued before this change lack `workflows:*`, so QA clients must re-authorize to pick up the new DCR defaults — acceptable on an unreleased epic.)
- **`registerScopedTool`** — a single registration path where the required scope is a *mandatory* argument and the gate runs before the handler. A tool literally can't be registered without declaring a scope, so a new tool inherits enforcement by construction (this is how the gap happened — a per-tool `guardMcpScopes` call is easy to forget). All survey, workflow, and workspace tools are migrated onto it; the survey/workspace migration is refactor-only (no behavior change) but closes the "forgot the guard" class for good.
- Read tools require `<resource>:read`, mutating tools `<resource>:write`; a token missing the scope gets `403 insufficient_scope` before any mutation or audit log runs.
- Docs updated (overview, setup, technical handbook) to list the workflow scopes and state the rule.

The v3 workspace-authorization layer is untouched — this only adds the token-scope gate in front of it, so a `:write` call now has to satisfy both.

## How should this be tested?

Automated (all green locally):

```bash
pnpm --filter=@formbricks/web exec vitest run \
  modules/mcp/tools/workflows.test.ts \
  modules/mcp/tools/surveys.test.ts \
  modules/mcp/tools/workspaces.test.ts \
  modules/mcp/tools/guard-scopes.test.ts \
  modules/mcp/auth.test.ts \
  modules/auth/lib/oauth-client-metadata.test.ts \
  modules/auth/lib/mcp-oauth-dcr.test.ts \
  app/.well-known/oauth-protected-resource/route.test.ts \
  app/api/mcp/route.test.ts
```

Coverage on the new code: `guard-scopes.ts` 100%, `workflows.ts` ~93% (the uncovered lines are pre-existing audit-log branches, not the scope gate). New/updated cases assert the allowed **and** denied paths for read-only vs write-scoped tokens, at both the tool layer and the JSON-RPC route layer.

Manual (MCP, API-key path, like the ENG-1776 smoke test):

1. Connect with an API key that has only `read` on the workspace → its token carries `workflows:read` but not `workflows:write`. `create_workflow` / `delete_workflow` return `403` (`OAuth token does not include the required MCP scope`); `list_workflows` succeeds.
2. Connect with a `write`/`manage` key → workflow write tools succeed.
3. `GET /.well-known/oauth-protected-resource/api/mcp` and `tools/list` advertise `workflows:read` + `workflows:write`.

### Required

- [ ] Ran `pnpm build` (ran targeted typecheck + lint + the affected Vitest suites instead)
- [ ] Merged the latest changes from main onto my branch with `git pull origin main` (stacked on `epic/workflows-v1`, rebased on its latest head)
- [x] My changes don't cause any responsiveness issues (backend + docs only, no UI)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (no UI change)
- [x] Updated the Formbricks Docs if changes were necessary

### DIFF
--- a/apps/web/app/api/mcp/route.test.ts
+++ b/apps/web/app/api/mcp/route.test.ts
@@ -43,7 +43,7 @@ vi.mock("@formbricks/database", () => ({
 }));
 
 vi.mock("@/modules/auth/lib/oauth-urls", () => ({
-  MCP_RESOURCE_SCOPES: ["surveys:read", "surveys:write"],
+  MCP_RESOURCE_SCOPES: ["surveys:read", "surveys:write", "workflows:read", "workflows:write"],
   getAuthIssuerUrl: () => "http://localhost/api/auth",
   getMcpOrigin: () => "http://localhost",
   getMcpProtectedResourceMetadataUrl: () => "http://localhost/.well-known/oauth-protected-resource/api/mcp",
@@ -161,7 +161,7 @@ describe("POST /api/mcp", () => {
     expect(response.status).toBe(401);
     expect(response.headers.get("Content-Type")).toBe("application/problem+json");
     expect(response.headers.get("WWW-Authenticate")).toBe(
-      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read surveys:write"'
+      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read surveys:write workflows:read workflows:write"'
     );
     expect(applyIPRateLimit).toHaveBeenCalled();
   });
@@ -425,7 +425,7 @@ describe("POST /api/mcp", () => {
     expect(authenticateApiKeyFromHeaders).not.toHaveBeenCalled();
     expect(applyIPRateLimit).toHaveBeenCalled();
     expect(response.headers.get("WWW-Authenticate")).toBe(
-      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read surveys:write"'
+      'Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp" scope="surveys:read surveys:write workflows:read workflows:write"'
     );
   });
 
@@ -467,6 +467,49 @@ describe("POST /api/mcp", () => {
       code: "forbidden",
       detail: "OAuth token does not include the required MCP scope",
       requestId: "req_read_only",
+    });
+  });
+
+  test("blocks workflow write tools for tokens without workflows:write", async () => {
+    // A write-capable user whose OAuth token was only granted read scopes (surveys:read + workflows:read)
+    // must not be able to reach a workflow mutation — the ENG-1967 token-scope boundary.
+    verifyAccessTokenMock.mockResolvedValueOnce({
+      sub: "user_1",
+      email: "person@example.com",
+      scope: "openid profile email surveys:read workflows:read",
+      exp: Math.floor(Date.now() / 1000) + 900,
+      azp: "client_wf_read_only",
+    });
+
+    const response = await POST(
+      createMcpRequest(
+        {
+          jsonrpc: "2.0",
+          id: 10,
+          method: "tools/call",
+          params: {
+            name: "delete_workflow",
+            arguments: {
+              workflowId: "wf1234567890123456789012ab",
+            },
+          },
+        },
+        {
+          authorization: "Bearer eyJhbGciOiJFZERTQSJ9.wfreadonly.signature",
+          "x-api-key": "",
+          "x-request-id": "req_wf_read_only",
+        }
+      )
+    );
+
+    expect(response.status).toBe(200);
+    const message = await readMcpResponse(response);
+    expect(message.result.isError).toBe(true);
+    expect(message.result.structuredContent.error).toMatchObject({
+      status: 403,
+      code: "forbidden",
+      detail: "OAuth token does not include the required MCP scope",
+      requestId: "req_wf_read_only",
     });
   });
 
--- a/apps/web/app/api/mcp/route.test.ts
+++ b/apps/web/app/api/mcp/route.test.ts
@@ -1,6 +1,7 @@
 import { NextRequest } from "next/server";
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { ApiKeyPermission } from "@formbricks/database/prisma";
+import { buildV3AuditLog, queueV3AuditLog } from "@/app/api/v3/lib/audit";
 import {
   createdResponse,
   problemBadRequest,
@@ -511,6 +512,10 @@ describe("POST /api/mcp", () => {
       detail: "OAuth token does not include the required MCP scope",
       requestId: "req_wf_read_only",
     });
+    // The scope gate must fire BEFORE any mutation side effect: no audit log is built or queued for a
+    // request that never reaches the workflow handler.
+    expect(buildV3AuditLog).not.toHaveBeenCalled();
+    expect(queueV3AuditLog).not.toHaveBeenCalled();
   });
 
   test("calls create_survey through the MCP route", async () => {
--- a/apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
+++ b/apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
@@ -20,26 +20,19 @@ export const getMcpOauthProviderOptions = (): TOauthProviderOptions => ({
   validAudiences: [getMcpResourceUrl()],
   allowDynamicClientRegistration: true,
   allowUnauthenticatedClientRegistration: true,
-  // Register MCP clients with read + write by default so the consent screen offers write and the
-  // write tools are reachable (clients derive their DCR/authorize scopes from what we advertise, and
-  // the plugin validates authorize against the client's registered scopes). Granting the write scope
-  // is safe: actual write access is still enforced downstream by the user's workspace permissions.
-  clientRegistrationDefaultScopes: [
-    "openid",
-    "profile",
-    "email",
-    "offline_access",
-    "surveys:read",
-    "surveys:write",
-    "workflows:read",
-    "workflows:write",
-  ],
+  // Register MCP clients with the full advertised scope set by default so the consent screen offers
+  // write and the write tools are reachable (clients derive their DCR/authorize scopes from what we
+  // advertise, and the plugin validates authorize against the client's registered scopes). Granting
+  // write is safe: actual write access is still enforced downstream by the user's workspace
+  // permissions. Spread the single source of truth so the defaults can't drift from MCP_OAUTH_SCOPES.
+  clientRegistrationDefaultScopes: [...MCP_OAUTH_SCOPES],
   accessTokenExpiresIn: 15 * 60,
   refreshTokenExpiresIn: 30 * 24 * 60 * 60,
-  scopeExpirations: {
-    "surveys:write": "15m",
-    "workflows:write": "15m",
-  },
+  // Every write scope gets the 15-minute step-up expiry, derived from the scope list so a new
+  // `<resource>:write` scope inherits it automatically (no separate hand-edit to keep in sync).
+  scopeExpirations: Object.fromEntries(
+    MCP_OAUTH_SCOPES.filter((scope) => scope.endsWith(":write")).map((scope) => [scope, "15m"])
+  ),
   // Store opaque access-token and refresh-token lookup values as hashes. JWT access tokens are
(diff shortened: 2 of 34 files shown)


apps/web/modules/mcp/tools/workflows.test.ts
```diff
@@ -414,3 +426,85 @@ describe("registerWorkflowTools", () => {
...
+  });
+
+  const INSUFFICIENT_SCOPE = {
+    status: 403,
+    code: "forbidden",
+    detail: "OAuth token does not include the required MCP scope",
+  };
+
+  // Every mutating tool + a representative valid input for it.
+  const writeTools: { name: string; handlerKey: keyof typeof workflowsHandlers; input: unknown }[] = [
```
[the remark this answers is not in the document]
xernobyl: Skipping this one — it's a false positive. `writeTools` is declared exactly **once** (line 444); there are no duplicate declarations. The file compiles and its suite passes (`workflows.test.ts`, 29 tests green), which wouldn't be the case with a duplicate block-scoped variable. Looks like the suggestion diff rendered the single anchor line three times. No change made.

apps/web/app/api/mcp/route.test.ts
```diff
@@ -470,6 +470,49 @@ describe("POST /api/mcp", () => {
...
+    const message = await readMcpResponse(response);
+    expect(message.result.isError).toBe(true);
+    expect(message.result.structuredContent.error).toMatchObject({
+      status: 403,
+      code: "forbidden",
+      detail: "OAuth token does not include the required MCP scope",
+      requestId: "req_wf_read_only",
+    });
+  });
+
```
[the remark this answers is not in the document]
xernobyl: Done in ffe1a9ded — the test now asserts `buildV3AuditLog` and `queueV3AuditLog` were **not** called on the denied request, proving the scope gate fires before any mutation or audit side effect.

apps/web/modules/mcp/tools/guard-scopes.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    string,
+    { config: Record<string, unknown>; handler: (input: unknown, extra: unknown) => Promise<CallToolResult> }
+  >();
+  const server = {
+    registerTool: vi.fn((name: string, config: Record<string, unknown>, handler: any) => {
+      tools.set(name, { config, handler });
+    }),
+  };
+  return { server, tools };
+}
```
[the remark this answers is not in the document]
xernobyl: Addressed in ffe1a9ded — narrowed the `structuredContent` access to a typed `ScopeErrorContent` instead of `any`. I kept the `registerTool`/handler test-double as `any` deliberately: it mirrors the existing `createToolServer` convention in `surveys.test.ts` and `workflows.test.ts`, and precisely typing the SDK's generic `ToolCallback` for a fake adds noise without real safety. Happy to revisit if we want a shared typed harness across all three.

apps/web/modules/mcp/tools/guard-scopes.ts
```diff
@@ -21,3 +23,39 @@ export async function guardMcpScopes(
...
+ * mutating tools pass `["<resource>:write"]`. Enforcing scope structurally (vs. a per-tool
+ * `guardMcpScopes` call that's easy to forget) is what prevents the ENG-1967 class of gap from
+ * recurring as new tools are added.
+ */
+export function registerScopedTool<InputArgs extends ZodRawShape>(
+  server: McpServer,
+  name: string,
+  config: ScopedToolConfig<InputArgs>,
+  requiredScopes: string[],
+  handler: ToolCallback<InputArgs>
```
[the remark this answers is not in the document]
xernobyl: Done in ffe1a9ded — `requiredScopes` is now `[string, ...string[]]`, so a tool can't be registered gating on `[]`. All call sites pass non-empty literals, so no call-site changes were needed.

apps/web/modules/mcp/tools/surveys.ts
```diff
@@ -61,8 +62,52 @@ export function buildListSurveysSearchParams(input: TMcpListSurveysInput): URLSe
...
+    return await responseToMcpToolResult(response, requestId);
+  } catch (error) {
+    if (auditLog) {
+      auditLog.eventId = requestId;
+      await queueV3AuditLog(auditLog, requestId, log);
+    }
+
+    throw error;
+  }
+}
```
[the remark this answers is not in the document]
xernobyl: Done in ffe1a9ded — extracted `runMcpMutation` (`run-mcp-mutation.ts`). The survey tools call it directly (the local `runSurveyMutation` is gone) and `runWorkflowMutation` delegates to it, keeping only its context-building. The try/catch/status/queue audit lifecycle now lives in one place.

docs/platform/mcp/overview.mdx
```diff
@@ -36,16 +36,20 @@ The whole flow uses Authorization Code + PKCE, the standard the [MCP authorizati
...
+The MCP server exposes survey and workflow tools, grouped into scopes you approve on the consent
+screen. Read tools require the matching `:read` scope; tools that create, update, or delete require
+the `:write` scope:
 
 | Scope | Tools | What it allows |
 | --- | --- | --- |
 | `surveys:read` | `list_surveys`, `get_survey`, `validate_survey` | Read surveys and validate survey documents |
 | `surveys:write` | `create_survey`, `patch_survey`, `delete_survey` | Create, update, and delete surveys |
+| `workflows:read` | `list_workflows`, `get_workflow`, `list_workflow_runs`, `get_workflow_run`, `test_workflow` | Read workflows and their runs |
+| `workflows:write` | `create_workflow`, `patch_workflow`, `duplicate_workflow`, `delete_workflow`, `enable_workflow`, `disable_workflow`, `archive_workflow`, `unarchive_workflow` | Create, update, and delete workflows |
```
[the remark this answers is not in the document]
xernobyl: Done in ffe1a9ded — verified `testWorkflow` authorizes at `readWrite` (workflows.handlers.ts), and documented the exception in `overview.mdx` (footnote on the `workflows:read` row) and `setup.mdx` troubleshooting: `test_workflow` needs workspace write/manage despite its read scope.

apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
```diff
@@ -31,11 +31,14 @@ export const getMcpOauthProviderOptions = (): TOauthProviderOptions => ({
     "offline_access",
     "surveys:read",
     "surveys:write",
+    "workflows:read",
+    "workflows:write",
```
Dhruwang: This is `MCP_OAUTH_SCOPES` spelled out a second time — `scopes` and `scopes_supported` above already spread the constant. Spread it here too (`[...MCP_OAUTH_SCOPES]`) so the DCR defaults can't drift out of sync.

apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
```diff
@@ -31,11 +31,14 @@ export const getMcpOauthProviderOptions = (): TOauthProviderOptions => ({
...
   ],
   accessTokenExpiresIn: 15 * 60,
   refreshTokenExpiresIn: 30 * 24 * 60 * 60,
   scopeExpirations: {
     "surveys:write": "15m",
+    "workflows:write": "15m",
```
Dhruwang: Adding a scope means hand-editing this, the two scope lists, and the consent label switch, with nothing linking them. Could derive the `:write` expirations from the scope list so a new write scope picks up the 15m step-up automatically.

apps/web/modules/mcp/tools/guard-scopes.ts
```diff
@@ -21,3 +23,40 @@ export async function guardMcpScopes(
...
+  config: ScopedToolConfig<InputArgs>,
+  // Non-empty tuple: a tool cannot be registered with `[]`, which would gate on nothing.
+  requiredScopes: [string, ...string[]],
+  handler: ToolCallback<InputArgs>
+): void {
+  const guardedHandler = (async (input: unknown, extra: { authInfo?: AuthInfo }) => {
+    const requestId = getMcpRequestId(extra.authInfo);
+    const scopeError = await guardMcpScopes(extra.authInfo, requiredScopes, requestId);
+    if (scopeError) {
+      return scopeError;
```
Dhruwang, line 56: This cast erases `InputArgs` — a handler whose input type drifts would still compile clean. If the SDK type forces the cast, a one-line note on why it's safe would help.

apps/web/modules/mcp/tools/workspaces.ts
```diff
@@ -21,14 +23,9 @@ export function registerWorkspaceTools(server: McpServer): void {
         openWorldHint: true,
       },
     },
+    ["surveys:read"],
```
Dhruwang, line 31: workflows:read is advertised as a standalone scope now, but a token holding only workflows:read can't get past here (needs surveys:read) — same with the outer surveys:read gate in auth.ts. So a workflows-only token can't reach any workflow tool. Either accept workflows:read here too, or document that workflows:* has to be granted alongside surveys:read.

Dhruwang (changes requested):

apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
xernobyl: Done in 2a8ae68f0 — `clientRegistrationDefaultScopes: [...MCP_OAUTH_SCOPES]` now, so the DCR defaults track the single source of truth and can't drift.

apps/web/modules/auth/lib/mcp-oauth-provider-options.ts
xernobyl: Done in 2a8ae68f0 — derived from the scope list: `Object.fromEntries(MCP_OAUTH_SCOPES.filter((s) => s.endsWith(":write")).map((s) => [s, "15m"]))`, so a new `<resource>:write` scope inherits the 15m step-up automatically. (The two scope lists + the consent-label switch still take a per-scope edit, but the expiry no longer does — one fewer place to sync.)

apps/web/modules/mcp/tools/guard-scopes.ts
xernobyl: Added a note in 2a8ae68f0. The cast is needed only because `ToolCallback<InputArgs>` is an overloaded/conditional signature TS won't let us call with the erased `unknown` params. It's safe: the SDK validates `input` against this tool's `inputSchema` (InputArgs) before invoking `guardedHandler`, and we forward the same `input`/`extra` through unchanged — the runtime value already conforms; nothing is reshaped.

apps/web/modules/mcp/tools/workspaces.ts
xernobyl: Went with the "document" option (2a8ae68f0). The governing gate is actually the auth.ts baseline — `DEFAULT_OAUTH_SCOPE = "surveys:read"` (auth.ts:369) rejects any token lacking `surveys:read` *before* it reaches `list_workspaces`, so a workflows-only OAuth token can't authenticate at all. In practice every token carries `surveys:read` (API keys hardcode `surveys:read`+`workflows:read`; DCR clients default to the full advertised set), so `workflows:*` is always additive. I documented that contract in the `list_workspaces` comment + the handbook. True workflows-only support would mean relaxing the baseline to "any resource read scope" — happy to file that as a follow-up if we want to support narrowed tokens, but it felt out of scope for this write-enforcement PR.

Dhruwang (approved):

PR #8622, opened 2026-07-23
fix: stop echoing raw error.message in v1 API 500 responses

xernobyl:
## What does this PR do?

Several v1 API handlers passed a raw `error.message` straight into their 500 response on the unexpected-error path:

```ts
responses.internalServerErrorResponse(error instanceof Error ? error.message : "Unknown error occurred")
```

When the thrown error was a `DatabaseError` (which wraps the underlying Prisma message — schema/column/constraint detail, sometimes query fragments) or any other unexpected `Error`, those internals were returned to the client. Some of the affected endpoints (e.g. `POST /api/v1/client/[workspaceId]/responses`) are public/unauthenticated, so this was a real information-disclosure hazard (OWASP: never return internals in an API response). Pre-existing debt, surfaced during the ENG-1660 Prisma-error review.

The fix is a single shared error boundary that v1 handlers route through:

- **`handleApiError(error, { cors })`** (`apps/web/app/lib/api/handle-api-error.ts`) maps a caught error to the `{ response, error }` shape `withV1ApiWrapper` already expects. Expected business errors with a **4xx** status keep their (domain-authored, safe) message; everything **5xx**/unknown returns a fixed generic message — `"Something went wrong. Please try again."` — while the *real* error is threaded back so the wrapper's `reportApiError` still logs it (and reports to Sentry on 5xx). The gate is `isExpectedError(error) && statusCode < 500`, so a `QueryExecutionError` (an "expected" name but a 500) is genericized too.
- This mirrors what the **v2** responses route already does — v1 was the straggler.
- `auth.ts`'s shared `handleErrorResponse` now delegates its default branch to the same boundary, which closes the leak for every route that uses it (`surveys/[surveyId]`, `action-classes/[actionClassId]`, `responses/[responseId]`, `singleUseIds`) in one place.

While consolidating I found one more leak the ticket's line-grep missed — a **multi-line** `internalServerErrorResponse(error.message)` in `management/responses` — now closed. A multi-line sweep confirms no `internalServerErrorResponse(…error.message…)` remains anywhere in `apps/web`.

Intentional 4xx validation/domain messages (e.g. `InvalidInputError → 400` with its message) are preserved on purpose — this is not a blanket genericize.

Linear: https://linear.app/formbricks/issue/ENG-1808

## Before / After

Public response-submission endpoint hitting an unexpected DB error (illustrative):

**Before** — raw Prisma internals returned to an unauthenticated client:

```http
POST /api/v1/client/{workspaceId}/responses  →  500
```
```json
{
  "code": "internal_server_error",
  "message": "Invalid `prisma.response.create()` invocation: Foreign key constraint failed on the field: `Response_surveyId_fkey (index)`",
  "details": {}
}
```

**After** — generic message to the client; the real error is logged + sent to Sentry server-side:

```http
POST /api/v1/client/{workspaceId}/responses  →  500
```
```json
{
  "code": "internal_server_error",
  "message": "Something went wrong. Please try again.",
  "details": {}
}
```

Expected client errors are unchanged — the domain message is still surfaced (no over-genericizing):

```http
POST /api/v1/webhooks   (invalid body)  →  400
```
```json
{
  "code": "bad_request",
  "message": "Fields are missing or incorrectly formatted",
  "details": { "...": "..." }
}
```

## QA / Test Plan

**How to test**

- [ ] Unit: `pnpm --filter=@formbricks/web test "app/lib/api/handle-api-error.test.ts"` → passes; asserts every 5xx/`DatabaseError`/unknown case returns the generic message (never the input message) and threads the real error, and that expected 4xx errors keep their status + message.
- [ ] Unit: `pnpm --filter=@formbricks/web test "app/api/v1/client/[workspaceId]/responses/[responseId]/lib/put-response-handler.test.ts"` → the DB-error cases now assert the generic body, not the raw message.
- [ ] Happy path: normal v1 requests still work unchanged — e.g. `GET /api/v1/webhooks` with a valid API key (200), `POST /api/v1/client/{workspaceId}/responses` with a valid payload (201).
- [ ] Leak check: force an unexpected server/DB error on a v1 endpoint (e.g. temporarily break a query) → the response body is exactly `{"code":"internal_server_error","message":"Something went wrong. Please try again.","details":{}}`, and the real error shows up in the server log / Sentry (5xx) — not in the HTTP body.
- [ ] Edge: an expected validation error (malformed webhook / survey / response input) still returns its **4xx** with the domain message.
- [ ] Regression sweep: `grep -rnE "internalServerErrorResponse\([^)]*error\.message" apps/web/app/api/v1` returns nothing.

**Preconditions / test data**

- A management API key (for `/api/v1/management/*` and `/api/v1/webhooks`) and a published link survey + workspace (for `/api/v1/client/*`). Works on Cloud or self-host; no feature flags.

**Risks & regressions**

- **Status change:** `GET /api/v1/management/surveys`, `GET /api/v1/management/responses`, and `GET /api/v1/management/action-classes` now return a generic **500** on a DB failure instead of a **400** carrying the raw message (a DB failure is a server error, so 500 is the correct status). Same for the `management/responses` create outer catch.
- **Headers:** a few client-route error responses now carry CORS headers they inconsistently lacked (additive).
- **Observability:** DB errors on webhooks / management-surveys / management-responses are now reported to Sentry on 5xx — they previously returned without threading the error, so nothing was reported. Expect more signal there, not less.
- No change to success responses or to intended 4xx messages.

- none

### Required

- [x] Filled out the "QA / Test Plan" section in this PR (behavior, edge cases, preconditions, risks, migrations/env)
- [ ] Ran `pnpm build`
- [x] Merged the latest changes from main onto my branch (branch cut from current `main`)

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR (no UI changes)
- [ ] Updated the Formbricks Docs if changes were necessary (no doc-facing changes)

### DIFF
--- a/apps/web/app/api/v1/auth.ts
+++ b/apps/web/app/api/v1/auth.ts
@@ -1,6 +1,6 @@
 import { NextRequest } from "next/server";
 import { TAuthenticationApiKey } from "@formbricks/types/auth";
-import { handleApiError } from "@/app/lib/api/handle-api-error";
+import { type ApiErrorResult, handleApiError } from "@/app/lib/api/handle-api-error";
 import { responses } from "@/app/lib/api/response";
 import {
   type AuthenticateApiKeyOptions,
@@ -14,15 +14,17 @@ export const authenticateRequest = async (
   return await authenticateApiKeyFromHeaders(request.headers, options);
 };
 
-export const handleErrorResponse = (error: any): Response => {
-  switch (error.message) {
+export const handleErrorResponse = (error: unknown): ApiErrorResult => {
+  const message = error instanceof Error ? error.message : undefined;
+  switch (message) {
     case "NotAuthenticated":
-      return responses.notAuthenticatedResponse();
+      return { response: responses.notAuthenticatedResponse() };
     case "Unauthorized":
-      return responses.unauthorizedResponse();
+      return { response: responses.unauthorizedResponse() };
     default:
-      // Delegate to the shared boundary: expected 4xx errors keep their status/message; a
-      // DatabaseError or anything unexpected becomes a generic 500 (never echo internals).
-      return handleApiError(error).response;
+      // Delegate to the shared boundary and return its full { response, error } result — not just
+      // `.response` — so the wrapper's reportApiError receives the real 5xx error (e.g. a
+      // DatabaseError) instead of a synthetic one. Expected 4xx errors keep their status/message.
+      return handleApiError(error);
   }
 };
--- a/apps/web/app/api/v1/auth.ts
+++ b/apps/web/app/api/v1/auth.ts
@@ -1,11 +1,6 @@
 import { NextRequest } from "next/server";
 import { TAuthenticationApiKey } from "@formbricks/types/auth";
-import {
-  DatabaseError,
-  InvalidInputError,
-  ResourceNotFoundError,
-  UniqueConstraintError,
-} from "@formbricks/types/errors";
+import { handleApiError } from "@/app/lib/api/handle-api-error";
 import { responses } from "@/app/lib/api/response";
 import {
   type AuthenticateApiKeyOptions,
@@ -26,15 +21,8 @@ export const handleErrorResponse = (error: any): Response => {
     case "Unauthorized":
       return responses.unauthorizedResponse();
     default:
-      if (error instanceof UniqueConstraintError) {
-        return responses.conflictResponse(error.message);
-      }
-      if (error instanceof ResourceNotFoundError) {
-        return responses.notFoundResponse(error.resourceType, error.resourceId);
-      }
-      if (error instanceof DatabaseError || error instanceof InvalidInputError) {
-        return responses.badRequestResponse(error.message);
-      }
-      return responses.internalServerErrorResponse("Some error occurred");
+      // Delegate to the shared boundary: expected 4xx errors keep their status/message; a
+      // DatabaseError or anything unexpected becomes a generic 500 (never echo internals).
+      return handleApiError(error).response;
   }
 };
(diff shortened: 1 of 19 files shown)


apps/web/app/api/v1/auth.ts
```diff
@@ -26,15 +21,8 @@ export const handleErrorResponse = (error: any): Response => {
...
-      if (error instanceof ResourceNotFoundError) {
-        return responses.notFoundResponse(error.resourceType, error.resourceId);
-      }
-      if (error instanceof DatabaseError || error instanceof InvalidInputError) {
-        return responses.badRequestResponse(error.message);
-      }
-      return responses.internalServerErrorResponse("Some error occurred");
+      // Delegate to the shared boundary: expected 4xx errors keep their status/message; a
+      // DatabaseError or anything unexpected becomes a generic 500 (never echo internals).
+      return handleApiError(error).response;
```
Dhruwang: Taking just `.response` here drops the error that `handleApiError` threads back. Callers do `{ response: handleErrorResponse(error) }`, so `result.error` is undefined and `reportApiError` ends up sending a synthetic "API V1 error, id: …" to Sentry instead of the real DatabaseError — the ~8 routes still on this path (management [id] routes, ee contacts) lose their real 5xx errors in Sentry, while the 13 routes migrated to `handleApiError` directly keep theirs. Could return `{ response, error }` from here and spread it at the call sites, or move those routes onto `handleApiError` too.

xernobyl: Good catch, thanks 🙏 Fixed in bb803ad79.

`handleErrorResponse` now returns the full `{ response, error }` result (typed as a shared `ApiErrorResult`) and the call sites `return handleErrorResponse(error)` directly, so the real error reaches the wrapper's `reportApiError` for the ~8 routes still on this path (management [id] routes + ee contacts) — same signal the 13 directly-migrated routes already get.

I went with your first option (keep the helper) rather than moving each route onto `handleApiError` directly: it preserves the `NotAuthenticated`/`Unauthorized` sentinel handling and keeps a single boundary. Added a unit test asserting the DatabaseError is threaded through.

Dhruwang (approved):

PR #8625, opened 2026-07-23
feat: add current authorization contract

BhagyaAmarasinghe:
Fixes: ENG-1711
## What changed

- Adds the engine-independent, server-only authorization contract for the current Formbricks permission model.
- Derives the 29 valid actions and their resource types from one immutable permission map.
- Exposes Formbricks-owned actor, action, resource, and action-constrained resource types without importing AuthZed, Prisma, configuration, or client code.
- Documents the current organization roles, team/workspace/API-key ladders, inherited survey/dashboard/response rules, and `USER_MANAGEMENT_MINIMUM_ROLE` as evaluator configuration.
- Identifies the TypeScript contract as the application source of truth while keeping `authzed/schema.zed` unchanged as a downstream implementation.

## Why

[ENG-1711](https://linear.app/formbricks/issue/ENG-1711/define-current-model-authorization-types-and-permission-vocabulary) establishes the stable current-model vocabulary that ENG-1712 can use for centralized `can()` and `assertCan()` APIs before any evaluator, AuthZed deployment, relationship projection, or route migration is introduced.

## Public server-only interface

- `TAuthorizationActor`
- `TAuthorizationAction`
- `TAuthorizationResource`
- `TAuthorizationResourceType`
- `TAuthorizationResourceForAction<TAction>`

ENG-1712 should use `NoInfer<TAction>` on the resource parameter so TypeScript cannot widen mismatched action/resource pairs.

## Scope boundaries

This PR intentionally adds no evaluator, SDK calls, AuthZed configuration, schema changes, relationship writes, permission enforcement, database migration, endpoint, or UI behavior. It does not add future capabilities such as survey sharing, dashboard ACL management, or audit-log access.

## Validation

- `pnpm --dir apps/web test lib/authorization` — 1 file / 2 tests passed
- `pnpm --filter @formbricks/web typecheck` — passed, including exact 29-action and all action/resource compile-time mappings
- `pnpm authzed:validate` — 27 relationships, 131 assertions, and 2 expected relations passed
- `pnpm build --filter=@formbricks/web... --force` — passed
- Targeted ESLint, Prettier, and `git diff --check` — passed
- Temporary Client Component import — production build failed at the intended `server-only` boundary; fixture removed and the final production build passed


xernobyl (approved):
Thank you for the PR. Looks great!

PR #8630, opened 2026-07-23
feat: add central can/assertCan authorization interface (ENG-1712)

xernobyl:
## What does this PR do?

Adds the engine-independent central authorization interface — `can()` / `assertCan()` — for Phase 0 of the Authorization & Access Refinement migration. This is the single choke point product code will use to ask "can this actor do this action on this resource?", so the scattered auth checks can later funnel through one place, and the backend can move to SpiceDB without any call site changing.

It's **migration-first / behavior-preserving**: the shipped evaluator (`legacyEvaluator`) delegates to today's org-role / workspace-team / API-key helpers rather than reimplementing them, so it returns exactly what the app enforces today — nobody gains or loses access. AuthZed/SpiceDB is intentionally *not* involved yet (no SDK, no network); a SpiceDB evaluator drops in behind the same `AuthorizationEvaluator` interface later, via the single selection point in `index.ts`.

**Stacked on #8625** (ENG-1711, the authorization contract). Base is `bhagya/eng-1711-…`; GitHub will retarget it to `epic/authzed` once #8625 merges. This PR adds only the evaluator layer on top of her `contract.ts`:

- `evaluator.ts` — the `AuthorizationEvaluator` interface (so the backend is swappable)
- `legacy-evaluator.ts` — the current-rules evaluator (delegates to the existing helpers)
- `resolvers.ts` — small cached lookups that walk a resource to its enforcement boundary (survey → workspace, response → survey → workspace, etc.)
- `index.ts` — the public `can` / `assertCan` + the engine selection point
- `legacy-evaluator.test.ts` — 22 unit tests

A few decisions worth a look during review:

- **Fail-closed, and failures propagate.** A missing resource or no matching grant → `false`; real DB errors are re-thrown, never turned into a silent deny (or allow).
- **`organization.manage_access` honors `USER_MANAGEMENT_MINIMUM_ROLE`** (via the existing `getUserManagementAccess`) instead of a hardcoded owner/manager check — otherwise deployments that set the floor to `owner`/`disabled` would silently get broader member-management access. Locked with a test.
- **API keys are first-class principals** (resolved by id), and the action-first `can<TAction>(…, resource: …ForAction<NoInfer<TAction>>)` signature makes a mismatched action/resource pair a compile error.

Ticket: https://linear.app/formbricks/issue/ENG-1712/implement-can-and-assertcan-authorization-api

## How should this be tested?

Behavior-preserving and not yet wired into any route (call-site migration is ENG-1714), so there's nothing user-facing to click through — it's covered by unit tests:

```bash
# from the repo root
pnpm --filter=@formbricks/web exec vitest run lib/authorization
```

- The 22 tests cover allow/deny, org roles (owner / manager / member / billing), workspace-team levels, API-key scoping, survey/dashboard/response inheritance, the `USER_MANAGEMENT_MINIMUM_ROLE` floor, mismatched action/resource rejection, and operational-failure propagation.
- Because the evaluator delegates to the existing helpers, parity with today's behavior is by construction; the SpiceDB schema's own `pnpm authzed:validate` suite (from the contract PR) asserts the same model.

### Required

- [ ] Ran `pnpm build` (ran `typecheck` + targeted `vitest` + `eslint`; full web build not run)
- [x] Checked for warnings, there are none (eslint clean)
- [ ] Merged the latest changes from main (stacked on `epic/authzed` via #8625, not `main`)

### DIFF
--- a/apps/web/lib/authorization/legacy-evaluator.ts
+++ b/apps/web/lib/authorization/legacy-evaluator.ts
@@ -1,0 +1,267 @@
+import "server-only";
+import { OrganizationAccessType } from "@formbricks/types/api-key";
+import { USER_MANAGEMENT_MINIMUM_ROLE } from "@/lib/constants";
+import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
+import { getAccessFlags, getUserManagementAccess } from "@/lib/membership/utils";
+import { type WorkspaceAction, hasUserWorkspaceAccessForAction } from "@/lib/workspace/auth";
+import { getTeamRoleByTeamIdUserId } from "@/modules/ee/teams/lib/roles";
+import { hasOrganizationAccess, hasPermission } from "@/modules/organization/settings/api-keys/lib/utils";
+import {
+  AUTHORIZATION_PERMISSION_MAP,
+  type TAuthorizationAction,
+  type TAuthorizationActor,
+  type TAuthorizationResourceForAction,
+  type TAuthorizationResourceType,
+} from "./contract";
+import type { AuthorizationEvaluator } from "./evaluator";
+import {
+  getApiKeyAuthById,
+  getApiKeyOrganizationId,
+  getDashboardWorkspaceId,
+  getResponseSurveyId,
+  getSurveyWorkspaceId,
+  getTeamOrganizationId,
+} from "./resolvers";
+
+/**
+ * The Phase-0 authorization evaluator. It reproduces the authorization behavior
+ * Formbricks enforces today by delegating to the existing org/workspace/team/
+ * API-key helpers — it does not reimplement the rules, so it stays in lockstep
+ * with current behavior and with the `./contract` vocabulary (and the
+ * `authzed/schema.zed` mirror the future SpiceDB evaluator implements).
+ *
+ * Operational failures (DB errors) raised by the helpers/resolvers propagate;
+ * only a genuine "no matching grant" (or a missing resource) resolves to
+ * `false`. It never translates an error into a denial.
+ */
+
+/** Split a dotted action (e.g. `"survey.response_export"`) into its parts. */
+const parseAction = (
+  action: TAuthorizationAction
+): { resourceType: TAuthorizationResourceType; permission: string } => {
+  const separator = action.indexOf(".");
+  return {
+    resourceType: action.slice(0, separator) as TAuthorizationResourceType,
+    permission: action.slice(separator + 1),
+  };
+};
+
+/**
+ * Exhaustiveness guard for the actor discriminated union: an unknown or future
+ * principal type is rejected outright — never implicitly treated as an API key.
+ * Adding a new actor type to `TAuthorizationActor` makes every call site that
+ * doesn't handle it a compile error.
+ */
+const rejectUnknownActor = (actor: never): never => {
+  throw new Error(`Unsupported authorization actor type: ${(actor as { type: string }).type}`);
+};
+
+/** Workspace permission level required by each workspace-derived action (mirrors the contract). */
+const WORKSPACE_ACTION_LEVEL: Partial<Record<TAuthorizationAction, WorkspaceAction>> = {
+  // read level → GET
+  "workspace.read": "GET",
+  "survey.read": "GET",
+  "survey.response_read": "GET",
+  "survey.response_export": "GET", // export shares the read gate today
+  "dashboard.read": "GET",
+  "response.read": "GET",
+  "response.export": "GET",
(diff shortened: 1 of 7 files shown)


apps/web/lib/authorization/resolvers.ts
```diff
@@ -0,0 +1,132 @@
+import "server-only";
+import { cache as reactCache } from "react";
```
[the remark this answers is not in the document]
xernobyl: Good catch on the docstring — softened it so it no longer overstates the dedup scope.

Keeping `reactCache` rather than moving to `cache.withCache()` though, deliberately:

- These are **authorization** lookups, so they must not be cached across requests. `withCache` is Redis/cross-request, which would risk serving stale permissions (a security bug); `React.cache()` is request-scoped, so it's the safe choice here.
- Per our caching guideline, `withCache`/Redis is for *expensive* data — these are single indexed PK lookups (just resolving a parent id), which is exactly what `React.cache()` is for.
- It also keeps these consistent with the sibling DAL resolvers (`getSurvey`, `getMembershipByUserIdOrganizationId`, `getWorkspacePermissionByUserId`, …), which all use `reactCache` the same way.

So: docstring fixed, `reactCache` stays.

apps/web/lib/authorization/legacy-evaluator.ts
```diff
@@ -0,0 +1,231 @@
...
+  workspaceId: string
+): Promise<boolean> => {
+  const method = WORKSPACE_ACTION_LEVEL[action];
+  if (!method) return false;
+
+  if (actor.type === "user") {
+    return hasUserWorkspaceAccessForAction(actor.id, workspaceId, method);
+  }
+
+  const auth = await getApiKeyAuthById(actor.id);
```
BhagyaAmarasinghe: [legacy-evaluator.ts:81](https://github.com/formbricks/formbricks/blob/464287b510d2856f190af4ed7bb4c9f3bea276fa/apps/web/lib/authorization/legacy-evaluator.ts#L81-L87) treats every actor other than "user" as an API key. The same pattern appears in organization and team evaluation. An invalid or future principal could therefore receive API-key permissions if its ID matches a key. Use an exhaustive actor switch and reject unknown discriminants.
Note that the same implicit fallback also occurs at lines 122 and 161.

apps/web/lib/authorization/legacy-evaluator.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (resourceType !== resource.type) {
+      throw new Error(`Authorization action "${action}" does not match resource type "${resource.type}"`);
+    }
+
+    switch (resource.type) {
+      case "organization":
+        return canOrganization(actor, permission, resource.id);
+      case "team":
+        return canTeam(actor, permission, resource.id);
+      case "apiKey":
```
BhagyaAmarasinghe, line 245: [canApiKeyResource()](https://github.com/formbricks/formbricks/blob/464287b510d2856f190af4ed7bb4c9f3bea276fa/apps/web/lib/authorization/legacy-evaluator.ts#L176-L185) ignores the parsed permission. Consequently, a runtime call such as apiKey.delete passes the resource-prefix check and returns true for an owner or manager. Validate actions against the permission map or explicitly switch over read and manage. Add runtime-invalid actor and action tests.

BhagyaAmarasinghe (changes requested):

apps/web/lib/authorization/legacy-evaluator.ts
xernobyl: Good call. Made the actor dispatch exhaustive in `canWorkspaceScoped`, `canOrganization`, and `canTeam`: `user` and `apiKey` are now handled explicitly, and anything else hits a `rejectUnknownActor(actor: never)` guard that throws instead of falling through to the API-key path. Because it takes `never`, adding a future principal type to `TAuthorizationActor` becomes a compile error at each site until it's handled. Added a test for an unknown actor type across all three.

apps/web/lib/authorization/legacy-evaluator.ts
xernobyl: Fixed generally rather than per-function: added a vocabulary guard at the top of `can()` that rejects any action not in `AUTHORIZATION_PERMISSION_MAP` before dispatch — so a bogus `apiKey.delete` (or any out-of-vocabulary action) throws instead of reaching an evaluator that would grant it. That closes it for `canApiKeyResource` and every other path in one place. Added invalid-action tests (`apiKey.delete`, `survey.frobnicate`).

BhagyaAmarasinghe (approved):

PR #8631, opened 2026-07-23
fix(editor): support nested lists via Tab in the rich text editor [ENG-1747]

itsjavi:
## What does this PR do?

Fixes [ENG-1747](https://linear.app/formbricks/issue/ENG-1747/fix-tabs-in-rich-text-editor)

In the rich text editor (Lexical), pressing Tab moved browser focus to the next form field, so nested lists could not be created. The editor registered `ListPlugin` and the theme/CSS for nested lists already existed, but nothing handled `KEY_TAB_COMMAND`.

**What changes:**

- **New `ListTabIndentationPlugin`** (`apps/web/modules/ui/components/editor/components/list-tab-indentation-plugin.tsx`): Tab indents and Shift+Tab outdents **only while the selection is inside list items** (both endpoints). Everywhere else Tab keeps its default focus-move behavior, so the editor never traps keyboard users (dashboard a11y guideline §7). Indentation is capped at depth 7, mirroring Lexical's stock `TabIndentationPlugin` mechanics. The plugin mounts unconditionally: `disableLists` only strips the markdown `- ` shortcuts, while all product editors still offer the toolbar list buttons.
- **One list model for every surface** (`styles-editor-frontend.css`, which `packages/surveys/src/lib/styles.ts` injects into the runtime, so it already reaches the editor, the studio preview and respondents): the editor and the runtime had drifted apart, each carrying its own list rules on top of a shared `.fb-editor-listitem { margin: 0 32px }`. Measured on the real stylesheets, a level-1 item sat **48px** from the text in the editor and **56px** in the preview, while each nesting level advanced only 1.5em — and the preview added a 0.5em gap above every list that the editor did not have. List geometry now lives in one place: the container owns the indent (`padding-inline-start: 1.5em`), items add no offset, and the structural wrapper `<li>` Lexical puts around a nested list shows no marker. Result: **1.5em per level on both surfaces, first level included** (21px in the editor at 14px, 24px in the runtime at 16px), and no extra line before a list. The editor-only `.editor li/ul/ol` rules that caused the drift were removed, as were the runtime overrides that were compensating for it.
- **RTL**: the runtime's list fallback used a physical `padding-left`, which left a stray 24px on the wrong side for RTL content; it is now `padding-inline-start` (verified: `padding-left: 0px / padding-right: 24px` in RTL, unchanged in LTR).
- **Email preview** (`apps/web/modules/email/lib/preview-email-template-styles.ts` + `preview-email-template.tsx`): email clients don't load the runtime stylesheet, so a unit-tested `suppressNestedListMarkers()` transform inlines `list-style-type:none` on the wrapper `<li>` for the embed-survey email preview and the sent preview email (best-effort on legacy Outlook).
- **Single source for the nested-list contract**: `NESTED_LIST_ITEM_CLASS` and `NESTED_LIST_ITEM_MARKER_STYLE` are exported from `lib/example-theme.ts` (the Lexical theme that emits the class) and imported by the email transform and its tests. The stylesheets can't import TS, so each carries a reference comment **and** `example-theme.test.ts` reads `styles-editor-frontend.css` and fails if a rename leaves the marker-suppression selector behind.
- **Toolbar accessibility** (`toolbar-plugin.tsx`): the icon-only toolbar buttons had no accessible name; they now expose their translated tooltip text as `aria-label` (no new strings).
- **Playwright coverage** (`apps/web/playwright/survey-editor-rich-text-tabs.spec.ts`): asserts Tab nests / Shift+Tab flattens, that further Tabs stop at the `MAX_INDENT` ceiling (7 levels) without escaping the editor, and that Tab in plain text still moves focus out (keyboard-trap guard). This spec caught a real bug during development (an over-eager render gate that disabled the feature in all product editors), so it guards both halves of the behavior.

**Known, deliberately out of scope** (pre-existing, discovered while testing):

- Sent follow-up emails strip `ul`/`ol`/`li` entirely via their `sanitize-html` allowlist, so lists authored in the follow-up editor silently flatten to text — needs a product decision (the marker bug cannot occur there).
- The live survey preview steals focus from the editor pane when a card mounts (separate fix in progress).
- Toolbar buttons remain `tabIndex={-1}` (not keyboard-tabbable), as before.

## QA / Test Plan

**How to test**

- [ ] Open a survey in the editor → Welcome card → "Welcome message" field (or any question description) → click the "Bulleted list" toolbar button → type `First`, press Enter, type `Second`.
- [ ] With the caret on `Second`, press Tab → the item indents under `First` as a nested list, focus stays in the editor, exactly one bullet per item.
- [ ] Press Shift+Tab → the list flattens back to two top-level items.
- [ ] Repeat with "Numbered list": after nesting, visible numbering stays continuous (1, 2 — the structural wrapper shows no ordinal).
- [ ] Place the caret in plain (non-list) text in the same editor and press Tab → focus moves to the next form field (no keyboard trap).
- [ ] Edge: press Tab repeatedly on a nested item → indentation stops at 7 levels; further Tabs are swallowed (focus stays in the editor).
- [ ] Edge: Shift+Tab on a top-level list item → no-op; focus stays in the editor.
- [ ] Indentation: a flat list sits one small step (1.5em) from the surrounding text — noticeably tighter than before — and every nesting level adds exactly one more step of the same size, in the editor and in the preview. Note: nested ordered lists restart numbering at "1." per level (standard Lexical behavior, not a bug).
- [ ] Editor vs preview: put a paragraph directly above a list. The vertical gap and the left indent match between the editor pane and the live preview — no extra blank line in the preview.
- [ ] Runtime: preview the survey (or answer it as a respondent) → the nested list renders with exactly one marker per item — no stray bullet/ordinal at the nesting point — for both bullet and numbered lists.
- [ ] Email: survey summary → share/embed → Email tab → the preview (and "send preview email") shows the nested list without a stray marker. Legacy Outlook (Word engine) is best-effort.
- [ ] Screen reader / a11y spot check: editor toolbar buttons now announce names ("Bold", "Italic", "Bulleted list", …) matching their tooltips.
- [ ] RTL spot check: in RTL content, lists indent from the right at every level, with no leftover padding on the left.

**Preconditions / test data**

- Any survey with a rich text field (welcome card or question description); no flags or plan requirements.
- Local runtime check only: rebuild the surveys bundle first (`pnpm build --filter=@formbricks/surveys... --force`) and hard-refresh — the package is pre-compiled into `apps/web/public/js/`. CI/production build it fresh.

**Risks & regressions**

- Tab handling in the editor is new: re-check that Tab still leaves the editor from plain paragraphs/headings (covered by the new e2e a11y-guard test) and that `disableLists` editors (follow-up modal, translation inputs) behave unchanged when no list is present.
- List indentation changes everywhere rich text is rendered, flat lists included — that is the intent (they were over-indented), but it is a visible change on existing surveys, not only on nested lists.
- Removing the `.editor li/ul/ol` rules means bullets/numbers in the editor now come from `.fb-editor-list-ul/-ol`. Only Lexical-authored lists live inside `.editor` (the toolbar renders no lists), so nothing else depended on them.
- The runtime's generic `#fbjs … ul/ol/li` rules are untouched apart from the RTL padding fix; they remain the fallback for markup without the editor's theme classes.
- Email preview transform is regex-based; it is idempotent, preserves `class`/`value`/existing `style` attributes, and is covered by unit tests.
- All editor toolbar buttons now have `aria-label`s (translated tooltip text) — no visual change expected.

- none

### Required

- [x] Filled out the "QA / Test Plan" section in this PR (behavior, edge cases, preconditions, risks, migrations/env)
- [ ] Read [How we Code at Formbricks](<[https://github.com/formbricks/formbricks/blob/main/CONTRIBUTING.md](https://formbricks.com/docs/contributing/how-we-code)>)
- [ ] Ran `pnpm build`
- [ ] My changes don't cause any responsiveness issues

### Appreciated

---

### Verification evidence

Latest run (review follow-up commit):

| Command | Scope | Result |
| --- | --- | --- |
| `pnpm exec prettier --check <8 touched files>` | formatting | PASS |
| `pnpm exec eslint <5 touched TS files>` (no `--fix`) | lint | PASS |
| `pnpm exec vitest run modules/email modules/ui/components/editor modules/survey` (in `apps/web`) | unit tests | PASS — 63 files / 839 tests |
| `pnpm exec vitest run` (in `packages/surveys`) | unit tests | PASS — 26 files / 659 tests |
| `pnpm typecheck` (in `apps/web`: `next typegen` + `tsc --noEmit`) | types | PASS |
| `tsc --noEmit` over `playwright/**` + the two touched test files (the repo config excludes both) | types | PASS — 0 errors in touched files (11 pre-existing errors in `survey-accessibility.spec.ts`, untouched here) |
| `pnpm build --filter="./packages/*"` | packages incl. runtime bundle | PASS — built `surveys.umd.cjs` contains `.fb-editor-list-ul,.fb-editor-list-ol{list-style-position:outside!important;margin-block:0 12px!important;padding-inline-start:1.5em!important}` and no longer `margin:0px 32px` |
| Indentation measurement (headless Chromium over the real `preflight.css` + `global.css` + `styles-editor.css` + `styles-editor-frontend.css`, `getBoundingClientRect` per level) | editor + runtime, ol + ul + RTL | see table below |

Measured indent per nesting level (px from the surrounding paragraph):

| Surface | level 1 | level 2 | level 3 | gap above list |
| editor — before | 48.5 | +21 | +21 | 0 |
| editor — after | 21 | +21 | +21 | 0 |
| runtime — before | 56 | +24 | +24 | 8px |
| runtime — after | 24 | +24 | +24 | 0 |

`pnpm test:e2e` was not re-runnable in this checkout (no app build), so the updated spec — including the new depth-7 ceiling assertion — was first exercised by CI: **Run E2E Tests passed in 19m09s**, alongside Build, Linters, Unit Tests and SonarQube (0 open issues).

### DIFF
--- a/apps/web/modules/email/lib/preview-email-template-styles.ts
+++ b/apps/web/modules/email/lib/preview-email-template-styles.ts
@@ -73,8 +73,12 @@ const EMAIL_PREVIEW_ACCENT_COLORS = {
 } as const;
 
 const RICH_TEXT_PARAGRAPH_TAG_REGEX = /<p\b([^>]*)>/gi;
+const RICH_TEXT_LIST_ITEM_TAG_REGEX = /<li\b([^>]*)>/gi;
 const RICH_TEXT_STYLE_ATTRIBUTE_REGEX = /\sstyle=(["'])(.*?)\1/i;
 const RICH_TEXT_STYLE_ATTRIBUTE_REPLACE_REGEX = /\sstyle=(["'])(.*?)\1/gi;
+const RICH_TEXT_CLASS_ATTRIBUTE_REGEX = /\sclass=(["'])(.*?)\1/i;
+const NESTED_LIST_ITEM_CLASS = "fb-editor-nested-listitem";
+const NESTED_LIST_ITEM_MARKER_STYLE = "list-style-type:none";
 
 export const importantStyle = (value: string): string => `${value} !important`;
 
@@ -93,6 +97,39 @@ export const normalizeRichTextSpacing = (html: string): string =>
     }
 
     return `<p${attributes} style="margin:0">`;
+  });
+
+/**
+ * Lexical wraps a nested list in a structural <li class="fb-editor-nested-listitem"> that must not
+ * show its own bullet/number. Email clients ignore the app stylesheets, so the suppression is
+ * inlined per list item (best effort: legacy Outlook's Word engine ignores list-style-type).
+ */
+export const suppressNestedListMarkers = (html: string): string =>
+  html.replaceAll(RICH_TEXT_LIST_ITEM_TAG_REGEX, (tag: string, attributes: string = "") => {
+    const classMatch = RICH_TEXT_CLASS_ATTRIBUTE_REGEX.exec(attributes);
+    const classNames = classMatch ? classMatch[2].split(/\s+/) : [];
+    if (!classNames.includes(NESTED_LIST_ITEM_CLASS)) {
+      return tag;
+    }
+
+    if (RICH_TEXT_STYLE_ATTRIBUTE_REGEX.test(attributes)) {
+      return `<li${attributes.replaceAll(
+        RICH_TEXT_STYLE_ATTRIBUTE_REPLACE_REGEX,
+        (_styleAttribute, quote: string, styleValue: string) => {
+          const trimmedStyle = styleValue.trim();
+          if (trimmedStyle.includes(NESTED_LIST_ITEM_MARKER_STYLE)) {
+            return ` style=${quote}${styleValue}${quote}`;
+          }
+          const styleWithMarker = trimmedStyle
+            ? `${trimmedStyle};${NESTED_LIST_ITEM_MARKER_STYLE}`
+            : NESTED_LIST_ITEM_MARKER_STYLE;
+
+          return ` style=${quote}${styleWithMarker}${quote}`;
+        }
+      )}>`;
+    }
+
+    return `<li${attributes} style="${NESTED_LIST_ITEM_MARKER_STYLE}">`;
   });
 
 const getPreviewDimension = (value: PreviewStyleValue, fallback: PreviewStyleValue): string => {
--- a/apps/web/modules/email/lib/preview-email-template-styles.ts
+++ b/apps/web/modules/email/lib/preview-email-template-styles.ts
@@ -2,6 +2,10 @@ import type { CSSProperties } from "react";
 import type { TSurveyStyling } from "@formbricks/types/surveys/types";
 import { COLOR_DEFAULTS, STYLE_DEFAULTS } from "@/lib/styling/constants";
 import { isLight, mixColor } from "@/lib/utils/colors";
+import {
+  NESTED_LIST_ITEM_CLASS,
+  NESTED_LIST_ITEM_MARKER_STYLE,
+} from "@/modules/ui/components/editor/lib/example-theme";
 
 export interface PreviewEmailStyleTokens {
   accentBackgroundColor: string;
@@ -77,8 +81,6 @@ const RICH_TEXT_LIST_ITEM_TAG_REGEX = /<li\b([^>]*)>/gi;
 const RICH_TEXT_STYLE_ATTRIBUTE_REGEX = /\sstyle=(["'])(.*?)\1/i;
 const RICH_TEXT_STYLE_ATTRIBUTE_REPLACE_REGEX = /\sstyle=(["'])(.*?)\1/gi;
 const RICH_TEXT_CLASS_ATTRIBUTE_REGEX = /\sclass=(["'])(.*?)\1/i;
-const NESTED_LIST_ITEM_CLASS = "fb-editor-nested-listitem";
-const NESTED_LIST_ITEM_MARKER_STYLE = "list-style-type:none";
 
 export const importantStyle = (value: string): string => `${value} !important`;
 
@@ -100,9 +102,9 @@ export const normalizeRichTextSpacing = (html: string): string =>
   });
 
 /**
- * Lexical wraps a nested list in a structural <li class="fb-editor-nested-listitem"> that must not
- * show its own bullet/number. Email clients ignore the app stylesheets, so the suppression is
- * inlined per list item (best effort: legacy Outlook's Word engine ignores list-style-type).
+ * Lexical wraps a nested list in a structural <li> (NESTED_LIST_ITEM_CLASS) that must not show its
+ * own bullet/number. Email clients ignore the app stylesheets, so the suppression is inlined per
+ * list item (best effort: legacy Outlook's Word engine ignores list-style-type).
  */
 export const suppressNestedListMarkers = (html: string): string =>
   html.replaceAll(RICH_TEXT_LIST_ITEM_TAG_REGEX, (tag: string, attributes: string = "") => {
--- a/apps/web/playwright/survey-editor-rich-text-tabs.spec.ts
+++ b/apps/web/playwright/survey-editor-rich-text-tabs.spec.ts
@@ -1,0 +1,111 @@
+import { type Page, expect } from "@playwright/test";
+import { type Fixtures, test } from "./lib/fixtures";
+import { createSurveyFromScratch } from "./utils/helper";
+
+/**
+ * Tab behavior of the survey editor's rich text editor (ENG-1747).
+ *
+ * Regression: Tab always moved focus out of the Lexical editor, so nested lists
+ * could not be created. The fix (ListTabIndentationPlugin) makes Tab indent and
+ * Shift+Tab outdent only while the caret is inside a list item. Outside list
+ * items Tab must keep its default behavior and move focus out of the editor —
+ * a deliberate a11y requirement (no keyboard trap).
+ *
+ * Both scenarios use the welcome card panel, which renders two rich text
+ * editors: "Note*" (headline) and "Welcome message" (description).
+ */
+
+const openWelcomeCardPanel = async (page: Page): Promise<void> => {
+  await expect(page.locator("#welcome-toggle")).toBeVisible();
+  await page.getByText("Welcome Card").click();
+  await page.locator("#welcome-toggle").check();
+  await expect(page.locator('label:has-text("Welcome message")')).toBeVisible();
+
+  // Enabling the card remounts the live preview's card, which focuses its first
+  // control (the welcome card's "Next" button) once it re-renders. Wait for that
+  // one-time focus steal to complete, otherwise it lands mid-typing and both
+  // drops characters and breaks focus assertions. Stacked off-screen preview
+  // cards render dummy buttons with tabindex="-1"; the real one is tabbable.
+  const previewNext = page
+    .locator("#formbricks-survey-container")
+    .getByRole("button", { name: "Next", exact: true })
+    .and(page.locator('[tabindex="0"]'));
+  await expect(previewNext).toBeFocused({ timeout: 15000 });
+};
+
+/**
+ * Same label -> container walk as fillRichTextEditor (utils/helper.ts); the
+ * container also holds the editor's toolbar, whose icon buttons expose their
+ * translated tooltip text as the accessible name ("Bulleted list" is
+ * workspace.surveys.edit.bulleted_list in en-US.json).
+ */
+const editorField = (page: Page, labelText: string) => {
+  const container = page.locator(`label:has-text("${labelText}")`).locator("..").locator("..");
+  return {
+    input: container.locator(".editor-input").first(),
+    bulletListButton: container.getByRole("button", { name: "Bulleted list", exact: true }),
+  };
+};
+
+const gotoFreshSurveyEditorWelcomeCard = async (page: Page, users: Fixtures["users"]): Promise<void> => {
+  const user = await users.create();
+  await user.login();
+  await page.waitForURL(/\/workspaces\/[^/]+\/surveys/);
+  await createSurveyFromScratch(page);
+  await openWelcomeCardPanel(page);
+};
+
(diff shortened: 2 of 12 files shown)


apps/web/modules/email/lib/preview-email-template-styles.ts
```diff
@@ -73,8 +73,12 @@ const EMAIL_PREVIEW_ACCENT_COLORS = {
 } as const;
 
 const RICH_TEXT_PARAGRAPH_TAG_REGEX = /<p\b([^>]*)>/gi;
+const RICH_TEXT_LIST_ITEM_TAG_REGEX = /<li\b([^>]*)>/gi;
 const RICH_TEXT_STYLE_ATTRIBUTE_REGEX = /\sstyle=(["'])(.*?)\1/i;
 const RICH_TEXT_STYLE_ATTRIBUTE_REPLACE_REGEX = /\sstyle=(["'])(.*?)\1/gi;
+const RICH_TEXT_CLASS_ATTRIBUTE_REGEX = /\sclass=(["'])(.*?)\1/i;
+const NESTED_LIST_ITEM_CLASS = "fb-editor-nested-listitem";
```
Dhruwang: **Class name is a stringly-typed contract with no single source.** `fb-editor-nested-listitem` is re-declared here as a literal, but its origin is the Lexical theme (`example-theme.ts` → `list.nested.listitem`), and it's also depended on by `styles-editor-frontend.css` and `packages/surveys/src/styles/global.css`. Nothing links them, and `preview-email-template-styles.test.ts` re-asserts this same literal — so renaming the theme class would silently break marker suppression in **both** the email preview and the survey runtime with zero failing tests.

Suggest exporting the class from `example-theme.ts` and importing it here (and in the test) instead of re-declaring:
```ts
import { NESTED_LIST_ITEM_CLASS } from "@/modules/ui/components/editor/lib/example-theme";
```
The two CSS files can't import it, so a `/* source of truth: example-theme.ts list.nested.listitem */` comment on each selector block keeps them anchored to the origin.

apps/web/modules/email/lib/preview-email-template-styles.ts
```diff
@@ -73,8 +73,12 @@ const EMAIL_PREVIEW_ACCENT_COLORS = {
...
+const NESTED_LIST_ITEM_MARKER_STYLE = "list-style-type:none";
```
Dhruwang: Same drift risk for the suppression **value**: `list-style-type:none` is also hardcoded in both stylesheets (`styles-editor-frontend.css`, `global.css`). If you extract the class name above, worth exporting this one too (e.g. `NESTED_LIST_ITEM_MARKER_STYLE`) so the marker-suppression rule has a single definition shared by the TS transform and — via a reference comment — the CSS.

Dhruwang (changes requested):
Thanks for the PR, looks good. Also I noticed two very small issues which might not be related to this PR but exist for the list. Can we fix them as well?

1. Too much space on the left side when using list

Screenshot 2026-07-28 at 12 52 15 PM

2. The text in the editor and preview is somewhat different. There is a extra line in the preview.

Screenshot 2026-07-28 at 12 53 41 PM

3. Fix sonar issue

apps/web/playwright/survey-editor-rich-text-tabs.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    await expect(input).toBeFocused();
+    await expect(input.locator("li ul li")).toHaveText("Second item");
+    await expect(input.locator("li.fb-editor-nested-listitem")).toBeVisible();
+
+    // Shift+Tab returns it to a single-level list.
+    await input.press("Shift+Tab");
+    await expect(input).toBeFocused();
+    await expect(input.locator("li ul li")).toHaveCount(0);
+    await expect(input.locator("ul > li")).toHaveText(["First item", "Second item"]);
+  });
```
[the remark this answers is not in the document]
itsjavi: Added in cf62f189e. The bullet-list test now keeps pressing Tab past the ceiling and asserts where nesting settles:

```ts
for (let press = 0; press < 10; press++) {
  await input.press("Tab");
}
await expect(input).toBeFocused();
await expect.poll(() => listNestingDepth(input)).toBe(7);
```

`listNestingDepth` walks every `li` up to the contenteditable and counts `ul`/`ol` ancestors, so the equality covers both halves you asked for: an eighth level would fail it just as a missing seventh would. The `toBeFocused()` guards the other half of the ceiling — past `MAX_INDENT` the plugin still calls `preventDefault()`, so a swallowed Tab must not escape the editor mid-list.

Kept it inside the existing test rather than a new one, as you suggested: the state right after the Shift+Tab assertions (two top-level items, caret on the second) is exactly the fixture this needs, and a separate test would re-create a user and a survey to rebuild it.

apps/web/modules/email/lib/preview-email-template-styles.ts
itsjavi: Good catch — done in cf62f189e. `NESTED_LIST_ITEM_CLASS` is exported from `example-theme.ts`, the theme object itself now consumes it (`list.nested.listitem`), and both the transform and its test import it.

For the stylesheets, I added the reference comments you proposed, but backed them with an actual guard — a comment doesn't fail a build, and the whole point of your comment is that a rename fails silently today. `example-theme.test.ts` now reads the stylesheet and asserts the rule is still keyed on the constant:

```ts
const stylesheet = readFileSync(
  fileURLToPath(new URL("../styles-editor-frontend.css", import.meta.url)),
  "utf8"
).replaceAll(/\s+/g, "");

expect(stylesheet).toContain(`.${NESTED_LIST_ITEM_CLASS}{${NESTED_LIST_ITEM_MARKER_STYLE}`);
```

Verified it actually bites — renaming the theme class gives:

```
AssertionError: expected '.fb-editor-text-bold{font-weight:bold…' to contain '.fb-editor-nested-item{list-style-typ…'
```

One simplification that fell out of this: it's now *one* stylesheet, not two. `styles-editor-frontend.css` is injected into the survey runtime by `packages/surveys/src/lib/styles.ts`, so the single marker-suppression rule already covers the editor, the studio preview and the respondent view — the `#fbjs …` copy I had added to `global.css` was redundant and is gone (verified in a browser against the real stylesheets: no stray marker on either surface). `global.css` keeps only the pointer comment back to `example-theme.ts`.

apps/web/modules/email/lib/preview-email-template-styles.ts
itsjavi: Agreed — `NESTED_LIST_ITEM_MARKER_STYLE = "list-style-type:none"` is exported next to the class in `example-theme.ts` and imported by the transform and its tests in cf62f189e.

The stylesheet guard above asserts the pair together (`.${NESTED_LIST_ITEM_CLASS}{${NESTED_LIST_ITEM_MARKER_STYLE}`), so the CSS declaration can't drift from the inlined one either — changing one without the other fails the test.

itsjavi:
Thanks — all three are fixed in cf62f189e / 7aae70080, and both of your screenshots turned out to be the same root cause.

### 1 + 2. Too much space on the left, and the extra line in the preview

The editor and the runtime had drifted into two different list models stacked on a shared `.fb-editor-listitem { margin: 0 32px }`: the editor added its own `.editor li` hanging indent, the runtime its own `ul` padding and a `0.5em` block margin. So a **level-1** item sat far off the text while each *nesting* level advanced only `1.5em`, and the runtime alone put a gap above every list — that gap is the extra line your arrow points at.

Rather than patch each surface, list geometry now lives in one place: `styles-editor-frontend.css`, which `packages/surveys/src/lib/styles.ts` already injects into the runtime. The container owns the indent, items add no offset of their own, and the editor-only `.editor li/ul/ol` rules that caused the drift are gone (nothing else inside `.editor` renders lists, so they had no other job).

Measured with headless Chromium over the real stylesheets (`preflight.css` + `global.css` + `styles-editor.css` + `styles-editor-frontend.css`), px from the surrounding paragraph:

| Surface | level 1 | level 2 | level 3 | gap above list |
| editor — before | 48.5 | +21 | +21 | 0 |
| editor — after | **21** | +21 | +21 | 0 |
| runtime — before | 56 | +24 | +24 | **8px** |
| runtime — after | **24** | +24 | +24 | **0** |

One uniform `1.5em` step per level on both surfaces, first level included (21px in the editor at 14px, 24px in the runtime at 16px), and no gap the editor doesn't also show. Checked bulleted + numbered and LTR + RTL.

Worth flagging for QA: this changes **flat** lists too, everywhere rich text renders — that's the intent (they were over-indented), but it isn't limited to the nested case this PR started from.

While measuring I also hit a genuine RTL bug in the runtime fallback: it used a physical `padding-left`, so RTL content got `1.5em` of dead space on the *left* on top of the correct right indent. Now `padding-inline-start` (verified `padding-left: 0px` / `padding-right: 24px` in RTL, LTR unchanged).

### 3. Sonar

`typescript:S5976` is fixed in 7aae70080 — the near-identical cases are two `test.each` tables now. I folded in the four transform tests as well, not just the three Sonar named, so the rule can't refire on the leftovers; the `value="3"` case now asserts the full output instead of `toContain`, which is a stronger check anyway. SonarQube reports **0 open issues** on the PR.

### Verification

Build, Linters, Unit Tests and SonarQube pass on the pushed commits; E2E is still running. Locally: `prettier`, `eslint`, `tsc --noEmit` (plus a scoped run over `playwright/**` and the touched test files, which the repo's typecheck config excludes — 0 errors in the touched files), 839 unit tests in `apps/web`, 659 in `packages/surveys`, and the rebuilt `surveys.umd.cjs` confirmed to carry the new rules and no longer `margin:0px 32px`.

I could not run `pnpm test:e2e` locally — the spec changed (new depth-7 ceiling assertion) and this checkout has no app build, so CI is the first real run of it. The PR description is updated to match the diff, including that caveat.

Dhruwang (approved):

PR #8636, opened 2026-07-24
fix(workflows): translate untranslated strings in workflow editor (ENG-2036)

Dhruwang:
## What & why

Several strings in the workflow editor rendered in English regardless of the app language (reported in ENG-2036 with Chinese screenshots). Two root causes:

1. **Hardcoded strings never went through `t()`**
   - The default Send-email node set a literal `label: "Send email"`, so `node-registry`'s `node.label ?? t("workspace.workflows.send_email")` fallback never fired → the card title was always English.
   - The multi-email input (shared by follow-ups **and** the workflow email action) hardcoded its placeholder and two validation error messages.
2. **Existing keys were present in the non-English locales but left as English** (locked with English values, so Lingo skipped them) — e.g. `pan_mode`, `pointer_mode`, and ~37 other `workspace.workflows.*` keys.

## Changes

- `editor.ts`: drop the hardcoded `label` on the default Send-email node so the title resolves to the translated `workspace.workflows.send_email` key. A label is still only set when the user renames the node.
- `follow-up-action-multi-email-input.tsx`: add `useTranslation` and wrap the placeholder + both error messages; add three `workspace.surveys.edit.follow_ups_modal_action_email_*` keys to `en-US.json`.
- Regenerated every non-English locale via Lingo so all previously-English `workspace.workflows.*` strings are now translated, and rebuilt `i18n.lock`. Only `en-US.json` source change is the three new keys.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2036/untranslated-strings-in-workflows

## QA / Test Plan

**How to test** 

- [ ] Set the app language to Chinese (or any non-English locale) → open the workflow editor → the Send-email action card title is translated (not "Send email") → expected: e.g. 发送邮件.
- [ ] Hover the canvas toolbar tools → the "Pan mode" / "Pointer mode" tooltips are translated.
- [ ] Open the Send-email action's inspector → the reply-to / recipient field placeholder ("Write an email & press space bar") is translated.
- [ ] In that recipient field, type an invalid email then space → the error message is translated; add a duplicate email → the "already added" error is translated.
- [ ] Switch the app back to English → all of the above read as the original English copy.

**Preconditions / test data**

- A workspace with Workflows enabled; app language set to a non-English locale.

**Risks & regressions**

- Behavior-neutral: no logic changes beyond string sourcing. The multi-email input is shared with survey follow-ups — verify the follow-ups modal recipient field still behaves the same (placeholder + errors now localized).

- none


apps/web/locales/zh-Hans-CN.json
```diff
@@ -4170,46 +4173,46 @@
...
+      "node_needs_survey": "选择一个调查问卷以开始",
+      "pan_mode": "平移模式",
+      "pointer_mode": "指针模式",
       "read_only": "只读",
-      "relative_date": "{date}, {time}",
-      "relative_days_ago": "{count, plural, one {# day ago} other {# days ago}}, {time}",
-      "relative_today": "Today, {time}",
-      "relative_yesterday": "Yesterday, {time}",
+      "relative_date": "{date} {time}",
+      "relative_days_ago": "{count, plural, other {# 天前}},{time}",
```
xernobyl: nit: the sibling `relative_*` strings in this file use a space before `{time}` (e.g. `今天 {time}`, `{date} {time}`), but this one has a comma. aligning it for consistency:

```suggestion
      "relative_days_ago": "{count, plural, other {# 天前}} {time}",
```

minor — and yes, a manual touch on a Lingo-managed file, but it should be stable: the en-US source is unchanged, so Lingo skips (won’t revert) it on the next run. same one-char fix applies to `zh-Hant-TW.json`.

apps/web/locales/zh-Hant-TW.json
```diff
@@ -4170,46 +4173,46 @@
...
+      "node_needs_survey": "選擇一份問卷以開始",
+      "pan_mode": "平移模式",
+      "pointer_mode": "指標模式",
       "read_only": "唯讀",
-      "relative_date": "{date}, {time}",
-      "relative_days_ago": "{count, plural, one {# day ago} other {# days ago}}, {time}",
-      "relative_today": "Today, {time}",
-      "relative_yesterday": "Yesterday, {time}",
+      "relative_date": "{date} {time}",
+      "relative_days_ago": "{count, plural, other {# 天前}},{time}",
```
xernobyl: nit: the sibling `relative_*` strings in this file use a space before `{time}` (e.g. `今天 {time}`, `{date} {time}`), but this one has a comma. aligning it for consistency:

```suggestion
      "relative_days_ago": "{count, plural, other {# 天前}} {time}",
```

minor — and yes, a manual touch on a Lingo-managed file, but it should be stable: the en-US source is unchanged, so Lingo skips (won’t revert) it on the next run. same one-char fix applies to `zh-Hans-CN.json`.

apps/web/locales/zh-Hans-CN.json
Dhruwang: Done in ccf26fe3b — swapped the comma for a space before `{time}` to match the sibling `relative_*` strings. en-US source is unchanged so Lingo won't revert it. Thanks!

apps/web/locales/zh-Hant-TW.json
Dhruwang: Done in ccf26fe3b — swapped the comma for a space before `{time}` to match the sibling `relative_*` strings. en-US source is unchanged so Lingo won't revert it. Thanks!

xernobyl (approved):
Reviewed — LGTM. 👍

Verified against the head:
- all 37 previously-English workflow strings are now translated across the 14 locales, placeholders + ICU plurals intact (nice to see ru `one/few/many/other` and ro `one/few/other` + the "de" large-number form come out correct)
- dropping the hardcoded `Send email` label is right — the node title falls back to the translated `workspace.workflows.send_email` via node-registry, and `label` is optional
- the 3 new follow-up keys match their `t()` calls, and `i18n.lock` stays clean (net +3; the source-locked workflow keys re-stamp identically)

Thanks for also catching the hardcoded follow-up strings and fixing the zh `relative_days_ago` separator.

PR #8638, opened 2026-07-25
fix: resolve pnpm audit dependency vulnerabilities

mattinannt:
## What & why

`pnpm audit` flagged 6 vulnerabilities (3 high, 3 moderate). This resolves 4 of them:

- **postcss** (high, [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849)) — path traversal via `sourceMappingURL` auto-loading could disclose arbitrary `.map` files. Override bumped `8.5.14` → `8.5.18`.
- **better-auth** (high, [GHSA-qq9h-g4jm-xgf3](https://github.com/advisories/GHSA-qq9h-g4jm-xgf3)) — account takeover via pre-account hijacking on magic-link/email-OTP sign-in. Direct dependency bumped `1.6.18` → `1.6.23` (safe patch upgrade per the advisory, no breaking changes).
- **tar** (moderate, [GHSA-r292-9mhp-454m](https://github.com/advisories/GHSA-r292-9mhp-454m)) — uncontrolled recursion in `mapHas`/`filesFilter` allows an uncatchable stack-overflow DoS via a crafted tar entry. Override bumped `7.5.19` → `7.5.21`.
- **valibot** (moderate, [GHSA-5qjj-4xww-7phc](https://github.com/advisories/GHSA-5qjj-4xww-7phc)) — `record()` issue paths can make `flatten()` throw for inherited `Object` property names. New override pinned to `1.4.2` (transitive, via Prisma's dev-only tooling).

`@better-auth/oauth-provider`'s companion `@better-auth/utils` was bumped `0.4.1` → `0.4.2` to satisfy its peer requirement alongside the `better-auth` bump.

**2 issues are intentionally deferred, documented inline in `pnpm-workspace.yaml`:**

- **brace-expansion@5** (high, [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg)) — the only patched release, `5.0.8`, was published 2026-07-23 and is still inside the repo's `minimumReleaseAge` (3-day) supply-chain cooldown, which clears 2026-07-26T11:39 UTC. It's a dev-tooling-only path (`@redocly/cli` → `minimatch`), not a runtime dependency. Left pinned at `5.0.7` with a comment to bump once the cooldown clears; did not add a `minimumReleaseAgeExclude` bypass since that requires going through the repo's documented exception process.
- **@better-auth/oauth-provider** (moderate, [GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528)) — resource-indicator/audience binding for OAuth access tokens. The fix only lands starting in the not-yet-stable `1.7.0` line (`beta.4`+) and requires a breaking `npx auth migrate` schema migration plus a `better-auth` major bump in lockstep. Given the blast radius (this package backs the app's MCP OAuth authorization server) and the lack of a stable patched release, this needs a dedicated, reviewed upgrade rather than an automated dependency bump — filing as a follow-up.

Also removed the `next@16.2.11` entries from `minimumReleaseAgeExclude`: their cooldown window cleared 2026-07-24T16:00 UTC, and the entry's own comment says to remove it once past that date.

**Follow-up commit:** bumping `better-auth` to 1.6.23 turned up a real schema gap — its two-factor plugin added a brute-force-lockout feature (`failedVerificationCount` / `lockedUntil` on `TwoFactor`) that the Prisma schema didn't have columns for, which failed CI's "Better Auth integration tests" job with `PrismaClientValidationError`. Added both columns to `schema.prisma` plus an additive schema migration, verified locally against a real Postgres (all 15 integration test files / 51 tests pass, including the 2 files that were failing).

## Linear ticket

No ticket — this is a scheduled dependency-audit maintenance run.

## How this was tested

- `pnpm install` ✅ — lockfile re-resolves cleanly, passes supply-chain policy checks
- `pnpm audit` — down from 6 vulnerabilities (3 high, 3 moderate) to 2 (1 high, 1 moderate), both documented/deferred above
- `pnpm test` ✅ — 523 test files / 6480 tests passed
- `pnpm lint` ✅ — 0 errors (pre-existing warnings only, unrelated to this change)
- `pnpm build` ✅ — all packages/apps build successfully
- `pnpm test:integration` ✅ (run locally against a real Postgres, matching CI's "Better Auth integration tests" job) — 15 test files / 51 tests passed, including `auth-two-factor.integration.test.ts` and `cutover/reencode-two-factor.integration.test.ts`, which failed in CI before the schema fix

## QA / Test Plan

**How to test**

- [ ] Run `pnpm audit` on this branch → expect 2 vulnerabilities reported (1 high: brace-expansion@5, 1 moderate: @better-auth/oauth-provider), down from 6 on `main`
- [ ] Sign in via email/password on `/auth/login` → expect normal login to still succeed
- [ ] Sign in via magic-link or email-OTP (if enabled) → expect normal passwordless sign-in to still succeed, no regression
- [ ] Enable 2FA (TOTP) on an account, sign out, sign back in → expect the 2FA challenge + verification to work normally (exercises the new `failedVerificationCount`/`lockedUntil` columns)
- [ ] Enter a wrong TOTP/backup code several times in a row → expect eventual temporary lockout (better-auth's new brute-force protection) rather than an error page or crash
- [ ] Exercise an MCP OAuth flow (connect an MCP client against `/api/v3/gateway/token` or the `.well-known` OAuth endpoints) → expect token issuance to work as before (`@better-auth/oauth-provider` only moved a patch version, `1.6.18` → `1.6.23`, to stay in lockstep with `better-auth`'s peer requirement; the audience-binding CVE fix itself is deferred, see above)

**Preconditions / test data**

- Any environment with email/password, magic-link/OTP, and 2FA auth enabled

**Risks & regressions**

- `better-auth` bumped from `1.6.18` → `1.6.23` (5 patch releases) — low risk, but re-check magic-link/email-OTP verification flows and session revocation behavior since that's the area GHSA-qq9h-g4jm-xgf3 touches
- The new `TwoFactor.failedVerificationCount`/`lockedUntil` columns enable better-auth's brute-force lockout by default — re-check that legitimate users aren't locked out unexpectedly under normal retry patterns
- `postcss`/`tar`/`valibot` are build-tooling/transitive dependencies — no runtime behavior change expected

- New DB migration: `packages/database/migration/20260725070000_add_two_factor_lockout_fields` — additive (`failedVerificationCount Int @default(0)`, `lockedUntil DateTime?` on `TwoFactor`), no backfill needed, safe to run on a live DB.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -23,8 +23,8 @@
     "i18n:generate": "dotenv -e ../../.env -- npx lingo.dev@latest run && dotenv -e ../../.env -- npx lingo.dev@latest lockfile --force"
   },
   "dependencies": {
-    "@better-auth/oauth-provider": "1.6.18",
-    "@better-auth/utils": "0.4.1",
+    "@better-auth/oauth-provider": "1.6.23",
+    "@better-auth/utils": "0.4.2",
     "@boxyhq/saml-jackson": "26.2.0",
     "@cubejs-client/core": "1.6.6",
     "@dnd-kit/core": "6.3.1",
@@ -98,7 +98,7 @@
     "@tanstack/react-table": "8.21.3",
     "ai": "6.0.177",
     "bcryptjs": "3.0.3",
-    "better-auth": "1.6.18",
+    "better-auth": "1.6.23",
     "boring-avatars": "2.0.4",
     "class-variance-authority": "0.7.1",
     "clsx": "2.1.1",
--- a/packages/database/migration/20260725070000_add_two_factor_lockout_fields/migration.sql
+++ b/packages/database/migration/20260725070000_add_two_factor_lockout_fields/migration.sql
@@ -1,0 +1,7 @@
+-- Better Auth 1.6.23 two-factor plugin brute-force lockout: adds the failedVerificationCount /
+-- lockedUntil columns it reads and writes on every 2FA verification (see better-auth's
+-- plugins/two-factor/schema.ts). Additive/nullable-or-defaulted, non-breaking.
+
+-- AlterTable
+ALTER TABLE "public"."TwoFactor" ADD COLUMN     "failedVerificationCount" INTEGER NOT NULL DEFAULT 0,
+ADD COLUMN     "lockedUntil" TIMESTAMP(3);
--- a/packages/database/schema.prisma
+++ b/packages/database/schema.prisma
@@ -884,12 +884,16 @@ model VerificationToken {
 /// User.backupCodes columns (kept until cutover). `secret` and `backupCodes` are encrypted
 /// at rest by Better Auth using BETTER_AUTH_SECRET.
 model TwoFactor {
-  id          String  @id @default(cuid())
-  secret      String
-  backupCodes String
-  userId      String
-  user        User    @relation(fields: [userId], references: [id], onDelete: Cascade)
-  verified    Boolean @default(true)
+  id                      String    @id @default(cuid())
+  secret                  String
+  backupCodes             String
+  userId                  String
+  user                    User      @relation(fields: [userId], references: [id], onDelete: Cascade)
+  verified                Boolean   @default(true)
+  /// Better Auth 1.6.23+ two-factor brute-force lockout: consecutive failed verification attempts.
+  failedVerificationCount Int       @default(0)
+  /// Better Auth 1.6.23+ two-factor brute-force lockout: set once failedVerificationCount trips the limit.
+  lockedUntil             DateTime?
 
   @@unique([userId])
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,7 +17,7 @@ overrides:
   node-forge: 1.4.0
   ajv@6: 6.14.0
   minimatch@10: 10.2.5
-  postcss: 8.5.14
+  postcss: 8.5.18
   fast-xml-parser: 5.7.0
   body-parser: 2.3.0
   typeorm: 0.3.31
@@ -33,7 +33,7 @@ overrides:
   esbuild: 0.28.1
   undici@7: 7.28.0
   form-data@4: 4.0.6
-  tar: 7.5.19
+  tar: 7.5.21
   js-yaml@4: 4.3.0
   sharp: 0.35.0
   engine.io@6: 6.6.7
@@ -43,6 +43,7 @@ overrides:
   fast-uri@3: 3.1.4
   linkify-it: 5.0.2
   '@opentelemetry/propagator-jaeger': 2.9.0
+  valibot: 1.4.2
 
 importers:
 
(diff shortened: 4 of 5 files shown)


Dhruwang (commented):
Thank you for the PR. Dependency changes looks good, but are the migration changes also intended?

mattinannt:
Yes, the migration is intended. It's a follow-up to the `better-auth` bump in the first commit, not scope creep.

Bumping `better-auth` 1.6.18 → 1.6.23 (for the account-takeover fix) pulled in a new brute-force-lockout feature in the two-factor plugin. That plugin now reads/writes `failedVerificationCount` and `lockedUntil` on the `TwoFactor` row, but our Prisma schema didn't have those columns — CI's "Better Auth integration tests" job caught this immediately (`PrismaClientValidationError: Unknown argument failedVerificationCount`).

The second commit adds both columns (additive: `Int @default(0)` and nullable `DateTime?`) plus a matching schema migration (`packages/database/migration/20260725070000_add_two_factor_lockout_fields`), verified locally against a real Postgres — all 15 integration test files / 51 tests pass now, including the two that were failing.

Happy to split it into a separate PR if you'd prefer, but since it's a direct consequence of the dependency bump (not a new feature), I kept it in this PR so the `better-auth` version bump doesn't land in a broken state.

---

Dhruwang:
Thanks for the explanation, sounds good !

Dhruwang:
But then how are these values set in DB?

mattinannt:
better-auth's two-factor plugin sets them itself — we don't write to these columns anywhere in our own code. Looking at `better-auth/dist/plugins/two-factor/verify-two-factor.mjs`:

- **On a failed TOTP/backup-code verification** (`recordTwoFactorFailure`): atomically increments `failedVerificationCount` via `adapter.incrementOne`. If the new count reaches `maxFailedAttempts` (default 10), it sets `lockedUntil = now + durationMs` (default 15 min).
- **On a successful verification** (`resetTwoFactorFailures`): unconditionally resets `failedVerificationCount: 0, lockedUntil: null`.
- **On a locked account whose lock has expired** (`assertTwoFactorNotLocked`): lazily clears the stale lock back to `0`/`null` the next time that row is touched.

For existing rows, the migration's `ADD COLUMN "failedVerificationCount" INTEGER NOT NULL DEFAULT 0` backfills every current row to `0` automatically (Postgres applies the column default to existing rows as part of the `ALTER TABLE`), and `lockedUntil` — being nullable with no default — backfills to `NULL`. So every existing 2FA user starts unlocked with a clean counter, and better-auth takes it from there on the next verification attempt.

---

Dhruwang:
Got it. Thanks for the explanation.

Dhruwang (approved):

PR #8644, opened 2026-07-27
fix(workflows): add Reply To helper text, align email labels (ENG-2033)

Dhruwang:
## Problem

In the Workflows **Send email** action, the **Reply To** field had no helper description, unlike **To** and **From** — its meaning was unclear.

## Solution

Add the helper description under **Reply To**, reusing the existing follow-ups copy (`follow_ups_modal_action_replyTo_description`: _"If the recipient hits reply, the following email address will receive it"_).

Single-line, copy-only change — no logic, no new strings, no label changes.

Screenshot 2026-07-27 at 12 32 22 PM

Closes ENG-2033
https://linear.app/formbricks/issue/ENG-2033


itsjavi (approved):
thanks, LGTM!

PR #8645, opened 2026-07-27
fix: declare js-core's real dependency, drop dead database coupling (#ENG-1681)

mattinannt:
## What & why

`packages/js-core` consumed `@formbricks/vite-plugins` in its `vite.config.ts` without declaring it as a workspace dependency. That only worked because of the hoisted `node_modules` layout, and it hid the dependency edge from Turborepo's task graph — a change in `vite-plugins` did not invalidate `@formbricks/js-core#build`.

Investigating that turned up a second, larger problem: js-core's `tsconfig.json` `include` also listed `../database/src/index.ts` and `../types/surveys.d.ts`, neither of which js-core actually consumes. `src/` has zero references to `@formbricks/database`, prisma, or `@formbricks/types`, and the `Window.formbricksSurveys` / `Window.__formbricksNonce` globals it needs are already declared locally in `src/vite-env.d.ts`. That one dead include dragged the entire prisma + types graph into js-core's typecheck and was the standing justification for three `@formbricks/database#build` edges in `turbo.json`.

Separately, `vite.config.ts` declared a build-time `define` for `import.meta.env.VERSION`, sourcing the value by reaching into `apps/web/package.json`.

This PR:

- **Declares `@formbricks/vite-plugins`** as a `devDependency` of `@formbricks/js-core` — the one genuinely load-bearing edge, so Turborepo recognizes it.
- **Deletes the dead `@formbricks/database` / `@formbricks/types` coupling**: both `include` entries, the two devDeps that existed only to formalize them, and the three `@formbricks/database#build` edges in `turbo.json`. js-core's build, lint, and typecheck no longer wait on a database build.
- **Removes the `import.meta.env.VERSION` define entirely.** It had no readers anywhere in `packages/js-core/src` — the define inlined nothing, so the emitted bundle is unaffected. Keeping it (pointed at either package.json) was the real hazard: `apps/web` is *downstream* of `js-core`, so `apps/web/package.json` can never be a Turborepo input for `@formbricks/js-core#build` without creating a dependency cycle, meaning a version bump there does not invalidate `js-core`'s build cache. Release Docker builds only get the right value today because they start cold. Repointing the define at `js-core`'s own `package.json` would have been worse — that file is `private: true`, pinned at `1.0.0`, and never stamped by the release pipeline (`.github/actions/build-and-push-docker/action.yml` stamps only `apps/web/package.json`). Either way, the first consumer added would have silently inlined a stale or placeholder version. Deleting the define leaves nothing to drift.

If SDK version reporting is wanted later, the correct wiring is explicit: a `VERSION` env var declared under `env` (not `passThroughEnv`, which is excluded from the task hash) on `@formbricks/js-core#build`, injected as a Docker build arg, and read at a real call site such as the request headers in `src/lib/common/api.ts`.

### Cache-invalidation effect, measured

Dropping the per-package `#lint` / `#typecheck` blocks lets both inherit the root task config. A `pkg#task` block in `turbo.json` *replaces* the root `task` config rather than merging with it, and neither of those blocks had a `^`-prefixed entry — so neither had any topological dependency. Measured with `turbo run build lint typecheck --filter=@formbricks/js-core --dry=json`, touching `packages/vite-plugins/node-next-dts.ts`:

| task | `dependsOn` after this PR | invalidates on a `vite-plugins` change? |
| --- | --- | --- |
| `@formbricks/js-core#build` | `["^build"]` | ✅ `bf6e45b2a051939d` → `426be777ac1c5ab7` |
| `@formbricks/js-core#typecheck` | `["@formbricks/database#generate", "^typecheck"]` (root) | ✅ `1ebeda4b81491393` → `f9b56e596859810f` — **newly fixed**, previously did not move |
| `@formbricks/js-core#lint` | `[]` (root) | ❌ unchanged at `7fc0aaac4b244c80` |

`@formbricks/database#build` no longer appears in the js-core task graph at all.

**Still open, not addressed here:** `#lint` does not invalidate. The root `lint` task declares no `dependsOn` for *any* package, so this is a repo-wide gap rather than something specific to js-core, and fixing it means changing shared build config for every package. Left for a follow-up.

Also worth stating plainly, since an earlier revision of this description overclaimed it: `@formbricks/types` was **already** invalidating before this PR via the then-existing `@formbricks/database#build` → `^build` → `@formbricks/types#build` chain. It was never a missing-cache-edge problem, which is part of why declaring it was the wrong fix.

`@formbricks/js-core#go` still depends on `@formbricks/database#db:setup`. That is the local dev-server task, left as-is deliberately.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1681/declare-js-cores-cross-package-dependencies

## How this was tested

In `packages/js-core`, after the cleanup:

- `tsc --noEmit` ✅ exits 0 — it previously reported prisma errors pulled in through the dead `../database/src/index.ts` include. Confirmed tsc is genuinely checking, not vacuously passing, by injecting `const __probe: number = "not-a-number"` into `src/index.ts` and seeing it caught (`TS2322`).
- `eslint` ✅ clean. This matters because js-core lints type-aware against its `tsconfig.json`, so narrowing the `include` could have changed lint results.
- `vitest run` ✅ 20 test files / 269 tests passed.
- `vite build` ✅ and the resulting `dist/index.umd.cjs` is **byte-identical** to a baseline built from `main` (29.02 kB / 9.60 kB gzip). Also confirmed the pre-change bundle contains no version string at all, and that `import.meta.env` has zero occurrences across `src/` and `vitest.setup.ts`.

Repo-level:

- `pnpm i` ✅, `pnpm test` ✅, `pnpm lint` ✅, `pnpm build` ✅
- Cache-invalidation table above measured directly via `turbo … --dry=json` hash comparison before/after touching `packages/vite-plugins/node-next-dts.ts`.
- **Lockfile is the 3-line importer addition only.** Two earlier revisions also carried `@better-auth/utils` 0.4.1→0.4.2 and `@better-fetch/fetch` 1.3.0→1.3.1 under `better-call@1.3.7`; that is a deterministic pnpm re-resolution artifact, not intended here, and has been reduced out. The lockfile passes `pnpm install --frozen-lockfile` on the pinned pnpm 11.7.0, and a normal non-frozen `pnpm install --lockfile-only` leaves it byte-unchanged, so it is stable and will not re-bump.

**Re-verified after merging `main`** (which landed the repo-wide ESLint flat-config migration, replacing `packages/js-core/.eslintrc.cjs` with `eslint.config.mjs`):

- Merge of `origin/main` (`322fff7`) into this branch resolved cleanly; the diff against the new base is still exactly the same 5 files, +6/−12.
- `pnpm install --frozen-lockfile` ✅ against the merged lockfile — both sides had touched `pnpm-lock.yaml`.
- `tsc --noEmit` ✅ 0, `eslint .` ✅ 0 under main's new flat config, `vitest run` ✅ 269 tests, and the UMD bundle is **still byte-identical** to the same pre-change baseline.

## QA / Test Plan

This is an internal monorepo build-configuration change. No UI, no API, no runtime logic is touched, the shipped JS bundle is byte-identical, and no dependency versions change — so there is nothing observable for a manual tester to exercise in the running app.

**How to test**

- [ ] Load any page with an in-app survey trigger in a running Formbricks instance → the survey widget renders and submits exactly as before (the `/js/formbricks.umd.cjs` bundle is byte-unchanged, so this is a sanity check that the asset is still built and served, not a behavior change)
- [ ] Sign in with email/password and with an SSO provider → auth works as before (the auth dependency path was touched by earlier revisions of this PR and has since been reverted; this confirms no residue)

**Preconditions / test data**

- None; no accounts, seed data, feature flags, or Cloud vs self-host distinction involved.

**Risks & regressions**

- Runtime: effectively none. The served bundle is byte-identical, verified by diff, and the lockfile changes no dependency version.
- Build config: js-core's `build`/`lint`/`typecheck` no longer wait on `@formbricks/database#build`. If anything in js-core ever does need prisma or `@formbricks/types`, it must declare that dependency properly rather than inheriting it through a tsconfig `include`. Nothing in `src/` references them today.
- `import.meta.env.VERSION` no longer exists in js-core's build config. Nothing referenced it, so nothing breaks; anyone wanting an SDK version constant must add it deliberately (see "What & why").

- none


pnpm-lock.yaml
```diff
@@ -19948,8 +19945,8 @@ snapshots:
 
   better-call@1.3.7(zod@4.3.6):
     dependencies:
-      '@better-auth/utils': 0.4.1
-      '@better-fetch/fetch': 1.3.0
+      '@better-auth/utils': 0.4.2
+      '@better-fetch/fetch': 1.3.1
```
itsjavi, line 19427: **Must fix: unrelated transitive bump on the auth path — please revert.**

These two lines, plus the four removed `resolution`/`snapshots` blocks earlier in this file, re-resolve `better-call@1.3.7`'s transitive deps: `@better-auth/utils` 0.4.1 → 0.4.2 and `@better-fetch/fetch` 1.3.0 → 1.3.1.

This is not a byproduct of the change. I applied this PR's exact `packages/js-core/package.json` edit to a clean tree and ran `pnpm install --lockfile-only` on the pinned pnpm 11.7.0 (root `packageManager`): the result was **only** the 9-line `packages/js-core` importer block at the top of this file, with zero `better-call` lines. A clean tree with no `package.json` change produces an empty diff, so the lockfile is in sync on `main` and this drift came from somewhere else in the authoring environment.

So this is an undocumented transitive bump under better-auth landing in a PR whose QA section states no runtime logic is touched — and CI's `--frozen-lockfile` will not flag it, because the lockfile is self-consistent either way.

Please restore `0.4.1` / `1.3.0` here and re-add the two `@better-auth/utils@0.4.1` / `@better-fetch/fetch@1.3.0` `resolution` blocks and their two `snapshots` blocks, so the lockfile diff is just the importer addition. If the bump is actually wanted, it belongs in a deps PR where it gets looked at on its own merits.

packages/js-core/package.json
```diff
@@ -45,7 +45,10 @@
   "author": "Formbricks <hola@formbricks.com>",
   "devDependencies": {
     "@formbricks/config-typescript": "workspace:*",
+    "@formbricks/database": "workspace:*",
```
itsjavi: **Must fix: js-core does not consume `@formbricks/database` — this formalizes a dead include.**

`packages/js-core/src` has zero references to `@formbricks/database`, prisma, or anything derived from them. The only link is `../database/src/index.ts` in `packages/js-core/tsconfig.json`'s `include`, which came in as leftover from #4815 ("chore: js-core sdk refactor").

Verified locally: delete that include entry → `tsc --noEmit` in `packages/js-core` exits 0. Deleting `../types/surveys.d.ts` as well also exits 0, because js-core already declares the same `Window.formbricksSurveys` / `Window.__formbricksNonce` globals itself in `src/vite-env.d.ts`. (I confirmed tsc was actually checking by injecting a type error into `src/index.ts` and seeing it caught, so those clean exits are not a no-op.)

That one dead entry is what drags the entire prisma + `@formbricks/types` graph into js-core's typecheck — `tsc --listFiles` shows `packages/types/*.ts` pulled in transitively via `../database/src/index.ts` → `../json-types` — and it is the standing justification for the three `@formbricks/database#build` edges on `@formbricks/js-core#build` / `#lint` / `#typecheck` in `turbo.json`, meaning all three tasks currently wait on a full prisma generate + database build. `.eslintrc.cjs` sets `parserOptions.project: "tsconfig.json"`, so lint pulls it in too.

Declaring the dep here locks in a dependency that should be going away. Either drop `@formbricks/database` from these devDeps and delete the dead include plus those three `turbo.json` edges — a bigger win for ENG-1681 than this PR as written — or keep it and say in the description why it is staying.

For contrast, `@formbricks/vite-plugins` on line 51 is the one declaration here that is genuinely load-bearing.

itsjavi (changes requested):
Reviewed by reproducing each claim with `turbo run … --dry=json` hash comparisons rather than from the description.

## What holds up

The `@formbricks/vite-plugins` declaration is a real fix. On `main`, touching `packages/vite-plugins/node-next-dts.ts` or `copy-compiled-assets/index.ts` leaves `@formbricks/js-core#build` at hash `c50d2f291f524dcc` — unchanged — and declaring the dep makes it move (`e74e81e3…` → `118df332…` / `5b8dd756…`). Worth noting turbo 2.9.14 creates the graph node and hashes the package files even though `vite-plugins` has no `build` script, which is why this works at all.

Dropping the `import.meta.env.VERSION` define is also right: zero readers repo-wide (the define was the only occurrence of `import.meta.env` in the package, tests included), `apps/web` does depend on `@formbricks/js-core` (`apps/web/package.json:42`) so the cycle reasoning holds, and `.github/actions/update-package-version` defaults to `./apps/web/package.json`, confirming js-core's own `1.0.0` is never stamped.

Three must-fixes — two inline, plus this one.

## The description overclaims the cache fix

It says the undeclared edges meant "a change in one of those consumed packages wouldn't correctly invalidate `js-core`'s build/lint/typecheck cache." Only **build** is fixed:

- `@formbricks/js-core#lint` and `#typecheck` are declared in `turbo.json` with `dependsOn: ["@formbricks/database#build"]` and no `^`-prefixed entry. A `pkg#task` block fully replaces the root `task` config in turbo rather than merging with it, so neither task has a topological dependency that can pick these packages up. Measured: with all three deps declared, touching `packages/vite-plugins/node-next-dts.ts` leaves `@formbricks/js-core#lint` at `61dbf8f9e2fa9609` — unchanged, even though lint does cover `vite.config.ts`.
- `@formbricks/types` was already invalidating before this PR, through the existing `@formbricks/database#build` → `^build` → `@formbricks/types#build` chain. With no declaration present, touching `packages/types/surveys.d.ts` already moved `js-core#build` `c50d2f29…`→`03e8cc8c…` and `#typecheck` `24021385…`→`58755e8a…`.

`@formbricks/js-core#build` is the only one of the three tasks with `^build`, so it is the only one that benefits. Please narrow the claim to `build`, and either fix `#lint`/`#typecheck` in `turbo.json` or note them as still open.

Unrelated, for anyone skimming: the CodeRabbit walkthrough at the top is stale — it describes commit 1 (repointing the define at js-core's own `package.json`), not head, which deletes the define outright.

pnpm-lock.yaml
mattinannt: Fixed in `fa94f77` — the lockfile diff is now just the 9-line `packages/js-core` importer block, with the `better-call` path untouched. Agreed it shouldn't ride along here.

Two things I verified before reducing it, because the reduced file needs to be stable and not just smaller:

- It passes `pnpm install --frozen-lockfile` on the pinned pnpm 11.7.0.
- A normal, non-frozen `pnpm install --lockfile-only` on top of it leaves it **byte-unchanged**, so it won't silently re-bump on the next install. The pre-commit hook's own full `pnpm install` also reported `Lockfile is up to date, resolution step is skipped` against it.

One correction on the mechanism, since it changes what to watch for rather than the conclusion: **this wasn't authoring-environment drift.** Starting from the base lockfile at `41dc62f` with only this PR's `packages/js-core/package.json` edit applied, `pnpm install --lockfile-only` on pnpm 11.7.0 regenerated a lockfile **byte-identical** to the one that was committed here — `better-call` bumps included. It's a deterministic consequence of pnpm re-resolving when the importer entries are missing, not something stray from my machine.

What most likely produced your 9-line result is a trap I hit twice myself: pnpm short-circuits on `node_modules/.pnpm-workspace-state-v1.json` and prints `Already up to date` without resolving at all. In an already-installed tree that no-ops even under `--force`, and — the tell — it also let `pnpm install --frozen-lockfile --lockfile-only` "pass" against a lockfile that was genuinely missing all three importer entries. Deleting that state file makes pnpm actually resolve, at which point frozen-lockfile correctly fails with:

```
ERR_PNPM_OUTDATED_LOCKFILE
* 3 dependencies were added: @formbricks/database@workspace:*, @formbricks/types@workspace:*, @formbricks/vite-plugins@workspace:*
```

and a real regeneration then reproduces the `better-call` lines.

The practical consequence: anyone regenerating this lockfile from a state that lacks the importer entries will reproduce the bump. Once the importer block is present — as it now is — resolution is stable, which is what makes the hand-reduced file safe to land rather than a diff that fights the next `pnpm install`.

---

packages/js-core/package.json
mattinannt: Took the first option — full cleanup, in `1f19215`. Reproduced your findings first: `src/` has zero references to `@formbricks/database` or prisma, `tsc --noEmit` exits 0 with both `include` entries deleted, and I confirmed tsc wasn't vacuously passing the same way you did (injected `const __probe: number = "not-a-number"` into `src/index.ts`, caught as `TS2322`).

One thing worth adding: **`@formbricks/types` was dead in exactly the same way.** Zero references anywhere in js-core, and as you noted the `Window.formbricksSurveys` / `Window.__formbricksNonce` globals are already declared locally in `src/vite-env.d.ts`. So it went too — leaving `@formbricks/vite-plugins` as the only declaration, which matches your read that it's the one load-bearing one.

What landed:

- Both dead `include` entries removed from `packages/js-core/tsconfig.json`
- `@formbricks/database` and `@formbricks/types` dropped from devDeps
- All three `@formbricks/database#build` edges removed from `turbo.json`

**A side effect worth flagging, because it partly answers your third must-fix.** Dropping the per-package `#lint` / `#typecheck` blocks lets both inherit the root config, and that fixes typecheck invalidation:

| task | `dependsOn` now | touching `vite-plugins/node-next-dts.ts` |
| --- | --- | --- |
| `#build` | `["^build"]` | ✅ `bf6e45b2a051939d` → `426be777ac1c5ab7` |
| `#typecheck` | `["@formbricks/database#generate", "^typecheck"]` | ✅ `1ebeda4b81491393` → `f9b56e596859810f` — newly fixed |
| `#lint` | `[]` | ❌ unchanged at `7fc0aaac4b244c80` |

`@formbricks/database#build` is now absent from the js-core task graph entirely. Typecheck still picks up `@formbricks/database#generate`, but that comes from the root `typecheck` task and applies to every package, so I left it rather than re-introducing a js-core-specific override.

`#lint` is still not invalidating, and I've left it open rather than papering over it: the root `lint` task declares no `dependsOn` for *any* package, so it's a repo-wide gap, not a js-core one. A js-core-only `#lint` block with `^build` would fix this package while leaving the same hole everywhere else — better as a follow-up that fixes it at the root.

Verification after the change: `tsc --noEmit` exits 0 (previously it reported prisma errors through the dead include), `eslint` clean under its type-aware `parserOptions.project`, 20 files / 269 tests pass, and the UMD bundle is still byte-identical to a `main` baseline. I left `@formbricks/js-core#go` → `@formbricks/database#db:setup` alone, since that's the local dev-server task rather than part of this graph.

Description updated to match the new scope.

---

itsjavi (approved):
thanks LGTM

PR #8646, opened 2026-07-27
fix: proxy feedback records in local development

BhagyaAmarasinghe:
Fixes: ENG-1773
## What & why

Local Next.js development bypasses the Traefik/Envoy gateways that normally expose Hub FeedbackRecords, so `/api/v3/feedbackRecords` and `/v1/feedback-records` were unavailable on `localhost:3000`.

This adds development/test-only optional catch-all Route Handlers that reuse the existing FeedbackRecords gateway authorizer, clone body-bearing requests before authorization, translate the v3 path, replace client credentials with the internal Hub credential, strip fixed and `Connection`-nominated hop-by-hop headers, and proxy requests without caching. Production app handlers return 404, leaving deployed Traefik/Envoy routing unchanged. Shared path normalization keeps the proxy and gateway authorizer aligned.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1773/make-apiv3feedbackrecords-work-in-local-development-setup

## How this was tested

- Focused Vitest suites for the proxy, path routing, shared gateway request auth, Traefik auth, and Envoy auth: 5 files / 55 tests passed, including a regression test for a custom `Connection: X-Client-Context` token ✅
- Targeted ESLint for all changed TypeScript files passed ✅
- `pnpm --filter @formbricks/web typecheck` passed ✅
- `pnpm --filter @formbricks/web build` passed; the route manifest contains both optional catch-all proxy routes and the existing `/api/v3/feedbackRecords/token` route ✅
- Manual local Next.js checks: both new routes returned `401 Unauthorized` without credentials, and `/api/v3/feedbackRecords/token` continued to reach its existing handler and returned `401` while signed out ✅
- A full authenticated Hub round trip was not run because the local Docker daemon/Hub dependencies were unavailable.

## QA / Test Plan

**How to test**

- [ ] Start the local stack with the existing `.env.example` Hub defaults (`HUB_API_URL=http://localhost:8080`, `HUB_API_KEY=dev-api-key`) and request `GET /api/v3/feedbackRecords?tenant_id=<feedback-directory-id>` with a read-capable Formbricks API key → `200` with Hub records.
- [ ] Request the equivalent `GET /v1/feedback-records?tenant_id=<feedback-directory-id>` → `200` with the same records.
- [ ] Send a valid write request with a write-capable key, including `POST /api/v3/feedbackRecords` with `tenant_id` in the JSON body → the request reaches Hub and preserves the body/content type.
- [ ] Request a nested record operation with query parameters through either prefix → the record path and query string reach Hub unchanged, except for the expected v3-to-v1 prefix rewrite.
- [ ] Repeat without credentials → `401 Unauthorized`; repeat with insufficient permissions → `403 Forbidden` and Hub is not called.
- [ ] Sign in and call `POST /api/v3/feedbackRecords/token` → the existing token handler returns a gateway token, not a proxied Hub response.
- [ ] Run a production build/deployment behind the supported external gateway → FeedbackRecords continues through Traefik/Envoy; the app-level fallback itself is unavailable in production.

**Preconditions / test data**

- Local Formbricks and Hub services running, with the existing Hub URL/key variables configured.
- A feedback directory containing at least one record.
- Formbricks API keys with read-only and write permissions, plus a signed-in user for the token-route check.

**Risks & regressions**

- Re-check body-bearing authorization to ensure reading tenant data does not consume the body sent to Hub.
- Re-check client `Authorization`, `X-API-Key`, `Cookie`, fixed hop-by-hop, and `Connection`-nominated headers are never forwarded to Hub.
- Re-check the static token route wins over the optional catch-all and deployed production routing remains gateway-owned.

- No migrations or new environment variables. Local use requires the existing `HUB_API_URL` and `HUB_API_KEY` values.

### DIFF
--- a/apps/web/app/api/v3/feedbackRecords/[[...path]]/route.ts
+++ b/apps/web/app/api/v3/feedbackRecords/[[...path]]/route.ts
@@ -1,0 +1,14 @@
+import { NextRequest } from "next/server";
+import { proxyFeedbackRecordsRequest } from "@/modules/hub/feedback-records-proxy";
+
+const handler = async (request: NextRequest): Promise<Response> => {
+  return await proxyFeedbackRecordsRequest(request);
+};
+
+export const GET = handler;
+export const POST = handler;
+export const PUT = handler;
+export const PATCH = handler;
+export const DELETE = handler;
+export const HEAD = handler;
+export const OPTIONS = handler;
--- a/apps/web/app/v1/feedback-records/[[...path]]/route.ts
+++ b/apps/web/app/v1/feedback-records/[[...path]]/route.ts
@@ -1,0 +1,14 @@
+import { NextRequest } from "next/server";
+import { proxyFeedbackRecordsRequest } from "@/modules/hub/feedback-records-proxy";
+
+const handler = async (request: NextRequest): Promise<Response> => {
+  return await proxyFeedbackRecordsRequest(request);
+};
+
+export const GET = handler;
+export const POST = handler;
+export const PUT = handler;
+export const PATCH = handler;
+export const DELETE = handler;
+export const HEAD = handler;
+export const OPTIONS = handler;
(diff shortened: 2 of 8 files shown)


xernobyl (approved):
Thank you for the PR. Looks good to me.

PR #8044, opened 2026-05-18
fix: bind storage uploads to survey elements

BhagyaAmarasinghe:
fixes ENG-871

## Summary
- Bind private storage presigned upload keys to the exact survey and file upload element.
- Scope new private upload URLs under `/storage/{workspaceId}/private/surveys/{surveyId}/elements/{elementId}/...`.
- Keep legacy flat storage URLs readable and deletable by using the catch-all storage route.
- Preserve legacy `survey.questions` file upload validation while using `elementId` for the new client upload API.

## Validation
- `CUBEJS_API_SECRET=test CUBEJS_API_URL=http://localhost:4000 HUB_API_URL=http://localhost:3001 HUB_API_KEY=test pnpm --dir apps/web exec vitest run modules/storage/service.test.ts modules/storage/utils.test.ts`
- `pnpm --dir packages/surveys exec vitest run src/lib/api-client.test.ts`
- `pnpm --dir packages/surveys exec eslint src/lib/api-client.ts src/lib/api-client.test.ts src/components/elements/file-upload-element.tsx`
- `CUBEJS_API_SECRET=test CUBEJS_API_URL=http://localhost:4000 HUB_API_URL=http://localhost:3001 HUB_API_KEY=test pnpm --dir apps/web exec eslint 'app/api/v1/client/[workspaceId]/storage/route.ts' modules/storage/service.test.ts modules/storage/utils.test.ts modules/storage/utils.ts`
- `pnpm --dir packages/types exec eslint storage.ts`
- `pnpm --dir packages/js-core exec eslint src/types/storage.ts`
- `pnpm exec prettier --check 'apps/web/app/api/v1/client/[workspaceId]/storage/route.ts' apps/web/modules/storage/service.test.ts apps/web/modules/storage/utils.test.ts apps/web/modules/storage/utils.ts apps/web/playwright/utils/helper.ts packages/js-core/src/types/storage.ts packages/surveys/src/components/elements/file-upload-element.tsx packages/surveys/src/lib/api-client.test.ts packages/surveys/src/lib/api-client.ts packages/types/storage.ts`
- `rm -rf packages/surveys/dist apps/web/public/js/surveys.* node_modules/.cache/turbo`
- `pnpm build --filter=@formbricks/surveys... --force`

## Notes
- The survey rebuild completed with Vite's local Node warning: Node `22.1.0` is below Vite's preferred `22.12+` patch level.
[...]

PR #8647, opened 2026-07-27
fix(api): bind management response file uploads to authorized workspace [ENG-1981]

Dhruwang:
## What & why

The v1 and v2 **management** create/update response routes validated file-upload URLs with `validateFileUploads`, which only checked the file extension and array shape — it never verified the URL belonged to the caller's workspace, survey, and file-upload element.

That left a cross-tenant storage-deletion path: a response could persist a storage URL pointing at **another workspace's** file (e.g. `/storage/<otherWorkspace>/private/.../file.pdf`). On response deletion, `findAndDeleteUploadedFilesInResponse` reparses the persisted URL and calls `deleteFile(storageId, …)` using the URL's `storageId` as the **primary** deletion key (`survey.workspaceId` is only a not-found fallback) — so the other tenant's file gets deleted.

The client submission paths already defend against this with `validateClientFileUploads` (→ `isScopedPrivateUploadUrl`), which binds each URL to `workspaceId` + `surveyId` + `elementId`. The management routes simply weren't using it.

This PR switches all four management routes (v1 POST/PUT, v2 POST/PUT) to `validateClientFileUploads`, passing the survey-authoritative workspace + survey + blocks/questions in scope at each call site, and removes the now-unused `validateFileUploads` helper and its unit tests.

Scope note: this is the persist-time binding fix (the ticket's stated root cause). The deletion cleanup path is left unchanged — it must keep trusting the URL's `storageId` to support legacy `environmentId`-prefixed uploads, so hardening it further is a separate, higher-risk change tracked as defensive follow-up.

**Behavior:** management responses referencing a file URL that is not a private, correctly-scoped survey-upload URL for the target workspace/survey/element are rejected with `400`. Accepted forms: relative `/storage/…` URLs **and** the app's own absolute URLs (the shape a response GET returns, so a read → re-submit round-trip still works); foreign-origin absolute URLs are rejected. This is stricter than the previous extension-only check, which accepted arbitrary absolute URLs.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1981/prevent-cross-tenant-storage-deletion-through-management-response-file

Related: ENG-1972 (source report).

## How this was tested

- `pnpm vitest run modules/storage/utils.test.ts app/api/v1/management/responses/route.test.ts modules/api/v2/management/responses/[responseId]/route.test.ts` ✅ 77 passed
- `pnpm typecheck` (next typegen + tsc) ✅ 0 errors
- `eslint --no-fix` + `prettier --check` on all touched files ✅ clean
- Tests added/updated:
  - `validateClientFileUploads`: cross-tenant workspace rejection; app-origin absolute URL accepted (round-trip); foreign-origin absolute rejected; legacy questions-only happy + cross-tenant; non-array / non-string / no-data guards.
  - New v1 POST route test and a new v2 PUT assertion: the route hands the validator the **server-authorized** `workspaceId`/`surveyId`, not caller-supplied values.
- No manual/browser QA in this session (management-API-only change, no UI surface).

## QA / Test Plan

**How to test**

- [ ] Upload a file into survey A (workspace W1), then `POST /api/v1/management/responses` for survey A with that scoped URL in the file-upload field → **201/200**, response created.
- [ ] Read that response back (`GET`), then re-`POST`/`PUT` it with the **absolute** file URL exactly as returned → **still accepted** (no 400 regression).
- [ ] `POST` for survey A with a file URL from another workspace (`/storage/<W2>/private/surveys/<other>/elements/<other>/f.pdf`) → **400**; nothing persisted.
- [ ] `POST` with a foreign-origin absolute URL whose path looks correctly scoped (`https://attacker.example.com/storage/<W1>/private/surveys/<A>/elements/<el>/f.pdf`) → **400**.
- [ ] Same cases against `PUT /api/v1/management/responses/:id` and both v2 endpoints → identical results.
- [ ] Edge: URL scoped to the right workspace+survey but a **different element id**, and an external `https://…` URL with no `/storage/` path → both **400**.
- [ ] Regression: create a response with a valid scoped file, then `DELETE` it → only that workspace's file is removed; no other-workspace file is touched.

**Preconditions / test data**

- Two workspaces (W1, W2), each with a survey containing a file-upload element, and a valid management API key for W1.
- A known storage URL belonging to W2 to attempt the cross-tenant case.

**Risks & regressions**

- Tighter validation on the management response API: integrations that previously posted responses with non-scoped, foreign, or arbitrary external file URLs will now get `400`. This is intended. The app's own relative and absolute URLs (the normal upload → read → re-submit flow) are unaffected.
- No change to the client submission paths, the `/test` endpoint, or the deletion cleanup logic.

- none


pandeymangg (changes requested):
Old file URLs in existing responses now get a 400

The new check counts the parts of the path, and wants exactly 8:

/storage/{workspaceId}/private/surveys/{surveyId}/elements/{elementId}/{file}

Anything that isn't 8 parts is rejected — `if (segments?.length !== 8) return false`.

We've shipped three shapes over the years:

before Formbricks 5:
/storage/env_abc/private/resume--fid--u1.pdf                             → 4 parts

Formbricks 5:         
/storage/wsp_xyz/private/resume--fid--u1.pdf                             → 4 parts

after #8044: 
/storage/wsp_xyz/private/surveys/srv_1/elements/el_1/resume--fid--u1.pdf → 8 parts

Only the last one passes. #8044 (d066d2386) landed on 26 May 2026, so anything uploaded before that is one of the short shapes. Those are real files, still sitting in S3, belonging to real customers. 20260204174943_migrate_storage_urls_to_relative exists to clean up storage URLs inside Response.data, and its own comment shows what they look like:
```
http://localhost:3000/storage/env123/public/image.png  →  /storage/env123/public image.png
```
An environment id, and 4 parts.

Example

A customer has a response from March 2026 with a resume attached:

{ "el_resume": ["/storage/env_abc/private/resume--fid--u1.pdf"] }

They read it, change something else, and write it back:

1. GET /api/v1/management/responses/resp_99 → gets the response, resume URL included.
2. Fixes a typo in a different answer.
3. PUT /api/v1/management/responses/resp_99 with the whole object.

Before this PR that was a 200. Now it's a 400 Invalid file upload response…, because the URL has 4 parts instead of 8. They never touched the file. The file is genuinely theirs.

Anything that replays old responses hits this — backfills, re-imports, integrations that sync both ways.

One detail so this isn't overstated: it only happens when the PUT body actually contains the file-upload answer. A PUT of just {"finished": true} has no data, so the check returns early and passes. But a read-then-write does contain it, and that's the normal way an integration updates a response.

Why this looks like a bug and not a decision

We treat the same URL as perfectly fine everywhere else:

- The download route serves the file. getFileStreamForDownload(..., fallbackId) exists purely to retry under the old prefix.
- Deleting the response deletes the file correctly. deleteFile(.ame retry.

So after this PR we will happily serve that file and delete that file, but refuse to save the URL we just handed the caller. If those old URLs were really
invalid, both fallbacks would be dead code.

The PR's risk section describes the rejected set as "non-scoped, foreign, or arbitrary external file URLs". It doesn't mention that our own old URLs are in
that set, so QA and the changelog would miss it.

Worth heading off: the environment → workspace migration doesn't cover this

Easy to assume it does, so to be clear on two separate things:

- Environment records were migrated. Workspace.legacyEnvironmentrkspace used to be env_abc.
- The URLs inside Response.data were not. The only migration that touches them is the relative-URL one above, and it just strips the http://host part.
Nothing ever rewrote the path into the 8-part shape.

And the validator never reads legacyEnvironmentId anyway — it only counts parts. So the migration is invisible to this check.

Suggested fix (keeps the tenant guarantee)

The prefix is still right there in the old URL:                                                                                                          
/storage/env_abc/private/resume.pdf
          ^^^^^^^ the only part that decides whose file gets deleted                                                                                     
So accept a 4-part private URL, but only when that prefix is one the caller's workspace owns: its own workspaceId, or its Workspace.legacyEnvironmentId (schema.prisma:603, already selected by /api/v1/management/me and the client environment route). You can never name another tenant's prefix, so cross-tendeletion stays closed.

What you lose is the survey/element binding, because the old URL never recorded which survey or element the file came from. That's a small loss: today,  within your own workspace, survey and element, you can already p file, since nothing checks the file is actually yours (see P3).

This covers every old file, not just most, which I checked because it wasn't obvious. legacyEnvironmentId is filled in from production environments only (20260403000000_remove_environment_model/migration.sql:11). But _environments runs first, and any development environment that had data was moved into its own new workspace with its type flipped to production (migration.ts:141) — so it gets picked up too. Development environments that were skipped had no surveys, so they have no responses and no file URLs either way.

If we'd rather take the break: please say so in the PR body and the QA plan, and add a test asserting the old shape is rejected, so it reads as a deliberate call rather than an accident.

pandeymangg (approved):
LGTM! 🚀

PR #8648, opened 2026-07-27
fix(hub): bind feedback records gateway API key access to target organization [ENG-1980]

Dhruwang:
## What & why

The feedback records gateway (`apps/web/modules/hub/feedback-records-gateway.ts`) authorized API-key requests against the target feedback directory, but the org-level shortcut in `hasApiKeyImplicitFeedbackDirectoryAccess` returned `true` whenever the key had org-level access control (`organizationAccess.accessControl.read`/`write`) — **without checking that the key's organization matched the directory's organization**.

So an API key with org-level read/write in organization **A** could read or write feedback records for a directory in organization **B** if the directory (tenant) id was known — a cross-tenant access path.

**Fix — API-key feedback-record access now comes solely from the key's per-workspace permissions (`workspacePermissions`); the org-level `organizationAccess.accessControl` flag is no longer consulted at all.** That flag governs organization **management** (members, teams, roles), not workspace data — the API-key UI describes it as *"Members and teams only — not workspace data"* (`apps/web/locales/en-US.json`), and every other consumer (`hasOrganizationAccess`, `/api/v2/me`, the org users/teams endpoints) uses it only for org-admin operations. Granting record access from it would let a members-management key read/write feedback data. This also lines the gateway up with the v3 API-key model, which is already workspace-permissions-only (`apps/web/app/api/v3/lib/auth.ts`).

As defense-in-depth, access additionally requires `authentication.organizationId === directory.organizationId`. A matching workspace permission already implies the same organization (a key's `workspacePermissions` only ever contain workspaces in its own org — see `authenticateApiKeyFromHeaders`, ENG-1749), so the two layers are independent: the org-id check keeps the decision correct on its own and blocks cross-tenant access even if that upstream invariant ever regresses. Unknown permission values fail closed (`Math.max()` of an empty set → `-Infinity` → denied).

To make this security-critical decision unit-testable, the pure logic (`hasApiKeyImplicitFeedbackDirectoryAccess` + its permission-weight tables) is extracted into `feedback-records-gateway-authz.ts`, free of `server-only`/Prisma/env imports. The gateway imports it back; the session (user) authorization branch is unchanged.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1980/bind-feedback-records-gateway-api-key-access-to-target-organization

Related: ENG-1972 (source report).

## How this was tested

- `pnpm vitest run modules/hub/feedback-records-gateway-authz.test.ts` ✅ (org-level access never grants; foreign-org denied; workspace-permission grants for read/write/manage; insufficient weight and no-match denied)
- `pnpm typecheck` ✅ touched files (`feedback-records-gateway*.ts`) are error-free
- `eslint --no-fix` + `prettier --check` on all touched files ✅ clean

Tests added: org-level write with no matching workspace permission → **denied** (read and write); foreign-org key with a matching workspace `manage` permission → **denied** (ENG-1980); workspace `write` → write allowed; workspace `read` → read allowed; workspace `manage` → write allowed; workspace `read` → write denied (insufficient weight); no matching workspace → denied.

## QA / Test Plan

**How to test**

- [ ] As org A, create an API key with a **workspace permission** on a workspace assigned to a feedback directory in org A. Call the gateway for that directory (e.g. `GET /api/v3/feedbackRecords?tenant_id=<orgA_directory_id>`, and a `POST`/`PATCH`/`DELETE`) → **allowed** per the permission level (`write`/`manage` for mutations, `read` for reads).
- [ ] Same key, a directory in **org B** (known directory id) → **403 Forbidden**; no records returned or mutated (ENG-1980 cross-tenant path).
- [ ] Org-level access-control key (read and/or write) with **no** workspace permission on the target directory's workspace → **403** for both read and write. Org-level access no longer grants feedback-record access.
- [ ] Workspace `read`-only permission on a write op (`POST`/`PATCH`/`DELETE`) → **403**; on a read op (`GET`) → **allowed**.
- [ ] Regression: user (session) principal access is unchanged — owners/managers and workspace-team members of the directory's org still authorized as before.

**Preconditions / test data**

- Two organizations (A, B), each with at least one feedback directory that has **≥1 assigned workspace**; feedback directories license/entitlement enabled.
- An org-A API key scoped to a workspace shared with the target directory, and an org-A API key with org-level access control but no matching workspace permission.
- A known org-B feedback directory id for the cross-tenant attempt.

**Risks & regressions**

- **Behavioral tightening (intended):** same-org API keys that relied on **org-level** access control to reach feedback records now get **403** — access requires a matching workspace permission. This is a deliberate correction; any integration built against the old org-level behavior must be granted the appropriate workspace permission.
- **Accepted side effect:** a feedback directory with **zero assigned workspaces** is unreachable by every API key (read and write), since access derives solely from workspace permissions. `workspaceIds` is optional on directory create/update, so this state is reachable from the UI. Blast radius is small — the composite FK on `FeedbackSource` means such a directory can't have sources and therefore no ingested records — but `POST /api/v3/feedbackRecords` into a not-yet-assigned directory returns 403. Session users still reach it via the org owner/manager branch. Accepted for this fix; assigning ≥1 workspace resolves it.
- No change to gateway routing, tenant resolution, JWT verification, or the user/session authorization branch.

- none

### DIFF
--- a/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
+++ b/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "vitest";
-import { TAuthenticationApiKey } from "@formbricks/types/auth";
+import type { TAuthenticationApiKey } from "@formbricks/types/auth";
 import { hasApiKeyImplicitFeedbackDirectoryAccess } from "./feedback-records-gateway-authz";
 
 const DIRECTORY_ORG_ID = "org_directory";
@@ -82,10 +82,11 @@ describe("hasApiKeyImplicitFeedbackDirectoryAccess", () => {
     });
   });
 
-  describe("workspace-permission path (org-scoped by construction)", () => {
-    test("grants when a workspace permission matches a directory workspace with sufficient weight", () => {
+  describe("workspace-permission path (same-organization keys only)", () => {
+    test("grants a same-organization key (no org-level access) via a matching workspace permission", () => {
+      // Fall-through: org-level accessControl is all-false, so access comes solely from the
+      // per-workspace permission.
       const workspaceKey = makeApiKeyAuth({
-        organizationId: "org_attacker",
         workspacePermissions: [
           { workspaceId: DIRECTORY_WORKSPACE_ID, workspaceName: "Shared", permission: "write" },
         ],
@@ -99,6 +100,27 @@ describe("hasApiKeyImplicitFeedbackDirectoryAccess", () => {
           "write"
         )
       ).toBe(true);
+    });
+
+    test("denies a foreign-organization key even with a matching workspace permission (defense-in-depth, ENG-1980)", () => {
+      // Upstream (authenticateApiKeyFromHeaders) filters workspacePermissions to the key's own org,
+      // so this shape shouldn't occur in production — but the authz function must deny it on its own
+      // rather than rely on that invariant holding.
+      const foreignKey = makeApiKeyAuth({
+        organizationId: "org_attacker",
+        workspacePermissions: [
+          { workspaceId: DIRECTORY_WORKSPACE_ID, workspaceName: "Shared", permission: "manage" },
+        ],
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          foreignKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(false);
     });
 
     test("denies when the read-only workspace permission is below the required write weight", () => {
--- a/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
+++ b/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
@@ -1,0 +1,138 @@
+import { describe, expect, test } from "vitest";
+import { TAuthenticationApiKey } from "@formbricks/types/auth";
+import { hasApiKeyImplicitFeedbackDirectoryAccess } from "./feedback-records-gateway-authz";
+
+const DIRECTORY_ORG_ID = "org_directory";
+const DIRECTORY_WORKSPACE_ID = "wsdirectory0000000000000";
+
+const makeApiKeyAuth = (overrides: Partial<TAuthenticationApiKey> = {}): TAuthenticationApiKey => ({
+  type: "apiKey",
+  apiKeyId: "key_1",
+  organizationId: DIRECTORY_ORG_ID,
+  organizationAccess: { accessControl: { read: false, write: false } },
+  workspacePermissions: [],
+  ...overrides,
+});
+
+describe("hasApiKeyImplicitFeedbackDirectoryAccess", () => {
+  describe("organization-level access control is bound to the API key's own organization (ENG-1980)", () => {
+    test("denies a foreign-organization key that has org-level write, for read and write", () => {
+      const foreignKey = makeApiKeyAuth({
+        organizationId: "org_attacker",
+        organizationAccess: { accessControl: { read: true, write: true } },
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          foreignKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "read"
+        )
+      ).toBe(false);
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          foreignKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(false);
+    });
+
+    test("grants a same-organization key with org-level write, for read and write", () => {
+      const ownerKey = makeApiKeyAuth({
+        organizationAccess: { accessControl: { read: false, write: true } },
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(ownerKey, DIRECTORY_ORG_ID, [DIRECTORY_WORKSPACE_ID], "read")
+      ).toBe(true);
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          ownerKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(true);
+    });
+
+    test("grants a same-organization key with org-level read only for read", () => {
+      const readerKey = makeApiKeyAuth({
+        organizationAccess: { accessControl: { read: true, write: false } },
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          readerKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "read"
+        )
+      ).toBe(true);
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          readerKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(false);
+    });
+  });
+
+  describe("workspace-permission path (org-scoped by construction)", () => {
+    test("grants when a workspace permission matches a directory workspace with sufficient weight", () => {
+      const workspaceKey = makeApiKeyAuth({
+        organizationId: "org_attacker",
+        workspacePermissions: [
+          { workspaceId: DIRECTORY_WORKSPACE_ID, workspaceName: "Shared", permission: "write" },
+        ],
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          workspaceKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(true);
+    });
+
+    test("denies when the read-only workspace permission is below the required write weight", () => {
+      const workspaceKey = makeApiKeyAuth({
+        workspacePermissions: [
+          { workspaceId: DIRECTORY_WORKSPACE_ID, workspaceName: "Shared", permission: "read" },
+        ],
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          workspaceKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(false);
+    });
+
+    test("denies when no workspace permission matches a directory workspace", () => {
+      const workspaceKey = makeApiKeyAuth({
+        workspacePermissions: [
+          { workspaceId: "wsother00000000000000000", workspaceName: "Other", permission: "manage" },
+        ],
+      });
+
+      expect(
+        hasApiKeyImplicitFeedbackDirectoryAccess(
+          workspaceKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "read"
+        )
+      ).toBe(false);
+    });
+  });
+});
--- a/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
+++ b/apps/web/modules/hub/feedback-records-gateway-authz.test.ts
@@ -15,39 +15,18 @@ const makeApiKeyAuth = (overrides: Partial<TAuthenticationApiKey> = {}): TAuthen
 });
 
 describe("hasApiKeyImplicitFeedbackDirectoryAccess", () => {
-  describe("organization-level access control is bound to the API key's own organization (ENG-1980)", () => {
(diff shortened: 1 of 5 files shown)


apps/web/modules/hub/feedback-records-gateway-authz.ts
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+
+  const matchingWeights = authentication.workspacePermissions
+    .filter((permission) => workspaceIds.includes(permission.workspaceId))
+    .map((permission) => apiKeyPermissionWeight[permission.permission]);
+
+  if (matchingWeights.length === 0) {
+    return false;
+  }
```
xernobyl, line 55: **[P3]** Side effect worth deciding on explicitly: a feedback directory with **zero assigned workspaces** is now unreachable by every API key, for both read and write.

`workspaceIds` is optional on create and update (`types/feedback-directory.ts:46,53`), `createFeedbackDirectory` passes `undefined` when the list is empty (`lib/feedback-directory.ts:330`), and the settings modal never requires one — so that state is reachable straight from the UI.

Session users still get in via the org owner/manager branch (`feedback-records-gateway.ts:280`), so the API-key path is now strictly narrower than the UI path for the same person.

Blast radius is small in practice — the composite FK on `FeedbackSource` means such a directory can't have sources, hence no ingested records — but `POST /api/v3/feedbackRecords` into a not-yet-assigned directory will now 403. Either call it out in the risks section as accepted, or require ≥1 workspace at directory create.

apps/web/modules/hub/feedback-records-gateway-authz.test.ts
```diff
@@ -0,0 +1,122 @@
...
+          foreignKey,
+          DIRECTORY_ORG_ID,
+          [DIRECTORY_WORKSPACE_ID],
+          "write"
+        )
+      ).toBe(false);
+    });
+  });
+
+  describe("workspace-permission path (same-organization keys only)", () => {
```
xernobyl, line 68: **[P3]** Small gap: none of the five cases assert a *grant* for `requiredPermission: "read"` — they either assert `false`, or assert `true` for `"write"`.

So if `gatewayPermissionToApiKeyPermissionWeight.read` were ever wrong, this file wouldn't catch it; only the envoy integration test (`envoy-auth/service.test.ts`, "allows similar feedback…", `permission: "read"` → 200) would. Given the module was extracted specifically to make this decision unit-testable, two more cases would close it:

- `read` workspace permission + `read` op → `true`
- `manage` workspace permission + `write` op → `true`

xernobyl (commented):
The security fix is correct — and it's actually stronger than the description claims. Nice one. A few things to sort before merge, mostly around the description itself.

### What I checked

The final code doesn't just org-gate the shortcut; `247f5501` **removes it entirely**, so feedback-record access now comes solely from `workspacePermissions`. I think that's the right call:

- The UI copy for that flag literally says *"Members and teams only — not workspace data."* (`apps/web/locales/en-US.json:1859`)
- Every other consumer (`hasOrganizationAccess`, `/api/v2/me`, the org users/teams endpoints) uses it only for org-admin ops.
- It lines the gateway up with the v3 API-key model, which is already workspace-permissions-only (`apps/web/app/api/v3/lib/auth.ts:99`).

The org-id check is genuine defense-in-depth as the comment says — ENG-1749's filter in `authenticateApiKeyFromHeaders` already drops foreign-org workspace permissions, so the two layers are independent. Fail-closed on an unknown permission value too (`Math.max(NaN)` → false). Extracting to a `server-only`-free module is clean.

---

### [P2] The description and QA plan document the abandoned first-commit approach

The body says *"gate the org-level shortcut on `authentication.organizationId === directory.organizationId`"* and *"the workspace-permission path below it is left unchanged"*. Neither is true of the head any more. That makes three QA steps wrong in a way that'll read as failures:

- *"Call the gateway for a directory in the key's **own** org A with the same key → allowed"* → now **403** unless the key also holds a matching workspace permission.
- *"Org-level **read-only** key against an own-org directory: read ops **allowed**"* → now **403**.
- *"Legitimate same-org keys are unaffected"* → not quite; same-org org-level-only keys lose access. That's the intended tightening, but it's an undeclared breaking change for anyone who built against the old behaviour.

Since this is a security fix that someone will audit (and QA) from that plan later, probably worth rewriting the body to match the head.

---

### Bot reconciliation

- **CodeRabbit is stale, not clean.** Its "No actionable comments 🎉" was produced against `41dc62fb…b57bf9cd` — commit 1 only — and it notes *"Review skipped: incremental reviews are disabled."* The three commits that actually changed the security semantics (`615a8c22`, `247f5501`, `b3fbe721`) were never bot-reviewed. Might be worth a `@coderabbitai full review` before merge.
- **SonarQube green:** quality gate passed, 0 new issues, 0 hotspots, 100% coverage on new code.
- **CI:** all 17 checks passing.

Leaving this as a comment rather than an approval only because of the description mismatch — the code itself looks good to me.

apps/web/modules/hub/feedback-records-gateway-authz.test.ts
Dhruwang: Added both cases in `0749d717`: `read` workspace permission + `read` op → `true`, and `manage` workspace permission + `write` op → `true`. The `gatewayPermissionToApiKeyPermissionWeight.read` path is now exercised directly by the unit suite.

apps/web/modules/hub/feedback-records-gateway-authz.ts
Dhruwang: Called out explicitly in the QA plan's **Risks & regressions** as an accepted side effect (zero-workspace directory unreachable by every API key; blast radius bounded by the `FeedbackSource` composite FK; session users still reach it via the org owner/manager branch; assigning ≥1 workspace resolves it). Left directory-create validation out of scope for this security fix.

Dhruwang:
Thanks for the thorough review @xernobyl 🙏

Addressed all three in `0749d717` (which also merges `main` and resolves the conflict — `main` extracted routing into `feedback-records-routing`, this branch extracted authz into `feedback-records-gateway-authz`; both are kept):

- **[P2] Description/QA mismatch:** rewritten the body + QA plan to match the head. It now states access is **workspace-permissions-only** (org-level `accessControl` no longer consulted), the org-id check is defense-in-depth, and the same-org org-level-only tightening is called out as an intended breaking change. The three formerly-wrong QA steps are rewritten.
- **[P3] Test grant gap:** added `read` perm + `read` op → `true` and `manage` perm + `write` op → `true`.
- **[P3] Zero-workspace directory:** documented in **Risks & regressions** as an accepted side effect (bounded by the `FeedbackSource` composite FK); left create-time validation out of scope.

Will kick off a fresh CodeRabbit pass now that the security-changing commits are in.

Dhruwang:
@coderabbitai full review

apps/web/modules/hub/feedback-records-gateway.ts
```diff
@@ -271,6 +236,7 @@ const authorizeFeedbackRecordsGatewayRequest = async (
   if (principal.type === "apiKey") {
     return hasApiKeyImplicitFeedbackDirectoryAccess(
       principal.authentication,
+      feedbackDirectory.organizationId,
```
xernobyl, line 239: **[P3]** Nothing pins this argument. The unit suite calls `hasApiKeyImplicitFeedbackDirectoryAccess` directly with an explicit `directoryOrganizationId`, so if this line ever became `principal.authentication.organizationId`, the org check turns into a tautology, the cross-org path reopens, and every test in the repo still passes.

Wiring it to some other *wrong* value would break the existing positive envoy/traefik cases, so this is narrowly about that one mutation — but it happens to be the exact exploit path ENG-1980 was filed for. One integration-level case would close it: in `envoy-auth/service.test.ts`, a key in `org_1` against a directory whose `organizationId` is `org_2` with a matching workspace permission → 403.

xernobyl (commented):
Rereviewed at `0749d717`. All three of my earlier points are addressed:

- **[P2] description/QA mismatch** — the body now leads with workspace-permissions-only, frames the org-id check as defense-in-depth, and lists the same-org tightening as an intended breaking change. Matches the code.
- **[P3] zero-workspace directory** — documented under Risks with the `FeedbackSource` composite-FK blast-radius argument and the session-user escape hatch. Fine as an accepted side effect.
- **[P3] missing `read` grant assertion** — both cases added; `gatewayPermissionToApiKeyPermissionWeight.read` is now exercised directly.

CodeRabbit's full review came back with no actionable comments, Sonar's gate passed with 0 new issues, and all checks are green. The security change itself looks right to me: dropping the `organizationAccess.accessControl` shortcut is the correct reading of that flag, and the org-id match is a sensible independent layer.

One inline nit below, plus a description nit:

**[P3] description** — "Unknown permission values fail closed (`Math.max()` of an empty set → `-Infinity` → denied)" describes the *no-match* path rather than the unknown-value one. An unrecognised `permission` produces `[undefined]` → `Math.max` → `NaN`, and `NaN >= n` is false. Both deny, so the code is fine — only the stated mechanism is off.

### Not blocking on this PR, but worth knowing before it merges

Two other open PRs change the same authz decision, and they disagree with each other:

- #8675 (ENG-1770) makes `update`/`delete`/`bulkDelete` authorizable **only** via `orgAccessControl.write`, with the workspace path returning `false` for mutations. This PR removes that org shortcut. Applied together, no API key can mutate a feedback record at all — not even one holding org read+write *and* workspace `manage`. #8675 also keeps the cross-org path open (its `orgAccessControl?.write → true` still runs before the mutation gate), so it doesn't close ENG-1980 on its own.
- #8682 requires `manage` weight *from a workspace permission* for deletes — the opposite premise to #8675 — and explicitly defers the org scoping here.

Both of those also add `modules/hub/feedback-records-gateway.test.ts` as a new file, so they conflict with each other as well as with this PR's extraction. My suggestion: land this one first (it's the one that actually closes the cross-tenant hole), then have the other two rebase onto `feedback-records-gateway-authz.ts` — and settle "may a workspace permission mutate a shared directory's records?" between #8675 and #8682 first, since that's a product decision rather than a merge conflict.

apps/web/modules/hub/feedback-records-gateway.ts
Dhruwang: Added a wiring-level regression test in `modules/envoy-auth/service.test.ts` (commit 8c2938da0): a key in `org_1` against a directory whose `organizationId` is `org_2`, with a matching workspace permission → 403. It runs the full `authorizeEnvoyRequest` path so it exercises this call site, not just the pure function.

Verified it closes the exact mutation you described: temporarily changing this arg to `principal.authentication.organizationId` flips the test to 200 (fails); it passes on the real code. Thanks for catching it.

xernobyl (approved):
Approving at `8c2938da0`. My last open point is closed, and I verified the fix rather than taking the claim at face value.

### [P3] wiring guard — resolved ✅

The new envoy case is exactly what was needed. I reproduced your mutation check independently:

- Baseline at head: **33/33** across `feedback-records-gateway-authz.test.ts`, `envoy-auth/service.test.ts`, `traefik-auth/service.test.ts`.
- Changing the arg at `feedback-records-gateway.ts:239` to `principal.authentication.organizationId`: **1 failed / 32 passed**, and the only failure is `denies a cross-org API key even with a matching workspace permission (ENG-1980)`. The pure unit suite still passes — so that test really is the sole guard on this call site.

The `expect(mockGetIsFeedbackDirectoriesEnabled).toHaveBeenCalledWith("org_2")` assertion is a nice addition I hadn't asked for: it pins the directory's org driving the entitlement gate too, not just the authz call.

Also checked that traefik deliberately has no duplicate — the authorizer is shared, so the envoy case covers the single call site and a copy would just be a low-signal duplicate. Agreed with leaving it.

### Two non-blocking nits — not gating this

**[P3] The fail-closed sentence in the description is still off**, and now in two ways rather than one:

> Unknown permission values fail closed (`Math.max()` of an empty set → `-Infinity` → denied)

`feedback-records-gateway-authz.ts:53` has an explicit `if (matchingWeights.length === 0) return false;`, so the empty-set → `-Infinity` path is unreachable. And the unknown-value path is `[undefined]` → `Math.max` → `NaN`, with `NaN >= n` false. Both deny, so the code is correct — only the stated mechanism is wrong.

**[P3] `## How this was tested` doesn't mention the new envoy test.** It still lists only the authz unit suite, and the "foreign-org key with a matching workspace `manage` permission → denied" line refers to the unit case that already existed at `0749d717`. Worth adding, since the wiring test is the most valuable guard in the PR — a release-QA reader has no way to know to check it otherwise.

### Bots

Sonar re-ran after the new commit: gate passed, 0 new issues, 100% coverage on new code. Worth noting CodeRabbit's "no actionable comments" badge predates this head — its last pass covered `e54641b0..0749d717`, so it hasn't seen `8c2938da0`. Test-only, so I'm not worried.

### Merge order still matters

Restating from my last review, since it's still unresolved and it's the real risk here rather than anything in this PR: #8675 and this PR disagree about whether a workspace permission may authorize mutating a shared directory's records. I checked what happens if both land as written — a maximally-privileged same-org key (org read+write *and* workspace `manage`) gets denied PATCH, DELETE and bulkDelete, so API-key mutation stops working entirely. #8675 and #8682 also both add `feedback-records-gateway.test.ts` as a new file, so they'll conflict with each other regardless.

This one should go in first. Happy for it to merge as-is; the two nits above can ride along or follow.

PR #8649, opened 2026-07-27
fix(workflows): restrict send_email recipients to organization members [ENG-2029]

Dhruwang:
## What & why

A workflow `send_email` action accepted an arbitrary external `to` address with **no recipient allowlist**. A workspace member with write access could create and **enable** a live workflow that, on every completed response, emails the full response data (including hidden fields + variables) to any outside inbox — a silent, continuous data-egress surface (ENG-2029). `from` is not spoofable (the runner always sends from `MAIL_FROM`); the issue is the arbitrary **recipient** + attached response data.

The `to` field is Follow-Ups-parity: it is either a literal email address, or a survey element id that resolves against the response at send time. Only the **literal** case is the egress vector — a respondent-field recipient resolves to an address carried by that same submission.

This PR binds literal recipients to the workflow's organization (org-member allowlist) at every enforcement point:

- **Enable** (and, through the same `@formbricks/workflows` handler, the MCP `create`/write surface): a new injected `verifyRecipientsAllowed` context capability checks every literal `to` on the definition's `send_email` nodes against the organization's member emails. A non-member recipient fails enable with `422 workflow_not_executable` (`invalid_params` names `definition.nodes.config.to`, and `detail` names the recipient rule so a UI surfacing only `detail` still shows the cause). Implemented in the v3 adapter (reused by MCP) via a new focused `getOrganizationMemberEmails` query.
- **Test / dry run**: `testWorkflow` runs the same check and reports a collected `recipient_not_allowed` problem (still a `200` with `ok: false`). The dry run exists to catch exactly this before going live — without it a workflow addressing an external inbox would test green and then hard-fail enable.
- **Send** (runner): the run job loads the member-email allowlist and skips (fails the step, no email) any send whose literal recipient is not a member. This is the backstop for a member removed after enable, or a definition that reached the runner some other way; it **fails closed** when the organization can't be resolved. The allowlist query is skipped entirely when no planned step has a literal recipient, so the common respondent-field run costs nothing extra.

A respondent-field `to` is always allowed and never allowlist-checked — only literal addresses are gated. The literal-vs-field test is shared via `isLiteralEmailRecipient` so enable, test, send, and the adapter agree; the reported field path and per-recipient message are shared between enable and test so the two reports can't diverge. Matching is case-insensitive and restricted to active members.

The org-member rule is now documented where callers actually look: the enable/test endpoint descriptions, `WorkflowSendEmailActionConfig.to` in the OpenAPI spec, and the Zod `to` field (previously undescribed, so an MCP agent had no way to learn the constraint before tripping it). A drift guard pins the dry-run problem-code enum to the shared schema.

Out of scope (deliberate, per triage): gating the response-data attachment (`attachResponseData` / `includeHiddenFields` / `includeVariables`) behind higher permission, and inline **live** validation in the editor as the author types — the server enable + send checks close the egress hole, and the author now gets an actionable reason from both the dry run and the enable `422`. Also out of scope, and worth its own ticket: survey **Follow-Ups** (`apps/web/modules/survey/follow-ups/lib/email.ts`) has the same arbitrary-`to` + response-data shape with no allowlist, in GA code.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2029/workflows-send-email-action-allows-arbitrary-external-recipient-silent

## How this was tested

- `pnpm vitest run` on the affected suites ✅
  - `packages/workflows` full suite — 218 passed (12 files). `workflows.handlers.test.ts` 80 passed, incl. enable rejecting an external literal recipient (asserting the `detail` text), allowing a member, skipping the check for a respondent-field recipient, and the three new dry-run cases: reports `recipient_not_allowed` instead of testing green, skips the check for a respondent-field recipient, skips it for a non-executable definition. `spec-drift.test.ts` 14 passed, incl. the new problem-code enum guard.
  - `apps/web` `context.test.ts` + `process-workflow-run-job.test.ts` — 52 passed (adapter allowlist resolves membership case-insensitively; runner blocks an off-allowlist literal send, fails closed on an unresolvable org, still delivers respondent-field recipients, and the two new cases: no membership query when no step has a literal recipient, exactly one when a step does).
  - Wider `apps/web` sweep over `modules/ee/workflows`, `app/api/v3/workflows`, `modules/email`, `lib/organization` — 469 passed (46 files).
- `pnpm turbo typecheck --filter=@formbricks/workflows --filter=@formbricks/web` ✅ 0 errors.
- `eslint --no-fix` + `prettier --check` on all touched files ✅ clean.
- Not run locally: the render-parity **integration** test (`*.integration.test.ts`) — its global setup needs a live Postgres; the change there is a mechanical mock addition and runs in CI.

## QA / Test Plan

**How to test**

- [ ] Create a workflow (`response.completed` trigger) with a `send_email` node whose `to` is an **external** literal address (`attacker@external-evil.example`), then **Test** it (canvas button, v3 `test` endpoint, or MCP) → `ok: false` with a `recipient_not_allowed` problem on `definition.nodes.config.to` naming the address.
- [ ] `enable` that same workflow → **422** `workflow_not_executable`; the error message names the recipient rule (not just "definition is not executable"); workflow stays draft/disabled.
- [ ] Change `to` to the email of an **organization member** → Test reports `ok: true` and enable **succeeds** (workflow goes live).
- [ ] Set `to` to a survey **email-question / hidden-field element id** (respondent's own email) → Test reports `ok: true` and enable **succeeds** regardless of the member allowlist.
- [ ] Enable a workflow with a member recipient, then **remove that member** from the org (or deactivate them), then fire a completed response → the send step **fails** (no external email); run is marked failed, no email delivered.
- [ ] Fire a completed response for a live workflow whose `to` is a respondent field → the respondent still receives their email.
- [ ] Confirm delivery of the original repro: a live workflow pointing `to` an external mailbox can no longer be enabled, and even a pre-existing one no longer sends externally.

**Preconditions / test data**

- An organization with ≥2 members and a survey with a contact-info/email element; a write API key (or MCP write scope).
- SMTP configured (`MAIL_FROM`) to observe actual delivery, or inspect the run/step status for the send outcome.

**Risks & regressions**

- Tightens `send_email`: workflows that legitimately emailed a **non-member** literal address (e.g. a shared external distro) can no longer be enabled and will fail to send. This is the intended security posture; such cases must use a member address or a respondent field. Feature is pre-release on `epic/workflows-v1`, so no migration of existing live workflows is required.
- The dry-run contract gains a `recipient_not_allowed` problem code. Additive, but a client that exhaustively switches on `ZWorkflowTestProblemCode` will see a new variant, and a workflow that previously tested `ok: true` with an external literal recipient now correctly tests `ok: false`.
- No change to `from` handling, the email template, response-data attachment, or the recipient-resolution logic itself.

- none


packages/workflows/src/handlers/context.ts
```diff
@@ -74,5 +74,16 @@ export interface WorkflowApiContext {
     surveyId: string;
     endingCardIds: string[];
   }) => Promise<TriggerSurveyCheck>;
+  /**
+   * Injected recipient allowlist check (the adapter queries the organization's members) so the
+   * package stays user/organization-agnostic. `send_email` actions may address a literal email; to
+   * prevent a workflow from silently forwarding response data to an arbitrary external inbox
+   * (ENG-2029), enable/test confirm every literal recipient belongs to the workspace's organization.
```
xernobyl, line 81: **[P2]** This says enable **and test** confirm every literal recipient, but the check was only added to `enable` — `testWorkflow` (`workflows.handlers.ts:489`) validates executability and the trigger survey, then returns without consulting the allowlist. Same claim in `apps/web/app/api/v3/workflows/lib/context.ts:96` ("Enable/test use it"), and `testWorkflow`'s own docstring still says it "Reuses the same executability + survey checks as `enable`".

Concrete consequence: author sets `to: attacker@external-evil.example`, hits **Test** in the canvas (`workflow-canvas.tsx:203`) → `ok: true`, zero problems → hits Enable → hard 422. The dry run exists to catch exactly this before going live.

Either add a `recipient_not_allowed` code to `ZWorkflowTestProblemCode` and collect it in `testWorkflow` (with the matching spec update), or narrow both comments to enable-only so the contract and the docs agree.

packages/workflows/src/handlers/workflows.handlers.ts
```diff
@@ -391,6 +419,21 @@ export const createWorkflowsHandlers = (service: WorkflowsService): WorkflowsHan
...
+      // an arbitrary external inbox on every completed response (ENG-2029). Respondent-field
+      // recipients resolve at send time and are not checked here.
+      const literalRecipients = collectLiteralRecipientEmails(executable.data);
+      if (literalRecipients.length > 0) {
+        const { disallowedEmails } = await ctx.verifyRecipientsAllowed({
+          workspaceId: loaded.workspaceId,
+          emails: literalRecipients,
+        });
+        if (disallowedEmails.length > 0) {
+          throw new WorkflowNotExecutableError(buildDisallowedRecipientParams(disallowedEmails));
```
xernobyl: **[P2]** This throws with `WorkflowNotExecutableError`'s default detail — *"The workflow definition is not executable."* — and the editor's `getV3ApiErrorMessage` (`apps/web/modules/api/lib/v3-client.ts:46-49`) returns only `detail`, dropping `invalid_params` entirely. So the author gets a toast that never mentions the recipient, which is the one thing they need to know to fix it.

That matters because the PR deliberately scopes out editor live-validation on the grounds that "the editor will show the `422` on enable" — it shows the 422, just not the reason. `WorkflowNotExecutableError` already takes a message as its second arg, so this is a one-liner:

```ts
throw new WorkflowNotExecutableError(
  buildDisallowedRecipientParams(disallowedEmails),
  "A send_email recipient is not a member of this organization."
);
```

The per-field `reason` strings stay where they are for API/MCP callers.

apps/web/modules/ee/workflows/lib/runner/process-workflow-run-job.ts
```diff
@@ -452,8 +475,13 @@ const loadRunEmailContext = async (
 
   const organization = await getOrganizationByWorkspaceId(workspaceId);
   const logoUrl = organization?.whitelabel?.logoUrl ?? "";
+  // Fail closed: no resolvable organization means an empty allowlist, so a literal external
+  // recipient is rejected rather than allowed through unchecked.
+  const allowedRecipientEmails = organization
```
xernobyl: **[P3]** This unbounded `membership.findMany` runs on **every** workflow run, including the common respondent-field case where the allowlist is never consulted — one wasted query per completed response, and the result set grows with org size.

`planExecutableSteps` has already run by the time `loadRunEmailContext` is called, so the steps are available to gate on:

```ts
const needsAllowlist = steps.some((s) => isLiteralEmailRecipient(s.node.config.to));
```

and skip the fetch when false (or make the set lazy). The fail-closed behaviour is unaffected — with no literal recipient there's nothing to check.

apps/web/modules/email/lib/survey-response-email.ts
```diff
@@ -153,6 +153,10 @@ export const buildSurveyResponseEmailHtml = async ({
 /** Outcome of resolving a `to` value (literal email or question/hidden-field id) against a response. */
 export type ResolveRecipientResult = { ok: true; email: string } | { ok: false; error: string };
 
+// Re-exported from `@formbricks/workflows` so the enable-time gate and this send-time path share a
+// single literal-vs-field classifier that cannot drift (ENG-2029).
+export { isLiteralEmailRecipient } from "@formbricks/workflows";
```
xernobyl: **[P3]** This re-export is pure indirection: nothing in this module uses `isLiteralEmailRecipient`, and the only consumer — `process-workflow-run-job.ts` — already imports from `@formbricks/workflows` a few lines up in the same import block. Import it directly there and drop the re-export, so the email module doesn't own a workflows-package symbol it has no use for.

xernobyl (commented):
The security fix itself is correct, and I checked the things that could have quietly undone it rather than assuming:

- `enable` really is the single choke point — `patch` refuses definition edits while `enabled`, `duplicate` and `unarchive` both land in `draft`, and the editor and MCP both route through `buildWorkflowApiContext` (the only construction site).
- No TOCTOU: `enableWorkflow` snapshots the **validated** `executable.data` into `workflowVersion`, and the runner prefers that snapshot, so a concurrent draft `patch` can't slip an unchecked definition into the published version.
- The two literal-vs-field classifiers can't disagree — `isLiteralEmailRecipient` and `resolveResponseRecipient` both key off `z.email()`.
- `to` is the only delivery target: no `cc`/`bcc` on the config, and `replyTo` doesn't receive the mail.
- Runner fails closed on an unresolvable org, and `isActive: true` correctly drops deactivated members.

Four inline findings below (2 × P2, 2 × P3) — none of them weaken the gate. I've skipped the duplicate-predicate point in `resolveResponseRecipient` and the `__mocks__` nit, since CodeRabbit already raised both (on the first commit — its "centralize the classifier" Major is now largely addressed by `e1d8cc5`; the `__mocks__` one is safe to dismiss, there's no shared mock module for `@/lib/organization/service` and inline `vi.mock` of it is the convention across dozens of test files).

Three more that can't be anchored inline because the files aren't in the diff:

**[P3] v3 spec drift.** `docs/api-v3-reference/src/paths/api_v3_workflows_{workflowId}_enable.yml` enumerates every *other* enable-time validation (acyclic graph, reachability, Scope 1 node types, survey/ending-card existence) but not this one. `WorkflowSendEmailActionConfig.yml`'s `to` description still reads "either a literal email address (e.g. a teammate) or…" with no org-member rule, and the Zod `to: z.string()` has no `.describe()` at all — so an MCP agent has no way to learn the constraint before tripping it.

**Worth its own ticket: survey Follow-Ups has the identical egress hole, in GA code.** `apps/web/modules/survey/follow-ups/lib/email.ts` sends to an arbitrary `to` with the same response-data attachment and no allowlist. If ENG-2029 is real for pre-release workflows, it's real for shipped Follow-Ups.

**Minor, on the reasoning rather than the code:** "a respondent-field recipient is, by definition, the respondent's own address" is a bit stronger than reality — a hidden-field `to` is whatever the link-crafter put in the URL. It holds in substance today because the mail only carries that same submission's data, but it stops holding if `includeVariables`/`includeHiddenFields` ever pull in cross-response data.

packages/workflows/src/handlers/context.ts
Dhruwang: Took the first branch rather than narrowing the comments — the dry run is the right place to catch this, and a "tests green, then hard 422" sequence is worse than a slightly wider `test` surface. `df6b9dc` adds `recipient_not_allowed` to `ZWorkflowTestProblemCode` and collects it in `testWorkflow`, gated on `executable.success` like the survey check so a malformed definition doesn't reach it. The field path and per-recipient message are now shared constants with `enable`, so the two reports can't drift. Spec, `WorkflowTestResult.yml` enum, and `testWorkflow`'s docstring updated to match, plus a drift test pinning the inline enum to the shared schema.

The other three are in the same commit as suggested. On the spec-drift note: `to` now carries a `.describe()` (it had none) and the org-member rule is in the enable/test descriptions and `WorkflowSendEmailActionConfig.yml`. Follow-Ups is left alone and called out in the PR description as needing its own ticket — agreed it's the same hole in GA code. I also softened the "by definition, the respondent's own address" line in the description, since you're right that a hidden-field `to` is whatever the link-crafter put in the URL.

---

Dhruwang:
@xernobyl , Thank you for the review. I addressed all your comments. Can you please recheck?

packages/workflows/src/contracts/resources.ts
```diff
@@ -82,6 +82,7 @@ export const ZWorkflowTestProblemCode = z.enum([
   "definition_not_executable",
   "survey_not_found",
   "ending_card_not_found",
+  "recipient_not_allowed",
```
xernobyl, line 85: **[P2]** Adding a problem code obliges a matching UI string, and this one doesn't have one — so the dry run detects the non-member recipient but the editor won't say so.

`workflow-test-result-dialog.tsx:35-39` localizes each problem **by `code`** and deliberately never renders the API's `message`. Its `problemMessages` is typed `Record<string, string>`, so TS doesn't enforce exhaustiveness, and `recipient_not_allowed` falls through to `?? t("…test_problem_generic")` — *"This workflow has a configuration problem."* — plus the raw field path.

That's the same failure mode the explicit 422 `detail` just fixed on the enable path, so right now Enable names the cause and Test doesn't. "Validate Translation Keys" can't catch it either: it checks that used keys exist, not that every code has one.

Three small pieces:

1. `recipient_not_allowed: t("workspace.workflows.test_problem_recipient_not_allowed")` in the map;
2. the key in `apps/web/locales/en-US.json` **only** — the other locales are Lingo.dev-generated via `pnpm i18n:web:generate`, never hand-edited;
3. type the map as `Record<TWorkflowTestProblemCode, string>` so the next code added here is a compile error instead of a silent generic fallback.

docs/api-v3-reference/src/paths/api_v3_workflows_{workflowId}_test.yml
```diff
@@ -3,9 +3,10 @@ post:
   summary: Dry-run (validate) a workflow
   description: >
     Validates a workflow without running it or causing any side effect. It checks that the
-    workflow's current definition is executable and that the trigger's referenced survey and ending
-    cards still exist in the workspace. **No** workflow run is created and **no** email or webhook is
-    sent.
+    workflow's current definition is executable, that the trigger's referenced survey and ending
+    cards still exist in the workspace, and that every `send_email` action addressing a *literal*
```
xernobyl, line 7: **[P3]** Since you updated this description, the MCP `test_workflow` tool's equivalent is now the odd one out — `apps/web/modules/mcp/tools/workflows.ts:241` still says it validates "its live definition would execute and resolve the trigger's survey + ending cards", with no mention of the recipient check. Cheap to align given every other description got there.

xernobyl (commented):
Re-reviewed at `e71c755`. All four findings are properly fixed, and I checked the substance rather than the commit messages:

- **`test` gate** — `recipient_not_allowed` collected in `testWorkflow`, with the field path and message hoisted to shared constants so enable's `invalid_params` and test's `problems` can't diverge. Both stale comments corrected.
- **422 detail** — explicit detail passed, with a test asserting the exact string.
- **Lazy allowlist query** — this was the one that could have quietly reopened the hole, so I checked the predicate: `steps.some(s => isLiteralEmailRecipient(s.node.config.to))` evaluates the same predicate on the same `config` object that `sendResolvedEmail(step.node.config, …)` later re-tests, over the same `steps` array, and `TWorkflowExecutableStep` is always a send-email step so `.to` is always a string. Fail-closed intact, and the negative test pins it.
- **Re-export** — gone; `survey-response-email.ts` is untouched by the PR now.

The spec updates go beyond what I raised, and the spec-drift test pinning the problem-code enum to the YAML is a nice touch — it prevents the next instance of this class rather than just this one. Bundle diff contains exactly the source edits, and the "Lint source & verify bundle freshness" job is green.

Two follow-ons below, one of which is a real gap the `test` fix opened up. Neither is a blocker for the security posture.

xernobyl (approved):
Approving at `5196e35`. Every finding from both rounds is addressed, and the core security work is unchanged from the head I reviewed in full — `workflows.handlers.ts`, `process-workflow-run-job.ts`, `handlers/context.ts`, `recipients.ts` and the spec files are byte-identical to `e71c755`.

What I verified rather than assumed, for the record:

- **`enable` is the single choke point** — `patch` refuses definition edits while `enabled`, `duplicate` and `unarchive` both land in `draft`, and the editor and MCP share the one `buildWorkflowApiContext` construction site.
- **No TOCTOU** — `enableWorkflow` snapshots the *validated* definition into `workflowVersion`, and the runner prefers that snapshot, so a concurrent draft `patch` can't publish an unchecked definition.
- **The lazy allowlist query can't skip a check it should run** — `steps.some(s => isLiteralEmailRecipient(s.node.config.to))` evaluates the same predicate on the same `config` object that `sendResolvedEmail(step.node.config, …)` re-tests, over the same `steps` array, and `TWorkflowExecutableStep` is always a send-email step. Fail-closed intact.
- **The classifiers can't drift** — `isLiteralEmailRecipient` and `resolveResponseRecipient` both key off `z.email()`.
- `to` is the only delivery target (no `cc`/`bcc`; `replyTo` doesn't receive the mail), and `isActive: true` correctly drops deactivated members.

The spec work went beyond what I raised, and the spec-drift test pinning the problem-code enum to the YAML prevents the next instance of that class rather than just this one.

**One informational note so nobody re-adds it:** `bdfbde53`'s localization payload (`test_problem_recipient_not_allowed` across all 16 locales + `i18n.lock`, plus the dialog wiring) was reverted by the `epic/workflows-v1` merge, and that resolution is correct — ENG-2030 deleted `workflow-test-result-dialog.tsx`, dropped the whole `test_problem_*` key family, and removed `testWorkflow` from the API client, so the key would now be an orphan. I checked for dangling references and there are none. Net effect: `recipient_not_allowed` is live for API/MCP callers but unreachable from the editor, since client-side live validation can't check organization membership. That's fine — the enable path surfaces *"A send_email recipient is not a member of this organization."* in the toast, which is exactly what the PR description already promises.

Two things for separate tickets, neither blocking: survey Follow-Ups has the identical arbitrary-recipient egress surface in GA code (`modules/survey/follow-ups/lib/email.ts`), and CodeRabbit's still-open point about `resolveResponseRecipient` inlining the literal-email predicate is worth folding in whenever that file is next touched.

PR #8650, opened 2026-07-27
feat: add feedbackRecord MCP tools to the XM Suite MCP server

xernobyl:
## What & why

The Hub was going to get its own MCP server, but a second server only creates ambiguity about which one an agent should reach for — and the Hub's API is good enough on its own. So feedback records become tools on the **existing** Formbricks MCP server instead.

Ten new tools, behind new `feedbackRecords:read` / `feedbackRecords:write` scopes:

| Tool | Scope | Purpose |
| --- | --- | --- |
| `list_feedback_datasets` | `feedbackRecords:read` | discovery — a workspace's feedback datasets |
| `list_feedback_records` | `feedbackRecords:read` | cursor-paginated records, with the Hub's full filter set |
| `count_feedback_records` | `feedbackRecords:read` | how many records match a filter set, without fetching them |
| `get_feedback_record` | `feedbackRecords:read` | one record by Hub UUID |
| `search_feedback_records` | `feedbackRecords:read` | semantic search — match by meaning, not keywords |
| `find_similar_feedback_records` | `feedbackRecords:read` | nearest neighbours of one record |
| `create_feedback_record` | `feedbackRecords:write` | create a record (audit-logged) |
| `create_feedback_records` | `feedbackRecords:write` | create up to 50 records in one call (audit-logged per record) |
| `update_feedback_record` | `feedbackRecords:write` | correct a record's values (audit-logged with both sides) |
| `delete_feedback_record` | `feedbackRecords:write` | permanently delete one record (audit-logged) |

The two searches return scored matches — `feedback_record_id`, `score`, `field_label`, `value_text` —
rather than hydrated records, so a whole result set costs one Hub call and the caller can pull the full
record for the hits it cares about. `limit` (1–100) and `minScore` (0–1) are validated here rather than
passed straight through, because the Hub silently coerces out-of-range values to its own defaults; our
`minScore` default is 0.5 against the Hub's 0.7, favouring recall since the caller sees the matched text
and can judge relevance itself.

The tools are thin delegations to a new shared operations layer (`app/api/v3/feedbackRecords/lib`), matching how the survey tools wrap the v3 survey operations. That layer is the single **tenant-isolation choke point**: the Hub tenant is the id of the feedback dataset assigned to the caller's workspace, and it is always resolved server-side from `workspaceId` — no tool input accepts a tenant/dataset id for writing, and the create payload is built as an explicit allowlist so one can't be smuggled in. Outward-facing, that id is called `dataset_id` (the Hub's own `tenant_id` never appears in a response). It reuses the same Hub service, dataset resolver and `feedbackDirectories` entitlement check as the Unify Feedback UI, so the two can't disagree about who may read what.

Four things worth calling out beyond the tools themselves:

- **The Hub's `GET /v1/feedback-records/{id}` is not tenant-scoped.** `get_feedback_record` therefore verifies the fetched record's dataset against the caller's datasets, and returns the *same* generic 403 for a cross-tenant hit and for not-found — otherwise record ids become an existence oracle across tenants.
- **Authenticating no longer requires `surveys:read` specifically.** Advertising `feedbackRecords:*` as an independent grant group while still demanding `surveys:read` at the door would reject a legitimate feedback-only client. Any resource scope now gets you in; the per-tool guards decide what you can reach. `list_workspaces` accepts either read scope, since it's the only way to discover the `workspaceId` every feedback tool needs.
- **Errors are as actionable as they can safely be.** The Hub enforces content rules we don't duplicate (NUL bytes, its own limits) and reports them as 400 — `invalid_params` and the problem detail now pass through for 400/422, bounded to 20 params / 512 chars. A duplicate `(submission_id, field_id)` or in-progress tenant purge (409) and an oversized body (413) map through as themselves, so an agent doesn't retry-loop against a healthy service. Outages, rejected Hub credentials and upstream 5xx still collapse to a generic 502 and are only logged.
- **We validate what the Hub doesn't.** The Hub accepts a record with no value at all — and MCP strips unknown keys before our schema sees them — so a mistyped field name (`valueText`) would have stored a permanently empty record and reported success. `create_feedback_record` now requires *a* value appropriate to `field_type` and names the expected field in `invalid_params`, and `update_feedback_record` rejects a `value_*` the stored `field_type` does not accept; `metadata` is bounded below the Hub's body cap.
- **The outward-facing name is `dataset_id`, not `tenant_id`.** "Tenant" is Hub-internal vocabulary; the product, its UI and its docs all say Feedback **Dataset**, so that is what the tools speak — inputs take `datasetId`, responses emit `dataset_id`, and the discovery tool is `list_feedback_datasets`. Same value (a `FeedbackDirectory` id), renamed at the boundary only: the serializer is the single outbound mapping point and internal code keeps the `feedbackDirectory*` names, so the Prisma model and the taxonomy routes are untouched.
- **Dead ends tell you what to do.** An agent has no upgrade prompt to fall back on, so "no feedback dataset is assigned to this workspace" now names who can fix it and where (an org owner or manager, under Settings → Organization → Feedback Datasets), and the unlicensed 403 says it needs an Enterprise plan or licence.
- **The access + entitlement gate is shared.** This PR initially added a second copy of the workspace-access + `feedbackDirectories` pair alongside the one the Unify taxonomy surface already had; both now call `requireUnifyFeedbackWorkspaceAccess` in `app/api/v3/lib/feedback-access.ts`, keeping one implementation of the gate. One deliberate behaviour change falls out of that: the shared unlicensed 403 detail is now actionable, so the taxonomy routes return that new wording too (no consumer reads it — the taxonomy UI never surfaces upstream detail).

**Filtering and counting.** `list_feedback_records` originally exposed 4 of the Hub's 12 list parameters;
it and `count_feedback_records` now take the full set — `source_id`, `field_id`, `field_group_id`,
`submission_id`, `user_id` and `value_id` join `source_type`, `field_type`, `since` and `until`. The two that
were actually missed were `user_id` ("everything this customer told us") and `submission_id` ("the rest of
this submission", the natural follow-up to any single record). The Hub documents its `/count` endpoint as
accepting the same parameters as list, so both operations share one mapper — a count can't end up
describing a different set than the list it is supposed to be counting. `count_feedback_records` returns
only the total and the dataset it came from, which is the point of having it: an agent can report "42
responses" without any end-user text entering the conversation.

**Update.** `update_feedback_record` covers the Hub's mutable set — `value_text`, `value_number`,
`value_boolean`, `value_date`, `value_id`, `user_id`, `language`, `metadata` — with the schema `.pick()`ed
from the create fields so bounds can't drift between creating and correcting a record. A record's
**provenance is immutable** (source, question, submission, `collected_at`): correcting those means deleting
and recreating, and the allowlist builder is what enforces it — those fields simply have nowhere to go. The
derived enrichment fields aren't accepted from callers either; the Hub owns them.

`PATCH /{id}` is IDOR-shaped upstream like get/delete/similar, so it goes through the same ownership guard —
and the record that guard returns *is* the pre-update state the audit log needs, so the check costs nothing
extra. Audited with both sides, because an edit is only reviewable if the previous value is recorded.

Three caller-facing behaviours came out of reading the Hub's contract and its update SQL. Each fails
silently, so each is in the tool description (where an agent reads it) as well as the handbook:

- **`metadata` is replaced, not merged** (`metadata = $n`), so adding one key without resending the others
  wipes them. Verified live — a `{tier}` patch removed the record's existing `device` and `plan` keys.
- **Editing `value_text` clears the derived `sentiment`, `sentiment_score`, `emotions`,
  `value_text_translated` and `translation_lang_key`** and re-queues enrichment, so the response comes back
  without them. That means "being recomputed", not "none" — an agent that reads it the other way will report
  the wrong thing.
- **Editing `value_text` re-queues the embedding**, so semantic search follows the edit asynchronously, and
  clearing a record's text removes its embedding entirely. Verified live: after an edit the old topic stopped
  matching and the new one matched at 0.80.

**Batch create.** The Hub has no bulk-create endpoint — its only bulk write is the delete-by-user erasure
path — so `create_feedback_records` fans out to one create per record, in parallel, capped at 50. The cap
is an amplification bound as much as a payload one: one authorized request must not become an unbounded
burst of upstream writes. Validation is all-or-nothing so a malformed batch never leaves half of itself
stored (`invalid_params` names the index, e.g. `records.3.value_text`); partial *upstream* failure is
reported per index so only those records need retrying; and if nothing could be created the real upstream
failure is returned, because an empty success would read as "there was nothing to do". Each record actually
created gets its own `created` audit event — N creations are N events.

Three more things about the tools added in the second round:

- **Delete and similar are IDOR-shaped upstream, by design.** The Hub's `DELETE /{id}` and
  `GET /{id}/similar` take a bare record id and derive the tenant from the stored record — its own source
  comments say the product gateway owns record-level authorization (ENG-1289). So both go through the same
  retrieve-then-assert guard as `get_feedback_record`, now extracted as `requireOwnedFeedbackRecord`: a
  foreign id is refused *before* anything is deleted or any neighbour is read, with the same generic 403 as
  an unknown id. Without that, `similar` would be a cross-tenant read primitive and `delete` a cross-tenant
  write primitive.
- **Two failure modes that would otherwise be unactionable.** Embeddings are optional in the Hub, so on
  most self-hosted installs both searches fail — a bare 503 becomes "set `EMBEDDING_PROVIDER` and
  `EMBEDDING_MODEL` on both the Hub API and the Hub worker". That wording is passed in by the two search
  operations only; every other operation's 503 names no subsystem, so a plain Hub outage on a list or a
  create does not send an operator after an unrelated setting. And because embedding is asynchronous, a
  record can exist without one; since we've already proven ownership by then, the Hub's 404 can only mean
  "not embedded yet", so it is reported as a retryable 409 rather than the generic 403 — which would have
  told the caller it lacks access to a record it owns.
- **Delete is deliberately narrow.** One record per call, `destructiveHint: true`, write scope, and the
  pre-delete record is captured as the audit log's `oldObject` — the only trace left afterwards. The Hub's
  bulk-delete and delete-by-user (GDPR erasure) endpoints stay off the agent surface; their blast radius
  doesn't belong behind a tool call.

Renamed the MCP server from `formbricks-v3-surveys` to `formbricks-xm-suite`, since it now covers more than surveys.

`operations.ts` was carrying three concerns by this point, so it is split the way the rest of the v3 layer
already does it (`unify-feedback/taxonomy/lib/access.ts`): `lib/access.ts` for tenant resolution and the
ownership guard, `lib/errors.ts` for the Hub-error mapping, `lib/operations.ts` for the operations. That
refactor is its own commit (`refactor: extract feedback-record access and error mapping`, no behaviour
change) so the review can separate moved code from new code.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1260/add-feedbackrecord-mcp-actions-to-the-existing-formbricks-xm-suite-mcp

## How this was tested

- **Automated:** `vitest run` across the touched surface (`app/api/v3`, `modules/mcp`, `modules/hub`, `app/api/mcp`, `app/.well-known`, `modules/auth/lib`, `lib/feedback-source`, `modules/ee/unify-feedback`) → **93 files, 1283 tests ✅**
- **Tests added** (new files in **bold**):
  - `app/api/v3/feedbackRecords/lib/operations.test.ts` — auth/entitlement denial, dataset auto-resolve, cross-tenant 403 with *identical* bodies, tenant injection ignoring a body-supplied tenant id, the `dataset_id` rename, Hub error mapping, 4xx relay, no 5xx leak. Plus, for the later rounds: the cross-tenant guard on delete *and* similar (including "the record still exists" after a refused delete), post-ownership 404 → embedding-pending 409, 503 → the actionable configuration message, locally-rejected `limit`/`minScore` bounds, tenant injection on search, the full filter set reaching the Hub under its own parameter names, count returning no record content, batch partial failure per index, batch validation being all-or-nothing, and that a batch does **not** implicitly group records into one submission.
  - **`app/api/v3/feedbackRecords/lib/errors.test.ts`** — the relay bounds (512-char detail, 20 `invalid_params`), no 5xx detail, no 401/403/404 detail, and the unexpected-throw mapping that no operation test reaches.
  - `app/api/v3/feedbackRecords/lib/serializers.test.ts` — pins the whole record DTO allowlist, so a silently dropped field fails, plus the shared match serializer.
  - `modules/mcp/tools/feedback-records.test.ts` — registration order and annotations (only `delete` destructive), scope gating per tool, delegation, and the audit lifecycle including one event per created record in a batch.
  - `modules/hub/service.test.ts` — the count and similar wrappers, and that the Hub's RFC 9457 problem members survive on delete/search (the reason those wrappers changed).
  - `modules/hub/utils.test.ts` — problem-body parsing incl. malformed input.
  - Extended the existing MCP route, MCP auth, workspaces, protected-resource-metadata and OAuth-scope-label suites.
- **Lint / types:** `eslint` clean on every changed file; `tsc --noEmit` reports no errors in any changed file (the repo's pre-existing Prisma/Node typecheck noise is unchanged).
- **Verified end-to-end against a running Hub** (local dev stack, real API keys, records really written to and read from the Hub):
  - `tools/list` → all 17 tools with the expected schemas and annotations
  - `list_feedback_datasets` → the workspace's dataset; `list_feedback_records` → that dataset's records, each `dataset_id` equal to that dataset id
  - `create_feedback_record` → 201; the dataset **server-injected** and `submission_id` generated; the record then appears in `list_feedback_records`
  - Full lifecycle: create → get → delete (via the Hub API — there is no delete tool) → get again returns the generic 403, and the record is gone from `list`
  - **Cross-tenant guard, for real:** created a record in a *different* tenant directly in the Hub, then fetched it through `get_feedback_record` with our own workspace → 403, with a body byte-identical to the not-found 403 (only `requestId` differs). Our logs distinguish the two cases; the caller cannot
  - **Tenant smuggling defeated:** passing `"tenant_id":"attacker-tenant-9999"` in the create arguments produced a record in *our* dataset
  - **Hub 4xx relay:** a NUL byte in `value_text` surfaced `invalid_params: [{name: "value_text", reason: "must not contain NULL bytes"}]` rather than a generic message
  - Cursor pagination round-trip (`limit:1` then that `nextCursor` → a different record); read-only key blocked from `create` while reads work; workspace with no dataset → 422; a dataset belonging to another workspace → 403; unknown workspace → 403
  - Server logs for these paths carry only `requestId`/`workspaceId`/`statusCode`/`hubStatus`/`hubCode` — no record content, no keys
  - Smoke testing also disproved an assumption this PR originally shipped with: the Hub accepts a `field_type`/`value_*` mismatch (and a record with no value at all), which is why that validation now lives here
- **The search/delete round was re-verified against the same running Hub** (records really embedded by `hub-worker`, using `google` / `gemini-embedding-001`):
  - semantic search for "users having trouble paying" ranked the two checkout/payment records above an unrelated crash report, best-first (0.707 / 0.700 / 0.588)
  - `similar` on the checkout record returned the payment record at 0.943 and **never the anchor itself** — confirmed in the Hub's SQL too (`e.feedback_record_id != $4`)
  - `similar` with `min_score=1.1` returned 0 rows with no error, which is exactly the silent coercion our schema now rejects locally
  - delete → 204, the record then 404s and is gone from `list`; fixtures purged afterwards
  - measured score distributions differ sharply between the two searches — query→record landed at 0.59–0.71, record→record at 0.80–0.94 — so no single threshold is "correct"; see the note below
- **Update was verified live end to end** (against the same pinned Hub): a partial patch changed only the fields sent and left `user_id`/`metadata` untouched; a `{tier}`-only metadata patch **wiped the existing keys**, confirming the replace semantics; a patch carrying `source_type`, `field_id`, `submission_id`, `collected_at`, `tenant_id` and `sentiment` changed the text and **silently dropped every one of them** (provenance re-read afterwards and unchanged); an empty patch → 422; a read-only key → 403; a **cross-tenant record → 403 with its text byte-for-byte unchanged afterwards**; and after editing the text, semantic search stopped matching the old topic and matched the new one at 0.80. Sentiment/emotions enrichment isn't configured on my local Hub, so the enrichment-clearing behaviour is taken from the Hub's documented contract and its `clearedEnrichmentFields` source rather than observed.
- **Count, filters and batch create were verified against the pinned Hub too**, after confirming the local container's digest matches the digest `charts/formbricks/values.yaml` pins (so `/count` really is in the version we ship — the 0.8.1 tag's HEAD commit *is* the PR that added it):
  - every new filter narrows correctly (`user_id`, `submission_id`, `field_id`, `field_type`, `source_type`, a future `since` → 0), and **count matched `list().length` for every filter tried** — the invariant the shared mapper exists to protect
  - an invalid `since` surfaced the Hub's own `invalid_params: [{name: "since", reason: "must be in RFC3339 (ISO 8601) format"}]`; a dataset from another workspace → 403
  - batch create: 3 records in one call → `created: 3`; a batch whose second record duplicated an existing `(submission_id, field_id)` → `created: 1, failed: 1` with the Hub's own duplicate message at `index: 1`; an invalid record → the whole batch rejected **and a follow-up count proved nothing was written**; 51 records → rejected; read-only key → 403
  - confirmed empirically that the Hub's error body names an offending *parameter* but never echoes its *value* (`user_id=alice@example.com` did not appear anywhere in the 400 body, and `instance` is path-only), so the service logging the raw error cannot leak a filter value
- **Translations:** the two new OAuth consent-screen strings are present in all 14 locales and the lockfile is rebuilt, so the translation check passes. Note the whole `auth.oauth.scopes` block is currently English in every locale (it is frozen in `i18n.lock`), so these follow their siblings rather than introducing a lone translated pair — a later `pnpm i18n` with a working Lingo key will translate the block together.

## Review round 2 (@Dhruwang)

Four findings from exercising the tools end-to-end against a local Hub. All four were valid, all four
reproduced, and all four are fixed in 967522af2:

- **`update_feedback_record` skipped the `field_type` cross-check** that `create` enforces. `{value_number,
  value_id}` on a `text` record returned 200 with text, number *and* id set. Now checked after the ownership
  guard against the *stored* `field_type` — it is immutable, so it is not in the patch — as a 422 naming the
  offending field. `value_id` stays updatable: it is the value for a `categorical` record, and the per-type
  table rejects it everywhere else.
- **Relayed Hub detail leaked `tenant_id`.** A duplicate create came back as "a feedback record with this
  tenant_id, submission_id, and field_id already exists", pointing callers at a parameter this API does not
  have. Word-bounded rewrite to `dataset_id` over relayed detail *and* `invalid_params`.
- **A 503 blamed embeddings on every operation** (see above).
- **Filters were camelCase while bodies and responses are snake_case**, so filtering by a `user_id` read off a
  response silently no-op'd — and a dropped filter widens the result set while reporting success. The eight
  record-field filters now carry the names of the fields they filter; `workspaceId`/`datasetId`/`limit`/
  `cursor`/`minScore` stay camelCase, naming nothing in the record. Plus `.strict()`, which hardens the
  operation contract — though note the MCP SDK strips unknown keys against its own non-strict copy of the
  shape before an operation runs, so the *naming* is what fixes that path.

### Self-review after that round (9778bd169)

Re-reviewing the fixes above turned up two regressions they introduced, both now fixed with tests
confirmed to fail without them:

- The `field_type` cross-check **threw** on a record the Hub returned without a `field_type`
  (`VALUE_FIELD_BY_TYPE[undefined]` → `accepted.includes` → TypeError → generic 500), failing an otherwise
  valid patch. The SDK types the field as a required union, but `TV3FeedbackRecord.field_type` is optional
  and the serializer copies it only when present — so the codebase already treats absence as possible. Now
  guarded: with no table to check against, the patch goes through.
- The `tenant_id` → `dataset_id` rewrite ran **after** the 512-char slice, and `dataset_id` is a character
  longer — so a capped detail came back at 563. Rewritten before slicing.

## QA / Test Plan

**How to test**

All steps are HTTP `POST`s to `/api/mcp` with `Authorization: Bearer <api-key>`, `Content-Type: application/json` and `Accept: application/json, text/event-stream`. Body is JSON-RPC: `{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"<tool>","arguments":{…}}}`. Use `{"method":"tools/list","params":{}}` to list tools.

- [ ] `tools/list` → 17 tools, including `list_feedback_datasets`, `list_feedback_records`, `count_feedback_records`, `get_feedback_record`, `search_feedback_records`, `find_similar_feedback_records`, `create_feedback_record`, `create_feedback_records`, `update_feedback_record`, `delete_feedback_record`
- [ ] `list_feedback_datasets` with `{"workspaceId":"<ws>"}` → 200, the workspace's feedback dataset (`id`, `name`)
- [ ] `list_feedback_records` with `{"workspaceId":"<ws>"}` → 200, records for that dataset; `meta.limit` and `meta.nextCursor` present. `dataset_id` on each record equals the id from the previous step
- [ ] Pagination: repeat with `{"workspaceId":"<ws>","limit":1}` → one record + a `meta.nextCursor`; call again passing that `cursor` → the *next* record, not the same one
- [ ] Filters: `{"workspaceId":"<ws>","field_type":"text"}` → only text records; `{"workspaceId":"<ws>","since":"<ISO date in the future>"}` → empty list
- [ ] `get_feedback_record` with `{"workspaceId":"<ws>","feedbackRecordId":"<id from the list>"}` → 200, that record
- [ ] **Cross-tenant guard:** take a record id belonging to a *different* organisation's dataset and call `get_feedback_record` with your own `workspaceId` → **403** "You are not authorized to access this feedback record"
- [ ] **No existence oracle:** call `get_feedback_record` with a well-formed but non-existent UUID → **the same 403, byte-identical body** to the previous step (not a 404)
- [ ] `create_feedback_record` with `{"workspaceId":"<ws>","source_type":"call_notes","field_id":"note","field_type":"text","value_text":"Customer asked for SSO","field_label":"Call summary"}` using a `manage` key → 201 with the created record (a `submission_id` was generated); it then appears in `list_feedback_records` and in **Unify → Feedback data** in the UI
- [ ] **Tenant can't be smuggled:** repeat the create with an extra `"tenant_id":"<some other dataset id>"` in the arguments → still 201, and the created record's `dataset_id` is *your* dataset
- [ ] **Value must match the field type:** create with `{"field_type":"text"}` and no `value_text` (or with only `value_number`) → 422 naming `value_text` in `invalid_params`. Same for `nps`/`rating`/`number` → `value_number`, `boolean` → `value_boolean`, `date` → `value_date`. A `categorical` record may carry `value_text`, `value_id`, or both
- [ ] **Hub validation is actionable:** create with a `value_text` containing a NUL byte (0x00) → 400 whose body names the offending field in `invalid_params` ("must not contain NULL bytes"), not a generic "could not process the request"
- [ ] **Duplicate submission:** create twice with the same explicit `submission_id` + `field_id` → the second returns **409** (not 502), with a detail mentioning the duplicate
- [ ] **Oversized payload:** create with a `metadata` object larger than ~32 KB → 422 rejected locally (it never reaches the Hub)
- [ ] `search_feedback_records` with `{"workspaceId":"<ws>","query":"<a phrase describing something in your data, worded differently>"}` → 200, scored matches ordered best-first, `meta.datasetId` / `meta.datasetName` / `meta.minScore` present. Each `feedback_record_id` works in `get_feedback_record`
- [ ] **Search matches meaning, not keywords:** query with wording that shares no words with the record text (e.g. "trouble paying" against "the checkout page is confusing") → the record still comes back
- [ ] `find_similar_feedback_records` with a `feedbackRecordId` from the search results → 200, neighbours ordered best-first, and **the anchor record is not in its own results**
- [ ] **Similar respects the ownership guard:** an anchor id from *another* organisation's dataset → 403, and no neighbours in the body
- [ ] **Embedding pending:** create a record and immediately call `find_similar_feedback_records` on it → 409 saying the embedding is generated in the background (not a 403, not an empty list). Retry a moment later → 200
- [ ] **Bounds are rejected locally:** `minScore: 1.1`, `minScore: -1`, `limit: 999`, `limit: 0` or an empty `query` → 422 with `invalid_params`, and nothing reaches the Hub
- [ ] **No embedding model configured** (self-hosted without `EMBEDDING_PROVIDER` / `EMBEDDING_MODEL`): both search tools → 503 whose detail names both variables and both processes; the other feedback tools keep working. With the Hub stopped instead, a `list_feedback_records` 503 must **not** mention embeddings
- [ ] `delete_feedback_record` with `{"workspaceId":"<ws>","feedbackRecordId":"<id>"}` using a `manage` key → success with no content; the record then 404s/403s on `get_feedback_record` and is gone from `list_feedback_records` and from **Unify → Feedback data**
- [ ] **Delete respects the ownership guard:** a record id from *another* organisation's dataset → 403, **and that record still exists** afterwards (check it directly in the Hub or in the other org's UI)
- [ ] **Delete is audited:** the org's audit log has a `deleted` `feedbackRecord` event with the record id and the deleted content
- [ ] `count_feedback_records` with `{"workspaceId":"<ws>"}` → 200 with `data.count`, `data.dataset_id`, `data.dataset_name`, **and no record content**
- [ ] **Count agrees with list:** call `count_feedback_records` and `list_feedback_records` with the same filters → the count equals the number of records listed
- [ ] **Each new filter narrows:** `user_id`, `submission_id`, `source_id`, `field_id`, `field_group_id`, `value_id` each return a smaller, correct subset on both `list_feedback_records` and `count_feedback_records`; `{"since":"<a future date>"}` → count 0
- [ ] **Bad filter value:** `{"since":"not-a-date"}` → 400 whose `invalid_params` names `since` (the Hub's own message), not a generic error
- [ ] `create_feedback_records` with 3 valid records → success with `meta.requested/created/failed` and the created records in `data`; all three then appear in `list_feedback_records`
- [ ] **A batch is not a submission:** create 2 records in one call *without* `submission_id` → they get different generated ones (check `submission_id` on the returned records); repeat with the same explicit `submission_id` on both → both are stored under it, and `list_feedback_records` filtered by that `submission_id` returns both
- [ ] **Batch partial failure:** include a record duplicating an existing `(submission_id, field_id)` → the others are still created, `meta.failed` is 1, and `meta.failures[0]` gives that record's `index` and a reason
- [ ] **Batch validation is all-or-nothing:** include one record with a `field_type` and no matching value → the whole call fails with the index named in `invalid_params`, and a follow-up count/list shows **none** of the batch was written
- [ ] **Batch cap:** 51 records → rejected; 0 records → rejected
- [ ] `update_feedback_record` with `{"workspaceId":"<ws>","feedbackRecordId":"<id>","value_text":"corrected"}` → 200 with the new text; the fields you didn't send are unchanged
- [ ] **`metadata` is replaced, not merged:** on a record with `metadata: {"a":1,"b":2}`, patch `metadata: {"c":3}` → the record ends up with **only** `c`. (Intended, and warned about in the tool description — confirm the warning matches the behaviour.)
- [ ] **Provenance is immutable:** patch with `source_type`, `field_id`, `submission_id`, `collected_at` (and a `tenant_id` for good measure) → they are ignored, and a follow-up `get_feedback_record` shows them unchanged
- [ ] **Enrichment resets on a text edit:** patch `value_text` on a record that has `sentiment` → the response comes back **without** sentiment/emotions/translation, and they reappear once re-enrichment completes (needs enrichment configured on the Hub)
- [ ] **Search follows an edit:** patch `value_text` to a different topic, wait a moment, then `search_feedback_records` → the old wording no longer matches and the new one does
- [ ] **Cross-tenant update:** a record id from another organisation's dataset → 403, **and that record's value is unchanged** afterwards
- [ ] **Empty patch:** `update_feedback_record` with no updatable field → 422 saying at least one field is required
- [ ] **Least privilege:** with a `read`-only API key, `create_feedback_record`, `create_feedback_records`, `update_feedback_record` and `delete_feedback_record` → 403 insufficient scope; the six read tools still work
- [ ] **Entitlement:** as an org *without* the Enterprise feedback-directories entitlement, any feedback tool → 403 "Unify Feedback is not enabled for this organization"
- [ ] **Wrong workspace:** `list_feedback_records` with a `workspaceId` the key has no access to → 403
- [ ] **Explicit dataset:** `list_feedback_records` with a `datasetId` belonging to another workspace → 403
- [ ] **No regression on surveys:** `list_surveys`, `get_survey` and `create_survey` still behave as before with the same key (the authentication scope gate changed for all MCP tools)
- [ ] **No regression in Unify:** Unify → Feedback data still lists records and paginates, and Unify → Taxonomy still loads (both share code this PR touched)
- [ ] **OAuth (if exercising it):** the consent screen shows "Read feedback records" and "Create feedback records" as separate items
- [ ] **Client tool refresh:** an already-connected client needs to reconnect to see the three tools added in the second round — MCP tool registrations are captured once at module scope, so a running dev server keeps serving the old list until it restarts

**Preconditions / test data**

- An organisation on a plan/licence with the **feedback-directories (Unify Feedback) entitlement**, plus a second organisation without it for the entitlement check
- A **feedback dataset assigned to the test workspace**, and at least a few feedback records in it — e.g. a feedback source mapping a survey, then submit a couple of responses (text + a rating), or a CSV import
- API keys on the workspace: one with `read`, one with `manage`
- A record id from a **different** organisation's directory, for the cross-tenant checks (`get`, `similar`, `delete`)
- For the search tools: the Hub API **and** `hub-worker` configured with `EMBEDDING_PROVIDER` / `EMBEDDING_MODEL`, and a few text records that have had a moment to be embedded. Also worth a pass with embeddings *off*, for the 503
- Hub reachable by the app (`HUB_API_URL` / `HUB_API_KEY` configured — already required today)

**Risks & regressions**

- The MCP **authentication scope gate** changed for every tool, not just the new ones: a token holding any resource scope now authenticates. Re-check the existing survey tools with both API-key and OAuth clients, and that a token with no resource scope at all is still refused.
- `list_workspaces` now accepts either read scope — confirm it still returns only workspaces the caller can actually access.
- The **shared access gate** extraction touches the Unify taxonomy v3 routes; and the shared entitlement 403 detail is now actionable for both surfaces. Re-check taxonomy pages load and that a non-entitled org is still blocked.
- `listFeedbackRecords` in the Hub service now returns errors through the shared helper — re-check the Unify Feedback table's empty and error states. `deleteFeedbackRecord` and `semanticSearchFeedbackRecords` moved to the same helper in the second round: their `error` objects now carry the Hub's problem members as well, which the Unify UI paths reading `error.message` are unaffected by.
- **The shared tenant resolver changed.** `resolveWorkspaceFeedbackTenant` now normalises the dataset id it returns, instead of one operation trimming it at the call site. Every feedback-record tool goes through that resolver, so it is worth a look — it is contained to `feedbackRecords/lib/access.ts` and does not touch the taxonomy surface, which has its own access module.
- **`minScore` default is a judgement call, not a tuned value.** Measured on `gemini-embedding-001`, query→record scores clustered at 0.59–0.71 and record→record at 0.80–0.94, so 0.5 is permissive (an unrelated record scored 0.588) while the Hub's 0.7 would have dropped a genuine match sitting at 0.7001. We favour recall because the caller gets `value_text` and can filter, and because an over-strict threshold looks identical to "nothing matched". Easy to revisit per tool if QA finds the noise unhelpful.
- **Batch create amplifies one request into up to 50 upstream writes.** Bounded by the 50-record cap (enforced by both the MCP input schema and the v3 body schema) and by the existing 2 MiB request-body limit, and it needs the write scope — but it does mean one rate-limit token can carry 50 record creations. Worth a look during QA if the Hub shows write pressure.
- **`user_id` is now a filter.** It filters within the caller's own dataset only (the tenant is still injected, never accepted), so it exposes nothing new — but it does make "show me everything this person submitted" a one-call operation, which is worth knowing when reasoning about what an agent with read scope can assemble.
- **End-user feedback text now reaches the audit log.** `create` records the new record and `delete` records the deleted one — both include `value_text`, which is written by an end user and can contain anything. `redactPII` is key-name based (`email`, `name`, `token`, …), so it cannot scrub PII sitting *inside* a free-text value. This is inherent to auditing feedback records at all, it is org-scoped and EE/`AUDIT_LOG_ENABLED`-gated, and `delete` is the case where the content is the entire point of the entry — but it is a retention/erasure question worth a deliberate answer (a user erased via the Hub's delete-by-user path could still appear in Formbricks audit logs). Worth noting `update` is the *cheapest* of the three: it supplies both sides, so the handler stores `deepDiff(old, new)` — only the fields that changed, not a snapshot.
- **`metadata` replace-not-merge is the sharpest edge in this PR.** It is the Hub's semantics, not ours, and it is silent: an agent "adding a tag" drops everything else. Mitigated only by the tool description and docs. If QA finds agents getting it wrong, the fix would be to make the tool read-modify-write, which trades a race for the surprise.
- **Update is irreversible too** (the previous value survives only in the audit log), which is why it carries the same `destructiveHint` as `patch_survey`.
- **Delete is irreversible and now reachable by an agent.** Worth deliberate attention in QA: the ownership guard, the read-only-key refusal, and that the audit entry captures what was removed.
- **Breaking for already-connected OAuth clients — confirmed by manual test.** The oauth-provider plugin validates `/authorize` against the *client's registered* scopes, so once the protected-resource metadata advertises `feedbackRecords:*`, an existing DCR client fails outright with `invalid_scope: The following scopes are invalid: feedbackRecords:read, feedbackRecords:write`. It does **not** degrade to the old scope set, and re-consenting does not fix it — the stale `client_id` is reused, so the client has to re-register. Remedy: remove and re-add the MCP server (`claude mcp remove/add`), or delete the `oauthClient` row so the next connection triggers a fresh DCR. This needs a release note; it will hit every user with a connected MCP client on upgrade.

- none — no schema changes, and `HUB_API_URL` / `HUB_API_KEY` are already required. Non-English locales for the two new consent-screen strings are generated from `en-US` by the usual translation step.


xernobyl:
@coderabbitai — went through all seven. **All valid, all addressed** in f9103d6.

| # | Finding | What I did |
| --- | --- | --- |
| 1 | `datasetId` compared against the untrimmed directory id | **Real bug, fixed.** Since the normalised id is what we echo back as `dataset_id`, a caller replaying that value would have been 403'd on the next call. Both sides trimmed, plus a test for the round trip. |
| 2 | 413 never relays the Hub detail | **Fixed.** 413 is in `RELAYABLE_HUB_STATUSES` and the batch path already relayed it per record, so the whole-request branch was the odd one out — a single create/update never surfaced the Hub's actual size limit. Test added. |
| 3 | list/count don't validate their filters | **Fixed.** These were the only operations not validating their own input, which contradicts this layer's stated contract — it's transport-independent, so it can't assume an MCP schema screened the input — and an out-of-range `limit` reached the Hub, which silently coerces it rather than rejecting. Both validate now via one `parseQueryInput` helper (renamed from `parseSimilarityFilters`, no longer similarity-specific). Side benefit: `ZV3FeedbackRecordListFilters` was exported but unused, and now isn't. |
| 4 | the eight-key `.pick()` written twice | **Fixed** exactly as proposed — refined body derived from the raw shape. |
| 5 | no test for `writeHandler`'s throw path | **Fixed.** Added for both write handlers (single and batch). The invariant is "a failed mutation is never silently unaudited" and the catch branches upholding it were unguarded. |
| 6 | `feedbackRecords:write` scope row missing `update_feedback_record` | **Fixed — and thanks, this one was a miss on my side.** I'd attempted it in an earlier commit, but the edit matched an exact column padding that prettier had since reflowed, so it silently no-op'd; my verification then counted occurrences file-wide and passed on other rows. Fixed and verified against that row specifically. |
| 7 | `clientRegistrationDefaultScopes` duplicates `MCP_OAUTH_SCOPES` (outside diff) | **Fixed.** Verified the literal was identical in members *and* order before switching to `[...MCP_OAUTH_SCOPES]` — this list drives DCR, so a reordering would have been a behaviour change. Checked against `mcp-oauth-dcr.test.ts`, which exercises the advertised-scope → DCR → authorize chain that would break if this list drifted. |

No false positives in the run — #1 and #3 were genuine gaps in code I'd already reviewed twice.

Gates: 93 files / 1283 tests, `tsc` and `eslint` clean on every changed file.

One thing deliberately **not** changed: the ~11 duplicated lines Sonar reports against `feedbackRecords/lib/errors.ts`. That's `handleUnexpectedError` matching six inline copies in `surveys/lib/operations.ts`, and those copies aren't identical — they differ in log messages, log-field order, and two carry an extra `InvalidInputError → 400` branch. Unifying needs a helper with ~three knobs across a file this PR doesn't otherwise touch, so it's tracked as ENG-2048 alongside a related error-handling inconsistency in the taxonomy routes.

apps/web/app/api/v3/feedbackRecords/lib/access.ts
```diff
@@ -0,0 +1,194 @@
...
+    if (!requested) {
+      return {
+        ok: false,
+        response: problemForbidden(
+          requestId,
+          "You are not authorized to access this feedback dataset",
+          instance
+        ),
+      };
+    }
```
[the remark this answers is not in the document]
xernobyl: Confirmed and fixed in f9103d6 — this was a real bug, not just an inconsistency.

The resolver normalises the id it returns, and that value is what we echo to callers as `meta.datasetId` / `dataset_id`. So a caller replaying the id we just gave it would have hit this raw `===` and been 403'd on a dataset it owns.

Trimmed on both sides (`directory.id.trim() === datasetId.trim()`) — the input too, since it is the other half of that round trip. Test added: `accepts the normalised datasetId it would have echoed back`. Also verified live: the `datasetId` returned in `meta` is now accepted by both `list_feedback_records` and `count_feedback_records`, and matches the `dataset_id` on the records themselves.

apps/web/app/api/v3/feedbackRecords/lib/errors.ts
```diff
@@ -0,0 +1,149 @@
...
+      requestId,
+      relayableHubDetail(error, "The feedback service reported a conflict."),
+      instance
+    );
+  }
+
+  // The Hub's body cap is lower than ours, so this is reachable with a large (but locally valid) payload.
+  if (status === 413) {
+    return problemPayloadTooLarge(requestId, "The feedback record is too large.", instance);
+  }
```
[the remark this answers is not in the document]
xernobyl: Fixed in f9103d6. You are right that this was the odd branch out — 413 is in `RELAYABLE_HUB_STATUSES`, and the per-record batch-create path already relayed it, so a single create/update was the only place that swallowed the Hub's real size-limit detail.

Now `relayableHubDetail(error, "The feedback record is too large.")`, so the Hub's message comes through when it sends one and the fixed fallback applies when it does not. Test added: `relays the Hub's own size-limit detail on a 413`, alongside the existing test that pins the fallback.

apps/web/app/api/v3/feedbackRecords/lib/operations.ts
```diff
@@ -0,0 +1,1019 @@
...
+    });
+    if (!resolution.ok) {
+      return resolution.response;
+    }
+
+    const listParams: FeedbackRecordListParams = {
+      ...buildHubFilterParams(resolution.tenantId, filters),
+      limit: limit ?? DEFAULT_LIST_LIMIT,
+    };
+    if (cursor) listParams.cursor = cursor;
```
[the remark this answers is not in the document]
xernobyl: Fixed in f9103d6. This was the most substantive of the six: these were the only two operations not validating their own input, which contradicts what this layer claims to be — the docstring in `schemas.ts` says the schemas live in the v3 layer "so the operations can validate independently of any transport", and list/count were not holding up that contract.

Both now validate — `list` against `ZV3FeedbackRecordListFilters` (filters plus pagination, and it forwards the validated `limit`/`cursor`), `count` against `ZV3FeedbackRecordFilters` — and the Hub params are built only from the parsed result. `parseSimilarityFilters` became `parseQueryInput` with a label parameter, since it is no longer similarity-specific. Side benefit: `ZV3FeedbackRecordListFilters` was an exported-but-unused schema and now is not.

Two things I checked before landing it, since adding validation can only break existing callers: nothing outside `modules/mcp/tools/feedback-records.ts` calls these operations, and the MCP layer already validates with the same schemas — so the new 422s are unreachable today and act as defence for a future REST caller. Then I exercised every filter at its valid extremes live (`limit: 1000`, `limit: 1`, all six ids, `fieldType`, a `since`/`until` range) to confirm nothing legitimate now gets rejected.

apps/web/app/api/v3/feedbackRecords/lib/schemas.ts
```diff
@@ -0,0 +1,294 @@
...
+export const ZV3FeedbackRecordUpdateBodyFields = ZV3FeedbackRecordCreateBodyFields.pick({
+  value_text: true,
+  value_number: true,
+  value_boolean: true,
+  value_date: true,
+  value_id: true,
+  user_id: true,
+  language: true,
+  metadata: true,
+});
```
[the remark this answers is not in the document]
xernobyl: Fixed in f9103d6, exactly as proposed — `ZV3FeedbackRecordUpdateBodyFields` is defined first and `ZV3FeedbackRecordUpdateBody` derives from it via `.refine()`, so the eight-key list exists once.

Worth noting why both forms exist at all, in case it reads as redundant: `inputSchema` needs a raw `.shape`, which a refined schema does not expose. The `.pick()` from the create fields is also load-bearing rather than stylistic — it is what makes update inherit create's field-level refinements (the 30k text cap and the 32 KB metadata byte cap). I had not actually proven that carried through `.pick()`, so there are now tests for it.

apps/web/modules/mcp/tools/feedback-records.test.ts
```diff
@@ -0,0 +1,521 @@
...
+    );
+    const { tools } = createToolServer();
+
+    const result = await tools.get("create_feedback_record")!.handler(body, { authInfo });
+
+    expect(result.isError).toBe(true);
+    expect(auditLog.status).toBe("failure");
+    expect(auditLog.eventId).toBe("req_tool");
+    expect(queueV3AuditLog).toHaveBeenCalledWith(auditLog, "req_tool", expect.anything());
+  });
```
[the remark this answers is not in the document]
xernobyl: Added in f9103d6 — for both write handlers, since the batch tool has its own `catch` with the same invariant.

Your framing is the right one: "a failed mutation is never silently unaudited" was the stated guarantee and the branch upholding it was unguarded. `still queues a failure audit event when the operation throws, and rethrows` covers `writeHandler`; `still queues a failure event when the batch operation throws, and rethrows` covers the batch handler, where it also pins that exactly one event is queued rather than one per record.

docs/development/technical-handbook/mcp-server.mdx
```diff
@@ -65,35 +66,53 @@ https://app.formbricks.com/api/auth
...
+| Scope                   | Use                                                                                                                                                            |
+| ----------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `surveys:read`          | `list_surveys`, `get_survey`, and read-only validation paths                                                                                                   |
+| `surveys:write`         | `create_survey`, `patch_survey`, `delete_survey`, and write validations                                                                                        |
+| `feedbackRecords:read`  | `list_feedback_datasets`, `list_feedback_records`, `count_feedback_records`, `get_feedback_record`, `search_feedback_records`, `find_similar_feedback_records` |
+| `feedbackRecords:write` | `create_feedback_record`, `create_feedback_records`, `delete_feedback_record`                                                                                  |
```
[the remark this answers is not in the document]
xernobyl: Fixed in f9103d6 — and thank you, this one was a genuine miss on my side rather than an oversight in the original write-up.

I had already attempted this exact fix in an earlier commit, but the edit matched a specific column padding that prettier had since reflowed, so it silently did nothing. My verification then counted `update_feedback_record` occurrences across the whole file, which passed on the tool section and the permission table while this row stayed stale — so the check confirmed the wrong thing.

Fixed with a padding-insensitive match and verified against this row specifically. I also added a programmatic cross-check that every registered tool appears in the correct scope row in both this table and `docs/platform/mcp/overview.mdx`, plus the permission table — currently zero mismatches across all ten tools.

apps/web/app/api/v3/feedbackRecords/lib/operations.ts
```diff
@@ -0,0 +1,1044 @@
...
+      datasetId,
+      minPermission: "readWrite",
+      requestId,
+      instance,
+    });
+    if (!resolution.ok) {
+      return resolution.response;
+    }
+
+    const parsed = ZV3FeedbackRecordUpdateBody.safeParse(body);
```
Dhruwang, line 738: Update only validates the picked body — unlike create it never re-checks `value_*` against the record's `field_type`, and it accepts `value_id`. Confirmed live: one PATCH put `value_number:99` + `value_id` onto a `text` record that already had `value_text`, leaving text+number+id on one record (create 422s that). `owned.record.field_type` is available right here — worth re-validating against it (mirror create's `superRefine`). `value_id` is also UI-locked for the same integrity reason (ENG-1673).

apps/web/app/api/v3/feedbackRecords/lib/errors.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * The one place that decides what a Hub failure may say to a caller.
+ *
+ * The Hub owns content rules we deliberately don't duplicate (NULL bytes, its own length limits), so for
+ * the statuses above its message is relayed — bounded — because without it an agent cannot correct its own
+ * request. Everything else is replaced by a fixed string. Used both for whole-request problem responses and
+ * for the per-record failures of a batch write, so neither can drift into leaking more than the other.
+ */
+export function relayableHubDetail(error: HubError | null, fallback: string): string {
```
Dhruwang, line 90: Relayed Hub detail leaks the internal term `tenant_id`: a batch duplicate 409 comes back as "...with this tenant_id, submission_id, and field_id already exists". Everywhere else the boundary is `dataset_id`. Worth scrubbing/renaming `tenant_id` in relayed detail.

apps/web/app/api/v3/feedbackRecords/lib/errors.ts
```diff
@@ -0,0 +1,153 @@
...
+      requestId,
+      relayableHubDetail(error, "The feedback record is too large."),
+      instance
+    );
+  }
+
+  // Embeddings are optional in the Hub, and the search endpoints are the only ones that need them. A
+  // deployment-level "not enabled", not an outage — so it must not collapse into the generic 502 below,
+  // which would read as "retry later" for something no retry can fix.
+  if (status === 503) {
```
Dhruwang, line 147: 503 maps to the embeddings-unavailable message for every op, not just search/similar. A plain Hub outage on list/create/etc. would return a misleading "configure EMBEDDING_*". Consider gating this to the two search ops and collapsing 503 to the generic 502 elsewhere.

apps/web/app/api/v3/feedbackRecords/lib/schemas.ts
```diff
@@ -0,0 +1,287 @@
...
+ * here and mapped to Hub params by one function in `operations.ts`.
+ *
+ * Field semantics follow the Hub's own `GET /v1/feedback-records` parameter documentation (hub 0.8.1
+ * `openapi.yaml`). Every filter is a plain equality match, combined with AND, and always scoped to the
+ * resolved dataset. Length caps mirror the Hub's (1–255); it additionally rejects NULL bytes and relays
+ * that as a 400 we pass through.
+ */
+const ZFeedbackRecordFilterId = z.string().trim().min(1).max(255);
+
+export const ZV3FeedbackRecordFilters = z.object({
```
Dhruwang, line 35: Filters are camelCase (`sourceType`, `userId`) while create/update bodies are snake_case (`source_type`, `user_id`), and unknown keys are stripped — so a wrong-cased filter silently no-ops instead of erroring. Align the casing, or `.strict()` the filter schema so a bad key is rejected.

Dhruwang (commented):
A few notes from exercising the feedback-record tools end-to-end against a local Hub. Details inline; bug summary in a separate comment.

Dhruwang:
Exercised the 10 tools end-to-end against a local Hub (embeddings off). Happy paths, pagination, the count/list invariant, partial-batch reporting, the no-oracle 403s, and the 503 message all behaved as designed. Tenant isolation held up (IDOR guards, injection, scope gate).

Bugs worth a look (detail in the inline comments):

- **[P2] update accepts values that contradict `field_type`, and writes `value_id`.** Live: one PATCH left `value_text` + `value_number` + `value_id` on a `text` record — a state `create` rejects. `operations.ts` update path.
- **[P1] filter vs body casing.** Filters camelCase, bodies snake_case; unknown keys stripped, so a mistyped filter silently returns wrong results. `schemas.ts`.
- **[P2] Hub 503 always returns the embeddings message**, even for non-search ops. `errors.ts`.
- **[P3] `tenant_id` leaks in relayed error detail** (batch-dup 409). `errors.ts`.
- **[P2] batch create has no concurrency cap** on the Hub fan-out (`Promise.all` over ≤50 + SDK retries ≈ ~150 concurrent). `service.ts` `createFeedbackRecordsBatch`.
- **[P1] new MCP scopes break already-connected OAuth (DCR) clients** until they re-register — worth a release note.
- **[P2] free-text `value_text` reaches the audit log**; `redactPII` is key-name based and can't scrub in-text PII — a retention/erasure call, not a code bug.
- **Maintainability:** `operations.ts` is ~1044 lines, mostly a repeated tenant/try-catch scaffold; a `withFeedbackTenant` wrapper would cut it down.

apps/web/app/api/v3/feedbackRecords/lib/operations.ts
xernobyl: Good catch, and thanks for the live repro — I reproduced it too (`{value_number, value_id}` on a `text` record → 200 with text, number *and* id set).

Fixed in 967522af2. The cross-check now runs after `requireOwnedFeedbackRecord`, reusing create's `VALUE_FIELD_BY_TYPE` against `owned.record.field_type`, and returns a 422 naming the offending field. It has to be there rather than in the schema: `field_type` is immutable, so it is never in the patch and can only come from the stored record.

One thing I did differently — I kept `value_id` in the updatable set rather than dropping it. For a `categorical` record it *is* the value (the option's stable id), so a blanket ban would make a mis-mapped choice id uncorrectable via the API; the per-type table already rejects it on every other `field_type`, which is the case you hit. On the ENG-1673 UI-lock point: the record drawer is read-only end to end — every field in it is `disabled`, including `value_text` — so I do not think it singles `value_id` out. Happy to reverse if you know of a reason specific to it.

apps/web/app/api/v3/feedbackRecords/lib/errors.ts
xernobyl: Agreed. Confirmed the exact string against a local Hub:

```
{"status":409,"detail":"a feedback record with this tenant_id, submission_id, and field_id already exists", ...}
```

Fixed in 967522af2 with a word-bounded `tenant_id` → `dataset_id` rewrite applied to every relayed string — the detail *and* `invalid_params` name/reason, since the Hub names fields in both. Went with `dataset_id` because that is what the product calls it (45 user-facing strings say "dataset"; `FeedbackDirectory` is only the internal name) and the record serializer already renames the field outward.

apps/web/app/api/v3/feedbackRecords/lib/errors.ts
xernobyl: Right, and this one is a real misdirection — a Hub outage on a list or a create telling the operator to go configure `EMBEDDING_*`.

Fixed in 967522af2. I kept the 503 rather than collapsing it to a 502, but made the *detail* the caller's to supply: it defaults to wording that names no subsystem, and only `searchV3FeedbackRecords` and `findSimilarV3FeedbackRecords` pass `EMBEDDINGS_UNAVAILABLE_DETAIL`. Status-wise 503 still seems the honest answer — it is genuinely "unavailable, retry later", where a 502 reads as "the upstream is broken".

Worth flagging why the shape matters beyond this PR: I hit the same thing on the taxonomy routes in ENG-2048, where a Hub 503 has *four* causes and only one is embeddings — and the embeddings one is unreachable through the UI, since `/fields` disables the Generate button first. So the per-caller wording is load-bearing there, not just tidier. That stack (#8666) now builds on this instead of carrying its own copy.

apps/web/app/api/v3/feedbackRecords/lib/schemas.ts
xernobyl: Agreed, and I went with aligning the casing.

Two things pushed me to snake_case rather than camelCase: the create/update bodies use it, and so do the *responses* (`user_id`, `source_type`, `dataset_id`) — so the realistic failure is reading `user_id` off a record and filtering by it. v3's camelCase query-param convention (`workspaceId`, `sortBy`, `includeTotalCount`) does not really apply here either, since these operations have no HTTP route; MCP is the only caller.

In 967522af2: the eight record-field filters are renamed to the fields they filter, while our own parameters (`workspaceId`, `datasetId`, `limit`, `cursor`, `minScore`) stay camelCase — they name nothing in the record. `buildHubFilterParams` is now near-identity but stays a field-by-field allowlist, same reason as the create builder.

On `.strict()` — added, but I want to be straight about what it does *not* buy: I checked the MCP SDK (`validateToolInput` → `normalizeObjectSchema`), and because we pass `inputSchema: Schema.shape`, the SDK builds its own plain `z.object` from the raw shape and strips unknown keys before the operation is ever called. So strict hardens the operation contract for any future caller, but it cannot make a mis-cased key error at the MCP boundary — the rename is what actually fixes that path. Making MCP itself strict would mean passing the schema instance instead of `.shape` (the SDK does support that), which changes the advertised JSON Schema for every tool; happy to do it as a follow-up if you think it is worth it.

xernobyl:
Thanks @Dhruwang — all four were valid and all four reproduced against a local Hub. Addressed in 967522af2; details inline on each thread.

Summary of what changed, and where I deviated from the suggestion:

| Finding | Fix | Deviation |
|---|---|---|
| Update skipped the `field_type` cross-check | Checked after the ownership guard against the stored `field_type`, reusing create's table → 422 naming the field | Kept `value_id` updatable — it *is* the value for `categorical`; the per-type table rejects it elsewhere |
| Relayed detail leaked `tenant_id` | Word-bounded `tenant_id` → `dataset_id` over detail **and** `invalid_params` | — |
| 503 blamed embeddings everywhere | 503 detail is caller-supplied, default names no subsystem; only the 2 search ops pass the embeddings message | Kept 503 rather than collapsing to 502 — it is honestly "retry later" |
| Filters camelCase vs snake_case bodies | Renamed the 8 record-field filters to snake_case; `.strict()` so unknown keys 422 | Chose snake_case (not camelCase) — responses are snake_case too, so it's the direction that removes the trap |

Two notes worth surfacing beyond the threads:

- **`.strict()` does not fix the MCP path.** The SDK validates against its own non-strict copy of the raw shape and strips unknown keys before an operation runs, so the *rename* is what prevents the silent no-op there. Making MCP strict means passing schema instances instead of `.shape`, which changes the advertised JSON Schema for every tool — flagged as a possible follow-up.
- **The 503 finding reshaped the stacked PR.** #8666 (ENG-2048) needed the same per-caller wording for taxonomy, where a Hub 503 has four causes and the embeddings one is unreachable through the UI. That fix now lives here, and #8666 rebased onto it as a pure module move — so `main` is never wrong, rather than being fixed one PR later.

Verification: `902 passed` across `app/api/v3`, `modules/mcp`, `modules/hub`, `modules/ee/unify-feedback`; eslint + `tsc` clean on changed files; gitleaks and semgrep clean on the diff. Docs (tool reference + filter table) updated, and `update_feedback_record`'s tool description now states the `value_*`/`field_type` rule so an agent learns it from the schema rather than from a 422.

Dhruwang (approved):

PR #8652, opened 2026-07-27
fix: disable external pre-aggregations for bundled Cube

### LINKED ISSUE #8418: Cube Store is not found.
### Issue Summary

**Cube Store is not found**

Installed Formbricks today via [https://formbricks.com/docs/self-hosting/setup/docker] (**not** the one-click install).
However, upon verifying the stack I get:

> Warning. Cube Store is not found. Please follow this documentation to configure Cube Store https://cube.dev/docs/caching/running-in-production
 🚀 Cube API server (1.6.6) is listening on 4000

I've read that linked page, but can't seem to work out what I need to change/update.

Thank you

_I also commented on a previously closed out issue, but not sure that was the best place for this so I'm reposting here. Feel free to delete if the other discussion is a better place._

### Expected Behavior

Successful verification of the stack

_No response_

### Your Environment

Self-hosted Formbricks

BhagyaAmarasinghe:
Fixes: ENG-1637

## What & why

The bundled production Docker and Helm configurations run a single Cube v1.6.6 replica with in-memory cache and queue storage, but Cube defaults external pre-aggregations to enabled. Without Cube Store, that mismatch emits a misleading production warning even though Cube is healthy and Formbricks does not define external pre-aggregations.

This change defaults `CUBEJS_EXTERNAL_DEFAULT` to `false` for both bundled production surfaces while preserving an explicit operator override. It keeps the existing single-replica memory-mode guard, documents when Cube Store is required, and explains the one-time update needed for existing Docker Compose installations.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1637/cube-store-is-not-found

Fixes #8418

## How this was tested

- `./node_modules/.bin/vitest run docker/__tests__/formbricks-script.test.ts` ✅ — 5/5 tests passed
- `docker compose config` ✅ — resolved the unset default to `false` and an explicit override to `true`
- Helm chart validation ✅ — linted successfully, rendered default/override values correctly, and retained the multi-replica memory-mode failure
- Isolated production Compose runtime ✅ — Cube `/readyz` returned 200, an authenticated `FeedbackRecords.count` query returned 200 with `external: false`, and Cube logs did not contain the Cube Store warning
- Prettier and `git diff --check` ✅

## QA / Test Plan

**How to test**

- [ ] Deploy a new self-hosted Docker installation with the default Cube configuration, then open the Cube service logs in the deployment UI → Cube starts without the `Cube Store is not found` warning.
- [ ] Send an HTTP request to the Cube `/readyz` endpoint from within the deployment network → it returns HTTP 200 with a healthy response.
- [ ] Open Formbricks feedback analytics after ingesting a feedback record → analytics load successfully and reflect the record.
- [ ] Set `CUBEJS_EXTERNAL_DEFAULT=true` in an installation with an operator-managed Cube Store → the override reaches the Cube container and normal analytics remain available.

**Preconditions / test data**

- A self-hosted Docker or Helm installation with valid `CUBEJS_API_SECRET` and database configuration.
- At least one feedback record for the analytics verification.
- An operator-managed Cube Store is required only for the explicit `true` override scenario.

**Risks & regressions**

- Custom external pre-aggregations require Cube Store and must override the new default.
- Helm installations must continue rejecting multiple Cube replicas while the cache and queue driver is `memory`.
- Existing Docker installations retain their current Compose file and need the documented environment entry before recreating Cube.

- Adds the optional `CUBEJS_EXTERNAL_DEFAULT` Docker environment variable, defaulting to `false`.
- Adds `cube.env.CUBEJS_EXTERNAL_DEFAULT: "false"` to Helm defaults.
- No database migrations, services, ports, or persistent-volume changes.

### DIFF
--- a/.github/workflows/helm-chart-validation.yml
+++ b/.github/workflows/helm-chart-validation.yml
@@ -39,6 +39,29 @@ jobs:
 
           helm template qa "$chart" \
             --set formbricks.webappUrl=https://qa.example.com \
+            --show-only templates/cube-deployment.yaml > "$render_dir/cube-default.yaml"
+          grep -A1 'name: CUBEJS_EXTERNAL_DEFAULT' "$render_dir/cube-default.yaml" \
+            | grep -q 'value: "false"'
+
+          helm template qa "$chart" \
+            --set formbricks.webappUrl=https://qa.example.com \
+            --set-string cube.env.CUBEJS_EXTERNAL_DEFAULT=true \
+            --show-only templates/cube-deployment.yaml > "$render_dir/cube-external.yaml"
+          grep -A1 'name: CUBEJS_EXTERNAL_DEFAULT' "$render_dir/cube-external.yaml" \
+            | grep -q 'value: "true"'
+
+          if helm template qa "$chart" \
+            --set formbricks.webappUrl=https://qa.example.com \
+            --set cube.replicas=2 \
+            --show-only templates/cube-deployment.yaml > "$render_dir/cube-invalid.yaml" 2>&1; then
+            echo "Expected the in-memory Cube configuration with multiple replicas to fail" >&2
+            exit 1
+          fi
+          grep -q 'CUBEJS_CACHE_AND_QUEUE_DRIVER=memory is only supported when cube.replicas=1' \
+            "$render_dir/cube-invalid.yaml"
+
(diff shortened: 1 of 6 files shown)


Dhruwang (approved):

PR #8654, opened 2026-07-28
feat(support): migrate in-product chat from Chatwoot to Plain [ENG-1984]

Dhruwang:
## What & why

Cloud in-product support chat ran on **Chatwoot**. This migrates it to **Plain** (the new unified support platform). The Chatwoot live-chat widget is replaced with Plain's Chat JS SDK, verified customer identity, and a paying-customer thread label; all `CHATWOOT_*` wiring is swapped to `PLAIN_*` and `app/chatwoot/` is removed.

- `app/plain/components/plain-chat.tsx` — loads `https://chat.cdn-plain.com/index.js`, calls `Plain.init()` once, then pushes later changes through Plain's in-place APIs (identity via `setCustomerDetails`, launcher via `update({ hideLauncher })`). No script re-injection, no dangling listeners.
- `app/plain/lib/identity.ts` — server-only `computePlainEmailHash()` (HMAC-SHA256 of the email with `PLAIN_CHAT_HMAC_SECRET`); the hash is a bearer credential, computed server-side and passed to the client already hashed.
- `app/plain/lib/customer.ts` — server-only `getIsActiveCustomer()`. Resolved in the layout so the paying-customer label is attached in `Plain.init({ threadDetails })` before the first thread. **Only `pro` and `scale` count** as paying (not `hobby`, not `custom`). The check reads the `Organization.billing.stripe` snapshot in Postgres (kept in sync by the Stripe webhook), across all of the user's organizations. **Cloud-only:** Stripe plans never exist on self-hosted, so it returns `false` early when `IS_FORMBRICKS_CLOUD` is unset instead of querying the user's organizations on every `(app)` render.
- Env/constants/turbo/.env.example: `CHATWOOT_*` → `PLAIN_*`. PostHog session-recording `blockSelector` now targets the Plain iframe.

Screenshot 2026-07-28 at 12 17 50 PM

## Linear ticket

https://linear.app/formbricks/issue/ENG-1984/move-in-product-chat-to-new-support-platform

## How this was tested

- `pnpm vitest run app/plain/lib/*.test.ts` ✅ — `utils` (4), `identity` (3), `customer` (9). 16 tests.
- Unit tests added: `identity.test.ts` (hash null-when-unset / correct HMAC / deterministic), `customer.test.ts` (pro & scale + trialing = paying; hobby, custom, free, canceled, no-stripe, empty = not paying; multi-org `.some`; self-hosted returns false **without** calling `getOrganizationsByUserId`).
- `tsc --project tsconfig.typecheck.json` — no errors in touched files. `eslint` and `prettier --check` on touched files — clean.
- Manual (local): widget loads for a logged-in user, thread created in Plain (identity populated); active-customer label correctly withheld for a `hobby`/`custom` account after narrowing paid plans to `pro`/`scale`.

## QA / Test Plan

**How to test**

- [ ] Set `PLAIN_APP_ID` (+ add the app domain to Plain's allowed origins) → log in → the Plain launcher appears on app pages.
- [ ] Open the widget, send a message → a thread appears in Plain with your name/email.
- [ ] With `PLAIN_CHAT_HMAC_SECRET` set (and matching Plain's identity secret) → the customer shows as **verified** in Plain.
- [ ] Navigate to onboarding (`/organizations/:id/workspaces/new`, `/organizations/:id/landing`) → launcher is **hidden**; leave onboarding → launcher returns.
- [ ] On Cloud (`IS_FORMBRICKS_CLOUD=1`) with `PLAIN_ACTIVE_CUSTOMER_LABEL_TYPE_ID` set: a user in a `pro`/`scale` active (or trialing) org → new threads carry the label. A `hobby`/`custom`/free-only user → **no** label.
- [ ] Self-hosted (`IS_FORMBRICKS_CLOUD` unset) with `PLAIN_ACTIVE_CUSTOMER_LABEL_TYPE_ID` set → widget works and new threads carry **no** active-customer label; app pages load normally.
- [ ] Unset `PLAIN_APP_ID` → widget does not mount (no errors).

**Preconditions / test data**

- Cloud only (Plain replaces the Cloud chat widget). A Plain Chat app (app id, identity secret, optional label type id). For the label: `IS_FORMBRICKS_CLOUD=1` plus an org on a `pro` or `scale` plan with an active/trialing Stripe subscription.

**Risks & regressions**

- Chat widget presence in the `(app)` layout; PostHog session-recording masking of the chat widget (selector changed from the Chatwoot id to the Plain iframe).
- Active-customer labeling is per-user across all their orgs (unchanged semantic from Chatwoot), now scoped to `pro`/`scale` only, and skipped entirely off Cloud. An instance running paid plans without `IS_FORMBRICKS_CLOUD=1` would get no label — that combination is not a supported setup.

- New env: `PLAIN_APP_ID` (required to enable), `PLAIN_CHAT_HMAC_SECRET` (verified identity), `PLAIN_ACTIVE_CUSTOMER_LABEL_TYPE_ID` (optional label).
- Removed env: `CHATWOOT_WEBSITE_TOKEN`, `CHATWOOT_BASE_URL`.
- No DB migration. Cutover: provision the Plain Chat app + set the env vars in the deploy environment, then redeploy.


apps/web/app/plain/lib/customer.ts
```diff
@@ -0,0 +1,21 @@
...
+import { getOrganizationsByUserId } from "@/lib/organization/service";
+
+const PAID_BILLING_PLANS = new Set<TCloudBillingPlan>(["pro", "scale"]);
+
+/**
+ * Whether the user belongs to at least one organization on a paid, active plan.
+ * Resolved server-side so the active-customer label can be attached to Plain
+ * threads at init time, before the user opens their first thread.
+ */
+export const getIsActiveCustomer = async (userId: string): Promise<boolean> => {
```
mattinannt, line 16: This is only relevant for Formbricks Cloud and not for self-hosting instances. I think it's best if this returns early if `IS_FORMBRICKS_CLOUD !== 1`

mattinannt (changes requested):
Thank you for the PR. I only have one concern

apps/web/app/plain/lib/customer.ts
Dhruwang: Good catch — done in f5c0c2e. `getIsActiveCustomer()` now returns `false` immediately when `IS_FORMBRICKS_CLOUD` is unset, so self-hosted instances skip the `getOrganizationsByUserId()` query entirely instead of walking every org just to always get `false`. Added a test asserting both the result and that the org query is never called.

---

Dhruwang:
@mattinannt Thank you for the review. I addressed your suggestion. Can you please recheck?

mattinannt (approved):

PR #8655, opened 2026-07-28
fix(workflows): reconcile stale trigger ending-card ids

Dhruwang:
## What does this PR do?

Fixes [ENG-2035](https://linear.app/formbricks/issue/ENG-2035/workflows-stale-ending-card-id-sticks-in-trigger-after-the-ending-is).

Deleting an ending card in the survey editor left its id behind on every workflow trigger that referenced it. `WorkflowTriggerForm.toggleEnding` appended onto `node.config.endingCardIds` verbatim and nothing anywhere compared those stored ids against the survey's actual endings — so picking the replacement ending produced the phantom "trigger on 2 ending cards", and **Enable** then failed on the server's `verifyTriggerSurvey` → `missingEndingCardIds` check. The only escape was the undiscoverable *All endings → Specific endings* toggle dance.

Server-side validation was already correct; this is the UI/state half.

**`lib/trigger-ending-cards.ts`** — pure, order-preserving reconcile of stored ids against the survey's real endings. Splits into kept vs. missing, and drops duplicates *without* counting them as missing (they would inflate the "N removed" signal). A definition-level wrapper returns `null` when the selection is already clean, so callers get a cheap "nothing to write" signal instead of marking the editor dirty on every render.

**`hooks/use-reconcile-trigger-ending-cards.ts`** — mounted by the builder page rather than the trigger form, so the canvas summary and the enable gate are correct without the inspector ever being opened; autosave persists the repair, which is what stops the drift recurring. It acts only when the definition is editable *and* the endings query has settled for that exact survey — a pending or failed lookup must never read as "the survey has no endings" and wipe a valid selection.

**Trigger form** — reconciles before applying a checkbox click, so the selection is exactly what the list shows.

### The widening case, and why the form needed changing too

Pruning the *last* stale id empties `endingCardIds`, and an empty list means **all endings** — so a repair can silently widen what fires the workflow. The form made this worse: `endingScope` initialised from the (now empty) config, so it opened on "All endings" and presented the widened state as though the user had chosen it.

So the reconcile publishes the dropped ids to a client-only `prunedTriggerEndingCardIdsAtom`, and the form treats them as "specific endings" intent even with an empty list. It opens on the checkbox list with its **existing** hint — *"Select at least one ending. With none selected, every ending fires this workflow."* — which is simultaneously the notice that something was removed and the warning that the trigger is now broader, shown exactly where the user has to act. The atom is deliberately not part of the definition: an empty `endingCardIds` cannot express "they wanted specific endings and the ones they picked are gone".

**`useWorkflowSurveyEndings`** — now reports `resolvedSurveyId`, so any caller that *deletes* stored ids can prove the endings belong to the current survey and not a previously-selected one.

### Deliberately out of scope

- The issue's "consider" item about cascading ending-card deletion from the survey editor to dependent workflows. That's a survey-editor → workflows dependency well outside trigger config, and the editor-side reconcile already unblocks the reported flow.
- A dedicated banner for **locked** workflows (enabled/archived/read-only) carrying stale ids, which can't be auto-repaired. The Test dry-run already reports that condition via the existing `test_problem_ending_card_not_found`, and reusing a `test_`-prefixed key in the trigger form would be worse than leaving it out.

### No new translation keys

Worth calling out because it shaped the design: this PR adds **zero** i18n keys, reusing `trigger_ending_cards_select_at_least_one`. `pnpm run scan-translations` passes.

## How should this be tested?

Automated — 25 unit tests:

- `lib/trigger-ending-cards.test.ts` — kept/missing split, order preservation, duplicate handling, empty selection, immutability of the input definition.
- `hooks/use-reconcile-trigger-ending-cards.test.ts` — prunes and records the dropped ids; records the widening case; duplicate-only drift reports nothing; quiet when clean; leaves locked definitions alone; keeps stale ids when the survey lookup fails; no fetch for "all endings"; refuses endings that resolved for a different survey. The last one was verified by deleting the guard and confirming the test goes red. The harness feeds the atom's definition back into the hook, so a reconcile that failed to converge would loop rather than pass.

Manual — the original repro:

1. Create a survey with an ending card. Create a workflow: trigger = that survey, **specific ending card** = that ending.
2. In the survey, delete that ending card and add a new one.
3. Open the workflow editor. The stale id is pruned on load — the canvas summary reads "Trigger on any survey response", not a phantom count.
4. Open the trigger config: it opens on **Specific endings** with nothing checked and the "select at least one / with none selected, every ending fires this workflow" hint, rather than silently showing "All endings".
5. Pick the new ending → summary reads **1** ending card, not 2.
6. **Enable** succeeds without the all-endings toggle workaround.

Also worth checking: switching the trigger survey still clears the selection, and opening a *different* workflow doesn't inherit the previous one's pruned state (reset on hydrate).

### Required

- [ ] Ran `pnpm build` — **not run locally**: the web build needs secrets this environment doesn't have (`ENCRYPTION_KEY`, `CUBEJS_*`, `HUB_API_*`, `REDIS_URL`). Every workspace package it depends on builds clean, `tsc --noEmit` over `apps/web` reports **zero** errors in `modules/ee/workflows`, and CI's *Build Formbricks-web* job passed on the first commit.
- [x] Checked for warnings, there are none — ESLint clean over all of `modules/ee/workflows`, Prettier clean
- [x] Removed all `console.log`s
- [x] Merged the latest changes from the base branch (`epic/workflows-v1`) onto my branch
- [ ] First PR at Formbricks? Please sign the CLA — n/a

### Appreciated

- [ ] Screen recording / screenshots — not captured; the app can't be booted here for the same missing-secrets reason
- [ ] Updated the Formbricks Docs — not needed, no user-facing API or documented behavior changed


Dhruwang:
The red **PR Check Summary** on run [30335951522](https://github.com/formbricks/formbricks/actions/runs/30335951522) is a superseded run, not a test failure. Lint, Unit Tests, Build and Helm all passed there; E2E was `cancelled` mid-step by `cancel-in-progress` when run [30336715252](https://github.com/formbricks/formbricks/actions/runs/30336715252) started on the same SHA, and the summary gate trips on `cancelled` as well as `failure`. Nothing to fix — the replacement run is in progress and I'll confirm it goes green.


---

Dhruwang:
CodeRabbit's 3 actionable comments were raised against `7a51fcc2` and are all already resolved on the current head (`6f4f64e8`) — no action needed, and the Autofix checkbox would be a no-op:

1. **Consolidate the duplicate `reconcileEndingCardIds` calls** (`workflow-trigger-form.tsx`) — there is only one call site left, inside `toggleEnding`. The second one was `staleEndingCardIds`, which went away in `b3edec88` along with the stale-id alert for locked workflows. Memoizing a single call would just be the sort of unnecessary `useMemo` AGENTS.md warns against.
2. **Move the `react-hot-toast` / `react-i18next` mocks into `__mocks__`** (`use-reconcile-trigger-ending-cards.test.ts`) — both `vi.mock` factories are gone. `b3edec88` dropped the prune toast in favour of the existing `trigger_ending_cards_select_at_least_one` hint, which removed the need for either mock; the only stub left is `fetch`.
3. **Throw rather than return `[]` when the endings shape is malformed** (`use-trigger-survey-picker.ts`) — implemented in `847d2044`, and `6f4f64e8` updated the picker test that was still asserting the old `isSuccess` contract (that one stale assertion had been failing both Unit Tests and SonarQube).

All 12 checks are green on `6f4f64e8`.


---

itsjavi (changes requested):
Thanks for the changes, I found 2 issues, both in the new guide-the-re-pick UX, neither affecting stored data:

1 - after a prune, choosing All endings is discarded on trigger-form remount (snaps back to Specific endings + hint). Reproduced twice.
2 - when the bound survey has zero endings, the prune widens the trigger to every-response with no notice: the hint lives after renderEndingChoices's empty-endings early return, so it can never render.

apps/web/modules/ee/workflows/components/inspector/workflow-trigger-form.tsx
itsjavi: The widening notice can't render when the bound survey has zero endings:

 `renderEndingChoices`
   returns at `endings.length === 0` before the scope select. That case prunes to `[]` and silently
   widens to every response, pill still "Valid".

apps/web/modules/ee/workflows/components/inspector/workflow-trigger-form.tsx
```diff
@@ -24,22 +27,29 @@ interface WorkflowTriggerFormProps {
...
+  // Only an authority once the query has SETTLED for the current survey; null until then, so a
+  // pending fetch never reads as "no endings" and the stored ids are taken at face value.
+  const surveyEndingIds =
+    endingsQuery.isSuccess && endingsQuery.resolvedSurveyId === triggerSurveyId
+      ? endingsQuery.endings.map((ending) => ending.id)
+      : null;
+  // Open in "specific" scope when ids are set OR the builder page pruned this trigger's picks: the
+  // user chose specific endings that are now gone, so ask for a fresh pick instead of showing the
+  // widened "all endings" state. Atom (not a prop) since it's trigger-specific.
+  const prunedEndingCardIds = useAtomValue(prunedTriggerEndingCardIdsAtom);
```
itsjavi, line 50: `prunedTriggerEndingCardIdsAtom` is never cleared once the user resolves the prompt, and the form
   reads it only in the `useState` initializer. After a full prune, picking "All endings" is a no-op
   write (guarded by `endingCardIds.length > 0`), so nothing clears the atom: remount the trigger form
   (click another node and back) and it snaps to "Specific endings" with the hint again. Clear the atom
   in `handleScopeChange`/`handleSurveyChange`

.github/workflows/breaking-change-label.yml
```diff
@@ -0,0 +1,81 @@
+name: Breaking Change Label
```
itsjavi: this file seems unrelated to this PR, it should be removed I think

Dhruwang:
Both reproduced and fixed in `0510435f`, plus the unrelated workflow file removed. Thanks — the second one was a real hole in my reasoning.

**1. "All endings" discarded on remount.** Confirmed exactly as you described: after a full prune `endingCardIds` is already `[]`, so `handleScopeChange("all")` hit the `length > 0` guard and wrote nothing, the atom stayed set, and because the form reads it only in the `useState` initializer the next remount snapped back to "Specific endings". `handleScopeChange` now clears the atom on **both** directions and unconditionally — a deliberate scope pick is the user answering the prompt, whether or not it happens to write to the config. `handleSurveyChange` clears it too, since pruned ids belong to the survey being switched away from.

**2. Notice unreachable with zero endings.** You're right, and your point about the pill is the part I'd missed: by the time the pill looks, the stale ids are gone, so `deriveTriggerEndingProblems` returns empty and reads "Valid" — there was no signal anywhere. The `endings.length === 0` early return now covers only the checkbox list; the scope select stays rendered (which also makes the notice dismissible via "All endings"), and the notice is hoisted out of that branch, keyed off the pruned atom so it shows even when there's nothing left to check.

**3. `breaking-change-label.yml`** — agreed, unrelated. It reached this branch via `1f605fd1` ("ci: auto-sync breaking-change label from PR body", ENG-2087), which is not on `epic/workflows-v1`. Removed here, so **@Dhruwang** it'll need re-landing in its own PR rather than being lost.

Two notes on verification. The changes are confined to `workflow-trigger-form.tsx`, and per AGENTS.md `.tsx` doesn't get component unit tests, so these two paths aren't covered by new unit tests — the 309 workflows tests still pass, ESLint and `tsc` are clean over `modules/ee/workflows`, and both fixes want the manual walk-through (prune, pick "All endings", switch nodes and back; and a survey with no endings). And the notice still reuses `trigger_ending_cards_select_at_least_one` to keep this PR at zero new i18n keys — with no endings on the survey its "Select at least one ending" opening is admittedly a bit off, since there's nothing to select. If you'd rather have wording written for that case, it needs a new key and someone with a Lingo key to run `pnpm i18n`; say the word and I'll add it.


---

itsjavi (approved):
thanls for the changes! it works

PR #613, opened 2023-07-26
Run Production Instance with a Single Command

ShubhamPalriwala:
## What does this PR do?
Run Formbricks instance (along with Postgres and SSL certificates configured internally) with a single script!

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How should this be tested?
```sh
/bin/sh -c "$(curl -fsSL https://raw.githubusercontent.com/formbricks/formbricks/prod-setup/docker/production.sh)"
```

## Checklist
- [ ] Added a screen recording or screenshots to this PR
- [x] Filled out the "How to test" section in this PR
- [x] Read the [contributing guide](https://github.com/formbricks/formbricks/blob/main/CONTRIBUTING.md)
- [x] Self-reviewed my own code
- [ ] Commented on my code in hard-to-understand bits
- [ ] Ran `pnpm build`
- [x] Checked for warnings, there are none
- [x] Removed all `console.logs`
- [x] Merged the latest changes from main onto my branch with `git pull origin main`
- [x] My changes don't cause any responsiveness issues
- [x] Updated the Formbricks Docs if changes were necessary

Issue #612, opened 2023-07-26
[FEATURE] Advanced User Targeting


mattinannt:
### Is your feature request related to a problem? Please describe.

The current approach for user filtering / targeting to qualify them for an in-product survey is very simple and a user can only be targeted based on their attribute values. But one of the great values of Formbricks is that it should allow very fine granular user targeting based on the user behaviour in the past. To solve this we want to introduce a new user targeting.

### Describe the solution you'd like

You can inspect the planned UI for the Advanced User Targeting in this Figma file: https://www.figma.com/file/YnPME8Jo5HWCE5Uvzvxs7X/Advanced-User-Targeting?type=whiteboard&node-id=0-1&t=Ww7cecUOp8V9XYZp-0

We introduce `User Segments` that are a set of rules how to filter users. The rules for the User Segments are very flexible and vary from simple attribute filtering to fine granular action filtering (past behaviour within the app).

The idea with the segment is that it is a separate model in the database where the filters are stored. So every survey has a segment, but only when a user saves this segment it is visible in the list of segments (indicated by a flag inside the model).
If we really allow segment filtering (e.g. user is in "Power User"), we will have to see later if this is too complicated to implement. But we want to plan with it first, so we know what the ideal structure should look like.

Please inspect the Figma file in detail as it includes all the structure of this feature, different rules and qualifiers.

### Additional context

Because this feature is very complex and introduces not only a complex UI but also new database entries and data structures within the Formbricks application, we split this issue into two steps:

1. Data Model Concept for User Segments
2. Implementation of Concept and UI

For the first step, please come up with a Zod Schema that we can use internally to store the User Segment filter information. If its easier for you, you can also first create a Json representation of an example User Segment that we discuss and in the next step derive a Zod Schema from it.

After the Zod Schema is set we can move over to the UI implementation.

Ankur-Datta-4:
I'd like to work on this issue @mattinannt

jobenjada:
PR is up

PR #8656, opened 2026-07-28
docs: document accepted risk for pnpm audit oauth-provider finding

mattinannt:
## What & why

`pnpm audit` reports a moderate finding for `@better-auth/oauth-provider` 1.6.23 ([GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528)), the plugin backing our MCP OAuth flow (ENG-1055). The advisory is that access-token audiences aren't bound to their authorization grant (RFC 8707 resource-indicator confusion): a client can complete a normal authorization flow and then request a JWT scoped to a resource server its authorization never covered.

It can't be closed with a pin like the other entries in `pnpm-workspace.yaml`. The whole 1.6.x stable line is in range and the patch ships only from `1.7.0-beta.4` — a line still in pre-release. Adopting it isn't a plugin bump either: the pre-release declares `better-auth` and `@better-auth/core` as matching `^1.7.0-rc.x` peers, making it a coupled auth-core upgrade touching login/2FA/SSO/sessions, plus new DDL for its resources model.

So this PR does the only thing available: it records the accepted risk in the file the next auditor actually opens, inside `overrides:`, under an explicit `# === Accepted risk, no override possible yet ===` header. The note states what the mitigation actually rests on (single-entry `validAudiences` — the resource-server `audience` pin hardens the MCP endpoint but is *not* a substitute, because it resolves to jose's membership-style `aud` check), that the RFC 8707 binding gap remains even for a single-audience deployment so the note can't be read as "compliant", which of the two legs is covered by a test and which isn't, and a realistic scope for the eventual 1.7.0 upgrade.

Four follow-up commits correct the neighbouring brace-expansion paragraph, which review found wrong on four independent points — each one a stale or mis-transcribed fact rather than a behavioural defect:

1. `GHSA-mh99-v99m-4gvg` was re-split upstream into per-major ranges and the lines we resolve have shipped patches, so the note's "no patch coming" and "a pin cannot close it" were false.
2. It called the affected paths "devDependency-only". They aren't — the chains are production-rooted via `googleapis` and `@boxyhq/saml-jackson` (`apps/web` `dependencies`) and `@ai-sdk/google-vertex` (a `packages/ai` `dependency`), and `pnpm audit` marks them `dev: false`. The conclusion is unchanged but now rests on the input (glob patterns from library code and developer config, never request data) rather than on the dependency type.
3. The per-line contributor lists were attributed to the wrong lines — `vite-plugin-dts>@vue/language-core`, `@redocly/cli` and both prettier plugins are 2.x, not 1.x, and 1.x is the eslint plugin graph.
4. A **second advisory**, [`GHSA-rgw5-rvv9-x895`](https://github.com/advisories/GHSA-rgw5-rvv9-x895), now covers all three lines we resolve **including 5.0.8**. So `brace-expansion` is dropped from the "resolve to the patched version on their own now" enumeration entirely, and that paragraph moved out from under `# === Removed once upstream caught up ===` — a note whose thesis is "re-pinning closes this" doesn't belong under "what no longer needs a pin".

The oauth-provider note is also deliberately version-agnostic now. Naming the `latest` and `rc` tags is what made it rot within two days of its own re-verification stamp; the rc-specific claims are pinned to the release actually inspected.

**Comment-only change.** No dependency versions, no lockfile entries, no runtime behavior.

## Linear ticket

[ENG-1703](https://linear.app/formbricks/issue/ENG-1703/dependabot-613-better-authoauth-provider-better-authoauth-provider-may) — filed from Dependabot alerts for this GHSA (#612 and #613, both dismissed as tolerable risk) and closed `Done` on 2026-07-16 with the same "accept as documented risk" conclusion (verified via Linear). This PR surfaces that reasoning in the file rather than leaving it only in a closed ticket; `pnpm audit` is a separate channel from the Dependabot alerts, which is why the finding still shows up locally.

Follow-ups this PR points at, both filed during review: [ENG-2234](https://linear.app/formbricks/issue/ENG-2234/re-pin-brace-expansion12-to-close-ghsa-mh99-v99m-4gvg) (re-pin brace-expansion; that PR also deletes the paragraph this one corrects) and ENG-2182 (the vite + sanitize-html audit findings).

## How this was tested

Re-verified on 2026-08-05 against `main` at `4aabd0b`, since the branch sat for a week and the facts have a short half-life:

- `pnpm audit` ✅ — `@better-auth/oauth-provider` 1.6.23 still reported, first patched `1.7.0-beta.4`. No pins are touched by this PR, so it cannot change any count. Deliberately not quoting totals here: they moved twice during review (vite dropped out via #8724, then main's lockfile drift added `postcss`, `socket.io-parser`, `undici`, `fast-uri`, `ip-address` and `hono` findings). See "Risks & regressions".
- npm registry checked directly ✅ — no stable 1.7.0 exists; the `latest` and `rc` tags both moved during review, which is exactly why the note no longer names them.
- `1.7.0-rc.4` tarball inspected ✅ — peers are `better-auth: ^1.7.0-rc.4` and `@better-auth/core: ^1.7.0-rc.4`; `validAudiences` is gone; `oauthResource`, `oauthClientResource` and `oauthClientAssertion` are declared, and none of the three exist in `packages/database/schema/main.prisma` or migration `20260702093000_add_better_auth_oauth_provider`.
- Every code claim re-checked against the current tree ✅ — single-entry `validAudiences` (`mcp-oauth-provider-options.ts:20`), the `audience`/`issuer` pin (`modules/mcp/auth.ts:319-322`), `customAccessTokenClaims` returning only email/name, no SCIM anywhere, and the migration having created `oauthClient` with `oauthAccessToken.token` and `sessionId`.
- jose's audience semantics verified from source ✅ — `verifyOptions` flows into `jwtVerify` (`@better-auth/core/dist/oauth2/verify.mjs:64`) and `checkAudiencePresence` is a membership test, so a multi-audience `aud` passes. This is why the resource-server pin is described as hardening rather than as a substitute.
- The single-audience caveat checked against the advisory text ✅ — it states the binding gap remains "even where cross-audience escalation does not apply", which is what the note now says.
- Test-coverage claim ✅ — `apps/web/modules/mcp/auth.test.ts:315-321` asserts the `audience`/`issuer` pin; nothing in `apps/web` asserts `validAudiences`, so that's the assert worth adding.
- brace-expansion re-derived from the audit paths, the manifests and the registry ✅ — both advisories and every range confirmed; 1.1.18 / 2.1.4 / 5.0.9 all published 2026-07-30 and all past the 3-day `minimumReleaseAge` cooldown. Contributor roots counted from the second hop of every path: 1.1.16 is the eslint plugin graph (`typescript-eslint` 33, `@vitest/eslint-plugin` 30, `eslint-config-next` 13, `eslint-plugin-storybook` 9) plus `@boxyhq/saml-jackson` 5 and `@redocly/cli` 1; 2.1.2 is `vite-plugin-dts` 6, `@redocly/cli` 3, the two prettier plugins 2 each, and the three prod chains. `pnpm audit` caps `paths` at 100 per finding and 1.1.16 hits exactly 100, so the 1.x counts are a floor — hence "dominated by" in the note rather than an enumeration.
- YAML parses and no added line exceeds the 110-char print width.
- `pnpm test` / `pnpm lint` / `pnpm build`: delegated to CI on this head — a comment-only YAML change can't affect them and the lockfile is byte-identical. Green on every earlier head, including E2E, Docker Build, Better Auth integration tests and the Sonar gate.

Rebased onto current `main` twice during review, so every placement and attribution call was made against the current file.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Nothing to exercise in the running app or API — this PR only adds and edits YAML comments. Verification is by reading the diff.
- [ ] Read the accepted-risk block and check it against [GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528) and ENG-1703: the advisory range, the two mitigations and which one is load-bearing, the "not exploitable here, not compliant" caveat, the covered/uncovered test legs, and the 1.7.0 upgrade scope.
- [ ] Read the brace-expansion block and check it against [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) and [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895): both advisories are named, all three lines are described as still vulnerable and closable by re-pinning, and the deferral points at ENG-2234.
- [ ] Confirm `brace-expansion` no longer appears in the "Removed once upstream caught up" enumeration, and that the deferral paragraph sits under its own header rather than under that one.

**Preconditions / test data**

- None — no runtime behavior, no env vars, no plan or entitlement dependency, identical on Cloud and self-host.

**Risks & regressions**

- None from this change: comments only, no code, dependency or lockfile change.
- Deliberately **not** fixed here, all real dependency changes wanting their own reviewed PR and test cycle:
  - `brace-expansion` 1.1.16 / 2.1.2 / 5.0.8 — [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) and [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895), both high. This PR corrects the *note*; the pin refresh to 1.1.18 / 2.1.4 / 5.0.9 that closes both is **ENG-2234**, which also deletes the paragraph.
  - `sanitize-html` 2.17.4 — [GHSA-vccv-cmxp-4j9h](https://github.com/advisories/GHSA-vccv-cmxp-4j9h), patched in 2.17.5. Tracked in **ENG-2182**. (`vite` was the other half of that ticket and is already fixed on main via #8724.)
  - Newly surfaced by main's lockfile drift during this review, none of them tracked yet: `postcss` 8.5.18, `socket.io-parser` 4.2.6, `undici` 6.27.0 / 7.28.0 (five advisories between them), `fast-uri` 3.1.4, `ip-address` 10.2.0 (three advisories) and `hono` 4.12.27. Worth a triage sweep of their own — flagging rather than folding in.

- none

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -138,7 +138,8 @@ overrides:
   # typeorm>sqlite3. `pnpm audit` marks these findings `dev: false`. By contributor, the 1.x line is
   # dominated by the eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next,
   # eslint-plugin-storybook) and the 2.x line by vite-plugin-dts>@vue/language-core across six workspace
-  # packages, plus both prettier plugins. @redocly/cli is the only root on both lines, mostly on 2.x.
+  # packages, plus both prettier plugins. Two roots sit on both lines: @redocly/cli (mostly 2.x) and
+  # @boxyhq/saml-jackson (mostly 1.x, via the node-gyp chain above).
   # What makes it tolerable is the input, not the dependency type: every one of these call sites expands
   # glob patterns supplied by library code or developer config (gaxios' cache cleanup, typeorm's entity
   # discovery, build and lint globs), never by request data.
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -117,11 +117,15 @@ overrides:
   # linkify-it, protobufjs@7, undici@7 and uuid@11 all resolve to the patched version on their own now.
   # js-cookie, shell-quote and form-data@4 were retired by bumping react-use to 17.6.1 (js-cookie ^3),
   # concurrently to 9.2.4 (shell-quote 1.9.0) and @redocly/cli to 1.34.17 (form-data 4.0.6) instead.
-  # GHSA-mh99-v99m-4gvg (brace-expansion <=5.0.7) is still reported by `pnpm audit` as high: the range
-  # also matches 1.1.16 / 2.1.2, each the latest release in its line with no patch coming. Those paths
-  # are devDependency-only (eslint, prettier plugins and @redocly/cli, all via minimatch) and never run
-  # against untrusted input, so the residual risk is accepted until their consumers move to
-  # brace-expansion@5 — a pin cannot close it.
+  # GHSA-mh99-v99m-4gvg is still reported by `pnpm audit` as high, but it is now FIXABLE — re-verified
+  # 2026-08-03. The advisory was re-split upstream into per-major ranges, and both lines we resolve have
+  # since shipped a patch: `<1.1.17` (we resolve 1.1.16; 1.1.17 published 2026-07-29) and
+  # `>=2.0.0 <2.1.3` (we resolve 2.1.2; 2.1.3 published 2026-07-28). The earlier note here — "no patch
+  # coming", "a pin cannot close it" — stopped being true on those dates: re-pinning brace-expansion@1
+  # and @2 closes both. That is a dependency change needing its own lockfile and test run, so it is
+  # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
+  # risk stays acceptable for the same reason as before: these paths are devDependency-only (eslint,
+  # prettier plugins and @redocly/cli, all via minimatch) and never run against untrusted input.
   # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
   # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
   # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -113,39 +113,49 @@ overrides:
   # load-bearing: @prisma/dev pins 1.2.0.
   valibot: 1.4.2
   # === Removed once upstream caught up (kept as a record of what no longer needs a pin) ===
-  # ajv@6, @babel/core, body-parser, brace-expansion@1/@2/@5, dompurify, engine.io@6, fast-uri@3, hono,
-  # linkify-it, protobufjs@7, undici@7 and uuid@11 all resolve to the patched version on their own now.
+  # ajv@6, @babel/core, body-parser, dompurify, engine.io@6, fast-uri@3, hono, linkify-it, protobufjs@7,
+  # undici@7 and uuid@11 all resolve to the patched version on their own now.
   # js-cookie, shell-quote and form-data@4 were retired by bumping react-use to 17.6.1 (js-cookie ^3),
   # concurrently to 9.2.4 (shell-quote 1.9.0) and @redocly/cli to 1.34.17 (form-data 4.0.6) instead.
-  # GHSA-mh99-v99m-4gvg is still reported by `pnpm audit` as high, but it is now FIXABLE — re-verified
-  # 2026-08-03. The advisory was re-split upstream into per-major ranges, and both lines we resolve have
-  # since shipped a patch: `<1.1.17` (we resolve 1.1.16; 1.1.17 published 2026-07-29) and
-  # `>=2.0.0 <2.1.3` (we resolve 2.1.2; 2.1.3 published 2026-07-28). The earlier note here — "no patch
-  # coming", "a pin cannot close it" — stopped being true on those dates: re-pinning brace-expansion@1
-  # and @2 closes both. That is a dependency change needing its own lockfile and test run, so it is
-  # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
-  # risk stays acceptable — but NOT for the "devDependency-only" reason this note used to give, which was
-  # wrong. Both lines reach brace-expansion via minimatch, and both are rooted in production chains, not
-  # only dev ones: the 2.x line through `googleapis` and `@boxyhq/saml-jackson` (both apps/web
-  # `dependencies`, via gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex` (a packages/ai
-  # `dependency`); the 1.x line through node-gyp under that same saml-jackson's typeorm>sqlite3.
-  # `pnpm audit` marks both findings `dev: false`. By contributor, the 1.x line is dominated by the
-  # eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next, the storybook
-  # eslint plugins) and the 2.x line by vite-plugin-dts>@vue/language-core across several workspace
-  # packages, with @redocly/cli and the prettier plugins mostly on the 2.x side.
+  # === Fixable, deferred to its own PR (ENG-2234) — brace-expansion ===
+  # NOT in the list above: every brace-expansion line we resolve (1.1.16, 2.1.2, 5.0.8) is still
+  # vulnerable, and all three are closable by re-pinning. Re-verified 2026-08-05. Two advisories now:
+  # - GHSA-mh99-v99m-4gvg (unbounded expansion length → OOM), re-split upstream into per-major ranges:
+  #   `<1.1.17` and `>=2.0.0 <2.1.3`. The 5.x line is clear of this one.
+  # - GHSA-rgw5-rvv9-x895 (unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation) covers
+  #   all three: `<1.1.18`, `>=2.0.0 <2.1.4` and `>=4.0.0 <5.0.9`.
+  # Pinning the current maintenance releases — 1.1.18 / 2.1.4 / 5.0.9, all published 2026-07-30 and all
+  # past the `minimumReleaseAge` cooldown below — closes both advisories on all three lines. That is a
+  # dependency change needing its own lockfile and test run, so it is deferred to ENG-2234 rather than
+  # folded into an audit-comment sweep; the note here goes away with that PR. (An earlier version of this
+  # note claimed no patch was coming and that a pin could not close it. Both stopped being true on
+  # 2026-07-28.)
+  # Until then the residual risk stays acceptable — but NOT for the "devDependency-only" reason this note
+  # used to give, which was wrong. Every line reaches brace-expansion via minimatch, and they are rooted
+  # in production chains, not only dev ones: the 2.x line through `googleapis` and `@boxyhq/saml-jackson`
+  # (both apps/web `dependencies`, via gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex`
+  # (a packages/ai `dependency`); the 1.x line through node-gyp under that same saml-jackson's
+  # typeorm>sqlite3. `pnpm audit` marks these findings `dev: false`. By contributor, the 1.x line is
+  # dominated by the eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next,
+  # eslint-plugin-storybook) and the 2.x line by vite-plugin-dts>@vue/language-core across six workspace
+  # packages, plus both prettier plugins. @redocly/cli is the only root on both lines, mostly on 2.x.
   # What makes it tolerable is the input, not the dependency type: every one of these call sites expands
   # glob patterns supplied by library code or developer config (gaxios' cache cleanup, typeorm's entity
   # discovery, build and lint globs), never by request data.
   # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
   # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
-  # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.
+  # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-05.
   # @better-auth/oauth-provider 1.6.23 (used for the MCP OAuth flow, ENG-1055) fails to bind
   # access-token audiences to their authorization grant (RFC 8707 resource-indicator confusion) — a
   # client can complete a normal authorization flow and then request a JWT scoped to a resource server
-  # the authorization never covered. No pin can close it: the whole 1.6.x line is in range (latest is
-  # 1.6.25) and the patch ships only from 1.7.0-beta.4, a line that is still unreleased (currently at
-  # 1.7.0-rc.2).
-  # Why this deployment isn't exposed today — and what that rests on:
+  # the authorization never covered. No pin can close it: the whole 1.6.x stable line is in range, and the
+  # patch ships only from 1.7.0-beta.4 — a line still in pre-release. (Deliberately version-agnostic: the
+  # `latest` and `rc` tags move about weekly, so naming them here just guarantees this note rots. The
+  # rc-specific claims below are pinned to 1.7.0-rc.4, the release actually inspected.)
+  # Why this deployment isn't exposed to the escalation — and what that rests on. Note the advisory is
+  # explicit that a single-audience deployment still does not bind the resource to the authorization
+  # grant, so the spec-compliance gap remains regardless of the below; this is "not exploitable here",
+  # not "compliant":
   # - apps/web/modules/auth/lib/mcp-oauth-provider-options.ts sets a single-entry validAudiences (the
   #   MCP resource URL), so there is no second audience for a client to escalate into. This is the
   #   load-bearing mitigation, and it holds only while that array stays single-entry.
@@ -163,11 +173,11 @@ overrides:
   # to us (migration 20260702093000_add_better_auth_oauth_provider already created `oauthClient`, not
   # `oauthApplication`, with an `oauthAccessToken.token` column, not `.accessToken`, plus the `sessionId`
   # that session-bound token validity needs — and we have no SCIM), and customAccessTokenClaims.resource
-  # does not apply either (ours only returns email/name). But there is still real DDL to write:
-  # 1.7.0-rc.2 replaces `validAudiences` with a resources model and declares three tables we do not have
-  # — `oauthResource`, `oauthClientResource` and `oauthClientAssertion`. The bigger blocker is that
-  # @better-auth/oauth-provider@1.7.0-rc.2 declares `better-auth` and `@better-auth/core` as
-  # `^1.7.0-rc.2` peers, making this a coupled auth-core upgrade touching login/2FA/SSO/sessions rather
+  # does not apply either (ours only returns email/name). But there is still real DDL to write: as
+  # inspected at 1.7.0-rc.4, the line replaces `validAudiences` with a resources model and declares three
+  # tables we do not have — `oauthResource`, `oauthClientResource` and `oauthClientAssertion`. The bigger
+  # blocker is that the same release declares `better-auth` and `@better-auth/core` as matching
+  # `^1.7.0-rc.x` peers, making this a coupled auth-core upgrade touching login/2FA/SSO/sessions rather
   # than a plugin-only bump. That needs its own reviewed PR once Better Auth ships 1.7.0 stable, not an
   # audit sweep or an unattended dependency-only bump.
 
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -125,14 +125,17 @@ overrides:
   # and @2 closes both. That is a dependency change needing its own lockfile and test run, so it is
   # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
   # risk stays acceptable — but NOT for the "devDependency-only" reason this note used to give, which was
-  # wrong. Both lines reach brace-expansion via minimatch, and the 2.x line does so through production
-  # chains as well: `googleapis` and `@boxyhq/saml-jackson` (both apps/web `dependencies`, via
-  # gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex` (a packages/ai `dependency`). Only
-  # the 1.x line is dev-tooling-only (eslint, prettier plugins, @redocly/cli, and node-gyp under
-  # typeorm's sqlite3); the bulk of the 2.x line is vite-plugin-dts>@vue/language-core across several
-  # workspace packages. What makes it tolerable is the input, not the dependency type: every one of these
-  # call sites expands glob patterns supplied by library code or developer config (gaxios' cache cleanup,
-  # typeorm's entity discovery, build and lint globs), never by request data.
+  # wrong. Both lines reach brace-expansion via minimatch, and both are rooted in production chains, not
+  # only dev ones: the 2.x line through `googleapis` and `@boxyhq/saml-jackson` (both apps/web
+  # `dependencies`, via gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex` (a packages/ai
+  # `dependency`); the 1.x line through node-gyp under that same saml-jackson's typeorm>sqlite3.
+  # `pnpm audit` marks both findings `dev: false`. By contributor, the 1.x line is dominated by the
+  # eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next, the storybook
+  # eslint plugins) and the 2.x line by vite-plugin-dts>@vue/language-core across several workspace
+  # packages, with @redocly/cli and the prettier plugins mostly on the 2.x side.
+  # What makes it tolerable is the input, not the dependency type: every one of these call sites expands
+  # glob patterns supplied by library code or developer config (gaxios' cache cleanup, typeorm's entity
+  # discovery, build and lint globs), never by request data.
   # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
   # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
   # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -122,6 +122,40 @@ overrides:
   # are devDependency-only (eslint, prettier plugins and @redocly/cli, all via minimatch) and never run
   # against untrusted input, so the residual risk is accepted until their consumers move to
   # brace-expansion@5 — a pin cannot close it.
+  # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
+  # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
+  # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.
+  # @better-auth/oauth-provider 1.6.23 (used for the MCP OAuth flow, ENG-1055) fails to bind
+  # access-token audiences to their authorization grant (RFC 8707 resource-indicator confusion) — a
+  # client can complete a normal authorization flow and then request a JWT scoped to a resource server
+  # the authorization never covered. No pin can close it: the whole 1.6.x line is in range (latest is
+  # 1.6.25) and the patch ships only from 1.7.0-beta.4, a line that is still unreleased (currently at
+  # 1.7.0-rc.2).
+  # Why this deployment isn't exposed today — and what that rests on:
+  # - apps/web/modules/auth/lib/mcp-oauth-provider-options.ts sets a single-entry validAudiences (the
+  #   MCP resource URL), so there is no second audience for a client to escalate into. This is the
+  #   load-bearing mitigation, and it holds only while that array stays single-entry.
+  # - apps/web/modules/mcp/auth.ts additionally pins `audience: getMcpResourceUrl()` on the resource
+  #   server's own verifyAccessToken call. That hardens the MCP endpoint but is NOT a substitute for the
+  #   above: the option flows straight into jose's jwtVerify, whose `aud` check is a membership test, so
+  #   a token whose `aud` array also lists other audiences still passes. The advisory's workaround also
+  #   requires rejecting multi-audience tokens, which verifyOptions.audience cannot express — that would
+  #   need an explicit assert on the returned payload. And if a second audience were ever added, the
+  #   party at risk is that other resource server, which our own pin does nothing for.
+  # - Test coverage is uneven: apps/web/modules/mcp/auth.test.ts asserts the audience/issuer pin, so
+  #   that leg is guarded. Single-entry validAudiences is asserted nowhere — that is the assert worth
+  #   adding before anyone touches this config.
+  # Upgrade cost, for whoever picks up 1.7.0: the renames the 1.7 upgrade guide warns about do not apply
+  # to us (migration 20260702093000_add_better_auth_oauth_provider already created `oauthClient`, not
+  # `oauthApplication`, with an `oauthAccessToken.token` column, not `.accessToken`, plus the `sessionId`
+  # that session-bound token validity needs — and we have no SCIM), and customAccessTokenClaims.resource
+  # does not apply either (ours only returns email/name). But there is still real DDL to write:
+  # 1.7.0-rc.2 replaces `validAudiences` with a resources model and declares three tables we do not have
+  # — `oauthResource`, `oauthClientResource` and `oauthClientAssertion`. The bigger blocker is that
+  # @better-auth/oauth-provider@1.7.0-rc.2 declares `better-auth` and `@better-auth/core` as
+  # `^1.7.0-rc.2` peers, making this a coupled auth-core upgrade touching login/2FA/SSO/sessions rather
+  # than a plugin-only bump. That needs its own reviewed PR once Better Auth ships 1.7.0 stable, not an
+  # audit sweep or an unattended dependency-only bump.
 
 autoInstallPeers: true
 linkWorkspacePackages: true
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -124,8 +124,15 @@ overrides:
   # coming", "a pin cannot close it" — stopped being true on those dates: re-pinning brace-expansion@1
   # and @2 closes both. That is a dependency change needing its own lockfile and test run, so it is
   # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
-  # risk stays acceptable for the same reason as before: these paths are devDependency-only (eslint,
-  # prettier plugins and @redocly/cli, all via minimatch) and never run against untrusted input.
+  # risk stays acceptable — but NOT for the "devDependency-only" reason this note used to give, which was
+  # wrong. Both lines reach brace-expansion via minimatch, and the 2.x line does so through production
+  # chains as well: `googleapis` and `@boxyhq/saml-jackson` (both apps/web `dependencies`, via
+  # gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex` (a packages/ai `dependency`). Only
+  # the 1.x line is dev-tooling-only (eslint, prettier plugins, @redocly/cli, and node-gyp under
+  # typeorm's sqlite3); the bulk of the 2.x line is vite-plugin-dts>@vue/language-core across several
+  # workspace packages. What makes it tolerable is the input, not the dependency type: every one of these
+  # call sites expands glob patterns supplied by library code or developer config (gaxios' cache cleanup,
+  # typeorm's entity discovery, build and lint globs), never by request data.
   # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
   # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
   # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.


xernobyl (dismissed):
Verified the claims independently and they hold: npm `latest` for `@better-auth/oauth-provider` is 1.6.25, still inside the advisory's `>=1.4.8, <1.7.0-beta.4` range (so "no fix on the 1.6.x line" is correct — bumping to latest wouldn't help), 1.7.0 is still unreleased (`rc` = 1.7.0-rc.2), and `validAudiences: [getMcpResourceUrl()]` is genuinely single-entry since the helper returns one string. Deferring is the right call. Approving — notes below are all non-blocking.

**1. The comment cites the weaker half of our defense.** It rests the argument entirely on single-entry `validAudiences`, but `apps/web/modules/mcp/auth.ts` already pins `audience: getMcpResourceUrl()` on `verifyAccessToken` — that's the advisory's *second* documented workaround, and it's the durable one. The `validAudiences` argument stops holding the moment a second audience is added; the resource-server `aud` check keeps holding regardless. Worth a sentence, since it turns a contingent acceptance into a robust one. (Neither invariant is asserted by a test today — a one-line assert in `mcp-oauth-dcr.test.ts` would make it durable rather than incidental.)

**2. This GHSA is already tracked: ENG-1703** (Dependabot #613, same advisory, closed Done 2026-07-16 with the same conclusion), so the "No ticket" line in the description isn't quite right. Not an argument against the PR — a `pnpm audit` rationale belongs in the file the next auditor actually opens, not only in a closed ticket — but the comment should reference ENG-1703, matching the `# ENG-XXXX / GHSA-...` convention used by every other entry in this file. (Also note `pnpm audit` here is a separate channel from the Dependabot alerts, which are already dismissed as tolerable risk.)

**3. The migration cost is overstated.** `customAccessTokenClaims.resource becomes a resources[] array` doesn't apply to us — our `customAccessTokenClaims` only returns `email`/`name`, it never touches `resource`. And the schema migration is cheaper than the 1.7 upgrade guide's generic warning suggests: we have no SCIM, and migration `20260702093000_add_better_auth_oauth_provider` already created `oauthClient` plus an `oauthAccessToken` whose column is already named `token`, so the `oauthApplication` → `oauthClient` move and the `accessToken` → `token` rename are both no-ops here.

Conversely, one cost the comment *doesn't* mention is the bigger one: `@better-auth/oauth-provider@1.7.0-rc.2` declares `better-auth: ^1.7.0-rc.2` and `@better-auth/core: ^1.7.0-rc.2` as peers, so this is a coupled auth-core upgrade touching login/2FA/SSO/sessions — not a plugin bump. That's the strongest reason it needs its own reviewed PR.

**4. Placement nit:** the block sits after `valibot: 1.4.2` with a blank line between, so it reads as attached to the top-level `autoInstallPeers:` / `linkWorkspacePackages:` keys rather than to `overrides:`. Every other comment here sits directly above its pin; since this one deliberately has no pin, an explicit `# === Accepted risk, no override ===` header would remove the ambiguity.

CodeRabbit's `npx auth migrate` thread is already addressed in the current head. LGTM 👍

xernobyl (changes requested):
Re-reviewed against head `3b5663b3`.

`3b5663b3` addressed all four points from the last round — the durable-workaround sentence, the ENG-1703 reference, the migration-cost correction plus the peer-coupling blocker, and the explicit `# === Accepted risk, no override ===` header. CodeRabbit's inline thread (soften `npx auth migrate`) is moot: the current text doesn't mention that command at all. Sonar gate green, all checks pass.

Nothing below is a live exposure — the deployment is fine today. These are accuracy defects in the note itself, which matters because its job is to be the record someone trusts later. Two of them trace back to my own wording last time, which this head transcribed faithfully — apologies for sending you down that path.

**[P2] `pnpm-workspace.yaml:158-160` — "the durable one: it keeps holding even if a second audience is ever added" isn't right.** My error from the last round. `verifyOptions.audience` flows into jose's `jwtVerify` (`@better-auth/core/dist/oauth2/verify.mjs`), and jose's `checkAudiencePresence` is a *membership* test — for an array `aud` it passes if any expected value is present. The advisory's workaround #2 is two-part: accept only when its own identifier is the expected audience, **and** reject tokens whose `aud` is an array that also lists other audiences. We implement the first half only, and `verifyOptions.audience` can't express the second — that needs an explicit assert on the returned payload. Separately, if a second audience were ever added, the party at risk is *that* resource server, which our own pin does nothing for. Accurate framing: both mitigations are contingent on `validAudiences` staying single-entry; the resource-server pin hardens the MCP endpoint but isn't a substitute for it.

**[P2] `pnpm-workspace.yaml:164` — "the schema migration is largely a no-op here" is overstated**, and sits oddly next to the same sentence's admission that the resources model is new. `@better-auth/oauth-provider@1.7.0-rc.2` declares three tables `20260702093000` never created: `oauthResource` (unique `identifier`, `name`, `accessTokenTtl`, `refreshTokenTtl`, `signingAlgorithm`), `oauthClientResource` (join table, `clientId` FK on-delete-cascade + `resourceId` FK) and `oauthClientAssertion`. The rename half of the claim is solid — our migration did create `oauthClient` and `oauthAccessToken.token`, and `sessionId` already exists on that table, so the session-binding column is there. So the true statement is narrower: the renames the upgrade guide warns about don't apply to us, but the resources model still needs real DDL. As written, whoever picks up the 1.7 upgrade budgets for nothing and finds three tables.

**[P3] `pnpm-workspace.yaml:160` — "(Neither invariant is asserted by a test today.)" is half wrong, and backwards from what a reader would act on.** Also mine. `apps/web/modules/mcp/auth.test.ts:268` asserts `verifyAccessTokenMock` was called with `audience: "https://app.example.com/api/mcp"` — drop or change the pin and that test fails, so the resource-server invariant *is* guarded. `validAudiences` is the uncovered one (no test in `apps/web` references it; `apps/web/modules/auth/lib/mcp-oauth-dcr.test.ts:44` is jwt-plugin setup, not an assertion). Which inverts the suggested follow-up: the assert worth adding is on single-entry `validAudiences`, since that's the leg nothing is holding.

**[P3] Placement and staleness.** The block sits at top level after the `overrides:` map, while the other accepted-risk-without-a-pin note (brace-expansion, lines 120-124 on main) lives *inside* `overrides:` under `# === Removed once upstream caught up ===`. Post-merge the file carries two conventions for the same thing. The header I asked for last time does resolve the ambiguity I raised, so this is a consolidation preference, not a defect. Worth rebasing first regardless — the branch is 22 commits behind and `pnpm-workspace.yaml` moved on main via #8663, so the placement call is better made against that file.

**Verified clean:** `1.6.23` still pinned; `latest` = 1.6.25, inside `>=1.4.8, <1.7.0-beta.4`; `rc` = 1.7.0-rc.2; patched version is genuinely `1.7.0-beta.4`; single-entry `validAudiences` (`mcp-oauth-provider-options.ts:20`); the `audience`/`issuer` pin (`modules/mcp/auth.ts:314`); `customAccessTokenClaims` returns only email/name; no SCIM anywhere; `validAudiences` is indeed gone from rc.2 and replaced by the resources model; rc.2's peers are exactly `better-auth: ^1.7.0-rc.2` and `@better-auth/core: ^1.7.0-rc.2`.

One I couldn't check: "ENG-1703 … closed Done" on line 148 — I didn't have Linear access this session. (Dependabot has two alerts for this GHSA, #612 and #613, both dismissed `tolerable_risk`; the note cites only #613.)

Fix the two P2 sentences and this is good to merge — the underlying judgment to defer is still right.

mattinannt:
@xernobyl thanks — I checked all four points independently before changing anything, and all four hold. Rebased onto current `main` and force-pushed; head is now `a72fdcf`.

**[P2] the "durable one" claim** — confirmed wrong, removed. `verifyOptions` goes straight into `jwtVerify` (`@better-auth/core/dist/oauth2/verify.mjs:64`), and jose's `checkAudiencePresence` is literally `audOption.some((aud) => audPayload.includes(aud))` — a membership test, so a token whose `aud` array also lists other audiences passes. Reframed as you described: single-entry `validAudiences` is the load-bearing mitigation and holds only while that array stays single-entry; the resource-server pin hardens the MCP endpoint but isn't a substitute, `verifyOptions.audience` can't express the reject-multi-audience half, and if a second audience were added the party at risk is that other resource server, which our pin does nothing for.

**[P2] "largely a no-op"** — confirmed overstated, narrowed. Pulled the `1.7.0-rc.2` tarball: `oauthResource`, `oauthClientResource` and `oauthClientAssertion` are all declared, and none of the three exist in `main.prisma` or `20260702093000`. The note now says the *renames* don't apply to us (we already have `oauthClient`, `oauthAccessToken.token`, `sessionId`, no SCIM) but that the resources model still needs real DDL — three tables, named.

**[P3] test coverage** — confirmed half wrong, inverted. `apps/web/modules/mcp/auth.test.ts:315-321` does assert the `audience`/`issuer` pin, so that leg is guarded; nothing in `apps/web` asserts `validAudiences`. The note now says so and points the follow-up assert at the uncovered leg.

**[P3] placement** — moved inside `overrides:`, directly after the brace-expansion note, under `# === Accepted risk, no override possible yet ===`. One convention in the file now.

CodeRabbit's inline thread is resolved as moot — `npx auth migrate` isn't in the text.

**Re-verified today (2026-08-03), since this sat a few days:** `latest` still 1.6.25, `rc` still `1.7.0-rc.2`, no stable 1.7.0, peers still `^1.7.0-rc.2` on both `better-auth` and `@better-auth/core`. Also worth noting the 3-day `minimumReleaseAge` cooldown is *not* what blocks the upgrade — rc.2 was published 2026-07-22 and clears it; `pnpm install` re-verified all 2352 lockfile entries against the policy and passed.

**Two things moved on `main` that the description now reflects:** `pnpm audit` is at moderate: 3, high: 3 (not 1/1). The new ones are `vite` 7.3.3 (high `GHSA-fx2h-pf6j-xcff` + moderate `GHSA-v6wh-96g9-6wx3`, patched 7.3.5) and `sanitize-html` 2.17.4 (moderate `GHSA-vccv-cmxp-4j9h`, patched 2.17.5). Separately, `GHSA-mh99-v99m-4gvg` has been re-split upstream and **now has patches in both lines** (1.1.17 / 2.1.3), which makes the existing brace-expansion note's "no patch coming — a pin cannot close it" stale and that finding closable. All three are real dependency changes, so I left them out to keep this comment-only; they're listed under "Risks & regressions" and want their own PR.

Your Linear point I couldn't verify this session (no Linear access) — the note still asserts ENG-1703 is closed Done, worth a glance before merge. I did add the `#612` / `#613` alert refs.

---

xernobyl (dismissed):
Re-reviewed `a72fdcf`. Merge-base is now exactly `main` (`34d5e468`), so the rebase took. Both P2s are genuinely resolved — clearing my changes-request.

I re-verified each claim against current `main` rather than the summary:

- **Durability claim properly inverted.** Single-entry `validAudiences` is now the load-bearing mitigation "and it holds only while that array stays single-entry", the resource-server pin is explicitly "NOT a substitute", and the membership-test reasoning, the can't-express-reject-multi-audience point, and the other-resource-server-is-the-victim point are all stated correctly.
- **Migration cost narrowed correctly.** Renames confirmed non-applicable (`oauthClient`, `oauthAccessToken.token`, `sessionId` all present, no SCIM), and the three tables are real: `oauthResource`, `oauthClientResource` and `oauthClientAssertion` are declared in rc.2 and absent from our schema.
- **Test coverage inverted correctly.** `apps/web/modules/mcp/auth.test.ts:315-320` asserts the `audience`/`issuer` pin; `validAudiences` appears in no test under `apps/web`.
- **Placement** now inside `overrides:`, and the Dependabot refs (#612 *and* #613, both dismissed) are right.

Re-checked today: `1.6.23` still pinned, `latest` still 1.6.25 (inside `>= 1.4.8, < 1.7.0-beta.4`), `rc` still `1.7.0-rc.2`, no stable 1.7.0, peers still `^1.7.0-rc.2` on both packages. The description's audit numbers are exact — moderate: 3, high: 3, with vite 7.3.3 (`GHSA-fx2h-pf6j-xcff` high + `GHSA-v6wh-96g9-6wx3` moderate, patched ≥7.3.5) and sanitize-html 2.17.4 (`GHSA-vccv-cmxp-4j9h`, patched ≥2.17.5).

**[P3] One thing I'd fold in before merge.** Your brace-expansion finding is right, and it has a consequence for this PR. The advisory was re-split upstream on 2026-07-31 into four ranges (`< 1.1.17`, `>= 2.0.0 < 2.1.3`, `>= 3.0.0 < 3.0.3`, `>= 4.0.0 < 5.0.8`), and `pnpm audit` now reports it as **fixable** — `patched >= 1.1.17` against our 1.1.16, `patched >= 2.1.3` against our 2.1.2. That makes lines 120-124 false on two counts: "each the latest release in its line with no patch coming", and "a pin cannot close it". The `(brace-expansion <=5.0.7)` label on line 120 is stale too.

Your reasoning for deferring — "all three are real dependency changes, so I left them out to keep this comment-only" — is right for the *bumps*; those want their own PR. But correcting that sentence is itself a comment edit: no dependency change, same file, the paragraph immediately above the block you're adding. As it stands this lands a "Re-verified 2026-08-03" stamp directly beneath a claim that stopped being true on 2026-07-28. I'd fix the one sentence here and leave the pins to the follow-up. Not blocking.

Minor and self-resolving: "one convention in the file now" isn't quite there — the new `# === Accepted risk, no override possible yet ===` header sits *after* the brace-expansion note, which stays under `# === Removed once upstream caught up ===`. Once brace-expansion becomes a pin it stops being an accepted risk anyway, so I wouldn't restructure further.

One I still couldn't check: "the ticket is closed Done" on line 148 — no Linear access in my session either, so worth your glance before merge.

SonarQube, Docker Build and E2E were still pending on this head when I looked — merge once they're green.

xernobyl (changes requested):
Re-reviewed `1f2f3a4`. The correction is accurate — but the rewritten paragraph carries forward one clause that isn't, so holding here rather than re-approving.

**Verified good in the new commit:** the advisory really was re-split upstream (updated 2026-07-31) into per-major ranges `< 1.1.17`, `>= 2.0.0 < 2.1.3`, `>= 3.0.0 < 3.0.3`, `>= 4.0.0 < 5.0.8`; both dates are right (1.1.17 → 2026-07-29, 2.1.3 → 2026-07-28); we do resolve 1.1.16 and 2.1.2; "now FIXABLE" matches `pnpm audit` (`patched >= 1.1.17` / `>= 2.1.3`); and re-pinning genuinely closes it — 1.1.17 satisfies minimatch@3's `^1.1.7`, 2.1.3 satisfies minimatch@5/9's `^2.0.1`/`^2.0.2`. Deferring the pin to its own PR is right, and this stays comment-only. Both patched versions also clear the 3-day `minimumReleaseAge` gate already (as do 1.1.18 / 2.1.4 from 07-30), so the follow-up isn't cooldown-blocked.

**Please fix before merge — two clauses in the closing sentence are wrong.**

> Until then the residual risk stays acceptable for the same reason as before: these paths are devDependency-only (eslint, prettier plugins and @redocly/cli, all via minimatch) and never run against untrusted input.

1. **"devDependency-only" is false.** From the same `pnpm audit` run behind the rest of these numbers, brace-expansion 2.1.2 is reached through production chains:

```
apps/web    > googleapis > google-auth-library > gaxios@7.1.3 > rimraf@5 > glob > minimatch@9 > brace-expansion
apps/web    > @boxyhq/saml-jackson > @googleapis/admin > googleapis-common > google-auth-library > gaxios > rimraf > ...
apps/web    > @boxyhq/saml-jackson > typeorm > glob > minimatch > brace-expansion
packages/ai > @ai-sdk/google-vertex > google-auth-library > gaxios > rimraf > ...
```

`googleapis` and `@boxyhq/saml-jackson` are in `apps/web` `dependencies`, and `@ai-sdk/google-vertex` is in `packages/ai` `dependencies` — all on current `main`.

2. **The package list is wrong too.** The largest contributor to the 2.1.2 finding is `vite-plugin-dts > @vue/language-core`, across seven workspace packages, plus `typeorm > glob`. Neither is eslint, a prettier plugin, or redocly.

The *conclusion* still holds, so this is a wording fix rather than a re-assessment: the glob patterns under gaxios' rimraf cleanup and typeorm's entity discovery are library- and developer-supplied, not attacker-controlled. Keep "never run against untrusted input", drop "devDependency-only", and either name the real contributors or just say "all via minimatch, in both dev and prod dependency chains".

Flagging it now, rather than when I approved the previous head, because this commit rewrote that paragraph and re-dated it to 2026-08-03 — so the sentence reads as a current finding, and anyone auditing prod-vs-dev exposure will take it at face value.

Everything else on this head is green, including the three checks that were pending earlier.

mattinannt:
Right on both counts — fixed in `f3d9645`. I checked before changing anything:

- `googleapis` (171.4.0) and `@boxyhq/saml-jackson` (26.2.0) are both in `apps/web` **`dependencies`**, and `@ai-sdk/google-vertex` (4.0.126) is in `packages/ai` **`dependencies`**, all on current `main`. The 2.1.2 finding reaches them via `gaxios>rimraf>glob>minimatch` and `typeorm>glob>minimatch`. `pnpm audit` also reports `dev: false` on both findings. "devDependency-only" was plainly false.
- The package list was wrong too, as you said. For the 2.x line the bulk is `vite-plugin-dts > @vue/language-core`, plus `@redocly/cli` and the prettier sort-imports plugin. The eslint/prettier/redocly naming actually describes the **1.x** line (which also picks up `node-gyp` under typeorm's `sqlite3`), so the note had the two lines' contributors swapped.

The rewrite drops "devDependency-only", splits the two lines' contributors correctly, and rests the conclusion on the input rather than the dependency type — the glob patterns come from library code and developer config (gaxios' cache cleanup, typeorm's entity discovery, build and lint globs), never from request data. Your framing, essentially.

Worth flagging since this note's whole job is to be accurate: I got **six** workspace packages for `vite-plugin-dts>@vue/language-core`, not seven — `cache`, `database`, `jobs`, `js-core`, `survey-ui`, `workflows`. Rather than assert a number that might drift, the note names the chain without a count.

Also confirming your cooldown check independently: 1.1.17 (2026-07-29) and 2.1.3 (2026-07-28) both clear the 3-day `minimumReleaseAge`, so the follow-up pin PR isn't gated on it.

Fair call on flagging it at this head rather than the last one — re-dating the paragraph to 2026-08-03 is exactly what would have made a stale claim read as a fresh finding.

---

xernobyl (dismissed):
Re-reviewed `f3d9645`. Both clauses are fixed correctly, and the rewrite is better than what I asked for — resting the conclusion on the input rather than the dependency type is the right basis. Clearing the changes-request.

And you're right about the count: six workspace packages on the `vite-plugin-dts > @vue/language-core` chain, not seven — `cache`, `database`, `jobs`, `js-core`, `survey-ui`, `workflows`. My error; naming the chain without a number is the better call anyway.

Verified against the audit run and current `main`:

- The three prod chains are exactly as stated — `googleapis` and `@boxyhq/saml-jackson` in `apps/web` `dependencies` (via `gaxios>rimraf>glob` and `typeorm>glob`), `@ai-sdk/google-vertex` in `packages/ai` `dependencies`.
- The two lines' contributors are now split correctly: the 1.1.16 finding has zero `vite-plugin-dts` paths and five `node-gyp`-under-`sqlite3` paths; the 2.1.2 finding has six `vite-plugin-dts` paths and zero `node-gyp`. That swap was real.
- `dev: false` on both findings, as you say.

Two precision nits, neither blocking, both on descriptive text rather than the load-bearing argument:

1. **`@redocly/cli` straddles both lines.** It's listed under the 1.x contributors, but it actually appears on both — three of its paths land on 2.1.2 via `@redocly/openapi-core > minimatch@5`, and only one on 1.1.16 via its own `glob > minimatch@3`. So if anything it belongs on the 2.x side.
2. **"Only the 1.x line is dev-tooling-only"** is true about the *tools* but not the graph: `node-gyp` is reached through `@boxyhq/saml-jackson`, which is an `apps/web` `dependency`, and `pnpm audit` marks both findings `dev: false`. Since the very next sentence disclaims dependency type as the basis, the simplest fix is to describe the 1.x contributors as build/lint tooling without the "only".

Neither changes the assessment, and the follow-up pin PR will delete this paragraph anyway — fold them in or leave them.

All checks green. Still worth your glance before merge: "the ticket is closed Done" on the ENG-1703 line is the one claim I've never been able to verify — no Linear access in my session.

mattinannt:
Folded both in — `38d3a46`. I re-derived the contributor roots by counting the second hop of every audit path rather than eyeballing, and that turned up a third error of the same kind:

| root | 1.1.16 | 2.1.2 |
| --- | --- | --- |
| `typescript-eslint` | 33 | – |
| `@vitest/eslint-plugin` | 30 | – |
| `eslint-config-next` | 13 | – |
| `eslint-plugin-storybook` | 9 | – |
| `@boxyhq/saml-jackson` | 5 | 2 |
| `eslint` / `-react-hooks` / `-react-refresh` | 3 / 3 / 3 | – |
| `@redocly/cli` | 1 | 3 |
| `vite-plugin-dts` | – | 6 |
| `@trivago/prettier-plugin-sort-imports` | – | 2 |
| `prettier-plugin-tailwindcss` | – | 2 |
| `googleapis` | – | 2 |
| `@ai-sdk/google-vertex` | – | 1 |

1. **`@redocly/cli`** — 1 path on 1.x, 3 on 2.x, exactly as you said. Moved to the 2.x side.
2. **"Only the 1.x line is dev-tooling-only"** — dropped. Those 5 `@boxyhq/saml-jackson` paths are the `node-gyp`-under-`typeorm>sqlite3` chain, and saml-jackson is an `apps/web` `dependency`, so *both* findings are production-rooted. The note now says so and cites `dev: false` on both.
3. **The prettier plugins were on the wrong line too** — `@trivago/prettier-plugin-sort-imports` (2) and `prettier-plugin-tailwindcss` (2) appear only on 2.1.2, but the note had them under 1.x. Same swap, one package further. The 1.x line is now described as what it actually is: the eslint plugin graph.

One caveat worth recording since it bounds the table: `pnpm audit` caps `paths` at 100 per finding, and 1.1.16 hits exactly 100. So the 1.x counts are a floor and that line's list may be incomplete — which is why the note says "dominated by" rather than enumerating it. The 2.1.2 list is complete at 18, so the six `vite-plugin-dts` packages and the three prod chains are exact.

Also confirming your read that this paragraph is temporary: the follow-up pin PR deletes it, so none of this outlives the bump.

Flagging that this push may have dismissed your approval depending on the branch setting — sorry for the extra round. You'd said "fold them in or leave them", and given this note's only job is to be the accurate record, three wrong attributions felt worth the round trip.

ENG-1703's "closed Done" remains the one unverified claim, on both our sides — over to @mattinannt for that glance.

---

pnpm-workspace.yaml
```diff
@@ -117,11 +117,59 @@ overrides:
   # linkify-it, protobufjs@7, undici@7 and uuid@11 all resolve to the patched version on their own now.
```
xernobyl: **[P2] This enumeration now contradicts the paragraph directly below it.**

Line 116, which the PR doesn't touch, still reads:

> `# ajv@6, @babel/core, body-parser, brace-expansion@1/@2/@5, dompurify, …`

…landing on this line's "all resolve to the patched version on their own now", under a header that says *"kept as a record of what no longer needs a pin"*. The rewritten paragraph immediately below says 1.1.16 and 2.1.2 are both still vulnerable and closable **only** by re-pinning. Both can't be true.

Only `@5` belongs there — we resolve 5.0.8, which is past the `>=4.0.0 <5.0.8` range. So `brace-expansion@1/@2` needs dropping from line 116.

The header compounds it: a paragraph whose whole thesis is "re-pinning closes both, deferred to a follow-up" is sitting under *"what no longer needs a pin"*. Either move the paragraph under its own header, or fold it in next to the new `=== Accepted risk ===` block.

Flagging it because it's the same error class the PR set out to fix, two lines above the fix — and an auditor reading top-to-bottom hits the false claim first.

pnpm-workspace.yaml
```diff
@@ -117,11 +117,59 @@ overrides:
...
+  # discovery, build and lint globs), never by request data.
+  # === Accepted risk, no override possible yet (pnpm audit will keep flagging this) ===
+  # ENG-1703 / GHSA-p2fr-6hmx-4528 (Dependabot alerts #612 and #613, both dismissed as tolerable risk;
+  # the ticket is closed Done with the same conclusion as here). Re-verified 2026-08-03.
+  # @better-auth/oauth-provider 1.6.23 (used for the MCP OAuth flow, ENG-1055) fails to bind
+  # access-token audiences to their authorization grant (RFC 8707 resource-indicator confusion) — a
+  # client can complete a normal authorization flow and then request a JWT scoped to a resource server
+  # the authorization never covered. No pin can close it: the whole 1.6.x line is in range (latest is
+  # 1.6.25) and the patch ships only from 1.7.0-beta.4, a line that is still unreleased (currently at
+  # 1.7.0-rc.2).
```
xernobyl: **[P2] These numbers went stale two days after the `Re-verified 2026-08-03` stamp.**

- `latest is 1.6.25` → npm `latest` is now **1.6.26**, published 2026-08-04
- `currently at 1.7.0-rc.2` → the `rc` tag is now **1.7.0-rc.4**, published 2026-08-05
- and lines 167-170 reason from `1.7.0-rc.2` specifically

Every *conclusion* survives — I pulled the rc.4 tarball to be sure: 1.6.26 is still inside `>= 1.4.8, < 1.7.0-beta.4`, there's still no stable 1.7.0, `validAudiences` is still gone, `oauthResource` / `oauthClientResource` / `oauthClientAssertion` are all still declared, and the peers are still coupled (now `^1.7.0-rc.4`). So the reasoning is untouched; only the numbers rotted.

Which is the thing this PR is itself about — a fresh date stamp over facts with a roughly one-week half-life. Suggest making it version-agnostic so it stops needing re-dating:

> No pin can close it: the whole 1.6.x stable line is in range, and the patch ships only from 1.7.0-beta.4 — a line still in pre-release.

Then the rc-specific claims below can stay pinned to the version you actually inspected, as long as they say "as inspected at rc.2" rather than reading as "the current rc".

pnpm-workspace.yaml
```diff
@@ -117,11 +117,59 @@ overrides:
...
+  # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
+  # risk stays acceptable — but NOT for the "devDependency-only" reason this note used to give, which was
+  # wrong. Both lines reach brace-expansion via minimatch, and both are rooted in production chains, not
+  # only dev ones: the 2.x line through `googleapis` and `@boxyhq/saml-jackson` (both apps/web
+  # `dependencies`, via gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex` (a packages/ai
+  # `dependency`); the 1.x line through node-gyp under that same saml-jackson's typeorm>sqlite3.
+  # `pnpm audit` marks both findings `dev: false`. By contributor, the 1.x line is dominated by the
+  # eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next, the storybook
+  # eslint plugins) and the 2.x line by vite-plugin-dts>@vue/language-core across several workspace
+  # packages, with @redocly/cli and the prettier plugins mostly on the 2.x side.
```
xernobyl: **[P3] Two small things in this sentence.**

1. `the storybook eslint plugins` is plural, but there's one — `eslint-plugin-storybook`. The other eslint devDep in `apps/storybook` is `eslint-plugin-react-refresh`, which isn't a storybook plugin.
2. `with @redocly/cli and the prettier plugins mostly on the 2.x side` — the clause is redundant (the sentence is already describing the 2.x line), and "mostly" is only true of `@redocly/cli` (1 path on 1.x, 3 on 2.x). Both prettier plugins are **exclusively** 2.x, with zero paths on 1.x.

`@redocly/cli` being the only root that straddles both lines is the more useful fact, and it's the one that keeps tripping this paragraph up. Maybe:

> …and the 2.x line by `vite-plugin-dts>@vue/language-core` across several workspace packages, plus both prettier plugins. `@redocly/cli` is the only root on both lines, mostly on 2.x.

xernobyl (commented):
Re-reviewed at `38d3a46`. CI green, Sonar gate passed, and CodeRabbit's thread is resolved + moot — confirmed there's no `npx auth migrate` anywhere in the text.

**The one claim we'd both left open checks out:** ENG-1703 is genuinely `Done` (completed 2026-07-16, `security` label, assignee me), so the note's parenthetical is accurate. That was the last blocker from my side.

Since the whole PR is text, I re-derived the facts instead of reading for plausibility — rebuilt the dependency graph from the resolved lockfile rather than re-reading npm metadata. **The corrected attribution holds.** All four named chains reproduce exactly:

```
googleapis > google-auth-library > gaxios > rimraf > glob > minimatch > 2.1.2
@boxyhq/saml-jackson > typeorm > glob > minimatch > 2.1.2
@ai-sdk/google-vertex > google-auth-library > gaxios > rimraf > glob > minimatch > 2.1.2
@boxyhq/saml-jackson > typeorm > sqlite3 > node-gyp > glob > minimatch > 1.1.16
```

1.x is the eslint plugin graph, 2.x is `vite-plugin-dts>@vue/language-core` (six packages) plus the prettier plugins, `@redocly/cli` is the only root on both, and both lines are production-rooted via `apps/web` `dependencies`. Identical on current `main`, so the lockfile drift since your rebase didn't invalidate any of it.

Also confirmed against the tree and the tarballs: both advisory ranges and both brace-expansion patch dates to the day; single-entry `validAudiences` at `mcp-oauth-provider-options.ts:20`; the pin at `modules/mcp/auth.ts:319-323`; `opts.verifyOptions` flowing into `jwtVerify` at `@better-auth/core@1.6.23` `verify.mjs:64`, exactly the line you cite; jose 6.2.2's `checkAudiencePresence` being a membership test (it's `audOption.some(Set.prototype.has.bind(new Set(audPayload)))` in this version rather than the `.includes` form, same semantics — so the "not a substitute" reasoning is right); the test-coverage claim in both directions; `oauthClient` / `oauthAccessToken.token` / `sessionId` present, no SCIM, the three rc tables absent; and `customAccessTokenClaims` returning only email/name. The placement fix works too — the block is contiguous with the `overrides:` comments and separated from `autoInstallPeers:`, so the ambiguity I flagged is gone. All added lines are within 110 chars.

Two P2s inline, both accuracy rather than anything behavioural, plus two smaller ones. Neither P2 is your text — the first is a paragraph this PR didn't touch, and the second is just the world moving.

Two things outside the diff:

- **The description's vite bullet is stale.** #8724 landed vite 7.3.5 on `main` this morning, so "deliberately not fixed here — vite 7.3.3 … appeared after this branch was opened" and the `moderate: 3, high: 3` counts no longer match `main`. `sanitize-html` 2.17.4 is still there. Doesn't touch the committed text.
- **Optional half-sentence:** "Why this deployment isn't exposed today" is right about *escalation*, but the advisory notes the RFC 8707 binding gap remains even for single-audience deployments. Worth saying so nobody reads the note as "we're compliant".

I filed ENG-2234 for the brace-expansion re-pin so the "deferred to a follow-up PR" line points at something real, and noted there that its PR should delete the paragraph. Happy to push the fixes for the two P2s myself if you'd rather not do another round — say the word.

pnpm-workspace.yaml
```diff
@@ -117,11 +117,59 @@ overrides:
...
-  # are devDependency-only (eslint, prettier plugins and @redocly/cli, all via minimatch) and never run
-  # against untrusted input, so the residual risk is accepted until their consumers move to
-  # brace-expansion@5 — a pin cannot close it.
+  # GHSA-mh99-v99m-4gvg is still reported by `pnpm audit` as high, but it is now FIXABLE — re-verified
+  # 2026-08-03. The advisory was re-split upstream into per-major ranges, and both lines we resolve have
+  # since shipped a patch: `<1.1.17` (we resolve 1.1.16; 1.1.17 published 2026-07-29) and
+  # `>=2.0.0 <2.1.3` (we resolve 2.1.2; 2.1.3 published 2026-07-28). The earlier note here — "no patch
+  # coming", "a pin cannot close it" — stopped being true on those dates: re-pinning brace-expansion@1
+  # and @2 closes both. That is a dependency change needing its own lockfile and test run, so it is
+  # deferred to a follow-up PR rather than folded into an audit-comment sweep. Until then the residual
```
xernobyl: **[P3] Filed this as ENG-2234** so the "follow-up PR" here points at something real: https://linear.app/formbricks/issue/ENG-2234

Two notes for whoever picks it up, both in the ticket:

- Pin the **current maintenance releases** — 1.1.18 and 2.1.4, both published 2026-07-30 and both past the cooldown — not the minimum 1.1.17 / 2.1.3 this paragraph names. `maintenance-v1` and `maintenance-v2` point at them.
- That PR should **delete this whole paragraph**, since it exists only to explain the deferral, and fix the line 116 enumeration at the same time.

Swept for overlap before filing: no open PR touches brace-expansion, and ENG-1953/1954/1955 are a different brace-expansion advisory (the `{}`-group one), all Done.

mattinannt:
All four addressed in `55708a0`, rebased onto `4aabd0b`. Thanks for re-deriving from the lockfile rather than reading for plausibility — and for chasing ENG-1703 down; I confirmed it independently via Linear (completed 2026-07-16, `security` label), so that parenthetical stands.

**One correction to your line-117 P2, and it goes the other way.** You said only `@5` belongs in the enumeration because we resolve 5.0.8, past `>=4.0.0 <5.0.8`. That was true on 2026-08-03. It isn't now: a **second advisory** landed, [`GHSA-rgw5-rvv9-x895`](https://github.com/advisories/GHSA-rgw5-rvv9-x895) — *"DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation"* — and it covers all three lines we resolve:

| line | we resolve | GHSA-mh99-v99m-4gvg | GHSA-rgw5-rvv9-x895 |
| --- | --- | --- | --- |
| 1.x | 1.1.16 | `<1.1.17` | `<1.1.18` |
| 2.x | 2.1.2 | `>=2.0.0 <2.1.3` | `>=2.0.0 <2.1.4` |
| 5.x | 5.0.8 | clear | **`>=4.0.0 <5.0.9`** |

So `brace-expansion` is dropped from that enumeration **entirely**, not just `@1/@2` — no line resolves to a patched version on its own. Your instinct in ENG-2234 to pin the maintenance releases rather than the minimums turns out to be exactly right for a reason neither of us had yet: 1.1.18 / 2.1.4 are precisely this second advisory's patched versions. **5.0.9 needs adding to that ticket** (also 2026-07-30, also past cooldown) — I've noted it there.

The rest:

- **[P2] line 117 / header** — enumeration fixed as above, and the paragraph moved out from under *"Removed once upstream caught up"* into its own `# === Fixable, deferred to its own PR (ENG-2234) — brace-expansion ===`. You were right that this was the same error class the PR set out to fix, two lines above the fix.
- **[P2] line 147 rot** — made version-agnostic, close to your wording: *"the whole 1.6.x stable line is in range, and the patch ships only from 1.7.0-beta.4 — a line still in pre-release"*, with an explicit note that naming the moving tags just guarantees rot. I re-pulled the rc.4 tarball myself rather than take it on trust: peers are `^1.7.0-rc.4` on both packages, `validAudiences` still gone, all three tables still declared. The rc-specific claims are now pinned to "as inspected at 1.7.0-rc.4".
- **[P3] line 135** — both fixed. `eslint-plugin-storybook` singular; and the clause now reads *"…across six workspace packages, plus both prettier plugins. `@redocly/cli` is the only root on both lines, mostly on 2.x"*, which is your suggested phrasing — the prettier plugins really are 2.x-exclusive (0 paths on 1.x).
- **Optional caveat** — taken, and the advisory backs it word for word: *"Such a deployment still does not bind the resource to the authorization grant, so the spec-compliance gap remains even where cross-audience escalation does not apply."* The heading now reads "not exposed to the escalation" and says explicitly this is "not exploitable here", not "compliant".

**On the description:** you're right that the vite bullet was stale — #8724 landed 7.3.5, and vite is gone from the audit. But the counts moved much further than that. Fresh run on the rebased branch is **moderate: 13, high: 9**, because main's lockfile drifted: new findings in `postcss`, `socket.io-parser`, `undici` (×5 advisories), `fast-uri`, `ip-address` (×3) and `hono`, plus the second brace-expansion advisory. `sanitize-html` 2.17.4 is still outstanding as you say. Rather than enumerate a number with a one-day half-life, the description now points at ENG-2182 and ENG-2234 and drops the exact counts — same lesson as the P2 above.

No thanks needed on pushing the fixes yourself — but genuinely appreciated, and if a seventh round appears I may take you up on it.

---

pnpm-workspace.yaml
```diff
@@ -113,15 +113,73 @@ overrides:
...
+  # 2026-07-28.)
+  # Until then the residual risk stays acceptable — but NOT for the "devDependency-only" reason this note
+  # used to give, which was wrong. Every line reaches brace-expansion via minimatch, and they are rooted
+  # in production chains, not only dev ones: the 2.x line through `googleapis` and `@boxyhq/saml-jackson`
+  # (both apps/web `dependencies`, via gaxios>rimraf>glob and typeorm>glob) and `@ai-sdk/google-vertex`
+  # (a packages/ai `dependency`); the 1.x line through node-gyp under that same saml-jackson's
+  # typeorm>sqlite3. `pnpm audit` marks these findings `dev: false`. By contributor, the 1.x line is
+  # dominated by the eslint plugin graph (typescript-eslint, @vitest/eslint-plugin, eslint-config-next,
+  # eslint-plugin-storybook) and the 2.x line by vite-plugin-dts>@vue/language-core across six workspace
+  # packages, plus both prettier plugins. @redocly/cli is the only root on both lines, mostly on 2.x.
```
xernobyl: **[P3] `@redocly/cli is the only root on both lines` — not quite, and this one's on me.**

Two roots are on both:

| root | 1.1.16 | 2.1.2 |
|---|---|---|
| `@boxyhq/saml-jackson` | 5 (via `typeorm>sqlite3>node-gyp`) | 2 (via `typeorm>glob`) |
| `@redocly/cli` | 1 | 3 |

I asserted the "only root" framing last round without checking it against the graph I'd just built — it was in my own output, and I read past it. Apologies for the churn.

The sentence three lines up already implies it, too: *"the 1.x line through node-gyp under that same saml-jackson's typeorm>sqlite3"* — the "that same" is doing exactly this work.

Either drop the clause (the by-contributor sentence stands without it) or:

> `@redocly/cli` and `@boxyhq/saml-jackson` are the only roots on both lines.

Noted in ENG-2234 so it doesn't get carried forward, though that PR deletes the block outright.

xernobyl (dismissed):
Approving. All four are fixed, and re-verified at `55708a0` rather than taken on the commit message.

**The new advisory claim is correct, and it's a genuine catch I missed.** GHSA-rgw5-rvv9-x895 = CVE-2026-69152, high, "DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation", published 2026-08-03. Ranges are exactly as the note states — `<1.1.18`, `>=2.0.0 <2.1.4`, `>=4.0.0 <5.0.9` — so 5.0.8 really is vulnerable and dropping brace-expansion from the enumeration wholesale was the right call, not just pulling `@1/@2` as I'd suggested. Publish dates check out to the minute too: 5.0.9 at 10:00, 2.1.4 at 10:15, 1.1.18 at 10:17, all 2026-07-30, all clear of `minimumReleaseAge: 4320`.

Rest of the re-check:

- **[P2] enumeration contradiction** — fixed, and better than asked. The paragraph moved out from under "Removed once upstream caught up" into its own `=== Fixable, deferred to its own PR ===` header, so a note arguing "re-pinning closes this" no longer sits under "what no longer needs a pin".
- **[P2] rotted versions** — fixed. The `latest`/`rc` tags are gone from the prose and the rc-specific claims read "as inspected at 1.7.0-rc.4". I'd verified rc.4 myself last round: peers coupled at `^1.7.0-rc.4`, `validAudiences` gone, all three tables declared. The parenthetical explaining *why* it's version-agnostic is a nice touch — it'll stop the next person "helpfully" re-adding the numbers.
- **[P3] prose** — `eslint-plugin-storybook` singular is right, "both prettier plugins" 2.x-only is right, and the newly-asserted **six** workspace packages for `vite-plugin-dts>@vue/language-core` is correct (cache, database, jobs, js-core, survey-ui, workflows). One clause is wrong — inline below, and it's my fault.
- **[P3] follow-up** — ENG-2234 referenced, and the maintenance releases named. I've updated that ticket: it now covers both advisories and all three lines, since my original version repeated the "5.x is already patched" error and would have sent the implementer in to pin only `@1/@2`.
- **Optional nit** — taken, and sharpened past what I suggested. "this is 'not exploitable here', not 'compliant'" is the sentence that makes the note un-mis-readable.

Everything else I'd verified last round still holds on this head: resolved versions unchanged (1.1.16 / 2.1.2 / 5.0.8), all four production chains reproduce identically, and every line in the block is within 110 chars.

Not blocking on the inline nit — it's one clause, it's my wording, and ENG-2234's PR deletes the whole block anyway. Fix it in passing if you touch this again. Three checks (Docker Build, Sonar, E2E) were still running when I approved, so give those a glance before merging.

pnpm-workspace.yaml
mattinannt: Correct, and fixed in `8b44332`. My own root counts said the same thing and I read past it too — `@boxyhq/saml-jackson` is 5 paths on 1.1.16 and 2 on 2.1.2, sitting right next to `@redocly/cli`'s 1 and 3 in the table I posted two rounds ago. Intersecting the two root sets gives exactly those two, and your full-graph derivation agrees, so:

> Two roots sit on both lines: `@redocly/cli` (mostly 2.x) and `@boxyhq/saml-jackson` (mostly 1.x, via the node-gyp chain above).

I went with naming both rather than dropping the clause — `@redocly/cli` straddling is the fact that kept tripping this paragraph up, so it's worth stating, just accurately this time.

Fixing it despite your "not blocking" and despite ENG-2234 deleting the block, because a knowingly-false factual claim is the one thing this PR can't land — that's its entire thesis. Sorry for the extra approval round; that cost is worth paying here.

No apology needed on the churn, incidentally. Six rounds found six real defects in a 50-line comment, four of which predated this PR. That's the process working.

---

xernobyl (approved):
Re-approving — the push dismissed the previous approval, and `8b4433292` is a one-line fix to exactly the clause I flagged.

Verified the new comparative rather than trusting the path table from earlier in the thread (that shortcut is what produced the wrong clause in the first place). Counted distinct paths from each root myself:

| root | 1.1.16 | 2.1.2 | note says |
|---|---|---|---|
| `@boxyhq/saml-jackson` | 5 | 2 | mostly 1.x ✅ |
| `@redocly/cli` | 1 | 3 | mostly 2.x ✅ |

Both correct, and "via the node-gyp chain above" is the right cross-reference — that's the `typeorm>sqlite3>node-gyp>glob>minimatch` route described two sentences up. Lines are 100 and 68 chars.

Nothing else moved in the diff against the sha I approved before, so everything from that review stands. No open findings from me.

CI restarted from scratch on this head, so merge is gated on that rather than on review now.

PR #8659, opened 2026-07-28
feat(billing): charge trial upgrades immediately at full price [ENG-1968]

Dhruwang:
## What & why

On a Pro trial, the paid-only features (follow-ups, custom links) are gated on trial status, so they stay locked for the duration of the trial and upgrading out of a trial did not bill the card at conversion. This PR makes converting a trial to paid an **immediate, card-validated, full-price** action and gives trial users a one-click way to **pay the full plan price now and unlock immediately**. It also makes leaving a trial (**Return to Hobby**) an immediate switch rather than a scheduled one.

> **Background & security context is tracked internally in the Linear tickets below** (ENG-2056, and ENG-1968 for the related earlier fix). The abuse-vector details and reproduction are intentionally kept out of this public PR — reviewers who want the full "why" should read the tickets.

> **Review follow-up:** an earlier revision of this PR credited unused trial days against the first invoice via customer-balance transactions (~500 lines of custom apply/verify/reverse logic). Per review feedback, that credit is **removed entirely** — every conversion and upgrade is billed at full price through Stripe-native flows only, and the custom credit machinery is gone.

What changed:

- **Charge immediately on trial → paid, at full price.** Ending the trial and switching plans happens in a **single** Stripe update (`trial_end: "now"` + `always_invoice` + `error_if_incomplete`), so the card is invoiced exactly once for the target plan and a declined card blocks the upgrade instead of granting access on an unpaid invoice. (A prior two-step update double-invoiced — fixed here.) The billing cycle starts at conversion; the next full charge is one period later.
- **No proration, no credits, no custom money math.** Same-plan conversion ("Unlock all Pro features"), card-on-file "Upgrade now", and tier changes out of a trial (Pro → Scale) are all billed the identical way: the full target-plan price. The confirmation modal quotes Stripe's **invoice preview** (tax and metered usage included); if the preview fails (it can, on usage-based line items) the modal falls back to amount-less copy with the list price from the plan card — never a fabricated number.
- **Return to Hobby from a Pro trial is now immediate.** It opens a confirmation dialog; on confirm the trial ends now and the subscription drops straight to the free Hobby plan (single Stripe update: swap to Hobby items + `trial_end: "now"` + `proration_behavior: "none"`). Because the Pro trial was created with `missing_payment_method: "cancel"`, ending it on a no-card org would otherwise **cancel** the subscription and strand a dead `subscriptionId` (which then broke the next upgrade with a generic error until a page refresh) — the same update overrides the end-behavior to `create_invoice` so a no-card trial lands on an active $0 Hobby subscription. **Paid** plans keep the existing schedule-at-period-end downgrade (they've paid for the month), so scheduling now only applies where it makes sense.

### UI

- The no-card trial CTA is **"Unlock all Pro features"** (was "Continue with Pro after trial"). It opens a **single-action** confirm modal (**"Start Pro now?"** → **Pay $X now**) stating that trial-locked features (follow-ups, custom links) unlock immediately and the exact full-price charge. The earlier two-option modal (a secondary "Keep trial & add card" button) was **removed** — the choice was confusing and the pay-now action is the primary intent.
- The CTA relabel is split into its own i18n key so it no longer contradicts the Pro card's **"Current plan"** badge or falsely imply "after trial".
- **The billing-page trial banner CTA goes through the same pay-now modal.** "Continue with Pro after trial" (which opened an add-card-only checkout and kept the trial) is gone; the banner now reads "Some features like follow-ups and custom links stay locked during the trial. Pay now to unlock everything immediately." with an **"Unlock all Pro features"** button opening the same confirm modal. There is **no** add-a-card-and-keep-trialing path anywhere anymore; the trial-warning modal's CTA (a plain link to the billing page) is relabeled "Unlock all features".
- **The card-on-file "You're all set" banner variant is removed.** Features stay locked until paid, so "Your plan will continue automatically after the trial ends" contradicted the model. Card-backed trials (e.g. after a declined pay-now attempt) see the same locked-features banner + pay-now CTA as everyone else.

### Review follow-ups (earlier commits)

- Quote the **invoice total** (tax + metered usage) from Stripe's preview instead of the bare list price.
- Stop sending `trial_end: "now"` in the **non-trial** upgrade preview — the real update for an active subscription sends none, and setting it re-anchors the billing cycle, so the preview showed a full fresh period where the charge is a mid-cycle proration.
- Distinct error for **authentication-required** cards, pending state on the modal's actions, and a saving toast during the post-checkout card resync.
- **Removed the unused-trial credit and all supporting logic** (daily-rate computation, customer-balance transactions, landed-verification, reversal bookkeeping, idempotency keys, and their tests) in favor of full-price billing — net **−517 lines**.

## Linear ticket

- https://linear.app/formbricks/issue/ENG-2056/charge-trialpaid-conversions-immediately-prorated-early-unlock-close (this PR — internal context + reproduction)
- https://linear.app/formbricks/issue/ENG-1968/reverse-trial-can-be-laundered-into-an-active-paid-pro-subscription (related earlier fix, #8602)

## How this was tested

- **Unit (vitest):** `pnpm vitest run modules/ee/billing` — ✅ **104/104** (`organization-billing.test.ts` alone: **66**). Covers full-price trial conversion (single update, `trial_end: "now"` + `error_if_incomplete`, no discounts/credits attached), declined-charge rollback (trial intact, nothing to clean up), `previewImmediateUpgradeCharge` (full invoice total; `null` → amount-less copy on preview failure; no `trial_end` in non-trial previews), and the **immediate trial → Hobby** switch (Hobby items + `trial_end: "now"` + `proration_behavior: "none"` + `missing_payment_method: "create_invoice"`; stray-schedule release; no schedule/`cancel_at_period_end`; pending snapshot nulled).
- **Typecheck:** `tsc --noEmit` ✅ clean for `apps/web`.
- **Lint / format:** `eslint` ✅ on all changed files; `prettier` ✅ (pre-commit).
- **i18n:** `pnpm i18n` (Lingo) run — all 14 non-English locales regenerated and the lockfile rebuilt. Reworded keys: `confirm_trial_continue_body` / `confirm_trial_continue_body_fallback` (full price, no credit line). Translation-keys check ✅.
- **Stripe sandbox (real API, test mode):** single full-price invoice on trial → paid, declined card leaves the subscription trialing. Immediate trial → Hobby lands on an active $0 Hobby subscription (not cancelled) and a subsequent "Upgrade now" works without a refresh. *(Re-verified before the credit removal; the removed credit had no other effect on these flows.)*

## QA / Test Plan

**How to test**

- [ ] Cloud org on a **Pro trial with days remaining** → Settings → Billing → **Pro** card → **"Unlock all Pro features"** → a modal opens (**"Start Pro now?"**) with a single **Pay $X now** action quoting the **full plan price** (e.g. **Pay $89.00 now**, plus tax where applicable). There is **no** credit line and **no** "keep trial" button in this modal.
- [ ] **Pay now** with a valid card → charged the full amount **once**, subscription becomes **active**, follow-ups + custom links unlock immediately; invoice history shows a single charge; billing cycle starts today.
- [ ] **In a tax region (e.g. DE, 19% VAT):** the amount charged equals `list price + tax` and matches the modal quote.
- [ ] Confirm the **next** billing cycle is the same full price, one period after conversion.
- [ ] **No card on file** → Pay now routes to add-card checkout first, then charges on return; the confirm modal (with the amount) is shown **before** the redirect, and the CTA shows a pending state while the Checkout session is created.
- [ ] **Trial banner on the billing page** (no card) → shows the locked-features copy and an **"Unlock all Pro features"** button → opens the same **"Start Pro now?"** modal as the Pro card CTA (not an add-card checkout). There is no way left to add a card while keeping the trial.
- [ ] **Card already on file, still trialing** → the Pro card shows **"Upgrade now"** → a single-action confirm showing the full price → charges immediately.
- [ ] **Return to Hobby from a Pro trial** → **Return to Hobby plan** CTA → confirmation dialog (**"Switch to the Hobby plan now?"**) → on confirm the org is on **Hobby immediately** (not "scheduled"), no charge.
- [ ] **Regression for the end-behavior fix:** right after returning to Hobby (no card), click **Upgrade now** on the Pro card → it works on the **first** click (no "something went wrong" toast, no refresh needed).
- [ ] **Paid** Pro → Hobby (not a trial) → still **schedules** the downgrade for period end (pending-change banner), unchanged.
- [ ] **Ordinary (non-trial) paid upgrade, mid-cycle** (active Pro → Scale) → the confirm modal shows the **prorated** amount for the rest of the cycle, matching the invoice. *(Regression for removing `trial_end` from the non-trial preview.)*
- [ ] Edge — **card that fails at charge time** on Pay now → upgrade rejected, org stays on the trial, no leftover credit/coupon/balance on the Stripe customer.
- [ ] Edge — **3DS / SCA-mandated card** on Pay now → a specific "payment authentication" message (not a generic failure), org stays on the trial.
- [ ] Edge — **Pro trial → Scale** → billed the full Scale amount immediately.

**Preconditions / test data**

- Formbricks **Cloud** (Stripe billing); a Pro trial subscription; Stripe **test** cards that (a) succeed, (b) fail at charge time, (c) require 3DS. A customer in a VAT region for the tax check. Trial-restricted entitlements: follow-ups, custom links, custom redirect URL, bulk invite.

**Risks & regressions**

- Billing-critical path. Re-check normal (non-trial) paid upgrades/downgrades and the Scale upgrade behave as before — in particular the **mid-cycle proration amount** in the confirm modal (the shared preview helper no longer sends `trial_end` for non-trialing subscriptions), and that **paid** Pro → Hobby still schedules at period end.
- The immediate trial → Hobby path overrides the trial `end_behavior` to `create_invoice`; verify a no-card trial lands on an **active** Hobby subscription (not cancelled) and that a subsequent upgrade works without a refresh.
- **SCA/3DS cards**: the trial-conversion path uses `error_if_incomplete`, so a card needing off-session 3DS errors rather than completing on-session, surfaced as a distinct actionable message. Such cards still cannot convert a trial — worth a manual check in a 3DS test region.
- Confirm-modal wording changed for **all** upgrade paths (period-neutral, full price) and the trial-continue modal is single-action; non-English locales were regenerated by Lingo in this PR.

- none


apps/web/modules/ee/billing/components/pricing-table.tsx
```diff
@@ -912,23 +1043,23 @@ export const PricingTable = ({
       return t("workspace.settings.billing.downgrade_to_hobby");
     }
 
+    if (isTrialingWithPayment && plan !== "hobby") {
```
Dhruwang, line 1027: **The card-on-file trial path charges full price behind copy that describes something else.**

A card-backed trial on day 1 now sees **"Upgrade now"** on the *Pro card they are already trialing*, the button is enabled (`isCurrentSelection && !isTrialing` no longer disables it), and one confirm click ends the trial and bills the full $89. The modal it goes through is the existing `confirm_upgrade_body_with_charge`:

> "You're about to **upgrade to the Pro plan** ({period}). You'll be charged {chargeNow} now **for the rest of your current billing period**…"

Three things are wrong for this path:
1. They are not upgrading to Pro — they are already on Pro.
2. It is not "the rest of your current billing period" — `trial_end: "now"` resets the anchor and bills a **fresh full period**.
3. It never says the free trial ends immediately and the remaining 13 days are forfeited **with no credit**.

Meanwhile the no-card user in the identical position is offered $47.47. Same product, same day of the same trial, ~2x price difference decided by whether a card happened to be added earlier — and the more-committed customer is the one who pays more. That will land in support.

Two asks:
- Give the trial conversion its own copy (trial ends now, full period billed now, days forfeited) rather than reusing the mid-cycle-proration string. This is a payment consent screen; the current text misstates both the reason and the period.
- Confirm with whoever owns pricing that the card-on-file/no-card asymmetry is intended. If it is not, the simplest resolution is to apply the credit to any same-plan trial conversion and drop the `applyTrialCredit` flag entirely.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +660,146 @@ export type TUpgradePaymentConfirmation = {
...
+ * the catalog price is what actually gets billed, and it keeps both the credit and the confirmation
+ * amount reliable and identical.
+ */
+const getTrialConversionGrossCents = async (
+  targetPlan: Exclude<TStandardCloudPlan, "hobby">,
+  targetInterval: TCloudBillingInterval
+): Promise<{ cents: number; currency: string }> => {
+  const catalogItem = await getCatalogItemForPlan(targetPlan, targetInterval);
+  return {
+    cents: catalogItem.basePrice.unit_amount ?? 0,
```
Dhruwang: **The amount shown is a tax-exclusive, usage-exclusive base price — but it is presented as the exact final charge.**

The old source was `invoice preview.amount_due`, which is the *invoice total*: base price **+ tax + metered lines**. This replaces it with the catalog `unit_amount`, which is base price only. Two gaps follow, and the new copy has no hedge for either:

- **Tax.** Checkout sets `billing_address_collection: "required"` + `tax_id_collection` (organization-billing.ts:584, create-setup-checkout-session.ts:43) — i.e. Stripe Tax. A German customer at 19% VAT is charged `89.00 × 1.19 − 41.53 ≈ $64.38`, while the modal states **"you'll be charged $47.47 today"** and the button reads **"Pay $47.47 now"**. Note the pre-existing `confirm_upgrade_body_with_charge` deliberately hedges with *"with any applicable taxes calculated at payment"*; `confirm_trial_continue_body` dropped that hedge and instead asserts an exact figure.
- **Metered responses.** `getCatalogItemsForPlan` puts the metered `responses` price on the subscription, and `trial_end: "now"` + `always_invoice` closes the period now, so response overage accrued during the trial is invoiced on the same invoice. Not reflected in the number either.

An unhedged exact price in a purchase confirmation that is then exceeded is the kind of thing that generates chargebacks, and in the EU it is a price-transparency problem, not just a UX one.

Minimum fix: restore the "plus applicable taxes" hedge (and a usage caveat) in `confirm_trial_continue_body`. Better: keep `invoices.createPreview` as the primary source for the net amount and fall back to `catalog − credit` only when the preview throws, so the common case shows the real total.

Separately, `unit_amount ?? 0` silently degrades: if the base price is ever tiered, the modal renders "Pay $0.00 now" / "the Pro price of $0.00" and applies no credit. Prefer returning `null` here and letting the caller fall through to `confirm_trial_continue_body_fallback`.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +660,146 @@ export type TUpgradePaymentConfirmation = {
...
+    trialEndSeconds: subscription.trial_end,
+    nowSeconds: Math.floor(Date.now() / 1000),
+    interval: input.targetInterval,
+  });
+  if (creditCents <= 0) return null;
+
+  const currency = gross.currency || subscription.currency || "usd";
+  // Idempotency key scoped to this trial: a double-submit cannot double-credit. (A rare retry after
+  // a failed+reversed attempt falls back to charging full, which is the safe direction.)
+  const idempotencyKey = `trial-credit-${subscription.id}-${subscription.trial_end ?? "na"}`;
```
Dhruwang: **This idempotency key makes a retry either silently charge full price, or fail hard. Both are reachable.**

The key is `trial-credit-<sub>-<trial_end>` — stable across attempts *within a trial* — while the amount it guards is a function of `Date.now()`.

1. **Retry after a failed+reversed attempt silently drops the credit.** Card declines → `updateSubscriptionItemsImmediately` throws → reversal posts `+4153`. The user retries. `createBalanceTransaction` replays the *original* (already-reversed) transaction instead of executing, so no credit is on the balance — but `previewImmediateUpgradeCharge` recomputes the credit from scratch and still shows **"Pay $47.47 now"**. They get charged **$89**. The inline comment calls this "the safe direction", which is true for us and not for the customer: the price they consented to is not the price they pay. It is also the *most likely* retry path, since after the first attempt `hasPaymentMethod` is now true and the UI routes them to the full-price "Upgrade now" button anyway.

2. **Retry across a day boundary throws.** `remainingDays` steps down every 24h from `trial_end`, so a retry that crosses that boundary sends a *different* `amount` under the *same* key. Stripe rejects that ("Keys for idempotent requests can only be used with the same parameters…"), the error propagates out of `applyUnusedTrialCredit`, and the conversion is blocked with an opaque failure. Stripe keys live ~24h, so both windows overlap.

**Suggestion — drop the balance-credit mechanism entirely and use a one-time coupon.** Create an `amount_off` coupon for the credit and pass it via `discounts` on the *same* `subscriptions.update` that ends the trial. That makes it atomic with the charge, scoped to that one invoice, and removes: the pre-apply/reverse dance, `reverseTrialCreditSafely`, the stranded-balance failure mode, the "may be stranded" log line, both idempotency keys, and the currency-bucket issue flagged below. It is strictly less machinery for strictly better guarantees on a billing-critical path.

If the balance approach stays, at minimum make the key vary with the amount (`…-${creditCents}`) so case 2 becomes a fresh credit rather than an error, and make the preview aware that a reversal happened so case 1 cannot show a stale price.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +660,146 @@ export type TUpgradePaymentConfirmation = {
...
+  const existingDeletions = subscription.items.data.map((item) => ({ id: item.id, deleted: true as const }));
+  const preview = await stripeClient.invoices.createPreview({
+    customer: customerId,
+    subscription: subscription.id,
+    subscription_details: {
+      items: [...existingDeletions, ...targetItems],
+      proration_behavior: "always_invoice",
+      // Ending the trial resets the cycle to now and bills a full period; mirror that here so the
+      // previewed gross matches the real charge before the unused-trial credit is applied.
+      trial_end: "now",
```
Dhruwang: **`trial_end: "now"` leaks into the *non-trial* upgrade preview, where the real charge does not use it.**

`previewFullConversionChargeCents` is the fallback for **every non-trial** upgrade (previewImmediateUpgradeCharge:1351), but the real update for a non-trialing subscription (updateSubscriptionItemsImmediately:841) sends **no** `trial_end`. So preview and charge no longer mirror each other on the ordinary paid Pro→Scale path:

- `trial_end` resets the billing-cycle anchor, so the preview bills a **fresh full period**, whereas the actual update keeps the anchor and bills a **mid-cycle proration**. Halfway through a month those differ by close to a full month's price — the confirm modal would over-state the charge.
- If Stripe instead rejects `trial_end` on a non-trialing subscription, the preview throws, `upgradePreview` is `null`, and the modal silently degrades to the amount-less `confirm_upgrade_body` — the exact symptom this PR set out to fix, moved onto the more common path.

Fix is one line — gate it:
```ts
...(subscription.status === "trialing" ? { trial_end: "now" as const } : {}),
```

Worth noting `previewImmediateUpgradeCharge` has **no test coverage at all** (`invoicesCreatePreview` is only ever set up as a default mock; no test invokes the function). That is the function that produces the number the customer consents to before being charged, and it is now the only untested new branch. Two small tests would cover it: trialing → `catalog − credit` and no `createPreview` call; non-trialing → `createPreview` called **without** `trial_end`, `trialCreditApplied === 0`.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +660,146 @@ export type TUpgradePaymentConfirmation = {
...
+
+  const creditCents = computeUnusedTrialCreditCents({
+    fullChargeCents: gross.cents,
+    trialEndSeconds: subscription.trial_end,
+    nowSeconds: Math.floor(Date.now() / 1000),
+    interval: input.targetInterval,
+  });
+  if (creditCents <= 0) return null;
+
+  const currency = gross.currency || subscription.currency || "usd";
```
Dhruwang: **Currency precedence is backwards: customer balances are per-currency, so this can post the credit into a bucket the invoice never draws from.**

Stripe keeps a **separate** credit balance per currency, and the invoice will be denominated in `subscription.currency`. This prefers the *catalog price's* currency and only falls back to the subscription's. `create-setup-checkout-session.ts:31` explicitly threads `subscription.currency ?? "usd"` into Checkout, so non-USD subscriptions are a real shape here (multi-currency prices via `currency_options` share one `Price`, whose top-level `currency` is the default, not the customer's).

When they diverge: the credit lands in, say, the USD bucket, the EUR invoice is charged in **full**, and the customer is left holding a stranded credit that silently discounts some unrelated future invoice. The reversal will not save it either — it reverses the same wrong bucket.

```suggestion
  const currency = subscription.currency || gross.currency || "usd";
```

(And note `gross.currency` is what gets returned to the client as the preview currency at :1344, so the modal would also format the amount in the wrong currency symbol for those customers.)

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +660,146 @@ export type TUpgradePaymentConfirmation = {
...
+  interval?: TCloudBillingInterval;
+}): number => {
+  if (!Number.isFinite(fullChargeCents) || fullChargeCents <= 0) return 0;
+  if (!trialEndSeconds) return 0;
+  const billingPeriodDays = TRIAL_CREDIT_BILLING_PERIOD_DAYS[interval];
+  const remainingDays = Math.ceil((trialEndSeconds - nowSeconds) / 86_400);
+  if (remainingDays <= 0) return 0;
+  const cappedDays = Math.min(remainingDays, billingPeriodDays);
+  const creditCents = Math.round((fullChargeCents * cappedDays) / billingPeriodDays);
+  return Math.max(0, Math.min(creditCents, fullChargeCents));
```
Dhruwang: **The clamp permits `credit === fullCharge`, i.e. a $0 invoice — which means the card is never charged.**

The test at organization-billing.test.ts asserts exactly this:
```ts
// capped at the billing-period days -> full charge
expect(credit).toBe(8900);
```
A $0 invoice is settled without a payment attempt, so `payment_behavior: "error_if_incomplete"` has nothing to validate — the subscription flips to **active paid** on an unvalidated card. That is precisely the ENG-1968 property this PR exists to enforce, defeated arithmetically rather than by a Stripe behavior.

With `trial_period_days: 14` (organization-billing.ts:542) against a 30-day basis it cannot fire today, so this is hardening, not a live bug. But the invariant is one config change away from breaking silently: bump the trial to 30 days, or extend a `trial_end` by hand in the Stripe dashboard for a support case, and the guard is gone with no test failing. The credit is also computed from `subscription.trial_end`, which support can move.

Make the invariant explicit rather than incidental — clamp so a positive charge always remains:
```ts
return Math.max(0, Math.min(creditCents, fullChargeCents - 1));
```
and/or assert post-conversion that the resulting invoice was actually paid. A one-cent floor is enough to force a real authorization.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -682,6 +822,22 @@ const updateSubscriptionItemsImmediately = async (
...
+  // in a SINGLE update so the card is invoiced exactly once for the target plan. Doing it as two
+  // updates (end trial, then change items) double-invoices — ending the trial bills the OLD plan and
+  // the item change then bills the new one. error_if_incomplete charges synchronously and throws on
+  // a decline, so a bad card blocks the upgrade instead of granting access on an unpaid invoice.
+  if (subscription.status === "trialing") {
+    await stripeClient.subscriptions.update(subscription.id, {
+      items: [...existingDeletions, ...targetItems],
+      trial_end: "now",
+      proration_behavior: "always_invoice",
+      payment_behavior: "error_if_incomplete",
```
Dhruwang: **`error_if_incomplete` turns every 3DS card into a decline — on a flow that is fully on-session.**

The PR body accepts this as a trade-off of "validate the card now", but the trade-off isn't actually forced: the machinery to authenticate is already in this codebase and already wired to this exact call site. The non-trial branch 5 lines below uses `pending_if_incomplete`, pulls `invoice.confirmation_secret`, and the client completes 3DS via `settleUpgradeConfirmation` → `confirmUpgradeSca`. `trial_end` is one of the attributes pending updates support, so the trialing branch can use the same shape — and `pending_if_incomplete` gives you the *same* security property, since the plan is not granted until the invoice is paid.

As written, a customer on a card that mandates SCA (i.e. much of the EU/UK — the plurality of this product's paying base) gets: an error toast, no upgrade, and the credit reversed. They cannot convert their trial at all, and nothing tells them why. The PR's own test plan flags this as "worth a manual check in a 3DS test region" — I'd treat a guaranteed hard-fail for a whole card population as a blocker rather than a QA note.

If `pending_if_incomplete` genuinely can't carry `trial_end` in this API version, please confirm that in a comment here, and at minimum surface a specific error for `requires_action` so these users are told to use a different card instead of seeing a generic failure.

---

apps/web/modules/ee/billing/components/pricing-table.tsx
```diff
@@ -804,39 +857,113 @@ export const PricingTable = ({
...
+    if (hasPaymentMethod) {
+      void handlePlanAction(plan, interval); // card on file -> convert + credit immediately
+    } else {
+      void redirectToPlanCheckout(plan, interval); // add card -> applySetupCheckoutUpgrade converts
+    }
+  };
+
+  // Secondary action: keep the free trial. No card on file -> add one (billed at trial end); card
+  // already on file -> nothing changes, stay on the trial.
+  const handleTrialKeepTrial = () => {
```
Dhruwang: **"Keep free trial" redirects to a card-collection page.**

The modal only opens when `!hasPaymentMethod`, so this branch always runs `openTrialPaymentCheckout()` → Stripe Checkout asking for card details. A user who read "Keep your free trial: **pay nothing now**" and clicked the button labelled "Keep free trial" lands on a payment form. `Cancel` is the only control that actually does nothing — so the modal has three buttons where two sound like "don't charge me" and only the quietest one is.

The body copy does explain it ("Add a card and you'll move to {plan} automatically…"), but button labels are what people act on, and this one promises the opposite of what it does. Either label it for the action (`confirm_trial_continue_keep_trial` → "Keep trial & add card") or make it a true no-op and leave card collection to the existing standalone CTA.

The `if (!hasPaymentMethod)` guard is also dead as written — worth either dropping it or opening this modal for card-backed trials too (see the pricing/asymmetry point on `getCtaLabel`), which would make the guard meaningful.

---

apps/web/modules/ee/billing/components/pricing-table.tsx
```diff
@@ -804,39 +857,113 @@ export const PricingTable = ({
...
+    }
+    if (willChargeImmediately(plan, interval) || willChargeAfterAddingCard(plan, interval)) {
+      openConfirmation(plan, interval, "upgrade");
       return;
     }
     void handlePlanAction(plan, interval);
   };
 
+  // Primary action of the trial-continue modal: pay the prorated amount now and convert to paid.
+  const handleTrialPayNow = (plan: Exclude<TStandardPlan, "hobby">, interval: TCloudBillingInterval) => {
```
Dhruwang, line 930: **No pending state on either modal action, on a button that charges a card.**

`onConfirm` calls `closeUpgradeConfirmation()` and then this. The no-card branch goes straight to `redirectToPlanCheckout`, bypassing `handlePlanAction`, so `isPlanActionPending` is never set: the modal vanishes and the page looks idle for the round-trip that creates the Checkout session (`createTrialPaymentCheckoutAction`, ~1–2s) before navigation. A user who clicks "Pay $47.47 now" and sees nothing happen will click the plan card again.

The card-on-file branch does go through `handlePlanAction`, but the underlying `subscriptions.update` carries **no** idempotency key (only the credit does), so two concurrent conversions race on the same trialing subscription — the second one having read `status: "trialing"` before the first landed.

Cheapest fix: set `setIsPlanActionPending(...)` before dispatching either branch (or keep the modal open with `buttonLoading` until navigation), and pass an idempotency key on the conversion `subscriptions.update` keyed on subscription + target plan.

Same shape one level up in the `checkout_success` effect: `waitForBillingPaymentMethodAction` can take 5 × 1200ms plus Stripe latency before the hard reload, with no spinner or toast — the upgrade path right below it shows `toast.loading(...)` for exactly this wait. Worth mirroring so the page doesn't just sit there looking broken.

---

Dhruwang (commented):
## In-depth review — billing-critical

The core mechanic is right and the hard part is done well: collapsing the trial conversion into a **single** `subscriptions.update` (`trial_end: "now"` + `always_invoice`) is the correct fix for the double-invoice, the credit/reversal pairing is thought through, and `computeUnusedTrialCreditCents` is a clean pure function with genuinely good unit tests (interval basis, tapering, expiry, clamp). The `canCancelCurrentPaidPlanAtPeriodEnd(isTrialing)` correction and clearing the pending downgrade unconditionally are both real bug fixes with tests behind them.

What I'd want resolved before this bills a real card. Nine inline comments; the ones I'd treat as blocking:

**1. Amount shown ≠ amount charged (three independent ways).** All on the number the customer consents to:
- `getTrialConversionGrossCents` drops **tax and metered usage** (the old `preview.amount_due` included both), yet `confirm_trial_continue_body` removed the "plus applicable taxes" hedge the existing upgrade copy has and asserts an exact figure. At 19% VAT the modal says $47.47 and the card sees ~$64.
- **A retry silently charges full price.** The credit's idempotency key replays an already-reversed transaction, so no credit lands — while the preview still recomputes and shows $47.47. Charge: $89. A retry across a day boundary instead throws a Stripe idempotency error and blocks the conversion entirely.
- **`trial_end: "now"` leaked into the non-trial preview**, which the real non-trial update does not send. The ordinary Pro→Scale confirm now previews a full period instead of a mid-cycle proration (or throws and shows no amount at all).

**2. `error_if_incomplete` hard-fails every 3DS card**, on a flow that is entirely on-session and in a codebase that already has `pending_if_incomplete` + `confirmation_secret` + `confirmUpgradeSca` wired to the adjacent branch. `trial_end` is pending-update-supported. Guaranteed decline for much of the EU/UK card base reads as a blocker to me, not a QA note.

**3. Credit currency precedence is backwards.** Customer balances are per-currency; preferring the catalog price's currency over `subscription.currency` can post the credit into a bucket the invoice never draws from — full charge *and* a stranded credit.

Then: the clamp permits `credit === fullCharge` → a $0 invoice → no card authorization at all, which is arithmetically the ENG-1968 property this PR enforces (latent at 14 trial days, one config change from live); "Keep free trial" opens a card-collection page; and the pay-now buttons have no pending state.

**Design suggestion worth considering seriously:** replace the customer-balance credit with a one-time `amount_off` coupon passed in the *same* `subscriptions.update`. It is atomic with the charge, scoped to that single invoice, and deletes the entire apply→reverse→re-reverse surface: `applyUnusedTrialCredit`'s pre-apply, `reverseTrialCreditSafely`, both idempotency keys, the "balance may be stranded" log, the currency-bucket hazard, and the retry-charges-full bug. Strictly less machinery for strictly stronger guarantees.

**Testing.** 66 unit tests and the Stripe-sandbox verification are the right instincts, and the reversal-on-failure and no-credit-for-Pro→Scale tests are the valuable ones. Two gaps: `previewImmediateUpgradeCharge` has **zero** coverage despite being the function that produces the price the user agrees to (its `createPreview` mock is only ever a default), and nothing covers a retry after a reversal — which is where the silent full-price charge lives. Both are pure-logic, cheap to add. No E2E needed here; the flow can't be exercised without Stripe, and per AGENTS.md the modal itself is `.tsx` so unit tests are correctly out of scope.

**Smaller notes** (non-blocking, no inline comment):
- `whitespace-pre-line` on `ConfirmationModal`'s body changes rendering for all 16 consumers. Only `save_changes_confirmation_body` contains a `\n`, and it reads better with the break — worth knowing it changed. No story added for the ReactNode body.
- `unit_amount ?? 0` degrades to "Pay $0.00 now" if a base price is ever tiered; prefer `null` + the fallback copy.
- `getCtaKey` / `getCtaLabel` duplicate the same branch ladder and three of the new conditions had to be edited in both. Pre-existing, but this PR widens the divergence risk on a screen where a wrong label means a wrong charge.
- `TRIAL_CREDIT_BILLING_PERIOD_DAYS` (30/365) vs. actual period length: acknowledged in the comment, rounds in the customer's favour, fine.
- `applyTrialCredit` flowing from the client into the preview is harmless (read-only, and both mutating call sites hardcode the flag server-side) — but it does mean the preview can be asked for a credited quote that the charge path will never honor, which is the same class of mismatch as #1. A server-derived flag would close it.
- Authorization (`owner`/`manager`/`billing`) and audit logging on the new action look right; locale files are properly Lingo-generated.

Nothing here is a rewrite — items 1 and 3 are a few lines each, and the coupon suggestion would shrink the diff.

---

apps/web/modules/ee/billing/lib/organization-billing.ts
```diff
@@ -660,6 +655,218 @@ export type TUpgradePaymentConfirmation = {
...
+ * caller MUST invoke on a failed conversion, so a stranded credit never silently discounts a future
+ * invoice. Returns null when no credit applies.
+ *
+ * Throws OperationNotAllowedError("trial_credit_unavailable") if the credit doesn't land — e.g. Stripe
+ * replays an idempotency key from an already-reversed attempt, returning the old transaction without
+ * re-executing it. Charging anyway would bill full price behind a modal quoting the discount, so this
+ * fails loudly instead. (The key includes the amount, so a later-day retry with a tapered credit gets
+ * a fresh key.)
+ */
+const applyUnusedTrialCredit = async (
```
mattinannt: In this PR I'm mostly concerned about the custom logic for the credit handling.

The `organization-billing.ts‎` has nearly 500 new lines that are necessary for this change, but it seems to be going against stripe default ways they have API options for. Are there other options do solve this that don't require this custom logic and let Stripe do the heavy lifting?

mattinannt (dismissed):

jobenjada (dismissed):
UX is great ✅

mattinannt:
@coderabbitai please review

jobenjada (approved):

PR #8663, opened 2026-07-28
chore(deps): drop 17 obsolete pnpm overrides

mattinannt:
> [!NOTE]
> **Created by an AI agent under human supervision.** This PR was authored by Claude Code (Anthropic's coding agent) in a Claude Code on-the-web session, from a task scoped and supervised by @mattinannt.

## What & why

`pnpm-workspace.yaml` had accumulated 39 `overrides` entries, each added at some point to close a Dependabot / `pnpm audit` alert on a transitive dependency. Nothing ever re-checked them, so a pin survives long after the upstream chain adopts a patched version — and a stale pin is not free: it silently *downgrades* packages (`fast-xml-parser` is pinned to 5.7.0 while the tree already offers 5.7.3), keeps major-version coercions alive, and buries which upstream is actually still holding us back.

Every entry was re-verified by dropping it, re-resolving the workspace (`pnpm install --lockfile-only`) and re-running `pnpm audit`. 17 of the 39 were no longer doing anything:

**Obsolete — upstream now resolves the patched version on its own (14 removed):**
`ajv@6`, `@babel/core`, `body-parser`, `brace-expansion@1`, `brace-expansion@2`, `brace-expansion@5`, `dompurify`, `engine.io@6`, `fast-uri@3`, `hono`, `linkify-it`, `protobufjs@7`, `undici@7`, `uuid@11`.

**Retired by bumping the dependency that held the vulnerable version back (3 removed):**

| override dropped | dependency bumped | why it works |
| --- | --- | --- |
| `js-cookie: 3.0.7` | `react-use` 17.6.0 → 17.6.1 (apps/web) | 17.6.1 moved from `js-cookie@^2` to `^3.0.0`, so the v2→v3 major coercion is gone |
| `shell-quote: 1.9.0` | `concurrently` 9.2.1 → 9.2.4 (packages/surveys, dev) | 9.2.4 pins `shell-quote@1.9.0` itself |
| `form-data@4: 4.0.6` | `@redocly/cli` 1.34.3 → 1.34.17 (root, dev) | 1.34.17 pins `form-data@4.0.6`; 1.34.3 pinned the vulnerable 4.0.0 |

The `@redocly/cli` bump also moves the `REDOCLY_NPX_FALLBACK` pin in `docs/api-v3-reference/scripts/bundle.mjs`, which that file's comment asks to keep in sync with the root devDependency.

The remaining 22 entries are still load-bearing — dropping any of them puts a version the advisory covers back in the tree. Each now carries a `load-bearing:` note naming the consumer responsible (e.g. `axios` — `@boxyhq/saml-jackson` pins 1.13.5; `postcss` — `next` pins 8.4.31), so the next pass can check one upstream instead of re-deriving the graph. `lodash` and `minimatch@10` had no rationale recorded at all and now carry their advisory IDs.

Net effect on the dependency graph: **none**. The resolved version of every previously overridden package is byte-identical before and after; the only lockfile movement is the three bumped packages and their pinned subtrees.

Not changed, and why: `@sentry/nextjs` 10.43.0 → 10.68.0 would retire `uuid@9` (the newer `@sentry/webpack-plugin` no longer depends on uuid), but `uuid@8` would still be needed via `@azure/msal-node`, so it buys no override removal on its own — left out to keep this diff to dependency hygiene. Separately, `react-use` is not imported anywhere in the repo; dropping it entirely (rather than bumping it) looks possible as a follow-up.

## Linear ticket

No ticket — dependency hygiene follow-up to the audit-pin batches (#8612, #8623, #8638, #8641). The individual advisories referenced in the file keep their existing ENG-* IDs.

## How this was tested

- `pnpm audit` before vs. after: identical, exactly 2 advisories in both runs — `GHSA-p2fr-6hmx-4528` (`@better-auth/oauth-provider`, patched only in a `1.7.0-beta.4` pre-release) and `GHSA-mh99-v99m-4gvg` (the documented `brace-expansion` `<=5.0.7` range that also matches the terminal 1.1.16 / 2.1.2 releases). No new advisory, no resolved advisory. ✅
- Resolved-version diff of all 34 affected package names, parsed out of the lockfile before and after: every one identical (`@grpc/grpc-js` 1.14.4, `dompurify` 3.4.12, `js-cookie` 3.0.7, `form-data` 4.0.6, `shell-quote` 1.9.0, …). ✅
- `pnpm install --frozen-lockfile` — clean, "Already up to date". ✅
- `pnpm lint` ✅ (includes `pnpm api:v3:lint`, which is what exercises the `@redocly/cli` bump)
- `pnpm api:v3:check` ✅ — committed `openapi.yml` bundle still in sync under redocly 1.34.17, so the generated spec is unchanged.
- `pnpm test` ✅ (all workspace vitest suites)
- `pnpm peers check`: same 3 pre-existing unmet peers (`@testing-library/dom`, `tailwindcss`, `mongodb` via typeorm), none related to this change.

No UI change, so no screenshots.

## QA / Test Plan

**How to test**

- [ ] Sign in to the app and load the surveys list → app boots and renders normally (this PR changes no runtime dependency version; it only removes redundant version pins).
- [ ] Open a survey link (public survey page) and submit a response → survey renders and the response is saved, confirming the `@formbricks/surveys` bundle still builds and serves correctly.
- [ ] Upload an image on a survey question (sharp/storage path) → upload succeeds and the thumbnail renders.
- [ ] Log in via a SAML SSO connection if one is configured → login completes (this is the chain behind the `@boxyhq/saml-jackson` pins that were kept).
- [ ] Open the API v3 reference docs → the spec renders and matches the committed `docs/api-v3-reference/openapi.yml`.

**Preconditions / test data**

- Any environment (Cloud or self-host); no feature flags, plan or entitlement requirements.
- A SAML connection is only needed for the optional SSO step; skip it if none is set up.

**Risks & regressions**

- The blast radius is dependency resolution, not app code. The three bumped packages are: `react-use` (not imported anywhere in the repo), `concurrently` (only used by `packages/surveys` dev/watch scripts) and `@redocly/cli` (dev tooling for the v3 OpenAPI spec — covered by `pnpm api:v3:lint` / `--check` in CI).
- Highest-signal re-check is CI itself: install with `--frozen-lockfile`, build, unit tests and E2E all run against the regenerated lockfile.
- If a future Dependabot alert reappears for one of the 17 dropped names, the fix is to re-add that single pin — the `load-bearing:` notes explain how each entry was judged.

- none

## Model & settings

- Model: Anthropic Claude, Opus-class model (largest tier), knowledge cutoff May 2026
- Reasoning effort: high · fast mode: off · temperature/sampling: harness defaults
- Harness: Claude Code CLI via the Claude Agent SDK, remote web session (ephemeral container, fresh clone)
- Tools used: shell, file read/write, GitHub API; no auto-merge, no force-push
- Human oversight: task, scope and final review by @mattinannt

---

https://claude.ai/code/session_01VdyBSihuBwULyAbGWnDfKQ

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -142,7 +142,7 @@
     "react-hot-toast": "2.6.0",
     "react-i18next": "16.5.8",
     "react-turnstile": "1.1.5",
-    "react-use": "17.6.0",
+    "react-use": "17.6.1",
     "recharts": "2.15.3",
     "redis": "4.7.1",
     "sanitize-html": "2.17.4",
--- a/docs/api-v3-reference/scripts/bundle.mjs
+++ b/docs/api-v3-reference/scripts/bundle.mjs
@@ -20,7 +20,7 @@ import { fileURLToPath } from "node:url";
 
 // Fallback pin for environments without the workspace install. Keep in sync with the
 // @redocly/cli entry in the root package.json devDependencies (the preferred, lockfile-pinned path).
-const REDOCLY_NPX_FALLBACK = "@redocly/cli@1.34.3";
+const REDOCLY_NPX_FALLBACK = "@redocly/cli@1.34.17";
 const here = dirname(fileURLToPath(import.meta.url));
 const srcRoot = resolve(here, "../src/openapi.yml");
 const artifact = resolve(here, "../openapi.yml");
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@azure/playwright": "1.1.5",
     "@formbricks/eslint-config": "workspace:*",
     "@playwright/test": "1.58.2",
-    "@redocly/cli": "1.34.3",
+    "@redocly/cli": "1.34.17",
     "@trivago/prettier-plugin-sort-imports": "6.0.2",
     "dotenv": "17.3.1",
     "eslint": "9.39.5",
--- a/packages/surveys/package.json
+++ b/packages/surveys/package.json
@@ -64,7 +64,7 @@
     "@types/node": "25.4.0",
     "@types/react": "19.2.14",
     "autoprefixer": "10.4.27",
-    "concurrently": "9.2.1",
+    "concurrently": "9.2.4",
     "cross-env": "10.1.0",
     "dotenv-cli": "11.0.0",
     "fake-indexeddb": "6.2.5",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,41 +7,24 @@ settings:
 overrides:
   '@grpc/grpc-js': 1.14.4
   '@hono/node-server': 2.0.10
-  hono: 4.12.27
-  protobufjs@7: 7.6.5
   protobufjs@8: 8.6.6
   '@tootallnate/once': 2.0.1
   '@xmldom/xmldom': 0.9.10
   axios: 1.18.0
   lodash: 4.18.1
   node-forge: 1.4.0
-  ajv@6: 6.14.0
   minimatch@10: 10.2.5
   postcss: 8.5.18
   fast-xml-parser: 5.7.0
-  body-parser: 2.3.0
   typeorm: 0.3.31
-  js-cookie: 3.0.7
   uuid@8: 11.1.1
   uuid@9: 11.1.1
-  uuid@11: 11.1.1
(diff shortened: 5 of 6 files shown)


Dhruwang (approved):

PR #8653, opened 2026-07-27
refactor: route authorization checks through central interface

BhagyaAmarasinghe:
## What does this PR do?

Fixes [ENG-1714](https://linear.app/formbricks/issue/ENG-1714/route-existing-authorization-checks-through-the-central) on top of the merged ENG-1712 central authorization interface.

This PR routes the first shared authorization paths through Formbricks-owned `can()` / `assertCan()` decisions while keeping the legacy evaluator authoritative:

- translates canonical action-client organization, workspace-team, and team-admin signatures into semantic actions;
- preserves a legacy fallback for ambiguous or unusual compatibility signatures, including central-denied OR cases where legacy workspace-team grants are broader;
- splits the database-backed workspace permission ladder from its public compatibility wrapper to prevent evaluator recursion;
- migrates organization helpers, shared analysis and V3 session workspace access, organization/workspace settings, team actions, and user-managed API-key settings;
- documents the remaining API-key-principal work for ENG-1731 and long-tail migration for ENG-1737.

Authorization behavior is unchanged. There are no AuthZed network calls, schema changes, relationship writes, database migrations, environment changes, public endpoints, or new permissions. API-key principals remain out of scope.

## How should this be tested?

Precondition: run with AuthZed disabled and credentials absent; the central interface continues to select the legacy evaluator.

Automated validation completed:

- `pnpm --filter=@formbricks/web exec vitest run lib/authorization lib/utils/action-client/action-client-middleware.test.ts lib/workspace/auth.test.ts lib/organization/auth.test.ts lib/auth.test.ts modules/ee/analysis/lib/access.test.ts app/api/v3/lib/auth.test.ts 'app/(app)/workspaces/[workspaceId]/settings/organization/general/actions.test.ts' modules/ee/teams/team-list/actions.test.ts modules/organization/settings/api-keys/actions.test.ts modules/workspaces/settings/actions.test.ts` — 16 files, 145 tests passed, including the billing-role + WorkspaceTeam parity regression.
- `pnpm --filter @formbricks/web typecheck` — passed.
- `pnpm authzed:validate` with checksum-verified `zed` v1.1.1 — 27 relationships, 131 assertions, and 2 expected relations passed.
- `pnpm build --filter=@formbricks/web... --force` — 12/12 tasks passed.
- ESLint on every changed TypeScript file — passed with no findings.
- SonarCloud follow-up passed after extracting central access-item evaluation and adding an explicit role comparator.

Manual QA matrix:

[...]

PR #8665, opened 2026-07-28
feat: project organization memberships to SpiceDB

BhagyaAmarasinghe:
## What does this PR do?

Implements [ENG-1715](https://linear.app/formbricks/issue/ENG-1715/project-organization-memberships-and-roles-into-spicedb) on top of the merged central authorization interface from #8653.

PostgreSQL remains the source of truth and the legacy evaluator remains authoritative. This PR adds post-commit, best-effort projection of the current organization membership model into SpiceDB:

- Extends the server-only Formbricks AuthZed facade with SDK-independent, idempotent relationship `touch`/`delete` batches and safely narrowed bulk deletion.
- Reconciles each `Membership` to exactly one `organization` role relation (`owner`, `manager`, `member`, or `billing`) and deletes stale roles atomically.
- Preserves current behavior by projecting accepted and pending membership rows identically; invites alone are not projected.
- Re-reads PostgreSQL after each write and retries reconciliation for up to three passes when a role changes concurrently.
- Projects only after source commits across shared membership creation, SSO transactions, role updates, explicit deletion, API v2 nested writes, organization deletion, and both user-deletion paths.
- Contains AuthZed failures at the projection boundary so an outage never misreports a committed PostgreSQL mutation as failed. Logs and results are sanitized and contain no identifiers, tokens, SDK objects, requests, responses, or raw errors.
- Defensively isolates Better Auth post-delete cleanup from an unexpected AuthZed cleanup rejection, ensuring Brevo cleanup, audit emission, and analytics still execute after the user deletion has committed.
- Extends the isolated Docker smoke test with the application relationship client, owner-to-billing transition, idempotent repeat/delete, bounded outage failure, recovery, persistence, test-only guard, and secret scanning.
- Documents the source-of-truth, accepted-membership parity, mutation inventory, and ENG-1718 repair dependency.

This deliberately does **not** add schema changes, backfill/repair, team/workspace/API-key projection, shadow evaluation, enforcement, public endpoints, database migrations, or ZedToken persistence. ENG-1718 must repair/backfill source relationships before shadow mode or enforcement can be enabled.

## How should this be tested?

- `pnpm --filter @formbricks/web exec vitest run lib/authzed lib/membership/service.test.ts modules/ee/role-management/lib/membership.test.ts modules/organization/settings/teams/lib/membership.test.ts modules/api/v2/organizations/'[organizationId]'/users/lib/tests/users.test.ts modules/ee/sso/lib/sso-provisioning.test.ts modules/ee/sso/lib/tests/sso-handlers.test.ts lib/organization/service.test.ts lib/user/service.test.ts modules/account/lib/better-auth-account-deletion.test.ts`
  - 18 files and 230 tests passed.
- `pnpm --filter @formbricks/web exec vitest run modules/account/lib/better-auth-account-deletion.test.ts`
  - 13 tests passed, including unexpected AuthZed cleanup rejection isolation.
- `pnpm --filter @formbricks/web typecheck`
  - Passed.
- `pnpm authzed:validate`
  - 27 relationships, 131 assertions, and 2 expected relations passed.
- `pnpm build --filter=@formbricks/web... --force`
  - Passed with AuthZed disabled and credentials absent.
- `pnpm --filter @formbricks/web build`
  - Passed with AuthZed enabled, syntactically valid credentials, and an unreachable endpoint; no startup RPC was attempted.
- `pnpm authzed:smoke`
  - Passed schema lifecycle, relationship projection, role transition, idempotent deletion, authentication failure, bounded outage handling, recovery, migrations, persistence, and secret scanning.
- Focused ESLint, Prettier, `bash -n`, and `git diff --check`
  - Passed.

### Required

- [x] Ran `pnpm build`
- [ ] Checked for warnings, there are none — local commands emitted only the expected engine warning because the workstation uses unsupported Node 25; CI uses the repository-supported Node 22 runtime
- [x] Created the branch from the latest `epic/authzed` target (`eb0987ac6`); merging `main` is not applicable to this stacked epic branch
- [x] CLA is already signed

### Appreciated

- [x] No UI change; screenshots or a recording are not applicable
- [x] Updated the AuthZed architecture documentation in `authzed/README.md`

### DIFF
--- a/apps/web/lib/authzed/__mocks__/client-dependencies.ts
+++ b/apps/web/lib/authzed/__mocks__/client-dependencies.ts
@@ -3,9 +3,11 @@ import { vi } from "vitest";
 export const sdkMocks = {
   close: vi.fn(),
   deadlineInterceptor: vi.fn((timeoutMs: number) => ({ timeoutMs })),
+  deleteRelationships: vi.fn(),
   diffSchema: vi.fn(),
   newClient: vi.fn(),
   readSchema: vi.fn(),
+  writeRelationships: vi.fn(),
   writeSchema: vi.fn(),
 };
 
@@ -33,6 +35,10 @@ vi.mock("@authzed/authzed-node", () => ({
       SECURE: 0,
     },
     NewClient: sdkMocks.newClient,
+    RelationshipUpdate_Operation: {
+      DELETE: 3,
+      TOUCH: 2,
+    },
   },
 }));
 
--- a/apps/web/lib/authzed/client.test.ts
+++ b/apps/web/lib/authzed/client.test.ts
@@ -9,9 +9,11 @@ describe("AuthZed client facade", () => {
     closeAuthzedClient();
     sdkMocks.close.mockReset();
     sdkMocks.deadlineInterceptor.mockClear();
+    sdkMocks.deleteRelationships.mockReset();
     sdkMocks.diffSchema.mockReset();
     sdkMocks.newClient.mockReset();
     sdkMocks.readSchema.mockReset();
+    sdkMocks.writeRelationships.mockReset();
     sdkMocks.writeSchema.mockReset();
     configMocks.isAuthzedEnabled.mockReset();
     retryMocks.execute.mockClear();
@@ -23,8 +25,10 @@ describe("AuthZed client facade", () => {
     sdkMocks.newClient.mockReturnValue({
       close: sdkMocks.close,
       promises: {
+        deleteRelationships: sdkMocks.deleteRelationships,
         diffSchema: sdkMocks.diffSchema,
         readSchema: sdkMocks.readSchema,
+        writeRelationships: sdkMocks.writeRelationships,
         writeSchema: sdkMocks.writeSchema,
       },
     });
@@ -95,9 +99,11 @@ describe("AuthZed client facade", () => {
     expect(sdkMocks.newClient).toHaveBeenCalledTimes(1);
     expect(Object.keys(first).sort()).toEqual([
       "consistency",
+      "deleteRelationships",
       "diffSchema",
       "readSchema",
       "systemKey",
+      "writeRelationships",
       "writeSchema",
     ]);
     expect(first).not.toHaveProperty("token");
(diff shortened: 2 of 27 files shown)


xernobyl (approved):
LGTM 🚀 Really solid — this is the part of the migration I was most wary about (drift on the write path) and it's handled carefully.

Things I checked specifically and liked:

- **Bulk-delete blast radius is contained at two levels** — the `TAuthzedRelationshipFilter` union makes a filter with neither `resourceId` nor `subject` unconstructible, and `validateRelationshipFilter` re-checks at runtime with non-empty guards so `""` can't sneak through. No way to mass-delete a resource type.
- **Delete scoping is right**: org-wide delete only on org deletion, the broad user-wide delete only in the two user-deletion paths (both post-delete), and membership deletion goes through the narrow per-(org,user) reconcile.
- **Atomic role reconciliation** — all four relations in one batch, so there's never a window with zero or two roles. The read → write → re-read → retry loop self-heals concurrent role changes and raises `authzed_projection_unstable` instead of leaving a silent stale write.
- **Post-commit discipline enforced by the type** — the explicit `{ projection: "deferred", transaction }` option is much harder to forget than an optional `tx`, and the projection sits outside the try, only after a successful write.
- **Coverage is complete**: I grepped every `prisma.membership.{create,update,delete,…}` in the app and the only hits are the three helpers that already project (plus one Playwright fixture) — no bypass.
- **`accepted` parity** matches the legacy evaluator (`membership !== null`, ungated on `accepted`), which is exactly right for shadow parity.
- Logs are clean (no ids/tokens/raw errors), and it's inert with AuthZed disabled.

A few non-blocking notes for the follow-ups:

1. **Failed *downgrades* are the one to keep an eye on** — if owner→member projection fails during an outage, SpiceDB keeps `owner`. Harmless while we're inert/shadow, but it means ENG-1718 repair should be a hard gate before we act on shadow results, not just before enforcement.
2. Sanitized logs mean a failed projection can't be repaired surgically — only via full backfill. `organizationId` probably isn't PII; logging it (or a correlation id) in ENG-1719 would let repair be targeted.
3. Projection is awaited in the request path — bounded and outage-tested, fine for v1; an outbox is the sturdier long-term shape if latency ever shows up.
4. For ENG-1716/1717: `deleteOrganizationRelationships` only clears edges where the org is the *resource*. Once workspace/team/api-key projection lands, edges where the org is the *subject* (`workspace:X#organization@organization:…`, etc.) will orphan on org deletion.

CodeRabbit's three nits are all trivial; the projection-isolation failure-path test is the only one with real value if you feel like grabbing it.

PR #8666, opened 2026-07-28
fix(api-v3): stop the taxonomy routes leaking raw Hub error text, and map upstream status

xernobyl:
> [!NOTE]
> Was stacked on #8650, which has now merged. Rebased onto `main` — the diff is these 7 commits and 11 files only. Still a draft pending a look-over.

## What & why

`app/api/v3/unify-feedback/taxonomy/lib/operations.ts` was putting raw upstream Hub error text into its own HTTP responses in nine places. `HubError.message` is the SDK error string, and for the Hub SDK that is the *whole* upstream problem body. Captured from the local Hub:

```
GET /api/v3/.../taxonomy/runs/<gone>  →  502
detail: "404 {\"type\":\"https://hub.formbricks.com/problems/not-found\",\"title\":\"Not Found\",
  \"status\":404,\"detail\":\"taxonomy run not found\",
  \"instance\":\"/v1/taxonomy/runs/00000000-0000-0000-0000-000000000000\",\"code\":\"not_found\",
  \"request_id\":\"019fa935-82ab-7cb6-9a34-80e364ddb9d3\",\"details\":{...}}"
```

So: the Hub's internal endpoint path and its own `request_id`, in a browser-visible response. `NO_CONFIG_ERROR` rode the same path, putting the literal `HUB_API_KEY is not set; Hub integration is disabled.` on the wire. That contradicts the rule the v3 layer documents for itself (`response.ts`'s `problemNotFound` JSDoc, and the OpenAPI Security section on not leaking existence).

Two things turned up alongside it:

- **The file logged nothing at all** — the diagnostic went to the client instead of to us.
- **Every Hub failure collapsed to 502**, discarding `error.status`. That hid a live bug: a run reaped by the orphan cleaner returns Hub 404 → surfaced as 502 → `useTaxonomyRun` never treats it as terminal, so the UI showed "generating…" indefinitely while polling every five seconds.

Found during the #8650 security review, checking whether that PR's relay rules were repeated correctly elsewhere.

### What changed

**Status mapping** for the set the Hub can actually emit here (`0/400/401/404/409/500/503`):

| Hub | → | detail |
| --- | --- | --- |
| 400 | 400 | fixed string + the Hub's `invalid_params` relayed (bounded) |
| 409 | 409 | Hub's detail, relayed (bounded 512) — retryable tenant-purge conflict |
| 404 | 404 | static, `resource_type` and **no** `resource_id` — the terminal signal the client needs |
| 503 | 503 | taxonomy's own wording (below) |
| 0 / 401 / 500 / other | 502 | fixed `"The feedback service is unavailable."` |

A **503 deliberately does not reuse the feedback-records "embeddings are not configured" copy.** The Hub returns 503 for taxonomy when embeddings **or** the compute service is unconfigured, and the embeddings case can't be reached through the UI at all — unset embeddings make `/fields` report `unavailable`, which disables the Generate button first. Confirmed live: this deployment has `EMBEDDING_MODEL` set and no starter, and its taxonomy 503 is `"Taxonomy service is not available."` — precisely the case the embeddings wording would have described wrongly. `hubErrorToProblemResponse`'s 503 detail is therefore a caller-supplied option whose default names no subsystem — that part landed on the base branch, where the same problem was raised in review for `feedbackRecords`, so this PR just uses it.

**The three `200 + unavailable: true` responses keep their shape** — the Generate gate depends on it — but carry a fixed message instead of `result.error.message`.

**Logging, where there was none.** `warn` for the statuses that really describe the caller's own request, `error` with the serialised `err` key for everything else. Context is ids and statuses only — never request bodies, node labels, or record content. A Hub **401** is in the error bucket on purpose: that's our `HUB_API_KEY` failing, it surfaces to the caller as an opaque 502, and logging it quietly would leave that outage explainable from nowhere.

**Client:** a `V3ApiError` with status 404 now ends the run poll. Everything else, 5xx included, keeps the existing self-recovery behaviour.

**Spec:** `conflict` and `service_unavailable` were emitted by response helpers but missing from the `Problem.code` enum. The spec CI job only fires on `docs/api-v3-reference/**`, so nothing caught the drift.

Out of scope: **part 2 of the ticket** (consolidating the duplicated `handleUnexpectedError` blocks). ENG-2048 is updated to say part 1 only.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2048

## How this was tested

**Automated** — `892 passed (61 files)` across `app/api/v3`, `modules/ee/unify-feedback`, `modules/mcp`, `modules/hub`. eslint clean on every changed file; `tsc --noEmit` clean for them. `api:v3:lint` valid and `api:v3:check` in sync.

Tests added:
- A status table over the whole reachable set, asserting status, `code` **and `detail`**. Every prior fixture used `status: 500`, so the old status-only assertions would have passed unchanged and given zero regression signal.
- A **disclosure guard**: `HUB_API_KEY`, an internal host, a fake upstream `request_id` and a Go source path are planted in the Hub error, and asserted absent from every response of all eight operations.
- A **503 regression guard** — a taxonomy 503 must not carry the embeddings copy — so nobody re-points this at the shared branch later.
- Logging assertions, including the 401-is-loud case.
- First tests for `getV3TaxonomyNodeRecordCounts`, which had none (`listTaxonomyNodeRecordCounts` wasn't even in the `vi.mock` factory) despite owning one of the nine sites.
- The 404-stops-the-poll test, verified to fail without the fix.

Coverage on changed code: taxonomy `operations.ts` 91.3% stmts / 98.9% lines, `hub-errors.ts` 87.9% / 90.6%.

**Against the live Hub** (`localhost:8080`), driving the real SDK → `HubError` → mapper chain with no mocks:
- A real Hub 404 → the `HubError.message` quoted at the top; mapped body carries none of it.
- A real Hub 503 (`"Taxonomy service is not available."`) → mapped to 503 with taxonomy's wording.
- A real Hub 400 → `detail` is a fixed `"One or more request parameters are invalid"` with the actionable message in `invalid_params[0].reason` and its own `code: "validation"`. That corrected a test fixture that had assumed the message lived in `detail`: relaying detail alone would have dropped the only text explaining a refused run, so the params relay is now asserted.

**Security** — gitleaks and semgrep (`p/typescript`, `p/secrets`) clean on the diff. No Prisma queries added and `requireUnifyDirectoryAccess` still runs before every Hub call, so tenant scoping is untouched; no response gains a field.

### Note on the rebase

#8650's review raised the same 503 problem for `feedbackRecords`. Rather than leave `main` briefly wrong and
fix it one PR later, that fix went into #8650 itself (now on `main`) and this branch rebased onto it, so
commit 1 here is a pure module relocation. One knock-on: the shared mapper's `tenant_id` → `dataset_id` rewrite (also added on
the base) now applies to the taxonomy routes too. That is correct at the product level — "dataset" is what the
UI calls a feedback directory, in 45 user-facing strings — though it does highlight that the taxonomy routes
spell their own query parameter `directoryId`. That inconsistency between v3 surfaces predates this PR and is
left alone here.

## QA / Test Plan

**How to test**

- [ ] Open Topics & Subtopics for a feedback directory on a licensed EE workspace → the page loads as before. Regenerating and renaming/removing nodes all behave unchanged on the happy path.
- [ ] With the Hub's taxonomy compute service unconfigured, press Generate → the request fails with **503** and a message naming *both* the embedding model and the taxonomy service, and suggesting a retry. It must **not** say only "embeddings".
- [ ] On a directory with too few embedded records (needs a fully configured Hub), press Generate → **400**, and the response still tells you *"at least N embedded text feedback records are required; found M"* (now in `invalid_params`). If that has gone generic, the relay is broken.
- [ ] **The poll fix:** start a generation, note the run id, delete that run row in the Hub DB, and leave the page open for 60s → the browser must **stop** issuing `/api/v3/.../taxonomy/runs/{id}` (count them in the Network tab). Before this change it kept polling every 5s forever.
- [ ] Stop the Hub entirely and load the page → **200** with the empty/unavailable state (not an error page), and the response body contains no `HUB_API_KEY`, host/port, or stack trace.
- [ ] Inspect every error body from the steps above for Hub internals — host/port, `HUB_API_URL`, `HUB_API_KEY`, `request_id` from the Hub, Go source paths, `/v1/...` paths. There should be none, including in `unavailableMessage`.
- [ ] Check the server logs for the failing cases: the upstream diagnostic must now be *there* instead. A bad `HUB_API_KEY` should log at **error** level with the Hub error attached.

**Preconditions / test data**

- Licensed EE org with Unify Feedback enabled, and a feedback directory with text records.
- The 400 case needs a Hub with both `EMBEDDING_MODEL` and the taxonomy compute service configured, plus a directory below the embedded-record minimum. The 503 case needs the compute service *un*configured.
- The record threshold is the Hub's `minimumEmbeddingCount` — **please confirm the live value** rather than trusting the 750 used in test fixtures.

**Risks & regressions**

- Status codes on these routes are now wire-visible where they were uniformly 502. Nothing in the Unify UI reads `V3ApiError.status` and retry is a flat `retry: 1` with no status predicate, so apart from the 404 fix this is hardening — but it is a behaviour change for any direct API consumer.
- The `200 + unavailable` shape is load-bearing for the Generate gate; it is unchanged, only its message is fixed text. Re-check that the gate still enables and disables correctly.
- No E2E coverage exists on this surface (`apps/web/playwright/` has no Unify specs), so the route-level tests are the safety net.

- none

## Follow-ups (not in this PR)

- **ENG-2048 part 2** — the `handleUnexpectedError` duplication (six variants in `surveys/lib/operations.ts` with drifting log messages, plus `list-resource.ts`, two in `lib/auth.ts`, and the api-wrapper catch).
- `surveys/generate/route.ts:112` returns `problemNotFound` **with** the `organizationId` in the body, against `response.ts`'s own JSDoc.
- `list-resource.ts:83` forwards an `InvalidInputError` message into `problemBadRequest` and `invalid_params[0].reason`.
- `api-client.test.ts:192` simulates `409 "already running"`, but the Hub returns **200 + `in_progress: true`** for that; 409 is `tenant_write_conflict`. It also has a pre-existing type error. Left alone to keep this diff scoped.
- Client hardening: a `retry` predicate so 4xx isn't retried, poll backoff on 429/503, and honouring `Retry-After` — nothing in `apps/web` does today.

### DIFF
--- a/apps/web/app/api/v3/feedbackRecords/lib/access.ts
+++ b/apps/web/app/api/v3/feedbackRecords/lib/access.ts
@@ -6,7 +6,7 @@ import type { TV3Authentication } from "@/app/api/v3/lib/types";
 import { getFeedbackDirectoriesByWorkspaceId } from "@/modules/ee/feedback-directory/lib/feedback-directory";
 import { retrieveFeedbackRecord } from "@/modules/hub/service";
 import type { FeedbackRecordData } from "@/modules/hub/types";
-import { hubErrorToProblemResponse } from "./errors";
+import { hubErrorToProblemResponse } from "@/app/api/v3/lib/hub-errors";
 
 /**
  * Tenant isolation for the feedback-records surface. Two guards live here, and every operation goes
--- a/apps/web/app/api/v3/feedbackRecords/lib/errors.ts
+++ b/apps/web/app/api/v3/feedbackRecords/lib/errors.ts
@@ -1,37 +1,16 @@
 import "server-only";
-import type { z } from "zod";
-import type { logger } from "@formbricks/logger";
-import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
-import {
-  type InvalidParam,
-  problemBadGateway,
-  problemBadRequest,
-  problemConflict,
-  problemForbidden,
-  problemInternalError,
-  problemPayloadTooLarge,
-  problemServiceUnavailable,
-  problemTooManyRequests,
-  problemUnprocessableContent,
-} from "@/app/api/v3/lib/response";
-import type { HubError } from "@/modules/hub/utils";
 
 /**
- * Error mapping for the feedback-records surface: Hub failures and unexpected throws → controlled v3
- * problem responses. Split out of `operations.ts` so the operations read as a dispatcher and the mapping
- * rules — the part with the disclosure risk — can be tested on their own.
+ * Wording specific to the feedback-records surface. The Hub-error *mapping* is shared and lives in
+ * `@/app/api/v3/lib/hub-errors`; these are the two messages only this surface can produce, passed into it.
  */
-
-// Bounds on what a Hub 4xx may contribute to our response body (see `hubErrorToProblemResponse`).
-const MAX_RELAYED_INVALID_PARAMS = 20;
-const MAX_RELAYED_DETAIL_LENGTH = 512;
 
 /**
  * Semantic search and similarity need embeddings, which are optional in the Hub. Our own static message,
  * not the upstream body: it names the setting to change, on both processes that need it.
  *
  * Passed explicitly by the two search operations rather than being the 503 default, because those are the
- * only ones for which it is true — see `GENERIC_SERVICE_UNAVAILABLE_DETAIL`.
+ * only ones for which it is true; every other operation gets the generic default from `hub-errors`.
  */
 export const EMBEDDINGS_UNAVAILABLE_DETAIL =
   "Semantic search is not available: the feedback service has no embedding model configured. A self-hosting administrator can enable it by setting EMBEDDING_PROVIDER and EMBEDDING_MODEL on both the Hub API and the Hub worker.";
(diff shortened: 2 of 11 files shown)


xernobyl:
Superseded by #8707 (ENG-1886), which landed the same core fix from a different angle. Closing rather than rebasing — a rebase would actively undo improvements #8707 made.

**What #8707 already does on main, better or equal:**
- Removes the raw `result.error.message` from all nine sites — and goes further than this branch did by **deleting `unavailableMessage` outright** instead of sanitising it, with a good reason: the UI renders a localized alert off the boolean, so any string sent from there is either unused or shipped untranslated. Rebasing this branch would re-add that field.
- Maps Hub 404 → 404 on the five read/mutate paths, correctly skipping it on the create (a 404 says nothing about a create).
- Maps NO_CONFIG → 503 via a new `isHubNotConfigured` helper.
- Its `hubFailureResponse` is documented at least as well as the helper here.

**One premise of this PR was simply wrong:** it claimed the taxonomy file "logs nothing at all, so the diagnostic goes to the client instead of to our logs". The diagnostic was already being logged one layer down — `modules/hub/service.ts` logs all 21 Hub failures with `logger.warn({ err, … })`. #8707 says so explicitly and is right. The structured logging added here was largely duplication, and is dropped with this PR.

**What is still genuinely live on main** (carved out of this branch; not filed anywhere yet):

1. **A run that no longer exists still polls forever.** `use-taxonomy-run.ts`'s `refetchInterval` treats "unknown because the last poll errored" as non-terminal, so a reaped or stale run leaves the UI on "generating…" indefinitely at ~12 requests/minute. #8707 made this *more* fixable by returning a real 404, but did not change the hook. ~5 lines plus a test.
2. **`service_unavailable` is emitted but is not in the `Problem.code` enum** — from `feedbackRecords/lib/errors.ts` (×2) and `taxonomy/lib/operations.ts` (×2, new in #8707). `conflict` did get added, via the workflows epic. The spec CI job only fires on `docs/api-v3-reference/**`, so nothing catches this. One line plus a bundle regen.
3. **A Hub 400's actionable detail is still lost.** Pressing Generate below the record threshold answers "Failed to start taxonomy generation" instead of the Hub's *"at least N embedded text feedback records are required; found M"* — the most useful message on this surface. It arrives in `invalid_params[0].reason` (the Hub keeps `detail` generic), so relaying `detail` alone would not recover it.
4. Minor: a genuine Hub **503** — taxonomy compute service unconfigured, distinct from NO_CONFIG — still collapses to 502. Verified live against a local Hub: it answers `503 "Taxonomy service is not available."`
5. Note, not a defect: there are now two independent Hub-error mappers (`feedbackRecords/lib/errors.ts` and taxonomy's local `hubFailureResponse`). Consolidating them is the DRY case this branch was making; it has no user-visible benefit, so it should stand on its own merits rather than ride along.

xernobyl:
Closing as superseded by #8707. Remaining live items documented above.

PR #8670, opened 2026-07-28
feat: project team and workspace grants to SpiceDB

BhagyaAmarasinghe:
## What does this PR do?

Implements [ENG-1716 — Project team membership and workspace grants into SpiceDB](https://linear.app/formbricks/issue/ENG-1716/project-team-membership-and-workspace-grants-into-spicedb) on top of ENG-1715.

PostgreSQL and the legacy evaluator remain authoritative. This PR extends the existing synchronous, post-commit, best-effort projection pattern to the current team and workspace permission model:

- Projects team ownership, `TeamUser.admin`, and `TeamUser.contributor`.
- Projects workspace ownership and `WorkspaceTeam.read`, `readWrite`, and `manage` as team-subject grants.
- Preserves multi-team access as independent relationships so SpiceDB computes the effective union without precomputing a highest permission.
- Deletes stale alternate role/grant relations atomically with the current relation update.
- Reconciles complete source snapshots in sequential batches of at most 1,000 relationship updates, retrying up to three passes when PostgreSQL changes concurrently.
- Cleans both resource-side and subject-side relationships for team, workspace, user, membership, and organization deletion cascades.
- Extracts ENG-1715's projection result/error runner into a shared server-only utility while preserving the organization projector's direct-path contract.

Mutation coverage includes:

- UI and API v2 team create, update, and delete.
- Workspace create, update healing, delete, and API v2 workspace-team create/update/delete.
- Invite, signup, SSO default-team, role-management, and membership-removal paths.
- API v2 organization-user team assignment changes.
- Legacy and Better Auth user deletion, plus organization cascade deletion.

Transaction-owned writes use an explicit deferred projection option and project only after the outer transaction commits. Multi-step non-transactional writes track only successfully committed targets and reconcile them in `finally`; projection failures never replace the source operation result or skip later cleanup.

Reviewer follow-up:

- The client validator and relationship packer now share one 1,000-update constant.
- Large deletion cascades run in bounded batches of at most 10 parallel relationship-deletion RPCs instead of one long serial chain.
- Serializable membership deletion retries Prisma `P2034` transaction conflicts at most three times before preserving the existing database-error behavior.
- The current-model migration explicitly resolves survey, dashboard, and response parents from PostgreSQL and checks the equivalent workspace permission in SpiceDB during ENG-1738 shadow evaluation. It does not require one high-cardinality SpiceDB parent edge per response.
- Removed the pre-existing discarded organization-membership query in role management.

The AuthZed smoke test verifies team roles, all workspace grant levels, multi-team permission union, grant downgrade/removal, membership revocation, idempotent team/workspace cascade deletion, outage recovery, and persistence. The AuthZed README documents mappings, transaction behavior, cascade behavior, the current-model parent-resolution strategy, and the mandatory ENG-1718 repair/backfill dependency.

This PR deliberately does **not** add backfill, shadow evaluation, enforcement, schema changes, database migrations, API-key projection, or new permission capabilities.

Fixes ENG-1716

## How should this be tested?

- `pnpm --filter @formbricks/web exec vitest run lib/authzed modules/ee/teams modules/api/v2/organizations modules/auth/invite/lib/team.test.ts modules/auth/signup/lib/__tests__/team.test.ts modules/ee/role-management/lib/membership.test.ts modules/organization/settings/teams/lib/membership.test.ts modules/workspaces/settings/lib/workspace.test.ts lib/organization/service.test.ts lib/user/service.test.ts modules/account/lib/better-auth-account-deletion.test.ts`
  - 26 files and 320 tests passed.
- `pnpm --filter @formbricks/web typecheck`
- `pnpm authzed:validate`
  - 27 relationships, 131 assertions, and 2 expected relations validated.
- `pnpm authzed:smoke`
  - Passed the isolated Docker lifecycle, projection, revocation, deletion, outage, recovery, persistence, and secret-log scan.
- Production web build with AuthZed disabled and credentials absent.
- Production web build with AuthZed enabled and a syntactically valid but unreachable endpoint.
  - Both builds passed, confirming there is no startup or build-time AuthZed RPC.
- ESLint on every changed TypeScript file and `git diff --check`.
- CodeRabbit follow-up (`560851d0f`):
  - Focused Vitest: 16 files and 175 tests passed.
  - `pnpm --filter @formbricks/web typecheck` passed.
  - Targeted ESLint and `git diff --check` passed.
- SonarQube follow-up (`28eaa641d`):
  - Fixed all 3 reported findings: deterministic string sorting and SSO provisioning complexity.
  - Focused Vitest: 3 files and 69 tests passed.
  - `pnpm --filter @formbricks/web typecheck` passed.
  - Targeted ESLint, Prettier, and `git diff --check` passed.
  - SonarCloud quality gate passed with 0 unresolved issues and 97.6% new-code coverage.
  - Full PR CI passed, including build, unit, E2E, Better Auth integration, lint, translations, and AuthZed schema validation.
- Reviewer follow-up (`c87f38b5c`):
  - AuthZed and affected membership Vitest: 12 files and 139 tests passed.
  - `pnpm --filter @formbricks/web typecheck` passed.
  - `pnpm authzed:validate` passed with 27 relationships, 131 assertions, and 2 expected relations.
  - `pnpm authzed:smoke` passed.
  - Forced production web build passed.
  - Targeted ESLint, Prettier, and `git diff --check` passed.

Local validation used Node 25.2.1, which produces the repository's package-manager engine warning because supported CI versions stop below Node 25; the checks themselves passed.

After merge, validate in `authzed-sandbox` by assigning one user to multiple teams with different workspace grants, changing each grant level, removing membership, and deleting a team and workspace. Existing records still require ENG-1718 before shadow evaluation.

### Required

- [x] Ran `pnpm build`
- [x] Checked for warnings; no change-specific warning remains
- [x] Created from the latest `origin/epic/authzed` target (`662d9d509`); merging `main` is not applicable to this epic-targeted PR
- [x] First PR at Formbricks? N/A; contributor CLA is already on file

### Appreciated

- [x] If a UI change was made: N/A; this is server-only projection work
- [x] Updated the AuthZed documentation for mappings, transactions, cascades, parent resolution, and repair/backfill


xernobyl (approved):
LGTM 🚀 This is a big generalisation of the ENG-1715 pattern and it holds up well. Thanks for picking up the org-deletion orphan note from the last one — cleaning the org's teams/workspaces on delete is exactly what I was after.

What I checked and liked:

- **Delete narrowing survives the generalisation.** All three new filters are properly scoped, and the workspace-team filter correctly carries `relation: "member"` so it only targets the `team#member` userset rather than everything on that workspace.
- **Atomicity per entity.** `packUpdateGroups` flushes before adding a group that would overflow, so a team's two role relations and a grant's three permission relations always land in one write — no window with zero or two roles.
- **`pairKey` is length-prefixed**, so `("ab","c")` and `("a","bc")` can't collide. Easy detail to get wrong.
- **Coverage is complete.** I grepped every `prisma.{team,teamUser,workspaceTeam,workspace}` write: the only untouched sites are `environmentState.ts` and `remove-branding`, both non-authz fields, so nothing bypasses projection.
- **The `deleteMembership` transaction change is my favourite bit** — moving the target-list read inside the transaction closes a real TOCTOU where a concurrently-added team membership would be deleted in Postgres but never projected.
- Refactor into `projection.ts` / `projection-boundary.ts` is behaviour-preserving, and it de-duplicated the ad-hoc try/catch in account deletion.
- I also read through the `sso-handlers.ts` extraction since that file is security-sensitive: pure restructuring, and the `email` → `inviteEmail` rename was required by the new parameter — the `inviteEmail !== email` mismatch check is intact (and clearer).

Non-blocking notes:

1. **Batch cap is duplicated** — `team-workspace.ts` has a local `MAX_RELATIONSHIP_UPDATES = 1_000` and `client.ts` has its own private `AUTHZED_MAX_RELATIONSHIP_UPDATES = 1_000`. They agree today, but lowering the client's cap would silently make the packer emit batches the client rejects. Worth exporting one constant.
2. **Serializable without a retry** — the new `deleteMembership` transaction can raise a serialization failure (`P2034`) under concurrent deletes and there's no retry, so it surfaces as an error. Loud rather than wrong, but a bounded retry is the usual pairing.
3. **Org deletion does its cleanup deletes serially** (2 RPCs per team + 1 per workspace), so a large org means a long chain of round trips in the request path. Candidate for batching, or for the async path in ENG-1719.
4. **Question worth settling before shadow mode:** this PR establishes the "project parent edges" pattern (`team→organization`, `workspace→organization`), which makes it noticeable that nothing projects `survey→workspace`, `dashboard→workspace` or `response→survey` — and no ticket covers them. Either the SpiceDB evaluator resolves those parents from Postgres (like `resolvers.ts` does today) and only asks SpiceDB at workspace level, or those edges need projecting before ENG-1738 can shadow survey/response actions. Both are fine, just worth deciding explicitly.

Also spotted in passing (pre-existing, not this PR): `role-management/lib/membership.ts` has a `prisma.membership.findMany({ where: { organizationId }, select: { userId: true } })` whose result is discarded — dead query, worth deleting sometime.

BhagyaAmarasinghe:
Addressed all follow-up notes in `c87f38b5c`:

1. Exported one shared `AUTHZED_MAX_RELATIONSHIP_UPDATES` constant for both the client validator and projector packer.
2. Added a bounded three-attempt retry for Prisma `P2034` conflicts around the serializable membership-deletion transaction.
3. Replaced the large-organization serial cleanup chain with deletion batches capped at 10 concurrent AuthZed RPCs.
4. Settled the parent-edge strategy for the current-model migration: ENG-1738 resolves survey/dashboard/response parents through the existing PostgreSQL resolvers and performs the equivalent workspace-level SpiceDB check. Direct resource-level checks remain blocked until a later projector/backfill covers those schema edges.
5. Removed the discarded `membership.findMany` query in role management.

Validation:

- AuthZed and affected membership Vitest: 12 files, 139 tests passed.
- Web typecheck passed.
- AuthZed schema validation passed: 27 relationships, 131 assertions, 2 expected relations.
- Isolated AuthZed Docker smoke test passed.
- Forced production web build passed.
- Targeted ESLint, Prettier, and `git diff --check` passed.

The PR description has been updated with the implementation and validation evidence.

PR #8673, opened 2026-07-28
fix: count only finished responses for survey response limit

jobenjada:
## What & why

A customer reported that a survey limited to 500 responses closed at **500 starts**, not 500 completed responses.

The "Close survey on response limit" setting is described in the UI as *"Automatically mark the survey as complete after N **completed** responses."* But the auto-complete check in the response pipeline compared the limit against `getResponseCountBySurveyId(surveyId)`, which counts **every** response row for the survey — including partial starts (`finished: false`). So the survey closed once *starts* (partial + finished) reached the limit, contradicting the label.

This fix makes the auto-complete threshold count **finished responses only**. The count now runs lazily — only when a limit is actually set — and the notification email's total-response count is deliberately left unchanged.

This is a long-standing bug (present in the pre-BullMQ pipeline too), not a recent regression.

### Follow-up fixes from review

Review turned up that the pipeline was only half the feature. Three further fixes:

**The survey editor validated the limit against total responses too.** `isSurveyValid` aborted the save when `survey.autoComplete <= responseCount`, where `responseCount` was the unfiltered count. Once the pipeline correctly keeps a survey open past its limit in starts, that state becomes normal — and the editor then refused **every** save and publish with *"response limit needs to exceed number of received responses (N)"* until the limit was raised above the start count. Fixed by adding `getFinishedResponseCountBySurveyId` and using it everywhere that compares against `survey.autoComplete`: the save/publish validation and the limit input's `min`, its default, and its on-blur warning.

The total count deliberately stays where *"has this survey been answered at all"* is the real question — the caution banner about editing a live survey, and the block-editing alert — since a partial start is still data recorded against the old structure.

**Already-closed surveys are left alone.** Auto-complete now skips a survey whose status is already `completed`. Previously every response arriving after closure re-issued the same survey update and queued another audit event.

**Response counts are no longer loaded when nothing consumes them.** The total count ran on every finished response even though its only remaining consumer is the notification email, so on the common path (nobody subscribed to response notifications) it was pure waste — and two queries rather than one, since the response service loads the full survey to build its where-clause and `cache()` does not dedupe that in a worker. Both counts are now conditional and run concurrently with the integration and follow-up work instead of blocking on it, so the auto-complete count costs no extra round trip either.

## Linear ticket

No Linear ticket — customer-reported bug filed directly. Not closing a GitHub issue.

## How this was tested

- `vitest run modules/response-pipeline modules/survey/lib/response.test.ts modules/survey/editor/lib/validation.test.ts` → **178 passed** ✅ (the pipeline suite went 24 → 28).
- Wider sweep over `modules/survey`, `modules/response-pipeline` and `lib/response` → **1097 passed** ✅, no collateral breakage.
- Regression tests added:
  - Auto-complete ignores starts: total 500 / finished 3 against a limit of 5 → survey stays open, and the completed-only count is the one queried.
  - Auto-complete fires once finished responses reach the limit.
  - No limit set → the completed-response count is never queried.
  - Survey already `completed` → no further count, no survey update, no audit event.
  - No notification subscribers → the total count is never queried and no email is sent.
  - Either count failing is logged and skipped without failing the job.
  - `getFinishedResponseCountBySurveyId` filters on `finished: true` and maps Prisma errors to `DatabaseError`.
- Each new behaviour was **negative-controlled**: reverting the status guard, the finished-vs-total counting, and the lazy total count each fails exactly its own test and nothing else.
- The total and completed counts now use **distinct values in test fixtures**, so a test can no longer pass while confusing the two; the notification email is asserted against the total specifically.
- `eslint` on all changed files → clean ✅ (2 pre-existing `react-hooks/exhaustive-deps` warnings in `survey-editor.tsx`, untouched by this diff).
- `prettier --check` on all changed files → clean ✅.
- `tsc --noEmit` for `apps/web` → no errors in any changed file ✅.

No UI change, so no screenshots.

## Changelog

This is a user-visible behaviour change, so it needs a changelog line — a survey with a response limit now stays open longer and collects more responses before closing, which on paid plans consumes more of the monthly response allowance.

Drafted into the Linear 5.3 release notes: https://linear.app/formbricks/pipeline/formbricks/release/53-67c924b9ec85/release-notes/53-128c0c9c0996

## QA / Test Plan

**How to test**

- [ ] Create and publish a **link survey**. In the survey editor → **Settings → Response Options**, enable **"Close survey on response limit"** and set the limit to a small number, e.g. **2**.
- [ ] **Only partial starts stay open:** Open the survey link and answer the first question but **do not** submit/finish. Repeat so there are ~3 partial starts. → Survey status stays **In progress** and the link keeps accepting responses (it must **not** close).
- [ ] **The editor still saves in that state:** with those ~3 partial starts and the limit still **2**, change anything in the editor (e.g. a question headline) and hit **Save**. → It saves. Before this fix it failed with *"Response limit needs to exceed number of received responses (3)"*.
- [ ] **Completions close it:** Now fully complete the survey **2 times** (reach the end / submit). → Shortly after the 2nd *completed* response, the survey auto-closes: status becomes **Completed** and opening the link shows the "survey completed / no longer accepting responses" screen. (Closing is processed by an async worker, so allow a few seconds.)
- [ ] **Edge — mixed:** With the limit still 2, on a fresh survey submit **1** complete response and leave several partials. → Survey stays open (only completions count toward the limit).
- [ ] **Edge — limit input:** With ~3 partial starts and 0 completions, open the response-limit field. → It accepts a limit of **1** or **2** (the minimum tracks *completed* responses, not starts).
- [ ] **Notification email unchanged:** with response notifications enabled for your user, complete a response. → The "View N more responses" email still reports the **total** response count (partials included).

**Preconditions / test data**

- A survey with the response-limit option enabled and a small `autoComplete` limit. Works on both Cloud and self-host. No special plan/entitlement or feature flag required.

**Risks & regressions**

- **Surveys that already closed early are not reopened** — the pipeline only ever sets status *to* `completed`, so anyone already affected has a survey stuck in "Completed" that must be reopened by hand. Deliberately out of scope here: reopening them automatically would be a new feature, not part of this fix.
- **Surveys with a limit now run longer than before**, since they close at N completions rather than N starts. For a survey with a low completion rate that is materially more responses (and, on Cloud, more billable volume) before it closes. Probably worth a changelog note so the higher count does not read as a regression.
- The response-finished **notification email** intentionally still uses the total response count; verify that email still renders a sensible number.
- Re-check that surveys with a limit still close (at the completed-response count) and that closing still writes the audit-log event.
- The editor's "this survey already has responses" caution still triggers on *any* response including partials — that is deliberate and unchanged.

- none
- No locale changes. Note that `response_limit_needs_to_exceed_number_of_received_responses` now reports a *completed*-response count while its wording still says "received responses" — rewording it needs a `pnpm i18n` regeneration pass, so it is left as a cosmetic follow-up rather than holding this fix.

### DIFF
--- a/apps/web/modules/response-pipeline/lib/process-response-pipeline-job.test.ts
+++ b/apps/web/modules/response-pipeline/lib/process-response-pipeline-job.test.ts
@@ -9,6 +9,7 @@ const {
   mockCreatePinnedDispatcher,
   mockDispatcherDestroy,
   mockGetIntegrations,
+  mockGetFinishedResponseCountBySurveyId,
   mockGetResponseCountBySurveyId,
   mockHandleIntegrations,
   mockLoggerError,
@@ -34,6 +35,7 @@ const {
     mockCreatePinnedDispatcher: vi.fn(() => ({ destroy: dispatcherDestroy })),
     mockDispatcherDestroy: dispatcherDestroy,
     mockGetIntegrations: vi.fn(),
+    mockGetFinishedResponseCountBySurveyId: vi.fn(),
     mockGetResponseCountBySurveyId: vi.fn(),
     mockHandleIntegrations: vi.fn(),
     mockLoggerError: vi.fn(),
@@ -103,6 +105,10 @@ vi.mock("@/lib/integration/service", () => ({
 
 vi.mock("@/lib/response/service", () => ({
   getResponseCountBySurveyId: mockGetResponseCountBySurveyId,
+}));
+
+vi.mock("@/modules/survey/lib/response", () => ({
+  getFinishedResponseCountBySurveyId: mockGetFinishedResponseCountBySurveyId,
 }));
 
 vi.mock("./posthog", () => ({
@@ -210,7 +216,8 @@ describe("processResponsePipelineJob", () => {
     mockGetIntegrations.mockResolvedValue([]);
     mockPrismaWebhookFindMany.mockResolvedValue([]);
     mockPrismaUserFindMany.mockResolvedValue([]);
-    mockGetResponseCountBySurveyId.mockResolvedValue(1);
+    mockGetResponseCountBySurveyId.mockResolvedValue(7);
+    mockGetFinishedResponseCountBySurveyId.mockResolvedValue(1);
     mockHandleIntegrations.mockResolvedValue(undefined);
     mockValidateAndResolveWebhookUrl.mockResolvedValue({ ip: "93.184.216.34", family: 4 });
     mockDispatcherDestroy.mockResolvedValue(undefined);
@@ -393,7 +400,9 @@ describe("processResponsePipelineJob", () => {
       "workspace_123",
       expect.objectContaining({ id: "survey_123" }),
       baseData.response,
-      1
+      // The notification email deliberately reports the *total* response count, not the
+      // completed-only count the response limit uses.
+      7
     );
     expect(mockPrismaSurveyUpdate).toHaveBeenCalledWith({
       data: {
@@ -417,10 +426,9 @@ describe("processResponsePipelineJob", () => {
   });
 
   test("only counts finished responses towards the auto-complete response limit", async () => {
-    // Total responses (starts) exceed the limit, but finished responses do not.
-    mockGetResponseCountBySurveyId.mockImplementation((_surveyId, filterCriteria) =>
-      Promise.resolve(filterCriteria?.finished ? 3 : 500)
-    );
+    // Total responses (starts) far exceed the limit, but finished responses do not.
+    mockGetResponseCountBySurveyId.mockResolvedValue(500);
+    mockGetFinishedResponseCountBySurveyId.mockResolvedValue(3);
     mockPrismaSurveyFindUnique.mockResolvedValue({
       ...survey,
       autoComplete: 5,
@@ -437,9 +445,92 @@ describe("processResponsePipelineJob", () => {
     ).resolves.toBeUndefined();
 
     // The auto-complete decision must be based on finished responses only.
-    expect(mockGetResponseCountBySurveyId).toHaveBeenCalledWith("survey_123", { finished: true });
+    expect(mockGetFinishedResponseCountBySurveyId).toHaveBeenCalledWith("survey_123");
     // 3 finished responses < limit of 5 → the survey must stay open.
     expect(mockPrismaSurveyUpdate).not.toHaveBeenCalled();
+  });
+
+  test("auto-completes the survey once finished responses reach the limit", async () => {
+    mockGetFinishedResponseCountBySurveyId.mockResolvedValue(5);
+    mockPrismaSurveyFindUnique.mockResolvedValue({
+      ...survey,
+      autoComplete: 5,
+    });
+
+    await expect(
+      processResponsePipelineJob(
+        {
+          ...baseData,
+          event: "responseFinished",
+        },
+        baseContext
+      )
+    ).resolves.toBeUndefined();
+
+    expect(mockPrismaSurveyUpdate).toHaveBeenCalledWith({
+      data: {
+        status: "completed",
+      },
+      where: {
+        id: "survey_123",
+      },
+    });
+  });
+
+  test("does not count finished responses when no response limit is set", async () => {
+    // The default survey fixture has autoComplete: null.
+    await expect(
+      processResponsePipelineJob(
+        {
+          ...baseData,
+          event: "responseFinished",
+        },
+        baseContext
+      )
+    ).resolves.toBeUndefined();
+
+    expect(mockGetFinishedResponseCountBySurveyId).not.toHaveBeenCalled();
+    expect(mockPrismaSurveyUpdate).not.toHaveBeenCalled();
+  });
+
+  test("does not re-close a survey that is already completed", async () => {
+    mockGetFinishedResponseCountBySurveyId.mockResolvedValue(50);
+    mockPrismaSurveyFindUnique.mockResolvedValue({
+      ...survey,
+      autoComplete: 5,
+      status: "completed",
+    });
+
+    await expect(
+      processResponsePipelineJob(
+        {
+          ...baseData,
+          event: "responseFinished",
+        },
+        baseContext
+      )
+    ).resolves.toBeUndefined();
+
+    expect(mockGetFinishedResponseCountBySurveyId).not.toHaveBeenCalled();
+    expect(mockPrismaSurveyUpdate).not.toHaveBeenCalled();
+    expect(mockQueueAuditEventWithoutRequest).not.toHaveBeenCalled();
+  });
+
+  test("does not count total responses when nobody subscribes to notifications", async () => {
+    // No user has response notifications enabled, so nothing consumes the total count.
+    await expect(
+      processResponsePipelineJob(
+        {
+          ...baseData,
+          event: "responseFinished",
+        },
+        baseContext
+      )
+    ).resolves.toBeUndefined();
+
+    expect(mockPrismaUserFindMany).toHaveBeenCalled();
+    expect(mockGetResponseCountBySurveyId).not.toHaveBeenCalled();
+    expect(mockSendResponseFinishedEmail).not.toHaveBeenCalled();
   });
 
   test("skips auto-complete when the finished response count cannot be loaded", async () => {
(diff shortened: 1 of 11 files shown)


apps/web/modules/response-pipeline/lib/process-response-pipeline-job.ts
```diff
@@ -538,29 +538,36 @@ const sendNotificationEmailsSafely = async ({
...
+  if (!survey.autoComplete) {
+    return;
+  }
+
+  // The response limit is defined in terms of completed responses, so only finished
+  // responses may count towards the auto-complete threshold. Counting all responses would
+  // close the survey once the number of starts (partial + finished) hits the limit.
+  let finishedResponseCount: number;
+  try {
+    finishedResponseCount = await getResponseCountBySurveyId(survey.id, { finished: true });
```
xernobyl: **[P3] Perf — this could join the existing parallel fetch, and the total count is now dead weight on the common path**

Two small things now that auto-complete owns its own count:

1. This runs serially, after the emails. `survey.autoComplete` is already known up in `runResponseFinishedSideEffects`, so it could go into the existing `Promise.allSettled` in `loadResponseFinishedContext` and save a round trip.
2. `loadResponseFinishedContext` still fetches the total count unconditionally (line 339), but its only remaining consumer is the notification email — and `sendNotificationEmailsSafely` returns early when nobody has notifications enabled (line 499). On that path it's pure waste, and it's two queries rather than one, since `getResponseCountBySurveyId` also does an internal `getSurvey()` (`apps/web/lib/response/service.ts:727`).

Net for a limit-enabled survey today is 4 queries (2× getSurvey + 2× count) where 2 would do. Worth flagging that the description's "count now runs lazily — only when a limit is actually set" holds for this new count but not for the total one.

(Also, `reactCache` won't dedupe the inner `getSurvey` here — there's no React request scope in the BullMQ worker, and the two calls have different args anyway.)

apps/web/modules/response-pipeline/lib/process-response-pipeline-job.ts
```diff
@@ -538,29 +538,36 @@ const sendNotificationEmailsSafely = async ({
...
+        err: error,
+      },
+      "Response pipeline survey auto-complete skipped because the finished response count could not be loaded"
+    );
 
     return;
   }
 
-  if (!survey.autoComplete || responseCount < survey.autoComplete) {
+  if (finishedResponseCount < survey.autoComplete) {
```
xernobyl: **[P3, nit] No `status !== "completed"` short-circuit**

Pre-existing rather than introduced here, but this is the function being rewritten so I'll mention it: once the survey is closed, every further finished response still re-issues the `prisma.survey.update` at line 577 and queues another audit event at line 597. A `survey.status === "completed"` early return alongside the `!survey.autoComplete` check would skip both.

xernobyl (commented):
Had a look at this — the core fix is right. I verified it beyond the tests, since they mock the response service and so a wrong filter key would have passed CI silently:

- `finished` is a real key on `ZResponseFilterCriteria` (`packages/types/responses.ts:220`) and `buildWhereClause` maps it into the Prisma count where-clause (`apps/web/lib/response/where-clause.ts:48`), so `{ finished: true }` does reach SQL.
- `handleSurveyAutoCompleteSafely` has exactly one call site, inside `runResponseFinishedSideEffects`, so only completions can advance the counter. Semantically right.
- Moving the error handling inside the function also means a transient failure of the *total* count no longer suppresses auto-complete, which is a small improvement on before.

One thing I think is worth fixing, plus two smaller notes inline.

### [P2] The survey editor still validates the limit against total responses

`apps/web/modules/survey/editor/lib/validation.ts:293` aborts the save when `survey.autoComplete <= responseCount`, and that `responseCount` is the unfiltered count from `apps/web/modules/survey/lib/response.ts:6` (`count({ where: { surveyId } })`). It's threaded `editor/page.tsx:62` → `survey-editor.tsx:219` → `survey-menu-bar.tsx:411` (also `:495` and `:553` — save, save & close, publish).

So with a limit of 500, 800 starts and 200 completions — a state this PR deliberately keeps open — the editor refuses every save and publish with "response limit needs to exceed number of received responses (800)" until the limit is raised above 800, even though only 200 of the 500 completions are in. Before this change the survey auto-closed at 500 starts, so that state was effectively unreachable; making it a normal steady state is what makes this reachable.

Same total-based assumption in `response-options-card.tsx:248` (default limit `Math.max(25, responseCount + 5)`) and the blur toast at `:270`.

Passing a `{ finished: true }` count into the editor would put both sides on one definition. Happy to be told this is fine as a follow-up rather than a blocker — but as it stands the same feature disagrees with itself about what "response limit" counts.

---

Also +1 to CodeRabbit's open nit on the test: both counts come from the same mock, so a refactor that switched the notification email to the finished count would still pass all 24 tests. Probably worth the one line.

Nothing else from me — Sonar gate is clean, and the QA plan is thorough; the partial-starts-stay-open step is exactly the right manual check. 👍

apps/web/modules/response-pipeline/lib/process-response-pipeline-job.test.ts
```diff
@@ -416,6 +416,65 @@ describe("processResponsePipelineJob", () => {
...
+        },
+        baseContext
+      )
+    ).resolves.toBeUndefined();
+
+    // The auto-complete decision must be based on finished responses only.
+    expect(mockGetResponseCountBySurveyId).toHaveBeenCalledWith("survey_123", { finished: true });
+    // 3 finished responses < limit of 5 → the survey must stay open.
+    expect(mockPrismaSurveyUpdate).not.toHaveBeenCalled();
+  });
```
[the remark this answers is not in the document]
xernobyl: Addressed in 210fadb7c, a bit differently than suggested. The total and completed counts now resolve to **different** values in the test fixtures (7 vs 1), so the notification-email assertion pins the total count specifically. Asserting the unfiltered call on its own would still have passed if the email were later switched to the completed-only count, since both came from the same mock returning the same number.

There is also a new test asserting the total count is not queried at all when nobody subscribes to notifications, which is now the case.

apps/web/modules/response-pipeline/lib/process-response-pipeline-job.ts
xernobyl: Fixed in 210fadb7c.

Both counts are now conditional — the total only when a notification recipient will consume it, the completed-only count only when a limit actually applies — and both are started before the integration and follow-up work so they overlap with it rather than adding a round trip.

The completed count also moved to a lightweight `getFinishedResponseCountBySurveyId` helper, which drops the redundant full-survey fetch that `getResponseCountBySurveyId` does to build its where-clause. So a limit-enabled survey now costs 1 query here instead of 2, and the no-subscriber path costs 0 instead of 2.

apps/web/modules/response-pipeline/lib/process-response-pipeline-job.ts
xernobyl: Fixed in 210fadb7c. `getAutoCompleteThreshold` now returns `null` for a survey that is already `completed`, so it skips the count, the update and the audit event in one place.

Covered by a new test, and negative-controlled — removing the guard fails exactly that test and nothing else.

xernobyl:
@coderabbitai can you rereview?

xernobyl:
## Manual verification (local, 2026-07-30)

Ran the two cases that no automated test covers, against a dev server on this branch with the real BullMQ worker in-process and a real Postgres. Fixture: a link survey with `autoComplete = 2`, 3 partial starts, 0 completions.

**Editor saves while starts exceed the limit** — edited a question headline and saved. Save succeeded and persisted; survey stayed `inProgress` at `limit=2, total=3, finished=0`. Before this fix the same action failed with *"Response limit needs to exceed number of received responses (3)"*, because the guard compared the limit against total responses.

**Limit still closes on completions** — submitted 2 completed responses through the public link. Survey auto-closed at `status=completed, limit=2, total=5, finished=2`, i.e. it closed on the 2nd *completed* response while carrying 5 total; the 3 partial starts never counted. The link then correctly served the "survey completed" screen.

Both behaved as intended. Leaving the QA / Test Plan checkboxes unticked deliberately — those are for release QA on staging against a release build, and this was a local dev run.

Unrelated observation while testing: every survey in the local seed data has `endings = {}`, so finishing one renders a blank screen rather than an ending card. Not caused by this PR (reproduces on the untouched seed surveys) and no console error, but worth noting that `endings` is `prefault([])` with no minimum length, so a survey with zero endings is schema-valid and a respondent who finishes it sees nothing.

BhagyaAmarasinghe (approved):
Changes LGTM!

PR #8674, opened 2026-07-28
fix: consistent edit-caution modal CTAs across survey list & summary

jobenjada:
## What & why

The "edit a published survey" caution modal showed **opposite primary CTAs** depending on where it was opened:

- From the **survey list** (Edit on a published survey): primary CTA was **"Edit"**.
- From the **summary page** (pencil icon): primary CTA was **"Duplicate & edit"**, with "Edit" demoted to secondary.

Same modal, same decision, inverted default — so the destructive-ish path was the one-click default in one place and not the other.

This makes **"Edit" the primary CTA in both places**. On the summary page the duplicate option is kept, just demoted to the secondary button; its behavior is unchanged (duplicates the survey and opens the copy in the editor).

That swap required a fix in the shared dialog: the loading spinner was hardcoded to the primary button, so once the async duplicate action moved to the secondary slot the spinner would have appeared on the wrong button. `EditPublicSurveyAlertDialog` now tracks which action was clicked and spins that button while disabling the other. The pending action is keyed on a stable `ActionId` (`"primary" | "secondary" | "close"`) rather than the button's translated label, which is optional and can collide across locales.

**Changes**

- `SurveyAnalysisCTA.tsx` — swap CTAs: primary **Edit** (closes the modal, opens the editor), secondary **Duplicate & edit** (unchanged duplicate-and-edit action).
- `edit-public-survey-alert-dialog/index.tsx` — footer buttons carry a stable `ActionId`; spinner follows the clicked action, the other button is disabled while it runs.

No copy or i18n changes: the label keeps its existing `caution_edit_duplicate` string, so all 15 locales stay consistent.

### Screenshots — summary-page modal

| Before | After |
| :---: | :---: |
| before | after |

<sub>These are faithful component renders (real Tailwind tokens, dialog/button classes and labels), not live captures — reaching this modal needs a published survey with responses. The `assets-survey-edit-modal` branch holds only these two PNGs and can be deleted after merge.</sub>

## Linear ticket

No ticket — small UI-consistency fix spotted while working in the survey summary.

## How this was tested

- `pnpm scan-translations` (i18n gate) ✅ — all keys valid, all locales complete.
- `tsc --noEmit` on `apps/web` ✅ — no errors in changed files (only pre-existing stale-Prisma-client noise from `packages/database` in this worktree).
- `eslint` on both changed files ✅ — clean.
- No unit tests added: both changed files are `.tsx`, which per AGENTS.md are covered by Playwright E2E rather than component unit tests. The dialog has no existing E2E spec; the behavior below is manual QA.
- Screenshots above for the summary-page modal; the survey-list modal is visually unchanged (its buttons were already Edit / Cancel).

## QA / Test Plan

**How to test**

- [ ] Create a survey, publish it, and collect at least 1 response.
- [ ] Surveys list → open the survey's **⋯** menu → **Edit** → caution modal opens with **Edit** as the primary (dark) button and **Cancel** as secondary → clicking **Edit** opens the survey editor.
- [ ] Same modal → click **Cancel** → modal closes, you stay on the list, no navigation.
- [ ] Open the survey's **Summary** page → click the **edit (pencil)** icon → caution modal opens with **Edit** as the primary (dark) button and **Duplicate & edit** as the secondary (light) button — note the primary is **Edit**, not "Duplicate & edit" as before.
- [ ] From that modal click **Edit** → modal closes and the editor for the *existing* survey opens (URL `/workspaces/<id>/surveys/<surveyId>/edit`, same survey id).
- [ ] From that modal click **Duplicate & edit** → a spinner appears **on the Duplicate & edit button** (not on Edit), **Edit** is disabled while it runs, then a "Survey duplicated successfully" toast shows and the editor opens for a *new copy* (different survey id in the URL). The original survey and its responses are untouched.
- [ ] Edge case — survey with **0 responses**: clicking Edit (from either list or summary) goes straight to the editor with no caution modal.
- [ ] Edge case — a viewer/read-only member who cannot manage surveys: no edit entry point, so no modal.
- [ ] Edge case — switch the app language (e.g. Settings → language → Deutsch), reopen the summary-page modal, click the secondary button → the spinner still lands on the clicked button (this is what the stable-`ActionId` change protects; previously the label-based match was locale-dependent).

**Preconditions / test data**

- A published survey with ≥1 response, in a workspace where your role can manage surveys.
- No feature flag or entitlement involved; applies to Cloud and self-host alike.

**Risks & regressions**

- `EditPublicSurveyAlertDialog` is shared by three call sites: the surveys-list dropdown, the summary CTA, and the survey editor (`survey-editor.tsx`, which renders it with no action props and gets the single "Close" button). Re-check all three still render the right buttons and that the editor's close-only variant still dismisses.
- The summary-page primary action now navigates directly instead of going through the async duplicate flow — confirm no double-navigation or modal left open behind the editor.
- Duplicate-and-edit itself is unchanged code; the risk is only which button hosts it.

- none


Dhruwang (dismissed):
Thanks for the PR, it looks good and works well. Could you please review the code rabbit suggestion? I think that makes sense. Also, the translation changes are only limited to English. Do we also need to update other languages to reflect the same copy?

Dhruwang (approved):

PR #8675, opened 2026-07-29
fix(unify-feedback): restrict record delete/update to org owners and managers

pandeymangg:
## What & why

A feedback record's only tenancy is its dataset (the Hub tenant), and a dataset is shared by every workspace it is assigned to — the record itself carries no workspace. Every guard asked the same question, "is this record's tenant one of my workspace's datasets?", which is also true for records another workspace's surveys ingested. So a `readWrite` member of workspace B could delete or edit records produced by workspace A's surveys in a shared dataset.

Per the ticket discussion, changing a record is treated as an organization-level act until Hub records carry a workspace dimension. Applied on all three paths that can mutate a record:

- **Server action** — `deleteFeedbackRecordAction` authorizes on the organization role alone: owners and managers, no workspace-team fallback. The dataset-membership check stays on top of it as the cross-*organization* boundary (the record id is caller-supplied and the Hub resolves any id globally), and moves to `unify-feedback/lib/access.ts` with tests — including the deliberate not-found-instead-of-forbidden error shape that keeps record ids unprobeable.
- **Gateway** (envoy/traefik ext-authz, and the dev proxy) — for session users, `update`, `delete` and `bulkDelete` lose the workspace-team fallback. `create`, reads and semantic search are unchanged: adding records to a shared dataset is ordinary workspace work, like a CSV import.
- **v3 feedback-records API**, and so the `update_feedback_record` / `delete_feedback_record` MCP tools — new `requireFeedbackRecordMutationRole`, called before the record is fetched so a refusal cannot double as an existence oracle. This surface landed in #8650 while this PR was open and had the same gap.

Alongside:

- **Taxonomy** runs, node renames and node removals get the same gate. A dataset has one shared taxonomy tree, so rewriting it changes what every other workspace reads. Everyone else keeps the existing read-only view.
- **Record deletes now emit an audit event** under a new `feedbackRecord` target. The logged object identifies the record but omits `value_*`, which is end-user feedback that `redactPII` would not strip.
- **UI** — the records page derives `canDeleteRecords` (owners and managers) separately from `canWrite`, so `readWrite` members lose the row checkboxes, bulk-delete bar and the drawer's Delete button while keeping CSV import.

**API keys are deliberately not gated.** #8648 removed org-wide write as a grant, so a key now authorizes purely on its per-workspace permissions. Applying the organization rule to keys as well would leave them no way to mutate a record at all — a bigger change than this ticket asks for, and the opposite of what #8682 proposes. That question is left to #8682 on both the gateway and the v3 surface, and is called out in code at each site.

Trade-off worth a second opinion: a workspace lead who owns their own workspace's data now has to ask an owner or manager to delete a record. That is unavoidable while records have no workspace column — the server genuinely cannot tell one workspace's records from another's inside a shared dataset.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1770

## How this was tested

- `pnpm test` in `apps/web` → **545 files / 6998 tests pass** ✅ (after merging `main`)
- `tsc --noEmit` reports no new errors; ESLint and Prettier clean on every changed file.
- Tests added: `modules/ee/unify-feedback/lib/access.test.ts` (6) for the dataset guard and its error shape; `modules/hub/feedback-records-gateway.test.ts` (10, the first test for that authorizer) covering per-operation gating for session principals, the API-key carve-out, and that a key from another organization is refused on a mutation *and* a read; 6 cases in `app/api/v3/feedbackRecords/lib/operations.test.ts` for the new v3 gate, including that it runs before the record lookup.
- Tests updated: `taxonomy/lib/access.test.ts` and `taxonomy/lib/operations.test.ts` for the taxonomy mutation gate; `envoy-auth/service.test.ts` and `traefik-auth/service.test.ts`, whose `PATCH` cases asserted the old permissive access list.
- Merge note: resolving `feedback-records-gateway.ts` the mechanical way would have reinstated a pre-#8648 copy of `hasApiKeyImplicitFeedbackDirectoryAccess` without its cross-organization guard. Main's version is now the only definition, and the new cross-org test fails if it is ever dropped.
- No E2E: Unify Feedback is entitlement-gated and needs a live Hub service, neither of which the Playwright environment has. The QA plan below covers that gap manually.
- ⚠️ Still to attach: before/after screenshots of the records table and drawer for a `readWrite` member (missing checkboxes, bulk-delete bar and Delete button) and of the read-only taxonomy page.

## QA / Test Plan

**How to test**

Setup below gives you dataset D assigned to workspaces A and B, holding records created by a survey in A.

- [ ] As an org **owner or manager**, open `/workspaces/{A}/unify/feedback-records` → rows have checkboxes. Select one → a bar appears with "Delete" → confirm → the row disappears with a success toast.
- [ ] Same page, click a row → the details drawer opens with a **Delete** button in the footer → confirm → the drawer closes and the list refreshes without that record.
- [ ] As a plain org **member with `readWrite` on B only**, open `/workspaces/{B}/unify/feedback-records` → records created by A's survey are **still listed** (expected, shared dataset), but there are **no row checkboxes, no select-all, no bulk-delete bar**, and opening a row shows a drawer with **no Delete button**.
- [ ] Same user, same page → **"Import feedback"** is still offered when the workspace has a CSV source, and importing still works. This must not regress.
- [ ] Same user → `/workspaces/{B}/unify/taxonomy` shows the **"Read only"** badge with no generate, rename or remove controls. As an owner or manager the same page keeps all of them.
- [ ] Repeat the two records-page checks as a member with **`manage`** on B (not just `readWrite`) → also blocked from deleting.
- [ ] **MCP tools** as the `readWrite` member: `update_feedback_record` and `delete_feedback_record` against a record in D → both refused with **403** and a message naming owners and managers. `list_feedback_records`, `get_feedback_record`, `count_feedback_records` and `create_feedback_record` still succeed. As an owner or manager, update and delete succeed.
- [ ] Direct API as the `readWrite` member (browser session), against a record in D: `DELETE /api/v3/feedbackRecords/{recordId}` → **403**; `PATCH /api/v3/feedbackRecords/{recordId}` with `{"metadata":{"qa":"1"}}` → **403**; `DELETE /api/v3/feedbackRecords?tenant_id={D}` → **403**. As an owner, the single-record `DELETE` succeeds.
- [ ] API non-regression: `POST /api/v3/feedbackRecords` with `{"tenant_id":"{D}", …}` and `GET /api/v3/feedbackRecords?tenant_id={D}` still succeed for the `readWrite` member.
- [ ] **API keys are unchanged by this PR** — a workspace-scoped key with write on B can still create, read, update and delete records in D, exactly as on `main`. Verify it still behaves as before rather than expecting a new denial. A key belonging to a *different* organization must be refused (403) on every one of those.
- [ ] Taxonomy API as the `readWrite` member → `POST /api/v3/unify-feedback/taxonomy/runs`, `PATCH /api/v3/unify-feedback/taxonomy/nodes/{nodeId}` and `DELETE /api/v3/unify-feedback/taxonomy/nodes/{nodeId}` all return **403**, while `GET /api/v3/unify-feedback/taxonomy/state` and `/taxonomy/fields` still return **200**.
- [ ] With audit logging on, delete a record as an owner → an audit entry appears with action `deleted`, target type `feedbackRecord`, the record id, and **no feedback text** anywhere in the payload. Needs log access.

⚠️ Needs a human check, inferred rather than observed: how a tester should reach `/api/v3/feedbackRecords/*` on Cloud versus self-host. The in-app Next proxy for those paths is dev-only and returns 404 when `IS_PRODUCTION`, so on staging and Cloud these requests go through the envoy/traefik ext-authz path. Confirm the right host/URL with whoever owns that config.

**Preconditions / test data**

- Enterprise entitlement for `feedbackDirectories` (Unify Feedback). Without it every route above reports the feature is not enabled.
- A reachable Hub service (`HUB_API_URL`, `HUB_API_KEY`). If Hub is down the records list renders empty and every action fails.
- One organization with **two workspaces, A and B**, and **one dataset D assigned to both** (organization settings → feedback datasets → assign workspaces).
- A Formbricks feedback source in **workspace A** mapping a survey to D, plus a handful of submitted responses, so D holds records with `source_type: formbricks_survey` that belong to A.
- A CSV feedback source in **workspace B**, so the import affordance is testable there.
- Users: an org owner (or manager); a plain org **member** with `readWrite` on B through a team and no access to A; optionally another member with `manage` on B.
- For the MCP steps: an OAuth token for each of those users with the `feedbackRecords:write` scope.
- For the API-key steps: a workspace-scoped key with write on B, plus a key from a second organization.
- For the audit step: `AUDIT_LOG_ENABLED` set, the audit-logs entitlement, and access to application logs.

**Risks & regressions**

- Over-restricting is the main risk. Re-check that `readWrite` members can still list records, open the drawer, import CSV, and create or edit **feedback sources** — sources are workspace-scoped and deliberately untouched.
- The MCP tools are the newest surface here (#8650, merged during this PR). Worth exercising every feedback-record tool once as an owner to confirm nothing else was caught by the new gate.
- Taxonomy is the widest product change: `readWrite` members can no longer generate topics. If that proves too strict, the run trigger can be relaxed on its own without touching rename and remove.
- API-key authorization is intentionally untouched, so nothing there should change. If a key starts getting 403 on a mutation, that is a bug in this PR, not the intent.
- The records table renders one fewer column for non-deleters, since the checkbox column drops. Check the header, the empty-state row and horizontal scrolling for both roles.
- Bulk delete still issues one request per record (ENG-2047, unchanged). A large selection is slow but should complete without errors, and a partial failure should report how many succeeded.

- none


apps/web/modules/ee/unify-feedback/lib/access.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  test("collects the assigned directory ids", async () => {
+    vi.mocked(getFeedbackDirectoriesByWorkspaceId).mockResolvedValue([
+      { id: sharedDirectoryId, name: "Shared" },
+      { id: "clfd9876543210987654321098", name: "Support" },
+    ]);
+
+    const ids = await getWorkspaceDirectoryIds(workspaceId);
+
+    expect(ids).toEqual(new Set([sharedDirectoryId, "clfd9876543210987654321098"]));
```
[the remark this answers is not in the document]
pandeymangg: Valid, fixed in 2483dfde7.

The mock ignores its argument, so a wrong or omitted workspace id inside `getWorkspaceDirectoryIds` would have kept the test green — and all the candidates are `string`, so the types wouldn't have caught it either. The set it returns is what the cross-organization guard compares a record's `tenant_id` against, so pinning the lookup scope is worth the line.

apps/web/modules/ee/unify-feedback/actions.ts
```diff
@@ -34,27 +43,37 @@ const ensureAccess = async (
...
+ *
+ * A record's only tenancy is its feedback directory (the Hub tenant), and a directory is shared by
+ * every workspace it is assigned to — the record itself carries no workspace. A workspace
+ * `readWrite` check therefore cannot tell this workspace's records from another workspace's, so a
+ * member of workspace B could delete records that workspace A's surveys ingested. Until Hub records
+ * carry a workspace, deleting stays with the roles that own org-level data.
+ *
+ * Returns the organization id so callers can attribute the audit event without resolving it again.
+ */
+const ensureDeleteAccess = async (userId: string, workspaceId: string): Promise<string> => {
```
xernobyl: **[P2]** This is the actual new rule — owners/managers, no workspace fallback — and it's the one piece with no coverage. The PR moved the *directory* guard into `lib/access.ts` and tested it well (6 tests, including the deliberate not-found shape), but the *role* gate stays module-private here, so nothing catches a future edit that re-adds a `workspaceTeam` entry to that access array. Which is exactly the bug being fixed.

Moving `ensureUnifyEnabled` / `ensureReadAccess` / `ensureDeleteAccess` into the `lib/access.ts` you already created makes them testable and matches the convention of testing the lib rather than the action file. That also closes CodeRabbit's remaining open comment, which asked for the tests but not the move — and the move is what makes them possible.

apps/web/modules/hub/feedback-records-gateway.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      expect(decision.status === "deny" && decision.response.status).toBe(403);
+    });
+
+    test("denies an update for a workspace-scoped write key", async () => {
+      const decision = await authorize("PATCH", `/api/v3/feedbackRecords/${recordId}`, workspaceWriteKey);
+
+      expect(decision.status).toBe("deny");
+    });
+
+    test("allows a delete for an organization-wide write key", async () => {
```
xernobyl: **[P3]** `orgWriteKey` inherits `workspacePermissions: []` from the `apiKey()` helper on line 65, and that's a principal production can never produce: `authenticateGatewayRequest` calls `authenticateApiKeyFromHeaders` with no options (`modules/gateway-auth/lib/request.ts:63`), which rejects org-only keys outright — `if (!options.allowOrganizationOnlyApiKey && workspacePermissions.length === 0) return null` (`modules/api/lib/api-key-auth.ts:74`) → 401 before the authorizer ever runs.

So this asserts an unreachable state, and the matching QA step ("With an org-wide key (Access control → write) → allowed") will come back **401, not 200**, unless the key also carries a workspace grant. Worth giving `orgWriteKey` a workspace permission and adding that caveat to the QA line, so a tester doesn't read the 401 as a correct denial.

xernobyl (commented):
Traced every mutation path — server action, taxonomy v3 operations, and the gateway authorizer shared by envoy/traefik and the dev proxy — and the gate is applied consistently. All five remaining `requireUnifyDirectoryAccess` calls in `operations.ts` are genuinely reads, `canDeleteRecords` derives from the same org role the server checks (`getWorkspaceAuth`'s `isOwner`/`isManager` are org membership flags, so UI and server agree), `colSpan` matches the conditional column, and `updateFeedbackRecord` has no dangling references left. Nice call on the not-found-over-forbidden error shape and on keeping `value_*` out of the audit payload.

Two non-blocking notes inline.

One thing this review turned up that is **not** yours to fix here: `hasApiKeyImplicitFeedbackDirectoryAccess` grants on `organizationAccess.accessControl` without comparing the key's org to the directory's org, so a cross-org API key with org-wide write can reach another organization's records. I diffed it against `main` — pre-existing, not a regression from this PR. But by routing API-key `update`/`delete`/`bulkDelete` past the (org-scoped) workspace branch and onto org-wide write, this PR makes that unscoped branch the sole authorization for API-key mutations. Filed as ENG-2088 with the repro and a two-line fix. Your call whether to fold it in here or take it separately — the rest of the change stands either way.

xernobyl:
Heads-up from reviewing #8648: three open PRs now change `hasApiKeyImplicitFeedbackDirectoryAccess`, and this one and #8682 take opposite positions on the same question.

**This PR vs #8682.** Here, `update`/`delete`/`bulkDelete` are org-level only for API keys — the workspace path returns `false` for mutations. In #8682, `delete`/`bulkDelete` require `manage` weight *from a workspace permission*. So a key holding workspace `manage` is exactly what #8682 wants for a delete and exactly what this PR refuses. Only one of those can be the rule.

**This PR vs #8648 (verified, this is the sharp one).** #8648 removes the `organizationAccess.accessControl` shortcuts entirely — that flag governs org management, not workspace data, so record access derives solely from `workspacePermissions`. With both applied, an API-key mutation has no route left: `orgAccessControl?.write` no longer grants, and `isRecordMutation` then returns `false` before the workspace check. A key with org read+write *and* workspace `manage` gets 403 on PATCH/DELETE/bulkDelete. API-key mutation of feedback records dies completely.

Two smaller things:

- On its own, this PR doesn't close ENG-1980 — the `orgAccessControl?.write → return true` shortcut still runs *before* the new mutation gate, so a cross-org key with org-level write can still read, create and update another organization's records. #8648 is what closes that.
- Both this PR and #8682 add `apps/web/modules/hub/feedback-records-gateway.test.ts` as a new file, so they conflict with each other, and both conflict with #8648 (which moves the function out into `feedback-records-gateway-authz.ts`). This PR also edits the envoy/traefik cases #8648 rewrote.

Suggested order, for what it's worth: land #8648 first since it's the one that actually closes the cross-tenant hole and it's green, then rebase this onto `feedback-records-gateway-authz.ts`. But the workspace-permission-vs-org-level question between this and #8682 is worth settling explicitly first — it's a product decision about who may edit records in a shared directory, not something a rebase resolves. @mattinannt for the #8682 side.

xernobyl:
Heads-up: this went `CONFLICTING` when #8648 merged this morning, and the conflict is in a spot where the mechanical resolution is the wrong one — so flagging before you rebase rather than after.

**What moved.** #8648 lifted `hasApiKeyImplicitFeedbackDirectoryAccess` out of `modules/hub/feedback-records-gateway.ts` into a new `modules/hub/feedback-records-gateway-authz.ts` (the gateway now imports it and calls it at ~line 237). Its first statement is the ENG-1980 cross-org guard:

```ts
if (authentication.organizationId !== directoryOrganizationId) {
  return false;
}
```

**Why that matters here.** This PR edits that same function in its *old* in-gateway location — adding the `isRecordMutation: boolean` parameter with the early `return false`, and gating the `workspaceTeam` fallback on it. So resolving `feedback-records-gateway.ts` by keeping this branch's side would reinstate an in-gateway copy of the function **without** the organization check, silently reopening the cross-org API-key BOLA that #8648 just closed (ENG-1980, and its duplicate ENG-2088). Tests wouldn't necessarily catch it: #8648's coverage lives in `feedback-records-gateway-authz.test.ts`, which a "keep ours" resolution on the gateway file leaves untouched and passing.

**Resolution that keeps both fixes:**

1. Port the `isRecordMutation` parameter and its early `return false` **into** `feedback-records-gateway-authz.ts`, on top of the org check — so the function ends up with both guards rather than either.
2. Keep this PR's `workspaceTeam`-fallback change in the gateway (~lines 248-261) — that part didn't move.
3. Worth a test asserting both hold *together*: an API key from another org must be refused on a mutation **and** on a plain read, so neither guard can be dropped without a failure.

One design note while you're in there: #8648's docstring justifies the org check as defense-in-depth, on the grounds that a workspace-permission match already implies the same organization. This PR removes the workspace-team fallback for mutating operations, which changes which check is actually load-bearing on that path — so the org check stops being purely redundant there. Might be worth updating that comment so the next person doesn't read it as safely removable.

No action needed from me — just didn't want a rebase to quietly undo this morning's fix. (Unrelated, and no rush: #8669 is up for your review when you get to it — same cross-tenant class, `createWorkspace` teamIds plus the v2 response update path.)

apps/web/modules/ee/unify-feedback/actions.ts
xernobyl: **[P2]** Still open from my last review, so flagging once more rather than letting it slide: this is the actual new rule — owners/managers, no workspace fallback — and it's still the one piece with no coverage.

`lib/access.ts` holds only `getWorkspaceDirectoryIds` and `assertRecordBelongsToWorkspace`, there's no `actions.test.ts`, and `ensureUnifyEnabled` / `ensureReadAccess` / `ensureDeleteAccess` stay module-private here. So nothing catches a future edit that re-adds a `workspaceTeam` entry to this access array — which is exactly the bug this PR fixes.

The gap did narrow: `148e92154` gives the v3 path real coverage for the same rule (`operations.test.ts`, "asks for an organization owner or manager, with no workspace-team fallback"). That makes the asymmetry more obvious rather than less — the v3 gate is pinned, the server-action gate isn't.

Moving these three into the `lib/access.ts` you already created makes them testable and matches the convention of testing the lib rather than the action file. It also closes the one actionable comment still open from the bot, which asked for the tests but not the move — and the move is what makes them possible.

apps/web/app/api/v3/feedbackRecords/lib/access.ts
```diff
@@ -131,6 +133,65 @@ export async function resolveWorkspaceFeedbackTenant({
...
+ */
+export async function requireFeedbackRecordMutationRole({
+  authentication,
+  organizationId,
+  log,
+  requestId,
+  instance,
+}: TRequireMutationRoleParams): Promise<{ ok: true } | { ok: false; response: Response }> {
+  const userId = authentication && "user" in authentication ? authentication.user?.id : undefined;
+  if (!userId) {
```
xernobyl, line 177: **[P3]** Optional, and I checked before raising it: this guard's default branch is *allow*. `TV3Authentication` is `TAuthenticationApiKey | Session | null`, so both an API key and `null` land here with no `userId` and pass through.

Not reachable today — `requireV3WorkspaceAccess` returns 401 on `null` (`app/api/v3/lib/auth.ts:87`), this only runs after `resolveWorkspaceFeedbackTenant` has already succeeded, and the `"user" in authentication` idiom matches what that function does, so it's consistent with the module. No failure path, hence P3 and not higher.

It's the shape I'd push back on for an authorization check: keying "allowed" off the *absence* of a field means a new principal type, or a session whose user lookup ever comes back empty, is admitted silently. An explicit `"apiKeyId" in authentication` allowlist with everything else denied would fail closed and read as deliberate, which matters more than usual here given the doc comment says keys pass through *by design*. Fine to leave if you'd rather settle key semantics in one place later.

xernobyl (commented):
Rereviewed at `148e92154`. The conflict I flagged last time is resolved, and resolved the right way.

`148e92154` scopes the mutation rule to person-shaped principals, so the API-key branch now calls #8648's `hasApiKeyImplicitFeedbackDirectoryAccess` with the directory's organization, untouched, and `isRecordMutation` only drops the `workspaceTeam` fallback in the session branch. The short-circuit that would have denied every API-key mutation once #8648 landed is gone. Extending the gate to the v3 surface (and so the `update_feedback_record` / `delete_feedback_record` tools) is placed correctly too — after tenant resolution, before the record fetch, so a refusal can't double as an existence oracle. Description and QA plan track the new commit.

Also confirmed resolved: my earlier note about `orgWriteKey` asserting a principal production can't produce. Every API-key principal in the gateway suite now carries a real workspace grant, and the new "refuses a key from another organization" case pins #8648 at this level as well.

Two notes inline — one is my [P2] from last time, still open.

**Needs a rebase before it can merge.** Conflicts with `main` in `taxonomy/lib/operations.ts` and `operations.test.ts` against #8707 (ENG-1886, Hub 404 → 404 mapping). Both sides touch the same three functions and the same import line, so the resolution needs to keep this branch's `requireUnifyDirectoryMutationAccess` swap *and* #8707's `hubFailureResponse` mapping — dropping one would be easy and silent.

### Not yours to fix, but it changes what ships here

I went to check whether this branch had regressed #8682's `manage` tier for deletes and found `main` never got it. The last commit touching `feedback-records-gateway.ts` on `main` is `b6e10d32d` (#8648), and #8682's merge commit `7ac7a932f` touched no `hub/` file at all — so the delete/bulkDelete-requires-`manage` change (ENG-2083) was dropped while resolving against #8648, even though #8682 merged.

Net effect once this lands: sessions are owner/manager-gated for mutations, but an API key still deletes a record with workspace `write`. That's the gap ENG-2083 was filed for, so it probably wants reopening — and the rest of that stack is worth a look for anything else lost the same way. On the upside, it's also why the add/add conflict on `feedback-records-gateway.test.ts` I warned about is no longer a problem.

apps/web/modules/ee/unify-feedback/actions.ts
pandeymangg: Done in 34bc49905 — the three gates moved into `lib/access.ts` and the delete access array is now asserted exactly, so re-adding a `workspaceTeam` entry fails a test.

apps/web/modules/hub/feedback-records-gateway.test.ts
pandeymangg: Fixed in 2db7c9e4e — explicit `apiKeyId` allowlist now, with anything that resolves to no user refused.

xernobyl (approved):
Approving at `847df389`. Green across all 18 checks, and the earlier E2E red was a stale merge base — the merge of main brought in the `survey-ui` calendar fix and it went away, nothing to do with this branch.

Re-reviewed the merged result as full files rather than hunks this time, since a merge across #8648 and #8682 can be textually clean and still semantically wrong. It isn't:

- The `#8648` signature change (`hasApiKeyImplicitFeedbackDirectoryAccess` gaining `directoryOrganizationId` and moving into `feedback-records-gateway-authz.ts`) is wired correctly at the call site.
- No mutation path slipped through the merge. The four mutating v3 operations are exactly what the MCP tools call; `update` and `delete` are gated, creates are deliberately not, and the gateway's `RECORD_MUTATING_OPERATIONS` covers its side. Coverage is complete and consistent across the server action, the v3/MCP layer and the gateway.
- The UI gate matches the server gate — `getWorkspaceAuth`'s `isOwner`/`isManager` are org membership flags, so `canDeleteRecords` and `checkAuthorizationUpdated` can't drift apart.

Both of my earlier notes are closed, and the first one better than I asked: `ensureDeleteAccess` now asserts the `access` array exactly, so re-adding a `workspaceTeam` entry fails a test rather than quietly reopening the bug. The new v3 layer is the strongest part of the diff — tenant never taken from caller input, indistinguishable 403 for missing vs. foreign records, and the role check running before the record lookup so a non-mutator can't probe ids.

**One thing to do before this closes ENG-1770.** The API-key carve-out is sound reasoning, but the four comments defer it to #8682, which merged on 2026-07-31 without the change — Matti dropped it there precisely because he'd concluded this PR's model was right. So the deferral currently points at nothing.

I've filed ENG-2189 for it, with the trace and the options laid out: https://linear.app/formbricks/issue/ENG-2189/workspace-scoped-api-keys-can-mutate-another-workspaces-records-in-a

Could you repoint these four at ENG-2189 before merging? Comment-only, no logic change:

- `modules/hub/feedback-records-gateway.ts:55`
- `app/api/v3/feedbackRecords/lib/access.ts:157`
- `modules/hub/feedback-records-gateway.test.ts:163`
- `app/api/v3/feedbackRecords/lib/operations.test.ts:1668`

Worth being explicit in the ticket-closing note too: this fixes ENG-1770 for people, not for API keys. A `readWrite` member of workspace B can no longer touch workspace A's records; a key with `write` on B still can, on both the gateway and MCP.

Non-blocking nit: three files are pure formatter churn unrelated to the ticket — `ui/components/card-table/index.tsx` and `ui/components/textarea/index.tsx` (Tailwind class reordering only) and a chunk of `feedbackRecords/lib/schemas.ts`. The Tailwind reordering suggests your local `prettier-plugin-tailwindcss` disagrees with CI's, which might be worth sorting out separately. Fine to leave in.

PR #8679, opened 2026-07-29
fix(security): scope historical response import to the authorized workspace [1/4 critical]

mattinannt:
## What & why

**1 of 4** from the pre-release security review, split out of #8667 by severity. **Severity: Critical** — cross-tenant disclosure of respondent PII, reachable by any authenticated user holding `readWrite` on a single workspace.

`importHistoricalResponsesAction` authorized the caller against the organization that owns the **feedback source**, then loaded an arbitrary caller-supplied `surveyId` with no ownership check and copied every one of that survey's responses into the caller's own feedback directory, where they are readable.

```ts
const organizationId = await getOrganizationIdFromFeedbackSourceId(parsedInput.feedbackSourceId);
await checkAuthorizationUpdated({ /* … authorizes the caller's OWN org/workspace … */ });
const survey = await getSurvey(parsedInput.surveyId);   // never checked against that org
return importHistoricalResponses(feedbackSource, survey);
```

So anyone with `readWrite` on one workspace could name a survey belonging to another organization and exfiltrate its responses. Survey ids are not secret — they appear in every `/s/<surveyId>` link.

Enterprise-licensed (Unify Feedback), so it affects Cloud and enterprise self-hosted — the multi-tenant deployments where the isolation matters.

## The fix

```ts
if (survey.workspaceId !== parsedInput.workspaceId) {
  throw new ResourceNotFoundError("Survey", parsedInput.surveyId);
}
```

Two deliberate choices, both differing from the reporter's suggested remediation:

- **Bound to the workspace, not the organization.** The reporter proposed comparing organizations. But the caller here may hold only `workspaceTeam` `readWrite` rather than org owner/manager, so an org-level check would still let a team member with access to one workspace import from a survey in a *different* workspace of the same organization. The survey must belong to the workspace actually authorized — which is also all the picker ever offers (`getSurveysForUnifyAction` → `getSurveys(workspaceId)`).
- **`ResourceNotFoundError`, not `AuthorizationError`.** "You may not read this" confirms the id exists, turning the endpoint into a cross-tenant survey-existence oracle. Not-found is indistinguishable from a bad id.

## Follow-up hardening in the same file

Review of the surrounding handlers turned up **no second vulnerability** — but it did turn up two places where the code was looser than the database, which is worth closing while we are here. Neither of these was exploitable.

**Mapped surveys were checked against the organization.** `createFeedbackSourceWithMappingsAction` and `updateFeedbackSourceWithMappingsAction` validated each mapped `surveyId` with `getOrganizationIdFromSurveyId(...) !== organizationId`. Mapping rows already carry a composite FK `(surveyId, workspaceId) → Survey(id, workspaceId)`, so a survey from another workspace could never actually be persisted — it passed the app check and then died on insert as `DatabaseError` wrapping a raw Prisma message. The guard now lives in one place, `resolveFormbricksMappingsInput`, right where the survey is loaded:

```ts
if (!survey || survey.workspaceId !== workspaceId) {
  throw new ResourceNotFoundError("Survey", surveyId);
}
```

This is strictly stronger than the organization comparison it replaces — in create, `organizationId` is itself derived from `parsedInput.workspaceId`, so workspace equality implies org equality — and it removes one org lookup per mapped survey.

**The update handler never pinned the feedback source to the request workspace.** Its `organizationId` comes from the feedback source while the `workspaceTeam` check uses the caller-supplied `workspaceId`, and only the CSV branch confirmed the two matched. That lookup is now hoisted to the top of the handler, so every branch — including a plain name/status update, which had no source check at all — is bound to one tenant, and a mismatch fails as not-found rather than as a Prisma error from the update.

All four write handlers in the file now agree: the feedback source and any survey must live in the workspace the caller was authorized on.

## Linear ticket

[ENG-1915](https://linear.app/formbricks/issue/ENG-1915) — In Progress. Reported privately by Kaivalya Ahir (github.com/default-cybe); worth coordinating advisory credit.

## How this was tested

- **New `apps/web/lib/feedback-source/mappings.test.ts`** — 6 tests over `resolveFormbricksMappingsInput`, the helper that now carries the tenancy guard: a survey from another workspace is refused, one foreign survey rejects the whole input, a missing survey is refused, unknown element ids and element types with no Hub field are skipped, and nothing mappable throws `InvalidInputError`. Mutation-checked — removing `survey.workspaceId !== workspaceId` fails exactly the two tenancy tests.
- **Removed `apps/web/lib/feedback-source/actions.test.ts`** (added in an earlier revision of this PR). `**/actions.ts` is excluded from coverage in `apps/web/vite.config.mts`, and the file mocked `authenticatedActionClient` so `.action(fn)` returned the raw function — no middleware, no zod, no audit logging. It asserted one `if` against a shell the test built itself. The logic worth covering moved to `mappings.ts` and is tested there.
- `pnpm test` (`apps/web`) — 529 files / 6602 tests ✅
- `tsc --noEmit`, `eslint` and `prettier` clean for all touched files.
- Not run here: Playwright E2E and manual QA — no Docker daemon or database in the authoring environment. CI covers E2E; the manual checks are below. The import guard itself is covered by code review and the schema, not by an executed test.

## QA / Test Plan

**How to test**

- [ ] With two organizations, each holding a feedback source and a survey with responses: as a user with `readWrite` in org A, call the import with A's `feedbackSourceId`/`workspaceId` and **B's** `surveyId` → rejected as not-found, and no records appear in A's feedback directory.
- [ ] Same call with a survey from a *different workspace of org A* → also rejected (this is the workspace-level part, stricter than an org check).
- [ ] Normal path: importing historical responses for a survey in the authorized workspace still works and imports the expected count.
- [ ] The survey picker in the feedback-source UI still lists only the current workspace's surveys.
- [ ] Create a Formbricks feedback source from the picker and confirm the mappings save; edit it and change the selected questions → still saves.
- [ ] Rename a feedback source and toggle its status → still saves (this path now revalidates that the source belongs to the workspace).
- [ ] Create and edit a CSV feedback source with field mappings → unchanged behaviour.

**Preconditions**

- Two organizations with separate workspaces, surveys and responses; an Enterprise license for Unify Feedback / feedback directories.

**Risks & regressions**

- The import behaves differently only for a request naming a survey outside the authorized workspace, which the UI never produces.
- Create and update now reject a cross-workspace mapped survey with `ResourceNotFoundError` where they previously raised `AuthorizationError` (cross-org) or a `DatabaseError` from the foreign-key violation (same org, other workspace). Both were already failures; only the error type changes. Nothing in the UI switches on those — `getTranslatedFeedbackSourceError` matches only `FEEDBACK_SOURCE_*` codes.
- No schema, API-shape or permission-level change.

- None.

## Known gaps, not addressed here

- `getResponseCountAction` in this same file authorizes its `workspaceTeam` item against a caller-supplied `workspaceId` while operating on a `surveyId`. That is the Medium finding from #8667 and is fixed in the Medium PR of this stack, not here.
- `checkWorkspaceTeamAccess` in `action-client-middleware.ts` never verifies that the workspace belongs to the `organizationId` it was called with. Not exploitable today — `getMembershipRole` throws for non-members before any access item is evaluated — but it applies to every action using the `workspaceTeam` access type, so it deserves its own ticket rather than a fix buried here.

## Merge order

Part of a four-PR stack, in severity order. This one targets `main` and is independent — it can merge on its own.

1. **#PENDING_CRITICAL — this PR** (Critical)
2. #PENDING_HIGH (High) — based on this branch
3. #PENDING_MEDIUM (Medium)
4. #PENDING_LOW (Low)


mattinannt:

| 1 | **#8679 ← this PR** | 🔴 Critical | `main` | 1 |

Merge in order. Each PR retargets to `main` automatically once the one below it lands, so each diff stays scoped to its own severity. This PR is the only one based directly on `main`, so it can merge on its own without waiting for the others.

All four replace #8667, which is now closed. 29 findings total, all four branches cut from current `main`.

---

mattinannt:
@pandeymangg reviewed your `b312f29` — it's a better fix than mine, and I've kept it as-is. Two follow-ups pushed, plus one gap worth your call.

**Your change, and why I agree with it**

Moving the guard into `resolveFormbricksMappingsInput` is strictly stronger than the org comparison it replaced, and matching the composite FK `(surveyId, workspaceId)` is the right level — my version left the create/update mapping paths at org scope, so a survey from a sibling workspace still passed authorization and failed on insert. Hoisting the source lookup to the top of the update handler is the part I'd missed entirely: the plain name/status update had no source check at all.

On dropping `actions.test.ts` — you're right, and I checked rather than took it on faith. `apps/web/vite.config.mts:91` excludes `**/actions.ts` from coverage, which is also why the earlier Sonar run read 0.0% new-code coverage on this PR and now reads 100%. Mocking `authenticatedActionClient` did mean the test exercised a hand-assembled function rather than the real pipeline, which is the kind of implementation-detail test AGENTS.md warns against. No objection.

**Pushed on top (`9b74ff3`)**

SonarCloud flagged one new MINOR issue on your commit (`typescript:S6582`, `mappings.ts:21`): `!survey || survey.workspaceId !== workspaceId` → `survey?.workspaceId !== workspaceId`. Behaviour is identical — `workspaceId` is a non-optional `string`, so a missing survey gives `undefined !== workspaceId` and still throws — and TypeScript still narrows `survey` past the guard for the `survey.blocks` read below it, which I verified with an isolated `tsc --strict` run rather than assuming. Your "rejects a survey that does not exist" case covers exactly that branch and still passes.

I also rebased #8680 / #8681 / #8682 onto this tip, so the rest of the stack carries your commit. The Medium `getResponseCountAction` fix in #8681 touches the same file and rebased without conflict.

**One gap I did not fix, because it's your call**

The Critical finding this PR was opened for — the `survey.workspaceId !== parsedInput.workspaceId` guard in `importHistoricalResponsesAction` — did not move to `mappings.ts`. It's still an inline check in `actions.ts`, which is coverage-excluded, and its only regression test was the one we just dropped. So that specific guard now has no automated coverage, even though it's the highest-severity item in the stack.

Two ways to close it, and I'd rather you pick than guess:

1. Extract the survey load + workspace check into a covered helper, mirroring what you did for the mappings — consistent with the precedent your commit sets, but it's another file.
2. Leave it. It's three lines next to the authorization block, and the import flow is a candidate for E2E coverage instead.

Happy to do either; say which and I'll push it. Otherwise this is green and merge-ready — the `PR Check Summary` failure you may have seen on `b312f29` was that run getting cancelled when I pushed on top, not a real failure.

---

pandeymangg (approved):

PR #8680, opened 2026-07-29
fix(security): close the high-severity auth, authz and exposure gaps [2/4 high]

mattinannt:
## What & why

**2 of 4** from the pre-release security review, split out of #8667 by severity. **Severity: High** — 10 findings. Each is reachable by a low-privilege or unauthenticated caller.

> ⚠️ **Stacked on #8679.** Base is `claude/security-critical`, so this PR's diff shows only the high-severity work. Merge #8679 first, then this retargets to `main`.

### Auth and identity

- **Cross-purpose token reuse.** Every token in `lib/jwt.ts` is signed with the same `NEXTAUTH_SECRET`, and `verifyTokenForLinkSurvey` accepted a token carrying **no `surveyId` claim at all**. The unauthenticated `createEmailTokenAction` mints exactly such a token for any registered address, so it could be replayed as `?verify=<token>` to clear *any* link survey's email gate as that person — no account, no inbox access. Tokens are now bound to the survey, carry explicit `purpose` claims and TTLs, and that action is rate-limited (it also answers "is this address registered?"). Confirmed as a failing test against the real mint/verify pair before the fix.
- **Invite acceptance not bound to the invited address or expiry.** The signature was verified; the token's email was never compared to the account being created, and expiry was never checked. Anyone holding a forwarded invite link — they get pasted into tickets and land in referrer logs — could redeem it with an address of their own **at the invited role, up to `owner`**. `resolveInviteMatch` already implemented the check and was wired only to the domain exemption.
- **`EMAIL_AUTH_DISABLED=1` didn't disable credential auth.** It only hid the login form; `emailAndPassword.enabled` was hardcoded `true`, leaving `POST /api/auth/sign-in/email` live on an instance configured as SSO-only. Any account still carrying a password could sign in **around the IdP** — and around whatever the IdP enforces: MFA, conditional access, deprovisioning.
- **Deactivation didn't end existing sessions.** `rejectInactiveUserOnSessionCreate` gates session *creation* only, and the session is served from Redis secondary storage behind a 5-minute signed cookie cache. Someone offboarded through the org-users API kept full access — server actions, v3 session routes, storage — for the life of a rolling 1-day session. `getSession` now reads `isActive`, the check the middleware's `getProxySession` already did.

### Privilege escalation

- **Team admin could grant itself `manage` on every workspace.** `updateTeamDetailsAction` is reachable by a team admin, who may be a plain org `member`, and its input carries the team's whole workspace grant list. `manage` satisfies every downstream `read`/`readWrite`/`manage` check, so the forged grant is honoured by the normal authorization layer. The UI already restricted these controls to owners/managers (`workspace-row.tsx`) — client-side enforcement only. Implemented as "did the grant set change at all", not "is `manage` being added": a level-only check would let a team admin add themselves at `read` and escalate in a second request, and would miss removals.
- **Integration OAuth routes used the wrong authorization helper.** Notion / Airtable / Slack / Google Sheets gated credential *writes* with `hasUserWorkspaceAccess`, which its own docstring says must not be used for mutating routes: it admits the `billing` role and read-only team members. Either could bind **their own** provider account as the workspace integration and start receiving another team's responses — an exfiltration channel granted by the lowest-privilege roles, that looks like a legitimate integration in the UI.

### Exposure

- **OAuth tokens in the RSC payload.** The four integration settings pages passed `config.key` — the provider access token, plus a long-lived **refresh** token for Google Sheets and Airtable — into `"use client"` wrappers, so those credentials were serialized into the page payload and readable in the page source. A refresh token survives password changes and session revocation. Redaction matches credential-shaped field *names* rather than a fixed pair, so a token a provider adds later is blanked by default.
- **Public link-survey page leaked internal config.** The whole survey object went to a client component, so `followUps` — internal recipient addresses, subjects and email bodies — and the target segment's filters were serialized for anonymous visitors. The renderer reads neither.
- **`ZStorageUrl` accepted `javascript:` / `data:` / `vbscript:`.** It guards survey image and video URLs, organization logos and favicons, and styling background URLs, all rendered into `src`/`href` — so an editable survey field became **stored XSS**, executing on the Formbricks origin for a link survey and on the customer's own site for an embedded one. Constrained to http(s) or a relative `/storage/` path, plus `isSafeMediaUrl` at the anchor sink, since the renderer receives survey JSON from the API and rows written before this validation can still carry an unsafe URL. The sink check rejects protocol-relative `//host` and `/\host`, which a browser resolves cross-origin.

### Rate limiting

- **Client IP taken from client-supplied headers.** `getClientIpFromHeaders` believed `cf-connecting-ip`, then the **leftmost** `X-Forwarded-For` entry. XFF is appended to by each proxy, so that entry is whatever the client sent: rotating it minted a fresh bucket per request and defeated every IP limit (login 10/15min, forgot-password 5/hr, sign-up, public client API), and forged the `ipAddress` stored on responses and audit entries. The address now comes from the rightmost `TRUSTED_PROXY_HOP_COUNT` entries. `cf-connecting-ip` is **not consulted at all** — a hop count cannot establish Cloudflare provenance, and Traefik/Envoy/nginx pass that header through untouched, so trusting it would hand the spoof straight back.

## Linear ticket

[ENG-2042](https://linear.app/formbricks/issue/ENG-2042) covers the team-admin escalation (reported by D0HY30N, CVSS 8.8, PoC available — worth taking them up on validating the fix). The others are documented here; see the tracking comment on #8667 for the full mapping.

## How this was tested

- `pnpm test` (`apps/web`) — 531 files / 6614 tests ✅; `packages/surveys` 26/670 ✅; `packages/survey-ui` 4/73 ✅.
- `eslint` and `tsc --noEmit` clean for every touched file. `TRUSTED_PROXY_HOP_COUNT` added to `turbo.json` as `turbo/no-undeclared-env-vars` requires.
- Regression tests: cross-purpose token replay (`lib/jwt.test.ts`), deactivated-user sessions (`session.test.ts`), team workspace-grant escalation (`workspace-access.test.ts` — add/remove/level-change/reorder/duplicate/no-op), credential redaction incl. unknown future token names (`redact-credentials.test.ts`), unsafe and protocol-relative media URLs (`video-upload.test.ts`), trusted-hop IP resolution and the one-shot misconfiguration warning (`client-ip.test.ts`), invite email/expiry binding (`signup/actions.test.ts`).
- `pnpm build --filter=@formbricks/surveys... --force` after the `packages/surveys` change, per AGENTS.md.

## QA / Test Plan

**How to test**

- [ ] Link survey with "Verify email before submission": a valid emailed link still lets you through and records the address. Then take a token minted by the email-verification action for survey A and replay it as `?verify=` on survey B → rejected.
- [ ] Invite flow: invite `a@example.com`, open the link, try to sign up as `b@example.com` → rejected; signing up as `a@example.com` still joins at the invited role. An expired invite is now rejected too.
- [ ] Set `EMAIL_AUTH_DISABLED=1`: the login form hides the credential fields **and** `POST /api/auth/sign-in/email` no longer authenticates. Unset → both work again.
- [ ] Deactivate a user via `PATCH /api/v2/organizations/<orgId>/users` while they have an active session → their next request is unauthenticated, rather than working for up to a day.
- [ ] Teams: as a **team admin who is a plain org member**, submit a team update with an extra workspace grant → rejected. Renaming the team and changing its membership still work. As an org owner/manager, workspace grants still save.
- [ ] Integrations: as an org `owner`, connect and disconnect Notion / Slack / Airtable / Google Sheets → unchanged. View the settings page source → no `access_token` / `refresh_token`. As a `billing`-role member, hit the OAuth start URL → 401.
- [ ] Survey editor: set an ending-card image URL to `javascript:alert(1)` → rejected by validation. A YouTube/Vimeo/Loom URL still renders.
- [ ] Public link survey page source: no follow-up recipients/subjects/bodies and no segment filter rules present.
- [ ] Client IP: with `TRUSTED_PROXY_HOP_COUNT` unset (default 1), send a forged `X-Forwarded-For` prefix **and** a forged `cf-connecting-ip` → the recorded `ipAddress` and the rate-limit bucket both follow the address the real proxy appended. Login and forgot-password limits still trip for a genuine client.

**Preconditions**

- Two organizations; a team whose workspace permission is exactly `readWrite`; a `billing`-role member; an Enterprise license for teams and SSO paths; both a Cloud and a self-hosted pass (`EMAIL_AUTH_DISABLED` and the proxy setting differ).

**Risks & regressions**

- **`TRUSTED_PROXY_HOP_COUNT` defaults to 1** — one trusted reverse proxy, which matches the Helm chart's Traefik/Envoy ingress, docker-compose behind a proxy, and Cloud. **A longer chain must raise it:** with a CDN in front of that proxy it is 2, and until it is raised the address resolves to the *inner* proxy's, so every client behind the CDN shares one rate-limit bucket. Setting it *higher* than the real chain re-opens the spoof. `0` is an explicit opt-out that identifies no client at all, because Next 16 gives route handlers no socket peer address to fall back on.
- Link-survey verification tokens now expire after 7 days and carry a purpose claim. Already-issued tokens have no `exp` and keep working, but any flow that stored one long-term should be re-checked.
- `ZStorageUrl` is stricter, so a pre-existing row holding a non-http(s) media URL will fail validation on the next save of that survey. Worth an audit query; this PR stops such rows rendering but does not clean them up.
- Integration credentials were readable by anyone who could open the settings page. Worth deciding whether to rotate the affected provider tokens as part of the release.
- Deactivation adds one indexed `user.findUnique` per `getSession`; `getSession` is wrapped in React `cache()`, so once per request.

- No DB migrations.
- New env var `TRUSTED_PROXY_HOP_COUNT` (integer 0–10, **default 1**): documented in `.env.example`, declared in `turbo.json`, pinned to `"1"` in `charts/formbricks/values.yaml`. No action for a single-proxy deployment.
- New rate-limit bucket `auth.emailToken` (10/hour per IP); no configuration needed.
- New error code `invite_token_invalid`; no locale changes (the client maps it to existing translated copy).

## Merge order

1. #8679 (Critical)
2. **#PENDING_HIGH — this PR** (High)
3. #PENDING_MEDIUM (Medium)
4. #PENDING_LOW (Low)

---

### DIFF
--- a/apps/web/lib/env.ts
+++ b/apps/web/lib/env.ts
@@ -315,6 +315,18 @@ const parsedEnv = createEnv({
       .optional()
       .or(z.string().refine((str) => str === "")),
     RATE_LIMITING_DISABLED: z.enum(["1", "0"]).optional(),
+    // Number of reverse proxies in front of the app whose X-Forwarded-For entries can be believed.
+    // Unset falls back to 1 (see TRUSTED_PROXY_HOP_COUNT in lib/constants.ts); an explicit 0 trusts no
+    // forwarding header at all. See resolveClientIp in lib/utils/client-ip.ts.
+    // Preprocessed because `z.coerce.number()` turns "" into 0, and 0 is a *valid* value here (the
+    // explicit "trust nothing" opt-out) rather than something `.min()` would reject. A deployment that
+    // renders the variable empty when unset — the common docker-compose / Helm shape — would otherwise
+    // silently opt out of trusting any forwarding header, collapsing every request into one rate-limit
+    // bucket. The neighbouring numeric vars only fail loudly on "" because their minimums exceed 0.
+    TRUSTED_PROXY_HOP_COUNT: z.preprocess(
+      (value) => (value === "" ? undefined : value),
+      z.coerce.number().int().min(0).max(10).optional()
+    ),
     TELEMETRY_DISABLED: z.enum(["1", "0"]).optional(),
     S3_ACCESS_KEY: z.string().optional(),
     S3_BUCKET_NAME: z.string().optional(),
@@ -480,6 +492,7 @@ const parsedEnv = createEnv({
     PASSWORD_RESET_TOKEN_LIFETIME_MINUTES: process.env.PASSWORD_RESET_TOKEN_LIFETIME_MINUTES,
     PRIVACY_URL: process.env.PRIVACY_URL,
     RATE_LIMITING_DISABLED: process.env.RATE_LIMITING_DISABLED,
+    TRUSTED_PROXY_HOP_COUNT: process.env.TRUSTED_PROXY_HOP_COUNT,
     TELEMETRY_DISABLED: process.env.TELEMETRY_DISABLED,
     S3_ACCESS_KEY: process.env.S3_ACCESS_KEY,
     S3_BUCKET_NAME: process.env.S3_BUCKET_NAME,
--- a/apps/web/lib/utils/client-ip.ts
+++ b/apps/web/lib/utils/client-ip.ts
@@ -1,5 +1,98 @@
 import { headers } from "next/headers";
 import { logger } from "@formbricks/logger";
+import { TRUSTED_PROXY_HOP_COUNT } from "@/lib/constants";
+
+/**
+ * Returned when no forwarding header may be believed, so the client is genuinely unidentifiable.
+ *
+ * Next 16 exposes no socket peer address to route handlers or server actions (`NextRequest.ip` was
+ * removed and `headers()` carries only HTTP headers), so with no trusted proxy there is nothing else to
+ * fall back to.
+ */
+export const UNTRUSTED_CLIENT_IP = "untrusted-client-ip";
+
+/**
+ * Throttle window for the misconfiguration warning. Time-based rather than once-per-process: a warning
+ * that fires only on the first request goes silent for the life of a long-running deployment, so an
+ * ongoing misconfiguration disappears from monitoring after one line. Raised by CodeRabbit on #8680.
+ */
+const UNTRUSTED_IP_WARNING_INTERVAL_MS = 10 * 60 * 1000;
+
+let lastUntrustedIpWarningAt: number | null = null;
+
+const warnAboutUntrustedIp = (): void => {
+  const now = Date.now();
+  if (
+    lastUntrustedIpWarningAt !== null &&
+    now - lastUntrustedIpWarningAt < UNTRUSTED_IP_WARNING_INTERVAL_MS
+  ) {
+    return;
+  }
+  lastUntrustedIpWarningAt = now;
+  logger.error(
+    "TRUSTED_PROXY_HOP_COUNT is set to 0 but the request carries forwarding headers. IP-based rate " +
+      "limiting and IP capture cannot identify individual clients while no hop is trusted. Unset it to " +
+      "take the default of 1, or set it to the number of reverse proxies actually in front of this app."
+  );
+};
+
+/**
+ * Resolves the client IP from forwarding headers, trusting only as many proxy hops as configured.
+ *
+ * `X-Forwarded-For` is *appended* to by each proxy, so its leftmost entry is whatever the client itself
+ * sent — reading that, as this used to, let a caller supply any value. Because the result keys the
+ * IP-based rate limits (login, forgot-password, signup, the public client API), a caller could rotate
+ * the header to mint a fresh bucket per request and bypass all of them, and could also forge the
+ * `ipAddress` recorded on responses and in audit-log entries.
+ *
+ * With `hopCount` proxies in front, the address the outermost trusted proxy observed is the
+ * `hopCount`-th entry from the right; everything left of it is client-supplied and ignored.
+ * `TRUSTED_PROXY_HOP_COUNT` defaults to 1, matching every supported topology; `0` is an explicit
+ * opt-out that trusts nothing and therefore cannot identify a client at all.
+ *
+ * `cf-connecting-ip` is deliberately *not* consulted. It is only trustworthy when the request provably
+ * came from Cloudflare's edge, and a hop count cannot establish that: `hopCount >= 1` says "one proxy is
+ * in front", which for most deployments is Traefik, Envoy or nginx — none of which strip
+ * `cf-connecting-ip`. Preferring it would therefore hand the spoof straight back to the caller. Nothing
+ * is lost by dropping it: Cloudflare also puts the visitor address into `X-Forwarded-For`, so a
+ * Cloudflare deployment is just `hopCount = 1` (or 2 with a proxy of its own behind it).
+ *
+ * Exported separately from {@link getClientIpFromHeaders} so the parsing is unit-testable without
+ * mocking `next/headers`.
+ */
+export const resolveClientIp = (headersList: Headers, hopCount: number): string => {
+  const xForwardedFor = headersList.get("x-forwarded-for");
+
+  if (hopCount <= 0) {
+    if (xForwardedFor || headersList.get("cf-connecting-ip") || headersList.get("x-real-ip")) {
+      warnAboutUntrustedIp();
+    }
+    return UNTRUSTED_CLIENT_IP;
+  }
+
+  if (xForwardedFor) {
+    const entries = xForwardedFor
+      .split(",")
+      .map((entry) => entry.trim())
+      .filter(Boolean);
+
+    if (entries.length > 0) {
+      // Clamped rather than falling back to the leftmost entry: a request with fewer hops than
+      // configured arrived through a shorter path than expected, and the earliest entry present is the
+      // closest thing to what a trusted proxy saw.
+      const index = Math.max(0, entries.length - hopCount);
+      return entries[index];
+    }
+  }
+
+  // No `x-real-ip` fallback. It is only reachable when no `X-Forwarded-For` arrived at all, and a proxy
+  // that is genuinely in front appends to XFF — so the absence of XFF means the request did not come
+  // through the trusted hop this app is configured for, and `x-real-ip` is then whatever the caller
+  // chose to send. Reading it would restore exactly the per-request bucket rotation this function
+  // exists to stop. This is the same argument the docstring makes against `cf-connecting-ip`; applying
+  // it to one header and not the other was inconsistent. Raised by @pandeymangg on #8680.
+  return UNTRUSTED_CLIENT_IP;
+};
 
 export async function getClientIpFromHeaders(): Promise<string> {
   let headersList: Headers;
@@ -7,16 +100,8 @@ export async function getClientIpFromHeaders(): Promise<string> {
     headersList = await headers();
   } catch (e) {
     logger.error(e, "Failed to get headers in getClientIpFromHeaders");
-    return "::1";
+    return UNTRUSTED_CLIENT_IP;
   }
 
-  // Try common proxy headers first
-  const cfConnectingIp = headersList.get("cf-connecting-ip");
-  if (cfConnectingIp) return cfConnectingIp;
-
-  const xForwardedFor = headersList.get("x-forwarded-for");
-  if (xForwardedFor) return xForwardedFor.split(",")[0].trim();
-
-  // Fallback (may be undefined or localhost in dev)
-  return headersList.get("x-real-ip") || "::1"; // NOSONAR - We want to fallback when the result is ""
+  return resolveClientIp(headersList, TRUSTED_PROXY_HOP_COUNT);
 }
--- a/apps/web/modules/auth/lib/auth.ts
+++ b/apps/web/modules/auth/lib/auth.ts
@@ -10,6 +10,7 @@ import { prisma } from "@formbricks/database";
 import { logger } from "@formbricks/logger";
 import type { TUserLocale } from "@formbricks/types/user";
 import {
+  EMAIL_AUTH_ENABLED,
   EMAIL_VERIFICATION_DISABLED,
   PASSWORD_RESET_TOKEN_LIFETIME_MINUTES,
   RATE_LIMITING_DISABLED,
@@ -90,7 +91,12 @@ export const auth = betterAuth({
   socialProviders: ssoSocialProviders,
 
   emailAndPassword: {
-    enabled: true,
+    // EMAIL_AUTH_DISABLED=1 has to switch the credential endpoints off here, not just hide the form on
+    // the login/signup pages (its only other use). Hardcoding `true` left
+    // POST /api/auth/sign-in/email live on an instance the operator had configured as SSO-only, so any
+    // account that still carried a password could sign in around the IdP — and around whatever the IdP
+    // enforces, such as MFA, conditional access, or deprovisioning.
+    enabled: EMAIL_AUTH_ENABLED,
     // Matches ZUserPassword (min 8 / max 128); the upper+digit composition rule stays enforced
     // by ZUserPassword at the app layer (deferred policy modernization → design doc §10.6).
     minPasswordLength: 8,
--- a/apps/web/modules/auth/signup/actions.test.ts
+++ b/apps/web/modules/auth/signup/actions.test.ts
@@ -1,6 +1,10 @@
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
-import { SIGNUP_EMAIL_DOMAIN_BLOCKED_ERROR_CODE } from "@formbricks/types/errors";
+import {
+  INVITE_TOKEN_INVALID_ERROR_CODE,
+  SIGNUP_EMAIL_DOMAIN_BLOCKED_ERROR_CODE,
+} from "@formbricks/types/errors";
 import { verifyInviteToken } from "@/lib/jwt";
+import { createMembership } from "@/lib/membership/service";
 import { getUserByEmail } from "@/lib/user/service";
 import { auth } from "@/modules/auth/lib/auth";
 import { getInvite, resolveInviteMatch } from "@/modules/auth/signup/lib/invite";
@@ -12,6 +16,7 @@ import { createUserAction } from "./actions";
 vi.mock("@formbricks/logger", () => ({
   logger: {
     error: vi.fn(),
+    warn: vi.fn(),
   },
 }));
 
@@ -89,7 +94,9 @@ describe("createUserAction — signup verification email callbackURL", () => {
     notificationSettings: { alert: {} },
   };
 
-  const baseInput = { name: "Ada", email: "Ada@Example.com", password: "Password123!" };
+  // Mirrors what signup-form.tsx actually posts: it always sends the field, as `inviteToken ?? ""`,
+  // so an uninvited sign-up arrives with an empty string rather than undefined.
+  const baseInput = { name: "Ada", email: "Ada@Example.com", password: "Password123!", inviteToken: "" };
 
   const newCtx = () => ({ auditLoggingCtx: { organizationId: "", userId: "" } });
 
@@ -115,6 +122,7 @@ describe("createUserAction — signup verification email callbackURL", () => {
   });
 
   test("does not point the verification callback at /invite for invite signups (ENG-1527)", async () => {
+    vi.mocked(resolveInviteMatch).mockResolvedValue("valid");
     vi.mocked(verifyInviteToken).mockReturnValue({ inviteId: "invite-1", email: "ada@example.com" } as never);
     vi.mocked(getInvite).mockResolvedValue({
       id: "invite-1",
@@ -141,6 +149,50 @@ describe("createUserAction — signup verification email callbackURL", () => {
     });
   });
 
+  // Regression: signup-form.tsx sends `inviteToken: inviteToken ?? ""`, so an uninvited sign-up arrives
+  // with an empty string. Guarding the "unusable invite" check on `!== undefined` rather than
+  // truthiness rejected every public sign-up with INVITE_TOKEN_INVALID_ERROR_CODE.
+  test.each(["", "   ", undefined])(
+    "treats %p as no invite and completes an ordinary sign-up",
+    async (inviteToken) => {
+      vi.mocked(resolveInviteMatch).mockResolvedValue("missing");
+
+      const result = await createUserAction({
+        ctx: newCtx(),
+        parsedInput: { ...baseInput, inviteToken },
+      } as never);
+
+      expect(result).toEqual({ success: true });
+      expect(auth.api.signUpEmail).toHaveBeenCalled();
+      // No invite means no membership grant — the org-creation path handles it instead.
+      expect(createMembership).not.toHaveBeenCalled();
+    }
+  );
+
+  // Regression: an invite binds a role in an organization to one address. Accepting it only on a valid
+  // signature meant a leaked or forwarded invite link could be redeemed with any address.
+  test.each(["email_mismatch", "invalid_or_expired", "verification_error", "missing"] as const)(
+    "refuses to grant membership when the invite resolves as %s",
+    async (inviteMatch) => {
+      vi.mocked(resolveInviteMatch).mockResolvedValue(inviteMatch);
+      vi.mocked(verifyInviteToken).mockReturnValue({
+        inviteId: "invite-1",
+        email: "someone-else@example.com",
+      } as never);
+
+      await expect(
+        createUserAction({
+          ctx: newCtx(),
+          parsedInput: { ...baseInput, inviteToken: "invite-jwt-123" },
+        } as never)
+        // The stable code, not just "some error": the sign-up form maps this to localized copy, so a
+        // regression to a raw message would break the UI while still passing a bare toThrow().
+      ).rejects.toThrow(INVITE_TOKEN_INVALID_ERROR_CODE);
+
+      expect(createMembership).not.toHaveBeenCalled();
+    }
+  );
+
   test("treats a duplicate email as already-existed without post-creation side effects", async () => {
     vi.mocked(auth.api.signUpEmail).mockRejectedValue(new Error("user already exists"));
     vi.mocked(getUserByEmail).mockResolvedValue(createdUser as never);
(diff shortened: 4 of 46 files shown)


mattinannt:

| 2 | **#8680 ← this PR** | 🟠 High | `claude/security-critical` | 10 |

Merge in order. **#8679 first** — this PR's base is `claude/security-critical`, so GitHub retargets it to `main` once that lands, and the diff shown here stays limited to the high-severity work.

---

mattinannt:
**SonarCloud quality gate was failing — fixed in `51f5223`** (force-pushed; #8681 and #8682 rebased onto it).

Everything else was green; the gate failed on one condition:

```
ERROR  new_duplicated_lines_density   actual=4.0%   threshold=3%
```

The cause was mine, and precise — `isSafeMediaUrl` was **100% duplicated**, 20 new lines in each of:

- `packages/surveys/src/lib/video-upload.ts`
- `packages/survey-ui/src/lib/video.ts`

I'd added the same predicate to both packages because both render survey media. That is also the drift risk CodeRabbit flagged on the closed #8667, where it noted the two `element-media.tsx` components had already diverged into near-copies.

**Fix:** `@formbricks/survey-ui` is now the single owner — it exports `isSafeMediaUrl` from its public entry point, and `packages/surveys` imports it there rather than keeping a copy. `packages/surveys` already depends on `@formbricks/survey-ui` and imports components from it, so this adds no new dependency edge.

Two alternatives I rejected:

- **Putting it in `@formbricks/types/common`, beside the `ZStorageUrl` schema it mirrors.** Conceptually the best home, and I built it that way first — but `packages/survey-ui` has *zero* runtime `@formbricks/*` dependencies today, only `config-typescript` and `eslint-config`. Adding one to share ten lines crosses a package boundary that looks deliberate, and that is not my call to make silently. Happy to move it there if you'd prefer the tighter cohesion — it's a one-line dependency addition.
- **Leaving both copies.** The code is correct either way; only the gate objects. But the duplication is genuinely new and genuinely a drift hazard on a security predicate, so it was worth fixing rather than waiving.

The tests moved with it: the `isSafeMediaUrl` suite now imports from `@formbricks/survey-ui` and stays in `packages/surveys/src/lib/video-upload.test.ts`, because `packages/survey-ui` has no test runner configured of its own. Noted in a comment there so it doesn't look misplaced.

Verified: `packages/surveys` 26 files / 670 ✅, `packages/survey-ui` 4 / 73 ✅, `apps/web` 533 / 6689 ✅, `tsc --noEmit` clean in all three packages, and the `surveys` UMD bundle rebuilt per AGENTS.md.

---

mattinannt:
@coderabbitai addressed in `09c2a4ed` — 6 of 7 fixed, 1 declined with reasoning. The two rated highest were both real; thanks.

**1. `element-media.tsx` — unvalidated `src` sinks (🔴 Critical). Fixed, and you were right that the anchor guard alone was incoherent.** `imgUrl` went straight to ``, and `getVideoUrlWithParams` returned the URL unchanged for anything not YouTube/Vimeo/Loom, so it reached `<iframe src>` — which does execute. Both sinks in both packages now use `isSafeMediaUrl`-validated values, and the component renders nothing when neither survives.

One correction to the finding: a `javascript:`/`data:` URL in `` does **not** execute — image loading doesn't run URL schemes, and SVG loaded via `` has scripting disabled. The iframe was the live sink. I fixed the img anyway, because the stated rationale for the anchor guard (stored rows predating `ZStorageUrl`) applies identically.

Worth knowing for context you couldn't have: #8681 replaces `getVideoUrlWithParams` so it returns `undefined` unless the URL is one of the three platforms, emitting only the normalized `convertToEmbedUrl` output. That's the stronger fix, but it lands one PR later — so this change closes the window where `main` would carry the unvalidated iframe.

**2. `workspace-access.ts` — duplicate `workspaceId` (🟠 Major). Fixed, though not exploitable.** Your reading of `Map` is right: `[{ws1, manage}, {ws1, read}]` collapses to `read`, compares equal to a stored `read` grant, and skips the owner/manager gate. But it can't be reached, because `updateTeamDetails` counts the submitted ids against a distinct `in` query:

```ts
const orgWorkspacesCount = await prisma.workspace.count({ where: { id: { in: workspaceIds }, ... } });
if (orgWorkspacesCount !== workspaceIds.length) throw new Error(...);
```

Any duplicate makes `workspaceIds.length` exceed the distinct count, so it throws before the upsert. It fails closed today — but that puts the outcome in a different module, so `hasWorkspaceAccessChanges` now returns `true` for any repeated id, with a regression test for exactly your case (last entry matching, earlier one escalated).

**3. `signup/actions.ts` — orphaned user if the invite changes mid-signup (🟠 Major). Declining, with reasons.**

- It's pre-existing, not introduced here. This PR changed *which* invites are accepted (binding to the invited address and checking expiry); the create-then-validate ordering is untouched.
- It isn't a security issue. The failure leaves a user row with **no membership** — no organization, no workspace, no access. That's an orphaned record, not a privilege. The "admitted through the domain exemption" concern doesn't hold either: the exemption governs whether sign-up is *allowed*, and without a membership the account still reaches nothing.
- The fix you describe is genuinely a heavy lift: `signUpUserSafely` goes through `auth.api.signUpEmail`, so making creation and invite consumption atomic means wrapping a Better Auth call in our transaction, and the compensating-delete alternative introduces a delete-user path on a failed signup — new blast radius on an auth flow, in a PR that already carries 10 security fixes.

Widening scope here trades a real regression risk for an orphaned row. It deserves its own ticket with the concurrency question handled properly — the same call was made on the non-atomic `user.count() === 0` fresh-instance check you raised on #8667, which is recorded in #8681's risks section.

**4–5. Test assertions (🟡 Minor). Both fixed** — the invalid-invite case asserts `INVITE_TOKEN_INVALID_ERROR_CODE`, and the rate-limit test asserts `emailToken`'s interval, quota and namespace rather than just the key.

**6. Colocate the `isSafeMediaUrl` suite (🟠 Major). Fixed** — moved to `packages/survey-ui/src/lib/video.test.ts`, importing from `./video`, all cases preserved. It was in `packages/surveys` only because of an earlier de-duplication step; `survey-ui` has its own runner, so there was no reason to leave it there.

**7. Warn-once flag (🔵 Trivial). Fixed** — 10-minute throttle window instead of once per process, so a standing `TRUSTED_PROXY_HOP_COUNT=0` keeps producing a signal. Test uses fake timers to assert it stays quiet at 9 minutes and fires again after 11.

On the two pre-merge warnings: **Docstring Coverage 0.00% appears to be a measurement error** — the changed files carry 33 JSDoc blocks (10 in `packages/types/errors.ts`, 6 in `survey/link/lib/data.ts`, 3 each in `workspace/auth.ts` and `redact-credentials.ts`, 2 each in `client-ip.ts`, `session.ts`, `auth.ts`). I'm not adding docstrings to chase the threshold; AGENTS.md explicitly discourages coverage-driven additions. On **screenshots**: the `.tsx` files here change what data is *passed* to client components (credential redaction) — the rendered output is byte-identical, so there's no before/after to show. Noting that here rather than restating it in the description.

Verification: 529 files / 6707 tests ✅ (`apps/web`), 84 ✅ (`survey-ui`), 663 ✅ (`surveys`), `pnpm build --filter=@formbricks/surveys... --force` clean, eslint clean. #8681 and #8682 rebased onto this tip.

---

xernobyl:
### Heads up: the invite-binding fix (ENG-2071) is also on `fix/ENG-2091_invite-signup-verification-email`

Not a review comment — a duplication warning, so we don't discover it at merge time.

**Why.** ENG-2091 (invite sign-up silently sends no verification email) is blocking users on Cloud right now, and its fix lands in the same function as this PR's invite-binding change — `createUserAction` / `handleInviteAcceptance` in `apps/web/modules/auth/signup/actions.ts`. Rather than ship ENG-2091 with a known auth hole sitting one line away, I took the ENG-2071 fix onto that branch too.

**What I did to keep this cheap for you.** I ported your fix rather than writing my own, close to verbatim:

- the `resolveInviteMatch(inviteToken, user.email)` gate in `handleInviteAcceptance`, same comment
- the normalized `const inviteToken = parsedInput.inviteToken?.trim() || undefined` hoist
- the pre-creation rejection so a bad token can't orphan an account
- `INVITE_TOKEN_INVALID_ERROR_CODE` in `packages/types/errors.ts`, same docstring
- the client mapping to `auth.invite.invite_not_found_description`
- your `test.each(["", "   ", undefined])` blank-token regression and the `email_mismatch` / `invalid_or_expired` / `verification_error` / `missing` cases

So the overlap should resolve to "the same change", not a real conflict. Your version is the reference; if the two differ anywhere, take yours.

**What is genuinely different on that branch**, and will still need reconciling:

1. `signUpUserSafely` returns a discriminated union (`{status:"created"|"already_existed"}`) instead of `{user, userAlreadyExisted}`, and `handlePostUserCreation` takes the `"created"` variant. That's the ENG-2091 fix — Better Auth answers a duplicate email with a synthetic 200 rather than throwing, so the old catch-based detection was dead and every post-creation side effect was running against the pre-existing account.
2. `createUserAction` returns a `nextStep` alongside `success`, so the form can route an invited user who already has an account to login instead of to a verification screen for an email that never arrives.
3. `emailVerification.sendVerificationEmail` in `auth.ts` now treats a falsy return as a failure and rethrows — `sendEmail` returns `false` without throwing when SMTP is unconfigured, and Better Auth ignores the return value. Same one-line gap in `sendResetPassword`.
4. `betterAuthLogger` redacts email addresses from Better Auth's log messages. `sign-up.mjs` logs `Sign-up attempt for existing email: <address>` at `info`; `level: "warn"` is currently the only thing suppressing it, and raising the level is exactly what you'd do while debugging a sign-up problem.

**One thing worth knowing for this PR specifically:** your `#8681` "closed sign-up" change also touches `createUserAction` and `signup-form.tsx`, so that's a three-way overlap on one function. Also note `autoSignIn: false` alone triggers Better Auth's synthetic-duplicate branch, so the ENG-2091 defect reaches self-hosters with `EMAIL_VERIFICATION_DISABLED=1` too — worth a thought for how far the stack gets backported.

Happy to rebase mine behind the stack instead and drop the duplicated hunks if you'd rather keep ENG-2071 solely here — say the word. Otherwise, whichever merges second should take the other's invite hunks as-is.

mattinannt:
Thanks for flagging this before merge time rather than after — and for porting rather than rewriting.

**My answer: drop the ported ENG-2071 hunks and ship ENG-2091 standalone, ahead of this stack. I'll take the reconciliation.** Reasoning below, including one thing you couldn't have known about.

### The invite hunk isn't stable across the stack

This is the part that changes the calculus. Your inventory is accurate — every item you listed comes from #8680, nothing from #8681 — but the pre-creation rejection **moves** in #8681:

- #8680 adds it inline in `createUserAction`.
- #8681 extracts it into `assertSignupPolicyAllows(inviteToken, inviteMatch)`, wrapping it with the `SIGNUP_ENABLED && multi-org`/fresh-instance gate, to keep `createUserAction` inside the cognitive-complexity budget.

So "whichever merges second takes the other's invite hunks as-is" is ambiguous — the hunk has two different shapes depending on which of my PRs you compare against. A verbatim copy of the #8680 form lands on a line #8681 then deletes, and the failure mode isn't a clean conflict: #8681's diff could apply and leave the check duplicated, once inline and once inside the helper. That's a second reconciliation, not a saved one.

### Shipping ENG-2091 without the fix isn't a regression

The invite hole is on `main` right now. ENG-2091 landing without ENG-2071 leaves `main` exactly as exposed as it is today — the port isn't buying safety that's currently present. That removes the reason to carry a duplicate of a security hunk in a second branch, which is the shape where one copy silently wins a merge and nobody re-reviews the result. ENG-2091 is blocking Cloud users; it shouldn't wait behind a four-PR security stack for a hole it doesn't widen.

### What I'll do

Your point 1 rewrites the exact line my change sits on — `handlePostUserCreation(ctx, user, inviteToken)` is inside `if (!userAlreadyExisted && user)`, which is the `{user, userAlreadyExisted}` shape your discriminated union replaces. Small and mechanical: once ENG-2091 is on `main` I'll rebase #8680 onto it and convert that call site to the `status === "created"` variant, then let #8681/#8682 follow up the stack. No need for you to hold anything.

### Two things that came out clean

- **`signup-form.tsx` is safe to port.** #8680 adds the `INVITE_TOKEN_INVALID_ERROR_CODE` branch; #8681 only adds a comment above the existing `else`. The branch itself is identical in both, so if you do keep any hunk, this is the one that won't fight anything.
- **`auth.ts` is a three-way overlap on disjoint regions.** #8680 changes `emailAndPassword.enabled` → `EMAIL_AUTH_ENABLED`; #8681 adds `WEBAPP_URL` to the `USE_SECURE_COOKIES` chain; yours touches `emailVerification.sendVerificationEmail`/`sendResetPassword`. Three separate parts of one file — that should merge without conflict, though it's worth a look after the fact rather than trusting me on it.

### On your last two points

Your point 4 is a real one and worth doing — `Sign-up attempt for existing email: <address>` at `info` means the only thing standing between a support-debugging session and addresses in logs is the log level. Yours to own; I'm not touching it.

The `autoSignIn: false` / `EMAIL_VERIFICATION_DISABLED=1` observation is the more consequential one for release planning: it puts ENG-2091 in front of self-hosters, not just Cloud. #8679 (Critical) is already backported to `release/5.2` as #8686. Whether the High stack and ENG-2091 follow it there is @mattinannt's call, not mine — flagging rather than assuming.

If you'd rather keep the duplicate route after all, push the branch (draft PR is fine) and I'll diff your hunks against mine and confirm equivalence properly — I can't do that today, since `fix/ENG-2091_invite-signup-verification-email` isn't on the remote and there's no PR for ENG-2091 yet, so "close to verbatim" is the one claim here I've taken on trust rather than checked.

---

apps/web/modules/auth/signup/actions.test.ts
pandeymangg, line 1: this file should be removed as we don't write action tests

apps/web/modules/auth/lib/auth.ts
pandeymangg, line 1: AI Review:

Better Auth runs its own rate limiter holding the real brute-force limits (/sign-in/email 5/min, /two-factor/* 5/min), and it gives up on working out the visitor's address whenever X-Forwarded-For has more than one entry, because trustedProxies is unset (@better-auth/core/src/utils/ip.ts:329). When it gives up, every request falls into one shared counter (rate-limiter/index.mjs:287) — so a CDN-plus-ingress deployment gets 5 logins per minute for the whole instance, and even behind a single proxy any visitor can force that shared counter by sending the header themselves. TRUSTED_PROXY_HOP_COUNT can't be reused here (Better Auth wants a list of proxy addresses, not a count), so this needs advanced.ipAddress.trustedProxies set to the CDN's egress ranges — or the proxy overwriting X-Forwarded-For instead of appending to it, so there's only ever one entry.

apps/web/lib/env.ts
```diff
@@ -315,6 +315,10 @@ const parsedEnv = createEnv({
       .optional()
       .or(z.string().refine((str) => str === "")),
     RATE_LIMITING_DISABLED: z.enum(["1", "0"]).optional(),
+    // Number of reverse proxies in front of the app whose X-Forwarded-For entries can be believed.
+    // Unset falls back to 1 (see TRUSTED_PROXY_HOP_COUNT in lib/constants.ts); an explicit 0 trusts no
+    // forwarding header at all. See resolveClientIp in lib/utils/client-ip.ts.
+    TRUSTED_PROXY_HOP_COUNT: z.coerce.number().int().min(0).max(10).optional(),
```
pandeymangg: `z.coerce.number()` turns `""` into `0`, which passes `.min(0)` — unlike every other numeric var here (`BULLMQ_WORKER_CONCURRENCY` `.min(1)` L231, `PASSWORD_RESET_TOKEN_LIFETIME_MINUTES` `.min(5)` L312), which fail loudly. At 0 every request collapses to one rate-limit bucket, so 10 failed logins lock out the instance. Suggest `z.preprocess((v) => (v === "" ? undefined : v), …)`.

apps/web/lib/utils/client-ip.ts
```diff
@@ -1,22 +1,104 @@
...
+    if (entries.length > 0) {
+      // Clamped rather than falling back to the leftmost entry: a request with fewer hops than
+      // configured arrived through a shorter path than expected, and the earliest entry present is the
+      // closest thing to what a trusted proxy saw.
+      const index = Math.max(0, entries.length - hopCount);
+      return entries[index];
+    }
+  }
+
+  const xRealIp = headersList.get("x-real-ip")?.trim();
```
pandeymangg: This is only reachable when no XFF arrived, i.e. when no proxy is actually in front — exactly when `X-Real-IP` is caller-controlled. On the default `hopCount=1` with a directly reachable app, rotating that header gives a fresh bucket per request. The docstring's own argument against `cf-connecting-ip` ("a hop count cannot establish provenance") applies here too.

pandeymangg (changes requested):
looks good, please take a look at my comments 🙏 
Also, this is from AI review but I think its pretty real:

The client sends `config.key` straight back on save (`AddIntegrationModal.tsx:197-203` → `actions.ts:45` → `service.ts:44` does `update: { ...integrationData }`), so the blanked `access_token`/`refresh_token`
overwrite the real ones — and `z.string()` with no `.min(1)` means it validates silently. It fires on add, edit **and** delete of any mapping (`deleteLink()` reuses the same action), and the UI still shows "connected" because the wrappers test `config?.key` as an object. Fix server-side: drop `key` from `ZIntegrationInput` and have `createOrUpdateIntegration` preserve the stored value.

*(Same applies to the `google-sheets`, `notion` and `slack` pages.)*

mattinannt:
Thanks — the integration one is a genuine regression I introduced, and I'd have shipped it. Verified all five against the code rather than taking them on trust; three are fixed, one I'm pushing back on, one is real but I don't think belongs in this PR. New tips: **high `331479b5`** · medium `5e8cd08a` · low `191523d5`.

### 1. Redacted `config.key` overwriting the stored tokens — confirmed, fixed

Traced it end to end and it's exactly as described. `AddIntegrationModal` builds `config: { key: notionIntegration?.config?.key, … }` from the prop this PR now redacts, then `createOrUpdateIntegration` does `update: { ...integrationData }` — a full overwrite. It fires on add, edit **and** `deleteLink()`, and all of Notion, Google Sheets and Airtable echo the key back the same way.

I did **not** take the suggested "drop `key` from `ZIntegrationInput`" route, because the OAuth callbacks legitimately write credentials through that same schema (`notion/callback/route.ts` sets `config.key = tokenData` → `createOrUpdateIntegration`), so removing it would break connecting. What makes the fix clean is that the two paths already differ at the boundary: **the modal goes through the server action, the callbacks call the service directly.** So the action is where client input gets neutralised — it now reads the stored integration and substitutes the stored `key`, via a `withStoredIntegrationKey` helper paired with `redactIntegrationCredentials` (outbound redact / inbound preserve, one file). Credentials are never taken from the request on that path.

Mutation-tested: making the helper a no-op fails 2 of the new tests; restoring it passes 17.

### 3. `z.coerce.number()` turning `""` into `0` — confirmed, fixed

Your reasoning is exactly right, including why the neighbours don't have the problem: `.min(5)`/`.min(1)` reject the coerced `0` loudly, but `0` is a *legitimate* value here (the explicit trust-nothing opt-out), so `.min(0)` waves it through. An empty-string render — the ordinary docker-compose/Helm shape for an unset optional — would silently opt the instance out of trusting any hop. Now `z.preprocess((v) => (v === "" ? undefined : v), …)` as you suggested.

### 4. `x-real-ip` fallback — confirmed, fixed

Agreed, and the inconsistency was mine: that fallback is only reachable when no XFF arrived, a real proxy in front always appends to XFF, so the only way to reach it is a request that didn't come through the trusted hop — where the header is caller-controlled. My own docstring argues precisely this against `cf-connecting-ip`. Fallback removed; the existing test that asserted the old behaviour is flipped, plus one pinning that a present XFF still wins.

### 2. `actions.test.ts` — not deleting, and I'd like your call

Happy to follow the convention, but I don't think this PR is the place, for three reasons:

- **The file isn't mine.** It pre-exists on `main` at 247 lines; my diff only adds cases to it. Deleting it here removes existing coverage under cover of a security PR.
- **There's a lot of precedent** — 10 `actions.test.ts` files in `apps/web` today, including `integrations/google-sheets/actions.test.ts`. If the convention is now "none", that's a sweep of its own.
- **#8691 is actively rewriting this exact file.** xernobyl replaced its duplicate-signup mock with the real Better Auth contract. Deleting it here would collide with that.

The cases I added are the mutation-checked regressions for the invite-binding fix (blank-token normalization, `email_mismatch` / `invalid_or_expired` / `verification_error` / `missing`). If you want them out of an action test, say where they should live and I'll move them — otherwise I'd rather leave the file for a dedicated cleanup.

### 5. Better Auth's own rate limiter — verified, real, and worse than stated; proposing a follow-up

I checked this against the installed **1.6.23**, not the description, and it holds. `getIPFromHeader` does `if (forwardedIps.length !== 1) return null` whenever `trustedProxies` is unset — which is us, we set only `ipAddressHeaders`. `resolveRateLimitConfig` then does `createRateLimitKey(ip ?? NO_TRUSTED_IP_KEY, path)`, so everyone shares `no-trusted-ip|/sign-in/email` at our configured 5/min.

One correction to the framing: it isn't only CDN-plus-ingress. A proxy *appends*, so **behind a single proxy any visitor can force the shared bucket by sending one XFF entry themselves** — two entries, resolution fails, instance-wide login lockout at 5/min. That's an unauthenticated DoS on sign-in.

Two things make me want it separate rather than folded in here:

- **It's pre-existing, not from this PR.** My `auth.ts` diff is only `emailAndPassword.enabled`; the `rateLimit`/`ipAddress` block is already on `main`. This PR fixes *our* limiter's IP resolution (`applyIPRateLimit`), which is a different subsystem.
- **The fix needs a design call, not a line.** `trustedProxies` wants CIDRs, so it can't reuse `TRUSTED_PROXY_HOP_COUNT` — either a new env var operators must populate with their proxy ranges, or middleware that overwrites a dedicated header with our already-correct `resolveClientIp` output and points `ipAddressHeaders` at it. The second reuses the hop-count model and needs no operator knowledge, but it's request plumbing I'd rather not bolt onto a 45-file security PR at review time.

Better Auth does log a one-time warning when it can't resolve an IP, so it isn't silent. Happy to take it as the next ticket and implement the middleware approach — or if you'd rather it block this PR, say so and I'll do it here.

Full suite green on the stack tip after all of the above: **534 files / 6737 tests**, eslint and prettier clean, and `tsc --noEmit` reports nothing in any touched file (the remaining errors are the pre-existing `@/images/*` asset declarations and one Prisma `archivedAt`, both from not running `typegen` first).

---

pandeymangg (approved):
LGTM

PR #8585, opened 2026-07-20
fix(surveys): return structured 422 when AI generation hits the output token limit (ENG-1891)

itsjavi:
## What does this PR do?

Long or highly structured prompts in the AI survey editor made `POST /api/v3/surveys/generate` fail with an opaque 502 and no actionable feedback. The generation budget was `maxOutputTokens: 3000`; Gemini 2.5 Flash spends most of that on reasoning tokens (observed: 2,664 reasoning, 322 left for the survey JSON), so generation stops with finish reason `"length"`. In AI SDK v6, structured output is only parsed when the finish reason is `"stop"`, so `result.output` threw a bare `NoOutputGeneratedError` that fell through to the route's 502 fall-through.

This PR:

- **Detects output-token exhaustion at the source**: `generateObject` in `@formbricks/ai` checks `finishReason === "length"` before accessing `result.output` and throws a new exported `AIOutputTokenLimitError` carrying plain token counts (max/output/reasoning) for logging — AI SDK internals stay encapsulated in the package.
- **Returns a structured 422** instead of a 502: the generate route maps the error to `problemUnprocessableContent` with `code: "ai_output_too_long"` and a detail asking the caller to simplify or split the prompt.
- **Shows a clear message in the AI survey editor**: the create-with-AI hook maps `ai_output_too_long` to a new translated string — "This prompt asks for more than the AI can generate in one go. Simplify it or split it into smaller surveys." (en-US hand-edited; other locales generated via `pnpm i18n`.)
- **Raises the generation budget** from 3,000 to 8,192 output tokens (`V3_SURVEY_GENERATION_MAX_OUTPUT_TOKENS`), leaving headroom for reasoning + JSON. A control run of the failing prompt with a 7,000-token limit succeeded in 19.7s, well within the existing 45s timeout.
- **Keeps the v3 API contract in sync**: the OpenAPI reference (`docs/api-v3-reference`) documents the new 422 example and adds `ai_output_too_long` to the shared `Problem.code` enum; bundle regenerated with `pnpm api:v3:bundle`.

Other `generateObject` callers (EE chart query, AI translation, summary) now receive the more informative error on token exhaustion instead of the opaque one; their behavior is otherwise unchanged.

Fixes ENG-1891

## How should this be tested?

Real token exhaustion needs a configured AI provider, so the definitive check is on staging:

1. Open **Create with AI** and paste the long music-distribution-form prompt from the ENG-1891 thread (or any prompt requesting many pages/questions with conditional branching).
[...]

PR #8681, opened 2026-07-29
fix(security): close the medium-severity bypass and disclosure gaps [3/4 medium]

mattinannt:
## What & why

**3 of 4** from the pre-release security review, split out of #8667 by severity. **Severity: Medium** — 13 findings: scoped bypasses, cross-tenant integrity, and disclosure that needs a valid account or a narrower precondition than the High set.

> ⚠️ **Stacked on #8680.** Base is `claude/security-high`, so this PR's diff shows only the medium-severity work. Merge #8679 and #8680 first.

### Verification bypasses on the public response endpoints

- **Forged `verifiedEmail`.** `isVerifyEmailEnabled` was enforced only in the page renderer, so posting straight to `/api/v{1,2}/client/{workspaceId}/responses` let a caller submit with any address they chose — the survey then displays it as a verified identity. Now enforced server-side on both versions through one shared `enforceVerifiedEmailGate`, with the address taken from the **verified token** rather than the request body, and the PUT handler stripping a client-supplied value (otherwise you could submit clean and patch it in). Same class of defect already fixed for PIN surveys in ENG-1579 / #8506. The token is read from `meta.url`, the channel v2 already trusts for `suId`/`suToken`, so no widget or bundle change was needed and older embedded snippets keep working.
- **reCAPTCHA only on v2.** Posting to the v1 URL opted out of a survey's spam protection entirely, while the setting still read as "on" in the UI. The client already sent `recaptchaToken` there — `ZResponseInput` was silently stripping it. Both endpoints now share one implementation, which is what stops them drifting apart again.

### Cross-tenant scoping and integrity

- **`displayId` connected without an ownership check** in v2 management response-create. `Display`↔`Response` is one-to-one, so naming another tenant's display *moved* it onto the attacker's response and corrupted the victim's display and completion counts.
- **Nested cross-workspace segments.** `buildSegmentFilterWhereClause` resolved a nested `segmentId` through `getSegment`, which looks up by id alone, while the id comes from a caller-authored filter tree. Another workspace's segment could be nested and its filters evaluated against contacts the caller controls — seeding candidate attribute values and reading back membership reveals the other team's targeting rules.
- **`getResponseCountAction`** authorized its `workspaceTeam` item against a caller-supplied `workspaceId` while operating on a `surveyId`, so the two arguments could disagree. The workspace is now derived from the survey.
- **`getSurveysForUnifyAction`** listed `member` among its organization roles, which made the `workspaceTeam` check below it dead for every member.

### Policy enforcement and account safety

- **Closed sign-up was a page-render check.** `createUserAction` enforced nothing, so anyone could POST the server action and create an account on a self-hosted instance the operator had deliberately closed. The action now mirrors the page, keeping both legitimate exceptions: an invited user, and the first administrator during fresh-instance setup.
- **2FA re-enrollment stripped the second factor with only a password.** `setupTwoFactorAuth` wrote `twoFactorEnabled: false` and replaced the secret and backup codes, reaching the same end state `disableTwoFactorAuth` requires a current TOTP code for. Anyone holding a stolen password could remove 2FA without ever satisfying it.
- **Session cookie lost `Secure`** on deployments setting only `WEBAPP_URL=https://…` — the primary documented variable — because the flag was derived from `BETTER_AUTH_URL ?? NEXTAUTH_URL`, both optional, falling through to `""`.

### Injection and enumeration

- **Respondent-controlled markup in follow-up emails.** Recalled values were spliced into the body *before* `sanitize-html` ran. The sanitizer sees only the **combined** string, so it cannot tell the author's intended markup from markup a respondent injected — an allowlist that legitimately permits `<a href>` in an author-written body passes an anchor spliced in from an open-text answer identically. An anonymous respondent could put an arbitrary clickable link, with text of their choosing, into a Formbricks-branded email delivered to the survey owner. Recalled values are data and are now HTML-escaped as they are substituted, via a flag used only by the email path (the React callers escape for themselves; escaping twice would surface literal `&amp;`).
- **`singleUseIds` gated at `read` but it mints, not reads.** Every link it returns carries a fresh HMAC-signed `suId`/`suToken` pair the unauthenticated response endpoint accepts, so a reporting-only key — the kind you hand a BI tool or a partner — could generate valid submission links in bulk and inject responses into a one-response-per-invitee survey. Now gated at write level despite being a GET.
- **Contact enumeration on the public client endpoint.** `POST /api/v1/client/{workspaceId}/user` is unauthenticated and `workspaceId` is public — it ships in the widget snippet on the customer's own site. The response echoed "the email already exists for this environment", making the endpoint answer *"is this address a contact of this organization?"* for any address an attacker chose.
- **Arbitrary URL in `<iframe src>`.** `getVideoUrlWithParams` fell through to returning the raw URL for anything that was not YouTube, Vimeo or Loom — enough to frame a phishing page inside a survey on a customer's site. It now emits only the normalized embed URL built from a hardcoded origin.

## Linear ticket

[ENG-1913](https://linear.app/formbricks/issue/ENG-1913) (forged `verifiedEmail`, reported by Yunseong Ryu / @ryuyunseong) is In Progress. The `displayId` half of [ENG-1923](https://linear.app/formbricks/issue/ENG-1923) is fixed here; the `contactId` half remains with @tiago. The contact-enumeration finding relates to the canceled [ENG-1762](https://linear.app/formbricks/issue/ENG-1762) — worth reopening, its cancel rationale addressed only the attribute-overwrite half.

## How this was tested

- `pnpm test` (`apps/web`) — 532 files / 6670 tests ✅; `packages/surveys` 26/670 ✅; `packages/survey-ui` 4/73 ✅.
- `eslint` and `tsc --noEmit` clean for every touched file.
- Regression tests: the email-verification gate (`verify-email-gate.test.ts`), recall escaping incl. record-shaped matrix/address answers (`recall.test.ts`), cross-workspace nested segments (`filter/prisma-query.test.ts`), display scoping (`responses/lib/tests/response.test.ts`), 2FA re-enrollment (`two-factor-auth.test.ts`), closed-instance sign-up policy (`signup/actions.test.ts`).

## QA / Test Plan

**How to test**

- [ ] Link survey with email verification on: `POST /api/v1/client/<workspaceId>/responses` with `data.verifiedEmail` set to an arbitrary address and no verify token in `meta.url` → 403. Same against `/api/v2/client/...`. A genuine widget submission through the emailed link still records the address.
- [ ] Submit normally, then `PUT` the response with `data.verifiedEmail` changed → the stored value does not change.
- [ ] reCAPTCHA-enabled survey: widget submission succeeds; `POST /api/v1/client/<workspaceId>/responses` without `recaptchaToken` → 400 `recaptcha_verification_failed` (previously accepted).
- [ ] PIN-protected and single-use link surveys still submit normally — both share the changed handlers.
- [ ] Follow-up email: recall an open-text answer, submit `<a href="https://evil.tld">click me</a>` → the received email shows that text literally, not a clickable link, and the author's own formatting still renders.
- [ ] Self-hosted with public sign-up closed: `/auth/signup` without an invite still 404s, and POSTing the sign-up action directly is rejected. Fresh-instance setup still creates the first administrator; invited sign-up still works.
- [ ] 2FA: with 2FA enabled, starting setup again → "already enabled". Disable (password + code) then re-enroll → works.
- [ ] Cookie flags: on an instance with only `WEBAPP_URL=https://…` set, the session cookie now carries `Secure`; sign-in over `http://localhost` in dev still persists.
- [ ] Segments: a segment nesting one from the **same** workspace still evaluates its filters; nesting one from a different workspace contributes no conditions.
- [ ] Single-use links: with a `read` key → 401; with `write`/`manage` → still returns links, and those links still work for submission.
- [ ] Contact identify: `attributes.email` colliding with another contact → the attribute is still not overwritten, but the response no longer contains "already exists". The widget still identifies users and applies other attributes; genuine `errors` (e.g. an invalid attribute key) still appear.
- [ ] Survey editor: a YouTube/Vimeo/Loom video still renders; a non-platform URL renders no iframe.

**Preconditions**

- Two organizations with separate workspaces, surveys, segments and API keys; keys at `read`/`write`/`manage`; an Enterprise license for contacts/segments and Unify Feedback.

**Risks & regressions**

- **`read`-scoped API keys can no longer fetch single-use links.** Any integration doing that will start getting 401 and needs its key upgraded to `write`. Worth a changelog entry.
- The contact-identify response no longer carries the two "already exists" messages. Behaviour is unchanged (the attribute was never overwritten) and the SDK only debug-logged them, but anything parsing those strings needs updating.
- The email-verification gate reads the token from `meta.url`. If an embedding integration strips or rewrites that, submissions to email-gated surveys would start failing — worth one pass over an embedded link survey.
- The fresh-instance sign-up exception is checked with a non-atomic `user.count() === 0`, so two concurrent uninvited sign-ups on a brand-new closed instance could both pass. Pre-existing — `signup/page.tsx` has always done the same — and far narrower than the unbounded exposure this closes, but worth its own fix (a unique bootstrap marker rather than a lock). Raised by CodeRabbit on #8667.

- No DB migrations.
- New error code `signup_disabled`; no locale changes (it is not reachable from the UI — the sign-up page 404s unless a valid invite is present, and a valid invite returns before that check).

## Merge order

1. #8679 (Critical)
2. #8680 (High)
3. **#PENDING_MEDIUM — this PR** (Medium)
4. #PENDING_LOW (Low)

---

### DIFF
--- a/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/put-response-handler.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/responses/[responseId]/lib/put-response-handler.ts
@@ -14,6 +14,7 @@ import { validateOtherOptionLengthForMultipleChoice } from "@/modules/api/v2/lib
 import { createQuotaFullObject } from "@/modules/ee/quotas/lib/helpers";
 import { validateClientFileUploads } from "@/modules/storage/utils";
 import { verifyLinkSurveyPinToken } from "@/modules/survey/link/lib/pin-token";
+import { VERIFIED_EMAIL_RESPONSE_KEY } from "@/modules/survey/link/lib/verify-email-gate";
 import { updateResponseWithQuotaEvaluation } from "./response";
 import { getValidatedResponseUpdateInput } from "./validated-response-update-input";
 
@@ -221,6 +222,13 @@ export const putResponseHandler = async ({
         surveyId: survey.id,
       }),
     };
+  }
+
+  // The update payload carries no `meta`, so there is no token to re-verify here. The address was
+  // established authoritatively from the verification token when the response was created, so drop any
+  // client-supplied value rather than letting an update overwrite it with an arbitrary address.
+  if (survey.isVerifyEmailEnabled && responseUpdateInput.data) {
+    delete responseUpdateInput.data[VERIFIED_EMAIL_RESPONSE_KEY];
   }
 
   const validationResult = validateUpdateRequest(existingResponse, survey, responseUpdateInput, workspaceId);
--- a/apps/web/app/api/v1/client/[workspaceId]/responses/route.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/responses/route.ts
@@ -15,10 +15,12 @@ import { getClientIpFromHeaders } from "@/lib/utils/client-ip";
 import { getOrganizationIdFromWorkspaceId } from "@/lib/utils/helper";
 import { resolveClientApiIds } from "@/lib/utils/resolve-client-id";
 import { formatValidationErrorsForV1Api, validateResponseData } from "@/modules/api/lib/validation";
+import { verifyResponseRecaptcha } from "@/modules/api/lib/verify-response-recaptcha";
 import { getIsContactsEnabled } from "@/modules/ee/license-check/lib/utils";
 import { createQuotaFullObject } from "@/modules/ee/quotas/lib/helpers";
 import { validateClientFileUploads } from "@/modules/storage/utils";
 import { verifyLinkSurveyPinToken } from "@/modules/survey/link/lib/pin-token";
+import { enforceVerifiedEmailGate } from "@/modules/survey/link/lib/verify-email-gate";
 import { createResponseWithQuotaEvaluation } from "./lib/response";
 
 export const OPTIONS = async (): Promise<Response> => {
@@ -150,6 +152,29 @@ export const POST = withV1ApiWrapper({
       return {
         response: responses.forbiddenResponse("Survey is protected by a PIN", true, { surveyId: survey.id }),
       };
+    }
+
+    // Email verification, like the PIN above, has to be enforced here and not only in the renderer:
+    // this endpoint is public, so a caller could otherwise submit with any `verifiedEmail` they like.
+    // Shared with the v2 endpoint so the two versions cannot drift apart.
+    const verifiedEmailErrorResponse = enforceVerifiedEmailGate({
+      survey,
+      responseData: responseInputData.data,
+      metaUrl: responseInputData.meta?.url,
+    });
+    if (verifiedEmailErrorResponse) {
+      return { response: verifiedEmailErrorResponse };
+    }
+
+    // Same gate the v2 endpoint applies — without it, posting to the v1 URL opts the caller out of the
+    // survey's spam protection entirely.
+    const recaptchaErrorResponse = await verifyResponseRecaptcha({
+      survey,
+      workspaceId,
+      recaptchaToken: responseInputData.recaptchaToken,
+    });
+    if (recaptchaErrorResponse) {
+      return { response: recaptchaErrorResponse };
     }
 
     const singleUseValidationResult = validateSingleUseResponseInput(survey, workspaceId, responseInputData);
--- a/apps/web/app/api/v1/management/surveys/[surveyId]/singleUseIds/route.ts
+++ b/apps/web/app/api/v1/management/surveys/[surveyId]/singleUseIds/route.ts
@@ -24,7 +24,12 @@ export const GET = withV1ApiWrapper({
           response: responses.notFoundResponse("Survey", params.surveyId),
         };
       }
-      if (!hasPermission(authentication.workspacePermissions, survey.workspaceId, "GET")) {
+      // Checked at write level despite being a GET: this endpoint *mints* credentials rather than
+      // reading anything. Each returned link carries a fresh HMAC-signed suId/suToken pair that the
+      // unauthenticated POST /api/v1/client/{workspaceId}/responses accepts, so at "read" a
+      // reporting-only key — the level you would hand an external analyst or BI tool — could generate
+      // thousands of valid submission links and inject responses with them.
+      if (!hasPermission(authentication.workspacePermissions, survey.workspaceId, "POST")) {
         return {
           response: responses.unauthorizedResponse(),
         };
--- a/apps/web/app/api/v2/client/[workspaceId]/responses/lib/utils.test.ts
+++ b/apps/web/app/api/v2/client/[workspaceId]/responses/lib/utils.test.ts
@@ -45,6 +45,7 @@ vi.mock("@/lib/utils/helper", () => ({
 vi.mock("@formbricks/logger", () => ({
   logger: {
     error: vi.fn(),
+    warn: vi.fn(),
   },
 }));
 
@@ -171,7 +172,8 @@ describe("checkSurveyValidity", () => {
     const result = await checkSurveyValidity(survey, "ws-1", responseInputWithoutToken);
     expect(result).toBeInstanceOf(Response);
     expect(result?.status).toBe(400);
-    expect(logger.error).toHaveBeenCalledWith("Missing recaptcha token");
+    // warn, not error: a missing token on a public endpoint is caller-controlled, not a fault.
+    expect(logger.warn).toHaveBeenCalledWith("Missing recaptcha token");
     expect(responses.badRequestResponse).toHaveBeenCalledWith(
       "Missing recaptcha token",
       { code: "recaptcha_verification_failed" },
@@ -190,7 +192,8 @@ describe("checkSurveyValidity", () => {
     });
     expect(result).toBeInstanceOf(Response);
     expect(result?.status).toBe(404);
-    expect(responses.notFoundResponse).toHaveBeenCalledWith("Organization", null);
+    // cors: true, matching every other response this path returns.
+    expect(responses.notFoundResponse).toHaveBeenCalledWith("Organization", null, true);
     expect(getOrganizationBillingByWorkspaceId).toHaveBeenCalledWith("ws-1");
   });
 
@@ -204,7 +207,7 @@ describe("checkSurveyValidity", () => {
       recaptchaToken: "test-token",
     });
     expect(result).toBeNull();
-    expect(logger.error).toHaveBeenCalledWith("Spam protection is not enabled for this organization");
+    expect(logger.warn).toHaveBeenCalledWith("Spam protection is not enabled for this organization");
   });
 
   test("should return badRequestResponse if recaptcha verification fails", async () => {
--- a/apps/web/app/api/v2/client/[workspaceId]/responses/lib/utils.ts
+++ b/apps/web/app/api/v2/client/[workspaceId]/responses/lib/utils.ts
@@ -1,17 +1,15 @@
 import { logger } from "@formbricks/logger";
 import { TSurvey } from "@formbricks/types/surveys/types";
-import { getOrganizationBillingByWorkspaceId } from "@/app/api/v2/client/[workspaceId]/responses/lib/organization";
-import { verifyRecaptchaToken } from "@/app/api/v2/client/[workspaceId]/responses/lib/recaptcha";
 import { TResponseInputV2 } from "@/app/api/v2/client/[workspaceId]/responses/types/response";
 import { responses } from "@/app/lib/api/response";
 import { ENCRYPTION_KEY } from "@/lib/constants";
 import { symmetricDecrypt } from "@/lib/crypto";
-import { getOrganizationIdFromWorkspaceId } from "@/lib/utils/helper";
 import { validateSurveySingleUseLinkParams } from "@/lib/utils/single-use-surveys";
-import { getIsSpamProtectionEnabled } from "@/modules/ee/license-check/lib/utils";
+import { verifyResponseRecaptcha } from "@/modules/api/lib/verify-response-recaptcha";
 import { verifyLinkSurveyPinToken } from "@/modules/survey/link/lib/pin-token";
+import { enforceVerifiedEmailGate } from "@/modules/survey/link/lib/verify-email-gate";
 
-export const RECAPTCHA_VERIFICATION_ERROR_CODE = "recaptcha_verification_failed";
+export { RECAPTCHA_VERIFICATION_ERROR_CODE } from "@/modules/api/lib/verify-response-recaptcha";
 
 export const checkSurveyValidity = async (
   survey: TSurvey,
@@ -95,41 +93,22 @@ export const checkSurveyValidity = async (
     responseInput.singleUseId = canonicalSingleUseId;
   }
 
-  if (survey.recaptcha?.enabled) {
-    if (!responseInput.recaptchaToken) {
-      logger.error("Missing recaptcha token");
-      return responses.badRequestResponse(
-        "Missing recaptcha token",
-        {
-          code: RECAPTCHA_VERIFICATION_ERROR_CODE,
-        },
-        true
-      );
-    }
-    const billing = await getOrganizationBillingByWorkspaceId(workspaceId);
-
-    if (!billing) {
-      return responses.notFoundResponse("Organization", null);
-    }
-
-    const organizationId = await getOrganizationIdFromWorkspaceId(workspaceId);
-    const isSpamProtectionEnabled = await getIsSpamProtectionEnabled(organizationId);
-
-    if (!isSpamProtectionEnabled) {
-      logger.error("Spam protection is not enabled for this organization");
-    }
-
-    const isPassed = await verifyRecaptchaToken(responseInput.recaptchaToken, survey.recaptcha.threshold);
-    if (!isPassed) {
-      return responses.badRequestResponse(
-        "reCAPTCHA verification failed",
-        {
-          code: RECAPTCHA_VERIFICATION_ERROR_CODE,
-        },
-        true
-      );
-    }
+  // Email verification, like the PIN above, must be enforced here and not only in the page renderer:
+  // this endpoint is public, so a caller could otherwise submit any `verifiedEmail` they like.
+  // Shared with the v1 endpoint so the two versions cannot drift apart.
+  const verifiedEmailErrorResponse = enforceVerifiedEmailGate({
+    survey,
+    responseData: responseInput.data,
+    metaUrl: responseInput.meta?.url,
+  });
+  if (verifiedEmailErrorResponse) {
+    return verifiedEmailErrorResponse;
   }
 
-  return null;
+  // Shared with the v1 endpoint so the two versions cannot drift apart again.
+  return await verifyResponseRecaptcha({
+    survey,
+    workspaceId,
+    recaptchaToken: responseInput.recaptchaToken,
+  });
 };
--- a/apps/web/lib/feedback-source/actions.ts
+++ b/apps/web/lib/feedback-source/actions.ts
@@ -20,6 +20,7 @@ import {
   getOrganizationIdFromFeedbackSourceId,
   getOrganizationIdFromSurveyId,
   getOrganizationIdFromWorkspaceId,
+  getWorkspaceIdFromSurveyId,
 } from "@/lib/utils/helper";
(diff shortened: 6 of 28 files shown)


mattinannt:

| 3 | **#8681 ← this PR** | 🟡 Medium | `claude/security-high` | 13 |

Merge in order: **#8679, then #8680, then this one.** GitHub retargets each PR to `main` as the one below it lands, so the diff shown here stays limited to the medium-severity work.

---

xernobyl:
Heads-up from the ENG-1922/1923 side (#8669): I trimmed my PR's overlapping create-response `displayId` change so this PR owns that fix — no more duplicate or conflict on `responses/lib/response.ts`. Two small things worth aligning:

1. **The v2 *management* update path isn't covered here.** `updateResponse` (`modules/api/v2/management/responses/[responseId]/lib/response.ts`) does `data: { ...responseInput }`, mass-assigning the client-supplied `contactId` **and** `displayId` with no ownership check — the same class as the create-path display link. #8669 now guards that sink; flagging so we don't assume the responses API is fully covered by the create-path fix here.

2. **Create-path error shape.** This PR returns `400 bad_request` "does not belong to the given survey", which confirms the display exists but is in a different survey — a mild cross-tenant existence oracle. #8669 uses a uniform `404` for foreign-vs-nonexistent (RFC 9110 §15.5.4; OWASP API1:2023 anti-enumeration) and also rejects a display already linked to another response / a contact mismatch (via the shared `assertDisplayOwnership`). Might be worth matching for consistency — not blocking.

mattinannt:
@xernobyl thanks for trimming #8669 — and for the timing. I was about to do the opposite of what you'd want: on your #8682 comment I'd concluded #8669 was the more thorough of the two on `displayId` and had just got @mattinannt's sign-off to **drop** my create-path fix in favour of yours. Your note arrived before I acted on it. If we'd both trimmed, the create-path display link would have ended up guarded by neither PR. So: I've kept it, and this PR owns the create path.

**On (2), the error shape — you're right and I've fixed it in `f9ba4fe`.** `400 bad_request` + "does not belong to the given survey" was exactly the oracle I removed from the historical-response import in #8679 (not-found rather than forbidden, so the response can't confirm an id exists), and then I reintroduced the same shape two PRs later. It now returns the uniform shape the sibling handlers already use:

```ts
return err({ type: "not_found", details: [{ field: "display", issue: "not found" }] });
```

I also added a test that asserts the *indistinguishability* rather than just the rejection — the two prior tests only checked `ok === false`, so they'd have passed with the oracle intact:

```ts
test("report a foreign and a nonexistent displayId identically", ...)
  expect(missing.error).toEqual(foreign.error);
```

Mutation-checked: reintroducing the split shape fails that test and only that test (1 failed / 15 passed).

**On (1), the v2 management update path — agreed, and thanks for covering it.** I'd confirm the sink is as you describe: `updateResponse` spreads `data: { ...responseInput }`, so both `contactId` and `displayId` are mass-assigned with no ownership check. I'm deliberately **not** adding a competing guard here — you have it in #8669 along with `assertDisplayOwnership`, the already-linked-to-another-response case and the contact mismatch, none of which I cover. So the split is: this PR owns create, #8669 owns update plus the contact side. Worth stating plainly in both descriptions so nobody reads either as "the responses API is now covered".

One thing I'd flag for whichever of us lands second: if #8669 ends up sharing `assertDisplayOwnership` across create and update, it'd be better for the create path to call that too rather than keep my inline check — one implementation is what stops these two drifting apart the way reCAPTCHA did across the v1/v2 endpoints (that's the other Medium finding in this PR). Happy to rebase onto your helper once #8669 is in, if you'd rather own the shared version.

---

mattinannt:
@pandeymangg — thanks for #8680 and #8682. This one is the middle of the stack and is the only link still unreviewed, so nothing can land until it goes through: **#8680 (approved) → #8681 (this) → #8682 (approved)**, each base auto-retargeting as the one below merges.

Current tip is `5e8cd08a` (rebased onto #8680's post-review tip `331479b5`). Green: Sonar 0 new issues / 83.8% coverage, 0 duplication, `mergeable_state: clean`.

It's the smallest of the three — four findings:

- **Closed sign-up enforced server-side.** `SIGNUP_ENABLED` + multi-org were only checked in `signup/page.tsx`, so a direct POST to `createUserAction` walked around them. Now in `assertSignupPolicyAllows`, with the fresh-instance case preserved for the initial admin.
- **`USE_SECURE_COOKIES`** now includes `WEBAPP_URL` in its chain — all three URL vars are optional, so a deployment setting only `WEBAPP_URL=https://…` served the session cookie without `Secure`.
- Plus the display-id and uniform `not_found` changes we discussed with @xernobyl earlier.

One thing to flag while you're in here: the `createUserAction` / `signup-form.tsx` hunks overlap #8691, so whichever of the two merges second will need a small reconciliation. Details in [my note on #8691](https://github.com/formbricks/formbricks/pull/8691#issuecomment-5120063245) — I've committed to doing that rebase, so it isn't extra work for you.

No rush if you're mid-something; just flagging that the stack is parked on this one.

---

xernobyl:
Correcting myself on one detail above, since it affects the rebase you offered.

I wrote that #8669 rejects the already-linked and contact-mismatch cases "via the shared `assertDisplayOwnership`". That was true of the *create-path* version I wrote and then dropped in favour of this PR — it isn't what's left. #8669's update path calls `getDisplayForResponseValidation` and compares the fields itself. (To be clear, `assertDisplayOwnership` isn't something I added: it's been in `apps/web/lib/display/service.ts` all along, and the v1/v2 *client* response paths already use it. Your instinct to converge on it is reasonable.)

The reason the update path doesn't just call it, which is worth knowing before you plan a rebase:

```ts
if (display.responseId) throw new InvalidInputError(`Display ${displayId} is already linked to a response`);
```

That rejects *any* existing response link. Create has no self case, so it's correct there. Update does: re-sending a response's own current `displayId` is a legitimate idempotent `PUT`, so the update path needs `responseId === null || responseId === <this response>`. As written, `assertDisplayOwnership` would fail that. So a genuinely shared helper needs the self-link allowance threaded through (an optional `allowResponseId`, or the caller passing its own response id) — happy to own that extraction once both land, rather than either of us blocking on it now.

One caution if the create path does move onto it later: its messages distinguish not-found / different-workspace / different-survey / different-contact. Fine as an internal exception, but the boundary has to collapse them into the single uniform 404 — otherwise the oracle you removed in `f9ba4fe` comes back in through the helper.

Last thing, low severity and not blocking: your create-path check is `display?.surveyId !== surveyId`, which covers workspace transitively (a display's survey determines its workspace, so it's the tighter check — agreed with your comment). What it doesn't cover is a display in the *same* survey that's already linked to another response; `Display↔Response` being one-to-one, that either 500s on the unique constraint or moves the display off the other response. Same tenant either way, so no cross-tenant exposure — just noting it's the one check that differs between the two paths today.

mattinannt:
Thanks for going back and correcting it — that changes the plan, so it was worth saying.

**Dropping the converge-on-`assertDisplayOwnership` idea for this PR.** Your self-link point is the reason, and it's sharper than "the helper needs a flag": create genuinely has no self case, update genuinely does, so `if (display.responseId) throw` is *correct* in one and *wrong* in the other. A helper that papers over that with an optional `allowResponseId` is the kind of parameter that reads as configuration but is really two different rules wearing one name. Better that you extract it once both land, with the update path's requirements in hand, than that I guess at the shape now.

**On the oracle caution — agreed, and worth pinning down.** My boundary currently collapses everything to one `not_found` at the point of the check itself, not downstream, so there's no internal exception carrying not-found / wrong-workspace / wrong-survey / wrong-contact for a mapper to leak. If the extraction later moves that distinction *into* a shared helper, the collapse has to move with it — the helper can distinguish internally, but every boundary that calls it needs to emit the single 404. Flagging it in the extraction PR is fine; I don't think it needs anything from me here.

**On your last point — you're right, and I checked rather than assumed.** My create-path check is:

```ts
const display = await (tx ?? prisma).display.findUnique({
  where: { id: displayId },
  select: { surveyId: true },      // ← no responseId
});
if (display?.surveyId !== surveyId) { /* uniform not_found */ }
```

So a display in the same survey that's already linked to another response passes it, and then hits exactly the one-to-one behaviour my own comment on that block describes — unique-constraint 500, or the display moves off the other response. Same failure mode I documented, minus the cross-tenant part.

**I'm not adding it here**, for two reasons rather than one:

1. It isn't purely additive. `responseId === null` is the right rule for create *in principle*, but it changes what happens to a client retrying a create whose display already got linked — today that silently succeeds by stealing the link, and rejecting it is better behaviour but still a behaviour change, on the release critical path, in a PR whose remit is cross-tenant binding.
2. It's the same question your extraction has to settle anyway. Create wants `responseId === null`; update wants `responseId === null || responseId === self`. Deciding that once, in the PR that owns both call sites, beats me hardcoding the create half here and you having to reconcile it.

So: real, same-tenant, and yours to fold in — with my agreement on record that create should end up rejecting a linked display, not just a foreign-survey one.

**Rebase status, since you asked about it:** #8680 merged (as `42734f85`, squashed rather than a true merge despite the queue notice), so this PR is rebased onto `main` and retargeted — new tip **`554b2ad5`**, one commit beyond `main`, its own 27 files. #8682 followed onto it at `0e66e81c`. Suites green: 663 in `packages/surveys`, 84 in `survey-ui`, and 6778 passing in `apps/web`.

One thing worth knowing if you see red CI on any of these: `app/api/v3/surveys/generate/error-mapping.test.ts` has 4 failures that **reproduce on a clean `origin/main` checkout** with none of my branches involved — they came in with #8585 this morning. Not from this stack, and not something I've touched.

---

xernobyl:
Agreed on all three — extraction is mine to own, with your create-side agreement (`responseId === null`) on record, and I'll carry the 404-collapse requirement into that PR rather than asking anything of this one. Your framing of the flag is better than mine: an optional `allowResponseId` would be two rules wearing one name, so the shape should come from having both call sites in hand.

**On the `error-mapping.test.ts` failures — I chased these down, and I don't think `main` is red. It's a stale local build.**

They reproduce for me too, same 4, and the trace is the giveaway:

```
TypeError: Right-hand side of 'instanceof' is not an object
  error-mapping.ts:74  if (error instanceof AIOutputTokenLimitError)
```

`instanceof` fails because the right-hand side is `undefined`, i.e. the import resolved to nothing. `AIOutputTokenLimitError` is a *new* export that arrived with #8585 in `packages/ai/src/errors.ts`, and `@formbricks/ai` resolves through its **built** output (`"exports": { "import": "./dist/index.js" }`), not source. In my worktree:

```
packages/ai/dist/index.js   built 2026-07-28 12:38   ← predates #8585
packages/ai/src/errors.ts   2026-07-30              ← has the class
$ grep -c AIOutputTokenLimitError packages/ai/dist/index.js
0
```

So the stale `dist` exports `AIConfigurationError`, `classifyAIProviderError`, … and no `AIOutputTokenLimitError`. Rebuilding that one package fixes it with nothing else changed:

```bash
pnpm build --filter=@formbricks/ai
```

→ `grep -c` goes 0 → 2, and the file goes **4 failed / 4 passed → 8 passed**.

That's consistent with the CI signal: `Run Unit Tests` is green on #8669 and the post-merge runs on `main` are green, because CI builds the workspace fresh and never sees a stale `dist`. So I'd expect nothing to chase here — worth a `pnpm build` (or `pnpm install && pnpm build`) before anyone digs further, so the release isn't held on a phantom.

Only flagging it because "4 failures on clean `main` from #8585" is the kind of note that gets someone to revert a good PR.

mattinannt:
You're right and I was wrong — `main` is not red. Retracting that note.

Verified your diagnosis exactly:

```
packages/ai/dist/index.js   2026-07-28   grep -c AIOutputTokenLimitError → 0
packages/ai/src/errors.ts   2026-07-30   grep -c AIOutputTokenLimitError → 5
packages/ai/package.json    exports.import → ./dist/index.js
```

`pnpm build --filter=@formbricks/ai` → dist count 0 → 2, and that test file goes **4 failed / 4 passed → 8 passed**. Full `apps/web` suite is now **537 files / 6782 tests, zero failures** — the 4 were the entire delta.

**My check was bad methodology, not bad luck.** I "confirmed pre-existing on main" by checking out `origin/main` and re-running — which swaps the source tree but leaves `packages/ai/dist` exactly as it was. So the stale artifact was present in both runs, and the test failed identically in both, which is precisely the observation that made it look like main's problem. Checking out a different commit proves nothing about a workspace whose packages resolve through built output; the control I needed was a rebuild, not a different ref.

That's a general trap in this repo, not specific to `@formbricks/ai` — `AGENTS.md` warns about it for `@formbricks/surveys`, and the same reasoning covers any workspace package resolving via `dist`. I've noted it so I don't repeat it.

Thanks for chasing the trace instead of taking my word for it. "4 failures on clean main from #8585" is exactly the sort of claim that gets a good PR reverted, and it was mine to get right.

Nothing outstanding on this PR from my side — tip `554b2ad5`, Sonar green (0 new issues, 83.8%), and #8682 green behind it at `0e66e81c` (0 issues, 100%). Still waiting on @pandeymangg's review here; it's the only thing the stack is parked on.

---

mattinannt:
Worked through all 8. Four fixed, one deferred with a concrete reason, three declined. New tip **`e000aa41`** (#8682 follows at `666dcd77`).

### Fixed

**1. `recall.ts` — `stringifyRecallValue` only ran on the escape path (Major).** Correct, and it was my bug: line 325 was literally `escapeValues ? escapeHtml(stringifyRecallValue(value)) : (value as string)`, so every caller outside the follow-up-email flow still cast a record straight to string. Now stringifies unconditionally and escapes conditionally. Added the two `recall.test.ts` cases you asked for — a record-shaped answer on the default path, and one asserting escaped and unescaped agree. Mutation-checked: restoring the old line fails both.

**2. `email.ts` — hardcoded `"en-US"` (Minor).** Right, and it contradicted the repo's own locale rule while the correctly-resolved `locale ?? DEFAULT_LOCALE` sat two lines above. Now reuses it.

**3. `verify-response-recaptcha.ts` — missing `cors: true` (Minor).** Confirmed against the signature: `notFoundResponse(resourceType, resourceId, cors = false, cache)`, while the sibling `badRequestResponse` calls all pass `true`. That branch would have answered a browser without CORS headers, surfacing as an opaque network error rather than a 404. Fixed.

**4. Log levels (Trivial).** Both are now `warn`. A public endpoint receiving no token is caller-controlled — at `error` anyone could flood the log by posting an empty body.

On your question about `isSpamProtectionEnabled`: **it is intentional, and it's fail-closed.** The branch only logs; `verifyRecaptchaToken` still runs immediately after, so an organization without the entitlement gets the stricter path, not a bypass. Gating there would make it fail-*open*. I've written that into the comment so the next reader doesn't have to re-derive it.

### Deferred, with the reason

**6. Localize `SIGNUP_DISABLED_ERROR_CODE` (Minor).** You're right and my existing comment was wrong. It claimed the code is only reachable by a direct POST — but sign-up can be open when the page renders and closed before submit, which is the *same* render-then-revoke race that made me localize the invite branch immediately above. Inconsistent reasoning on my part.

I'm not shipping it here because the fix needs a new `en-US` key, and adding one without a Lingo run fails the translation gate — I verified: `pnpm run scan-translations` reports the key missing from all 14 target locales plus `LOCKFILE OUT OF SYNC`. `pnpm i18n` can't run in this environment (no `LINGO_API_KEY`), and `AGENTS.md` forbids hand-writing the non-English locales. Reddening `Validate Translation Keys` on a release-critical PR for a toast string is the wrong trade, so the corrected reasoning is now in the code comment and the change is a follow-up for someone with the key.

### Declined

**5. Invite validation + account creation not atomic (Major).** Same finding you raised on #8680, declined for the same reason: pre-existing, and it grants no privilege — `autoSignIn: false` means no session is minted for the orphaned account. Worth noting #8691 is the PR that actually fixes this area, and its author reached the sharper version of it independently.

**7. Fresh-instance bootstrap not atomic (Major).** Real race, but the window is "two people race to be the very first user of an empty instance, during the seconds before anyone has signed up". Both would land as separate org owners, which is the same outcome as an operator creating two accounts. Making it atomic means a transactional claim around Better Auth's own user creation, which I'm not doing inside a security PR on the release path.

**8. `as any` in the 2FA test fixture (Trivial).** That file has 23 other `as any` casts. Fixing only the line I touched leaves a mixed style and buys no real type safety; it's a file-wide cleanup, not this PR.

Green on the new tip: **537 files / 6784 tests**, eslint and prettier clean, `scan-translations` passing.

---

pandeymangg (approved):

PR #8682, opened 2026-07-29
fix(security): harden the low-severity storage, authz and abuse paths [4/4 low]

mattinannt:
## What & why

**4 of 4** from the pre-release security review, split out of #8667 by severity. **Severity: Low** — 4 findings. None is a live bypass on a stock deployment; each removes a way for the next change to become one, or brings a permission into line with the rest of the API.

> ⚠️ **Stacked on #8681.** Base is `claude/security-medium`, so this PR's diff shows only the low-severity work. #8679 has merged; merge #8680 and #8681 before this one.

- **Relative path segments in storage keys.** Download and delete keys are built as `${id}/${accessType}/${fileName}`, and `fileName` comes from the `[...filePath]` route param, so `..` segments named an object outside the caller's own `workspaceId/accessType/` prefix — including a `private/` one through the unauthenticated `public/` route. This does **not** resolve on a stock S3/MinIO backend, which treats keys as opaque and signs the unnormalized path, so it is hardening rather than an exploit. But whether one tenant can name another's file should not depend on how the storage backend, or a proxy in front of it, happens to treat dot segments. Rejected on read and delete, and checked both before *and* after the decode because the download handler decodes twice (`%252e%252e` only becomes `..` on the second pass).

- **`testEndpointAction` unauthorized when `webhookId` was omitted.** Any logged-in user — including a billing-only member — could make the server POST a signed payload to a URL of their choosing. SSRF was already blocked by the pinned dispatcher (IP-pinned, `redirect: "manual"`); this closes the "who may use it" half.

- **Unsplash actions were unauthenticated** while spending the instance's `UNSPLASH_ACCESS_KEY`. Now behind `authenticatedActionClient` and rate-limited **per user** (30/min) — neither action carries a workspace or survey id to scope against, and the quota belongs to the whole instance, so a session is the only thing to key on.

- **`deleteWorkspaceAction` name-confirmation oracle.** It compared the typed workspace name *before* authorizing, so the mismatch error distinguished "wrong name" from "workspace you may not touch" — a cross-tenant workspace-name oracle.

## Removed from this PR: feedback-record deletion at `manage` level

An earlier revision raised feedback-record deletion from `write`/`readWrite` to `manage` on both the v3 gateway and the UI server action. **That has been dropped in favour of #8675 (ENG-1770)**, after @xernobyl [flagged the conflict](https://github.com/formbricks/formbricks/pull/8682#issuecomment-5116736738) and @mattinannt confirmed the call.

The reason is not merge convenience — #8675 is simply correct where mine was not. My change kept the *workspace* as the dimension and only raised the bar within it. But a feedback record's only tenancy is its feedback directory, and a directory is shared by every workspace assigned to it, so the record carries no workspace of its own. No workspace permission can therefore distinguish workspace A's records from workspace B's: a workspace-`manage` holder in B could still delete records that A's surveys ingested. #8675 removes the workspace path for record mutations entirely, which is the only thing that actually draws that line, and it also covers `update`, both the gateway and the server action, the UI affordances and the taxonomy routes.

Consequences of the removal, recorded so nothing is assumed:

- `apps/web/modules/hub/feedback-records-gateway.ts` and `apps/web/modules/ee/unify-feedback/actions.ts` are now byte-identical to their `claude/security-medium` versions, and `apps/web/modules/hub/feedback-records-gateway.test.ts` is gone from this PR. That also clears the add/add conflict with #8675, which creates the same test file, and means this PR no longer needs porting into the `feedback-records-gateway-authz.ts` module that #8648 extracts.
- This PR previously and incidentally closed the cross-org `delete`/`bulkDelete` paths, because org-level access control only models read/write and so could never satisfy `manage`. That is no longer closed here. It is **#8648's** to close for every operation at once — and note it is not closed by #8675 either, whose `hasApiKeyImplicitFeedbackDirectoryAccess` still returns `true` on `orgAccessControl?.write` before its mutation check. If #8675 lands before #8648, the cross-org delete path is open in the interim. Flagged on #8648 as a reason to keep it first.

## Linear ticket

Documented here rather than filed individually; see the tracking comment on #8667.

## How this was tested

- `npx vitest run modules/storage modules/core/rate-limit modules/integrations/webhooks modules/hub modules/ee/unify-feedback` (`apps/web`) — 32 files / 484 tests ✅ after the removal, including the untouched gateway suites.
- Whole-stack run after rebasing onto the post-#8679 `main`: 402 files / 5040 tests ✅ across every area the four PRs touch.
- `eslint` and `tsc --noEmit` clean for every touched file.
- Regression tests: storage path traversal on read and delete, raw and double-encoded (`storage/service.test.ts`); the rate-limit config surface (`rate-limit-configs.test.ts`).

## QA / Test Plan

**How to test**

- [ ] Storage: uploading, downloading and deleting a survey file upload all still work. Requesting a download path containing an encoded `..` segment (e.g. `/storage/<workspaceId>/public/..%252f..%252f<otherWorkspaceId>%252fprivate%252ffile.pdf`) → rejected as invalid input rather than reaching the backend.
- [ ] Webhooks: "Test endpoint" still works when creating a webhook and on an existing webhook's settings tab. As a `billing`-role member, calling the action directly → rejected.
- [ ] Unsplash: the survey editor's image search and selection still work; hammering the search past 30 requests/minute as one user returns a rate-limit error. Signed-out callers are rejected.
- [ ] Workspace deletion: the confirm-by-name flow still works for an owner. For a workspace you have no access to, the error does not distinguish a wrong name from missing permission.
- [ ] Regression from the removal above: deleting a feedback record still behaves exactly as it does on `main` today (workspace `readWrite` in the UI, `write` over the gateway). The tightening now arrives via #8675, not this PR.

**Preconditions / test data**

- A `billing`-role member; a survey with a file-upload question and both public and private stored files; `UNSPLASH_ACCESS_KEY` configured; a second workspace the test user cannot access, for the deletion-oracle check.

**Risks & regressions**

- No user-visible permissions change remains in this PR. The feedback-record tightening that previously needed a release note has moved to #8675.
- Unsplash is now authenticated, so any non-editor caller of those actions (there should be none — they are only used by the survey editor) would break.
- The storage path check runs on every download and delete. It rejects only keys containing dot segments, which no legitimate client produces, but it is on a hot read path.

- No DB migrations.
- New rate-limit bucket `actions.unsplash` (30/minute per user); no configuration needed.

## Merge order

1. ~~#8679 (Critical)~~ — merged as `f1114c1`
2. #8680 (High) — now targets `main`
3. #8681 (Medium)
4. **#8682 — this PR** (Low)

### DIFF
--- a/apps/web/modules/core/rate-limit/rate-limit-configs.test.ts
+++ b/apps/web/modules/core/rate-limit/rate-limit-configs.test.ts
@@ -97,10 +97,20 @@ describe("rateLimitConfigs", () => {
         "isSurveyResponsePresent",
         "validateSurveyPin",
         "licenseRecheck",
+        "unsplash",
         "inviteMember",
         "bulkInviteMembers",
         "generateExampleResponses",
       ]);
+
+      // Exact values, not just presence: this quota is the only thing bounding one account from
+      // exhausting the instance-wide UNSPLASH_ACCESS_KEY, so a loosened interval, allowance or
+      // namespace is a security regression and should fail here rather than in production.
+      expect(rateLimitConfigs.actions.unsplash).toEqual({
+        interval: 60,
+        allowedPerInterval: 30,
+        namespace: "action:unsplash",
+      });
     });
 
     test("should have all storage configurations", () => {
@@ -171,6 +181,7 @@ describe("rateLimitConfigs", () => {
         { config: rateLimitConfigs.api.clientEnvironment, identifier: "environment-id" },
         { config: rateLimitConfigs.actions.emailUpdate, identifier: "user-profile" },
         { config: rateLimitConfigs.actions.accountDeletion, identifier: "user-account-delete" },
+        { config: rateLimitConfigs.actions.unsplash, identifier: "user-unsplash" },
         { config: rateLimitConfigs.storage.upload, identifier: "storage-upload" },
         { config: rateLimitConfigs.storage.uploadPerWorkspace, identifier: "storage-upload-workspace" },
         { config: rateLimitConfigs.storage.delete, identifier: "storage-delete" },
--- a/apps/web/modules/core/rate-limit/rate-limit-configs.ts
+++ b/apps/web/modules/core/rate-limit/rate-limit-configs.ts
@@ -48,6 +48,7 @@ export const rateLimitConfigs = {
       namespace: "action:validate-survey-pin",
     }, // 10 per minute — prevents brute-force PIN guessing
     licenseRecheck: { interval: 60, allowedPerInterval: 5, namespace: "action:license-recheck" }, // 5 per minute
+    unsplash: { interval: 60, allowedPerInterval: 30, namespace: "action:unsplash" }, // 30 per minute per user — bounds one account exhausting the instance-wide UNSPLASH_ACCESS_KEY quota
     inviteMember: { interval: 3600 * 24, allowedPerInterval: 20, namespace: "action:invite-member" }, // 20 per day  — bounds invite-spam abuse
     bulkInviteMembers: {
       interval: 3600 * 24,
--- a/apps/web/modules/integrations/webhooks/actions.ts
+++ b/apps/web/modules/integrations/webhooks/actions.ts
@@ -141,6 +141,10 @@ export const updateWebhookAction = authenticatedActionClient.inputSchema(ZUpdate
 const ZTestEndpointAction = z.object({
   url: z.string(),
   webhookId: ZId.optional(),
+  // Required so the not-yet-created-webhook path has something to authorize against; without it this
+  // action did no authorization at all and any logged-in user — including a billing-only member with no
+  // product access — could make the server POST a signed test payload to any URL they named.
+  workspaceId: ZId,
   secret: z.string().optional(),
 });
 
@@ -173,6 +177,23 @@ export const testEndpointAction = authenticatedActionClient
 
       secret = webhookResult.data.secret ?? undefined;
     } else {
+      // No webhook yet: authorize against the workspace the webhook is being created in.
+      await checkAuthorizationUpdated({
+        userId: ctx.user.id,
+        organizationId: await getOrganizationIdFromWorkspaceId(parsedInput.workspaceId),
+        access: [
+          {
+            type: "organization",
+            roles: ["owner", "manager"],
+          },
+          {
+            type: "workspaceTeam",
+            minPermission: "readWrite",
+            workspaceId: parsedInput.workspaceId,
+          },
+        ],
+      });
+
       // New webhook, use the provided secret or generate a new one
       secret = parsedInput.secret ?? generateWebhookSecret();
     }
--- a/apps/web/modules/integrations/webhooks/components/add-webhook-modal.tsx
+++ b/apps/web/modules/integrations/webhooks/components/add-webhook-modal.tsx
@@ -75,6 +75,7 @@ export const AddWebhookModal = ({
       const testEndpointActionResult = await testEndpointAction({
         url: testEndpointInput,
         secret: webhookSecret,
+        workspaceId,
       });
 
       if (!testEndpointActionResult?.data) {
--- a/apps/web/modules/integrations/webhooks/components/webhook-settings-tab.tsx
+++ b/apps/web/modules/integrations/webhooks/components/webhook-settings-tab.tsx
@@ -76,6 +76,7 @@ export const WebhookSettingsTab = ({
       const testEndpointActionResult = await testEndpointAction({
         url: testEndpointInput,
         webhookId: webhook.id,
+        workspaceId: webhook.workspaceId,
       });
       if (!testEndpointActionResult?.data?.success) {
         const errorMessage = getFormattedErrorMessage(testEndpointActionResult);
--- a/apps/web/modules/storage/service.test.ts
+++ b/apps/web/modules/storage/service.test.ts
@@ -405,6 +405,29 @@ describe("storage service", () => {
       expect(result).toEqual(mockError);
       expect(deleteFileFromS3).toHaveBeenCalledTimes(1);
     });
+
+    // Regression: the key is `${id}/${accessType}/${fileName}`, so a `..` segment would delete
+    // objects outside the workspace prefix the caller was authorized against.
+    test.each([
+      "../../ws-victim/private/secret.pdf",
+      "sub/../../../ws-victim/private/secret.pdf",
+      "./file.jpg",
+    ])("should reject traversal in the file name: %s", async (fileName) => {
+      const result = await deleteFile("ws-456", "public" as TAccessType, fileName);
+
+      expect(result.ok).toBe(false);
+      expect(deleteFileFromS3).not.toHaveBeenCalled();
+    });
+
+    test("should still allow nested file paths without dot segments", async () => {
+      const mockSuccess = { ok: true, data: undefined } as MockedDeleteFileReturn;
+      vi.mocked(deleteFileFromS3).mockResolvedValue(mockSuccess);
+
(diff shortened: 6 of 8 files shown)


mattinannt:

| 4 | **#8682 ← this PR** | ⚪ Low | `claude/security-medium` | 5 |

Merge last, after #8679 → #8680 → #8681. GitHub retargets each PR to `main` as the one below it lands, so the diff shown here stays limited to the low-severity work.

One extra dependency for this PR specifically: **#8648 (ENG-1980) should merge before it.** That PR moves `hasApiKeyImplicitFeedbackDirectoryAccess` into a new `feedback-records-gateway-authz.ts`, and the `manage`-for-deletes change here edits the same function — so this branch will need a rebase onto it.

---

xernobyl:
Heads-up from reviewing #8648: three open PRs now change `hasApiKeyImplicitFeedbackDirectoryAccess`, and this one and #8675 disagree on the same question.

**This PR vs #8675.** Here, `delete`/`bulkDelete` require `manage` weight *from a workspace permission*. In #8675 (ENG-1770), `update`/`delete`/`bulkDelete` are authorizable **only** at the org level and the workspace path returns `false` for mutations outright — so a key holding workspace `manage` is precisely what this PR requires and precisely what #8675 refuses. Worth settling between the two of you: may a workspace permission mutate records in a directory that's shared across workspaces? That's a product call, and a rebase won't decide it.

**This PR vs #8648.** Deferring the org binding to #8648 is the right split, but note that #8648 doesn't just add an org check — it deletes the `organizationAccess` shortcuts and moves the function plus both weight tables into a new `feedback-records-gateway-authz.ts` (kept free of `server-only`/Prisma/env imports so the decision is unit-testable). So this PR's `manage` entry in `gatewayPermissionToApiKeyPermissionWeight`, and the `TFeedbackRecordsGatewayPermission` widening to `"read" | "write" | "manage"`, will need porting into that module rather than the gateway. Also, this PR and #8675 both add `apps/web/modules/hub/feedback-records-gateway.test.ts` as a new file, so there's an add/add conflict between them on top of the conflict with #8648.

One consequence worth recording while the ordering is undecided: **this PR alone** incidentally closes the cross-org *delete* and *bulkDelete* paths, since org-level access control only models read/write and so can't satisfy `manage`. Cross-org read, create and update stay open until #8648 lands.

Suggested order: #8648 first (it's the one that closes the cross-tenant hole, review-clean and green), then rebase this and #8675 onto the extracted module.

mattinannt:
@xernobyl thanks — this is the right catch, and having read #8675 properly I think it settles more of it than "product call" suggests.

**Agreed on ordering and on the port.** #8648 first. And you're right that my `manage` entry in `gatewayPermissionToApiKeyPermissionWeight` plus the `TFeedbackRecordsGatewayPermission` widening to `"read" | "write" | "manage"` belong in the extracted `feedback-records-gateway-authz.ts`, not the gateway — I'd only planned for a line-shift rebase, not a move into a new module. Confirmed the add/add on `feedback-records-gateway.test.ts` too: both this PR and #8675 create it.

**On this PR vs #8675 — I don't think it's symmetric, and I think #8675 is simply right.** My change raises the bar from workspace `readWrite` to workspace `manage`, which keeps the *workspace* as the dimension. #8675's rationale is that the dimension itself is wrong, and that's a fact about the data model rather than a preference:

> A record's only tenancy is its feedback directory (the Hub tenant), and a directory is shared by every workspace it is assigned to — the record itself carries no workspace.

If a directory is shared by workspaces A and B, then no workspace permission can distinguish A's records from B's. So requiring workspace `manage` doesn't fix the problem it looks like it fixes: a `manage` holder in workspace B can still delete records that workspace A's surveys ingested. It just makes that harder to obtain. #8675 removes the workspace path for mutations entirely, which is the only thing that actually draws the line — and it covers `update` as well as the two deletes, both the gateway and the server action, plus the UI affordances and the taxonomy routes. Mine covers `delete`/`bulkDelete` on two paths.

So my answer to "may a workspace permission mutate records in a directory that's shared across workspaces?" is no, and #8675 already implements that. I'd rather drop the feedback-record permission change from this PR than have you two reconcile two half-answers. I'm confirming with @mattinannt before I remove it, since he and I settled on `manage`-for-deletes earlier and this reverses that — I'll update here either way, and if it comes out the other way I'll port into #8648's module as you describe.

**One correction to record, in the direction that argues against keeping mine.** You noted this PR incidentally closes cross-org `delete`/`bulkDelete` because org-level access control only models read/write and so can't satisfy `manage` — that's right, and it's in the code comment. But it isn't a reason to keep it: in #8675, `hasApiKeyImplicitFeedbackDirectoryAccess` still returns `true` on `orgAccessControl?.write` *before* the mutation check, so an org-write key from org A can still delete in org B's directory there. That gap is #8648's to close for every operation at once, not something to buy piecemeal by picking the stricter of two overlapping PRs. If #8675 lands before #8648, that cross-org delete path reopens in the interim — worth flagging on #8648 as a reason to keep it first.

Nothing else in this PR touches that function, so the remaining four findings here are independent of all of this.

---

pandeymangg (dismissed):
LGTM

mattinannt:
@pandeymangg — your approval got dismissed by a rebase, not by a change. Re-approval needed, and it should be a formality.

**What happened:** #8681 squash-merged, which put its content on `main` as a new commit rather than as this branch's parent. That left this PR sitting two commits above `main`, the lower one being a duplicate of work already merged — so its diff would have shown #8681's changes as if they were new. I rebased to replay only this PR's own commit, and the force-push dismissed the stale review. Same thing happened after #8680; it's the cost of a squash-merged stack.

**Nothing you reviewed has changed.** I checked each file's blob hash against `191523d5`, the commit you approved:

| File | vs. approved |
|---|---|
| `modules/core/rate-limit/rate-limit-configs.ts` | identical |
| `modules/core/rate-limit/rate-limit-configs.test.ts` | identical |
| `modules/integrations/webhooks/actions.ts` | identical |
| `modules/integrations/webhooks/components/add-webhook-modal.tsx` | identical |
| `modules/integrations/webhooks/components/webhook-settings-tab.tsx` | identical |
| `modules/storage/service.ts` | identical |
| `modules/storage/service.test.ts` | identical |
| `modules/survey/editor/actions.ts` | identical |

All 8, byte for byte. Only the base moved.

New tip `ca4159de`, base auto-retargeted to `main`, exactly 1 commit beyond it. Verified the diff no longer re-shows any of #8681's medium-severity work. Local suite green at 543 files / 6969 tests; CI is re-running now and I'll confirm it lands green.

This is the last of the four security tranches — #8679, #8680 and #8681 are all on `main`, and #8686 backported the Critical fix to `release/5.2`.

---

mattinannt:
Good catch — fixed. New tip **`2ee7d81d`**.

You're right that the assertion only proved `"unsplash"` was present and correctly ordered. That's the weakest possible check for this particular entry: the 30/60 quota is the only thing stopping one account from draining the instance-wide `UNSPLASH_ACCESS_KEY`, so a loosened value is a security regression, and nothing in the suite would have failed.

Two changes:

1. **Exact-value assertion**, matching the shape #8680 used for `auth.emailToken` (the other config where the number *is* the control):
   ```ts
   expect(rateLimitConfigs.actions.unsplash).toEqual({
     interval: 60,
     allowedPerInterval: 30,
     namespace: "action:unsplash",
   });
   ```
2. **Added to the behavioral matrix** alongside the other action configs, so enforcement is actually exercised rather than just the shape.

Mutation-checked rather than assumed: loosening the quota to `300/60` fails the new assertion, and restoring it passes. Before this change the same mutation was silent.

Suite green at 543 files / 6969 tests, eslint and prettier clean.

Worth noting the timing made this free — @pandeymangg's approval was already dismissed by the rebase off #8681's squash-merge, so folding this in costs no additional review round. The 8 files otherwise remain byte-identical to what was approved at `191523d5`; this is the only content change since.

---

pandeymangg (approved):

PR #8683, opened 2026-07-29
fix(workflows): inline field validation on the Send email step [ENG-2032]

itsjavi:
## What & why

The Send email config panel gave no per-field feedback. The canvas card read "Set recipient & contents" while the panel showed nothing highlighted, no inline error text, and no indication of which fields were even required — so with only Subject empty, there was no way to tell what the editor was unhappy about.

**Required vs optional** — Send to, Subject, Body and the trigger's Survey carry an asterisk plus a screen-reader-only "Required". Reply to and the toggles stay unmarked.

**Field-level errors** — a required field shows `border-red-500` + a red label + inline message once the user has engaged with it and left it empty, mirroring the dashboard's `isInvalid` + `FormError` visual contract without the react-hook-form dependency (the inspector has no submit; validity is derived from the definition and autosaved). Deliberately touched-based, so a freshly added node opens clean instead of painting three errors:

- Subject: on blur
- Send to: on closing the dropdown without picking (a chosen recipient can't be un-chosen, so that's the only "touched while empty" moment)
- Body: once it has held content and then been emptied
- Send to + Body also reveal immediately when a trigger survey change wipes them, since that happens without the user touching either

**Actionable problems** — a row in the validation-problems dialog that resolves to a single config field is now a button: it closes the dialog, opens that step's panel, focuses the offending field, and reveals every missing field on that step. Rows fixed elsewhere (naming the workflow, adding a trigger, connecting the flow) stay passive — there's nothing to focus. `step_incomplete` keeps its step-level grouping so the problem count still matches the one unfinished step the user perceives; the specific field is resolved at click time as the first still-blank one.

### The bug found while investigating

Not in the ticket, and the more serious half of this PR. The email body is authored in a Lexical editor whose serializer always emits the enclosing block, so a body the user emptied out comes back as `<p class="fb-editor-paragraph"><br></p>` — not `""`. Both the client's canvas issue flag and the server's `ZWorkflowExecutableDefinition` tested `body.trim().length === 0`, so **an empty body read as filled**: the warning cleared, the workflow could be enabled, and it would send a blank email.

I confirmed the serialization directly (rendered the editor in jsdom and inspected what `setText` receives) rather than inferring it. Both sites now share `getBlankSendEmailContentFields`, which strips markup and whitespace entities before testing the body, and preserves recall-token-only bodies (`RecallNode.exportDOM` writes the token as text content). One rule, so the schema, the canvas flag, and the panel's errors can't disagree about what's missing.

### Autosave — dropped, now covered on the base branch

An earlier revision of this PR added a "Saving…" state to the header pill. [ENG-1970](https://linear.app/formbricks/issue/ENG-1970) (#8697) has since landed on `epic/workflows-v1` with a pill that already reports the in-flight state and adds a failed state on top, so the merge defers to it wholesale and this PR no longer touches the indicator. The three i18n keys it had introduced (`autosave_on`, `autosave_saving`, `changes_saved`) are gone with it — `autosave_saving` was byte-identical to base's `saving_changes` (same lockfile checksum), so no copy was lost.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2032/workflows-editor-no-inline-field-level-validation-errors-unclear

Two of the ticket's asks were already covered by the ENG-2030 work that landed since it was filed (the live "N problems" pill and its dialog, and an autosave pill). This PR closes the rest.

## How this was tested

- `packages/workflows`: **13 files, 229 tests passed** (`pnpm exec vitest run`)
- `apps/web` workflows: **37 files, 403 tests passed** (`pnpm exec vitest run modules/ee/workflows`)
- `pnpm turbo typecheck --filter=@formbricks/web`: ✅ clean
- `pnpm exec eslint modules/ee/workflows --quiet`: ✅ clean
- `pnpm i18n:validate`: ✅ 4127 keys used == 4127 in translations, 0 missing, 0 unused, all locales complete

Tests added: `content.test.ts` (blank-rich-text and per-field rules, including the recall-token case), an executable-schema case rejecting the empty-markup body, a canvas-flag case for the same, `getWorkflowValidationProblemFocusTarget` (field resolution order, non-jumpable problems), and the focus-request atoms.

**Not browser-verified.** Port 3000 was held by another session's dev server, and this app's auth is pinned to `WEBAPP_URL=http://localhost:3000`, so serving on another port would have exercised a different worktree. The QA plan below is the manual pass this still needs — specifically the touched-state timing and the jump-to-field focus.

image

## QA / Test Plan

**How to test**

- [ ] Open a draft workflow → editor. Add a **Send email** step and select it → panel opens with **Send to**, **Subject** and **Body** each showing a red `*`, and **no** red borders or error text yet (a brand-new step must open clean).
- [ ] Click into **Subject**, type nothing, click elsewhere (blur) → Subject's label turns red, its border turns red, and "Add a subject line." appears beneath it. Subject only.
- [ ] Open the **Send to** dropdown and close it without picking anyone → Send to shows the red state and "Pick who should receive this email."
- [ ] Type something into **Body**, then delete it all → Body shows the red state and "Add the message to send." (an untouched, never-filled Body must stay clean until then)
- [ ] Fill all three → red states and messages clear, and the canvas card stops reading "Set recipient & contents".
- [ ] With a required field still blank, click the **"N problems"** pill in the header → dialog lists the problems. Click the row for the incomplete Send email step → dialog closes, that step's panel opens, the first still-blank field receives focus, and **every** missing field on that step is now showing its error.
- [ ] In the same dialog, a row for an un-named workflow / missing trigger / disconnected flow is plain text, **not** clickable — nothing to focus.
- [ ] Trigger step: **Survey** carries a red `*`. **Reply to** and the toggles on the email step carry **no** `*`.
- [ ] Change the trigger's survey to a different one so it wipes the recipient/body → Send to and Body reveal their errors **immediately**, with no click needed on either.

**Edge cases**

- [ ] Empty-body regression (the main bug): in **Body**, type a character then delete it so the editor is visually empty, and fill Send to + Subject. Expected: the step still reads as incomplete, the "N problems" pill still counts it, and **Enable** is refused. Before this change the workflow enabled and sent a blank email.
- [ ] Body containing **only a recall token** (insert a recall/`@` token and nothing else) → treated as filled, no error, workflow can be enabled.
- [ ] Body of only spaces → treated as blank, error shown.

**Preconditions / test data**

- Workflows are Enterprise-gated: the org's license must include the `workflows` feature, and the workflow must be a **draft** (an enabled workflow shows "Disable the workflow to make changes here" and the panel is read-only).
- At least one survey with an ending card, and ≥2 surveys if you want to exercise the trigger-survey-change step.
- At least one other organization member, so the **Send to** dropdown has a pickable recipient (recipients are restricted to org members per ENG-2029).

**Risks & regressions**

- The blank-body rule is now shared by the executable schema, the canvas issue flag and the panel's errors — a bad rule would misfire in all three. Re-check that a **legitimately filled** body (plain text, a list, bold text, a recall token) still enables the workflow; ENG-2034's list-formatting fix is nearby.
- Enabling a workflow goes through `ZWorkflowExecutableDefinition`, which is now stricter about the body. Confirm existing workflows that are already enabled still load and still run.
- Header pill: this PR no longer touches it, but it merged with ENG-1970's version — sanity-check that Saving… → All changes saved → Save failed still behaves.

- none


apps/web/modules/ee/workflows/lib/definition-to-flow.test.ts
```diff
@@ -119,6 +119,18 @@ describe("workflowDefinitionToFlowNodes / Edges", () => {
     });
   });
 
+  test("flags an email node whose body is only the editor's empty markup", () => {
+    // The rich-text editor keeps the enclosing block when the body is emptied, so a trim-based
+    // check would clear the flag on a step that has nothing to send.
+    const def = buildDefinition();
+    (def.nodes[0] as { config: { body: string } }).config.body = '<p class="fb-editor-paragraph"><br></p>';
```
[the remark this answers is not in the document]
itsjavi: Not reproducible — line 126 is a single assignment:

```ts
(def.nodes[0] as { config: { body: string } }).config.body = "<p class=\"fb-editor-paragraph\"><br></p>";
```

The cast expression `(x as T).config.body` appears to have been parsed as two assignments. Skipping; no dead code to remove.

apps/web/modules/ee/workflows/components/inspector/workflow-trigger-form.tsx
```diff
@@ -35,6 +41,25 @@ export const WorkflowTriggerForm = ({ node, isEditable, onChange }: Readonly<Wor
...
+    // The request is cleared inside the frame, not before it: clearing is what re-runs this effect,
+    // and an earlier clear would let the re-run's cleanup cancel the frame before it ever fired.
+    const frame = requestAnimationFrame(() => {
+      const target = document.getElementById("workflow-trigger-survey");
+      target?.focus();
+      target?.scrollIntoView({ block: "nearest" });
+      clearFocusRequest();
+    });
+    return () => cancelAnimationFrame(frame);
+  }, [focusRequest, node.id, clearFocusRequest]);
```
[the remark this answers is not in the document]
itsjavi: Valid — fixed in ed3914ae6, though not gated on a blank `surveyId`.

A blank `surveyId` is nearly unreachable here: `addWorkflowTriggerAtom` seeds a placeholder cuid, so the real broken state is "set but does not resolve to a survey". That is what `hasBoundTriggerSurveyAtom` tracks, so the picker is invalid on `!hasBoundSurvey` — covering both blank and dangling ids.

No `touched` gating either, unlike the email step: an unbound survey means the stored config points at something that no longer exists, which is never a normal mid-edit state.

On not marking a selected survey invalid for stale-ending requests — agreed, and handled by moving the target rather than suppressing the styling. `trigger.config.endingCardIds` now routes to the endings list (`workflow-trigger-ending-scope`) with its own error, so the survey picker is never implicated in an endings problem. Staleness is derived with the same `endingsQuery.isSuccess` gate `WorkflowValidationStatus` uses, so a loading/403/404 query cannot render as an error.

apps/web/modules/ee/workflows/lib/display.ts
```diff
@@ -105,3 +109,34 @@ export const getWorkflowValidationProblemLocation = (
...
+): TWorkflowNodeFieldFocusRequest | null => {
+  if (!definition) return null;
+
+  const nodeMatch = NODE_FIELD_PATTERN.exec(problem.field);
+  if (nodeMatch) {
+    const node = definition.nodes[Number(nodeMatch[1])];
+    if (!node || node.type !== "action") return null;
+    const [firstBlankField] = getBlankSendEmailContentFields(node.config);
+    return firstBlankField ? { nodeId: node.id, field: firstBlankField } : null;
+  }
```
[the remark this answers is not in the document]
itsjavi: Valid — fixed in ed3914ae6 with a dedicated exact-match pattern:

```ts
const NODE_CONFIG_FIELD_PATTERN = /^nodes\.(\d+)\.config$/;
```

The `node.type !== "action"` guard already caught `nodes.N.type` for `if_else`, but you are right that it did not cover a non-config path on an *action* node — a `definition_invalid` fallback at `nodes.N.label` would have jumped to "first blank content field" with a message that never mentioned it. New test `ignores non-config paths on an action node` covers `nodes.0.type`, `nodes.0.label`, `nodes.0.config.to` and bare `nodes.0` against a node with all three fields blank, so a regression cannot pass silently.

Trigger paths moved to an explicit map at the same time, so `trigger.config.*` is no longer a prefix match either.

packages/workflows/src/types/content.ts
```diff
@@ -0,0 +1,42 @@
...
+ * enclosing block, so a body the user emptied out comes back as `<p class="…"><br></p>` — 11+
+ * characters — rather than `""`. A plain `.trim()` check therefore reads a blank body as filled,
+ * which let an empty-bodied workflow clear its "needs contents" flag and be enabled.
+ *
+ * Recall tokens survive this check: `RecallNode.exportDOM` writes the
+ * `#recall:<id>/fallback:<value>#` token as the element's text content, so a body consisting only
+ * of a recall token is correctly treated as filled.
+ */
+export const isBlankWorkflowRichText = (value: string): boolean =>
+  value.replaceAll(HTML_TAG_PATTERN, "").replaceAll(HTML_WHITESPACE_ENTITY_PATTERN, " ").trim().length === 0;
```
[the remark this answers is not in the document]
itsjavi: Confirmed real, and fixed as far as this PR can reasonably go — the full remedy is a separate change.

Reproduced before fixing: `<p><3</p>` stripped to `""`, and `<p>5 < 6</p>` to `"5"`. So a body of `<3` was treated as blank and could not be enabled.

Fix in ed3914ae6 gates the pattern on a tag name so it can only match something tag-shaped:

```ts
const HTML_TAG_PATTERN = /<\/?[a-zA-Z][^>]*>/g;
```

| body | before | after |
|---|---|---|
| `<p><3</p>` | blank ❌ | `<3` ✅ |
| `<p>5 < 6</p>` | `5` ❌ | `5 < 6` ✅ |
| `<p>a > b</p>` | `a > b` ✅ | `a > b` ✅ |
| `<p><br></p>` | blank ✅ | blank ✅ |
| `<p><b>x</b></p>` | `x` ✅ | `x` ✅ |

Regression case added (`treats visible text containing an angle bracket as filled`).

On the heavy-lift half — preserving escaped text or editor state before extracting visible content: not doing it here. That means changing what the shared `Editor` persists (`toolbar-plugin.tsx` un-escapes `&lt;`/`&gt;` on every save), which affects survey follow-ups and every other consumer, and would need a migration for existing stored bodies. Out of scope for a validation-feedback PR.

The residual case is narrow and documented in the helper: a body whose only visible text is `<` immediately followed by a letter with a later `>` (e.g. `<b`) still reads as blank. It fails conservatively — the author is asked for more content rather than an empty email being sent — which is the right direction for this check. Happy to file it as a follow-up against the editor if you want it tracked.

apps/web/modules/ee/workflows/components/inspector/workflow-email-action-form.tsx
```diff
@@ -63,11 +76,69 @@ export const WorkflowEmailActionForm = ({
...
+  // point of the jump is to answer "which field is wrong") and focus the one it pointed at.
+  useEffect(() => {
+    if (focusRequest?.nodeId !== node.id) return;
+    const { field } = focusRequest;
+    markTouched(...getBlankSendEmailContentFields(node.config));
+
+    // One frame late so the inspector's width transition has laid the panel out before we scroll.
+    // The request is cleared inside the frame, not before it: clearing is what re-runs this effect,
+    // and an earlier clear would let the re-run's cleanup cancel the frame before it ever fired.
+    const frame = requestAnimationFrame(() => {
```
Dhruwang: **Duplicated focus-jump effect**

This `requestAnimationFrame` → focus → `scrollIntoView` → `clearFocusRequest` effect is duplicated near-verbatim in `workflow-trigger-form.tsx` (the rAF block around its focus handler), including the "one frame late" rationale. Two copies of the same lifecycle will drift.

Suggest extracting a small shared hook, e.g. `useWorkflowNodeFieldFocusJump(nodeId, resolveElement)`, that owns the rAF/focus/scroll/clear lifecycle once, and have both forms call it with their own field→element resolver.

Minor, same spot: the rAF has no visibility guard — in a backgrounded tab the callback is deferred, so the jump can land late. Not blocking, worth a comment if intentional.

apps/web/modules/ee/workflows/lib/display.ts
```diff
@@ -105,3 +109,49 @@ export const getWorkflowValidationProblemLocation = (
 
   return null;
 };
+
+// Exactly the step-level path `step_incomplete` reports at — deliberately not the looser
+// NODE_FIELD_PATTERN above. A problem somewhere else on an action node (a shape issue at
+// `nodes.N.label`, an unsupported `nodes.N.type`) must not resolve to "the first blank content
+// field", which would land the user on a field the message never mentioned.
+const NODE_CONFIG_FIELD_PATTERN = /^nodes\.(\d+)\.config$/;
```
Dhruwang: **Silent string-path coupling**

`NODE_CONFIG_FIELD_PATTERN` and `TRIGGER_CONFIG_FIELD_TARGETS` here re-derive the field-path strings that `state/editor.ts` (`categorizeDefinitionIssue`) produces — `nodes.N.config`, `trigger.config.surveyId`, `trigger.config.endingCardIds` — with no shared constant tying producer and consumer together. `editor.test.ts` and `display.test.ts` also hardcode these literals independently.

So a producer-side rename (e.g. `nodes.N.config` → `nodes.N.settings`) compiles cleanly, both test suites stay green, and the dialog's "Fix" button just silently stops resolving at runtime — no error, no failing test.

Suggest a single shared source of truth: export the path builders/constants from `editor.ts` (or a small shared module) and have both the producer and this matcher consume them, so drift is a compile error.

apps/web/modules/ee/workflows/components/inspector/workflow-email-action-form.tsx
```diff
@@ -243,26 +240,49 @@ export const WorkflowEmailActionForm = ({
...
       {/* Body (recall editor) */}
       {/* The editor defaults to a 2-line min-height (48px); a 4-line body (24px line-height)
           better matches the amount of content an email body usually holds. */}
-      <div className="flex flex-col gap-2 [--editor-min-height:96px]">
-        <Label>{t("workspace.workflows.email_body_label")}</Label>
+      <div className="flex flex-col gap-2 [--editor-min-height:96px]" ref={bodyWrapperRef}>
+        <WorkflowFieldLabel id="workflow-email-body-label" isRequired isInvalid={invalidFields.has("body")}>
+          {t("workspace.workflows.email_body_label")}
+        </WorkflowFieldLabel>
         <Editor
```
Dhruwang, line 255: **Body field's inline error isn't wired for assistive tech (P2, accessibility — also CodeRabbit's still-open comment).**

Every other field in this form associates its error: the Subject `Input` and the recipient/survey/endings `Select`s all set `aria-invalid` + `aria-describedby` pointing at their `WorkflowFieldError` id. The body is the exception — this `<Editor>` only forwards `isInvalid` (which just toggles the red border) and its `ContentEditable` sets only `aria-labelledby`, so the `WorkflowFieldError id="workflow-email-body-error"` rendered just below is referenced by nothing. A screen-reader user on a focused, empty, invalid body gets no "invalid" state and never hears the error.

Fix requires threading it through the shared editor component: add an `ariaDescribedBy` (and `aria-invalid`) prop to `TextEditorProps`, forward both to `ContentEditable` in `apps/web/modules/ui/components/editor/components/editor.tsx`, then pass `workflow-email-body-error` here when `invalidFields.has("body")` — mirroring the Subject/Send-to pattern.

Dhruwang (commented):
A few structural/accessibility notes from a review pass. None block the core fix (the shared blank-body rule is solid and well-tested) — details inline.

Dhruwang (changes requested):
Hey Javi, thanks for the PR overall looks good, but I noticed one horizontal scroll issue Can you please check on this ?

You can reproduce it like this as shown in the video: 

https://github.com/user-attachments/assets/e1de8ca7-244d-4872-80c0-8e5409b13e84

apps/web/modules/ee/workflows/components/inspector/workflow-email-action-form.tsx
itsjavi: Agreed — extracted as `useWorkflowNodeFieldFocus` in 07d9a0e56. Both forms now pass a field→element resolver and nothing else:

```ts
useWorkflowNodeFieldFocus({
  nodeId: node.id,
  onRequest: () => markTouched(...getBlankSendEmailContentFields(node.config)),
  resolveElement: (field) => field === "body" ? /* contenteditable */ : document.getElementById(...),
});
```

`onRequest` exists because the email step has to reveal its touched errors before the focus lands; the trigger form omits it. Callbacks are read through a ref (refreshed in an effect — writing one during render trips `react-hooks/refs`) so callers can pass inline closures without the effect re-running and re-focusing on every keystroke.

The hook doc also captures the ordering constraint that made the duplication dangerous, which is the real reason this shouldn't have been copied: the request must be cleared INSIDE the rAF, never before it. Clearing is what re-runs the effect, so an earlier clear lets the re-run's cleanup cancel the frame before it fires and focus silently never happens. I hit exactly that while writing the second copy.

On the visibility guard: intended, now documented. rAF does not fire in a backgrounded tab, so a jump queued just before the tab was hidden lands when it is next shown — which is what you want for something whose only job is to move the user's focus.

apps/web/modules/ee/workflows/lib/display.ts
itsjavi: Valid, and fixed in 07d9a0e56 — this was a real gap, not just a style point.

`state/editor.ts` now exports the paths as builders + a matcher, and both sides go through them:

```ts
export const workflowProblemFields = {
  name: "name",
  trigger: "trigger",
  triggerSurveyId: "trigger.config.surveyId",
  triggerEndingCardIds: "trigger.config.endingCardIds",
  nodeConfig: (nodeIndex: number) => `nodes.${nodeIndex}.config`,
  parseNodeConfigIndex: (field: string): number | null => { /* … */ },
} as const;
```

The producer emits `workflowProblemFields.nodeConfig(...)` instead of `path.slice(0, 3).join(".")`, and `display.ts` keys `TRIGGER_CONFIG_FIELD_TARGETS` off the same constants and matches via `parseNodeConfigIndex`.

You were also right that hardcoded literals in both test suites meant a rename stayed green. Rather than swap the literals for constants — which would have made the tests pass by construction and assert nothing — I added round-trip tests that feed the **producer's own output** back to the consumer:

```ts
const { problems } = deriveWorkflowValidation({ workflowName: "wf", definition, hasBoundTriggerSurvey: true });
const stepProblem = problems.find((p) => p.code === "step_incomplete");
expect(getWorkflowValidationProblemFocusTarget(stepProblem!, definition)).toEqual({ nodeId: "email-1", field: "to" });
```

Three of them, covering `step_incomplete`, `trigger_survey_unbound` and `trigger_ending_not_found`. A rename now either fails to compile or fails these, and the existing literal-based tests stay as independent assertions about the path format itself.

Writing them surfaced something worth noting: my first fixture used a partial `send_email` config, and the discriminated union then fails at `nodes.0` and reports `definition_invalid` rather than `step_incomplete` — so the fixture has to be otherwise-valid for the content check to be reached at all. Called out in a comment.

apps/web/modules/ee/workflows/components/inspector/workflow-email-action-form.tsx
itsjavi: Correct, and fixed in 07d9a0e56 along the route you outlined.

`TextEditorProps` gains an `ariaDescribedBy`, and `editor.tsx` forwards both it and `aria-invalid` to the `ContentEditable`:

```tsx
<ContentEditable
  aria-labelledby={props.id}
  aria-invalid={props.isInvalid ? true : undefined}
  aria-describedby={props.ariaDescribedBy}
  …
/>
```

The workflow form passes `workflow-email-body-error` when the body is invalid, mirroring Subject and Send to. Additive and optional, so the other `Editor` consumers (survey follow-ups included) are unaffected.

Verified in a real browser rather than by inspection — driving the editor through Playwright and reading the attributes off the contenteditable:

| state | aria-invalid | aria-describedby |
|---|---|---|
| body filled | `null` | `null` |
| body typed then emptied | `"true"` | `"workflow-email-body-error"` |

One related thing you may want to weigh in on: I also gave the Body label an id and pointed the editor's `aria-labelledby` at it. Before this PR that label was associated with nothing at all (a bare `<Label>` next to a contenteditable), so the field had no accessible name — a pre-existing gap this happened to sit next to.

itsjavi:
thanks @Dhruwang I fixed that

Dhruwang (approved):

PR #8687, opened 2026-07-29
feat(embedded-data): add EmbeddedData and SurveyEmbeddedData tables

pandeymangg:
## What & why

Embedded Data lives inside each survey today as JSON — `survey.variables` (keyed by cuid) and `survey.hiddenFields.fieldIds` (keyed by name) — so the same field has to be re-created by hand in every survey that needs it. This adds the storage that lets a field be defined once at workspace level and reused, which is the foundation the rest of Milestone 1 builds on.

Two new tables:

- **`EmbeddedData`** — one field definition, owned by a workspace. A field is either **local** (used by a single survey — how existing variables and hidden fields get migrated in ENG-1835) or **shared** (part of the workspace library). There is no stored flag for that: `surveyId` is the owner and `key` is the library name, so a local field has `surveyId` set and `key` null, and a shared field the other way round. `isLocalEmbeddedData()` derives it in one place.
- **`SurveyEmbeddedData`** — links a survey to a field and records the **`storageKey`** its value lives under inside that survey. This is the piece that makes the migration safe: a migrated field keeps the exact address its recall tokens, logic conditions and stored responses already use, so nothing has to be rewritten.

`source` (`computed` | `ingested` | `reserved`) replaces the old variable-vs-hidden-field split. `reserved` exists in the enum so every layer shares one type, but reserved fields are never stored as rows — they are a code catalog (ENG-1839), and `ZEmbeddedData` rejects the value for that reason.

Schema, types and validation only. **Nothing reads or writes these tables yet** — surveys keep using `hiddenFields` / `variables` until ENG-1837 repoints readers — and **no response data is touched**.

**Tenancy is enforced by the database, not by callers.** Every foreign key that could cross a workspace boundary carries `workspaceId`:

- `SurveyEmbeddedData` → `Survey(id, workspaceId)` and → `EmbeddedData(id, workspaceId)`
- `EmbeddedData` → `Survey(id, workspaceId)` for its owning survey

On their own, `surveyId` and `embeddedDataId` are independent, so nothing would stop a survey linking a field defined in another workspace, or a local field naming a survey from one. Referencing `(id, workspaceId)` makes a cross-workspace pair unrepresentable, and puts the workspace scope every query needs on the row instead of behind a join. `EmbeddedData.surveyId` is nullable and Postgres skips a multi-column foreign key when any of its columns is NULL, so shared fields are unaffected. `@@unique([id, workspaceId])` is the existing pattern here — the schema already uses it in three other places.

One thing worth a reviewer's eye: **`storageKey` carries exactly the rules `ZSurveyHiddenFields.fieldIds` already imposes on stored surveys, and no more.** A migrated field keeps its original address — a cuid for a computed field, a legacy name that may hold uppercase letters or hyphens for an ingested one — so anything stricter would fail the ENG-1835 backfill on a name a survey is using today.

- **In:** non-blank and `isLegacyIdCharset`. The survey load path already enforces that charset, so no survey that still loads can hold a name this rejects, and it rules out the case with a concrete failure: for an ingested field `storageKey` *is* the URL param name, so a padded `" plan "` never matches `?plan=` while `@@unique([surveyId, storageKey])` counts it as distinct from `"plan"` — one survey, two fields, one of which can never hold a value.
- **Out:** `isSafeIdentifier`, any length cap, and the reserved names. Reserved is the subtle one — `RESERVED_DECLARED_FIELD_NAMES` is `FORBIDDEN_IDS` **plus** the link-survey system params, so a survey stored before those params were reserved can still hold a hidden field named `lang` and still load. Rejecting it here would strand that survey mid-backfill.

The strict create-time rule stays on the create path (`isSafeIdentifier` from #8721, plus the reserved-name check). `key` — always newly authored — does go through it.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1833/schema-add-embeddeddata-surveyembeddeddata-tables

Part of Milestone 1 of the [Embedded Data](https://linear.app/formbricks/project/embedded-data-3d184a43d2f9) project. Targets the `epic/embedded-data-v1` branch — all of v1 (M1–M4) ships as one release. Rebased on #8721, so `isSafeIdentifier` now comes from the shared `@formbricks/types/safe-identifier` rather than a copy added here.

## How this was tested

**Unit — `pnpm test` in `packages/types`: 89 passed**, of which **45** are new in `embedded-data.test.ts`: local/shared derivation, the `key` naming rule, every reserved name in both casings, reserved-as-a-row rejection, the blank-`name` guard, the `locked` and computed-`dataType` guards, the full `defaultValue` × `dataType` matrix, and both sides of the `storageKey` rule.

**Integration — `pnpm test:integration` in `apps/web`: 83 passed across 22 files**, 6 of them new in `integration/embedded-data-tenancy.integration.test.ts`. The tenancy guarantee is pure DDL — `ZSurveyEmbeddedData` checks that `workspaceId` is present, never that the two foreign keys agree on it, and the unit suite mocks the database — so nothing but a real Postgres can see it. That harness already runs on `packages/**` changes and in the merge queue.

| Case | Result |
| --- | --- |
| Shared field with no owning survey | accepted |
| Local field owned by a survey in the same workspace | accepted |
| Local field owned by a survey in **another** workspace | rejected by `EmbeddedData_surveyId_workspaceId_fkey` |
| Survey links a field defined in another workspace | rejected by `SurveyEmbeddedData_embeddedDataId_workspaceId_fkey` |
| Link names a survey outside its own workspace | rejected by `SurveyEmbeddedData_surveyId_workspaceId_fkey` |
| Delete the owning survey | local field and its links cascade away, shared library field untouched |

Each rejection is asserted **by constraint name**, not just "it threw" — a later migration that recreates one of these single-column has to fail loudly. Checked that it does: weakening `SurveyEmbeddedData`'s second relation to a single column on the test database makes the cross-workspace link succeed and the suite fail with `Expected the write to be rejected, but it succeeded`. `prisma validate` and all 89 unit tests stay green through that same change, which is why these tests exist.

**Other checks**

- `pnpm typecheck` — clean in `packages/types` and `packages/database`
- `pnpm lint` — 0 errors (the 4 warnings in `packages/types` are pre-existing, in `surveys/*.ts`)
- `npx prisma validate` — schema valid; `prisma migrate diff` against the applied database is empty, so the migration and the models agree
- **Migration applied through the repo's runner** — both tables and enums created, migration recorded, 8 indexes and 4 cascading foreign keys match the models. Re-ran the runner: reports "already applied", nothing re-runs

## QA / Test Plan

**How to test**

There is nothing user-visible in this PR — it adds tables that no code reads or writes yet. The useful check is that nothing regressed around the fields it will eventually replace:

- [ ] Open a survey with variables and hidden fields in the editor → both cards load and list their fields exactly as before
- [ ] Add, rename and remove a hidden field and a variable, save → all succeed, no new errors
- [ ] Use a variable and a hidden field in recall (`@`) and in a logic condition, then submit a response → the values resolve as before
- [ ] Export that survey's responses → variable and hidden-field columns appear with the same headers and values as before
- [ ] Duplicate the survey → duplicate opens with the same fields
- [ ] Delete a survey that has variables and hidden fields → deletes cleanly, no foreign-key error in the logs

**Preconditions / test data**

- Any workspace with an existing survey that has at least one variable and one hidden field, plus a few responses collected before this change.

**Risks & regressions**

- Low: additive tables only, no reads/writes, no changes to survey or response payloads.
- The only shared-code touch is one file in `packages/types` plus one added `PrismaJson` type, so the area to re-check is that the app still builds and existing survey/response flows behave unchanged.
- The new file under `apps/web/integration/` is test-only and ships no runtime code. It is the first non-Better-Auth spec in that harness; it reuses `resetDb()` unchanged, and both new tables are FK-cascade-reachable from `Organization`, so the harness's isolation invariant still holds.

- One new schema migration: `20260806000000_add_embedded_data_tables` — creates two tables and two enums. Additive and non-breaking; no data is written or backfilled (that is ENG-1835).
- No env changes.

### DIFF
--- a/packages/database/migration/20260806000000_add_embedded_data_tables/migration.sql
+++ b/packages/database/migration/20260806000000_add_embedded_data_tables/migration.sql
@@ -1,14 +1,3 @@
--- ENG-1833: Embedded Data — workspace-level field definitions + per-survey links.
---
--- Additive only. Nothing reads or writes these tables yet: surveys keep using
--- Survey.hiddenFields / Survey.variables until ENG-1837 repoints readers, and no response
--- data is touched at any point.
---
--- EmbeddedData holds one field definition. A field is either local (bespoke to a single
--- survey, which is how existing variables and hidden fields get migrated in ENG-1835) or
--- shared (part of the workspace library). SurveyEmbeddedData links a survey to a field and
--- records the storage key its value lives under inside that survey.
-
 -- CreateEnum
 CREATE TYPE "EmbeddedDataSource" AS ENUM ('computed', 'ingested', 'reserved');
 
@@ -27,7 +16,6 @@ CREATE TABLE "EmbeddedData" (
     "dataType" "EmbeddedDataType" NOT NULL DEFAULT 'string',
     "defaultValue" JSONB,
     "locked" BOOLEAN NOT NULL DEFAULT false,
-    "isLocal" BOOLEAN NOT NULL DEFAULT true,
     "surveyId" TEXT,
     "workspaceId" TEXT NOT NULL,
 
@@ -37,6 +25,7 @@ CREATE TABLE "EmbeddedData" (
 -- CreateTable
 CREATE TABLE "SurveyEmbeddedData" (
     "id" TEXT NOT NULL,
+    "workspaceId" TEXT NOT NULL,
     "surveyId" TEXT NOT NULL,
     "embeddedDataId" TEXT NOT NULL,
     "storageKey" TEXT NOT NULL,
@@ -45,18 +34,13 @@ CREATE TABLE "SurveyEmbeddedData" (
 );
 
 -- CreateIndex
-CREATE INDEX "EmbeddedData_workspaceId_idx" ON "EmbeddedData"("workspaceId");
-
--- CreateIndex
 CREATE INDEX "EmbeddedData_surveyId_idx" ON "EmbeddedData"("surveyId");
 
 -- CreateIndex
--- Only shared fields carry a key. Postgres treats NULLs as distinct, so any number of local
--- fields can coexist in a workspace while the shared library itself stays de-duplicated.
 CREATE UNIQUE INDEX "EmbeddedData_workspaceId_key_key" ON "EmbeddedData"("workspaceId", "key");
 
 -- CreateIndex
-CREATE INDEX "SurveyEmbeddedData_surveyId_idx" ON "SurveyEmbeddedData"("surveyId");
+CREATE UNIQUE INDEX "EmbeddedData_id_workspaceId_key" ON "EmbeddedData"("id", "workspaceId");
 
 -- CreateIndex
 CREATE INDEX "SurveyEmbeddedData_embeddedDataId_idx" ON "SurveyEmbeddedData"("embeddedDataId");
@@ -65,18 +49,16 @@ CREATE INDEX "SurveyEmbeddedData_embeddedDataId_idx" ON "SurveyEmbeddedData"("em
 CREATE UNIQUE INDEX "SurveyEmbeddedData_surveyId_embeddedDataId_key" ON "SurveyEmbeddedData"("surveyId", "embeddedDataId");
 
 -- CreateIndex
--- No two fields inside one survey may share a storage key, which stops a newly linked field
--- from colliding with an existing one and overwriting its response values.
 CREATE UNIQUE INDEX "SurveyEmbeddedData_surveyId_storageKey_key" ON "SurveyEmbeddedData"("surveyId", "storageKey");
 
 -- AddForeignKey
-ALTER TABLE "EmbeddedData" ADD CONSTRAINT "EmbeddedData_surveyId_fkey" FOREIGN KEY ("surveyId") REFERENCES "Survey"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE "EmbeddedData" ADD CONSTRAINT "EmbeddedData_surveyId_workspaceId_fkey" FOREIGN KEY ("surveyId", "workspaceId") REFERENCES "Survey"("id", "workspaceId") ON DELETE CASCADE ON UPDATE CASCADE;
 
 -- AddForeignKey
 ALTER TABLE "EmbeddedData" ADD CONSTRAINT "EmbeddedData_workspaceId_fkey" FOREIGN KEY ("workspaceId") REFERENCES "Workspace"("id") ON DELETE CASCADE ON UPDATE CASCADE;
 
 -- AddForeignKey
-ALTER TABLE "SurveyEmbeddedData" ADD CONSTRAINT "SurveyEmbeddedData_surveyId_fkey" FOREIGN KEY ("surveyId") REFERENCES "Survey"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE "SurveyEmbeddedData" ADD CONSTRAINT "SurveyEmbeddedData_surveyId_workspaceId_fkey" FOREIGN KEY ("surveyId", "workspaceId") REFERENCES "Survey"("id", "workspaceId") ON DELETE CASCADE ON UPDATE CASCADE;
 
 -- AddForeignKey
-ALTER TABLE "SurveyEmbeddedData" ADD CONSTRAINT "SurveyEmbeddedData_embeddedDataId_fkey" FOREIGN KEY ("embeddedDataId") REFERENCES "EmbeddedData"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE "SurveyEmbeddedData" ADD CONSTRAINT "SurveyEmbeddedData_embeddedDataId_workspaceId_fkey" FOREIGN KEY ("embeddedDataId", "workspaceId") REFERENCES "EmbeddedData"("id", "workspaceId") ON DELETE CASCADE ON UPDATE CASCADE;
--- a/packages/database/schema/main.prisma
+++ b/packages/database/schema/main.prisma
@@ -405,6 +405,8 @@ model Survey {
   pin                             String?
   displayPercentage               Decimal?
   languages                       SurveyLanguage[]
+  embeddedData                    EmbeddedData[]
+  embeddedDataLinks               SurveyEmbeddedData[]
   showLanguageSwitch              Boolean?
   followUps                       SurveyFollowUp[]
   /// [SurveyRecaptcha]
@@ -617,6 +619,7 @@ model Workspace {
   clickOutsideClose           Boolean                      @default(true)
   overlay                     SurveyOverlay                @default(none)
   languages                   Language[]
+  embeddedData                EmbeddedData[]
   /// [Logo]
   logo                        Json?
   workspaceTeams              WorkspaceTeam[]
@@ -1141,6 +1144,88 @@ model SurveyLanguage {
 
   @@id([languageId, surveyId])
   @@index([surveyId])
+}
+
+/// Where an Embedded Data field's value comes from.
+/// `computed` is set by survey logic, `ingested` arrives from a URL param or the SDK, and
+/// `reserved` is auto-captured system metadata. Reserved fields are never stored as rows —
+/// they are a static catalog in code — but the value exists so every layer shares one type.
+enum EmbeddedDataSource {
+  computed
+  ingested
+  reserved
+}
+
+/// The declared type of an Embedded Data field's value.
+enum EmbeddedDataType {
+  string
+  number
+  boolean
+  date
+}
+
+/// Defines one Embedded Data field: the extra context stored on a response that isn't an
+/// answer to a question (brand, plan, page type, a computed score, ...).
+///
+/// A field is either **local** — bespoke to a single survey, which is how existing variables
+/// and hidden fields are migrated — or **shared**, meaning it is part of the workspace library
+/// and can be linked by many surveys. Those states move together:
+/// `isLocal = true` <=> `surveyId` is set <=> `key` is null.
+///
+/// @property key - Library name, unique per workspace and immutable. Null for local fields
+/// @property name - Human-readable display label
+/// @property source - Where the value comes from (computed | ingested)
+/// @property dataType - Declared value type, used for coercion at ingest
+/// @property defaultValue - Initial value (computed) or fallback when nothing arrives (ingested)
+/// @property locked - Ingested only: ignore external writes and use the default instead
+/// @property isLocal - True when the field belongs to one survey rather than the shared library
+/// @property survey - The owning survey for a local field; cascades so nothing is orphaned
+model EmbeddedData {
+  id          String   @id @default(cuid())
+  createdAt   DateTime @default(now()) @map(name: "created_at")
+  updatedAt   DateTime @updatedAt @map(name: "updated_at")
+  key         String?
+  name        String
+  description String?
+
+  source       EmbeddedDataSource
+  dataType     EmbeddedDataType   @default(string)
+  /// [EmbeddedDataDefaultValue]
+  defaultValue Json?
+  locked       Boolean            @default(false)
+  isLocal      Boolean            @default(true)
+
+  survey      Survey?              @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  surveyId    String?
+  workspace   Workspace            @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
+  workspaceId String
+  surveyLinks SurveyEmbeddedData[]
+
+  // Only shared fields carry a key, and Postgres treats NULLs as distinct, so any number of
+  // local fields can coexist in a workspace while the library itself stays de-duplicated.
+  @@unique([workspaceId, key])
+  @@index([workspaceId])
+  @@index([surveyId])
+}
+
+/// Links a survey to an Embedded Data field and records the key that field's value is stored
+/// under inside that survey.
+///
+/// @property storageKey - The key the value lives under: a cuid for computed fields, the field
+/// name for ingested ones. Migrated fields keep their original key, which is what lets existing
+/// recall tokens, logic conditions and stored responses keep resolving untouched.
+model SurveyEmbeddedData {
+  id             String       @id @default(cuid())
+  survey         Survey       @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  surveyId       String
+  embeddedData   EmbeddedData @relation(fields: [embeddedDataId], references: [id], onDelete: Cascade)
+  embeddedDataId String
+  storageKey     String
+
+  @@unique([surveyId, embeddedDataId])
+  @@unique([surveyId, storageKey])
+  @@index([surveyId])
+  @@index([embeddedDataId])
 }
 
 /// Represents a team within an organization.
--- a/packages/database/schema/main.prisma
+++ b/packages/database/schema/main.prisma
@@ -1167,18 +1167,20 @@ enum EmbeddedDataType {
 /// Defines one Embedded Data field: the extra context stored on a response that isn't an
 /// answer to a question (brand, plan, page type, a computed score, ...).
 ///
-/// A field is either **local** — bespoke to a single survey, which is how existing variables
-/// and hidden fields are migrated — or **shared**, meaning it is part of the workspace library
-/// and can be linked by many surveys. Those states move together:
-/// `isLocal = true` <=> `surveyId` is set <=> `key` is null.
+/// A field is either **local** — used by a single survey, which is how existing variables and
+/// hidden fields are migrated — or **shared**, meaning it is part of the workspace library and can
+/// be linked by many surveys. There is no stored flag for that: `surveyId` is the owner and `key`
+/// is the library name, so a local field has `surveyId` set and `key` null, and a shared field the
+/// other way round. Deriving it rather than storing a boolean removes a column that could drift
+/// out of step with the two fields it stands for.
 ///
 /// @property key - Library name, unique per workspace and immutable. Null for local fields
 /// @property name - Human-readable display label
-/// @property source - Where the value comes from (computed | ingested)
+/// @property source - Where the value comes from (computed | ingested). `reserved` is part of the
+/// shared enum but never stored as a row: those fields are a static catalog in code
 /// @property dataType - Declared value type, used for coercion at ingest
 /// @property defaultValue - Initial value (computed) or fallback when nothing arrives (ingested)
 /// @property locked - Ingested only: ignore external writes and use the default instead
-/// @property isLocal - True when the field belongs to one survey rather than the shared library
 /// @property survey - The owning survey for a local field; cascades so nothing is orphaned
 model EmbeddedData {
   id          String   @id @default(cuid())
@@ -1193,9 +1195,12 @@ model EmbeddedData {
   /// [EmbeddedDataDefaultValue]
   defaultValue Json?
   locked       Boolean            @default(false)
-  isLocal      Boolean            @default(true)
 
-  survey      Survey?              @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  // Composite for the same reason SurveyEmbeddedData's are: on its own, `surveyId` would let a
+  // local field name a survey from a different workspace. `surveyId` is nullable and Postgres
+  // skips a multi-column foreign key when any of its columns is NULL, so shared fields (no owning
+  // survey) are unaffected.
+  survey      Survey?              @relation(fields: [surveyId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
   surveyId    String?
   workspace   Workspace            @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
   workspaceId String
@@ -1204,27 +1209,35 @@ model EmbeddedData {
   // Only shared fields carry a key, and Postgres treats NULLs as distinct, so any number of
   // local fields can coexist in a workspace while the library itself stays de-duplicated.
   @@unique([workspaceId, key])
-  @@index([workspaceId])
+  // Target for SurveyEmbeddedData's composite foreign key, which is what stops a survey linking
+  // a field definition from another workspace.
+  @@unique([id, workspaceId])
   @@index([surveyId])
 }
 
 /// Links a survey to an Embedded Data field and records the key that field's value is stored
 /// under inside that survey.
 ///
+/// `workspaceId` is part of both foreign keys on purpose: `surveyId` and `embeddedDataId` alone are
+/// independent, so nothing would stop a survey in one workspace linking a field defined in another.
+/// Referencing `(id, workspaceId)` on both sides makes a cross-workspace pair unrepresentable, and
+/// puts the workspace scope every query needs on the row instead of behind a join through Survey.
+///
 /// @property storageKey - The key the value lives under: a cuid for computed fields, the field
 /// name for ingested ones. Migrated fields keep their original key, which is what lets existing
 /// recall tokens, logic conditions and stored responses keep resolving untouched.
 model SurveyEmbeddedData {
-  id             String       @id @default(cuid())
-  survey         Survey       @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  id             String @id @default(cuid())
+  workspaceId    String
   surveyId       String
-  embeddedData   EmbeddedData @relation(fields: [embeddedDataId], references: [id], onDelete: Cascade)
   embeddedDataId String
   storageKey     String
 
+  survey       Survey       @relation(fields: [surveyId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
+  embeddedData EmbeddedData @relation(fields: [embeddedDataId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
+
   @@unique([surveyId, embeddedDataId])
   @@unique([surveyId, storageKey])
-  @@index([surveyId])
   @@index([embeddedDataId])
 }
 
--- a/packages/types/embedded-data.ts
+++ b/packages/types/embedded-data.ts
@@ -1,5 +1,5 @@
 import { z } from "zod";
-import { isSafeIdentifier } from "./safe-identifier";
+import { isLegacyIdCharset, isSafeIdentifier } from "./safe-identifier";
 import { RESERVED_DECLARED_FIELD_NAMES } from "./surveys/validation";
 
 /** Longest allowed field key / display name. */
@@ -67,7 +67,13 @@ export const ZEmbeddedData = z
       // `lang`, and ingestion would then refuse to fill it — a field that can never hold a value.
       .refine((key) => !RESERVED_DECLARED_FIELD_NAMES.has(key.toLowerCase()), "Key is reserved")
       .nullable(),
-    name: z.string().min(1).max(EMBEDDED_DATA_NAME_MAX_LENGTH),
+    // Blank rather than empty: `name` is the label the library and the editor render, and a
+    // whitespace-only one draws a row with nothing to read or click. Checked, not trimmed, so the
+    // stored value is never silently rewritten.
+    name: z
+      .string()
+      .refine((value) => value.trim().length > 0, "Name must not be blank")
+      .max(EMBEDDED_DATA_NAME_MAX_LENGTH),
     description: z.string().nullable(),
     source: ZEmbeddedDataSource,
     dataType: ZEmbeddedDataType.prefault("string"),
@@ -164,19 +170,32 @@ export const isLocalEmbeddedData = (field: Pick<TEmbeddedData, "surveyId">): boo
  * `workspaceId` is part of both foreign keys in the schema, so a survey cannot link a field defined
  * in another workspace.
  *
- * `storageKey` is deliberately **not** validated as a safe identifier, has no length cap, and does
- * not exclude reserved names: migrated fields keep their original address, which for a computed
- * field is a cuid and for an ingested field can be a legacy name with uppercase letters or hyphens.
- * Any restriction here that `ZSurveyHiddenFields.fieldIds` doesn't also impose would fail the
- * ENG-1835 backfill on a name a survey is currently using. The strict rule belongs on the create
- * path — see `RESERVED_DECLARED_FIELD_NAMES` and the third enforcement point on ENG-1839.
+ * `storageKey` holds exactly the restrictions `ZSurveyHiddenFields.fieldIds` already imposes on
+ * stored surveys, and no more. Migrated fields keep their original address — a cuid for a computed
+ * field, and for an ingested field a legacy name that may carry uppercase letters or hyphens — so
+ * anything stricter would fail the ENG-1835 backfill on a name a survey is using today.
+ *
+ * What that leaves in: `isLegacyIdCharset`, because the load path enforces it, so no survey that
+ * still loads can hold a name outside it. That rules out the case with a concrete failure — for an
+ * ingested field `storageKey` is the URL param name, so a padded `" plan "` never matches `?plan=`
+ * while `@@unique([surveyId, storageKey])` counts it as distinct from `"plan"`, giving one survey
+ * two fields where one can never hold a value.
+ *
+ * What that leaves out: `isSafeIdentifier`, any length cap, and the reserved names. Reserved is the
+ * subtle one — `RESERVED_DECLARED_FIELD_NAMES` is `FORBIDDEN_IDS` plus the link-survey system
+ * params, so a survey stored before those params were reserved can still hold a hidden field named
+ * `lang` and still load. Rejecting it here would strand that survey mid-backfill. The strict rule
+ * belongs on the create path instead — see the enforcement points on ENG-1839.
  */
 export const ZSurveyEmbeddedData = z.object({
   id: z.cuid2(),
   workspaceId: z.cuid2(),
   surveyId: z.cuid2(),
   embeddedDataId: z.cuid2(),
-  storageKey: z.string().refine((key) => key.trim().length > 0, "Storage key must not be blank"),
+  storageKey: z
+    .string()
+    .refine((key) => key.trim().length > 0, "Storage key must not be blank")
+    .refine(isLegacyIdCharset, "Storage key must contain only letters, digits, hyphens and underscores"),
 });
 
 export type TSurveyEmbeddedData = z.infer<typeof ZSurveyEmbeddedData>;
--- a/packages/types/embedded-data.ts
+++ b/packages/types/embedded-data.ts
@@ -1,7 +1,8 @@
 import { z } from "zod";
 import { isSafeIdentifier } from "./safe-identifier";
+import { RESERVED_DECLARED_FIELD_NAMES } from "./surveys/validation";
 
-/** Longest allowed field key / name / storage key. */
+/** Longest allowed field key / display name. */
 export const EMBEDDED_DATA_NAME_MAX_LENGTH = 255;
 
 /**
@@ -9,8 +10,10 @@ export const EMBEDDED_DATA_NAME_MAX_LENGTH = 255;
  *
  * - `computed` — set inside the survey by logic (what used to be a "variable")
  * - `ingested` — arrives from outside via URL param or SDK (what used to be a "hidden field")
- * - `reserved` — auto-captured system metadata. Read-only, and never stored as a row:
- *   reserved fields are a static catalog in code (ENG-1839), globally available to every survey.
+ * - `reserved` — auto-captured system metadata. Read-only, globally available to every survey, and
+ *   **never stored as a row**: reserved fields are a static catalog in code (ENG-1839). The value
+ *   lives in this enum so the resolver and the catalog share one vocabulary; `ZEmbeddedData`
+ *   rejects it, because that schema describes a stored row.
  */
 export const ZEmbeddedDataSource = z.enum(["computed", "ingested", "reserved"]);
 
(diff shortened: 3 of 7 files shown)


pandeymangg:
@coderabbitai please review

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,141 @@
...
+    key: z
+      .string()
+      .max(EMBEDDED_DATA_NAME_MAX_LENGTH)
+      .refine(isSafeIdentifier, "Key must start with a lowercase letter and contain only a-z, 0-9 and _")
+      .nullable(),
+    name: z.string().min(1).max(EMBEDDED_DATA_NAME_MAX_LENGTH),
+    description: z.string().nullable(),
+    source: ZEmbeddedDataSource,
+    dataType: ZEmbeddedDataType.prefault("string"),
+    defaultValue: ZEmbeddedDataDefaultValue,
```
itsjavi, line 80: **`defaultValue` is never checked against `dataType`.**

`dataType` is documented as *"declared value type, used for coercion at ingest"*, but the two fields are validated completely independently, so every one of these parses successfully today:

| input | result |
| --- | --- |
| `dataType: "number"`, `defaultValue: "abc"` | ✅ accepted |
| `dataType: "boolean"`, `defaultValue: 0` | ✅ accepted |
| `dataType: "date"`, `defaultValue: "banana"` | ✅ accepted |
| `dataType: "string"`, `defaultValue: true` | ✅ accepted |

(ran against this branch's `ZEmbeddedData`, not inferred from reading)

This is the layer that's supposed to catch it — the ticket's acceptance criteria list validation as the guard, and there's no DB check constraint behind it. A row with `dataType: "number"` / `defaultValue: "abc"` will surface as a wrong value in `resolveEmbeddedValue()` (ENG-1836) or as a coercion failure at ingest (ENG-1845), far from where it was written.

`date` is the sharper edge: `ZEmbeddedDataDefaultValue` is `string | number | boolean | null`, so a date default can only ever be an unvalidated string. Either forbid `defaultValue` on `date` outright, or pin the format.

Suggest discriminating the default on `dataType` in the `superRefine`:

```ts
if (field.defaultValue !== null) {
  const expected = { string: "string", number: "number", boolean: "boolean" }[field.dataType];
  if (field.dataType === "date") {
    ctx.addIssue({ code: "custom", message: "Date fields cannot have a default value", path: ["defaultValue"] });
  } else if (typeof field.defaultValue !== expected) {
    ctx.addIssue({
      code: "custom",
      message: `A ${field.dataType} field's default value must be a ${expected}`,
      path: ["defaultValue"],
    });
  }
}
```

---

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,141 @@
...
+    // Library name — workspace-unique and immutable. Null for local fields, which
+    // aren't listed in the shared library and therefore have no library name.
+    key: z
+      .string()
+      .max(EMBEDDED_DATA_NAME_MAX_LENGTH)
+      .refine(isSafeIdentifier, "Key must start with a lowercase letter and contain only a-z, 0-9 and _")
+      .nullable(),
+    name: z.string().min(1).max(EMBEDDED_DATA_NAME_MAX_LENGTH),
+    description: z.string().nullable(),
+    source: ZEmbeddedDataSource,
```
itsjavi, line 78: **`ZEmbeddedData` accepts `source: "reserved"`, which the docs say can never exist as a row.**

Three places in this PR state that reserved fields are never stored — the docstring above (`reserved` … *"never stored as a row"*), the Prisma enum comment, and the migration header. But the row validator accepts one. Verified on this branch:

```
ACCEPTED  source=reserved as a LOCAL row   (isLocal: true, surveyId set, key: null)
ACCEPTED  source=reserved as a SHARED row  (isLocal: false, surveyId: null, key: "country")
```

`embedded-data.test.ts` actually locks this in — [`rejects a default value on a reserved field`](https://github.com/formbricks/formbricks/pull/8687/files#diff-packages/types/embedded-data.test.ts) asserts only that `defaultValue` is rejected, which implies the reserved row itself is otherwise fine.

Validation is the only guard here (no DB constraint), so the row schema should refuse the value the catalog owns:

```ts
if (field.source === "reserved") {
  ctx.addIssue({
    code: "custom",
    message: "Reserved fields are a code catalog and are never stored as rows",
    path: ["source"],
  });
}
```

That also makes the `defaultValue`-on-reserved guard below redundant, and it makes the intent unambiguous for ENG-1839 when the catalog lands. If you'd rather keep one schema for both, split it: `ZEmbeddedDataRow` (computed | ingested) vs. a separate catalog type that reuses `ZEmbeddedDataSource` — right now the shared enum is doing double duty and the stricter half isn't expressed anywhere.

Worth noting the local/shared invariant is meaningless for a reserved field too: it would have to either own a survey or claim a workspace-unique key, and neither is true of globally-available system metadata.

---

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * `storageKey` is deliberately **not** validated as a safe identifier: migrated fields keep
+ * their original address, which for a computed field is a cuid and for an ingested field can
+ * be a legacy name containing uppercase letters or hyphens.
+ */
+export const ZSurveyEmbeddedData = z.object({
+  id: z.cuid2(),
+  surveyId: z.cuid2(),
+  embeddedDataId: z.cuid2(),
+  storageKey: z.string().min(1).max(EMBEDDED_DATA_NAME_MAX_LENGTH),
```
itsjavi: **`storageKey` doesn't exclude `FORBIDDEN_IDS`, so it's a weaker guard than the `hiddenFields.fieldIds` it replaces.**

Agreed on not applying `isSafeIdentifier` here — the comment above explains that well, and the `Brand-Name` test case is the right call. But `ZSurveyHiddenFields.fieldIds` ([`packages/types/surveys/types.ts:165`](https://github.com/formbricks/formbricks/blob/main/packages/types/surveys/types.ts#L165)) also rejects `FORBIDDEN_IDS`, and that check didn't come along. Verified on this branch:

```
ACCEPTED  storageKey='userId'         ACCEPTED  storageKey='start'
ACCEPTED  storageKey='source'         ACCEPTED  storageKey='end'
ACCEPTED  storageKey='suid'           ACCEPTED  storageKey='embed'
ACCEPTED  storageKey='verifiedEmail'
```

For an ingested field `storageKey` *is* the URL param name — `survey-client-wrapper.tsx:126-133` builds `hiddenFieldsRecord` with `searchParams.get(field)` for each field id. So:

- `storageKey: "verifiedEmail"` is silently overwritten ~4 lines later, where `getVerifiedEmail` merges `{ verifiedEmail }` into the same record.
- `suid` / `embed` / `verify` / `start` / `end` collide with the link-survey URL contract; `getPrefillValue` ([`prefill/index.ts:48`](https://github.com/formbricks/formbricks/blob/main/apps/web/modules/survey/link/lib/prefill/index.ts#L48)) skips those keys outright, so the field is simply never ingestable.

`@@unique([surveyId, storageKey])` can't catch any of this — the colliding name isn't another row.

```ts
storageKey: z
  .string()
  .min(1)
  .max(EMBEDDED_DATA_NAME_MAX_LENGTH)
  .refine((v) => !FORBIDDEN_IDS.includes(v), "Storage key is reserved"),
```

Two smaller things in the same field while you're here:

- `.min(1)` doesn't trim, so `storageKey: "  "` is accepted. `.trim().min(1)` or a non-blank refine.
- `EMBEDDED_DATA_NAME_MAX_LENGTH` (255) is applied to `storageKey`, but `ZSurveyHiddenFields.fieldIds` has no length cap at all. A legacy hidden field id longer than 255 chars would fail the backfill and break the *"migrated fields keep the exact address"* promise. Unlikely in practice, but it's a silent 500 in ENG-1835 rather than a caught case, and the cap buys nothing here.

---

packages/database/schema.prisma
```diff
@@ -1138,6 +1141,88 @@ model SurveyLanguage {
...
+  key         String?
+  name        String
+  description String?
+
+  source       EmbeddedDataSource
+  dataType     EmbeddedDataType   @default(string)
+  /// [EmbeddedDataDefaultValue]
+  defaultValue Json?
+  locked       Boolean            @default(false)
+  isLocal      Boolean            @default(true)
```
itsjavi: **`isLocal` is fully derivable from `surveyId`, and the invariant tying them together only exists in Zod.**

`isLocal = true ⟺ surveyId != null ⟺ key = null` is enforced in `ZEmbeddedData.superRefine` and nowhere else. Every writer that reaches these tables next bypasses that: the ENG-1835 backfill, and the ENG-1978 editor write bridge, both go through Prisma/raw SQL, which never runs the Zod schema. So the column can drift, and both drift directions are silent and unrecoverable-by-constraint:

- **`isLocal = true`, `surveyId = NULL`** → an orphan. Not in the shared library (no `key`), owned by no survey, so `onDelete: Cascade` never reaches it. It stays in the workspace forever, invisible to both the `surveyId` lookup and the library query.
- **`isLocal = false`, `key = NULL`** → a "shared" row with no name. `@@unique([workspaceId, key])` can't dedupe it (NULLs distinct, as the comment below correctly notes), so you can accumulate unbounded unnamed library entries.

Cheapest fix is to delete the column and derive it — `isLocal` carries zero information that `surveyId IS NULL` doesn't, and a derived boolean stored alongside its source is a drift generator by construction. `surveyId === null` at the read boundary, or a getter on the domain type.

If you want to keep it (the ticket does specify it), then put the invariant in the migration while both tables are still empty:

```sql
ALTER TABLE "EmbeddedData" ADD CONSTRAINT "EmbeddedData_local_shared_check" CHECK (
  ("isLocal" = true  AND "surveyId" IS NOT NULL AND "key" IS NULL) OR
  ("isLocal" = false AND "surveyId" IS NULL     AND "key" IS NOT NULL)
);
```

Fair warning: there are currently zero `CHECK` constraints anywhere in `packages/database/migration/`, and Prisma doesn't model them, so this shows up as schema drift on `prisma migrate dev`. That's the trade-off — and it's the reason I'd lean toward dropping the column instead. Either way this is a decision to make now: adding it after ENG-1835 has backfilled means a migration plus a cleanup pass over real rows.

---

packages/database/schema.prisma
```diff
@@ -1138,6 +1141,88 @@ model SurveyLanguage {
...
+///
+/// @property storageKey - The key the value lives under: a cuid for computed fields, the field
+/// name for ingested ones. Migrated fields keep their original key, which is what lets existing
+/// recall tokens, logic conditions and stored responses keep resolving untouched.
+model SurveyEmbeddedData {
+  id             String       @id @default(cuid())
+  survey         Survey       @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  surveyId       String
+  embeddedData   EmbeddedData @relation(fields: [embeddedDataId], references: [id], onDelete: Cascade)
+  embeddedDataId String
```
itsjavi: **Nothing constrains the two FKs to the same workspace — a survey can link another workspace's field definition.**

`surveyId` and `embeddedDataId` are independent foreign keys. `Survey` has a `workspaceId`, `EmbeddedData` has a `workspaceId`, and nothing requires them to match. A link row pairing survey `S ∈ workspace A` with field `F ∈ workspace B` satisfies every constraint in this migration — both unique indexes are `surveyId`-scoped, so neither notices.

That's a cross-tenant read of a field definition and its `defaultValue`, in a table whose whole job is deciding which fields a survey resolves. AGENTS.md is explicit: *"Multi-tenancy: All data must be scoped by Organization or Environment."*

Same shape, second instance: for a **local** field, `EmbeddedData.surveyId` and `SurveyEmbeddedData.surveyId` both name a survey, and nothing forces them to be the same one. Survey B can link survey A's private local field — the field that ENG-1835 will migrate someone's existing variable into.

Both are enforceable at the DB level, and it's nearly free while these tables are empty:

```prisma
model Survey {
  // ...
  @@unique([id, workspaceId])
}

model EmbeddedData {
  // ...
  @@unique([id, workspaceId])
}

model SurveyEmbeddedData {
  id             String @id @default(cuid())
  workspaceId    String
  surveyId       String
  embeddedDataId String
  storageKey     String

  survey       Survey       @relation(fields: [surveyId, workspaceId],       references: [id, workspaceId], onDelete: Cascade)
  embeddedData EmbeddedData @relation(fields: [embeddedDataId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
  // ...
}
```

Now a mismatched pair can't be inserted at all, and the workspace scope every query needs is on the row instead of requiring a join through `Survey`.

If you'd rather not widen the table, that's defensible — but then the invariant needs to be a stated, tested service-layer rule rather than an unwritten one, because `ZSurveyEmbeddedData` structurally cannot express it (it never sees a `workspaceId`). Either way I'd settle it in this PR: retrofitting a composite FK after ENG-1835 means a backfill of `workspaceId` plus reconciling whatever rows already violate it.

---

packages/database/schema/main.prisma
```diff
@@ -1138,6 +1141,88 @@ model SurveyLanguage {
...
+  survey         Survey       @relation(fields: [surveyId], references: [id], onDelete: Cascade)
+  surveyId       String
+  embeddedData   EmbeddedData @relation(fields: [embeddedDataId], references: [id], onDelete: Cascade)
+  embeddedDataId String
+  storageKey     String
+
+  @@unique([surveyId, embeddedDataId])
+  @@unique([surveyId, storageKey])
+  @@index([surveyId])
+  @@index([embeddedDataId])
```
itsjavi, line 1241: **Two of the nine indexes are redundant.**

A Postgres btree index on `(a, b)` already serves lookups on `a` alone, so:

- `@@index([surveyId])` here is fully covered by *both* `@@unique([surveyId, embeddedDataId])` and `@@unique([surveyId, storageKey])`.
- `@@index([workspaceId])` on `EmbeddedData` (line 1202) is fully covered by `@@unique([workspaceId, key])`.

`@@index([embeddedDataId])` is genuinely needed — it's the only index leading with that column, and the FK reverse lookup and cascade delete both need it. Keep that one.

Two dead indexes on tables that will get a row per field per survey is pure write amplification for no read benefit, and it's cheaper to not create them than to drop them later. Worth dropping from both the model and the migration SQL.

---

packages/types/package.json
```diff
@@ -12,6 +12,7 @@
   "scripts": {
     "lint": "eslint .",
     "typecheck": "tsc --noEmit",
+    "test": "vitest run",
```
itsjavi, line 15: Minor, and the ask is to drop this rather than extend it: packages/types does need a test runner, but PR 8721 (ENG-1834, the blocker this PR sits on) already adds one, and more of one than this does. It ships test, test:coverage and test:watch scripts, the vitest and @vitest/coverage-v8 devDependencies at 4.1.6, and a packages/types/vitest.config.mts. Its PR body makes the same observation independently: packages/types had no test runner at all, so tests placed there would have silently never run.

This PR's version is a strict subset of that, on the same base commit, so the two conflict. Drop the package.json change and the vitest devDependency here, along with the pnpm-lock.yaml hunks that exist only to add it. Then embedded-data.test.ts runs on the inherited config with nothing added by this PR.

One thing worth recording so nobody re-derives it: the coverage angle is moot either way. packages/types sits outside sonar.sources, and the types glob is in sonar.coverage.exclusions, so the 0.0 percent Coverage on New Code in the SonarQube comment above is expected and not something this PR introduced. The test.yml workflow runs pnpm test via turbo run test, which is what actually gates these tests on every PR.

Formatting note: this comment was edited through Linear's diff sync, which stores plain text and escapes markdown, so it is unformatted by necessity. The reply below carries the same content properly formatted.

Generated by Claude Code https://claude.ai/code

itsjavi (changes requested):
Reviewed with the Embedded Data spec and ENG-1833 alongside. The shape is right and a few things are notably well handled: `storageKey` as a separate address from `key` is the piece that makes ENG-1835 safe, deliberately *not* running `isSafeIdentifier` over it is correct (legacy hidden field ids allow uppercase and hyphens — `/^[a-zA-Z0-9_-]+$/`), the NULLs-distinct reasoning on `@@unique([workspaceId, key])` is accurate, and `updated_at` with no DB default matches every other new table here. I confirmed the 21 tests pass and that the merged schema passes `prisma validate`.

Six things I'd want fixed before this lands. All of them are cheap now and expensive after ENG-1835 backfills real rows.

**Correctness — validation is the only guard here, and it has holes**

1. **`defaultValue` is never checked against `dataType`.** `number`/`"abc"`, `boolean`/`0`, `date`/`"banana"` all parse. `date` has no representable default at all.
2. **`source: "reserved"` is accepted as a stored row**, which all three docstrings in this PR say can never exist. The test suite implicitly locks the gap in.
3. **`storageKey` doesn't exclude `FORBIDDEN_IDS`**, unlike the `hiddenFields.fieldIds` validation it replaces. For an ingested field `storageKey` *is* the URL param name, so `verifiedEmail` gets silently overwritten and `suid`/`embed`/`start`/`end` are never ingestable.

**Schema integrity — decisions that get expensive later**

4. **`isLocal` is derivable from `surveyId`** and the invariant binding them lives only in Zod, which no writer on the roadmap goes through. Both drift directions produce rows no constraint can catch.
5. **Nothing scopes `SurveyEmbeddedData` to one workspace** — a survey can link another workspace's field definition, and survey B can link survey A's local field.
6. **Two of the nine indexes are redundant** with the unique indexes that already lead with the same column.

Details and suggested diffs are inline. #4 and #5 are the ones I'd settle before merge rather than defer, since they're structural — the rest are contained in `packages/types`.

**Three notes, no action needed in this PR**

- Linear has ENG-1833 **blocked by** ENG-1834, but this lands ahead of it, so `isSafeIdentifier` now exists twice (`packages/types/safe-identifier.ts` and `apps/web/lib/utils/safe-identifier.ts`, byte-identical logic). The PR body owns this and defers the consolidation — flagging only so the duplicate doesn't outlive ENG-1834.
- Legacy **variable names** are validated with `/^[a-z0-9_]+$/` ([`types.ts:210`](https://github.com/formbricks/formbricks/blob/main/packages/types/surveys/types.ts#L210)), which permits a leading digit or underscore, while `isSafeIdentifier` requires `/^[a-z]/`. Harmless for migrated fields (local ⇒ `key` is null), but a field named `1st_page` or `_score` can never be promoted to the shared library. Worth a known-limitation line in ENG-1835 or a normalization step at promotion time.
- The base branch has moved since this was opened — `epic/embedded-data-v1` was fast-forwarded to main (`f0234aa`), which split `schema.prisma` into `schema/main.prisma` + `schema/workflows.prisma` and added `prisma.config.ts` for Prisma 7. GitHub reports the PR as not cleanly mergeable, though `git merge-tree` resolves it via rename detection and the result validates, so the rebase should be mechanical. Re-run `pnpm install` afterwards — the lockfile hunks here predate that upgrade.

---

packages/types/embedded-data.ts
itsjavi: **Correcting this one — I reviewed without #8721 in scope, and it changes the remark substantially.**

**Retracting the suggested fix.** Adding `!FORBIDDEN_IDS.includes(v)` to `ZSurveyEmbeddedData.storageKey` would be wrong, and would break ENG-1835.

#8721 splits the rule by *when* a name is created, and `storageKey` is squarely on the load side. It keeps `ZSurveyHiddenFields` lenient precisely so already-stored surveys holding `Verify` or `UserId` keep parsing — it closes the capture hole at ingestion instead of rewriting stored surveys. Those surveys are exactly what ENG-1835 has to backfill. A row schema that rejects reserved names would fail on them, breaking the *"migrated fields keep the exact address"* guarantee that `storageKey` exists to provide. **Lenient here is correct as written.**

**My cited evidence is also stale.** #8721 replaces the inline `hiddenFieldsRecord` loop in `survey-client-wrapper.tsx` with [`getHiddenFieldsFromSearchParams`](https://github.com/formbricks/formbricks/pull/8721/files#diff-apps/web/modules/survey/link/lib/hidden-fields.ts), which skips any param whose lowercased key is in `RESERVED_DECLARED_FIELD_NAMES`. So the `?verify=` token capture and the `verifiedEmail` overwrite are already fixed there — not this PR's problem, and my `verifiedEmail` line was describing code that's on its way out.

**What still stands, redirected**

#8721 establishes strict-on-create / lenient-on-load with `RESERVED_DECLARED_FIELD_NAMES` as *"the single source of truth shared by the two ends that have to agree"* — and it's broader than `FORBIDDEN_IDS`, adding `sutoken`, `lang`, `preview`, `startat`, `skipprefilled`, `offlinesupport`, matched case-insensitively. This PR ships only the lenient half. Two consequences:

1. **`key` (line 46) is a create-time name and should use the reserved set.** It's the workspace library name — always newly authored, never migrated, since local fields carry `key: null`. `isSafeIdentifier("verify")` passes, so a shared field can be keyed `verify`, `lang`, `startat` or `sutoken`. For an ingested field the name *is* the address, so that field can never capture a value — the exact silent-empty-field failure #8721's reserved set was created to prevent, reintroduced one layer up:

   ```ts
   key: z
     .string()
     .max(EMBEDDED_DATA_NAME_MAX_LENGTH)
     .refine(isSafeIdentifier, "Key must start with a lowercase letter and contain only a-z, 0-9 and _")
     .refine((k) => !RESERVED_DECLARED_FIELD_NAMES.has(k.toLowerCase()), "Key is reserved")
     .nullable(),
   ```

2. **There's no strict counterpart for a newly created `storageKey`.** When ENG-1978 / ENG-1837 mint new ingested fields, nothing schema-level stops a reserved name, and whoever writes that check will re-derive the list by hand. Either add the strict create-side schema here, or at minimum say in the `ZSurveyEmbeddedData` docstring that this is deliberately the lenient half and point at `RESERVED_DECLARED_FIELD_NAMES` for the create path — the current comment explains the `isSafeIdentifier` exemption but not that a strict sibling is still owed.

The two smaller notes are unaffected: `.min(1)` doesn't trim (`storageKey: "  "` is accepted), and the 255 cap on `storageKey` has no counterpart on `ZSurveyHiddenFields.fieldIds`, so an over-long legacy id would fail the backfill.

---

packages/types/package.json
itsjavi: **Withdrawing this — #8721 already does it, and this change should be dropped rather than extended.**

[#8721's `packages/types/package.json`](https://github.com/formbricks/formbricks/pull/8721/files#diff-packages/types/package.json) already ships everything I was asking for, and more:

```json
"test": "vitest run",
"test:coverage": "vitest run --coverage",
"test:watch": "vitest",
```

plus `@vitest/coverage-v8: 4.1.6` and `vitest: 4.1.6` in `devDependencies`, and a `packages/types/vitest.config.mts`. Its PR body makes the same observation independently — *"`packages/types` had no test runner at all … tests placed there would have silently never run."*

So this PR's version is a strict subset of #8721's, on the same base commit, and the two will conflict. Once #8721 lands, **delete this `package.json` change and the `vitest` devDependency from this PR** — `embedded-data.test.ts` runs on the inherited config with nothing added here. Same for the `pnpm-lock.yaml` hunks, which exist only to add `vitest` to this package.

---

itsjavi:
## Correction to my review above — #8721 was not in scope, and it changes some of it

Thanks for the pointer. I reviewed this PR against `epic/embedded-data-v1` as it stands, and #8721 (ENG-1834, the blocker Linear records) isn't merged into that base yet, so I treated the validation layer as if it were complete. GitHub review bodies can't be edited, so the corrections are here and inline on the two affected threads.

**Retracted**

- **`storageKey` should reject reserved names** — [thread](https://github.com/formbricks/formbricks/pull/8687#discussion_r3721064676). My suggested fix would have broken ENG-1835. #8721 deliberately keeps the load path lenient so surveys already holding `Verify` / `UserId` keep parsing, and fixes the capture at ingestion instead. Rejecting reserved names in a row schema would fail the backfill on exactly those surveys. Lenient here is right, and the `?verify=` / `verifiedEmail` failure modes I cited are already closed by `getHiddenFieldsFromSearchParams`. Redirected to a narrower, still-real point about `key`.
- **Missing `test:coverage`** — [thread](https://github.com/formbricks/formbricks/pull/8687#discussion_r3721072407). #8721 already adds `test`, `test:coverage`, `test:watch`, `@vitest/coverage-v8` and a `vitest.config.mts`. This PR's version should be deleted, not extended.

**Unaffected — #8721 touches none of this, and all five still stand**

`defaultValue` never cross-checked against `dataType` (CodeRabbit found this independently at `embedded-data.ts:121`) · `source: "reserved"` accepted as a stored row · `isLocal` derivable + invariant only in Zod · no workspace scoping on `SurveyEmbeddedData` · two redundant indexes.

Worth being precise about one thing I said in the summary, since #8721 sharpens it rather than dissolving it: *"validation is the only guard here"* is still accurate for these five. #8721 fixes **naming-rule** validation — it doesn't add a `dataType`/`defaultValue` cross-check, doesn't constrain `source`, and adds no DB constraints. The four findings under "Schema integrity" and "Correctness" #1–#2 sit entirely in files #8721 never opens.

**Upgraded: three of this PR's eight files collide with #8721**

Both PRs branch from the same base (`0922e4e`), and #8721 is the further-along one — it's out of draft and already rebased.

| file | this PR | #8721 |
| --- | --- | --- |
| `packages/types/safe-identifier.ts` | adds, 17 lines, 1 export | adds, 120 lines, 7 exports |
| `packages/types/package.json` | +3 −1 | +6 −1 |
| `pnpm-lock.yaml` | +4 −4 | +6 −3 |

I flagged the `safe-identifier.ts` duplication as a process nit — that undersold it. It's a same-path collision, and #8721 also **deletes** `apps/web/lib/utils/safe-identifier.ts`, so there's no version of this where both files survive. #8721's copy is a superset: it adds `isLegacyFieldIdentifier`, `isLegacyVariableName` and `matchDeclaredFieldName`, and fixes a double-space bug in `formatSnakeCaseToTitleCase`. **All three files should come out of this PR once #8721 lands** — `embedded-data.ts`'s `import { isSafeIdentifier } from "./safe-identifier"` resolves unchanged against #8721's module, so nothing else needs touching.

**One note I can withdraw**

The legacy-variable-name gap I raised (`/^[a-z0-9_]+$/` permits `_score` / `1st_page`, `isSafeIdentifier` doesn't) is exactly what #8721's `isLegacyVariableName` shim exists for — I was describing a problem it already solves. The forward-looking half survives on its own terms: such a variable migrates fine as a local field (`key` is null), but can never be promoted to a shared library `key`, since `key` uses the strict rule. Still worth a known-limitation line in ENG-1835, and now that #8721 has named the strict/lenient split explicitly there's a clean place to hang it.

---

packages/database/schema.prisma
pandeymangg: Fixed — and the same gap was on `EmbeddedData.surveyId`, where a local field could name a survey from another workspace. Both sides are composite now:

- `SurveyEmbeddedData` → `Survey(id, workspaceId)` and → `EmbeddedData(id, workspaceId)`
- `EmbeddedData` → `Survey(id, workspaceId)`

`surveyId` is nullable and Postgres skips a multi-column FK when any of its columns is NULL, so shared fields are unaffected. Both rejections verified against the DB, plus that deleting a survey still cascades its local field and links while leaving shared library fields alone.

packages/database/schema.prisma
pandeymangg: Dropped. Derived from `surveyId` now, with an `isLocalEmbeddedData()` helper so call sites don't each re-derive it.

packages/types/embedded-data.ts
pandeymangg: Fixed. `superRefine` checks the runtime type of `defaultValue` against `dataType`, and a `date` default has to be an ISO 8601 date or datetime string. Full matrix in the tests.

packages/types/embedded-data.ts
pandeymangg: Fixed. `ZEmbeddedData` rejects `source: "reserved"` in either shape.

packages/types/embedded-data.ts
pandeymangg: Kept it lenient, and dropped the 255 cap as well — a migrated computed field's key is a cuid, an ingested one can be a legacy name with caps or hyphens, and a stored survey may already hold a hidden field named `verify`. Anything stricter here than `ZSurveyHiddenFields.fieldIds` breaks the ENG-1835 backfill on a name that works today. Only guard is non-blank.

The strict rule stays on the create path. `key` is always newly authored, so it does go through `isSafeIdentifier` plus a `RESERVED_DECLARED_FIELD_NAMES` check — otherwise a shared ingested field could be keyed `verify` and then never be fillable.

packages/database/schema/main.prisma
pandeymangg: Dropped both.

packages/types/package.json
pandeymangg: Dropped in the rebase — this now takes #8721's version of `packages/types`.

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    // The logic engine can only calculate strings and numbers.
+    if (field.source === "computed" && field.dataType !== "string" && field.dataType !== "number") {
+      ctx.addIssue({
+        code: "custom",
+        message: "Computed fields support only string or number",
+        path: ["dataType"],
+      });
+    }
```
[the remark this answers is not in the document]
pandeymangg: Fixed — `defaultValue` is now cross-checked against `dataType`, with `date` pinned to ISO 8601.

packages/database/schema/main.prisma
```diff
@@ -1138,6 +1141,88 @@ model SurveyLanguage {
...
+/// `isLocal = true` <=> `surveyId` is set <=> `key` is null.
+///
+/// @property key - Library name, unique per workspace and immutable. Null for local fields
+/// @property name - Human-readable display label
+/// @property source - Where the value comes from (computed | ingested)
+/// @property dataType - Declared value type, used for coercion at ingest
+/// @property defaultValue - Initial value (computed) or fallback when nothing arrives (ingested)
+/// @property locked - Ingested only: ignore external writes and use the default instead
+/// @property isLocal - True when the field belongs to one survey rather than the shared library
+/// @property survey - The owning survey for a local field; cascades so nothing is orphaned
```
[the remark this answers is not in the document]
pandeymangg: Fixed.

packages/database/schema/main.prisma
```diff
@@ -1138,6 +1141,88 @@ model SurveyLanguage {
...
+  workspace   Workspace            @relation(fields: [workspaceId], references: [id], onDelete: Cascade)
+  workspaceId String
+  surveyLinks SurveyEmbeddedData[]
+
+  // Only shared fields carry a key, and Postgres treats NULLs as distinct, so any number of
+  // local fields can coexist in a workspace while the library itself stays de-duplicated.
+  @@unique([workspaceId, key])
+  @@index([workspaceId])
+  @@index([surveyId])
+}
```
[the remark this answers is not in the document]
pandeymangg: Fixed, both dropped.

packages/database/migration/20260806000000_add_embedded_data_tables/migration.sql
```diff
@@ -0,0 +1,82 @@
...
+    "source" "EmbeddedDataSource" NOT NULL,
+    "dataType" "EmbeddedDataType" NOT NULL DEFAULT 'string',
+    "defaultValue" JSONB,
+    "locked" BOOLEAN NOT NULL DEFAULT false,
+    "isLocal" BOOLEAN NOT NULL DEFAULT true,
+    "surveyId" TEXT,
+    "workspaceId" TEXT NOT NULL,
+
+    CONSTRAINT "EmbeddedData_pkey" PRIMARY KEY ("id")
+);
```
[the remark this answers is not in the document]
pandeymangg: `isLocal` is gone, so the constraint as written no longer applies.

Not adding the remaining half (exactly one of `surveyId` / `key`) either: Prisma can't declare a CHECK, so it would live only in raw SQL and would silently not exist on databases created with `db:push`. The repo has 0 CHECK constraints across 240+ migrations, so this isn't the place to introduce the first one. The invariant stays in `ZEmbeddedData`.

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * One Embedded Data field definition, owned by a workspace.
+ *
+ * A field is either **local** (bespoke to a single survey) or **shared** (part of the
+ * workspace library). Those two states move together with `surveyId` and `key`:
+ * `isLocal === true` ⟺ `surveyId !== null` ⟺ `key === null`.
+ */
+export const ZEmbeddedData = z
+  .object({
+    id: z.cuid2(),
```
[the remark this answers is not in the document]
pandeymangg: Out of scope. `z.cuid2()` is loose about length, but it rejects real garbage, accepts every real id, and is what every id in the repo already uses. Tightening it is a repo-wide change, not this PR's.

itsjavi (changes requested):
thanks for the fixes, claude had other comments, can you check those?

packages/database/schema/main.prisma
```diff
@@ -1143,6 +1146,101 @@ model SurveyLanguage {
...
+/// recall tokens, logic conditions and stored responses keep resolving untouched.
+model SurveyEmbeddedData {
+  id             String @id @default(cuid())
+  workspaceId    String
+  surveyId       String
+  embeddedDataId String
+  storageKey     String
+
+  survey       Survey       @relation(fields: [surveyId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
+  embeddedData EmbeddedData @relation(fields: [embeddedDataId, workspaceId], references: [id, workspaceId], onDelete: Cascade)
```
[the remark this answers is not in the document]
pandeymangg: Fair, and the revert-stays-green check settled it. Went with option 1: `apps/web/integration/embedded-data-tenancy.integration.test.ts`, 6 tests — the four accept/reject cases plus the cascade split (owning survey's local field and links go, shared library field stays).

Each rejection asserts the **constraint name**, not just that it threw, so a migration that recreates one single-column fails loudly. Verified it actually catches that: weakening `SurveyEmbeddedData`'s second relation on the test DB makes the cross-workspace link succeed and the suite fail with `Expected the write to be rejected, but it succeeded`, while `prisma validate` and all 89 unit tests stay green.

On the harness fit — it's the first non-Better-Auth spec there, but it needed no changes: `resetDb()` truncates `Organization` CASCADE and both tables are FK-reachable from it, so the isolation invariant holds. `integration-tests.yml` already triggers on `packages/**` and in the merge queue, so it runs on this PR and on ENG-1835/1836/1837/1978. Full suite: 83 passed across 22 files.

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,182 @@
...
+    key: z
+      .string()
+      .max(EMBEDDED_DATA_NAME_MAX_LENGTH)
+      .refine(isSafeIdentifier, "Key must start with a lowercase letter and contain only a-z, 0-9 and _")
+      // A library key is always newly authored (local fields carry `key: null`), so it goes through
+      // the strict create-time rule. Without this a shared ingested field could be keyed `verify` or
+      // `lang`, and ingestion would then refuse to fill it — a field that can never hold a value.
+      .refine((key) => !RESERVED_DECLARED_FIELD_NAMES.has(key.toLowerCase()), "Key is reserved")
+      .nullable(),
+    name: z.string().min(1).max(EMBEDDED_DATA_NAME_MAX_LENGTH),
```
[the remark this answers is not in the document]
pandeymangg: Fixed. `.refine((value) => value.trim().length > 0)`, kept non-mutating as suggested so the stored value isn't rewritten. Tests cover empty, spaces and a tab, plus one asserting `" Score "` parses through untrimmed.

packages/types/embedded-data.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Any restriction here that `ZSurveyHiddenFields.fieldIds` doesn't also impose would fail the
+ * ENG-1835 backfill on a name a survey is currently using. The strict rule belongs on the create
+ * path — see `RESERVED_DECLARED_FIELD_NAMES` and the third enforcement point on ENG-1839.
+ */
+export const ZSurveyEmbeddedData = z.object({
+  id: z.cuid2(),
+  workspaceId: z.cuid2(),
+  surveyId: z.cuid2(),
+  embeddedDataId: z.cuid2(),
+  storageKey: z.string().refine((key) => key.trim().length > 0, "Storage key must not be blank"),
```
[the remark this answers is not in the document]
pandeymangg: Checked the load-path claim rather than taking it, and it holds: `ZSurveyHiddenFields.fieldIds` runs `isLegacyIdCharset` with the comment *"this schema also parses surveys loaded from the database"*, and `ZSurveyVariable.id` is `z.cuid2()`, so both address shapes the backfill can move already conform. Added the charset refine. The `" plan "` case is the one that sold it — never matches `?plan=`, and the unique index counts it as distinct from `"plan"`.

Reserved names stay allowed, and you were right that it's the narrower point: `RESERVED_DECLARED_FIELD_NAMES` is `FORBIDDEN_IDS` **plus** `LINK_SURVEY_SYSTEM_PARAMS`, so a survey stored before those params were reserved can hold a hidden field named `lang` and still load today. Rejecting it here would strand that survey mid-backfill.

Rewrote the docstring — you were right that it read as "no restriction is possible here". It now says what's in, what's out, and why for each.

pandeymangg:
All three of Claude's comments are addressed — details in the threads.

- **Composite FK had no test.** Added one: `apps/web/integration/embedded-data-tenancy.integration.test.ts`, 6 cases, each rejection asserted by **constraint name**. Confirmed it catches a revert — weakening the link's second relation to a single column makes the cross-workspace link succeed and the suite fail, while `prisma validate` and every unit test stay green.
- **`storageKey` was too lenient.** Now also holds `isLegacyIdCharset`. Checked the premise: the survey load path already enforces that charset on hidden field ids and variable ids are cuids, so nothing the ENG-1835 backfill can move is excluded. Rules out `" plan "`, which for an ingested field would never match `?plan=` while the unique index counts it as distinct. Reserved names stay allowed — that set is `FORBIDDEN_IDS` **plus** the link-survey system params, so a survey stored before those were reserved can hold a `lang` hidden field and still load.
- **Blank `name` was accepted.** `z.string().min(1)` counts characters, not content. Fixed, non-mutating.

Unit 89 passed, integration 83 across 22 files, full CI green. Ready for another look.

itsjavi (approved):
thanks LGTM

PR #8689, opened 2026-07-29
feat(license): add Workflows to the enterprise licensed features list

jobenjada:
## What & why

The self-hosted **Organization → Settings → Enterprise → Licensed Features** table is data-driven from the enterprise license feature schema (`TEnterpriseLicenseFeatures`). It listed every enterprise feature except Workflows.

This adds **Workflows** as a licensed feature so the table shows its access state based on the enterprise license server's response, exactly like every other row:

- New `workflows` boolean (default `false`) in both license feature Zod schemas (`ZEnterpriseLicenseFeatures` in `types/enterprise-license.ts`, `LicenseFeaturesSchema` in `lib/license.ts`) and in `DEFAULT_FEATURES`. Using `.default(false)` keeps older license-server responses that don't yet send the flag parsing safely (row renders as "Disabled").
- New row in `EnterpriseLicenseFeaturesTable`, with the docs link pointing to `/docs/workflows/overview`.
- New `workspace.settings.enterprise.workflows` en-US translation key.

Scope is intentionally minimal and **self-hosted only** — the Enterprise settings page `notFound()`s on Cloud. No Stripe entitlement/billing wiring and no `getIsWorkflows…` permission helper are added, because no product code gates a Workflows feature yet and that would be dead code. The access badge reuses the table's existing **Enabled / Disabled** wording for consistency with every other row; it flips purely on the license server's `workflows` value.

Licensed Features table with the new Workflows row

*Faithful render of the table component. The **NEW** badge and row highlight are annotations for this PR, not part of the UI. The live page requires a self-hosted instance with an active enterprise license (it 404s on Cloud).*

## Linear ticket

No ticket — direct request to surface Workflows in the licensed features list.

## How this was tested

- `pnpm scan-translations` ✅ — "All translation validations passed": 0 missing / unused / incomplete keys across all 15 web locales, and `i18n.lock` valid.
- Unit tests (run with the self-hosted `.env`):
  - `modules/ee/license-check/lib/license.test.ts` ✅ **33 passed**
  - `modules/ee/license-check/lib/utils.test.ts` ✅ **37 passed**
- `eslint` on changed source files ✅, `prettier --check` on changed files ✅.

> Non-English locale files carry `"workflows": "Workflows"` as a placeholder to satisfy the translation gate. Per the repo's en-US-only workflow, Lingo.dev replaces these with real translations on its next run.

## QA / Test Plan

**How to test**

- [ ] On a **self-hosted** instance with an **active enterprise license**, sign in as an owner/admin and open **Organization → Settings → Enterprise** (`/organizations/{organizationId}/settings/enterprise`) → the **Licensed Features** table now includes a **Workflows** row.
- [ ] Workflows **Access** badge shows **Enabled** when the license server returns `workflows: true`, and **Disabled** when it returns `false` or omits the flag → expected: badge reflects the license response.
- [ ] Click the Workflows **Read docs** link → opens `https://formbricks.com/docs/workflows/overview` in a new tab.
- [ ] Edge — **Formbricks Cloud**: the Enterprise settings page still returns **404** (unchanged).
- [ ] Edge — instance with **no license**: page shows the upgrade prompt; the Licensed Features table is not rendered (unchanged).

**Preconditions / test data**

- Self-hosted Formbricks with `ENTERPRISE_LICENSE_KEY` set and the license server reachable.
- Owner or admin account (the page 404s for members).

**Risks & regressions**

- Low; additive change to a data-driven table. Re-check that the Licensed Features table renders all existing rows plus Workflows, and that license parsing still succeeds for responses that omit `workflows` (covered by `default(false)`).

- none

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/enterprise/components/EnterpriseLicenseFeaturesTable.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/enterprise/components/EnterpriseLicenseFeaturesTable.tsx
@@ -91,6 +91,11 @@ const getFeatureDefinitions = (t: TFunction): TFeatureDefinition[] => {
       labelKey: t("workspace.settings.general.ai_smart_tools_enabled"),
       docsUrl: "https://formbricks.com/docs/self-hosting/configuration/ai",
     },
+    {
+      key: "workflows",
+      labelKey: t("workspace.settings.enterprise.workflows"),
+      docsUrl: "https://formbricks.com/docs/workflows/overview",
+    },
   ];
 };
 
--- a/apps/web/i18n.lock
+++ b/apps/web/i18n.lock
@@ -2560,6 +2560,7 @@ checksums:
     workspace/settings/enterprise/unlock_the_full_power_of_formbricks_free_for_30_days: a96f8df102a5d1f0367f3f82ac1f344b
     workspace/settings/enterprise/white_glove_onboarding: 79d5f04ad3decd659e63f5cf61628ea0
     workspace/settings/enterprise/whitelabel_email_follow_ups: 4d62c861b60c5dceb6ad72cd5d1186d7
+    workspace/settings/enterprise/workflows: b0c9c8615a9ba7d9cb73e767290a7f72
     workspace/settings/feedback_directories/archive: fa813ab3074103e5daad07462af25789
     workspace/settings/feedback_directories/archive_directory: c2b39feb0962b93f883a4bac6ffadd1e
     workspace/settings/feedback_directories/archive_not_allowed: 929b16f54260e321b0ebf5d458499d60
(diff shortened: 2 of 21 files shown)


xernobyl (approved):
Approving — the license plumbing and the table row look right, and the rest is translation churn. 👍

Two non-blocking notes for later:

**1. The non-English `"workflows": "Workflows"` placeholders won't be replaced by Lingo.** `i18n.lock` fingerprints the *en-US source* string, so once `workspace/settings/enterprise/workflows` is stamped, Lingo treats the key as done and the English targets stay frozen. That same hash (`b0c9c861…`) already exists on `epic/workflows-v1` for `common/workflows` and `workspace.settings.billing.comparison_row_workflows`, where it resolves to ワークフロー / 工作流 / Рабочие процессы — so the Enterprise row would end up as the one English label in a table whose other rows are all localized.

Targeted lock-bust whenever it's convenient:

```bash
cd apps/web && sed -i '' '/^    workspace\/settings\/enterprise\/workflows: /d' i18n.lock && cd ../.. && pnpm i18n:web:generate
```

In practice only ja / zh-Hans / zh-Hant / ru (and probably hu / ro) are actually wrong — for de/nl/sv/tr/es/fr/pt "Workflows" is a reasonable loanword, so those placeholders may be accidentally correct.

**2. The row will read "Disabled" for every self-hosted customer** until the license server learns the flag — `ee/` has no `workflows` field yet (nor `feedbackDirectories` / `dashboards`, so those two existing rows are in the same boat). Worth a small `ee/` follow-up once Workflows ships; not this PR's problem.

PR #8690, opened 2026-07-29
chore: merge main into epic/workflows-v1

itsjavi:
## What & why

`epic/workflows-v1` was 37 commits behind `main`. This syncs `main` into the epic branch so the epic
builds against current tooling (ESLint 9 flat config, the Prisma 7 facade, main's job registry) rather
than drifting further.

**Conflict resolutions (12 files)**

Most conflicts were both-sides-added collisions in the same registry — main added a
`survey-archive-purge` job while the epic added `workflow-run` jobs. Both were kept:

- `packages/jobs/src/{constants,types,definitions,index,queue}.ts` — both job families registered.
- `apps/web/instrumentation-jobs.ts` + test — both handlers and recurring schedules registered.
- `apps/web/package.json` — both sides' dependencies kept; `pnpm-lock.yaml` regenerated from the
  merged manifests (`pnpm install --frozen-lockfile` passes).
- `docs/api-v3-reference/src/openapi.yml` — main's survey archive/restore paths alongside the epic's
  workflow paths; the generated `openapi.yml` artifact was re-bundled (`pnpm api:v3:check` ✅).

One conflict needed a real decision:

- `packages/database/src/scripts/migration-runner.ts` — main replaced the per-migration
  `runSchemaMigration` with a batched `runSchemaMigrationBatch`, whose batch-level `allApplied` check
  subsumes the per-migration `copyOnly` flag it replaced. Took main's version, and kept the epic's
  package-local `prisma.config.ts` path.

**Follow-up commit: repairs the sync surfaced**

`main`'s tooling changes broke epic-only code that main's own migrations never touched:

- `packages/workflows` had no `eslint.config.mjs` and still used `eslint . --ext`, which ESLint 9
  rejects. Migrated it to flat config like its siblings. Its one finding — a registry read only via
  `typeof` — is now a type instead of a dead runtime value.
- `use-workflow-builder.ts` wrote a ref during render, which main's `eslint-plugin-react-hooks` bump
  now rejects (`react-hooks/refs`). Moved the flush-on-unmount closure into a dependency-less effect
  so it still refreshes on every commit; the existing unmount test covers it.

Two of the repairs are **latent bugs that predate this merge** and fail on `epic/workflows-v1` today:

- `packages/database/zod/workflows.ts` imported its model types from `@prisma/client`, which exports
  no model types under Prisma 7 — so `packages/database` typecheck fails on the epic branch. Fixed to
  import from `../src/prisma` like all 12 sibling zod modules, and added the four Workflow models to
  that facade's explicit re-export list, where they were missing.
- With that import fixed, the `satisfies z.ZodType<WorkflowRun>` constraint became live and caught a
  missing `dispatchedAt` field — added to the Prisma model in #8578 but never mirrored into the zod
  schema. The broken import had been silently disabling the check.

## Linear ticket

No ticket — routine branch sync to keep the epic current with `main`.

## How this was tested

- `pnpm turbo run typecheck` — **25/25 passed** ✅ (fails on `epic/workflows-v1` without the
  Prisma-facade fix above)
- `pnpm turbo run lint` — **20/20 passed, 0 errors** ✅ (remaining warnings are pre-existing and
  present on `main`)
- `pnpm turbo run test` — **24/24 tasks, 570 test files passed** ✅
- `pnpm install --frozen-lockfile` — passes, confirming the regenerated lockfile matches the merged
  manifests (this is what CI enforces)
- `pnpm api:v3:check` — bundled OpenAPI artifact is in sync with `src/` ✅
- Spot-checked that the merged `@formbricks/jobs` build exports both job families, and that both
  sides' dependencies resolve in the lockfile.

No UI change — no screenshots.

Note: two of my early local failures were stale build artifacts, not merge defects — a `packages/jobs`
`dist/` from before the merge, and a `database:build`/`database:typecheck` race that cleans
`generated/prisma` mid-typecheck. Both clear after a rebuild; results above are from clean serialized
runs.

## QA / Test Plan

**How to test**

This is a branch sync with no intended behavior change, so QA is regression-focused on the two areas
the resolutions touched.

- [ ] Open the workflow editor for an existing workflow, edit a step, and navigate away (browser back
      or tab switch) before the autosave debounce fires → the edit is persisted; reopening the
      workflow shows the change. This exercises the reworked flush-on-unmount.
- [ ] Edit a workflow and hard-refresh the page while unsaved → the browser's native "leave site?"
      confirm still appears.
- [ ] Open a workflow in read-only mode (archived workflow, or a user without edit rights), navigate
      away → no save is attempted and no error toast appears.
- [ ] Archive a survey, then confirm it disappears from the survey list and stops accepting responses
      → main's soft-delete feature still works after the merge.
- [ ] Trigger a workflow from a completed survey response → the run appears and completes. Confirms
      both job families coexist in the queue registry.
- [ ] `GET /api/v3/workflows` and `POST /api/v3/surveys/{surveyId}/archive` both respond → both sides'
      v3 routes survived the OpenAPI merge.

**Preconditions / test data**

- Workflows are Enterprise-licensed (EE) — needs a license/entitlement that enables them.
- Redis must be reachable for the background job queue, or the workflow-run and archive-purge
  schedules won't register.
- A workspace with at least one survey that has responses, for the trigger path.

**Risks & regressions**

- Highest risk is the background job registry: `survey-archive-purge` and `workflow-run` schedules
  were merged into shared registration code, so a mistake would silently drop one of them. Worth
  confirming both recurring schedules register on boot (they log on startup).
- Second is the Prisma migration runner, where main's batched deploy replaced per-migration deploys.
  A migration run against a non-empty database is the check that matters.
- `dispatchedAt` is now part of the `ZWorkflowRun` schema; anything validating a workflow run without
  that field would newly reject. It is non-nullable-optional in the DB (`DateTime?`), so existing rows
  are fine.

- No new migrations authored here, but this branch inherits `main`'s pending migrations (survey
  soft-delete/archive, interaction-based segment filters). Run `pnpm db:migrate:dev` after checkout.
- No new or changed env vars.


xernobyl (approved):
Clean union merge — nothing lost from either side as far as I can tell.

I reconstructed the mechanical merge and diffed it against the commit to isolate the 15 files carrying actual decisions, and reviewed those. Job registry and `instrumentation-jobs.ts` keep both families (all 5 handlers, all 3 recurring schedules), the OpenAPI paths are a proper union, and `main`'s two migrations sort after every epic one so there's no out-of-order hazard. Also checked that no dependency went backwards — `apps/web/package.json` is additions-only vs `main`, and the regenerated lockfile has no downgrades, so the CVE bumps are all intact.

Since CI doesn't run `typecheck`, I verified the repair locally: `packages/database` is clean on this head and fails with exactly the described errors when I swap the epic's `zod/workflows.ts` back, so that latent break was real. `react-hooks/refs` does reject the old render-phase ref write too, and the unmount-flush test still passes.

One non-blocking thing for later: `@formbricks/database#typecheck` has no build-order edge to `@formbricks/workflows#build` in `turbo.json`, so on a cold cache it can race and fail with "Cannot find module '@formbricks/workflows'" (hit it once). Pre-existing, but it makes the typecheck result order-dependent now that it actually passes.

Small note: `ZWorkflowRun` has no importers, so the `dispatchedAt` risk in the description has no callers — probably nothing to chase during QA. 👍

Dhruwang (approved):

PR #8691, opened 2026-07-29
fix(auth): send the verification email on invite sign-up and bind invites to the invited address [ENG-2091]

xernobyl:
## What & why

A user who was invited to a workspace, clicked the invite and created an account never got the verification email. Reproduced on staging, then locally against real Better Auth + Postgres. It turned out to be **two independent causes with the same symptom**, plus an auth hole sitting one line away from both.

**1. A duplicate email doesn't throw.** `signUpUserSafely` classified "this address is already taken" by catching an error from `auth.api.signUpEmail`. Better Auth never throws there: with `requireEmailVerification || autoSignIn === false` — `auth.ts` sets both — it takes an enumeration-safe branch that hashes the password for timing parity and returns **HTTP 200 carrying a synthetic user** (generated id, no row written, and, because it returns before the send, no verification email).

So `userAlreadyExisted` was always `false`, and every post-creation side effect ran against the *pre-existing* account from an unauthenticated endpoint: invite accepted and the invite row deleted, organization + owner membership + workspace created (Cloud, where multi-org is enabled), the user's locale overwritten from attacker-supplied input, mailing-list subscription, and a `created` audit event attributed to that user. Meanwhile the person saw "check your inbox" for an email that was never sent, in front of a resend button that returns early for an already-verified address. A complete dead end.

Now classified on the id Better Auth hands back — a synthetic id is never in the database. The `catch` stays, because flipping either option makes Better Auth throw `USER_ALREADY_EXISTS` instead, so both signals have to map to already-existed. `signUpUserSafely` returns a discriminated union and `handlePostUserCreation` accepts only the `"created"` variant, so dropping that guard is a compile error rather than a silent privilege hole.

This dates from the Better Auth cutover (#8384), i.e. **5.2.0 onward** — 1.6.18 and 1.6.23 are byte-identical here, so the recent bump isn't the trigger. 5.1.x (NextAuth) is unaffected. Worth noting `autoSignIn: false` alone is enough, so self-hosters running `EMAIL_VERIFICATION_DISABLED=1` are affected too.

**2. A failed send reported success.** Two ways it disappeared: a **throw** (on sign-up Better Auth calls the callback through `runInBackgroundOrAwait`, whose catch only logs, so `signUpEmail` resolves 200 regardless), and a **falsy return** (`sendEmail` returns `false` without throwing when SMTP isn't configured, and Better Auth ignores the return value — so on an instance with no SMTP, *every* verification email "succeeded"). Both are failures now, logged attributably (userId + email domain only, never the address, token, or verify URL) and rethrown. Since Better Auth swallows that throw on sign-up, the outcome comes back through a request scope and the form shows "we couldn't send it, try resend" instead. Same one-line gap fixed in `sendResetPassword`.

**3. Invite acceptance wasn't bound to the invited address (ENG-2071).** Only the token signature was checked — not that its email matched the account being created, nor that the invite was unexpired. Anyone holding a forwarded invite link could redeem it with an address of their own, at the invited role, up to `owner`. Verified against a real database: before the fix, Bob redeeming Alice's invite became an **owner** of her org and consumed her invite.

> **#8680 has since merged, so the gate itself is now on `main` and this branch is rebased on top of it.** What remains here is the coverage and the client half: two integration tests against real Better Auth + Postgres (an invite redeemed with a different address writes nothing and the invite survives; a DB-expired invite is refused even though its JWT is still signed — the only way to exercise the row's `expiresAt` independently of the token's own `exp`), plus mapping `INVITE_TOKEN_INVALID_ERROR_CODE` to the existing generic invite copy in the sign-up form. #8680 remains the reference for the gate; where the two differed I took theirs.

**Raised by @Dhruwang in review, and removed.** An earlier version of this PR routed an invited visitor with an existing account to the login page. That made the response a lookup: invite any address, run the invite sign-up, and read which screen came back to learn whether it is registered — available to anyone who can create an organization. I first argued it was acceptable because a stronger oracle already exists on the sign-in path. That reasoning holds for an inherited issue, but this one wasn't inherited: `origin/main` has no branch on account existence anywhere in the sign-up flow, so this PR would have added it.

Both cases now land on the same verification-requested screen, and an integration test asserts the two responses are **byte-identical**. The dead end is still fixed, because that screen was already built for it: the copy is conditional ("if there is an account associated with …") and it now carries an unconditional "have an account? log in" link with the invite callback attached, so logging in returns the visitor to the invite. The verification-disabled variant of the same screen already did exactly this, which is where the pattern came from. The remaining, pre-existing instance on `/sign-in/email` is ENG-2099.

The pre-existing one on `/sign-in/email` is left for that ticket: sign up with an address and then attempt sign-in — `EMAIL_NOT_VERIFIED` versus invalid-credentials answers the same question in two anonymous requests, with no org or invite needed. Fixing it means a generic "invalid credentials or unverified account" on that endpoint, which is its own change with its own blast radius. Nothing in this PR contributes to it any more.

Also, while in here: Better Auth's `sign-up.mjs` logs `Sign-up attempt for existing email: <address>` at `info`, and `level: "warn"` was the only thing suppressing it — raising the level to debug a sign-up problem is exactly what you'd do, and it would start writing customer addresses into our logs and Sentry. Now redacted in the log callback (domain kept).

**The login-page notice is gone.** Worth being explicit, because it existed in two earlier forms on this branch: first "log in with your existing **password**" (wrong for an SSO-only account — the reported user signs in with Azure and has no password), then "using the same method you signed up with". Removing the differential removed the notice with it, along with `notices.ts`, the `?notice=` plumbing and both locale strings. If that alert is still showing, it's a build at or before `3922167d4` — the current head renders no alert on `/auth/login` for this flow.

**Why no test caught any of this:** `actions.test.ts` mocked `signUpEmail` into *rejecting* on a duplicate — asserting a branch that cannot execute in production, and staying green the whole time. The framework contract is now pinned against live Better Auth + Postgres.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2091/corporate-email-invitation-flow-does-not-send-verification-email

Also fixes https://linear.app/formbricks/issue/ENG-2071/invite-acceptance-is-not-bound-to-the-invited-email-or-expiry-so-a (see the #8680 overlap note above — that ticket stays assigned to @mattinannt).

## How this was tested

- **Unit** — 6766 passed across all of `apps/web`; 358 of those in `modules/auth`. `actions.test.ts` rewritten to model the real synthetic-200 shape instead of a rejection, plus #8680's `test.each(["", "   ", undefined])` blank-token regression and the four invite-mismatch cases.
- **Integration (real Better Auth + real Postgres, no DB mock)** — the full suite passes, 66 tests across 19 files. Four are new: `signup-duplicate-email` pins the framework contract, `signup-invite` drives the real `createUserAction` end-to-end, `signup-verification-send` covers a healthy / throwing / falsy-returning mailer, and `signup-sso-existing-account` covers the reported user's actual shape — an `azuread` account with no credential row.
- **The enumeration invariant is asserted on both axes and their cross product.** `signup-invite` compares existence **under each mailer state**, and `signup-verification-send` compares mailer states against each other. That pairing is the lesson from the `verification_send_failed` regression: each axis passed on its own, and comparing existence under a *healthy* mailer — which is all the first version did — could never have seen it.
- **Mutation-tested the fixes**, because a suite that stays green without the fix proves nothing. Reverted each in turn: removing the synthetic-id check → the existing-account tests fail; accepting a falsy mailer return → the SMTP-unconfigured test fails. Then both shapes of the oracle, reconstructing the second through the only channel that could carry it: leaking existence as a response field → 4 tests fail; returning the send outcome → 8 fail. Restored: green.
- **ENG-2071 verified by direct repro**, not inference. Before: `bobJoinedVictimOrg: true, bobRole: "owner", aliceInviteStillPending: 0`. After: rejected, no account, no membership, invite intact.
- **Playwright** — new `invite-existing-account.spec.ts` covers the exact reported journey; ran it alongside `signup.spec.ts` (5 passed). That second one matters: it's the public uninvited path that the blank-token normalization could have broken. The invite spec is now run against a dev server in **both** verification configurations, since which success screen appears depends on `EMAIL_VERIFICATION_DISABLED` and CI sets it — that mismatch is what broke E2E on the first push.
- **Smoke-tested in a browser**, including SSO, which none of the automated tests reach: seeded a real `azuread` account with no credential row plus a pending invite, then drove invite → "Create an account" → submit. It and a brand-new address reach the **same path with the same rendered copy** (modulo the address), the log-in link carries `callbackUrl=…/invite?token=…`, and afterwards the DB shows 0 credential accounts on the SSO user, the invite still pending, and `emailVerified` / `identityProvider` untouched.
- **Re-walked @Dhruwang's exact flow on the current head** (existing credential account, invited from a second org, invite link → "Create an account" → same address): lands on "Please confirm your email address" with the conditional copy and a **Log in** link carrying the invite callback — no alert anywhere. DB after: one user, same id, password hash unchanged, name unchanged, invite still pending, 0 memberships in the inviting org, 1 credential account.
- **Open-redirect review of the new `callbackUrl` href**, since it originates in a search param — three layers, each verified rather than assumed: `proxy.ts` rejects any off-origin `callbackUrl` on auth routes with `400` before a page renders (the untouched verification-requested screen answers identically, so it is inherited platform behaviour); the page resolves through `resolveAuthCallbackUrl`, which is origin-allowlisted and strips embedded credentials; and `/auth/login` validates again before redirecting. Smoke-tested: invite callback carried, absent callback → plain `/auth/login`, `https://evil.example/phish` → `400 {"error":"Invalid callback URL"}`.
- `tsc --noEmit` 0 errors, eslint clean, prettier clean, gitleaks + Semgrep clean on the diff.

- **Rebased onto `d8d4feb`** (#8637 reworked the same `betterAuthLogger` block). Kept its ENG-2037 Sentry gating verbatim and the redaction on the logger calls only — #8637 removed the `new Error(message)` Sentry fallback, so that is all the redaction needs to cover now. Re-verified after.
- **Sonar** — Quality Gate passed, **100% coverage on new code** (was 60%), 0 new issues. The 60% was concentrated in two new lib helpers with no unit tests; Sonar's exclusions already drop `**/*.tsx`, `**/actions.ts` and `auth.ts`.
- **CodeRabbit** — `Readonly<>` props and the `@slow` tag applied; the regex finding was already fixed one commit earlier (details in the thread).
- **@BhagyaAmarasinghe** — three findings, two fixed and one blocked:
  - *[P1] invite callback lost when verification is disabled.* Correct and it was the common path, not an edge: `EMAIL_VERIFICATION_DISABLED=1` ships in both `.env.example` and `docker/docker-compose.yml`, so the default self-hosted flow preserved the invite and then stranded the invitee at the app root. `/auth/signup-without-verification-success` now carries `callbackUrl` into `BackToLoginButton`, both branches build their URL through `verification-links.ts`, and the e2e assertion is unconditional (mutation-checked — dropping the callback fails it).
  - *[P2] successful `created` audit event on the duplicate path.* Also correct: the attribution move didn't touch the wrapper, whose action name is fixed and whose status is success unless the handler throws. Added an opt-in `suppressEvent` on `AuditLoggingCtx`, honoured only on the success path so it can never mask a failure, plus a test that a real creation does **not** suppress.
  - *[P2] hand-edited target locales.* Right on policy, blocked in practice — `pnpm i18n:generate` is rejecting the `LINGO_API_KEY`, nothing in CI generates, and en-US alone turns the translation gate red. Agreed in the thread to merge as-is and let the next successful Lingo run overwrite them; the broken key gets its own ticket.

(An earlier round had 7 2FA integration tests and `organization.spec.ts` failing locally on a dev DB missing `Survey.archivedAt` / `TwoFactor.failedVerificationCount` — verified identical on a clean `origin/main` checkout at the time. Those migrations are applied now, so the whole suite runs locally.)

**Visual changes** — two changes to one screen, no redesign, so described rather than screenshotted:

- `/auth/verification-requested` shows a `warning` Alert in place of the usual "we sent a link to <email>" message **when the instance has no SMTP configured**: "Email is not configured / This Formbricks instance has no email server set up, so no verification link could be sent. Please contact your administrator." Config-derived, so it is the same for every visitor — see the note above on why it is not per-request. The copy deliberately talks about the *instance*, not the visitor's account: on a no-mailer instance it renders for someone whose address already had an account too, and an earlier wording ("your account was created… try again below") was false for them and pointed at a resend that cannot work. The resend button and its "you did not receive an email?" lead-in are hidden in that state too, leaving the log-in link as the only offered action.
- That same page gains an always-visible "Have an account? Log in" line, reusing the existing `auth.signup.*` strings, with the invite callback attached when there is one. Unconditional by design — see the note above.

Both use the shared `Alert`/`AlertTitle`/`AlertDescription` components with existing variants, so there's no new styling. Happy to attach captures if a reviewer wants them.

## QA / Test Plan

**How to test**

- [ ] **The reported bug.** As an org owner, invite an address that **already has a Formbricks account**. Open the invite link logged out → "Create an account" → enter that same address + any password → you reach the "please confirm your email address" screen (or "user created successfully" if `EMAIL_VERIFICATION_DISABLED=1`), which offers a **Log in** link. Follow it, log in → you land back on the invite and can accept it. *(Before: the same screen, but with no way forward and the invite already consumed.)*
- [ ] **Invite reachable on a verification-disabled instance (the default self-hosted setup).** With `EMAIL_VERIFICATION_DISABLED=1`, repeat the step above → you land on "user successfully created" and its **Log in** button must carry `?callbackUrl=…/invite?token=…`. Follow it, log in → back on the invite. *(Before: that button went to `/auth/login` bare, so the invite could only be reached by reopening the original email.)*
- [ ] **The invite survives.** After the step above, before logging in, check org settings → Members: the invite still shows **Pending**. *(Before: it was consumed, so re-opening the invite link showed "Invite Not Found".)*
- [ ] **Happy path unbroken.** Invite a brand-new address → "Create an account" → a verification email arrives; clicking the link lands you in the app as a member of the org.
- [ ] **Plain sign-up unbroken.** `/auth/signup` with no invite, a fresh address → account created as normal. (Regression-prone: the invite-token handling treats a blank token as "no invite".)
- [ ] **Invite bound to the address (edge).** Take an invite link issued to address A and sign up with a different address B → rejected with the generic invite error toast; no account for B, and B is **not** added to the org. *(Before: B joined at the invited role, up to owner.)*
- [ ] **Expired invite (edge).** An invite past its expiry, redeemed by the correct address → same generic invite error, no account created.
- [ ] **No-SMTP instance is honest (edge).** Unset `SMTP_HOST` / `SMTP_PORT` and sign up a fresh address → "Email is not configured" instead of "check your inbox", and no resend button (it cannot work). *(Before: it claimed success silently.)* With SMTP configured but the mail server stopped, the screen still says "check your inbox" — deliberate: a per-request failure state would differ by whether the address already had an account. The failure lands in the logs and Sentry, and pressing **Resend verification email** surfaces it.
- [ ] **Same screen either way (security).** Run the invite sign-up twice, once with an address that already has an account and once with a fresh one, and compare what comes back — the screen, the copy, and the log-in link must be identical apart from the address echoed in the text. Worth repeating with the mail server stopped.

**Preconditions / test data**

- An org owner account to send invites from, plus a second, **already-registered** address to accept with — that second account existing beforehand is the precondition for the whole bug.
- Cloud vs self-host: the organization-creation amplification (a duplicate sign-up creating an org owned by someone else's account) only reproduces where multi-org is licensed, i.e. Cloud. Everything else reproduces on both.
- `EMAIL_VERIFICATION_DISABLED=1` changes where plain sign-up lands (`/auth/signup-without-verification-success`) but not the invite paths — that branch is decided first.

**Risks & regressions**

- Public sign-up is the main one: the form always posts `inviteToken: ""`, so anything blank must mean "no invite" everywhere. Covered by unit tests and `signup.spec.ts`, but worth an explicit click-through.
- Invite acceptance is now stricter. If anyone was relying on being invited at one address and onboarding with another, that no longer works — deliberate, and it matches what the logged-in `/invite` page has always enforced.
- Sign-up still succeeds when the mailer fails (unchanged, now pinned by a test) — so a mail outage must not start failing sign-ups.
- `createUserAction`'s response is unchanged from `main` (`{ success: true }`). An intermediate commit on this branch added a `nextStep` field and it has been removed again, so nothing downstream needs to adapt.

- none — no migrations, no new env vars, no `.env.example` change.
- 2 new `en-US` strings (`auth.verification-requested.email_not_configured_{title,description}`). **These are hand-written in the 14 target locales, against repo policy, because the proper route is currently broken** — `pnpm i18n:generate` fails with `Invalid API key` on the `LINGO_API_KEY` in `.env`; nothing in CI generates (`translation-check.yml` only validates); and shipping `en-US` alone makes `scan-translations` exit 1 on both "missing translations" and "lockfile out of sync", so **Validate Translation Keys** would go red. Converting them to generated output is one command once a valid key exists: drop the two `auth/verification-requested/email_not_configured_*` lines from `apps/web/i18n.lock` (the lock freezes them, so this bust is required) and run `pnpm i18n:web:generate`. **Agreed with @BhagyaAmarasinghe in review** to merge as-is and let the next successful Lingo run overwrite them, with the broken key tracked separately. Original note: hand-translated with `i18n.lock` checksums, because nothing actually generates them: the CI job only validates (`scan-translations`) and the pre-commit hook only runs lint-staged. Register and terminology follow each file's existing neighbours (du-form for de-DE, formal for nl-NL/hu-HU/tr-TR/ro-RO). Worth a native-speaker glance at de / ja / tr before merge — and worth fixing the Lingo integration separately, since the local API key is also rejected.
- **Backport to 5.2.x** (5.2.0 and 5.2.1 both carry the duplicate-signup defect). 5.1.x is NextAuth and unaffected.


mattinannt:
Following through on what I said on #8680 — you pushed the branch, so I diffed your ENG-2071 port against #8680's version instead of taking "near-verbatim" on trust. **It checks out. The port is faithful and the security guarantee is intact.**

What I compared (`origin/claude/security-high` vs this PR's head, `fcfbf762`):

| Piece | Result |
|---|---|
| `handleInviteAcceptance` — `resolveInviteMatch(inviteToken, user.email)` gate | **byte-identical**, comment included |
| `createUserAction` — normalized hoist + pre-creation rejection | **byte-identical**, comment included |
| `packages/types/errors.ts` | **byte-identical** (empty diff) |
| `signup-form.tsx` — `INVITE_TOKEN_INVALID_ERROR_CODE` branch | identical content, **reordered** above the password branch |
| Token threading | equivalent — `handlePostUserCreation(ctx, outcome, inviteToken)` passes the normalized token to `handleInviteAcceptance`, same as mine |
| Regression tests | both `test.each(["", "   ", undefined])` and the four `email_mismatch`/`invalid_or_expired`/`verification_error`/`missing` cases present |

The `signup-form.tsx` reorder is semantically neutral — those are mutually exclusive `errorMessage === CONSTANT` branches, so order can't change which one fires. Isolating the port in its own commit (`464c3cb0`) is what made this checkable in one pass; thank you for that.

**This changes my position on #8680.** My objection there was two-part, and the stronger half was "I can't verify equivalence." Now I can, and it's equivalent — so keeping the duplicate is materially less risky than I assessed, and I'm no longer arguing you should strip the hunks. Your call, and either way works.

The #8681 relocation concern also resolves, for a reason worth spelling out: because your inline form is byte-identical to the one #8680 introduces, #8681's extraction into `assertSignupPolicyAllows` still finds exactly the block it expects to move. Had the port drifted even slightly, that extraction could have applied while leaving the check in two places. It didn't drift, so the sequence works: this lands → #8680's invite hunks are already-applied and drop out → #8681 relocates the block your commit put there.

One small thing that *will* need a hand at merge time: #8681 adds a comment immediately above the `else` in that same `signup-form.tsx` chain, and your reorder moves the surrounding lines, so expect a textual conflict there. Trivial to resolve — the comment belongs above the final `else`, wherever the branches end up.

Two notes, no action needed from you:

- Your ENG-2091 root cause makes the CodeRabbit finding I declined on #8680 look better-founded than I credited. I dismissed an orphaned-account concern as pre-existing and privilege-free; your point that every post-creation side effect was running against the pre-existing account — invite consumed, membership and org created, locale overwritten from attacker-supplied input, audit event misattributed — is a sharper version of the same area. It doesn't change my decision for #8680 (`autoSignIn: false` means no session is minted, so there's still no takeover), but your fix is the one that actually closes it, and it belongs in your PR rather than mine.
- I've left ENG-2071 assigned as-is per your note, and the backport question — you flag 5.2.0 and 5.2.1 both carry the duplicate-signup defect — is @mattinannt's call. #8679 went to `release/5.2` as #8686; whether this and the High tranche follow is above my pay grade.

---

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -101,6 +101,23 @@ export const signInAuditDatabaseHook: NonNullable<
...
+ * our logs. Redacting here rather than relying on the level means raising it stays safe. The domain is
+ * kept — it is the part that carries diagnostic value. (ENG-2091)
+ *
+ * Only the logger needs this: since ENG-2037 (below) Sentry receives a real `Error` object or nothing,
+ * never a string built from `message`.
+ */
+const EMAIL_IN_MESSAGE = /[\w.!#$%&'*+/=?^`{|}~-]+@([\w-]+(?:\.[\w-]+)+)/g;
+
+export const redactEmailsInLogMessage = (message: unknown): unknown =>
+  typeof message === "string" ? message.replace(EMAIL_IN_MESSAGE, "[redacted]@$1") : message;
```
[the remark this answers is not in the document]
xernobyl: Already fixed in 17dd15b26 — this review ran against 14f0022, one commit behind. SonarCloud reports 0 new issues on the current head.

I went a different way than suggested, though, because bounding the existing pattern narrows the ambiguity rather than removing it: `.` stays inside the local-part class, so `[\w.!#$%&'*+/=?^`{|}~-]{1,64}` still has many ways to split a long dotted run before failing on the `@`. Shipped instead:

```js
const EMAIL_IN_MESSAGE = /[^\s@]{1,64}@([\w-]{1,63}(?:\.[\w-]{1,63}){1,8})/g;
```

The local part is a negated class that **excludes `@`**, so the run can only end where the `@` actually is — there's no ambiguous split to backtrack through at all. Domain labels likewise exclude `.`, and both are RFC 5321-bounded (64 / 63) so the work per start position is capped.

Same output on every case the tests cover, and 56KB of address-free text goes from a backtracking risk to 8ms. There's now a test pinning that, so a future "simplification" back to an enumerated local-part class shows up as a failure rather than a Sonar warning.

apps/web/modules/auth/verification-requested/page.tsx
```diff
@@ -1,17 +1,20 @@
...
+import { VERIFICATION_SEND_FAILED_PARAM } from "@/modules/auth/lib/verification-links";
 import { RequestVerificationEmail } from "@/modules/auth/verification-requested/components/request-verification-email";
 import { VerificationMessage } from "@/modules/auth/verification-requested/components/verification-message";
+import { Alert, AlertDescription, AlertTitle } from "@/modules/ui/components/alert";
 
 export const VerificationRequestedPage = async ({
   searchParams,
 }: {
-  searchParams: Promise<{ token: string; callbackUrl?: string | string[] }>;
+  searchParams: Promise<{ token: string; callbackUrl?: string | string[]; sendFailed?: string }>;
```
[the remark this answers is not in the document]
xernobyl: Applied in c7b504f2c — good catch, and the guideline is unambiguous.

Did both sites you named plus the other two this branch touches, so the diff is internally consistent rather than half-applying it:

- `verification-requested/page.tsx` → `}: Readonly<{ … }>` (inline props)
- `login/components/login-form.tsx` → `Readonly<LoginFormProps>`
- `signup/components/signup-form.tsx` → `Readonly<SignupFormProps>`
- `login/page.tsx` → `}: Readonly<{ … }>` (inline props)

The latter three were pre-existing violations rather than something this PR introduced; I fixed them because the files are already in the diff, and left the rest of the repo alone.

apps/web/playwright/invite-existing-account.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * ENG-2091: an invitee who already has a Formbricks account used to be sent to "please confirm your
+ * email address" for a verification email that is never sent (Better Auth answers a duplicate address
+ * with a synthetic 200 and skips the send), in front of a resend button that no-ops for an
+ * already-verified address — and their invite was consumed on the way, so logging in afterwards
+ * showed "Invite Not Found".
+ *
+ * They should land on the login page instead, with the invite still pending.
+ */
+test.describe("Invite sign-up with an address that already has an account", async () => {
```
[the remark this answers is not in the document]
xernobyl: Applied in c7b504f2c, with one change to the syntax:

```js
test.describe("Invite sign-up with an address that already has an account @slow", async () => {
```

Plain suffix rather than the backticked `` `@slow` `` from the suggestion — Playwright reads the tag out of the title string, so backticks would become part of the tag and it'd stop matching `--grep @slow`. This matches how `survey-archive.spec.ts`, `survey-keyboard.spec.ts` and `survey-accessibility.spec.ts` already do it.

I checked before adding it that nothing filters on the tag: `pnpm test:e2e` is a bare `playwright test` with no `--grep-invert`, and `e2e.yml` doesn't add one, so the tag is informational today and this regression suite still runs in CI. Worth stating explicitly, since tagging a suite `@slow` in a repo that *did* filter would have quietly removed the coverage this PR exists to add.

apps/web/playwright/invite-existing-account.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+      // Either way there must be a way out, because for this visitor no email is coming and the resend
+      // button no-ops on an already-verified address. That link is the whole point of the fix: before it,
+      // the screen was a dead end and the invite had already been consumed.
+      const loginLink = inviteePage.getByRole("link", { name: "Log in" });
+      await expect(loginLink).toBeVisible();
+      if (inviteePage.url().includes("verification-requested")) {
+        // On the verification screen the link also carries the invite callback, so logging in lands the
+        // visitor back on the invite instead of the app root. The no-verification screen shares a plain
+        // "Log in" button with the other auth flows and has no callback to carry.
```
BhagyaAmarasinghe: [P1] Preserve the invite callback when verification is disabled

This branch explicitly accepts that `/auth/signup-without-verification-success` has no callback, but `EMAIL_VERIFICATION_DISABLED=1` is the default self-hosted configuration and is also the path CI exercises. That page's login button hardcodes `/auth/login`, so although this PR preserves the invite, an existing invitee who clicks **Log in** is not returned to it and cannot complete the flow without reopening the original link. Please carry the invite callback through the verification-disabled success route too and make this assertion unconditional.

apps/web/modules/auth/signup/actions.ts
```diff
@@ -398,13 +451,22 @@ export const createUserAction = actionClient.inputSchema(ZCreateUserAction).acti
...
         }
       }
-    }
 
-    if (user) {
+      // Inside the "created" branch: an audit record claiming a user was created must only exist when
+      // one was. Previously this ran unconditionally, so a duplicate sign-up wrote a `created` event
+      // attributed to the PRE-EXISTING user and carrying their object (ENG-2091 / S1).
+      // Note: withAuditLogging is wrapped around the whole action, so the already-existed path still
+      // emits a `created` event with targetId UNKNOWN_DATA and no newObject. That records an attempt
```
BhagyaAmarasinghe: [P2] Do not emit a successful `created` audit event for duplicate attempts

Moving `auditLoggingCtx.userId` and `newObject` into the `"created"` branch prevents attribution to the pre-existing account, but it does not suppress the wrapper. `withAuditLogging("created", "user", ...)` still sees the duplicate path return successfully and records a successful `created` event with an unknown target. On audit-enabled deployments that is still a false creation record. Please suppress the event for `already_existed` or represent it with accurate attempt semantics while keeping the external response invariant.

apps/web/locales/de-DE.json
```diff
@@ -126,6 +126,8 @@
       "user_successfully_created_info": "Wir haben nach einem Konto gesucht, das mit {email} verknüpft ist. Falls keines existierte, haben wir eines für dich erstellt. Falls bereits ein Konto vorhanden war, wurden keine Änderungen vorgenommen. Bitte melde dich unten an, um fortzufahren."
     },
     "verification-requested": {
+      "email_not_configured_description": "Für diese Formbricks-Instanz ist kein E-Mail-Server eingerichtet, daher konnte kein Bestätigungslink gesendet werden. Bitte wende dich an deinen Administrator.",
```
BhagyaAmarasinghe, line 129: [P2] Regenerate non-English locales instead of editing them manually

The repository policy permits manual changes only in `apps/web/locales/en-US.json`; all other locale files are generated by Lingo.dev via `pnpm i18n`. This PR changes all 14 non-English locales directly. Please keep only the `en-US.json` source change and regenerate the translated files through the repository workflow so they are not overwritten or drift on the next generation run.

BhagyaAmarasinghe (commented):
Overall looks good! Several suggestions:

apps/web/playwright/invite-existing-account.spec.ts
xernobyl: Fixed in 86be3ca70. You're right that this is the common path, not an edge case — both `.env.example` and `docker/docker-compose.yml` ship `EMAIL_VERIFICATION_DISABLED=1`, so I'd left the default self-hosted flow broken while claiming the fix.

`/auth/signup-without-verification-success` now takes `callbackUrl` and passes it to `BackToLoginButton`, and both branches build their URL through `verification-links.ts` (`buildSignupWithoutVerificationSuccessPath` next to the existing `buildVerificationRequestedPath`) rather than one going through a builder and the other string-concatenating. The e2e assertion is unconditional now, and it bites — dropping the callback from that branch fails it, so this can't silently regress.

I also checked the other five `BackToLoginButton` call sites before adding the prop: all password-reset / email-change screens, none reachable in the invite journey, so they correctly keep passing nothing.

On the open-redirect side of putting a search param into that href, there are three layers and I verified each:

- `proxy.ts` already rejects any off-origin `callbackUrl` on auth routes with `400` before a page renders — the untouched verification-requested screen answers identically, so it's inherited platform behaviour rather than something this adds
- the page resolves through `resolveAuthCallbackUrl`, which is origin-allowlisted and strips embedded credentials
- `/auth/login` validates again with the same helper before redirecting

Smoke-tested all three: invite callback carried, absent callback → plain `/auth/login`, `https://evil.example/phish` → `400 {"error":"Invalid callback URL"}`.

Enumeration-wise this is neutral: the callback derives from the `inviteToken` in the URL, never from whether the account exists, so both cases get a byte-identical link.

apps/web/modules/auth/signup/actions.ts
xernobyl: Fixed in 0187000e7. You're right — moving the attribution stopped the event pointing at the pre-existing account but did nothing about the event existing, and since the duplicate path returns successfully on purpose (the response has to be indistinguishable), the wrapper had no way to tell the branches apart.

`ZAuditAction` has no "attempted" value and adding one would ripple into every audit consumer, so I took the suppression option you offered: an opt-in `suppressEvent` on `AuditLoggingCtx` that `withAuditLogging` honours. Additive and default-off so no other action changes behaviour, and deliberately gated on `status === "success"` — a handler that throws is always audited, so the flag can't be used to hide a failure. The context is built per invocation so it can't leak between requests, and nothing in the Zod-validated input can reach it. Response stays byte-identical; only what we record changes.

Mutation-tested with one failing test each: the wrapper ignoring the flag (the defect you found), suppression leaking to the failure path, and the action setting it unconditionally. That last one is why there's now a test asserting a real creation does *not* suppress — otherwise a bug could silence every `created` event and nothing would notice.

One trade-off worth naming rather than burying: there's now no audit record of a duplicate attempt at all. I judged a false creation record worse than none, and the abuse case is covered by the endpoint's IP rate limit with the enumeration surface tracked on ENG-2099 — but if you'd rather have an accurate attempt event than nothing, that's a new `ZAuditAction` value and I'm happy to do it as a follow-up.

apps/web/locales/de-DE.json
xernobyl: Agreed on the policy, and I'd rather not hand-edit these either. I tried to do it properly and hit a wall — checking each link rather than assuming:

- `pnpm i18n:generate` fails at the auth step: `Invalid API key. Run lingo.dev login or check your LINGO_API_KEY.` The key in `.env` is set but rejected.
- Nothing in CI generates them. `translation-check.yml` only runs `scan-translations`, and the pre-commit hook only runs lint-staged — so whatever is committed is what ships.
- Shipping only the `en-US.json` change fails that gate. I reverted the 14 target locales to `main` and ran it: exit 1, on both counts — "Some keys from en-US are missing in target languages" for all 14, and "LOCKFILE OUT OF SYNC". So **Validate Translation Keys** would go red.

So it's hand-written strings or a red gate until someone with a working key runs the generator. I've kept them for now so the gate stays green, and the register in each file follows its existing neighbours (du-form for de-DE and sv-SE, tú for es-ES, vous for fr-FR, formal for hu-HU / nl-NL / pt-PT / ro-RO / ru-RU / tr-TR, você for pt-BR, 您 for zh) — but a native-speaker glance at de / ja / tr before merge would still be welcome.

To convert them to generated output is one command once a valid key exists: remove the two `auth/verification-requested/email_not_configured_*` lines from `apps/web/i18n.lock`, then `pnpm i18n:web:generate`. The lock-bust matters — the lock freezes a key against its en-US checksum, so without removing those lines Lingo treats them as current and leaves the hand-written text in place.

Leaving this thread open. Do you know of a working key, or is the interim acceptable with the next Lingo run overwriting it? Either way it seems worth fixing the key separately, since right now nobody can regenerate translations locally.

xernobyl: Following up on my own question: I don't have a working key either, so I can't regenerate these on this PR.

That leaves two options and I'd rather you pick than assume:

1. **Merge with the hand-written strings** (current state, gate green) and let the next successful Lingo run overwrite them — it will, once the two `i18n.lock` lines are busted, otherwise the lock keeps them frozen.
2. **Block this PR** until someone with a valid key regenerates.

I lean towards (1): the strings are two sentences on a screen that only renders on an instance with no mailer configured, and the alternative is holding a duplicate-signup fix behind unrelated tooling credentials. But it's your call.

Either way the key itself looks worth a separate ticket — as it stands nobody can regenerate translations locally, so the next person to add a string hits exactly this, and the pressure will be to hand-write again.

BhagyaAmarasinghe: yes, 1. seems good to me too.

xernobyl: Thanks — going with 1 then.

For whoever regenerates these later: bust the two `auth/verification-requested/email_not_configured_*` lines in `apps/web/i18n.lock` first, otherwise Lingo sees them as current against their en-US checksum and leaves the hand-written text in place. I'll note that on the key ticket so it doesn't have to be rediscovered.

Resolving this one.

BhagyaAmarasinghe (approved):
Thank you, LGTM!

PR #8692, opened 2026-07-29
feat: project API key scopes to SpiceDB

BhagyaAmarasinghe:
## What does this PR do?

Implements [ENG-1717](https://linear.app/formbricks/issue/ENG-1717/project-existing-api-key-workspace-scopes-into-spicedb) on top of the relationship-projection foundation merged in #8670.

This PR:

- adds a server-only API-key projector for organization ownership, organization access-control read/write flags, and existing workspace `read`/`write`/`manage` grants;
- reconciles complete PostgreSQL snapshots for up to three passes and deletes stale alternate permission relations atomically;
- cleans deleted API keys on both the resource and subject sides, including organization-delete cascades;
- extracts the existing 1,000-operation grouping and bounded deletion behavior into a shared internal projection utility;
- invokes projection after API-key creation/deletion and organization deletion without changing successful source mutations when AuthZed is unavailable;
- extends the isolated Docker smoke test to cover API-key permission ladders, multiple workspace scopes, downgrade/removal, deletion, outage recovery, and persistence;
- covers the independent `accessControl.read=true` plus `accessControl.write=true` state and compile-time completeness of the organization-access relation map;
- records the current immutable-workspace-scope invariant so future scope editing must reconcile pre-change workspace IDs and remove stale grants;
- documents the current API-key projection and repair contract.

PostgreSQL and the legacy evaluator remain authoritative. The canonical SpiceDB schema is unchanged.

Out of scope:

- API-key authorization through the central evaluator (ENG-1731);
- historical backfill and global repair (ENG-1718);
- SpiceDB shadow comparison or enforcement (ENG-1738);
- editable/new API-key scopes, UI changes, database migrations, or environment changes.

There are no visual changes.

## How should this be tested?

Automated validation completed locally with Node 22.22.0:

```bash
CUBEJS_API_URL=https://cube.formbricks.local \
HUB_API_URL=https://hub.formbricks.local \
HUB_API_KEY=test-hub-key \
pnpm --filter @formbricks/web exec vitest run \
  lib/authzed \
  modules/organization/settings/api-keys \
  lib/organization/service.test.ts \
  modules/workspaces/settings/lib/workspace.test.ts
```

Result: 17 test files and 228 tests passed.

```bash
pnpm --filter @formbricks/web typecheck
pnpm authzed:validate
pnpm authzed:smoke
pnpm build --filter=@formbricks/web... --force
```

Results:

- web typecheck passed;
- schema validation loaded 27 relationships and passed 131 assertions plus 2 expected-relations checks;
- the isolated Docker smoke test passed API-key organization access (including simultaneous read/write flags), all workspace grant levels, multiple scopes, downgrade/removal, idempotent deletion, outage recovery, migrations, and persistence;
- a forced production build passed with AuthZed disabled;
- a production build passed with AuthZed enabled and `127.0.0.1:1` as an unreachable endpoint, confirming no build-time RPC.

Review the projector tests for the four organization-access combinations, all three workspace permissions, malformed JSON, batching, bounded deletion, concurrent source changes, recreation, disabled behavior, and secret-safe failures.

### Required

- [x] Ran `pnpm build`
- [x] Checked for warnings introduced by this change; there are none
- [x] Started from and rechecked the latest `epic/authzed` target branch
- [x] Not my first PR at Formbricks

### Appreciated

- [x] No UI change was made, so screenshots are not applicable
- [x] Updated the AuthZed authorization documentation

### DIFF
--- a/apps/web/lib/authzed/api-key.ts
+++ b/apps/web/lib/authzed/api-key.ts
@@ -217,6 +217,9 @@ export const reconcileApiKeyRelationships = async (
       return 0;
     }
 
+    // Workspace scopes are immutable today, so current snapshots contain every workspace that needs
+    // reconciliation. If scope removal is added, callers must also target pre-change workspace IDs
+    // so stale subject-side grants are deleted.
     const observedWorkspaceIds = new Map(apiKeyIds.map((apiKeyId) => [apiKeyId, new Set<string>()]));
 
     for (let pass = 1; pass <= AUTHZED_MAX_RECONCILIATION_PASSES; pass++) {
--- a/apps/web/lib/authzed/api-key.ts
+++ b/apps/web/lib/authzed/api-key.ts
@@ -1,0 +1,234 @@
+import "server-only";
+import { prisma } from "@formbricks/database";
+import { ApiKeyPermission } from "@formbricks/database/prisma";
+import { type TAuthzedRelationshipFilter, type TAuthzedRelationshipUpdate, getAuthzedClient } from "./client";
+import {
+  AUTHZED_MAX_RECONCILIATION_PASSES,
+  AuthzedProjectionUnstableError,
+  type TAuthzedProjectionResult,
+  runBestEffortProjection,
+} from "./projection";
+import { deleteRelationshipsInBoundedBatches, packRelationshipUpdateGroups } from "./relationship-batches";
+
+const ORGANIZATION_ACCESS_RELATIONS = {
+  read: "api_key_reader",
+  write: "api_key_writer",
+} as const;
+
+const WORKSPACE_RELATIONS = {
+  [ApiKeyPermission.manage]: "manager",
+  [ApiKeyPermission.read]: "reader",
+  [ApiKeyPermission.write]: "writer",
+} as const satisfies Record<ApiKeyPermission, string>;
+
+const WORKSPACE_RELATION_NAMES = Object.values(WORKSPACE_RELATIONS);
+
+export type TApiKeyProjectionTargets = Readonly<{
+  apiKeyIds?: ReadonlyArray<string>;
+}>;
+
+type TOrganizationAccessSnapshot = Readonly<{
+  read: boolean;
+  write: boolean;
+}>;
+
+type TApiKeyWorkspaceSnapshot = Readonly<{
+  permission: ApiKeyPermission;
+  workspaceId: string;
+}>;
+
+type TApiKeySnapshot = ReadonlyArray<
+  Readonly<{
+    apiKeyWorkspaces: ReadonlyArray<TApiKeyWorkspaceSnapshot>;
+    id: string;
+    organizationAccess: TOrganizationAccessSnapshot;
+    organizationId: string;
+  }>
+>;
+
+const isRecord = (value: unknown): value is Readonly<Record<string, unknown>> =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
+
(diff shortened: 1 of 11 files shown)


apps/web/lib/authzed/api-key.ts
```diff
@@ -0,0 +1,234 @@
...
+export const reconcileApiKeyRelationships = async (
+  targets: TApiKeyProjectionTargets
+): Promise<TAuthzedProjectionResult> =>
+  runBestEffortProjection("reconcile_api_key_relationships", "api_key", async () => {
+    const apiKeyIds = normalizeTargets(targets);
+    if (apiKeyIds.length === 0) {
+      return 0;
+    }
+
+    const observedWorkspaceIds = new Map(apiKeyIds.map((apiKeyId) => [apiKeyId, new Set<string>()]));
```
xernobyl, line 223: Latent, not reachable today — flagging it so it isn't rediscovered later.

`observedWorkspaceIds` can only ever learn workspaces that are still in Postgres (it's fed from the snapshot). So if a workspace grant is *removed*, that workspace never enters the set, the `createWorkspaceUpdates` loop never runs for it, and the stale `workspace:W#reader|writer|manager@api_key:X` relation is never deleted — the key keeps that access in SpiceDB.

I checked and it's unreachable right now: `updateApiKey` only changes the label, and nothing anywhere mutates `ApiKeyWorkspace` rows, so scopes are immutable after creation. Create/delete/org-delete really is complete coverage today.

It becomes live with ENG-1729/1730 though, once scopes are editable — and then it's privilege-retention drift rather than a missing grant, which is the worse direction. Cheapest fix at that point is to accept the affected workspace ids as projection targets (or delete subject-side before rewriting). A short comment here noting the assumption would probably save someone the archaeology.

xernobyl (approved):
LGTM 🚀 And thanks for picking up both notes from #8670 — sharing `AUTHZED_MAX_RELATIONSHIP_UPDATES` out of `constants` and bounding the deletes in parallel batches. The extra guard that throws when a single group exceeds the cap is better than what I suggested.

Things I checked and liked:

- **Delete narrowing survives the generalisation** — the `api_key` resource filter plus the subject-scoped filters on `organization` and `workspace` all stay within the narrowing guard.
- **Cleans both sides on key deletion** (resource *and* subject), which closes the orphan class I flagged on the membership PR.
- **Fails closed on bad data** — `normalizeOrganizationAccess` defaults to `{read:false, write:false}` if the JSON column isn't the expected shape.
- **Independent access-control flags are handled correctly**: each relation is touched iff its flag is set and deleted otherwise, so `read=false, write=true` projects only `api_key_writer` and the schema's `read_access` union still grants read — matching the legacy write-implies-read behaviour. Both maps using `satisfies` for compile-time completeness is a nice touch.
- **Coverage is genuinely complete**: I grepped for it rather than assuming — `updateApiKey` only touches the label and nothing mutates `ApiKeyWorkspace` rows, so create/delete/org-delete really is every path today.

One latent thing left inline about `observedWorkspaceIds` (unreachable now, becomes real with ENG-1729/1730), plus two trivia: `snapshotsMatch` via `JSON.stringify` is key-order dependent like `team-workspace`, and the deletion filters get recomputed on each retry pass (idempotent, just a few wasted RPCs).

Worth recording for ENG-1738: this projects api-key scopes into SpiceDB while my ENG-1731 has the evaluator read the same scopes from Postgres per decision. They agree, with one mechanism difference — my legacy path implements "write implies read" in code, yours gets it from the schema union. Same outcome, but that's exactly the kind of thing parity testing should assert explicitly rather than assume.

PR #8699, opened 2026-07-30
refactor: route API-key authorization through the central interface (ENG-1731)

xernobyl:
## What does this PR do?

Routes API-key authorization through the central `can()` / `assertCan()` interface, so API keys are decided the same way user sessions already are after #8653. The legacy evaluator stays authoritative and no AuthZed configuration or network access is involved.

Migrated surfaces:

- **v1** management, webhooks and storage checks (~20 sites), plus both v1 and v2 body/workspace resolvers
- **v2** management checks (~13 sites) and the 14 organization access-control checks (`hasOrganizationIdAndAccess`)
- **v3** — the API-key branch of `requireV3WorkspaceAccess` now asks `can`; the `apiKeyPermissionAllows` rank comparator is gone
- **storage** route auth

Effective permissions are unchanged. The workspace ladder still maps `GET → read`, `POST/PUT/PATCH → write`, `DELETE → manage`, and organization `accessControl.write` still implies `read`. The route-level `allowOrganizationOnlyApiKey` opt-in stays at the authentication layer, untouched.

Three things worth a look in review:

**1. The pure ladders moved so the wrappers can delegate without recursing.** The evaluator itself called `hasPermission` / `hasOrganizationAccess`, so pointing those at `can` would have looped (`can → evaluator → helper → can`). They now live in `lib/authorization/legacy-api-key-access.ts`, the evaluator uses those, and the public helpers are thin `can` wrappers — the same shape as `legacy-workspace-access.ts` from #8653.

**2. The helpers were renamed rather than made async in place.** `hasPermission` → `hasApiKeyWorkspaceAccess`, `hasOrganizationAccess` → `hasApiKeyOrganizationAccess`. This is deliberate: `if (!hasPermission(...))` keeps compiling when the `await` is forgotten, `!Promise` is always false, and the guard would silently authorize — TypeScript won't catch it. Renaming turns every unmigrated call site into a compile error instead, which is how I found two local wrappers I'd otherwise have missed. Also deleted `hasWorkspacePermission`, a byte-identical dead duplicate of `hasPermission`.

**3. One intentional behaviour delta: keys are re-resolved per decision.** The evaluator reads a key's scopes from storage rather than trusting the snapshot captured at authentication, so a key that has been **revoked, deleted or narrowed** since it authenticated stops authorizing. That's stricter than before and satisfies the ticket's "revoked and deleted keys cannot authorize" requirement, but it is a change worth being explicit about. Cost is one extra `react.cache`-deduped query per request. The alternative — threading the resolved auth through the interface — would have kept byte-identical semantics at the price of widening the contract, which we chose not to do.

**MCP needed no change, and I want to be explicit about why** rather than leave it looking skipped. Its own checks are authentication and OAuth *scope* concerns: `getMcpScopes` derives `surveys:read` / `surveys:write` strings from the key's grants, and the OAuth path validates token, user-active and scope presence. Resource authorization happens downstream through the v3 path, which this PR migrates — so the applicable boundary is covered transitively.

Fixes [ENG-1731](https://linear.app/formbricks/issue/ENG-1731/route-api-key-authorization-through-the-central-interface)

## How should this be tested?

Behaviour-preserving, so there is nothing new to click through — it is covered by unit tests:

```bash
pnpm --filter=@formbricks/web exec vitest run lib/authorization modules/organization/settings/api-keys modules/api app/api app/storage modules/ee/contacts/api modules/mcp
```

- **1221 tests across 131 files pass**; typecheck and ESLint are clean.
- Tests were restructured to match the new layering: the ladder is covered purely in `lib/authorization/legacy-api-key-access.test.ts`, while the wrappers assert they delegate to `can` with the right action and resource.
- New coverage for the invariants the ticket calls out: the workspace `read`/`write`/`manage` ladder, organization `accessControl` combinations, the **organization-only route opt-in**, and **revoked/deleted keys cannot authorize** — the latter both at the evaluator and at the authentication layer, where `authenticateApiKeyFromHeaders` previously had no test coverage at all.

Manual sanity check for a reviewer who wants one: with an API key scoped `read` on a workspace, a `GET` management route should succeed and `POST`/`DELETE` should 401/403 exactly as before; an organization-only key should still be rejected everywhere except the routes that opt in.

### Required

- [ ] Ran `pnpm build` (ran `typecheck` + full `vitest` + `eslint`; full web build not run)
- [ ] Merged the latest changes from main (stacked on `epic/authzed`, not `main`)


apps/web/modules/organization/settings/api-keys/lib/utils.ts
```diff
@@ -1,84 +1,68 @@
...
+ */
+export const hasApiKeyWorkspaceAccess = async (
+  authentication: TAuthenticationApiKey,
   workspaceId: string,
-  method: "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
-): boolean => {
-  if (!permissions) return false;
+  method: LegacyApiKeyMethod
+): Promise<boolean> =>
+  can({ type: "apiKey", id: authentication.apiKeyId }, METHOD_WORKSPACE_ACTION[method], {
```
BhagyaAmarasinghe, line 46: [P1] Preserve the tenant boundary when reloading API-key grants

This now calls `can()` with only the `apiKeyId`, so the evaluator reloads scopes through `getApiKeyAuthById()`. Unlike `authenticateApiKeyFromHeaders`, that resolver neither selects `workspace.organizationId` nor filters grants to the key’s owning organization. A key with one legitimate grant plus a stale cross-organization `ApiKeyWorkspace` row can therefore pass authentication, have the foreign row reintroduced here, and authorize against another tenant. Please mirror the ENG-1749 organization filter in `getApiKeyAuthById()` and add a regression test for a foreign workspace grant.

apps/web/modules/api/v2/organizations/[organizationId]/lib/utils.ts
```diff
@@ -1,18 +1,18 @@
...
-): boolean => {
+): Promise<boolean> => {
   if (paramOrganizationId !== authentication.organizationId) {
     logger.error("Organization ID from params does not match the authenticated organization ID");
 
     return false;
   }
 
-  return hasOrganizationAccess(authentication, accessType);
+  return hasApiKeyOrganizationAccess(authentication, accessType);
```
Comment (BhagyaAmarasinghe) [inline apps/web/modules/api/v2/organizations/[organizationId]/lib/utils.ts:17]: [P1] Restore the team-route tests after making this boundary async

This change sends the organization team route tests through the real central evaluator, but those tests do not mock this authorization boundary. The required unit-test check currently has 6 failures across `teams/route.test.ts` and `teams/[teamId]/route.test.ts`, all ending at `prisma.apiKey.findUnique` because the test Prisma mock has no `apiKey` model. Please update both route tests to mock the central authorization decision and preserve their user-management-floor assertions.

BhagyaAmarasinghe (commented):
Two blocking findings from the exact-head review.

BhagyaAmarasinghe (approved):

PR #8703, opened 2026-07-30
feat(authzed): add relationship sync observability and runbook (ENG-1719)

xernobyl:
## What does this PR do?

Makes AuthZed relationship-sync failures observable, and documents the recovery path that points at #8702's backfill command.

The premise worth stating first, because everything here follows from it: **projection is best-effort by design.** An AuthZed outage must never turn a successful PostgreSQL mutation into an application error. That is the right trade-off, and it means sync failures are *silent* — nothing in the product breaks when a projection is dropped, so drift accumulates until something reads SpiceDB and disagrees with PostgreSQL.

Fixes [ENG-1719](https://linear.app/formbricks/issue/ENG-1719/add-authzed-relationship-sync-observability-and-runbook)

### Metrics — the first custom instruments in the repo

Four instruments on the OpenTelemetry metrics API. The app already exports through both the Prometheus and OTLP readers configured in `instrumentation-node.ts`, and `getMeter` returns a no-op meter when neither is enabled — so recording is safe with zero configuration and costs nothing. No new dependency; `@opentelemetry/api` is promoted from transitive to declared.

| Instrument | Attributes |
|---|---|
| `formbricks_authzed_projection_total` | `operation`, `projection`, `status` |
| `formbricks_authzed_projection_duration_seconds` | same |
| `formbricks_authzed_request_failures_total` | `operation`, `code`, `retryable` |
| `formbricks_authzed_request_retries_total` | `operation`, `code` |

Wired into exactly the two places that already compute a duration and a sanitized code — `projection.ts` and `retry.ts` — so nothing new is calculated on the request path.

> **Post-review update.** A best-practices pass verified against OpenTelemetry's Prometheus translation caught a real naming defect in the duration metric — see **[Review findings](#review-findings)**.

Three decisions worth a look:

**Retries are counted separately from failures.** A retry that later succeeds never reaches the failure counter, so without this a *degraded* SpiceDB is invisible until it starts dropping writes outright. The ratio of retries to projections is the early-warning signal; AuthZed's own guidance points at connection-pool starvation (`pgxpool_empty_acquire`) as the usual cause, which the runbook links to.

**`disabled` is recorded as its own projection outcome** rather than skipped. A deployment that believes AuthZed is on while `AUTHZED_ENABLED` is off looks healthy by every other signal — same reason the health CLI exits non-zero on `disabled`.

**Every attribute is a bounded enumerable value, never an identifier.** These leave the deployment when an OTLP endpoint is configured, so an organization or user ID would be both a cardinality explosion and a privacy leak. A test asserts the exact attribute key set rather than trusting review to catch a future addition.

Deliberately **no metrics from the backfill command**. A short-lived process has no scrape window and no flush; its observability is the counters in its own JSON result and its exit code. Better to say that than leave it looking like an omission.

### One real inconsistency in the failure logs

Auditing the existing log call sites turned up a gap that would break an alert query rather than merely look untidy: **`projection-boundary.ts` logged `code` where every other AuthZed failure logs `errorCode`**, so a single query could not cover both. That boundary is the one that fires when a projector *itself* has a bug — the path you least want to miss. Renamed, and given the `status` / `retryable` fields its siblings already carry. The projection failure log was also missing `status` while its success counterpart had it.

I deliberately left `attemptCount` (request layer, "which attempt is this") and `attempts` (projection layer, "how many in total") alone — they look similar but mean different things, and the runbook documents the distinction rather than papering over it with a rename.

Otherwise the existing logs were already sanitized and complete, so this is a narrow fix rather than a sweep through Bhagya's code.

### The runbook

New [`authzed/RUNBOOK.md`](https://github.com/formbricks/formbricks/blob/tiago/eng-1719-add-authzed-relationship-sync-observability-and-runbook/authzed/RUNBOOK.md), linked from the README: symptoms, diagnosis (metrics, the log field contract, and what to correlate on the SpiceDB side), the escalation from dry run → converge → prune, what to confirm before pruning, what to do after an outage, PromQL alert sketches, and escalation.

Two things are called out because getting them wrong is expensive:

- **The commands load `.env` and ignore `.env.local`.** Next.js prefers `.env.local`, so the instance the CLI rewrites is not necessarily the one your dev server talks to — which is exactly why `--expected-endpoint` exists.
- **A large orphan count is a symptom, not a workload.** Wrong endpoint, wrong database, a restore in progress. Establish why before reaching for `--max-prune`.

The alert rules are sketches with thresholds to tune. Shipping them as a Helm `PrometheusRule` belongs with the AuthZed deployment contract rather than the application, so that is left as a follow-up rather than bolted on here.

## How should this be tested?

```bash
pnpm turbo build --filter='@formbricks/web^...'
pnpm --dir apps/web test lib/authzed modules/account
```

- **370 tests across 22 files pass**; typecheck clean.
- New `metrics.test.ts` covers each instrument, that retries and failures are counted separately, and the attribute-cardinality assertion.
- `better-auth-account-deletion.test.ts` pinned the old `code` field, so it is updated to the new contract — which is the regression proof that the rename is complete.

To see the metrics for real: set `PROMETHEUS_ENABLED=1` with `AUTHZED_ENABLED=true`, exercise a membership change, and scrape `:9464/metrics` for `formbricks_authzed_*`. With neither exporter enabled the meter is a no-op, which is the default and is covered by the tests above.

## Review findings

**The duration metric was named wrong, twice, and the second attempt is worth reading.** It started as `..._duration_ms` with `unit: "ms"` — wrong, since the semantic conventions prescribe seconds. The first fix renamed it to `..._duration` on the claim that the Prometheus exporter appends the unit as a name suffix. **It does not.** `@opentelemetry/exporter-prometheus` appends only `_total`, to monotonic sums, and emits the unit as a `# UNIT` comment — so a deployment scraping `/metrics` exposed `..._duration_bucket` while an OTLP collector normalized to `..._duration_seconds_bucket`, and the runbook's `histogram_quantile` alert silently matched nothing on the scrape path.

It is now `formbricks_authzed_projection_duration_seconds` with `unit: "s"`, recording seconds: OTLP's translation skips a unit suffix the name already carries, so that spelling is the one both exporters agree on. The test pins it and rejects both the unit-less and `_ms` forms. Worth flagging as a review lesson — a unit test can only see the *instrument* name, so nothing in the suite could have caught this; it took reading the exporter we actually ship.

**`disabled` outcomes no longer pollute the latency histogram.** They short-circuit before any work, so the zero is structural rather than measured; on a deployment with AuthZed switched off it would drag every quantile of the one signal the runbook calls user-visible toward zero. The counter still records them, which is the part that matters.

**Deliberately not changed:** the `_total` suffix on the counter. I checked both translators and the spec — each trims an existing `_total` before re-appending, so there is no `_total_total`. It does leak an exporter concern into the instrument name, but the runbook documents the Prometheus names and renaming would churn them for nothing.

**Deferred:** dot-namespaced instrument and attribute names (`authzed.operation`, `error.type` for codes) per semconv, and exporting metrics from the CLI via an explicit `forceFlush`. Both are worth doing; neither is a defect.

### Required

- [ ] Ran `pnpm build` (ran `typecheck` + `vitest`; full web build not run)
- [x] Merged the latest changes onto my branch (stacked on #8702, which is on `epic/authzed`)

### Appreciated

- [x] Updated the Formbricks Docs if changes were necessary (new `authzed/RUNBOOK.md`, linked from `authzed/README.md`)

---

**Stacked on #8702** (ENG-1718) — the runbook's whole recovery path is that command, so this only makes sense on top of it. Kept as a draft until #8702 merges; GitHub will retarget it to `epic/authzed` then.

### DIFF
--- a/apps/web/lib/authzed/metrics.ts
+++ b/apps/web/lib/authzed/metrics.ts
@@ -52,6 +52,15 @@ const projectionTotal = meter.createCounter("formbricks_authzed_projection_total
  * nothing on the scrape path.
  */
 const projectionDuration = meter.createHistogram("formbricks_authzed_projection_duration_seconds", {
+  // The SDK's default boundaries are `[0, 5, 10, 25, … 10000]` — a millisecond scale. Recording seconds
+  // against them puts every healthy projection in the single `(0, 5]` bucket, and `histogram_quantile`
+  // interpolates within a bucket: a p95 over observations that are all ~100ms reports something close to
+  // 4.75s, so the runbook's `> 0.5` alert would fire continuously on healthy traffic. These are the
+  // semantic conventions' second-scale boundaries, which include 0.5 exactly so the alert threshold
+  // falls on a boundary rather than inside a bucket.
+  advice: {
+    explicitBucketBoundaries: [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10],
+  },
   description: "Duration of AuthZed relationship projections",
   unit: "s",
 });
--- a/apps/web/lib/authzed/metrics.ts
+++ b/apps/web/lib/authzed/metrics.ts
@@ -19,6 +19,16 @@ import { metrics } from "@opentelemetry/api";
  * Note this covers the always-on request path only. The backfill command is a short-lived process with
  * no scrape window and no flush, so its observability is the counters in its own JSON result and its
  * exit code.
+ *
+ * **A deliberate deviation from the semantic conventions, which prescribe dots as namespace delimiters
+ * (`formbricks.authzed.projection.duration`) and say a unit need not appear in the name.** This app
+ * configures the Prometheus reader *and* an OTLP reader at once, and the two derive a series name
+ * differently: the Prometheus exporter sanitizes dots to underscores and appends no unit, while OTLP's
+ * Prometheus translation appends the unit unless the name already carries it. Under the conventional
+ * spelling the same instrument would surface as `..._duration` on a scrape and `..._duration_seconds`
+ * through a collector — so the runbook could not name one series, which is exactly the defect this
+ * naming replaced. Prometheus-style names with the unit spelled out are the only form both paths agree
+ * on. Revisit if the Prometheus reader is ever dropped.
  */
 
 const meter = metrics.getMeter("formbricks.authzed");
--- a/apps/web/lib/authzed/metrics.ts
+++ b/apps/web/lib/authzed/metrics.ts
@@ -29,15 +29,19 @@ const projectionTotal = meter.createCounter("formbricks_authzed_projection_total
 });
 
 /**
- * Seconds, and no unit in the instrument name.
+ * Seconds, with the unit spelled out in the instrument name.
  *
- * OpenTelemetry's semantic conventions prescribe seconds for durations, and its Prometheus translation
- * appends the unit as a name suffix — so `..._duration_ms` with `unit: "ms"` would export as
- * `..._duration_ms_milliseconds` through the OTLP path while the JS Prometheus exporter appends nothing,
- * leaving the two exporters this app configures side by side disagreeing on the metric's name. Naming it
- * without a unit and measuring in seconds yields `..._duration_seconds` from both.
+ * The semantic conventions prescribe seconds for durations, which rules out the `_ms` this started as.
+ * The unit belongs in the *name* as well because the two exporters this app configures side by side
+ * derive the series name differently: `@opentelemetry/exporter-prometheus` appends only `_total`, to
+ * monotonic sums, and emits the unit as a `# UNIT` comment rather than a suffix, while OTLP's Prometheus
+ * translation appends the unit — skipping it when the name already carries it. Naming it `_seconds` is
+ * therefore the one spelling both paths agree on, and the runbook's alert queries can name a single
+ * series. Leaving the unit out of the name would export `..._duration` on a scrape and
+ * `..._duration_seconds` through a collector, which is how the runbook's histogram query came to match
+ * nothing on the scrape path.
  */
-const projectionDuration = meter.createHistogram("formbricks_authzed_projection_duration", {
+const projectionDuration = meter.createHistogram("formbricks_authzed_projection_duration_seconds", {
   description: "Duration of AuthZed relationship projections",
   unit: "s",
 });
--- a/apps/web/lib/authzed/metrics.ts
+++ b/apps/web/lib/authzed/metrics.ts
@@ -1,0 +1,88 @@
+import "server-only";
+import { metrics } from "@opentelemetry/api";
+
+/**
+ * Operational metrics for AuthZed relationship sync.
+ *
+ * Projection is best-effort by design: an outage never turns a successful PostgreSQL mutation into an
+ * application error. That is the right trade-off, and it is also why drift can accumulate silently —
+ * these counters are how an operator finds out, and what tells them to run `pnpm authzed:backfill`.
+ *
+ * Uses the OpenTelemetry metrics API, which the app already exports through both the Prometheus and
+ * OTLP readers configured in `instrumentation-node.ts`. When neither is enabled `getMeter` returns a
+ * no-op meter, so recording is safe with zero configuration and costs nothing.
+ *
+ * **Every attribute is a bounded, enumerable value — never an identifier.** Same rule as the logger:
+ * these leave the deployment when an OTLP endpoint is configured, and an organization or user ID here
+ * would be both a cardinality explosion and a privacy leak.
+ *
+ * Note this covers the always-on request path only. The backfill command is a short-lived process with
+ * no scrape window and no flush, so its observability is the counters in its own JSON result and its
+ * exit code.
+ */
+
+const meter = metrics.getMeter("formbricks.authzed");
+
+/** Projection outcomes, by operation and projector. Includes `disabled` so a misconfigured deployment is visible. */
+const projectionTotal = meter.createCounter("formbricks_authzed_projection_total", {
+  description: "AuthZed relationship projections by outcome",
+});
+
+const projectionDuration = meter.createHistogram("formbricks_authzed_projection_duration_ms", {
+  description: "Duration of AuthZed relationship projections",
+  unit: "ms",
+});
+
+/**
+ * Requests that exhausted their retry budget.
+ *
+ * The signal that distinguishes a blip from an outage: a sustained rate here means relationships are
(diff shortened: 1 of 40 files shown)


apps/web/lib/authzed/metrics.ts
```diff
@@ -0,0 +1,117 @@
...
+ * The unit belongs in the *name* as well because the two exporters this app configures side by side
+ * derive the series name differently: `@opentelemetry/exporter-prometheus` appends only `_total`, to
+ * monotonic sums, and emits the unit as a `# UNIT` comment rather than a suffix, while OTLP's Prometheus
+ * translation appends the unit — skipping it when the name already carries it. Naming it `_seconds` is
+ * therefore the one spelling both paths agree on, and the runbook's alert queries can name a single
+ * series. Leaving the unit out of the name would export `..._duration` on a scrape and
+ * `..._duration_seconds` through a collector, which is how the runbook's histogram query came to match
+ * nothing on the scrape path.
+ */
+const projectionDuration = meter.createHistogram("formbricks_authzed_projection_duration_seconds", {
```
BhagyaAmarasinghe, line 54: [P1] Configure second-scale histogram buckets. This histogram records seconds but provides no bucket advice or SDK view, so OpenTelemetry SDK 0.217.0 uses boundaries 0, 5, 10, 25, and so on. Every normal sub-five-second projection lands in one bucket; the runbook query can estimate roughly 4.75 seconds even when every observation is 100 ms, causing the p95 > 0.5 alert to fire on healthy traffic. Add explicit second-scale boundaries around the alert threshold and test the exported bucket configuration with the real SDK.

authzed/RUNBOOK.md
```diff
@@ -0,0 +1,285 @@
...
+# Remove the relationships of a workspace whose row is gone. This is a prune — every relationship on
+# that workspace goes, team and API-key grants included — so it takes the prune flags, not --apply
+# alone. Without them the run reports the orphans and removes nothing.
+pnpm authzed:backfill --apply --prune --confirm-prune --workspace-id=<cuid> \
+  --expected-endpoint=<host:port>
+```
+
+Narrow, but not hermetic: the API keys holding grants on that workspace are reconciled in full, which
+also converges their grants on *other* workspaces. That direction only ever writes what PostgreSQL
+says, so the effect is a wider repair than you asked for, never a deletion outside the workspace named.
```
BhagyaAmarasinghe: [P1] This blast-radius guarantee is false for the stacked implementation. A stale workspace grant causes normalizeTargets to infer its API key or team; if that principal is missing from PostgreSQL, the reconcilers issue subject-wide deletes across every organization and workspace. A single orphan on workspace W can therefore delete relationships outside W without those deletions being included in its prune budget. Fix the underlying #8702 behavior before documenting this guarantee, or explicitly warn about the wider destructive scope.

authzed/RUNBOOK.md
```diff
@@ -0,0 +1,285 @@
...
+
+Escalate in this order. Every step is safe to repeat — relationships are written with `TOUCH`.
+
+**1. See what is wrong. Writes nothing.**
+
+```bash
+pnpm authzed:backfill
+```
+
+Exit `0` clean, `2` drift remains, `1` failed. Read `counters`, `orphans`, and `failures` from the JSON.
```
BhagyaAmarasinghe, line 112: [P1] Exit zero does not currently mean clean. The stacked backfill calculates drift from orphaned, missing, mismatched-parent, and truncated state only; it ignores invalid and unmanaged even though those categories represent unrepaired authorization state. The command can therefore exit 0 and report reconciled while cross-organization source grants or unknown SpiceDB relationships remain. Because this runbook later treats a clean run as the gate for shadow evaluation and enforcement, #8702 must include these categories in final status and this runbook should explain how to handle invalid.

authzed/RUNBOOK.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Metrics
+
+All four carry only bounded attributes — never an organization, user, or relationship identifier.
+
+| Metric | Attributes | Read it as |
+| --- | --- | --- |
+| `formbricks_authzed_projection_total` | `operation`, `projection`, `status` | Projection outcomes. `status` is `projected` / `failed` / `disabled`. |
+| `formbricks_authzed_projection_duration_seconds` | same | Projection latency. It sits on the request path, so a rise here is user-visible. `disabled` outcomes are deliberately excluded — their duration is a structural zero, not a measurement. |
+| `formbricks_authzed_request_failures_total` | `operation`, `code`, `retryable` | Requests that exhausted their retry budget. **Each one is a dropped relationship.** |
```
BhagyaAmarasinghe: [P2] Request failures are not one dropped relationship each. This counter instruments every final facade failure, including schema operations, relationship reads, writes, and deletes; one failed write can also contain an entire relationship batch. A sample therefore represents a terminal AuthZed request failure, not exactly one dropped relationship. Describe it accordingly and use the failed projection counter as the signal that application projection drift was introduced.

BhagyaAmarasinghe (changes requested):
Requesting changes for three blocking observability and recovery-safety issues, plus one P2 metric-contract correction. Review is against exact head 26d82359de907ff5b90a51717ead2e5736339db1. The focused suite passed (21 files, 402 tests); the histogram defect was also reproduced against the exact installed OpenTelemetry SDK.

apps/web/lib/authzed/metrics.ts
xernobyl: Fixed in `1a5df70dc` — explicit second-scale `advice.explicitBucketBoundaries`, including 0.5 so the alert threshold falls on a boundary rather than inside one.

You were right about the default: 0/5/10/25 is a millisecond-scale ladder, so every sub-five-second projection collapsed into a single bucket and the p95 the runbook queried was an artefact of the boundary, not a measurement.

authzed/RUNBOOK.md
xernobyl: The underlying behaviour is fixed in #8702 (`50fcd5c37`) rather than only documented — grant refs whose inferred principal is missing are withheld from workspace scope. The runbook now also states the remaining cost outright instead of promising containment.

authzed/RUNBOOK.md
xernobyl: Fixed on both sides: #8702 includes `invalid` and `unmanaged` in the final status (`50fcd5c37`), and the runbook now says exit `0` covers every unrepaired category and explains what a non-zero `invalid` needs from a human.

authzed/RUNBOOK.md
xernobyl: Reworded — the counter is now described as one terminal request failure across any facade call, including schema operations and reads, with a note that one failed write can carry a whole batch. `formbricks_authzed_projection_total{status="failed"}` is named as the signal for "was projection drift introduced".

BhagyaAmarasinghe (approved):

PR #8704, opened 2026-07-31
chore: resolve all eslint warnings across the workspace

Dhruwang:
No Linear ticket — housekeeping to clear the ESLint warnings that `pnpm lint` prints across the workspace.

## What & why

`pnpm lint` printed warnings while every individual package `lint` task still exited 0, so they were easy to miss under Turbo's cache. This clears them: `pnpm lint` is now **21/21 tasks, 0 warnings, 0 errors**.

All of it is `react-hooks/exhaustive-deps` in `apps/web`, resolved case by case rather than blanket-suppressed:

- Added the missing dependency where it is a pure derivation and safe to include (`useMemo`/`useCallback` inputs, stable `t`).
- Removed deps ESLint flagged as genuinely unused — `onChange` in `multi-select`'s `handleKeyDown` (it only calls `setSelected`), `survey.workspaceId` in `share-survey-modal` (`survey` itself is already a dep), `workspace.id` in `MainNavigation`'s `handleSettingsOrganizationChange`.
- `survey-client-wrapper`: extracted the complex dependency expression into `hiddenFieldIdsKey`, so the dep is statically checkable while the memo stays keyed on the field ids' **contents** rather than the array identity — the semantics `main` deliberately chose. That also made `main`'s `react-hooks/use-memo` disable unused, so it goes.
- `useDocumentVisibility`: moved `onVisible` into a ref updated by an effect, so the listener always calls the current callback without re-subscribing. This replaced a suppression.
- `ResponsePage`: adding `survey` to the `filters` memo is the correct dep, but `survey` is an RSC prop, so every `router.refresh()` on the analysis page gave `filters` a new identity for unchanged content. `filters` is now held at a stable identity keyed on the serialized filter value (`filtersKey`, itself memoized), so nothing downstream reacts to a refresh: not the fetch effect that calls `setPage(1)` and would collapse a scrolled list back to page 1, and not `fetchNextPage` / `refetchResponses`, whose identities were re-registering the analysis refresh handler on every refresh. Both caught in review. `CustomFilter` has the same memo change but its `filters` feeds only the download handler — no effect, no callback deps — so it needs nothing.
- Notion `AddIntegrationModal`: `dbItems` / `elementItems` are back on `selectedDatabase?.id` / `selectedSurvey?.id`, as on `main`. The seeding effect hands them a fresh object literal whenever the `surveys` / `databases` server props change identity, so keying on the object recomputed both lists for identical content. Documented suppressions rather than dep arrays that assert something false.
- `survey-editor`: the init effect no longer depends on the state it sets. `localSurvey` is seeded in its own `useState` initializer, so `if (localSurvey) return;` was already true on the first render — adding `localSurvey` to the deps bought a re-run on every keystroke for a body that could never execute, in the shape that becomes a real loop if the guard is ever edited. Now `setLocalSurvey((current) => current ?? structuredClone(survey))` on `[survey]`, which keeps the (unreachable-today) recovery the `LoadingSkeleton` guard implies.
- `contact-data-view`: fixed a bug the review surfaced next to one of those suppressions. The workspace-sync effect armed `isResettingSearch` unconditionally and then called `setSearchValue("")`; on the common path — switching workspace without having searched — that setter is a no-op, so the `[searchValue]` effect never ran to lower the flag, and the user's *next* keystroke was consumed as the reset it was still waiting for. The flag is now armed only when there is a non-empty search to clear. Pre-existing on `main`.
- A documented `// eslint-disable-next-line react-hooks/exhaustive-deps -- <reason>` only for intentional mount-only / single-trigger effects, where adding the deps would loop, re-fire toasts and uploads, or clobber user state (element-form sync effects, localStorage column settings, OAuth-error toast, editor seeding).

`useDocumentVisibility`, `ResponsePage`, the `survey-editor` init effect and the `contact-data-view` reset flag are the only behavioural changes in the diff; everything else is dependency bookkeeping or a comment.

<details>
<summary>Why this rebase dropped the original "stale eslint-disable" half of the PR</summary>

The branch was cut in July and also removed 15 `eslint-disable` directives that `--report-unused-disable-directives` flagged at the time. Re-derived against current `main`:

- `@formbricks/i18n-utils` and `@formbricks/vite-plugins` — `main` has since removed those directives itself, so the hunks are gone.
- `packages/types` (`elements-validation.ts`, `types.ts`, `validation.ts`) and `packages/cache` (`service.ts`, `service.test.ts`) — the types these guard have changed on `main`, so the directives are **load-bearing again**. Removing them now produces 4 + 6 real ESLint errors, so they are restored and this PR leaves them alone.

</details>

## Breaking changes

None. Every change is internal to `apps/web` components and one `apps/web/lib` hook — no public API, route, env var, config key, or SDK export in `packages/js-core`, `packages/survey-ui` or `packages/surveys` is touched.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `useDocumentVisibility` calls the latest `onVisible` after a re-render, fires it exactly once, ignores a hidden document, and detaches on unmount | unit (mutation) | New `apps/web/lib/useDocumentVisibility.test.ts`, 4 tests. `npx vitest run lib/useDocumentVisibility.test.ts` from `apps/web`. Mutating `apps/web/lib/useDocumentVisibility.ts:15` back to `onVisible()` fails "invokes the latest callback after a re-render, exactly once" (1 failed / 3 passed), so the test fails if the fix is undone. Also the only file in this diff Sonar measures coverage for — new-code coverage 0% → 100%, quality gate `OK`, 0 new issues. |
| Responses list survives an unrelated `router.refresh()` (survey status toggle, share modal) without resetting to page 1, and without re-registering the analysis refresh handler | manual (inspection) | Traced the chain: `SurveyStatusDropdown.tsx:43` → `router.refresh()` → new `survey` prop identity → new `filters` identity → fetch effect's `setPage(1)`, plus new `fetchNextPage`/`refetchResponses` identities → the `registerAnalysisRefreshHandler` effect re-running. Holding `filters` at an identity keyed on `filtersKey` breaks both links while keeping the correct memo dep. Not exercised at runtime — see Open gaps. |
| `filtersKey` is not re-serialized on every render | inspection | `useMemo(() => JSON.stringify(computedFilters), [computedFilters])`, and `computedFilters` is memoized on `[survey, selectedFilter, dateRange]` — none of which a scroll fetch or a row/tag edit touches. |
| The survey editor still seeds `localSurvey` and the first active element | manual (inspection) | The seeding path is the `useState` initializer, unchanged. The effect is recovery-only and was already unreachable. The first active element comes from the `[localSurvey?.type]` effect at :156, which fires on mount and on any null → set transition, so removing that call from the init effect drops nothing. Not exercised at runtime — see Open gaps. |
| Notion field mapping is not recomputed for unchanged content | inspection | Both memos keyed back on ids; the effect at :160 that re-seeds `selectedDatabase`/`selectedSurvey` with fresh literals is what makes object-keying wrong here. Runtime flow still uncovered — see Open gaps. |
| Signup password checklist still flips per keystroke after `DEFAULT_VALIDATIONS` was memoized | manual | Local dev server, `/auth/signup` → "Continue with Email". Typing `abc` leaves all three rules unticked; extending it to `Defgh123` ticks all three. No visual change intended, so no screenshot. |
| ~20 dependency-array changes across the survey editor, rich-text toolbar, signup, segments and survey overview | e2e | CI Playwright suite (`Run E2E Tests`) green on this commit — same suite that was green before the rebase. |
| `hiddenFieldsRecord` memo after extracting `hiddenFieldIdsKey` | manual (inspection) | The dependency **value** is unchanged: `hiddenFieldIdsKey` is the identical `JSON.stringify(survey.hiddenFields.fieldIds \|\| [])` expression hoisted to a `const` on the line above. Not exercised at runtime — see Open gaps. |

`pnpm lint` 21/21 · `pnpm typecheck` 27/27 · `pnpm test` 26/26 (8542 web tests) locally, in a clean worktree rebased on `main`.

**Open gaps**

- **Notion / Airtable integration modals** — reviewed by inspection only; the dev environment has no third-party integration credentials to drive the field-mapping flow. The Notion dep changes have since been reverted to `main`'s id-keyed memos (see *What & why*), so what is uncovered there is now `main`'s behaviour plus two comments rather than a change. Those comments record the trade-off: id-keying is stale if a refresh returns different properties for the same database id, and the real fix is the seeding effect not replacing `selectedDatabase` with a fresh literal for unchanged content — a follow-up, since this modal cannot be exercised here.
- **Responses infinite scroll** — the `filtersKey` fix was reasoned from the call chain, not observed in a browser; verifying it needs a survey with enough responses to scroll past the first page, then toggling survey status.
- **Link-survey hidden fields** — no local survey has hidden fields configured, so the URL→hidden-field extraction in `survey-client-wrapper` was not exercised at runtime.
- **The contacts first-search fix is verified by reading, not by running** — it needs contacts seeded across two workspaces. Per the repo's testing rules the fix does not get a `.tsx` unit test, so it wants the manual pass: search in workspace A, switch to workspace B, type one character, confirm the list filters.
- **Not exercised at runtime, reviewed by inspection**: EE contacts/attributes localStorage column settings and debounced contact search, response filters, file-input image/video tab switching, `multi-select`, `preview-survey`, the chart dropdown's add-to-dashboard fetch, and the rich-text toolbar's seeding effects. The Airtable `AddIntegrationModal` `[isEditMode]` effect CodeRabbit flagged is pre-existing on `main` — this PR only documents the exception, and re-keying it to the selected record belongs in its own change.

**Risks**

- Review found four dep arrays that were correct by the linter's reading but keyed on RSC prop identities that change on every refresh (three in `ResponsePage`/Notion, one re-running the editor's init effect per keystroke). All four are fixed; the class is worth re-checking if anything else in this diff looks chatty.
- The failure mode of a wrong dependency array is a stale closure or a render loop, not a crash. If something regresses, look first at the effects that write back into the survey (`address-element-form`, `contact-info-element-form`, `cal-element-form`), the localStorage-backed contacts/attributes column settings, the debounced contacts search, and the one-time init effects (OAuth-error toast, editor seeding, script load).
- The restored `packages/types` / `packages/cache` disable comments are load-bearing against the current types. If a later change makes one unnecessary again, `--report-unused-disable-directives` will flag it rather than let it rot.

---

> [!NOTE]

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/airtable/components/AddIntegrationModal.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/airtable/components/AddIntegrationModal.tsx
@@ -216,6 +216,7 @@ export const AddIntegrationModal = ({
     } else {
       reset();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- only re-seed the form when edit mode toggles; adding the other deps would reset user edits
   }, [isEditMode]);
 
   const survey = watch("survey");
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
@@ -110,7 +110,11 @@ export const AddIntegrationModal = ({
         type: dbProperties[fieldKey].type,
       })) || []
     );
-  }, [selectedDatabase]);
+    // The effect below re-seeds `selectedDatabase` with a fresh object literal whenever the
+    // `databases`/`surveys` server props change identity, which an RSC refresh does on unchanged
+    // content. Keying on the id keeps identical content from recomputing this list.
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the database's identity, not the object's
+  }, [selectedDatabase?.id]);
 
   const elementItems = useMemo(() => {
     const mappedElements = selectedSurvey
@@ -155,7 +159,10 @@ export const AddIntegrationModal = ({
     }));
 
     return [...mappedElements, ...variables, ...hiddenFields, ...Metadata, ...createdAt, ...personAttributes];
-  }, [contactAttributeKeys, elements, selectedSurvey, t]);
+    // Same as `dbItems` above: `selectedSurvey` is re-seeded from the `surveys` server prop, so its
+    // identity changes on a refresh that changed nothing.
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the survey's identity, not the object's
+  }, [contactAttributeKeys, elements, selectedSurvey?.id, t]);
 
   useEffect(() => {
     if (selectedIntegration) {
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
@@ -110,7 +110,7 @@ export const AddIntegrationModal = ({
         type: dbProperties[fieldKey].type,
       })) || []
     );
-  }, [selectedDatabase?.id]);
+  }, [selectedDatabase]);
 
   const elementItems = useMemo(() => {
     const mappedElements = selectedSurvey
@@ -155,7 +155,7 @@ export const AddIntegrationModal = ({
     }));
 
     return [...mappedElements, ...variables, ...hiddenFields, ...Metadata, ...createdAt, ...personAttributes];
-  }, [selectedSurvey?.id, contactAttributeKeys]);
+  }, [contactAttributeKeys, elements, selectedSurvey, t]);
 
   useEffect(() => {
     if (selectedIntegration) {
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
@@ -113,6 +113,13 @@ export const AddIntegrationModal = ({
     // The effect below re-seeds `selectedDatabase` with a fresh object literal whenever the
     // `databases`/`surveys` server props change identity, which an RSC refresh does on unchanged
     // content. Keying on the id keeps identical content from recomputing this list.
+    //
+    // The trade-off, stated so it is a choice rather than an oversight: if a refresh brings back
+    // *different* properties for the same database id — someone edited the Notion database's schema
+    // while this modal was open on it — the list here stays stale until the database is reselected.
+    // Fixing that by keying on the object trades a rare staleness for a recompute on every refresh;
+    // fixing it properly means not re-seeding with a fresh literal when the content is unchanged,
+    // which belongs in the effect rather than in this dep array.
     // eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the database's identity, not the object's
   }, [selectedDatabase?.id]);
 
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
@@ -51,7 +51,7 @@ export const ResponsePage = ({
   const filters = useMemo(
     () => getFormattedFilters(survey, selectedFilter, dateRange),
 
-    [selectedFilter, dateRange]
+    [survey, selectedFilter, dateRange]
   );
 
   const searchParams = useSearchParams();
@@ -107,7 +107,7 @@ export const ResponsePage = ({
     } finally {
       setIsFetchingFirstPage(false);
     }
-  }, [filters, responsesPerPage, surveyId]);
+  }, [filters, responsesPerPage, surveyId, t]);
 
   useEffect(() => {
     return registerAnalysisRefreshHandler(refetchResponses);
@@ -162,6 +162,8 @@ export const ResponsePage = ({
     };
     fetchFilteredResponses();
     // page is intentionally omitted to avoid refetching after the initial page setup.
+    // hasFilters is derived from selectedFilter/dateRange which are already deps.
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- effect must run only on filter changes, not on page updates it sets internally
   }, [filters, responsesPerPage, selectedFilter, dateRange, surveyId]);
 
   return (
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
@@ -54,6 +54,12 @@ export const ResponsePage = ({
     [survey, selectedFilter, dateRange]
   );
 
+  // The fetch effect below is keyed on the filter *value*, not on `filters`' identity: `survey` is an
+  // RSC prop, so every `router.refresh()` (survey status dropdown, share modal, reset survey) hands
+  // down a freshly deserialized object. Re-running on identity alone would call setPage(1) and
+  // collapse the infinite-scroll list back to the first page on an unrelated refresh.
+  const filtersKey = JSON.stringify(filters);
+
   const searchParams = useSearchParams();
 
   const fetchNextPage = useCallback(async () => {
@@ -163,8 +169,8 @@ export const ResponsePage = ({
     fetchFilteredResponses();
     // page is intentionally omitted to avoid refetching after the initial page setup.
     // hasFilters is derived from selectedFilter/dateRange which are already deps.
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- effect must run only on filter changes, not on page updates it sets internally
-  }, [filters, responsesPerPage, selectedFilter, dateRange, surveyId]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- effect must run only when the filter value changes, not on page updates it sets internally nor on a new `filters` identity from an RSC refresh
+  }, [filtersKey, responsesPerPage, selectedFilter, dateRange, surveyId]);
 
   return (
     <>
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
@@ -48,17 +48,23 @@ export const ResponsePage = ({
   const [isFetchingFirstPage, setIsFetchingFirstPage] = useState<boolean>(false);
   const { selectedFilter, dateRange, resetState, registerAnalysisRefreshHandler } = useResponseFilter();
   const { t } = useTranslation();
-  const filters = useMemo(
+  const computedFilters = useMemo(
     () => getFormattedFilters(survey, selectedFilter, dateRange),
 
     [survey, selectedFilter, dateRange]
   );
 
-  // The fetch effect below is keyed on the filter *value*, not on `filters`' identity: `survey` is an
-  // RSC prop, so every `router.refresh()` (survey status dropdown, share modal, reset survey) hands
-  // down a freshly deserialized object. Re-running on identity alone would call setPage(1) and
-  // collapse the infinite-scroll list back to the first page on an unrelated refresh.
-  const filtersKey = JSON.stringify(filters);
+  // `survey` is an RSC prop, so every `router.refresh()` (survey status dropdown, share modal, reset
+  // survey) hands down a freshly deserialized object and `computedFilters` takes a new identity for
+  // unchanged content. Memoized so the serialization happens when the filters actually recompute
+  // rather than on every render of this page — each scroll fetch, each row or tag edit.
+  const filtersKey = useMemo(() => JSON.stringify(computedFilters), [computedFilters]);
+
+  // The identity everything downstream keys on, held steady while the filter *value* is unchanged.
+  // Without it a refresh re-creates `fetchNextPage` and `refetchResponses`, which re-registers the
+  // analysis refresh handler and would collapse the infinite-scroll list back to page 1.
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the filter value, not its identity
+  const filters = useMemo(() => computedFilters, [filtersKey]);
 
   const searchParams = useSearchParams();
 
@@ -169,7 +175,7 @@ export const ResponsePage = ({
     fetchFilteredResponses();
     // page is intentionally omitted to avoid refetching after the initial page setup.
     // hasFilters is derived from selectedFilter/dateRange which are already deps.
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- effect must run only when the filter value changes, not on page updates it sets internally nor on a new `filters` identity from an RSC refresh
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- effect must run only when the filter value changes, not on page updates it sets internally
   }, [filtersKey, responsesPerPage, selectedFilter, dateRange, surveyId]);
 
   return (
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx
@@ -158,7 +158,7 @@ export const CustomFilter = ({ survey }: CustomFilterProps) => {
   const filters = useMemo(
     () => getFormattedFilters(survey, selectedFilter, dateRange),
 
-    [selectedFilter, dateRange]
+    [survey, selectedFilter, dateRange]
   );
 
   const datePickerRef = useRef<HTMLDivElement>(null);
--- a/apps/web/lib/useDocumentVisibility.ts
+++ b/apps/web/lib/useDocumentVisibility.ts
@@ -1,11 +1,18 @@
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 
 // This hook will listen to the visibilitychange event and run the provided function whenever the document's visibility state changes to visible
 export const useDocumentVisibility = (onVisible: () => void) => {
+  // Keep the latest callback in a ref so the listener always calls the current
+  // `onVisible` without having to re-subscribe on every render.
+  const onVisibleRef = useRef(onVisible);
+  useEffect(() => {
+    onVisibleRef.current = onVisible;
+  }, [onVisible]);
+
   useEffect(() => {
     const listener = () => {
       if (document.visibilityState === "visible") {
-        onVisible();
+        onVisibleRef.current();
       }
     };
 
@@ -14,6 +21,5 @@ export const useDocumentVisibility = (onVisible: () => void) => {
     return () => {
       document.removeEventListener("visibilitychange", listener);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- subscribe once on mount; onVisible may be recreated each render and must not re-subscribe the listener
   }, []);
 };
--- a/apps/web/lib/useDocumentVisibility.ts
+++ b/apps/web/lib/useDocumentVisibility.ts
@@ -14,5 +14,6 @@ export const useDocumentVisibility = (onVisible: () => void) => {
     return () => {
       document.removeEventListener("visibilitychange", listener);
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- subscribe once on mount; onVisible may be recreated each render and must not re-subscribe the listener
   }, []);
 };
--- a/apps/web/modules/ee/contacts/components/contact-data-view.tsx
+++ b/apps/web/modules/ee/contacts/components/contact-data-view.tsx
@@ -105,6 +105,7 @@ export const ContactDataView = ({
     if (isResettingSearch.current) {
       isResettingSearch.current = false;
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- debounced search must fire only on searchValue change; `fetchContactsFromStart` is recreated each render and would re-trigger the search
   }, [searchValue]);
 
   useEffect(() => {
--- a/apps/web/modules/ee/contacts/components/contact-data-view.tsx
+++ b/apps/web/modules/ee/contacts/components/contact-data-view.tsx
@@ -43,11 +43,20 @@ export const ContactDataView = ({
       prevWorkspaceId.current = workspaceId;
       setContacts([...initialContacts]);
       setHasMore(initialHasMore);
-      isResettingSearch.current = true;
+      // Arm the skip only when there is a search to clear. `setSearchValue("")` on an already-empty
+      // value is a no-op, so the `[searchValue]` effect never runs to lower the flag again — and the
+      // next time the user does type, that effect consumes their first search as the "reset" it was
+      // waiting for. Switching workspace without having searched is the common path, so the flag was
+      // usually left armed.
+      if (searchValue) {
+        isResettingSearch.current = true;
+      }
       setSearchValue("");
       prevInitialContactsLength.current = initialContacts.length;
     }
-  }, [workspaceId, initialContacts, initialHasMore]);
+    // `searchValue` is read above, so it belongs here; the whole body is behind the workspace-change
+    // guard, which updates `prevWorkspaceId` immediately, so the extra runs are a comparison.
+  }, [workspaceId, initialContacts, initialHasMore, searchValue]);
 
   // Sync state when initialContacts changes from server refresh (e.g., after CSV upload)
   // Only update if we're viewing the first page without search
--- a/apps/web/modules/ee/contacts/components/contact-data-view.tsx
+++ b/apps/web/modules/ee/contacts/components/contact-data-view.tsx
@@ -105,7 +105,7 @@ export const ContactDataView = ({
     if (isResettingSearch.current) {
       isResettingSearch.current = false;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- debounced search must fire only on searchValue change; `fetchContactsFromStart` is recreated each render and would re-trigger the search
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- debounced search must fire only on searchValue change; `fetchContactsFromStart` is keyed on `searchValue` itself, so listing it would re-trigger the search whenever `workspaceId` or `itemsPerPage` changed too
   }, [searchValue]);
 
   useEffect(() => {
--- a/apps/web/modules/survey/editor/components/survey-editor.tsx
+++ b/apps/web/modules/survey/editor/components/survey-editor.tsx
@@ -119,20 +119,16 @@ export const SurveyEditor = ({
 
   useDocumentVisibility(fetchLatestWorkspaceData);
 
+  // Recovery only: `localSurvey` is seeded from `survey` in its `useState` initializer, so this is a
+  // no-op unless something ever resets it to null (the `LoadingSkeleton` guard below is the state it
+  // recovers from). Written as an updater rather than reading `localSurvey`, because depending on it
+  // would re-run this on every keystroke in the editor to do nothing — and that shape becomes a real
+  // loop the moment someone edits the guard. The active element is not set here: the
+  // `[localSurvey?.type]` effect below already picks the first element whenever `localSurvey`
+  // appears.
   useEffect(() => {
-    if (survey) {
-      if (localSurvey) return;
-
-      const surveyClone = structuredClone(survey);
-      setLocalSurvey(surveyClone);
-
-      // Set first element from first block
-      const firstBlock = survey.blocks[0];
-      if (firstBlock) {
-        setActiveElementId(firstBlock.elements?.[0]?.id);
-      }
-    }
-  }, [survey, localSurvey]);
+    setLocalSurvey((current) => current ?? structuredClone(survey));
+  }, [survey]);
 
   useEffect(() => {
     const listener = () => {
--- a/apps/web/modules/survey/editor/components/survey-editor.tsx
+++ b/apps/web/modules/survey/editor/components/survey-editor.tsx
@@ -132,7 +132,7 @@ export const SurveyEditor = ({
         setActiveElementId(firstBlock.elements?.[0]?.id);
       }
     }
-  }, [survey]);
+  }, [survey, localSurvey]);
 
   useEffect(() => {
     const listener = () => {
@@ -158,6 +158,7 @@ export const SurveyEditor = ({
     if (firstBlock) {
       setActiveElementId(firstBlock.elements[0]?.id);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- intentionally resets active element only when the survey type changes, not on every block edit
   }, [localSurvey?.type]);
 
   useEffect(() => {
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -109,7 +109,7 @@ export const ManageTranslationsModal = ({
       if (!aEmpty && bEmpty) return 1;
       return 0;
     });
-  }, [strings, missingFirst]);
+  }, [strings, missingFirst, isDraftEmpty]);
 
   // Merge draft translations into localSurvey so that the recall dropdown
   // can see in-progress translations (e.g. translated headlines).
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -109,7 +109,8 @@ export const ManageTranslationsModal = ({
       if (!aEmpty && bEmpty) return 1;
       return 0;
     });
-  }, [strings, missingFirst, isDraftEmpty]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- `isDraftEmpty` is intentionally excluded so rows don't re-sort on every keystroke; the order snapshots on strings/missingFirst changes
+  }, [strings, missingFirst]);
 
   // Merge draft translations into localSurvey so that the recall dropdown
   // can see in-progress translations (e.g. translated headlines).
--- a/apps/web/app/(app)/workspaces/[workspaceId]/components/MainNavigation.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/components/MainNavigation.tsx
@@ -353,7 +353,7 @@ export const MainNavigation = ({
         }
       });
     },
-    [router, organization.id, workspace.id]
+    [router, organization.id]
   );
 
   const switcherTriggerClasses = cn(
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/account/notifications/components/NotificationSwitch.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/account/notifications/components/NotificationSwitch.tsx
@@ -114,6 +114,7 @@ export const NotificationSwitch = ({
           break;
       }
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- run once on mount; re-running would re-fire the switch change and toast
   }, []);
 
   return (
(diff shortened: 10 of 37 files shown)


apps/web/lib/useDocumentVisibility.ts
```diff
@@ -14,5 +14,6 @@ export const useDocumentVisibility = (onVisible: () => void) => {
     return () => {
       document.removeEventListener("visibilitychange", listener);
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- subscribe once on mount; onVisible may be recreated each render and must not re-subscribe the listener
   }, []);
```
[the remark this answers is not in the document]
Dhruwang: Fixed in 71329f774a — `onVisible` now lives in a ref updated via an effect, and the listener calls `onVisibleRef.current`, so a later `fetchLatestWorkspaceData` (which depends on `localWorkspace.id`) is always current. Removed the eslint-disable.

apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
```diff
@@ -109,7 +109,7 @@ export const ManageTranslationsModal = ({
       if (!aEmpty && bEmpty) return 1;
       return 0;
     });
-  }, [strings, missingFirst]);
+  }, [strings, missingFirst, isDraftEmpty]);
```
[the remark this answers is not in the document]
Dhruwang: Fixed in 71329f774a — dropped `isDraftEmpty` from the `displayStrings` deps (kept `[strings, missingFirst]`), so the sort order snapshots when the string set or sort mode changes and no longer reshuffles on every keystroke. Good catch — the earlier revision added that dep and contradicted the snapshot comment.

apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/airtable/components/AddIntegrationModal.tsx
```diff
@@ -216,6 +216,7 @@ export const AddIntegrationModal = ({
     } else {
       reset();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- only re-seed the form when edit mode toggles; adding the other deps would reset user edits
   }, [isEditMode]);
```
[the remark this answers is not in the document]
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/airtable/components/AddIntegrationModal.tsx:220]: Skipping in this PR. This is pre-existing behavior — `main` already has `[isEditMode]` here; this PR only added the eslint-disable comment, not the dependency list. Re-keying the effect to the selected record is a behavior change beyond the scope of this lint-warning cleanup, and better tracked as its own fix.

itsjavi (commented):
Went through the diff (41 files, mostly React hook dependency arrays + a few stale `eslint-disable` removals). Spot-checked the ones with the highest regression risk:

- `MainNavigation.tsx`: `workspace.id` removed from `handleSettingsOrganizationChange` deps — confirmed unused in the callback body, safe.
- `share-survey-modal.tsx`: `survey.workspaceId` dropped but `survey` itself stays in the deps array, and the newly-added `enterpriseLicenseRequestFormUrl` is genuinely read inside the memo — correct fix.
- `survey-editor.tsx`: `localSurvey` added to the survey-init effect's deps — the effect early-returns via `if (localSurvey) return;`, so this only adds a redundant re-check, not a behavior change.
- `multi-select/index.tsx`: `onChange` removed from `handleUnselect` deps — confirmed it doesn't call `onChange` directly (goes through `setSelected`, synced separately), safe.
- Notion `AddIntegrationModal.tsx`: `selectedDatabase`/`selectedSurvey` swapped in for `.id` — both are plain `useState` values, so no memoization/re-render regression.

No issues found. The CodeRabbit threads on `AddIntegrationModal.tsx` (Airtable), `useDocumentVisibility.ts`, and `manage-translations-modal.tsx` are already resolved with reasonable responses (one explicitly deferred as pre-existing/out-of-scope, two fixed with a ref-based pattern and a corrected dep array).

One thing worth noting: SonarCloud's quality gate is currently failing on "0% coverage on new code" — expected for a pure lint/dependency-array cleanup with no new logic, but flagging so it's not overlooked before merge.

Since this only touches dependency arrays and disable comments with no new logic, this is reviewable from the diff alone — no server/DB needed to assess the changes I checked above. That said, actually confirming zero behavioral regressions across 20+ touched effects (survey editor, contacts tables, integrations, file upload) needs manual QA per the PR's own test plan checklist — that's for a human reviewer with a running app, @itsjavi.

_Reviewed by Claude (claude-sonnet-5, medium effort)._

---

itsjavi (commented):
Reviewed the diff (41 files — dependency-array fixes across `apps/web` plus cleanup of stale `eslint-disable` directives in `@formbricks/cache`, `@formbricks/i18n-utils`, `@formbricks/types`, `@formbricks/vite-plugins`). Spot-checked several of the hook-dependency changes for staleness/loop risk (`MainNavigation`, Notion `AddIntegrationModal`, `ResponsePage`/`CustomFilter`, `useDocumentVisibility`) against the actual source and didn't find any new correctness issues beyond what's already resolved in this thread. The one item CodeRabbit flagged and the author deferred (Airtable `AddIntegrationModal` re-seed effect, `[isEditMode]` deps) checks out as pre-existing on `main`, so pushing it to a follow-up is reasonable.

**Needs attention**
- SonarCloud Quality Gate is currently **failing**: 0.0% coverage on new code (required ≥80%). Worth a look even on a lint-only PR — some added lines (e.g. the new `useMemo` in `password-checks.tsx`) count as new code for that check.
- Still marked as a **draft** — flagging in case that's not intentional at this point.

**Needs human review**
- This PR touches ~20 `react-hooks/exhaustive-deps` fixes across effects/memos in the survey editor, EE contacts tables, Notion/Airtable integrations, and file/rich-text components. I can't spin up the app or a DB in this environment to exercise it, so the manual QA checklist already in the PR description (survey editor persistence, EE contacts column settings, integration field mapping, share-survey modal, file input tabs) needs a human pass. @itsjavi

---

itsjavi (commented):
Spot-checked the dependency-array changes most likely to hide a stale-closure bug or a silently-broken memo: `MainNavigation`'s `handleSettingsOrganizationChange`, `ShareSurveyModal`'s `linkTabs` memo (removed `survey.workspaceId`, added `enterpriseLicenseRequestFormUrl`), the Notion `AddIntegrationModal`'s `dbItems`/`elementItems` memos (dep swapped from an id to the object), and `MultiSelect`'s `handleKeyDown` (dropped `onChange`). In each case the new deps match exactly what's read in the function body, and the objects swapped in (`selectedDatabase`, `selectedSurvey`) are stable state references rather than values recreated every render, so these are genuine fixes rather than lint-satisfying regressions. CI is fully green (lint, unit tests, build, E2E) and there are no open Sonar or CodeRabbit findings.

Since this is still a draft and several of the changed effects back persisted UI state (contacts/attributes column settings in localStorage, debounced contact search, Notion/Airtable field-mapping modals) that I can't exercise without the running app + DB in this environment, the PR's own "How to test" checklist is worth a manual pass before it comes out of draft. @itsjavi

---

---

itsjavi (changes requested):
Pls resolve conflicts and move it out of Draft whenever it is ready for review.

Sonar seems to fail as well.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
```diff
@@ -51,7 +51,7 @@ export const ResponsePage = ({
   const filters = useMemo(
     () => getFormattedFilters(survey, selectedFilter, dateRange),
 
-    [selectedFilter, dateRange]
+    [survey, selectedFilter, dateRange]
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx:54]: Adding `survey` here is the right stale-closure fix, but it re-keys `filters` to a prop that changes identity on every RSC re-render — and `filters` is the first dependency of the effect at line 132, which calls `setPage(1)` and replaces `responses` with page 1.

`responses/page.tsx:78` renders `SurveyAnalysisCTA` next to `ResponsePage`, and that component calls `router.refresh()` (`SurveyAnalysisCTA.tsx:148`, `:171`, plus the share-modal tabs). After a refresh the server component hands down a freshly deserialized `survey` object, so:

`router.refresh()` → new `survey` identity → new `filters` identity → effect at :132 refires → `setPage(1)` + `setResponses(page 1)`.

Net effect: on a responses list the user has scrolled through, toggling survey status or using the share modal collapses the infinite-scroll list back to the first page and spends an extra `getResponsesAction` call. Before this PR, `survey` changes didn't re-key `filters`, so the refresh was invisible.

The dep is correct; the effect is the part that's too broadly keyed. Options: memoize the filter *value* (e.g. key the effect on a serialized `filters` string), or drop the `setPage(1)`/reset when the recomputed filters are value-equal to the previous ones.

Not necessarily blocking for a lint-cleanup PR — but worth either fixing here or noting under `Open gaps`, since it's the one dep change in the diff that has an observable behavioural consequence beyond `useDocumentVisibility`.

---

kiln-agent (commented):
Reviewed the rebased head `eedd4ab` (32 files, +148/−38, all in `apps/web`) at high effort. This is a much tighter diff than the July version — the `packages/*` disable-directive half is gone, and the reasoning for restoring the load-bearing ones in `packages/types` / `packages/cache` checks out.

**One finding** — inline on `ResponsePage.tsx:54`. Adding `survey` to the `filters` memo re-keys the fetch effect at :132 to a prop whose identity changes on every `router.refresh()`, and that effect resets the infinite-scroll list to page 1. `SurveyAnalysisCTA` (which calls `router.refresh()`) is rendered on the same page. Correct dep, too-broad effect key — details and options inline.

**Verified and cleared**

- **`useDocumentVisibility` ref refactor** — correct latest-callback pattern, single subscription, cleanup intact. One consumer (`survey-editor.tsx:120`), which is exactly the case that benefits. The new test file passes locally (4/4) and the `@vitest-environment jsdom` pragma is required and works, given `vite.config.mts` files `*.test.ts` under the `node` project.
- **The three removed deps** — all genuinely unread in their callback bodies: `workspace.id` in `MainNavigation`'s `handleSettingsOrganizationChange`, `survey.workspaceId` in `share-survey-modal` (`survey` itself stays), `onChange` in `multi-select`'s `handleKeyDown` (it only calls `setSelected`).
- **`t` added as a dep** (`chart-dropdown-menu`, `ResponsePage`, `elements-view`, `follow-up-item`, `template-tags`) — safe. react-i18next serves `t` via `useSyncExternalStore` with a cached snapshot, so it's referentially stable unless language or namespace changes. The `chart-dropdown-menu` fetch effect was the one that could have looped; it doesn't.
- **`survey-editor.tsx:135`** (`[survey, localSurvey]`) — the `if (localSurvey) return;` guard makes the extra runs no-ops, and it lets the component recover if `localSurvey` were ever reset to null. Net improvement.
- **`survey-client-wrapper` `hiddenFieldIdsKey`** — value-identical to the inline `JSON.stringify(...)` it replaced, evaluated at the same point in the render. Keeps the contents-keyed semantics `main` chose, and correctly retires the `react-hooks/use-memo` disable.
- **Notion `AddIntegrationModal`** — `selectedDatabase` / `selectedSurvey` are `useState` values, so stable identity; `dbItems`/`elementItems` feed render props only and `MappingRow` has no effects. No re-fetch loop.
- **`element-form-input` `elementId` memo** — the added deps fix a real stale value; I checked the out-of-range `localSurvey.endings[elementIdx - elements.length]` path and the parent updates `elementIdx` and `endings` in the same render, so no new crash window.
- **The 17 added `eslint-disable-next-line` directives** — all comment-only, none changes a dep array, all correctly placed on the line before the array. `npx eslint` across all 32 changed files reports zero warnings and zero errors, so none is stale under `reportUnusedDisableDirectives`.

**Nit (non-blocking, comment text only)**

`apps/web/modules/ee/contacts/components/contact-data-view.tsx:104` — the suppression says `fetchContactsFromStart` "is recreated each render", but it's a `useCallback` keyed on `[workspaceId, itemsPerPage, searchValue]`. Behaviour is unaffected; only the stated rationale is wrong. The equivalent claims in `preview-survey` and `ResponseFilter` are accurate.

**Housekeeping**

Sonar's quality gate now passes (100% coverage on new code) — the concern raised in the earlier review rounds is resolved by the `useDocumentVisibility` test. The PR is out of draft, but `mergeable_state` is `blocked` and there's a standing changes-requested review from @itsjavi asking for conflicts to be resolved; both look worth a re-check before merge. The `Open gaps` list is honest about what wasn't exercised at runtime, which is the right call for a diff of this shape.

---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx:54]: Adding `survey` here is correct in principle — `getFormattedFilters` does read it — but it hands `filters` a new identity on every RSC refresh, and `filters` is the first dep of the fetch effect at line 166. That effect calls `setPage(1)` and replaces `responses` wholesale, so infinite scroll resets to page 1 plus an extra `getResponsesAction` request.

This page renders `SurveyAnalysisCTA` (`responses/page.tsx:78`), which in turn renders `SurveyStatusDropdown` (`SurveyAnalysisCTA.tsx:266`). Pausing/resuming a survey from that dropdown calls `router.refresh()` (`SurveyStatusDropdown.tsx:43`), the server component re-renders and `survey` arrives as a fresh object across the RSC payload. Before this change the memo was keyed on `[selectedFilter, dateRange]` only, so a refresh left the list alone; now a user who has scrolled several pages into the responses list loses their position when they toggle survey status.

(The other two `router.refresh()` callers on this page — reset survey and generate example responses — genuinely *want* a refetch, so they're fine either way.)

Keying the memo on the fields `getFormattedFilters` actually reads instead of the whole object would keep the lint fix without the reset.


---

kiln-agent (commented):
Reviewed `fix/lint-warnings` at `eedd4ab` (32 files, `6fa67c6...eedd4ab`) — an ESLint `react-hooks/exhaustive-deps` cleanup across `apps/web`.

One finding, inline on `ResponsePage.tsx`: adding `survey` to the `filters` memo makes it change identity on every `router.refresh()`, which re-fires the fetch effect and resets infinite scroll to page 1. The status dropdown on this very page triggers that path. Correct dep, unwanted side effect — keying on the fields `getFormattedFilters` reads would keep the fix without the reset.

**What I checked and found clean**

- Ran ESLint over all 32 changed files: 0 errors, 0 warnings, no unused disable directives.
- `useDocumentVisibility.ts` — the ref-based rewrite is a genuine fix, not a lint workaround: the previous `[]` effect captured the first-render `onVisible` permanently. New test file passes 4/4.
- Every **removed** dep is genuinely unreferenced in its closure — `workspace.id` in `MainNavigation`'s `handleSettingsOrganizationChange`, `onChange` in `multi-select`'s `handleKeyDown`, `survey.workspaceId` in `share-survey-modal` (superseded by the `survey` dep itself).
- Every **added** dep checked for render/fetch loops. None besides the one above: `survey-editor`'s `[survey, localSurvey]` effect early-returns, `elements-view`'s toast effect is guarded by an id comparison, and the `chart-dropdown-menu` / Notion modal memos feed nothing that writes back into their own deps.
- The `t` identity added to ~6 dep arrays is stable — react-i18next 16.5.8 serves `t` via `useSyncExternalStore` with a cached snapshot, so it changes on language/namespace revision, not per render. Worth stating explicitly, because without that the `chart-dropdown-menu` fetch effect would loop.
- `survey-client-wrapper`'s `hiddenFieldIdsKey` extraction is value-identical to the inline `JSON.stringify(...)` it replaced; `password-checks` and `template-tags` are pure memo/callback hoists.
- The 17 comment-only hunks adding `eslint-disable-next-line` are correctly positioned and document pre-existing intentional behavior — none suppresses a defect this PR introduces.

**Needs a human**

Per the PR's own Open gaps, the Notion/Airtable field-mapping flows and link-survey hidden fields weren't exercised at runtime, and I can't drive them here either. The failure mode of a wrong dep array is a stale value or a redundant refetch rather than a crash, so those want a manual pass. @itsjavi — also note your `CHANGES_REQUESTED` from Aug 11 asked for conflict resolution and de-drafting; the branch is now rebased, out of draft, and Sonar's quality gate passes (100% coverage on new code), so that review looks addressed.

_Reviewed with Claude Opus 5 (high effort)._


---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx:54]: Fixed in 67213d9484. You're right that the dep is correct and the effect was the over-keyed part — `filters` was dep #1 of the fetch effect, which unconditionally calls `setPage(1)`, so a `router.refresh()` from the status dropdown or share modal reset a scrolled list. The effect is now keyed on `filtersKey` (the serialized filter value), so it re-runs on a real filter change and ignores a new object of equal value; the memo keeps `survey`. `CustomFilter` has the same memo change but its `filters` only feeds the download handler, so it needed nothing. Noted under Open gaps that this was traced from the call chain rather than observed in a browser.

Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx:54]: Fixed in 67213d9484. You're right that the dep is correct and the effect was the over-keyed part — `filters` was dep #1 of the fetch effect, which unconditionally calls `setPage(1)`, so a `router.refresh()` from the status dropdown or share modal reset a scrolled list. The effect is now keyed on `filtersKey` (the serialized filter value), so it re-runs on a real filter change and ignores a new object of equal value; the memo keeps `survey`. `CustomFilter` has the same memo change but its `filters` only feeds the download handler, so it needed nothing. Noted under Open gaps that this was traced from the call chain rather than observed in a browser.

Dhruwang:
All three addressed. Conflicts resolved by rebasing on `main` — that also invalidated half the original PR: the `packages/types` and `packages/cache` `eslint-disable` directives are load-bearing again (removing them is now 4 + 6 real errors), and `main` had already removed the `i18n-utils`/`vite-plugins` ones. So what's left is the `exhaustive-deps` work. Sonar was failing on one condition only — 0% coverage on new code — where the sole coverage-measured file in the diff was `apps/web/lib/useDocumentVisibility.ts`; it now has a unit test and the gate is green at 100%. Out of draft. Review since then caught one genuine regression in `ResponsePage` (a `router.refresh()` resetting the responses list to page 1), fixed in 67213d9484. Manual QA gaps are listed explicitly under Open gaps — integrations modals and the EE contacts column settings still need a human with credentials.

apps/web/modules/survey/editor/components/survey-editor.tsx
```diff
@@ -132,7 +132,7 @@ export const SurveyEditor = ({
         setActiveElementId(firstBlock.elements?.[0]?.id);
       }
     }
-  }, [survey]);
+  }, [survey, localSurvey]);
```
itsjavi: localSurvey was added to an effect that sets localSurvey.

It terminates because of the early return, but localSurvey gets a new identity on every editor edit, so this effect now re-runs on essentially every keystroke to do nothing. 

It's also the exact shape that becomes a real loop the moment someone touches that guard. setLocalSurvey(prev => prev ?? structuredClone(survey)) (or an init ref) satisfies the lint rule while keeping [survey] as the trigger.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
```diff
@@ -51,9 +51,15 @@ export const ResponsePage = ({
...
   );
 
+  // The fetch effect below is keyed on the filter *value*, not on `filters`' identity: `survey` is an
+  // RSC prop, so every `router.refresh()` (survey status dropdown, share modal, reset survey) hands
+  // down a freshly deserialized object. Re-running on identity alone would call setPage(1) and
+  // collapse the infinite-scroll list back to the first page on an unrelated refresh.
+  const filtersKey = JSON.stringify(filters);
```
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx]: const filtersKey = JSON.stringify(filters) is unmemoized, so it re-serializes on every render of the responses page (each scroll fetch, each row/tag edit). 

`useMemo(() => JSON.stringify(filters), [filters])` costs nothing.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx
```diff
@@ -158,7 +158,7 @@ export const CustomFilter = ({ survey }: CustomFilterProps) => {
   const filters = useMemo(
```
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx:158]: adding survey to the filters memo makes filters identity-unstable on every router.refresh(). 

The effect is protected by filtersKey, but refetchResponses (line 116) and fetchNextPage (L 84) still key on filters, so the registerAnalysisRefreshHandler effect now unregisters/re-registers on every refresh. 

Harmless (it's a ref write, no setState), but it's new churn the PR adds.

apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
```diff
@@ -110,7 +110,7 @@ export const AddIntegrationModal = ({
         type: dbProperties[fieldKey].type,
       })) || []
     );
-  }, [selectedDatabase?.id]);
```
Comment (itsjavi) [inline apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx:113]: [selectedDatabase?.id] → [selectedDatabase] de-memoizes dbItems: the effect at L106 always calls setSelectedDatabase({...}) with a fresh literal when the surveys/databases server props change, so identical content now recomputes where ?.id didn't.

itsjavi (changes requested):

apps/web/modules/survey/editor/components/survey-editor.tsx
Dhruwang: Taken in 19ccb147e — your suggested shape, `setLocalSurvey((current) => current ?? structuredClone(survey))` with `[survey]` as the only dep.

You're right on both counts, and there is a third thing I found while making the change that makes it clearer cut than the thread suggests: **the effect body was already unreachable.** `localSurvey` is seeded by its own initializer, `useState<TSurvey | null>(() => structuredClone(survey))` (line 91, and the same on `main`), so `if (localSurvey) return;` is true on the very first render. Adding `localSurvey` to the deps therefore bought a re-run on every keystroke for a body that could never execute.

That also settles what to do with the `setActiveElementId` half, which is why I did not simply delete the effect: the recovery path is unreachable *today*, but the `if (!localSurvey) return <LoadingSkeleton />` guard below says the component is meant to survive a null. The updater keeps that recovery while keying on `survey` alone. The active element is not set here any more — the `[localSurvey?.type]` effect at :156 already picks the first element whenever `localSurvey` appears, including a null → set transition, so that path is covered rather than dropped.

The comment now says all of this, so the next person does not re-add `localSurvey` to satisfy the linter.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/components/ResponsePage.tsx]: Fixed in 19ccb147e, and it turned out to be load-bearing rather than a free win — the memo is what your other thread needs too.

`filtersKey` is now `useMemo(() => JSON.stringify(computedFilters), [computedFilters])`. Since `computedFilters` is itself a memo on `[survey, selectedFilter, dateRange]`, the serialization now runs when the filters actually recompute — a filter change, or an RSC refresh handing down a new `survey` — instead of on every render of the page. Scroll fetches and row/tag edits re-render through `setResponses`/`setPage` without touching any of those three, so they no longer re-serialize.

The reason it is more than a micro-optimisation: it gives a stable key to hang the identity fix on. See the reply on the `CustomFilter` thread — `filters` is now `useMemo(() => computedFilters, [filtersKey])`, so downstream callbacks key on the filter *value* rather than on an identity that changes with every `router.refresh()`.

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/CustomFilter.tsx:158]: Real, and fixed in 19ccb147e rather than accepted as churn — the thread text is about `ResponsePage`, so that is where the fix landed.

You are right that the `filtersKey` guard only protected the fetch effect. `fetchNextPage` (`[filters, …]`) and `refetchResponses` (`[filters, …]`) still keyed on the identity, so every `router.refresh()` gave them new identities, which re-ran the `registerAnalysisRefreshHandler` effect. Harmless as you say — a ref write, no setState — but it was churn this PR introduced by adding `survey` to the memo, and "harmless today" is only true while that handler stays a ref write.

The fix stabilises the identity at the source instead of patching each consumer:

```ts
const computedFilters = useMemo(() => getFormattedFilters(survey, selectedFilter, dateRange), [survey, selectedFilter, dateRange]);
const filtersKey = useMemo(() => JSON.stringify(computedFilters), [computedFilters]);
// eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the filter value, not its identity
const filters = useMemo(() => computedFilters, [filtersKey]);
```

`filters` now changes identity only when the filter value changes, so `fetchNextPage`, `refetchResponses` and the register/unregister effect all stop reacting to refreshes. The dep on the memo that computes the filters stays exhaustive — `survey` is still read, so the correctness fix this PR is here for is intact; only the *adoption* of the new object is gated on the value. The one suppression is on the identity-holding memo, where the mismatch is the point, with the reason inline.

The fetch effect's own suppression comment lost the clause about a new `filters` identity from an RSC refresh, since that can no longer happen.

`CustomFilter`'s own `filters` memo I left as it is: it feeds only the CSV/Excel download handler at :244, no effect and no callback deps, so the extra identity change costs a recompute of a plain object on refresh and nothing else.

apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx:113]: Correct, and reverted in 19ccb147e — both memos are back on `?.id`, with the reason written down instead of the dep array quietly asserting something false.

I checked the mechanism you describe: the effect at :160 calls `setSelectedDatabase({ id, name, properties })` — a fresh object literal every time — and its deps are `[selectedIntegration, surveys, databases]`, all server props whose identity changes on any refresh. So keying `dbItems` on `selectedDatabase` recomputed it for identical content, exactly as you say.

`elementItems` had the same shape one memo below: `selectedSurvey` is re-seeded by the same effect from the `surveys` prop, and `main` keyed it on `selectedSurvey?.id` for the same reason. Fixed both rather than only the one flagged.

Both now carry an `eslint-disable-next-line react-hooks/exhaustive-deps` with the rationale, which is the pattern this PR uses for the other 17 intentional cases — the honest outcome here is a documented suppression, not a dep array that trades correct memoization for a green linter.

Dhruwang:
@itsjavi — your four inline comments are all addressed in 19ccb147e, each with a reply in its own thread. Every one was right, and three of them were the same class of mistake: a dep array that satisfies the linter by depending on an RSC prop's *identity*, which changes on every `router.refresh()` even when nothing changed.

- **`ResponsePage`** — `filters` now holds a stable identity keyed on the serialized filter value, so a refresh stops re-creating `fetchNextPage`/`refetchResponses` and re-registering the analysis refresh handler. `filtersKey` is memoized, so it no longer re-serializes on every scroll fetch and row edit.
- **Notion `AddIntegrationModal`** — `dbItems` and `elementItems` are back on `?.id`, as on `main`, with documented suppressions. I fixed `elementItems` too; it had the same shape one memo below the one you flagged.
- **`survey-editor`** — the init effect no longer depends on the state it sets. While making the change I found its body was already unreachable (`localSurvey` is seeded in the `useState` initializer), so the dep bought a re-run per keystroke for code that could never execute.
- **`CustomFilter`** — left as it is, deliberately: its `filters` feeds only the download handler, no effect and no callback deps. The churn you described was in `ResponsePage`, which is where the fix went.

Also fixed the nit from the automated pass: the suppression comment in `contact-data-view.tsx` claimed `fetchContactsFromStart` is recreated each render, when it is a `useCallback`. Wrong reason, right suppression — the reason now says what it actually is.

On your Aug 11 review: the branch is rebased on `main`, out of draft, and Sonar's gate passes (100% coverage on new code — `useDocumentVisibility.ts` is the only file in the diff it measures). The PR description is updated for all of the above, including the new coverage rows and what is still uncovered at runtime.

The `Open gaps` list is unchanged in substance and still wants a human with the app running: the responses infinite scroll, and the Notion/Airtable field-mapping flows. Worth noting the Notion dep changes are now reverted to `main`'s behaviour, so what is uncovered there is `main` plus two comments rather than a change of mine.

apps/web/modules/ee/contacts/components/contact-data-view.tsx
```diff
@@ -105,6 +105,7 @@ export const ContactDataView = ({
     if (isResettingSearch.current) {
       isResettingSearch.current = false;
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- debounced search must fire only on searchValue change; `fetchContactsFromStart` is keyed on `searchValue` itself, so listing it would re-trigger the search whenever `workspaceId` or `itemsPerPage` changed too
```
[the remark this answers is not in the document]
Dhruwang: Real, and fixed in 8f1013c1f0. I traced it rather than taking the description on trust, and it plays out exactly as written:

1. Switch workspace → the sync effect sets `isResettingSearch.current = true` and calls `setSearchValue("")`.
2. `searchValue` was already `""` — the common path, since you have to have searched for it not to be — so React bails on the setter and the `[searchValue]` effect never runs.
3. The flag is still armed. The user types `a` → `searchValue` changes → the effect runs → takes the `isResettingSearch` branch, lowers the flag, fetches nothing.
4. Their first search after switching workspace silently returns the unfiltered list.

Fixed as you suggest: arm the flag only when there is a non-empty search that `setSearchValue("")` will actually clear, and list `searchValue` in that effect's deps now that the body reads it. The extra runs are free — the whole body is behind `prevWorkspaceId.current !== workspaceId`, which it updates immediately, so a keystroke costs one comparison.

I did **not** add `fetchContactsFromStart` to the debounced effect, per your note and for the reason its suppression already gives: that callback is keyed on `searchValue` itself, so listing it would re-trigger the search whenever `workspaceId` or `itemsPerPage` changed too.

Two things worth stating plainly:

**This is pre-existing on `main`**, not something this PR introduced — `git show origin/main:…/contact-data-view.tsx` has the same unconditional `isResettingSearch.current = true`. I fixed it here rather than filing it because this PR owns the suppression that documents this effect pair, and leaving a known first-search-swallowing bug next to a comment explaining why the deps are deliberate seemed worse than the small scope increase.

**It is verified by reading, not by running.** The EE contacts view is in this PR's `Open gaps` for exactly that reason — no seeded contacts across two workspaces here — and per AGENTS.md the fix does not get a `.tsx` unit test. So it wants the same manual pass as the rest of that list: search in workspace A, switch to workspace B, type one character, confirm the list filters.

apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx
```diff
@@ -110,6 +110,10 @@ export const AddIntegrationModal = ({
         type: dbProperties[fieldKey].type,
       })) || []
     );
+    // The effect below re-seeds `selectedDatabase` with a fresh object literal whenever the
+    // `databases`/`surveys` server props change identity, which an RSC refresh does on unchanged
+    // content. Keying on the id keeps identical content from recomputing this list.
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- keyed on the database's identity, not the object's
```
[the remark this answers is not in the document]
Comment (Dhruwang) [reply apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/notion/components/AddIntegrationModal.tsx:123]: The staleness is real, but I am keeping `?.id` — and I want to be explicit that this is a judgement between two review comments that point in opposite directions, not an oversight.

@itsjavi asked for exactly the reverse on this line ([thread](https://github.com/formbricks/formbricks/pull/8704#discussion_r3864102552)): keying on `selectedDatabase` de-memoizes `dbItems`, because the effect at :160 re-seeds it with a fresh object literal whenever the `databases`/`surveys` server props change identity, which a refresh does on unchanged content. `main` keys on `?.id` for that reason. I reverted to it one commit ago on his review.

Weighing the two failure modes:

- **Keyed on the object** (your version): correct on a property change, recomputes both lists on every refresh that changed nothing. Guaranteed cost, on every refresh.
- **Keyed on the id** (current): free on a refresh, stale if a refresh brings back *different* properties for the same database id — someone editing the Notion database's schema while the modal is open on that database. Rare cost, on a narrow window.

Neither is free, and switching back would just re-open a thread a human maintainer already closed.

What I have done instead is stop the trade-off being invisible: the comment above the memo now names the staleness bound and says what the actual fix is — not either dep array, but the seeding effect, which should not replace `selectedDatabase` with a fresh literal when the content is unchanged. Do that and the memo can key on the object with exhaustive deps and no suppression, and both of your concerns go away together.

I did not make that change here because this modal is in the PR's `Open gaps` — there are no Notion credentials in this environment to exercise the field-mapping flow — and rewriting its seeding effect unverified, in a PR whose subject is dependency arrays, is a worse trade than a documented `?.id`. Happy to file it.

Same reasoning applies to `elementItems` directly below, which I reverted in the same commit and which has the identical shape.

itsjavi (approved):

PR #8705, opened 2026-07-31
fix(charts): wrap long x-axis labels on cartesian charts (ENG-2061)

Dhruwang:
## What & why

On dashboard cartesian charts (area / line / bar), long survey questions used as x-axis categories were unreadable:

- Recharts renders x-axis ticks as single-line SVG `<text>`, so a full question (e.g. *"How satisfied are you with the clarity of information provided before the match…"*) overflowed into the neighbouring data point.
- Recharts' default tick interval hides overlapping labels, so on area/line charts most data points ended up with **no** label at all (only the far edges showed, bleeding off the plot).

This adds a custom x-axis tick rendered inside a `foreignObject`, so each label is real HTML: it wraps, stays centred under its data point, uses `text-wrap: pretty`, and clamps to 3 lines with the full text available on hover (`title`). `interval={0}` forces every category to keep its label. The custom tick is scoped to category axes (`hasCategoryAxis`); measure-only charts, which hide the axis, are untouched. All cartesian chart types share `CartesianChart`, so this fixes area, line, and bar in one place.

Before: 

Screenshot 2026-08-06 at 4 00 19 PM

After:

Screenshot 2026-08-06 at 3 47 05 PM

**Follow-up fixes (from QA on many-category charts):**

- **Label overlap** — a fixed 140px label width overlapped neighbours once a chart had many categories. Each label is now sized to its category band (`width / visibleTicksCount`, both injected by Recharts' `CartesianAxis`), clamped between `X_AXIS_TICK_MIN_WIDTH` (32px) and `X_AXIS_TICK_MAX_WIDTH` (140px) with an 8px gap, so labels shrink to fit instead of colliding.
- **Clipped 3rd line** — reserved x-axis height raised 56px → 72px so the 3-line clamp isn't cut at the plot's bottom edge (tick margin + offset + 3×~15px lines).
- **Tooltip horizontal scroll** — `PolishedChartTooltip` had `min-w-[180px]` but no max-width, so a long question header grew unbounded and forced a horizontal scrollbar. It's now capped at `max-w-xs` with `break-words` on the header and row labels (and `shrink-0` on the value), so long text wraps inside the tooltip.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2061

## How this was tested

- `eslint` on the changed files → clean.
- Manual visual QA on a bar chart with 8 categories including very long question text:
  - Every category keeps its label; labels wrap ≤3 lines, centred under their bar, no overlap.
  - Long labels clamp to 3 lines; hover shows the full text via the `title` tooltip.
  - Hovering a bar → tooltip stays within `max-w-xs`, header/labels wrap, no horizontal scrollbar.
- Charts covered: Area, Line, Bar (shared `CartesianChart`).

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Dashboard chart with **8+ categories** and **long category text**, as **Area**, **Line**, then **Bar** → every category shows a label under its point, wrapped ≤3 lines, centred, no overlap, no dropped labels.
- [ ] Hover a truncated (3-line-clamped) label → the `title` tooltip shows the full text.
- [ ] Hover a bar/point → the chart tooltip stays contained (no bottom horizontal scrollbar), long header/labels wrap.
- [ ] Few categories (2–4) with long text → labels use the full max width (140px), still centred and readable.
- [ ] Short category labels (enum options, dates) → render centred, no excess whitespace.
- [ ] Measure-only chart (e.g. emotion counts, no real category) → x-axis label still hidden as before.

**Preconditions / test data**

- A workspace/dashboard with a survey containing several long-text questions and a CSAT/rating measure (8+ categories to exercise the band-width sizing).

**Risks & regressions**

- `interval={0}` forces every category tick to render; very dense axes shrink labels toward the 32px floor (readability degrades gracefully rather than overlapping).
- Reserved axis height increased to 72px, so the plot area is slightly shorter — confirm bars/lines still look right on short charts.

- none


xernobyl:
Tested this locally against a dashboard with a `collectedAt` day-granularity chart (45 daily buckets), and I think there's one regression worth sorting before this lands. Everything else looks good — the wrapped labels are a clear improvement on the four-long-questions case this was aimed at :slightly_smiling_face:

**[P2] Dense date axes come out worse than before.** With `interval={0}` every bucket now renders a label. On a 45-day chart at ~1200px the band is ~27px, and because `X_AXIS_TICK_MIN_WIDTH` (32px) floors each label *above* its band width, they overlap. `Jun 23, 2026` wraps to three lines in a 32px box, so the whole axis reads as one solid block of text. The same chart on `main` renders ~15 evenly spaced, readable dates, because recharts was thinning them.

To be clear, `interval={0}` is doing the right thing for the case it was added for — the issue is only that it's unconditional.

**Suggested fix** — let the tick thin itself. Recharts already passes `index` and `visibleTicksCount` to a custom tick, so this can live inside `WrappingXAxisTick` and doesn't need the plot width hoisted up to `<XAxis>`:

```diff
 const X_AXIS_TICK_GAP = 8;
+/** Narrowest a label may get before we drop labels instead of shrinking them. Below this a
+ * wrapped label is unreadable, so thinning beats cramming. */
+const X_AXIS_TICK_MIN_READABLE_WIDTH = 56;

-function WrappingXAxisTick({ x, y, payload, formatter, width, visibleTicksCount }: Readonly<{
+function WrappingXAxisTick({ x, y, payload, formatter, width, visibleTicksCount, index }: Readonly<{
   …
+  index?: number;
 }>) {
   const label = formatter(payload?.value);
   const band = width && visibleTicksCount ? width / visibleTicksCount : X_AXIS_TICK_MAX_WIDTH;
+  // Once bands are narrower than a readable label, keep only every Nth label and give the
+  // survivors the full width of the bands they span.
+  const step = Math.max(1, Math.ceil(X_AXIS_TICK_MIN_READABLE_WIDTH / band));
+  if (step > 1 && (index ?? 0) % step !== 0) return null;
+  const slot = band * step;
-  const tickWidth = Math.min(X_AXIS_TICK_MAX_WIDTH, Math.max(X_AXIS_TICK_MIN_WIDTH, band - X_AXIS_TICK_GAP));
+  const tickWidth = Math.min(X_AXIS_TICK_MAX_WIDTH, Math.max(X_AXIS_TICK_MIN_WIDTH, slot - X_AXIS_TICK_GAP));
```

I pulled the component into an isolated harness to compare the shapes we care about. The few-category cases come out identical — only the dense ones change:

| scenario | labels before → after | label width | overlaps before → after |
|---|---|---|---|
| 4 long questions (bar / area) | 4 → 4 | 140px | 0 → 0 |
| 8 long questions | 8 → 8 | 135px | 0 → 0 |
| 2 long questions | 2 → 2 | 140px | 0 → 0 |
| 6 short enum labels | 6 → 6 | 140px | 0 → 0 |
| measure-only (axis hidden) | 0 → 0 | — | 0 → 0 |
| **60 daily buckets** | **60 → 20** | 49px | **59 → 0** |
| **30 daily buckets** | **30 → 15** | 68px | 0 → 0 |
| **500 categories** | **500 → 20** | 49px | — |

It also helps the DOM cost: 500 categories goes from 1,501 x-axis nodes down to 541. Recharts still emits an empty `<g>` per tick, but the `foreignObject` and its layout cost go away.

Two judgement calls rather than bugs, so entirely up to you:

• `index % step === 0` anchors to the first tick, so the last bucket or two can end up unlabelled (my 60-day case stops at Jun 27 with data through Jun 29). `main`'s `preserveEnd` kept the end instead — anchoring to the last tick is a one-liner if you'd rather have that.
• Separately, `formatXAxisTick` could drop the year when the whole series sits in one year — `Jun 23` rather than `Jun 23, 2026` fits on a single line at these widths and would help every date chart, not just the dense ones. Complementary to the above, not a substitute.

xernobyl (commented):
Follow-up to the dense-axis comment above — three more things I ran into testing this locally. The area/line one is the one I'd want fixed; the other two are minor and won't block anything.

The bar chart case this was aimed at looks good, to be clear — all four labels present, wrapped and centred under their bars.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
```diff
@@ -37,6 +37,61 @@ export interface CartesianChartProps {
...
+  formatter: (value: unknown) => string;
+  width?: number;
+  visibleTicksCount?: number;
+}>) {
+  const label = formatter(payload?.value);
+  const band = width && visibleTicksCount ? width / visibleTicksCount : X_AXIS_TICK_MAX_WIDTH;
+  const tickWidth = Math.min(X_AXIS_TICK_MAX_WIDTH, Math.max(X_AXIS_TICK_MIN_WIDTH, band - X_AXIS_TICK_GAP));
+  return (
+    <foreignObject
+      x={(x ?? 0) - tickWidth / 2}
```
xernobyl: **[P2] On area and line charts the first and last labels get clipped by the plot edge.**

On a point scale (area / line) the first and last categories sit *on* the plot boundary, and since the tick is centred (`x - tickWidth / 2`) up to half the label falls outside the SVG viewport and gets clipped.

Measured on the four-question area chart at ~1200px: the last label overhangs the plot by 65px of its 140px, so roughly 46% of it is cut off, and the first overhangs by 5px and loses its leading character. It renders as `dow easy was the arrival…` on the left and `How satisfie / with the clea / availability o` on the right.

Worth fixing because this is the area chart from screenshot A on the ticket — the case ENG-2061 was opened with — so as it stands the reported bug is only half closed. Clamping the `foreignObject` x into the axis extent, or left/right-anchoring the edge ticks rather than centring them, should cover it.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
```diff
@@ -37,6 +37,61 @@ export interface CartesianChartProps {
   yAxisScale?: YAxisScale;
 }
 
+/** Upper bound (px) on a single x-axis label before wrapping. Keeps long question labels from
+ * bleeding into neighbouring data points on charts with few categories. */
+const X_AXIS_TICK_MAX_WIDTH = 140;
```
xernobyl: **[P3] The 140px cap truncates more than it needs to, on the exact case this was written for.**

With four categories at ~1200px the band is ~300px, but labels cap at 140px and clamp to three lines, so about a quarter of each question ends up hidden behind the hover title (`scrollHeight` 60 vs `clientHeight` 45) with ~160px of empty space either side of it.

The band-width clamp from the follow-up commit already stops neighbours colliding, so the max width is now mostly just truncating. Raising it, or deriving it from the band with a looser ceiling, would show noticeably more of each question. Not a blocker — just leaving a bit of value on the table.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
```diff
@@ -37,6 +37,61 @@ export interface CartesianChartProps {
...
+/** Lower bound (px) so a label keeps a little room to wrap even on very dense axes. Below the band
+ * width the label narrows to fit rather than overlapping its neighbours. */
+const X_AXIS_TICK_MIN_WIDTH = 32;
+/** Horizontal gap (px) reserved between adjacent labels so wrapped text never touches. */
+const X_AXIS_TICK_GAP = 8;
+/** Vertical space reserved for the x-axis labels: enough for 3 wrapped lines of `text-xs`
+ * (~15px each) plus the tick margin and top offset below the axis, so the third line is not
+ * clipped by the plot's bottom edge. */
+const X_AXIS_HEIGHT = 72;
```
xernobyl: **[P3] `X_AXIS_HEIGHT` is doing double duty.**

It's both the reserved `<XAxis height>` and the `foreignObject` height. The box ends up ~27px taller than the clamped text, so on bar charts (which have no legend) it overhangs the plot by 15px. Harmless today — the SVG clips it and I checked it isn't hit-testable — but it is why the reserve had to go to 72px. Two separate constants would probably read clearer.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
```diff
@@ -33,6 +33,45 @@ export interface CartesianChartProps {
...
+      x={(x ?? 0) - X_AXIS_TICK_WIDTH / 2}
+      y={(y ?? 0) + 4}
+      width={X_AXIS_TICK_WIDTH}
+      height={X_AXIS_HEIGHT}
+      // Let the tooltip cursor win the pointer; the label is decorative.
+      style={{ overflow: "visible", pointerEvents: "none" }}>
+      <div
+        title={label}
+        className="text-muted-foreground line-clamp-3 text-center text-xs leading-tight"
+        style={{ textWrap: "pretty" }}>
```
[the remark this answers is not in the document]
Dhruwang: Already resolved at head — `pointerEvents: "none"` was dropped in an earlier commit, so `style={{ overflow: "visible" }}` keeps the `foreignObject` hit-testable and the `title` hover works. The tick sits in the axis band below the plot, so it doesn't intercept hover over the bars/points.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
Dhruwang: Good catch, and agreed this left the ticket half-closed. Fixed in `2dd60e3f8`: line/area charts now pass a `pointScale` flag, and on those the first/last labels anchor inward (left-align the first, right-align the last) instead of centring, so the full label stays inside the plot. Bar charts (band scale, edges already inset) keep centred labels. Visual QA against the four-question area chart from screenshot A is still pending on my side — worth a re-check when you get a moment.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
Dhruwang: Raised to 220 in `2dd60e3f8`. The per-category band clamp already prevents neighbours colliding, so the cap now only bites on charts with lots of room, where it shows noticeably more of each question before truncating.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
Dhruwang: Split in `2dd60e3f8`: `X_AXIS_RESERVED_HEIGHT` for the `<XAxis height>` and `X_AXIS_LABEL_BOX_HEIGHT` (just the clamped text) for the `foreignObject`, so the box no longer overhangs the plot on legend-less bar charts.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
```diff
@@ -35,6 +35,98 @@ export interface CartesianChartProps {
...
+  const band = width && visibleTicksCount ? width / visibleTicksCount : X_AXIS_TICK_MAX_WIDTH;
+  const tickWidth = Math.min(X_AXIS_TICK_MAX_WIDTH, Math.max(X_AXIS_TICK_MIN_WIDTH, band - X_AXIS_TICK_GAP));
+  const tickX = x ?? 0;
+
+  const isFirst = pointScale && index === 0;
+  const isLast = pointScale && index === (visibleTicksCount ?? 0) - 1;
+  let boxX = tickX - tickWidth / 2;
+  let textAlign = "text-center";
+  if (isFirst) {
+    boxX = tickX; // left edge at the point; label extends inward (right)
```
xernobyl, line 116: **🟠 should · frontend · confidence: high — measured by rendering this component at 2dd60e3f; the arithmetic and the render agree**
Edge-anchored labels now collide with their neighbour on point-scale charts with 5+ categories

The anchoring fixes the clipping, but it shifts the edge label a full `tickWidth / 2` inward while its neighbour is still centred, so past a certain density the two boxes overlap.

No overlap requires `1.5 x tickWidth <= spacing`, where `spacing = width / (visibleTicksCount - 1)`. Since `tickWidth` is `band - 8` and `band = width / visibleTicksCount`, that inequality fails from about 5 categories up. Rendered at a 1140px plot area:

| categories | tickWidth | spacing | overlaps | worst |
|---|---|---|---|---|
| 4 | 220px | 380px | 0 | — |
| 5 | 220px | 285px | 2 | **45px** |
| 8 | 134.5px | 163px | 2 | **38.9px** |

It's the first/second and last/second-last pairs, both ends. Visually the text runs together — `…visa, procedures)?` overprinting `of information provided…`. Four categories is clean, which is why the ticket's own screenshots look right and this didn't show up.

Bar charts are unaffected (`pointScale` is false, centred labels, 0 overlaps at every count I measured), so this is confined to line and area.

The cap that makes it exact: an anchored edge box may only reach as far as its neighbour's left edge, so bound it by `spacing - tickWidth / 2` rather than letting it take the full `tickWidth`:

```ts
const spacing =
  width && visibleTicksCount && visibleTicksCount > 1
    ? width / (visibleTicksCount - 1)
    : Infinity;
const edgeWidth = Math.max(
  X_AXIS_TICK_MIN_WIDTH,
  Math.min(tickWidth, spacing - tickWidth / 2)
);
```

Then use `edgeWidth` for the box width (and for the `isLast` offset, `boxX = tickX - edgeWidth`) on the two edge ticks only. I patched exactly that into a local copy of the component and re-measured: 5 categories and 8 categories both go to 0 overlaps, edges narrow to 175px and 95.6px, everything stays inside the plot, and the band-scale charts are byte-identical.

What would make me wrong: if line/area dashboards in practice only ever carry three or four categories, this is theoretical. The 45px overlap at five is real either way.

xernobyl (commented):
**1 should** — reviewed at `2dd60e3f85`

Re-review after `2dd60e3f`. Three of my four previous findings are fixed at this head, verified by re-rendering the component from `2dd60e3f` against the same category shapes:

• **Edge clipping on line/area — fixed.** The four-question area chart now sits fully inside the plot (previously the last label overhung by 65px of its 140px). The `pointScale` split is the right call: band-scale charts inset their edge categories already, so leaving bars centred is correct.
• **140px cap — fixed.** 220px now, and the four-question chart uses the full width rather than clamping at 140.
• **`X_AXIS_HEIGHT` double duty — fixed.** Splitting the label box (48px) from the reserved axis height (72px) also removes the 15px overhang the old single constant caused on legend-less bar charts.

One new problem the edge fix introduces, inline below.

**Still open — the dense date axis.** Unchanged at this head: `interval={0}` still forces every tick, so 60 daily buckets render 60 labels at the 32px floor with 59 overlaps (worst 28.7px), against ~15 readable ones on `main`. Details and a verified patch are in https://github.com/formbricks/formbricks/pull/8705#issuecomment-5204343916 — not re-posting it here, just flagging it's still live. Happy to be told it's out of scope and should be its own ticket.

**Checked** — frontend/rendering (re-measured every category shape from 2 to 500), correctness, drift. **Skipped** — tenancy, security, conventions, tests: the diff is two presentational components with no data access, and `AGENTS.md` rules out `.tsx` unit tests, so there is no test gap to report here.
**Not verifiable here** — how the 220px labels look on a narrow/mobile viewport; I only measured at 1280px wide.

apps/web/modules/ee/analysis/charts/components/cartesian-chart.tsx
Dhruwang: Fixed — you nailed the arithmetic. Capped the edge box at `spacing - tickWidth / 2` (floored at `X_AXIS_TICK_MIN_WIDTH`) exactly as you suggested, and used that `edgeWidth` for both the box width and the `isLast` offset (`boxX = tickX - edgeWidth`); centred ticks are unchanged. Band-scale (bar) charts stay byte-identical since `pointScale` is false there.

Pushed in 8a2858cf36 (also merged latest `main` to unblock the format:check CI step).

xernobyl (approved):
Approving. The edge-label cap in `8a2858cf` is exactly right — I re-measured at `f60cc152` and point-scale charts go to 0 overlaps at 5 and 8 categories with everything inside the plot, and the band-scale (bar) charts are unchanged. Together with the earlier commits that closes the clipping, the 140px cap and the `X_AXIS_HEIGHT` double duty, and the ticket's original case (4 long CSAT questions on bar and area) renders properly now.

On my remaining open comment about the dense date axis — **treating that as out of scope for ENG-2061 and not blocking on it.** It's collateral from `interval={0}` rather than anything to do with long question labels, and this PR has been round the houses enough. Deliberately not filing a ticket either; we'd rather see whether it actually bites in practice before spending anything on it.

One ask so that decision is real rather than the concern just evaporating: could you add a line to the QA / Test Plan for it? Something like

- [ ] Open a chart with a `Collected at` time dimension at day granularity over a long range (30+ buckets) → check the x-axis labels are still readable.

Since the QA plan is what generates release QA, that's the only way it gets looked at. Worth flagging for whoever picks it up that it's a change against `main` — `main` thins those ticks to ~15 readable labels, this renders all of them — so it should be judged on whether it reads acceptably, not on whether it looks broken in isolation.

E2E and the Docker build are still running as I write this; the required gate covers those, so not holding the approval for them.

PR #8707, opened 2026-07-31
fix(taxonomy): map Hub 404 to 404 instead of a blanket 502 (ENG-1886)

pandeymangg:
## What & why

Every taxonomy v3 route collapsed **all** Hub failures into `problemBadGateway`, regardless of what the Hub actually said. So a benign **404** — a stale run id, a node someone else already removed — surfaced as **"502 Bad Gateway"**: reads like a server fault, and counts a perfectly normal outcome towards the 5xx rate.

The 502 also echoed `result.error.message`. The Hub SDK has no `message` field to read off an RFC 9457 problem body, so it stringifies the whole body into `message` — meaning internal Hub problem URLs and codes were landing in a customer-facing response:

```
502 Bad Gateway
detail: 404 {"type":"https://hub.formbricks.com/problems/not-found","title":"Not Found"}
```

`HubError.status` was already available; the taxonomy ops just weren't reading it. A new `hubFailureResponse` helper now routes on it:

| Hub returned | Before | After |
| --- | --- | --- |
| `404` | 502 | **404** (`Taxonomy run not found` / `Taxonomy node not found`) |
| NO_CONFIG | 502 | **503** (`problemServiceUnavailable`) |
| 5xx / timeout / connection / null payload | 502 | 502, with a **static** detail |

Notes on the edges:

- **No IDOR oracle.** Directory access is verified before the Hub call and every call is scoped by `tenant_id`, so a 404 only ever means "not in *your* directory".
- **Creates keep 502.** `POST /runs` gets no 404 mapping — "not found" says nothing useful about a create.
- **`isHubNotConfigured` matches the sentinel, not `status === 0`.** The SDK also reports connection failures and timeouts without a status, so matching on 0 alone would turn a dead socket into a 503. Those stay 502.
- **502 details are static.** The full Hub error is already logged server-side in `@/modules/hub/service`; correlate on `requestId`.

Two things found while in here:

- **`unavailableMessage` removed** from the `fields` / `state` responses. It carried the same raw Hub JSON, this time inside a **200** body. It is rendered nowhere — the UI switched to a localized alert in ENG-1917 and this field was left behind — and being outside `t()` it would ship untranslated English if anyone ever did render it. The `unavailable` boolean it sits next to is still live (drives the alert, the Generate gating, and `gate.ts`).
- **`getTaxonomyRun` logs a 404 at `debug`**, not `warn`. A run is polled every 5s while it generates, so a deleted run wrote a warning with a full stack trace on every tick. Matches what `getActiveTaxonomyTree` already does. The other four ops are user-click driven, so they keep `warn`.

Depends on nothing; no Hub change. Follow-up: the Hub currently returns 200-empty (not 404) on `nodes/{id}/records` for a tenant mismatch, so that one 404 mapping is unreachable until [ENG-1887](https://linear.app/formbricks/issue/ENG-1887) lands — after which this side needs no further change.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1886/triage-map-hub-404-404-instead-of-blanket-502-in-taxonomy-v3-routes

## How this was tested

- `pnpm vitest run app/api/v3/unify-feedback/taxonomy/ modules/hub/ modules/ee/unify-feedback/` → **347 passed (28 files)** ✅
- `pnpm eslint` on all six changed files → clean ✅
- `tsc --noEmit` → no new errors from this diff. One pre-existing error remains at `modules/ee/unify-feedback/topics-subtopics/lib/api-client.test.ts:196` (`triggerTaxonomyRun` call missing `scopeType`); verified it reproduces on a clean `main` via `git stash`, so it is not from this change.
- Tests added/updated in `operations.test.ts`: per call site, Hub `{status: 404}` → 404 with the right `resource_type`/`resource_id`; Hub 5xx → 502 with the static detail; NO_CONFIG → 503; connection failure (status 0, no sentinel) → 502; `{data: null, error: null}` → 502; create keeps 502 on a Hub 404. Every case also asserts the response body contains no `hub.formbricks.com/problems` marker.
- Tests added in `modules/hub/utils.test.ts` for `isHubNotConfigured`, including the connection-failure case that must **not** be treated as "not configured".

No UI change — the UI already rendered localized strings for all of these states and never showed the raw detail, so there is nothing visual to capture.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open **Workspace → Unify → Topics & Subtopics**, pick a feedback directory with an existing taxonomy. The tree, per-node record counts and the record drilldown all load as before — this change is invisible on the happy path.
- [ ] Generate a taxonomy and watch it run to completion. The spinner, the "Generating…" state and the hand-off to the finished tree all behave as before.
- [ ] Rename a topic and remove a topic. Both still succeed, and the success/error toasts are unchanged.
- [ ] **Stale run (the main case).** With a run polling, delete that run from the Hub (or point at a run id that no longer exists) and re-open the page. In DevTools → Network, `GET /api/v3/unify-feedback/taxonomy/runs/{runId}` now returns **404** with body `detail: "Taxonomy run not found"` and `code: "not_found"` — previously **502** `bad_gateway`. On screen the behaviour is unchanged: the soft "something went wrong / retry" warning.
- [ ] **Hub off.** Unset `HUB_API_KEY` and restart. Any taxonomy call that used to 502 now returns **503** with `code: "service_unavailable"`. Confirm the body does **not** contain the string `HUB_API_KEY`. The Topics & Subtopics page shows the localized "service unavailable" alert with a Retry button, and the Generate button is disabled.
- [ ] **Hub unreachable (must stay 502).** Point `HUB_API_URL` at a dead port and restart. Calls return **502** `bad_gateway` — *not* 503. This is the case that must not be confused with "not configured".
- [ ] **No internal detail leaks.** In any of the failure cases above, inspect the raw JSON response body. It must not contain `hub.formbricks.com` or a `https://.../problems/...` URL. The `detail` should be one of the fixed strings: "Failed to load taxonomy run", "Failed to load record counts", "Failed to load feedback records", "Failed to start taxonomy generation", "Failed to rename taxonomy node", "Failed to remove taxonomy node".
- [ ] **Fields/state shape.** With the Hub unreachable, inspect `GET /api/v3/unify-feedback/taxonomy/fields` and `/state`. Both still return **200** with `unavailable: true` (they must not become 4xx/5xx — that would trip a false "not enough feedback" gate). The `unavailableMessage` field is now absent; the on-screen alert text is unchanged because it comes from the app's own translations.

**Preconditions / test data**

- A workspace on a plan with Unify / feedback directories enabled, and at least one feedback directory holding enough open-text records to generate a taxonomy.
- A user with `readWrite` on the workspace (needed for generate / rename / remove).
- Ability to restart the app with `HUB_API_KEY` / `HUB_API_URL` changed — the 503 and 502 cases can't be produced from the UI alone.
- A running Hub for the happy path.

**Risks & regressions**

- The blast radius is the six taxonomy v3 endpoints (`fields`, `state`, `runs`, `runs/{id}`, `runs/{id}/record-counts`, `nodes/{id}`, `nodes/{id}/records`). Nothing else calls `hubFailureResponse`.
- Watch the run-poll loop specifically: a run that 404s must still surface the soft warning and must not hard-error the page or stop the page rendering the rest of the taxonomy.
- `isHubNotConfigured` is the one genuinely subtle bit — re-check that a **timeout / dead socket** is still a 502 and has not been reclassified as 503.
- `unavailableMessage` was removed from the response and from the client type. Re-check that the fields/state unavailable alert still renders with its Retry button and that Generate is still disabled while unavailable.
- Log-level change on `getTaxonomyRun`: a genuine Hub 5xx on a run poll must still write a `warn` with the error; only 404s dropped to `debug`.

- none

### DIFF
--- a/apps/web/app/api/v3/unify-feedback/taxonomy/lib/operations.test.ts
+++ b/apps/web/app/api/v3/unify-feedback/taxonomy/lib/operations.test.ts
@@ -5,14 +5,17 @@ import {
   getActiveTaxonomyTree,
   getTaxonomyRun,
   listTaxonomyFields,
+  listTaxonomyNodeRecordCounts,
   listTaxonomyNodeRecords,
   listTaxonomyRuns,
   removeTaxonomyNode,
   renameTaxonomyNode,
 } from "@/modules/hub/service";
 import type { FeedbackRecordData, TaxonomyNode, TaxonomyRun } from "@/modules/hub/types";
+import { NO_CONFIG_ERROR } from "@/modules/hub/utils";
 import { getSessionUserId, requireUnifyDirectoryAccess } from "./access";
 import {
+  getV3TaxonomyNodeRecordCounts,
   getV3TaxonomyNodeRecords,
   getV3TaxonomyRun,
   getV3TaxonomyState,
@@ -36,6 +39,7 @@ vi.mock("@/modules/hub/service", () => ({
   getTaxonomyRun: vi.fn(),
   createTaxonomyRun: vi.fn(),
   listTaxonomyNodeRecords: vi.fn(),
+  listTaxonomyNodeRecordCounts: vi.fn(),
   renameTaxonomyNode: vi.fn(),
   removeTaxonomyNode: vi.fn(),
 }));
@@ -80,6 +84,23 @@ const node: TaxonomyNode = {
   updated_at: "2026-07-01T00:00:00.000Z",
 };
 
+/**
+ * Hub failures, shaped the way the SDK actually produces them: it has no `message` field to read off an
+ * RFC 9457 body, so it stringifies the whole body — internal problem URLs included — into `message`.
+ * Nothing built from these may reach the response, which is what the `toContain` guards below check.
+ */
+const HUB_INTERNAL_MARKER = "hub.formbricks.com/problems";
+const hubNotFound = {
+  status: 404,
+  message: `404 {"type":"https://${HUB_INTERNAL_MARKER}/not-found","title":"Not Found"}`,
+  detail: "",
+};
+const hubServerError = {
+  status: 500,
+  message: `500 {"type":"https://${HUB_INTERNAL_MARKER}/internal","title":"Internal Server Error"}`,
+  detail: "",
+};
+
 const record: FeedbackRecordData = {
   id: "rec-1",
   collected_at: "2026-07-01T00:00:00.000Z",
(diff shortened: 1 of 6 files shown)


BhagyaAmarasinghe (approved):
LGTM!

PR #8721, opened 2026-07-31
refactor(surveys): one naming rule for declared field keys [ENG-1834]

itsjavi:
## What & why

Declared field keys had **three divergent naming rules**: `isSafeIdentifier` in `apps/web/lib/utils/safe-identifier.ts` (contact-attribute keys), an inline copy of nearly the same rule in the variable schema, and a looser `[a-zA-Z0-9_-]` rule for hidden fields. `packages/types` cannot import from `apps/web`, which is exactly why the variable schema carried its own copy — a structural blocker for the Embedded Data schema work.

This promotes `isSafeIdentifier`, `toSafeIdentifier` and `formatSnakeCaseToTitleCase` into `packages/types/safe-identifier.ts` and routes every declared-field-key call site through it. The old module is deleted and all 13 importers repointed — no re-export shim, since a second path to the same helper is the drift this change exists to remove.

The two halves of the rule are split by **when** a name is created, because the same Zod schemas parse both editor input and surveys loaded from the database:

- **New names are strict.** `validateId` gained an opt-in `requireSafeIdentifier`, passed by the hidden-fields editor; the variables editor's two separate checks (charset + leading letter) collapse into one `isSafeIdentifier` call. Element/question-ID renames deliberately stay lenient — tightening them would stop users naming a question `Q1`, which is out of scope. Under `requireSafeIdentifier` the reserved-name check also becomes case-insensitive and covers the link-survey system params, both read from a single shared `RESERVED_DECLARED_FIELD_NAMES` set — otherwise the editor accepts a name (`lang`, `userid`, …) that the ingestion guard then silently refuses to fill. The lenient path keeps its exact-case check, so question-ID renames are untouched.
- **The load path stays lenient.** `ZSurveyHiddenFields` and `ZSurveyVariable` now call named legacy predicates (`isLegacyIdCharset`, `isLegacyVariableName`) holding the exact former regexes, so existing surveys with caps/hyphen hidden fields or a `_legacy` variable keep parsing byte-for-byte.

Hidden-field URL ingestion also became case-tolerant (D12): a survey declaring `CustomerRef` is now filled by `?customerref=x` as well as `?CustomerRef=x`, normalised onto the declared spelling.

### The reserved-param guard

Case-tolerant matching needs a guard, and the claim about that guard in the first version of this description was wrong — corrected here, with thanks to @pandeymangg ([#discussion_r3726967023](https://github.com/formbricks/formbricks/pull/8721#discussion_r3726967023)).

`ZSurveyHiddenFields` rejects reserved names via `FORBIDDEN_IDS.includes()` — case-**sensitively** — so a survey can legally declare `Verify`, and verification emails link to `/s/?verify=<jwt>`, the server-side credential read by `verify-email-gate.ts`. That was **not** exploitable before this PR. `URLSearchParams.get()` is case-sensitive, so the old exact-match loop asked for `Verify` and read nothing from `?verify=`; the lowercase `verify` that the product actually emits is un-declarable, because `FORBIDDEN_IDS` matches that spelling exactly. Capturing the token would have required a `?Verify=` URL, which nothing in the product produces.

What makes the guard necessary is the case-insensitive matching **this PR introduces**: without it, a declared `Verify` would newly start reading `?verify=`. So this is a hazard avoided inside the same change, not a pre-existing hole being closed. Same for the `verifiedEmail` overwrite. Ingestion now skips any param whose lowercased key is reserved, derived from `FORBIDDEN_IDS` itself so the lists cannot drift, following the precedent already set by `getPrefillValue`.

The guard does remove one **real** pre-existing behaviour, which is why the breaking-change label stays: a declared field whose name matched a reserved param key *exactly* did capture it. `lang`, `startAt`, `preview`, `skipPrefilled`, `offlineSupport` and `suToken` are absent from `FORBIDDEN_IDS`, so they were freely declarable, and `?lang=de` genuinely did fill a hidden field named `lang`. Those stop capturing — see the table below.

`packages/types` had no test runner at all (no `test` script, no vitest config, zero test files), so tests placed there would have silently never run. Added the minimal runner matching sibling packages.

Unblocks ENG-1833 — the Embedded Data schema's `key`/`name` validation can now import the helper from `packages/types`.

### Known gap, deliberately not fixed here

`RESERVED_DECLARED_FIELD_NAMES` gates the **editor only**. `ZSurveyHiddenFields` stays lenient by design for the load path, so `PUT /api/v1/management/surveys/<id>` can still create a hidden field named `lang` or `userid`, which `getHiddenFieldsFromSearchParams` will then refuse to fill — the same silent-empty-field failure this PR prevents in the UI, reached through the API instead.

Closing it means splitting create-time validation from load-time validation at the schema layer, which is the same problem #8687 has for a newly created `storageKey`. Both want one answer, not two, so rather than solving it twice it is tracked as a third enforcement point on [ENG-1839](https://linear.app/formbricks/issue/ENG-1839/reserved-field-catalog-tier-1-name-collision-rules), which already owns the blocklist-for-new-fields rule.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1834/one-naming-rule-promote-issafeidentifier-to-a-shared-package

## How this was tested

Based on `epic/embedded-data-v1`. That branch is currently 26 commits behind `main`, including four `package.json` files and 409 lockfile lines, so the whole set below was re-run against the rebased base after a `pnpm install --frozen-lockfile` (which passed, confirming the rebased lockfile matches the rebased manifests) — not carried over from the earlier main-based run.

- `packages/types` — new suite, **44 tests ✅** (`pnpm exec vitest run`). Covers the strict rule, both legacy predicates, `matchDeclaredFieldName` precedence/tie-breaks, label formatting for repeated/leading/trailing underscores, and the reserved-name matrix (every name in `RESERVED_DECLARED_FIELD_NAMES` in both casings, plus proof the lenient path is unchanged).
- `apps/web` new + affected specs — **121 tests ✅** (`hidden-fields.test.ts`, `legacy-field-names.test.ts`, `editor/lib/validation.test.ts`).
- Import-repoint blast radius — **599 tests ✅** across `modules/ee/contacts` and `modules/api/v2/management/contact-attribute-keys`. (601 on a main-based run; `epic/embedded-data-v1` simply has two fewer tests in that path.)
- `turbo typecheck --filter=@formbricks/types --filter=@formbricks/database --filter=@formbricks/web --concurrency=1` — **12/12 successful ✅**, all cache misses so it genuinely executed.
- `eslint --quiet` on all changed paths — **exit 0 ✅** (`packages/types`, `packages/database/zod`, 21 changed `apps/web` files). `prettier --check` clean.
- `pnpm i18n` — **all validations passed ✅**, 0 unused keys, all locales complete.

Tests added: `packages/types/safe-identifier.test.ts`, `packages/types/surveys/validation.test.ts` (the `validateId` strict/lenient matrix, including `Q1` still accepted), `apps/web/modules/survey/link/lib/hidden-fields.test.ts`, `apps/web/lib/survey/legacy-field-names.test.ts` (a real survey with `Legacy-Field_1`, `UserID` and `_legacy` parsed through the exported `ZSurvey`, asserting names survive byte-for-byte).

The reserved-param guard was additionally checked by a throwaway sweep over 398,920 declared×param casing combinations across every `FORBIDDEN_ID` and system param: zero captures of any reserved name under any casing, while non-reserved names still resolve under every casing. That scratch file was deleted; the committed tests assert the same invariant over each reserved name in lower/upper/original casing on both sides.

Review follow-up (`isLegacyFieldIdentifier` → `isLegacyIdCharset`) re-verified: `packages/types` **44 ✅**, `apps/web` affected specs **121 ✅**, `tsc --noEmit` in `packages/types` exit 0, `eslint` + `prettier --check` clean on the four touched files. Pure rename, no behaviour change.

No UI change to screenshot — the editor validation messages change text only (noted under Breaking changes), and no layout, component or styling was touched.

## Breaking changes

One change to declare: response data that used to be captured no longer is. No migration or manual action, but it changes what an existing published survey records.

| Reserved URL params are never captured into hidden fields | A hidden field whose name matched a reserved param key **exactly** captured that param's value from the link-survey URL — `URLSearchParams.get()` is case-sensitive, so only an exact match ever read anything. In practice that meant the six link-survey system params (`lang`, `startAt`, `preview`, `skipPrefilled`, `offlineSupport`, `suToken`), which are absent from `FORBIDDEN_IDS` and so freely declarable; plus a `FORBIDDEN_IDS` name declared in a casing `includes()` missed (`Verify`, `Source`, `UserId`, …) **and** arriving in that same casing, which nothing in the product generates | Those params are never captured, in any casing | Link surveys declaring a hidden field that collides with a reserved name — realistically `lang` or `startAt` | Rename the hidden field to a non-reserved name if you relied on capturing that value |
| Reserved names can no longer be created in the editor | The editor accepted a new hidden field named `userid`, `lang`, `startat`, `preview`, `sutoken`, `skipprefilled`, `offlinesupport`, `verifiedemail`, `multilanguage` or `welcomecard` | Those names are rejected up front as reserved, because the guard above would refuse to fill them — a field that can never receive a value must not be creatable | Anyone creating a new hidden field with one of those names | Pick a non-reserved name. Already-stored surveys holding such a name keep loading and are not rewritten; they simply stop capturing, per the row above |

Both rows are the same behaviour change seen from the two ends. Neither is a security fix — see "The reserved-param guard" above for why the `?verify=` case was never actually reachable.

Three changes that are explicitly **not** breaking, listed so nobody has to go check:

- **The stricter naming rule is editor-only.** `requireSafeIdentifier` is passed from exactly one place, the hidden-fields editor card. `ZSurveyHiddenFields` — the schema behind the Management API, survey duplication and template import — stayed lenient, so `PUT /api/v1/management/surveys/` still accepts `hiddenFields.fieldIds: ["Customer-Ref"]`. No integration sees a new rejection; only the editor form does. (This leniency is also the known gap tracked on ENG-1839 above.)
- **Case-insensitive hidden-field matching is additive.** A declared `CustomerRef` newly picks up `?customerref=x` where it previously captured nothing. Nothing that captured a value stops capturing it, and an exact match still wins over a case-insensitive one.
- **`formatSnakeCaseToTitleCase` is not a pure move.** The `.filter()` dropping empty segments is new, so `a__b` now renders as `A B` where it previously gave `A  B`. Six call sites use the helper, all deriving a contact-attribute display `name`. Flagged because the diff otherwise reads as a relocation — the behaviour change is intentional (CodeRabbit [#discussion_r3692684237](https://github.com/formbricks/formbricks/pull/8721#discussion_r3692684237)) and covered by the QA item below. It only ever affects a generated label, never a stored key, and only for keys with repeated, leading or trailing underscores.

## QA / Test Plan

**How to test**

- [ ] **Legacy hidden fields still load and capture (highest-risk regression).** Open an existing survey that has a hidden field with uppercase or a hyphen (or create one via the Management API `PUT /api/v1/management/surveys/` with `hiddenFields.fieldIds: ["Legacy-Field_1"]`, which is still accepted). → The survey opens in the editor with the field listed unchanged, and publishing works. Open the link survey with `?Legacy-Field_1=hello` → response shows `Legacy-Field_1 = hello` in the responses table, spelled exactly as declared.
- [ ] **Legacy variables still load.** Open a survey with a variable named `_legacy` (API-created, since the editor never allowed it). → Survey loads and saves without a validation error.
- [ ] **New hidden field rejects a non-conforming ID.** Survey editor → Hidden Fields → type `Customer-Ref` → Add. → Toast error: "Hidden field ID must start with a lowercase letter and can only contain lowercase letters, numbers, and underscores." Then type `customer_ref` → Add → succeeds.
- [ ] **New hidden field rejects a reserved name in any casing.** Survey editor → Hidden Fields → type `lang` → Add. → Rejected as a reserved keyword, no pill added. Repeat with `userid`, `startat`, `preview`. Then type `language_pref` → Add → succeeds, confirming only exact reserved names are blocked and not merely names containing them.
- [ ] **Question IDs stay lenient (must NOT have tightened).** Survey editor → a question's Advanced settings → change its ID to `Q1`. → Accepted, no error. This is the main thing to re-check; if `Q1` is rejected, the strict rule leaked to the wrong path.
- [ ] **New variable name shows the full rule.** Survey editor → Variables → add a variable named `_score`. → Inline error: "Variable ID must start with a lowercase letter and can only contain lowercase letters, numbers, and underscores." (Previously said only "Only lower case letters, numbers, and underscores are allowed", which `_score` already satisfies.) Then `score` → accepted.
- [ ] **Case-insensitive hidden-field capture.** Survey with hidden field `CustomerRef`, published as a link survey. Open `?customerref=abc` → response records `CustomerRef = abc` (declared spelling, not `customerref`). Open `?CustomerRef=xyz` → records `CustomerRef = xyz`. With both `?customerref=a&CustomerRef=b` present → records `b`, the exact match.
- [ ] **A system param that really did capture, now stops (the actual behaviour change).** Via the Management API create a hidden field named `lang` (the editor now refuses it). Open the link survey with `?lang=de`. → The response does **not** record a `lang` value, and the survey still switches to German. Before this change `lang = de` was captured into the response. Repeat with `startAt`.
- [ ] **Reserved param stays uncaptured under the new case-insensitive matching (regression guard, not a fix).** Survey with hidden field named `Verify`, published as a link survey with email verification enabled. Request the verification email, then open the link from that email (URL contains `?verify=`). → The response does NOT contain a `Verify` value; no token appears in the responses table or in a CSV export of that response. Repeat with `?Verify=` and `?VERIFY=` → still not captured. Note this also captured nothing before the PR (`get("Verify")` never read `verify=`); the check is that case-tolerant matching did not open it up.
- [ ] **Other reserved params likewise.** Hidden field `Source`, open `?Source=newsletter` → not captured. Hidden field `UserId`, open `?userid=123` → not captured.
- [ ] **Contact attribute keys unchanged.** Contacts → Attributes → Create attribute. Key `Job-Title` → rejected with "Key must be a safe identifier: only lowercase letters, numbers, and underscores, and must start with a letter". The auto-fill runs **Label → Key**, not the other way round: type Label `Job Title` and Key fills in as `job_title` via `toSafeIdentifier`, which saves. (Label is required, so the `formatSnakeCaseToTitleCase(key)` fallback is unreachable from this modal — it only applies to the v1/v2 management routes with `name` omitted and to the CSV/bulk upsert paths.) Contacts CSV upload mapping to a new attribute column `Some Column` → still normalises to `some_column`. Existing contacts with default camelCase attributes (`userId`, `firstName`, `lastName`) still load in the contacts table and in the API.
- [ ] **Derived attribute label has no double space.** Via the v2 management API, create a contact attribute key with `key: "a__b"` and **no** `name` → stored display name is `"A B"`, not `"A  B"`. (Only reachable through the API and CSV/bulk import, per the note above.)

**Preconditions / test data**

- A workspace with at least one link survey and one app survey.
- Email verification enabled on one link survey, plus a reachable inbox, for the reserved-param check.
- Management API access (v1 or v2) to create the legacy-named hidden field, the `lang` hidden field and the `_legacy` variable — the editor now blocks those names by design, so they cannot be created through the UI.
- Contacts feature available (EE) for the attribute-key checks.
- No feature flag, env var, plan or entitlement gates any of this. Applies equally to Cloud and self-hosted.

**Risks & regressions**

- **Survey loading is the big one.** Every survey parses through `ZSurveyHiddenFields` / `ZSurveyVariable`, so a mistake in the legacy predicates would stop surveys loading entirely. Re-check that surveys with unusual hidden-field and variable names open, save, publish, and accept responses.
- **Question/element ID renaming** shares `validateId` (and `isLegacyIdCharset`) with hidden fields; confirm `Q1`-style IDs are still accepted.
- **Link-survey URL handling** — prefill (`?questionId=value`), `?startAt=`, `?lang=`, single-use links (`?suId=`/`?suToken=`), `?userId=` contact identification, and embed mode all read the same query string. Confirm each still works alongside hidden fields.
- **Contacts** — 13 files changed import path only, but that path feeds attribute-key creation, CSV upload mapping, and the bulk contacts API. Covered by 601 passing unit tests; a smoke test of CSV upload is still worth it.
- Two files in the contacts upload UI have unrelated Tailwind class-order reordering from `prettier --write` on pre-existing drift; no visual change.

- none. No DB migration, no schema change, no env var. Existing names are shimmed, never rewritten or backfilled.
- Adds `vitest` and `@vitest/coverage-v8` as devDependencies of `packages/types` (lockfile updated); no runtime dependency change.


packages/types/safe-identifier.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Example: "job_description" -> "Job Description"
+ *          "api_key" -> "Api Key"
+ *          "signup_date" -> "Signup Date"
+ */
+export const formatSnakeCaseToTitleCase = (key: string): string => {
+  return key
+    .split("_")
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(" ");
+};
```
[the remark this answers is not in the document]
itsjavi: Fixed in 5bb46839e — good catch on the output, though not via that path.

`formatSnakeCaseToTitleCase` is never called on hidden field ids or element ids; its only call sites are contact attribute keys (create-attribute modal, the v1/v2 management routes, and the CSV/API upsert paths), so `isLegacyFieldIdentifier` is not what makes this reachable. `isSafeIdentifier` is — it requires a leading lowercase letter but is indifferent to repeated and trailing underscores, so `a__b` and `api_key_` are keys the modal accepts outright, and leading underscores can arrive from the unvalidated upsert paths. Any of those rendered a label with stray spaces.

Applied the `.filter()` and added a case covering repeated, leading and trailing underscores plus a bare `_`. `packages/types` now at 39 tests ✅.

apps/web/lib/survey/legacy-field-names.test.ts
```diff
@@ -0,0 +1,67 @@
+import { mockSurvey } from "@/app/api/(internal)/pipeline/lib/__mocks__/survey-follow-up.mock";
+import { describe, expect, test } from "vitest";
+import { ZSurvey } from "@formbricks/types/surveys/types";
```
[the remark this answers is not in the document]
itsjavi: Skipping this one — the suggested order is the one the repo forbids.

`.prettierrc.js` runs `@trivago/prettier-plugin-sort-imports`, and the shared `importOrder` in `packages/config-prettier/prettier-preset.js` puts `__mocks__` imports *first*, ahead of `<THIRD_PARTY_MODULES>`, with the reason inline: "Mocks must be at the top as they contain vi.mock calls." `mockSurvey` comes from `@/app/api/(internal)/pipeline/lib/__mocks__/survey-follow-up.mock`, which matches that first group, so the current order is exactly what Prettier produces — `pnpm exec prettier --check` on the file passes as-is.

Moving that import last would fail `prettier --check` in CI. Leaving it.

packages/types/safe-identifier.ts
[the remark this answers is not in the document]
itsjavi: Correction to my reply above, from QA on this branch: I named the wrong reachable path.

The create-attribute modal **requires** a Label, so `formData.name` is never empty there and the `formatSnakeCaseToTitleCase(formData.key)` fallback at `create-attribute-modal.tsx:132` is unreachable from the UI. (The modal actually derives in the opposite direction — typing Label `Job Title` auto-fills Key `job_title` via `toSafeIdentifier`.)

The paths that do reach it are the v1/v2 management routes with `name` omitted (`name: data.name ?? formatSnakeCaseToTitleCase(data.key)`) and the bulk-contacts / CSV upsert paths (`contacts.ts:438`, `attributes.ts:379`). Those gate new keys on `isSafeIdentifier` (`contacts.ts:414`, `bulk/lib/contact.ts:538`), which accepts `a__b` — so the defect and the fix both stand, just on the API/import surface rather than the modal.

Everything else in the fix is unchanged.

apps/web/modules/survey/link/lib/hidden-fields.ts
```diff
@@ -0,0 +1,39 @@
...
+    const matchedParamKey = matchDeclaredFieldName(incomingParamKeys, declaredFieldId);
+    if (matchedParamKey === undefined) continue;
+
+    // `ZSurveyHiddenFields` rejects reserved names case-sensitively, so an already-stored survey can
+    // hold a field named `Verify` or `UserId` (the editor now refuses to create one). Matching
+    // case-insensitively here would turn that into a way to capture reserved params - most seriously
+    // `?verify=<jwt>`, the email-verification credential read by `verify-email-gate.ts`, which would
+    // then be written to the response and every export.
+    // Checked on the resolved param key so no casing on either side gets through.
+    if (RESERVED_DECLARED_FIELD_NAMES.has(matchedParamKey.toLowerCase())) continue;
```
pandeymangg, line 32: The guard is right, but I think the framing in the PR body is backwards — worth correcting before this is labelled a security fix.

`URLSearchParams.get()` is case-sensitive:

```js
new URLSearchParams("verify=TOKEN").get("Verify") // => null
```

So the old inline loop, with a field declared as `Verify` and the product's own link `?verify=<jwt>`, captured nothing. Verification emails generate lowercase `verify`, and `verify` is in `FORBIDDEN_IDS`, so a field with the matching casing can't be declared in the first place. The pre-existing hole needed the URL casing to match the declared field's casing, which nothing in the product produces.

What makes this guard necessary is the case-insensitive matching **this PR introduces** — without it, declared `Verify` would newly start reading `?verify=`. That's a hazard avoided in the same change, not an exploitable bug being closed. Same for the `verifiedEmail` overwrite.

Keep the code exactly as is; just narrow the claim in the body and the breaking-changes table.

packages/types/surveys/validation.ts
```diff
@@ -65,6 +66,31 @@ export const FORBIDDEN_IDS = [
...
+  "offlinesupport",
+];
+
+/**
+ * Every name a declared field must never take, lowercased. The single source of truth shared by the
+ * two ends that have to agree: `validateId` refuses to create such a name, and
+ * `getHiddenFieldsFromSearchParams` refuses to capture a param with such a key. When the two lists
+ * disagree the editor happily accepts a field that can never receive a value.
+ */
+export const RESERVED_DECLARED_FIELD_NAMES = new Set([
```
pandeymangg, line 89: `RESERVED_DECLARED_FIELD_NAMES` only gates the editor — `requireSafeIdentifier` is passed from one place, and `ZSurveyHiddenFields` stays lenient by design for the load path.

So `PUT /api/v1/management/surveys/<id>` can still create a hidden field named `lang`, `startat` or `userid`, and `getHiddenFieldsFromSearchParams` will then refuse to fill it — the exact silent-empty-field failure this set exists to prevent, just reached through the API instead of the UI.

This is the same gap you raised on #8687 ("no strict counterpart for a newly created `storageKey`"), so it's symmetric: the create side needs one guard, and it'd be good to settle where it lives once rather than twice. Fine as a follow-up — flagging so it doesn't fall between the two PRs.

packages/types/safe-identifier.ts
```diff
@@ -0,0 +1,120 @@
...
+    .filter((word) => word.length > 0)
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(" ");
+};
+
+/**
+ * Legacy shim for the load path: the character rule hidden field ids and element ids were
+ * stored under before `isSafeIdentifier` existed. Survey schemas validate already-persisted
+ * names with this, so tightening it would stop existing surveys from loading. Editors gate new
+ * names with `isSafeIdentifier`.
```
pandeymangg: Naming nit: via `validateId` this now also governs element and question ids, not just hidden fields, so the name reads narrower than the use. Something like `isLegacyIdCharset` would match its actual scope.

packages/types/safe-identifier.ts
```diff
@@ -0,0 +1,120 @@
...
+ *          "api_key" -> "Api Key"
+ *          "signup_date" -> "Signup Date"
+ *
+ * Empty segments are dropped, because `isSafeIdentifier` allows leading, trailing and repeated
+ * underscores: without the filter a key like `a__b` would render with a double space.
+ */
+export const formatSnakeCaseToTitleCase = (key: string): string => {
+  return key
+    .split("_")
+    .filter((word) => word.length > 0)
```
pandeymangg, line 79: Worth calling out that this one isn't a pure move: the `.filter()` is new, so `a__b` renders as `A B` where it previously gave `A  B`. Eight files use this helper. Intentional and QA'd per the body — just noting it's a behaviour change rather than a relocation, in case someone reads the diff as move-only.

pandeymangg (commented):
Reviewed this because ENG-1833 depends on it, so I checked the load-path claims against the base rather than taking them on trust. The parts that could break existing surveys are right:

- `LEGACY_FIELD_IDENTIFIER_REGEX` and `LEGACY_VARIABLE_NAME_REGEX` are byte-exact copies of the two regexes they replace — load path unchanged.
- `isSafeIdentifier` / `toSafeIdentifier` are untouched, so ENG-1833's `key` validation is unaffected.
- `validateId`'s new param is optional and defaulted, so no existing caller changes behaviour.
- I enumerated every search param the link-survey module reads (`offlineSupport`, `skipPrefilled`, `startAt`, `verify`, `embed`, `lang`, `preview`, `suId`, `suToken`) — `RESERVED_DECLARED_FIELD_NAMES` covers all nine, and deriving it from `FORBIDDEN_IDS` keeps it from drifting.
- Leaving `ZContactAttributeKey.key` un-refined is the right call and the cited `attribute-key-policy.ts` does exist.
- Both replaced i18n keys are gone from code and all 16 locale files, so the "0 unused keys" result holds.

Four comments below. Only the first two are worth acting on, and neither blocks the merge.

On ordering: this should land before #8687 — that PR ships a strict subset of this one's `packages/types/safe-identifier.ts` and `package.json`, and I'll drop both from it once this is in.

apps/web/modules/survey/link/lib/hidden-fields.ts
itsjavi: You're right, and I've rewritten the description. Verified the mechanism rather than taking it on trust:

```
new URLSearchParams("verify=TOKEN").get("Verify")  // => null
new URLSearchParams("lang=de").get("lang")         // => "de"
```

The old loop was `searchParams.get(field)`, so declared `Verify` read nothing from the product's own lowercase `?verify=`, and lowercase `verify` is un-declarable because `FORBIDDEN_IDS` matches that spelling exactly. Capturing the token needed a `?Verify=` URL that nothing generates. Hazard introduced and closed in the same change, not a pre-existing hole. Code unchanged, as you said.

Chasing it down turned up a second thing wrong in the body, in the same family. The other breaking-changes row claimed reserved names "stayed empty forever with no warning" **before** this PR — but there was no ingestion guard before, and the six link-survey system params (`lang`, `startAt`, `preview`, `skipPrefilled`, `offlineSupport`, `suToken`) are absent from `FORBIDDEN_IDS`, so they were freely declarable and exact-matched. `?lang=de` genuinely did fill a hidden field named `lang`. So there *is* a real behaviour removal here, just a mundane one — which is why the `breaking-change` label stays, now justified by `lang`/`startAt` rather than by `verify`.

Changes to the description:

- New "The reserved-param guard" section stating the hazard-avoided framing explicitly, including why `?verify=` was unreachable.
- Both table rows rewritten with the true "Before" — exact-case matching, and which params that actually covered.
- Added: "Neither is a security fix."
- QA plan: the `Verify` item is relabelled a regression guard and says outright that it captured nothing before either; added a new item for `lang`, which is the case that really did change.



---

packages/types/surveys/validation.ts
itsjavi: Agreed on both counts — real gap, and it should be settled once rather than in each PR separately.

Tracked as a third enforcement point on [ENG-1839](https://linear.app/formbricks/issue/ENG-1839/reserved-field-catalog-tier-1-name-collision-rules), which already owns "Blocklist for new fields" and listed exactly two: `EmbeddedData.key` (ENG-1833) and `storageKey` in the editor reconcile (ENG-1978). Adding the legacy `hiddenFields.fieldIds` create path there keeps all three under one ticket, so the shared answer — a create-time validation layer distinct from the lenient load-time schema — gets designed once. `RESERVED_DECLARED_FIELD_NAMES` is already the source of truth for *which* names to refuse; only the enforcement point is missing.

The PR description now carries a "Known gap, deliberately not fixed here" section spelling out the API path and pointing at ENG-1839, so it's visible to anyone reading this PR rather than living only in a review thread.

Leaving this unresolved for you to close if the ticket placement looks right.



---

packages/types/safe-identifier.ts
itsjavi: Renamed to `isLegacyIdCharset` in 479dd12. You're right that the old name described the first call site rather than the predicate — `validateId` routes element and question ids through it too, and the constant had the same problem (`LEGACY_FIELD_IDENTIFIER_REGEX` → `LEGACY_ID_CHARSET_REGEX`).

Took the chance to record *why* those ids never graduate to the strict rule, since that's the question the next reader will have:

```ts
/**
 * Legacy shim for the load path: the character rule every declared id was stored under before
 * `isSafeIdentifier` existed. Via `validateId` this governs element and question ids as well as
 * hidden field ids, hence the charset-level name. Survey schemas validate already-persisted names
 * with this, so tightening it would stop existing surveys from loading — and element/question ids
 * never graduate to the strict rule, because `Q1` has to stay a legal question id. Editors gate
 * new *declared field* names with `isSafeIdentifier` on top of this.
 */
```

Four files touched (definition, `surveys/types.ts`, `surveys/validation.ts`, the spec). Pure rename: `packages/types` 44 ✅, `apps/web` affected specs 121 ✅, `tsc --noEmit` exit 0, eslint + `prettier --check` clean.



---

packages/types/safe-identifier.ts
itsjavi: Fair — it was buried in a QA checkbox and a code comment, which is not where someone skimming a "promote helper to shared package" diff would look. Promoted to an explicit bullet in the Breaking changes section, alongside the two other not-breaking-but-worth-knowing items:

> **`formatSnakeCaseToTitleCase` is not a pure move.** The `.filter()` dropping empty segments is new, so `a__b` now renders as `A B` where it previously gave `A  B`. Six call sites use the helper, all deriving a contact-attribute display `name`. […] It only ever affects a generated label, never a stored key, and only for keys with repeated, leading or trailing underscores.

One correction to your count: it's six call sites, not eight — `rg` over non-test source gives `contacts.ts:438`, `attributes.ts:379`, `contact-attribute-keys.ts:46`, `create-attribute-modal.tsx:132`, v1 `contact-attribute-keys.ts:54` and v2 `contact-attribute-key.ts:67`. The other two hits are the definition and its spec. All six derive a display `name`, never a key, so the blast radius is labels only.



---

pandeymangg (approved):
LGTM! 🚀

PR #8727, opened 2026-08-03
fix(helm): bound Hub worker API wait

BhagyaAmarasinghe:
## What & why

A hub-worker init container could hang indefinitely when its first Hub API health request connected before the endpoint was ready but never returned. Because the request itself had no hard total deadline, the surrounding retry counter and maximum-attempt guard were ineffective.

This wraps each GNU Wget health request in the pinned image's `timeout` utility for a hard five-second process deadline while retaining Wget's DNS/connect/read timeouts. It also adds attempt-aware logs and replaces the inaccurate elapsed-time failure message. Chart documentation now describes the bounded request and retry behavior, including the disabled wait mode, and Helm validation covers the default, overridden, disabled, and shared-volume rendering paths independently.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2181/helm-hub-worker-can-hang-indefinitely-while-waiting-for-hub-api

## How this was tested

- Executed the complete `helm-chart-validation.yml` render-validation script locally ✅
- Rendered the default wait, `maxAttempts=3`, disabled wait, default-volume, and secret-backed shared-volume variants; the validation now checks `timeout 5`, `sleep 5`, and Hub API/worker volume propagation independently ✅
- Verified the pinned Hub 0.8.1 image provides BusyBox `/usr/bin/timeout` and `/usr/bin/wget --timeout`; a two-second process-timeout probe terminated in two seconds ✅
- Initial Minikube 1.38.1 / Kubernetes 1.35.1 verification: simulated both connection refusal and a socket that accepted but did not return HTTP headers; the same worker pod logged three bounded retries and started automatically when Hub returned HTTP 200 ✅
- Initial Minikube verification: Hub API and worker mounted the credentials file with matching SHA-256, with zero init/main-container restarts ✅
- Attempted to rerun the Kubernetes scenario after the review follow-up, but the recreated local Minikube control plane never started its kubelet/API server; the broken profile was removed. The exact post-review timeout command was instead verified in the pinned runtime image as noted above.
- `git diff --check` ✅

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Deploy the Helm chart while the Hub API endpoint remains unavailable or nonresponsive for more than five seconds → hub-worker stays in its wait init container, logs increasing bounded attempt numbers, and does not remain stuck in one request.
- [ ] Restore Hub API health before `hub.worker.waitForApi.maxAttempts` is exhausted → the existing hub-worker pod completes initialization and becomes Ready without being deleted or recreated.
- [ ] Configure a Secret through `hub.extraVolumes` and `hub.extraVolumeMounts` → Hub API and hub-worker both become Ready and can read the same mounted file.
- [ ] Set `hub.worker.waitForApi.enabled=false` → hub-worker starts without the API-wait init container.

**Preconditions / test data**

- A self-hosted Kubernetes test cluster with the Formbricks Helm chart.
- Ability to delay and restore the Hub API endpoint.
- A test Kubernetes Secret containing a non-sensitive credentials file.

**Risks & regressions**

- Hub worker startup timing and retry exhaustion.
- Compatibility of `timeout` and Wget health-check flags with the chart-pinned Hub image.
- Shared Hub API/worker volume and mount rendering.

- none

### DIFF
--- a/.github/workflows/helm-chart-validation.yml
+++ b/.github/workflows/helm-chart-validation.yml
@@ -64,11 +64,13 @@ jobs:
             --set formbricks.webappUrl=https://qa.example.com \
             --show-only templates/hub-worker-deployment.yaml > "$render_dir/hub-worker-default.yaml"
           grep -q 'max_attempts=120' "$render_dir/hub-worker-default.yaml"
-          grep -q 'wget --no-verbose --tries=1 --timeout=5 --spider' "$render_dir/hub-worker-default.yaml"
+          grep -q 'timeout 5 wget --no-verbose --tries=1 --timeout=5 --spider' \
+            "$render_dir/hub-worker-default.yaml"
           grep -q 'Waiting for Hub API migrations and health check (attempt \$attempts/\$max_attempts)' \
             "$render_dir/hub-worker-default.yaml"
           grep -q 'Hub API did not become healthy after \$max_attempts attempt(s)' \
             "$render_dir/hub-worker-default.yaml"
+          grep -q 'sleep 5' "$render_dir/hub-worker-default.yaml"
 
           helm template qa "$chart" \
             --set formbricks.webappUrl=https://qa.example.com \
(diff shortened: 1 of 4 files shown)


xernobyl (commented):
Resolve the Coderabbit comments and it should be good to go :)

pandeymangg (approved):

PR #8730, opened 2026-08-04
fix(web): exit non-zero when typegen env validation fails

Dhruwang:
## What does this PR do?

Fixes ENG-2183: `pnpm --filter=@formbricks/web typegen` now exits non-zero when environment validation fails.

`next typegen` loads `next.config.mjs`, which imports `lib/env`. When required variables are missing or invalid, `@t3-oss/env-nextjs` throws — but that failure surfaces only as an *unhandled promise rejection* during config load: the offending variables are printed, yet the process still exits `0`. CI and scripts therefore can't detect that type generation ran against an invalid environment.

This wraps typegen in `apps/web/scripts/typegen.ts`:

1. Loads the same `.env` files Next does, via `@next/env`'s `loadEnvConfig` (so the preflight sees exactly the environment Next would).
2. Imports `lib/env`, running the identical `@t3-oss/env-nextjs` validation up front.
3. On failure → prints the offending variables and exits `1` **before** `next typegen` is ever spawned.
4. On success → spawns `next typegen`, generates types, exits `0`, and propagates its exit code.

The package `typegen` script now runs `tsx scripts/typegen.ts` instead of `next typegen` directly (same injected env vars, unchanged otherwise).

Fixes ENG-2183.

## How should this be tested?

```sh
# Failure path — invalid var, exits non-zero, lists the variable, no success message
DATABASE_URL=not-a-valid-url pnpm --filter=@formbricks/web typegen; echo "exit: $?"

# Success path — valid root .env, generates types, exits 0
pnpm --filter=@formbricks/web typegen; echo "exit: $?"
```

Automated: `apps/web/scripts/typegen.test.ts` spawns the wrapper with an invalid `DATABASE_URL` and asserts a non-zero exit + "Invalid environment variables" in output, with no success message.

## QA / Test Plan

- [ ] `pnpm --filter=@formbricks/web typegen` exits non-zero when a required env var is missing/invalid.
- [ ] The validation output still lists the missing/invalid variables.
- [ ] No "Types generated successfully" message after a validation failure.
- [ ] A valid root `.env` generates types and exits `0` (existing behavior unchanged).
- [ ] `pnpm --filter=@formbricks/web typecheck` (which runs typegen) still passes.

### Required

- [x] Ran `pnpm build`

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -12,7 +12,7 @@
     "build:dev": "pnpm run build",
     "start": "next start",
     "typecheck": "pnpm typegen && tsc --noEmit --project tsconfig.typecheck.json",
-    "typegen": "dotenv -e ../../.env -- next typegen",
+    "typegen": "dotenv -e ../../.env -- tsx scripts/typegen.ts",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "test": "dotenv -e ../../.env -- vitest run",
--- a/apps/web/scripts/typegen.test.ts
+++ b/apps/web/scripts/typegen.test.ts
@@ -4,27 +4,32 @@ import { fileURLToPath } from "node:url";
 import { describe, expect, test } from "vitest";
 
 const scriptDir = path.dirname(fileURLToPath(import.meta.url));
-const scriptPath = path.join(scriptDir, "typegen.ts");
 const appDir = path.resolve(scriptDir, "..");
 
 // Regression guard for the failure exit-code path: `next typegen` used to swallow
 // environment-validation failures and still exit 0, so CI could not detect that
-// type generation ran against an invalid environment. The wrapper must now surface
-// that failure as a non-zero exit while still listing the offending variables.
+// type generation ran against an invalid environment. The `typegen` script must
+// now surface that failure as a non-zero exit while still listing the offending
+// variables.
+//
+// This drives the real package entrypoint (`pnpm typegen`) rather than the wrapper
+// module directly, so a regression in the script command itself — e.g. re-adding a
+// hardcoded `cross-env DATABASE_URL=...` that would mask a caller-supplied invalid
+// value — is also caught.
 //
 // The success path (valid env -> exit 0, "Types generated successfully") is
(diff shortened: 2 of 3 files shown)


apps/web/package.json
```diff
@@ -12,9 +12,8 @@
     "build:dev": "pnpm run build",
     "start": "next start",
     "typecheck": "pnpm typegen && tsc --noEmit --project tsconfig.typecheck.json",
-    "typegen": "dotenv -e ../../.env -- next typegen",
-    "lint": "eslint .",
-    "lint:fix": "eslint . --fix",
+    "typegen": "cross-env DATABASE_URL=postgresql://postgres:postgres@localhost:5432/formbricks ENCRYPTION_KEY=example REDIS_URL=redis://localhost:6379 tsx scripts/typegen.ts",
```
BhagyaAmarasinghe: The inline cross-env values overwrite DATABASE_URL, ENCRYPTION_KEY, and REDIS_URL supplied by the caller or root .env. I reproduced DATABASE_URL=not-a-valid-url pnpm typegen exiting 0 and printing ✓ Types generated successfully when the remaining variables were valid.
The unit test bypasses the package entrypoint by executing typegen.ts directly, so it cannot catch this regression. Rebase onto main and resolve the conflict as:
`"typegen": "dotenv -e ../../.env -- tsx scripts/typegen.ts"`

The test should also exercise the real package entrypoint.

BhagyaAmarasinghe (changes requested):

apps/web/package.json
Dhruwang: Good catch — fixed both. Rebased onto main and resolved the conflict to `dotenv -e ../../.env -- tsx scripts/typegen.ts`, dropping the hardcoded cross-env values that were masking a caller/`.env`-supplied invalid variable. Confirmed `DATABASE_URL=not-a-valid-url pnpm typegen` now exits 1, and a valid env still generates types + exits 0. The regression test now drives the real `pnpm typegen` entrypoint (not the wrapper module directly), so a reintroduced clobber would be caught.

BhagyaAmarasinghe (approved):

PR #8733, opened 2026-08-04
fix(workflows): report a blocked autosave state for an empty name

Dhruwang:
## What does this PR do?

Fixes ENG-2171: in the workflow editor, clearing the (pre-selected) title left the autosave pill stuck on **"Saving…"** forever while nothing was being saved and no inline error appeared; a rejected name also showed the raw `The request payload is invalid.` instead of the real reason.

**Root cause:** an empty name makes `buildWorkflowPatch` bail before any request (`{ invalid: { code: "name_required" } }`), so `save()` records no error and `isDirty` stays true. The pill's precedence (`hasFailed` → `isDirty || isSaving` → `saved`) then fell into the dirty branch and rendered "Saving…". Separately, `save()` stored only the generic `detail` and discarded the server's field-level `invalid_params`.

Screenshot 2026-08-04 at 1 57 56 PM

**Fixes (the four directions from the ticket):**

1. **Blocked pill state.** New amber `blocked` state ("Not saved"), shown when the draft is dirty but the name is empty (`isDirty && !isNameValid`), ranked above "Saving…". An unsendable draft never claims to be saving. `workflow-auto-save-indicator.tsx`
2. **Inline name error.** The title field shows a red border + `Give the workflow a name.` message with `aria-invalid` / `aria-describedby`, instead of the error living only in the validation-problems dialog. `workflow-page-title.tsx`
3. **Client-side max length.** `maxLength={120}` on the title input (mirrors the `name: z.string().min(1).max(120)` contract) so an over-long name never round-trips into a generic payload error.
4. **Field-level rejection copy.** `save()` maps `V3ApiError.invalid_params` into the pill tooltip, so a rejected name reads "Name must be between 1 and 120 characters." instead of "invalid payload". `use-workflow-builder.ts`

Description handling was intentionally left out — workflows have no description-editing UI (`setWorkflowDescription` is unused; the create CTA sets `description: null`).

## How should this be tested?

1. New workflow → editor opens with the title focused + selected.
2. Clear the title. Expect: amber **"Not saved"** pill (not "Saving…"), red title border, inline "Give the workflow a name." Add/edit a node — still not saving.
3. Type a name. Expect: pill → "Saving…" → green "All changes saved"; inline error clears.
4. (Contract) The title input caps at 120 chars.

## QA / Test Plan

- [ ] Empty name → pill shows "Not saved" (amber), never stuck on "Saving…".
- [ ] Empty name → inline error + red border on the title field; screen reader announces it.
- [ ] Typing a valid name clears the blocked state and autosaves (pill → saved).
- [ ] Title input cannot exceed 120 characters.
- [ ] A rejected save surfaces a field-level reason in the pill tooltip (not "The request payload is invalid.").

## Notes

- `pnpm i18n` run: new en-US keys (`autosave_blocked`, `autosave_blocked_tooltip`, `name_invalid`) translated to all locales via Lingo; `scan-translations` passes.
- `pnpm test` (workflows state/hook/api-error suites): 123 passed. `pnpm lint`: clean.
- Component behavior (the pill states) should also be covered by an editor e2e case — flagging for the e2e-tester.

### Required

- [x] Removed all console.logs


apps/web/modules/ee/workflows/components/workflow-auto-save-indicator.tsx
```diff
@@ -56,6 +65,11 @@ export const WorkflowAutoSaveIndicator = () => {
   if (hasFailed) {
     state = "failed";
     label = t("workspace.workflows.autosave_failed");
+  } else if (isBlocked) {
```
itsjavi, line 68: Nit — a stale `saveError` can mask the new `blocked` state.

`hasFailed` (checked above `isBlocked`) is `saveError !== null && isDirty`, and `hasWorkflowSaveFailedAtom`/`isWorkflowDirtyAtom` aren't scoped to the specific draft that failed. Sequence: user has a valid name, autosave gets rejected by the server (`saveError` set, kind `\"rejected\"`) → user then clears the name entirely. `buildWorkflowPatch`'s bail path (`invalid` in `built`) never touches `saveError`, so it's still non-null, and `isDirty` is still true (draft ≠ last persisted) → `hasFailed` stays true and the pill shows **\"Save failed\"** with the old rejection detail, not the new amber **\"Not saved\"** — even though nothing was attempted this time and the real reason is the empty name.

Narrow edge case (requires a prior real save failure this session before the name is cleared), and `hasFailed`-outranks-all is a deliberate pre-existing design (per the comment at line 59-62), so this may be intentionally out of scope. But since it's the same failure mode the ticket is about (pill misreporting an unsendable draft), might be worth clearing `saveError` in the invalid-bail path of `save()`, or gating `hasFailed` on `saveError?.draftSignature === draftSignature` matching the current draft.

---

itsjavi (commented):
Reviewed the diff (workflow autosave indicator, page title, and the save-error mapping in `use-workflow-builder.ts`, plus the i18n additions).

The four fixes described in the PR body all check out against the code:
- `isBlocked` (`isDirty && !isNameValid`) correctly matches `workflowValidityAtom`'s own `isNameValid = workflowName.trim().length > 0`, so the pill and the validation-problem list can't disagree.
- The inline error wires `aria-invalid`/`aria-describedby`/`role="alert"` correctly and reuses the existing `validation_problem_name_missing` key rather than introducing a duplicate string.
- `maxLength={120}` matches the actual contract (`z.string().min(1).max(120)` in `packages/workflows/src/contracts/inputs.ts`), so the comment tying it to `ZPatchWorkflowInput` is accurate.
- `describeSaveErrorDetail` correctly gates on `status < 500` (a 5xx/unreachable failure gets the generic tooltip, not a misleading "reason") and maps the `name` field-level rejection to the new `name_invalid` string.

Left one inline note on a narrow edge case: a stale `saveError` from an earlier rejected save can outrank the new "blocked" state if the name is cleared afterward, since `buildWorkflowPatch`'s bail path never touches `saveError`. Not a regression this PR introduces (the `hasFailed`-outranks-all precedence is pre-existing), just flagging since it's the same failure mode the ticket targets.

CI is fully green (lint, unit tests, build, E2E, CodeQL, SonarCloud) and there are no open CodeRabbit or Sonar findings. I can't run the dev server/DB in this environment to visually confirm the pill states and focus behavior described in the "How to test" steps, so that manual pass is for a human reviewer. @itsjavi

---

---

apps/web/modules/ee/workflows/components/workflow-auto-save-indicator.tsx
Dhruwang: Good catch — fixed in fc7be6e5. Went with the second option: `hasWorkflowSaveFailedAtom` is now gated on `saveError.draftSignature === workflowDraftSignatureAtom`, so a stale rejection from an earlier draft no longer masks the amber "Not saved" state once the name is cleared. This also mirrors the autosave effect's own no-retry guard (same signature check), and leaves the saveError object intact so that guard keeps working. Updated the two existing tests to record the failure against the live signature and added one covering the scoped behaviour.

itsjavi (commented):
Verified the follow-up fix in `fc7be6e5` ("scope the save-failed pill to the draft that failed"): `hasWorkflowSaveFailedAtom` now gates on `saveError.draftSignature === workflowDraftSignatureAtom`, which correctly closes the edge case I flagged — clearing the name after a prior rejected save now surfaces the amber "Not saved" state instead of a stale "Save failed" with an outdated reason. The two existing tests were updated to record the failure against the live signature, and a new test covers the scoped behavior directly. No remaining issues in the diff.

CI is fully green (lint, unit tests, build, E2E, CodeQL, SonarCloud) and there are no open CodeRabbit or Sonar findings.

I still can't spin up the dev server/DB in this environment to visually confirm the pill states and focus behavior from the "How to test" steps, so that manual pass remains for a human reviewer. @itsjavi

---

---

itsjavi (approved):

PR #8736, opened 2026-08-04
fix(workflows): scope send_email recipients to workspace access

itsjavi:
## What & why

Revoking a team's access to a workspace did not stop that team's members from receiving the workspace's workflow `send_email` messages — and the editor actively told the author otherwise.

Two different definitions of "allowed recipient" had drifted apart:

- **The recipient picker was workspace-scoped.** `email-authoring-context.ts` resolved options from the teams linked to the workflow's workspace, so unlinking the team made the member disappear — and because Radix `Select` falls back to its placeholder when the stored value matches no item, the field rendered **empty** while the definition still held the address.
- **The enable/dry-run gate and the runner's send-time backstop were organization-scoped.** Both called `getOrganizationMemberEmails`, which filtered only on `organizationId` + `user.isActive`. The member was still an active org member, so both checks passed and the mail kept going out — carrying response data (`attachResponseData`, `includeVariables`, `includeHiddenFields`) from a workspace the admin had explicitly cut them off from.

This resolves the two scopes onto one workspace-scoped source of truth, `getWorkspaceMembers` / `getWorkspaceMemberEmails` in `apps/web/lib/workspace/service.ts`, now used by the picker, the enable/test gate and the runner backstop alike. "Can access this workspace" mirrors what the real request path already enforces in `checkAuthorizationUpdated`: an active organization member who is either an `owner`/`manager` (they reach every workspace) or on a team linked to that workspace (at any `WorkspaceTeam` permission — `read` already grants access). Including owners/managers matters: the picker always offers the current user as "Yourself", so a team-only allowlist would have made an owner unable to address themselves.

Also fixed, per the ticket's secondary issue: the recipient field now renders a stored-but-unavailable recipient as its own flagged entry with an explanation, instead of silently showing nothing. Member options are matched case-insensitively, the same normalization the runtime allowlist uses, so a `to` written as `Manager@Example.com` through the v3 API or MCP is not falsely reported as unavailable.

Rounding it out: the enable/test/runner failure messages now say the recipient cannot access the workspace rather than "is not a member of the organization"; the now-callerless `getOrganizationMemberEmails` is removed; and the v3 API reference (which documented the old rule, including a verbatim error string the API no longer returns) is updated and re-bundled.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2186/fix-revoking-a-teams-workspace-access-doesnt-stop-workflow-emails-to

## How this was tested

**Automated**

- `pnpm exec vitest run` in `packages/workflows` — 13 files, 233 tests ✅
- `pnpm exec vitest run` in `apps/web` across the touched areas (`lib/workspace`, `lib/organization`, `app/api/v3/workflows`, `modules/ee/workflows`, `modules/survey/follow-ups`) — all passing ✅
- `pnpm exec turbo run typecheck --filter=@formbricks/web --filter=@formbricks/workflows --concurrency=1` — 11/11 ✅
- `pnpm exec eslint --quiet` on the changed files (run per package) — clean ✅
- `pnpm exec prettier --check` on the changed files — clean ✅
- `node docs/api-v3-reference/scripts/bundle.mjs --lint` and `--check` — spec valid, bundle in sync with `src/` ✅

Tests added/updated: workspace-access predicate (owner/manager vs team-linked, org-resolved-from-workspace, active-only, normalization, fail-closed empty set); the workspace-scoped enable gate; a runner case for a member whose team lost workspace access; `findEmailSendToOption` matching rules; and the reworded assertions in the runner and handler specs. No `.tsx` unit tests, per AGENTS.md.

**Manual — the ticket's repro, end to end on a local instance (mailhog)**

Seeded an org with a workspace, a team linked to it, and a plain `member` recipient on that team, then ran the workflow twice against a real response:

| Step | Result |
| --- | --- |
| Enable workflow addressed to the team-linked `member` | 200 — allowed by the new gate |
| Submit response (team still linked) | run `completed`, `send_email` succeeded, **mail delivered** to the member |
| Unlink the workspace from the team | — |
| Submit response again | run **`failed`**: `Recipient eng2186-manager-…@example.com cannot access this workspace`, **no mail delivered** (mailhog total stayed at 1) |

That is the ticket's "Expected: the send is refused (revocation takes effect)".

**Manual — the editor (screenshots below)**

- *Before*: with the recipient no longer available, "Send to" renders completely blank while the canvas node still reads "Send an email to …".
- *After*: the stored address renders with a warning icon and an explanation.
- *After (case variant)*: a recipient stored as `ENG2186-Owner-…@Example.com` resolves to the member row with no warning — the case that would have regressed into a false warning without normalized matching.

## Breaking changes

None.

Workflows ships in 5.3 and this is a pre-release QA fix, so no released contract changes. Worth knowing for testing, though: the recipient rule is genuinely tighter than it was on `main` — an organization member who cannot reach the workflow's workspace can no longer be a literal `send_email` recipient, and a workflow already enabled with such a recipient will now fail that step instead of sending. That is the behavior the ticket asks for.

## QA / Test Plan

**How to test**

- [ ] Preconditions below. In workspace B, create a workflow with a **Response completed** trigger bound to a survey, add a **Send email** action, and pick the manager from "Send to". Enable it → the workflow enables successfully.
- [ ] Submit a response to that survey → the manager receives the email. (Confirms the happy path still works.)
- [ ] Go to **Settings → Teams**, open the team, and remove **workspace B** from it. The team keeps workspace A only.
- [ ] Submit another response to the same survey → **no email is delivered to the manager**. Open the workflow's **Runs** tab: the latest run is **failed**, and the `send_email` step error reads `Recipient <email> cannot access this workspace`.
- [ ] Reopen the workflow in workspace B and click the Send email node → the "Send to" field shows the manager's address with a warning triangle and the message "This recipient is no longer available — they may have lost access to this workspace…". It must **not** be blank.
- [ ] Open the "Send to" dropdown → the manager is no longer offered as an option; members who still have access are.
- [ ] Re-add workspace B to the team, reload the workflow → the manager is offered again, the warning is gone, and a new response delivers the email.
- [ ] Edge — owner/manager without a team: as an org **owner** who belongs to no team, create a workflow in any workspace and address it to yourself ("Yourself" in the picker). Enable → succeeds, and a response delivers the mail. (Owners/managers reach every workspace, so revoking team access must not affect them.)
- [ ] Edge — external address: address a Send email action to an address belonging to nobody in the organization and try to enable → enable is refused with "A send_email recipient cannot access this workspace." (ENG-2029 protection must still hold.)
- [ ] Edge — respondent field: address the action to a survey question / hidden field holding the respondent's email instead of a person → still sends normally, unaffected by any team change.
- [ ] API — `POST /api/v3/workflows/{workflowId}/test` with a literal recipient who cannot access the workspace → `problems` contains `recipient_not_allowed` with the message naming workspace access; `POST /{workflowId}/enable` returns **422** with `detail` "A send_email recipient cannot access this workspace."

**Preconditions / test data**

- An organization with **two workspaces** (A and B), an **owner** and a **manager/member**, and a **team** containing both users, initially linked to both workspaces.
- Workflows entitlement (EE license / Cloud plan) — the feature is enterprise-gated.
- SMTP configured (or mailhog on a local instance) so delivery can be observed.
- A link survey in workspace B with at least one response path.

**Risks & regressions**

- Recipient selection in **survey Follow-Ups** shares `email-send-to-options.ts`. Follow-Ups behavior is deliberately unchanged (it applies no allowlist — see ENG-2029); re-check that its recipient dropdown still lists the team roster and that follow-up emails still send.
- Workflows already enabled before this change with a recipient who lacks workspace access will now fail that step. Intended, but worth watching for on staging data.
- The picker roster is now the workspace's full access list rather than just team members, so org owners/managers appear as options where they previously did not.
- `getOrganizationMemberEmails` was removed; confirm nothing else regressed around organization member lookups.

- none


pandeymangg (dismissed):
LGTM! 🚀

pandeymangg (approved):
LGTM! 🚀

PR #8744, opened 2026-08-04
fix(logger): resolve OTEL transport from runtime root

BhagyaAmarasinghe:
## Summary
- resolve the Pino OTEL transport from the standalone runtime root when the app process cwd is apps/web
- update the Docker image smoke check to simulate the real Next.js runtime cwd before loading the transport
- cover the standalone cwd case in the logger unit test

## Staging evidence
- Endpoint-only OTEL was enabled in staging with OTEL_LOGS_ENABLED=0 and remained healthy
- Enabling OTEL_LOGS_ENABLED=1 exposed the remaining runtime error: Pino looked for /home/nextjs/apps/web/node_modules/pino-opentelemetry-transport/... while the package exists at /home/nextjs/node_modules/...
- Logs were disabled again; staging returned to Synced/Healthy with public /health = 200

## Validation
- pnpm --filter @formbricks/logger test
- pnpm --filter @formbricks/logger typecheck
- pnpm --filter @formbricks/logger lint
- git diff --check
- local Pino OTEL transport smoke command from an apps/web cwd

### DIFF
--- a/apps/web/Dockerfile
+++ b/apps/web/Dockerfile
@@ -299,7 +299,9 @@ RUN chmod -R 755 \
       ./node_modules/@opentelemetry/exporter-logs-otlp-proto \
     && OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318 node -e "\
       const pino = require('pino'); \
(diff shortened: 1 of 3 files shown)


xernobyl (approved):
LGTM

PR #8762, opened 2026-08-05
docs: warn about stale package dist after a branch switch

xernobyl:
## What & why

`AGENTS.md` already warns that `packages/surveys` is consumed as a **build** and that Turborepo caches it aggressively. The same thing is true of every workspace package consumed through its built output, and that half isn't written down: `@formbricks/ai` and `@formbricks/database` both resolve via `dist/` in their `exports` map, so `apps/web` imports the build, not `src/`. A `git switch`, rebase, or pull updates `src/` and leaves `dist/` exactly as it was, silently.

What makes it worth documenting is that **the symptom doesn't look like a stale build**. A symbol added on the branch you just checked out is simply missing from `dist/`, so the import resolves to `undefined`:

- `TypeError: Right-hand side of 'instanceof' is not an object` — the class is in `src/`, not in `dist/`
- a missing named export, or `Cannot find module '@formbricks/…'`
- `tsc`/Vitest failures in files you never touched

This cost real time recently: four failing tests in `app/api/v3/surveys/generate/error-mapping.test.ts` looked like a broken `main` introduced by an unrelated PR, and were nearly reported as such. The actual cause was a `packages/ai/dist` built days before the export existed — rebuilding that one package turned 4 failed / 4 passed into 8 passed with no other change.

Adds a short `### Stale package builds after a branch switch` subsection next to the existing surveys one covering: the failure signatures, the rebuild command, a one-line way to confirm the diagnosis before chasing it, and the tell — CI builds from scratch every run, so **local-red plus CI-green is this until proven otherwise**.

**On the red `sync-labels` check:** unrelated to this diff. The workflow labelled this docs-only PR `breaking-change` because its "nothing here" test only accepted the bare word `None`; removing that label by hand then raced with the body edit that still observed it, so the job's `DELETE /issues/8762/labels/breaking-change` returned `404 Label does not exist` and the unhandled rejection failed the run. Both halves are fixed in #8763, merged to `main` on 7 Aug. Because `pr-label-sync.yml` runs on `pull_request_target`, the workflow definition comes from the base branch, so the fixed version governs the next event on this PR — the red run above is stale and re-running it would only replay the old code. Nothing to do on this branch.

## Linear ticket

No ticket — documentation-only fix for a recurring local-dev footgun, in the same spirit as the surveys build/cache note directly above it.

## How this was tested

Documentation only; no product code touched (`AGENTS.md`, +26/-0).

Verified the claims rather than asserting them:

- `pnpm build --filter=@formbricks/web^...` is not invented — it's the exact command `.github/workflows/integration-tests.yml:125` already runs before its suites, with a comment giving this same reason (deps resolve `exports` → `dist`). The doc now cross-references it.
- Confirmed both named packages really are dist-consumed: `@formbricks/ai` → `"./dist/index.js"`, `@formbricks/database` → `dist/index.js` / `dist/index.cjs`.
- Confirmed the described failure and its fix end to end: with a stale `packages/ai/dist`, `grep -c AIOutputTokenLimitError` was `0` in `dist/index.js` while present in `src/errors.ts`, and the suite was 4 failed / 4 passed; after rebuilding only that package, `2` and 8 passed.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Read the new `### Stale package builds after a branch switch` subsection in `AGENTS.md` — it should sit directly after `### Survey Packages Build & Cache`, both under `## Build, Test & Development Commands`.
- [ ] Optional, reproduces the trap: from a branch where `packages/ai/dist` predates a new export, run `grep -c "AIOutputTokenLimitError" packages/ai/src/errors.ts packages/ai/dist/index.js` → `src` non-zero, `dist` `0`. Run `pnpm build --filter=@formbricks/web^...`, repeat → both non-zero, and `pnpm test app/api/v3/surveys/generate/error-mapping` goes green.

**Preconditions / test data**

- None. A checkout whose `packages/*/dist` predates the current branch is what triggers it; a freshly built tree won't reproduce.

**Risks & regressions**

- None. No executable content changed — the only risk is guidance going stale if a package stops resolving through `dist/`, which is why the section names the mechanism (`exports` → `dist`) rather than a fixed package list.

- none


AGENTS.md
```diff
@@ -57,6 +57,32 @@ Tailwind is configured CSS-first everywhere. Only two JS/TS Tailwind configs rem
...
+
+(Dependencies only — `^...` excludes the Next app itself. This is the same command `.github/workflows/integration-tests.yml` runs before its suites, for exactly this reason.)
+
+To confirm the diagnosis first, check whether the symbol is actually in the built output — if `src/` has it and `dist/` doesn't, that's this problem and nothing else:
+
+```
+grep -c "MyNewExport" packages/<pkg>/src/*.ts packages/<pkg>/dist/index.js
+```
+
+CI never hits this — it builds from scratch every run — so a failure that reproduces locally but is green on CI is this until proven otherwise.
```
BhagyaAmarasinghe: **This tell is inverted, and it's the part readers will act on.**

CI's unit job has no build step at all — `test.yml:44` is `pnpm install` → `pnpm dev:setup` → `pnpm test`, nothing else. It stays green because `turbo run test` pulls the builds in itself, via root `turbo.json:184`:

```json
"@formbricks/web#test": {
  "dependsOn": ["@formbricks/ai#build", "@formbricks/logger#build", "@formbricks/database#build", ...]
}
```

`turbo run test --filter=@formbricks/web --dry=json` resolves 13 `#build` tasks ahead of the suite, and the log of a green run (31820585949) shows `@formbricks/ai:build` running *inside* `turbo run test`. That graph landed in #7611, so it predates this PR — the claim was already inaccurate when written, not drift.

A developer's `pnpm test` runs that identical graph, so the trap cannot bite there, and telling them to rebuild after a red `pnpm test` sends them to a no-op. The real boundary is **turbo vs. bypassing turbo**: `npx vitest run <file>` inside `apps/web` (how I reproduced it), `pnpm --filter @formbricks/web test`, bare `tsc`, and IDE test runners all skip the `dependsOn` graph. That's almost certainly how this was originally hit — iterating on a single test file.

Suggested reframe, which keeps everything valuable about the section and makes the trigger checkable:

> This bites when you bypass Turborepo — running `vitest`/`tsc` directly inside `apps/web`, `pnpm --filter @formbricks/web test`, or an IDE test runner. The root `pnpm test`/`pnpm typecheck` are safe: `@formbricks/web#test` in `turbo.json` already depends on `@formbricks/ai#build`, `@formbricks/database#build`, and friends — which is also why CI stays green.

AGENTS.md
```diff
@@ -57,6 +57,32 @@ Tailwind is configured CSS-first everywhere. Only two JS/TS Tailwind configs rem
 reached through an explicit `@config` bridge from the package's own stylesheet. Do not add a
 `tailwind.config.js` that nothing `@config`s — Tailwind v4 will not load it, and it will silently rot.
 
+### Stale package builds after a branch switch
+
+The same trap applies to **every** workspace package that is consumed through its built output rather than its source — `@formbricks/ai` and `@formbricks/database` resolve via `dist/` in their `exports` map, so `apps/web` imports the build, not `src/`. `git switch`, a rebase, or a pull changes `src/` but leaves `dist/` exactly as it was, and nothing warns you.
```
BhagyaAmarasinghe: **Dangling referent.** "The same trap" reads as pointing at the section directly above, which is `### Tailwind & Workspace Package CSS` — about a `tailwind.config.js` that nothing `@config`s silently rotting. The trap actually being generalised is the surveys build/cache one, two sections up.

Either name the referent explicitly ("The same trap as `### Survey Packages Build & Cache` applies to…") or move this section to sit directly after that one — which is where the PR description's QA checklist already claims it lives, so moving it fixes both at once.

AGENTS.md
```diff
@@ -57,6 +57,32 @@ Tailwind is configured CSS-first everywhere. Only two JS/TS Tailwind configs rem
...
+```
+pnpm build --filter=@formbricks/web^...
+```
+
+(Dependencies only — `^...` excludes the Next app itself. This is the same command `.github/workflows/integration-tests.yml` runs before its suites, for exactly this reason.)
+
+To confirm the diagnosis first, check whether the symbol is actually in the built output — if `src/` has it and `dist/` doesn't, that's this problem and nothing else:
+
+```
+grep -c "MyNewExport" packages/<pkg>/src/*.ts packages/<pkg>/dist/index.js
```
BhagyaAmarasinghe: `src/*.ts` is non-recursive, and `packages/ai/src/providers/` exists (`aws.ts`, `azure.ts`, `google.ts`, `openai-compatible.ts`). A symbol defined in a nested file reads as absent from **both** `src` and `dist`, which muddies exactly the diagnosis this snippet is meant to confirm.

```
grep -rc "MyNewExport" packages/<pkg>/src packages/<pkg>/dist/index.js
```

BhagyaAmarasinghe (changes requested):
## Review: stale package `dist` after a branch switch

The underlying mechanism is real and I reproduced it — with no `packages/ai/dist`, `apps/web/app/api/v3/surveys/generate/error-mapping.test.ts` dies with `Failed to resolve entry for package "@formbricks/ai"`. Both named packages really are dist-consumed, and `pnpm build --filter=@formbricks/web^...` is verbatim what `.github/workflows/integration-tests.yml:130` runs, with a comment giving the same `exports` → `dist` reason. Good cross-reference.

One substantive problem: the closing tell ("local red + CI green ⇒ stale `dist`") inverts the real axis, because local `pnpm test` runs the same Turborepo build graph CI does. Since `AGENTS.md` is the instruction file agents load every session, a wrong heuristic here propagates. Details inline.

### Also worth fixing

- The **QA checklist** in the PR description says the new section "should sit directly after `### Survey Packages Build & Cache`". It doesn't — it lands after `### Tailwind & Workspace Package CSS`. Either move the section (which would also fix the "same trap" referent below) or correct the checklist, so QA doesn't fail on step 1.
- Minor: the new section uses unwrapped long lines where the two neighbouring sections wrap at ~110 chars. `AGENTS.md` is in `.prettierignore`, so this won't trip `format:check` — purely a consistency call.

### What's good

- Leading with failure *signatures* rather than the cause is the right shape for a footgun doc; those symptoms genuinely don't read as a stale build.
- The `^...` gloss is accurate, and zero runtime risk.

Happy to re-review as soon as the tell is reframed — the rest of the section earns its place.

On the red `sync-labels` check: your reasoning holds. `pull_request_target` takes the workflow definition from the base branch, so the fix in #8763 governs the next event. Nothing to do on this branch.

AGENTS.md
xernobyl: Fixed in `7c1b37f32` — you're right, and I verified it against `origin/main` before rewriting rather than taking it on trust.

`@formbricks/web#test` does declare `dependsOn` on seven `#build` tasks (`@formbricks/ai#build`, `@formbricks/database#build`, `logger`, `storage`, `cache`, `jobs`, `surveys`), and `test.yml` really is just install → `dev:setup` → `pnpm test` with no build step. So the tell was not merely imprecise, it pointed the reader at a no-op: root `pnpm test` runs the same graph CI does.

Reframed to your axis — turbo vs bypassing turbo — and I kept your framing that iterating on a single test file is how you actually meet it, since that is almost certainly how I hit it. The section now says explicitly *not* to read "green on CI, red locally" as evidence of a stale `dist/`.

AGENTS.md
xernobyl: Fixed by moving the section rather than renaming the referent, per your suggestion — it now sits directly after `### Survey Packages Build & Cache`, so "the same trap" points at the section it is generalising, and the PR's QA checklist becomes accurate at the same time. That also closes the placement nit claude[bot] raised.

AGENTS.md
xernobyl: Fixed, and testing it turned up a second problem worth recording.

Your recursive form is in. Verified against a nested symbol: `grep -rc awsProviderAdapter packages/ai/src packages/ai/dist/index.js` reports `src/providers/aws.ts:1` and `dist/index.js:0`, while the old `src/*.ts` glob does not list the nested file at all.

The check as written was also over-confident in the other direction: `awsProviderAdapter` is genuinely absent from `dist/index.js` simply because the entry does not re-export it — so an internal helper would read as a stale `dist/` when nothing is wrong. Added a line telling the reader to pick a symbol the package really exports.

xernobyl:
All three addressed in `7c1b37f32`, replies inline. Summary:

1. **Inverted tell** — you were right, and I verified both halves against `origin/main` before rewriting: `@formbricks/web#test` declares `dependsOn` on seven `#build` tasks, and `test.yml` has no build step. Reframed to turbo-vs-bypassing-turbo, and the section now explicitly warns *against* reading "green on CI, red locally" as the signal.
2. **Dangling referent + placement** — fixed by moving the section to sit directly after `### Survey Packages Build & Cache`, which resolves the referent, the QA checklist, and claude[bot]'s nit together.
3. **Non-recursive grep** — your form is in, verified against a nested symbol. Testing it also exposed a false-positive mode I had missed (an internal helper is legitimately absent from `dist/index.js`), so the doc now says to pick a symbol the entry actually exports.

Also wrapped at ~110 to match the neighbouring sections.

Thanks for the review — the first finding was the one that mattered, and you're right that a wrong heuristic in `AGENTS.md` propagates further than a wrong heuristic anywhere else.

xernobyl:
The red Typecheck was this branch being 33 commits behind `main`, which is a pleasing way for *this* PR to fail.

`Cannot find module '@formbricks/jobs'` on a diff that touches one markdown file was obviously not the diff. At this branch's base (`88a38c0`), `@formbricks/web#typecheck` in `turbo.json` had 6 `dependsOn` entries and `@formbricks/jobs#build` was not among them; `main` has since added it (7 entries). So typecheck ran against an unbuilt `@formbricks/jobs` — a stale *task graph* rather than a stale `dist/`, but the same family of problem the section is about.

Merged `main` in (`f2041b5a1`) — no conflicts, and the diff against `main` is still just `AGENTS.md`. The merged file keeps main's newer content (the `format:check` line, the server-actions-are-legacy note) alongside the new section, still sitting directly after `### Survey Packages Build & Cache`.

Worth noting for the section's own accuracy: this failure would *not* have been caught by the reframed tell either, since it is a missing `dependsOn` edge rather than a bypassed graph. I have left the text as reviewed rather than widening it further — happy to add a sentence if you think the task-graph variant is worth naming.

AGENTS.md
```diff
@@ -29,6 +29,51 @@ The `@formbricks/surveys` package is pre-compiled (Vite → UMD + ESM) and the b
...
+
+This has cost real time: four failing tests read as a broken `main` from an unrelated PR, when the only
+problem was a `dist/` built days earlier. To confirm the diagnosis, check whether a symbol you expect
+the package to *export* is in the built output — present in `src/`, absent from `dist/`, is the
+signature. (Pick one the entry really re-exports; an internal helper is legitimately absent from
+`dist/index.js` and would read as a false positive.)
+
+```
+grep -rc "MyNewExport" packages/<pkg>/src packages/<pkg>/dist/index.js
+```
```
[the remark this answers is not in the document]
xernobyl: Applied in `b9c8dd8d7` — both fences are now `shell`.

Two notes for calibration rather than pushback: **markdownlint isn't wired into CI here** (no config in `.github/workflows/` or `package.json`), and the file's pre-existing fences are bare too — including the ones in the `Survey Packages Build & Cache` section directly above. So this is a rendering/highlighting improvement rather than a gate, and I've tagged only the fences this PR adds rather than reformatting a file that's in `.prettierignore` and outside the diff.

AGENTS.md
```diff
@@ -57,6 +57,32 @@ Tailwind is configured CSS-first everywhere. Only two JS/TS Tailwind configs rem
 reached through an explicit `@config` bridge from the package's own stylesheet. Do not add a
 `tailwind.config.js` that nothing `@config`s — Tailwind v4 will not load it, and it will silently rot.
 
+### Stale package builds after a branch switch
```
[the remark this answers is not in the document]
xernobyl: Addressed, though not in the way this suggested. Rather than updating the QA step to match where the section landed, I **moved the section** so it sits directly after `### Survey Packages Build & Cache` as the PR body always claimed.

That was the better fix because it also resolved a second problem raised separately: the section opens with "The same trap applies to…", and while it sat after the Tailwind section that referent pointed at the wrong place. Moving it makes the sentence true and the QA checklist accurate at the same time.

xernobyl:
Pushed `db9d4f776`. One finding from CodeRabbit, plus two I found doing a full re-read of the section against the code rather than only chasing the open threads. All three were accuracy problems in the half of the section that got rewritten last round, which is fair — that rewrite was where the new claims came in.

**1. The `undefined` lead-in contradicted its own bullet** (CodeRabbit, still open)

Valid, and sharper than it first looked. The paragraph said the missing symbol means "the import resolves to `undefined`", then listed `Failed to resolve entry for package "@formbricks/…"` two lines later — which is a *module-resolution* failure that only happens when `dist/` is missing altogether, i.e. the import throws and never binds anything. That is exactly how you reproduced it, @BhagyaAmarasinghe, by deleting `packages/ai/dist`, so the section was describing your repro with the wrong mechanism.

Now covers both outcomes, and the bullet says which one it is:

> …absent from `dist/`, so depending on what is missing the import either resolves to `undefined` or fails outright at module resolution
>
> - a missing named export, or `Failed to resolve entry for package "@formbricks/…"` when `dist/` is absent altogether

**2. "CI stays green with no build step of its own" was only true of one workflow — and the section cites the counterexample**

This one I'd have caught last round if I'd checked the second workflow as carefully as the first. `test.yml` has no build step, which is what I verified. But `integration-tests.yml` *does* carry one (line 130), and it needs it: line 165 runs `cd apps/web && pnpm test:integration`, whose script is `… vitest run --config vitest.integration.config.mts` — straight past the task graph. So the section asserted CI has no build step, then two paragraphs later pointed at a CI workflow that does, as the source of the rebuild command.

Annoyingly it was the *strongest* example in the repo of the thing the section is about, and I had it filed as a footnote. The no-build-step claim is now attributed to `test.yml` specifically, and the citation does the work:

> (Dependencies only — `^...` excludes the Next app itself. `integration-tests.yml` runs this before its suites for exactly this reason: it drives `cd apps/web && pnpm test:integration`, a plain `vitest run` that bypasses the task graph.)

**3. "The root commands are safe" is conditional on `dependsOn` being complete — and this branch proved it**

This is the task-graph variant I asked about above and then argued myself out of adding. Re-reading it, leaving it out made a *safety* claim unqualified, which is the same shape as the original inverted tell you flagged: a reader who trusts "root `pnpm typecheck` is safe" has no way to explain a red root typecheck.

It isn't hypothetical. At this branch's base, `@formbricks/web#typecheck` had **6** `dependsOn` entries and `@formbricks/jobs#build` was not one of them; `main` has 7. So `@formbricks/jobs` was never built at all — which fails identically to a stale `dist/`, from a different cause. One sentence, no new section:

> One caveat: a `dependsOn` list is only as good as its entries. A package missing from `@formbricks/web#test`/`#typecheck` is never built at all, which fails exactly like a stale `dist/` — so when a root command goes red, check the task graph too, not just the build.

That also let me drop "`@formbricks/database#build` and five more". The count was right (7 at the base, the head, and `main` today) but a bare number in `AGENTS.md` drifts silently on the next graph edit, which is precisely how #3 happened.

**Verified, not assumed** — re-checked every claim in the section against the head, since the last round showed how easy the plausible-but-wrong ones are here:

- both `exports` maps resolve to `dist/` (`ai` → `dist/index.js`; `database` → `dist/index.js` + `dist/index.cjs`)
- `pnpm --filter @formbricks/web test` really does bypass turbo — the script is `dotenv -e ../../.env -- vitest run`
- `apps/web/turbo.json` overrides only `build`, so the root `#test`/`#typecheck` `dependsOn` lists still govern (they'd be *replaced*, not merged, if it declared them)
- `#test` and `#typecheck` are identical 7-entry sets at the head, so "the workspace builds they need" is accurate for both
- `packages/ai/src/providers/` exists, so the recursive grep still earns its "on purpose"
- `AGENTS.md` is in `.prettierignore:31`, so none of this touches `format:check`

Still docs-only: `AGENTS.md`, +52/-0. @BhagyaAmarasinghe ready for another look when you have a moment 🙏 — and thanks for pushing back on the tell originally, since #2 and #3 are both the same failure of mine that you caught the first time, just further down the section.

AGENTS.md
```diff
@@ -29,6 +29,58 @@ The `@formbricks/surveys` package is pre-compiled (Vite → UMD + ESM) and the b
...
+usually meet it, iterating on one test file. The root `pnpm test` and `pnpm typecheck` go through
+turbo, and `@formbricks/web#test` / `@formbricks/web#typecheck` in `turbo.json` each declare
+`dependsOn` on the workspace builds they need (`@formbricks/ai#build`, `@formbricks/database#build`,
+and the rest), so turbo rebuilds those first. That is why the unit-test workflow (`test.yml`) needs no
+build step of its own — do **not** read "green on CI, red locally" as evidence of a stale `dist/`;
+it and the root commands run the same graph.
+
+One caveat: a `dependsOn` list is only as good as its entries. A package missing from
+`@formbricks/web#test`/`#typecheck` is never built at all, which fails exactly like a stale `dist/` —
+so when a root command goes red, check the task graph too, not just the build.
```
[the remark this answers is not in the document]
xernobyl: Valid, and it invalidated the general rule rather than just the wording. Fixed in `feabcb665`.

I'd written "a package missing from `@formbricks/web#test`/`#typecheck` is never built at all". Root `build` is `dependsOn: ["^build"]` — topological — so that only holds for a package nothing in the list reaches. Your example is exactly right: `@formbricks/ai` depends on `@formbricks/logger`, so `logger` builds even if its explicit `@formbricks/logger#build` entry were dropped. Same for `cache`, `database`, and `storage`, all of which depend on `logger`. As written, the sentence would have sent a reader hunting a missing edge that changes nothing.

The concrete example survives, which is why the error was easy to miss: `@formbricks/jobs` sits *downstream* (`jobs` → `database`, `logger`, `types`) and nothing in the list depends on it, so it really was unreachable and really did go unbuilt while absent from `#typecheck`. A true instance sitting under a false generalisation.

Now:

> One caveat: a `dependsOn` list is only as good as its entries — though `build` is topological (`dependsOn: ["^build"]`), so a package left out is still built when something in the list depends on it. Only one that nothing in the list reaches goes unbuilt, failing exactly like a stale `dist/` from a different cause — as `@formbricks/jobs` did while it was absent from `#typecheck`, nothing else in the list depending on it.

Verified by reading the `@formbricks/*` deps of all seven listed packages rather than reasoning about it: `ai` → `logger`, `types`; `cache` → `logger`; `database` → `logger`, `types`, `i18n-utils`, `workflows`; `storage` → `logger`; `surveys` → `types`, `i18n-utils`, `survey-ui`; `logger` → none; and no listed package depends on `jobs`.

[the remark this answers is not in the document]
xernobyl: Follow-up: I have removed the paragraph this thread was about entirely (`be202db75`), so the reworded version above is gone too.

Your finding was correct and I am not walking that back — root `build` is topological, so "a package missing from the list is never built at all" was wrong. But reviewing my own change, the paragraph should not have been in this PR at all: it did not correct anything inaccurate in the section, it introduced a second failure mode (an incomplete `dependsOn` list) adjacent to the actual topic, which is stale `dist/`. The fact that its first wording needed your correction is a reasonable signal it was not carrying its weight.

The PR is now the two places the section was genuinely wrong — the `undefined` vs module-resolution lead-in, and the `test.yml`-only CI claim — and nothing else.

xernobyl:
Scope correction on my own last two pushes — `be202db75` is what should be reviewed.

I had widened this beyond what needed fixing. Walked back to the two places the section was actually inaccurate:

1. **The `undefined` lead-in** contradicted its own second bullet. `Failed to resolve entry for package` is a module-resolution failure that happens when `dist/` is absent altogether, so the import throws rather than binding `undefined` — which is how @BhagyaAmarasinghe reproduced it. Now covers both outcomes, and the bullet says which is which.
2. **"CI stays green with no build step of its own"** held only for the unit workflow. `integration-tests.yml` carries one (it runs `cd apps/web && pnpm test:integration`, a plain `vitest run`, past the task graph). Attributed to `test.yml`.

Plus one of the same class: the sentence claims both root `pnpm test` and `pnpm typecheck` are safe, but only `@formbricks/web#test` was cited as evidence. Now cites `#typecheck` too — both are identical 7-entry `dependsOn` sets, so the claim holds for both.

**Dropped:** a caveat paragraph I had added about incomplete `dependsOn` lists. It corrected nothing — it introduced a second failure mode adjacent to the topic, and its first wording was over-broad (root `build` is topological, so a package left out is still built when something in the list depends on it; CodeRabbit caught that). Not worth a ticket in my view — the specific gap it described, `@formbricks/jobs` missing from `#typecheck`, is already fixed on `main`, so what's left is speculative. Happy to file if you disagree.

Net effect versus the version you reviewed: **+2 lines**, both corrections. `AGENTS.md` +47/-0, still docs-only.

BhagyaAmarasinghe (approved):

PR #8766, opened 2026-08-05
fix(survey-ui): announce validation errors to screen readers - a11y [ENG-1292]

itsjavi:
## What & why

Submitting a required field empty rendered "Please fill out this field" **visually only**. The error container had no live region and no programmatic association to its control, so a screen reader announced nothing: a blind respondent pressed Next, nothing appeared to happen, and the survey could not be completed. This affected every required element type.

- `ElementError` now requires an `id`, and its message container is a **permanently mounted** `aria-live="polite"` / `aria-atomic="true"` region. It no longer early-returns `null`. Regions that mount together with their content announce unreliably, and an `aria-describedby` target must never be a missing id. While empty it carries no children and no classes, so it paints nothing and takes no space. The decorative indicator bar moved outside the region and stays error-gated.
- The shared `getElementErrorAria(inputId, errorMessage)` helper now derives the error id, invalid state, and description association together. Every call site uses it, so the live-region id and control attributes cannot drift independently.
- `aria-invalid` was missing entirely on `open-text`, `date`, `cta`, `nps`, `ranking`, `file-upload` and the `picture-select` multi branch. It is now present on a semantically valid node in each case. `date` became a native `fieldset`; NPS was already a `fieldset` and now carries `role="radiogroup"` because its options are native radios.
- Replaced the modal overlay wrapper's `aria-live="assertive"` behavior with main's new `aria-live="polite"` behavior. The old assertive wrapper spanned the entire survey, so **every** DOM mutation interrupted the screen reader and could drown the scoped validation region. Modal question changes can still be announced politely without interrupting higher-priority speech.
- Marked the stuck-response retry panel `role="alert"`, matching the existing `error-component`.

Inline errors are **polite, not `role="alert"`**, on purpose. `block-conditional` already moves focus to the first invalid control at that same instant; an assertive region interrupts that announcement, leaving the user hearing the error but not *which field* failed.

Giving NPS a composite `radiogroup` role made its accessible name load-bearing, which exposed a latent bug: it was named by a hardcoded English `sr-only` legend reading "NPS rating options", omitting the question entirely, while its header's `htmlFor` pointed at a `div`, which is not labelable. It now names itself by the headline like every sibling grouping element, which also removes an untranslated user-facing string.

All four "Other" free-text inputs, single- and multi-select, dropdown and list, carry `aria-describedby`. The list variants were skipped in the first review pass on the grounds that their enclosing fieldset already described them; that reasoning was wrong (per accname, an ancestor's `aria-describedby` is not part of a descendant's accessible description, so focusing the box announced its invalid state with no reason), and b477d54d corrected it. The group-level description stays put: entering the group and focusing the input are separate events, so there is no duplicate announcement to suppress.

One deliberate non-change, from review:
- **contact-info / address** fields no longer point at the shared message. Those messages are field-scoped (`getDefaultErrorMessage` prefixes the field placeholder), so a shared reference made "First name" announce "Email: …".

### Known limitation

Contact-info and address are now the one element type where **only the live region** carries the description; every other type also has it on the control or its group. Announcement there depends on the screen reader flushing pending polite speech, rather than on a redundant `aria-describedby`. This is not a regression (those elements announced nothing at all before this PR), and the alternative, restoring the shared reference, is the defect described above.

The proper fix is per-field error nodes, so each input points at its own message and `aria-invalid` lands only on the field that failed. That changes how errors are **displayed**, not just how they are wired, and needs `rule.field` plumbed through `block-conditional` plus both element wrappers. This is a product decision worth scoping separately. Filed as follow-up.

### Follow-ups found but not fixed here

- Per-field errors for contact-info/address (above).
- Dropdown "Other" inputs have no `aria-label`, falling back to the placeholder for their accessible name, which disappears once the user types. The list variants do this correctly.
- `picture-select.tsx:120` is the last `role="group"` + `aria-invalid` site; its comment is silent rather than wrong.
- `Input`/`Textarea` declare an `errorMessage` prop they never use, so it lands in `...props` and renders a junk `errormessage="…"` attribute on the native element.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1292/validation-errors-are-not-announced-to-screen-readers-wcag-413

Also closes the F5 and F10 slices of https://linear.app/formbricks/issue/ENG-1070/follow-up-fix-wcag-21-aa-accessibility-issues-in-survey-runtime (F6, F7, F8, F12, F14 remain open there).

**Compliance relevance:** contributes to WCAG 2.0 **3.3.1 Error Identification** and **1.3.1 Info and Relationships**, both Level A and both inside the Section 508 baseline, plus WCAG 2.1 **4.1.3 Status Messages**. Relevant to the VPAT/ACR in progress.

## How this was tested

**Automated, all run locally against a full stack (PG16 + pgvector, Redis, `next dev`):**

- Full Chromium run of `apps/web/playwright/survey-keyboard.spec.ts` against the current built survey bundle and local stack: **9 passed**.
- `npx playwright test apps/web/playwright/survey-accessibility.spec.ts --project=chromium` → **9 passed**, zero WCAG AA violations across all variants (desktop / mobile / tablet / forced-colors / reduced-motion / dark / RTL Arabic / empty-submit / back-nav). Only warnings emitted were `color-contrast-enhanced` (AAA) and `region`, both in `WARN_TAGS`, informational by design.
- `pnpm --filter @formbricks/survey-ui test`: 84 passed. Current merged `@formbricks/surveys` suite: 667 passed.
- `eslint src` and `tsc --noEmit` clean in both packages; `prettier --check` clean on all changed files.
- After merging current main: forced `@formbricks/surveys` build passed, and the focused Chromium modal-announcement regression passed.

**Tests added**: 4 in the `Survey validation error announcement` suite, plus a fifth covering the "Other" free-text paths. Worth stating why they exist: **axe-core cannot detect this defect class.** It is a static scanner, and a missing live region is invisible to it, which is exactly why the ticket asked for a Playwright assertion. These are the only automated guard:

- the region exists and is **empty before** any submit, then carries the message on the same node (updated, not remounted);
- the focused control's `aria-describedby` is **resolved in-page per token** and the resolved node's text equals the visible error; a dangling *or empty* IDREF fails;
- the error id is **appended** to an existing `aria-describedby` rather than replacing it;
- the region is polite and carries **no** `role`, so a later switch to `role="alert"` fails deliberately;
- the actual modal rendering path contains a polite region, while neither its survey wrapper nor any descendant carries an assertive live region;
- consent, file upload, matrix, and contact-info keep their intended live-region and control-association behavior;
- all four "Other" boxes, single/multi × dropdown/list, are `aria-invalid` and resolve their own `aria-describedby` to the visible message, in one card so a single empty submit surfaces every error.

**Not verified, and needs a human:** a real screen-reader pass (NVDA/VoiceOver). Automated tests prove the semantics are present and resolvable; they cannot prove the announcement *sounds* right, and the polite-vs-assertive choice above specifically deserves a human ear. That belongs to ENG-1298.

During the final browser verification, one existing response-submission case timed out once while the development server compiled its response route. The isolated rerun passed, followed by a clean full 9-test run.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open any link survey with a required Open Text question → the error text is not visible, but a screen reader's element inspector shows an empty live region already present below the question.
- [ ] With a screen reader running (NVDA on Windows, VoiceOver on macOS), leave the required field empty and press Next → focus moves to the field and the error is announced; you should hear which field failed **and** the message, not the message alone.
- [ ] Type a valid answer → the error text disappears and the region goes silent; no announcement fires on clearing.
- [ ] Repeat on NPS, Ranking, Date, multi-select Picture Selection, Consent, File Upload, and Matrix questions: each announces the same way.
- [ ] On the NPS question specifically, tab into the rating scale → the screen reader announces the **question text** as the group's name (previously it announced "NPS rating options", in English regardless of survey language).
- [ ] Open a Contact Info or Address question, submit empty → the failing field reads as invalid, and the message is announced by the region. Confirm that focusing a *different* field (e.g. First name) does **not** announce another field's message (e.g. "Email: …").
- [ ] Open a Select question with an "Other" option, choose Other, leave the text empty, submit → focus lands in the Other field and its error is announced. Check all four combinations: single- and multi-select, in **dropdown** and in **list** mode.
- [ ] Open a CTA question card and compare against production → the button must sit in the same place; there should be no extra gap above it.
- [ ] Open an **app (modal) survey** and interact → the survey should no longer announce every change on screen. Previously any DOM update inside the modal was announced assertively.
- [ ] Take a link survey offline mid-submission (block the network) → the "Your feedback is stuck" panel is announced immediately as an alert.

**Preconditions / test data**

- A published link survey containing: required Open Text, NPS, Ranking, Date, Picture Selection (multi), Contact Info or Address, Select in dropdown mode with an "Other" option, and a CTA card.
- One app-type survey for the modal check.
- A screen reader: NVDA (Windows) or VoiceOver (macOS). The keyboard-only parts can be checked without one; the announcement parts cannot.

**Risks & regressions**

- **Layout:** `ElementError` now always renders a container. It was verified to have no class and no children when empty, and the one call site with a spacing utility (CTA) was fixed, but every question type is worth a visual glance against production, and Chromatic will flag any Storybook diff.
- **Focus order:** `aria-invalid` was added to several fieldsets, groups and buttons. `focusFirstControl` prefers `:is(input, textarea, select)[aria-invalid="true"]`, which those cannot match, and the existing focus test still passes, but a failed submit on every question type is worth re-checking by keyboard.
- **Date element type:** `date` changed from a `div` to a native `fieldset`, so check its UA-default resets (margin, padding, border, min-inline-size), including in dark mode and RTL. NPS was already a `fieldset`; its risk is limited to the new radiogroup role and headline-based accessible name.
- **Modal announcements:** the production survey bundle is now rendered through its modal entry point in Playwright. The test proves the modal dialog and a polite live region exist, then verifies that the modal wrapper and its descendants have no assertive live region. A real screen-reader pass is still required to assess how the resulting speech sounds.

- none


itsjavi:
Thanks — one of the two findings was right and is now fixed; the other is real but pre-existing and wider than this PR.

### `single-select.tsx` "Other" input — fixed in `b477d54dc`

Correct, and the accname point is the crux of it. An ancestor's `aria-describedby` is not part of a descendant's accessible description, so focusing the "Other" box announced it invalid with no reason. The earlier comment justifying the omission ("would announce it twice for this row") assumed a transfer that does not happen — that reasoning was wrong, and it left the list variants inconsistent with the dropdown variants, which had already been fixed for exactly this reason.

All four "Other" inputs now carry it — `single-select` and `multi-select`, dropdown and list.

I did not take the second half of the suggestion (dropping the group-level `aria-describedby` when Other is selected). Entering the group and focusing the text input are separate events, so there is no duplicate announcement to suppress, and removing it would leave the radios themselves undescribed.

Playwright coverage added in `54ff3eb56`, covering all four paths in one card. It was falsified rather than assumed — reconstructing the pre-fix DOM (box loses its `aria-describedby`, ancestor keeps its own), repointing the IDREF at a missing id, and dropping the region's `id` each fail the new assertions.

### `file-upload.tsx` duplicate IDs — valid observation, not this PR, and not file-upload specific

The mechanism checks out. `stacked-cards-container.tsx` mounts `[prevBlockIdx, currentBlockIdx, nextBlockIdx, nextBlockIdx + 1]` simultaneously, and `safePrevIdx = Math.max(0, blockIdxTemp - 1)` can equal `currentBlockIdx` when returning to the first block with an empty visited stack. `stacked-card.tsx` sets only `opacity` and `pointerEvents: none` — no `aria-hidden`, no `inert` — so inactive cards do stay in the accessibility tree.

But this is neither new nor specific to file upload. The collision surface already includes 19 `${inputId}-headline` ids (live `aria-labelledby` targets) and 13 `id={elementId}` DOM ids, all predating this branch. The `-error` ids join an existing pattern rather than creating one, and instance-scoping only file-upload would leave the same ambiguity everywhere else.

The better fix is one of: instance-scoped ids via `useId` across all elements, or marking inactive cards `inert` so they leave the accessibility tree entirely — which would resolve the ambiguity *and* the larger problem that non-current cards are currently reachable by screen reader at all. Both are wider than an ARIA-wiring ticket, so I have logged them as follow-ups rather than expanding this PR.

### One more found while testing

In the **multi-select list** variant the option checkboxes carry `aria-invalid="true"` but no `aria-describedby`, so a failed submit focuses the first checkbox and it announces "invalid" with no reason. Same class of gap, also pre-existing. The fix is arguably to drop `aria-invalid` from the individual checkboxes entirely — the group's selection is invalid, not any one checkbox — but that changes focus behaviour, so it wants its own ticket rather than a late change here.

### On the two pre-merge warnings

The **docstring coverage** one I am skipping deliberately: every changed file is `.tsx`, the repo has no docstring convention for components, and the changed lines are JSX attributes already carrying explanatory comments. Adding docstrings to satisfy a threshold would be noise.

The **screenshot** one is fair — there is real visual surface here (the CTA spacing fix, and `date`/`nps` moving to native `fieldset`). The intent in all three cases is that rendering is unchanged, and that is called out under Risks & regressions. I will attach before/after proof.

---

itsjavi:
Correcting my note above on visual proof: I captured the **after** state for the three cards that have visual surface (CTA, Date, NPS) and will attach them, but I could not produce matching **before** shots — the capture harness failed twice rendering the survey against a rebuild of `main`'s packages, and I stopped rather than keep sinking time into a throwaway script. So what I have is one-sided and I would rather say that than imply otherwise.

Two things that matter more than my screenshots would have:

**Chromatic does not run on pull requests.** `.github/workflows/chromatic.yml` triggers only on `push` to `main` (plus `workflow_dispatch`). All 13 element components in `packages/survey-ui` have stories, so Chromatic will diff every one of them against the baseline — but **after** this merges, not before. If someone wants a pre-merge visual gate, a `workflow_dispatch` run against this branch is the cheap way to get it.

**The intended visual delta is zero.** All three changes are meant to render identically to `main`:

- **CTA** — the always-mounted empty live region became child #1 of a `space-y-2` wrapper and picked up `margin-block-end: 0.5rem`, pushing the button down 8px. Dropping `space-y-2` cancels exactly that, so the net change against `main` is nothing. In the error state the message div's own `.mb-2` is the same `0.5rem` and wins on specificity, so that state is unchanged too. Worth noting the shift only ever manifested for external CTAs, since the button renders only when `buttonExternal` is true.
- **Date / NPS** — `div` to native `fieldset`, with the UA defaults (margin, padding, border, min-inline-size) reset.

Those are the three places to look if anyone does eyeball it against production, and they are already listed under Risks & regressions.

---

itsjavi:
Correcting myself on the `file-upload.tsx` duplicate-id finding above. I said the mechanism "checks out" and declined it as pre-existing and wider than this PR. The decline stands, but my reasoning was wrong in a way worth putting on the record, because I also said I had logged follow-ups off the back of it.

**Duplicate element ids cannot occur.** `packages/surveys/src/components/wrappers/stacked-card.tsx:132` renders `delayedOffset === 0 ? getCardContent(...) : getDummyCardContent()`, and `getDummyCardContent()` (`:56`) returns an empty `<div>`. Only the current card renders element content; the peeking cards are empty placeholders. So while the *card containers* are mounted simultaneously — which is the part I verified and reported accurately — no element ever renders twice, and neither `${inputId}-error` nor the pre-existing `${inputId}-headline` / `id={elementId}` ids can collide.

Confirmed empirically while testing #8776, probing the accessibility tree across the four mounted cards: `[{"id":"questionCard--2","bars":0},{"id":"questionCard--1","bars":1},{"id":"questionCard-0","bars":0},{"id":"questionCard-1","bars":0}]` — one rendered element, three empty.

Two consequences:

- The **`useId` instance-scoping follow-up I proposed is unnecessary.** There is no ambiguity to fix. Please disregard it rather than picking it up.
- The related worry I raised — that inactive cards sit in the accessibility tree because they carry no `aria-hidden` or `inert` — is also close to moot. They contain an empty div, so there is nothing for a screen reader to announce. Adding `inert` would still be tidier, but it is not fixing a live defect.

Net effect on this PR: unchanged, and the finding is a clean no-op rather than a deferred risk.

---

itsjavi:
Picking up the two "outside diff range" findings from CodeRabbit's review ([review 4869928479](https://github.com/formbricks/formbricks/pull/8766#pullrequestreview-4869928479)) — neither got an inline thread, so neither could be replied to or resolved in place. Both checked against the current head, `54ff3eb`.

### 1. `single-select.tsx` — "Associate the list 'Other' input with its error" → already done

Landed in b477d54dc, which came after that review. `SingleSelectListVariant`'s "Other" `Input` now carries `aria-describedby={errorMessage ? \`${inputId}-error\` : undefined}` next to its existing `aria-invalid` (`packages/survey-ui/src/components/elements/single-select.tsx:631`). The requested Playwright regression case landed in 54ff3eb56: it covers all four select paths (single/multi × dropdown/list), asserts each "Other" box is `aria-invalid`, and resolves its `aria-describedby` in-page against the document, so a dangling *or* empty IDREF fails rather than passing on attribute presence alone.

The half not applied is "remove the group-level description when Other is selected". The enclosing `<fieldset role="radiogroup">` keeps its own `aria-describedby` unconditionally (`single-select.tsx:739`), and that does not double-announce: an ancestor's description is not part of a descendant's accessible description, so focusing the "Other" box reads its own name/state/description only. The e2e test encodes exactly this — the select types assert a describing-control count of 2, the group and the box, each legitimately pointing at the region.

### 2. `file-upload.tsx` — "Make the file upload error ID unique or skip duplicate stacked cards" → not changing

The premise doesn't hold. `inputId` is `element.id` (`packages/surveys/src/components/elements/file-upload-element.tsx:353`), so it is scoped per element, not per card. `StackedCardsContainer` renders `[prevBlockIdx, currentBlockIdx, nextBlockIdx, nextBlockIdx + 1]` — distinct block indices, hence distinct elements and distinct ids. Two `FileUpload`s can only share `${inputId}-error` if the same block index occupies two slots at once, and in that case every id in the runtime collides — `elementId`, every element's `inputId`, the radio group `name`s — along with a duplicate React `key` on the cards. That is a container-level concern that predates this PR, not something `${inputId}-error` introduced.

It's also worth noting the finding singles out file-upload while `${inputId}-error` is the same pattern used across all 17 changed files; nothing about this call site is more collision-prone than the others.

Leaving this one open for a human call rather than acting on it.

### Flagging separately: the PR body is now stale on point 1

Under "Two deliberate non-changes, both from review", the description still says the list-variant "Other" inputs keep no `aria-describedby` because a second reference would double-announce. b477d54dc did the opposite, and the code comment there gives the opposite reasoning. Worth correcting so the description matches the diff — I did not edit the body.

---

itsjavi:
Closing out the stale-description item I flagged in [the previous comment](https://github.com/formbricks/formbricks/pull/8766#issuecomment-5236715691) — the PR body is now updated, no code change.

Verified against head `54ff3eb` first: `single-select.tsx:631` does carry `aria-describedby={errorMessage ? \`${inputId}-error\` : undefined}` on the list-variant "Other" input, so the "Two deliberate non-changes" bullet claiming the list variants keep no description was contradicted by b477d54d. That bullet is replaced with what the code actually does, including why the group-level description stays. Two other spots were drifting for the same reason and are corrected: the test-count line now accounts for the fourth announcement test added in 54ff3eb5, and the QA step for the "Other" box now asks for all four combinations (single/multi × dropdown/list) instead of dropdown only.

Nothing else in the body changed, and no threads were resolved.

---

apps/web/playwright/survey-keyboard.spec.ts
```diff
@@ -338,3 +343,611 @@ test.describe("Survey keyboard interaction @slow", () => {
...
+    await page.goto(url);
+    await expect(page.locator("#fbjs")).toBeVisible();
+    // Proves the scoped selector below can actually match live regions inside the
+    // survey, so its zero-count assertion cannot pass vacuously.
+    await expect(page.locator('#fbjs [aria-live="polite"]').first()).toBeAttached();
+
+    // The wrapper used to be aria-live="assertive", so every DOM mutation anywhere
+    // in the survey was announced assertively — which would talk over the polite
+    // error region asserted above.
+    await expect(page.locator('#fbjs [aria-live="assertive"]')).toHaveCount(0);
```
Dhruwang: This runs on the link path (`/s/`), which renders `mode:inline` — the `!isModal` branch never had `aria-live="assertive"` in the first place. The removal you're actually testing lives in the `isModal` branch, so this assertion passes even with the fix reverted. Could we render the modal path here (or assert absence on the modal wrapper directly)? As-is the headline change has no regression guard.

packages/survey-ui/src/components/elements/single-select.tsx
```diff
@@ -313,6 +318,7 @@ function SingleSelectDropdownVariant({
           disabled={disabled}
           aria-required
           aria-invalid={Boolean(errorMessage)}
+          aria-describedby={errorMessage ? `${inputId}-error` : undefined}
```
Dhruwang: Small thing: the description's "deliberate non-changes" says list-variant Other keeps no `aria-describedby`, but it's set right here (and in multi-select). The note reads as stale now — worth correcting so it doesn't mislead a reviewer.

packages/survey-ui/src/components/elements/form-field.tsx
```diff
@@ -98,6 +98,18 @@ function FormField({
   // Get visible fields
   const visibleFields = fields.filter((field) => field.show !== false);
 
+  // This element has no single `inputId` — its inputs are keyed off `elementId` — so the error
+  // region id follows the same `${elementId}-...` scheme the field ids already use.
+  //
+  // The region is deliberately NOT referenced by the individual inputs via aria-describedby. This
```
Dhruwang, line 104: This asymmetry is the subtlest part of the PR and nothing tests it. A contact-info/address case that submits empty and checks the region announces while a sibling field (First name) stays silent would lock in exactly the mislabel this avoids.

packages/survey-ui/src/components/elements/consent.tsx
```diff
@@ -74,7 +76,7 @@ function Consent({
 
       {/* Consent Checkbox */}
       <div className="relative" data-element-input>
-        <ElementError errorMessage={errorMessage} dir={dir} />
+        <ElementError errorMessage={errorMessage} dir={dir} id={errorId} />
```
Dhruwang: consent/cta/file-upload/matrix all got the same wiring but none are seeded in the new specs. Since axe can't see a missing region, those types are effectively unguarded — worth adding them to the announcement suite.

packages/survey-ui/src/components/general/element-error.tsx
```diff
@@ -7,26 +7,42 @@ interface ElementErrorProps {
   errorMessage?: string;
   /** Text direction: 'ltr' (left-to-right), 'rtl' (right-to-left), or 'auto' (auto-detect from content) */
   dir?: "ltr" | "rtl" | "auto";
+  /**
+   * Id placed on the live region so the invalid control can point at it with aria-describedby.
+   * Call sites use the `${inputId}-error` convention.
+   */
+  id?: string;
```
Dhruwang: `id` is optional but every call site now depends on it, and the `${inputId}-error` + invalid/describedby ternary is copied ~14 times with only the e2e suite catching a desync. A tiny `useElementErrorAria(inputId, errorMessage)` returning `{ id, aria-invalid, aria-describedby }` would remove the duplication and the silent-drift risk — and `id` could then be required.

apps/web/playwright/survey-keyboard.spec.ts
itsjavi: Fixed in 89abe66fb. The regression now loads the production survey bundle, renders the same fixture through the modal entry point, proves the modal dialog and a polite live region exist, then asserts that neither the modal wrapper nor its descendants contain an assertive live region.

packages/survey-ui/src/components/elements/single-select.tsx
itsjavi: Confirmed. The PR description now documents that all four Other inputs carry `aria-describedby`, including both list variants, and explains why the group-level association remains.

packages/survey-ui/src/components/elements/form-field.tsx
itsjavi: Added in 89abe66fb. The contact-info fixture submits empty, verifies the shared live region announces the required error, and proves that First name is invalid without pointing at the shared message. Address uses the same `FormField` component and therefore shares this association behavior.

packages/survey-ui/src/components/elements/consent.tsx
itsjavi: Added coverage for consent, file upload, and matrix in 89abe66fb. CTA is intentionally not seeded because it cannot produce a runtime validation error: `CTAElement` forces `required={false}`, passes no `errorMessage`, and both validation paths skip CTA elements. The shared wiring remains covered through the error-capable element types.

packages/survey-ui/src/components/general/element-error.tsx
itsjavi: Implemented in be090666c. `ElementError.id` is now required, and `getElementErrorAria` derives the id, invalid state, and description association together for every caller. Contact and address intentionally consume the derived id and invalid state while omitting the shared description association.

packages/survey-ui/src/components/elements/nps.tsx
```diff
@@ -159,20 +159,33 @@ function NPS({
...
+            untranslated <legend> that omitted the question. The roving-tabindex model lives on the
+            inputs, so the container role does not touch it. */}
+        <fieldset
+          className="w-full px-[2px]"
+          dir={dir}
+          role="radiogroup"
+          aria-labelledby={`${inputId}-headline`}
+          aria-required={required}
+          aria-invalid={Boolean(errorMessage)}
+          aria-describedby={errorMessage ? `${inputId}-error` : undefined}>
```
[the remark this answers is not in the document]
itsjavi: Corrected in the refreshed PR description. The risk section now states that only Date changed element type. NPS is documented as an existing fieldset whose change is limited to the radiogroup semantics and headline-based accessible name.

Dhruwang (approved):

PR #8771, opened 2026-08-06
fix: prevent duplicate otel log export (backport release/5.3)

BhagyaAmarasinghe:
Backport of #8770 for the 5.3 production release line.

## What & why

Production canary validation found that setting `OTEL_EXPORTER_OTLP_ENDPOINT` exported Pino logs through auto-instrumentation even when `OTEL_LOGS_ENABLED=0`. Enabling the dedicated Pino transport then duplicated those logs. Bundled logger instances also accumulated process signal/error handlers and emitted `MaxListenersExceededWarning`.

This keeps Pino trace-context injection while disabling its automatic log sending, leaves log export exclusively to the transport gated by `OTEL_LOGS_ENABLED`, and makes process-handler registration idempotent with rollback after partial registration failure.

## Linear ticket

No Linear ticket; this follows up directly on the production OTEL rollout incident.

## How this was tested

- `pnpm --filter @formbricks/logger exec vitest run src/logger.test.ts` - 16 tests passed
- `pnpm --filter @formbricks/logger typecheck` - passed
- `pnpm --filter @formbricks/logger lint` - passed
- `pnpm --filter @formbricks/web exec vitest run instrumentation-node-config.test.ts` - 10 tests passed
- targeted web coverage - 100% statements/lines/functions and 83.33% branches
- production-isolated canary - traces, metrics, and dedicated logs reached SigNoz with zero restarts and no public Service routing
- production rollout guardrail test - 95/95 public health probes returned 200

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Start with `OTEL_EXPORTER_OTLP_ENDPOINT` set and `OTEL_LOGS_ENABLED=0`, then load the login page -> traces and metrics appear in SigNoz, while Pino transport logs remain disabled.
- [ ] Set `OTEL_LOGS_ENABLED=1` and repeat the request -> each Pino event appears once through the dedicated transport.
- [ ] Exercise health URLs with query strings and `/metrics/*` -> those requests do not create spans; `/metrics-dashboard` remains instrumented.
- [ ] Restart the pod -> it terminates cleanly without `MaxListenersExceededWarning`, Pino worker exits, or missing-module errors.

**Preconditions / test data**

- SigNoz collector reachable over OTLP HTTP.
- Existing `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_LOGS_ENABLED` environment variables.

**Risks & regressions**

- Pino trace-context injection must remain active.
- Logger shutdown handlers must attach once and still flush on process termination.
- OTEL transport startup and shutdown must not affect web readiness.

- none

### DIFF
--- a/apps/web/instrumentation-node-config.test.ts
+++ b/apps/web/instrumentation-node-config.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, test } from "vitest";
+import { nodeAutoInstrumentationConfig } from "./instrumentation-node-config";
+
+describe("nodeAutoInstrumentationConfig", () => {
+  test("leaves Pino log export to the OTEL_LOGS_ENABLED transport", () => {
+    expect(nodeAutoInstrumentationConfig["@opentelemetry/instrumentation-pino"]).toMatchObject({
+      disableLogSending: true,
+    });
+  });
+
+  test.each(["/health", "/metrics", "/metrics/runtime", "/api/v2/health"])(
+    "ignores the %s endpoint",
+    (url) => {
+      const config = nodeAutoInstrumentationConfig["@opentelemetry/instrumentation-http"];
+      expect(config?.ignoreIncomingRequestHook?.({ url } as never)).toBe(true);
+    }
+  );
+
+  test("instruments application routes", () => {
+    const config = nodeAutoInstrumentationConfig["@opentelemetry/instrumentation-http"];
+    expect(config?.ignoreIncomingRequestHook?.({ url: "/auth/login" } as never)).toBe(false);
+  });
+});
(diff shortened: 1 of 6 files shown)


xernobyl (approved):
Approving as a faithful backport of #8770 👍

Checked it rather than rubber-stamping: the six touched files have the same pre-change baseline on `release/5.3` as they do on `main`, the resulting contents are byte-identical to the approved head, and the complete diff comes out identical to #8770's. So nothing unrelated hitched a ride from `main`.

CI is green here. The one non-blocking note I left on #8770 (the `new URL` parse in the `ignoreIncomingRequestHook`) applies equally, but it's a nit — no need to hold the backport for it; whatever lands on `main` can follow later.

PR #8773, opened 2026-08-06
feat(dashboards): add dashboard-level date filter (ENG-1553)

Dhruwang:
## What & why

Dashboards had no way to scope every widget to a shared date range — each chart carried its own saved "Collected at" range, so a viewer couldn't ask "show me the whole board for last 30 days".

This adds a date-range dropdown to the dashboard header that overrides each widget chart's `FeedbackRecords.collectedAt` range for the whole dashboard:

- **Opt-in.** With nothing selected, each chart keeps its own saved range. Selecting a range overrides all widgets; **All time** strips the collectedAt range so widgets show everything.
- **URL-param state** (`?dateRange=…&dateRangeFrom=…&dateRangeTo=…`) — shareable and reload-safe. The server page reads it and applies the override to every widget query before execution, so it works with SSR/streaming and no new client fetching. The custom-range keys are namespaced (`dateRangeFrom` / `dateRangeTo`) so toggling the filter never clears unrelated `from` / `to` params owned by other UI.
- **Per-dashboard persistence (cookie).** The last chosen range is stored in a per-dashboard cookie so it carries across sessions. The server reads the cookie on the first render pass, so a revisit with no URL param renders the widgets filtered directly — no client round trip and no flash of unfiltered data. A pinned URL param always wins and is intentionally **not** written back to the cookie, so opening a shared link never overwrites the viewer's own saved preference.
- **Presets** per the ticket: Last 24 hours (timestamp precision), Last 7 days, Last 30 days, This month, Last month, This quarter, Last quarter, Last 6 months, This year, Last year, plus a Custom range picker and All time.
- Override logic is a pure `applyDashboardDateFilter` helper (overrides an existing collectedAt time dimension in place preserving granularity, or appends a filter-only one; All time strips the range; other time dimensions untouched) with a matching `parseDashboardDateFilter` for the URL. Reuses the existing chart date-preset components and the `expandPresetDateRanges` resolver.
- **Grid alignment.** `containerPadding={[0, 0]}` on the widget grid drops the default container padding so the grid sits flush with the container's left/top edge, aligned with the new date filter above. This applies to every dashboard, not just ones using the filter.

Screenshot 2026-08-06 at 4 50 46 PM

Screenshot 2026-08-06 at 4 50 41 PM

## Linear ticket

https://linear.app/formbricks/issue/ENG-1553/add-dashboard-level-date-filter

## How this was tested

- `pnpm vitest run` on the two affected suites — **45/45 pass** (new `dashboard-date-filter.test.ts` covering the override + URL-parse edge cases incl. impossible/inverted ranges; extended `date-presets.test.ts` for the 4 new presets incl. the 24h UTC timestamp serialization).
- `tsc --noEmit` — clean on all touched files.
- `eslint` — clean on all touched files.
- Translation scan: the 6 new `en-US` keys are detected as used in code; target locales fill in on commit via Lingo.
- Manual UI: not yet run against a live dashboard — see QA plan below.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a dashboard with ≥2 chart widgets that use the "Collected at" time dimension → a **Date range** dropdown shows in the header (below the title), unselected.
- [ ] With nothing selected, each widget renders with its own saved date range (unchanged from before).
- [ ] Pick **Last 30 days** → all widgets reload scoped to the last 30 days; the URL gains `?dateRange=last+30+days`.
- [ ] Copy the URL to a new tab / reload → same range is applied (state survives reload and is shareable).
- [ ] Navigate away and reopen the dashboard **without** any URL param (e.g. via the dashboards list) → the last chosen range is restored from the cookie and the widgets render filtered on first paint (no flash of unfiltered data).
- [ ] Pick **All time** → all widgets show unrestricted data (`?dateRange=all+time`); charts that group by day keep their grouping, just without a range limit.
- [ ] Pick **Custom range** → two date pickers appear; the view only changes once both start and end are chosen; URL gains `dateRange=custom&dateRangeFrom=YYYY-MM-DD&dateRangeTo=YYYY-MM-DD`. Reload → the pickers display the same days you selected (verify on a machine set to a negative-UTC timezone, e.g. `America/Los_Angeles`).
- [ ] Pick **Last 24 hours** → range is scoped to the last 24 hours (timestamp precision, not whole-day); the applied window matches regardless of the server's timezone.
- [ ] Enter dashboard edit mode → the date filter is hidden; exit edit mode → it returns.
- [ ] Edge: hand-edit the URL to an unknown `dateRange` value → filter falls back to unselected (charts use their own ranges), no error.

**Preconditions / test data**

- A workspace on a plan with **Dashboards** enabled (EE / entitlement-gated).
- A dashboard containing at least one chart whose query has a `FeedbackRecords.collectedAt` time dimension, with feedback records spread across several months so ranges visibly differ.

**Risks & regressions**

- Widget query execution path (`executeWidgetQuery` / `expandPresetDateRanges`) — re-check charts that already pin a custom `collectedAt` range still render correctly when no dashboard filter is set.
- Charts with a non-`collectedAt` time dimension (e.g. `createdAt`) must be left untouched by the filter.
- Grid layout: `containerPadding={[0, 0]}` shifts widget alignment on **every** existing dashboard (flush left/top), not only ones using the new filter — eyeball a dashboard that doesn't use the filter.

- none


apps/web/modules/ee/analysis/dashboards/components/dashboard-date-filter.tsx
```diff
@@ -0,0 +1,149 @@
...
+    if (value?.type === "all-time") return ALL_TIME_VALUE;
+    if (value?.type === "preset") return value.value;
+    return "";
+  })();
+
+  // Custom range only produces a filter once both bounds are chosen; a half-picked range keeps the
+  // previous view rather than navigating to an invalid query.
+  const emitCustom = (start: Date | null, end: Date | null) => {
+    if (start && end) {
+      onChange({ type: "custom", range: [format(start, "yyyy-MM-dd"), format(end, "yyyy-MM-dd")] });
```
xernobyl, line 78: **[P2] The custom range shifts back a day for anyone west of UTC.**

This serializes with `format(date, "yyyy-MM-dd")`, which is **local**, but the value is read back at lines 34/37 and 46/47 with `new Date("YYYY-MM-DD")`, which per spec parses as **UTC midnight**. Verified against our own date-fns:

```
TZ=America/Los_Angeles  url=2026-08-01 -> Date=Fri Jul 31 2026 -> re-emitted=2026-07-31  SHIFTED
TZ=America/Los_Angeles  url=2026-08-06 -> Date=Wed Aug 05 2026 -> re-emitted=2026-08-05  SHIFTED
TZ=Europe/Lisbon        url=2026-08-01 -> re-emitted=2026-08-01
```

So after a reload or opening a shared link the pickers *display* a day earlier than the range actually applied (the server still uses the raw URL strings, which are correct). Then touching either picker re-emits the shifted bounds and the window really does move.

Parsing as local — `new Date(y, m - 1, d)`, or `parseISO` — would make it symmetric with the serializer. Won't reproduce for us here on UTC+1, only on negative offsets.

apps/web/modules/ee/analysis/lib/date-presets.ts
```diff
@@ -16,22 +30,53 @@ const PRESET_RESOLVERS: Record<string, (now: Date) => [Date, Date]> = {
...
   if (!query.timeDimensions?.length) return query;
 
   const expanded = query.timeDimensions.map((td) => {
     if (typeof td.dateRange !== "string") return td;
-    const resolver = PRESET_RESOLVERS[td.dateRange.toLowerCase().trim()];
+    const key = td.dateRange.toLowerCase().trim();
+    const resolver = PRESET_RESOLVERS[key];
     if (!resolver) return td;
     const [start, end] = resolver(now);
+    const pattern = TIME_PRECISION_PRESETS.has(key) ? "yyyy-MM-dd'T'HH:mm:ss" : "yyyy-MM-dd";
```
xernobyl: **[P2] `last 24 hours` serializes in server-local time with no offset** — this is CodeRabbit's remaining one, and I think it's right.

The naive pattern means the same instant emits a different string depending on the server's timezone:

```
UTC                    -> ["2026-08-05T12:00:00","2026-08-06T12:00:00"]
America/Los_Angeles    -> ["2026-08-05T05:00:00","2026-08-06T05:00:00"]
Pacific/Auckland       -> ["2026-08-06T00:00:00","2026-08-07T00:00:00"]
```

Nothing sets `timezone` on the query (`cube-client.ts` calls `expandPresetDateRanges(input.query)` bare), so Cube reads these as UTC. Correct on a UTC server, silently skewed anywhere else — and Auckland lands on the wrong calendar day entirely.

Worth flagging that the new test can't catch this: `NOW = new Date(2026, 4, 21, 14, 30, 0)` is a local-time constructor, so `formatDate` returns `14:30` in every timezone and the assertion is TZ-invariant by construction. The date-only presets share the local-formatting pattern, but they're day-granular so it doesn't really bite — it's the sub-day preset that exposes it.

apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx
```diff
@@ -387,8 +402,64 @@ export function DashboardDetailClient({
...
+  // hide the filter and have nothing to apply it to.
+  useEffect(() => {
+    if (isEmpty) return;
+
+    if (searchParams.get(DATE_FILTER_PARAM) !== null) {
+      writeStoredDateFilter(dashboard.id, dateFilter);
+      return;
+    }
+
+    const stored = readStoredDateFilter(dashboard.id);
```
xernobyl: **[P2] Every load with a persisted filter runs all widget queries twice.**

Arriving at `/dashboards/<id>` with no query param, the server renders and executes **every** widget's Cube query unfiltered, then this effect reads localStorage and `router.replace`s, and the server re-renders and executes them **all again** with the filter applied. So doubled Cube and Hub load on each visit, plus a visible flash of unfiltered data before it settles. Gets worse the more widgets a dashboard has.

A cookie rather than localStorage would let the server apply it on the first pass and collapse it back to one round trip.

Two smaller notes while I'm here:

• The localStorage persistence isn't mentioned in the description at all (it reads as URL-param state only) and the QA plan doesn't exercise it — worth adding so it actually gets tested.
• Opening a colleague's shared link **overwrites your own** stored preference for that dashboard, via the `writeStoredDateFilter(dashboard.id, dateFilter)` sync branch. Might well be what you want, just flagging it as a deliberate call rather than an accident.

apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx
```diff
@@ -429,6 +508,10 @@ export function DashboardDetailClient({
                 cols={{ lg: 12, md: 12, sm: 6, xs: 4, xxs: 2 }}
                 rowHeight={ROW_HEIGHT}
                 margin={[16, 16]}
+                // Keep 16px only between items; drop the default container padding (which mirrors
+                // margin) so the grid sits flush with the container's left/top edge, aligned with
+                // the date filter above instead of inset from it.
+                containerPadding={[0, 0]}
```
xernobyl, line 497: **[P3]** `containerPadding={[0, 0]}` is an unrelated grid-layout change riding along with the date filter — it shifts widget alignment on every existing dashboard, not just ones using the new filter. Fine if that's intended (the comment suggests it is), but it isn't mentioned in the description or the QA plan, so it'd be easy to miss in review.

xernobyl (commented):
Had a proper look through this one. Nice shape overall — the override helper being pure with 28 tests behind it makes it easy to reason about, and the SSR wiring is clean. Three things I'd want fixed before it lands, flagged inline, plus a couple of smaller notes :slightly_smiling_face:

A few things I checked specifically that are fine:

• **Presets round-trip through the URL correctly.** This was my main worry given the UI hasn't been exercised manually yet — `URLSearchParams` encodes the spaces as `+`, and every preset value has spaces in it. Next decodes via `searchParams.entries()`, so `dateRange=last+30+days` comes back out as `"last 30 days"`. No issue.
• **The `Select` change** is a no-op in computed CSS but removes a real Tailwind ordering hazard — if `overflow-hidden` happened to sort after `overflow-y-auto` in the generated sheet, the 12-item dropdown couldn't scroll at all. Strictly narrower, so safe for every other Select in the app.
• **`applyDashboardDateFilter`** is genuinely pure and immutable, preserves granularity, appends filter-only dimensions, and leaves non-`collectedAt` dimensions alone.

On the bot review — of CodeRabbit's six, five are already fixed at head (Readonly props, param namespacing, invalid/inverted range rejection, the custom-bounds re-sync effect, and the localized display formatter). Only the `last 24 hours` UTC one is still open, and I think it's valid, so I've picked it up inline rather than leaving it to get lost.

One small thing on the description: it still says `?dateRange=…&from=…&to=…`, and the QA plan says `dateRange=custom&from=…&to=…`, but after the rename the code uses `dateRangeFrom` / `dateRangeTo`. Worth updating so whoever QAs it isn't hunting for the wrong params.

apps/web/modules/ee/analysis/dashboards/components/dashboard-date-filter.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <SelectContent>
+          <SelectItem value={ALL_TIME_VALUE}>
+            {t("workspace.analysis.dashboards.date_filter_all_time")}
+          </SelectItem>
+          {DASHBOARD_DATE_PRESETS.map((preset) => (
+            <SelectItem key={preset} value={preset}>
+              {getTranslatedDatePresetLabel(preset, t)}
+            </SelectItem>
+          ))}
+          <SelectItem value={CUSTOM_VALUE}>{t("workspace.analysis.charts.custom_range")}</SelectItem>
```
xernobyl, line 119: **🟠 should · correctness · confidence: high — traced every onChange call site at this head; none can produce null**
Once a range is picked there is no way back to "nothing selected"

The description opens with the filter being opt-in — *"With nothing selected, each chart keeps its own saved range"* — and `selectValue` has a real unselected state (`return ""` at line 66, with a placeholder on the trigger). But nothing in the UI can get back to it.

`onChange` is only ever called with a value: `{ type: "custom" }` (line 73), `{ type: "all-time" }` (line 85), `{ type: "preset" }` (line 88). The dropdown offers All time, the ten presets and Custom — there is no "None" item, and Radix won't accept an empty-string item value anyway. So `handleDateFilterChange(null)` — which is wired up correctly and does clear both the cookie and the URL params — has no caller.

The cookie change makes this stickier than it was. Before, clearing the URL param got you one unfiltered render. Now a param-less visit falls back to the cookie server-side, so the only way out is devtools:

1. Open a dashboard whose charts have their own saved ranges, pick **Last 30 days**.
2. Decide you wanted the per-chart ranges after all. There's no menu item for it.
3. Strip `?dateRange=…` from the URL and reload — the cookie re-applies it on the server's first pass.
4. It stays that way for 30 days unless the cookie is deleted by hand.

**All time** isn't the escape hatch either: it strips the range from every widget, which is a different result from letting each chart keep its own.

Smallest fix is a "Default (per-chart ranges)" item at the top of the list that calls `onChange(null)` — the whole clearing path underneath it already works. Worth a translated label rather than reusing the placeholder text, since it'd now be a real choice.

What would make me wrong: if the intent is that a dashboard-level range, once chosen, *should* be permanent until changed to another one, then this is by design and only the description's "opt-in" framing needs a tweak.

xernobyl (commented):
**1 should** — re-reviewed at `150289d228`

Re-review after `bec490e5` and `150289d2`. All three of my previous findings are fixed, and well beyond the minimum — verified at this head:

• **Custom-range off-by-one — fixed.** `parseLocalDate` splits the ISO parts and builds a local date, so it's now symmetric with the `format(date, "yyyy-MM-dd")` serializer. No more UTC-midnight drift for negative offsets, and the QA plan now calls out testing it on `America/Los_Angeles`.
• **`last 24 hours` timezone skew — fixed.** `toISOString().slice(0, 19) + "Z"` makes the same instant produce the same string on any server. I checked the format is actually accepted rather than just well-formed: queried our Cube 1.6.6 against real seeded records with a `Z`-suffixed `dateRange` and it returns normally (`200`, same count as the naive and date-only forms over the same window). Also worth calling out that the test was rewritten to construct `NOW` as a UTC instant — the old one was TZ-invariant by construction and could never have caught this, and the new comment says exactly that.
• **Double widget-query on every persisted-filter load — fixed.** The cookie is read server-side in `dashboard-detail-page.tsx` before the widget queries are built, and the client restore effect is gone entirely, so there's no `router.replace` round trip and no flash. The side note about shared links clobbering the viewer's own preference was picked up too, deliberately and with a comment.

The PR description and QA plan were brought back in line as well (namespaced params, cookie persistence, the negative-UTC custom-range case). One thing I think is still worth a look, inline below.

**One nit, not posted inline:** the cookie is per dashboard at `path=/`, so a user who works across many dashboards accumulates one ~150-byte cookie each, all sent on every request to the origin — static assets and `/api/*` included. Scoping the path to the dashboards route, or holding one cookie keyed by dashboard id, would keep it bounded. Not worth blocking on.

**Checked** — correctness, frontend/state, drift (description vs diff), conventions. **Skipped** — tenancy and security: the filter is a presentational override on an already-authorized query, and `getWorkspaceAuth` is untouched. **Not verifiable here** — the UI itself; I can't sign in to the local instance, so the dropdown, the cookie round trip and the custom-range pickers are still on a human. The Cube format check above was done directly against the API.

_Reply per finding with the fix commit, a reason it's wrong, or a ticket if you're deferring it._

apps/web/modules/ee/analysis/dashboards/components/dashboard-date-filter.tsx
Dhruwang: Good catch — you're right, the clearing path had no caller. Added a "Default (per-chart ranges)" item at the top of the dropdown that calls `onChange(null)`, which already clears both the cookie and the URL params. Kept the opt-in framing from the description; used a dedicated translated label rather than the placeholder since it's now a real choice.

Pushed in dcb18df871.

xernobyl (approved):
Approving — everything I raised is addressed, and thoroughly.

Verified at `dbfad8d2`:

- **Clearing the filter** — the new `Default (per-chart ranges)` item calls `onChange(null)`, which was already wired through `handleDateFilterChange` to delete the cookie and strip both URL params. The `DEFAULT_VALUE = "default"` sentinel doesn't collide with anything in `DASHBOARD_DATE_PRESETS`, and it never reaches the URL (the param is deleted rather than set), so a hand-typed `?dateRange=default` still falls through to "no override" — which is the right outcome anyway.
- **Timezone-stable bounds** — `parseLocalDate` makes the custom-range round trip symmetric with the serializer, and `last 24 hours` now emits UTC with the `Z`. I checked our Cube 1.6.6 accepts that format against real records rather than just assuming it parses.
- **First-pass filtering** — cookie read server-side before the widget queries are built, restore effect gone, so no double query and no flash.
- **i18n** — label added to `en-US.json` only, 15 locales generated separately. Correct process.

The only thing I left unraised is a nit: one cookie per dashboard at `path=/` means they're attached to every request to the origin, static assets included. Bounded and cheap enough that it isn't worth another round — happy to leave it or file a follow-up, your call.

CI is green across the board. Nice work on this one :slightly_smiling_face:

jobenjada (commented):
image

UI / UX looks great ✅

PR #8774, opened 2026-08-06
feat(analytics): add PostHog AI and MCP observability

harshsbhat:
## What & why

None of the 4 in-repo AI/LLM features (survey generation, AI chart queries, translation, example-response generation) or the MCP server had any observability into token usage, cost, latency, or failures — only one emitted a basic usage-count event. This adds PostHog's AI Observability and MCP Analytics products, reusing the existing `lib/posthog` wrapper conventions (null-safe on `POSTHOG_KEY`, snake_case events, `{organizationId, workspaceId}` group context).

- `packages/ai` gains a generic `wrapModel` hook on `generateText`/`generateObject` so it can wrap the resolved AI SDK model without knowing anything about PostHog (dependency injection, matching the existing layering).
- `lib/posthog/ai-tracing.ts` wraps AI SDK models via `@posthog/ai`, tagging each generation with a typed `AI_TRACING_FEATURE` (no more free-text feature strings across 4 call sites). Privacy mode is always on — metadata only (model, tokens, cost, latency), never prompt/completion content.
- `lib/posthog/mcp-tracing.ts` instruments the MCP server via `@posthog/mcp`, stripping tool call parameters/responses in `beforeSend` for the same privacy guarantee.
- Both wrappers have a try/catch around the actual SDK call — a broken/incompatible analytics SDK must never fail a real AI generation or MCP tool call.
- `lib/posthog/server.ts` now exports two PostHog clients: the existing immediate-flush client for low-volume product events, and a new batched client (`flushAt: 20`, `10s` interval) for the much higher-volume AI/MCP tracing events, so tracing can't flood PostHog with unbatched requests.
- MCP's `identify()` resolves `distinctId`/org group from API-key or OAuth auth. OAuth sessions intentionally get no group — `@posthog/mcp` awaits `identify()` *before* the real tool handler runs (confirmed by reading its source), so this path has to stay free of DB/network calls; API keys get a group for free since `organizationId` is already resolved.
- Fixed a real circular-import crash along the way: `lib/posthog` is a low-level, widely-imported barrel, and an earlier version of `mcp-tracing.ts` imported `@/modules/mcp/auth` (which eagerly constructs an OAuth resource client at module scope), creating a cycle back through `lib/organization/service.ts` → billing → `lib/posthog`. Fixed via dependency injection (`identify` passed in from `modules/mcp/server.ts` instead).

Out of scope: the external `@formbricks/hub` microservice's internal LLM calls (embeddings, topic clustering) can't be instrumented from `apps/web`.

## Linear ticket

No ticket — internal observability improvement, not user-facing.

## How this was tested

- `pnpm --filter=@formbricks/ai test` and `pnpm --filter=@formbricks/web test` (full suite, not just touched files): **7633/7633 passing**.
- `pnpm typecheck` and `pnpm lint`: clean on every touched file.
- **Verified end-to-end against a live PostHog project**, not just mocks:
  - Fired a real MCP `list_surveys` tool call (real API key, real DB data) against a local dev server → confirmed a `$mcp_tool_call` event landed with correct `distinct_id` (`apiKey:<id>`), tool name, duration, zero errors, correct organization group, and — critically — `$mcp_parameters`/`$mcp_response` both `null`, confirming the privacy-mode content stripping works.
  - Fired a real Gemini call through the exact production `generateOrganizationAIObject` → `wrapAiModelWithTracing` path → confirmed a `$ai_generation` event landed with correct model/provider/feature/trace ID/groups, and `$ai_input`/`$ai_output_choices` both `null`, confirming privacy mode redacts prompt/completion content.
- Along the way, caught and fixed a real regression via the full-suite run: `@posthog/mcp`'s `instrument()` awaits `identify()` *before* the real tool handler executes (read its bundled source to confirm), which surfaced as 2 failing tests in `app/api/mcp/route.test.ts`. Fixed by draining the response before asserting (matching every other test in that file) and by removing a DB lookup I'd added to `identify()`, since that function is now confirmed to sit on the critical path of every tool call.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] With `POSTHOG_KEY` set, trigger AI survey generation (`POST /api/v3/surveys/generate` or the "Create with AI" UI flow) → a `$ai_generation` event should appear in PostHog with `properties.feature = "ai_survey_generation"`, redacted `$ai_input`/`$ai_output_choices`, and correct token/cost/latency fields.
- [ ] Call any MCP tool (e.g. `list_surveys`) via an authenticated MCP client against `/api/mcp` → a `$mcp_tool_call` event should appear with the correct tool name, duration, and no `$mcp_parameters`/`$mcp_response` (privacy mode).
- [ ] With `POSTHOG_KEY` unset, repeat both flows → both features work identically, no errors, and no events are sent (no-op path).
- [ ] Trigger AI translation, AI chart query, and AI example-response generation → each should tag its `$ai_generation` event with its own `feature` value (`ai_translation`, `ai_chart_query`, `ai_example_responses`); the chunked example-response flow should group all its chunk calls under one shared `$ai_trace`.
- Edge case: force an AI generation or MCP tool call to run while PostHog is unreachable/misconfigured → the real feature (survey generation, tool call, etc.) must still succeed; only a `logger.warn` should appear.

**Preconditions / test data**

- `POSTHOG_KEY` set to a valid project token to see events land; unset to verify the no-op path.
- An AI provider configured (`AI_PROVIDER`/`AI_MODEL` + credentials) to exercise the AI generation paths.
- A valid API key or OAuth session for MCP tool calls.

**Risks & regressions**

- Any change to `packages/ai`'s `generateText`/`generateObject` signatures (now take an optional 3rd `wrapModel` arg) — re-check any other consumers of these functions if added later.
- MCP tool-call latency: `@posthog/mcp`'s `identify()` runs before every tool executes: keep `identifyMcpUser` free of DB/network calls if it's ever extended.

- none (reuses the existing `POSTHOG_KEY` env var as the sole on/off switch; no new env vars)


apps/web/lib/posthog/mcp-tracing.ts
```diff
@@ -0,0 +1,53 @@
+import "server-only";
+import { PostHogMCPAnalyticsProperty, instrument } from "@posthog/mcp";
+import type { MCPAnalyticsOptions, McpAnalytics } from "@posthog/mcp";
+import { logger } from "@formbricks/logger";
+import { posthogTracingClient } from "./server";
+
+const CONTENT_PROPERTIES: string[] = [
```
xernobyl: **[P2] Denylist here should be an allowlist — and the test currently verifies it against a mock of itself.**

`CONTENT_PROPERTIES` deletes exactly two known keys, and `mcp-tracing.test.ts` mocks `PostHogMCPAnalyticsProperty` with exactly those two members. So neither the code nor the test would notice a third content-bearing property in the real SDK, now or after a version bump. The live check against a real project is stronger evidence, but it only proves those two keys are null for the one tool that was called (`list_surveys`).

Why this matters more than a usual privacy nicety: the feedback-record tools carry **end-user content**, not survey config. `get_feedback_record`, `list_feedback_records` and `search_feedback_records` all return `value_text` — the respondent's open-text answer — and `create_feedback_record` takes it as an *argument*. So if the strip list ever drifts from the SDK, what escapes isn't metadata, it's customer feedback going to a third party.

Suggestion, same effort and it fails safe: keep an allowlist of the metadata you actually want (tool name, duration, error, client info) and drop everything else, so a new SDK property defaults to *not sent* rather than sent. And assert against the real enum's full member list rather than a two-member mock — otherwise the test can only ever confirm the assumption it was written from.

Happy to be told this is over-cautious for a metadata-only product, but the feedback-records payloads are the one place on this surface where the downside is customer data rather than noise.

xernobyl (commented):
Reviewing this specifically against the feedback-record MCP tools (#8650, #8717), since those are the ones whose payloads carry customer content.

**Verdict: needed, and the wiring is good. One ask before it lands — see the inline note on the strip list.**

**No conflict with the feedback-record tools.** `instrument(server, …)` wraps above tool registration, so it doesn't touch `readOnlyHandler`/`writeHandler`, the `guardMcpScopes` gates, or the audit-log lifecycle. It's also orthogonal to ENG-2119: if those tools converge onto `registerScopedTool`, the instrumentation sits above that and is unaffected. `identifyMcpUser` is genuinely cheap, and `distinctId: apiKey:<apiKeyId>` is the ApiKey row id rather than the secret — worth saying explicitly since that's the obvious place to get it wrong. The circular-import diagnosis and the "no DB/network here, ever" comment on the `identify` critical path are both the kind of thing that saves the next person an afternoon.

**Two smaller notes, neither blocking:**

- `beforeSend` runs inside PostHog's send pipeline, *outside* the `try/catch` that guards `instrument()`. If `event.properties` were ever undefined, the `delete` throws there. Worst case is lost analytics rather than a failed tool call, so it's minor — but the defensiveness the rest of this file is careful about stops at that boundary.
- Worth knowing rather than changing: privacy mode strips `$mcp_parameters`, which is exactly the signal that would reveal agents sending *misnamed* arguments. That's the failure mode in ENG-2256 (the MCP SDK silently drops unknown keys, so a wrong filter name returns a confidently wrong answer). Not an argument against stripping — just don't expect tool-call analytics to surface that class of bug.

**On whether it's needed:** the AI half is an easy yes — four LLM features with no token, cost or latency visibility is a cost-control gap. The MCP half is a yes too, and the overlap with what we already have is smaller than it looks: audit logs are mutations only (and were silently dropping every event until #8743), `logger` is per-request, Sentry is errors. Aggregate per-tool and per-client usage and latency isn't available anywhere else today. The price is a third-party SDK awaited on every tool call, which you've documented rather than buried.

harshsbhat:
Addressed the [P2] allowlist ask in b868a5af5.

`beforeSend` now allowlists metadata (tool name, duration, error state/type, client/server/protocol info, session id) instead of denylisting two known keys — anything the SDK doesn't explicitly need is dropped by default, including a future version's additions.

Two things worth flagging from the implementation:

- `$mcp_error_message` (the raw exception text) is now also stripped — it wasn't on the original denylist, but it's genuinely content-bearing (can echo back invalid input), so it belongs on the deny side under an allowlist philosophy. This also closes an earlier open item from my own review.
- `$groups`, `$process_person_profile`, and `$set` aren't part of `PostHogMCPAnalyticsProperty` at all — I checked the SDK's bundled source (`sink.js`) and confirmed `$groups` is read straight off `properties.$groups` *after* `beforeSend` runs, so an allowlist built only from the enum would have silently broken organization/workspace grouping (which I'd already verified working end-to-end against a live project). Explicitly preserved all three.

Test now imports the real, installed `PostHogMCPAnalyticsProperty` enum (mocking only `instrument`) and iterates every member with a distinct sentinel value, asserting the allow/deny split per-property — so it can't just confirm the assumption it was written from, per your note.

The two smaller non-blocking notes (properties possibly undefined, and privacy mode masking the ENG-2256 misnamed-argument signal) — added a one-line guard for the former since it was free; the latter is a real, documented tradeoff of the privacy-mode-by-default decision, not something to change here.

apps/web/lib/posthog/ai-tracing.ts
```diff
@@ -0,0 +1,55 @@
...
+): AIResolvedLanguageModel {
+  if (!posthogTracingClient) return model;
+
+  try {
+    return withTracing(model, posthogTracingClient, {
+      posthogDistinctId: context.distinctId,
+      posthogTraceId: context.traceId,
+      posthogGroups: buildGroups(context),
+      posthogProperties: { feature: context.feature },
+      posthogPrivacyMode: true,
```
xernobyl, line 41: **🔴 blocker · security** · confidence: high
posthogPrivacyMode does not cover $ai_error, so a provider error ships the full prompt to PostHog

`posthogPrivacyMode: true` gates exactly two properties in `@posthog/ai@8.5.0`'s Vercel wrapper. `$ai_error` is not one of them:

```js
// node_modules/@posthog/ai/dist/vercel/index.cjs
$ai_output_choices: withPrivacyMode(client, privacyMode, safeOutput),   // gated
$ai_error: stringifyError(options.error),                               // NOT gated
```

`stringifyError` is `JSON.stringify(sanitizeValues(serializeError(error)))`. `serializeError` copies **every own enumerable key** off the Error (`for (const key of Object.keys(value)) out[key] = serializeError(value[key], depth - 1)`), and `sanitizeValues` only does a JSON round-trip plus lone-surrogate cleanup — there is no redaction anywhere in that path. The AI SDK's `APICallError` assigns `requestBodyValues` (the entire provider request body, i.e. the prompt) and `responseBody` as own enumerable instance fields (`@ai-sdk/provider/dist/index.js:105`).

**Failure scenario.** Any provider 4xx/5xx on any of the four wrapped features. Running the shipped `@posthog/ai@8.5.0` `stringifyError` against a real `@ai-sdk/provider@3.0.0` `APICallError` produces:

```
{"name":"AI_APICallError","message":"Invalid request","stack":"…",
 "url":"https://generativelanguage.googleapis.com/…:generateContent",
 "requestBodyValues":{"contents":[{"role":"user","parts":[{"text":"Translate: \"How did agent Jane handle your refund for order 55123?\""}]}], …},
 "statusCode":400,"responseBody":"{\"error\":{\"message\":\"payload rejected\"}}"}
```

The prompt text is present verbatim. That contradicts the guarantee this PR states in two places — the body ("metadata only … never prompt/completion content") and the docstring at `ai-tracing.ts:34-38` — and a Gemini 400 on a long survey-generation prompt is a routine event, not a corner case. No credential exposure: provider keys travel in headers, and `url` comes from operator config.

**Fix.** `posthogTracingClient` is now dedicated to tracing, so a `beforeSend` on the client itself in `lib/posthog/server.ts` is a single choke point that covers this and the MCP side together — posthog-node v5 supports it. Prune `$ai_error` there (keeping `$ai_is_error` for the failure-rate metric) rather than relying on `posthogPrivacyMode`.

**What would make me wrong:** if `@posthog/ai` gained privacy-gating for `$ai_error` in a version this PR does not pin — it pins `8.5.0` in `apps/web/package.json`, and I read that exact tarball. Also worth noting this is unmerged code, so nothing is deployed; happy to move it off a public thread if you'd rather.

apps/web/lib/posthog/mcp-tracing.ts
```diff
@@ -0,0 +1,88 @@
...
+  PostHogMCPAnalyticsProperty.ListedToolNames,
+  PostHogMCPAnalyticsProperty.DurationMs,
+  PostHogMCPAnalyticsProperty.ServerName,
+  PostHogMCPAnalyticsProperty.ServerVersion,
+  PostHogMCPAnalyticsProperty.ClientName,
+  PostHogMCPAnalyticsProperty.ClientVersion,
+  PostHogMCPAnalyticsProperty.ProtocolVersion,
+  PostHogMCPAnalyticsProperty.IsError,
+  PostHogMCPAnalyticsProperty.ErrorType,
+]);
```
xernobyl, line 41: **🟠 should · correctness** · confidence: high
The allowlist strips $exception_list, so every failed tool call sends an empty $exception event

The allowlist closes the `$mcp_error_message` leak, which is the right call. But `@posthog/mcp` also emits a **sibling `$exception` event** for every errored tool call — `enableExceptionAutocapture` defaults to `true` (`dist/types.d.ts:52-56`) and isn't disabled at `mcp-tracing.ts:78-83` — and that event's payload lives in properties the allowlist does not carry.

Evidence, in order:
- `buildPostHogCaptureEvents` pushes `buildExceptionEvent(event)` when `isError && error` (`dist/extensions/posthog-events.js`).
- `buildExceptionEvent` puts the payload in via `if (event.error) Object.assign(properties, event.error)`.
- `event.error` is `ErrorProperties` = `{ $exception_list, $exception_level? }` (`@posthog/core/dist/error-tracking/types.d.ts:18-21`).
- Neither key is in `ALLOWED_PROPERTIES`, so `beforeSend` deletes both.
- `applyBeforeSend` only drops an event when the callback returns nullish or throws; this one returns the event, so it is still sent (`dist/extensions/sink.js`).

**Failure scenario.** Any MCP tool returning `isError: true` — which is the normal path, since `responseToMcpToolResult` (`apps/web/modules/mcp/errors.ts`) returns `isError: true` for every non-2xx v3 response. PostHog then receives an `$exception` event carrying session id, groups, tool name and client info, and no exception payload at all: one ingested, billed event per failed tool call with zero diagnostic value, plus malformed input to error tracking.

**Why the new test doesn't catch it.** The test at `mcp-tracing.test.ts:87` iterates `Object.values(PostHogMCPAnalyticsProperty)`, and its comment reads "every property the real, installed @posthog/mcp SDK can ever write". `$exception_list` is not in that enum — it comes from `@posthog/core` — so the loop cannot reach it, and the only event shape constructed is `$mcp_tool_call`.

**Fix.** Add `enableExceptionAutocapture: false` alongside `context: false` at `mcp-tracing.ts:81`. MCP errors already go to Sentry, and `$mcp_is_error` + `$mcp_error_type` are both allowlisted, so the dashboard keeps its failure-rate and failure-reason breakdown. That also removes a second problem I'd flagged: `captureException`'s `extractCallToolResultMessage` folds the whole `JSON.stringify({error:{…,requestId,…}})` blob into the exception value, so the per-request `requestId` would have given every failure a unique fingerprint and made grouping useless. If you'd rather keep the exception stream, the alternative is allowlisting `$exception_list` and scrubbing its `value` — strictly more work for the same privacy position.

**What would make me wrong:** if PostHog silently discards `$exception` events lacking `$exception_list` server-side, this is dead weight rather than junk data — still worth the one-line disable, but a nit instead of a should.

apps/web/lib/posthog/index.ts
```diff
@@ -4,3 +4,6 @@ export { capturePostHogEvent, groupIdentifyPostHog, identifyPostHogPerson, getEm
 export type { PostHogGroupContext } from "./capture";
 export { getPostHogFeatureFlag } from "./get-feature-flag";
 export type { TPostHogFeatureFlagContext, TPostHogFeatureFlagValue } from "./types";
+export { wrapAiModelWithTracing, AI_TRACING_FEATURE } from "./ai-tracing";
+export type { AITracingContext, AITracingFeature } from "./ai-tracing";
+export { instrumentMcpServerWithTracing } from "./mcp-tracing";
```
xernobyl: **🟠 should · simplicity** · confidence: high
Re-exporting the tracing modules from the barrel pulls @posthog/ai and @posthog/mcp into 50 unrelated modules

50 modules import `@/lib/posthog`. Most want `capturePostHogEvent` or `getPostHogFeatureFlag`; four want `AI_TRACING_FEATURE`, a plain const object. After these three lines, all 50 carry `@posthog/ai` (2.2 MB unpacked) plus `@posthog/mcp` (912 KB) and the MCP SDK.

This will not tree-shake: `apps/web/package.json` declares no `sideEffects`, `mcp-tracing.ts:1` opens with `import "server-only"`, and `@posthog/mcp` does a runtime `require("@modelcontextprotocol/sdk/types.js")` (`dist/extensions/instrument-lowlevel.js:29`). Among the 50 is `app/api/v1/client/[workspaceId]/environment/lib/environmentState.ts:11` — the widget's environment-state endpoint, hit by every app survey — which imports the barrel purely for `capturePostHogEvent`.

**The sharper cost is the landmine, not the bytes.** This PR already had to diagnose and fix a circular-import crash *because* `lib/posthog` is a low-level, widely-imported barrel; the docstring at `mcp-tracing.ts:65-69` documents it. Routing the two heaviest new modules back through that same barrel re-arms it: the next top-level side effect added to either one, or a re-added `@/modules/...` import, breaks all 50 importers at module-init rather than at the four call sites that actually use tracing.

**Fix.** Keep both out of the barrel and deep-import at the four call sites (`@/lib/posthog/ai-tracing`, `@/lib/posthog/mcp-tracing`) — `service.ts`, `translate-fields.ts`, `ai-chart-query.server.ts`, `example-responses.ts` and `modules/mcp/server.ts`. Better still, move `AI_TRACING_FEATURE` into a dependency-free constants module, since four of those five want only the const and none of the SDK.

**What would make me wrong:** if Next.js's server bundling drops the unreferenced re-exports despite the missing `sideEffects` flag and the runtime `require`. Cheap to settle — compare `.next/standalone` size, or `NEXT_TURBOPACK_TRACING=1` on the environment route, with and without lines 7-9.

xernobyl (commented):
**1 blocker · 2 should** — reviewed at `77fef119c1`

Re-review at `77fef119c172eee1b38ac8f87fc647034bbc697d` (previous pass was `8221bca`). The two new commits change only `mcp-tracing.ts` and its test: the denylist became an allowlist, `event.properties` is now null-guarded, and the test drops its two-member mock in favour of the real installed enum via `importOriginal`. All 18 checks green; Sonar gate passed.

**The allowlist ask is fixed, and correctly.** I tried to break the `$groups` reasoning in the new comment at `mcp-tracing.ts:7-13` and couldn't — `@posthog/mcp`'s sink really does read `captureEvent.properties.$groups` *after* `beforeSend` runs (`dist/extensions/sink.js`, in `McpEventSink.capture`), so allowlisting it is load-bearing and org/workspace grouping would have silently broken without it. `$process_person_profile` and `$set` are likewise written outside the SDK's own enum (`buildCaptureEvent` in `dist/extensions/posthog-events.js`). The new test is a real test: it iterates every `PostHogMCPAnalyticsProperty` member with a distinct sentinel and asserts allow/deny per property, so a new SDK property defaults to stripped rather than sent.

One consequence of the allowlist wasn't accounted for, and the new test can't see it because the property doesn't come from that enum — see the `$exception_list` finding. The two findings below it are unchanged from `8221bca` and were never posted; both files are byte-identical at this head.

**Also worth a look, not anchored:** `identifyMcpUser` returns `distinctId: "apiKey:<id>"` (`modules/mcp/server.ts:39`), which sets `identifyActorGivenId`, so `$process_person_profile` stays on and PostHog creates a billed *person* per API key at `$mcp_tool_call` volume. Held back as a nit — worth a deliberate call rather than a side effect, but not merge-blocking.

**Checked** — correctness, tests, drift, security, simplicity. **Skipped** — tenancy (the delta touches only the PostHog beforeSend filter — no queries, no auth paths); frontend (no .tsx, packages/surveys or packages/survey-ui in the diff); conventions (no cache keys, Prisma pagination, action-result shapes or i18n in the delta).

**Not verifiable from a static review**
- Could not run `mcp-tracing.test.ts` locally: `@posthog/mcp` is not resolvable in this checkout, and installing it would mutate a working tree currently sitting on an unrelated branch (`pr-8705`). CI's Run Unit Tests passed at this head, which covers it.
- The claim that PostHog's error-tracking ingestion needs `$exception_list` to form an issue is from PostHog's documented event contract, not something I exercised against a live project — the stripping itself is verified in code.

<sub>Opus 5, high reasoning effort. Findings verified against the files at `77fef119c1` and against the shipped `@posthog/ai@8.5.0` / `@posthog/mcp@0.10.2` / `@posthog/core@1.46.1` tarballs.</sub>

harshsbhat:
All three fixed in 6501deaa5, each verified against the installed package source before touching code.

**🔴 blocker — `$ai_error` bypasses privacy mode.** Confirmed exactly as described in `@posthog/ai@8.5.0`'s bundled `vercel/index.cjs`: `$ai_output_choices`/`$ai_input` go through `withPrivacyMode`, `$ai_error: stringifyError(options.error)` does not. `stringifyError` → `serializeError` copies every own-enumerable key off the Error with no redaction, and the AI SDK's `APICallError` puts `requestBodyValues` (the full prompt) there as an instance field. This is a real gap in the upstream SDK, not something fixable from our side except by intercepting after the fact — which your suggested fix (a `before_send` on the client) does cleanly. Implemented in `lib/posthog/server.ts`, scoped to `posthogTracingClient` only, stripping `$ai_error` while leaving `$ai_is_error` (boolean, no content) untouched. Confirmed correct: this is unmerged code on an open PR, nothing was ever deployed — appreciate you flagging that explicitly rather than assuming.

**🟠 allowlist strips `$mcp_error_message` but not `$exception_list`, so failures send an empty `$exception` event.** Verified: `buildExceptionEvent` puts `event.error` (`{$exception_list, ...}`) into that event's properties via `Object.assign`, which our tool-call allowlist never touches since it's a structurally separate `$exception` event, not `$mcp_tool_call`. Rather than extending the allowlist to also cover the exception event's shape, went with your suggested `enableExceptionAutocapture: false` — MCP errors already go to Sentry, and `$mcp_is_error`/`$mcp_error_type` stay allowlisted so the failure-rate/reason breakdown survives on the PostHog side without a second, redundant error-reporting pipe.

**🟠 barrel re-export pulls both SDKs into ~50 unrelated modules.** Agreed on both the mechanism (no `sideEffects: false`, so no tree-shaking guarantee) and the sharper point — re-arming the exact landmine this PR already got bitten by once (the circular-import crash). Removed `wrapAiModelWithTracing`/`instrumentMcpServerWithTracing`/`AI_TRACING_FEATURE` from `lib/posthog/index.ts` entirely. `AI_TRACING_FEATURE` now lives in a new dependency-free `lib/posthog/ai-tracing-feature.ts` (four of the five callers only ever wanted the enum, not `@posthog/ai`). The two SDK-wrapping modules are deep-imported directly at their one real call site each (`lib/ai/service.ts` for AI, `modules/mcp/server.ts` for MCP) — both of which need the full SDK anyway, so this doesn't shrink their footprint, it just stops it leaking into the other ~48.

Also updated/added test coverage for all three: `server.test.ts` now asserts the tracing client is constructed with `before_send` and exercises it directly (strips `$ai_error`, keeps `$ai_is_error`, no-ops on missing properties); `mcp-tracing.test.ts` asserts `enableExceptionAutocapture: false` is passed. Full suite: 7653/7653 passing, typecheck and lint clean.

apps/web/lib/posthog/ai-tracing.ts
xernobyl: Verified fixed at `7b0fbd1`, and not just by reading the diff — the failure mode here was going to be a silently-ignored option, so I ran it.

`posthog-node@5.28.11` takes `before_send` (snake_case) and redefines `BeforeSendFn` over its own `EventMessage`; a camelCase guess would have type-checked against the wrong export and done nothing. Both are right. Instantiating the real client with your exact `stripAiError` and capturing the payload `@posthog/ai` emits on a failed generation:

```
before_send invoked: true
$ai_error present after hook: false
$ai_is_error preserved: true          $ai_input_tokens preserved: 120
LEAKS prompt text: false
```

So the hook is on the `capture()` path, the prompt is gone, and the failure-rate and token metrics survive.

I also closed the gap in my own report: I had flagged `$ai_error` without enumerating the rest. Every `$ai_*` property the pinned 8.5.0 vercel wrapper emits — `$ai_input`/`$ai_output_choices` gated, `$ai_error` now stripped, and the remainder is tokens, costs, latency, model/provider ids, plus `$ai_model_parameters`/`$ai_tools`/`$ai_provider_metadata`, which for the way we call these models carry call settings and safety ratings rather than prompt or completion text. Nothing else leaks.

One thing worth naming in the docstring: this is a denylist, so an unknown property defaults to *sent* — the opposite of the MCP allowlist two files over. Safe only because `@posthog/ai` is pinned at `8.5.0`. See the approval for why that matters at bump time.

apps/web/lib/posthog/mcp-tracing.ts
xernobyl: Verified fixed at `7b0fbd1`. `enableExceptionAutocapture: false` propagates correctly — the SDK reads `data.options.enableExceptionAutocapture ?? true` in `capture.js`, so an explicit `false` survives the nullish coalesce, and `buildPostHogCaptureEvents` then skips `buildExceptionEvent` entirely. No more empty `$exception` events, and `$mcp_is_error`/`$mcp_error_type` stay allowlisted so the failure-rate and reason breakdown is unaffected.

I re-checked the assumption the advice rested on, since `#8725` (move Sentry client init to instrumentation-client) came in via one of the main merges in between: it only moved the *client* init — `sentry.server.config.ts` and `instrumentation.ts` are intact — so server-side MCP errors still reach Sentry and the justification holds.

apps/web/lib/posthog/index.ts
xernobyl: Verified fixed at `7b0fbd1`. The resulting import graph is exactly the target: `@posthog/ai` is now reachable only from `lib/ai/service.ts`, `@posthog/mcp` only from `modules/mcp/server.ts`, and the ~50 modules that import the barrel for `capturePostHogEvent` carry neither.

Splitting `AI_TRACING_FEATURE` into a dependency-free `ai-tracing-feature.ts` is a better shape than what I suggested — four of the five call sites only ever wanted the enum, and they now get it without `server-only` or an SDK anywhere in their graph. Leaving the comment at `index.ts:8-14` explaining why the re-exports are deliberately absent is what stops someone re-adding them in six months.

(Grepping for importers turns up `index.ts` twice, but those are the quoted paths inside that comment, not imports — the barrel is clean.)

xernobyl (approved):
Approving at `7b0fbd1`. All four findings across the two rounds — the allowlist ask plus the three from the re-review — are closed, and I verified each against the code and the SDK plumbing rather than going off the resolved threads. Details are in the replies on each thread; the short version is that the `$ai_error` strip was the one that could plausibly have been fake (a camelCase `beforeSend` would have type-checked against the wrong export and silently done nothing), so I ran the real `posthog-node` client with your hook and watched the prompt disappear while `$ai_is_error` and the token counts survived.

Two things to carry forward. Neither is worth holding the PR for, but neither should be silent either.

**The two filters now have opposite defaults.** MCP is a strict allowlist — an unknown property is stripped. The AI side is a one-property denylist — an unknown property is sent. That's safe today, and I checked it properly rather than assuming: every `$ai_*` property the pinned `8.5.0` build emits is either privacy-gated, now-stripped, or metadata (tokens, costs, latency, model/provider ids, call settings, safety ratings). The exposure is the version bump. A Dependabot PR for `@posthog/ai` will look routine and carries no signal that it needs a privacy re-audit, and the person reviewing it won't have this thread in their head. Worth pinning the version in the `stripAiError` docstring, or a follow-up ticket to re-audit on bump — your pick.

**The live check predates the fixes.** The end-to-end PostHog verification in the description was run against the original code, and the last three rounds changed exactly the path it exercised. The hook is unit-tested and I proved it in isolation, so this isn't an approval condition — but forcing one provider 4xx and confirming `$ai_error` is absent on the real event is a couple of minutes and turns "proven in isolation" into "proven in the deployed shape." Worth doing before merge.

Two smaller things: `identifyMcpUser` returning `distinctId: "apiKey:<id>"` sets `identifyActorGivenId`, so PostHog creates a billed *person* per API key at `$mcp_tool_call` volume — a deliberate call worth making rather than inheriting. And the description still says "No ticket" while GTM-16 is open; linking it would satisfy the one-PR-one-ticket convention.

Given this touches `generateText`/`generateObject` signatures in `packages/ai`, the shared PostHog client, and the import graph for ~50 modules, a second approver from the observability side is worth having before merge — that's about blast radius, not about the change.

Nice work on the diagnosis quality throughout, particularly catching that the SDK's sink reads `$groups` off `event.properties` *after* `beforeSend` runs. That one would have silently broken org grouping with nothing failing anywhere.

Dhruwang (approved):

PR #8775, opened 2026-08-06
fix(helm): support graceful container shutdown

BhagyaAmarasinghe:
## What & why

Expose app-container lifecycle hooks and the Pod termination grace period through the Formbricks Helm chart. During an EKS production rolling update, a terminating Pod exited before the AWS target group completed deregistration, producing a transient 502 even though Kubernetes kept all desired replicas available.

This is a generic chart capability; no lifecycle hook is enabled by default. Environments can configure a preStop drain appropriate for their load balancer. The chart documents that the Pod grace budget must exceed the preStop drain because lifecycle hooks run inside that budget.

## Linear ticket

No Linear ticket; this is a production rollout follow-up.

## How this was tested

- Default Deployment render: 30-second grace period and no lifecycle block ✅
- Custom Deployment render: 60-second grace period and preStop exec command ✅
- Negative lifecycle assertion regression check: forbidden default lifecycle block fails as intended ✅
- PR-scope Prettier check after merging current `main` ✅
- `kubectl apply --dry-run=client` against the rendered Deployment ✅
- `git diff --check` ✅

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Deploy the chart with a preStop drain shorter than `deployment.terminationGracePeriodSeconds`, then trigger a rolling update → the old Pod remains available during load-balancer deregistration and the public app endpoint does not return transient 502 responses.
- [ ] Deploy with the default `deployment.lifecycle: {}` → the rollout completes with the existing 30-second Pod termination grace period and no preStop behavior.
- [ ] Supply an invalid lifecycle map → the deployment is rejected during manifest validation rather than silently applying an invalid hook.

**Preconditions / test data**

- A self-hosted Kubernetes staging environment behind a load balancer.
- Continuous requests to the public app or health endpoint during the rollout.
- For the configured case, a preStop drain shorter than the total Pod termination grace period.

**Risks & regressions**

- A preStop drain equal to or longer than the Pod grace period leaves insufficient time for the app process to shut down.
- Lifecycle configuration applies only to the main Formbricks app container in this PR.

- none


.github/workflows/helm-chart-validation.yml
```diff
@@ -37,6 +37,22 @@ jobs:
 
           helm lint "$chart" --set formbricks.webappUrl=https://qa.example.com
 
+          helm template qa "$chart" \
+            --set formbricks.webappUrl=https://qa.example.com \
+            --show-only templates/deployment.yaml > "$render_dir/deployment-default.yaml"
+          grep -q 'terminationGracePeriodSeconds: 30' "$render_dir/deployment-default.yaml"
+          ! grep -q '^          lifecycle:$' "$render_dir/deployment-default.yaml"
```
xernobyl: **[P2] This assertion can never fail the job.**

Bash exempts `!`-inverted commands from `set -e` (the "return value is being inverted with `!`" case in the manual), so if `lifecycle:` ever started rendering by default, the grep would match, the negation would return 1, and the script would carry on regardless. Repro:

```bash
bash -c 'set -euo pipefail; printf "lifecycle:\n" > /tmp/t; ! grep -q "^lifecycle:$" /tmp/t; echo REACHED'
```

prints `REACHED` and exits 0.

The working idiom is already in this file at lines 69-75:

```bash
if grep -q '^          lifecycle:$' "$render_dir/deployment-default.yaml"; then
  echo "Expected no lifecycle block in the default render" >&2
  exit 1
fi
```

In fairness this is a pre-existing pattern — lines 101, 108, 142 and 148 do the same on `main`, so this is following house style rather than introducing the bug. I'd only fix the line this PR adds and file the other four separately, unless you'd rather sweep them all here.

charts/formbricks/values.yaml
```diff
@@ -73,6 +73,12 @@ deployment:
     #   maxSurge: 25%
     #   maxUnavailable: 25%
 
+  # Time Kubernetes allows the app container to stop gracefully
+  terminationGracePeriodSeconds: 30
+
+  # Optional app container lifecycle hooks
+  lifecycle: {}
```
xernobyl: **[P3] The defaults quietly encode the footgun this PR exists to fix.**

`terminationGracePeriodSeconds: 30` combined with a preStop `sleep 30` — the exact example the new CI test uses — means the hook eats the entire grace budget. Kubelet then sends SIGTERM and SIGKILLs within its ~2s one-off extension, so in-flight requests still drop: the same 502 class the PR is addressing. The grace period covers preStop *plus* process shutdown, not preStop and then a fresh budget.

The CI test pairs `sleep 30` with grace `60`, so you clearly know this, but nothing an operator reads says the grace period must exceed the drain. A commented example would carry it, and matches the convention `strategy` already uses four lines up:

```yaml
  # Time Kubernetes allows the Pod to stop gracefully.
  # Must exceed any preStop drain below, since the hook runs inside this budget.
  terminationGracePeriodSeconds: 30

  # Optional app container lifecycle hooks
  lifecycle: {}
  #   preStop:
  #     sleep:
  #       seconds: 20   # requires Kubernetes >= 1.29; use exec sh -c "sleep 20" below that
```

The native `sleep` handler (K8s 1.29+) is worth preferring in the example since it needs no shell in the image. Folds in the Pod-vs-container wording CodeRabbit raised too.

xernobyl (commented):
Reviewed at `f6ec50d2`. Rendered the chart locally to check the assertions rather than reading them, so the notes below are verified against this head.

Overall this looks good — small, generic, no change to default rendering. Two inline notes (one P2, one P3) plus a scope thought, and a list of things I checked that are fine.

**On the CodeRabbit nit:** it's valid, and slightly more than pedantry. The template supports Istio sidecar injection (`.Values.deployment.disableIstioInject`), so the pod really can hold more than one container, and `terminationGracePeriodSeconds` covers all of them. It also correctly left the `lifecycle` wording alone — that one genuinely is app-container-scoped. Worth taking, but not a blocker.

**[P3] Scope:** `hub-deployment.yaml`, `hub-worker-deployment.yaml`, `cube-deployment.yaml` and `taxonomy-deployment.yaml` have neither field. Hub sits behind the same gateway path so the identical deregistration race applies to it, and `hub-worker` is probably the one that most wants a grace period, for in-flight jobs. Reads like a deliberate scoping to the incident rather than an oversight — flagging it as a follow-up ticket, not a change here.

<details>
<summary>Checked and cleared (no action needed)</summary>

- `terminationGracePeriodSeconds` is not dead config — `templates/deployment.yaml:79` already had `| default 30` on `main`, so this PR only surfaces and documents it. Confirms the "no behaviour change" claim in the description.
- No `Chart.yaml` bump needed: version is `0.0.0-dev` in-repo and stamped by `release-helm-chart.yml` at release time.
- No `values.schema.json` to keep in sync, and the README table isn't helm-docs-generated, so hand-editing it is correct.
- `sh -c "sleep 30"` will actually run — the runtime image is `node:24-alpine3.23`, busybox provides both.
- Rendered both paths: default gives grace 30 with no `lifecycle` block; the override gives grace 60 with correct indentation. `--set-json` needs Helm >= 3.10 and CI pins v3.15.4, so that's fine.
- Very minor: the grace-period greps are unanchored, so `terminationGracePeriodSeconds: 30` would also match `300`. Not worth a commit on its own.

</details>

.github/workflows/helm-chart-validation.yml
BhagyaAmarasinghe: fixed

charts/formbricks/values.yaml
BhagyaAmarasinghe: fixed

xernobyl (approved):
LGTM

Dhruwang (approved):

PR #8776, opened 2026-08-06
fix(surveys): expose progress and auto-close countdown to assistive tech [ENG-1070]

itsjavi:
## What & why

Survey progress and auto-close countdown bars were visual-only, so assistive technology could not identify their purpose or state.

This PR:

- exposes survey progress as a named, determinate progress bar whose clamped and floored value also drives the painted width;
- exposes the auto-close countdown as a named, indeterminate progress bar;
- mounts the polite countdown live region empty, then populates it after a tick so screen readers can observe the content change;
- localizes the new accessible names and announcement in every survey locale;
- adds Playwright coverage for stacked and cardless progress, hidden progress bars, countdown semantics, mutation order, and removal after interaction.

## Linear ticket

Fixes ENG-1070

## How this was tested

- `pnpm --filter @formbricks/surveys test` ✅ 27 files and 667 tests passed.
- `pnpm --filter @formbricks/surveys typecheck` ✅
- Scoped surveys ESLint ✅
- Scoped Prettier check ✅
- `pnpm i18n:validate` ✅ 88 survey keys in every locale.
- `pnpm build --filter=@formbricks/surveys... --force` ✅ surveys and dependencies rebuilt without cache.
- `survey-progress-a11y.spec.ts` previously passed 6 of 6 tests and 18 of 18 repeated runs. The latest empty-first mutation assertion was not rerun locally after the review fix and will run in CI.
- A real NVDA or VoiceOver pass is still required to verify spoken output.
- No screenshot is attached because the change is semantic and does not intentionally alter the painted UI.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a published multi-card link survey with the progress bar enabled. Confirm the bar is visible and a screen reader identifies it as "Survey progress" with the current percentage.
- [ ] Move through the welcome card, question cards, and ending card. Confirm the announced percentage matches the visual width, starts at 0%, increases between question cards, and reaches 100% on the ending card.
- [ ] Open a cardless survey. Confirm the welcome card has no progress bar and question cards expose the named progress bar with the expected percentage.
- [ ] Open a survey with the progress bar hidden. Confirm no visual or accessible survey progress bar is present.
- [ ] Open an app survey with a 30-second auto-close timeout. Confirm the countdown is identified as "Time remaining before the survey closes" and the screen reader announces "This survey closes automatically in 30 seconds unless you respond" once.
- [ ] Interact with the first survey field. Confirm the auto-close bar and its announcement are removed while the normal survey progress bar remains.

**Preconditions / test data**

- A published multi-card link survey with a welcome card and progress bar enabled.
- A published cardless link survey.
- A published link survey with the progress bar hidden.
- An app survey with a 30-second auto-close timeout and no welcome card.
- NVDA, VoiceOver, or another screen reader for announcement checks.

**Risks & regressions**

- Re-check that percentage rounding remains aligned with the painted width across all cards.
- Re-check that stacked and cardless layouts retain their existing visual spacing.
- Re-check that the countdown announcement fires once after mount and does not remain after interaction.
- Re-check right-to-left and dark-mode rendering because the progress bars remain visible UI, even though no visual change is intended.

- none


itsjavi:
Thanks — three of the four are fixed in `2e5ee1350`; the fourth is not from this PR.

### `Readonly<>` on both components — fixed

Correct on both counts, and required by `AGENTS.md` ("Always mark React component props as `Readonly<>`"). `Progress` and `AutoCloseProgressBar` both now use it.

### Russian plural — fixed, but at the source rather than in `ru-RU.json`

Good catch, and the reasoning is right: `one` is not "exactly 1" outside English. Russian selects it for 21, 31, 41…, so the hardcoded literal made a 21-second timeout announce "через 1 секунду".

I did not apply the suggested edit to `ru-RU.json` though, for two reasons. The generated locales are machine-written — `AGENTS.md` says never to hand-edit them, and the next `pnpm i18n` would have reverted it. More importantly, the bug is not in the Russian file: it is in my **en-US source**, which hardcoded `1 second` in the `one` branch, and that is what the translator mirrored.

Interpolating `{count}` in the source leaves the English output identical (English's `one` really is only 1) and gives every locale a number to place. After regenerating:

- **17 of 20** locales now interpolate `{count}` in their `one` branch, including Russian → `через {count} секунду`;
- the other three — `id`, `ja`, `zh-Hant` — correctly have no `one` branch at all, since CLDR gives them only `other`, and their single branch does interpolate.

So one source-level change fixed the Russian case plus 16 others that were not flagged.

### Italian `takes_x_plus_minutes` — not this PR

`Richiede più di {count} minuti` is on `main` unchanged. This PR touched each locale file with exactly **3 insertions and 0 deletions** — the three new keys — so that line is pre-existing and simply adjacent to the addition.

The observation looks right on its merits ("più di" reads as strictly more than, where the key means `{count}+`), but a translation correction to an unrelated key does not belong in a compliance PR. Worth its own ticket.

### One the review did not flag

The same plural bug still exists on `main` in `takes_x_minutes` (`one {Займёт 1 минуту}`) and `takes_less_than_x_minutes`, from the same root cause — hardcoded numerals in the en-US `one` branch, which is the existing convention across these keys. A Russian respondent currently gets told a 21-minute survey takes 1 minute. Also its own ticket, but worth knowing it is a pattern rather than a one-off.

### Verification after the fixes

`survey-progress-a11y.spec.ts` → **6 passed** against a rebuilt bundle, including both countdown tests that assert the ICU-resolved announcement text, so the string change did not invalidate them. `pnpm i18n` → 87/87 keys, all locales complete. `eslint`, `tsc --noEmit`, `prettier` clean; `@formbricks/surveys` unit → 665 passed.

---

packages/surveys/locales/it-IT.json
```diff
@@ -29,13 +29,16 @@
     "select_option": "Seleziona un'opzione",
     "select_options": "Seleziona opzioni",
     "sending_responses": "Invio risposte in corso...",
+    "survey_closes_automatically_in_x_seconds": "{count, plural, one {Questo sondaggio si chiude automaticamente tra 1 secondo se non rispondi} other {Questo sondaggio si chiude automaticamente tra {count} secondi se non rispondi}}",
     "survey_dialog": "Finestra di dialogo del sondaggio",
+    "survey_progress": "Avanzamento del sondaggio",
     "takes_less_than_x_minutes": "{count, plural, one {Richiede meno di 1 minuto} other {Richiede meno di {count} minuti}}",
     "takes_x_minutes": "{count, plural, one {Richiede 1 minuto} other {Richiede {count} minuti}}",
     "takes_x_plus_minutes": "Richiede più di {count} minuti",
```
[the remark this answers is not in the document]
itsjavi: Resolving without a change: this line is not part of this PR. `Richiede più di {count} minuti` is on `main` untouched — this PR added exactly 3 keys and deleted nothing from each locale file, so it is simply adjacent to the addition.

The observation looks correct on its merits ("più di" reads as strictly more than, where the key means `{count}+`), but it needs its own ticket rather than a translation change smuggled into an accessibility PR.

---

packages/surveys/src/components/general/auto-close-progress-bar.tsx
```diff
@@ -1,18 +1,36 @@
...
+          style={{
+            animation: `shrink-width-to-zero ${autoCloseTimeout.toString()}s linear forwards`,
+            width: "100%",
+          }}
+        />
+      </div>
+      {/* A shrinking bar conveys the deadline visually only. This states it once, politely, when the
+          countdown mounts. The bar itself stays silent — an indeterminate progressbar has nothing to
+          report — and role="timer" was avoided because it would re-announce on every tick. */}
+      <div className="sr-only" aria-live="polite" aria-atomic="true">
```
Dhruwang, line 43: This region mounts with its text already in place. A lot of screen readers (JAWS especially) only announce a live region that was already present-and-empty when the content changes — one inserted already-populated often stays silent, so the countdown may never actually be spoken. This is the exact reason #8766's ElementError mounts empty then fills. Suggest the same here: render the sr-only div empty and set the text in a useEffect a tick after mount. Worth a real NVDA/VoiceOver check either way — the spec can only prove the markup is present, not that it's announced.

Dhruwang (commented):
Nice change — the determinate/indeterminate split is well reasoned and the spec asserting aria-valuenow tracks the painted width (not just that it's valid) is exactly right. One real concern inline on the countdown announcement, plus one heads-up:

While reading the mount path I hit a pre-existing bug in `auto-close-wrapper.tsx` (not touched by this PR): the effect cleanup on `autoCloseTimeout` change can stale-mark `hasInteracted=true`, hiding the countdown UI/announcement while the underlying auto-close timer keeps running (shows up in editor live-preview when autoClose changes). Out of scope here, but worth a follow-up ticket since it undercuts this same feature.

packages/surveys/src/components/general/auto-close-progress-bar.tsx
itsjavi: Fixed in 7ba93a12c. The live region now mounts with an empty string, then a zero-delay effect populates the localized countdown text. The cleanup cancels the pending update if the component unmounts or the announcement changes. The Playwright regression now records the region's mutation sequence and verifies that the empty state precedes the populated announcement. A real NVDA or VoiceOver pass remains the appropriate manual check for actual speech output.

Dhruwang (approved):

PR #8778, opened 2026-08-06
chore: move CodeRabbit config into the repo as .coderabbit.yaml (ENG-2268)

mattinannt:
## What & why

Every CodeRabbit review footer read `Configuration used: Repository UI`. Our entire review configuration — profile, pre-merge check modes, path filters — lived in the CodeRabbit dashboard: not version-controlled, not reviewable in a PR, no change history, and alterable by anyone with dashboard access without a trace in the repo.

This adds `.coderabbit.yaml` at the repo root.

**`0eb3a47` — the migration, no behaviour change.** CodeRabbit can export its own resolved configuration (`@coderabbitai generate configuration`), so this is a transcription of what was already running rather than a hand-written approximation. Notable settings that were only visible in the dashboard: `profile: assertive`, `early_access: true`, `inheritance: false`, `request_changes_workflow: false`, all four pre-merge checks in `warning` mode, and most summary/label/reviewer surfaces switched off.

**`883ba31` — the changes the migration unblocks.** Docstring-coverage check off (it failed routinely — 28.57%, 16.67%, 66.67% against an 80% threshold — was ignored every time, and its detector reports *"No functions found in the changed files"* on PRs that plainly add functions; `mode: "off"` is quoted because unquoted `off` is a boolean in YAML 1.1). Path instructions encoding the repo rules a generic reviewer keeps getting wrong. `tone_instructions` to stop comments on formatting and import order, which Prettier and ESLint own. And `path_filters` re-including `apps/web/locales/en-US.json`, which the single existing filter had hidden along with the generated locales — it is the one that is hand-written and the source of truth for the rest.

**`db08896` — two custom pre-merge checks**, both advisory. *Breaking changes match the diff* (`pr-label-sync.yml` reads what the section **says**; nothing verified it was **true**) and *QA plan is executable by a tester*.

**`ee4e2fd` — first review round**, plus a merge of `main`. `base_branches` takes **regex**, not globs, so the transcribed `epic/*` also matched any branch containing "epic" — `feat/epicenter-cleanup` qualified; corrected to `^epic/`. The breaking-changes check demanded a table where the template accepts *"a table or prose"*, and counted every added env var as breaking, contradicting the `charts/formbricks/**` instruction in the same file. The QA check treated any file path as proof a step wasn't tester-executable, which fails a legitimate step pointing at a published docs page.

**`97089fc` — two inherited settings, decided rather than carried over.** `auto_incremental_review: true`, so the commits that fix a review round get reviewed — with it off, `auto_pause_after_reviewed_commits: 5` was dead config. `chat.allow_non_org_members: false`, since this is a public repo and the previous value let anyone outside the org drive the bot by comment.

**`bbcc045` and `fbfbb4a` — second review round.** Six findings, each verified against the schema or the checkout before applying:

- `review_status: true`. The schema calls this the notice posted "when a review is skipped". Enabling incremental review in `97089fc` made the pause live, so without this a paused review would look exactly like a clean one — a bug introduced by my own earlier commit.
- `learnings.approval_delay: 7`. At `0` the schema applies a learning "immediately without approval", making every learning a durable, unversioned review rule settable from a comment thread — the same failure this PR exists to end, one layer down. CodeRabbit recorded one during this very review.
- The `**/*.tsx` block now asks for effect and closure bugs. `apps/web/eslint.config.mjs` turns five `react-hooks` rules off and leaves `exhaustive-deps` at `warn` with no `--max-warnings 0`, so that class is unguarded by CI.
- The v3 block was spec-sync only — weaker than the rule for the legacy v1 surface above it. It now carries the security contract: `withV3ApiWrapper` (30 files), `requireV3WorkspaceAccess` (17), 403-not-404 so a missing resource doesn't leak existence, and the RFC 9457 builders. Adds a first instruction for `apps/web/app/api/(internal)/**`.
- **No new Server Actions** — new backend work belongs in an `/api/v3` route consumed with TanStack Query. This also updates `AGENTS.md`, which still presented Server Actions as the pattern: CodeRabbit applies `AGENTS.md` as a code guideline, so leaving it would have handed the reviewer the ban and the endorsement at once.
- `finishing_touches.autofix` and `fix_ci` off. Both let the reviewer author commits, and with incremental review on it would then review its own work. Reviewers here are read-only on the code. `resolve_merge_conflict` stays — mechanical, and visible in the diff.

**`3ec6f09` — the config was invalid, and the checks here did not catch it.** `coderabbit config validate`, run locally, rejected the file: `tone_instructions` was 319 characters against a `maxLength` of 250. That does not truncate the tone — CodeRabbit refuses the whole configuration, so this would have merged and gone on silently using the dashboard settings. Rewritten to 245 characters with the substance intact, with a few characters of headroom and a comment naming the limit, because landing exactly on 250 makes the next one-word edit a silent failure. Every other length-constrained field was audited: longest path instruction 1369 of 20000, longest custom check 1613 of 10000.

ENG-2268 asked for the behaviour changes as a separate follow-up PR. Combining them was a deliberate call so the noise reductions land immediately; each commit stays separately revertable if a review-quality regression needs attributing.

## Linear ticket

Fixes https://linear.app/formbricks/issue/ENG-2268/move-coderabbit-config-into-the-repo-as-coderabbityaml

## How this was tested

- **Full JSON Schema validation** with `ajv` against the published schema (`www.coderabbit.ai/integrations/schema.v2.json`): valid, including types, enums and length limits.

  Earlier revisions of this PR described their validation as schema validation when it was not. It was a recursive walk over key *names* plus a handful of values checked by hand — enough to prove there were no unknown keys, structurally incapable of catching a `maxLength`, a bad enum or a wrong type. That is why the `tone_instructions` overrun above survived six commits and was found by the CLI rather than here.
- `coderabbit config validate` (CLI v0.7.2) is the authoritative check and should be run before pushing this file. It cannot run in the agent sandbox — its bundled `fetch` ignores `HTTPS_PROXY`, so the schema fetch fails — so the `ajv` run above stands in, using the same schema document.
- Checked all 19 `path_instructions` globs and every `path_filters` pattern with `minimatch` against `git ls-files`. All match at least one real file — no dead rules, including the new `apps/web/app/api/(internal)/**` (4 files). Two candidate filters matched zero tracked files because they are gitignored, and were dropped rather than shipped dead.
- Checked the `base_branches` correction as a real regex: `^epic/` matches `epic/foo` and `epic/bar/baz`, and no longer matches `feat/epicenter-cleanup`, which the previous value did.
- Verified each review claim against the checkout before acting on it, rather than trusting it: the ESLint rule states, the four v3 helpers and their usage counts, and the `pull_request_template.md` wording on `main`.
- `prettier --check .coderabbit.yaml` passes — relevant since `#8777` added `pnpm format:check` to the lint job and it covers root YAML.
- No application code is touched, so no test suite exercises this change.

**This PR cannot verify its own effect.** CodeRabbit's first review reported:

> Ignoring CodeRabbit configuration file changes. For security, only the configuration from the base branch is applied for open source repositories.

So every review on this PR ran on the dashboard configuration — expected, not a failure. The configuration takes effect once it is on `main`. Note the consequence, given the bug above: an invalid file cannot be detected by anything on this PR either.

## Breaking changes

None

## QA / Test Plan

This changes AI review tooling and one guidance document. It has no user-facing surface, no runtime effect on the app, and nothing for a manual tester to click. Every step below is checked **after this PR merges**, on subsequent PRs.

**How to test**

- [ ] **Check this one first.** On the next PR that touches `apps/web/locales/en-US.json` → the file appears in CodeRabbit's reviewed-files list. The re-inclusion relies on a later `path_filters` pattern overriding an earlier `!` negation, which the reference does not state explicitly; if precedence turns out to be exclusion-wins, the source-of-truth locale file stays invisible to review and nothing announces it. `de-DE.json` and the other generated locales should stay excluded.
- [ ] Open the CodeRabbit review on the first PR after this merges, expand `Run configuration` → **Configuration used** no longer says `Repository UI`. If it still does, the file is being rejected as invalid — run `coderabbit config validate` on `main`.
- [ ] Same footer's pre-merge checks table → `Docstring Coverage` is absent; the two custom checks appear; `Title`, `Description` and `Linked Issues` still run. All non-blocking.
- [ ] Push a follow-up commit to any open PR → CodeRabbit reviews it without being asked, and posts a visible "review skipped" notice once the five-commit pause trips.
- [ ] Under any CodeRabbit review comment → no 🪄 Autofix checkbox is offered, and `@coderabbitai autofix` is declined.
- [ ] Someone outside the org comments `@coderabbitai review` → not honoured, while the automatic review on their PR still happens.
- [ ] On the next PR adding a new Server Action → CodeRabbit raises the action itself, not just its authorization wiring.
- [ ] On the next PR adding a Prisma migration → CodeRabbit comments on index locking or idempotency where relevant.
- [ ] On a PR that renames an API response field while leaving "Breaking changes" as "None" → the custom check warns and names the file and line. On one documenting a real break in prose rather than a table → it passes.

**Preconditions / test data**

- None to set up. Note that for open source repositories CodeRabbit reads the configuration from the **base** branch, never from the PR head — nothing is verifiable until this is merged to `main`.

**Risks & regressions**

- The `assertive` profile is preserved, so review volume should not change beyond the suppressed docstring and formatting comments. If reviews get noticeably worse, revert `883ba31` alone — `0eb3a47` keeps the configuration in the repo at exact dashboard parity.
- `auto_incremental_review: true` increases review volume by design. `auto_pause_after_reviewed_commits: 5` is the cap; lower it before reverting.
- The `AGENTS.md` change makes the Server Actions deprecation binding on every agent reading it, not only CodeRabbit. That is the intent, but it is the one change here with reach beyond review tooling.
- An invalid value fails closed and silently: CodeRabbit rejects the whole file and falls back to the dashboard, and nothing on the PR that introduced it says so. `coderabbit config validate` before pushing is the guard; the first QA step above is the backstop.
- The custom checks are LLM-judged and advisory. If either is noisy, set its `mode` to `off` rather than reverting.
- Dashboard settings become dead weight once this merges. With `inheritance: false` and this file present, any future dashboard edit is silently ignored — the dashboard should be cleared so nobody edits it expecting an effect. Clear it only after the first QA step confirms the file is actually being used.
- `drafts: false` is carried over deliberately and left for a separate decision once the incremental-review volume is known.

- none


.coderabbit.yaml
```diff
@@ -0,0 +1,500 @@
...
+  auto_review:
+    enabled: true
+    description_keyword: ""
+    auto_incremental_review: false
+    auto_pause_after_reviewed_commits: 5
+    ignore_title_keywords: []
+    labels: []
+    drafts: false
+    base_branches:
+      - "epic/*"
```
itsjavi: `base_branches` takes **regex**, not globs — the config reference is explicit about it ("Base branches (other than the default branch) to review. Accepts regex patterns."). `epic/*` as a regex is `epic` followed by zero-or-more `/`, and matching is unanchored, so it also matches any branch whose name merely *contains* `epic` — `feat/epicenter-cleanup` would qualify. Since this line is a transcription of the dashboard value it isn't a regression, but the migration is the moment to correct it:

```suggestion
      - "^epic/"
```

Good news on the related question: the default branch is always reviewed regardless of this list, so `main` PRs are unaffected either way.

---

.coderabbit.yaml
```diff
@@ -0,0 +1,500 @@
...
+          - Changes an exported signature of the public SDK surface in `packages/js-core` or
+            `packages/surveys`.
+
+          FAIL if the diff contains at least one of the above and the "Breaking changes" section
+          says "None", is empty, or still holds the unedited template comment. Name the specific
+          file and line that is breaking.
+
+          FAIL if the section claims a breaking change that has no corresponding change in the diff.
+
+          PASS if the section holds a table whose rows correspond to the breaking changes in the
```
itsjavi: This PASS condition requires a **table**, but `.github/pull_request_template.md` explicitly blesses prose for the same section:

> The `breaking-change` label is applied automatically from this section — **a table or prose both count**, so you should not need to touch it.

So a PR that documents a genuine breaking change in a sentence — exactly what the template permits, and what `pr-label-sync.yml` already labels correctly — fails this check. Since the check exists to make the section trustworthy rather than to enforce a shape, suggest relaxing it to something like *"PASS if the section describes, in a table or in prose, each breaking change present in the diff"*.

Two smaller notes on the same check while you're in here:

- Line 273 counts *adding* an env var or config key as breaking. That's the template's wording too, so it isn't your invention — but it does conflict with the `charts/formbricks/**` instruction you add above ("New values need a default that keeps existing installs working unchanged"), and will warn on every PR that adds a purely additive, defaulted variable. Scoping it to "adds a **required** env var, or one with no safe default" would remove that class of false positive without weakening the check.
- The `Return inconclusive rather than failing…` escape hatch (line 288) is the right instinct and reads well.

---

.coderabbit.yaml
```diff
@@ -0,0 +1,500 @@
...
+      - name: QA plan is executable by a tester
+        mode: warning
+        instructions: |
+          Check the PR description's "## QA / Test Plan" section. Its audience is a manual tester
+          with no repository or CLI access, working only from the running app or the live API.
+
+          FAIL if any of the following is true:
+          - The "How to test" subsection has no steps, or holds only the unedited template
+            placeholder text.
+          - A step cannot be performed through the app UI or an HTTP request — for example it
```
itsjavi, line 352: `a file path` is the one item in this list that will misfire. The other examples (`pnpm test`, `pnpm build`, `grep`, reading code) all describe things a tester genuinely cannot do; a file path does not, because docs are a user-facing surface. A perfectly valid QA step such as

> Open `docs/self-hosting/configuration/job-runner.mdx` → the workloads table lists `survey-archive-purge.process`

is executable by anyone with a browser, and would be failed by this rule as written. (#8765's QA plan has exactly that step today.)

Suggest narrowing it to the thing you actually mean — a step that requires reading or running repository *source*:

```suggestion
          - A step cannot be performed through the app UI or an HTTP request — for example it
            references `pnpm test`, `pnpm build`, `grep`, a source file, a function name, or reading
            code. Those belong in "How this was tested" instead. Published documentation pages are
            a user-facing surface and are fine to reference.
```

---

.coderabbit.yaml
```diff
@@ -0,0 +1,500 @@
...
+  abort_on_close: true
+  disable_cache: false
+
+  slop_detection:
+    enabled: true
+
+  auto_review:
+    enabled: true
+    description_keyword: ""
+    auto_incremental_review: false
```
itsjavi: Not a change you're making — but this migration is the first time it's visible in a diff, so worth a deliberate decision rather than inheriting it.

With `auto_incremental_review: false`, follow-up pushes are never re-reviewed. The commit status on this very PR reads *"Review skipped: incremental reviews are disabled"*, and on #8765 the two commits that **fixed** CodeRabbit's own findings got no re-review — someone has to remember to type `@coderabbitai review`. That's the weakest link in the review loop this PR is otherwise tightening.

Two related knobs: `auto_pause_after_reviewed_commits: 5` (line 218) is dead while incremental review is off, and `drafts: false` (line 224) means draft PRs get no review at all — relevant given how often this repo opens PRs as drafts (#8704 has been a draft for a week).

If the reason it's off is review volume, `auto_incremental_review: true` plus the pause-after-5 you already have may get you re-reviews without the noise.

---

itsjavi (commented):
Read the full 500-line file rather than skimming it, and verified the claims in the description that are checkable from the repo. The premise is right: review configuration that lives only in a dashboard is unversioned, unattributable, and silently editable, and the three-commit split (parity transcription → behaviour changes → custom checks) keeps each part separately revertable.

**Verified against the repo (all clean)**

- Every `path_instructions` glob and every `path_filters` pattern resolves to real tracked files: `charts/formbricks/values.yaml`, `apps/web/vendor/`, `packages/database/schema/*.prisma`, `packages/database/migration/**`, `apps/web/app/api/v1|v2|v3`, `apps/web/modules/api/v2`, `apps/web/modules/ee/LICENSE`, `docs/api-v3-reference/src/`, both `i18n.lock` files, `packages/surveys/locales/`. No dead rules — the description's claim holds.
- The `**/*.test.tsx` instruction is correctly scoped to *newly added* files: only 3 such files exist today, and they'd otherwise generate a comment each on every touch.
- `docstrings.mode: "off"` is quoted, so it parses as a string and not a YAML 1.1 boolean.
- CI green; merges cleanly onto `main` (5 commits behind, no conflicts).

**Comments inline** — one correctness fix (`base_branches` is regex, not glob), two likely false-positive sources in the new custom checks, and one inherited setting worth a conscious decision rather than a silent carry-over.

**One thing to confirm first, not block on**

The `path_filters` re-include of `en-US.json` depends on later patterns overriding earlier `!` negations. The configuration reference doesn't state that precedence rule explicitly — it says the patterns behave like `git sparse-checkout` patterns, where last-match-wins is the norm, so the reasoning is sound but undocumented. Your QA plan already covers it ("the next PR that touches `apps/web/locales/en-US.json` → the file appears in the reviewed-files list"); just make that the *first* thing checked after merge, because if precedence turns out to be exclusion-wins, the source-of-truth locale file stays invisible to review and nothing tells you.

**Needs human review** @itsjavi

- Nothing here can be verified before merge — for OSS repos CodeRabbit reads config from the base branch only, so the config on this PR's head is inert and the reviews you see on it ran on the dashboard settings. The description is upfront about this; the trade-off is real and worth an explicit call.
- The judgement calls are yours, not mine to settle from the diff: whether `profile: assertive` plus `request_changes_workflow: false` is the balance you want, and whether `chat.allow_non_org_members: true` (line 448) is intentional on a public repo — it lets any outside contributor drive the bot via comments, which combined with `finishing_touches.autofix`/`fix_ci` is worth a deliberate yes rather than an inherited default.
- Per the description, the dashboard should be cleared after merge: with `inheritance: false` and this file present, later dashboard edits are silently ignored, so leaving it populated is a trap for whoever edits it next.

---

mattinannt:
Thanks — three fixes pushed in `ee4e2fd`, each verified against the source rather than taken on trust.

**`base_branches` is regex.** Confirmed in the schema: *"Accepts regex patterns."* Checked the actual matching, and your example is exactly right:

| branch | `epic/*` | `^epic/` |
| --- | --- | --- |
| `epic/foo` | ✅ | ✅ |
| `feat/epicenter-cleanup` | ✅ (wrong) | ❌ |

Applied, with a comment recording why the transcribed value was changed.

**Table-vs-prose.** Confirmed — `main`'s template says *"a table or prose both count"*. My local checkout predated `#8763`, so I'd written the check against the older wording that only mentioned a table. Relaxed to accept either shape. Also took the second half: adding an env var now counts as breaking only when it is required or has no safe default, which resolves the contradiction with the `charts/formbricks/**` instruction.

**"a file path" in the QA check.** Agreed, and the `#8765` example makes it concrete. Narrowed to source files, with published docs called out explicitly as a user-facing surface.

**`auto_incremental_review`** — deferring to Matti rather than deciding it here. You're right that it's the weakest link, and that this diff is the first time it's been visible; it's also a review-volume and cost decision, which is his call, not something to settle inside a migration PR. Same for `chat.allow_non_org_members` and `drafts`. If he wants them changed I'll do it here; otherwise they stay as transcribed and get their own PR.

**On precedence:** taking your advice — the `en-US.json` re-inclusion is now the first item in the QA plan, since a silent exclusion-wins outcome is invisible.

Also merged `main`, which brings in the `prettier --check` enforcement from `#8777`. That now covers root YAML; this file passes unchanged.

---

.coderabbit.yaml
mattinannt: Applied in `ee4e2fd`. Confirmed against the schema before changing it — `"Base branches (other than the default branch) to review. Accepts regex patterns."` — and checked the actual matching rather than trusting the reading:

| branch | `epic/*` | `^epic/` |
| --- | --- | --- |
| `epic/foo` | ✅ | ✅ |
| `epic/bar/baz` | ✅ | ✅ |
| `feat/epicenter-cleanup` | ✅ | ❌ |
| `main` | ❌ | ❌ |

Left a comment on the line recording why a transcribed value was changed, so it doesn't read as a transcription error later.

---

.coderabbit.yaml
mattinannt: Both taken, in `ee4e2fd`.

You're right about the template, and I checked why I got it wrong: my checkout predated `#8763`, so I wrote the check against the older wording that only mentioned filling the table. The PASS condition now reads *"describes, in a table or in prose, each breaking change present in the diff"*, with the template named as the reason.

On the env var: agreed the contradiction with the `charts/formbricks/**` instruction is the tell — a check that fires on exactly the thing another rule in the same file calls correct is a bug. Adding one now counts only when it is required or has no default that keeps existing installs working. Removals and renames still count unconditionally.

---

.coderabbit.yaml
mattinannt: Agreed, and the `#8765` example settles it — a docs page is a user-facing surface, so pointing at one is a legitimate QA step. Took your wording verbatim in `ee4e2fd`: narrowed to "a source file", with published documentation pages called out explicitly as fine to reference.

---

.coderabbit.yaml
mattinannt: Matti made the call, and it's your recommendation: `auto_incremental_review: true` in `97089fc`, keeping the existing `auto_pause_after_reviewed_commits: 5` as the volume cap — which, as you noted, was dead config until now.

He also turned off `chat.allow_non_org_members` off the back of your point about it on a public repo with `autofix`/`fix_ci` enabled. Outside PRs are still reviewed automatically; non-members just can't invoke the bot.

`drafts` stays `false` for now — deliberately, not by omission. It's a bigger volume change and easier to judge once we can see what incremental review actually costs. Recorded in the PR body as an open decision rather than left silent.

---

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+        - Network and storage boundaries should be mocked via helpers from `@formbricks/*` and
+          mocks placed in a `__mocks__` directory.
+
+    - path: "apps/web/playwright/**/*.spec.ts"
+      instructions: |
+        - Flag timing hacks (`waitForTimeout`, arbitrary sleeps) and dependencies on data another
+          spec created — these are the repo's main source of flake.
+        - Specs must stay small and single-purpose; flag mega-specs. Slow suites need a `@slow` tag.
+
+    - path: "**/actions.ts"
```
itsjavi, line 114: **Nothing in this file says new Server Actions are banned — and this block coaches the opposite.**

Our review guidance treats "no new Server Actions; new backend work is an HTTP API under `/api/v3`" as the one rule that explicitly *overrides* `AGENTS.md`. And `AGENTS.md:42` still reads "Server actions wrap service calls and return `{ data }` or `{ error }` consistently" — no deprecation note. Since CodeRabbit auto-applies `AGENTS.md` as a code guideline, this file is the only place that can correct it. Instead this block describes what a *good* new action looks like, so a PR adding one gets reviewed for authorization and rate limiting and then passes.

That's the single rule a generic reviewer is most likely to get wrong here, which is exactly what these 18 blocks exist for.

Suggested addition to the block:

> A **newly added** exported action is itself the finding: Server Actions are legacy and no new ones should be added. New backend work belongs in an `/api/v3` route consumed from the client with TanStack Query. Raise this first, then review the points below for actions that already exist.

The checks already listed stay useful for the 58 existing `actions.ts` files.

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+
+  path_instructions:
+    - path: "**/*.test.tsx"
+      instructions: |
+        AGENTS.md forbids React component unit tests in this repo — components are covered by
+        Playwright specs in `apps/web/playwright/`. If this file is newly added, say so and ask for
+        the coverage to move to a Playwright `.spec.ts` instead. Do not raise this on edits to the
+        handful of `.test.tsx` files that already exist.
+
+    - path: "**/*.tsx"
```
itsjavi, line 79: **Two additions to this block.**

1. Same gap as `**/actions.ts` from the other side: nothing tells the reviewer that new client data flows go through `/api/v3` + TanStack Query, that server data lives in the query cache only (never mirrored into `useState` or Jotai), and that `router.refresh()` is not a data-refresh mechanism. Without it the reviewer will accept new action-based wiring in new components.

2. The block tells the reviewer what *not* to say about hooks (`useMemo`/`useCallback`) but not what to look for. `apps/web/eslint.config.mjs:14-21` turns `react-hooks/set-state-in-effect`, `incompatible-library`, `error-boundaries`, `immutability` and `preserve-manual-memoization` **off**, and keeps `exhaustive-deps` at `warn` — and warnings never fail the lint job. So effect and stale-closure bugs are entirely unguarded in the largest app in the repo, and this is the block that could ask for them.

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+        breaking: it must be intentional, and the PR's "Breaking changes" table must document it.
+        Prefer adding to v3 over changing v1.
+
+    - path: "apps/web/{app/api/v2,modules/api/v2}/**"
+      instructions: |
+        Public API v2. Route handlers stay thin; logic lives in `apps/web/modules/api/v2/`. Any
+        change to a request or response shape must be mirrored in the endpoint's `lib/openapi.ts`,
+        or the published `openapi.yml` silently goes stale. Response-shape changes are breaking.
+
+    - path: "apps/web/app/api/v3/**"
```
itsjavi, line 181: **v3 gets spec-sync only, while the deprecated surface above gets the full auth/rate-limit/audit contract.**

That inversion is backwards for the surface all new backend work is supposed to target. Spec drift is a real bug class and worth keeping, but the security-relevant half is missing. All of these exist in the checkout today:

- Handlers go through `withV3ApiWrapper` (`apps/web/app/api/v3/lib/api-wrapper.ts`), which owns auth, input parsing, rate limiting and audit logging — a hand-rolled handler is a finding.
- Authorization is `requireV3WorkspaceAccess` (`apps/web/app/api/v3/lib/auth.ts`, used by 17 routes today); every query is scoped by the resolved workspace/environment, never fetched by id alone.
- **403, not 404**, for a resource the caller cannot see — a 404 leaks existence.
- Responses come from the RFC 9457 builders in `apps/web/app/api/v3/lib/response.ts`; no invented envelopes, and no internal/DB types or timestamps leaking into a public shape.

Separately, there is no instruction for `apps/web/app/api/(internal)/**` at all. Worth adding the taxonomy rule to one of the two: **a `/v3` route with no public OpenAPI entry is internal by definition** and belongs in the internal group — shipping an internal route as public v3 is the mistake this rule exists to catch.

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+  profile: assertive
+  request_changes_workflow: false
+  high_level_summary: false
+  high_level_summary_instructions: ""
+  high_level_summary_placeholder: "@coderabbitai summary"
+  high_level_summary_in_walkthrough: false
+  auto_title_placeholder: "@coderabbitai"
+  auto_title_instructions: ""
+  review_status: false
+  review_details: false
```
itsjavi: **With incremental review now on, a skipped review becomes indistinguishable from a clean one.**

`review_status` is documented as "post review status messages (**e.g., when a review is skipped**)" and defaults to `true`. Until this PR, `auto_pause_after_reviewed_commits: 5` was dead config — you say so yourself at line 218. Turning on `auto_incremental_review` makes it live, so the pause now actually fires on the fifth commit. With `review_status: false`, plus `high_level_summary: false` and `review_details: false`, nothing is posted when it does.

The failure is concrete: an author pushes a sixth commit, sees no new comments, and reads that as reviewed-and-clean. Our review guidance is explicit in the other direction — a reader has to be able to tell "clean" from "never looked at", and silence must never stand in for a pass.

```suggestion
  review_status: true
  review_details: false
```

`review_details` is fine off; it is the ignored-files/suppressed-comments dump, not the skip notice.

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+    docstrings:
+      enabled: false
+    unit_tests:
+      enabled: false
+    simplify:
+      enabled: false
+    autofix:
+      enabled: true
+    fix_ci:
+      enabled: true
```
itsjavi: **These make the reviewer a writer, and with incremental review on it then reviews its own commits.**

`autofix` is triggered by the one-click 🪄 checkboxes under each review comment, not only by the `@coderabbitai autofix` command — so it is a low-friction path to bot-authored commits on the branch. Every reviewer definition we have is read-only on the code, on the principle that AI-generated code is an untrusted external contribution and the human who merges owns it. A commit the reviewer wrote and then reviewed has had no independent reviewer at all.

The PR already recognises this as a write surface — it is the stated reason for setting `allow_non_org_members: false` — but stops short of the conclusion.

```suggestion
    autofix:
      enabled: false
    fix_ci:
      enabled: false
```

`resolve_merge_conflict` is more defensible (mechanical, and the result is visible in the diff), so I would leave that one on if you want to keep something here.

.coderabbit.yaml
```diff
@@ -0,0 +1,514 @@
...
+knowledge_base:
+  opt_out: false
+  web_search:
+    enabled: true
+  code_guidelines:
+    enabled: true
+    filePatterns: []
+  learnings:
+    scope: auto
+    approval_delay: 0
```
itsjavi: **Learnings are unversioned review rules that apply immediately — the thing this PR exists to remove.**

`approval_delay: 0` is documented as "apply learnings immediately without approval". So every learning CodeRabbit takes from a PR comment becomes a durable review rule that lives in CodeRabbit, not in this file: no change history, no review, and alterable by anyone who replies to the bot. That is the same failure mode as the dashboard configuration described at the top of the description, and the closing note about clearing the dashboard does not reach it.

`scope: auto` resolves to `local` for a public repo, so the blast radius is this repo only — that part is fine. The zero delay is the problem.

```suggestion
    approval_delay: 7
```

Any non-zero value works; 7 gives an admin a week to reject a rule before it starts shaping reviews.

itsjavi (commented):
**6 findings, none blocking merge** — reviewed at `97089fc`.

The config holds up well against how our own review agents and skills work: comment-only reviews, the import-order and quote-style suppressions (both real withdrawn-finding classes here), tests-that-cannot-fail, tenant scoping, the `.tsx` unit-test ban, migration safety, spec sync. `code_guidelines.filePatterns: []` is not a gap — CodeRabbit's built-in defaults already include `**/AGENTS.md`, so the root file is applied as review criteria on top of these instructions. That is also what creates the first finding below.

Six things that should change before this merges. Four are contradictions with our review guidance; two are incompleteness on the surfaces that matter most. All are anchored inline.

**Checked** — consistency with `AGENTS.md` and our review agents/skills, glob validity against `git ls-files`, CodeRabbit schema and documented defaults for every key I comment on. **Not verified here** — the `path_filters` negation-precedence question, which the description already flags as the first post-merge QA item.

_Findings produced with Claude Opus 5 and verified against the checkout; the dispositions are mine._

.coderabbit.yaml
mattinannt: Done in `fbfbb4a`, and I've fixed `AGENTS.md` in the same commit.

Being straight about the process here: I went looking for the rule before encoding it and couldn't find it anywhere I can reach — not `AGENTS.md`, not the `pr-review` skill (whose `failure-patterns.md:355` actually repeats the `{ data }` / `{ error }` guidance), not `.agents/`. So I flagged it to Matti rather than encode an unverifiable rule, and he confirmed it's real.

Your point about `AGENTS.md` is what decided how to fix it. Since CodeRabbit applies it as a code guideline via the default `filePatterns`, encoding the ban here while line 40 still endorsed the pattern would have handed the reviewer both instructions at once. So `AGENTS.md` now reads:

> Server actions are legacy — do not add new ones. New backend work belongs in an `/api/v3` route consumed from the client with TanStack Query… The existing server actions wrap service calls and return `{ data }` or `{ error }` consistently; keep that contract when changing them.

The path instruction leads with the newly-added-action finding and keeps the existing checks for the 58 files already there.

---

.coderabbit.yaml
mattinannt: Both added — the data-flow half in `fbfbb4a`, the hooks half in `bbcc045`.

I verified the ESLint claim rather than taking it, and it's exact: `apps/web/eslint.config.mjs` turns off `set-state-in-effect`, `incompatible-library`, `error-boundaries`, `immutability` and `preserve-manual-memoization`, and the comment on line 19 even says the quiet part — *"the web lint script has no `--max-warnings 0`, so it surfaces violations without blocking"*. So `exhaustive-deps` cannot fail CI either. The block now asks for state set from an effect without a guard, missing or lying dependency arrays, stale closures, and effects that need cleanup and don't have it.

The data-flow rule went in alongside the Server Actions ban, phrased from the client side: `/api/v3` + TanStack Query for new flows, server data in the query cache and not mirrored into `useState` or Jotai, and `router.refresh()` flagged when used as a data-refresh mechanism.

---

.coderabbit.yaml
mattinannt: Applied in `bbcc045`. Checked every claim against the checkout first, and all four hold:

- `withV3ApiWrapper` — `app/api/v3/lib/api-wrapper.ts`, used in 30 files.
- `requireV3WorkspaceAccess` — `app/api/v3/lib/auth.ts`, 17 files, matching your count.
- 403-not-404 — `auth.ts:34` says it outright: *"We use 403 (not 404) when the workspace is not found to avoid leaking resource existence."*
- RFC 9457 — `response.ts:2`, with `problemForbidden`, `problemUnauthorized` and friends built on a shared `problemResponse`.

The inversion you spotted is the real point: the deprecated surface had the full contract and the one we're steering everything toward had spec-sync only. Fixed in both directions now, since the Server Actions ban makes v3 the target.

Also took the taxonomy rule, and added the missing `apps/web/app/api/(internal)/**` block — glob verified against `git ls-files` (4 files) so it isn't a dead rule. It covers internal auth and the reverse mistake: something in the internal group that should have been public v3.

---

.coderabbit.yaml
mattinannt: Applied in `bbcc045`, and this was the best catch of the six — it's a bug I introduced two commits earlier without noticing. Turning on `auto_incremental_review` made the pause live, and I didn't re-examine what that meant for a setting I'd transcribed as-is. The interaction only exists because of my own change.

Schema confirms the reading: *"Post review status messages (e.g., when a review is skipped) in the walkthrough summary comment."* `review_details` stays off — schema describes it as the ignored-files/extra-context/suppressed-comments dump, exactly as you said.

---

.coderabbit.yaml
mattinannt: Applied in `bbcc045` — `approval_delay: 7`.

The argument is unanswerable, and the thread two comments down proves it live: while reviewing this very PR, CodeRabbit recorded a learning about `autofix`/`fix_ci` from javi's comment. Under `approval_delay: 0` that rule was already in force. A PR whose entire premise is "review configuration must be versioned and reviewable" was, at that moment, still growing unversioned review rules from comment threads.

Schema confirms both halves of your reading: 0 means "apply learnings immediately without approval", and `scope: auto` is "local for public repos".

---

.coderabbit.yaml
mattinannt: Both off in `fbfbb4a`; `resolve_merge_conflict` kept, as you suggested — mechanical, and the resolution lands in the diff like any other change.

This one I put to Matti rather than deciding it myself, since it removes a capability rather than fixing an error, and he took your recommendation. The schema backs the mechanism you described: autofix is enabled "via the 🪄 Autofix checkboxes under review comments **or** the `@coderabbitai autofix` command", so disabling it hides the checkboxes rather than just declining the command. `fix_ci` is the sharper one — the schema says `@coderabbitai fix-ci commit` "commits directly" to the branch.

You're right that the PR had already made this argument and not followed it through: `allow_non_org_members: false` was set precisely because these are write surfaces, and then the write surfaces were left on.

---

itsjavi (dismissed):
thanks, after refining server actions usage LGTM

itsjavi (commented):
Re-reviewed the delta since my last pass — `fbfbb4a…3ec6f09`, which is the `tone_instructions` length fix and nothing else. My earlier review covered the rest of the file and all six threads from it are resolved with fixes.

**Independently verified, and this is the part that matters**

The PR's own argument is that an invalid file fails closed and silently — CodeRabbit rejects the whole configuration, falls back to the dashboard, and nothing on the PR says so. So I checked the file rather than the claim: fetched the published schema (`www.coderabbit.ai/integrations/schema.v2.json`, draft 2020-12) and ran `ajv` over `.coderabbit.yaml` at `3ec6f09`.

- **Valid.** No errors, whole document.
- `tone_instructions` is **245** characters against the schema's `maxLength: 250` — the constraint is real and the fix clears it with the headroom the commit claims.
- The rest of the length audit in the commit message is exact: 19 `path_instructions`, longest 1369 of 20000; 2 custom checks, longest 1613 of 10000. Every other `maxLength` in the schema is on a field this file doesn't set.

Good call putting the limit in a comment above the key. And the honesty in the description about the earlier "schema validation" being a key-name walk is the right instinct — that's exactly why this survived six commits.

**One non-blocking note**

The rewrite is described as keeping the substance intact, and it nearly does, but one clause did change meaning. Before:

> if you cannot point at the **specific** line that breaks, **say so or** stay quiet.

After:

> if you cannot point at the line that breaks, stay quiet.

The old wording offered two ways out — flag it as uncertain, or drop it. The new one mandates silence. In practice that means an uncertain-but-real concern now goes unsaid rather than said with a caveat, which is a slightly different reviewer than the one you had. With 5 characters of headroom it isn't free to restore, but a few characters are recoverable elsewhere ("Do not ask for docstrings or JSDoc" → "No docstring or JSDoc requests" buys 4) if you'd rather keep the escape hatch. Entirely your call — not worth another round on its own.

**State of the PR:** all 18 checks green at `3ec6f09`, SonarQube and SonarCloud clean, no unresolved review threads, no conflicts with `main`.

**Needs human review** @itsjavi

- Unchanged from my first pass and inherent to the change: for OSS repos CodeRabbit reads configuration from the base branch, so nothing here — including the validity I just confirmed against the schema — proves the file is what CodeRabbit actually loads. The first post-merge QA step (`Configuration used` no longer reading `Repository UI`) is the only real confirmation, and it's worth doing before the dashboard is cleared rather than after.
- The dashboard cleanup is still an out-of-band action nobody is tracking in code. With `inheritance: false` and this file present, a later dashboard edit is silently ignored.

---

mattinannt:
Fixed in `7003ea4` — you're right, and "keeps the substance intact" was wrong in the commit message I wrote.

The old clause offered two ways out, flag it as uncertain or drop it; the rewrite mandated silence. That's the opposite of what the surrounding sentence is asking for — fewer *speculative* comments, not fewer *honest* ones — so an uncertain-but-real finding would have gone unsaid rather than said with a caveat. It was an accident of hitting a character budget, not a decision, which is why I'd rather not leave it in.

Restored without spending the headroom, using your suggestion plus one more:

- `Do not ask for docstrings or JSDoc` → `No docstring or JSDoc requests` (−4)
- `Prefer one well-evidenced comment over three` → `One well-evidenced comment beats three` (−6)

Still 245 of 250, and `ajv` against the published schema is clean. `say so or stay quiet` is back.

On your two standing notes:

- **Base-branch config.** Agreed, and it's the reason the first QA step is the footer check rather than the dashboard cleanup. Neither your ajv run nor mine proves what CodeRabbit loads — only a review on `main` does.
- **Dashboard cleanup isn't tracked in code.** Also agreed, and it can't be from here. It's in the PR's risk list with the ordering constraint attached (clear it only after the footer check passes), but that's a note in a description, not a mechanism. Worth a Linear ticket if it shouldn't rest on someone remembering.

One thing for whoever merges: your approval was dismissed by the `3ec6f09` push, and this push dismisses it again. The PR is green but blocked on a fresh approval.

---

.coderabbit.yaml
```diff
@@ -0,0 +1,558 @@
...
+    skillspector:
+      enabled: true
+    psscriptanalyzer:
+      enabled: true
+
+chat:
+  art: true
+  # Public repo: only org members may drive the bot with commands. Outside PRs
+  # are still reviewed automatically — this only stops non-members invoking
+  # `@coderabbitai`, including the autofix and fix_ci recipes enabled above.
```
itsjavi: **These two are off, not "enabled above".**

`finishing_touches.autofix.enabled` and `fix_ci.enabled` are both `false` (lines 280-283), turned off in `fbfbb4a` — and the comment on *that* block says so explicitly. This comment still describes them as enabled, so as written it tells the next reader that restricting chat to org members is what stands between an outsider and a bot-authored commit, when in fact those recipes aren't available to anyone. The setting itself is right; only the reasoning attached to it went stale when the block above changed.

```suggestion
  # `@coderabbitai` at all. (The autofix and fix_ci recipes are off above.)
```

Comment-only, nothing behaves differently. Flagging it because the argument for this whole PR is that the configuration should be readable and reviewable in the repo, and this is the one line in it that now says something untrue.

---

itsjavi (commented):
Re-reviewed the delta since my last pass — `3ec6f09…7003ea4`, which is two lines inside `tone_instructions` and nothing else in the file.

**Verified rather than taken from the commit message**

- The escape hatch is genuinely back: *"if you cannot point at the line that breaks, **say so or** stay quiet."* That was the clause the length squeeze dropped, and it's the half that mattered — an uncertain-but-real finding now gets said with a caveat rather than swallowed.
- Re-ran full schema validation at `7003ea4` (`ajv` against the published `schema.v2.json`, draft 2020-12): **valid**, whole document, zero errors.
- `tone_instructions` is **245** characters against `maxLength: 250` — the same length as at `3ec6f09`, so "the headroom is unchanged" is exact rather than approximate. The compensating shortenings (`Do not ask for docstrings or JSDoc` → `No docstring or JSDoc requests`, `Prefer one … over` → `One … beats`) pay for the restored words precisely.
- The diff is confined to the block scalar's second and third lines; no other key moved.

Worth saying plainly: a commit whose job is to correct its predecessor's "substance intact" claim is the right instinct, and it's the one kind of follow-up that's cheaper to land than to argue about.

**One small thing**, inline and non-blocking: a comment further down still describes two settings as enabled that this PR turned off.

**State of the PR:** all 16 checks green at `7003ea4` — Linters, Unit Tests, E2E, Build, Docker, Helm, Translation Keys, CodeQL, SonarQube and SonarCloud. No unresolved review threads on either mine or CodeRabbit's findings, and no conflicts with `main`.

**Needs human review** @itsjavi

- Unchanged from my earlier passes and inherent to the change: for OSS repos CodeRabbit loads configuration from the **base** branch, so nothing here — including the schema validity I just confirmed — proves this is the file CodeRabbit actually runs. The first post-merge QA step (`Configuration used` no longer reading `Repository UI`) is the only real confirmation, and it's worth doing *before* the dashboard is cleared rather than after.
- The dashboard cleanup is still an out-of-band action nobody is tracking in code. With `inheritance: false` and this file present, a later dashboard edit is silently ignored.

---

.coderabbit.yaml
mattinannt: Fixed in `16913db`. You're right, and the framing is the part I'd underweighted: the whole case for this PR is that configuration should be readable in the repo, so a comment asserting the opposite of a setting three hundred lines up is precisely the drift a dashboard would have hidden. It earned its own commit rather than a fold-in.

I went slightly beyond your suggestion, because dropping the clause would leave the reader wondering why the setting is there at all now that nothing can reach those recipes:

```yaml
  # Public repo: only org members may drive the bot with commands. Outside PRs are
  # still reviewed automatically — this only stops non-members invoking
  # `@coderabbitai` at all. The autofix and fix_ci write recipes are off above, so
  # this is defence in depth rather than the thing keeping them out of reach.
```

Same correction, and it keeps the reason for the setting surviving the recipes being disabled. Revalidated at `16913db`: ajv clean against the published schema, `tone_instructions` still 245/250, `prettier --check` passes.

Worth noting the sequence, since it's the third of these: the stale comment was introduced by `fbfbb4a`, the commit that turned the recipes off — I changed the setting and its own comment, and didn't grep for other places referencing it. Same shape as the `review_status` interaction I missed in `97089fc`. Both were found by re-reading rather than by any check, which is the argument for these passes.

On your two standing notes — both still stand and neither is mine to close. The base-branch behaviour means nothing on this PR proves what CodeRabbit loads, and the dashboard cleanup has no owner in code. I'd suggest a Linear ticket for the cleanup rather than leaving it to the PR description.

---

itsjavi (approved):
LGTM

PR #8781, opened 2026-08-07
refactor(email): treat package as source-consumed and reconcile Tailwind on v4

itsjavi:
## What & why

`packages/email` declared a `build` that only ran `tsc --noEmit`, yet other tasks waited on it through
`^build` in the turbo graph. Its entry points (`main`/`types`) target raw source and `apps/web`
consumes it that way, so this makes the source-consumed contract explicit instead of pretending to
produce output: the no-op script and its `turbo.json` entry are gone (the separate `typecheck` task
already covers the same ground), and `packages/email/README.md` documents the contract.

Being consumed as source has a consequence the package was not honouring: its runtime imports have to
be real `dependencies`. `clsx` and `tailwind-merge` back `src/lib/cn.ts`, and `@formbricks/types`
supplies `TSurveyElementTypeEnum`, used as a *value* in `email-utils.tsx` switch cases. All three
reach `apps/web` through the exported `ElementHeader` and only worked via pnpm hoisting.

**The `tailwindcss@3.4.19` pin was never the render engine.** `@react-email/components@1.0.12` pulls
`@react-email/tailwind@2.0.7`, which itself depends on `tailwindcss ^4.1.18` and imports the v4 CSS
entry points — emails already compile with v4. So the ticket's premise, that v4-only classes would
silently not render, does not hold; `text-shadow-lg`, `inset-shadow-sm`, `shadow-xs` and
`bg-linear-to-r` all render correctly today.

The pin's only real consumer was `prettier-plugin-tailwindcss`, through a `tailwindConfig` override in
a package-local `.prettierrc.cjs`. That override resolved relative to the **process CWD**, so from the
repo root it looked for a root `tailwind.config.js` that does not exist and printed
`Unable to load your Tailwind CSS v3 config`. Where it *did* resolve, it forced the plugin down its
**v3** code path — sorting email classes by v3 rules while react-email rendered them with v4. Removing
the override drops the warning and aligns sorting with the engine. (#8777 has since made
`pnpm format:check` a CI gate, so that warning now prints on every lint run until this lands.)

That re-sorted Tailwind classes in six files. The re-order is render-inert: all twelve emails are
byte-identical once `style=""` declaration order is normalised, no property is declared twice, no
shorthand/longhand pair changed relative order, and no `<style>` block changed.

Also drops `postcss` and `autoprefixer`, which nothing in the package used, and adds the Vitest
coverage the package never had. Those tests surfaced two defects in `mock-translate.ts`: stale
`_par1`/`_par2` keys no template references, and missing `imprint`/`privacy_policy` keys that made the
legal footer render `emails.imprint` / `emails.privacy_policy` as its link labels. Both now mirror
`en-US.json` verbatim.

To be precise about blast radius: `mockT` is only a **default parameter** (`t = mockT`), and `apps/web`
always passes a real `t` from `getTranslate(locale)`. So no production email was ever affected — this
was broken copy in the `email dev` preview server that developers use to eyeball templates, which is
also why nothing caught it. Fixing it is what makes the new tests able to assert on rendered copy at
all.

## Linear ticket

Fixes ENG-1687

## How this was tested

- `pnpm --filter @formbricks/email test` — new suite, **PASS** (41 tests). Covers all 12 renderers
  (complete HTML document, no raw translation keys leaking, Tailwind compiled to inline styles),
  translation threading including the `responseCount > 1` CTA branch and its fallback, the
  legal-footer conditional, and a contract test pinning the v4 render engine.
- Mutation-checked rather than assumed: dropping a translation key fails the raw-key guard, and the
  guard was confirmed to **fail before** the `imprint`/`privacy_policy` fix (13 failed) and pass after.
- `cd apps/web && vitest run modules/email` — **46/46 PASS**, including the normalised-HTML fixture
  comparison, so the class re-order is proven not to change rendered output.
- Render-equivalence proof for the re-order: baseline `packages/email` reconstructed out-of-tree via
  `git archive`, all 12 emails (plus a legal-props variant) rendered from both trees, every render
  identical after normalising declaration order. `twMerge` output identical for both orderings of the
  `ElementHeader` strings across seven caller-`className` variants.
- `pnpm format:check` — **PASS** (repo-wide, after merging `main` to pick up #8777). The
  `Unable to load your Tailwind CSS v3 config` warning goes **1 → 0**, since this PR deletes the
  config that caused it.
- Sorting proven still active, not merely silent: a deliberately mis-ordered `className`
  (`text-black break-all text-sm p-6 m-0`) is rewritten to `m-0 p-6 text-sm break-all text-black`.
- `pnpm --filter @formbricks/email typecheck` and `lint` — **PASS**. Repo-wide `pnpm typecheck`
  **25/25** and `pnpm build` **12/12** after the `turbo.json` change.
- `turbo run build --dry` — `@formbricks/email#build` is now `<NONEXISTENT>`, the same harmless no-op
  node as `@formbricks/types#build`. No reference to it exists in `.github/workflows/**`, `docker/**`,
  or any script.
- Lockfile: the 30 removed entries are all Tailwind v3's dependency tree. Nothing added beyond the
  Vitest / `clsx` / `tailwind-merge` / `@formbricks/types` importer changes.

Not run: Playwright. This changes no product UI — email templates render identical HTML.

## Breaking changes

None — internal package wiring only.

## QA / Test Plan

This is a **no-visible-change PR**: every step below should look exactly as it did before. Tailwind
class order changed in six email templates, so the whole point of QA is confirming the rendered
emails are unaffected. Open each one in a real mail client, not just a preview pane.

**How to test**

- [ ] Request a password reset (login → "Forgot password") → email arrives and renders with the
      centred logo, the rounded white card on a light grey background, correct spacing, and a working
      CTA button. Identical to how it looks on `main`.
- [ ] Click the plain-text fallback link in that email → it wraps within the card instead of
      overflowing the card edge. (`break-all` was re-ordered on this link.)
- [ ] Invite a member (Organization → Members → invite) → invite email renders correctly with the
      invitee's name in the heading and the inviter's name in the body; CTA button works.
- [ ] Accept that invite → the inviter receives the "new organization member" email, rendered
      correctly.
- [ ] Complete a survey response with response notifications enabled → the response-finished email
      renders; long and multi-line answers still wrap inside the card rather than overflowing
      (`whitespace-pre-wrap break-words` was re-ordered here); the CTA reads "View N more responses",
      or "View survey summary" when this is the only response.
- [ ] Trigger a survey follow-up email carrying hidden fields and variables → values still wrap
      correctly and headings keep their weight and line spacing.
- [ ] Verify a new signup email address → verification email renders correctly.
- [ ] With `IMPRINT_URL`, `PRIVACY_URL` and `IMPRINT_ADDRESS` set, trigger any of the above → the
      footer shows the address plus **Imprint** and **Privacy Policy** links. Unset them → no legal
      footer appears. Behaviour is unchanged by this PR; it is listed because the footer branch is the
      one the new tests newly cover.

**Preconditions / test data**

- SMTP configured (or a mail catcher such as Mailpit) so emails are actually delivered.
- An organization where you can invite a second member, and a survey with response notifications on.
- For the footer step: `IMPRINT_URL`, `PRIVACY_URL`, `IMPRINT_ADDRESS` set, then unset.
- Applies to both Cloud and self-hosted; no plan or entitlement gating.
- Comparing against a `main` deployment side by side makes any visual delta obvious.

**Risks & regressions**

- Highest-risk area is **email visual rendering**, because Tailwind class order changed in six
  template/source files. The automated proof is strong (all twelve emails byte-identical once
  declaration order is normalised, no duplicated property, no shorthand/longhand pair reordered), but
  a human eyeball in a real client is the cheap confirmation.
- `ElementHeader` is consumed by `apps/web`, so the response-finished and follow-up emails are the
  best places to look for a heading or subheading spacing/weight change.
- The dependency reclassification (`clsx`, `tailwind-merge`, `@formbricks/types` → `dependencies`)
  changes install topology. A clean install followed by a successful `apps/web` build is the check;
  these previously resolved only through pnpm hoisting.
- Removing the `build` script changes the turbo graph. A green repo-level build is the check.
- Not user-facing, but worth a developer's glance: `pnpm --filter @formbricks/email dev` (port 3456)
  now shows real copy in the invite, invite-accepted and legal-footer previews, which previously
  displayed raw translation keys.

- none


mattinannt (commented):
Reviewed the full diff and reproduced the checks locally rather than just reading the description:

- `pnpm --filter @formbricks/email test` → 41/41 passing (new `render.test.ts`)
- `pnpm --filter @formbricks/email typecheck` → clean
- `pnpm --filter @formbricks/email lint` → clean
- `pnpm exec prettier --check packages/email` → clean

Specifically verified the two claims most likely to hide a real bug:

1. **`mock-translate.ts` key rename** (`invite_accepted_email_text_par1/2` → `invite_accepted_email_text`, etc.): checked against `apps/web/locales/en-US.json` — the email templates already call the new key names, so this PR is fixing a stale mock file, not introducing a mismatch.
2. **Runtime deps moved from `devDependencies` to `dependencies`** (`clsx`, `tailwind-merge`, `@formbricks/types`): confirmed `cn.ts` and `email-utils.tsx` genuinely import these at module scope in code reachable from `ElementHeader`/`email-utils`, so the "pnpm hoisting was masking a real dependency gap" claim holds.

No dangling references to the removed `packages/email/tailwind.config.js` / `@formbricks/email#build` turbo task elsewhere in the repo, and `apps/web`'s consumption of `@formbricks/email` as source is unaffected.

Nothing to flag — looks correct and well-tested.

---

mattinannt (approved):

PR #8782, opened 2026-08-07
refactor(storybook): single Tailwind v4 CSS entry and document the workspace CSS invariant

itsjavi:
## What & why

Tailwind v4 detects sources from the consuming app's own root and never descends into `node_modules`
— which is exactly where every `@formbricks/*` workspace package is linked. App styling therefore
rests on an unwritten invariant: consumed workspace UI packages ship their own compiled CSS. Nothing
stated it, so nothing stopped someone from breaking it.

Verified before documenting. The oxide scanner run for `apps/web` reads 2635 files, **none** under
`node_modules` and none outside `apps/web`; and five utilities that exist only in `survey-ui` source
(`aspect-[162/97]`, `[--cell-size:2rem]`, `h-[stretch]`, `grid-cols-[0_1fr]`,
`[font-size:var(--fb-input-font-size)]`) appear **zero** times in the compiled `apps/web` stylesheet.
The invariant is now recorded where it is load-bearing: a header comment on
`apps/web/modules/ui/globals.css` and a new section in `AGENTS.md` naming how each consumed package
ships its styles — `@formbricks/surveys` as a prebuilt bundle under `public/js/`,
`@formbricks/survey-ui` via its `./styles` export, `@formbricks/email` inlined at render time by
`@react-email/tailwind`.

`apps/storybook` mixed a v3-style `tailwind.config.js` with the v4 engine. Nothing loaded it — v4 does
not auto-discover JS configs and no `@config` referenced it. Moving it aside and rebuilding produced a
**byte-identical** stylesheet, so it is deleted rather than migrated.

Storybook's Tailwind actually arrived through `.storybook/preview.ts` reaching directly into
`packages/survey-ui/src/styles/globals.css`. Both are replaced by a single CSS-first entry at
`apps/storybook/src/index.css` that imports survey-ui's stylesheet and declares `@source` globs for
the app's own files. That also resolves a pre-existing broken `./index.css` import in `src/main.tsx`,
so `vite build` succeeds again.

Worth stating plainly, since an earlier draft of this change got it wrong: those `@source` globs are
**not** what makes the storybook chrome work. Tailwind v4 automatic source detection roots at the
storybook project root and scans it whole, so `.storybook/**` and `src/**` are already covered —
removing the globs produces byte-identical CSS. They are kept as a written-down declaration of intent
that survives someone moving the Vite root, and the comment in the file says exactly that rather than
implying the coverage would otherwise be accidental.

## Linear ticket

Fixes ENG-1692

## How this was tested

- `pnpm --filter storybook build-storybook` — **PASS**, and the built stylesheet is **byte-identical**
  to the pre-change baseline (md5 `db896b6b37774f4ceac26eae5f4ed7a3`, 58513 bytes, `cmp` clean,
  selector-set diff 0 added / 0 removed). All classes the `#fbjs` preview decorator needs are present
  (`min-h-screen`, `antialiased`, `font-sans`, `bg-background`, `text-foreground`, `w-full`, `p-4`,
  `h-full`), and survey-ui component styling is unchanged.
- Deleting `tailwind.config.js` was proven safe before doing it: the file was moved aside and storybook
  rebuilt, producing the same byte-identical CSS. It was dead code.
- `apps/web` compiled CSS is **unchanged** — baseline vs current `apps/web/modules/ui/globals.css`
  compiled through `@tailwindcss/postcss` gives byte-identical output (md5
  `89ec00cc8ac9d701e0fefac20d8fe40f`), since the added comment is stripped.
- `npx vite build` in `apps/storybook` — **PASS**. Confirmed it fails with
  `Could not resolve "./index.css" from "src/main.tsx"` when the new file is moved away, so the fix is
  real and not incidental.
- `pnpm --filter storybook typecheck` and `lint` (`eslint . --max-warnings 0`) — **PASS**.
- `pnpm format:check` — **PASS** (repo-wide, after merging `main` to pick up #8777).

Not run: Playwright. This changes no product UI — `apps/web`'s compiled CSS is byte-identical.

## Breaking changes

None — docs and a storybook-only build entry.

## QA / Test Plan

**How to test**

- [ ] Load the app and click through a few dense surfaces (survey editor, response summary, settings)
      → styling is visually identical to `main`. This PR asserts `apps/web` CSS is byte-identical, so
      any visible difference is a bug.
- [ ] Open a survey (link survey or in-app preview) → survey rendering and theming are unchanged; the
      `#fbjs`-scoped styles still apply and nothing leaks into the surrounding page.
- [ ] Run Storybook and open several `survey-ui` stories → components render with full styling, the
      light background and padding from the preview wrapper are present, and no story appears
      unstyled.
- [ ] Toggle the a11y addon panel on a story → still functions as before.

**Preconditions / test data**

- Storybook checks need it running locally (`pnpm --filter storybook storybook`, port 6006); there is
  no deployed surface for them. Chromatic also builds this app, so its diff is a second signal.
- No accounts, seed data, feature flags, or entitlements required. Applies equally to Cloud and
  self-hosted.

**Risks & regressions**

- The realistic failure mode is a **missing utility class** in Storybook — a component rendering
  unstyled or half-styled because a class stopped being generated. The byte-identical stylesheet makes
  this very unlikely, but scanning a handful of stories is the cheap confirmation.
- `apps/web` is documentation-only here (a CSS comment plus `AGENTS.md`), with compiled output proven
  unchanged. Any product styling regression would indicate something outside this diff.
- Nothing changes at runtime for end users; there is no server, data, or API surface involved.

- none


apps/web/modules/ui/globals.css
```diff
@@ -1,3 +1,11 @@
+/* Tailwind v4 automatic source detection is rooted at this app (apps/web) and never
```
itsjavi: @claude  remove this comment block which is redundant

mattinannt: Removed in 120b69c — the block duplicated what's now in `AGENTS.md` under "Tailwind &amp; Workspace Package CSS".

---

AGENTS.md
```diff
@@ -29,6 +29,34 @@ The `@formbricks/surveys` package is pre-compiled (Vite → UMD + ESM) and the b
 - The browser also caches the UMD bundle (`surveys.umd.cjs`) served from `public/js/`. After rebuilding, do a **hard refresh** (Cmd+Shift+R / Ctrl+Shift+R) or disable the browser cache via DevTools to pick up the new bundle.
 - If changes still don't appear, restart the Next.js dev server (`pnpm dev`).
 
+### Tailwind & Workspace Package CSS
+
+Tailwind v4 detects sources starting from the consuming app's own root (`apps/web` for the Next.js
+PostCSS build, the Vite root for `apps/storybook`) and **never descends into `node_modules`** — which
```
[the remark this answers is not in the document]
itsjavi: Not changing this, on both halves of the finding.

**The `apps/web/modules/ui/globals.css` half is moot.** That comment block was removed in 120b69c at @mattinannt's request. The file now starts at `@import "tailwindcss";` with no prose to qualify, and it is no longer part of this PR's diff.

**The `AGENTS.md` half is mechanically right but doesn't make the sentence wrong.** Yes — `@tailwindcss/postcss` is configured `{}` and its `base` defaults to `process.cwd()`. But every build path in this repo runs `next build` with cwd = `apps/web`:

- `pnpm build` / `pnpm dev` → `turbo run build|dev`, and Turborepo executes each package's script in that package's own directory.
- CI uses `pnpm build --filter=@formbricks/web...` (`.github/workflows/e2e.yml:151`, `integration-tests.yml:125`) — same thing.
- There is no invocation from the repo root, and there isn't really a path to one: `next.config` and `postcss.config.js` both live in `apps/web`.

So "`apps/web` for the Next.js PostCSS build" describes what actually happens, which is what this guide is for. Adding "when run from `apps/web`" would hedge against an invocation that doesn't exist.

**And the `base: "."` alternative would be a no-op.** `"."` is itself resolved against `process.cwd()`, so it pins exactly the default. A real pin would have to be `base: __dirname`, which is a build-config change for a hypothetical this repo doesn't have — out of scope for a docs-and-storybook PR.

Leaving the thread open for a human call rather than resolving it.

---

mattinannt (approved):

PR #8785, opened 2026-08-07
docs: clarify custom subpath storage setup

BhagyaAmarasinghe:
## What & why

The custom-subpath guide could be read as requiring a custom RustFS or MinIO image, even though `BASE_PATH` only affects the Formbricks web build. This update clarifies that customers must build the Formbricks web image, keep the S3 API at the root of a dedicated endpoint, and avoid arbitrary proxy prefixes that can invalidate SigV4 requests. It also documents the distinct app-root and auth-endpoint URL values, distinguishes Formbricks routes from external webhook destinations, and links directly to the file-upload configuration.

The guide now also documents the current browser auth-client limitation: login and sign-up requests still target the domain-root `/api/auth` path. It no longer presents onboarding/login as a successful verification step or suggests that changing server URL variables fixes that client-side path.

## Linear ticket

No Linear ticket; this is a documentation follow-up to a self-hosting configuration discussion.

## How this was tested

- `pnpm exec prettier --check docs/self-hosting/configuration/custom-subpath.mdx` ✅
- `git diff --check` ✅
- Confirmed the linked `docs/self-hosting/configuration/file-uploads.mdx` page exists.
- Confirmed Better Auth preserves a configured server path, so auth URLs must include `/custom-path/api/auth`.
- Confirmed the unconfigured browser client resolves its auth base to `https://yourdomain.com/api/auth`, without the custom prefix.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open the Custom Subpath documentation → the requirements identify the custom image as the Formbricks web image.
- [ ] Read the known limitation → it states that browser login/sign-up requests currently omit the custom prefix and that URL environment variables do not fix the client request path.
- [ ] Check the environment example → `WEBAPP_URL` ends at the app subpath, while `NEXTAUTH_URL` and `BETTER_AUTH_URL` use the full `/api/auth` endpoint for server-side callbacks and discovery.
- [ ] Check the URL guidance → external webhook targets are not instructed to inherit the Formbricks subpath.
- [ ] Read “Verify the deployment” → it checks prefixed pages/assets without claiming onboarding or login works.
- [ ] Read “Configure object storage separately” → the page recommends a dedicated root S3 endpoint and explains the SigV4 limitation.
- [ ] Select “File Uploads Configuration” → the file-upload configuration page opens.

**Preconditions / test data**

- Documentation preview or published documentation.

**Risks & regressions**

- Documentation-only change; verify the limitations, URL examples, internal link, and storage terminology remain accurate.

- none

### DIFF
--- a/docs/self-hosting/configuration/custom-subpath.mdx
+++ b/docs/self-hosting/configuration/custom-subpath.mdx
@@ -19,9 +19,24 @@ assets, and sign-in redirects honor the prefix.
 ### Requirements and limitations
 
 - `BASE_PATH` must be present during `pnpm build`; changing it afterward requires a rebuild.
-- Official Formbricks Docker images do **not** accept this flag for technical reasons, so you must build your own image.
+- Next.js inlines the base path into the client bundles, so setting `BASE_PATH` on an already-built official Formbricks
+  image has no effect. You must build your own Formbricks web image with the required value.
 - All Formbricks URLs (`WEBAPP_URL`, `NEXTAUTH_URL`, webhook targets, OAuth callbacks, MCP clients, etc.) need the same prefix.
 - Your proxy must rewrite `/custom-path/*` to the Formbricks container while keeping the prefix visible to clients.
+
+### Configure object storage separately
+
+`BASE_PATH` applies only to the Formbricks web application. It does not add the same prefix to RustFS, MinIO, or
+another S3-compatible storage service, and the custom image described below is a Formbricks image—not a custom
+storage image.
+
+Keep the S3 API at the root of a dedicated endpoint such as `https://files.example.com` and set `S3_ENDPOINT_URL`
+to that endpoint. Do not route it through an arbitrary prefix such as `https://example.com/custom-path/storage`:
+the request path is part of the AWS Signature Version 4 signature, so proxy path rewriting can invalidate signed and
+presigned requests. Provider settings that expose a management console under a subpath do not change the S3 API
+base path.
+
+For the complete storage and CORS setup, see [File Uploads Configuration](/self-hosting/configuration/file-uploads).
 
 ### Configure environment variables
 
--- a/docs/self-hosting/configuration/custom-subpath.mdx
+++ b/docs/self-hosting/configuration/custom-subpath.mdx
@@ -21,7 +21,10 @@ assets, and sign-in redirects honor the prefix.
 - `BASE_PATH` must be present during `pnpm build`; changing it afterward requires a rebuild.
 - Next.js inlines the base path into the client bundles, so setting `BASE_PATH` on an already-built official Formbricks
   image has no effect. You must build your own Formbricks web image with the required value.
-- All Formbricks URLs (`WEBAPP_URL`, `NEXTAUTH_URL`, webhook targets, OAuth callbacks, MCP clients, etc.) need the same prefix.
+- Formbricks-owned public URLs need the prefix, but each variable expects a different path: `WEBAPP_URL` uses the
+  app root, while `NEXTAUTH_URL` and optional `BETTER_AUTH_URL` use the full auth endpoint under
+  `/custom-path/api/auth`. External webhook target URLs do not need the prefix unless they point to a Formbricks
+  endpoint.
 - Your proxy must rewrite `/custom-path/*` to the Formbricks container while keeping the prefix visible to clients.
 
 ### Configure object storage separately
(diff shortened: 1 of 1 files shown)


xernobyl (dismissed):

xernobyl (commented):
The `NEXTAUTH_URL` / `BETTER_AUTH_URL` correction is right, and worth spelling out because it's the load-bearing part of the diff: Better Auth's `withPath` returns a configured URL verbatim if it carries *any* path, so the previously-documented `https://yourdomain.com/custom-path` made the auth base `/custom-path` with no `/api/auth` appended and every endpoint 404'd. `getAuthIssuerUrl`'s `endsWith` guard (`apps/web/modules/auth/lib/oauth-urls.ts:23`) keeps the OAuth issuer identical under both values, so there's no discovery regression either way.

I verified the rest of the new content: `S3_ENDPOINT_URL` is the real variable (`.env.example:190`), the SigV4 path-in-signature reasoning is correct, `apps/web/Dockerfile:42` does expose `ARG BASE_PATH`, the file-uploads link resolves and is in `docs.json` nav, and the `.well-known/oauth-authorization-server` route exists. Dropping "email links" is fine — those derive from `WEBAPP_URL`, so there's nothing to register.

One substantive comment inline, plus two nits:

- `charts/formbricks/templates/secrets.yaml:24` hardcodes `NEXTAUTH_URL = webappUrl`, which now contradicts the documented split. Not broken — a k8s `env` entry beats `envFrom`, so `deployment.env` overrides it, and the chart has no `BASE_PATH` knob anyway — but a Helm reader gets no hint. Optional sentence.
- Lines 36–37 switch to `files.example.com` / `example.com/custom-path/storage` while the env block and verification steps use `yourdomain.com` (`file-uploads.mdx` uses `files.yourdomain.com`).

docs/self-hosting/configuration/custom-subpath.mdx
```diff
@@ -19,23 +19,41 @@ assets, and sign-in redirects honor the prefix.
...
 
 - `BASE_PATH` must be present during `pnpm build`; changing it afterward requires a rebuild.
-- Official Formbricks Docker images do **not** accept this flag for technical reasons, so you must build your own image.
-- All Formbricks URLs (`WEBAPP_URL`, `NEXTAUTH_URL`, webhook targets, OAuth callbacks, MCP clients, etc.) need the same prefix.
+- Next.js inlines the base path into the client bundles, so setting `BASE_PATH` on an already-built official Formbricks
+  image has no effect. You must build your own Formbricks web image with the required value.
+- Formbricks-owned public URLs need the prefix, but each variable expects a different path: `WEBAPP_URL` uses the
+  app root, while `NEXTAUTH_URL` and optional `BETTER_AUTH_URL` use the full auth endpoint under
+  `/custom-path/api/auth`. External webhook target URLs do not need the prefix unless they point to a Formbricks
+  endpoint.
```
xernobyl, line 27: This bullet is accurate for the server, but it doesn't hold in the browser — so the page now reads as a complete recipe for a login that still won't work under a subpath.

`apps/web/modules/auth/lib/auth-client.ts:10` sets neither `baseURL` nor `basePath`. In the browser, `getClientConfig` → `getBaseURL` falls through to the window branch — `withPath(window.location.origin, "/api/auth")` → `https://yourdomain.com/api/auth`, **prefix dropped**. The `NEXTAUTH_URL` fallback that would carry the prefix (`resolvePublicAuthUrl`) is explicitly gated on `typeof window === "undefined"`, and `BETTER_AUTH_URL` isn't inlined into the client bundle (`next.config.mjs:461` exposes only `NEXTAUTH_URL` — which is what its "TODO: proper solution for the base path" refers to). So `apps/web/modules/auth/login/components/login-form.tsx:123` POSTs to `https://yourdomain.com/api/auth/sign-in/email`, which on a subpath deployment hits whatever else owns the root.

This is a pre-existing gap in the product, not something the diff introduced, and the "under internal review" note at the top softens it. The concrete problem is the troubleshooting bullet further down (line 105–106): *"If you cannot log in, double-check that `WEBAPP_URL`, `NEXTAUTH_URL`, and `BETTER_AUTH_URL` include the prefix"* now points operators at the one thing that isn't the cause.

Suggest not widening this PR — add a known-limitation line here, and let's track wiring `basePath` into `createAuthClient` on its own ticket.

BhagyaAmarasinghe: addressed

xernobyl (approved):
Re-reviewed at `a70b2f9`. The P2 is addressed properly — and addressed the honest way, by documenting the limitation rather than papering over it. LGTM.

Verified against the code:

- The "Known limitation" bullet matches the actual mechanism: `authClient` (`apps/web/modules/auth/lib/auth-client.ts:10`) sets neither `baseURL` nor `basePath`, so in the browser Better Auth resolves `window.location.origin + "/api/auth"`. And the claim that the env vars don't correct it is right — `resolvePublicAuthUrl`'s `NEXTAUTH_URL` branch is gated on `typeof window === "undefined"`, and `BETTER_AUTH_URL` isn't inlined into the client bundle.
- Dropping "sign-in redirects honor the prefix" from the intro was the right call — that was the sentence that made the page over-promise.
- The rewritten troubleshooting bullet is accurate: those variables genuinely do matter for the server-side issuer and callbacks (`getAuthIssuerUrl` in `apps/web/modules/auth/lib/oauth-urls.ts`), just not for the browser request path.
- New verification step 1 resolves — the login route is `apps/web/app/(auth)/auth/login/page.tsx`, so `/custom-path/auth/login` is correct.

Two optional things, neither blocking:

- "login and sign-up requests" is slightly narrower than reality — everything routed through `authClient` is affected (session refresh, 2FA, forgot-password, social sign-in). The bullet's next sentence generalizes correctly, so this only matters if a reader takes the first sentence as an exhaustive list.
- Still open from the last pass, both cosmetic: `charts/formbricks/templates/secrets.yaml:24` hardcodes `NEXTAUTH_URL = webappUrl` (overridable via `deployment.env`, so nothing is broken), and lines 36–37 use `example.com` while the rest of the page uses `yourdomain.com`.

E2E, SonarQube and Docker Build are still running on this head; everything else is green, including both Mintlify checks. I'll file the auth-client `basePath` fix as its own ticket so this doesn't stay documentation-only.

PR #8796, opened 2026-08-10
fix: canonicalize client IP resolution

BhagyaAmarasinghe:
## What & why

Client IP resolution diverged between Formbricks and Better Auth. Better Auth received a multi-value `X-Forwarded-For` header it could not trust and collapsed unauthenticated traffic into one shared rate-limit bucket, while Formbricks used unvalidated raw tokens that could vary by IPv6 spelling, host rotation, or Azure gateway source port.

This change makes Proxy the single client-IP trust boundary:

- select the configured trusted XFF hop exactly once and fail closed for missing, short, or invalid chains
- validate and normalize IPv4/IPv6, strip supported socket ports, convert mapped IPv6, and group IPv6 identities by `/64`
- overwrite or remove the private request-only `x-formbricks-client-ip` header before forwarding
- make Formbricks and Better Auth consume only that canonical identity
- keep asset exclusions segment-bound so asset-like dynamic route prefixes cannot bypass the private-header overwrite
- emit distinct, value-free diagnostics when XFF is absent but untrusted forwarding headers are present
- forward request headers into server-side sign-up and document the existing proxy topology requirement

No rate-limit thresholds, database schema, or configuration defaults change.

## Linear ticket

Fixes ENG-2097

## How this was tested

- `pnpm --filter @formbricks/web exec vitest run lib/utils/client-ip.test.ts proxy.test.ts modules/core/rate-limit/helpers.test.ts modules/auth/signup/actions.test.ts modules/traefik-auth/service.test.ts` — 6 files / 115 tests passed
- `pnpm --filter @formbricks/web typecheck` — passed after generating route types
- affected-file ESLint — passed
- targeted Prettier check and `git diff --check` — passed
- pre-commit lint-staged formatting and ESLint — passed
- fresh PR CI — 19 checks passed, including Better Auth integration tests, full unit tests, linters, web build, E2E, Docker build validation, CodeQL, and Sonar; 1 deployment check intentionally skipped

## Breaking changes

None — no public API/SDK contract, database migration, new environment variable, rate-limit threshold, or configuration-default change.

## QA / Test Plan

**How to test**

- [ ] Send repeated failed email sign-ins while changing only client-prepended XFF entries → all attempts for the same real client reach the existing 429 threshold in one bucket.
- [ ] Send the same IPv4 client through Azure-style `IP:port` values while changing the port → requests still reach 429 in one bucket.
- [ ] Rotate IPv6 host addresses within one `/64` → requests share one bucket; use a client in a different `/64` → it receives a separate bucket.
- [ ] Exhaust the limit for one real client, then sign in from a second real client → the second client is not locked out by the first.
- [ ] Complete a successful sign-in and submit a response → session, audit-log, and response metadata use the same canonical client identity.
- [ ] Send a malformed or shorter-than-configured forwarding chain with a spoofed `x-formbricks-client-ip` header → the spoofed value is removed and requests use the fail-closed shared bucket.
- [ ] Send only `x-real-ip` or `cf-connecting-ip` without XFF → the private header remains absent and diagnostics identify the missing trusted chain without logging header values.
- [ ] Load login pages and static assets, then exercise asset-like dynamic routes → auth/client requests traverse Proxy and assets continue loading normally.

**Preconditions / test data**

- A deployment with Redis-backed rate limiting enabled.
- `TRUSTED_PROXY_HOP_COUNT` set to the exact number of controlled proxies/load balancers.
- Two test clients with distinct real IPs, plus the ability to vary XFF values at the edge.
- A test account and access to the deployment's session/audit and response metadata views.

**Risks & regressions**

- An incorrect trusted-hop count can still choose an attacker-prepended entry when the forged chain is long enough; operators must configure the real topology.
- IPv6 values persisted in sessions, responses, and audit logs now use canonical `/64` network addresses.
- Auth routes, client response routes, and server actions depend on the Proxy matcher continuing to cover dynamic requests.

- No migration or new environment variable.
- Confirm the existing `TRUSTED_PROXY_HOP_COUNT` matches the deployed proxy topology before rollout.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -366,7 +366,9 @@ ENTERPRISE_LICENSE_KEY=
 # only the rightmost N entries were added by infrastructure you control. Defaults to 1, which is
 # correct for a single nginx/Traefik/Envoy in front of the app. Set it to 2 if Cloudflare (or another
 # CDN) also fronts that proxy, and so on. Setting it higher than your real topology lets callers spoof
-# their address again by prepending entries; setting it to 0 believes no forwarding header at all,
+# their address again by prepending entries. A chain shorter than this setting, or a malformed selected
+# hop, fails closed. IPv4 ports appended by gateways are removed and IPv6 addresses are grouped by /64.
+# Setting it to 0 believes no forwarding header at all,
 # which means IP-based rate limiting and the IP recorded on responses and audit logs cannot identify
 # individual clients. cf-connecting-ip is never trusted — Cloudflare also populates X-Forwarded-For.
 # TRUSTED_PROXY_HOP_COUNT=1
--- a/.env.example
+++ b/.env.example
@@ -367,8 +367,8 @@ ENTERPRISE_LICENSE_KEY=
 # correct for a single nginx/Traefik/Envoy in front of the app. Set it to 2 if Cloudflare (or another
 # CDN) also fronts that proxy, and so on. Setting it higher than your real topology lets callers spoof
 # their address again by prepending entries. A chain shorter than this setting, or a malformed selected
-# hop, fails closed. IPv4 ports appended by gateways are removed and IPv6 addresses are grouped by /64.
-# Setting it to 0 believes no forwarding header at all,
+# hop, fails closed. Ports appended by gateways are removed and IPv6 addresses are grouped by /64.
+# Setting it to 0 trusts no forwarding headers at all,
 # which means IP-based rate limiting and the IP recorded on responses and audit logs cannot identify
 # individual clients. cf-connecting-ip is never trusted — Cloudflare also populates X-Forwarded-For.
 # TRUSTED_PROXY_HOP_COUNT=1
--- a/apps/web/lib/utils/client-ip.ts
+++ b/apps/web/lib/utils/client-ip.ts
@@ -22,12 +22,15 @@ const VALID_DECIMAL_PORT = /^[1-9]\d{0,4}$/;
 const BRACKETED_ADDRESS = /^\[([^\[\]]+)\](?::([^:]+))?$/;
 const IPV4_SOCKET = /^([^:]+):(\d+)$/;
 
-type ClientIpWarningReason = "disabled" | "invalid-selected-hop" | "short-chain";
+type ClientIpWarningReason = "disabled" | "invalid-selected-hop" | "missing-chain" | "short-chain";
 
 const clientIpWarningMessages: Record<ClientIpWarningReason, string> = {
   disabled:
     "Client IP resolution is disabled because TRUSTED_PROXY_HOP_COUNT is 0 while forwarding headers " +
     "are present. IP-based limits and captured IP metadata will use the shared untrusted identity.",
+  "missing-chain":
+    "Client IP resolution failed because X-Forwarded-For is absent. Configure the proxy to append it; " +
+    "x-real-ip and cf-connecting-ip are not trusted client identity sources.",
   "short-chain":
     "Client IP resolution failed because X-Forwarded-For has fewer entries than " +
     "TRUSTED_PROXY_HOP_COUNT. Verify that every configured proxy appends to the forwarding chain.",
@@ -106,7 +109,7 @@ export const resolveClientIp = (headersList: Headers, hopCount: number): string 
 
   const xForwardedFor = headersList.get("x-forwarded-for");
   if (xForwardedFor === null) {
-    if (hasAnyForwardingHeader(headersList)) warnAboutClientIp("short-chain");
+    if (hasAnyForwardingHeader(headersList)) warnAboutClientIp("missing-chain");
     return null;
   }
 
--- a/apps/web/lib/utils/client-ip.ts
+++ b/apps/web/lib/utils/client-ip.ts
@@ -1,107 +1,140 @@
+import { isValidIP, normalizeIP } from "@better-auth/core/utils/ip";
 import { headers } from "next/headers";
 import { logger } from "@formbricks/logger";
-import { TRUSTED_PROXY_HOP_COUNT } from "@/lib/constants";
 
-/**
- * Returned when no forwarding header may be believed, so the client is genuinely unidentifiable.
- *
- * Next 16 exposes no socket peer address to route handlers or server actions (`NextRequest.ip` was
- * removed and `headers()` carries only HTTP headers), so with no trusted proxy there is nothing else to
- * fall back to.
- */
+/** Private request header written by Proxy after the trusted forwarding hop has been resolved. */
+export const FORMBRICKS_CLIENT_IP_HEADER = "x-formbricks-client-ip";
+
+/** Collapse IPv6 identities to a stable network prefix before rate limiting or persistence. */
+export const CLIENT_IP_IPV6_SUBNET_PREFIX = 64;
+
+/** Better Auth must consume exactly the same private request identity as the rest of Formbricks. */
+export const BETTER_AUTH_IP_ADDRESS_CONFIG = {
+  ipAddressHeaders: [FORMBRICKS_CLIENT_IP_HEADER],
+  ipv6Subnet: CLIENT_IP_IPV6_SUBNET_PREFIX,
+};
+
+/** Stable downstream identity when Proxy could not establish a trustworthy client address. */
 export const UNTRUSTED_CLIENT_IP = "untrusted-client-ip";
 
-/**
- * Throttle window for the misconfiguration warning. Time-based rather than once-per-process: a warning
- * that fires only on the first request goes silent for the life of a long-running deployment, so an
- * ongoing misconfiguration disappears from monitoring after one line. Raised by CodeRabbit on #8680.
- */
-const UNTRUSTED_IP_WARNING_INTERVAL_MS = 10 * 60 * 1000;
+const CLIENT_IP_WARNING_INTERVAL_MS = 10 * 60 * 1000;
+const VALID_DECIMAL_PORT = /^[1-9]\d{0,4}$/;
+const BRACKETED_ADDRESS = /^\[([^\[\]]+)\](?::([^:]+))?$/;
+const IPV4_SOCKET = /^([^:]+):(\d+)$/;
 
-let lastUntrustedIpWarningAt: number | null = null;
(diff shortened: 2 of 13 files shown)


apps/web/lib/utils/client-ip.ts
```diff
@@ -1,107 +1,140 @@
...
+  } catch (error) {
+    logger.error(error, "Failed to get headers in getClientIpFromHeaders");
     return UNTRUSTED_CLIENT_IP;
   }
 
-  return resolveClientIp(headersList, TRUSTED_PROXY_HOP_COUNT);
+  const internalClientIp = headersList.get(FORMBRICKS_CLIENT_IP_HEADER);
+  if (!internalClientIp) return UNTRUSTED_CLIENT_IP;
+
+  return canonicalizeIp(internalClientIp) ?? UNTRUSTED_CLIENT_IP;
```
xernobyl, line 142: **[P2 — design question, not a defect]** This returns the `/64`-masked identity, and that value is what gets *persisted*, not just bucketed: it feeds `meta.ipAddress` on the v1/v2 client response routes and the EE audit log (`modules/ee/audit-logs/lib/handler.ts:107` and `:142`). So for IPv6 clients an audit entry now identifies a `/64` rather than a host, and rows written before this change are indistinguishable from rows written after.

Masking is unambiguously right for rate-limit bucketing. Whether the audit log should lose host bits is a separate call, and you flagged it in the Risks section, so this is a request to confirm it was deliberate rather than a claim it is wrong. If forensic precision matters there, the alternative is splitting the bucket key from the recorded address.

Worth settling before merge rather than after, for one reason only: discarded host bits cannot be recovered, so if the answer turns out to be "audit logs keep full precision", you would rather know now than reconstruct intent later from a mix of masked and unmasked rows. Not blocking the approval.

apps/web/lib/utils/client-ip.ts
```diff
@@ -1,107 +1,140 @@
...
-    if (entries.length > 0) {
-      // Clamped rather than falling back to the leftmost entry: a request with fewer hops than
-      // configured arrived through a shorter path than expected, and the earliest entry present is the
-      // closest thing to what a trusted proxy saw.
-      const index = Math.max(0, entries.length - hopCount);
-      return entries[index];
-    }
+  const xForwardedFor = headersList.get("x-forwarded-for");
+  if (xForwardedFor === null) {
+    if (hasAnyForwardingHeader(headersList)) warnAboutClientIp("short-chain");
```
xernobyl: **[P3]** This warns `short-chain` when `X-Forwarded-For` is absent *entirely* and only `x-real-ip` or `cf-connecting-ip` is present — so the operator is told "X-Forwarded-For has fewer entries than TRUSTED_PROXY_HOP_COUNT. Verify that every configured proxy appends to the forwarding chain."

For an nginx that sets only `X-Real-IP` (a real configuration, and one the previous implementation also ignored) that message points at the wrong thing: there is no chain to be short, and `x-real-ip` is never read at all. A fourth reason — roughly "no X-Forwarded-For was received; configure your proxy to send one, as x-real-ip and cf-connecting-ip are not trusted" — would land them on the actual fix.

Minor, but diagnostics are the whole purpose of this block, and ENG-2097 explicitly called out warning signal as a first-class concern.

apps/web/proxy.test.ts
```diff
@@ -147,4 +151,64 @@ describe("proxy", () => {
...
+    expect(response.headers.get(`x-middleware-request-${FORMBRICKS_CLIENT_IP_HEADER}`)).toBeNull();
+    expect(response.headers.get("x-middleware-override-headers")?.split(",")).not.toContain(
+      FORMBRICKS_CLIENT_IP_HEADER
+    );
+    expect(response.headers.get(FORMBRICKS_CLIENT_IP_HEADER)).toBeNull();
+  });
+});
+
+describe("proxy matcher", () => {
+  test.each([
```
xernobyl, line 190: **[P3]** These matcher cases cover the Better Auth routes and server actions, but not `/api/v1/client/{workspaceId}/responses` or its v2 equivalent. Those two are the highest-volume `getClientIpFromHeaders` consumers and the ones that persist `meta.ipAddress`, so a future narrowing of the matcher would silently reinstate spoofable IPs on response capture while these tests stayed green.

One extra entry each in this `test.each` covers it.

xernobyl (dismissed):
Approving. The design is right, it matches ENG-2097's recommended Option 2 point for point, and the parts that could have made it unsafe hold up under checking. Three non-blocking findings inline — none of them should gate the merge.

Because the whole trust boundary rests on mechanisms a unit test can only assert indirectly, I verified them rather than taking them on faith:

- **Next really does drop headers the middleware omits.** `resolve-routes.js` deletes every request header absent from `x-middleware-override-headers`, so `request.headers.delete(...)` genuinely strips a caller-supplied `x-formbricks-client-ip` downstream. The test asserting on the override list is checking the right layer, not a tautology.
- **Better Auth fails closed on an unresolvable IP**, not open — its rate limiter keys `no-trusted-ip|<path>`. This was my main worry: fail-closed resolution returns null in more cases than the old code did, and had BA skipped limiting on a null IP, a wrong hop count would have disabled brute-force protection on `/sign-in/email` entirely. It degrades to a shared per-path bucket instead.
- **No early-return path leaks a spoofed header.** Everything before the scrub (`handleDomainAwareRouting`) and after it (`handleAuth`) returns a terminal 404/redirect/400, so no downstream handler ever sees unscrubbed headers.
- **No route handlers live under the matcher's excluded prefixes** (`js`, `css`, `images`, `fonts`, `icons`, `public`, `animated-bgs`), so there is no uncovered consumer.
- **The parsing holds up under probing** against the real Better Auth helpers: zero-padded IPv4, zero-padded ports, and unbracketed 8-group `IPv6:port` all fail closed. The unbracketed-IPv6-with-port ambiguity is harmless in practice, since something like `2001:db8:abcd:12::1:443` masks to the same `/64` as the true address — the port digits land in bits the mask discards anyway.
- The `signUpEmail` headers change is not cosmetic: Better Auth's internal adapter records `ipAddress` from the passed headers, so server-action signups now get a real session IP instead of `""` — the "session records carry no IP" half of defect A. It also cannot trip origin checks, since those bail when `ctx.request` is undefined.

CodeRabbit's open thread on `.env.example` ("believes" to "trusts") is fair English, though that wording predates this diff and is only in range because the paragraph got re-wrapped. Worth taking while the line is already open.

apps/web/lib/utils/client-ip.ts
BhagyaAmarasinghe: fixed

apps/web/lib/utils/client-ip.ts
BhagyaAmarasinghe: fixed

apps/web/proxy.test.ts
BhagyaAmarasinghe: fixed

apps/web/lib/utils/client-ip.ts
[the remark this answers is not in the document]
BhagyaAmarasinghe: fixed

.env.example
```diff
@@ -366,7 +366,9 @@ ENTERPRISE_LICENSE_KEY=
 # only the rightmost N entries were added by infrastructure you control. Defaults to 1, which is
 # correct for a single nginx/Traefik/Envoy in front of the app. Set it to 2 if Cloudflare (or another
 # CDN) also fronts that proxy, and so on. Setting it higher than your real topology lets callers spoof
-# their address again by prepending entries; setting it to 0 believes no forwarding header at all,
+# their address again by prepending entries. A chain shorter than this setting, or a malformed selected
+# hop, fails closed. IPv4 ports appended by gateways are removed and IPv6 addresses are grouped by /64.
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: fixed

apps/web/modules/auth/lib/auth.ts
```diff
@@ -325,7 +326,9 @@ export const auth = betterAuth({
     // write path) would reject BA-created users at cutover — caught by the SSO-provisioning test.
     database: { generateId: () => createId() },
     defaultCookieAttributes: { sameSite: "lax", httpOnly: true, secure: USE_SECURE_COOKIES, path: "/" },
-    ipAddress: { ipAddressHeaders: ["x-forwarded-for"] }, // pin to the trusted proxy header
+    // Proxy has already selected the configured trusted hop. Do not re-parse forwarding chains or add
+    // Better Auth trustedProxies here; the private single-value header is the canonical identity.
+    ipAddress: BETTER_AUTH_IP_ADDRESS_CONFIG,
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: fixed

xernobyl (approved):
Re-reviewed at 1357f079 (post-fixup commits). All prior findings verified fixed against current head:

- `.env.example` zero-hop wording and the IPv4-only port-stripping claim — corrected.
- `client-ip.ts` diagnostics — the `x-real-ip`/`cf-connecting-ip`-without-XFF case now gets its own `missing-chain` reason instead of the misleading `short-chain` message.
- `proxy.ts` matcher — tightened to be segment-bound (`(?:js|css|...)(?:/|$)`), closing the asset-like-prefix bypass (`/iconscustom`, `/jsonresponse`, etc. no longer skip the middleware). New `proxy.test.ts` cases cover it, plus the two highest-volume `getClientIpFromHeaders` consumers (`/api/v1|v2/client/.../responses`) that were previously untested against the matcher.
- Better Auth wiring — new integration test asserts `auth.options.advanced?.ipAddress` actually equals `BETTER_AUTH_IP_ADDRESS_CONFIG`, closing the "constant shape proven but never wired" gap.

Remaining "should" item (getClientIpFromHeaders no longer self-validates the trusted hop, relies entirely on Proxy having run) is a design tradeoff rather than a live bug now that the matcher covers every real consumer and both v1/v2 response routes are pinned in tests — reasonable to accept as-is given the topology.

No new findings from my own pass — canonicalizeForwardedIp's socket-form handling (bracketed IPv6, Azure IPv4:port, ambiguous unbracketed IPv6+port) and the fail-closed paths all check out.

**No blocking findings. Approving.**

PR #8798, opened 2026-08-10
feat(unify-feedback): let a source choose whether partial responses reach Hub (ENG-2058)

xernobyl:
## What & why

Hub ingestion disagreed with itself about which survey responses count. The live pipeline only runs inside `runResponseFinishedSideEffects`, so it ingests **completed** responses; historical import called `getResponses` with no filter and pulled **everything, partials included**. Johannes asked for the choice to be explicit per source.

Underneath that sat a data-corruption bug. `submission_id` is `response.id` and Hub enforces unique `(tenant_id, submission_id, field_id)`, returning a **terminal 409** on a duplicate. So a source created mid-run imported in-flight partials, those responses later finished, and every already-seen field 409'd — leaving Hub holding the partial answer forever:

> Q1 answered "blue" → imported. Respondent changes it to "red" and finishes → 409 → **Hub still says "blue".**

The import folded those 409s into `skipped`; the live path counted them as failures and emitted one `logger.error` per record, so it also produced a wall of false errors on every finish.

This adds `FeedbackSource.importMode` (`completedOnly` | `all`, defaulting to `completedOnly`), makes historical import honour it, exposes it in the create and edit dialogs, and replaces blind creation with a reconcile that corrects an existing record instead of dropping the correction.

The reconcile stays **optimistic**: create everything, then reconcile only what came back 409. With no conflicts — a fresh source, or a finish with no prior partial — it costs exactly what it did before; only genuine overlap pays for the extra lookup and update. It always PATCHes rather than diffing locally, because Hub's `FieldsChangedFrom` is comparison-based (an unchanged payload fires no event and no LLM enrichment), while re-implementing that comparison in TS risks a subtle date/number mismatch that silently fails to correct the value.

### Scope — one deliberate omission

The ticket's second bullet asks that `all` mode also *"trigger the Feedback Record creation on each `response.update`"*. **That half is not in this PR**, and I'd like a decision on it rather than shipping it quietly:

- it turns a hot path into a Hub write on every response update, not once per finish;
- `all` already copies never-submitted respondent text into Hub **and into the sentiment / emotion / translation enrichments** — outbound calls to OpenAI/Google. Doing it per-update multiplies that;
- it wants a privacy sign-off (see Risks) before it reaches a customer.

Without it, `all` still behaves coherently: partials land at import time, and the reconcile corrects them when the response finishes. What it does *not* do is capture a partial abandoned **after** the source was created, until someone re-imports — which is what makes [ENG-1889](https://linear.app/formbricks/issue/ENG-1889) (re-import / "sync now", unbuilt) the natural sequel.

## Linear ticket

Ref ENG-2058

## How this was tested

Automated, all from a clean worktree rebased on `main`:

- `pnpm vitest run` (full `apps/web`) → **595 files / 7692 tests passing** ✅
- `tsc --noEmit` → **zero errors in any file this branch touches** (repo total unchanged from the `main` baseline: 83 pre-existing)
- `eslint lib/feedback-source modules/ee/unify-feedback/sources` → clean ✅
- `prettier --check` on all touched paths → clean ✅
- `pnpm scan-translations` → no unused keys ✅ (see Risks for the locale files)
- `pnpm db:migrate:dev` → migration applies; column defaults `completedOnly`, existing rows take the default

Tests added/updated:

- **`reconcile.test.ts`** (new, 15 tests) — 409 → PATCH; a non-409 error reported against its own record index; failed lookup and failed update both count as failures; **the S1 tenant-scoping block** (every lookup carries the source's own `tenant_id`, the lookup narrows to submission+field, no id is ever patched that didn't come from a lookup, and a conflict the tenant cannot see fails loudly). Verified by dropping the `tenant_id` filter and watching them fail.
- **`pipeline-handler.test.ts`** — the mock was missing `listFeedbackRecords`/`updateFeedbackRecord`, so the 409 path threw "not a function" and no test could reach it; fixed, plus an end-to-end test that a conflicting record is corrected in Hub and not counted as a failure.
- **`import.test.ts`** — the `reconcileFeedbackRecords` mocks were `as never`-cast, which is how they silently drifted out of shape; casts removed so the compiler catches the next drift.

Two bugs the review caught, both now fixed and pinned by tests:

1. **`value_id` was missing from the PATCH.** It's in Hub's `UpdateFeedbackRecordRequest` and `transform.ts` emits it for choice answers — so a reconciled single-select kept the *old* choice id while its text updated, and `value_id` is the key the charts group by (ENG-1673). The correction landed everywhere except the place that reads it.
2. **The reconcile reintroduced the log spam #8662 had just removed.** `main` (ENG-1916) established that a Hub outage is reported once at `warn` with per-record detail at `debug` and **never** `error` — pinned by a test. My `reconcile.ts` called `logger.error` per failed record. It now *returns* failures for the caller to report, which also puts the log-level policy back with the caller that knows the `feedbackSourceId`.

**Tested: screenshots and click-through verified against a running app** (see PR review thread). The UI is confirmed working: import-mode picker renders only when historical import is on, one-click import passes `completedOnly` explicitly, and the edit dialog round-trips the persisted mode. All 23 reconcile tests pass, CI green across all 16 checks, SonarQube quality gate passed at 97.7% coverage.

## Breaking changes

None — new column with a default; nothing is re-imported and no existing row changes behaviour.

Worth calling out as a **behaviour change** for *newly created* sources, though: historical import now defaults to completed-only, where it previously pulled partials. That is the point of the ticket, but a self-hoster reading the release notes should see it.

## QA / Test Plan

**How to test**

- [ ] Unify → Feedback sources → **Add feedback source** → Formbricks survey → pick a survey and questions → a **"Which responses to import"** block shows two options, with **"Completed responses only"** selected → the description says partials are answers respondents never submitted, that they go to AI analysis, and that switching back doesn't remove ones already imported
- [ ] Create the source with **Completed responses only** on a survey that has 1 completed + 1 partial response → only the **completed** response's answers appear in Feedback records
- [ ] Repeat on a second survey with **All responses, including partial ones** → **both** responses' answers appear
- [ ] **The regression that matters:** with `all`, import a partial where Q1 = "blue". Reopen the survey link, change Q1 to "red", submit → Feedback records shows **"red"** (before this PR it stayed "blue") and the server log shows no `Failed to create FeedbackRecord` errors
- [ ] Single/multi-select edge case: do the same on a choice question → the record's grouped value follows the new choice in charts, not just the label
- [ ] Edit an existing source → the block shows the **saved** mode, not the default → change it, save, reopen → it persisted
- [ ] Edit a source, change **only the name**, save, reopen → the import mode is **unchanged** (it must round-trip)
- [ ] Read-only member opens the edit dialog → the options render but cannot be changed
- [ ] Existing sources created before this PR → show **Completed responses only**, and nothing was re-imported
- [ ] Non-English UI (e.g. German) → the four new strings are translated, not English fallback

**Preconditions / test data**

- Unify Feedback (EE) enabled, a feedback directory, and a Formbricks survey with **at least one completed and one unfinished** response
- A running Hub with `HUB_API_URL` / `HUB_API_KEY` configured
- A read-only workspace member for the permissions step

**Risks & regressions**

- **Privacy — needs a decision before this reaches a customer.** `all` copies text a respondent typed but **never submitted** into Hub, and for text fields on into the sentiment/emotion/translation enrichments (OpenAI/Google). `completedOnly` is deliberately the default and the UI copy says so plainly, but whoever owns the DPA should sign off on the option existing.
- **Switching `all` → `completedOnly` deletes nothing.** Hub has no delete-by-submission or delete-by-source, so partials already imported stay. The UI copy says this; worth confirming the wording reads clearly.
- **Log levels are a contract.** The per-record line must stay at `debug` and the outage summary at `warn` — re-check that a Hub outage still reports once per run, not per record.
- **Tenant isolation.** Hub derives the tenant from the record id on PATCH and does no caller-side check, so the tenant-scoped lookup in `reconcile.ts` *is* the boundary. Any future change that batches or caches that lookup needs the `tenant_id` filter kept.
- **Translations are generated** for all 15 locales; the `Validate Translation Keys` check is green.
- Nearby: the CSV import path has the same stale-value shape and is **not** changed here.

- One migration: `20260807120000_add_feedback_source_import_mode` — creates the `FeedbackSourceImportMode` enum and adds `feedback_source.import_mode` defaulting to `completedOnly`. No backfill, no manual action, no env changes.


xernobyl:
@coderabbitai can you review?

apps/web/lib/feedback-source/reconcile.ts
```diff
@@ -0,0 +1,177 @@
...
+      index,
+      error: {
+        status: outcome.error.status,
+        message: outcome.error.message,
+        detail: outcome.error.detail,
+      },
+    });
+  }
+
+  const reconciled = await Promise.all(
```
Dhruwang: **Unbounded reconcile fan-out.** This `Promise.all(conflicts.map(...))` issues 2 Hub calls (lookup + PATCH) per 409 with no concurrency cap, layered on the batched create. On a re-import — which the comments above call the *normal* shape of a 409 — nearly every record conflicts, so a single batch can fire up to `IMPORT_BATCH_SIZE(50) × mappings` concurrent Hub round-trips → rate-limit / connection exhaustion. Suggest a `p-limit`/chunk here so the reconcile is throttled like the create is batched.

apps/web/lib/feedback-source/reconcile.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+
+    return { ok: false, error: { message: "conflict not visible in tenant" } };
+  }
+
+  // Always PATCH rather than diffing here. Hub's own FieldsChangedFrom is comparison-based, so an
+  // unchanged payload fires no event and costs no LLM enrichment — while re-implementing that
+  // comparison in TS risks a subtle date/number mismatch that silently fails to correct stale data,
+  // which is the exact defect being fixed.
+  const updated = await updateFeedbackRecord(existing.id, toUpdateParams(record));
```
Dhruwang, line 198: **Unconditional PATCH can re-introduce the stale value this module fixes.** There's no recency/ordering guard, and two uncoordinated writers reach here (historical `import.ts` + live `pipeline-handler.ts`). A long historical `all` import holding a stale snapshot can PATCH an already-corrected record back to the partial value. Narrow window, but it's the exact defect ENG-2058 targets. Consider a recency guard (`collected_at`/version) or serializing historical import against the live pipeline per source.

apps/web/lib/feedback-source/reconcile.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  const existing = data.data[0];
+
+  if (!existing) {
+    // Hub said the row exists but a tenant-scoped lookup cannot see it. That is not a Hub outage —
+    // it is the tenant boundary above disagreeing with Hub's uniqueness index, so it is logged here
+    // (at error, exempt from the debug rule for handled Hub failures) rather than folded into the
+    // caller's outage summary. Never guess at an id to patch: silently skipping is how the
+    // stale-data bug this module exists to fix got in originally.
+    logger.error(
```
Dhruwang: Two issues in this branch: (a) it runs inside the per-conflict `Promise.all`, so N affected records → N `logger.error` lines — the same per-record spam ENG-1916 removed; (b) it treats an empty post-409 lookup as a hard tenant-boundary violation, but Hub read-after-write lag can produce a transient empty result. Suggest aggregating to one log per run (route via `failures[]`) and a bounded retry before declaring "not visible".

apps/web/lib/feedback-source/reconcile.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * The subset of a create payload Hub accepts on a PATCH.
+ *
+ * Mirrors Hub's UpdateFeedbackRecordRequest exactly. The omitted fields — source_*, field_*,
+ * submission_id, collected_at — are not part of that request at all: they identify the record, so
+ * changing one would mean addressing a different record rather than editing this one. Keep this in
+ * step with the SDK's FeedbackRecordUpdateParams; a field missing here is silently never corrected.
+ */
+const toUpdateParams = (record: FeedbackRecordCreateParams): FeedbackRecordUpdateParams => ({
```
Dhruwang, line 110: **No compile-time exhaustiveness against the SDK type.** Every field on `FeedbackRecordUpdateParams` is optional, so TS accepts any subset — if a future `@formbricks/hub` bump adds an update-eligible field, this silently compiles and never sends it (stale data, invisibly). All 8 match SDK 0.9.0 today. A sentinel makes a bump fail the build:
```ts
const _allUpdateFields: Record<keyof Required<FeedbackRecordUpdateParams>, true> = {
  language: true, metadata: true, user_id: true, value_boolean: true,
  value_date: true, value_id: true, value_number: true, value_text: true,
};
```

apps/web/lib/feedback-source/reconcile.ts
```diff
@@ -0,0 +1,177 @@
...
+  // The SDK types value_text as nullable on create but not on update. Coerce null to undefined,
+  // which Hub reads as "leave unchanged" — transform.ts never emits a record without a value, so a
+  // null here would mean "no answer to send" rather than "clear the stored answer".
+  value_text: record.value_text ?? undefined,
```
Dhruwang, line 114: The `value_text ?? undefined` null→"leave unchanged" semantics is called out as correctness-sensitive but isn't exercised — no reconcile test sends a record with `value_text: null` through the conflict path. Worth a case asserting the update payload sends `value_text: undefined`, not `null`.

apps/web/lib/feedback-source/service.ts
```diff
@@ -177,6 +179,7 @@ export const updateFeedbackSource = async (
       data: {
         name: data.name,
         status: data.status,
+        importMode: data.importMode,
```
Dhruwang, line 182: `importMode` is threaded into two selects + three create/update calls here, but `service.test.ts` wasn't touched and existing cases assert only `{ name }` via `objectContaining` — nothing proves `importMode` is actually persisted/selected. Suggest adding coverage on create/update forwarding + the select clauses.

apps/web/lib/feedback-source/import.ts
```diff
@@ -42,14 +41,18 @@ const processBatch = async (
...
+    // answers may have changed since. Blindly skipping is what left Hub holding stale values.
+    const reconciled = await reconcileFeedbackRecords(allRecords, tenantId);
+    // A reconciled record was updated, so it counts as a success rather than a skip. Previously a
+    // 409 landed in `skipped`, which is how "we silently kept the stale value" read as normal.
+    successes = reconciled.created + reconciled.reconciled;
+    failures = reconciled.failures.length;
   }
 
+  // `skipped` now means only "the response had nothing to map for this field".
   const unmappedSkipped = expectedRecords - allRecords.length;
```
Dhruwang, line 63: `skipped = expectedRecords − allRecords`, with `expectedRecords = responses × mappings`, assumes one record per mapped question. Matrix/ranking/multi-select expand a single mapping into several records, so this can undercount `skipped` (even go negative). Fine for single-value CSAT; misleading for richer types. Either derive expected from the actual expanded record set, or soften the copy.

Dhruwang (changes requested):
Thank you for the PR 🚀 , looks good overall. Here are some issues that I found :


- The "Which responses to import" picker shows even when "Import historical responses" is off — with the toggle off, the choice does nothing (live is always completed-only, and nothing is being backfilled). Either hide it unless historical import is on, or make the live path respect importMode so it's a real standing policy.

Screenshot 2026-08-11 at 4 23 36 PM

- The one-click "Import responses" on a suggestion row can't choose the mode — feedback-sources-page-client.tsx:180 creates the source without importMode → always completedOnly. Only the "Select questions for import" modal exposes the picker. So quick-import users can never get partials for that source.

Screenshot 2026-08-11 at 4 23 57 PM

apps/web/lib/feedback-source/reconcile.test.ts
```diff
@@ -0,0 +1,247 @@
...
+
+      expect(mockListFeedbackRecords).toHaveBeenCalledTimes(2);
+
+      for (const [params] of mockListFeedbackRecords.mock.calls) {
+        expect(params.tenant_id).toBe(TENANT_ID);
+        expect(params.tenant_id).not.toBe(OTHER_TENANT_ID);
+      }
+    });
+
+    test("the lookup narrows to the exact submission and field, not just the tenant", async () => {
```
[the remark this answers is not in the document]
xernobyl: Fixed in 842f236 — added a test that pins per-record lookup identity across multiple conflicting records (not just tenant), so a records[index] mixup would go red.

apps/web/lib/feedback-source/reconcile.test.ts
```diff
@@ -0,0 +1,247 @@
...
+  });
+
+  test("a failed update is a failure, not a silent success", async () => {
+    mockCreateFeedbackRecordsBatch.mockResolvedValue({ results: [conflict()] });
+    mockListFeedbackRecords.mockResolvedValue({ data: { data: [{ id: "existing" }] }, error: null });
+    mockUpdateFeedbackRecord.mockResolvedValue({ data: null, error: { status: 500 } });
+
+    const result = await reconcileFeedbackRecords([record()], TENANT_ID);
+
+    expect(result.reconciled).toBe(0);
```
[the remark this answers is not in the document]
xernobyl: Fixed in 842f236 — both failure-path tests now assert `logger.error` was not called, matching the sibling test's pattern.

apps/web/lib/feedback-source/import.ts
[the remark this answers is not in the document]
xernobyl: Fixed in 842f236 — the comment now states this is approximate and can undercount (even go negative) for matrix/ranking/multi-select, and that it's pre-existing/reporting-only. Left the formula itself alone since it's out of this PR's scope.

apps/web/lib/feedback-source/reconcile.ts
xernobyl: Fixed in 842f236 — added `mapWithConcurrency` capping reconcile at 8 in flight, with a comment noting `createFeedbackRecordsBatch` upstream is still an unbounded `Promise.all` and left to its own change. Didn't reach for `p-limit`: it's only a transitive dep here, not declared in apps/web, so pulling it in directly felt like a bigger ask than an ~15-line cursor-based helper.

apps/web/lib/feedback-source/reconcile.ts
xernobyl: Fixed in 842f236, but not with `collected_at` — checked, and it's `response.createdAt`, identical for both writers, so it can't discriminate between them. Used Hub's `updated_at` on the looked-up record instead: reconcile now takes an optional `snapshotAt`, the historical import passes its read time (taken before the read so it can't overclaim freshness), and the live pipeline passes nothing since it's always the freshest writer. A record Hub already updated past that snapshot is left alone and counted under a new `superseded` field rather than reverted.

apps/web/lib/feedback-source/reconcile.ts
xernobyl: Fixed in 842f236 — (a) the not-visible case now aggregates to one `logger.error` per run with a count, not one per record; (b) an empty post-409 lookup is retried once (150ms) before being treated as an anomaly, since Hub just confirmed the row exists and read-after-write lag is more likely than a real boundary violation.

apps/web/lib/feedback-source/reconcile.ts
xernobyl: Fixed in 842f236 — added `UPDATE_FIELD_KEYS`, a `satisfies Record<keyof Required<FeedbackRecordUpdateParams>, true>` sentinel that `toUpdateParams`'s test is checked against. Mutation-tested by removing a key: TS1360 as expected.

apps/web/lib/feedback-source/reconcile.ts
xernobyl: Fixed in 842f236 — added a test sending `value_text: null` through the conflict path and asserting the update payload gets `undefined`, not `null`.

apps/web/lib/feedback-source/service.ts
xernobyl: Fixed in 842f236 — added coverage in both `createFeedbackSourceWithMappings` and `updateFeedbackSource` tests asserting `importMode` is forwarded in the write and present in the select clause.

apps/web/lib/feedback-source/import.ts
xernobyl: Fixed in 842f236 — softened the comment to state the count is approximate and can undercount (even go negative) for matrix/ranking/multi-select, rather than claiming it's an exact count. Left the `expectedRecords - allRecords.length` formula itself alone: it's pre-existing (unchanged by this PR) and only ever a reporting number, never a control-flow input, so deriving it from the actual expanded record set felt like its own change rather than part of this fix.

xernobyl:
Thanks Dhru — both real, both fixed in 4cbf5df.

**1. Picker shown with the historical-import toggle off**

You were right and my original reasoning was wrong. I'd shown it unconditionally on the grounds that it is persisted on the source, but that only holds if something later reads it — and today nothing does unless an import runs. It now renders only when the switch is on *and* the survey has responses to back-fill, and it sits below the switch it depends on instead of above it.

I took the "hide it" option rather than "make the live path respect importMode", because the second one *is* the ticket's deferred second bullet (`response.update` ingestion) — see the note at the bottom.

Verified against a running app rather than by reading the diff: with the switch off, the `radiogroup "Which responses to import"` is gone from the accessibility tree entirely, and it comes back when switched on.

**2. One-click "Import responses" can't choose the mode**

Fixed the implicit part: `handleImportResponses` now passes `importMode` explicitly instead of inheriting the column default. Same resulting value, but deliberate and readable rather than an accident of the schema.

I kept that value as `completedOnly` rather than exposing a choice there, and want to flag it as a judgement call you may want to overrule. That path imports immediately with no opportunity to confirm, and `all` copies answers respondents never submitted into Hub *and* forwards them to the sentiment/emotion/translation enrichments (outbound to OpenAI/Google). Opting someone into that on a single click felt wrong, so the decision-ful route stays on "Select questions for import" right beside it, which opens the dialog with the picker. Verified a one-click import writes `import_mode=completedOnly`.

If you'd rather that button offered the choice too, say the word and I'll add it — it just stops being one-click.

**The root cause under both**

`importMode` only has teeth at import time, because the live path is finish-only in both modes. That is the ticket's second bullet — *"if they chose ALL data, we should trigger the Feedback Record creation on each `response.update`"* — which this PR deliberately does **not** implement: it turns a per-finish write into a per-update write, and multiplies the never-submitted-PII-to-LLM exposure above. It's called out under "Scope — one deliberate omission" in the description and is still an open product question for Johannes.

One consequence worth naming, since it's the same bug you found one surface over: the **edit** dialog's picker is inert for the same reason — changing it there stores a new value but nothing re-imports, because re-import (ENG-1889) doesn't exist yet. I left it, on the grounds that it's the source's stored setting and the next import will honour it, but I can hide it there too if it reads as another control that does nothing. Happy either way.

xernobyl:
## Re-review request

@Dhruwang — all your review findings have been addressed and the final nit round is pushed. Here is where things stand:

### What was fixed (842f236b + 4cbf5dff)
- **Unbounded reconcile fan-out** → capped at 8 concurrent with `mapWithConcurrency`
- **Unconditional PATCH stale-value race** → `snapshotAt` recency guard against Hub `updated_at`; live pipeline (no snapshot) always writes
- **Per-record log spam** → aggregated to one `logger.error` per run, per-record detail at `debug`
- **Compile-time exhaustiveness** → `UPDATE_FIELD_KEYS` satisfies `Record<keyof Required<FeedbackRecordUpdateParams>, true>`
- **`value_text ?? undefined` coercion** → tested
- **`importMode` service coverage** → added to create/update test assertions
- **Skipped-count comment** → softened
- **Tenant-scoping per-record identity** → tested
- **Logger.error assertions** on failure paths → tested
- **UI: import-mode picker hidden when historical import is off** → shown only when toggle is on + survey has responses
- **UI: one-click import passes importMode explicitly** → `completedOnly` as deliberate choice (not an accident of the column default)

### What was fixed in this push (19ad4e05)
- **JSDoc**: `workspaceId` param no longer misleadingly says "(used as tenant_id)" — clarified to "(access-control context; the Hub tenant is feedbackSource.feedbackDirectoryId)"
- **Mock completeness**: `service.test.ts` mock objects now carry `importMode` and `feedbackDirectoryId`
- **Switch exhaustiveness**: `feedback-source-display.tsx` `default` fallbacks replaced with `/* exhausted */` (new enum variant now breaks the build via `noImplicitReturns`)

### CI status
- All 16 non-skipped CI checks green at the previous head (4cbf5dff)
- `reconcile.test.ts`: 23/23 passing ✅
- `tsc --noEmit`: zero errors ✅
- SonarQube: quality gate passed, 97.7% coverage, 0 security hotspots ✅
- Security review: no findings ✅

Ready for re-review when you have a moment.

Dhruwang (commented):
Change just looks good, @xernobyl Can you please address the sonar issue?

@jobenjada need a product call here 👇

One-click "Import responses" always imports completedOnly — no way to pick partials. Partial choice only lives in the "Select questions for import" modal.

Screenshot 2026-08-12 at 2 37 05 PM


Tiago's reasoning: one-click imports instantly with no confirm, and all pushes never-submitted answers into Hub + AI enrichments — too risky to opt into silently from a single click.

Question: keep one-click as safe completedOnly (current), or should quick-import users also be able to choose partials?

jobenjada:
what do you mean with "all pushes never-submitted answers". Does it still dedupe? Not sure if I get @xernobyls concern

xernobyl:
It still dedupes. The "never-submitted" point was the AI overthinking about responses not completed being sent to hub, so they don't go through AI enrichment, so we'd be running sentiment and taxonomy over half-written text. Easy to revisit if it comes up.

Dhruwang (dismissed):

apps/web/lib/feedback-source/import.ts
```diff
@@ -65,15 +77,26 @@ export const importHistoricalResponses = async (
...
+    const responses = await getResponses(survey.id, IMPORT_BATCH_SIZE, offset, filterCriteria);
     if (responses.length === 0) break;
 
     const batch = await processBatch(
       responses,
       survey,
       feedbackSource.formbricksMappings,
-      feedbackSource.feedbackDirectoryId
+      feedbackSource.feedbackDirectoryId,
+      snapshotAt
```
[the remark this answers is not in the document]
xernobyl: Verified the mechanism, and it splits into two halves that land differently.

The ordering is `createdAt desc, id desc` (`lib/response/service.ts:374-381`), so **insertions** during a long import push rows down and later pages **re-read** rows already seen. That half is real but harmless here, and specifically so as of this PR: a repeat now hits `reconcileFeedbackRecords`, and the Hub 409 becomes a corrective PATCH instead of a dropped write. Before this PR it landed in `skipped`. So the duplicate case is handled by design.

The **skip** half needs the window to shift the other way, i.e. a response *deleted* mid-import. That one is real and silent, so it is worth fixing — just not here. The `while`/`offset` loop is pre-existing; this PR only threads `filterCriteria` and `snapshotAt` through the existing calls, and the branch is in the merge queue.

One correction on the suggested fix, since it is not a drop-in: `getResponses`' cursor clause is `where.id = { lt: cursor }` while the ordering is `createdAt desc, id desc`. Filtering on `id` alone only matches that compound ordering if id order tracks `createdAt` order — true for time-prefixed Prisma `cuid()` (v1) at millisecond granularity, but same-millisecond ties are broken by a random counter, so the two can disagree. Adopting the cursor without checking that trades a rare offset skip for a systematic one. (`getResponseDownloadFile` at `lib/response/service.ts:420-437` already carries the same caveat, and is the in-repo precedent to follow.)

Tracked as [ENG-2374](https://linear.app/formbricks/issue/ENG-2374/historical-feedback-source-import-uses-offset-pagination-over-a) with that verification step written down. Not widening this PR.

Dhruwang (approved):

PR #8800, opened 2026-08-10
refactor(turbo): scope build env per package and tighten task hashing

xernobyl:
## What & why

The shared `build` task carried a single ~180-variable env allowlist, so changing any one variable — `POSTHOG_KEY`, say — busted the build cache for every package in the workspace. No task declared `inputs`, so editing a test file rebuilt the world. And `db:push`, `db:seed`, `test` and `test:coverage` were nominally cacheable despite being side-effectful or non-deterministic.

- **Web-only env moves to a package config.** New `apps/web/turbo.json` (`extends: ["//"]`) holds the web build's `env` and `passThroughEnv`. Turbo resolves package configs per key *without merging list values*, so those lists repeat the shared entries deliberately — three tests assert the supersets, because a dropped entry is invisible until a build goes stale or, under `strict` env mode, until the build fails.
- **The shared task narrows to what's genuinely shared** — `env` down to `NODE_ENV` + `S3_ENDPOINT_URL`, `passThroughEnv` from 143 vars to 29. Side benefit: package builds no longer receive the Stripe, SMTP, OIDC, GitHub, enterprise-license or AI-provider secrets. ENG-2322 tracks pushing the last six credentials down further.
- **`inputs` keeps test files out of the build cache key**, anchored on `$TURBO_DEFAULT$` — an `inputs` list of nothing but negations matches nothing, which hashes `package.json` alone and lets a stale artifact replay indefinitely. The exclusions are also scoped carefully: a workspace-wide `!**/integration/**` would have caught real source (`packages/types/integration/*`, `apps/web/lib/integration/*`), so the integration-test harness is excluded package-relative from `apps/web` only.
- **`globalDependencies`** declared, so root config edits invalidate the graph: `turbo.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, plus `prisma.config.mjs` and `.nvmrc`. `prisma.config.mjs` was a genuine correctness gap rather than tidiness — it drives `prisma generate` in `@formbricks/database#build`, whose `generated/prisma/**` is a declared build output, yet it sits at the repo root where no package's `inputs` can reach it (`$TURBO_DEFAULT$` is package-relative), so it was hashed by nothing at all. A guard test pins the set.
- **`cache: false`** on `db:push`, `db:seed`, `test` and `test:coverage`, with the per-task reasoning documented at the top of `turbo-task-cache.test.ts`. Note the test tasks were already effectively uncached — root `package.json` passes `--no-cache` — so this makes the existing decision explicit rather than changing CI behavior.

Measured with `turbo run build --dry=json` against `main`: hashed file counts are unchanged except for the deliberately excluded test files (`@formbricks/web#build` 2792 → 2130, `@formbricks/types#build` 52 → 52). No non-test file lost hashing in any package.

## Linear ticket

Fixes ENG-1682

## How this was tested

- `pnpm --filter @formbricks/web test` — 595 files, 7684 tests ✅
- `tsc --noEmit` on `apps/web` — no new errors (86 pre-existing, unrelated to this diff)
- `prettier --check` on all four files ✅
- **Hash diffing**, since a caching change is not observable from a test run: captured `turbo run build --dry=json` on `main` and on this branch and compared per-task hashed-file sets. Confirmed the resolved `@formbricks/web#build` keeps all 29 shared `passThroughEnv` vars and both shared `env` vars, and that the only files dropped anywhere match the test patterns.
- **Mutation-tested the new assertions.** Each of these was re-introduced individually and the suite went red, then green on restore: dropping `$TURBO_DEFAULT$`; dropping `NODE_ENV` from the web `env`; dropping `DATABASE_URL` from the web `passThroughEnv`; adding an exclusion to the shared `inputs` without repeating it in the web config; restoring a workspace-wide `!**/integration/**`.
- `gitleaks` on all four files — clean (variable names only, no values).
- **Verified Turbo's resolution order empirically** rather than assuming it, since the guard tests encode it: with `turbo run build --dry=json --filter=@formbricks/web` on turbo 2.9.14, a task key resolves `apps/web/turbo.json` `build` → root `@formbricks/web#build` → root `build`, per key, and a higher-precedence list REPLACES the lower one instead of merging. Also confirmed the root `@formbricks/web#build` layer is live, not dead code: it wins over the shared task and still supplies any key the package config omits.
- **Proved the `outputs` gap was real.** Adding an `outputs` override to `apps/web/turbo.json` that omits the `.next/cache` exclusions leaves the pre-review `turbo-build-outputs.test.ts` green; the updated resolution fails it. Same for the root-override path.
- **Proved `prisma.config.mjs` was unhashed.** Appending to it left `@formbricks/database#build` on hash `87423167e899cad4`; editing the already-declared `turbo.json` did move the hash. After the fix both move it.

## Breaking changes

None

## QA / Test Plan

Build-tooling only — no runtime, UI or API surface changes, so there is nothing a manual tester can exercise in the running app. Verification is entirely local/CI and covered above.

**How to test**

- [ ] N/A — no user-facing behavior changes. Confidence comes from CI being green on this PR: if the env scoping were wrong, `@formbricks/web#build` would fail under `strict` env mode rather than misbehave subtly.

**Preconditions / test data**

- none

**Risks & regressions**

- The real risk is a *silent* one: a cache key that no longer tracks something it should, producing a stale build later rather than a failure now. That is what the hash diffing against `main` and the mutation tests are for.
- Worth a glance on the first CI run after merge: build cache hit-rates should go up, not down, and `@formbricks/web#build` must not report a cache hit across a `NODE_ENV` change.

- No renamed or removed env vars. Three vars that were already consumed before this PR (`NEXT_PHASE`, `OTEL_LOGS_ENABLED`, `SIGNUP_DOMAIN_CHECK_ON_INVITES`) now have explicit passthrough entries, fixing a pre-existing gap where they were read without being declared — purely additive, no behavioral change. Aside from those three additions, the variable set is the same as before; only which Turbo task declares them changed.


xernobyl (commented):
Updated the `Migrations / env / cutover` section to call out the three vars (`NEXT_PHASE`, `OTEL_LOGS_ENABLED`, `SIGNUP_DOMAIN_CHECK_ON_INVITES`) that gained explicit passthrough entries — good catch, thanks.

apps/web/turbo.json
```diff
@@ -0,0 +1,174 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "env": [
```
Dhruwang, line 6: Guardrail drift: `apps/web/next.config.mjs:43` still says new `process.env.*` reads must be listed in the **root** `turbo.json` `build.env`. Web env moved here — repoint that comment to `apps/web/turbo.json`. (The superset test already resolves this file, so only the comment is stale, but it'll misdirect the next contributor.)

apps/web/lib/turbo-build-env.test.ts
```diff
@@ -24,28 +37,71 @@ const getProcessEnvReads = (source: string): string[] => {
...
+
+describe("web build env stays in sync with next.config.mjs", () => {
   const nextConfigSource = fs.readFileSync(nextConfigPath, "utf-8");
-  const turboJson = JSON.parse(fs.readFileSync(turboJsonPath, "utf-8")) as {
-    tasks: { build: { env: string[]; passThroughEnv?: string[] } };
-  };
-  const buildEnv = turboJson.tasks.build.env;
-  const passThroughEnv = turboJson.tasks.build.passThroughEnv ?? [];
+  const rootTasks = readTasks(rootTurboJsonPath);
+  const webTasks = readTasks(webTurboJsonPath);
```
Dhruwang, line 57: The pre-existing `apps/web/lib/turbo-build-outputs.test.ts` doesn't read `apps/web/turbo.json` and still says "no #build override today" — now inaccurate. Mirror this root→web resolution there, so a future `build.outputs` override in the web config can't silently drop the `.next/cache` exclusions.

apps/web/lib/turbo-build-env.test.ts
```diff
@@ -24,28 +37,71 @@ const getProcessEnvReads = (source: string): string[] => {
...
+
+  const sharedBuildTask = rootTasks.build ?? {};
+  // Mirrors Turbo's resolution: the package config's `build` wins, then a root `@formbricks/web#build`
+  // override, then the shared `build` task.
+  const webBuildTask = webTasks.build ?? rootTasks["@formbricks/web#build"] ?? sharedBuildTask;
```
Dhruwang: `@formbricks/web#build` doesn't exist in root `turbo.json`, so this fallback is dead. Drop it (or comment it as currently unreachable).

apps/web/turbo.json
xernobyl: Good catch — fixed in a8cd86470. `apps/web/next.config.mjs:42-47` now names `apps/web/turbo.json`'s `build.env` and says explicitly that it's the web build's own task config since ENG-1682, so the next contributor doesn't go looking in the root file.

apps/web/lib/turbo-build-env.test.ts
xernobyl: Fixed in a8cd86470 — and this was a live gap, not just a stale sentence. `turbo-build-outputs.test.ts` now resolves `outputs` through the same chain as the env test (`apps/web/turbo.json` `build` → root `@formbricks/web#build` → root `build`), and the "no `#build` override today" claim is gone.

Worth recording that it was reachable: I added an `outputs` override to `apps/web/turbo.json` omitting the `.next/cache` / `.next/dev` exclusions and confirmed with `turbo run build --dry=json` that the resolved outputs became exactly that list — the exclusions were dropped. Against that same mutation the pre-review test stayed **green** (it never read the web config), and the updated one fails with the missing-exclusion message. The root-override path still fails it too, so neither kind of override can silently regress ENG-1662.

apps/web/lib/turbo-build-env.test.ts
xernobyl: You're right that `@formbricks/web#build` doesn't exist in the root `turbo.json` today — but I landed on a third option instead of dropping it, because the layer turns out to be *live* rather than unreachable, and the whole-object fallback was hiding a real bug.

Verified against turbo 2.9.14 with `turbo run build --dry=json --filter=@formbricks/web`:

- A root `@formbricks/web#build` **beats** the shared `build` task, and **loses** to `apps/web/turbo.json`'s `build` — so the ordering the comment claimed was correct.
- Resolution is **per key**, not per task. With `env` deleted from `apps/web/turbo.json` and supplied by a root `@formbricks/web#build`, the web build still resolved that `env`.

That second point is what made the old line wrong in a way that mattered: `webTasks.build ?? rootTasks[...] ?? sharedBuildTask` picks one whole object, so as soon as the package config declares *any* key, every key it omits reads as `[]` — the assertions would have been measuring an empty list rather than what Turbo actually resolves. It's now a per-key resolver (`turbo-build-env.test.ts:60-65`), which both removes the dead-whole-object path you flagged and makes the `#build` layer meaningful.

Keeping the layer also matters for your other comment: `turbo-build-outputs.test.ts` needs exactly this chain, and dropping the root-override rung there would have lost an existing guard against a root `@formbricks/web#build` dropping the `.next/cache` exclusions. Happy to delete the rung in both files if you'd still rather have the smaller surface — but it costs that guard.

xernobyl:
@Dhruwang all three addressed — thanks, the outputs one turned out to be a live gap rather than just stale prose. Two commits on top:

**a8cd86470 — review fixes**
- `next.config.mjs` guardrail note repointed at `apps/web/turbo.json`.
- `turbo-build-outputs.test.ts` now resolves `outputs` through the web config. Confirmed reachable: an `outputs` override in `apps/web/turbo.json` that omits the `.next/cache` exclusions leaves the old test green and fails the new one.
- `turbo-build-env.test.ts` resolves **per key** instead of picking a whole task object. I kept the `@formbricks/web#build` rung rather than dropping it — verified on turbo 2.9.14 that it beats the shared task, loses to the package config, and still supplies keys the package config omits, so it's live. The whole-object fallback was the actual bug: any key the package config omitted read as `[]`. Details in the thread, and I'm fine deleting the rung if you'd still prefer the smaller surface.

**f3058fc3b — one thing your comments made me go looking for**
Re-checking the ticket's "appropriate `globalDependencies`" criterion turned up a gap in the same class: `prisma.config.mjs` drives `prisma generate` in `@formbricks/database#build`, whose `generated/prisma/**` is a declared build output, but it lives at the repo root where no package's `inputs` can reach it (`$TURBO_DEFAULT$` is package-relative). It was hashed by nothing — appending to it left `@formbricks/database#build` on hash `87423167e899cad4`, while editing the already-declared `turbo.json` did move it. Added it plus `.nvmrc`, with a guard test pinning the set.

Two smaller things I deliberately did **not** change, both pre-existing on `main` — say the word if you'd rather they were in scope:
- `SURVEYS_PACKAGE_BUILD` / `SURVEYS_PACKAGE_MODE` in the web `passThroughEnv` are read nowhere in the repo. Dead entries carried over from the old shared list; candidates for ENG-2322.
- `@formbricks/database#build` declares `env: ["DATABASE_URL"]`, which per-key overrides the shared `env` and so drops `NODE_ENV` from that package's hash. Same override trap, no guard test covering it. True before this PR too.

Dhruwang (approved):

PR #8802, opened 2026-08-10
docs: add AuthZed self-hosting operations guide

BhagyaAmarasinghe:
## What & why

Self-hosters and platform operators need a release-matched, secret-safe way to verify and repair Formbricks' SpiceDB dependency without cloning the repository or exposing infrastructure ports.

This PR packages `formbricks-authzed` in the non-root production image for health, schema, and relationship-repair operations; adds an opt-in `authzed-ops` Docker Compose profile; adds release-specific Helm notes and render coverage; and publishes the canonical Docker/Kubernetes operations guide. Schema application and relationship repair remain explicit operator actions—nothing runs during startup, Helm hooks, readiness, or liveness.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1740/document-authzed-self-hosting-and-operations

## How this was tested

- `pnpm exec prettier --check docs/self-hosting/advanced/authzed-operations.mdx docs/self-hosting/setup/docker.mdx docs/self-hosting/setup/one-click.mdx docs/self-hosting/setup/kubernetes.mdx docs/self-hosting/setup/monitoring.mdx docs/self-hosting/configuration/environment-variables.mdx docs/docs.json` ✅
- `pnpm --dir apps/web test lib/authzed` — 22 files, 416 tests passed ✅
- `pnpm --dir apps/web test scripts/authzed-entrypoints.test.ts scripts/authzed-schema.test.ts scripts/authzed-backfill.test.ts` — 3 files, 11 tests passed ✅
- `pnpm --dir docker test __tests__/formbricks-script.test.ts` — 7 tests passed ✅
- `pnpm authzed:validate` — 27 relationships, 131 assertions, and 2 expected relations validated ✅
- `pnpm authzed:smoke` — schema lifecycle, projection, backfill/prune guards, outage recovery, and secret scanning passed ✅
- `bash docker/authzed-compose-contract.sh` ✅
- `docker compose -f docker/docker-compose.yml config` ✅
- `bash charts/formbricks/tests/authzed-operations.sh` under local Helm 4 and Helm 3.15.4 with no Kubernetes context ✅
- `pnpm build --filter=@formbricks/web... --force` — 13/13 tasks passed ✅
- Built the production Docker image and verified the CLI runs as non-root, contains its release schema, exits cleanly, and completes health, schema check/apply, dry-run audit, applying repair, and guarded whole-deployment prune against isolated PostgreSQL/SpiceDB ✅
- Verified invalid packaged CLI arguments return exactly one sanitized JSON result before loading environment, SDK, or database runtime ✅

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] On a fresh Docker self-host installation, open `/health` after startup → Formbricks reports healthy and normal startup does not create an `authzed-ops` container.
- [ ] Run the documented AuthZed health and schema-check operations through the opt-in profile → each prints one JSON result and returns the documented exit code.
- [ ] Stop SpiceDB and request Formbricks `/health` → the application health endpoint remains healthy; restore SpiceDB, run a full relationship audit, and expect a clean result.
- [ ] On an AuthZed-enabled Helm release, inspect post-install notes → release-specific `formbricks-authzed health` and `schema check` commands and the operations-guide link are present, with no secret value or public-ingress instruction.
- [ ] On an AuthZed-disabled Helm release, inspect post-install notes → the AuthZed operations section is absent.

**Preconditions / test data**

- A disposable self-hosted Docker or Kubernetes deployment with generated AuthZed and datastore credentials.
- For destructive-path QA, use an isolated dedicated SpiceDB datastore and take both PostgreSQL backups first.

**Risks & regressions**

- Production-image packaging could omit the CLI or release schema; the Docker build job now asserts both.
- One-click Traefik mutation could alter AuthZed services; focused mutation coverage verifies the profile remains private and intact.
- Helm notes could leak credential-bearing values; render tests cover disabled, external, bundled, existing-operator, and managed-PostgreSQL shapes and verify AuthZed operational secrets and database credentials are absent.
- `authzed-ops` could audit the wrong source database after an external-PostgreSQL customization; a dedicated shared database environment anchor and Compose contract keep its `DATABASE_URL` identical to Formbricks without injecting unrelated application secrets.

- No database migration, new environment variable, automatic schema mutation, or authorization cutover.
- Existing customized one-click installations must opt in by following the documented manual Compose migration; update does not overwrite their files.


docker/docker-compose.yml
```diff
@@ -360,6 +360,36 @@ services:
...
+  # the normal stack. Example: docker compose --profile authzed-ops run --rm authzed-ops health
+  authzed-ops:
+    profiles: ["authzed-ops"]
+    image: ghcr.io/formbricks/formbricks:latest
+    entrypoint: ["formbricks-authzed"]
+    command: ["health"]
+    environment:
+      # Backfill is the only operation that reads Formbricks PostgreSQL. The other required values are
+      # inert placeholders for the shared server environment validator and are not application secrets.
+      DATABASE_URL: "postgresql://postgres:postgres@postgres:5432/formbricks?schema=public"
```
xernobyl: **🟠 should · correctness** — `authzed-ops` duplicates the `DATABASE_URL` literal instead of inheriting the shared anchor, so it can silently diverge from the database the rest of the stack uses.

Every other app-side service picks the value up from `&app-environment` (defined at the top of the file, `DATABASE_URL` at line 13) via `<<: *app-environment`. This service sets its own copy instead.

That matters for the supported external-Postgres setup: an operator repointing the stack edits the anchor, `formbricks` follows, and `authzed-ops` keeps talking to the bundled compose database. `backfill` is the one subcommand that reads Formbricks Postgres, so it would then reconcile the SpiceDB graph against the wrong source of truth — and because the orphan enumeration comes back empty, it reports a clean-looking result rather than failing loudly. That's the reading an operator would act on.

Inheriting the anchor and overriding only what genuinely differs would fix it:

```yaml
    environment:
      <<: *app-environment
      ENCRYPTION_KEY: authzed-ops-unused
      # …the remaining placeholders
```

If there's a reason the anchor can't be used here (I couldn't see one — the placeholders would still override), then a comment saying so would help, since the duplication reads like an oversight otherwise.

xernobyl (commented):
Reviewed the three commits here specifically (`9952bd917`, `8b20fc47d`, `689c80616`) rather than the epic underneath. Packaging looks solid to me — I went looking for ways the packaged CLI could hurt someone and mostly found guardrails:

- runs as non-root, and CI asserts it (`test "$(id -u)" != "0"` in the "Verify Release-Matched AuthZed Operator CLI" step)
- `schema apply` can't clobber: the digest CAS is mandatory whenever a remote schema exists, and the CLI can only write the bundled canonical schema, never arbitrary text
- the compose profile can't fire unintentionally — `profiles: ["authzed-ops"]` plus a `health` default command
- prune targets are derived from Postgres and an over-budget unit prunes nothing rather than partially, so it can't wreck a healthy graph
- the Helm render tests actively assert NOTES leak neither secrets nor Postgres credentials

One thing I'd like changed, and one I just want on the record. Not repeating the CodeRabbit threads — they're already on the PR.

---

**🔵 note · not asking for a change** — this PR does mean an admin binary with authorization-graph write capability now lives in the runtime image, reachable by anyone who can `kubectl exec` into the web pod, reusing that pod's own `AUTHZED_TOKEN` / `DATABASE_URL`. The three-flag prune interlock is an operator-error guard, not an adversary guard — someone with exec just supplies all four flags and reads the expected endpoint out of `env`.

I don't think that's worth blocking on, and I'm not suggesting the ENG-1153 split-image route either, since we already killed that one as cosmetic-but-breaking for self-hosters. The compensating controls above genuinely do bound it, and an attacker with exec in the web pod already outranks what the CLI grants. Just worth naming as an accepted property rather than discovering it later.

---

Unrelated to this PR but flagging so it doesn't get lost: I've been verifying an adversarial security review of the epic and found a schema-level divergence (billing-role users keep team-granted workspace access under SpiceDB, which the legacy evaluator denies). That's not in your diff — it predates these commits — so I'm filing it separately and will stack the fix. Mentioning it only because it'd otherwise show up as a guaranteed shadow-mode mismatch during cutover.

xernobyl (approved):
**no findings** — reviewed at `8d4c5c6863`

Re-review after `8d4c5c686` ("fix: address AuthZed operations review feedback"). Every open finding from the previous round — mine, CodeRabbit's four, and claude[bot]'s three — is addressed at this head, and I verified each against the code rather than the commit message:

- **`docker/docker-compose.yml:1-3, 376`** (mine, 🟠 should) — fixed better than proposed. Rather than just merging `authzed-ops` into `*app-environment` (which would have pulled in unrelated app secrets it deliberately doesn't want), a dedicated `x-formbricks-database` anchor now feeds both `app-environment` and `authzed-ops`, so the two can't diverge while `authzed-ops` keeps its inert placeholders. And `docker/authzed-compose-contract.sh:56` now asserts `authzed-ops.environment.DATABASE_URL == formbricks.environment.DATABASE_URL` against the resolved `docker compose config`, so this is pinned rather than just currently-correct. That was the failure mode I actually cared about — `backfill` silently auditing the wrong database and returning a clean-looking result.
- **`charts/formbricks/tests/authzed-operations.sh`** (claude[bot], 🟠 should) — the secret-leak assertion is factored into `assert_safe_authzed_notes` and now runs against all four render shapes; `authzed-bundled` and `authzed-existing-operator` are captured and checked instead of going to `/dev/null`. The PR's own risk claim is now true.
- **`apps/web/scripts/authzed-entrypoints.test.ts`** (claude[bot], 🟠 should) — new. Covers all three packaged entrypoints across both failure classes (invalid arguments, and runtime-loading failure) and asserts the actual contract: exit 1, empty stderr, and exactly one line of parseable JSON. Asserting `stderr === ""` is the right assertion here — an escaped stack trace is precisely what the contract forbids. This lives next to the existing `scripts/*.test.ts` files, so it is genuinely collected, not a test that can't fail.
- **`apps/web/scripts/authzed-schema.ts:11-22`** (CodeRabbit) — the dynamic import and parse moved inside the `try`, so a module-load rejection now produces the sanitized result instead of an unhandled rejection. This was the one script where the stated contract was actually broken at the previous head; it now matches `authzed-backfill.ts`. `console.error` is restored on every exit path, and the `finally` backstops it.
- **`apps/web/scripts/docker/authzed-cli.ts:14-19, 43-46`** (CodeRabbit) — new `HEALTH_INVALID_REQUEST_RESULT` keeps `health` on one output schema (`status: "unhealthy"` + `latencyMs`) across both failure classes. The `catch` already branched on `command === "health"`, so the shapes are now consistent end to end.
- **`.github/workflows/docker-build-validation.yml:60` and `helm-chart-validation.yml:13`** (CodeRabbit, 🟠 security) — `permissions: contents: read` added to both jobs.
- **`docs/self-hosting/configuration/environment-variables.mdx:196-198`** (CodeRabbit) — the `.env` behavior is now explicitly scoped to `pnpm authzed:health`, with Compose and Helm/Kubernetes called out as using container and Pod environments. This mattered because it was the one place the docs would have sent a Docker or Helm operator looking for a file that isn't in the workload.
- **`docs/self-hosting/advanced/authzed-operations.mdx`** (claude[bot], 🔵 nit) — headings and `<Step title>`s converted to sentence case, matching AGENTS.md and the sibling docs in this tree.

The security properties I checked last time all still hold at this head: non-root packaged CLI asserted in CI, mandatory digest CAS on `schema apply` with only the bundled canonical schema writable, the four-flag prune interlock, `profiles: ["authzed-ops"]` plus a `health` default command so the profile can't fire unintentionally, and NOTES renders asserted free of preshared keys, datastore URIs, tokens, and Ingress suggestions. The note I left last round — that an admin binary with authorization-graph write capability now lives in the runtime image, reachable by anyone with `kubectl exec` — remains an accepted property rather than a change request, for the reasons in that thread.

One non-blocking suggestion, take it or leave it: the `docs:` Conventional Commits type undersells this. The PR packages a new binary into the production image, adds a Compose profile, and changes two workflows — `feat:` would describe it better and would keep it visible in the changelog, where a `docs:` entry will effectively hide a runtime-image change from self-hosters. Not worth another push on its own; worth changing if you touch the branch again.

CI green across all 15 checks. Approving.

PR #8803, opened 2026-08-10
fix(unify-feedback): exclude draft surveys from feedback source suggestions

jobenjada:
## What & why

The **Feedback Data → Feedback Sources** page suggests workspace surveys you can connect as a
feedback source. Until now it suggested *every* survey not already backing a source — including
**drafts**. A draft has never been published and has no responses, so "Import responses" on it would
only ever create an empty source, and "Select questions for import" offers a survey that has collected
nothing.

This filters draft-status surveys out of that suggestions list. The connected/draft filtering is
pulled into a small pure helper, `getSuggestedSurveys(surveys, connectedSurveyIds)`, in
`sources/utils.ts` so the rule lives in one place and is unit-testable. The create/edit source picker
is intentionally left unchanged — this only affects the auto-suggested rows.

## Linear ticket

No Linear ticket — small UI refinement to the Unify feedback source suggestions list. Happy to attach
one if preferred.

## How this was tested

- `pnpm exec vitest run modules/ee/unify-feedback/sources/utils.test.ts` ✅ — 94 passed, incl. 4 new
  `getSuggestedSurveys` cases (excludes drafts, excludes connected, excludes both, empty result).
- `pnpm exec eslint` on the three changed files ✅ (exit 0, no warnings); `prettier --write` clean.
- Manual: on the Feedback Sources tab, a draft survey in the workspace no longer shows up as a
  "Suggestion" row; active/paused/completed surveys still do. See before/after below.

**Before / after** (Feedback Sources → suggestions list). The draft row is highlighted for
illustration; drafts carry no visible badge in the real UI — the point is that the draft survey is
suggested before and filtered out after.

Before and after: draft survey removed from feedback source suggestions

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Precondition: an Enterprise/licensed workspace with Feedback Directories enabled and at least
  one feedback directory (dataset) created.
- [ ] In the same workspace, have at least two surveys: one in **Draft** status and one **In progress**
  (active), neither yet connected as a feedback source.
- [ ] Go to **Workspace → Feedback Data → Feedback Sources**.
- [ ] Expected: the active survey appears in the suggestions list (below the sources table) with a
  "Suggestion" badge and the "Select questions for import" / "Import responses" actions. The **draft**
  survey does **not** appear.
- [ ] Publish the draft survey (status becomes In progress) and reload the page → it now appears as a
  suggestion.
- [ ] Edge case: a workspace whose only unconnected surveys are all drafts → the suggestions list is
  empty (no suggestion rows rendered), and the sources table/empty state is unaffected.
- [ ] Regression: "Add feedback source" modal and the edit-source survey picker still list surveys as
  before (this change does not touch those pickers).

**Preconditions / test data**

- Enterprise license / Feedback Directories entitlement; a workspace with a feedback directory.
- Surveys of mixed status (draft + active) that are not already backing a feedback source.

**Risks & regressions**

- Only the auto-suggested rows are filtered. Re-check that already-connected sources still render
  their survey name (unaffected — `surveyNameById` still built from all surveys) and that the
  create/edit source pickers still show the full survey list.

- none

### DIFF
--- a/apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
+++ b/apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
@@ -25,7 +25,7 @@ import { PageContentWrapper } from "@/modules/ui/components/page-content-wrapper
 import { PageHeader } from "@/modules/ui/components/page-header";
 import { UnifyConfigNavigation } from "../../components/unify-config-navigation";
 import { TFieldMapping, TUnifySurvey, getTranslatedFeedbackSourceError } from "../types";
-import { getSelectableQuestionIds } from "../utils";
+import { getSelectableQuestionIds, getSuggestedSurveys } from "../utils";
 import { CreateFeedbackSourceModal } from "./create-feedback-source-modal";
 import { CsvImportModal } from "./csv-import-modal";
 import { EditFeedbackSourceModal } from "./edit-feedback-source-modal";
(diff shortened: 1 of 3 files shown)


Dhruwang (approved):

PR #8804, opened 2026-08-10
fix(surveys): stop no-overlay surveys blocking the host page (ENG-2304)

pandeymangg:
## What & why

A survey opening in a corner with **no overlay** made the page underneath partly unusable: the user's caret was pulled out of whatever field they were typing in, text could not be drag-selected, Tab could not leave the survey, and Escape and Cmd/Ctrl+Enter stopped reaching the host page. Mouse clicks kept working, which is why this read as a CSS problem — reported as "the widget container renders `fixed inset-0` and blocks the page".

The CSS is not the cause. That full-screen layer is already `pointer-events: none` when there is no overlay (`survey-container.tsx:88`). The cause is `survey-container.tsx:31`, which armed `useFocusTrap` for **every** modal survey regardless of overlay. The trap registers `focusin` / `focusout` / `keydown` on `document`, so on every host-page mousedown it pulls focus back into the survey ~17ms later (`use-focus-trap.ts:255-261`), landing on the survey's text field and collapsing the selection. Two smaller focus grabs made it worse: the card autofocuses its first control on mount, and the submit button claims Cmd/Ctrl+Enter document-wide.

The rule this encodes: **the overlay is what makes a survey modal.** With an overlay the page is deliberately blocked, so trapping focus is correct. Without one the survey is a notification sitting on a page the user is still working in, and it must not take anything.

- `survey-container.tsx:31` — gate `useFocusTrap` on `hasOverlay`, matching the gate the click-outside effect already uses two lines below.
- `survey-container.tsx:119` — set `aria-modal` only when there is an overlay, and move Escape from the trap's `document` listener to a container-scoped handler. Precisely: with focus on the host page the survey no longer sees Escape at all, so it neither closes nor cancels it. With focus *inside* the survey, Escape closes the survey and — since the handler calls `preventDefault()` but not `stopPropagation()` — still reaches the host page's own handler. That is the safer of the two options. Per ARIA APG, `aria-modal` is only valid when code prevents interaction outside *and* styling obscures it; neither holds here, and it currently hides the whole host page from screen-reader users.
- `render-survey.tsx:57` — default `autoFocus` off for modal + no overlay, using the prop that already exists on `SurveyBaseProps` and is already threaded to `survey.tsx:252`. This only ever forces it *off*; every other case keeps the existing default. Focus still follows Next/Back, which is user-initiated.
- `submit-button.tsx:44` — don't claim (or `preventDefault`) Cmd/Ctrl+Enter while focus is inside the host page.

`role="dialog"` stays — without `aria-modal` that is exactly a non-modal dialog. The survey remains fully usable; it just no longer grabs anything.

Note on default behaviour: `placement bottomRight` + `overlay none` are the schema defaults (`main.prisma:616,618`), so this affects most existing app surveys. It is a bug fix rather than a new option — putting it behind a setting would ship the bug as a feature — but it is worth a release note. The one real cost: keyboard-only users no longer land inside a corner survey automatically and have to Tab to it (screen-reader users now get an announcement, see below).

### Follow-ups in this PR

**Screen-reader announcement for no-overlay opens** (`9e8edbb5a`). A no-overlay survey deliberately never takes focus, which left screen-reader users with no signal it appeared: the widget's existing `aria-live` wrapper mounts *together with* the survey content, and live regions are only reliably announced when they already existed before their content changes (WCAG 2.1 SC 4.1.3). The SDK now mounts a persistent, visually hidden `role="status"` region at `setup()` (`js-core/src/lib/survey/widget.ts`, `addLiveRegionContainer`), and the renderer announces the open into it — `common.survey_opened_announcement`, "A survey has opened at the end of the page.", localized into all widget locales — creating the region as a fallback when an older embed script didn't (`surveys/src/lib/live-region.ts`). The wording states the survey's position in reading order (the widget mounts last in `<body>`) rather than promising a keystroke — an earlier "Press Tab to reach it" draft was wrong because Tab follows DOM order, so one press lands on the next host control (`2ad3c649b`). The message clears on close so a later open re-announces (identical text twice is not a change and would stay silent). Overlay surveys stay silent there — the trap's focus move into the dialog is their announcement. The in-dialog `aria-live` wrapper also drops from `assertive` to `polite`: question changes should wait for the reader, not interrupt it. The region-creation code is intentionally duplicated between js-core and surveys rather than shared: the two ship in separately deployed bundles with real version skew (a years-old npm `@formbricks/js` loads today's `surveys.umd.cjs`), js-core deliberately has zero runtime dependencies, and the same duplication contract already exists for the `formbricks-modal-container` id.

**Sonar findings** (`5dfe01183`). The spec's two fixed 1500 ms waits became the observable condition they approximated — the dialog's fade-in reaching `opacity: 1` (S2925 ×2). The no-overlay Escape handler moved from a JSX `onKeyDown` prop on the `role="dialog"` div to an imperative listener on the same node (S6847), which also removed the deprecated `JSX.TargetedKeyboardEvent` type (S1874). Semantics unchanged: Escape closes only while focus is inside the survey.

## Linear ticket

Fixes ENG-2304

## How this was tested

- `pnpm test` ✅ — 24/24 turbo tasks, 598 test files. `packages/surveys` alone: 665 tests pass.
- `tsc --noEmit` (surveys + web) ✅, `eslint` on changed files ✅, `pnpm format:check` ✅.
- **New e2e spec** `apps/web/playwright/survey-modal-non-blocking.spec.ts` ✅ — 2/2. Covers both directions: `overlay:none` must not block, `overlay:dark` must still be modal. New Prisma seeder `playwright/utils/app-survey.ts` provisions an app survey + code action without driving the editor UI.
- **Negative control, per assertion.** A whole-spec revert only proves the *first* assertion bites, because Playwright aborts the test there. So each assertion was isolated into its own test and run against a bundle rebuilt from `b204241c9` (verified pre-fix: the `closest("#fbjs")` guard absent from the built bundle). Result: **every isolated assertion fails pre-fix and passes post-fix** — host-input focus, Escape `defaultPrevented`, full-string drag selection, focus-outside-survey after the drag, and the Cmd/Ctrl+Enter chord.
- Review (thanks @itsjavi) caught that the original Cmd/Ctrl+Enter assertion could not fail: both keydown listeners sit on `document` in the bubble phase and the host page registers first, so an inline `defaultPrevented` read always saw `false`. Auditing the rest of the spec for the same pattern found three more weak assertions. All fixed in `85965d457`; details in the review thread.
- **Manual, against the real JS playground on a local stack** (real `formbricks.umd.cjs` + real `surveys.umd.cjs`, app survey, bottomRight, overlay none) — 10/10 checks:

  | Check | Before | After |
  | --- | --- | --- |
  | Drag-select host prose | 0 chars | full string |
  | Focus after survey opens | survey `textarea` | host `button` (untouched) |
  | Typing after survey opens | goes to the survey | goes to the host input |
  | Tab ×6 from host page | all 6 inside survey | all 6 on host controls |
  | `aria-modal` | `"true"` | absent |
  | Escape on host page | closed the survey | host keeps it, survey stays open |
  | Cmd+Enter in host textarea | survey submitted, `defaultPrevented: true` | host handler runs, survey unchanged |
  | Survey still usable | ✅ | ✅ (types, Escape closes it) |

- **Overlay regression check on the real bundle** — 8/8: `overlay:dark` still sets `aria-modal="true"`, still takes focus on open, still snaps focus back when pulled out, layer still `pointer-events: auto`. `overlay:none` inverts on all four.
- **Follow-up commits** (`5dfe01183`, `9e8edbb5a`): `packages/js-core` vitest 40/40 (setup + widget), `packages/surveys` focus-trap 13/13, `tsc --noEmit` (js-core) ✅, eslint on every changed file ✅, `pnpm i18n` regenerated all widget locales for the new string. E2E spec re-run locally after both commits: 2/2 ✅ — it now also pins the announcement contract: the region exists (`role="status"`) before any survey opens, receives the message on open, clears on close, and stays silent for `overlay:dark`.
- **Sonar coverage follow-up** (`8ef4efc6a`): added direct unit coverage for the SDK and renderer live-region helpers, including creation, accessibility attributes, existing-region reuse, setup invocation, and SSR safety. Full package coverage passes at 667/667 surveys tests and 291/291 js-core tests. The new `live-region.ts` reports 12/12 covered lines and 2/2 covered branches; all added executable lines and branches in Sonar-included `.ts` code are covered.
- **Manual re-check on the local playground with rebuilt bundles**: region mounted at `setup()`; announcement lands ~100 ms after open; cleared on close; re-open re-announces; Escape-inside still closes after the S6847 listener refactor; overlay mode silent with `aria-modal` and the focus trap intact.

No screenshots: there is **no visual change**. The widget renders pixel-identically in both states — only focus and keyboard behaviour changed, which the measurements above capture.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] In a workspace whose widget settings are Placement = Bottom Right and Overlay = None, publish an app survey triggered by a code action. On a test page running the SDK, click into a text field, type some text, then fire the action → the survey appears bottom-right and **the cursor stays in your field**; continuing to type keeps going into your field, not the survey.
- [ ] With the survey open, drag across a paragraph of page text → the text highlights normally and the highlight follows the mouse.
- [ ] With the survey open, press Tab repeatedly from a field on the page → focus moves through the page's own controls instead of cycling inside the survey.
- [ ] With the survey open, click into the survey's answer box and type → it works normally. Press Escape → the survey closes.
- [ ] With the survey open and focus on the page (not the survey), press Escape → the survey stays open and the page's own Escape behaviour (e.g. closing its own dialog) works.
- [ ] With focus in a page textarea that has its own Cmd/Ctrl+Enter shortcut, press Cmd/Ctrl+Enter → the page's shortcut runs and the survey does **not** advance.
- [ ] Set Overlay to Dark or Light and repeat → the old behaviour is expected here: the page is dimmed, the survey takes focus on open, focus stays inside it, and Escape closes it.
- [ ] Announcement, no screen reader needed: after the host page loads the SDK, DevTools → Elements shows an empty `<div id="formbricks-live-region" role="status">` at the end of `<body>` (present **before** any survey opens). Open the no-overlay survey → its text becomes "A survey has opened at the end of the page.". Close the survey → the text clears. Reopen it → the text is set again on every open, not just the first.
- [ ] Screen reader (VoiceOver / NVDA): with Overlay = None, opening the survey plays that announcement politely (it waits for current speech, does not interrupt) and focus does not move; the rest of the page stays reachable while the survey is open. With an overlay, focus moves into the survey, the page outside is not reachable, and no extra announcement plays.
- [ ] Bundle freshness when testing locally: `/js/*.umd.cjs` is served with `max-age=3600`, so hard-refresh (or DevTools → disable cache) before judging — a stale `formbricks.umd.cjs` shows the old behaviour (no region at all).
- [ ] Link surveys (`/s/<id>`) and the survey preview in the editor: unchanged — autofocus and Cmd/Ctrl+Enter still behave as before.

**Preconditions / test data**

- An app-type survey with status In Progress, a code action trigger, and Display Option set to "respond multiple times" so it can be reopened.
- A host page loading the JS SDK that has selectable text, at least one input, and ideally its own Escape / Cmd+Enter shortcut to prove we no longer swallow them.
- Widget Placement and Overlay live on the workspace's widget settings, not on the survey.

**Risks & regressions**

- Highest risk is the overlay path: confirm surveys **with** an overlay are unchanged (focus taken on open, focus contained, Escape closes). The e2e spec pins this, but it is worth eyes.
- `center` placement combined with Overlay = None is now treated as non-modal. It still looks like a centered modal to a sighted user. Flagging for a product decision — the editor labels it "Centered Modal" while overlay is an independent control.
- Keyboard-only discoverability: a no-overlay survey is no longer auto-focused, so a keyboard user must Tab to it (the container is appended last to `<body>`, so it is last in tab order). Screen-reader users now hear the open announced through the status region; sighted keyboard users still get no cue beyond the survey appearing.
- Not addressed here, pre-existing: below 640px the survey is a full-width bar pinned to the bottom (`survey-container.tsx:106`, placement classes are `sm:`-only), so it can cover a sticky footer on mobile. Deliberate design from 2023; separate ticket.
- Not addressed here, found while investigating and worth their own tickets: `hasYieldedToPage` never resets once the trap gives up (`use-focus-trap.ts:184/190`); `focusScopesStack.add` pauses the outgoing scope but nothing resumes it (`:27-39`); `removeWidgetContainer()` detaches the container without unmounting Preact, leaking the trap's `document` listeners (`js-core/src/lib/survey/widget.ts:277-279`).

- No migrations or env changes.
- On deploy, purge the CDN entries for **both** `/js/surveys.umd.cjs` and `/js/formbricks.umd.cjs` (the live-region commit changed js-core too). `apps/web/next.config.mjs` serves `/js/(.*)` with `s-maxage=2592000`, so embedding sites would otherwise keep the old bundles for up to 30 days. Sites loading the SDK from npm instead of `/js/` get the region via the surveys-side fallback until they upgrade.

### DIFF
--- a/apps/web/playwright/survey-modal-non-blocking.spec.ts
+++ b/apps/web/playwright/survey-modal-non-blocking.spec.ts
@@ -1,0 +1,188 @@
+import { expect } from "@playwright/test";
+import http from "http";
+import { type AddressInfo } from "net";
+import { test } from "./lib/fixtures";
+import { seedAppSurvey } from "./utils/app-survey";
+
+/**
+ * A corner survey with no overlay must not take anything from the page it sits on.
+ *
+ * The widget used to arm a full focus trap for every modal survey regardless of
+ * overlay (packages/surveys/src/components/wrappers/survey-container.tsx), so a
+ * bottom-right survey stole the caret, wiped text selections mid-drag, trapped Tab,
+ * claimed Escape and Cmd/Ctrl+Enter document-wide, and told screen readers to ignore
+ * the host page. Mouse clicks kept working the whole time, which is what made it hard
+ * to spot. These specs pin the non-blocking contract, and the overlay case pins the
+ * modal behaviour that must survive it — otherwise someone "simplifies" the gate away.
+ */
+
+declare global {
+  interface Window {
+    formbricks: {
+      setup: (config: { workspaceId: string; appUrl: string }) => Promise<void>;
+      track: (name: string) => Promise<void>;
+    };
+    __hostEscapes: number;
+    __hostChordDefaultPrevented: boolean | null;
+  }
+}
+
+// Prose long enough that a drag across it selects an unmistakable number of characters.
+const HOST_PROSE =
+  "The quick brown fox jumps over the lazy dog while the host page stays fully usable underneath.";
+
+const hostPage = (workspaceId: string) => `<!doctype html>
+<html>
+  <head>
+    <script type="text/javascript">
+      !(function () {
+        var t = document.createElement("script");
+        (t.type = "text/javascript"), (t.async = !0), (t.src = "http://localhost:3000/js/formbricks.umd.cjs");
+        t.onload = function () {
+          window.formbricks.setup({ workspaceId: "${workspaceId}", appUrl: "http://localhost:3000" });
+        };
+        var e = document.getElementsByTagName("script")[0];
+        e.parentNode.insertBefore(t, e);
+      })();
+    </script>
+  </head>
+  <body style="background-color: #fff; font-family: sans-serif">
+    <p id="prose" style="max-width: 480px">${HOST_PROSE}</p>
+    <input id="host-input" aria-label="Host input" />
+    <textarea id="host-textarea" aria-label="Host textarea"></textarea>
+    <button id="host-button" type="button" onclick="window.__hostButtonClicks=(window.__hostButtonClicks||0)+1">
+      Host button
+    </button>
+    <script>
+      window.__hostEscapes = 0;
+      window.__hostChordDefaultPrevented = null;
+      document.addEventListener("keydown", function (e) {
+        if (e.key === "Escape") window.__hostEscapes++;
+        if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
+          window.__hostChordDefaultPrevented = e.defaultPrevented;
+        }
+      });
+    </script>
+  </body>
+</html>`;
+
+test.describe("App survey widget does not block the host page", () => {
+  let server: http.Server;
+  let hostUrl = "";
+
+  test.setTimeout(3 * 60 * 1000);
+
+  // The workspace comes in on the query string so the two tests can share one server
+  // without sharing mutable state, and the port is ephemeral so parallel workers do
+  // not fight over it.
+  test.beforeAll(async () => {
+    server = http.createServer((req, res) => {
+      const workspaceId = new URL(req.url ?? "/", "http://localhost").searchParams.get("workspaceId") ?? "";
+      res.writeHead(200, { "Content-Type": "text/html" });
+      res.end(hostPage(workspaceId));
+    });
+    await new Promise<void>((resolve) => {
+      server.listen(0, () => {
+        hostUrl = `http://localhost:${(server.address() as AddressInfo).port}`;
+        resolve();
+      });
+    });
+  });
+
+  test.afterAll(async () => {
+    await new Promise((resolve) => server.close(resolve));
+  });
+
+  test("overlay:none leaves the host page fully usable", async ({ page, users }) => {
+    const seeded = await seedAppSurvey(users, { overlay: "none", placement: "bottomRight" });
+
+    await page.goto(`${hostUrl}/?workspaceId=${seeded.workspaceId}`);
+    await page.waitForFunction(() => Boolean(window.formbricks), null, { timeout: 120000 });
+
+    // Park the caret in a host field first — this is the customer's scenario: a survey
+    // firing mid-form must not pull the user out of what they are typing.
+    await page.locator("#host-input").fill("BEFORE");
+
+    await page.evaluate((key) => window.formbricks.track(key), seeded.actionKey);
+
+    const dialog = page.locator("#fbjs [role='dialog']");
+    await expect(dialog).toBeVisible({ timeout: 120000 });
+    // Let the card transition settle so any deferred focus has had its chance to land.
+    await page.waitForTimeout(1500);
+
+    // The survey must not have taken the caret.
+    await expect(page.locator("#host-input")).toBeFocused();
+    await page.keyboard.type("-AFTER");
+    await expect(page.locator("#host-input")).toHaveValue("BEFORE-AFTER");
+
+    // A survey that does not block the page must not claim to be a modal dialog:
+    // aria-modal makes assistive tech ignore everything outside it.
+    await expect(dialog).not.toHaveAttribute("aria-modal", /.*/);
+
+    // Text on the host page must stay selectable — the focus trap used to wipe the
+    // selection ~17ms into the drag by pulling focus back into the survey.
+    const prose = page.locator("#prose");
+    const box = await prose.boundingBox();
+    if (!box) throw new Error("host prose has no bounding box");
+    const midY = box.y + box.height / 2;
+    await page.mouse.move(box.x + 2, midY);
+    await page.mouse.down();
+    for (let step = 1; step <= 6; step++) {
+      await page.mouse.move(box.x + 2 + ((box.width - 6) * step) / 6, midY, { steps: 3 });
+    }
+    await page.mouse.up();
+    const selectedLength = await page.evaluate(() => String(window.getSelection() ?? "").length);
+    expect(selectedLength).toBeGreaterThan(20);
+
+    // Tab must be able to move through the host page instead of cycling in the survey.
+    await page.locator("#host-input").focus();
+    await page.keyboard.press("Tab");
+    await expect(page.locator("#host-textarea")).toBeFocused();
+
+    // Escape belongs to the host page while focus is on the host page.
+    await page.keyboard.press("Escape");
+    await expect(dialog).toBeVisible();
+    expect(await page.evaluate(() => window.__hostEscapes)).toBeGreaterThan(0);
+
+    // Same for the Cmd/Ctrl+Enter chord: the survey must not swallow it.
+    await page.locator("#host-textarea").focus();
+    await page.keyboard.press("ControlOrMeta+Enter");
+    expect(await page.evaluate(() => window.__hostChordDefaultPrevented)).toBe(false);
+    await expect(dialog).toBeVisible();
+
+    // The survey itself is still fully usable — it just does not grab anything.
+    const surveyInput = page.locator("#fbjs input[type='text'], #fbjs textarea").first();
+    await surveyInput.click();
+    await surveyInput.fill("still works");
+    await expect(surveyInput).toHaveValue("still works");
+
+    // Escape closes the survey when focus is inside it.
+    await page.keyboard.press("Escape");
+    await expect(dialog).toBeHidden();
+  });
+
+  test("overlay:dark keeps the modal behaviour", async ({ page, users }) => {
+    const seeded = await seedAppSurvey(users, { overlay: "dark", placement: "center" });
+
+    await page.goto(`${hostUrl}/?workspaceId=${seeded.workspaceId}`);
+    await page.waitForFunction(() => Boolean(window.formbricks), null, { timeout: 120000 });
+
+    await page.locator("#host-input").fill("BEFORE");
+    await page.evaluate((key) => window.formbricks.track(key), seeded.actionKey);
+
+    const dialog = page.locator("#fbjs [role='dialog']");
+    await expect(dialog).toBeVisible({ timeout: 120000 });
+    await page.waitForTimeout(1500);
+
+    // A survey that does block the page is a real modal: it announces itself as one,
+    // takes focus, and keeps it.
+    await expect(dialog).toHaveAttribute("aria-modal", "true");
+    await expect(page.locator("#host-input")).not.toBeFocused();
+
+    await page.locator("#host-input").focus();
+    const focusStayedInSurvey = await page.evaluate(() =>
+      Boolean(document.getElementById("fbjs")?.contains(document.activeElement))
+    );
+    expect(focusStayedInSurvey).toBe(true);
+  });
+});
--- a/apps/web/playwright/survey-modal-non-blocking.spec.ts
+++ b/apps/web/playwright/survey-modal-non-blocking.spec.ts
@@ -149,7 +149,9 @@ test.describe("App survey widget does not block the host page", () => {
 
     // It takes nothing from the page, so the open is announced through the pre-existing status
     // region instead of a focus move — otherwise screen-reader users get no signal it appeared.
-    await expect(liveRegion).toHaveText("A survey has opened. Press Tab to reach it.");
+    // The message states where the survey is (the widget mounts last in <body>) rather than
+    // promising a keystroke: Tab follows DOM order, so one Tab press would not reach it.
+    await expect(liveRegion).toHaveText("A survey has opened at the end of the page.");
 
     // Text on the host page must stay selectable — the focus trap used to wipe the
     // selection ~17ms into the drag by pulling focus back into the survey.
--- a/apps/web/playwright/survey-modal-non-blocking.spec.ts
+++ b/apps/web/playwright/survey-modal-non-blocking.spec.ts
@@ -127,8 +127,9 @@ test.describe("App survey widget does not block the host page", () => {
 
     const dialog = page.locator("#fbjs [role='dialog']");
     await expect(dialog).toBeVisible({ timeout: 120000 });
-    // Let the card transition settle so any deferred focus has had its chance to land.
-    await page.waitForTimeout(1500);
+    // The card fades in over 500ms (survey-container.tsx); opacity 1 marks the survey fully open, by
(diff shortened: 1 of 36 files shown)


apps/web/playwright/survey-modal-non-blocking.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    // Escape belongs to the host page while focus is on the host page.
+    await page.keyboard.press("Escape");
+    await expect(dialog).toBeVisible();
+    expect(await page.evaluate(() => window.__hostEscapes)).toBeGreaterThan(0);
+
+    // Same for the Cmd/Ctrl+Enter chord: the survey must not swallow it.
+    await page.locator("#host-textarea").focus();
+    await page.keyboard.press("ControlOrMeta+Enter");
+    expect(await page.evaluate(() => window.__hostChordDefaultPrevented)).toBe(false);
+    await expect(dialog).toBeVisible();
```
itsjavi: **The Cmd/Ctrl+Enter assertion passes with or without the fix.**

Both keydown listeners sit on `document` in the bubble phase: the host page's inline script registers at page load, `SubmitButton`'s registers when the survey mounts (after `formbricks.track()`). Same target, same phase, so they fire in registration order — host first. `e.defaultPrevented` is therefore read *before* the survey's handler has had a chance to call `preventDefault()`, and reads `false` either way.

I verified this in Chromium rather than reasoning about it, reproducing the spec's exact registration order:

```js
document.addEventListener('keydown', e => { window.hostSaw = e.defaultPrevented; });   // host, first
document.addEventListener('keydown', e => { e.preventDefault(); window.surveyRan = true; }); // survey, later
```
```
host observed defaultPrevented = false
survey listener DID preventDefault = true
```

The negative control in the description doesn't cover this, because the reverted run fails earlier at line 118 (`#host-input` not focused) and never reaches line 154.

Line 155 doesn't rescue it either: before the fix, Cmd+Enter submits the open-text element and advances to the ending card, and `#fbjs [role='dialog']` is still visible there — so `toBeVisible()` also passes in the broken case. As written, the chord regression is the one behaviour in this PR with no test behind it.

Two ways to make it bite, either is fine:

- Read `defaultPrevented` after dispatch completes, from inside the host handler:
  ```js
  document.addEventListener("keydown", function (e) {
    if ((e.metaKey || e.ctrlKey) && e.key === "Enter") {
      setTimeout(function () { window.__hostChordDefaultPrevented = e.defaultPrevented; }, 0);
    }
  });
  ```
  The event object survives dispatch, so this observes the final state — `true` pre-fix, `false` post-fix.
- Or assert the user-visible contract instead: the survey did not advance. Checking that the survey's own text input is still on screen (rather than just *some* dialog) distinguishes the question card from the ending card.

---

itsjavi (commented):
Read the diff against the branch rather than the description, and re-derived the load-bearing claims instead of trusting them. The diagnosis is right and unusually well argued: the full-screen layer really is already `pointer-events: none` without an overlay (`survey-container.tsx:107`), so the reported "container blocks the page" was never the CSS — it was `useFocusTrap` arming for every modal survey regardless of overlay, and its `document`-level listeners pulling focus back in. Gating the trap on `hasOverlay` to match the gate the click-outside effect two lines below already used is the minimal correct fix.

**Verified rather than assumed**

- `useFocusTrap`'s effect early-returns on `!enabled` (`use-focus-trap.ts:145`), so the no-overlay path registers no listeners at all — nothing left behind to clean up.
- The `hasOverlay` derivations don't disagree across the two files, which is what I most expected to be wrong. `render-survey.tsx` computes `props.overlay && props.overlay !== "none"` (falsy when `overlay` is undefined) while `survey-container.tsx` computes `overlay !== "none"` — but the container defaults the param to `"none"` (`:22`), so an undefined overlay lands on "no overlay" on both sides. The gate holds for the schema default.
- "Focus still follows Next/Back" holds. `autoFocusEnabled` (`survey.tsx:252`) is now pinned false for no-overlay surveys, and `block-conditional.tsx:439` passes it straight through to the first element — but the card-level `shouldFocusOnMount` keeps the `|| hasUserNavigatedRef.current` term (`survey.tsx:1238`) and calls `focusFirstControl` on the container (`block-conditional.tsx:123-128`), so after Next/Back focus still lands on the first control. The two paths are redundant, not contradictory.
- `#fbjs` is emitted by both the modal and inline branches of `SurveyContainer`, so `submit-button.tsx`'s `closest("#fbjs")` resolves in every render mode. No shadow root anywhere in `packages/surveys` or `js-core`, so `document.activeElement` is directly comparable — the containment check can't be defeated by a shadow boundary.
- `overlay: none` + `placement: bottomRight` really are the schema defaults (`main.prisma:616,618`), so this does change behaviour for most existing app surveys — correctly called out as needing a release note.

**One finding**, inline on the e2e spec: the Cmd/Ctrl+Enter assertion can't fail. The host page's `document` listener is registered before the survey's, so it reads `defaultPrevented` before the survey's handler runs and gets `false` either way — I reproduced the ordering in Chromium. The negative control doesn't catch it because the reverted run fails earlier at line 118 and never reaches that assertion. Details and two fixes in the inline comment. Everything else in the spec is solid, and testing both directions (`overlay:none` must not block, `overlay:dark` must still be modal) is the right shape — that second test is what stops someone "simplifying" the gate away later.

Not a finding, just noting the tradeoff you've taken: the new container-scoped Escape handler calls `preventDefault()` but not `stopPropagation()`, so while focus is *inside* the survey, Escape both closes the survey and still reaches the host page's own Escape handler. That's the safer default of the two, and the reverse would be worse — just worth knowing it's the behaviour, since the PR frames Escape as no longer being taken from the host page.

CI is green on all 17 checks at `1b50226a`, including the CodeQL fix for the spec's host page, it merges cleanly and is not behind `main`. No open review threads.

**Needs human review** @itsjavi

- The new Playwright spec can't be run here — it needs a dev server, seeded DB and the built `formbricks.umd.cjs` / `surveys.umd.cjs` bundles, none of which are up in this environment. The static reasoning above holds, but the manual checklist in the description is still a human's.
- The screen-reader step specifically: `aria-modal` disappearing is the right call per ARIA APG, but "the rest of the page is still reachable with Overlay = None, and not reachable with an overlay" wants a real VoiceOver/NVDA pass, not a DOM assertion.
- The `center` + `overlay: none` question the author raises is a product decision, not a code one — a survey that still *looks* like a centred modal but no longer behaves as one needs someone to say whether that's acceptable or whether `center` should imply an overlay.
- The CDN purge for `/js/surveys.umd.cjs` on deploy (`s-maxage=2592000`) is a real operational step — embedding sites keep the old bundle for up to 30 days otherwise.

---

apps/web/playwright/survey-modal-non-blocking.spec.ts
pandeymangg: Confirmed and fixed in 85965d457. You were right on both counts, and I reproduced the ordering rather than taking it on trust:

```
syncRead: false        <- what the spec read
deferredRead: true     <- the real final state
surveyListenerRan: true
```

Applied both of your suggestions, since they pin different things: the host page now defers the read with `setTimeout(…, 0)`, and the assertion moved from "some dialog is visible" to "the survey's text input is still visible" — your ending-card point was exactly right.

**Your note about the negative control was the more valuable half of the finding.** A test that aborts at its first failing assertion only ever proves that *first* assertion bites. I re-ran the audit that way, with each assertion isolated in its own test against a rebuilt pre-fix bundle, and found three more of the same class:

- `__hostEscapes > 0` — same bug one assertion lower. A keydown counter on `document` increments before the trap's handler runs and a later `preventDefault()` cannot retract it, so it read `>= 1` either way. Now a deferred `defaultPrevented` read, which is `true` pre-fix (`use-focus-trap.ts:274-277` preventDefaults every unmodified Escape document-wide).
- The drag-select threshold was a latent flake, not a vacuity bug — it does fail pre-fix (4 chars). But `midY` samples only the last wrapped line, so the count is pure font metrics. Measured: 30 chars at 16px, 24 at 15px, **18 at 14px** — one wrap step from failing `> 20` on a CI box with narrower fonts. Now drags corner to corner and compares the whole string; measured 94/94 at every size from 13-18px, still 3-4 pre-fix.
- `not.toBeFocused()` in the `overlay:dark` test is satisfied by focus being on `<body>`, so it did not assert what its comment claimed. Now the positive containment check.

Also fixed a real trap in the seeder that had nothing to do with your comment: `recontactDays` was left null, which falls through to `Workspace.recontactDays` (`main.prisma:613`, default 7). The current tests survive only because each gets a fresh context and `lastDisplayAt` is null. Now seeded as `0`.

Verification: all four assertions isolated, run against a pre-fix bundle rebuilt from `b204241c9` — **4 failed**; against the fixed bundle — **4 passed**.

On your Escape note — not a finding, but you're right that the PR description overstated it. `preventDefault()` without `stopPropagation()` means that with focus *inside* the survey, Escape both closes the survey and still reaches the host page's handler. I've corrected the description rather than change the behaviour, since as you say the reverse would be worse.

Your four "needs human review" items all stand and I'm not claiming any of them: the VoiceOver/NVDA pass, the `center` + `overlay: none` product call, the CDN purge on deploy, and a human running the manual checklist.

itsjavi (dismissed):
[eng-2304-smoke.html](https://github.com/user-attachments/files/30928172/eng-2304-smoke.html)
[eng2304-smoke-seed.ts](https://github.com/user-attachments/files/30928252/eng2304-smoke-seed.ts)

I reviewed it using a smoke test app. The issues are gone, and the focus trap is only present on modals.

image


The focus trap now arms only when the survey has an overlay: enabled: isModal && isOpen && hasOverlay (survey-container.tsx:34), which is correct.

For anything that appears without a user gesture (e.g. a popover in the corner), a proper ARIA announcement is the appropriate mechanism, and moving focus (focus trap) is the anti-pattern. So your fix is correct, except that is missing some ARIA props

itsjavi:
I am going to fix sonar issues and add proper ARIA announcements that are missing

itsjavi (approved):
I applied some fixes. LGTM

PR #8806, opened 2026-08-10
feat(embedded-data): add the single embedded-value resolver and readable-fields helper

itsjavi:
## What & why

Recall, logic, export and the pickers each carry their own private answer to "where does a field's value live" — and with Embedded Data definitions moving into the workspace-level tables (ENG-1833, #8687), every one of those surfaces would need to relearn it. This PR builds the single read seam they will all use instead, as pure functions in `@formbricks/types/embedded-data-resolver`. **Nothing calls it yet**: ENG-1837 repoints the call sites, ENG-1839 supplies the reserved catalog contents.

- **`resolveEmbeddedValue(ref, response)`** — one field's value, by source: `computed` → `response.variables[storageKey]`, `ingested` → `response.data[storageKey]`, reserved catalog entry → its typed accessor (`meta.*`, `ttc._total`, top-level fields; an accessor that throws reads as "nothing captured" rather than crashing the caller's loop over the catalog). A `locked` ingested field ignores `response.data` entirely — the default IS the value; honoring a stored value would let a crafted URL override what locking promised. Missing or uncoercible values fall back to `defaultValue` (`null` = no default) → unset, and every tier passes through one exported `coerceToEmbeddedDataType`, so the result always agrees with the declared `dataType` or is `undefined`. The ref is a two-member union with `?: never` exclusion props — a stored field is only addressable with its link's `storageKey`, a reserved entry can never carry one, and a mixed literal is a compile error (pinned by a `@ts-expect-error` test).
- **`projectReservedValues(entries, response)`** — resolves catalog entries into a `Record<string, string | number>` that is statically assignable into both maps recall and logic already read (`variables[...]` / `responseData[...]`; compile-asserted in the tests). This is the ticket's "projection, not a special branch" design: reserved values become reachable by bare name with zero per-consumer special-casing — which matters because `response.meta` is not otherwise passed into those resolvers at all. Absent values are omitted (recall falls to its fallback text, logic sees `undefined` — exactly like an absent hidden field today); falsy-but-present values (`false` → `"false"`, `""`, `0`) stay in the map.
- **`RESERVED_FIELD_CATALOG`** — deliberately empty. This PR ships the mechanism (typed entries whose `read` accessor makes a bad "path" a compile error instead of a silent `undefined`); ENG-1839 decides the contents. Call sites can wire the constant now and light up when entries land.
- **`listReadableFields(input)`** — the grouped "what can I reference, and by which key" enumeration: **question · embeddedData · reserved · contactAttribute** (contact attributes stay their own clearly-labelled group). Each entry carries the exact reference key — element id, link `storageKey` (never the library `key`; the two can differ), catalog entry name, contact attribute key — and labels fall back to the key whenever blank.
- **`deriveLegacyEmbeddedData(survey)`** — maps un-backfilled `variables` / `hiddenFields.fieldIds` into the same field/link pairs per migration-spec §8, so ENG-1837 has a fetch-free fallback for surveys whose rows don't exist yet (local dev before the backfill).

**Four deliberate behavioral deltas are documented as an "ENG-1837 swap checklist"** on `resolveEmbeddedValue`, so the repointing inherits them consciously instead of discovering them in production: (a) today `getVariableValue` evaluates a number variable as `Number(value) || 0`, so a non-numeric string sitting in a number variable (a state `performCalculation`'s `assign` produces today) is `0` to logic, while here it is a coercion failure resolving to defaultValue-or-unset; (b) both recall paths render raw strings type-obliviously today; (c) on a key present in both maps today's consumers disagree with each other (packages/surveys lets `data` win, apps/web lets `variables` win) while here the declared source is exclusive; (d) a response missing a variable's key entirely resolves to the field's `defaultValue` here, where today logic yields `0`/`""` and recall falls to fallback text. The date rule is also pinned precisely: ISO date or UTC datetime only — the exact subset `ZEmbeddedData` accepts for date defaults, excluded offsets and padding included — with a cross-schema corpus test that goes red if either side drifts.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1836/single-resolver-resolveembeddedvalue-list-all-readable-fields-helper

## How this was tested

- `pnpm --filter @formbricks/types test` — 4 files, **192 passed** (103 new in `embedded-data-resolver.test.ts`, 89 pre-existing). The new suite covers: per-source resolution incl. dual-key precedence cases; `locked` (incl. locked-with-no-default and locked-ignored-on-computed); missing → default → unset; the full coercion matrix (string 12 / number 15 / boolean 10 / date 11 cases) incl. uncoercible inputs per type, partially-numeric strings (`"25 seats"`), both non-scalar response shapes, and a corpus test binding the read-side date rule to `ZEmbeddedData`'s default-value rule; projection over nested `meta.userAgent.*`, `ttc._total`, top-level fields, absent-key omission, Date→ISO coercion, uncoercible-entry omission, falsy-but-present retention, and throwing-accessor isolation; four-group enumeration with per-entry reference keys, blank-label and blank-locale fallbacks, multi-token recall flattening, dirty-i18n degradation, and `de` localization; §8 legacy derivation fed end-to-end through resolver and enumerator; and a `@ts-expect-error` pin that the mixed ref shape stays unrepresentable.
- `pnpm --filter @formbricks/types typecheck` — clean; this package's tsconfig includes `*.test.ts`, so the pin above is live on every typecheck run. Note `pnpm typecheck` runs in no CI workflow, so the compile-level pins bite on local/agent runs, not in the merge gate.
- `pnpm typecheck` (root, all packages) — clean.
- `pnpm --filter @formbricks/types lint` — 0 errors; 4 pre-existing unused-disable warnings in untouched `surveys/*` files.
- `npx prettier --check` on both new files — pass. (The root has no `format:check` script, only the mutating `format`, so the check was scoped.)
- Nothing imports the module: the module name appears only in its own test's import; no other tracked file references it and none was modified, so there is no production behaviour change.
- Two adversarial review rounds. Round 1 (pre-PR): consumer-semantics comparison against `replaceRecallInfo`, `parseRecallInfo`, `getLeftOperandValue`/`performCalculation` — findings applied (swap checklist, blank-label fallbacks, `?: never` exclusion props). Round 2 (on the PR head, three parallel finders incl. a sandboxed mutation battery with red controls): 10 suite-surviving mutations → the 13 tests above; executed zod probes → the date-rule and swap-checklist doc corrections; verified-against-source citations → five JSDoc corrections (boolean provenance, label-mirror attribution, the `hiddenFields.enabled` link-survey gap, `blocks`-on-TSurvey, caller tense) plus the throwing-accessor hardening and the `toElementLabel` dirty-default guard.

## Breaking changes

None

## QA / Test Plan

**How to test**

Nothing user-visible ships here — two new files that no code imports yet. The useful manual check is that nothing regressed around the surfaces this seam will later serve:

- [ ] Open a survey with variables and hidden fields in the editor → both cards load and list their fields exactly as before
- [ ] Use a variable and a hidden field in recall (`@`) and in a logic condition, then submit a response → the values resolve as before
- [ ] Export that survey's responses → variable and hidden-field columns appear with the same headers and values as before

**Preconditions / test data**

- Any workspace with a survey that has at least one variable and one hidden field, plus a response or two.

**Risks & regressions**

- Low: additive files only; no existing file changed; the module is unreachable from production code paths. The area to re-check is simply that the app still builds and existing survey/response flows behave unchanged.

- none

### DIFF
--- a/packages/types/embedded-data-resolver.ts
+++ b/packages/types/embedded-data-resolver.ts
@@ -1,6 +1,7 @@
 import { z } from "zod";
 import type { TContactAttributeKey } from "./contact-attribute-key";
 import type { TEmbeddedData, TEmbeddedDataType, TSurveyEmbeddedData } from "./embedded-data";
+import { toDesiredEmbeddedFields } from "./embedded-data-mapping";
 import type { TI18nString } from "./i18n";
 import type { TResponse } from "./responses";
 import { formatSnakeCaseToTitleCase } from "./safe-identifier";
@@ -278,6 +279,9 @@ export const resolveEmbeddedValue = (
  * maps: absent values are omitted (recall then falls back to its fallback text and logic sees
  * `undefined`, exactly like an absent hidden field today), and booleans are stringified to
  * `"true"`/`"false"` — what `String()` would render anyway, in a map type that has no boolean slot.
+ *
+ * An entry name that matches an element id or hidden field name shadows it in the merged map, and
+ * `FORBIDDEN_IDS` guards none of the names §9b proposes — so ENG-1839 picks names knowing that.
  */
 export const projectReservedValues = (
   entries: readonly TReservedFieldCatalogEntry[],
@@ -329,7 +333,10 @@ export interface TListReadableFieldsInput {
   languageCode?: string;
 }
 
-/** Matches one full recall token (`#recall:<id>/fallback:<text>#`) — the editor's storage syntax. */
+/**
+ * Matches one full recall token (`#recall:<id>/fallback:<text>#`) — the editor's storage syntax.
+ * `.replace()` only: `.test()`/`.exec()` would carry `lastIndex` across calls on this shared instance.
+ */
 const RECALL_TOKEN_REGEX = /#recall:[A-Za-z0-9_-]+\/fallback:[^#]*#/g;
 
 /**
@@ -402,9 +409,10 @@ export const listReadableFields = (input: TListReadableFieldsInput): TReadableFi
  * ENG-1837 serve surveys whose rows haven't been backfilled (migration spec §8), with no fetch
  * logic here.
  *
- * Per §8: a variable becomes a `computed` field of its declared type whose `defaultValue` is the
- * variable's value, addressed by its existing cuid (recall tokens and stored responses already use
- * it); a hidden field becomes an `ingested` string field with no default, addressed by its name.
+ * The §8 rules live in `toDesiredEmbeddedFields`, shared with ENG-1978's write bridge and ENG-1835's
+ * backfill; this only reshapes them into `{field, link}` pairs and adds `locked: false`, which has no
+ * legacy equivalent.
+ *
  * `hiddenFields.enabled` is deliberately ignored: recall and logic consult `fieldIds` alone today,
  * and ingestion is split on the flag — the js-core SDK drops hidden fields when disabled, while the
  * link-survey URL path fills them regardless (`getHiddenFieldsFromSearchParams` receives only
@@ -414,32 +422,8 @@ export const listReadableFields = (input: TListReadableFieldsInput): TReadableFi
 export const deriveLegacyEmbeddedData = (survey: {
   variables: TSurveyVariables;
   hiddenFields: TSurveyHiddenFields;
-}): TLinkedEmbeddedField[] => {
-  const fromVariables = survey.variables.map(
-    (variable): TLinkedEmbeddedField => ({
-      field: {
-        name: variable.name,
-        source: "computed",
-        dataType: variable.type === "number" ? "number" : "string",
-        defaultValue: variable.value,
-        locked: false,
-      },
-      link: { storageKey: variable.id },
-    })
-  );
-
-  const fromHiddenFields = (survey.hiddenFields.fieldIds ?? []).map(
-    (fieldId): TLinkedEmbeddedField => ({
-      field: {
-        name: fieldId,
-        source: "ingested",
-        dataType: "string",
-        defaultValue: null,
-        locked: false,
-      },
-      link: { storageKey: fieldId },
-    })
-  );
-
-  return [...fromVariables, ...fromHiddenFields];
-};
+}): TLinkedEmbeddedField[] =>
+  toDesiredEmbeddedFields(survey).map(({ storageKey, ...field }) => ({
+    field: { ...field, locked: false },
+    link: { storageKey },
+  }));
--- a/packages/types/embedded-data-resolver.ts
+++ b/packages/types/embedded-data-resolver.ts
@@ -10,9 +10,13 @@ import { getTextContent } from "./surveys/validation";
 
 /**
  * What a resolved Embedded Data value can be. A `date` field resolves to its ISO 8601 string, not a
- * `Date` — recall and logic operate on strings and numbers, and recall already date-formats ISO
- * strings for display. Booleans exist only for reserved fields (e.g. `finished`); the projection
- * into the recall/logic maps stringifies them, see {@link projectReservedValues}.
+ * `Date` — machine-facing values stay ISO/UTC; display formatting is the consumer's job. One gap to
+ * inherit consciously (ENG-1837/1839): today's recall formatters recognize date-ONLY strings
+ * (`isValidDateString` in packages/surveys; `formatStoredDateForDisplay` in apps/web, which runs
+ * only on the responseData branch), so a datetime resolved from a `Date` renders verbatim until the
+ * consumers learn to format it. Booleans come from `boolean`-dataType stored fields (legal on
+ * ingested rows) as well as reserved reads; {@link projectReservedValues} stringifies them for the
+ * recall/logic maps, and direct resolver callers own their own boolean handling.
  */
 export type TResolvedEmbeddedValue = string | number | boolean;
 
@@ -60,7 +64,11 @@ export interface TReservedFieldCatalogEntry {
    */
   name: string;
   dataType: TEmbeddedDataType;
-  /** Reads the raw value off a response. Coercion to `dataType` is the resolver's job, not the accessor's. */
+  /**
+   * Reads the raw value off a response. Coercion to `dataType` is the resolver's job, not the
+   * accessor's. Accessors should be total functions; one that throws is treated as "nothing
+   * captured" by the resolver rather than crashing the caller.
+   */
   read: (response: TEmbeddedValueResponse) => TReservedFieldRawValue;
 }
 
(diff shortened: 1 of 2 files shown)


packages/types/embedded-data-resolver.ts
```diff
@@ -0,0 +1,445 @@
...
+ * Per §8: a variable becomes a `computed` field of its declared type whose `defaultValue` is the
+ * variable's value, addressed by its existing cuid (recall tokens and stored responses already use
+ * it); a hidden field becomes an `ingested` string field with no default, addressed by its name.
+ * `hiddenFields.enabled` is deliberately ignored: recall and logic consult `fieldIds` alone today,
+ * and ingestion is split on the flag — the js-core SDK drops hidden fields when disabled, while the
+ * link-survey URL path fills them regardless (`getHiddenFieldsFromSearchParams` receives only
+ * `fieldIds`). Deriving from `fieldIds` alone therefore preserves today's read behavior exactly:
+ * whatever either path stored still resolves, and what nothing stored reports as unset.
+ */
+export const deriveLegacyEmbeddedData = (survey: {
```
pandeymangg, line 422: This encodes the same §8 rules as `toDesiredEmbeddedFields` in #8801 (`packages/types/embedded-data-mapping.ts`): variable → cuid / declared type / value, hidden field → name / string / null, variables first, `enabled` ignored. Identical rules, different shape — yours nests into `{field, link}` and adds `locked: false`.

The two files don't overlap, so the merge won't flag it. That's the risk: if §8 ever changes, one side gets updated and the other silently doesn't — and those two rules *are* the migration's non-breakage guarantee.

#8801 is expected to land first, so `toDesiredEmbeddedFields` will already be on the epic when you rebase. It's pure and exported from the same package, so this becomes:

```ts
toDesiredEmbeddedFields(survey).map(({ storageKey, ...field }) => ({
  field: { ...field, locked: false },
  link: { storageKey },
}));
```

The shapes line up exactly — nothing needs to change on my side.

packages/types/embedded-data-resolver.ts
```diff
@@ -0,0 +1,445 @@
...
+ * already read two maps (`variables[...]`, `responseData[...]`); merging this projection into
+ * either one makes reserved values reachable by name exactly like every other field, which matters
+ * because `response.meta` is not otherwise passed into the recall/logic resolvers at all.
+ *
+ * The value type is deliberately `string | number` so the result is assignable into both existing
+ * maps: absent values are omitted (recall then falls back to its fallback text and logic sees
+ * `undefined`, exactly like an absent hidden field today), and booleans are stringified to
+ * `"true"`/`"false"` — what `String()` would render anyway, in a map type that has no boolean slot.
+ */
+export const projectReservedValues = (
```
pandeymangg, line 286: Worth a line on what happens when an entry name collides with an element id or hidden field name in whichever map this gets merged into.

Element ids are user-editable (`update-element-id.tsx`) and only checked against `FORBIDDEN_IDS`, which contains none of the names §9b proposes — `country`, `url`, `browser`, `os`, `language` are all legal element ids today. Latent while the catalog is empty, but ENG-1839 should pick names knowing this rather than find out at 1840.

pandeymangg (commented):
Ran the suite on `657b5c58`: 192 pass, typecheck and lint clean. No blockers.

The date corpus test is the right call — two files declaring the same ISO rule privately is only safe because something binds them, and that test does. Same for the `?: never` exclusions with the `@ts-expect-error` pin: I checked the mixed ref really is a compile error.

Two notes inline, neither blocking.

Nit: `RECALL_TOKEN_REGEX` is a module-level `/g` regex. Correct with `.replace()`, which resets `lastIndex` per spec — but a later `.test()`/`.exec()` on it would be the classic stateful-regex bug. One comment would prevent that.

pandeymangg (approved):
LGTM! 🚀

packages/types/embedded-data-resolver.ts
itsjavi: Agreed, and done — #8801 merged onto the epic while this sat open, so I merged the epic in and took your snippet verbatim:

```ts
export const deriveLegacyEmbeddedData = (survey: TLegacyEmbeddedFields): TLinkedEmbeddedField[] =>
  toDesiredEmbeddedFields(survey).map(({ storageKey, ...field }) => ({
    field: { ...field, locked: false },
    link: { storageKey },
  }));
```

The shapes did line up exactly — `locked: false` is the only thing left to add, since legacy fields have no lock and the flat write-side row carries no such column.

I also moved the parameter to your `TLegacyEmbeddedFields` rather than keeping a local `{ variables; hiddenFields }` literal, so the two sides can't drift on the *input* shape either — and the optional/nullable cases your suite already covers come along for free instead of being re-tested here.

The JSDoc now points at `toDesiredEmbeddedFields` instead of restating §8; otherwise the prose becomes the new duplicate. It keeps only the read-side-specific half — why ignoring `hiddenFields.enabled` is also right for reads, given the js-core SDK drops hidden fields when disabled while the link-survey URL path fills them regardless.

Worth noting as evidence: all seven `deriveLegacyEmbeddedData` tests pass **unchanged** against the delegating implementation, which is what confirms the reshape preserves the mapping rather than approximating it. `packages/types`: 203 passed (192 + your 11), typecheck clean, lint 0 errors.

One logistics note — this session pushes to `claude/pr-8806-cr-remarks-sknlm6`, not this PR's head, so the commit (`dbae4bd`) isn't on `javier/eng-1836-…` yet. It merges cleanly; @itsjavi can fast-forward it in.

---

packages/types/embedded-data-resolver.ts
itsjavi: Good catch — documented, and your read of the validation checks out when traced:

- **Element ids** go through `ZSurveyElementId` (`packages/types/surveys/elements.ts`), which screens against `FORBIDDEN_IDS` only — `userId`, `source`, `suid`, `end`, `start`, `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`. Case-sensitive exact match, and `validateId` stays on that lenient path for renames.
- **Hidden field names** go through `validateId`'s strict path → `RESERVED_DECLARED_FIELD_NAMES`, i.e. the above lowercased plus `LINK_SURVEY_SYSTEM_PARAMS` (`sutoken`, `lang`, `preview`, `startat`, `skipprefilled`, `offlinesupport`).

Neither list holds `country`, `url`, `browser`, `os` or `language` — and note `lang` is reserved while `language` is not, so even the near-miss doesn't help.

The note lives on `projectReservedValues`, since that's where the consequence actually lands: a bare map is returned precisely so the merge stays visible at the call site — spread the projection last and a reserved value shadows the survey's own field of that name, spread it first and the survey's wins. `RESERVED_FIELD_CATALOG` carries a pointer to it, so ENG-1839 hits the constraint at the moment it's picking names rather than at ENG-1840. If a colliding name gets chosen deliberately anyway, ENG-1837's merge order is the place that decides precedence on purpose.

No test pins it: the catalog is empty, and the merge is a call-site decision that doesn't exist yet — a test here would only be asserting object-spread semantics.

Same logistics note as the other thread: the commit (`dbae4bd`) is on `claude/pr-8806-cr-remarks-sknlm6`, not yet on this PR's head.

---

packages/types/embedded-data-resolver.ts
itsjavi: Correction to my logistics note above: @itsjavi had already landed this on the PR head as `bd123ee` ("derive legacy fields through the shared §8 mapping") — I was working from the pre-force-push head and didn't see it. The delegation there is the same implementation, keeping the narrower `{ variables; hiddenFields }` parameter rather than reusing `TLegacyEmbeddedFields`.

Verified on the current head: `packages/types` 203 passed, typecheck clean, with the seven `deriveLegacyEmbeddedData` tests unchanged. Nothing outstanding here — my parallel branch is superseded and won't be pushed.

---

packages/types/embedded-data-resolver.ts
itsjavi: Same correction as the other thread: this is already on the PR head (`bd123ee`) — I was reading the pre-force-push head. The note there is the two-line version on `projectReservedValues`, which covers the ask: an entry name matching an element id or hidden field name shadows it in the merged map, and `FORBIDDEN_IDS` guards none of the §9b candidates.

The trace above still stands as the backing detail, and the `lang` / `language` near-miss is the part worth remembering when ENG-1839 picks names.

---

PR #8809, opened 2026-08-11
fix: clean up transient Prisma migrations directory after each run

xernobyl:
## What & why

The database package had two near-identically-named directories — `migration/` (singular, checked-in source of truth with schema + data migrations) and `migrations/` (plural, gitignored scratch dir for Prisma) — that caused developer confusion. The similar names made it easy to read or edit the wrong one.

**Fix**: Rename the scratch directory to `.prisma-migrations/` (hidden). The dot-prefix makes it invisible in directory listings and clearly signals it is a generated, transient artifact. Developers should only work in `migration/`.

Document the dual-directory architecture in the README with a "Why two directories?" callout.

**Bonus**: Fix `.github/labeler.yml` which globbed the gitignored `migrations/**` instead of the checked-in `migration/**` — the label has never fired.

> **Future**: When [Prisma Next](https://www.prisma.io/blog/data-migrations-in-prisma-next) (or Prisma 8) ships with native data-migration support, we can collapse the dual-directory setup into a single canonical `migrations/` directory and remove the copy logic from the custom runner. This PR is the minimal fix to eliminate confusion now; the real consolidation is deferred until Prisma natively supports the pattern.

_Thanks to Anshuman for catching that the original try/finally approach broke `pnpm create-migration` — `prisma migrate dev --create-only` needs the scratch directory to mirror the applied history._

## Linear ticket

Fixes ENG-1683

## How this was tested

- `node --check` on source — ✅ passes
- `prettier --check` on all TS/MD files — ✅ passes
- `pnpm build --filter=@formbricks/database --force` — ✅ compiles, compiled dist uses new path
- E2E: built dist, ran 186 migrations against Docker Postgres — ✅ `create-migration` flow works (scratch directory stays on disk)
- Verified no stale `migrations` (plural) path references remain in any source file
- Verified all production features from #8627 intact: fresh-DB fast path, schema batching, atomic data-migration transactions

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] `ls packages/database/` → only `migration/` visible (`.prisma-migrations/` is hidden)
- [ ] `ls -a packages/database/` → `.prisma-migrations/` visible
- [ ] Run `pnpm db:migrate:dev` → migrations apply normally, scratch dir populated
- [ ] Run `pnpm create-migration` after → migration created successfully (no drift error)
- [ ] Open `packages/database/README.md` → "Why two directories?" callout visible

**Preconditions / test data**

- Running Postgres, configured DATABASE_URL

**Risks & regressions**

- None — the scratch directory behaviour is unchanged; only the name differs

- Developers with a stale `migrations/` dir locally may want to `rm -rf packages/database/migrations` (it's gitignored)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,5 @@
 "❗️ migrations":
-  - packages/database/migrations/**/migration.sql
+  - packages/database/migration/**/migration.sql
 
 "❗️ .env changes":
   - .env.example
--- a/.gitignore
+++ b/.gitignore
@@ -63,7 +63,7 @@ secrets/
 packages/lib/uploads
 apps/web/public/js
 packages/database/generated
-packages/database/migrations
+packages/database/.prisma-migrations
 branch.json
 .vercel
 
--- a/apps/web/Dockerfile
+++ b/apps/web/Dockerfile
@@ -117,7 +117,7 @@ COPY --from=installer /app/apps/web/public ./apps/web/public
 RUN chown -R nextjs:nextjs ./apps/web/public && chmod -R 755 ./apps/web/public
 
 # Create packages/database directory structure with proper ownership for runtime migrations
-RUN mkdir -p ./packages/database/migrations && chown -R nextjs:nextjs ./packages/database
+RUN mkdir -p ./packages/database/.prisma-migrations && chown -R nextjs:nextjs ./packages/database
 
 COPY --from=installer /app/packages/database/package.json ./packages/database/package.json
 RUN chown nextjs:nextjs ./packages/database/package.json && chmod 644 ./packages/database/package.json
--- a/packages/database/README.md
+++ b/packages/database/README.md
@@ -30,7 +30,7 @@ packages/database/
 │   │   └── migration.sql      # Schema migration file
 │   └── [timestamp_name]/      # Data migration folder
 │       └── migration.ts       # Data migration file
-├── migrations/                # Prisma internal migrations
+├── .prisma-migrations/        # Transient Prisma scratch dir (gitignored)
 ├── schema/                    # Prisma schema folder
 │   ├── main.prisma            # Shared models, datasource, and generators
 │   └── workflows.prisma       # Workflows models and enums
@@ -51,6 +51,8 @@ packages/database/
 - **Single Type per Subdirectory**: A migration subdirectory can only contain one file type—either `migration.sql` or `migration.ts`
 - **Custom Naming Convention**: Subdirectories follow the format `timestamp_name_of_the_migration` (e.g., `20241214112456_add_users_table`)
 - **Order of Execution**: Migrations are executed sequentially based on their timestamps, enabling precise control over the execution sequence
+
+> **Why two directories?** `migration/` (singular, checked in) is the source of truth — it contains both schema and data migrations interleaved by timestamp. `.prisma-migrations/` (hidden, gitignored) is a transient scratch directory generated at runtime: the runner copies only schema migrations into it and feeds them to `prisma migrate deploy`. The hidden name prevents accidental edits; developers should only work in `migration/`.
 
 ### Database Tracking
 
--- a/packages/database/prisma.config.ts
+++ b/packages/database/prisma.config.ts
@@ -11,7 +11,7 @@ config({ path: path.resolve(__dirname, "../../.env"), quiet: true });
 export default defineConfig({
   schema: "./schema",
   migrations: {
-    path: "./migrations",
+    path: "./.prisma-migrations",
   },
   datasource: {
     url: env("DATABASE_URL"),
--- a/packages/database/src/scripts/create-migration.ts
+++ b/packages/database/src/scripts/create-migration.ts
@@ -52,7 +52,7 @@ async function main(): Promise<void> {
   const migrationName = await promptForMigrationName();
   const migrationNameUnderscored = migrationName.replace(/\s+/g, "_");
 
-  const migrationsDir = path.resolve(__dirname, "../../migrations");
+  const migrationsDir = path.resolve(__dirname, "../../.prisma-migrations");
   const customMigrationsDir = path.resolve(__dirname, "../../migration");
 
   // Check if migration already exists in custom migrations
--- a/packages/database/src/scripts/migration-runner.ts
+++ b/packages/database/src/scripts/migration-runner.ts
@@ -45,7 +45,7 @@ const isBuilt = __filename.split(path.sep).includes("dist");
 const MIGRATIONS_DIR = isBuilt
   ? path.resolve(__dirname, "../migration") // From dist/scripts to dist/migration
   : path.resolve(__dirname, "../../migration"); // From src/scripts to migration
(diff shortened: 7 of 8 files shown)


xernobyl:
@coderabbitai can you review this?

pandeymangg (changes requested):
Reviewed this and ran it against a throwaway DB, because one thing worried me. It breaks `pnpm create-migration`.

## 🔴 Blocker: this breaks the "add a schema migration" workflow

`packages/database/migrations/` isn't only a hand-off buffer for the current `migrate deploy`. After a successful run it holds **the full mirror of applied schema history** — that's deliberate, and the runner says so at `migration-runner.ts:373-376`:

> Always copy every folder in the batch into the scratch migrations dir — even the already-applied ones. A later batch's `migrate deploy` runs against this accumulated directory, and Prisma treats a migration recorded in `_prisma_migrations` but missing from the directory as a divergence error. So the directory must always mirror the applied history.

`create-migration.ts:68-72` then runs `prisma migrate dev --create-only` against that directory. `migrate dev` replays the local migration folder into a shadow DB and compares it to the real one. Delete the folder and it sees 162 migrations recorded in `_prisma_migrations` with nothing on disk → drift → it refuses and asks to wipe the database.

Verified both sides on a throwaway DB seeded with the real 162 migrations.

**With the dir present (today's `main`)** — works:

```
Prisma Migrate created the following migration without applying it 20260811110149_probe_test
You can now edit it and apply it by running prisma migrate dev.
```

**With the dir deleted (this PR)** — fails:

```
- The following migration(s) are applied to the database but missing from the local
  migrations directory: 20230329205933_init, ... 20260725070000_add_two_factor_lockout_fields

We need to reset the "public" schema at "localhost:5432"

You may use prisma migrate reset to drop the development database.
All data will be lost.
```

So after this merges, the sequence every dev follows — `pnpm db:migrate:dev`, then `pnpm create-migration` — fails on step 2, every time, and the suggested remedy is dropping your dev database. CI is green because nothing in CI runs `create-migration`; it's a local-only workflow. The test plan in the description doesn't cover it either.

## 🟡 The fix doesn't quite solve the stated problem

Two more things, even setting the blocker aside:

- Both confusingly-named directories still exist. One is just usually absent. A dev who runs `ls` mid-migration, or reads `prisma.config.ts`, hits the same "which one is real?" question.
- "Usually absent" isn't "absent". `finally` doesn't run on SIGINT/SIGKILL, so Ctrl+C during a long migration — exactly when someone is most likely to go poking at the directory — leaves it behind anyway. QA item #5 in the description acknowledges this.

That makes the new README line ("If you see `migrations/` on disk, a migration is (or was recently) running") not a guarantee, which is a slightly misleading thing to teach.

## ✅ Suggested alternative: rename the scratch dir

Give it a name nobody can confuse with `migration/`, e.g. `packages/database/.prisma-migrations/`. That removes the confusion permanently, keeps the directory on disk so `create-migration` keeps working, and is about the same size as this diff:

| File | Change |
|---|---|
| `packages/database/src/scripts/migration-runner.ts:48` | `PRISMA_MIGRATIONS_DIR` |
| `packages/database/src/scripts/create-migration.ts:55` | `migrationsDir` |
| `packages/database/prisma.config.ts` | `migrations.path` |
| `prisma.config.mjs:27` | `path` |
| `.gitignore:58` | ignore path |
| `apps/web/Dockerfile:120` | `mkdir -p` (keep the `chown` — the runtime user needs write access) |
| `.github/labeler.yml:2` | see below |
| `packages/database/README.md` | tree + callout |

The "Why two directories?" callout is genuinely useful — worth keeping either way.

## 🟢 Bonus bug, in scope for ENG-1682

`.github/labeler.yml:2` globs `packages/database/migrations/**/migration.sql` — the **gitignored** directory. No file there is ever part of a diff, so the `❗️ migrations` label has never fired (git has no history under that path at all). It should point at `packages/database/migration/**`. Same root cause as this ticket, so it fits naturally here.

## Minor

- Deleting the scratch dir in `finally` also deletes it on failure, so you lose the exact SQL set that was fed to `migrate deploy` when triaging a failed production migration. Low impact since `migration/` is the source of truth, but it's a small step backwards.
- `migration-runner.ts:276` — `fs.rm(..., { force: true })` doesn't throw on a missing path, so the surrounding `catch` is close to unreachable. Fine as belt-and-braces, but worth a word on why it's there.
- Nothing pins the new control flow (the early `return` in the fresh-DB fast path now passing through `finally`). The runner has no tests today, so not a blocker.

---

The `create-migration` breakage needs fixing before this can go in. My suggestion is to drop the `try/finally` entirely and do the rename instead — it actually ends the confusion and doesn't change the runner's behaviour at all.

xernobyl (commented):
Thanks for the thorough review! All points addressed:

- **🔴 create-migration breakage** — reverted the try/finally approach entirely. The scratch directory now stays on disk, so `prisma migrate dev --create-only` will always find the applied-history mirror.
- **🟡 naming confusion** — renamed to `.prisma-migrations/` (hidden). `ls` won't show it; the name clearly signals it's generated.
- **✅ rename approach** — implemented as suggested: 8 files, `migrations/` → `.prisma-migrations/` everywhere.
- **🟢 labeler.yml bonus bug** — fixed the glob to `migration/**/migration.sql` (checked-in dir).
- **Minor points** — all moot since the try/finally is gone (no cleanup on failure, no unreachable catch, no new control flow).
- **README** — "Why two directories?" callout updated for the hidden name.

@coderabbitai full review

xernobyl (commented):
Re-requesting review after addressing all feedback:

- ✅ Switched from try/finally cleanup to the rename approach you suggested
- ✅ migrations/ → .prisma-migrations/ across all 8 files
- ✅ Reverted try/finally entirely — no behavior changes to the runner
- ✅ labeler.yml glob fixed (bonus bug)
- ✅ README updated
- ✅ Smoke tested: dist rebuilt, 186 migrations applied, .prisma-migrations/ stays on disk with 162 schema entries, create-migration flow unbroken
- ✅ CI is green

@coderabbitai full review

pandeymangg (approved):
Re-reviewed — blocker is fixed. The `try/finally` is gone and the runner's original behaviour is back, so `create-migration` works again; I separately confirmed a dot-prefixed dir handles both `migrate deploy` (162 migrations) and `migrate dev --create-only`.

Two small things, both non-blocking:

**`.dockerignore:56`** still points at `packages/database/migrations`. With `COPY . .` at `Dockerfile:54`, a local docker build on a machine that has run migrations now pulls the stale folders into the build context — CI (fresh checkout) and the runtime image (own `mkdir -p`) are unaffected.

**`.gitignore`** replaced the old entry instead of adding to it, so anyone with a leftover `packages/database/migrations/` will now see it as untracked and `git add -A` will stage it. Keeping both lines, in both files, covers it:

```
packages/database/.prisma-migrations
packages/database/migrations   # legacy scratch dir name — keeps stale local copies ignored
```

**Nit:** the PR title still describes the dropped try/finally approach — the commit message has it right.

Last thing: deferring ENG-1683's "one canonical migration location" criterion is the correct call — Prisma 7 still hard-errors `P3015` on the mixed dir and offers no ignore/exclude option. Worth flagging to Matti so the ticket isn't closed as fully delivered.

PR #8811, opened 2026-08-11
fix(workflows): stop read-only members getting an interactive editor

Dhruwang:
## What & why

A read-only workspace member opened the workflow editor and, despite the "Read-only" pill, got a broadly interactive editor. Three defects, one theme — the UI advertised edits the server can never accept and explained them wrongly:

- **Lifecycle dropdown opened despite being disabled.** `disabled` was passed only to the child `Button`, not to `DropdownMenuTrigger`; Radix's open guard reads the trigger's own prop, so the DOM attribute and the JS guard disagreed and the menu opened. Enable / Unarchive / Delete then fired real mutations (all 403s).
- **Canvas add/delete/drag were live.** `workflow-canvas-node.tsx` reads `canMutateCanvasAtom` directly, and that atom checked status only — no permission. On a draft/disabled workflow the status check passed, so structural edits went live against local state that autosave never persists (silently lost on reload).
- **Wrong blocked-edit copy.** The inspector always showed "Disable the workflow to make changes here.", even for read-only members and archived workflows where that instruction is wrong.

Fix:

- Fold `isReadOnly` into the editor state and `canEditWorkflowDefinitionAtom`, seeded at hydrate time, so the node card gates on permissions, not just status — one source of truth instead of two.
- Pass `disabled` to `DropdownMenuTrigger` **and** withhold the menu content from read-only members (the list page hides its actions the same way).
- Guard `transition()` in the hook so lifecycle ops never fire for a read-only member, independent of UI gating.
- Replace the single `edit_blocked_active` string with reason-aware copy: no-permission / archived / enabled (permission wins).

No security or data impact: server-side authz already rejects every lifecycle call with 403 and autosave never fired for read-only users. This is a presentation/affordance fix.

## Linear ticket

Fixes ENG-2176

## How this was tested

- `pnpm vitest run` on the two affected suites — **102 passed**:
  - `state/editor.test.ts` (64) — added cases: `canEditWorkflowDefinitionAtom` and `canMutateCanvasAtom` are `false` for a read-only member on an editable/unlocked draft.
  - `hooks/use-workflow-builder.test.ts` (38) — added: a read-only member fires none of enable/disable/archive/unarchive.
- `eslint` + `prettier` clean on all changed files.
- Typecheck: no errors in the changed files (the worktree's unbuilt-package module-resolution noise is unrelated).

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Preconditions below. As a **read-only** member, open `/workspaces/<workspaceId>/workflows/<workflowId>` on a **draft** workflow.
- [ ] Header shows the "Read-only" pill; click the status dropdown trigger → **it does not open** (no menu, no items).
- [ ] On the canvas: the trigger card shows **no `+`** affordance; nodes have **no delete/drag** and no context menu.
- [ ] Select a node → inspector alert reads **"You don't have permission to edit this workflow."** (not "Disable the workflow…").
- [ ] Repeat on an **archived** workflow as an editor (owner/manager) → inspector alert reads **"Unarchive the workflow to make changes here."**
- [ ] Repeat on an **enabled** workflow as an editor → inspector alert still reads **"Disable the workflow to make changes here."**
- [ ] Regression (editor role): owner/manager on a draft can still open the lifecycle menu, add/delete/drag nodes, and Enable/Disable/Archive/Unarchive as before.

**Preconditions / test data**

- EE-entitled org (Workflows is Enterprise-gated). A member with **read-only** workspace access, plus an owner/manager to verify the editable path. Workflows in draft, enabled, and archived status.

**Risks & regressions**

- Lifecycle menu gating and canvas node affordances for editor roles — confirm nothing was over-gated. Inspector field editing for editors unchanged (field-level gating already lived in `handleChange`).

- none

### DIFF
--- a/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
+++ b/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
@@ -5,8 +5,10 @@ import { useTranslation } from "react-i18next";
 import type { TWorkflowDefinition, TWorkflowNode } from "@formbricks/workflows";
 import { getNodeRegistryEntry } from "@/modules/ee/workflows/lib/node-registry";
 import {
+  isWorkflowReadOnlyAtom,
   selectedWorkflowNodeIdAtom,
   setWorkflowDefinitionAtom,
+  workflowAtom,
   workflowDefinitionAtom,
 } from "@/modules/ee/workflows/state/editor";
 import { Alert, AlertDescription } from "@/modules/ui/components/alert";
@@ -50,10 +52,22 @@ export const WorkflowNodeConfigPanel = ({ isEditable }: Readonly<WorkflowNodeCon
   const { t } = useTranslation();
   const definition = useAtomValue(workflowDefinitionAtom);
   const selectedNodeId = useAtomValue(selectedWorkflowNodeIdAtom);
+  const workflow = useAtomValue(workflowAtom);
+  const isReadOnly = useAtomValue(isWorkflowReadOnlyAtom);
   const setDefinition = useSetAtom(setWorkflowDefinitionAtom);
 
   const selectedNode = findSelectedNode(definition, selectedNodeId);
   if (!selectedNode || !definition) return null;
+
+  // `isEditable` collapses three distinct block reasons (no permission, archived, enabled) into one
+  // boolean; name the actual reason instead of always blaming an "active" workflow. Permission wins
+  // — a read-only member sees the permission message even on an enabled/archived workflow.
+  const blockedReason = (() => {
+    if (isEditable) return null;
+    if (isReadOnly) return t("workspace.workflows.edit_blocked_read_only");
+    if (workflow?.status === "archived") return t("workspace.workflows.edit_blocked_archived");
+    return t("workspace.workflows.edit_blocked_active");
+  })();
 
   const registryEntry = getNodeRegistryEntry(selectedNode);
   const ConfigForm = registryEntry.ConfigForm;
@@ -77,9 +91,9 @@ export const WorkflowNodeConfigPanel = ({ isEditable }: Readonly<WorkflowNodeCon
       </header>
       <div className="scroll-bar min-h-0 flex-1 overflow-y-auto">
         <div className="flex flex-col gap-3 px-3 pt-3 pb-4">
-          {!isEditable && (
+          {blockedReason && (
             <Alert variant="info" size="small">
-              <AlertDescription>{t("workspace.workflows.edit_blocked_active")}</AlertDescription>
+              <AlertDescription>{blockedReason}</AlertDescription>
             </Alert>
           )}
           {ConfigForm ? (
--- a/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
+++ b/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
@@ -2,14 +2,14 @@
 
 import { useAtomValue, useSetAtom } from "jotai";
 import { useTranslation } from "react-i18next";
-import type { TWorkflowDefinition, TWorkflowNode } from "@formbricks/workflows";
+import type { TWorkflowDefinition, TWorkflowNode, TWorkflowResource } from "@formbricks/workflows";
 import { getNodeRegistryEntry } from "@/modules/ee/workflows/lib/node-registry";
 import {
   isWorkflowReadOnlyAtom,
   selectedWorkflowNodeIdAtom,
   setWorkflowDefinitionAtom,
-  workflowAtom,
   workflowDefinitionAtom,
+  workflowStatusAtom,
 } from "@/modules/ee/workflows/state/editor";
 import { Alert, AlertDescription } from "@/modules/ui/components/alert";
 
@@ -24,6 +24,21 @@ const findSelectedNode = (
   if (!definition || !selectedNodeId) return null;
   if (definition.trigger?.id === selectedNodeId) return definition.trigger;
   return definition.nodes.find((node) => node.id === selectedNodeId) ?? null;
+};
+
+// `isEditable` collapses three distinct block reasons (no permission, archived, enabled) into one
+// boolean; name the actual reason instead of always blaming an "active" workflow. Precedence:
+// permission wins — a read-only member sees the permission message even on an enabled/archived
+// workflow — then archived, then active/enabled. Returns the translation key; the caller runs t().
+const getBlockedReasonKey = (
+  isEditable: boolean,
+  isReadOnly: boolean,
+  status: TWorkflowResource["status"] | undefined | null
+): string | null => {
+  if (isEditable) return null;
+  if (isReadOnly) return "workspace.workflows.edit_blocked_read_only";
+  if (status === "archived") return "workspace.workflows.edit_blocked_archived";
+  return "workspace.workflows.edit_blocked_active";
 };
 
 const replaceNode = (definition: TWorkflowDefinition, node: TWorkflowNode): TWorkflowDefinition => {
@@ -52,22 +67,18 @@ export const WorkflowNodeConfigPanel = ({ isEditable }: Readonly<WorkflowNodeCon
   const { t } = useTranslation();
   const definition = useAtomValue(workflowDefinitionAtom);
   const selectedNodeId = useAtomValue(selectedWorkflowNodeIdAtom);
-  const workflow = useAtomValue(workflowAtom);
+  // Read only the status primitive, not the whole workflow: autosave replaces the workflow atom
+  // with the API response every ~2s, and subscribing to it here would re-render this panel (and the
+  // rich-text email editor it hosts) on every save. Jotai bails out when the primitive is unchanged.
+  const status = useAtomValue(workflowStatusAtom);
   const isReadOnly = useAtomValue(isWorkflowReadOnlyAtom);
   const setDefinition = useSetAtom(setWorkflowDefinitionAtom);
 
   const selectedNode = findSelectedNode(definition, selectedNodeId);
   if (!selectedNode || !definition) return null;
 
-  // `isEditable` collapses three distinct block reasons (no permission, archived, enabled) into one
-  // boolean; name the actual reason instead of always blaming an "active" workflow. Permission wins
-  // — a read-only member sees the permission message even on an enabled/archived workflow.
-  const blockedReason = (() => {
-    if (isEditable) return null;
-    if (isReadOnly) return t("workspace.workflows.edit_blocked_read_only");
-    if (workflow?.status === "archived") return t("workspace.workflows.edit_blocked_archived");
-    return t("workspace.workflows.edit_blocked_active");
-  })();
+  const blockedReasonKey = getBlockedReasonKey(isEditable, isReadOnly, status);
+  const blockedReason = blockedReasonKey ? t(blockedReasonKey) : null;
 
   const registryEntry = getNodeRegistryEntry(selectedNode);
   const ConfigForm = registryEntry.ConfigForm;
--- a/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
+++ b/apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
@@ -29,16 +29,19 @@ const findSelectedNode = (
 // `isEditable` collapses three distinct block reasons (no permission, archived, enabled) into one
 // boolean; name the actual reason instead of always blaming an "active" workflow. Precedence:
 // permission wins — a read-only member sees the permission message even on an enabled/archived
-// workflow — then archived, then active/enabled. Returns the translation key; the caller runs t().
-const getBlockedReasonKey = (
+// workflow — then archived, then active/enabled. Returns a reason discriminant, not the i18n key,
+// so the caller resolves it with literal `t("…")` calls the translation scanner can detect.
+type TBlockedReason = "readOnly" | "archived" | "active";
+
+const getBlockedReason = (
   isEditable: boolean,
   isReadOnly: boolean,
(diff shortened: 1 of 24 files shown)


apps/web/modules/ee/workflows/state/editor.ts
```diff
@@ -475,9 +489,11 @@ export const hydrateWorkflowEditorAtom = atom(
     {
       workflow,
       flowNodes,
+      isReadOnly = false,
```
itsjavi, line 492: A permission flag with an optional `= false` default fails in the permissive direction: any future `hydrateWorkflowEditorAtom` call site that forgets it silently gets the full set of edit affordances back, and nothing — type error, test, lint — says so. Today there is exactly one production caller (`use-workflow-builder.ts`), so this is latent rather than broken.

Consider making it required (`isReadOnly: boolean`) so a new hydrate site has to state the answer. The cost is ~20 `isReadOnly: false` additions in `editor.test.ts` / `use-workflow-node-url-sync.test.ts`, which is churn — your call whether that trade is worth it, but the default is worth being deliberate about either way.

---

apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
```diff
@@ -50,11 +52,23 @@ export const WorkflowNodeConfigPanel = ({ isEditable }: Readonly<WorkflowNodeCon
   const { t } = useTranslation();
   const definition = useAtomValue(workflowDefinitionAtom);
   const selectedNodeId = useAtomValue(selectedWorkflowNodeIdAtom);
+  const workflow = useAtomValue(workflowAtom);
```
itsjavi, line 55: Subscribing to the whole `workflowAtom` to read one field re-renders this panel — and with it the whole `ConfigForm` subtree, including the rich-text email editor — every time the server snapshot is replaced. `save()` calls `setWorkflowAtom` with the API response on each autosave (2s debounce), so that's a new object identity and a new panel render after every save, where previously the panel only tracked `definition` + `selectedNodeId`.

`state/editor.ts` already follows a one-field-per-derived-atom convention (`workflowNameAtom`, `workflowLastSavedAtAtom`, …). A `workflowStatusAtom` fits that grain and lets Jotai bail out on an unchanged primitive:

```ts
export const workflowStatusAtom = atom((get) => get(workflowEditorAtom).workflow?.status ?? null);
```

Also, minor: two sources of read-only truth now meet in this file — `isEditable` comes from the hook's prop-derived `canEditDefinition`, while the message reads the hydrate-seeded `isWorkflowReadOnlyAtom`. They agree today (one hydrate site, same value), so this is not a bug; but the PR body's "one source of truth instead of two" would be literally true if the hook derived `canEditDefinition` from `canEditWorkflowDefinitionAtom` instead of recomputing the same rule from the prop.

---

apps/web/modules/ee/workflows/components/inspector/workflow-node-config-panel.tsx
```diff
@@ -50,11 +52,23 @@ export const WorkflowNodeConfigPanel = ({ isEditable }: Readonly<WorkflowNodeCon
...
 
+  // `isEditable` collapses three distinct block reasons (no permission, archived, enabled) into one
+  // boolean; name the actual reason instead of always blaming an "active" workflow. Permission wins
+  // — a read-only member sees the permission message even on an enabled/archived workflow.
+  const blockedReason = (() => {
+    if (isEditable) return null;
+    if (isReadOnly) return t("workspace.workflows.edit_blocked_read_only");
+    if (workflow?.status === "archived") return t("workspace.workflows.edit_blocked_archived");
+    return t("workspace.workflows.edit_blocked_active");
+  })();
```
itsjavi, line 70: Nit, style only: an IIFE inside the component body reads oddly next to this file's existing pattern of module-level pure helpers (`findSelectedNode`, `replaceNode`). Same logic as a plain function keeps the component body flat and makes the precedence rule (permission → archived → enabled) testable on its own:

```ts
const getBlockedReasonKey = (
  isEditable: boolean,
  isReadOnly: boolean,
  status: TWorkflowResource["status"] | undefined
): string | null => {
  if (isEditable) return null;
  if (isReadOnly) return "workspace.workflows.edit_blocked_read_only";
  if (status === "archived") return "workspace.workflows.edit_blocked_archived";
  return "workspace.workflows.edit_blocked_active";
};
```

Returning the key rather than the translated string keeps it a pure function and leaves `t()` at the call site.

---

apps/web/modules/ee/workflows/components/workflow-header-cta.tsx
```diff
@@ -99,53 +99,60 @@ export const WorkflowHeaderCta = ({ workflowId, isReadOnly }: Readonly<WorkflowH
       {isEditTab && (
         <>
           <DropdownMenu>
-            <DropdownMenuTrigger asChild>
+            {/* `disabled` must sit on the trigger, not only the child Button: Radix's open guard
+                reads the trigger's own prop, so a value passed solely to the Button leaves the menu
+                openable (the DOM attribute and the JS guard disagree). Belt-and-suspenders, the
+                content is also withheld from read-only members, so a bypassed trigger has nothing
+                to select — the list page hides its actions the same way. */}
+            <DropdownMenuTrigger asChild disabled={isReadOnly || isBusy}>
```
itsjavi, line 107: The fix is right, and the comment above it is accurate — I checked it against the installed Radix source and a real browser rather than assuming (details in the review summary).

One consequence worth a deliberate QA pass, though: `isBusy` is now genuinely effective here, not just decorative. Before this change the pointer-down leak meant the menu opened even while the button was rendered `disabled`; now `isBusy` (`isTransitioning || isSaving || isDeleting`) truly blocks opening it. For an **editor**, that adds a short window — one autosave round-trip after the 2s debounce — where clicking the status control does nothing at all, and `loading` is only set for `isTransitioning`, so during a *save* the button just looks dim with no spinner to explain it.

Probably fine (and arguably what the `disabled` prop always intended), but it's the one behavior change this PR makes for users who *do* have permission, so it deserves a look during QA: edit a field, then immediately click the status dropdown. If it feels like a dead control, gating the trigger on `isReadOnly || isTransitioning` (dropping `isSaving`) would keep the menu reachable while still serializing the actual writes — `transition()` already refuses to run when a save is in flight.

---

itsjavi (commented):
**1 gap in the canvas fix + 3 non-blocking notes** — reviewed at `e2e4590d`.

The three defects are real and the diagnosis is precise. I verified the load-bearing claim rather than trusting it, because it looked wrong at first: a natively-disabled `<button>` should swallow clicks, so "the menu opened anyway" seemed like a misdiagnosis. It isn't.

- `@radix-ui/react-dropdown-menu` (installed build) opens on **`onPointerDown`** and guards on the trigger's *own* prop: `if (!disabled && event.button === 0 && event.ctrlKey === false) context.onOpenToggle()`. With `asChild`, Radix's Slot gives the child's props precedence, so `disabled` reached the DOM but never the guard.
- In headless Chromium, a disabled button still **receives `pointerdown`** — I probed it directly: a real mouse click on `<button disabled>` fires `pointerdown` and nothing else (`mousedown`/`click` are suppressed, and Tab skips it, so keyboard was never a path). That is exactly the one event Radix listens to.

So the pre-fix trigger really was openable for a read-only member, and moving `disabled` onto `DropdownMenuTrigger` is a substantive fix, not belt-tightening. The added code comment is accurate.

Also ran the suites myself instead of trusting the numbers (had to build `@formbricks/workflows` first — `dist/` isn't checked in): `editor.test.ts` + `use-workflow-builder.test.ts` → **102 passed**, matching the PR body exactly. Then mutation-checked the three new assertions by reverting each guard: dropping `if (state.isReadOnly) return false;` fails both new atom tests, dropping the hook's `if (isReadOnly) return;` fails the lifecycle test. Exactly those three, no others — real tests, not can't-fail ones. ESLint clean on all four changed source files.

---

### The canvas is still deletable by keyboard

The dropdown path is closed, but pressing **Backspace on a selected node** still removes it, for a read-only member and for anyone on an enabled/archived workflow. `workflow-canvas.tsx` isn't in this diff, so I couldn't anchor an inline comment:

```ts
const handleNodesChange: OnNodesChange<Node<TWorkflowNodeData>> = useCallback(
  (changes) => setFlowNodes((currentNodes) => applyNodeChanges(changes, currentNodes)),
  [setFlowNodes]   // ← no canMutate guard, unlike handleNodeDragStop / handleAutoLayout
);
```

The chain, traced through the installed `@xyflow/react@12.9.3` rather than assumed:

1. `deleteKeyCode` defaults to `'Backspace'` (`deleteKeyCode = 'Backspace'` in the dist bundle) — the canvas doesn't override it.
2. `useGlobalKeyHandler` calls `deleteElements({ nodes: nodes.filter(selected), … })` on that keypress.
3. `getElementsToRemove` excludes a node only when `node.deletable === false`; `workflowDefinitionToFlowNodes` sets no such flag, so every node is deletable.
4. Nodes are selectable and focusable for a read-only member — `elementsSelectable={!isLocked}` / `nodesFocusable={!isLocked}`, and `isLocked` starts `false`. Clicking a node selects it (and opens the inspector, correctly showing the new read-only alert).
5. `deleteElements` emits `remove` changes to `onNodesChange` → `applyNodeChanges` drops the node from `flowNodes`.

The definition atom is untouched, so nothing is persisted and no request fires — which makes this precisely the failure mode the PR body describes for the other paths: *"structural edits went live against local state that autosave never persists (silently lost on reload)"*. The re-sync effect is keyed on `derivedFlowNodes`, so the node stays visually gone until the definition changes or the page reloads. `flowEdges` is projected straight from the definition with no `onEdgesChange`, so the edge into the deleted node survives — the canvas is left visibly inconsistent, not merely missing a card.

This predates the PR (it also hits editors on enabled/archived workflows today), but it's inside the stated scope — *"Canvas add/delete/drag were live"* — and the QA plan asserts "nodes have **no delete**/drag", which a tester following it by mouse alone will confirm while the keyboard path still works. One line either way:

```tsx
// simplest: don't bind the key at all when mutations are gated (null disables it in RF)
deleteKeyCode={canMutate ? "Backspace" : null}
```

```tsx
// or drop just the structural changes, keeping selection/dimension changes flowing
(changes) =>
  setFlowNodes((currentNodes) =>
    applyNodeChanges(canMutate ? changes : changes.filter((c) => c.type !== "remove"), currentNodes)
  )
```

I'd take the first — it also stops the delete from being *attempted*, so React Flow never emits changes that have to be filtered downstream. Worth a unit test on the filtering variant if you pick the second.

To be clear about my confidence: this is verified by reading the library source and the call site, not by clicking in a running editor with an EE license and a read-only member — one keystroke on the canvas will confirm or refute it.

### Follow-up worth considering (not this PR)

The header pill has the same conflation this PR removes from the inspector: it renders **"Read-only"** whenever `!builder.canEditDefinition`, so an owner on an *enabled* workflow is told they lack permission while the inspector — now reason-aware — correctly tells them to disable the workflow. Two different explanations of the same state, three lines apart. Naming the reason in the pill too (or softening it to something status-neutral) would finish the theme; fine as a separate change.

### Checks

All 8 CI runs green at `e2e4590d` (Check PR, SonarQube, Docker Build Validation, Integration Tests, Translation Validation, PR Size Check, PR Update, PR Label Sync). Sonar quality gate passed: 0 new issues, 0 security hotspots, 100% coverage and 0% duplication on new code. No conflicts with `main` (`mergeable_state: blocked` is the required-review gate, not a merge conflict). Nothing outstanding to address — with one caveat: **CodeRabbit skipped this PR entirely** ("Review skipped due to path filters" — all 22 files, including every `modules/ee/workflows/**` source file, matched none of its include patterns), so there is no bot review coverage on this diff beyond Sonar. Might be worth checking whether `.coderabbit.yaml` is meant to cover `apps/web/modules/**`.

**Checked** — the Radix trigger fix (verified against library source + a real-browser pointer-event probe), the permission gating chain end to end (`getWorkflowsRouteAuth` → layout → hook → hydrate → atoms → node card; `hydrateWorkflowEditorAtom` has exactly one production caller, so the seeded flag can't silently diverge), canvas mutation paths (drag, auto-layout, trigger picker and node-card affordances are all gated; keyboard delete is not — above), the reason-aware copy precedence, tests (ran them; mutation-checked the new ones), i18n mechanics (`en-US.json` edited with generated locales and `i18n.lock` regenerated, per repo policy — key naming follows the existing `edit_blocked_*` convention), lint, CI, Sonar, merge state.

### Needs human review / input @itsjavi

- **The QA checklist itself.** Every step needs an EE-entitled org, a read-only member and an owner, and workflows in draft/enabled/archived — I can't provision that here, and all of it is browser interaction. Please add "select a node, press Backspace" as a step regardless of how the finding above is resolved, since that's the one path the mouse-driven checks miss.
- **No E2E coverage, and no obvious home for it.** `apps/web/playwright/` has no workflows spec at all and no read-only-member fixture, so per AGENTS.md (React components are covered by Playwright, not component unit tests) the *primary* fix — the Radix trigger — ships guarded only by a code comment. Adding the fixture is its own piece of work; worth a ticket rather than blocking this.
- **Two generated translations read wrong**, and I can't hand-fix them (repo policy: only `en-US.json` is editable, the rest come from Lingo). `hu-HU`'s `edit_blocked_archived` — "Kérem, archiválja vissza a munkafolyamatot" — reads as *re-archive it*, the opposite of the instruction; `es-ES` uses the infinitive "Desarchivar" where the neighbouring strings use the imperative ("Desactiva"). Both need a Lingo glossary entry or a re-run, not a manual edit.

Not approving — the keyboard-delete gap is worth a decision first, and the rest of the verification I can't do from here.

<sub>Reviewed with Claude Code — model `claude-opus-5`, reasoning effort: max.</sub>

---

itsjavi (changes requested):
thanks, please check open CR remarks

itsjavi (approved):
thanks for the fix LGTM

PR #8813, opened 2026-08-11
fix(security): restrict the onboarding invite path to organization owners [ENG-2169]

pandeymangg:
## What & why

The onboarding invite screen (`/setup/organization/{organizationId}/invite`) creates **owner** invites — `inviteUser` hardcodes `role: "owner"` and `inviteOrganizationMemberAction` has no role input — but it authorized `["owner", "manager"]`. Nothing ties the route to the moments right after organization creation, so a manager could navigate straight to it and mint an owner, going around the `Managers can only invite users as members` rule enforced in `modules/organization/settings/teams/actions.ts`. Acceptance trusts the stored role, so the invitee lands as a real owner membership — able to manage billing, remove other owners and delete the organization.

Exposure is organizations where managers exist at all, i.e. those with the access-control license; without it every invite is an owner by design.

- `invite/lib/authorization.ts` (new) — `SETUP_INVITE_ROLES` is the single source of truth for who may use this path (`["owner"]`), with `checkSetupInviteAuthorization` for the action and `hasSetupInviteAccess` for the page gate, so the two cannot drift.
- `invite/actions.ts` — authorization narrowed to owners via `checkSetupInviteAuthorization`. This is the fix; the action is the boundary. It still goes through `checkAuthorizationUpdated`, so the authorization pattern and the `AuthorizationError` contract are unchanged.
- `invite/page.tsx` — gate swapped from `verifyUserRoleAccess().hasCreateOrUpdateMembersAccess` (true for managers) to `hasSetupInviteAccess`, so a manager gets a 404 instead of a form that fails on submit.
- `invite/lib/invite.ts` — the hardcoded owner role stays (it matches the default for organizations with no access-control license) with a comment on why it is only safe behind an owner-gated caller.
- `lib/organization/auth.ts` — `verifyUserRoleAccess` deleted along with its test block. The invite page was its last production call site, and only `hasCreateOrUpdateMembersAccess` was ever read from it. `canUserAccessOrganization` stays; the organization layout still uses it.

Nothing legitimate is lost: the only entry to this screen is the redirect right after `createOrganizationAction`, which makes the creator an owner.

Reported through responsible disclosure by [meifukun](https://github.com/meifukun).

## Linear ticket

Fixes ENG-2169

## How this was tested

- **New unit test** `invite/lib/authorization.test.ts` — 10 tests over the role decision: an owner is allowed, a manager, member, billing user and a user with no membership are all rejected, for both the action helper and the page helper. Only the membership lookup is mocked, so the real `checkAuthorizationUpdated` role matching executes and the assertions are about outcomes per role, not about the arguments a mock received. ✅
- **New Playwright spec** `apps/web/playwright/setup-invite-role-access.spec.ts` — a manager navigating to the invite URL gets a 404; the owner still gets the onboarding form (guards against a fix that over-blocks). Both pass ✅
- **Bug-catch checks** — each guard was verified to actually fail on the unfixed code:
  - Reverted only the page gate and re-ran the Playwright manager test → fails (the manager sees "Invite your Organization members").
  - Added `"manager"` back to `SETUP_INVITE_ROLES` and re-ran the unit test → `rejects a manager` and `is false for a manager` both fail.
  - Both restored; green again.
- `pnpm vitest run` across `lib/organization` and the touched module — 11 files / 138 tests ✅ (after deleting `verifyUserRoleAccess`, confirmed no references remain anywhere)
- `pnpm typecheck` (apps/web) ✅, Prettier ✅
- No unit test for the action file itself — `**/actions.ts` is coverage-excluded, which is why the role decision lives in `lib/authorization.ts` where it can be tested directly.

No UI change: the screen is unchanged for the owner who legitimately reaches it.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] As an organization **owner**, sign up and create a new organization → you land on "Invite your Organization members" and can send invites as before.
- [ ] As an owner, copy that URL (`/setup/organization/{organizationId}/invite`) and open it again later → the invite form still loads.
- [ ] Invite a second user, set their organization role to **Manager** in Settings → Teams (needs the access-control entitlement), and log in as them.
- [ ] As the manager, open `/setup/organization/{organizationId}/invite` for that same organization → **404 Not Found**, no invite form.
- [ ] As the manager, go to Settings → Teams → Invite → the normal invite flow still works and still offers **Member** only.
- [ ] Check the invite created by the manager in Settings → Teams shows role **Member**, and any invite created from the onboarding screen by an owner shows **Owner**.
- [ ] As a **member** or **billing** user, open the invite URL → 404 (unchanged behavior).

**Preconditions / test data**

- Two accounts in the same organization: one owner, one manager.
- Promoting someone to manager needs the access-control (role management) entitlement — on a plan/licence without it, everyone is an owner and no manager exists to test with.
- SMTP configured if you want the invite emails to actually send; not required for the authorization checks.

**Risks & regressions**

- Onboarding right after organization creation — the owner must still reach the invite step; a too-strict gate would 404 the happy path.
- Organization Settings → Teams invites for owners, managers and team admins are untouched and must keep working, including the manager → member restriction.
- Accepting an existing invite (both sign-up and already-logged-in paths) is untouched; roles on already-issued invites are unchanged.

- none

### DIFF
--- a/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
+++ b/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
@@ -2,11 +2,10 @@ import { Metadata } from "next";
 import { notFound } from "next/navigation";
 import { AuthenticationError } from "@formbricks/types/errors";
 import { SMTP_HOST, SMTP_PASSWORD, SMTP_PORT, SMTP_USER } from "@/lib/constants";
-import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
-import { getAccessFlags } from "@/lib/membership/utils";
 import { getTranslate } from "@/lingodotdev/server";
 import { getSession } from "@/modules/auth/lib/session";
 import { InviteMembers } from "@/modules/setup/organization/[organizationId]/invite/components/invite-members";
+import { hasSetupInviteAccess } from "@/modules/setup/organization/[organizationId]/invite/lib/authorization";
 
 export const metadata: Metadata = {
   title: "Invite",
@@ -28,14 +27,10 @@ export const InvitePage = async (props: InvitePageProps) => {
   const session = await getSession();
   if (!session) throw new AuthenticationError(t("common.session_not_found"));
 
-  // Owner-only, matching `inviteOrganizationMemberAction`: this screen creates owner invites, so a
-  // manager must not reach it (the previous `hasCreateOrUpdateMembersAccess` flag is true for
-  // managers too). Not the security boundary — the action is — but keep the two in sync so managers
-  // get a 404 instead of a form that fails on submit.
-  const membership = await getMembershipByUserIdOrganizationId(session.user.id, params.organizationId);
-  const { isOwner } = getAccessFlags(membership?.role);
-
-  if (!isOwner) return notFound();
+  // Not the security boundary — `inviteOrganizationMemberAction` is — but this shares the action's
+  // role list so a manager gets a 404 instead of a form that fails on submit. The previous
+  // `verifyUserRoleAccess().hasCreateOrUpdateMembersAccess` flag was true for managers too.
+  if (!(await hasSetupInviteAccess(session.user.id, params.organizationId))) return notFound();
 
   return <InviteMembers IS_SMTP_CONFIGURED={IS_SMTP_CONFIGURED} organizationId={params.organizationId} />;
 };
--- a/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
+++ b/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
@@ -28,8 +28,7 @@ export const InvitePage = async (props: InvitePageProps) => {
   if (!session) throw new AuthenticationError(t("common.session_not_found"));
 
   // Not the security boundary — `inviteOrganizationMemberAction` is — but this shares the action's
-  // role list so a manager gets a 404 instead of a form that fails on submit. The previous
-  // `verifyUserRoleAccess().hasCreateOrUpdateMembersAccess` flag was true for managers too.
+  // role list so a manager gets a 404 instead of a form that fails on submit.
   if (!(await hasSetupInviteAccess(session.user.id, params.organizationId))) return notFound();
 
   return <InviteMembers IS_SMTP_CONFIGURED={IS_SMTP_CONFIGURED} organizationId={params.organizationId} />;
--- a/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
+++ b/apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
@@ -2,7 +2,8 @@ import { Metadata } from "next";
 import { notFound } from "next/navigation";
 import { AuthenticationError } from "@formbricks/types/errors";
 import { SMTP_HOST, SMTP_PASSWORD, SMTP_PORT, SMTP_USER } from "@/lib/constants";
-import { verifyUserRoleAccess } from "@/lib/organization/auth";
+import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
+import { getAccessFlags } from "@/lib/membership/utils";
 import { getTranslate } from "@/lingodotdev/server";
 import { getSession } from "@/modules/auth/lib/session";
 import { InviteMembers } from "@/modules/setup/organization/[organizationId]/invite/components/invite-members";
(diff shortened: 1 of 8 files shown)


apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
```diff
@@ -27,12 +27,10 @@ export const InvitePage = async (props: InvitePageProps) => {
...
 
-  const { hasCreateOrUpdateMembersAccess } = await verifyUserRoleAccess(
-    params.organizationId,
-    session.user.id
-  );
-
-  if (!hasCreateOrUpdateMembersAccess) return notFound();
+  // Not the security boundary — `inviteOrganizationMemberAction` is — but this shares the action's
+  // role list so a manager gets a 404 instead of a form that fails on submit. The previous
+  // `verifyUserRoleAccess().hasCreateOrUpdateMembersAccess` flag was true for managers too.
```
Comment (kiln-agent) [inline apps/web/modules/setup/organization/[organizationId]/invite/page.tsx]: **This removes the last production caller of `verifyUserRoleAccess` (`lib/organization/auth.ts`), leaving it dead code.**

Before this PR, this page was the *only* production call site of `verifyUserRoleAccess` — the other flags on its return object (`hasCreateOrUpdateAccess`, `hasDeleteAccess`, `hasDeleteMembersAccess`, `hasBillingAccess`) were already unused elsewhere, only `hasCreateOrUpdateMembersAccess` was read, and only here. Swapping this call for `hasSetupInviteAccess` makes `verifyUserRoleAccess` unreachable from any production code path — only its own `auth.test.ts` still exercises it.

AGENTS.md's Quality Checklist: "Keep code DRY and small; remove dead code and unused imports." Since this diff is what makes the function fully dead, consider deleting `verifyUserRoleAccess` from `apps/web/lib/organization/auth.ts` (and its now-orphaned describe block in `auth.test.ts`) in this same PR rather than leaving it in place.

(Single-pass review without the Agent tool / subagent verify pass — flagging for a human look rather than a fully cross-checked finding. No other correctness bugs found; the authorization fix itself — `SETUP_INVITE_ROLES`, `checkSetupInviteAuthorization`, `hasSetupInviteAccess`, and the accompanying unit/e2e tests — looks correct and well covered.)

---

kiln-agent (commented):
Single-pass review (Agent tool unavailable this run, so no multi-agent fan-out / subagent verify pass). The core security fix — narrowing the onboarding invite path to owners via `SETUP_INVITE_ROLES` / `checkSetupInviteAuthorization` / `hasSetupInviteAccess`, shared between the action and the page gate — checked out correctly against every angle I traced (removed-behavior re-establishment, callers of `checkAuthorizationUpdated`/`getMembershipByUserIdOrganizationId`, the new unit and Playwright tests, import order, denied-role exhaustiveness). One cleanup finding posted inline: this diff makes `verifyUserRoleAccess` fully dead code.

---

apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
Comment (Dhruwang) [reply apps/web/modules/setup/organization/[organizationId]/invite/page.tsx]: I think this makes sense

Dhruwang (commented):

Can you once check on the open comment from kiln?

apps/web/playwright/setup-invite-role-access.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  test("keeps the onboarding invite screen for the organization owner", async ({ page, users }) => {
+    const { owner, inviteUrl } = await setupOrgWithManager(users);
+
+    await owner.login();
+    await page.goto(inviteUrl, { waitUntil: "domcontentloaded" });
+
+    await expect(page.getByText("Invite your Organization members")).toBeVisible();
+  });
+});
```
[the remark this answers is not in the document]
pandeymangg: Fair catch — the spec only covered the page gate, and a server action is its own POST, so `notFound()` there proves nothing about a direct action call.

Went a different route than an action test: `**/actions.ts` is coverage-excluded in `sonar-project.properties`, and asserting which args a mocked `checkAuthorizationUpdated` received tests the wiring, not the decision. So the decision moved to `invite/lib/authorization.ts` — `SETUP_INVITE_ROLES` is the single source of truth, shared by the action (`checkSetupInviteAuthorization`) and the page gate (`hasSetupInviteAccess`).

`authorization.test.ts` mocks only the membership lookup, so the real role matching runs and the assertions are per role: owner allowed, manager/member/billing/no-membership rejected. Verified it fails on the regression — adding `"manager"` back to the list reddens two assertions.

apps/web/modules/setup/organization/[organizationId]/invite/page.tsx
Comment (pandeymangg) [reply apps/web/modules/setup/organization/[organizationId]/invite/page.tsx]: Correct, and done in 2062e1c0e. `verifyUserRoleAccess` and its describe block are gone; `canUserAccessOrganization` stays since the organization layout still uses it.

Confirmed no references left, and typecheck plus the unit suite are green.

Dhruwang (approved):

PR #8818, opened 2026-08-11
fix(settings): hide the workspace breadcrumb on organization and account settings

itsjavi:
## What & why

`/organizations/[organizationId]/settings/**` and `/account/settings/**` are workspace-agnostic routes — there is no `workspaceId` in the URL. The top bar nevertheless rendered a **workspace breadcrumb** there, for a workspace the shell only inferred from the `formbricks-workspace-id` cookie added in #8471. A page that is not scoped to a workspace should not claim one, so that crumb is gone; the organization breadcrumb remains.

- `WorkspaceAndOrgSwitch` and `TopControlBar` take `showWorkspaceBreadcrumb` (default `true`); the workspace-agnostic settings shell passes `false`. Workspace-scoped routes are untouched.
- The organization crumb's closed-state chevron pointed **right**, which worked as a separator while a workspace crumb followed it and read as a dangling separator once nothing did. It now points **down** when it is the last crumb, matching the workspace breadcrumb's own closed-state affordance.

The organization breadcrumb still receives the resolved workspace id, so its menu keeps the entry that depends on it.

**The settings sidebar is deliberately untouched.** It still shows its Workspace section — selector pill and the eight workspace-scoped links — on organization and account settings, exactly as before. Every test below asserts that positively, so this change cannot quietly leak into the sidebar.

**Scope note:** ENG-1705 also proposed removing the sidebar's workspace selector and splitting the settings shell in two. That is **not** in this PR — the decision was to keep the sidebar as-is and narrow this change to the breadcrumb. The ticket's "Done when" allows for that (*"or a documented decision to keep it"*), so this is linked with `Ref` rather than `Fixes` and should not auto-close the ticket.

## Linear ticket

Ref ENG-1705

## Screenshots

Captured on a local stack at `88a38c0` (before) and `a111ed1` (after), same viewport and seed data. **The sidebar is identical in every pair** — the only difference is the top bar.

**Organization settings** — `/organizations/{orgId}/settings/general`. Top bar goes from `Seed Organization › Seed Workspace ⌄` to `Seed Organization ⌄`.

| Before | After |
| --- | --- |
|  |  |

**Account settings** — `/account/settings/profile`. Same treatment; it is the same workspace-agnostic shell.

| Before | After |
| --- | --- |
|  |  |

**Workspace settings** — `/workspaces/{workspaceId}/settings/workspace/general`. Unchanged: `Seed Organization › Seed Workspace ⌄` and the full sidebar.

| Before | After |
| --- | --- |
|  |  |

## How this was tested

Verified against a full local stack (Postgres + migrations + seed, dev server, real Chromium). **CI is green on every check.**

- **New Playwright coverage** — `apps/web/playwright/settings-workspace-breadcrumb.spec.ts`, 4 tests, all passing locally and in CI: organization settings hides the workspace breadcrumb and keeps the sidebar section; account settings the same; workspace settings keeps **both**; and the back arrow still returns to the workspace you came from.
  - Every test asserts the **sidebar Workspace section is present** — pill, heading, and all eight links in order. That is the guard that this change stays in the top bar.
  - The breadcrumb-absence assertions are scoped to the top bar, since the workspace name legitimately appears in the sidebar pill; an unscoped assertion would fail for the wrong reason.
  - They were verified **non-vacuous**: flipping `showWorkspaceBreadcrumb` back to `true` fails the organization test, so it detects the breadcrumb rather than passing on an empty page. Each is also gated on a positive assertion first, since `.count()` does not retry.
  - The back-arrow test creates a second workspace *after* the fixture's, so it is never the first-workspace fallback; landing back on it proves the #8471 cookie still wins (ENG-1700).
- **Build** — `pnpm build --filter=@formbricks/web`: 12/12 tasks, TypeScript clean, 72 static pages generated. This is the check that reproduces CI, since `next build` runs `tsc`; an earlier revision of this branch failed both the E2E and Docker Build jobs on a type error in the build step, before any test ran.
- **Unit** — full `apps/web` suite: **598 files, 7780 tests, all passing**. `navigation-data.test.ts` and `proxy.test.ts` (the #8471 cookie tests) pass **unedited**.
- **Format** — `pnpm format:check` ✅. **Lint** — `eslint` on every touched file: **0 problems**.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Log in and open **Organization Settings** (`/organizations/{orgId}/settings/general`) → the top bar shows **only** `Your Organization ⌄`. There is no `› Your Workspace` after it, and no trailing chevron pointing at nothing. Clicking it opens the **organization** switcher.
- [ ] On that same page, the left sidebar is **unchanged**: it still shows the **WORKSPACE** heading, the workspace name pill, and all 8 workspace links (General, Team Access, Survey Languages, Connect Your App, Integrations, Appearance, User Actions, Tags). Clicking the pill still switches workspace.
- [ ] Open **Account Settings** (`/account/settings/profile`) → same: organization breadcrumb only in the top bar, full sidebar including the Workspace section.
- [ ] Open **Workspace Settings** (`/workspaces/{workspaceId}/settings/workspace/general`) → the top bar **still** shows `Your Organization › Your Workspace ⌄` and the workspace switcher still works there.
- [ ] Open any non-settings workspace page (e.g. the surveys list) → the workspace breadcrumb is still present and functional. *(Only the settings shell opts out.)*
- [ ] **Back arrow / ENG-1700:** with **two or more workspaces**, open workspace B, go to Organization Settings, click the **Back** arrow → you return to **workspace B's** surveys, not workspace A's.
- [ ] Edge case — a user with **exactly one** workspace: all of the above still behave.

**Preconditions / test data**

- Any account with at least one workspace; the back-arrow check needs **two or more workspaces** in the same organization to be meaningful.
- On single-organization self-hosted setups the organization dropdown hides its switcher list — confirm the organization crumb still opens and behaves as the last crumb there.
- Both Cloud and self-hosted; no feature flag or entitlement gates this.

**Risks & regressions**

- `WorkspaceAndOrgSwitch`, `TopControlBar` and the organization breadcrumb are **shared with every workspace-scoped page**. The default is `true`, so only the settings shell changes — but re-check the breadcrumb on the surveys list, a survey editor, and workspace settings.
- The chevron direction change affects the organization crumb wherever it is the last crumb.
- Workspace switching from organization and account settings now happens via the **sidebar pill** rather than the top bar. Worth confirming that reads naturally, since the sidebar pill is the less-used of the two.

- none


Dhruwang (approved):

PR #8824, opened 2026-08-11
feat: auto-update feedback source mappings when survey changes (ENG-2064)

xernobyl:
Fixes ENG-2064

## What & why

A `formbricks_survey` feedback source publishes to Hub by iterating its stored `FeedbackSourceFormbricksMapping` rows — one per mapped question. Nothing kept those rows in step with the survey, and every failure mode was silent: nothing errors, answers just stop arriving or arrive mistyped.

| Survey change | Before | Now |
| --- | --- | --- |
| Question removed | zombie row keeps publishing | row deleted |
| Question retyped to another supported type | publishes under the stale `hubFieldType` | `hubFieldType` corrected |
| Question retyped to an unsupported type (Contact Info, File Upload…) | row survives and keeps publishing — the id is preserved across a retype, so a contactInfo answer lands in Hub as `value_text: "name, email, phone"` | row deleted |
| Question added | **never mapped — answers dropped for good** | mapped, if the source tracks every question |

Reconciliation runs after the survey row is committed, from the editor (`updateSurveyInternal`) and from the v3 API (`executeV3SurveyPatch`), deferred with `after()` so no save pays for it on the request path. It is best-effort: it can never turn a successful save into an error.

**Tracking added questions safely.** Mapping a new question automatically is only sound if we know the operator wanted everything, and stored state cannot tell us — survey elements carry no timestamps, so a question added last week is indistinguishable from one deliberately unticked at setup. Inferring it is impossible too: "the source already maps every supported element" is false *exactly* when there is something to add, because the new element is itself supported and unmapped.

So the intent is recorded where it **is** known — the moment a selection is submitted — in a new `FeedbackSource.elementScope`:

- **`all`** — the submitted selection covered every mappable element of every mapped survey → new questions are picked up as they appear.
- **`specific`** — something mappable was left out → reconciliation never adds anything.

Derived server-side in `resolveFormbricksMappingsInput` and written in the same transaction as the rows it describes, so it cannot disagree with them and a client cannot set it. This is the house pattern for the same decision: webhooks treat `surveyIds: []` as "all surveys"; segment survey-interaction filters carry `surveyScope: "any" | "specific"`.

<details>
<summary>Invariants a reviewer may want to check directly</summary>

- **Scoped by survey.** A source can map several surveys, so the reconciler filters by `surveyId` before diffing and every write carries `surveyId` — mirroring `transform.ts`, which already filters by `survey.id` when publishing. Without it, saving survey A deleted survey B's mappings.
- **Diffs persisted blocks, never the request payload.** An update that omits `blocks` deliberately leaves the stored questions alone; diffing its empty payload would read as "no questions" and delete every mapping. The deferred callback re-reads the blocks rather than trusting a captured snapshot, so a late reconcile cannot undo a newer save.
- **Inert vs leaking rows.** A row whose element is *gone* publishes nothing (the publish path looks the answer up by id and skips it), so it may be held back to keep the source discoverable. A row whose element was *retyped to an unmappable type* is still exporting, and is always deleted.
- **PII types can never be auto-mapped.** `contactInfo`, `address`, `cal`, `cta`, `fileUpload` and `consent` have no Hub field, so they are excluded from both the create arm and the scope derivation.
- **Recoverable failure.** A source stripped of its last mapping is flagged `status: "error"` (already a rendered badge, and it leaves the publish path). Reconciliation clears the flag again once it re-creates a row, and the edit modal now lets a zero-row source pick a survey so the state is repairable rather than terminal.
</details>

## Breaking changes

None

## Migrations & env

- `20260812120000_add_feedback_source_element_scope` — creates the `FeedbackSourceElementScope` enum and adds `FeedbackSource.elementScope NOT NULL DEFAULT 'specific'`. Additive and metadata-only (`ADD COLUMN … NOT NULL DEFAULT` does not rewrite the table on PG 11+).
- No data backfill, deliberately. Existing sources default to `specific`, so their behaviour is unchanged; they pick up their real scope the next time an operator saves the source. Backfilling to `all` would have to guess, and guessing wrong pushes questions someone deliberately excluded into a dataset.
- No env vars.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Editing survey A no longer deletes survey B's mappings | unit (mutation) | `mapping-reconciliation.test.ts` "ignores mappings that belong to another survey" + "scopes the delete to the reconciled survey". Removing the `surveyId` filter at `mapping-reconciliation.ts:66` fails 3; removing it from the `deleteMany` where fails 2 |
| Same, against the real schema | manual | Two-survey source on a local Postgres: replaying the pre-fix query shapes deletes the sibling survey's row while that survey is untouched; post-fix shapes leave it intact |
| Reconciles persisted blocks, not the request payload | unit (mutation) | `survey/service.test.ts` "reconciles against the persisted blocks, not the caller's payload". Reverting to `updatedSurvey.blocks ?? []` fails it |
| A question retyped to an unmappable type stops publishing | unit (mutation) | `mapping-reconciliation.test.ts` `test.each` over all six unsupported types, plus "deletes a lone mapping retyped to %s even though it empties the survey". Restoring the old single-list guard fails 4 |
| Same, plus the inert/leaking split, on the real schema | manual | Element retyped to `contactInfo` in a scratch survey: leaking row deleted and source flagged, while a source also holding an inert row keeps that one and stays healthy — 5/5 assertions |
| Added questions are mapped only when `elementScope` is `all` | unit (mutation) | `mapping-reconciliation.test.ts` "elementScope gate" block. Ignoring the stored scope fails 1; forcing the derivation to `all` in `mappings.ts` fails 3 |
| Same gate, real schema | manual | Two sources on one survey differing only in `elementScope`: `all` adopted the unmapped question, `specific` left it alone, `all` dropped its stale row, `specific` kept its last row — 4/4 |
| `elementScope` is persisted on create and on update | unit (mutation) | `service.test.ts`; deleting either write spread fails 2 |
| An errored source is not un-errored by a csv save | unit (mutation) | `service.test.ts` "replaces field mappings when provided" asserts the tracked status is still `error`. Reverting the type scoping fails with `Expected: "error" / Received: "active"` |
| Reconciliation clears `error` once it re-creates a row | unit (mutation) | `mapping-reconciliation.test.ts` "clears an error flag once it has given the source something to publish again"; removing the block fails it |
| Deferred survey re-read is workspace-scoped | unit (mutation) | `mapping-reconciliation.test.ts` "scopes the deferred survey re-read to the workspace"; dropping `workspaceId` from the where fails it |
| Reconciliation covers paused sources, so a resume cannot publish a stale row | unit (mutation) | `service.test.ts` "returns sources of every status"; restoring the `status: "active"` filter fails it. Confirmed on the real DB: the publish-path query returns 0 for a paused source, the reconcile query returns it |
| Cross-workspace mapping rows are impossible | manual | Attempted insert rejected by `FeedbackSourceFormbricksMapping_surveyId_workspaceId_fkey` — DB-enforced, not convention |
| `skipDuplicates` really no-ops on the concurrent-create path | manual | Same insert without `ON CONFLICT` raises `unique_violation`; with it, `INSERT 0 0` and no error |
| `updateMany` scoping survives the hardest case | manual | Same `elementId` mapped in two surveys of one source: retyping in survey A left survey B's row untouched |
| Whole flow from a real editor save | e2e | `playwright/feedback-source-reconciliation.spec.ts`, 2 tests — scope gate, sibling-survey isolation, stale-row cleanup and the retype correction. Both confirmed executing and passing in CI from the `playwright-report` artifact, not inferred from the check mark |

**Open gaps**

- **`elementScope` has no UI.** It is derived from the selection, so unticking anything opts out, but nothing shows an operator which mode a source is in and there is no way to say "track exactly these" while everything is ticked. The read-only copy for this was written and dropped: new keys need the other 14 locales generated and `pnpm i18n` fails on an invalid `LINGO_API_KEY`, and hand-writing locale files is against AGENTS.md. Approach recorded on ENG-2358.
- **The scope is per source, not per (source, survey).** A source mapping several surveys carries one value. Today's UI only creates single-survey sources; ENG-2341 tracks whether multi-survey should be supported at all.
- **Not transactional with the survey write.** Reconciliation runs after the row is committed, so a crash in between leaves mappings stale until the next save. Deliberate: it must never fail a save.
- **No E2E for the surrounding source surface** (create/edit/CSV/import) — Unify Feedback had none at all before this PR. ENG-2357.

**Risks**

- Every survey save now runs one extra query to find sources, off the request path via `after()`. The transaction only opens on a non-empty delta, and `hubFieldType` updates batch by target type rather than one statement per element.
- Concurrent edits: last writer wins; `createMany` uses `skipDuplicates`, so a concurrent create is a no-op rather than a failed transaction.
- If reconciliation strips a source's last mapping it is flagged `error` and stops publishing. That is intended, but it is the one path that changes a source's status without an operator action — re-check the sources table if a source unexpectedly stops syncing.

## Follow-ups filed

- ENG-2341 — `updateFeedbackSourceWithMappings` replaces mappings unscoped by survey (same bug class, pre-existing).
- ENG-2357 — no E2E coverage for the surrounding Unify Feedback surface.
- ENG-2358 — surface and directly control `elementScope`.

---

> [!NOTE]
> **AI model used** — `claude-opus-5`. Reasoning effort not recorded by the harness.


apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,314 @@
...
+ * that were actually written — diffing against a request payload would delete mappings for questions
+ * that are still in the database.
+ */
+export const reconcileFeedbackSourcesForSurvey = async (
+  surveyId: string,
+  blocks: TSurveyBlock[]
+): Promise<void> => {
+  try {
+    const feedbackSources = await getFeedbackSourcesToReconcile(surveyId);
+    for (const source of feedbackSources) {
```
mattinannt: Non-blocking efficiency note: this loop `await`s `applyReconciliationToFeedbackSource` once per matching feedback source, even though each source's reconciliation is scoped to a distinct `feedbackSourceId` and opens its own independent transaction. Both `updateSurveyInternal` and `executeV3SurveyPatch` await `reconcileFeedbackSourcesForSurvey` synchronously before responding, so a survey mapped by several sources pays N sequential round-trips on every save instead of being bounded by the slowest one. Running them concurrently (e.g. `Promise.allSettled(feedbackSources.map(...))`, since each call already swallows its own errors) would avoid the linear latency growth.

---

mattinannt (commented):
Reviewed the reconciliation logic (`classifyStoredMappings`, `resolveDeletions`, `elementScope` derivation, the transaction in `applyReconciliationToFeedbackSource`) against the extensive test suite and traced every branch — didn't find a correctness defect; the multi-survey scoping, the "never empty a survey's last mapping unless something is being created" guard, and the leaking-vs-inert row distinction all check out. One non-blocking efficiency note left inline.

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * A source with zero rows for a survey is unrecoverable: every other write path requires min(1), and
+ * getFeedbackSourcesToReconcile matches on `formbricksMappings: { some: { surveyId } }`, so the source
+ * would never be found for this survey again. Where the only rows left are inert, keeping them costs
+ * nothing and preserves that handle — a pending create rescues the source too, so the hold only
+ * applies when nothing is being added.
+ *
+ * Leaking rows are never held back: stopping the export outranks keeping the source discoverable.
+ */
+const resolveDeletions = (
```
Dhruwang, line 106: **P1:** This hold-back rests on "the row is inert" — but `transformResponseToFeedbackRecords` has no `if (!element) continue`. It guards on the *response value*, then tolerates `element === undefined`, so any response still carrying this element id publishes a record labelled `"Untitled"` — and `importHistoricalResponses` replays every historical response through the same loop. Nothing purges `response.data` when a question is deleted, so held-back rows keep publishing deleted questions.

Either guard the publish loop (`if (!element) continue`), or drop the hold-back and find the source another way.

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,314 @@
...
+      // a source still serving a sibling survey is working, and erroring it would stop that too.
+      if (toDelete.length > 0) {
+        const remainingMappings = await tx.feedbackSourceFormbricksMapping.count({
+          where: { feedbackSourceId, workspaceId },
+        });
+
+        if (remainingMappings === 0) {
+          await tx.feedbackSource.update({
+            where: { id: feedbackSourceId, workspaceId },
+            data: { status: "error" },
```
Dhruwang, line 293: **P1:** Nothing ever clears this. `handleUpdateFeedbackSource` sends only `{name, importMode}` and `ZFeedbackSourceUpdateInput.status` is optional, so re-mapping leaves the column untouched — while `getFeedbackSourcesBySurveyId` filters `status: "active"`. A repaired source stays dark; only the unrelated pause/resume toggle escapes. Suggest resetting to `active` in `updateFeedbackSourceWithMappings` when the input resolves to >=1 row.

**P2:** Also unconditional, so it overwrites `paused`. The resume toggle then flips it to `active`, silently resuming a source the operator had deliberately disabled.

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  const wouldRemoveEveryMapping =
+    !hasPendingCreates && surveyMappingCount > 0 && removed.length + unmappable.length === surveyMappingCount;
+
+  if (!wouldRemoveEveryMapping) {
+    return [...removed, ...unmappable];
+  }
+
+  if (removed.length > 0) {
+    logger.warn(
```
Dhruwang: **P2:** Both branches dead-end, but only one is visible. This one holds rows back with a server-side `logger.warn` and leaves `status: "active"` — green "Live sync" badge, nothing published. The other deletes everything and flags `error`, but the edit modal takes its survey from `formbricksMappings[0]?.surveyId` with the survey `Select` hardcoded `disabled`, so a zero-row source can only be fixed by delete+recreate.

apps/web/lib/survey/service.ts
```diff
@@ -622,6 +623,12 @@ export const updateSurveyInternal = async (
       select: selectSurvey,
     });
 
+    // ENG-2064: keep feedback-source mappings in sync with the survey's questions. Diff against the
+    // blocks that were actually persisted, not the caller's payload — a partial update that omits
+    // blocks leaves the stored questions untouched, and diffing its empty payload would delete every
+    // mapping. Best-effort: a failure logs inside the helper and never blocks the save.
+    await reconcileFeedbackSourcesForSurvey(surveyId, persistedSurvey.blocks);
```
Dhruwang: **P2:** Awaited inline, and the editor autosaves drafts every 10s — so any survey backed by a source pays an extra `findMany` (plus a transaction on a real delta) on routine typing. It's already best-effort and swallows its own errors, so nothing depends on awaiting it.

apps/web/lib/feedback-source/survey-elements.ts
```diff
@@ -0,0 +1,42 @@
...
+ * is the one place that cast lives.
+ */
+export const indexSurveyElements = (blocks: TSurveyBlock[]): TSurveyElementIndex => {
+  const elementIds = new Set<string>();
+  const supportedHubFieldTypes = new Map<string, THubFieldType>();
+
+  for (const element of getElementsFromBlocks(blocks)) {
+    elementIds.add(element.id);
+
+    const hubFieldType = getHubFieldTypeFromElementType(element.type) as THubFieldType | undefined;
```
Dhruwang: **P2:** `getHubFieldTypeFromElementType` has exactly one caller repo-wide, and this PR already edits `packages/types/feedback-source.ts` — widening its return to `THubFieldType | undefined` is a one-liner that removes both this cast and the comment explaining it.

apps/web/playwright/feedback-source-reconciliation.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+    where: { id: surveyId },
+    select: { blocks: true },
+  });
+
+  return (survey.blocks as unknown as { elements: TSeededElement[] }[]).flatMap(
+    (block) => block.elements ?? []
+  );
+};
+
+const UNMAPPABLE_ELEMENT_TYPES = new Set(["contactInfo", "address", "cal", "cta", "fileUpload", "consent"]);
```
Dhruwang: **P2:** `UNSUPPORTED_FEEDBACK_SOURCE_ELEMENT_TYPES` is already exported from `@formbricks/types/feedback-source` and imported by two other call sites — import it rather than re-declaring the six strings.

apps/web/playwright/feedback-source-reconciliation.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    await expect
+      .poll(() => mappedElementIds(trackAllSourceId, surveyId), { timeout: 15000 })
+      .toEqual(supported.map((element) => element.id).sort((a, b) => a.localeCompare(b)));
+
+    // Sibling survey's row survives — editing one survey must not touch another's mappings.
+    expect(await mappedElementIds(trackAllSourceId, siblingSurvey.id)).toEqual([SIBLING_ELEMENT]);
+
+    // Curated source: no questions adopted, and it was not left with zero rows.
+    expect(await mappedElementIds(curatedSourceId, surveyId)).toEqual([STALE]);
```
Dhruwang: **P2:** Seeded `[STALE]`, asserted `[STALE]` — this passes identically if reconciliation never ran, never found the source, or threw. It also doesn't poll, unlike the `trackAll` assertion above. Seed a row that *should* be removed so the guard is what makes it pass. (Same point coderabbitai raised.)

apps/web/app/api/v3/surveys/patch.test.ts
```diff
@@ -280,6 +287,15 @@ describe("patchV3Survey", () => {
     );
   });
 
+  // ENG-2064: this route writes blocks directly instead of going through updateSurveyInternal, so
+  // without its own call a survey edited over the v3 API would leave its feedback-source mappings
+  // stale — and the API is exactly the surface an automation changes questions from.
+  test("reconciles feedback-source mappings against the persisted blocks", async () => {
+    await patchV3Survey(currentSurvey, { name: "renamed via v3" }, "req_qa", "org_1");
+
+    expect(reconcileFeedbackSourcesForSurvey).toHaveBeenCalledWith(currentSurvey.id, expect.any(Array));
```
Dhruwang: **P2:** `expect.any(Array)` doesn't pin the thing worth pinning — that the blocks are the *persisted* ones, not the request payload. `survey/service.test.ts` has the fixture pattern for this (`persistedBlocks` distinguishable from the payload).

Dhruwang (commented):
Review of the reconciliation logic. Two P1s worth resolving before merge — the hold-back guard's inertness premise, and the `status: "error"` flag having no clearing path. The rest are P2. Architecture itself looks sound: multi-survey scoping, persisted-vs-payload diffing and PII-retype deletion all check out.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Confirmed and fixed in `79095ebab`. You were right on the mechanism, and the consequence was worse than "labelled Untitled".

`transformResponseToFeedbackRecords` guarded only the response **value**, then every branch tolerated a missing element: the three type checks were optional-chained (`element?.type === …`) so they all fell through, and the generic path ended at `getHeadlineFromElement(element)`, which returns `"Untitled"` for `undefined`. There was no `if (!element) continue` anywhere in the loop.

I took your first option — guard the publish loop rather than drop the hold-back — because the hold-back is a real property worth keeping and the missing guard is the actual defect. It also makes `resolveDeletions`' comment true for the first time: it keeps orphaned rows "where the only rows left are inert", and they were not inert until this guard existed.

Also tightened the three `element?.type` chains to `element.type` now that they are provably non-null, so the guard cannot be deleted without a type error somewhere.

**Coverage, mutation-tested.** Two new cases in `transform.test.ts`: an orphaned mapping publishes nothing, and an orphaned mapping alongside a live one still publishes the live one (the guard must skip the row, not abandon the response). Removing the guard fails exactly those two and nothing else.

**One thing that needed correcting rather than accommodating.** Two existing tests — "joins array values to comma-separated text for non-choice elements" and its empty-array sibling — went red. They used `elementId: "el-array"`, an element **absent from the survey**, and their own comment said so: *"An element absent from the survey definition has no type, so the generic path stores the raw array joined."* They were encoding the exact behaviour you reported as a bug. Their real intent is value conversion, so I repointed them at the real `el-text` openText element; they still cover the generic array-joining path.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Both halves confirmed and fixed in `79095ebab`.

**P1 — the source stays dark.** Verified the whole chain rather than the symptom: `handleUpdateFeedbackSource` sends `feedbackSourceInput: { name, importMode }` only; `ZFeedbackSourceUpdateInput.status` is `.optional()`, so `data.status` is `undefined` and Prisma treats it as "leave alone"; `getFeedbackSourcesBySurveyId` filters `status: "active"` (`service.ts:149`). Nothing else writes the column. So a source reconciliation flagged `error` was repaired by a re-map and then never reconciled or published again.

`updateFeedbackSourceWithMappings` now resets it, on your suggested condition — the input resolves to at least one row — plus two deliberate narrowings: only when the caller did not set `status` itself, and only from `error`.

Worth recording how I got that wrong first: my initial version put `status: "error"` in the `where` of the main `update`. That would have failed to match every **healthy** source and thrown P2025 on the normal save path. It has to be a separate `updateMany`, which no-ops when nothing matches.

**P2 — clobbering `paused`.** Also right. The reconciler's write was unconditional once `remainingMappings === 0`, so a paused source became `error`, and the resume toggle would then flip it to `active` — silently resuming something an operator had deliberately disabled. Scoped to `status: "active"`, and the "flagged as errored" log now only fires when a row actually changed, so the log stops claiming something that did not happen.

A paused source with no mappings stays paused: it is not publishing either way, and `error` is a claim about the source being broken rather than about it being switched off.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Agreed on the diagnosis, and I would rather not fix it here.

You are right that the two dead-ends are asymmetric: the hold-back leaves `status: "active"` with a green "Live sync" badge and nothing published, while the zero-row path flags `error` but is unrecoverable through the UI, because the edit modal takes its survey from `formbricksMappings[0]?.surveyId` with the survey `Select` hardcoded `disabled` — so a source with no rows has no survey to render and no way to pick one.

Two of the three parts are better now. The `error` path is genuinely recoverable: re-mapping resets the status (see the P1 thread), so it is no longer a one-way door at the data layer. And the held-back rows are inert for real now that the publish loop skips orphans — the badge is still optimistic, but it is no longer green-while-exporting-a-deleted-question.

What is left is the UI: making a zero-row source re-mappable needs the survey `Select` enabled and a sensible empty state, which is a component change with its own design question (which surveys should it offer?) rather than anything this diff touches. I would rather file it than widen a reconciliation PR into the edit modal. @xernobyl — worth a ticket, and I will open it once Linear is reachable from my session.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Fixed in `79095ebab`. Each source reconciles a distinct `feedbackSourceId` in its own transaction with no data dependency on the others, exactly as you say, so the loop now builds the non-empty reconciliations first and runs them through `Promise.allSettled`.

`allSettled` rather than `all` for the reason you gave — `applyReconciliationToFeedbackSource` already swallows its own errors — so it is belt-and-braces rather than load-bearing, and one source cannot take out another's write.

It also stopped being on the request path entirely: see the survey-service thread. Between the two, a survey mapped by several sources no longer pays anything on save.

apps/web/lib/survey/service.ts
xernobyl: Fixed in `79095ebab`, though not by simply dropping the `await`.

You are right that nothing depends on it: it is best-effort and swallows its own errors, and the response does not contain its result. But an un-awaited promise in a serverless runtime can be killed when the response is sent, which would turn "slower saves" into "mappings that silently stop reconciling" — a worse failure than the one being fixed, and an invisible one.

So both call sites now go through `scheduleFeedbackSourceReconciliation`, which defers the work with `next/server`'s `after()` and falls back to awaiting inline when `after()` throws — scripts, jobs and tests have no request to run after, and there the inline await is the correct behaviour anyway.

Combined with the concurrency fix on the sibling thread, the editor's 10-second autosave now pays nothing for reconciliation on the request path.

apps/web/lib/feedback-source/survey-elements.ts
xernobyl: Fixed in `79095ebab` — `getHubFieldTypeFromElementType` now declares `THubFieldType | undefined`, and the cast plus the comment explaining it are both gone from `survey-elements.ts`. You were right that it is a one-liner: one caller repo-wide, and this PR was already editing `packages/types/feedback-source.ts`.

apps/web/playwright/feedback-source-reconciliation.spec.ts
xernobyl: Fixed in `79095ebab`. The spec now imports `UNSUPPORTED_FEEDBACK_SOURCE_ELEMENT_TYPES` and builds its `Set` from it, so a type added to the product constant reaches the spec instead of leaving it asserting an old six.

apps/web/app/api/v3/surveys/patch.test.ts
xernobyl: Fixed in `79095ebab`. It now asserts `currentSurvey.blocks` — the stored value — rather than `expect.any(Array)`.

Worth noting why the value matters and not just the shape: that patch sends no `blocks` at all, so reconciling against the request payload would pass `undefined`, which `expect.any(Array)` would have caught — but so would an empty array, and reconciling an empty block list deletes every mapping the survey has. Added `expect(currentSurvey.blocks.length).toBeGreaterThan(0)` so the fixture cannot quietly become empty and make the assertion vacuous again.

apps/web/playwright/feedback-source-reconciliation.spec.ts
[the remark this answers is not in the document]
xernobyl: Valid, and fixed in `79095ebab` — but not with the proposed fixture, which would have quietly removed the behaviour it was meant to preserve.

The suggestion adds a second mapping row to the curated source and then asserts *"the stale row must still survive"*. It cannot: `resolveDeletions` holds rows back only when `removed.length + unmappable.length === surveyMappingCount`, so a second row makes that false and the stale row is deleted. The fixture would have retargeted the assertion at a different scenario while still reading as if it covered the hold-back.

Your underlying point stands, so the fix uses **two** curated sources:

- `curatedHeldBackSourceId` — one stale row only, so the hold-back still applies and the assertion still means what it says.
- `curatedRetypedSourceId` — a stale row plus a real element with a deliberately wrong `hubFieldType`. The hold-back does not apply here, so reconciliation must delete the stale row *and* correct the type. That is the discriminating signal, and it is what the poll now waits on: seeded `[STALE, supported[0]]`, asserted `[supported[0]]`, so it cannot pass if reconciliation never ran, never matched the source, or threw.

The hold-back assertion then runs after that poll, so it is credible rather than racy — and it is independently unit-covered in `mapping-reconciliation.test.ts`, which is the better home for it.

xernobyl: Fixed in `79095ebab`, and you and coderabbitai landed on the same line from different angles.

Seeded `[STALE]` / asserted `[STALE]` was the whole problem: it passed if reconciliation never ran, never matched the source, or threw. It is now two curated sources — one keeping the hold-back scenario intact, one seeded with a row that **must** be removed plus a real element that must be retyped, which is the signal the poll waits on. Your framing ("seed a row that should be removed so the guard is what makes it pass") is exactly what the second source does.

One correction worth recording: coderabbitai's suggested fixture put both rows on the *same* source. That would have made `removed.length + unmappable.length === surveyMappingCount` false, disabling the hold-back and deleting the stale row — so the assertion it proposed ("the stale row must still survive") could not have passed. Splitting into two sources is what keeps both properties testable.

It also polls now, like the `trackAll` assertion above it.

packages/database/migration/20260812120000_add_feedback_source_element_scope/migration.sql
```diff
@@ -0,0 +1,9 @@
+-- CreateEnum
+CREATE TYPE "FeedbackSourceElementScope" AS ENUM ('all', 'specific');
+
+-- AlterTable
+-- Existing sources default to 'specific' so their behaviour does not change: reconciliation will not
+-- add mappings for questions added after they were connected until an operator next saves the source,
+-- at which point the scope is derived from the selection they submit. Backfilling to 'all' would have
+-- to guess, and guessing wrong pushes questions the operator deliberately excluded into the dataset.
+ALTER TABLE "FeedbackSource" ADD COLUMN "elementScope" "FeedbackSourceElementScope" NOT NULL DEFAULT 'specific';
```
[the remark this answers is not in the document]
xernobyl: Valid, and fixed in `79095ebab` — this one is a documented path rule (`packages/database/migration/**/migration.sql`: *"The migration is idempotent and convergent — re-running it, or running it against a database created with `db:push`, must not fail"*), and both statements broke it.

`CREATE TYPE` is now wrapped in a `DO $$ … EXCEPTION WHEN duplicate_object THEN NULL` block and the column is `ADD COLUMN IF NOT EXISTS`, as proposed. The explanatory comment about why existing sources default to `specific` is kept — it is the part a reader needs when deciding whether to backfill.

xernobyl:
Pushed `26b075529` — CI Typecheck was red on my previous push, and the cause was not the review fixes.

`@formbricks/web#typecheck` failed with 33 errors, all `Cannot find module '@formbricks/jobs'` and its fallout, in files this PR never touches. Turbo reported "24 successful, 25 total" — the dependency builds were fine and only the typecheck task failed, which is the tell: `@formbricks/jobs` was never built *for that task*.

Root cause: `turbo.json` declares an explicit `dependsOn` for `@formbricks/web#typecheck`, and Turbo **replaces** such lists rather than merging them, so a package missing from it is not built even though `apps/web` depends on it (`@formbricks/jobs: workspace:*`). `main` has `@formbricks/jobs#build` in that list. This branch did not.

It was not dropped by me: the only commit touching `turbo.json` on this branch is the earlier `Merge origin/main` (`4f7931ef0`), which resolved the file to the branch side and silently reverted several of main's entries — `@formbricks/jobs#build`, the `@formbricks/database#generate` task and `@formbricks/jobs#typecheck` were all lost. The branch had no intentional changes to that file at all.

Rather than re-add one line, I merged current `main`, since the branch was behind it anyway and a merge that silently reverts upstream config is exactly the kind of thing that comes back. The merge was clean — no conflicts — and I verified afterwards that `turbo.json` matches main and every review fix is still in place.

Local state after the merge: `transform.test.ts` 60/60 including the two mutation-verified guard tests, typecheck clean, Prettier clean.

Worth flagging for whoever reviews: this also means anything else that merge dropped is still dropped on other branches cut from it.

BhagyaAmarasinghe (dismissed):

apps/web/lib/feedback-source/mapping-reconciliation.test.ts
```diff
@@ -0,0 +1,561 @@
...
+    const tx = mockTx();
+    tx.feedbackSourceFormbricksMapping.count.mockResolvedValue(0);
+
+    await applyReconciliationToFeedbackSource(SOURCE_ID, WORKSPACE_ID, SURVEY_ID, {
+      toCreate: [],
+      toDelete: ["el-last"],
+      toUpdate: [],
+    });
+
+    expect(tx.feedbackSource.update).toHaveBeenCalledWith({
```
kiln-agent: **Failing test** — this assertion was not updated when the flagging write changed from `update` to `updateMany`.

`applyReconciliationToFeedbackSource` now calls `tx.feedbackSource.updateMany({ where: { id, workspaceId, status: "active" }, ... })` (mapping-reconciliation.ts:273), but `mockTx()` at line 50 only stubs `feedbackSource: { update: vi.fn() }`. So `tx.feedbackSource.updateMany` is `undefined`, the call throws `TypeError` inside `prisma.$transaction`, the `catch` in `applyReconciliationToFeedbackSource` swallows it as "failed to apply reconciliation", and `tx.feedbackSource.update` is never called.

Verified locally:

```
FAIL lib/feedback-source/mapping-reconciliation.test.ts > applyReconciliationToFeedbackSource >
  flags the source as errored when the delete leaves it with no mappings at all
AssertionError: expected "vi.fn()" to be called with arguments: [ { where: {…}, …(1) } ]
Number of calls: 0
```

Worse than a red suite: because the TypeError is swallowed, the *entire* error-flag path — including the new `status: "active"` scoping that is the stated reason for switching to `updateMany` — is now unexercised. Add `updateMany: vi.fn().mockResolvedValue({ count: 1 })` to the tx mock and assert on `updateMany` with the `status: "active"` predicate.

---

apps/web/lib/feedback-source/service.ts
```diff
@@ -427,6 +488,16 @@ export const updateFeedbackSourceWithMappings = async (
         );
       }
 
+      if (clearsErrorStatus) {
+        // updateMany, not update: putting `status: "error"` in a `where` would fail to match every
+        // healthy source and throw P2025 on the normal save path. This must be a no-op unless the
+        // source really was errored.
+        await tx.feedbackSource.updateMany({
```
kiln-agent, line 503: **Two failing tests** — this new `updateMany` is missing from the transaction mock in `service.test.ts`.

`setupTransaction()` for the `updateFeedbackSourceWithMappings` block (service.test.ts:736) stubs only `feedbackSource: { update, findUniqueOrThrow }`. `clearsErrorStatus` is true in both of those tests (they pass mappings and no `status`), so this line dereferences a missing mock:

```
FAIL lib/feedback-source/service.test.ts > updateFeedbackSourceWithMappings > replaces formbricks mappings when provided
FAIL lib/feedback-source/service.test.ts > updateFeedbackSourceWithMappings > replaces field mappings when provided
TypeError: tx.feedbackSource.updateMany is not a function
 ❯ lib/feedback-source/service.ts:495:33
```

Add `updateMany: vi.fn().mockResolvedValue({ count: 0 })` to that tx mock (and ideally one test that pins the `status: "error"` → `active` transition, which currently has no coverage).

---

apps/web/lib/feedback-source/service.ts
```diff
@@ -378,13 +421,31 @@ export const updateFeedbackSourceWithMappings = async (
...
+      // clears the column — the edit modal sends only `{name, importMode}` and `status` is optional on
+      // the input, so Prisma treats it as "leave alone" — while `getFeedbackSourcesBySurveyId` filters
+      // `status: "active"`. Re-mapping a broken source repaired its rows and left it dark forever,
+      // reachable only through the unrelated pause/resume toggle.
+      //
+      // Scoped deliberately: only when this save supplies mappings (so there is something to publish
+      // again), only when the caller did not set `status` itself, and only from `error` — a `paused`
+      // source stays paused, because pausing is an operator decision and re-mapping is not a request
+      // to resume.
+      const clearsErrorStatus = Boolean(mappingsInput?.mappings.length) && data.status === undefined;
```
kiln-agent: `clearsErrorStatus` is true for **any** `mappingsInput`, including `type: "field"`, but the only thing that ever writes `status: "error"` is the formbricks mapping reconciler. So a save that supplies *field* mappings clears an error that only broken *formbricks* mappings can have caused.

Concrete path: `updateFeedbackSourceWithMappingsAction` takes `fieldMappings` regardless of source type — it only uses `feedbackSource.type === "csv"` to decide whether to sanitize (actions.ts:243-249). Reconciliation flags a `formbricks_survey` source `error` after removing its last mapping; a subsequent update carrying `fieldMappings` (and no `formbricksMappings`) then takes the `else if (mappingsInput?.type === "field")` branch, leaves `feedbackSourceFormbricksMapping` completely untouched, and still flips `error` → `active`. The source renders as healthy with a green badge while it still has zero formbricks mappings and publishes nothing.

The comment above says "only when this save supplies mappings (so there is something to publish again)", which is exactly the invariant the check misses. Gate it on the mapping kind that can actually repair the source:

```suggestion
      const clearsErrorStatus =
        mappingsInput?.type === "formbricks_survey" &&
        mappingsInput.mappings.length > 0 &&
        data.status === undefined;
```

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,361 @@
...
+ * be killed when the response is sent, which would turn "slower saves" into "mappings that silently
+ * stop reconciling". Outside a request — scripts, jobs, tests — `after()` throws, and there the inline
+ * await is correct anyway, so that path falls back to it.
+ */
+export const scheduleFeedbackSourceReconciliation = async (
+  surveyId: string,
+  blocks: TSurveyBlock[]
+): Promise<void> => {
+  try {
+    after(() => reconcileFeedbackSourcesForSurvey(surveyId, blocks));
```
kiln-agent: Deferring with `after()` removes the only thing that serialized two consecutive saves, and the deferred work carries a **captured `blocks` snapshot** — so a stale reconcile can now delete a mapping a newer save just created.

Before this change the reconcile completed before the response, so a client that waits for save N before issuing save N+1 could not interleave them. Now:

1. `PATCH /api/v3/surveys/{id}` (or an editor save) removes question `q`; the response is sent; reconcile-1 is queued with `blocks₁` (no `q`).
2. The caller, having received the 200, immediately sends save 2 re-adding `q`. It persists and its reconcile creates the mapping for `q` (scope `all`).
3. Reconcile-1 finally runs `getFeedbackSourcesToReconcile` → now sees the row for `q`, diffs it against `blocks₁` where `q` does not exist, classifies it as `removed`, and deletes it.

The mapping for a question that *is* in the survey is gone, and nothing re-creates it until the next save — i.e. answers to `q` silently stop reaching Hub, the exact failure mode this PR exists to fix. Only the queue-vs-run ordering makes this narrow, not any guard in the code.

Two things would close it without giving up the off-request-path win: re-read the survey's blocks *inside* `reconcileFeedbackSourcesForSurvey` instead of capturing them at schedule time (the caller has just committed them, so a fresh read is always at least as new as the snapshot), and/or make the reconcile per-survey mutually exclusive (an advisory lock keyed on `surveyId`, or `updatedAt`-based staleness check). At minimum the "Concurrent edits: last writer wins" note in the PR body understates this: with a snapshot, the *last* writer is not necessarily the winner.

---

kiln-agent (commented):
Reviewed the full diff (20 files) for correctness. The reconciliation logic itself holds up well — the survey-id scoping, the `elementScope` gate, the removed/unmappable split and the never-empty-a-survey hold-back all check out against the callers, and `UNSUPPORTED_FEEDBACK_SOURCE_ELEMENT_TYPES` + `ELEMENT_TYPE_TO_HUB_FIELD_TYPE` are exact complements over all 18 `TSurveyElementTypeEnum` members, so the UI's filter and the server's scope derivation can't disagree today.

Four findings, two of them confirmed by running the suite:

1. **`mapping-reconciliation.test.ts:452` — failing test.** The last commit changed the error-flag write to `updateMany`; the tx mock still only stubs `update`, so the call throws, gets swallowed by the best-effort `catch`, and the assertion sees zero calls. The error-flag path is now entirely unexercised.
2. **`service.ts:495` — two failing tests.** Same root cause on the `service.test.ts` side: `tx.feedbackSource.updateMany is not a function`.
3. **`service.ts:434`** — `clearsErrorStatus` isn't gated on `formbricks_survey`, so a field-mapping save clears an error only broken formbricks mappings could have caused.
4. **`mapping-reconciliation.ts:357`** — `after()` plus a captured `blocks` snapshot lets a stale reconcile delete a mapping a newer save just created.

Repro for 1 and 2 (after `prisma generate` + building `@formbricks/database`, `logger`, `storage`, `cache`):

```
$ npx vitest run lib/feedback-source lib/survey/service.test.ts app/api/v3/surveys/patch.test.ts
Tests  3 failed | 289 passed (292)
```

`prettier --check` is clean on every changed `.ts`/`.tsx`, and `tsc` reports nothing new for the files in this diff.

---

apps/web/lib/feedback-source/service.ts
kiln-agent, line 503: **Breaks two existing unit tests.** `setupTransaction()` in `apps/web/lib/feedback-source/service.test.ts` (line 734) builds the tx client as `feedbackSource: { update, findUniqueOrThrow }` — there is no `updateMany`.

Both `"replaces formbricks mappings when provided"` and `"replaces field mappings when provided"` pass a `mappingsInput` with one mapping and no `status`, so `clearsErrorStatus` is `true` and this line evaluates `tx.feedbackSource.updateMany(...)` → `TypeError: tx.feedbackSource.updateMany is not a function`. `updateFeedbackSourceWithMappings` only special-cases Prisma errors, so it rethrows and both tests fail.

Commit `79095eb` introduced the `updateMany` without touching `service.test.ts`. Adding `updateMany: vi.fn().mockResolvedValue({ count: 0 })` to that mock (plus an assertion for the new behaviour) fixes it.

---

apps/web/lib/feedback-source/mapping-reconciliation.test.ts
```diff
@@ -0,0 +1,561 @@
...
+
+const buildBlocks = (elements: { id: string; type: string }[]): TSurveyBlock[] =>
+  [{ id: "block-1", name: "Block 1", elements }] as unknown as TSurveyBlock[];
+
+const mapping = (elementId: string, hubFieldType: string, surveyId = SURVEY_ID) =>
+  ({ surveyId, elementId, hubFieldType }) as Parameters<typeof reconcileMappingsAgainstSurvey>[0][number];
+
+const mockTx = () => {
+  const tx = {
+    feedbackSource: { update: vi.fn() },
```
kiln-agent: **Same problem here — this test fails.** `mockTx()` stubs only `feedbackSource.update`, but `applyReconciliationToFeedbackSource` now calls `tx.feedbackSource.updateMany({ where: { id, workspaceId, status: "active" }, ... })`.

In `"flags the source as errored when the delete leaves it with no mappings at all"` (line 442) the count is mocked to `0`, so the implementation hits the missing `updateMany`; the resulting `TypeError` is swallowed by `applyReconciliationToFeedbackSource`'s own try/catch, and the assertion `expect(tx.feedbackSource.update).toHaveBeenCalledWith({ where: { id, workspaceId }, data: { status: "error" } })` fails.

Worth noting the collateral damage: the two sibling tests that assert `tx.feedbackSource.update` was **not** called (lines 469 and 486) now pass vacuously — they can no longer distinguish "did not flag" from "flagged via a method the mock doesn't have". `updateMany: vi.fn().mockResolvedValue({ count: 1 })` plus repointing the three assertions restores the coverage, including the new `status: "active"` scoping and the `flagged.count === 0` early return, which is currently untested.

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,361 @@
...
+    logger.warn(
+      { surveyId, mappingCount: surveyMappingCount, unmappableCount: unmappable.length },
+      "Keeping inert feedback-source mappings: deleting them would leave this survey with none"
+    );
+    return unmappable;
+  }
+
+  // Only leaking rows are left, so they all go and the source loses its handle on this survey. That is
+  // the lesser harm, but it needs a human, so say so loudly.
+  logger.error(
```
kiln-agent, line 133: **A source that lands here can't be repaired from the UI, so `error` is terminal.**

Scenario — the PR's own QA step, on a single-question source: the mapped question is retyped to Contact Info. `removed` is empty, `unmappable` is everything, so every row is deleted, `applyReconciliationToFeedbackSource` counts `0` remaining and flips the source to `error`.

The operator now opens **Edit source**, and `edit-feedback-source-modal.tsx` cannot recover it:

- `mappedSurveyId = feedbackSource.formbricksMappings[0]?.surveyId ?? ""` → empty, because there are no rows left
- the survey `<Select>` is hard-coded `disabled`, so they cannot pick the survey again
- `saveChangesDisabled` returns `true` while `!formbricksValues.surveyId`

So the error-clearing path added in `service.ts:491` is unreachable for exactly the state that produces the error, and `getFeedbackSourcesBySurveyId` filters `status: "active"` so nothing publishes. The only way out is delete + recreate the source (the pause/resume toggle flips it to `active`, which just hides the badge on a source that still has no mappings).

Note the asymmetry with the branch above: the hold-back keeps a row precisely so the edit modal can still prefill and repair. This branch removes that ability. Either keep one row here too, or make the survey selector usable when a formbricks source has no mappings.

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,361 @@
...
+          where: { feedbackSourceId, workspaceId },
+        });
+
+        if (remainingMappings === 0) {
+          // updateMany scoped to `active`: this write was unconditional, so it also overwrote
+          // `paused`. The resume toggle would then flip that source to `active` and silently resume
+          // something an operator had deliberately disabled. A paused source with no mappings is
+          // still paused — it is not publishing either way, and `error` is a claim about the source
+          // being broken rather than about it being off.
+          const flagged = await tx.feedbackSource.updateMany({
```
kiln-agent, line 291: Reconciliation sets `error` but never clears it, even when reconciliation itself has fixed the source.

Concrete case with `elementScope: "all"`: the survey's only question is retyped to File Upload → row deleted, source flagged `error`. The operator spots the mistake and retypes it back to Open Text, then saves. `collectUnmappedElements` re-creates the mapping row, so the data is correct again — but the status is still `error`, `getFeedbackSourcesBySurveyId` filters `status: "active"`, and nothing publishes. Nothing in the UI says a further step (re-saving the source in the edit modal) is needed, and for a single-survey source that re-save is itself blocked until the row exists.

Since you already count remaining mappings here, the symmetric write is cheap: when `toCreate.length > 0` and `remainingMappings > 0`, `updateMany({ where: { id, workspaceId, status: "error" }, data: { status: "active" } })` — scoped to `error` for the same reason the flag write is scoped to `active`, so a `paused` source stays paused.

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,361 @@
...
+
+  const wouldRemoveEveryMapping =
+    !hasPendingCreates && surveyMappingCount > 0 && removed.length + unmappable.length === surveyMappingCount;
+
+  if (!wouldRemoveEveryMapping) {
+    return [...removed, ...unmappable];
+  }
+
+  if (removed.length > 0) {
+    logger.warn(
```
kiln-agent: Minor, but this warns on a hot path for a condition that is permanent.

For a `specific`-scoped source whose only rows for a survey are orphaned, the delta is empty (`toDelete` is `unmappable` = `[]`, no updates, no creates), so `reconcileFeedbackSourcesForSurvey` filters it out and no DB work happens — yet `resolveDeletions` still emits this `warn` on every single survey write, including the editor's periodic draft autosave. Nothing in the code path can ever change the condition (a `specific` source never gains rows on its own), so it fires for as long as the drift exists.

Consider emitting it only from the branch that actually writes, or dropping it to debug.

---

kiln-agent (commented):
Careful pass over the whole diff. The reconciliation logic itself holds up well — the `surveyId` scoping, the persisted-blocks-vs-payload distinction, the ordering of the remaining-mappings count after the creates, and the no-overlap between `toCreate`/`toDelete`/`toUpdate` all check out, and the UI's selectable set (`getSelectableQuestionIds`) is an exact complement of `ELEMENT_TYPE_TO_HUB_FIELD_TYPE`, so `elementScope: "all"` is actually reachable from a default create.

Five findings, most severe first:

1. `apps/web/lib/feedback-source/service.ts:495` — the new `clearsErrorStatus` `updateMany` breaks two existing tests in `service.test.ts` (tx mock has no `updateMany`).
2. `apps/web/lib/feedback-source/mapping-reconciliation.test.ts:51` — same missing `updateMany` on `mockTx()`; the error-flag test fails and two sibling tests now pass vacuously.
3. `apps/web/lib/feedback-source/mapping-reconciliation.ts:129` — a source stripped of every mapping is flagged `error` and cannot be re-mapped through the edit modal.
4. `apps/web/lib/feedback-source/mapping-reconciliation.ts:273` — `error` is never cleared, even when reconciliation itself restores the mappings.
5. `apps/web/lib/feedback-source/mapping-reconciliation.ts:120` — `warn` fires on every survey save (autosave included) for a permanently held-back source.

(1) and (2) both trace to commit `79095eb` switching `update` → `updateMany` without updating the mocks — CI should be red on this branch.

---

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,361 @@
...
+ * await is correct anyway, so that path falls back to it.
+ */
+export const scheduleFeedbackSourceReconciliation = async (
+  surveyId: string,
+  blocks: TSurveyBlock[]
+): Promise<void> => {
+  try {
+    after(() => reconcileFeedbackSourcesForSurvey(surveyId, blocks));
+  } catch {
+    await reconcileFeedbackSourcesForSurvey(surveyId, blocks);
```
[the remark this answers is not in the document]
xernobyl: Confirmed, and this one was mine — introduced by the `after()` change in the previous commit. Fixed in `7c448d6cc`.

Before that commit the work was awaited inline, so it ran against blocks persisted moments earlier. Deferring it made the callback close over a snapshot captured at request time, and `after()` callbacks from concurrent requests are not serialized — so an older save's callback can run after a newer one and apply a stale snapshot, deleting mappings the newer save created or recreating ones it removed. The editor autosaves every 10 seconds, so that interleaving is routine rather than exotic. I traded request latency for a lost-update race and did not see it.

**Fix:** the deferred task re-reads the survey's persisted blocks instead of using the snapshot, so a late callback converges on current state — whichever runs last reconciles against what is actually stored. Two details worth naming:

- a survey deleted between the save and the callback is skipped, because diffing against nothing reads as "this survey has no questions" and would delete every mapping for it;
- the read is wrapped in its own try/catch, since Next swallows throws from an `after()` callback and the failure would otherwise vanish.

The inline fallback (scripts, jobs, tests, where `after()` throws) keeps using the caller's blocks: there is no interleaving to lose to there, and they are the freshest thing available.

**Regression test, as you asked** — two callbacks fired in reverse order. I first wrote it asserting that `findUnique` had been called, then rewrote it, because that asserts the implementation rather than the bug. It now asserts the behaviour: the stale callback must not delete the mapping the newer save created. Reverting the fix reproduces exactly the data loss you described:

```
AssertionError: expected [ 'el-new' ] to not include 'el-new'
```

All three new cases (reverse order, deleted survey, failing read) go red on that mutation and green with the fix.

apps/web/playwright/feedback-source-reconciliation.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    await expect
+      .poll(() => mappedElementIds(curatedRetypedSourceId, surveyId), { timeout: 15000 })
+      .toEqual([supported[0].id]);
+
+    // ...and it was retyped rather than left stale.
+    const retyped = await prisma.feedbackSourceFormbricksMapping.findFirstOrThrow({
+      where: { feedbackSourceId: curatedRetypedSourceId, surveyId, elementId: supported[0].id },
+      select: { hubFieldType: true },
+    });
+    expect(retyped.hubFieldType).not.toBe("nps");
```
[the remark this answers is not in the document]
xernobyl: Valid — fixed in `7c448d6cc`.

`not.toBe("nps")` passed for any value other than `"nps"`, so a reconciler writing the wrong Hub field type would have gone unnoticed. That is the same weak-assertion shape this PR spent several rounds removing, so thank you for catching it in my own fix.

The retype target is now pinned to an element of a known type (`openText`) rather than `supported[0]`, and the assertion is the literal it maps to (`"text"`). The fixture also asserts that such an element exists, so the test fails loudly if the seeded survey ever stops containing one instead of silently skipping.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Fixed in ea0222d5b. You were both right that this dead-ends, and the modal was the reason.

`mappedSurveyId` came from `formbricksMappings[0]?.surveyId ?? ""` and the survey `Select` was hardcoded `disabled`, so a zero-row source showed an empty **and** locked picker — `error` was unrecoverable.

The picker is now enabled and lists every survey exactly when the source has no mapping rows. That is the only state that needs repairing, and it is also the only state where enabling it is safe: there are no published FeedbackRecords to orphan. A source that still has rows keeps the lock.

On the invisibility half: the hold-back branch no longer logs at warn (it is a permanent condition re-evaluated on every survey write), and reconciliation now clears `error` itself once it re-creates a row, so the badge tracks reality rather than only ever accumulating.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Fixed in ea0222d5b — and your walkthrough was exactly the reproduction.

The survey picker is now enabled, listing all surveys, when the source has no mapping rows. Nothing is orphanable in that state, so the lock has nothing to protect, and it is precisely the state reconciliation leaves behind after the retype-to-Contact-Info path.

`error` is also no longer one-way: reconciliation clears it when it re-creates a mapping, so if the operator retypes the question back the source recovers without opening the modal at all.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Fixed in ea0222d5b. Reconciliation now clears the flag when it creates mappings:

```ts
if (toCreate.length > 0) {
  await tx.feedbackSource.updateMany({
    where: { id: feedbackSourceId, workspaceId, status: "error" },
    data: { status: "active" },
  });
}
```

Scoped to `error` for the same reason the flag write is — a deliberately `paused` source stays paused; a survey edit must not silently resume something an operator turned off. Three tests cover it, including that a paused source is untouched, and removing the block fails two of them.

apps/web/lib/feedback-source/service.ts
xernobyl: Good catch, fixed in ea0222d5b. It now requires the mappings to be formbricks ones:

```ts
const clearsErrorStatus =
  mappingsInput?.type === "formbricks_survey" &&
  mappingsInput.mappings.length > 0 &&
  data.status === undefined;
```

Your reasoning is the reason: `status: "error"` is written by exactly one thing, the formbricks mapping reconciler, so a save carrying *field* mappings cannot be repairing an error it could have caused — and the action takes `fieldMappings` regardless of source type, so the unscoped version could un-error a formbricks source that was still broken. `"replaces field mappings when provided"` now asserts `updateMany` is never called, and it fails if I revert the scoping.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Agreed, fixed in ea0222d5b — moved to `logger.debug`.

Your analysis holds: for a `specific` source whose only rows are orphaned the delta is empty, the caller filters it out and does no DB work, yet the warn fired on every survey write including saves that touch nothing. At warn it reported a standing condition forever and buried the one-off `error` next to it.

apps/web/lib/feedback-source/mapping-reconciliation.ts
xernobyl: Fixed in 7c448d6cc, before this reply — worth recording why, since it was a real ordering hazard and not just an outdated line.

The deferred callback no longer trusts the captured `blocks`. It re-reads them inside `after()`:

```ts
const persisted = await readPersistedSurveyBlocks(surveyId);
await reconcileFeedbackSourcesForSurvey(surveyId, persisted);
```

So a reconcile that runs late reconciles against whatever is actually stored at that moment, and cannot delete a mapping a newer save just created. The `blocks` argument is now only used on the non-deferred path, where it is synchronous with the request that produced it.

apps/web/lib/feedback-source/mapping-reconciliation.test.ts
xernobyl: Already fixed in 7c448d6cc — both transaction mocks now stub `updateMany`:

- `mockTx()` in `mapping-reconciliation.test.ts`: `feedbackSource: { update: vi.fn(), updateMany: vi.fn().mockResolvedValue({ count: 1 }) }`
- `setupTransaction()` in `service.test.ts`: `updateMany: vi.fn().mockResolvedValue({ count: 0 })`, with a comment on why the error→active reset is its own `updateMany` rather than a `status` filter on the main update (a filter there would P2025 on every healthy save).

Verified locally on ea0222d5b: **713 tests pass** across `lib/feedback-source/`, `lib/survey/` and `app/api/v3/surveys/`. The diagnosis was right for the commit it was written against.

apps/web/lib/feedback-source/service.ts
xernobyl: Already fixed in 7c448d6cc — both transaction mocks now stub `updateMany`:

- `mockTx()` in `mapping-reconciliation.test.ts`: `feedbackSource: { update: vi.fn(), updateMany: vi.fn().mockResolvedValue({ count: 1 }) }`
- `setupTransaction()` in `service.test.ts`: `updateMany: vi.fn().mockResolvedValue({ count: 0 })`, with a comment on why the error→active reset is its own `updateMany` rather than a `status` filter on the main update (a filter there would P2025 on every healthy save).

Verified locally on ea0222d5b: **713 tests pass** across `lib/feedback-source/`, `lib/survey/` and `app/api/v3/surveys/`. The diagnosis was right for the commit it was written against.

xernobyl: Already fixed in 7c448d6cc — both transaction mocks now stub `updateMany`:

- `mockTx()` in `mapping-reconciliation.test.ts`: `feedbackSource: { update: vi.fn(), updateMany: vi.fn().mockResolvedValue({ count: 1 }) }`
- `setupTransaction()` in `service.test.ts`: `updateMany: vi.fn().mockResolvedValue({ count: 0 })`, with a comment on why the error→active reset is its own `updateMany` rather than a `status` filter on the main update (a filter there would P2025 on every healthy save).

Verified locally on ea0222d5b: **713 tests pass** across `lib/feedback-source/`, `lib/survey/` and `app/api/v3/surveys/`. The diagnosis was right for the commit it was written against.

apps/web/lib/feedback-source/mapping-reconciliation.test.ts
xernobyl: Already fixed in 7c448d6cc — both transaction mocks now stub `updateMany`:

- `mockTx()` in `mapping-reconciliation.test.ts`: `feedbackSource: { update: vi.fn(), updateMany: vi.fn().mockResolvedValue({ count: 1 }) }`
- `setupTransaction()` in `service.test.ts`: `updateMany: vi.fn().mockResolvedValue({ count: 0 })`, with a comment on why the error→active reset is its own `updateMany` rather than a `status` filter on the main update (a filter there would P2025 on every healthy save).

Verified locally on ea0222d5b: **713 tests pass** across `lib/feedback-source/`, `lib/survey/` and `app/api/v3/surveys/`. The diagnosis was right for the commit it was written against.

apps/web/lib/feedback-source/mapping-reconciliation.ts
```diff
@@ -0,0 +1,400 @@
...
+ * Deliberately a fresh read rather than anything cached: its whole purpose is to observe writes that
+ * landed after the caller captured its own snapshot.
+ */
+const readPersistedSurveyBlocks = async (surveyId: string): Promise<TSurveyBlock[] | null> => {
+  const survey = await prisma.survey.findUnique({
+    where: { id: surveyId },
+    select: { blocks: true },
+  });
+
+  return survey ? (survey.blocks as unknown as TSurveyBlock[]) : null;
```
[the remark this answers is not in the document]
xernobyl: Done in 4d6b5761e. The read now goes through Survey's composite `@@unique([id, workspaceId])`, with the workspaceId threaded from both call sites (they already hold it as the tenant anchor):

```ts
where: { id: surveyId, workspaceId },
```

One correction to the severity, for the record: this was defense in depth rather than an open door. The `surveyId` is the survey the caller just persisted and was authorized on, `blocks` never reaches a response, and every mapping write is FK-bound to `(surveyId, workspaceId)` — I verified at the DB level earlier that a cross-workspace mapping row is rejected outright. So there was no reachable cross-tenant path.

That said, the point stands on its merits: the deferred read runs outside the request that authorized it, so the constraint belongs in the query rather than resting on FKs in another file. There is now a test pinning the where clause, and it fails if the scoping is removed.

Threading the argument also surfaced something the finding did not: four calls in `mapping-reconciliation.test.ts` were still on the old two-argument form, silently passing `blocks` into the `workspaceId` slot. Nothing would have caught it — vitest does not typecheck, and CI's typecheck project excludes test files.

apps/web/playwright/feedback-source-reconciliation.spec.ts
```diff
@@ -0,0 +1,249 @@
...
+      select: { hubFieldType: true },
+    });
+    // The exact type openText maps to, not merely "something other than nps" — that weaker form
+    // passes for any wrong value reconciliation might write.
+    expect(retyped.hubFieldType).toBe(RETYPE_EXPECTED_HUB_FIELD_TYPE);
+
+    // Neither curated source adopted the questions it was never mapped to.
+    expect(await mappedElementIds(curatedRetypedSourceId, surveyId)).not.toContain(
+      supported.find((element) => element.id !== retypeTarget!.id)?.id
+    );
```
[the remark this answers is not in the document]
xernobyl: Correct, and removed in 4d6b5761e rather than guarded.

Your first suggestion was the better one: the `.poll(...).toEqual([retypeTarget!.id])` above is an exact-set equality, so no other element can be present and the extra `not.toContain` added nothing even in the cases where it was not vacuous. Adding a `toBeDefined()` guard would have kept a redundant assertion alive.

I moved the intent into the surviving assertion's comment so the "adopted nothing new" half is still stated where it is actually checked.

apps/web/lib/feedback-source/mapping-reconciliation.test.ts
```diff
@@ -0,0 +1,719 @@
...
+
+    await applyReconciliationToFeedbackSource(SOURCE_ID, WORKSPACE_ID, SURVEY_ID, {
+      toCreate: [{ elementId: "el-recreated", hubFieldType: "rating" }],
+      toDelete: [],
+      toUpdate: [],
+    });
+
+    const [[args]] = tx.feedbackSource.updateMany.mock.calls as any;
+    expect(args.where.status).toBe("error");
+  });
```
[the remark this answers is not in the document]
xernobyl: Fixed in 506680f7c, though not by the suggested diff — asserting the full `updateMany` input there would have made this test a duplicate of the one directly above it, which already asserts exactly that object.

The `where.status` peek was a strict subset of that assertion plus an `any`, so I removed the test and folded its intent into the other one's comment: the `status: "error"` filter in the where is load-bearing beyond the happy path, because it is what stops a survey edit silently resuming a source an operator had paused.

apps/web/lib/feedback-source/service.test.ts
```diff
@@ -783,6 +888,11 @@ describe("updateFeedbackSourceWithMappings", () => {
       where: { feedbackSourceId: FEEDBACK_SOURCE_ID, workspaceId: ENV_ID },
     });
     expect(tx.feedbackSourceFieldMapping.create).toHaveBeenCalledTimes(1);
+    // `status: "error"` is written by exactly one thing, the formbricks mapping reconciler, so saving
+    // FIELD mappings cannot be repairing an error it could have caused. The update action accepts
+    // fieldMappings regardless of source type, so an unscoped clear would un-error a formbricks source
+    // that is still broken.
+    expect(tx.feedbackSource.updateMany).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
xernobyl: Agreed, fixed in 506680f7c.

`setupTransaction` now tracks the source's status across both `update` and `updateMany`, honouring the `where` filter the way Postgres would (a scoped write only lands if the current status matches). The test starts from an errored source and asserts it is *still* errored after saving field mappings:

```ts
const tx = setupTransaction("error");
...
expect(persistedStatus()).toBe("error");
```

So it survives the clear moving between `update` and `updateMany`. Reverting the type scoping now fails with `Expected: "error" / Received: "active"` rather than a call count.

BhagyaAmarasinghe (approved):

PR #8828, opened 2026-08-12
fix(deps): remediate container image vulnerabilities

BhagyaAmarasinghe:
## What & why

The current Formbricks ECR image was reported with vulnerabilities in `undici`, `ip-address`,
`postcss`, `dompurify`, `sanitize-html`, and `nanoid`. Upgrade the direct dependencies and add
load-bearing workspace overrides so every transitive path resolves to a patched release. The Better
Auth OAuth provider upgrade remains deferred because its release-candidate migration is coupled; add
a regression test for the single-audience restriction that mitigates the reported token-confusion
scenario in our deployment.

## Linear ticket

No Linear ticket — this remediation comes from the Oneleet/ECR image vulnerability alert.

## How this was tested

- `pnpm install --frozen-lockfile` ✅
- `pnpm audit --json` ✅ all alert-targeted package findings cleared except the documented Better Auth
  advisory; unrelated repository audit findings remain
- `pnpm test` ✅ 25 tasks, including 598 web test files / 7,781 web assertions
- Focused survey UI and survey bundle suites ✅ 84 + 667 assertions
- `pnpm lint` ✅ zero errors (existing warnings only)
- `pnpm typecheck --filter=@formbricks/web --filter=@formbricks/surveys --filter=@formbricks/survey-ui --filter=@formbricks/vite-plugins` ✅
- `pnpm format:check` ✅
- `pnpm build --filter=@formbricks/surveys... --force` ✅
- Production ARM64 Docker image build ✅ `sha256:b271ce10b2a2b1689922784916a4dfda3beddf3c6c8c931b1401d3d961185d4e`
- Trivy image scan ✅ zero OS or Node package vulnerabilities in the built image
- Runtime image inspection ✅ no global npm installation or `npm`/`npx` commands

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Sign in to staging and open an existing project → authentication completes and the project loads without an error.
- [ ] Publish a survey with formatted text, open its public link, and submit a response → content renders correctly and the response completes.
- [ ] Configure a follow-up email containing bold text and a link, then trigger it with a response → the email arrives with the allowed formatting intact and no unexpected HTML rendered.
- [ ] Configure a survey webhook and submit a response → the endpoint receives one valid delivery and Formbricks records it as successful.
- [ ] With an MCP client registered against staging, authorize and call an MCP tool → authorization succeeds for the MCP resource and the tool call completes.

**Preconditions / test data**

- A staging account with project and survey management access.
[...]

PR #8829, opened 2026-08-12
fix(security): bind MCP access-token audience to this resource server

xernobyl:
## What & why

`@better-auth/oauth-provider` does not bind an access token's audience to the resource approved during authorization ([GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528), CVSS 6.4). The token endpoint stamps `aud` with the **client-supplied** `resource`, gated only by the static `validAudiences` allow-list — the authorization code stores no resource, so nothing ties the token request back to what the user consented to.

This deployment is not exploitable today only because `validAudiences` has exactly one entry, and **nothing asserted that**. It was one config edit away from being real, and the suite would have stayed green.

The upstream fix ships only in the unreleased 1.7 line (still no stable — npm `latest` is `1.6.27`, itself in the affected range), and that bump is coupled to `better-auth` core, touching login/sessions/2FA/SSO. That upgrade is tracked separately in ENG-2343, gated on a stable release. This PR closes the gap on our side with work that is correct on either version.

**The MCP resource server now asserts the verified payload's audience itself.** `verifyOptions.audience` cannot do this: it is handed to jose, whose `aud` check is a *membership* test, so a token naming our resource **and** a foreign one passes it. RFC 9068 §4 puts that duty on the resource server, so this assertion is required no matter which provider version issued the token — it is not scaffolding to be removed after the upgrade. (1.7 in fact drops `verifyOptions.audience` from its own verification, which makes it *more* load-bearing.)

The check is an **allow-list** of `{MCP resource, the AS's UserInfo endpoint}` rather than "exactly one audience". The provider appends UserInfo to `aud` whenever `openid` is in the *granted* scopes — current 1.6.23 behaviour, not something the pending upgrade introduces — so a single-value rule would reject those tokens outright.

### Commits

1. `fix(security)` — the audience assertion, the single-entry config assertion, and real-JWT audience tests.
2. `test` — closes two coverage gaps SonarCloud surfaced (see below).
3. `refactor` — extracts the six-times-duplicated "rate-limit, log, 401 + challenge" block that adding a fourth rejection pushed over Sonar's cognitive-complexity limit.
4. `docs` — records a latent assumption found while reviewing, and pins the logged `statusCode`.
5. `docs` — corrects the checked-in accepted-risk note in `pnpm-workspace.yaml`, which this branch made stale.

Rebased onto `main` after #8828 landed (it touches the same note). Worth knowing for review: **#8828 independently added the same single-entry `validAudiences` assertion**, in `mcp-oauth-dcr.test.ts`. Both are kept, and the duplication is now called out in a comment — it is the invariant this whole acceptance rests on, and the two suites can be edited independently.

## Linear ticket

Ref ENG-2339

## How this was tested

- `pnpm test` (full `apps/web` suite) ✅ — **599 files / 7823 tests passed**.
- Targeted: `modules/mcp/`, `modules/auth/lib/`, `app/api/mcp/`, `app/.well-known/` ✅ — 32 files / 376 tests.
- `eslint` ✅ clean, `prettier --check` ✅ clean, `tsc --noEmit -p tsconfig.typecheck.json` ✅ — zero errors in touched files.
- **Fault injection, run twice** (before and after the refactor, since it moved the guard): with the assertion short-circuited, exactly the three audience tests fail and the rest of the suite stays green. The multi-audience case fails with `expected true to be false` — i.e. the vulnerable token was *accepted*. Guard restored, suite re-run green.

Tests added:

- `modules/auth/lib/mcp-oauth-provider-options.test.ts` (new) — the single-entry audience invariant, plus advertised-vs-registration scope agreement and the derived write-scope step-up expiry.
- `modules/mcp/auth.test.ts` — a `MCP OAuth access token audience binding` block that signs **real** ES256 tokens and verifies them through jose with the production `verifyOptions`. Covers `aud` as a string, `[MCP]`, `[MCP, UserInfo]` (accepted), and `[MCP, foreign]`, foreign-only, absent (rejected). The multi-audience case also asserts jose *itself* accepts the token, so the test documents why the library layer is not sufficient.
- Two further cases drive the check through the stubbed verifier, exercising it with no jose `aud` enforcement upstream — the configuration 1.7 creates.
- `modules/auth/lib/oauth-urls.test.ts` — pins the UserInfo audience derivation, including a subpath `BETTER_AUTH_URL`.

Existing fixtures gained the `aud` claim every real token carries — without it they described a token the authorization server cannot issue, which is how this class of bug stayed invisible.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Connect an MCP client (Claude Code, MCP Inspector) to `/api/mcp` and complete the OAuth flow → connects as before, tools list and calls work. This is the regression that matters: the audience rule must not turn away legitimate clients.
- [ ] Repeat with a client that requests `openid` → still connects (this is the two-audience shape).
- [ ] `POST /api/mcp` with `Authorization: Bearer <garbage>` → `401` with a `WWW-Authenticate: Bearer resource_metadata="…", scope="…"` challenge, unchanged.
- [ ] `POST /api/mcp` with an API key (`x-api-key`) → unaffected.
- [ ] `POST /api/mcp` with no credentials at all → `401` with the same challenge, and the message still reads `API key or OAuth access token required` (the refusal helper is shared now, so this wording is worth eyeballing).
- [ ] Sign in normally, exercise 2FA and SSO → unchanged. No auth-library version moved in this PR.

**Preconditions / test data**

- A workspace with the MCP surface reachable and an MCP client able to complete Dynamic Client Registration.
- No feature flag, no new env var, no plan/entitlement gate.

**Risks & regressions**

- Blast radius is the OAuth-bearer path of `/api/mcp`. If the allow-list were wrong the symptom would be legitimate MCP clients getting `401` instead of connecting — hence the first two steps.
- The refactor touches the shared refusal path for **all** MCP auth failures, including the two API-key ones, so the "no credentials" and "bad API key" responses are worth a look even though their behaviour is unchanged.
- API-key MCP auth, the v3 REST API and browser session auth are otherwise untouched.
- A rejected token is logged distinctly (`audience is not bound to this resource server`) but the client receives the same opaque `Invalid OAuth access token`, so this adds no oracle.

- none

### DIFF
--- a/apps/web/app/api/mcp/route.test.ts
+++ b/apps/web/app/api/mcp/route.test.ts
@@ -58,6 +58,10 @@ const { MCP_CHALLENGE_SCOPE } = await import("@/modules/auth/lib/oauth-urls");
 // The auth-params are comma-separated per RFC 9110 §11.6.1 (#8718): asserting the whole string is what
 // keeps the separator from regressing, since a strict client parser needs it to read `resource_metadata`.
 // The scope list interpolates the real constant, so this stays honest as the advertised scopes change.
+// Real tokens always carry `aud`; the resource server rejects any token not minted for it, so the
+// fixtures have to look like something the authorization server would actually issue.
+const MCP_AUDIENCE = "http://localhost/api/mcp";
+
 const EXPECTED_CHALLENGE = `Bearer resource_metadata="http://localhost/.well-known/oauth-protected-resource/api/mcp", scope="${MCP_CHALLENGE_SCOPE}"`;
 
 vi.mock("@/modules/api/lib/api-key-auth", () => ({
@@ -144,6 +148,7 @@ describe("POST /api/mcp", () => {
     vi.mocked(applyIPRateLimit).mockResolvedValue({ allowed: true });
     userFindUniqueMock.mockResolvedValue({ isActive: true });
     verifyAccessTokenMock.mockResolvedValue({
+      aud: MCP_AUDIENCE,
       sub: "user_1",
       email: "person@example.com",
       name: "Person",
@@ -449,6 +454,7 @@ describe("POST /api/mcp", () => {
 
   test("blocks write tools for read-only OAuth tokens", async () => {
     verifyAccessTokenMock.mockResolvedValueOnce({
+      aud: MCP_AUDIENCE,
       sub: "user_1",
       email: "person@example.com",
       scope: "openid profile email surveys:read",
@@ -494,6 +500,7 @@ describe("POST /api/mcp", () => {
     // A write-capable user whose OAuth token was only granted read scopes (surveys:read + workflows:read)
     // must not be able to reach a workflow mutation — the ENG-1967 token-scope boundary.
     verifyAccessTokenMock.mockResolvedValueOnce({
+      aud: MCP_AUDIENCE,
       sub: "user_1",
       email: "person@example.com",
       scope: "openid profile email surveys:read workflows:read",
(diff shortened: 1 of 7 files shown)


BhagyaAmarasinghe (approved):
Thank you. LGTM

PR #8830, opened 2026-08-12
fix(unify-feedback): render feedback record value text as a textarea

jobenjada:
## What & why

The **Value (Text)** field in the Feedback Record drawer rendered as a single-line `Input` in a half-width grid cell. Open-text answers are the longest values on a record, so anything past ~40 characters was clipped and only readable by scrolling inside the input — the common case for verbatim feedback.

`Value (Text)` is now a full-width, 5-row `Textarea` (vertically resizable, still read-only/disabled like every other field in this drawer), so long verbatims wrap and are readable at a glance. `Value ID` moves out of the shared two-column grid onto its own full-width row, matching the other single-field rows (`Source type`, `Field group label`).

No behavior change: the field stays disabled, the value binding is unchanged, and the translation / AI-enrichment panels below it are untouched.

## Linear ticket

No ticket — drive-by UI readability fix on the Unify Feedback record drawer, reported directly while reviewing a Swedish verbatim that was clipped in the field.

## How this was tested

- `vitest run modules/ee/unify-feedback` → **226 tests passed**, 19/20 files passed. The one failing file (`lib/contacts.test.ts`) fails to resolve `@formbricks/database` in this worktree — a pre-existing environment issue (no generated Prisma client), unrelated to this diff. Per AGENTS.md, `.tsx` components are not unit-tested.
- ESLint + Prettier via the pre-commit lint-staged run ✅
- Verified visually with a component render (Playwright, using this repo's compiled Tailwind CSS and the real `Input`/`Textarea`/`Label` classes). **These are component renders, not live app captures** — the web app can't be stood up in this worktree (no DB / Hub backend), so the surrounding drawer chrome is omitted and the rows that changed are shown in isolation.

Left: single-line input clipping a Swedish verbatim (the reported bug). Right: full-width textarea, whole answer readable.

Value (Text) before and after

<sub>Screenshot lives on the orphan branch `assets-value-text-textarea` (PNG only, not part of this diff) — safe to delete after merge.</sub>

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a workspace with Unify Feedback enabled → **Feedback** → **Records**, click any record row → the details drawer opens from the right.
- [ ] Pick a record whose answer is a long open text (2+ sentences) → **Value (Text)** spans the full drawer width as a multi-line box, the whole answer is visible/wrapped, and no text is cut off horizontally.
- [ ] Drag the bottom-right corner of the **Value (Text)** box → it resizes vertically; the rest of the drawer reflows without overlap.
- [ ] Confirm **Value (Text)** is still read-only: clicking into it does not allow typing.
- [ ] **Value ID** now sits on its own row directly below **Value (Text)**, full width, read-only (empty when the record has no value ID).
- [ ] Open a record with a non-text answer (number / date / rating / boolean) → **Value (Text)** is empty, and the `Translated text` and `AI enrichment` panels behave exactly as before (they only show for text records).
- [ ] Open a record with a translation → the `Translated text` panel still renders below **Value (Text)**.

**Preconditions / test data**

- Workspace with the Unify Feedback (Hub) surface available, plus at least one feedback record with a long free-text value — ideally one non-English so the `Translated text` panel also shows.
- Owner/manager role to see the Delete action in the drawer footer (unrelated to this change, but the drawer footer is in the same view).

**Risks & regressions**

- Drawer layout only: the two-column grid that previously held `Value (Text)` + `Value ID` was split into two full-width rows. Re-check the drawer's vertical rhythm and that the following rows (`Value (Number)` / `Value (Date)`, `Value (Boolean)`, `Language` / `User identifier`, `Metadata`) are unchanged.
- Narrow viewports: verify the drawer on a small screen — the grid rows were already stacking, so only the textarea height should differ.

- none

### DIFF
--- a/apps/web/modules/ee/unify-feedback/components/feedback-record-form-drawer.tsx
+++ b/apps/web/modules/ee/unify-feedback/components/feedback-record-form-drawer.tsx
@@ -34,6 +34,7 @@ import {
   SheetTitle,
 } from "@/modules/ui/components/sheet";
 import { Switch } from "@/modules/ui/components/switch";
+import { Textarea } from "@/modules/ui/components/textarea";
 import { deleteFeedbackRecordAction, retrieveFeedbackRecordAction } from "../actions";
 import { FIELD_TYPE_OPTIONS, type TFeedbackRecordFormValues } from "../lib/types";
 import {
@@ -379,29 +380,35 @@ export const FeedbackRecordFormDrawer = ({
                   )}
                 />
 
-                <div className="grid grid-cols-2 gap-3">
-                  <FormField
-                    control={form.control}
-                    name="value_text"
-                    render={({ field }) => (
-                      <FormItem>
-                        <FormLabel>{t("workspace.unify.value_text")}</FormLabel>
-                        <FormControl>
-                          <Input {...field} value={field.value ?? ""} disabled />
-                        </FormControl>
-                      </FormItem>
-                    )}
-                  />
+                {/* Full width, multi-line: open-text answers are the longest values on the record and
+                    have to stay readable without horizontal scrolling. */}
+                <FormField
+                  control={form.control}
+                  name="value_text"
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel>{t("workspace.unify.value_text")}</FormLabel>
+                      <FormControl>
+                        <Textarea
+                          {...field}
+                          value={field.value ?? ""}
+                          rows={5}
+                          className="resize-y"
+                          disabled
+                        />
+                      </FormControl>
+                    </FormItem>
+                  )}
+                />
 
-                  {/* ENG-1673: read-only — the stable option identity is managed by ingestion/backfill;
-                      hand-editing it would desync the record from its source option. Always shown so the
-                      canonical Value ID is visible even when the record has none. */}
-                  <div className="space-y-2">
-                    <Label htmlFor="value_id" className="block text-slate-800">
-                      {t("workspace.unify.value_id")}
-                    </Label>
-                    <Input id="value_id" value={record.value_id ?? ""} disabled readOnly />
-                  </div>
+                {/* ENG-1673: read-only — the stable option identity is managed by ingestion/backfill;
+                    hand-editing it would desync the record from its source option. Always shown so the
+                    canonical Value ID is visible even when the record has none. */}
+                <div className="space-y-2">
+                  <Label htmlFor="value_id" className="block text-slate-800">
+                    {t("workspace.unify.value_id")}
+                  </Label>
+                  <Input id="value_id" value={record.value_id ?? ""} disabled readOnly />
                 </div>
 
                 {translatedText && (


apps/web/modules/ee/unify-feedback/components/feedback-record-form-drawer.tsx
```diff
@@ -379,29 +380,35 @@ export const FeedbackRecordFormDrawer = ({
...
+                    <FormItem>
+                      <FormLabel>{t("workspace.unify.value_text")}</FormLabel>
+                      <FormControl>
+                        <Textarea
+                          {...field}
+                          value={field.value ?? ""}
+                          rows={5}
+                          className="resize-y"
+                          disabled
+                        />
```
itsjavi, line 398: `disabled` pulls against the goal of this change.

The base `Textarea` carries `disabled:opacity-50`, so the verbatim this PR exists to make readable is painted at half contrast. Beyond contrast, a disabled control is removed from the tab order, is not announced by screen readers, cannot be selected/copied in Chrome or Firefox, and cannot be scrolled by keyboard — which matters here precisely because an answer longer than the 5 rows still needs scrolling (or a resize drag, which is also mouse-only).

`readOnly` buys everything `disabled` was doing on this field — the value can't be edited, and RHF's `onChange` never fires — while restoring contrast, focus, selection and screen-reader access:

```suggestion
                        <Textarea
                          {...field}
                          value={field.value ?? ""}
                          rows={5}
                          className="resize-y"
                          readOnly
                        />
```

Worth noting the `Translated text` panel ~25 lines below already solves the same readability problem the other way: a plain `<p className="whitespace-pre-wrap">` at full contrast. Keeping every control in this drawer visually `disabled` for consistency is a defensible call — but then the original verbatim ends up as the one value on the record a support user can't copy, while the machine translation of it they can.

---

itsjavi (commented):
Reviewed the diff plus the surrounding drawer, the shared `Textarea`/`Input` primitives and `cn`'s tailwind-merge behaviour.

**What holds up**

- `resize-y` does override the primitive's `resize-none`, because `cn` is `twMerge(clsx(...))` — the override the comment claims is real, and the `Textarea` docblock explicitly sanctions overriding via `className`.
- Splitting `Value ID` out of the two-column grid onto its own full-width row matches the rows above it (`Field group label`) and the ones below, so the drawer's rhythm is preserved rather than disturbed.
- `rows={5}` and the primitive's `min-h-20` don't fight: the taller intrinsic height wins.
- No new i18n keys, no behaviour change to the translation / AI-enrichment panels, no new props needing `Readonly<>`.

**One point to address** — see the inline comment: `disabled` gives the field half contrast, no focus, no text selection and no keyboard scrolling, which is at odds with a change whose whole purpose is making long verbatims readable. `readOnly` is a one-word swap that keeps the field non-editable and fixes all four.

**Checks at time of review**

- All 19 checks green, SonarQube quality gate passed (0 new issues, 0 hotspots).
- CodeRabbit **skipped this PR entirely** — the only changed file was excluded by its path filters — so nothing automated has looked at this diff apart from lint/Sonar. Worth keeping in mind when weighing how much the green checks say here.
- No conflict with `main`; the PR is only blocked on approvals.

Not approving — leaving the call to a human reviewer.

<sub>Reviewed with Claude Code (model `claude-opus-5`, reasoning effort: max).</sub>

---

itsjavi:
**Needs human review — @itsjavi**

Things I could not check from this environment, all of them the parts that actually decide whether this fix lands well:

- **The rendered drawer.** The screenshots in the description are isolated component renders, not the real drawer, and the author says as much. Somebody should open a real Feedback record and look at the drawer end to end: the vertical rhythm after the two-column grid was split, the 5-row box against a short one-word answer (a lot of empty space for `pro` or `4`), and behaviour on a narrow viewport.
- **Long-answer behaviour.** How a verbatim that overflows 5 rows behaves in practice — whether the resize handle is discoverable, and whether scrolling inside the box works for a keyboard user (see the inline comment about `disabled`).
- **Unify Feedback is EE / Hub-backed**, so I can't reach the surface or the data from here; the QA checklist in the description needs a workspace with the feature enabled.

The code-level review is in the submitted review above.

---

itsjavi (approved):
thanks LGTM!

Issue #40, opened 2022-08-10
Reset password


matthiasnannt:
### Password Reset

As a user I want to reset my password when I forgot it.

What needs to be done:
- build a new "forgot password" view where a user can enter his email address and request a new password.

PR #8832, opened 2026-08-12
feat(embedded-data): backfill existing variables and hidden fields into the tables [ENG-1835]

pandeymangg:
## What & why

ENG-1833 added the tables, ENG-1978 keeps them current when someone edits a survey, ENG-1836 built the read seam. What's missing is the surveys nobody has edited — the vast majority. This migrates them, which is what lets ENG-1837 point readers at the tables at all.

Every survey's legacy `variables` and `hiddenFields.fieldIds` become local `EmbeddedData` rows plus `SurveyEmbeddedData` links, per migration spec §8:

| Legacy | Row | Link's `storageKey` |
| --- | --- | --- |
| variable (number) | `source=computed, dataType=number, defaultValue=<value>, name=<variable name>` | the variable's **existing cuid** |
| variable (text) | same with `dataType=string` | the variable's **existing cuid** |
| hidden field | `source=ingested, dataType=string, defaultValue=null, name=<field id>` | the field's **existing name** |

`key` and `locked` fall to their column defaults — a migrated field is local (no library key) and unlocked (nothing could refuse an outside write before this).

**The mapping itself is not restated here.** `toDesiredEmbeddedFields` already holds it, shared with the write bridge and the read seam, so the two invariants that make this safe — a variable keeps its cuid, a hidden field keeps its name — are stated once. Those are the addresses recall tokens, logic operands and stored responses already use, which is why every existing survey keeps resolving with zero rewrites.

**No `Response` row is read or written.** A response is already keyed by the same `storageKey` a definition moves under. That is the whole reason this migration can exist without touching the largest, busiest table in the database.

### Two things the bulk case needs that a single survey save does not

The runner wraps every data migration in **one transaction**, so anything that throws rolls back the entire backfill — not just the offending survey. Two guards follow from that:

- **A survey declaring the same address twice is skipped and logged, not inserted.** `@@unique([surveyId, storageKey])` would reject it and take every other survey down with it. Reachable through the legacy API rather than the editor, which prevents duplicate names. `toDesiredEmbeddedFields` passes duplicates through rather than merging them precisely so this decision belongs to the caller.
- **Candidate selection runs in SQL behind a `jsonb_typeof` guard.** `jsonb_array_length` raises on a non-array, so one malformed row would abort everything.

### Idempotency and pagination

A survey that already has links is never a candidate, so re-running does nothing. That also covers surveys the ENG-1978 write bridge migrated on its own when someone edited them — in local development that is the normal state, since the bridge merged first.

Keyset pagination on `id`, not `OFFSET`: inserting links removes a survey from the candidate set, so an offset would step over unprocessed rows. Walking forward past the last id also means a skipped survey is passed once rather than re-read forever.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1835/backfill-migrate-existing-variables-hidden-fields-into-the

Milestone 1 of the [Embedded Data](https://linear.app/formbricks/project/embedded-data-3d184a43d2f9) project. Targets `epic/embedded-data-v1` — all of v1 (M1–M4) ships as one release, so this does not reach production until the epic merges. Unblocks ENG-1837.

## How this was tested

**Unit — 7 new** in `utils.test.ts`: the two storage-key invariants, row/link cross-references scoped to the survey's workspace, a variable's declared type and value carried onto the row, and the three collision shapes (same hidden field twice, a hidden field named like a variable's cuid, each colliding key reported once).

**Integration — 8 new** against real Postgres in `apps/web/integration/embedded-data-backfill.integration.test.ts`:

| Case | Asserted |
| --- | --- |
| Survey with a variable and a hidden field | rows + links match, both addresses preserved |
| Second run | no-op, no duplicates |
| Survey the write bridge already migrated | untouched |
| Survey with nothing to migrate | ignored |
| A duplicated survey and its original | both migrate; sharing a variable cuid is fine, only `(surveyId, storageKey)` is unique |
| Surveys in two workspaces | each row lands in its own workspace |
| Survey declaring the same address twice | skipped and reported; the other survey still migrates |
| Response | `data` and `variables` byte-identical after the run |

**The collision guard was verified, not assumed.** Removing it makes the migration throw `PrismaClientKnownRequestError` on the unique constraint — which, in one transaction, is exactly the "one bad survey rolls back every good one" failure the guard exists to prevent:

```
× skips a survey whose declarations collide, and migrates the rest
  PrismaClientKnownRequestError:
```

**Run against a local database** holding 2889 responses and a partially-migrated state (two surveys already carried rows from the write bridge):

- migrated 2 surveys / 3 fields; the already-migrated surveys were left alone
- `Response` count unchanged, 2889 → 2889
- a drift query comparing every survey's legacy JSON to its rows returned **0 rows**
- deleting the `DataMigration` record and re-running reported 0 surveys, 0 fields, with counts unchanged

**Full suites:** `packages/database` 42 passed · `apps/web` unit 7638 passed · `apps/web` integration 103 passed · typecheck, lint and prettier clean.

## QA / Test Plan

**How to test**

Still nothing user-visible — no code reads these tables until ENG-1837. The check is that the surveys this rewrote behave exactly as before.

- [ ] Before deploying, note a survey that has variables and hidden fields, plus its response count
- [ ] After the migration runs, open that survey in the editor → both cards list the same fields
- [ ] Use a variable and a hidden field in recall (`@`) and in a logic condition, then submit a response → values resolve as before
- [ ] Export that survey's responses → same headers and values as before
- [ ] Response count unchanged
- [ ] Check the deploy logs for `Skipped survey with duplicate embedded data storage keys` — each one names a survey whose declarations collide and which therefore keeps running on its legacy JSON

**Preconditions / test data**

- A workspace with surveys that have variables and hidden fields, and responses collected before the change. A duplicated survey is worth including, since the original and the copy share variable cuids.

**Risks & regressions**

- Medium. This is the only step in v1 that writes across every survey, and it runs in one transaction — so it either completes or leaves the database exactly as it was. It is additive: no existing column, row or response is modified.
- The failure mode to watch is a survey whose declarations collide. Those are skipped by design and logged by survey id, rather than failing the deploy.

- One data migration: `20260812121944_backfill_embedded_data`. No schema change, no env change.
- Nothing reads the tables yet, so this can run well ahead of ENG-1837 serving traffic.


packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
```diff
@@ -0,0 +1,123 @@
...
+        AND NOT EXISTS (
+          SELECT 1 FROM "SurveyEmbeddedData" l WHERE l."surveyId" = s."id"
+        )
+        AND (
+          (jsonb_typeof(s."variables") = 'array' AND jsonb_array_length(s."variables") > 0)
+          OR (
+            jsonb_typeof(s."hiddenFields" -> 'fieldIds') = 'array'
+            AND jsonb_array_length(s."hiddenFields" -> 'fieldIds') > 0
+          )
+        )
```
itsjavi, line 78: **The malformed-row guard only covers half the path.**

The comment above is explicit that one malformed row must not abort the whole migration, and in SQL that holds. But the two branches are OR'd, so a row qualifies on *either* one — and then the whole row goes to `planSurveyBackfill` → `toDesiredEmbeddedFields`, which does `(variables ?? []).map(...)` and `(hiddenFields?.fieldIds ?? []).map(...)` with no shape check.

Concretely: a survey with `variables = '{}'::jsonb` (an object, not an array) and a perfectly good `hiddenFields.fieldIds = ["plan"]` passes the second branch, is selected, and then throws `variables.map is not a function` in JS — inside the runner's single transaction, so every survey migrated before it rolls back and the deploy fails. The mirror case (`hiddenFields.fieldIds` a non-array while `variables` is a valid non-empty array) fails the same way. `?? []` only catches `null`/`undefined`, not a wrong JSON type.

Same class, one level down: an entry inside `variables` that isn't an object, or lacks `id`, yields `storageKey: undefined` → Prisma rejects the `createMany` on a required column → same all-or-nothing rollback. The duplicate-key guard already establishes the right pattern for this — report the survey and skip it, don't let it take the batch down.

Cheapest fix that closes both, in `planSurveyBackfill` before mapping:

```ts
const variables = Array.isArray(survey.variables) ? survey.variables : [];
const fieldIds = Array.isArray(survey.hiddenFields?.fieldIds) ? survey.hiddenFields.fieldIds : [];
```

…or, if you'd rather stay loud, extend `TSurveyBackfillPlan` with a `malformed` status so those surveys surface in the logs next to the collisions instead of silently migrating half of themselves. Either way, worth one unit test with a non-array `variables` plus a valid hidden field — the guard is currently asserted only through the SQL, which is the half that already works.

If you're confident no such row can exist in production (the column defaults are `[]` / `{"enabled": false}` and every write path is zod-validated), then the SQL `jsonb_typeof` guard is arguably also unnecessary — the inconsistency is what stands out: one path defends against malformed JSON, the very next one trusts it.

---

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
```diff
@@ -0,0 +1,123 @@
...
+  migratedFields: number;
+  skippedSurveys: { surveyId: string; duplicateStorageKeys: string[] }[];
+}
+
+/** Only the client surface this backfill needs, so a test can pass the real prisma client. */
+interface TMigrationTx {
+  $queryRaw: <T = unknown>(query: TemplateStringsArray, ...values: readonly unknown[]) => Promise<T>;
+  embeddedData: { createMany: (args: { data: TEmbeddedDataInsert[] }) => Promise<unknown> };
+  surveyEmbeddedData: { createMany: (args: { data: TSurveyEmbeddedDataInsert[] }) => Promise<unknown> };
+}
```
itsjavi: **This hand-rolled client type buys testability you already have, and costs you Prisma's input checking.**

The runner hands `run` a `DataMigrationContext["tx"]` (`Omit<PrismaClient, "$connect" | ... >`), and the real `PrismaClient` the integration test passes is assignable to that — so typing the parameter as `DataMigrationContext["tx"]` keeps the test working exactly as it does now, with no structural stand-in.

What you get back is that `createMany` is checked against the generated Prisma input types rather than against `TEmbeddedDataInsert`, which is a local interface asserting whatever this file says it asserts. That matters immediately for one field: `defaultValue` is a nullable `Json` column, and the write bridge deliberately routes it through `toStoredDefaultValue` → `Prisma.DbNull`, with a comment explaining that Prisma distinguishes JSON `null` from SQL `NULL` on such a column. This migration passes a bare `null` for every ingested field. The integration test asserts `defaultValue: null` and passes, so it round-trips through the client either way — but a bridge-created row and a backfilled row may not be byte-identical in the column, and anything later that filters `"defaultValue" IS NULL` in SQL, or diffs the two paths, would see the difference. With the real type in place the compiler would have raised it; as written, nothing does.

Worth either switching to the runner's own tx type, or reusing `toStoredDefaultValue`'s convention here so both writers agree.

---

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
```diff
@@ -0,0 +1,123 @@
...
+      `Embedded Data backfill progress: ${stats.migratedSurveys.toString()} surveys, ${stats.migratedFields.toString()} fields`
+    );
+  }
+
+  for (const survey of stats.skippedSurveys) {
+    // Not a failure of the migration: these surveys declare the same address twice, which the unique
+    // constraint has always forbidden and which nothing could have written through the editor. They
+    // keep working on their legacy JSON until someone renames the collision.
+    logger.warn(survey, "Skipped survey with duplicate embedded data storage keys");
+  }
```
itsjavi: Two small things about the skipped set, both about what an operator can do with it after the fact.

**The warnings only land if the whole loop finishes.** They're accumulated in memory and flushed after the `for (;;)`, so if survey #40,000 throws for any reason, the diagnostic naming the surveys skipped before it goes down with the transaction — exactly the run where you'd most want to know which surveys were involved. Logging each one at the point it's skipped (and keeping the end-of-run summary line) costs nothing and makes the log useful on the failure path too.

**The recovery path is worth writing down.** Once this commits, `DataMigration` records it applied and it never runs again, so a skipped survey is not picked up by a later deploy. As far as I can tell the recovery is: rename the colliding declaration in the editor → `updateSurveyInternal` calls `reconcileEmbeddedData` → the survey migrates on save. And until then it keeps resolving through `deriveLegacyEmbeddedData`, so ENG-1837 won't break it. If that's the intended story, it belongs in the QA / rollout notes next to the log-grep step — right now the description says the survey "keeps running on its legacy JSON" without saying what makes it stop being skipped, and an operator reading `Skipped survey with duplicate embedded data storage keys` in the deploy log has no next action.

---

packages/database/migration/20260812121944_backfill_embedded_data/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  for (const desiredField of desired) {
+    const embeddedDataId = newId();
+    fields.push({
+      id: embeddedDataId,
+      workspaceId: survey.workspaceId,
+      surveyId: survey.id,
+      name: desiredField.name,
+      // `reserved` is never a stored row, and `toDesiredEmbeddedFields` never produces one.
+      source: desiredField.source === "computed" ? "computed" : "ingested",
```
itsjavi: Nit: this ternary turns "impossible" into "silently wrong". If `toDesiredEmbeddedFields` ever did produce a `reserved` field, it would be written as an `ingested` row — a real row for something the schema says is never stored — and the comment above would be the only thing that noticed.

Since it can't happen today, make it say so in code rather than in prose, e.g. skip-and-report the field, or `throw` on the unexpected source. Then a future change to the shared mapping fails loudly here instead of quietly manufacturing rows.

---

itsjavi (commented):
Reviewed the migration, its utils and both test suites, and then checked the load-bearing claims against the rest of the epic branch rather than taking the description's word for them.

**What I verified, and that holds**

- **The `EmbeddedData` uniqueness story.** `@@unique([workspaceId, key])` with `key` left to its `NULL` default is genuinely safe for unlimited local rows (Postgres treats NULLs as distinct), and nothing constrains `name` — so a variable and a hidden field sharing a display name is fine, and only `@@unique([surveyId, storageKey])` can actually fire. The duplicate guard is aimed at the right constraint.
- **The single-transaction risk is bounded.** The runner sets `TRANSACTION_TIMEOUT` to 30 minutes, so the batch walk isn't sitting under Prisma's 5s interactive default. Worth confirming against production volume all the same (see the separate comment).
- **Skipped surveys really do keep working.** ENG-1836's `deriveLegacyEmbeddedData` serves a survey with no rows straight from its legacy JSON, so a skipped survey degrades to the pre-migration path rather than losing its fields once ENG-1837 flips.
- **The new integration tests actually gate.** The job is named "Better Auth integration tests" but `vitest.integration.config.mts` includes `**/*.integration.test.ts`, so the 8 new cases ran against real Postgres in CI and passed on this head — this isn't local-only coverage.
- **Keyset pagination is right, and for the stated reason.** Inserting links removes rows from the candidate set, so `OFFSET` really would step over unprocessed surveys, and walking past the last id is what stops a skipped survey being re-read forever.
- The tests are well chosen — the duplicated-survey case (shared variable cuids across two surveys) and the byte-identical `Response` assertion are the two things I'd have asked for if they weren't there.

**Findings** — four inline, one of which I'd want addressed before merge:

1. The malformed-JSON guard is applied in SQL but not in the JS that consumes the same row, so a survey that qualifies on one branch while the *other* field is a non-array still crashes the whole transaction. This is the exact failure the guard exists to prevent.
2. `TMigrationTx` opts this file out of Prisma's generated input types; the `defaultValue: null` vs the bridge's deliberate `Prisma.DbNull` is the concrete thing it hides.
3. The skipped-survey warnings are flushed only after a successful loop, and there's no stated recovery path for a skipped survey once the migration records itself applied.
4. Nit on the `source` ternary silently coercing a hypothetical `reserved` field into an `ingested` row.

**Checks at time of review**

Unit, Linters, Build, Better Auth integration, Helm, translation keys, PR size and title all green; SonarQube and SonarCloud both passed with 0 new issues and 0 hotspots. E2E still running. CodeRabbit posted a walkthrough with no findings — but note its own ESLint run was skipped in its sandbox on all four files (missing dependency), so its pass is weaker evidence than it looks. No conflict with `epic/embedded-data-v1`.

Not approving — leaving that to a human reviewer.

<sub>Reviewed with Claude Code (model `claude-opus-5`, reasoning effort: max).</sub>

---

itsjavi:
**Needs human review — @itsjavi**

I could read the code and the epic branch around it, but the things that decide whether this migration is safe to run live in production data and a production-sized database, which I can't reach from here:

**1. How many surveys will actually be skipped?** The collision guard is only reassuring if the number is small and known before the deploy, not discovered in a log afterwards. Worth running this against production (and staging) first — it's the same rule the migration applies, in SQL, read-only:

```sql
SELECT s.id, s.name, k.storage_key, COUNT(*) AS declarations
FROM "Survey" s
CROSS JOIN LATERAL (
  SELECT v->>'id' AS storage_key
  FROM jsonb_array_elements(s."variables") v
  WHERE jsonb_typeof(s."variables") = 'array'
  UNION ALL
  SELECT f #>> '{}'
  FROM jsonb_array_elements(s."hiddenFields" -> 'fieldIds') f
  WHERE jsonb_typeof(s."hiddenFields" -> 'fieldIds') = 'array'
) k(storage_key)
GROUP BY s.id, s.name, k.storage_key
HAVING COUNT(*) > 1;
```

Every row is a survey this migration will skip. While you're in there, the companion check for my first inline finding — rows that would crash the run rather than be skipped:

```sql
SELECT id FROM "Survey"
WHERE (jsonb_typeof("variables") <> 'array'
       OR (jsonb_typeof("hiddenFields" -> 'fieldIds') NOT IN ('array', 'null')
           AND "hiddenFields" ? 'fieldIds'));
```

If that returns nothing on production, the crash path is theoretical and the fix is cheap insurance. If it returns anything, it's a blocker.

**2. Runtime against the real Survey table, inside one transaction.** The 30-minute ceiling is generous, but nobody has measured this against production volume — the local run in the description was 2 surveys. A `pg_stat_statements`/`EXPLAIN` sanity check of the candidate query on a production-sized table, or just a timed dry run on a restored snapshot, would settle it.

**3. The QA checklist itself.** Recall tokens, logic conditions, response export and the editor cards after the migration all need a real environment with pre-existing data — that's the whole point of the plan and none of it is verifiable from a diff.

The code-level review is in the submitted review above.

---

itsjavi (changes requested):

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
pandeymangg: Right, and the inconsistency was the tell — fixed in `faa62d639`.

`planSurveyBackfill` now reads both columns as the raw JSON they are (`TLegacySurveyRow` says `unknown`, so it's checked rather than assumed), reports the shape problem, and skips the survey the same way a colliding one is skipped. Covers your case and the element-level one CodeRabbit raised in the same round: non-object element, element without a string id, non-string `fieldIds` entry.

Verified by reverting the guard — the new integration test fails with exactly the `.map` crash, healthy survey and all.

On your last paragraph: I ran the crash-path check against production and it returns **0 rows**, so the path is theoretical today. Kept the guard anyway — the epic ships later and the legacy API keeps writing until then, and "one path defends, the next trusts" was the real problem regardless.

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
pandeymangg: Fixed, and the `defaultValue` thing was real — I checked the rows rather than reasoning about it:

```
 name    | source   | is_sql_null | json_type
 hfone   | ingested | t           |            <- write bridge
 emailhf | ingested | f           | null       <- this backfill
```

Same logical state, two different column values. `WHERE "defaultValue" IS NULL` returns one and not the other, and both parse back to `null` in JS so nothing in the app would ever have noticed.

Now typed as `DataMigrationContext["tx"]` and routed through the same `Prisma.DbNull` convention as the bridge. Integration test asserts it in raw SQL, since the Prisma-level assertion passes either way — and reverting the conversion makes it fail.

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
pandeymangg: Both done in `faa62d639`.

Each skip is now logged where it happens, with the survey id and reason, so the diagnostic survives a later throw. The end-of-run summary line stays.

Recovery path is what you described, and it's now in the function's docstring rather than only in the PR: a skipped survey keeps resolving through `deriveLegacyEmbeddedData`, and it migrates itself the next time someone saves it — the write bridge runs the same mapping. Fixing the declaration and saving is the whole recovery. Adding it to the QA notes too.

packages/database/migration/20260812121944_backfill_embedded_data/utils.ts
pandeymangg: Fair — it now throws with the offending storage key. If the shared mapping ever starts producing `reserved`, every survey in the run is suspect, so failing loudly is the right blast radius.

packages/database/migration/20260812121944_backfill_embedded_data/migration.ts
```diff
@@ -0,0 +1,123 @@
...
+      WHERE s."id" > ${cursor}
+        AND NOT EXISTS (
+          SELECT 1 FROM "SurveyEmbeddedData" l WHERE l."surveyId" = s."id"
+        )
+        AND (
+          (jsonb_typeof(s."variables") = 'array' AND jsonb_array_length(s."variables") > 0)
+          OR (
+            jsonb_typeof(s."hiddenFields" -> 'fieldIds') = 'array'
+            AND jsonb_array_length(s."hiddenFields" -> 'fieldIds') > 0
+          )
```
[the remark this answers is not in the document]
pandeymangg: Fixed in `faa62d639` — same fix as itsjavi's thread above, which covered the container level; yours is the level below it.

`planSurveyBackfill` now rejects a non-object element, an element without a string `id`, and a non-string `fieldIds` entry, reporting the survey instead of attempting it. `[null]` was the case that would have thrown on `.id`; a missing `id` would have hit the NOT NULL on `storageKey`.

Both suites cover it: six unit cases for the shapes, and an integration test with a malformed survey alongside a healthy one asserting the healthy one still migrates and nothing rolls back.

pandeymangg:
Ran both queries against production. **Both return 0 rows.**

- **Surveys that will be skipped: 0.** No survey declares the same address twice, so the collision guard never fires on today's data.
- **Surveys on the crash path: 0.** By your own framing that makes it theoretical rather than a blocker — but the fix is in `faa62d639` anyway. The epic ships later and the legacy API keeps writing until then, and "one path defends against malformed JSON, the very next one trusts it" was worth closing regardless of the count.

One addition while I was in there: your crash query checks the containers, and CodeRabbit's inline finding is one level below that — inside a well-formed array. Those cases don't show up in either version, so I widened it. Seeded each shape and confirmed the wider query catches all five while leaving a healthy control alone:

```sql
SELECT id, "workspaceId", name, reason FROM (
  SELECT s."id", s."workspaceId", s."name",
         concat('variables is ', jsonb_typeof(s."variables"), ', not an array') AS reason
  FROM "Survey" s WHERE jsonb_typeof(s."variables") NOT IN ('array', 'null')
  UNION ALL
  SELECT s."id", s."workspaceId", s."name",
         concat('fieldIds is ', jsonb_typeof(s."hiddenFields" -> 'fieldIds'), ', not an array')
  FROM "Survey" s WHERE s."hiddenFields" ? 'fieldIds'
    AND jsonb_typeof(s."hiddenFields" -> 'fieldIds') NOT IN ('array', 'null')
  UNION ALL
  SELECT DISTINCT s."id", s."workspaceId", s."name", 'variables contains a non-object element'
  FROM "Survey" s CROSS JOIN LATERAL jsonb_array_elements(
    CASE WHEN jsonb_typeof(s."variables") = 'array' THEN s."variables" ELSE '[]'::jsonb END) v
  WHERE jsonb_typeof(v) <> 'object'
  UNION ALL
  SELECT DISTINCT s."id", s."workspaceId", s."name", 'variables element has no string id'
  FROM "Survey" s CROSS JOIN LATERAL jsonb_array_elements(
    CASE WHEN jsonb_typeof(s."variables") = 'array' THEN s."variables" ELSE '[]'::jsonb END) v
  WHERE jsonb_typeof(v) = 'object' AND jsonb_typeof(v -> 'id') IS DISTINCT FROM 'string'
  UNION ALL
  SELECT DISTINCT s."id", s."workspaceId", s."name", 'fieldIds contains a non-string entry'
  FROM "Survey" s CROSS JOIN LATERAL jsonb_array_elements(
    CASE WHEN jsonb_typeof(s."hiddenFields" -> 'fieldIds') = 'array'
         THEN s."hiddenFields" -> 'fieldIds' ELSE '[]'::jsonb END) f
  WHERE jsonb_typeof(f) <> 'string'
) crashers ORDER BY reason, id;
```

The `CASE` inside each `LATERAL` is load-bearing — `jsonb_array_elements` raises on a non-array and a `WHERE` doesn't run early enough to stop it, so without it the diagnostic itself errors on exactly the rows it's looking for.

**Your point 2 (runtime) and point 3 (QA on real data) are still open and still need a human.** Nothing I can do from here changes that — the local run was 2 surveys. A timed dry run on a restored snapshot is the thing that settles the transaction question.

Worth noting for whoever does it: production has **zero** surveys needing the collision or crash handling, so the runtime question is purely about volume — how many surveys carry declarations at all, and how long the keyset walk plus two `createMany` calls per batch take on that. The candidate query is the part to `EXPLAIN`.

itsjavi (approved):
thanks LGTM

PR #8835, opened 2026-08-12
chore(deps): upgrade Better Auth to 1.7.0 and migrate MCP OAuth to the resource model

xernobyl:
Fixes ENG-2343

## What & why

`@better-auth/oauth-provider` 1.6.x has [GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528): the authorization server never persists the resource a user approved, so an access token's audience is not bound to its grant. #8829 closed the exploitable half from our side (the MCP resource server refuses foreign-audience tokens); it could not fix the AS, because the fix only exists on the 1.7 line.

**1.7.0 went stable on 2026-08-18**, so this now ships the real fix rather than a mitigation. The AS binds each token to the resources the authorization actually covered, and `pnpm audit` no longer reports the advisory.

It is a coupled auth-core upgrade — `better-auth` + `@better-auth/core` + `@better-auth/oauth-provider` move together and the plugin declares the other two as matching peers — so it touches login, sessions, 2FA and SSO, not just MCP:

- **Resource model.** `validAudiences` is gone; the MCP resource is registered in `resources` and linked to every DCR client. Three new tables plus new columns on the four existing `oauth*` tables.
- **Account identity.** 1.7 keys `Account` on `(issuer, accountId)`. `issuer` is added nullable and backfilled — this is the part that takes production down if it is wrong, since `findCredentialAccount` (password sign-in) and `updatePassword` both filter on it.
- **Generic OAuth rebuilt onto the social-provider path.** `genericOAuthClient()` no longer exists and `signIn.oauth2()` → `signIn.social()`. Upstream also moves the callback to `/api/auth/callback/{providerId}`, which would have made every self-hoster re-register a redirect URI — **so the callback URL is now pinned instead** and nobody has to touch their IdP. See below.
- **JWT keyring.** `alg`/`crv` on `jwks`. See Migrations & env; this one would have broken MCP OAuth on any fresh install.
- **Two 1.7 behaviours are normalized in the `/api/auth/*` catch-all**, because neither is configurable and both break a flow that worked on 1.6. `modules/auth/lib/legacy-sso-callback.ts` serves the pinned SSO callback path (below), and `modules/auth/lib/mcp-dcr-application-type.ts` fills in `application_type: "native"` on a dynamic client registration that omits it and asks for a loopback redirect — 1.7 hardcodes the default to `"web"` at the call site and then refuses `web` + loopback, which is every local MCP client. Both are narrow, both leave an explicitly-supplied value alone, and both are the kind of thing to re-verify on the next auth-core bump — see Risks.

### The SSO callback URL is pinned, not moved

Better Auth has moved this path twice with no choice of ours: 1.6's `genericOAuth` plugin mounted its own `/oauth2/callback/:providerId` route, and 1.7 rebuilt that plugin onto the built-in `/callback/:id` route. OAuth requires `redirect_uri` to match exactly (RFC 6749 §3.1.2.2, no wildcards), so each move means every operator edits every identity provider by hand. v5.2 already did that to them; this would have been the second time in three minor releases. There is no standard path for a relying party's callback — OAuth/OIDC standardise only the *provider's* discovery paths and make everything else discovered or registered — so nothing external pulls us toward upstream's default.

Two halves, deliberately asymmetric:

- **`redirectURI` is pinned** on the three generic providers to the v5.2 URL their IdPs already have. Upstream resolves `options.redirectURI || redirectURI` in *both* legs that must agree — the authorization request and the token exchange — so the IdP never sees a mismatch. The option is not new: 1.6 honoured it with identical precedence, so this is not a 1.7 affordance we might lose next minor. Built off `getAuthIssuerUrl()` rather than `WEBAPP_URL` so the host receiving authorization codes is the one Better Auth itself considers its base; this also removes the Host-header influence that exists today when `BETTER_AUTH_URL`/`NEXTAUTH_URL` are unset.
- **We serve that path**, because `redirectURI` does not move the route the handler is mounted on. `modules/auth/lib/legacy-sso-callback.ts` maps it onto the path the installed version serves, inside the existing `/api/auth/*` catch-all. Matched as a suffix with the provider id exact and no second `/api/auth` in the prefix, so the only path it can emit is `<basePath>/api/auth/callback/<pinned-id>` — it cannot reach an endpoint the caller could not already reach, and cannot shadow the 22 sibling `/oauth2/*` routes of our own MCP OAuth server.

`accountSubject` is also pinned per provider, for a related reason: 1.7 resolves the subject as `isOidc ? profile.sub : profile.id`, and `isOidc` is set only inside the discovery branch — so a provider configured with explicit endpoints instead of discovery (Azure `common`, and the SAML bridge) silently resolves `profile.id`. Microsoft Graph's userinfo does not return that field, which is a broken Azure sign-in. Pinning it to the field each provider actually sends makes identity derivation ours rather than a side effect of whether discovery ran.

The external URL stays `/api/auth/oauth2/callback/{providerId}`; Better Auth's internal `ctx.path` is `/callback/:id`, which is what the hook, audit and observability matchers key off — those are correct as they are and must not be "restored".

## Breaking changes

| Access tokens no longer outlive the session | Token valid until its own TTL | Introspection returns `active: false` and UserInfo rejects it once the bound session ends | Anything holding an MCP access token across a sign-out | None for us — MCP clients re-authorize. Upstream behaviour change, recorded because it is observable |

## Migrations & env

- **No new env vars.**
- **Dependency declaration, after merging main's pnpm catalog migration.** main moved every dependency
  used by two or more workspaces into the `catalog:` block of `pnpm-workspace.yaml`, enforced by
  `scripts/check-catalog.mjs` in `pnpm lint`. The Better Auth trio is declared by **one** workspace
  (`apps/web`) only, so it correctly stays a literal version rather than a catalog entry — `check-catalog`
  passes (64 catalogued dependencies). Nothing here needs to move to the catalog, and a future Better
  Auth bump still edits `apps/web/package.json`.
- **Lockfile:** taking main's `pnpm-lock.yaml` through the merge resolves the trio at `1.6.23` against a
  `package.json` asking for `1.7.0`, which `pnpm install --frozen-lockfile` rejects — so it is
  regenerated in its own commit. `--frozen-lockfile` exits 0 on the merged tree. This is why the diff
  carries lockfile churn beyond the three version lines.
- **The Better Auth supply-chain cooldown exemption is removed in this PR** (ENG-2349, `6e89625ba`). 1.7.0 published 2026-08-18T00:07Z and `minimumReleaseAge` is 72h, so it became installable unaided at 2026-08-21T00:07Z; the nine-package `minimumReleaseAgeExclude` entries that bridged that window are gone, `@formbricks/hub` stays. `pnpm install` reports `Lockfile passes supply-chain policies (2337 entries)` and `pnpm-lock.yaml` is byte-identical — the same 1.7.0 resolves, the waiver just is not load-bearing any more. Removed here rather than in its own PR because those entries matched by **name, not version**, so while they stood any later Better Auth release would have skipped the cooldown silently.
- `20260812110000_eng_2343_better_auth_17_resource_model` (schema): adds `oauthResource`, `oauthClientResource`, `oauthClientAssertion`; new columns on `oauthClient`, `oauthAccessToken`, `oauthRefreshToken`, `oauthConsent`; `alg`/`crv` on `jwks`; `Account.issuer` nullable **plus its backfill**, which runs before the new unique index so a collision fails the migration instead of surfacing later as a login bug. Legacy `oauthClient.public`/`.type` are kept, so the schema half stays additive and rollback-safe.
- `20260812110001_eng_2343_backfill_oauth_resource_links` (data): seeds the MCP `oauthResource` row and links existing clients, refresh tokens and consents to it. Needs `WEBAPP_URL` (it derives the resource identifier) and no-ops on an empty DB.
- **`jwks.alg`/`.crv` are not optional in practice.** `createJwk` always writes `alg` and, under the default EdDSA/Ed25519 keyring, `crv`. Without the columns Prisma rejects the insert, so the first key mint fails and JWT signing — and all of MCP OAuth — goes down with it. Nullable because pre-1.7 rows have no value and upstream reads `alg: null` as the configured default, so existing keys keep signing untouched.
- Rolling deploys: `Account.issuer` is nullable specifically so a 1.6 pod still serving can keep inserting. A 1.6 pod will write NULL-issuer rows that 1.7 cannot see until re-backfilled.
- **Rolling deploys are also why the callback pin matters beyond operator convenience.** A 1.6 pod advertises `/api/auth/oauth2/callback/{providerId}` as its own default; a 1.7 pod advertises the same URL because it is pinned. So both pods send the IdP the same `redirect_uri` and both can serve the callback — the old one natively, the new one through the mapper. Had the path tracked the upstream default instead, mid-rollout sign-ins would have been split across two URLs with only one of them registered at the IdP, so a share of SSO logins would fail for the length of the deploy. Pinning removes that window.
- The `Account.issuer` backfill cannot collide, so it cannot abort a deploy: the provider→issuer mapping is injective (`credential`, `google`, else `local:oauth:<provider>`), and `@@unique([provider, providerAccountId])` already holds, so `(issuer, providerAccountId)` uniqueness follows structurally. Legacy NextAuth `azure-ad` rows are not a case here either — `20260416110000_backfill_legacy_sso_accounts` renamed them to `azuread` long before this migration runs.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `jwks.alg`/`.crv` exist, so a key mint succeeds | unit (red on main) | New `modules/auth/lib/better-auth-schema-contract.test.ts` reads each plugin's own declared schema and diffs it against `main.prisma`. Removing the two fields fails it with `expected [ 'alg', 'crv' ] to deeply equal []` |
| No schema drift on any of the 8 models Better Auth owns | unit (guard) | Same test, parametrised per model — `oauthClient` alone has 35 declared fields. Reads the plugins' exposed schema rather than a hand-copied list, and asserts our model is a superset (our extra columns — the legacy `oauthClient.public`/`.type` kept for rollback, Prisma relations — are expected). Verified it binds by deleting `oauthClient.clientCredentialsScopes`: fails with `expected [ 'clientCredentialsScopes' ] to deeply equal []` |
| The same, against a real database | manual | Ran better-auth's real key-mint write shape through the generated Prisma client on a freshly migrated Postgres: `MINT OK -> {"alg":"EdDSA","crv":"Ed25519"}`. Dropped both columns and reran: `PrismaClientKnownRequestError`. So the failure was real and the fix is what closes it |
| Existing users still sign in after the `Account.issuer` backfill | manual | Applied the full migration history to a throwaway Postgres seeded with pre-upgrade `credential`, `google` and `github` accounts, then signed in. `google` needed its own case — the built-in provider declares `accountIssuer: "https://accounts.google.com"`, not the synthetic form — which is in the migration |
| Credential sign-in keyed on `(issuer, accountId)` | unit (red on main) | Fixtures that create credential accounts by raw Prisma insert were leaving `issuer` unset, so sign-in returned `INVALID_EMAIL_OR_PASSWORD` for a *correct* password. Fixed in `playwright/fixtures/users.ts` + three integration suites; all 12 previously-failing tests pass against real Postgres |
| Token audience bound to the approved resource | unit (guard) | `mcp-oauth-provider-options.test.ts` pins exactly one registered resource and the full advertised scope set; `mcp-oauth-dcr.test.ts` drives a real Better Auth instance through DCR |
| The resource server still refuses foreign audiences | unit (guard) | `modules/mcp/auth.test.ts` mints real signed JWTs. Kept deliberately: 1.7 drops `verifyOptions.audience` from its own `jwtVerify`, so this is now the only audience check on that path |
| rc.5 → 1.7.0 stable is a no-op for this branch | manual | `npm pack`ed both and compared semantically rather than by chunk filename. The oauth-provider `schema` object is byte-identical (MD5 `a6e95a09572bf66d303024be1deaf601`), `OAuthOptions` differs only in dts reflow, and `createLocalAccountIssuer`/`createOAuthAccountIssuer`/`validateClientRedirectUri`/`verifyBearerToken` are unchanged |
| The pinned URL is what Better Auth advertises — on **both** legs | unit (guard) | `better-auth-redirect-uri-pin.test.ts` drives a real `betterAuth` instance. One case reads the authorization URL's `redirect_uri`; another runs a full two-leg flow (sign-in through `auth.handler` for the genuine `state` + signed cookie, then the callback with `fetch` stubbed at the IdP boundary) and asserts the **token request's** `redirect_uri` and that both legs agree. That second case exists because a release keeping the option on one leg and dropping it on the other would otherwise leave the guard green while every sign-in died at the token endpoint. Unpinning fails all three |
| The pinned path reaches Better Auth's callback route | integration (real Postgres) | `legacy-sso-callback.integration.test.ts` asserts the pinned path is handled *identically* to the path this version serves, comparing responses rather than hardcoding today's OAuth error. Mutation-tested: without the mapper the pinned path is 404 while the native path is 403 |
| The mapper cannot shadow our own MCP OAuth routes | unit (guard) | `legacy-sso-callback.test.ts` — 24 cases: each pinned provider, `code`/`state` preservation, basePath form, `form_post` body, and negatives for the sibling `/oauth2/*` routes, unpinned ids, deeper paths, percent-encoded and case-varied spellings, and a crafted double-auth-segment path |
| The route actually applies the mapper | unit (guard) | New cases in `app/api/auth/[...all]/route.test.ts`. Note the pre-existing `toHaveBeenCalledWith(request)` assertions there **could not fail** — a `Request` has no own properties, so any two compare deep-equal (verified). All three now assert identity, and a rebuilt pass-through equivalent fails them |
| A real sign-in reaches the pinned URL through Next's router | manual | Dev server on a real database: the three pinned paths return the *same* status as the native path (403 state rejection), while `not-pinned`, `google` and `openid/extra` 404 and `/oauth2/userinfo` (MCP) still 401. Percent-encoded, upper-cased and crafted double-auth-segment paths 404; dot segments normalise before matching. Next 308-canonicalises trailing/doubled slashes *before* the handler, so that shape is mapped in production — the code comment said otherwise and was corrected |
| The GHSA fix itself, end to end on a running instance | manual | Full handshake against a dev server: DCR (`application_type: "native"` + loopback redirect accepted) → `/authorize` with `resource` → consent → token. The access token came back `typ: at+jwt`, `alg: EdDSA`, and `aud: ["…/api/mcp", "…/api/auth/oauth2/userinfo"]` — the approved resource plus the documented implicit UserInfo audience, exactly the pair `hasAcceptedMcpAudience` allow-lists. `tools/list` returned 40 tools. Negative case, which is the advisory itself: refreshing with `resource=https://evil.example.com/api` is refused `400 invalid_target`. `alg: EdDSA` also proves the key mint succeeded, which is the `jwks.alg`/`.crv` failure this PR fixes. **Caveat, since it matters:** this ran on a database migrated *from empty*, and the data migration returns early at `clients === 0` — so this run took the path that SKIPS the backfill and could not have caught the `allowedScopes` blocker. The row below is the one that covers it |
| **The upgrade path, on a non-empty database** | manual | The case the row above cannot reach. Built a pre-1.7 instance — users with `credential`/`google`/`github`/`openid`/`azuread`/`saml` accounts all at `issuer = NULL`, an existing `oauthClient`, a live refresh token and a consent with empty `resources` — then ran the migrations' own SQL. Every provider maps correctly (`github → local:oauth:github` confirmed against `account-key.mjs:25`, which falls back to `createOAuthAccountIssuer` when a provider declares none), `nulls=0`, no `(issuer, providerAccountId)` collisions, resource scopes 10, client link present, zero refresh tokens or consents left without resources, and re-running is idempotent |
| A pre-upgrade password user can still sign in | manual | Live against that upgraded database: `POST /sign-in/email` → 200, session established |
| A pre-upgrade user's password change actually applies | manual | The nastiest failure mode, since `updatePassword` also filters on `issuer` and would report success while matching zero rows. Rotated it: change → 200, new password signs in → 200, **old password → 401**. Not a silent no-op |
| The `allowedScopes` blocker, demonstrated both ways | manual | On the upgraded instance with a pre-existing client: with the fix `/authorize` proceeds to consent; with `allowedScopes` emptied back to the shipped state it returns `error=invalid_scope` — "none of the requested scopes are allowed for resource …", the exact message from `resolveResourcePolicy` |
| Azure keeps working with no tenant configured | unit | Both strategies asserted: `AZUREAD_TENANT_ID` set → `discoveryUrl` (id_token verified against a concrete issuer, stronger than 1.6); unset → explicit `login.microsoftonline.com/common/oauth2/v2.0/*` + `graph.microsoft.com/oidc/userinfo` endpoints, so no `idTokenConfig` is built and identity comes from UserInfo as in 1.6 |
| Every `Account` write path sets `issuer` | unit | `account-linking.ts` create + both normalize branches, and `packages/database/src/seed.ts`. The existing account-linking tests caught all three changes before I updated them |
| The migration's scope literal cannot drift from the app's | unit | `mcp-oauth-resource-seed.test.ts` pins it against `MCP_OAUTH_SCOPES` (`packages/database` cannot import `apps/web`). Mutation-tested: dropping one scope fails it |
| `pnpm audit` no longer reports the advisory | manual | Gone from the report; only the separately-documented `brace-expansion` and `deepmerge-ts` findings remain. The ~50-line accepted-risk note it justified is removed |
| Azure sign-in after the id_token split | manual, live | 1.7 resolves the account subject as `isOidc ? profile.sub : profile.id`, and `isOidc` is set only inside the discovery branch — so configuring Azure `common` with explicit endpoints (which the id_token fix requires) flipped resolution to `profile.id`, which Microsoft Graph's userinfo does not return. Reproduced locally with a stub OIDC provider whose discovery omits `jwks_uri`, which yields the same `isOidc = false` state: **with `accountSubject` pinned** the flow lands on `/` and writes `providerAccountId = stub-subject-1` (the `sub`) with `issuer = local:oauth:openid`; **with the pin removed and nothing else changed** it ends at `/auth/login?error=unable_to_get_user_info`. Control and proof, on the real request path |
| Loopback DCR on 1.7 | manual, live + unit | 1.7's `validateClientRedirectUri` refuses any loopback redirect for a `web` client, and DCR hardcodes `web` as the default. Live matrix against a running instance: loopback-only → `native` ✅, loopback + https → `native` ✅, https-only → `web` ✅, loopback + non-loopback http → `invalid_redirect_uri` ❌, explicit `web` + loopback → refused (never overridden) ❌, explicit `native` → preserved ✅. A registered loopback client is also usable, not just registrable — `/authorize` 302s to login carrying the `resource` param, no `invalid_client`. `mcp-oauth-dcr.test.ts` drives the real validator for both the refusal and the acceptance, plus the boundary case that `native` does not make a non-loopback http redirect acceptable |
| Core-table schema drift | unit, mutation-tested | The contract test only read the plugins' declared schemas, so it could not have caught `Account.issuer` — a core field no plugin declares. It now also resolves the merged core schema via upstream's `getAuthTables`. Dropping `issuer` from the Prisma model fails two tests. `user.image` is excluded (we dropped that column) and the exclusion is bound to the SSO hook that strips the field, so removing the strip fails too |

| **SSO sign-in as a pre-upgrade user** | manual, live | The case this PR's risk section leads with, and it turned out to be reproducible locally rather than staging-only. Seeded a user + `openid` `Account` exactly as a 1.6 pod leaves it (`issuer` NULL, `access_token = 'old-token'`), ran the migration's own backfill statement verbatim, then signed in through the stub IdP with the same `sub`. Result: **the existing rows are reused** — same `userId`, same `Account.id`, still 1 user / 1 account, and `access_token` updated to the fresh value, which proves 1.7 *found and updated* the backfilled row rather than inserting alongside it. Control: setting `issuer` back to NULL and retrying does **not** duplicate or hard-fail — it routes into Formbricks' own verify-before-link recovery (`/auth/verification-requested?…&purpose=sso_recovery`). Restoring the backfill returns it to a clean sign-in |

**Open gaps**

- **No SSO E2E exists anywhere in the repo** (ENG-2445), so nothing automated completes a handshake. A stub IdP has since closed more of this by hand than expected — the callback routing, the pinned `redirect_uri` on both legs, the `accountSubject` resolution, and the pre-upgrade-user backfill are all now verified live against a running instance (rows above). What a stub **cannot** cover is the part that depends on a real provider's own behaviour: Azure AD against real Entra (its userinfo shape is the thing that broke, and `AZUREAD_TENANT_ID` unset vs set take different code paths), and the SAML bridge end to end through BoxyHQ. Both still need staging, and the claim to check there is unchanged: sign-in works **without touching the IdP configuration**.
- The full integration suite is verified by CI, not locally. Running it here needs `TEST_DB_*`/Redis overrides plus the env CI seeds from `.env.example`; with those, four password-reset tests still fail locally on an `Invalid URL` built from `WEBAPP_URL`. That is local env — CI passes those same files — but it means the local reproduction of that suite is incomplete, and the next person will hit the same wall.
- No *automated* test mints a JWK through the running app. It is now proven by hand end to end — the live handshake above issued an `EdDSA`-signed token, which cannot happen without a successful mint on a database migrated from empty — and guarded against schema drift by the contract test. Still worth automating.
- ~~The MCP OAuth handshake needs a smoke test.~~ **Done** — driven end to end against a running instance (coverage row above), including the `invalid_target` refusal. Still worth repeating against the deployed image, since that exercises the real ingress and TLS.
- ~~Whether real MCP clients send `application_type: "native"` during DCR.~~ **No longer a dependency** — 1.7 refuses a loopback redirect for a `web` client, and defaults the field to `web` as a literal at the call site, so a client that omits it would 400 before consent. `mcp-dcr-application-type.ts` now infers `native` from a loopback redirect per RFC 8252 §7.3, so the flow works whether or not the client sends the field, and an explicit value is never overridden. What Claude Code and MCP Inspector actually send is still worth observing on staging, but nothing depends on the answer.

**Risks**

- `Account.issuer` is the one that can take production down, but the two account kinds fail differently, and it is worth knowing which is which before deploying. **Credential accounts are the dangerous half**: `findCredentialAccount` and `updatePassword` both filter on `issuer`, so a missing value locks password users out *and* makes their reset silently match zero rows, with no self-service way back. **SSO accounts degrade gracefully** — verified live: a NULL-issuer SSO row does not duplicate or error, it routes the user into the existing verify-before-link recovery flow. So re-check password login first after deploying; SSO is the more forgiving surface, not the more fragile one.
- `modules/auth/lib/session-cookie.ts` reimplements better-call's cookie HMAC and hardcodes the cookie names. A change there fails as an auth redirect loop, not an error.
- ~25 endpoint-path string literals back rate-limit rules, HIBP paths and audit allow-lists. A silent rename disables a security control rather than erroring. Note the split the pin introduces: the **external** URL is `/api/auth/oauth2/callback/{id}` while Better Auth's **internal** `ctx.path` is `/callback/:id`, so those matchers correctly stay on `/callback/` — a future reader "restoring" `/oauth2/` there would break the audit `authMethod` and the SSO hooks.
- **We now own two normalizers in the auth catch-all** (the SSO callback path and the DCR `application_type`), which is a deliberate divergence from upstream defaults rather than an accident — worth stating so nobody "simplifies" it back. The cost is that each future Better Auth upgrade has to re-verify both halves; the guard tests make that cheap (the redirect-URI pin drives a real `betterAuth` instance, and the DCR pair drives the real validator), but they only fail if someone runs them. Both fail closed rather than open: a broken pin yields `state_not_found`, and a dropped DCR inference yields `invalid_redirect_uri`.
- `@better-auth/utils` stays at 0.4.2 (verified unmoved). `base32.decode` sits on the 2FA cutover path, so a bump there would need its own check.

---

> [!NOTE]


xernobyl:
Known conflict to reapply at rebase time: [#8846](https://github.com/formbricks/formbricks/pull/8846) (`epic/authzed`) changes the same line in `apps/web/modules/mcp/auth.ts` — it replaces `jwksUrl: \`${getAuthIssuerUrl()}/jwks\`` with a new `getMcpOAuthJwksUrl()` helper (internal JWKS override, ENG-2360). This PR rewrites that same call to `verifyBearerToken` + `typ: at+jwt`.

Note it lands on `epic/authzed`, not `main`, so it won't appear here until that epic merges. Two things to watch:

1. When rebasing this branch, keep **both** changes — the 1.7 rename *and* `jwksUrl: getMcpOAuthJwksUrl()`.
2. If this merges to `main` before the epic does, the epic's merge carries the older `verifyAccessToken` form of that line and could revert the 1.7 change on conflict resolution. Verify by content after the epic lands, not by ancestry.

xernobyl:
Merged `main` in (`2dc07bf23`) — this PR is `MERGEABLE` again after 23 commits of drift.

One conflict, `modules/ee/sso/lib/better-auth-hooks.test.ts`, where both sides had changed the same region for unrelated reasons.

**Kept main's additions:** the `@/lib/constants` mock with a mutable `SIGNUP_ENABLED`, `@/lib/instance/service`, and the license-check mock it moved into that block.

**Kept this branch's callback path**, deliberately. `main` and the merge base both use `/oauth2/callback/:providerId`; this branch uses `/callback/:providerId` because **Better Auth 1.7 routes the generic-OAuth plugin through the social-provider callback** rather than its own endpoint — the reason is already in the hook's own comment, which this branch rewrote:

> Better Auth sets `context.path` to the ROUTE PATTERN — `/callback/:id` for both built-in social … (generic-OAuth now goes through the social-provider route instead of its own `/oauth2/callback/:providerId`)

Taking main's value would have exercised a path 1.7 no longer produces. Worth knowing the matcher (`/\/callback\/([^/?:]+)$/`) accepts both shapes, so this would **not** have failed — it would just have stopped testing the real 1.7 path.

**Checked the duplicate this shape invites:** main *moved* the license-check mock rather than adding a second, and the auto-merge applied that removal from its old position, so the resolved file has exactly one `vi.mock` per module. Verified by count.

**The lockfile got the most scrutiny**, since an auto-merged `pnpm-lock.yaml` on a dependency-upgrade PR is where this would quietly go wrong: `pnpm install --lockfile-only` leaves it byte-identical, and the supply-chain policy check passes across 2333 entries. `better-auth` and `@better-auth/*` are still pinned at `1.7.0-rc.5`.

Verification: typecheck 12/12 · `format:check` clean · full web suite **8020 passing, zero failures** · of those, 697 across the `auth`, `mcp`, `sso` and `api/v3/lib` suites this upgrade actually touches.

Still a draft — the gating question (shipping on an `rc`) is unchanged by this merge.

xernobyl:
Pushed one docs fix after opening for review, so the diff you're reading is current: the v5.4 migration section enumerated the schema changes but predated the `jwks` `alg`/`crv` discovery, so those two columns weren't mentioned. The section's guarantee still held (both additive and nullable, neither needs a backfill), but a self-hoster would have met two unexplained columns on the table holding their signing keys. [e2eb34858](https://github.com/formbricks/formbricks/commit/e2eb34858) — docs only, no code.

Two things worth your attention as reviewer, both already in the description:

1. **`Account.issuer` is the one that can take production down.** A wrong or missing value locks password users out *and* makes their reset silently match zero rows. The migration backfills before adding the unique index so a collision fails the migration rather than surfacing later as a login bug, and `google` gets its upstream-declared `https://accounts.google.com` rather than the synthetic form. Verified against a throwaway Postgres seeded with pre-upgrade `credential`/`google`/`github` accounts.

2. **The staging SSO pass is the real open gap.** There is no SSO E2E anywhere in the repo (now tracked as ENG-2445), so Azure AD / OIDC / SAML — and specifically **signing in as a user that existed before the migration** — is the one claim here with neither a test nor a real-Postgres proof behind it. That last case is what proves the issuer backfill.

packages/database/migration/20260812110000_eng_2343_better_auth_17_resource_model/migration.sql
```diff
@@ -0,0 +1,173 @@
...
+-- built-in social provider must be checked the same way before it is enabled.
+UPDATE "Account"
+SET "issuer" = CASE
+  WHEN "provider" = 'credential' THEN 'local:credential'
+  WHEN "provider" = 'google' THEN 'https://accounts.google.com'
+  ELSE 'local:oauth:' || "provider"
+END
+WHERE "issuer" IS NULL;
+
+CREATE UNIQUE INDEX "Account_issuer_providerAccountId_key" ON "Account"("issuer", "providerAccountId");
```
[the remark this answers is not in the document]
xernobyl: Valid, and it is repo policy rather than preference — `.coderabbit.yaml` requires migrations to be idempotent and convergent, including against a `db:push` database. Fixed in 632d3035c.

Every added column, table and index is now guarded. The two foreign keys use `DO $$ ... EXCEPTION WHEN duplicate_object` because Postgres has no `IF NOT EXISTS` for `ADD CONSTRAINT`. Style follows `20260812120000_add_feedback_source_element_scope`, the most recent migration to land under the same rule.

One part needed no change: the `Account.issuer` backfill was already convergent via `WHERE "issuer" IS NULL`, so it fills only nulls and never overwrites a value a previous run — or a newer pod — already wrote.

Proven rather than asserted: applied the full migration history to a fresh database, then re-applied this file twice more directly with `ON_ERROR_STOP=1`. All three exit 0, the second and third emit only `already exists, skipping` notices, the backfill reports `UPDATE 0`, and the end state is identical each time (3 indexes on `oauthClientResource`, both `jwks` columns, zero null issuers).

On the SQLFluff `CONCURRENTLY` hints: not applicable here. Prisma runs migration files inside a transaction and `CREATE INDEX CONCURRENTLY` cannot run in one. Every index except `Account_issuer_providerAccountId_key` is on a table this migration just created, so there is nothing to lock; that one is on an existing table and is a deliberate trade — it must exist before 1.7 serves, and the alternative is a separate out-of-transaction migration for a lock measured in seconds at our row counts.

packages/database/schema/main.prisma
```diff
@@ -1013,10 +1077,66 @@ model oauthConsent {
...
+  clientId   String
+  client     oauthClient   @relation(fields: [clientId], references: [clientId], onDelete: Cascade)
+  resourceId String
+  resource   oauthResource @relation(fields: [resourceId], references: [identifier], onDelete: Cascade)
+  metadata   Json?
+  createdAt  DateTime?
+
+  @@unique([clientId, resourceId])
+  @@index([clientId])
+  @@index([resourceId])
```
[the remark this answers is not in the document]
xernobyl: Correct, fixed in 632d3035c. `@@unique([clientId, resourceId])` is a btree led by `clientId`, so the separate index only added write amplification. Removed from both the Prisma model and the migration so they cannot drift, with a comment recording why the sibling `oauthClient*` models legitimately keep theirs — none of them has a composite unique starting with `clientId`.

Verified on a migrated database: `oauthClientResource` now has exactly `pkey`, the `(clientId, resourceId)` unique, and `resourceId_idx`.

pnpm-workspace.yaml
```diff
@@ -260,3 +223,24 @@ minimumReleaseAgeExclude:
...
+  # `pnpm install` to confirm the cooldown accepts 1.7.0 on its own. That is the whole of ENG-2349.
+  - "better-auth"
+  - "@better-auth/core"
+  - "@better-auth/oauth-provider"
+  - "@better-auth/prisma-adapter"
+  - "@better-auth/drizzle-adapter"
+  - "@better-auth/kysely-adapter"
+  - "@better-auth/memory-adapter"
+  - "@better-auth/mongo-adapter"
+  - "@better-auth/telemetry"
```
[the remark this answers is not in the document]
xernobyl: Fair, and it is the more durable half of the risk. Recorded in 632d3035c.

The block now says explicitly that the entries match by NAME, not version, so until ENG-2349 lands this also exempts any later Better Auth release — 1.7.1, 1.8.0 — not just the 1.7.0 it was added for, and a bump landing in that window would bypass the cooldown with nothing in CI saying so.

Not adding a date-triggered failing check: it would be a test whose only job is to go red on a date, and the thing it guards is already a dated, assigned ticket (ENG-2349, due 2026-08-21) whose entire remaining scope is deleting this block. A second mechanism would be more surface than the risk it covers, for the three days it exists.

apps/web/modules/auth/lib/mcp-oauth-dcr.test.ts
```diff
@@ -170,13 +250,42 @@ describe("MCP OAuth Dynamic Client Registration → authorize (real-client shape
...
+
     const registration = await registerClient(auth, ["surveys:read"]);
     const clientId = registration.body.client_id;
     expect(clientId).toBeTruthy();
 
+    expect(registration.body.scope?.split(" ")).toEqual(expect.arrayContaining(["surveys:write"]));
+
+    // Consequence: a scope the client never requested is now accepted at authorize.
     const authorize = await requestAuthorize(auth, clientId as string, ["surveys:read", "offline_access"]);
+    expect(authorize.location).not.toContain("error=invalid_scope");
```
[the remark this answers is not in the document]
xernobyl: Right, and it is the sharper version of the problem: `requestAuthorize` defaults `location` to `""` when the header is absent, and `""` satisfies every `not.toContain` — so that assertion would also have passed on a different error, or on no redirect at all. Fixed in 632d3035c.

Rather than guess the accepted outcome I ran it: unauthenticated authorize bounces to the configured `loginPage` with no `error` param. So it now asserts that positively — location is truthy, pathname is `/auth/login`, and `searchParams.get("error")` is null.

Checked that it actually binds by asking for a scope outside the advertised set, which must be refused: it fails with `expected '/callback' to be '/auth/login'` — naming the outcome rather than the absence of a substring.

packages/database/migration/20260812110001_eng_2343_backfill_oauth_resource_links/migration.ts
```diff
@@ -0,0 +1,143 @@
...
+  try {
+    const url = new URL(configured);
+    url.hash = "";
+    url.search = "";
+    const basePath = trimTrailingSlash(url.pathname);
+    return `${url.origin}${basePath.endsWith("/api/mcp") ? basePath : `${basePath}/api/mcp`}`;
+  } catch {
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
xernobyl: Split verdict — the second bullet is a real gap and is now documented; the first is based on a premise I could not reproduce.

**Bullet 1, the precedence mismatch: not reproducible.** The two derivations already agree. `getMcpResourceUrl()` is `appendPath(getWebAppBaseUrl(), "/api/mcp")`, and `getWebAppBaseUrl()` is `normalizeConfiguredUrl(env.WEBAPP_URL)` — `WEBAPP_URL` only. The `BETTER_AUTH_URL ?? NEXTAUTH_URL ?? WEBAPP_URL` chain belongs to `getAuthIssuerUrl()`, a different function for a different value (the issuer, not the resource). `resolveMcpResourceIdentifier` in the migration reads `WEBAPP_URL` and normalizes identically: clears hash and search, trims the trailing slash, appends `/api/mcp` unless the path already ends in it. Same input, same normalization, same output.

The one deliberate difference is the unset case: the app falls back to `http://localhost:3000`, the migration returns null and throws `WEBAPP_URL is not set`. That asymmetry is intentional — a migration that silently seeds a localhost resource identifier into a production database is worse than one that refuses to run.

So no change there; adopting the issuer chain would have introduced the mismatch it was meant to prevent.

**Bullet 2, the base-URL change: valid, documented in 632d3035c.** `mcp-oauth-provider-options.ts` now spells out the consequence — the identifier is derived from `WEBAPP_URL`, so a new URL is a new resource. With `insertOnly`, boot inserts a second `oauthResource` and repoints `clientRegistrationDefaultResources` at it, while every already-registered client keeps its link row for the old identifier and then fails `invalid_target` at the token endpoint, after the user has consented. Nothing self-heals it, because insertOnly is exactly what stops boot rewriting operator-owned rows.

The note gives the procedure rather than only the warning: update the existing `oauthResource.identifier` in place — the FK from `oauthClientResource.resourceId` is `ON UPDATE CASCADE`, verified in the migration, so link rows follow — instead of letting a second row appear. The alternative, asking every MCP user to re-register, is the thing the ENG-2343 backfill exists to avoid.

apps/web/modules/auth/lib/mcp-oauth-dcr.test.ts
```diff
@@ -101,7 +122,63 @@ const requestAuthorize = async (
...
+          response_types: ["code"],
+          token_endpoint_auth_method: "none",
+          scope: "surveys:read",
+        }),
+      })
+    );
+    const body = (await response.json()) as { error?: string };
+
+    expect(response.status).toBe(400);
+    expect(body.error).toBe("invalid_redirect_uri");
```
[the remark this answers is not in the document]
xernobyl: Not taking this one, and the reason is that the premise does not hold against the installed package.

The quoted web result says 1.7 "may default to native in certain contexts (such as legacy or MCP-related flows)". It does not. In `@better-auth/oauth-provider@1.7.0` the DCR path is `applyOAuthClientRegistrationDefaults(client, isClientMetadataDocument ? null : "web")` — dynamic and administrative registrations default to `"web"`, and only a Client ID Metadata Document preserves an omitted value as `null`. I extracted that function and `validateClientRedirectUri` from both `1.7.0-rc.5` and `1.7.0` and diffed them: byte-identical, so this is not an rc-versus-stable difference either. The test pins real behaviour.

The suggested fix is also a security change rather than a test change. Normalizing an omitted `application_type` to `"native"` is what makes loopback redirect URIs legal, and `web`-versus-`native` is precisely the distinction RFC 8252 and OIDC DCR draw to keep a browser-based client from registering `http://127.0.0.1`. Silently reclassifying every unlabeled DCR client as native to permit that is a real weakening, and 1.7 exposes no option to scope it to MCP — the only registration knobs are `clientRegistrationDefaultResources` and `clientRegistrationDefaultScopes`. It would have to be our own interception of the DCR endpoint, applied to all clients.

The underlying operational risk is real, though, and is already tracked on this PR rather than dismissed: whether real MCP clients send `application_type: "native"` is listed under **Open gaps**, precisely because only staging answers it. If Claude Code or MCP Inspector turn out to omit it, the fix is a scoped normalization with a written security rationale — not a default flipped now on a search summary the code contradicts. The sibling test already pins the refusal so the behaviour is visible either way.

mattinannt (commented):
@xernobyl Let's discuss: We just changed the URLs in 5.2 ([Migration guide](https://formbricks.com/docs/self-hosting/advanced/migration#v5-2)) from `/api/auth/callback/{provider}` to `/api/auth/oauth2/callback/{providerId}` and made every self-hoster (+ Formbricks Cloud) update every SSO provider configuration.
Now we change back. Why this back and forth?

xernobyl:
@mattinannt when we were on NextAuth, the callback path was /api/auth/callback/{provider}. When we migrated to Better Auth (1.6) on Formbricks 5.2, we used its genericOAuth plugin, which mounts its own callback route at /api/auth/oauth2/callback/{providerId}.

In 1.7 they rebuilt genericOAuth on top of Better Auth's core social-provider path, so it now shares the built-in callback route and the path goes back to /api/auth/callback/{providerId}. So the change comes from there, rather than from a choice on our side.

There is an override setting, that originally I decided against using it because I got under the impression that it could be removed (maybe I saw something about the plugin being merged and got that confused at the time), so we would have to change the path at some point in the future, so it felt like it was better to do it then. The setting it's still there in 1.7, so use can use it to keep the current path without requiring change from our customers.

Working on it.

xernobyl:
I addressed the breaking callback URL, stays `/api/auth/oauth2/callback/{providerId}`, so nothing changes for self-hosters.

docs/self-hosting/advanced/migration.mdx
```diff
@@ -73,6 +77,54 @@ This change adds no migrations and touches no schema, so rolling the application
 to v5.3 undoes it with nothing to restore. An API key raised to `manage` in step 2 keeps working on v5.3 —
 `manage` was always accepted wherever `write` was — so that step does not need reverting either.
 
+### Better Auth 1.7 Upgrade (no action required)
```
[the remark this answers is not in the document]
xernobyl: Fixed in 53bc2eea2 — the three headings this PR owns are now sentence case with the product/feature names kept: `Better Auth 1.7 upgrade (no action required)`, `SSO callback URLs are unchanged`, `MCP access tokens now end with their session`.

Left the surrounding v5.2/v5.3 headings alone: they use title case too, but they are not this PR's to rename and doing so would bury the actual diff.

xernobyl:
Re-reviewed the callback pin end to end after CodeRabbit's heading fix. One finding was mine and worth calling out.

**`signal` was never committed.** The commit message on `8f939ffa4` says the rewrite carries the abort signal through; `git log -S"signal: request.signal"` returns nothing — the edit was lost to a file restore during mutation testing and I wrote the message without re-reading the diff. Effect was narrow but real: on the pinned callback path a client disconnect stopped aborting `auth.handler` and its outbound IdP calls, while the pass-through path kept the original request object and did abort. Fixed in `1541eae9d`, with a behavioural test (abort the original, assert the mapped one aborts) rather than an identity check, since the spec lets an implementation wrap the signal.

Two guard corrections in the same commit:

- Non-http(s) URLs are now rejected. On a cannot-be-a-base URL the `pathname` setter is a silent no-op, so `data:text/plain,/api/auth/oauth2/callback/openid` came back unchanged but non-null — a non-rewrite escaping as a rewrite. Unreachable through Next, but the docblock asserts the invariant unconditionally.
- The crafted-prefix guard matched `/api/auth` without a trailing slash, so a legitimate `basePath` merely starting with it (`/api/authority`) would have silently 404'd SSO.

And one test that could not fail: *"still maps inside the SSO request context"* asserted only ordering, making it a duplicate of the test above it — green with the mapper call deleted. It now asserts the mapped URL; removing the mapper fails three route tests instead of two. All three new guards were mutation-tested individually.

Separately, on the earlier `at+jwt` rolling-deploy note: keeping the check strict, and `53bc2eea2` records why. A 1.6-minted token (no `typ`) hitting a 1.7 pod is rejected — but the check is on the resource server only, not the refresh path, and `20260812110001_eng_2343_backfill_oauth_resource_links` backfills `oauthRefreshToken.resources` so existing refresh tokens keep working. A client takes one 401, refreshes, retries with a typed token: self-healing in a round trip, using the 401 handling MCP clients already implement. Relaxing it to "absent is fine" would permanently weaken a cross-JWT-confusion defence to smooth a window that closes on its own.

Gates: 630 files / 8341 tests, typecheck, lint, format all green.

mattinannt (dismissed):

packages/database/migration/20260812110001_eng_2343_backfill_oauth_resource_links/migration.ts
```diff
@@ -0,0 +1,143 @@
...
+      );
+    }
+
+    // Matches what the plugin would seed from its `resources` option, so `insertOnly` seeding at boot
+    // is a no-op afterwards. `name` is descriptive only.
+    await migrationTx.$executeRaw`
+      INSERT INTO "oauthResource" ("id", "identifier", "name", "createdAt", "updatedAt")
+      VALUES (${createId()}, ${resourceIdentifier}, 'Formbricks MCP', NOW(), NOW())
+      ON CONFLICT ("identifier") DO NOTHING
+    `;
```
kiln-agent, line 130: **The seeded `oauthResource` row has no `allowedScopes`, which breaks MCP OAuth on exactly the instances this migration exists for.**

This INSERT omits `allowedScopes`, so the column takes its DDL default from the sibling schema migration (`"allowedScopes" TEXT[] DEFAULT ARRAY[]::TEXT[]`) — an **empty array**, not NULL. The plugin's own seeder writes the configured list instead (`buildSeedRow`: `allowedScopes: input.allowedScopes ?? null`, and `mcp-oauth-provider-options.ts` sets it to the full `MCP_OAUTH_SCOPES`).

Empty array and NULL are not equivalent to the resolver. `resolveResourcePolicy` (`@better-auth/oauth-provider/dist/introspect-*.mjs`):

```js
for (const row of resolved) {
  if (row.allowedScopes === null || row.allowedScopes === void 0) continue;
  const allowed = new Set(row.allowedScopes);
  const intersection = effectiveScopes.filter((s) => allowed.has(s));
  if (intersection.length === 0) throw new APIError("BAD_REQUEST", {
    error: "invalid_scope", ... });
  effectiveScopes = intersection;
}
```

`[]` does not `continue` — it intersects to nothing and throws.

And because `resourceSeedMode: "insertOnly"` is pinned, boot never repairs it: `seedResources` finds the existing row and `if (mode === "insertOnly") continue`. The empty list is permanent.

Concrete scenario — an instance upgrading to v5.4 that already has ≥1 `oauthClient` row (the only case where this migration does anything at all):

1. Migration inserts `oauthResource(identifier = WEBAPP_URL + /api/mcp, allowedScopes = {})`.
2. App boots, `insertOnly` skips the row.
3. An MCP client hits `/api/auth/oauth2/authorize?...&resource=<identifier>` → `resolveResourcePolicy` → `400 invalid_scope` ("none of the requested scopes are allowed for resource …"), redirected to the client.
4. The refresh path fails the same way (`resolveResourceGrantIssuance` → same resolver with the backfilled `oauthRefreshToken.resources`), so the "one 401 then self-heal via refresh" story in `modules/mcp/auth.ts` for the `at+jwt` enforcement cannot complete either.

MCP OAuth is down for every pre-existing client, which is the opposite of what the backfill is for. The live handshake in the PR description was run "on a database migrated **from empty**" — that path has `clients === 0`, returns early, and lets the plugin seed the row correctly, so it cannot catch this.

```suggestion
    // Matches what the plugin would seed from its `resources` option, so `insertOnly` seeding at boot
    // is a no-op afterwards. `allowedScopes` is NOT optional here: the column defaults to an empty
    // array, and an empty (as opposed to NULL) allow-list intersects every token request down to zero
    // scopes -> `invalid_scope`. `insertOnly` means boot would never repair it. `name` is descriptive only.
    await migrationTx.$executeRaw`
      INSERT INTO "oauthResource" ("id", "identifier", "name", "allowedScopes", "createdAt", "updatedAt")
      VALUES (${createId()}, ${resourceIdentifier}, 'Formbricks MCP', ${MCP_OAUTH_SCOPES}::TEXT[], NOW(), NOW())
      ON CONFLICT ("identifier") DO NOTHING
    `;
```

(The scope list has to be a local literal in this file — `packages/database` cannot import `apps/web` — same treatment as `resolveMcpResourceIdentifier` above, with a test pinning it against `MCP_OAUTH_SCOPES`. Setting it to NULL would also work and would inherit the plugin default, but it diverges from what a fresh install gets.)

---

packages/database/schema/main.prisma
```diff
@@ -850,7 +850,22 @@ model Account {
...
+  /// Better Auth 1.7 (ENG-2343): the account's identity namespace. 1.7 keys accounts on
+  /// (issuer, accountId) and filters EVERY lookup on it — findAccountByKey, findAccountOwnerByKey,
+  /// findCredentialAccount and updatePassword. A null here is not cosmetic: password sign-in stops
+  /// finding the credential row (rejects as if the password were wrong) and a password reset matches
+  /// zero rows, so it silently succeeds while changing nothing.
+  ///
+  /// NULLABLE on purpose, even though better-auth declares it required. A NOT NULL column would fail
+  /// every INSERT from a 1.6 pod still serving during a rolling deploy. The migration backfills every
+  /// existing row; see the migration SQL for the value derivation.
+  issuer String?
```
kiln-agent, line 862: **SSO recovery still writes `Account` rows with a NULL `issuer`, so a recovered link is invisible to 1.7 and the user is pushed back through recovery on every subsequent sign-in.**

Not in this diff, which is why I'm anchoring it here: `apps/web/modules/ee/sso/lib/account-linking.ts` is the one place in app code that creates/updates an SSO `Account` row itself, and it was not updated for `(issuer, accountId)`:

- `account-linking.ts:124` — `tx.account.create({ data: { userId, type, provider, providerAccountId, ...tokens } })`
- `account-linking.ts:96` and `:106` — the legacy-normalize `tx.account.update` branches, same omission

`issuer` is `String?`, so nothing complains at compile time. The only caller is `syncSsoIdentityForUser` from `sso-recovery.ts:323`, i.e. the verify-before-link completion path.

Concrete scenario, OIDC:

1. Existing local user `a@x.com` signs in with OIDC → `handleOAuthUserInfo` finds no linked account, `accountLinking.enabled: false` → `"account not linked"` → our recovery flow starts.
2. User verifies the email, recovery completes, `syncSsoIdentityForUser` inserts `Account(provider='openid', providerAccountId=<sub>, issuer=NULL)`. They get a session, so this looks like it worked.
3. Next sign-in: the callback resolves the account key to `{ issuer: 'local:oauth:openid', accountId: <sub> }` and `handleOAuthUserInfo` does

   ```js
   dbUser.accounts.find((acc) => acc.issuer === account.issuer && acc.accountId === account.accountId)
   ```

   `NULL !== 'local:oauth:openid'` → no match → `"account not linked"` again → recovery again.

So recovery never actually sticks after this upgrade: every OIDC/Azure/SAML sign-in for a recovered user requires a fresh email verification. The migration backfill does not save them either — it runs once, before the row exists.

Fix: have `syncSsoIdentityForUser` set `issuer` from the same synthetic form the migration and `better-auth-providers.ts` agree on (`local:oauth:<provider>`), on both the `create` and the two `update` branches. Worth noting that `ssoAccountIssuer` in `better-auth-providers.ts` is exactly this value and is already described as "load-bearing … byte-identical to what the migration writes" — this is a third site that has to agree with it, so it probably wants to be one exported helper rather than a fourth literal.

---

apps/web/playwright/fixtures/users.ts
```diff
@@ -103,12 +104,21 @@ export const createUsersFixture = (page: Page, workerInfo: TestInfo): UsersFixtu
...
+      // sign-in silently returns INVALID_EMAIL_OR_PASSWORD (no session, no cookie) for a correct
+      // password, and every helper that expects a post-login redirect (loginAndGetApiKey et al.) times
+      // out waiting for a navigation that never happens.
       await prisma.account.create({
         data: {
           userId: user.id,
           type: "credential",
           provider: "credential",
           providerAccountId: user.id,
+          issuer: createLocalAccountIssuer("credential"),
```
kiln-agent, line 121: **The same fix is missing in `packages/database/src/seed.ts` — every `db:seed` user loses the ability to sign in.**

This fixture is right, and the reasoning in the comment above applies verbatim to the dev/demo seed, which was not touched:

```ts
// packages/database/src/seed.ts:769
await prisma.account.upsert({
  where: { provider_providerAccountId: { provider: "credential", providerAccountId: id } },
  update: { password: passwordHash },
  create: {
    userId: id,
    type: "credential",
    provider: "credential",
    providerAccountId: id,
    password: passwordHash,
    // no issuer
  },
});
```

`sign-in.mjs` filters on it exactly as described here:

```js
const credentialIssuer = createLocalAccountIssuer("credential");
const credentialAccount = userRecord?.accounts.find((a) =>
  a.providerId === "credential" && a.issuer === credentialIssuer && a.accountId === userRecord.user.id);
if (!userRecord || !credentialAccount) { ... throw INVALID_EMAIL_OR_PASSWORD }
```

Concrete scenario: `pnpm db:up && pnpm db:migrate:dev && pnpm db:seed`, then log in as a seeded user with the correct seeded password → `INVALID_EMAIL_OR_PASSWORD`, no session. `updatePassword` filters on the same tuple, so the password-reset escape hatch silently updates zero rows too. Every fresh dev environment and any self-hoster who used the seed hits this.

```suggestion
          issuer: createLocalAccountIssuer("credential"),
```
(and the same line added to the `create` block in `packages/database/src/seed.ts`, imported from `@better-auth/core/db` as here — or a local `"local:credential"` literal if `packages/database` should not take that dependency.)

---

apps/web/modules/ee/sso/lib/better-auth-providers.ts
```diff
@@ -92,26 +184,26 @@ export const ssoGenericOAuthConfig: GenericOAuthConfig[] = ENTERPRISE_LICENSE_KE
               clientSecret: AZUREAD_CLIENT_SECRET ?? "",
               discoveryUrl: `https://login.microsoftonline.com/${AZUREAD_TENANT_ID || "common"}/v2.0/.well-known/openid-configuration`,
```
kiln-agent: **`AZUREAD_TENANT_ID` unset → `common` → 1.7 will reject every Azure id_token, because the multi-tenant discovery issuer is a `{tenantid}` placeholder.**

`AZURE_OAUTH_ENABLED` is `!!(AZUREAD_CLIENT_ID && AZUREAD_CLIENT_SECRET)` (`lib/constants.ts:37`) and `AZUREAD_TENANT_ID` is `z.string().optional()`, so `|| "common"` is a reachable production configuration.

What changes in 1.7 is that `genericOAuth` now builds an id_token verification config from the discovery document:

```js
if (discovered.jwks_uri && discovered.issuer) {
  idTokenConfig = { jwks: createRemoteJWKSet(jwksUrl), issuer: discovered.issuer, audience: c.clientId, ... };
}
...
requiresIdTokenNonce: idTokenConfig !== void 0 && c.disableIdTokenNonceBinding !== true,
```

and `getUserInfo` hard-fails on it:

```js
if (oauthTokens.idToken && provider.idToken) {
  if (!await verifyProviderIdToken(provider, oauthTokens.idToken, expectedIdTokenNonce)) { ...; return null; }
}
```

`verifyProviderIdToken` passes `issuer: config.issuer` into jose's `jwtVerify`. Microsoft's `common` (and `organizations`/`consumers`) discovery document returns the literal `"issuer": "https://login.microsoftonline.com/{tenantid}/v2.0"`, while the real id_token carries the concrete tenant GUID. jose raises `JWTClaimValidationFailed`, `verifyProviderIdToken` returns `false`, `getUserInfo` returns `null`, and the callback redirects with `FAILED_TO_GET_USER_INFO`.

Upstream treats this as a hard error rather than a subtlety — `better-auth/dist/plugins/generic-oauth/providers/microsoft-entra-id.mjs` refuses to construct at all:

```js
if (!/^[0-9a-f]{8}-...-[0-9a-f]{12}$/.test(tenantId))
  throw new Error("The generic microsoftEntraId helper requires a concrete Microsoft Entra tenant GUID. Use the built-in Microsoft provider for common, organizations, or consumers.");
```

Concrete scenario: a self-hoster with `AZUREAD_CLIENT_ID`/`SECRET` set and no `AZUREAD_TENANT_ID` (works on v5.3) upgrades to v5.4 → every Azure sign-in dies after the token exchange with a generic "failed to get user info", including for users who signed in yesterday.

Flagging as **plausible** rather than confirmed since I could not diff 1.6's generic-oauth id_token handling offline — it is possible 1.6 was already doing this — but the PR's own Open gaps say no SSO handshake is tested anywhere, so nothing in CI would tell you. Worth resolving one of three ways before merge: make the single-tenant GUID a hard requirement when Azure is enabled, set `disableIdTokenSignIn`/a custom `verifyIdToken` for the multi-tenant case, or move `azuread` to the built-in Microsoft social provider (which handles `common` and declares `accountIssuer: ({ profile }) => profile.iss` — note that would also change the `Account.issuer` value the backfill has to write, so it is not a drop-in).

---

apps/web/modules/mcp/auth.ts
```diff
@@ -395,10 +401,28 @@ async function authenticateMcpOAuthBearer(
   let payload: JWTPayload;
 
   try {
-    payload = await oauthResourceClient.getActions().verifyAccessToken(token, {
+    // Renamed from `verifyAccessToken` in Better Auth 1.7 (ENG-2343). 1.7 also stops passing
+    // `verifyOptions.audience` into its own jwtVerify, validating instead that each `aud` value
+    // resolves to a registered resource — which is not the same question as "is this token for ME".
```
kiln-agent: **The stated reason for keeping `hasAcceptedMcpAudience` is factually wrong: 1.7 does still pass `verifyOptions.audience` into jose.**

`@better-auth/core/dist/oauth2/verify.mjs`:

```js
async function verifyAccessTokenPayload(token, opts) {
  if (opts.jwksUrl && !opts?.remoteVerify?.force) try {
    payload = await verifyJwsAccessToken(token, { jwksFetch: opts.jwksUrl, verifyOptions: opts.verifyOptions });
  ...
async function verifyJwsAccessToken(token, opts) {
  jwt = await jwtVerify(token, createLocalJWKSet(resolved.jwks), opts.verifyOptions);
```

`opts.verifyOptions` — audience included — goes straight to `jwtVerify`, and the type even makes it mandatory (`JWTVerifyOptions & Required<Pick<JWTVerifyOptions, "audience" | "issuer">>`).

The conclusion ("keep both") is still the right one, for the ENG-2339 reason: jose's `aud` check is a *membership* test, so a multi-audience token passes it, and `hasAcceptedMcpAudience` is the allow-list that rejects it. But the justification as written is load-bearing in two places — here and in `pnpm-workspace.yaml` ("1.7 drops `verifyOptions.audience` from its own jwtVerify, so it is now the only audience check on that path") — and the next reader who checks it against the source will find it false and may conclude the whole note is stale. Worth restating as the membership-vs-allow-list argument in both spots.

---

apps/web/modules/ee/sso/lib/better-auth-providers.ts
```diff
@@ -82,6 +85,95 @@ export const ssoSocialProviders = ENTERPRISE_LICENSE_KEY
...
+ * Better Auth 1.7 keys accounts on (issuer, accountId). Left to itself a provider with a
+ * `discoveryUrl` adopts the DISCOVERED issuer, which is tenant-specific — different for every
+ * self-hoster, and therefore impossible to reproduce in a portable backfill. Pinning the synthetic
+ * form Better Auth itself uses for providers without their own issuer (`local:oauth:<id>`) keeps
+ * identity scoped to the provider id, exactly as 1.6 keyed it, so existing accounts keep matching
+ * after the upgrade and the migration backfill is one portable UPDATE.
+ *
+ * This value is load-bearing: it must stay byte-identical to what
+ * migration/20260812110000_eng_2343_better_auth_17_resource_model writes into Account.issuer.
+ */
```
kiln-agent: Orphaned docblock: this one documents `ssoAccountIssuer` (declared 34 lines below, with its own, longer docblock) but sits immediately above `toAccountSubject`, which has its own docblock right after it. Two stacked `/** */` blocks before one declaration — the first is dead and misattributes the "account-identity namespace" description to the subject coercion helper. Looks like a leftover from moving `ssoAccountIssuer` down; safe to delete.

---

kiln-agent (commented):
Correctness review of the Better Auth 1.7 upgrade (effort: high). The routing/pin half holds up — I traced `options.redirectURI || redirectURI` through both `create-authorization-url.mjs` and `validate-authorization-code.mjs`, confirmed `/callback/:id` is the only mounted callback in 1.7, and confirmed the oauth-provider declares no `/oauth2/callback/*` route the mapper could shadow. `Account.issuer` values check out against upstream (`createLocalAccountIssuer`/`createOAuthAccountIssuer`, google's declared `https://accounts.google.com`, github's absence), and the new admin resource endpoints are all `SERVER_ONLY` so they add no HTTP surface.

Six findings, most-severe first. Two are blocking in my reading:

1. **`oauthResource` seeded without `allowedScopes`** (data migration) — the DDL default is `{}`, not NULL, and an empty allow-list intersects every token request down to zero scopes → `invalid_scope`. `resourceSeedMode: "insertOnly"` means boot never repairs it. Breaks MCP OAuth on every instance that already has clients, i.e. exactly the instances the backfill targets. The end-to-end handshake in the description was run against a DB migrated from empty, which takes the `clients === 0` early return and so cannot reach this.
2. **SSO recovery writes `Account.issuer = NULL`** (`account-linking.ts`, not in the diff) — 1.7's `(issuer, accountId)` match fails, so a recovered link is invisible and every later SSO sign-in re-enters recovery.
3. **`packages/database/src/seed.ts` missing `issuer`** — the same bug the playwright fixture fixed; `db:seed` users can no longer sign in, and their password reset matches zero rows.
4. **Azure with `AZUREAD_TENANT_ID` unset** (`common`) — 1.7 verifies the id_token against the discovery issuer, which for `common` is the literal `{tenantid}` placeholder. Marked plausible, not confirmed; upstream's own `microsoftEntraId` helper throws rather than allow a non-GUID tenant, which is suggestive. The Open gaps note that no SSO handshake is tested, so CI will not answer this.
5. The "1.7 drops `verifyOptions.audience` from its own `jwtVerify`" justification (here and in `pnpm-workspace.yaml`) is wrong — it is still passed. The conclusion to keep the explicit check is right for a different reason.
6. An orphaned docblock in `better-auth-providers.ts`.

Details, sources and suggested fixes are inline.

---

xernobyl:
Addressed all six review findings in `87c5ed02b`. I verified each against real code before touching anything; four were real, two of those were release blockers, and one of them invalidates a claim I made earlier in this PR.

## Blocker — the resource backfill seeded an empty scope allow-list

Confirmed at every link. The INSERT omitted `allowedScopes`, so it took the DDL default `ARRAY[]::TEXT[]`; `resolveResourcePolicy` skips only NULL/undefined, so an empty array intersects every request to zero scopes and throws `invalid_scope`; and `resourceSeedMode: "insertOnly"` means boot never repairs the row. MCP OAuth would have been **permanently** down for exactly the instances this migration exists to protect — those with at least one pre-existing `oauthClient`.

Proven against a fully migrated Postgres, old shape vs new:

| row | shape | result |
| --- | --- | --- |
| `r_old` | what shipped | `cardinality = 0`, `is_null = false` |
| `r_new` | the fix | `cardinality = 10` |

**And it retracts something I claimed.** The live MCP handshake in the description was run on a database migrated *from empty* — where the migration returns early at `clients === 0`. So that test passed **because it took the path that skips this bug**; it could never have caught it. I have added a from-a-pre-existing-client check, which is the case that matters.

Worth noting the docblock already said "insertOnly, so the plugin will not fight this row" — it had the right mechanism and drew the opposite conclusion from it.

## Blocker — Azure multi-tenant would have rejected every sign-in

Verified against Microsoft's live endpoint: `common` discovery advertises `issuer: "https://login.microsoftonline.com/{tenantid}/v2.0"` — a documented **template** — while 1.7 compares `iss` literally whenever discovery yields both `issuer` and `jwks_uri`. Every real id_token carries the tenant GUID, so all of them fail. 1.6 never parsed the id_token, so this is a regression from the upgrade, and it bites the documented default (`AZUREAD_TENANT_ID` unset).

Fixed so **no self-hoster has to change anything**, which ruled out the obvious option:

- **Concrete `AZUREAD_TENANT_ID`** → keeps `discoveryUrl`. Issuer is a real value, so the id_token *is* verified — stronger than 1.6.
- **`common`** → explicit endpoints, no discovery, so no `idTokenConfig` is built and identity comes from UserInfo over a client-authenticated back-channel call: the 1.6 posture exactly.

Making the tenant mandatory was rejected deliberately — it demands action from every self-hoster who has not set it, and drops multi-tenant app registrations, which have no single issuer by construction. Microsoft's own guidance is to substitute the token's `tid` and validate that; Better Auth 1.7 cannot express a dynamic issuer, so the built-in check is not merely strict here, it is unimplementable.

## Two more real ones

**`account-linking.ts`** wrote `Account` rows with a NULL `issuer` on all three write paths. Recovery appeared to work, then the next sign-in did not match on `(issuer, accountId)` and re-triggered recovery — every time. The migration cannot help; it runs before that row exists.

**`packages/database/src/seed.ts`** had the same omission, so every `db:seed` user got INVALID_EMAIL_OR_PASSWORD with the correct password, and the reset path matched zero rows.

`ssoAccountIssuer` now lives in `modules/ee/sso/lib/constants.ts` — dependency-free, and already imported by both call sites — since three places must agree byte-for-byte and the third is a SQL literal that cannot call TypeScript.

## The other two

The orphaned docblock is gone (my consolidation made it worse before it made it better). And the `verifyOptions.audience` claim in `mcp/auth.ts` was right to flag: I could not substantiate it — `verifyBearerToken` is re-exported through `better-auth/oauth2` and its body is not readable in the published dist — so the comment now gives the reason that *is* checkable, namely that "every `aud` resolves to a registered resource" and "this token is for me" are different questions.

## Coverage

A drift guard pinning the migration's scope literal against `MCP_OAUTH_SCOPES` (mutation-tested: dropping one scope fails it), both Azure strategies asserted, and the three account-linking write paths. The existing account-linking tests caught all three of my changes before I updated them, which is what I want from them.

632 files / 8351 tests, typecheck 27/27, lint and format clean.

Two incidental notes: the `WEBAPP_URL` lint error and a dead `no-console` directive surfaced only through lint-staged, because the database package's lint script covers `./src` — so nothing lints `./migration`. And the earlier E2E red was a 15-minute action timeout with all 107 tests passing, not a failure; it is green on re-run.

apps/web/modules/mcp/auth.ts
xernobyl: Fixed in `87c5ed02b`, and you were right to flag it. I tried to verify the claim either way and could not: `verifyBearerToken` is re-exported through `better-auth/oauth2`, and the symbol does not appear anywhere in the published `better-auth` dist, so its body is not readable — even though it demonstrably works at runtime.

Rather than swap one unverifiable assertion for another, the comment now gives the reason that is actually checkable: "every `aud` resolves to a registered resource" and "this token is for ME" are different questions, and only the second is the one a resource server must answer. `hasAcceptedMcpAudience` answers it, so it stays regardless of what upstream does with `verifyOptions.audience`. Behaviour unchanged — this was only ever a wrong justification for correct code.

apps/web/modules/ee/sso/lib/better-auth-providers.ts
xernobyl: Fixed in `87c5ed02b` — deleted, not moved. Its content now lives on the shared helper in `modules/ee/sso/lib/constants.ts`, because the same finding turned out to matter in a second place: `account-linking.ts` was writing `Account` rows with a NULL `issuer`, making it a third site that has to produce a byte-identical value. So the docblock and the helper both moved somewhere both call sites already reach, and this file imports it.

Worth noting my consolidation briefly made the orphan worse before removing it, which is exactly why this was worth flagging.

xernobyl:
Another review pass over the branch, and this one was worth doing — it turned up two more upgrade blockers, both live, plus a handful of smaller things. All addressed in the three commits above.

### Blockers

**Azure sign-in was broken by my own previous fix.** 1.7's generic-OAuth resolves the account subject as `isOidc ? profile.sub : profile.id`, and `isOidc` is only ever set inside the discovery branch. The id_token fix in the commit before this had to configure Azure `common` with explicit endpoints instead of discovery — which silently flipped subject resolution to `profile.id`, a field Microsoft Graph's userinfo does not return. Every Azure sign-in would have ended at `error=unable_to_get_user_info`.

`accountSubject` is now pinned per provider to the field that provider actually sends (azuread/openid `sub`, the SAML bridge `id`), so identity no longer depends on whether discovery happened to run.

I could reproduce this end to end locally, because a discovery document without `jwks_uri` produces the same `isOidc = false` state as the Azure `common` path. Against a running instance with a stub OIDC provider:

| | result |
| --- | --- |
| with `accountSubject` pinned | lands on `/`, `Account` row written with `providerAccountId = stub-subject-1` (the `sub`) |
| pin removed, nothing else changed | `→ /auth/login?error=unable_to_get_user_info` |

**Loopback DCR was refused.** 1.7 added `validateClientRedirectUri` and defaults `application_type` to `"web"` — as a literal at the call site, not an option we can set. A `web` client is refused any loopback redirect, which is exactly what a local MCP client registers, so a registration omitting `application_type` (MCP Inspector's shape) would 400 before consent. Registrations that omit the field and ask for a loopback redirect are now labelled `native` per RFC 8252 §7.3 before Better Auth validates them; an explicit `application_type` is never touched.

The third commit narrows this further: my first cut required *every* redirect URI to be loopback, which still broke a native client pairing a loopback URI with an https one. Widening to "at least one" can't widen what upstream accepts, and I checked that rather than assuming — upstream refuses a non-loopback http redirect under `native` too, so being labelled native is never a route to registering one.

Live matrix on the running app:

| registration | result |
| --- | --- |
| loopback only, no `application_type` | ✅ `native` |
| loopback + https, no `application_type` | ✅ `native` |
| https only, no `application_type` | ✅ `web` |
| loopback + non-loopback http | ❌ `invalid_redirect_uri` |
| explicit `web` + loopback | ❌ `invalid_redirect_uri` (not overridden) |
| explicit `native` | ✅ preserved |

And a registered loopback client is usable, not just registrable: `/authorize` returns a 302 to login carrying the resource param, no `invalid_client`.

### Smaller items

- The schema contract test only read the *plugins'* declared schemas — so it could never have caught the bug it was written for, since `Account.issuer` is a core field no plugin declares. It now also resolves the merged core schema through upstream's own `getAuthTables`. Mutation-tested: dropping `issuer` from the Prisma model fails two tests. `user.image` is excluded because we dropped that column, and the exclusion is tied to the SSO hook that strips the field, so removing the strip fails the suite too.
- `lib/account/service.ts` created and upserted `Account` rows with no `issuer`, by mass-assigning the whole input. It had no callers left, and it couldn't be fixed in place — a generic helper can't know whether a row wants the credential or the OAuth issuer. Deleted.
- Retracted a claim about upstream dropping `verifyOptions.audience` that I couldn't substantiate on a second look. I'd already retracted it in `modules/mcp/auth.ts` and missed the copy in `pnpm-workspace.yaml`. The reasoning it supported still stands on its own.
- Migration guide now notes the SSO callback is derived from the auth base URL, not `WEBAPP_URL`, which matters for anyone who sets `BETTER_AUTH_URL` to a different origin.

### Still true

No action required for self-hosters. The SSO callback URLs are the v5.2 ones, verified live — the IdP receives `http://localhost:3100/api/auth/oauth2/callback/openid` on both the authorization and token legs.

Gates: 8391 tests, typecheck 27/27, 0 lint errors, format clean.

@mattinannt this needs a re-approval when you get a chance — sorry, the push dismissed yours.

xernobyl:
Follow-up: I'd listed "SSO sign-in as a pre-upgrade user" as staging-only, and that was wrong — the stub IdP makes it reproducible locally, so I've now done it rather than deferring it. This is the case the risk section leads with, so it seemed worth proving.

Seeded a user and an `openid` `Account` exactly as a 1.6 pod leaves them — `issuer` NULL, `access_token = 'old-token'` — ran the migration's own backfill statement verbatim, then signed in through the stub with the same `sub`:

| | result |
| --- | --- |
| after the backfill | existing rows reused — same `userId`, same `Account.id`, still 1 user / 1 account, `access_token` updated to the fresh value |
| `issuer` set back to NULL (backfill skipped) | no duplicate, no error — routes into verify-before-link recovery, `/auth/verification-requested?…&purpose=sso_recovery` |
| backfill restored | clean sign-in again |

The updated `access_token` is the part that matters: it proves 1.7 **found and updated** the backfilled row rather than inserting alongside it, which a row count alone wouldn't show.

**This corrects the risk note in the description.** I had said a missing `issuer` "locks password users out *and* makes their reset silently match zero rows" — true for **credential** accounts, and that stays the dangerous half, since there's no self-service way back. But SSO accounts degrade gracefully into the existing recovery flow. So the deploy check is: password login first; SSO is the more forgiving surface, not the more fragile one.

What a stub still can't cover, and what I'd genuinely like staging for: **Azure AD against real Entra** — its userinfo shape is exactly what broke, and `AZUREAD_TENANT_ID` set vs unset take different code paths — and the **SAML bridge through BoxyHQ**. Narrowed the open gap to just those two.

xernobyl:
Added the ENG-2349 cooldown restore here (`6e89625ba`) now that the window has passed — 1.7.0 turned 72h old at 00:07Z today, so the nine-package `minimumReleaseAgeExclude` exemption is no longer doing anything and is removed. `@formbricks/hub` stays, for the unrelated first-party-SDK reason its own note gives.

Verified rather than assumed: `pnpm install` with the exemption gone reports `✓ Lockfile passes supply-chain policies (2337 entries in 16.8s)`, and **`pnpm-lock.yaml` is byte-identical** — the same 1.7.0 resolves, the waiver simply stopped being load-bearing. Typecheck 27/27, 8391 tests, lint 0 errors, format clean.

Folded in here rather than opened as its own PR for two reasons: it would otherwise race this branch on the same file, and those entries matched by **name, not version** — so for as long as they stood, any later Better Auth release (1.7.1, 1.8.0) would have bypassed the 72h cooldown silently, with nothing in CI to say so. Worth closing on the day rather than at leisure.

No behaviour change, and it does not affect anything already reviewed here.

xernobyl:
Merged `origin/main` (12 commits) and resolved the conflicts. Re-reviewed and re-smoke-tested — nothing broke, but the resolution had one trap worth writing down.

## What conflicted, and how it was resolved

main migrated most of `apps/web/package.json` to pnpm **`catalog:`** references. It left the Better Auth packages as literal versions, so the two sides only collided on adjacent lines: main's catalog refs are kept, our `1.7.0` pins are kept, and `@better-auth/utils` stays at `0.4.2` unmoved.

**The trap:** taking main's `pnpm-lock.yaml` wholesale produces a lockfile resolving `better-auth: 1.6.23` against a `package.json` asking for `1.7.0` — a pair `pnpm install --frozen-lockfile` rejects, which is exactly what CI runs. Caught by diffing the committed lockfile rather than the working tree, and fixed in its own commit. `--frozen-lockfile` now exits 0.

Also content-verified that the merge did not quietly revert anything of ours — `accountSubject`, `ssoLegacyRedirectUri`, the DCR normalizer, `legacy-sso-callback.ts`, `ssoAccountIssuer`, the `getAuthTables` contract test, the migration's scope literal, and the ENG-2349 cooldown removal (`pnpm-workspace.yaml` auto-merged, so that one especially wanted checking) are all still present.

## Smoke test on the merged result

The merge changes how nearly every dependency resolves, so this re-ran the flows that matter rather than trusting the suite:

| check | result |
| --- | --- |
| App boots on the merged tree | ✅ catalog migration didn't break resolution |
| SSO: pinned callback URL | `redirect_uri = http://localhost:3100/api/auth/oauth2/callback/openid` — the v5.2 URL, with PKCE |
| SSO: **both** OAuth legs agree | IdP log shows the identical pinned URL on `AUTHORIZE` **and** `TOKEN` — the RFC 6749 exact-match requirement |
| SSO: full callback | lands on `/`; `Account` row `openid / stub-subject-1 / local:oauth:openid`, so `accountSubject` and `ssoAccountIssuer` both still hold |
| DCR: loopback only / loopback+https / https only | `native` / `native` / `web` ✅ |
| DCR: loopback + non-loopback http | `invalid_redirect_uri` ✅ still refused |
| DCR: explicit `web` + loopback | refused, not overridden ✅ |
| MCP PRM | `resource = http://localhost:3100/api/mcp` ✅ |
| Mapper doesn't shadow sibling `/oauth2/*` | DCR POSTs to `/api/auth/oauth2/register` still reach upstream ✅ |

Gates on the merged tree: typecheck 27/27, lint 0 errors, format clean, 8488 tests passing.

## One local-only test failure, not from this PR

`modules/ee/analysis/dashboards/lib/widget-view.test.ts` fails 8/8 on my machine after the merge — `window.localStorage` undefined. It is **not** from this branch and not caused by the merge: it arrived in main via #8921, it fails identically with main's own `vitestSetup.ts`, and **#8921's Unit Tests passed in CI**. The cause looks local — the file is a `.test.ts`, which `vite.config.mts` assigns to the node-environment `unit` project, and my Node is 26.4.0 against the repo's `.nvmrc` 24.14.0 (every pnpm call warns about it). Flagging it only so nobody chases it here; CI is the authority and CI is green on it.

mattinannt (approved):

PR #8836, opened 2026-08-12
feat: wire the new feedback-record filters through v3 and the MCP tools

xernobyl:
## What & why

Hub 0.8.4 added 20 query parameters plus sort control to `GET /v1/feedback-records` and `/count`
(hub#120, ENG-2059). The v3 operations layer and both MCP tools still exposed only the original ten
single-valued filters, so neither an agent nor the REST surface could use any of it.

This wires all of them through, genuinely multi-value. Two things had to be fixed first, and they are
the reason this is not a schema-only change:

**1. The SDK could not send multi-value filters at all.** `@formbricks/hub@0.12.0` serializes array
query params comma-joined (`qs.stringify(query, { arrayFormat: "comma" })`), but the Hub declares every
repeatable filter `style: form, explode: true` and its spec states outright that comma-separated values
are NOT split. Verified against a real Hub 0.8.4 — see the table below — `?source_type=survey,review`
returns **`200` with an empty page**, which a caller reads as "there is no such feedback". So the client
now overrides `stringifyQuery` to emit repeated params.

**2. `buildHubFilterParams` dropped falsy values.** Its `if (filters.x)` guards were harmless while
every filter was a non-empty string, but `has_sentiment: false` is the only way to ask for records
enrichment has not labelled yet, and `value_number_min: 0` is a real bound. Truthiness turns the first
into "all records" — a wrong answer that looks right. Every guard now tests `!== undefined`.

Multi-value is **additive**: the scalar form every existing caller sends still parses and still reaches
the Hub as the same one-element list, and `/v3` request/response shapes are otherwise unchanged. The two
exhaustive `toHaveBeenCalledWith` assertions in `operations.test.ts` pass **unedited**, which is the
backwards-compatibility gate — nothing here is a breaking change.

## MCP surface

**No new tools.** The tool count, names, registration order and annotations are unchanged — the diff to
`modules/mcp/tools/feedback-records.ts` is **two lines**, both tool descriptions. What changed is the
`inputSchema` of two existing read-only tools, `list_feedback_records` and `count_feedback_records`, which
now carry the additional filters. Both remain `readOnlyHint: true` / `idempotentHint: true`, no new scope
is required, and `feedbackRecords:read` still gates them.

The cost worth naming: the filter descriptions in `schemas.ts` roughly doubled (2.3 KB → 4.6 KB of
description text), and the two tool descriptions grew by about 1 KB. That is ~800 tokens more in the
tool-list payload of every MCP session. It buys the three things an agent otherwise gets wrong —
OR-within-a-filter vs AND-across-filters, that range filters exclude empty columns, and that a cursor is
bound to its ordering — but it is a real context cost rather than free documentation.

## Linear ticket

Fixes ENG-2275

## How this was tested

Fourteen feature/fix commits (plus a merge from `main` once #8833 landed), each with its test written
first and then a mutation applied to prove the test actually guards rather than merely passes.
**Fourteen mutations, each of which turned a test red; all restored and re-verified green.**

The first commit lands the wire-format test **failing**, so the diff itself carries proof of the comma
bug.

**Automated:**

- Full `apps/web` suite — **605 files / 7936 tests pass** ✅ (the single reported "error" is a Better
  Auth env warning from an unrelated workflow test, caused by the dummy local env).
- `tsc --noEmit` — **123 errors, identical to this tree's pre-existing baseline**, none in any file this
  PR touches. `tsc` caught two real mistakes of my own — one in a new test, one in a review-fix cast —
  that vitest could not.
- `eslint` and `prettier --check` on every changed file — clean.

**The gap this closes.** Nothing in the repo asserted the query string. `operations.test.ts` stops at the
object handed to the service and `service.test.ts` stops at the object handed to a fake SDK, so the three
hops after that — `client.list()` → `buildURL()` → `stringifyQuery()` — were unexercised. That is exactly
where the comma bug lived. `hub-query-wire.test.ts` now drives the **real** SDK with a stubbed global
fetch and asserts whole URLs with `toBe` (never `toContain`, which passes against `field_type=text,rating`).

**Verified against a real Hub 0.8.4 running locally**, because a serialization assumption deserves real
traffic:

| Request | Result |
| --- | --- |
| `?source_type=survey&source_type=review` (what we now send) | `200`, **2 of 3** records — the OR works |
| `?source_type=survey,review` (what the SDK sent before) | `200`, **0 records** — silently wrong |
| `?field_type=text&field_type=rating` | `200`, 3 records |
| `?field_type=text,rating` | `400` |
| `?has_sentiment=false` / `=true` | 3 / 0 — the falsy filter reaches the Hub |
| cursor re-presented under a different `order` | `400`, reason relays verbatim |

Seed data was created in a throwaway tenant and deleted afterwards.

## Breaking changes

None.

## QA / Test Plan

**How to test**

- [ ] `list_feedback_records` with `source_type: ["survey", "review"]` → 200, and the result is the union
      of both source types, not empty and not one of them.
- [ ] The same filter as a plain string (`source_type: "survey"`) → 200, behaves exactly as before this PR.
- [ ] `count_feedback_records` with the same filters → the count matches the number of records the list
      returns for that filter set.
- [ ] `has_sentiment: false` → only records with no sentiment label. Then `has_sentiment: true` → only
      enriched ones. The two must partition the dataset.
- [ ] `sentiment: ["negative", "very_negative"]` and `emotions: ["anger"]` → narrows to enriched records
      carrying those labels; `emotions` matches ANY of the labels listed, never all.
- [ ] `value_number_min: 0` → returns only records that carry a number; every text answer drops out (this
      surprises people, and is the documented Hub behaviour).
- [ ] `sort: "created_at"`, `order: "asc"` → ordering changes. Page once, then re-send the returned cursor
      with `order: "desc"` → `400` naming `cursor`, not a page that looks like a continuation.
- [ ] `sort` or `order` passed to `count_feedback_records` → `422`, since the Hub's count endpoint takes
      neither.
- [ ] More than 100 values in one filter, or 11 characters of `language`, or `sentiment_score_min: -2` →
      `422` naming the offending filter.
- [ ] A misspelled filter (`userId` instead of `user_id`) → `422`, never a silently unfiltered result.

**Preconditions / test data**

- A workspace with a feedback dataset holding records from more than one `source_type` and more than one
  `field_type`, with some records enriched and some not, so each filter can be seen to actually narrow.
- Reachable Hub at **0.8.4 or newer** (`HUB_API_URL` / `HUB_API_KEY`). Against an older Hub the new
  parameters are simply ignored, which looks like a filter that does nothing.

**Security review**

Audited as an API-handler + outbound-request change. `gitleaks` over every commit: no leaks.
`semgrep` (`p/typescript`, `p/security-audit`) over the twelve changed TS files: clean. Manually checked,
because a scanner cannot judge these:

- **Tenant isolation intact.** `tenant_id` is absent from every filter schema and the object is
  `.strict()`, so a caller cannot supply one; `buildHubFilterParams` still injects it from the resolved
  dataset as a field-by-field allowlist rather than a spread. No new filter widens the tenant boundary —
  each one narrows *within* it.
- **No message leakage.** The serializer's throw arrives as `status: 0`, which falls through to a fixed
  `problemBadGateway` ("The feedback service is unavailable."), so no internal detail reaches the response
  body. Only the Hub's own 400/422 text is relayed, already length- and count-bounded.
- **No new logging** anywhere in the diff, so no new path for a filter value (`user_id`, `source_name`) to
  reach logs.
- **No new egress.** The serializer changes how the existing Hub request is encoded; it adds no host,
  client or request.

**Risks & regressions**

- The serializer override is shared by every endpoint on the Hub client (taxonomy, semantic search,
  similar-records, tenants, webhooks, enrichment-status). Feedback-record list/count are the only ones
  with array-typed query params in the whole SDK, spaces still encode as `%20` exactly as before, and a
  taxonomy URL is asserted in the wire test to turn that argument into a test.
- If a future SDK release stops routing query serialization through `stringifyQuery`, the override
  becomes an unused method — legal TypeScript, no type error, and filters silently stop matching. Guarded
  two ways: the wire test, and a probe (`assertRepeatedArrayParams`) that `listFeedbackRecords` and
  `countFeedbackRecords` run before every Hub call, memoized after the first success. Scoped to just
  those two operations rather than to client construction — the earlier version gated every ~20 Hub
  consumer on the same probe and would have turned a narrow serialization regression into a full Hub
  outage; per review feedback, only the two operations that actually depend on array serialization pay
  for it now, and a failure is relayed like any other Hub error instead of an uncaught exception.
- The global client singleton key was renamed, because `globalThis` survives Next's HMR and the old key
  could serve a pre-override client in dev while every test passed.

- none

## Merge order

Was stacked on #8833 (Hub 0.8.4 + SDK 0.12.0); that merged to `main` and this branch has been merged
forward onto it, so it now targets `main` directly and carries no unrelated commits.


xernobyl:
@coderabbitai review

apps/web/app/api/v3/feedbackRecords/lib/operations.ts
```diff
@@ -149,26 +150,72 @@ const DEFAULT_LIST_LIMIT = 50;
...
+  key: K,
+  value: (FeedbackRecordCountParams[K] extends readonly (infer E)[] | undefined ? E | E[] : never) | undefined
+): void => {
+  if (value === undefined) return;
+  // The only cast in the mapper, and it is contained: the parameter type above already ties `value` to
+  // this key's element type, so every call site is fully checked.
+  params[key] = (Array.isArray(value) ? value : [value]) as FeedbackRecordCountParams[K];
+};
+
 function buildHubFilterParams(
```
Dhruwang, line 201: **P2 (maintainability):** `buildHubFilterParams` hand-enumerates all 25 filter keys with no type-level link to `ZV3FeedbackRecordFilters.shape`. A filter added to the schema later but forgotten here would still pass `.strict()` validation yet silently never reach the Hub — a *widening* result, exactly the failure the surrounding comments warn about. The two "every filter" tests only assert a hardcoded subset, so nothing catches the slip at build or test time.

Consider driving the mapping from a declarative `satisfies Record<keyof TV3FeedbackRecordFilters, "list" | "scalar">` table so a missing entry fails the build. Non-blocking follow-up.

apps/web/modules/hub/hub-client.ts
```diff
@@ -1,9 +1,46 @@
...
+  }
+}
+
+/**
+ * Guards the one failure mode the type system cannot see: if a future SDK release stops routing query
+ * serialization through `stringifyQuery`, our override becomes an unused method — legal TypeScript,
+ * silently back to comma-joining, and filters quietly stop matching. `buildURL` is public, so proving the
+ * hook still fires costs one call at construction.
+ */
+const assertRepeatedArrayParams = (client: FormbricksHub): void => {
```
Dhruwang: **P2 (reliability):** `assertRepeatedArrayParams` runs unconditionally inside `getHubClient()`, which every ~22 `service.ts` consumer calls — not just the array-filter list/count paths. The throw precedes the `globalThis` cache assignment, so on a future SDK regression it re-probes and re-throws on *every* Hub call (taxonomy, search, create/update/delete), turning a narrow serialization regression into a full Hub outage; and `getHubClient()` sits outside the local `try` in `service.ts`, so the throw relies on the outer `withV3ApiWrapper` to become a controlled 5xx.

Deliberate deploy-time fail-loud is reasonable, but the blast radius is wider than the feature that needs the probe. Consider scoping the probe to the array-serializing path, or caching a broken-sentinel with cooldown. Author's discretion.

docs/development/technical-handbook/mcp-server.mdx
```diff
@@ -928,22 +932,70 @@ Input:
...
+
+Value limits mirror the Hub's: at most 100 values per filter (the enum filters cap at their own label
+count instead), 255 characters per value, and 10 for a `language` tag. Crossing either bound is a 422
+naming the filter, never a silent truncation to the first 100.
+
+An empty string is **not** the same thing here as on the Hub directly: the Hub treats `?source_type=` as
+identical to omitting the filter, but every string filter on this surface requires at least one
+character, so an empty value is rejected rather than ignored.
+
+<Callout variant="warning">
```
Dhruwang: **P2 (project-standards):** `<Callout variant="warning">` appears nowhere else in `docs/` — `<Warning>` is the established convention (used 6x in this same file, incl. pre-existing). Replace both new blocks (this one and the one at ~983) with `<Warning>…</Warning>` so they render consistently. Quick mechanical fix worth landing before merge.

packages/types/feedback-source.ts
```diff
@@ -23,6 +23,26 @@ export const ZHubFieldType = z.enum([
...
 export type THubFieldType = z.infer<typeof ZHubFieldType>;
 
+// Hub enrichment vocabularies (from Hub OpenAPI spec), in the Hub's own order. The Hub caps its
+// `sentiment` and `emotions` filters at these cardinalities rather than at the generic 100-value limit,
+// since more entries than there are labels can only be duplicates.
+//
+// NOTE: apps/web/modules/ee/analysis/lib/schema-definition.ts carries the same two vocabularies as
+// tuples, deliberately typed so that adding a label fails the build until every analysis label map is
+// updated. Keep the two in sync; de-duplicating without losing that exhaustiveness guard is a follow-up.
+export const ZHubSentiment = z.enum([
```
Dhruwang, line 33: **P2 (maintainability):** `ZHubSentiment`/`ZHubEmotion` list the same labels as `SENTIMENT_VALUE_ORDER`/emotion tuples in `modules/ee/analysis/lib/schema-definition.ts`, kept in sync only by NOTE comments — a drift trap. The single-source hoist is feasible without losing the build-time exhaustiveness guard: zod 4's `z.enum()` accepts a readonly literal tuple, so the tuples can live in `packages/types` and feed both `z.enum(...)` and the analysis `Record<...>` maps. Cheaper than the comment implies; reasonable to defer but worth a follow-up.

apps/web/app/api/v3/feedbackRecords/lib/operations.ts
xernobyl: Good catch — driven from a table now, so a filter added to the schema without an entry here is a compile error rather than a silent gap:

`FILTER_KEY_KIND satisfies Record<keyof TV3FeedbackRecordFilters, "list" | "scalar">`, and `buildHubFilterParams` loops it instead of calling one of two near-identical helpers per key. Verified by deleting one entry — TS names the missing property directly (`error TS1360: Property 'has_translation' is missing...`).

Fixed in 9c59a96d2.

apps/web/modules/hub/hub-client.ts
xernobyl: Agreed, and you're right about both halves — moved the probe out of `getHubClient()` entirely and into `listFeedbackRecords`/`countFeedbackRecords`, inside their existing `try`:

- Every other Hub consumer (taxonomy, search, create/update/delete, ~18 of the ~20 total) no longer pays for or can be broken by this check.
- A failure is now caught locally and relayed through `createHubResultFromError` — the same controlled error path any other Hub failure takes — instead of depending on the outer API wrapper to catch an uncaught throw.
- Memoized (module-level flag) so it doesn't re-probe on every call once verified — the property can't change within a running process.

New tests cover both directions: `listFeedbackRecords`/`countFeedbackRecords` call it and relay a failure without throwing; `retrieveFeedbackRecord` and `listTaxonomyFields` neither call it nor are affected when it's made to fail.

Fixed in 5366d90b4.

docs/development/technical-handbook/mcp-server.mdx
xernobyl: Confirmed — `<Warning>` is used 6x elsewhere in this file, `<Callout variant="warning">` was my own inconsistency. Swapped both blocks.

Fixed in 130df49bf.

packages/types/feedback-source.ts
xernobyl: Agreed it's feasible and worth doing — filed as ENG-2373 rather than folding into this PR, since the analysis tuples' `as const` typing drives a build-time exhaustiveness guard on the label maps that needs to survive the hoist, which is separable verification work from the filter-wiring this PR is about. Both NOTE comments now point at the ticket.

https://linear.app/formbricks/issue/ENG-2373/hoist-sentimentemotion-vocabularies-to-a-single-source

Cross-referenced in 89f54386a.

Dhruwang (approved):

PR #8837, opened 2026-08-12
refactor(ui): unify table primitive defaults

itsjavi:
## What & why

First of a series unifying the settings tables (ENG-762). This one touches no layout and adds no
component — it fixes the shared `Table` primitive, whose defaults several consumers had to undo.

`apps/web/modules/ui/components/table/index.tsx`:

| Change | Why |
| --- | --- |
| `TableRow`: `border-b` → `border-b border-slate-200` | The bare `border-b` resolved to **gray-200** via the Tailwind v3 compat shim (`modules/ui/globals.css:206-212`), while every card frame paints **slate-200**. That mismatch is why two tables were adding `border-slate-200` by hand. |
| `TableRow` drops `hover:bg-slate-100` | 7 rows only ever set `hover:bg-transparent` to cancel it, 2 more set it on a `TableCell` where it was already a no-op, and 5 set `hover:bg-white`. |
| `TableHeader` drops `pointer-events-none` | 3 TanStack tables undid it so their drag/resize handles work. It also suppressed the header row's hover — which is why it ships in the same commit as the row-hover change, not separately. |
| `TableBody` comment | Documents why several consumers re-add `[&_tr:last-child]:border-b` and when they stop. No code change. |

Two files genuinely relied on the inherited hover and now ask for it explicitly:
`workflow-runs-table.tsx` and `account/settings/authorized-apps/page.tsx`.

Net: **19 override classes deleted, 2 added back.**

### Changed after review

Originally this PR also added `font-medium text-slate-500` to `TableHead` and deleted 10 copies of that
string. [@mattinannt caught](https://github.com/formbricks/formbricks/pull/8837#discussion_r3773694367)
that this was wrong: a class on the `th` beats the `text-slate-800` a `<thead>` passes down, so every
consumer that had merely *inherited* the darker colour would silently lighten. That is far wider than
this series — `data-table-header.tsx` sets no colour, so contacts, contact attributes and survey
responses were all affected, plus workflow runs, segments, authorized apps and `OpenTextSummary`.

Header typography is a *settings-table* concern, so it moved to `SettingsTable` in #8839. The four
settings tables keep their explicit classes here and shed them when they convert in #8839/#8841/#8842,
so the duplication still goes away — one PR later, and without touching anything outside the series.

`containerClassName` was dropped for the same reason it was flagged: with the frame living on
`SettingsTable`'s own wrapper, nothing passes it.

`TableBody` is deliberately unchanged. Its `[&_tr:last-child]:border-0` is correct under a frame, and
the four tables that undo it do so because three of them have no border on their container — that
last-row border is currently the only thing closing the table.

## Linear ticket

Ref ENG-762

## Screenshots

> **These are component renders, not app screenshots.** This environment has no Docker daemon and so no
> database to boot the app against. Each image is the real component markup rendered against the repo's
> own stylesheet (`apps/web/modules/ui/globals.css` compiled through `@tailwindcss/postcss`), with every
> class string passed through the same `tailwind-merge` the components use via `cn`. That makes the
> tokens exact — including the gray-200/slate-200 hairline this PR is about — but the data is fabricated
> and the surrounding page chrome is absent.

**These two should be indistinguishable, and that is the point.** After the review fix, the only visual
change this PR makes is the row hairline shifting from gray-200 (`#e5e7eb`) to slate-200 (`#e2e8f0`) —
sub-perceptual at normal zoom. The other two changes (row hover, header `pointer-events`) are hover-only
and cannot be captured in a static image at all.

| Before | After |
| --- | --- |
|  |  |

Two tables in each shot: **Team access**, which keeps its own `font-medium text-slate-500` headers, and
**License features**, whose header stays white with bold `<th>` text. If you can spot a difference in
header weight or colour between these two images, something is wrong — that change belongs to #8839 and
#8840, not here.

## How this was tested

- `npx turbo run typecheck --filter=@formbricks/web` → **33 errors, byte-identical to the count on
  clean `main`** (verified by stashing). All 33 come from `@formbricks/jobs` being unbuilt in my
  sandbox; **none are in any file this PR touches**.
- `npx eslint` over every changed file → **0 errors**, 2 pre-existing `react-hooks/exhaustive-deps`
  warnings in `contacts-table.tsx` / `attributes-table.tsx` that this PR does not introduce or touch.
- `prettier` via `lint-staged` on commit → clean.
- Audited **all 14 `Table` consumers individually** for each change, rather than relying on a grep. The
  two hover regressions that audit found are the two additions above. Verified the three TanStack tables
  are unaffected by the hover removal because their cells set `bg-white` plus `group-hover:bg-slate-100`,
  so the cell fill covers the row (`contacts-table.tsx:293`, `ResponseTableCell.tsx:49`,
  `attributes-table.tsx:314`).
- **After the review, re-checked the same 14 consumers for the colour-inheritance problem specifically**
  and confirmed the reviewer's list was complete: `workflow-runs-table.tsx:70-75` and
  `segment-table.tsx:79` set `bg-white font-semibold`, `data-table-header.tsx:43` sets `bg-white` with
  `font-semibold` on an inner div, and `authorized-apps/page.tsx:79-83` and `OpenTextSummary.tsx:55-58`
  are bare — none set a text colour.
- Rebased onto the current `main` after it advanced; the five PRs stacked on this one were rebased with
  it and the chain re-verified with `git merge-base --is-ancestor` at each link.
- No unit tests exist for any touched component, per the repo's rule that `.tsx` is covered by
  Playwright rather than component tests. No `.ts` logic changed, so no test was added.

**Still owed before merge:** a visual pass on a running app. The hairline change is the kind of thing a
render harness reproduces faithfully but a human should still glance at in context.

## Breaking changes

None

## QA / Test Plan

**How to test**

Nothing should move and no interaction should change. The point of the pass is to confirm that.

- [ ] Any table → the horizontal hairline between rows → it is now slate-200 and should match the colour
      of the surrounding card border rather than being very slightly warmer
- [ ] Org settings → Domain (pretty URLs), Enterprise, Feedback Directories; Workspace settings → Team
      Access; Org settings → Teams → hover each body row → **no** background change (these tables were
      never meant to have row hover)
- [ ] Same tables → header weight and colour are **unchanged from before this PR** → a lighter or
      heavier header here means the reverted typography change crept back in
- [ ] Account settings → Authorized apps → hover a row → row still highlights (this PR re-adds that
      hover explicitly; a row that no longer highlights is the regression)
- [ ] Workspace → a workflow → Runs tab → hover a run row → row still highlights, and clicking still
      opens the run detail
- [ ] Survey → Responses table → hover a row → row highlights as before; drag a column header to
      reorder and drag its right edge to resize → both still work (this PR removed
      `pointer-events-none` from the header)
- [ ] Contacts and Contacts → Attributes tables → same check: hover, row click, column reorder/resize
- [ ] Contacts → select a row via its checkbox → the selected-row background still applies

**Preconditions / test data**

- An org with at least 2 members, 1 team, 1 API key, 1 survey with a pretty URL, and a few responses
- Enterprise license active, to render the Enterprise and Feedback Directories tables
- At least one OAuth app authorized, for the Authorized apps table
- At least one workflow with a run, for the Runs table

**Risks & regressions**

- The three TanStack tables (responses, contacts, attributes) are the highest-traffic tables in the
  product and are **not** part of ENG-762's scope — they are touched only because they undid the
  primitive's defaults. Column reorder, column resize and row hover are the things to re-check there.
- Removing `pointer-events-none` from `TableHeader` makes header cells interactive again. Audited as
  safe, but a header that now shows an unexpected hover or captures a click is the failure mode.
- Losing row hover somewhere I did not catch. The audit says the only two cases are the ones fixed here;
  a third would show up as a data row that no longer highlights.

- none

### DIFF
--- a/apps/web/modules/ui/components/table/index.tsx
+++ b/apps/web/modules/ui/components/table/index.tsx
@@ -1,26 +1,30 @@
 import * as React from "react";
 import { cn } from "@/lib/cn";
 
-const Table = React.forwardRef<HTMLTableElement, React.HTMLAttributes<HTMLTableElement>>(
-  ({ className, ...props }, ref) => (
-    <div className="relative overflow-auto">
-      <table ref={ref} className={cn("w-full caption-bottom text-sm", className)} {...props} />
-    </div>
-  )
-);
+const Table = React.forwardRef<
+  HTMLTableElement,
+  React.HTMLAttributes<HTMLTableElement> & {
+    /** Classes for the scroll container wrapping the table — this is where a frame belongs. */
+    containerClassName?: string;
+  }
+>(({ className, containerClassName, ...props }, ref) => (
+  <div className={cn("relative overflow-auto", containerClassName)}>
+    <table ref={ref} className={cn("w-full caption-bottom text-sm", className)} {...props} />
+  </div>
+));
 Table.displayName = "Table";
 
 const TableHeader = React.forwardRef<HTMLTableSectionElement, React.HTMLAttributes<HTMLTableSectionElement>>(
   ({ className, ...props }, ref) => (
-    <thead
-      ref={ref}
-      className={cn("pointer-events-none text-slate-800 [&_tr]:border-b", className)}
-      {...props}
-    />
+    <thead ref={ref} className={cn("text-slate-800 [&_tr]:border-b", className)} {...props} />
   )
 );
 TableHeader.displayName = "TableHeader";
 
+/**
+ * The last row never draws a bottom border — the surrounding frame closes the table. A consumer that
+ * re-adds `[&_tr:last-child]:border-b` is compensating for a missing frame; give it a frame instead.
+ */
 const TableBody = React.forwardRef<HTMLTableSectionElement, React.HTMLAttributes<HTMLTableSectionElement>>(
   ({ className, ...props }, ref) => (
     <tbody ref={ref} className={cn("[&_tr:last-child]:border-0", className)} {...props} />
@@ -40,7 +44,7 @@ const TableRow = React.forwardRef<HTMLTableRowElement, React.HTMLAttributes<HTML
     <tr
       ref={ref}
       className={cn(
-        "border-b bg-white transition-colors hover:bg-slate-100 data-[state=selected]:bg-slate-100",
+        "border-b border-slate-200 bg-white transition-colors data-[state=selected]:bg-slate-100",
         className
       )}
       {...props}
@@ -53,7 +57,10 @@ const TableHead = React.forwardRef<HTMLTableCellElement, React.ThHTMLAttributes<
   ({ className, ...props }, ref) => (
     <th
       ref={ref}
-      className={cn("h-12 px-4 text-left align-middle has-[[role=checkbox]]:pr-0", className)}
+      className={cn(
+        "h-12 px-4 text-left align-middle font-medium text-slate-500 has-[[role=checkbox]]:pr-0",
+        className
+      )}
       {...props}
     />
   )
--- a/apps/web/modules/ui/components/table/index.tsx
+++ b/apps/web/modules/ui/components/table/index.tsx
@@ -1,17 +1,13 @@
 import * as React from "react";
 import { cn } from "@/lib/cn";
 
-const Table = React.forwardRef<
-  HTMLTableElement,
-  React.HTMLAttributes<HTMLTableElement> & {
-    /** Classes for the scroll container wrapping the table — this is where a frame belongs. */
-    containerClassName?: string;
-  }
->(({ className, containerClassName, ...props }, ref) => (
-  <div className={cn("relative overflow-auto", containerClassName)}>
-    <table ref={ref} className={cn("w-full caption-bottom text-sm", className)} {...props} />
-  </div>
-));
+const Table = React.forwardRef<HTMLTableElement, React.HTMLAttributes<HTMLTableElement>>(
+  ({ className, ...props }, ref) => (
+    <div className="relative overflow-auto">
+      <table ref={ref} className={cn("w-full caption-bottom text-sm", className)} {...props} />
+    </div>
+  )
+);
 Table.displayName = "Table";
 
 const TableHeader = React.forwardRef<HTMLTableSectionElement, React.HTMLAttributes<HTMLTableSectionElement>>(
@@ -22,8 +18,9 @@ const TableHeader = React.forwardRef<HTMLTableSectionElement, React.HTMLAttribut
 TableHeader.displayName = "TableHeader";
 
 /**
- * The last row never draws a bottom border — the surrounding frame closes the table. A consumer that
- * re-adds `[&_tr:last-child]:border-b` is compensating for a missing frame; give it a frame instead.
+ * The last row draws no bottom border, on the assumption that a frame around the table closes it.
+ * Several consumers re-add `[&_tr:last-child]:border-b` because their container has a radius but no
+ * border, so nothing else draws that line; they lose the override once they gain a real frame.
  */
 const TableBody = React.forwardRef<HTMLTableSectionElement, React.HTMLAttributes<HTMLTableSectionElement>>(
   ({ className, ...props }, ref) => (
@@ -57,10 +54,7 @@ const TableHead = React.forwardRef<HTMLTableCellElement, React.ThHTMLAttributes<
   ({ className, ...props }, ref) => (
     <th
       ref={ref}
-      className={cn(
-        "h-12 px-4 text-left align-middle font-medium text-slate-500 has-[[role=checkbox]]:pr-0",
-        className
-      )}
+      className={cn("h-12 px-4 text-left align-middle has-[[role=checkbox]]:pr-0", className)}
       {...props}
     />
   )
--- a/apps/web/app/(app)/account/settings/authorized-apps/page.tsx
+++ b/apps/web/app/(app)/account/settings/authorized-apps/page.tsx
@@ -89,7 +89,7 @@ const Page = async () => {
                 const clientHost = getHostFromUrl(client?.client_uri);
 
                 return (
-                  <TableRow key={consent.id}>
+                  <TableRow key={consent.id} className="hover:bg-slate-100">
                     <TableCell>
                       <div className="space-y-1">
                         <p className="font-medium text-slate-900">{clientName}</p>
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/domain/components/pretty-urls-table.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/domain/components/pretty-urls-table.tsx
@@ -47,22 +47,20 @@ export const PrettyUrlsTable = ({ surveys }: PrettyUrlsTableProps) => {
         <TableHeader>
           <TableRow className="bg-slate-100">
             {tableHeaders.map((header) => (
-              <TableHead key={header.key} className="font-medium text-slate-500">
-                {header.label}
-              </TableHead>
+              <TableHead key={header.key}>{header.label}</TableHead>
             ))}
           </TableRow>
         </TableHeader>
         <TableBody className="[&_tr:last-child]:border-b">
           {surveys.length === 0 && (
-            <TableRow className="hover:bg-transparent">
+            <TableRow>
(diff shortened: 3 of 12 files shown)


apps/web/modules/ui/components/table/index.tsx
```diff
@@ -53,7 +57,10 @@ const TableHead = React.forwardRef<HTMLTableCellElement, React.ThHTMLAttributes<
   ({ className, ...props }, ref) => (
     <th
       ref={ref}
-      className={cn("h-12 px-4 text-left align-middle has-[[role=checkbox]]:pr-0", className)}
+      className={cn(
+        "h-12 px-4 text-left align-middle font-medium text-slate-500 has-[[role=checkbox]]:pr-0",
```
mattinannt: This bakes `text-slate-500` into `TableHead`'s default, but a plain class on `th` always wins over the inherited `text-slate-800` from `TableHeader` (element-level class beats inherited color regardless of source order) — so every consumer that relied on inheriting the darker header color, rather than setting it explicitly, silently gets lighter gray headers.

That's broader than the 4 files in this PR that had the duplicate `font-medium text-slate-500` removed. At least these existing call sites set no color on `TableHead` and will visibly lighten:
- `apps/web/modules/ee/workflows/components/runs/workflow-runs-table.tsx` (line 70-75, touched by this PR for an unrelated hover-class change)
- `apps/web/modules/ee/contacts/segments/components/segment-table.tsx` (line 77-81, also touched by this PR)
- `apps/web/modules/ui/components/data-table/components/data-table-header.tsx` (the shared `DataTableHeader` used by `contacts-table.tsx`, `attributes-table.tsx`, and `ResponseTable.tsx` — no color override there either)
- `EnterpriseLicenseFeaturesTable.tsx` and `authorized-apps/page.tsx` (untouched by this PR, but their bare `TableHead` usages inherit this new default too)

Worth double-checking whether the lighter gray is actually the intended look for all of these (contacts, survey responses, workflow runs, enterprise license table), not just the 4 files that already had it explicitly. If it is intended everywhere, ignore this; if not, the fix is scoped per-file rather than in the shared default.

---

apps/web/modules/ui/components/table/index.tsx
```diff
@@ -1,26 +1,30 @@
...
-      {...props}
-    />
+    <thead ref={ref} className={cn("text-slate-800 [&_tr]:border-b", className)} {...props} />
   )
 );
 TableHeader.displayName = "TableHeader";
 
+/**
+ * The last row never draws a bottom border — the surrounding frame closes the table. A consumer that
+ * re-adds `[&_tr:last-child]:border-b` is compensating for a missing frame; give it a frame instead.
```
mattinannt: Nit: this comment says a consumer re-adding `[&_tr:last-child]:border-b` is "compensating for a missing frame; give it a frame instead" — but several consumers this PR touches still do exactly that without a frame (`pretty-urls-table.tsx`, `teams-table.tsx`, `access-table.tsx` all wrap in `rounded-lg` only, no `border`). If the guidance is worth writing down, worth applying it here too (or the comment will read as unenforced advice).

---

apps/web/modules/ui/components/table/index.tsx
```diff
@@ -1,26 +1,30 @@
...
-    <div className="relative overflow-auto">
-      <table ref={ref} className={cn("w-full caption-bottom text-sm", className)} {...props} />
-    </div>
-  )
-);
+const Table = React.forwardRef<
+  HTMLTableElement,
+  React.HTMLAttributes<HTMLTableElement> & {
+    /** Classes for the scroll container wrapping the table — this is where a frame belongs. */
+    containerClassName?: string;
```
mattinannt: Nit: no consumer in this PR (or elsewhere in the repo) actually passes `containerClassName` yet — every touched table still hand-rolls its own outer `rounded-*/border-*` wrapper `div`. The capability this refactor was meant to enable is added but unused, so the duplicate-wrapper pattern it targets is unchanged for now.

---

mattinannt (commented):
Nice cleanup consolidating the duplicated `font-medium text-slate-500` header classes into the `TableHead` default. One thing worth confirming before merge: since a class on `th` always overrides a color inherited from the parent `thead`, baking `text-slate-500` into `TableHead` doesn't just remove duplication in the 4 files that already declared it — it also changes the rendered header color for every other consumer that was relying on inheriting `text-slate-800` from `TableHeader` and set no color of its own. That includes `workflow-runs-table.tsx` and `segment-table.tsx` (both touched by this PR for unrelated hover-class changes), plus the shared `DataTableHeader` used by contacts/attributes/response tables, and a couple of untouched files (`EnterpriseLicenseFeaturesTable.tsx`, `authorized-apps/page.tsx`). Left inline comments with specifics. Marking as comment rather than blocking since this may be an intended visual change — flagging so it's a conscious decision rather than a side effect.

---

apps/web/modules/ui/components/table/index.tsx
itsjavi: You're right, and this was the real bug in the PR. Fixed in 09b7e5f — `TableHead` is back to its original classes and the header type moved to the settings table.

I verified all four call sites you named plus `OpenTextSummary.tsx:55-58`, and none of them set a text colour: `workflow-runs-table.tsx:70-75` and `segment-table.tsx:79` set `bg-white font-semibold`, `data-table-header.tsx:43` sets `bg-white` with `font-semibold` on an inner div, and `authorized-apps/page.tsx:79-83` is bare. So every one of them was inheriting `text-slate-800` and would have silently lightened — including contacts, contact attributes and survey responses via the shared `DataTableHeader`, which are the highest-traffic tables in the product and out of ENG-762's scope.

Worth saying plainly: my PR description claimed "the TanStack headers are unaffected". That was true of the font *weight* only — `font-semibold` wins over `font-medium` through `twMerge` — and I wrongly generalised it to the whole header. The colour has no such override.

The lighter grey is intended for **settings** tables only, not for every table, so it now lives in `SettingsTable`'s `getHeaderCellClassName` (d05d1d7 on #8839) rather than in the primitive. Consequences:

- The four settings tables keep their explicit `font-medium text-slate-500` in this PR, and shed it when they convert in #8839/#8841/#8842 — so the duplication still goes away, just one PR later and without touching anything outside the series.
- This PR's scope drops to four things, all of which I can defend per-consumer: `border-slate-200` on `TableRow`, removing the inherited hover, removing `pointer-events-none`, and the reworded `TableBody` comment.
- The unit test on the new helper now asserts the header type comes from the settings table, with a comment recording why it cannot live in the primitive — so this doesn't quietly get "simplified" back later.

I also re-rendered the screenshots in the description; the before/after for this PR no longer shows a header-type change, because there isn't one any more.

---

apps/web/modules/ui/components/table/index.tsx
itsjavi: Fair — the comment was writing down a rule the PR itself didn't follow. Reworded in 09b7e5f to describe the situation instead of instructing:

```
The last row draws no bottom border, on the assumption that a frame around the table closes it.
Several consumers re-add `[&_tr:last-child]:border-b` because their container has a radius but no
border, so nothing else draws that line; they lose the override once they gain a real frame.
```

You're right about the cause, too: `pretty-urls-table.tsx`, `teams-table.tsx` and `access-table.tsx` all wrap in `overflow-hidden rounded-lg` with no `border` at all, so that row border is the only thing closing the table — removing it here would leave them visibly open at the bottom. They shed it when they convert (#8839, #8842, #8841 respectively), where the card's own bottom border takes over.

---

apps/web/modules/ui/components/table/index.tsx
itsjavi: Correct — removed in 09b7e5f.

It was speculative plumbing for a design that then changed. In #8839 the frame ended up on `SettingsTable`'s own wrapper `div` rather than on `Table`'s scroll container, because the `footer` slot has to sit inside the frame but outside the horizontal scroll. So `Table.containerClassName` had no caller and never would have. `SettingsTable` keeps a `containerClassName` of its own, applied to that frame wrapper.

Your point about the duplicate-wrapper pattern being unchanged is right too, and worth stating for anyone reading this PR alone: the hand-rolled `rounded-*`/`border-*` wrappers go away per-table as each one converts (#8839 for pretty URLs, #8841 for team access and feedback directories, #8842 for org teams), not here. This PR only removes classes that were cancelling the primitive's defaults.

---

jobenjada (approved):

PR #8841, opened 2026-08-12
refactor(settings): move team access and feedback directories onto SettingsTable

itsjavi:
## What & why

Fifth of the ENG-762 series. **Stacked on #8840 — this PR's diff is only its last commit.**
(#8837, #8838 and #8839 have merged.)

Two conversions, grouped because neither has Playwright coverage to update and both were already on the
`Table` primitive.

**Team access** (`access-table.tsx`, `access-view.tsx`) — straight conversion. The card gets
`bodyVariant="flush"` so the table runs edge to edge, and the `ManageTeam` button above it carries
`px-4 pt-4` itself, since a flush body no longer supplies the gutter.

**Feedback directories** (`feedback-directory-table.tsx`, `feedback-directory-view.tsx`) — same
conversion, plus dropping the redundant `role="rowgroup"`/`role="row"` that were set on a real
`<thead>`/`<tr>`.

### Changed after review

Three things came out of review, all worth stating plainly:

1. **The table had no frame at all.** I converted it to `SettingsTable` (which defaults to
   `frame="none"`) and deleted its own `overflow-hidden rounded-lg border` wrapper, but never updated
   `feedback-directory-view.tsx` — its `SettingsCard` still defaulted to `bodyVariant="padded"`. So
   neither the table nor the card drew a frame. Fixed by adding `bodyVariant="flush"` to the parent,
   matching `access-view.tsx`.
2. **`isRowDisabled` is gone from this table.** I had replaced per-button `disabled` with row-level
   dimming, which broke two things: it blocked the "View data" navigation link (never disabled before,
   and it has nothing to race), and `pointer-events-none` only stops the mouse — a keyboard user could
   still Tab to another row's Manage and press Enter, starting exactly the concurrent request the guard
   was meant to prevent. `disabled={loadingDirectoryId !== null}` is back on Manage/Unarchive,
   byte-identical to the pre-PR behaviour, and "View data" is untouched.
3. **Columns moved to a module-level factory.** An inline `cell` arrow returning JSX reads as a nested
   component definition to Sonar (`typescript:S6478`), which flagged 2 issues here. The factory pattern
   the pretty-URLs and enterprise tables already use has 0 — so this aligns the series rather than
   suppressing anything.

Also un-exported `getWorkspaceTeamColumns`: the factory-export pattern exists so a `loading.tsx` can
share the columns, and this one's skeleton isn't a table, so nothing consumed it.

**Not changed:** `TeamPermissionMapping` renders English literals (`"Read"`, `"Manage"`). Real bug, but
pre-existing — the old code rendered the identical literal — and fixing it means new i18n keys, a shared
util's contract, and its unit test. Out of scope for a restyle; flagged as follow-up.

## Linear ticket

Ref ENG-762

## Screenshots

> **Component renders, not app screenshots** — no Docker here, so no database to boot the app against.
> Real markup, the repo's own compiled `globals.css`, and every class string through the same
> `tailwind-merge` the components use.
>
> **These are now known to be misleading in one respect**, and it is worth saying why: the harness drew
> the "after" state I *intended*, not the state the code produced. It models the table but not
> `feedback-directory-view.tsx`, so it could not show the missing frame that review caught. Treat them as
> illustrating the target, not as evidence the code reaches it.

| Before | After (intended) |
| --- | --- |
|  |  |

The in-flight dimming shot that was here has been removed — that behaviour no longer exists.

## How this was tested

- **Rebased after #8839 was squash-merged.** The squash gave `main` a different SHA from the commit this
  chain was stacked on, so the whole `settings-table/` primitive replayed as new files in #8840's diff
  (8 files, +622/−107, conflicted). #8840 was rebased onto the squashed `main` and this branch replayed
  on top; its diff is back to its own **4 files, +152/−106**. One drift adopted from `main` in passing:
  `getPrettyUrlColumns` is exported there, so that export stays.
- `npx turbo run typecheck --filter=@formbricks/web` → **12/12 tasks successful, 0 errors.** (Earlier
  runs in this series reported 33 pre-existing errors from `@formbricks/jobs` being unbuilt in my
  sandbox; turbo has since built it, so there is a genuinely clean baseline now — which matters more
  because `main` just gained a `pnpm typecheck` CI gate.)
- `npx eslint` on all four changed files → clean.
- `npx vitest run` on the settings-table suite → **25/25 pass**.
- `prettier --check` on every changed file → clean.
- **Queried the SonarCloud API across the whole series** rather than reading one warning in isolation:
  `#8839` 0 issues, `#8840` 0, `#8841` 2, `#8842` 2 — perfectly correlated with whether the columns were
  inline or in a module-level factory. That is what identified the fix as a pattern rather than a
  one-off.
- **Then verified the fix locally instead of waiting on CI.**
  `react/no-unstable-nested-components` is ESLint's analogue of `typescript:S6478` — it emits the same
  sentence Sonar does (*"move this component definition out of the parent component and pass data as
  props"*). Ran it over all six files in the series: **0 errors.** Also confirmed the rule genuinely
  fires on the flagged pattern, by running it against a four-line fixture holding an inline
  JSX-returning `cell` arrow — it flags that. So the check is meaningful, not vacuously passing.
- Verified every i18n key used already exists; no new keys, so `pnpm i18n` has nothing to regenerate.
- Confirmed no Playwright spec references either table's DOM.

**Still owed before merge:** a visual pass on a running app. This PR is the reason to insist on it — a
missing prop in a parent file is exactly what a component-render harness cannot catch.

## Breaking changes

None

## QA / Test Plan

**How to test**

Team access:

- [ ] Open Workspace settings → Team access → the table fills the card edge to edge, header band flush
      against the card's title divider, last row meeting the card's bottom border
- [ ] On that page, check the "Manage" button above the table → still inset from the card's edges with
      the same spacing as before → a button touching the card edge is the regression
- [ ] Read across a row → team name, size, team ID badge, permission (capitalised)
- [ ] Click the copy icon on a team ID badge → the ID is copied
- [ ] Hover a row → no highlight (rows are not interactive)
- [ ] Open the page for a workspace with no teams → "No teams found" as a single centred row, headers
      still visible

Feedback directories:

- [ ] Open Org settings → Feedback directories → **the table has a visible border and rounded corners,
      filling the card edge to edge** → no border at all is the bug this PR fixes; two parallel lines is
      the nested-frame bug
- [ ] Check the "Show archived" switch and "Create" button above the table → both inset from the card
      edges, not touching them
- [ ] Click "Manage" on a directory → its button shows a spinner and the other rows' Manage/Unarchive
      buttons become disabled → **the other rows do not dim, and "View data" on them stays clickable**
- [ ] While that request is in flight, Tab to another row's "Manage" and press Enter → nothing happens
      (the button is genuinely disabled, not just unclickable by mouse)
- [ ] While that request is in flight, click "View data" on a different row → it navigates normally
- [ ] Toggle "Show archived" → archived directories appear with a grey "Archived" badge and an
      "Unarchive" button
- [ ] Click "Unarchive" → same spinner and disabling behaviour → directory becomes Active
- [ ] Click "Unarchive" on a directory whose workspace already has another directory → the conflict
      toast appears and nothing changes
- [ ] Open the page for an org with no directories → the empty-state row shows, headers still visible

**Preconditions / test data**

- Enterprise license (both surfaces are EE)
- An org with at least 3 feedback directories, one archived, at least one linked to a workspace the test
  user can reach (so "View data" renders)
- A workspace with 2+ teams at different permission levels, plus one workspace with none
- For the unarchive conflict: two directories assigned to the same workspace, one archived

**Risks & regressions**

- The frame fix touches `feedback-directory-view.tsx`, which was outside the original diff. Confirm the
  card's own padding still looks right around the controls above the table.
- Action-button behaviour is now back to pre-PR parity, so the risk is the opposite of before: a button
  that *isn't* disabled during another row's request, or a "View data" link that *is*.
- The columns moved to a module-level factory. Same rendered output, but the actions column is the one
  with real branching (archived vs not, owner/manager vs not), so it's worth clicking through both
  states.

- none

### DIFF
--- a/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
+++ b/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
@@ -82,7 +82,10 @@ const getFeedbackDirectoryColumns = ({
     header: null,
     srLabel: t("common.actions"),
     headerClassName: "w-[25%]",
-    align: "right",
+    // No `align: "right"`. The buttons are block-level, so `text-align` cannot move them — the flex here
+    // is what right-aligns them, and it is also what right-aligns a skeleton bar if this array is ever
+    // fed to `SettingsTableSkeleton`. Declaring `align` too would leave two mechanisms for one intent,
+    // with the inert one looking authoritative.
     cellClassName: "flex justify-end gap-2",
     stopRowClick: true,
     cell: (directory) => (
--- a/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
+++ b/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import type { TFunction } from "i18next";
 import Link from "next/link";
 import { useRouter } from "next/navigation";
 import { useState } from "react";
@@ -22,8 +23,8 @@ import {
 import { TOrganizationWorkspace } from "@/modules/ee/teams/team-list/types/workspace";
 import { Badge } from "@/modules/ui/components/badge";
 import { Button } from "@/modules/ui/components/button";
+import { SettingsTable, type TSettingsTableColumn } from "@/modules/ui/components/settings-table";
 import { Switch } from "@/modules/ui/components/switch";
-import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@/modules/ui/components/table";
 
 interface FeedbackDirectoryTableProps {
   directories: TFeedbackDirectory[];
@@ -32,6 +33,95 @@ interface FeedbackDirectoryTableProps {
   workspaceAccessByWorkspace: TWorkspaceFeedbackDirectoryAccess[];
   membershipRole: TOrganizationRole;
 }
+
+/**
+ * Defined at module level rather than inside the component: an inline `cell` that returns JSX reads as a
+ * nested component definition to Sonar (typescript:S6478), and re-declaring the array per render buys
+ * nothing.
+ */
+const getFeedbackDirectoryColumns = ({
+  t,
+  isOwnerOrManager,
+  loadingDirectoryId,
+  viewDataWorkspaceIdByDirectory,
+  onManage,
+  onUnarchive,
+}: Readonly<{
+  t: TFunction;
+  isOwnerOrManager: boolean;
+  loadingDirectoryId: string | null;
+  viewDataWorkspaceIdByDirectory: Map<string, string>;
+  onManage: (directoryId: string) => void;
+  onUnarchive: (directoryId: string) => void;
+}>): TSettingsTableColumn<TFeedbackDirectory>[] => [
+  {
+    id: "name",
+    header: t("workspace.settings.feedback_directories.directory_name"),
+    headerClassName: "w-[45%]",
+    cell: (directory) => directory.name,
+  },
+  {
+    id: "workspaceCount",
+    header: t("common.workspaces"),
+    headerClassName: "w-[15%]",
+    cell: (directory) => directory.workspaceCount,
+  },
+  {
+    id: "status",
+    header: t("common.status"),
+    headerClassName: "w-[15%]",
+    cell: (directory) =>
+      directory.isArchived ? (
+        <Badge type="gray" size="tiny" text={t("common.archived")} />
+      ) : (
+        <Badge type="success" size="tiny" text={t("common.active")} />
+      ),
+  },
+  {
+    id: "actions",
+    header: null,
+    srLabel: t("common.actions"),
+    headerClassName: "w-[25%]",
+    align: "right",
+    cellClassName: "flex justify-end gap-2",
+    stopRowClick: true,
+    cell: (directory) => (
+      <>
+        {/* Never disabled: it is plain navigation with nothing to race. */}
+        {!directory.isArchived && viewDataWorkspaceIdByDirectory.has(directory.id) && (
+          <Button size="sm" variant="ghost" asChild>
+            <Link
+              href={`/workspaces/${viewDataWorkspaceIdByDirectory.get(directory.id)}/unify/feedback-records`}>
+              {t("workspace.settings.feedback_directories.view_data")}
+            </Link>
+          </Button>
+        )}
+        {/* `disabled` on the button itself, not `pointer-events-none` on the row: the latter blocks the
+            mouse but leaves the button focusable, so Enter would still start a second request. */}
+        {isOwnerOrManager && !directory.isArchived && (
+          <Button
+            size="sm"
+            variant="secondary"
+            loading={loadingDirectoryId === directory.id}
+            disabled={loadingDirectoryId !== null}
+            onClick={() => onManage(directory.id)}>
+            {t("common.manage")}
+          </Button>
+        )}
+        {isOwnerOrManager && directory.isArchived && (
+          <Button
+            size="sm"
+            variant="secondary"
+            loading={loadingDirectoryId === directory.id}
+            disabled={loadingDirectoryId !== null}
+            onClick={() => onUnarchive(directory.id)}>
+            {t("workspace.settings.feedback_directories.unarchive")}
+          </Button>
+        )}
+      </>
+    ),
+  },
+];
 
 export const FeedbackDirectoryTable = ({
   directories,
@@ -126,7 +216,8 @@ export const FeedbackDirectoryTable = ({
   return (
     <>
       {isOwnerOrManager && (
-        <div className="mb-4 flex items-center justify-between">
+        // The table runs edge to edge, so the controls above it carry the card's gutter themselves.
+        <div className="mb-4 flex items-center justify-between px-4 pt-4">
           <div className="flex items-center gap-2">
             <Switch checked={showArchived} onCheckedChange={setShowArchived} />
             <span className="text-sm text-slate-500">
@@ -139,72 +230,20 @@ export const FeedbackDirectoryTable = ({
         </div>
       )}
 
-      <div className="overflow-hidden rounded-lg border" aria-label="Feedback directories list">
-        <Table>
-          <TableHeader role="rowgroup">
-            <TableRow className="bg-slate-100" role="row">
-              <TableHead className="font-medium text-slate-500">
-                {t("workspace.settings.feedback_directories.directory_name")}
-              </TableHead>
-              <TableHead className="font-medium text-slate-500">{t("common.workspaces")}</TableHead>
-              <TableHead className="font-medium text-slate-500">{t("common.status")}</TableHead>
-              <TableHead></TableHead>
-            </TableRow>
-          </TableHeader>
-          <TableBody className="[&_tr:last-child]:border-b">
-            {filteredDirectories.length === 0 && (
-              <TableRow>
-                <TableCell colSpan={4} className="text-center">
-                  {t("workspace.settings.feedback_directories.empty_state")}
-                </TableCell>
-              </TableRow>
-            )}
-            {filteredDirectories.map((directory) => (
-              <TableRow key={directory.id}>
-                <TableCell>{directory.name}</TableCell>
-                <TableCell>{directory.workspaceCount}</TableCell>
-                <TableCell>
-                  {directory.isArchived ? (
-                    <Badge type="gray" size="tiny" text={t("common.archived")} />
-                  ) : (
-                    <Badge type="success" size="tiny" text={t("common.active")} />
-                  )}
-                </TableCell>
-                <TableCell className="flex justify-end gap-2">
-                  {!directory.isArchived && viewDataWorkspaceIdByDirectory.has(directory.id) && (
-                    <Button size="sm" variant="ghost" asChild>
-                      <Link
-                        href={`/workspaces/${viewDataWorkspaceIdByDirectory.get(directory.id)}/unify/feedback-records`}>
-                        {t("workspace.settings.feedback_directories.view_data")}
-                      </Link>
-                    </Button>
-                  )}
-                  {isOwnerOrManager && !directory.isArchived && (
-                    <Button
-                      size="sm"
-                      variant="secondary"
-                      loading={loadingDirectoryId === directory.id}
-                      disabled={loadingDirectoryId !== null}
-                      onClick={() => handleManageDirectory(directory.id)}>
-                      {t("common.manage")}
-                    </Button>
-                  )}
-                  {isOwnerOrManager && directory.isArchived && (
-                    <Button
-                      size="sm"
-                      variant="secondary"
-                      loading={loadingDirectoryId === directory.id}
-                      disabled={loadingDirectoryId !== null}
-                      onClick={() => handleUnarchiveDirectory(directory.id)}>
-                      {t("workspace.settings.feedback_directories.unarchive")}
-                    </Button>
-                  )}
-                </TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </div>
+      <SettingsTable
+        columns={getFeedbackDirectoryColumns({
+          t,
+          isOwnerOrManager,
+          loadingDirectoryId,
+          viewDataWorkspaceIdByDirectory,
+          onManage: handleManageDirectory,
+          onUnarchive: handleUnarchiveDirectory,
+        })}
+        rows={filteredDirectories}
+        getRowId={(directory) => directory.id}
+        emptyMessage={t("workspace.settings.feedback_directories.empty_state")}
+        aria-label={t("workspace.settings.feedback_directories.title")}
+      />
 
       {openCreateModal && (
         <FeedbackDirectorySettingsModal
--- a/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
+++ b/apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
@@ -82,14 +82,15 @@ const getFeedbackDirectoryColumns = ({
     header: null,
     srLabel: t("common.actions"),
     headerClassName: "w-[25%]",
-    // No `align: "right"`. The buttons are block-level, so `text-align` cannot move them — the flex here
-    // is what right-aligns them, and it is also what right-aligns a skeleton bar if this array is ever
-    // fed to `SettingsTableSkeleton`. Declaring `align` too would leave two mechanisms for one intent,
-    // with the inert one looking authoritative.
-    cellClassName: "flex justify-end gap-2",
     stopRowClick: true,
+    // The flex belongs on a wrapper inside the cell, never on `cellClassName` — that class lands on the
+    // `<td>`, and `display: flex` there stops it being a table-cell, which silently kills the shared
+    // `align-middle` (`vertical-align` applies only to inline-level and table-cell boxes) and makes the
+    // browser wrap the cell in an anonymous table-cell defaulting to baseline. The buttons would sit high
+    // in the row instead of centred. `align: "right"` is not an alternative here: the wrapper is
+    // block-level, so `text-align` cannot move it.
     cell: (directory) => (
-      <>
+      <div className="flex justify-end gap-2">
         {/* Never disabled: it is plain navigation with nothing to race. */}
         {!directory.isArchived && viewDataWorkspaceIdByDirectory.has(directory.id) && (
           <Button size="sm" variant="ghost" asChild>
@@ -121,7 +122,7 @@ const getFeedbackDirectoryColumns = ({
             {t("workspace.settings.feedback_directories.unarchive")}
           </Button>
         )}
-      </>
+      </div>
     ),
   },
 ];
--- a/apps/web/modules/ee/teams/workspace-teams/components/access-table.tsx
+++ b/apps/web/modules/ee/teams/workspace-teams/components/access-table.tsx
@@ -1,51 +1,55 @@
 "use client";
 
+import type { TFunction } from "i18next";
 import { useTranslation } from "react-i18next";
 import { TeamPermissionMapping } from "@/modules/ee/teams/utils/teams";
 import { TWorkspaceTeam } from "@/modules/ee/teams/workspace-teams/types/team";
 import { IdBadge } from "@/modules/ui/components/id-badge";
-import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@/modules/ui/components/table";
+import { SettingsTable, type TSettingsTableColumn } from "@/modules/ui/components/settings-table";
 
 interface AccessTableProps {
   teams: TWorkspaceTeam[];
 }
 
-export const AccessTable = ({ teams }: AccessTableProps) => {
+const getWorkspaceTeamColumns = (t: TFunction): TSettingsTableColumn<TWorkspaceTeam>[] => [
+  {
+    id: "name",
+    header: t("workspace.teams.team_name"),
+    headerClassName: "w-[30%]",
+    cellClassName: "font-medium",
+    cell: (team) => team.name,
+  },
+  {
+    id: "size",
+    header: t("common.size"),
+    headerClassName: "w-[15%]",
+    cell: (team) => t("common.count_members", { count: team.memberCount }),
+  },
+  {
+    id: "teamId",
+    header: t("common.team_id"),
+    headerClassName: "w-[30%]",
+    cell: (team) => <IdBadge id={team.id} />,
+  },
+  {
+    id: "permission",
+    header: t("workspace.teams.permission"),
+    headerClassName: "w-[25%]",
+    cellClassName: "capitalize",
+    cell: (team) => TeamPermissionMapping[team.permission],
+  },
+];
+
+export const AccessTable = ({ teams }: Readonly<AccessTableProps>) => {
   const { t } = useTranslation();
 
   return (
-    <div className="overflow-hidden rounded-lg">
-      <Table>
-        <TableHeader>
-          <TableRow className="bg-slate-100">
-            <TableHead className="font-medium text-slate-500">{t("workspace.teams.team_name")}</TableHead>
-            <TableHead className="font-medium text-slate-500">{t("common.size")}</TableHead>
-            <TableHead className="font-medium text-slate-500">{t("common.team_id")}</TableHead>
-            <TableHead className="font-medium text-slate-500">{t("workspace.teams.permission")}</TableHead>
-          </TableRow>
-        </TableHeader>
-        <TableBody className="[&_tr:last-child]:border-b">
-          {teams.length === 0 && (
-            <TableRow>
-              <TableCell colSpan={4} className="text-center">
-                {t("workspace.teams.no_teams_found")}
-              </TableCell>
-            </TableRow>
-          )}
-          {teams.map((team) => (
-            <TableRow key={team.id}>
-              <TableCell className="font-medium">{team.name}</TableCell>
-              <TableCell>{t("common.count_members", { count: team.memberCount })}</TableCell>
-              <TableCell>
-                <IdBadge id={team.id} />
-              </TableCell>
-              <TableCell>
-                <p className="capitalize">{TeamPermissionMapping[team.permission]}</p>
-              </TableCell>
-            </TableRow>
-          ))}
-        </TableBody>
-      </Table>
-    </div>
+    <SettingsTable
+      columns={getWorkspaceTeamColumns(t)}
+      rows={teams}
+      getRowId={(team) => team.id}
+      emptyMessage={t("workspace.teams.no_teams_found")}
+      aria-label={t("common.team_access")}
+    />
   );
 };
--- a/apps/web/modules/ui/components/settings-table/types.ts
+++ b/apps/web/modules/ui/components/settings-table/types.ts
@@ -15,7 +15,15 @@ export type TSettingsTableColumn<TRow> = {
   /** Already-translated header label. `null` renders an empty header cell — pair it with `srLabel`. */
   header: ReactNode;
   cell: (row: TRow, index: number) => ReactNode;
-  /** Applies to the header cell and every body cell in this column. Defaults to `left`. */
+  /**
+   * Sets `text-align` on the header cell and every body cell in this column. Defaults to `left`.
+   *
+   * It moves **inline** content only. A cell holding block-level content — buttons, a link, a badge —
+   * is unaffected, and needs a flex `cellClassName` (`flex justify-end`) to move instead. Don't declare
+   * both for one column: the flex wins, and the `align` sitting next to it reads as load-bearing while
+   * being inert. The skeleton draws its bar through the same two class sources, so whichever mechanism
+   * aligns the real cell is also the one aligning the skeleton.
+   */
   align?: TSettingsTableAlign;
   /** Hides the column below this breakpoint. Applied to the header and body cells together. */
   hideBelow?: TSettingsTableBreakpoint;
--- a/apps/web/modules/ui/components/settings-table/types.ts
+++ b/apps/web/modules/ui/components/settings-table/types.ts
@@ -18,11 +18,10 @@ export type TSettingsTableColumn<TRow> = {
   /**
    * Sets `text-align` on the header cell and every body cell in this column. Defaults to `left`.
    *
-   * It moves **inline** content only. A cell holding block-level content — buttons, a link, a badge —
-   * is unaffected, and needs a flex `cellClassName` (`flex justify-end`) to move instead. Don't declare
-   * both for one column: the flex wins, and the `align` sitting next to it reads as load-bearing while
-   * being inert. The skeleton draws its bar through the same two class sources, so whichever mechanism
-   * aligns the real cell is also the one aligning the skeleton.
+   * It moves **inline-level** content, which covers text, badges and `Button`/`Link` (both `inline-flex`).
+   * It cannot move a block-level child, so a cell whose content is a wrapper `<div>` — or the skeleton's
+   * placeholder bar — is unaffected by it. Declaring it alongside a flex wrapper is therefore inert; pick
+   * one and let it be the only alignment source for the column.
    */
   align?: TSettingsTableAlign;
   /** Hides the column below this breakpoint. Applied to the header and body cells together. */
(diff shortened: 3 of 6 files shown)


apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
```diff
@@ -139,72 +139,80 @@ export const FeedbackDirectoryTable = ({
...
-                      {t("workspace.settings.feedback_directories.unarchive")}
-                    </Button>
-                  )}
-                </TableCell>
-              </TableRow>
-            ))}
-          </TableBody>
-        </Table>
-      </div>
+      <SettingsTable
```
mattinannt, line 237: **Table loses its border/frame — the caller wasn't updated to match.**

`SettingsTable` defaults to `frame="none"` (per its own doc: "the table draws no frame, because a `<SettingsCard bodyVariant=\"flush\">` already frames it"), and this call doesn't pass `frame="card"` either. The old code supplied its own frame directly: `<div className="overflow-hidden rounded-lg border" ...><Table>...`, which is now removed.

The parent that renders this component, `feedback-directory-view.tsx`, wasn't touched by this PR and still does:
```tsx
<SettingsCard
  title={t("workspace.settings.feedback_directories.title")}
  description={t("workspace.settings.feedback_directories.description")}>
  <FeedbackDirectoryTable ... />
</SettingsCard>
```
— no `bodyVariant="flush"`, so it defaults to `"padded"` (`px-4 pt-4`, meant for "ordinary form content", not a table).

Net effect: the feedback directories table will render with no border/rounded corners and an inconsistent padding gutter around it — unlike the other two tables converted in this PR series, where the sibling `access-view.tsx` was explicitly updated in this same PR to add `bodyVariant="flush"`, and the enterprise-features table already had it. This one was missed.

Fix: either add `bodyVariant="flush"` to the `SettingsCard` in `feedback-directory-view.tsx` (matching the `access-view.tsx` pattern in this PR), or pass `frame="card"` here.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
```diff
@@ -139,72 +139,80 @@ export const FeedbackDirectoryTable = ({
...
+            ),
+          },
+        ]}
+        rows={filteredDirectories}
+        getRowId={(directory) => directory.id}
+        emptyMessage={t("workspace.settings.feedback_directories.empty_state")}
+        // While one directory's action is in flight, the *other* rows are blocked so a second request
+        // cannot race it — this replaces `disabled={loadingDirectoryId !== null}` on each Button. The
+        // acting row is deliberately excluded, so its spinner stays at full opacity.
+        isRowDisabled={(directory) => loadingDirectoryId !== null && loadingDirectoryId !== directory.id}
```
mattinannt: `isRowDisabled` sets `pointer-events-none` on the whole `<tr>` (see `SettingsTable`'s `isDisabled && "pointer-events-none opacity-60"` on the row), so this blocks *every* interactive element in a non-acting row — including the "View data" `Link`/`Button` in the `actions` column.

Before this change, only the "Manage"/"Unarchive" buttons carried `disabled={loadingDirectoryId !== null}`; the "View data" button was never disabled by another row's loading state:
```tsx
<Button size="sm" variant="ghost" asChild>
  <Link href={...}>{t("workspace.settings.feedback_directories.view_data")}</Link>
</Button>
```
So for an owner/manager, clicking "Manage" on directory A now also blocks navigating via "View data" on an unrelated, non-archived directory B until A's request resolves — a real (if narrow) regression, since "View data" is a pure navigation link with nothing to race. The PR description flags this exact tradeoff as worth confirming; it reads as unintended rather than a deliberate product decision.

If the intent is only to prevent a second Manage/Unarchive request from racing the first, consider disabling just those two buttons based on `loadingDirectoryId !== null && loadingDirectoryId !== directory.id`, and leaving "View data" (and the row's `pointer-events`) untouched.

---

mattinannt (commented):
Single-pass review (high effort, no Agent-tool fan-out / subagent verify pass available in this context) of the actual PR diff — `git diff 306013f...8b7f5fc` (the PR's one commit against its stacked base), scoped to `access-table.tsx`, `access-view.tsx`, `feedback-directory-table.tsx`. `EnterpriseLicenseFeaturesTable.tsx` is out of scope (that's the prior stacked PR #8840, already in this PR's base). Two findings posted inline; both concrete, both in the feedback-directories table.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
mattinannt: `isRowDisabled` dims a non-acting row with `pointer-events-none opacity-60` on the `<tr>`, but the Manage/Unarchive `Button`s inside it no longer get a native `disabled` (that prop was dropped when moving off per-button `disabled={loadingDirectoryId !== null}`). `pointer-events: none` only suppresses pointer/mouse hit-testing — it does not remove a button from the tab order or block the synthetic `click` a keyboard Enter/Space fires on a focused element. So a keyboard user can still Tab to another row's Manage/Unarchive button while a request is in flight and activate it, racing the in-flight request — exactly what this comment says can't happen.

This exact caveat is already documented in this codebase, just for a different code path: `apps/web/modules/ui/components/settings-table/lib/column-classes.ts` (`isRowActivatable`) calls out that a disabled row's activator button "stays in the tab order" and must be excluded from being clickable for this reason. That protection covers the primitive's own row-activator button, not these per-cell `Button`s in the actions column — so the same failure mode reappears here without the same guard.

Worth adding `disabled={isDisabled}` (or passing it down) to the Manage/Unarchive buttons themselves, in addition to the row dimming.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
mattinannt, line 237: `SettingsTable` here uses the default `frame="none"` (no border/box of its own — see `settings-table/types.ts`: "the `<SettingsCard bodyVariant="flush">` it usually sits in already provides the frame"). But `feedback-directory-view.tsx` (not touched by this PR) still renders `<SettingsCard title=... description=...>` with no `bodyVariant`, i.e. the default `"padded"` (`px-4 pt-4`, no `overflow-hidden`).

Previously the table supplied its own `overflow-hidden rounded-lg border` wrapper, so it always had a visible box. Now, with neither the table nor the card providing a frame, the table renders with no border/box at all, inset by the card's `px-4 pt-4` — not edge-to-edge as the PR description's "after" screenshot and `AccessTable`/`access-view.tsx` (which *did* get `bodyVariant="flush"` in this same PR) would suggest.

Looks like `feedback-directory-view.tsx`'s `SettingsCard` needs the same `bodyVariant="flush"` treatment `access-view.tsx` got, or this table needs `frame="card"`.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
```diff
@@ -139,72 +139,80 @@ export const FeedbackDirectoryTable = ({
...
+            id: "actions",
+            header: null,
+            srLabel: t("common.actions"),
+            headerClassName: "w-[25%]",
+            align: "right",
+            cellClassName: "flex justify-end gap-2",
+            stopRowClick: true,
+            cell: (directory) => (
+              <>
+                {!directory.isArchived && viewDataWorkspaceIdByDirectory.has(directory.id) && (
```
mattinannt: Minor correction to the PR description's "Risks & regressions" note: it says the "View data" link being blocked by `isRowDisabled` "was already `disabled` before" — but in the pre-PR code (left side of this diff), the "View data" `Button`/`Link` never had a `disabled` prop at all, only the Manage/Unarchive buttons did (`disabled={loadingDirectoryId !== null}`). So blocking "View data" on other rows while an unrelated Manage/Unarchive request is in flight is new behavior introduced by moving the guard to the whole row via `isRowDisabled`, not a preserved one. Not blocking, since it's already flagged as something to confirm in QA — just flagging the premise so it gets tested with the right expectation.

---

mattinannt (commented):
Reviewed the `SettingsTable` conversion for team access and feedback directories. Two confirmed issues around the new `isRowDisabled` row-blocking behavior in `feedback-directory-table.tsx` (a keyboard-activation gap, and a missing `bodyVariant="flush"` on the parent card that leaves the table frameless with no box), plus a factual correction to the PR description. `access-table.tsx`/`access-view.tsx` conversion looked solid.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
mattinannt: `isRowDisabled` only adds `pointer-events-none opacity-60` (+ `aria-disabled`) to the ancestor `<TableRow>` (`settings-table/index.tsx` lines 78-82) — it never disables the `<Button>` elements themselves, since this actions column uses `stopRowClick: true` with raw `<Button>`s rather than going through the table's own row-activator mechanism. `pointer-events-none` only blocks mouse hit-testing; it does not remove a button from the tab order or stop a keyboard-triggered click.

Concretely: click "Manage" on directory A (`loadingDirectoryId = A.id`). While that request is in flight, Tab to directory B's "Manage"/"Unarchive" button — it has no `disabled` attribute (the removed `disabled={loadingDirectoryId !== null}` was the only thing enforcing this before) — and press Enter/Space. That fires `handleManageDirectory(B.id)` concurrently with A's still-pending request, which is exactly the race the PR description says is preserved ("Same guarantee as before — a second request cannot race the first").

This is the same pitfall the codebase documents for the table's built-in row-activator button in `settings-table/lib/column-classes.ts` lines 58-63 ("Dimming it with `pointer-events-none` only stops the mouse... so a 'disabled' row would still be operable from the keyboard") — but that guard doesn't cover this case, because it only applies to the activator button on non-`stopRowClick` columns.

Worth restoring `disabled={loadingDirectoryId !== null}` on these buttons alongside `isRowDisabled` (or otherwise gating the click handlers) so keyboard users get the same protection as mouse users.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
mattinannt, line 237: `SettingsTable` defaults to `frame="none"` here, and per its own doc comment (`settings-table/index.tsx` lines 20-21) that's meant to be used inside a `<SettingsCard bodyVariant="flush">`, which supplies the border/rounding instead.

`FeedbackDirectoryView` (the `SettingsCard` wrapping this table — `feedback-directory-view.tsx`) wasn't updated to `bodyVariant="flush"` the way `AccessView` was in this same PR (`access-view.tsx` now passes `bodyVariant="flush"`). `SettingsCard`'s default `bodyVariant` is `"padded"`, which doesn't draw the frame `FRAME_CLASSES` would otherwise supply. Net effect: this table loses its previous `overflow-hidden rounded-lg border` box entirely with no replacement, contradicting the PR's own QA checklist item "table renders edge to edge inside its card" and the PR description's claim that "the table gains the card's frame."

Since `feedback-directory-view.tsx` isn't in this PR's diff, flagging here — it'll need `bodyVariant="flush"` added there (and its own padding audit, mirroring what `access-view.tsx` did) for this to render correctly.

---

apps/web/modules/ee/teams/workspace-teams/components/access-table.tsx
```diff
@@ -1,51 +1,55 @@
...
 import { IdBadge } from "@/modules/ui/components/id-badge";
-import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from "@/modules/ui/components/table";
+import { SettingsTable, type TSettingsTableColumn } from "@/modules/ui/components/settings-table";
 
 interface AccessTableProps {
   teams: TWorkspaceTeam[];
 }
 
-export const AccessTable = ({ teams }: AccessTableProps) => {
+export const getWorkspaceTeamColumns = (t: TFunction): TSettingsTableColumn<TWorkspaceTeam>[] => [
```
mattinannt: Nit: `getWorkspaceTeamColumns` is exported but has no consumer outside this file. The `SettingsTable` doc comment recommends this factory pattern specifically so `loading.tsx` can reuse the same columns for its skeleton, but `apps/web/modules/ee/teams/workspace-teams/loading.tsx` still renders its own hand-rolled skeleton instead of `SettingsTableSkeleton` fed by this factory. Either wire it up there or drop the export until something needs it.

---

mattinannt (changes requested):
Mechanical conversion mostly checks out, but two things regress behavior the PR description claims is unchanged:

1. **Keyboard users can still race the in-flight request** — `isRowDisabled` only dims the row via `pointer-events-none`, which doesn't stop keyboard-triggered clicks on the action buttons (the removed per-button `disabled` was the only thing enforcing that). See inline comment.
2. **The feedback-directories table loses its border entirely** — `SettingsTable` defaults to frameless on the assumption the parent `SettingsCard` is `bodyVariant="flush"`, but `FeedbackDirectoryView` wasn't updated to `flush` the way `AccessView` was for the team-access table. See inline comment.

Plus one minor nit on a dead export in `access-table.tsx`.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
itsjavi: Confirmed and fixed in 588eb06 — this was a real bug and the diagnosis is exactly right.

I converted the table to `SettingsTable` (default `frame="none"`) and removed its own `overflow-hidden rounded-lg border` wrapper, but never touched `feedback-directory-view.tsx`. Verified: its `SettingsCard` had no `bodyVariant`, so it defaulted to `"padded"`. Net result was a table with **no frame at all**, inset by a `px-4 pt-4` gutter — not what the PR description or the screenshot claimed.

I updated the parent rather than passing `frame="card"`, matching `access-view.tsx`: the card owning the frame is the whole point of #8838, and `frame="card"` would have reintroduced the nested double border for a table that lives inside a card.

Also gave the switch/create-button row above the table its own `px-4 pt-4`, since a flush card body no longer supplies the gutter — same trade as `access-view.tsx` and `teams-table.tsx`.

Worth noting how this slipped through: my screenshots are component renders, not app screenshots, so the harness drew the "after" state I *intended* rather than the state the code actually produced. A render harness can't catch a missing prop in a file it doesn't model. That's a real limitation of the evidence I attached, and this is the second thing in this series a running app would have caught first.

Two of the other threads on this PR make the same point — resolving those as duplicates and pointing at this one.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
itsjavi: Both of your findings on this cell are right, and they have the same root cause — I've reverted the mechanism entirely in 588eb06 rather than patching around it.

**The keyboard race (this thread):** correct, and it's the same bug I'd just fixed one layer up in `isRowActivatable`. I moved the guard from the buttons to the row, which swapped a real `disabled` attribute for `pointer-events-none` — mouse-only. Tab to another row's Manage and press Enter and you'd start a concurrent request. You even pointed at my own comment documenting exactly this pitfall; I wrote that guard for the primitive's activator button and then reintroduced the same hole for per-cell buttons.

**The "View data" regression ([sibling thread](https://github.com/formbricks/formbricks/pull/8841#discussion_r3782132068)):** also correct. `pointer-events-none` on the `<tr>` blocked a pure navigation link that was never disabled before. Nothing to race there.

Fix: **`isRowDisabled` is gone from this table.** `disabled={loadingDirectoryId !== null}` is back on Manage/Unarchive — byte-identical to the pre-PR behaviour — and "View data" is untouched, with a comment saying why it stays enabled. That fixes both threads at once and restores exact parity, which is what the PR was supposed to preserve.

I also corrected the PR description, which claimed "View data" "was already `disabled` before". [Your third thread](https://github.com/formbricks/formbricks/pull/8841#discussion_r3782139728) is right that it never had a `disabled` prop at all — my premise was wrong, so the QA note was steering a tester toward the wrong expectation.

Consequence worth flagging: `isRowDisabled` now has **no consumer** anywhere in the series. Given I removed `containerClassName` from #8837 for exactly that reason, consistency says this should probably go too — along with the `isDisabled` branch of `isRowActivatable`. I've left it for now only because that branch *is* the a11y guard you asked for on #8839, and deleting it would look like reverting a fix. Happy to strip both if you'd rather the primitive carried no speculative API; say the word.

---

apps/web/modules/ee/teams/workspace-teams/components/access-table.tsx
itsjavi: Fair — dropped the export in 588eb06; `getWorkspaceTeamColumns` is now module-local.

I took the "drop it until something needs it" option rather than wiring `loading.tsx`, because that skeleton isn't a table at all — it's avatar circles and stacked bars (`workspace-teams/loading.tsx`), so feeding it the column array would mean redesigning the loading state, which is a bigger change than this PR should carry. Worth doing, just not here.

Same nit applied to `getPrettyUrlColumns` in #8839, so I un-exported that too (ce69643) rather than leaving the pair inconsistent. The enterprise one was already module-local.

The factory-with-export pattern only earns its keep once a `loading.tsx` actually consumes it — Tags (PR 9 in the series) is the first table where that's true, since its skeleton already mirrors the real header markup and currently drifts from it.

---

apps/web/modules/ee/teams/workspace-teams/components/access-table.tsx
```diff
@@ -1,51 +1,55 @@
...
+    header: t("common.team_id"),
+    headerClassName: "w-[30%]",
+    cell: (team) => <IdBadge id={team.id} />,
+  },
+  {
+    id: "permission",
+    header: t("workspace.teams.permission"),
+    headerClassName: "w-[25%]",
+    cellClassName: "capitalize",
+    cell: (team) => TeamPermissionMapping[team.permission],
```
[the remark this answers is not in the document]
itsjavi: Real bug, but **not introduced here — declining for this PR** and flagging it as follow-up.

Verified: `TeamPermissionMapping` in `modules/ee/teams/utils/teams.ts:4-8` is a hardcoded English map (`"Read"`, `"Read & write"`, `"Manage"`), and the pre-PR code rendered the identical literal — `<p className="capitalize">{TeamPermissionMapping[team.permission]}</p>`. The conversion moved that expression into a `cell`, unchanged. So non-English locales showed untranslated permissions before this PR and still do; nothing regressed.

Reasons to fix it separately rather than here:

- It changes a **shared util's contract**, not this table. `TeamPermissionMapping` is also asserted against the English strings in `modules/ee/teams/utils/teams.test.ts:8-10`, so a proper fix means new i18n keys, an i18n-aware helper, and updating that test.
- ENG-762's guardrail is explicit that copy and i18n keys stay unchanged unless a new label is genuinely required. Slipping a localisation change into a visual-unification PR makes the diff harder to reason about and puts a copy change behind a "restyle" title.
- `TeamRoleMapping` right below it has the same problem (`"Team Admin"`, `"Contributor"`), so the honest fix covers both — which is clearly its own change.

Happy to open that as a small follow-up if you want it tracked; it's roughly 4-6 new keys plus a `getTeamPermissionLabel(permission, t)` helper.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
itsjavi: You're right and thanks for flagging the premise rather than just the code — a wrong "Risks" note is worse than none, because it tells the tester to expect the regression as intended behaviour.

Confirmed against the pre-PR code: only Manage/Unarchive carried `disabled={loadingDirectoryId !== null}`; the "View data" `Button`/`Link` had no `disabled` prop at all. So blocking it was new behaviour I introduced, not a preserved one.

Both the claim and the cause are gone in 588eb06: `isRowDisabled` is removed, per-button `disabled` is back, and "View data" is never disabled. I rewrote that section of the description so QA is no longer pointed at the wrong expectation.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
```diff
@@ -33,6 +34,95 @@ interface FeedbackDirectoryTableProps {
...
+        <Badge type="gray" size="tiny" text={t("common.archived")} />
+      ) : (
+        <Badge type="success" size="tiny" text={t("common.active")} />
+      ),
+  },
+  {
+    id: "actions",
+    header: null,
+    srLabel: t("common.actions"),
+    headerClassName: "w-[25%]",
```
mattinannt, line 84: `align: "right"` is dead configuration here — `cellClassName: "flex justify-end gap-2"` already right-aligns the buttons via flexbox, and `text-right` has no effect on a `display:flex` cell with block-level Button/Link children. Not a bug today, but if this column definition is ever reused for a `SettingsTableSkeleton` (the shared-factory pattern this PR's own doc comment recommends), a future maintainer could reasonably expect `align: "right"` to be load-bearing for the skeleton bar's position — removing `cellClassName` while trusting `align` to still right-align would silently left-align it.

---

mattinannt (commented):
Re-reviewed at the latest commit. Both changes-requested items from the prior review are correctly fixed: `bodyVariant="flush"` is now on `feedback-directory-view.tsx`'s `SettingsCard`, and `disabled={loadingDirectoryId !== null}` is back on Manage/Unarchive while "View data" stays clickable during an in-flight request — verified against the actual keyboard-activation and frame concerns raised earlier. Traced every hunk against `SettingsTable`/column-classes, confirmed no dangling references to the old DOM structure (`role="row"`/`role="rowgroup"`), and checked that all i18n keys referenced already exist in `en-US.json`. One cosmetic-only nit left inline; nothing blocking.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-table.tsx
itsjavi: Valid, and the hazard is worse than "not a bug today" — fixed in `78caec9`.

I traced it rather than taking it on faith, and the mechanism is exactly as you describe. `getBodyCellClassName` composes `align` and `cellClassName` into one string, so this column's cell renders as:

```
text-right flex justify-end gap-2
```

`text-align` moves inline content. The children are block-level `Button`/`Link`, so `text-right` is inert and `justify-end` does all the work — on the `<th>` too, since `header: null` means there is no visible text for it to move either. Dead in both places.

**The skeleton point is the important half, and it's sharper than "a future maintainer could expect".** `SettingsTableSkeletonRows` builds its cell from `getBodyCellClassName(column)` — the *same* function — and puts a block-level `<Skeleton>` inside. So the skeleton bar's position is *already* governed by `justify-end` and not by `align`. Dropping `cellClassName` while trusting `align` wouldn't just left-align the buttons; it would left-align the skeleton bar too, in the one code path whose entire purpose is to stand in for the real thing. Two mechanisms, one intent, and the inert one looking authoritative.

So I removed `align: "right"` and left a comment saying the flex owns the alignment, including for the skeleton.

**Also fixed the same pairing in `teams-table.tsx` on #8842**, which had `align: "right"` + `cellClassName: "flex justify-end"` and `header: null` — identical case, and it would have been odd to fix one and ship the other.

**And documented it on the prop itself**, since a comment on two call sites doesn't stop the third:

```ts
/**
 * Sets `text-align` on the header cell and every body cell in this column. Defaults to `left`.
 *
 * It moves **inline** content only. A cell holding block-level content — buttons, a link, a badge —
 * is unaffected, and needs a flex `cellClassName` (`flex justify-end`) to move instead. Don't declare
 * both for one column: the flex wins, and the `align` sitting next to it reads as load-bearing while
 * being inert. The skeleton draws its bar through the same two class sources, so whichever mechanism
 * aligns the real cell is also the one aligning the skeleton.
 */
```

That's a change to `types.ts`, which #8841 didn't otherwise touch — slightly outside the diff, but a two-call-site fix for a trap that's a property of the API seemed like the wrong shape.

**One column keeps both, deliberately:** the members actions column added to #8842 has `header: t("common.actions")` — *visible* text — so there `align: "right"` is genuinely load-bearing on the `<th>` and nothing else can do that job. I gave it an explicit `cellClassName: "flex justify-end"` as well, so the cell no longer depends on `MemberActions`' internal flex, and commented that the two props are doing two different jobs rather than competing for one. That's the distinction: your finding is about `align` being *inert*, not about the two props being mutually exclusive.

Verification: typecheck 12/12 tasks 0 errors, 25/25 unit tests, eslint and `prettier --check` clean on both branches. This is a class-only change with no visual delta — `justify-end` was already doing the aligning, so the rendered output is identical.

---

apps/web/modules/ui/components/settings-table/types.ts
```diff
@@ -15,7 +15,15 @@ export type TSettingsTableColumn<TRow> = {
...
+  /**
+   * Sets `text-align` on the header cell and every body cell in this column. Defaults to `left`.
+   *
+   * It moves **inline** content only. A cell holding block-level content — buttons, a link, a badge —
+   * is unaffected, and needs a flex `cellClassName` (`flex justify-end`) to move instead. Don't declare
+   * both for one column: the flex wins, and the `align` sitting next to it reads as load-bearing while
+   * being inert. The skeleton draws its bar through the same two class sources, so whichever mechanism
+   * aligns the real cell is also the one aligning the skeleton.
+   */
   align?: TSettingsTableAlign;
```
mattinannt, line 26: Nit, non-blocking: this guard against pairing `align` with a flex `cellClassName` is enforced only by this comment, not by the type system. Nothing in `TSettingsTableColumn` or in `column-classes.ts`'s class builders flags the redundant/inert combination, so a future column author who authors both (or drops the flex from `cellClassName` during a later refactor while keeping `align`) can silently reintroduce the exact bug this commit fixes. Not asking for a change here — just flagging it as a possible follow-up (e.g. a runtime dev-mode warning, or narrowing the type so `align` and a flex `cellClassName` can't coexist).

---

mattinannt (commented):
Re-reviewed at the latest commit (`78caec9`), which only adds the `align`-vs-flex-`cellClassName` JSDoc and the corresponding fix in `feedback-directory-table.tsx` (dropping the now-redundant `align: "right"`). Diffed both migrated tables and the `SettingsTable`/`column-classes` infrastructure again — no correctness bugs, no dangling references to the old table markup, i18n keys all resolve. One non-blocking design nit left inline: the align/flex conflict this commit documents is comment-enforced rather than type- or runtime-enforced. Nothing blocking merge.

---

apps/web/modules/ui/components/settings-table/types.ts
itsjavi: Taking you up on it rather than filing it — implemented in `37dd3b2a`. You're right that a guard enforced only by the comment describing it is not a guard, and this one has the strongest possible evidence against it: **I documented the constraint and then violated it in four column definitions.** Prose demonstrably didn't hold.

**First, the rule changed under this comment**, which is worth knowing before reading the fix. You commented at 13:22 against `78caec9`, where the rule was "don't pair `align` with a flex `cellClassName`". `091796ca` (two minutes later, from [a finding on #8871](https://github.com/formbricks/formbricks/pull/8871#discussion_r3783780718)) replaced it with something simpler and stricter: **no display utility on `cellClassName` or `headerClassName`, ever.** The flex was never legitimate there — it stops the `<td>` being a table cell, which kills `align-middle` and leaves content baseline-aligned. So there are now zero flex `cellClassName`s in the repo, and the pairing you describe can't arise.

That reframes enforcement usefully: instead of expressing "these two props are mutually exclusive in one direction", the check is "neither prop may contain a display utility" — which subsumes your case and is far easier to verify.

**On your two suggestions, and why I took a third route:**

- *Narrowing the type* — I don't think it's reachable. Detecting `flex` inside an arbitrary `string` needs template-literal gymnastics that would also reject legitimate strings containing those substrings, and it can't see through a variable or a `cn()` call either.
- *A runtime dev-mode warning* — blocked by this repo's own rules. `NODE_ENV` is **deliberately** excluded from the `FRAMEWORK_INJECTED_ENV_VARS` allowlist in `eslint.config.mjs:21-23`, and `lib/env-access.test.ts:65` asserts that `process.env.NODE_ENV` in a `modules/` file is a lint error. So there's no sanctioned way to gate a warning to dev from inside `column-classes.ts`, and an ungated `console.warn` would fire in production.

**So: a test, in the pattern this repo already uses for exactly this kind of constraint.** `lib/turbo-build-env.test.ts` and `lib/env-access.test.ts` both enforce invariants by reading source, so this isn't a new idea here. It walks `apps/web/modules` and `apps/web/app`, pulls every `cellClassName`/`headerClassName` string literal, and fails on any Tailwind display utility — stripping variant prefixes, so `sm:flex` is caught as well as bare `flex`.

**Mutation-checked, because a guard that passes vacuously is worse than none.** Reintroducing the original bug:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+ ["modules/ee/feedback-directory/components/feedback-directory-table.tsx —
   cellClassName: \"flex justify-end gap-2\" sets flex"]
```

`headerClassName: "w-[25%] sm:flex"` is caught the same way. There's also a second test asserting the walk actually reaches `feedback-directory-table.tsx` and `column-classes.ts`, so a silently empty directory scan can't make the first test pass for the wrong reason — that failure mode is exactly what makes source-scanning tests untrustworthy otherwise.

27 tests in that file now, up from 25.

**What it does not do**, stated in the test's own comment so nobody mistakes its reach: it reads source text, so it only catches the literal form every call site actually uses. A computed value or a `cn()` call slips past. It's a guard against the easy mistake — which is the one that was made four times — not a proof.

Verification: 27/27 tests, typecheck 12/12 tasks 0 errors, eslint and `prettier --check` clean. Also ran it against #8842's branch, which carries the teams and members column definitions on top of this one — passes there too.

---

Dhruwang (approved):

mattinannt (approved):

mattinannt:
Re-reviewed the commits since my last pass at `78caec9` (`e4dc2b5`, `47cbc1b`, `535ec93`, `33c29ce`), up to head `33c29ce`. Note: my prior review attempt this session ended up submitting an empty approval (`#pullrequestreview-4949793641`) due to a tool-side hiccup on my end while trying to attach an inline comment — apologies for the noise; this comment carries the actual analysis.

Verified the three items from my earlier reviews are still correctly fixed at this commit: keyboard-race disable on Manage/Unarchive, `bodyVariant="flush"` wiring, and the `align`/flex conflict (inert `align: "right"` removed in `47cbc1b`; the real `table-cell`-display bug then fixed properly in `535ec93` by moving the flex to a wrapper `<div>` inside `cell()` instead of `cellClassName`). `33c29ce`'s new `column-classes.test.ts` guard passes and genuinely reaches the affected files.

One non-blocking gap survived that verification, in `535ec93`:

**`apps/web/modules/ui/components/settings-table/types.ts` (near the `align` doc, ~line 22-26):** Moving the flex off `cellClassName` and onto a wrapper `<div>` inside `cell()` correctly fixed the table-cell/`align-middle` bug, but it also silently removed the *only* mechanism that right-aligned this actions column's loading-skeleton bar — and nothing replaced it. `SettingsTableSkeletonRows` (used both by `SettingsTable`'s own `isLoading` prop and by `SettingsTableSkeleton` in `loading.tsx` files) never calls `cell()`; it renders a bare `<Skeleton>` positioned only by `getBodyCellClassName(column)`, i.e. `column.align`/`column.cellClassName`. `feedback-directory-table.tsx`'s `actions` column now has neither. `47cbc1b`'s own commit message explicitly called out keeping this working for the skeleton case ("it is also what right-aligns a skeleton bar if this array is ever fed to `SettingsTableSkeleton`") — that property was dropped a commit later without a replacement (e.g. a `skeletonClassName`/`ml-auto`-style hook).

Not urgent — no column in this PR currently passes `isLoading`/uses a skeleton — but worth a tracked follow-up before the next actions-style column reaches for it, since it's now a live capability of `SettingsTable` itself (see `settings-table/index.tsx`'s `isLoading` branch), not just a hypothetical.

Nothing here blocks merge.

---

Issue #121, opened 2022-10-21
Matrix Question Type


mattinannt:
Support Matrix Question Types

Examples from the web:

![Image](https://user-images.githubusercontent.com/675065/197175607-c801c40a-56be-4743-9c75-8e30d08795f5.png)

![Image](https://user-images.githubusercontent.com/675065/197175673-c1907771-d562-4771-a989-bd8d13768379.png)

mattinannt:
I am closing this issue due to the ongoing migration to Formbricks with a completely new code base.

PR #8843, opened 2026-08-12
feat(helm): add scalable background embedding pool

BhagyaAmarasinghe:
## What & why

Self-hosted embedding jobs and interactive semantic search currently share one TEI deployment, so a backfill saturates the same CPU service and can progress at less than one record per second.

This adds an opt-in worker-only TEI pool as a StatefulSet with one retained RWO cache PVC per replica, a load-balancing service, a headless identity service, and CPU HPA support. Hub API stays on the foreground endpoint while hub-worker receives the background URL and micro-batch settings. It also adds a guarded, disabled-by-default embedding backfill Job that requires a unique run ID.

The batching variables and packaged backfill binary require [formbricks/hub#121](https://github.com/formbricks/hub/pull/121). The pool itself remains disabled by default, and the legacy single-TEI render is covered by CI.

## Linear ticket

No Linear ticket was provided for this incident fix.

## How this was tested

- Rendered legacy single-TEI mode and verified API plus worker both use the foreground endpoint with batch size 1 ✅
- Rendered enabled background mode and verified API/worker endpoint separation, `24 / 8 / 25 / 3` worker settings, StatefulSet, normal/headless services, retained per-replica PVCs, and HPA `1..6` with 900-second scale-down stabilization ✅
- Rendered persistence-disabled mode and verified `emptyDir` replaces volume claim templates ✅
- Verified the backfill Job is absent by default, rejects missing/invalid `runId`, and renders taxonomy/tenant/max-record arguments when enabled ✅
- Rendered the chart with staging, EU production, and KSA GitOps values; staging enables the pool while both production targets remain gated ✅
- Added the same render assertions to the Helm chart validation workflow.

## Breaking changes

| New `hub.embeddings.background.*` values and worker `EMBEDDING_BATCH_*` settings | Hub worker shared the foreground TEI endpoint | Operators can opt into a worker-only scalable TEI pool and batching | Self-hosters enabling the new pool | Use a Hub release containing formbricks/hub#121 before enabling batching or the backfill Job; no action when left disabled |

## QA / Test Plan

**How to test**

- [ ] Deploy with `hub.embeddings.background.enabled=false` → Hub API and hub-worker remain healthy and both use the existing foreground TEI service.
- [ ] Deploy with the background pool enabled → a separate background StatefulSet becomes Ready; Hub API continues serving semantic search through the foreground TEI service while hub-worker drains embedding jobs through the background service.
- [ ] Scale the background HPA to six ready replicas and enqueue representative staging records → sustained embedding throughput reaches at least 5 inputs/s and semantic-search p95 remains within 20% of baseline.
- [ ] Enable a backfill Job with a new `runId`, `taxonomy: true`, and `countOnly: true` → the Job reports the missing count without increasing the embedding queue.
- [ ] Re-run with `maxRecords: 500` and count-only disabled → no more than 500 candidates are enqueued; changing `runId` creates the next deliberate Job.

**Preconditions / test data**

- A Hub image release containing formbricks/hub#121.
- Self-hosted TEI enabled with the same model, arguments, and authentication for both pools.
- Representative staging taxonomy feedback and metrics export enabled.

**Risks & regressions**

- Worker/API endpoint separation, TEI model compatibility, RWO PVC scheduling, HPA scale behavior, River retries, and foreground semantic-search latency.
- If batching regresses, set batch size to 1 while retaining the isolated pool. If the pool fails, disable it and restore foreground worker concurrency to 1.

- No database migration.
- New optional chart values under `hub.embeddings.background.*` and `hub.embeddingBackfill.*`.
- New chart-managed worker variables: `EMBEDDING_BATCH_SIZE`, `EMBEDDING_BATCH_MAX_WAIT_MS`, and `EMBEDDING_BATCH_MAX_IN_FLIGHT`.

### DIFF
--- a/charts/formbricks/templates/_helpers.tpl
+++ b/charts/formbricks/templates/_helpers.tpl
@@ -404,6 +404,17 @@ Hub embeddings runtime resource name.
 {{- printf "%s-hub-embeddings" $base | trimSuffix "-" }}
 {{- end }}
 
+{{/* Worker-only background embeddings runtime resource name. */}}
+{{- define "formbricks.hubEmbeddingsBackgroundName" -}}
+{{- $base := include "formbricks.name" . | trunc 37 | trimSuffix "-" }}
+{{- printf "%s-hub-embeddings-background" $base | trimSuffix "-" }}
+{{- end }}
+
+{{/* Headless service for stable background StatefulSet identities. */}}
+{{- define "formbricks.hubEmbeddingsBackgroundHeadlessName" -}}
+{{- printf "%s-headless" (include "formbricks.hubEmbeddingsBackgroundName" .) | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
 {{/*
 Secret used by Hub and the embeddings runtime for the embeddings API key.
 */}}
@@ -436,6 +447,15 @@ OpenAI-compatible embeddings base URL used by Hub.
 {{- end -}}
 {{- end }}
 
+{{/* Worker-only OpenAI-compatible background embeddings base URL. */}}
+{{- define "formbricks.hubEmbeddingsBackgroundBaseURL" -}}
+{{- if .Values.hub.embeddings.background.baseUrl -}}
+{{- .Values.hub.embeddings.background.baseUrl -}}
+{{- else -}}
+{{- printf "http://%s:%v/v1" (include "formbricks.hubEmbeddingsBackgroundName" .) (.Values.hub.embeddings.background.service.port | default .Values.hub.embeddings.port) -}}
+{{- end -}}
+{{- end }}
+
 {{/*
 Embedding API key value for the generated embeddings secret.
 */}}
@@ -459,22 +479,39 @@ self-hosted runtime is enabled so Hub API and Hub worker cannot drift.
 */}}
 {{- define "formbricks.hubEmbeddingEnv" -}}
 {{- $root := .root -}}
+{{- $worker := .worker | default false -}}
 {{- if $root.Values.hub.embeddings.enabled }}
 - name: EMBEDDING_PROVIDER
   value: "openai"
 - name: EMBEDDING_MODEL
   value: {{ include "formbricks.hubEmbeddingsServedModelName" $root | quote }}
 - name: EMBEDDING_BASE_URL
+  {{- if and $worker $root.Values.hub.embeddings.background.enabled }}
+  value: {{ include "formbricks.hubEmbeddingsBackgroundBaseURL" $root | quote }}
+  {{- else }}
   value: {{ include "formbricks.hubEmbeddingsBaseURL" $root | quote }}
+  {{- end }}
 - name: EMBEDDING_PROVIDER_API_KEY
   valueFrom:
     secretKeyRef:
       name: {{ include "formbricks.hubEmbeddingsSecretName" $root }}
       key: {{ $root.Values.hub.embeddings.auth.secretKey | default "EMBEDDING_PROVIDER_API_KEY" }}
 - name: EMBEDDING_MAX_CONCURRENT
+  {{- if and $worker $root.Values.hub.embeddings.background.enabled }}
+  value: {{ $root.Values.hub.embeddings.background.maxConcurrent | quote }}
+  {{- else }}
   value: {{ $root.Values.hub.embeddings.maxConcurrent | quote }}
+  {{- end }}
 - name: EMBEDDING_NORMALIZE
   value: {{ $root.Values.hub.embeddings.normalize | quote }}
+{{- if $worker }}
+- name: EMBEDDING_BATCH_SIZE
+  value: {{ ternary $root.Values.hub.embeddings.background.batchSize "1" $root.Values.hub.embeddings.background.enabled | quote }}
+- name: EMBEDDING_BATCH_MAX_WAIT_MS
+  value: {{ $root.Values.hub.embeddings.background.batchMaxWaitMs | quote }}
+- name: EMBEDDING_BATCH_MAX_IN_FLIGHT
+  value: {{ ternary $root.Values.hub.embeddings.background.batchMaxInFlight "1" $root.Values.hub.embeddings.background.enabled | quote }}
+{{- end }}
 {{- end }}
 {{- end }}
 
@@ -483,9 +520,15 @@ Returns true when an env var is managed by hub.embeddings and should not be rend
 */}}
 {{- define "formbricks.hubEmbeddingEnvManaged" -}}
 {{- $key := .key -}}
-{{- if has $key (list "EMBEDDING_PROVIDER" "EMBEDDING_MODEL" "EMBEDDING_BASE_URL" "EMBEDDING_PROVIDER_API_KEY" "EMBEDDING_MAX_CONCURRENT" "EMBEDDING_NORMALIZE") -}}
+{{- if has $key (list "EMBEDDING_PROVIDER" "EMBEDDING_MODEL" "EMBEDDING_BASE_URL" "EMBEDDING_PROVIDER_API_KEY" "EMBEDDING_MAX_CONCURRENT" "EMBEDDING_NORMALIZE" "EMBEDDING_BATCH_SIZE" "EMBEDDING_BATCH_MAX_WAIT_MS" "EMBEDDING_BATCH_MAX_IN_FLIGHT") -}}
 true
 {{- end -}}
+{{- end }}
+
+{{/* Name of one deliberate embedding backfill run. */}}
+{{- define "formbricks.hubEmbeddingBackfillName" -}}
+{{- $runID := regexReplaceAll "[^a-z0-9-]+" (.Values.hub.embeddingBackfill.runId | lower) "-" | trunc 20 | trimAll "-" -}}
+{{- printf "%s-embedding-backfill-%s" (include "formbricks.hubname" . | trunc 22 | trimSuffix "-") $runID | trunc 63 | trimSuffix "-" -}}
 {{- end }}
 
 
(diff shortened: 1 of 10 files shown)


charts/formbricks/templates/_helpers.tpl
```diff
@@ -459,22 +479,39 @@ self-hosted runtime is enabled so Hub API and Hub worker cannot drift.
...
+  {{- else }}
   value: {{ $root.Values.hub.embeddings.maxConcurrent | quote }}
+  {{- end }}
 - name: EMBEDDING_NORMALIZE
   value: {{ $root.Values.hub.embeddings.normalize | quote }}
+{{- if $worker }}
+- name: EMBEDDING_BATCH_SIZE
+  value: {{ ternary $root.Values.hub.embeddings.background.batchSize "1" $root.Values.hub.embeddings.background.enabled | quote }}
+- name: EMBEDDING_BATCH_MAX_WAIT_MS
+  value: {{ $root.Values.hub.embeddings.background.batchMaxWaitMs | quote }}
```
xernobyl: [P4] Nit: `EMBEDDING_BATCH_SIZE` and `EMBEDDING_BATCH_MAX_IN_FLIGHT` both fall back to `"1"` via `ternary ... background.enabled` when the background pool is disabled, but this one always renders `background.batchMaxWaitMs` unconditionally. Harmless in practice — Hub disables the whole batcher whenever `BatchSize <= 1`, so the wait value is a no-op when background is off — but worth the same ternary guard for consistency with its siblings.

charts/formbricks/templates/hub-embeddings-background-statefulset.yaml
```diff
@@ -0,0 +1,170 @@
+{{- if and .Values.hub.enabled .Values.hub.embeddings.enabled .Values.hub.embeddings.background.enabled }}
+{{- if and .Values.hub.embeddings.auth.existingSecret .Values.hub.embeddings.huggingFace.token (not .Values.hub.embeddings.huggingFace.existingSecret) }}
+  {{- fail "hub.embeddings.huggingFace.token cannot be stored when hub.embeddings.auth.existingSecret is set; put HF_TOKEN in the existing auth secret or set hub.embeddings.huggingFace.existingSecret" }}
```
xernobyl: [P4] Nit: this HuggingFace-token/existingSecret conflict guard is copy-pasted verbatim from `hub-embeddings-deployment.yaml`. Not new debt from this PR, but it's now duplicated twice — worth factoring into a shared `_helpers.tpl` define at some point so the two copies can't drift.

xernobyl (dismissed):
Reviewed and security-reviewed (CodeRabbit skipped this one — Helm paths are blocked by the default path filters, so this is otherwise unreviewed). Backfill Job is well-guarded (disabled by default, validated runId, non-root/read-only-root securityContext, args passed as argv not shell), background TEI stays ClusterIP-only, and the CI additions match the PR description's test claims. No blockers — two trivial nits inline, non-blocking.

charts/formbricks/templates/hub-embeddings-background-statefulset.yaml
BhagyaAmarasinghe: Fixed in 936912292. The Hugging Face token/existing-secret conflict check is now a shared _helpers.tpl definition used by both foreground and background embedding templates.

charts/formbricks/templates/_helpers.tpl
```diff
@@ -483,11 +527,17 @@ Returns true when an env var is managed by hub.embeddings and should not be rend
 */}}
 {{- define "formbricks.hubEmbeddingEnvManaged" -}}
 {{- $key := .key -}}
-{{- if has $key (list "EMBEDDING_PROVIDER" "EMBEDDING_MODEL" "EMBEDDING_BASE_URL" "EMBEDDING_PROVIDER_API_KEY" "EMBEDDING_MAX_CONCURRENT" "EMBEDDING_NORMALIZE") -}}
+{{- if has $key (list "EMBEDDING_PROVIDER" "EMBEDDING_MODEL" "EMBEDDING_BASE_URL" "EMBEDDING_PROVIDER_API_KEY" "EMBEDDING_MAX_CONCURRENT" "EMBEDDING_NORMALIZE" "EMBEDDING_BATCH_SIZE" "EMBEDDING_BATCH_MAX_WAIT_MS" "EMBEDDING_BATCH_MAX_IN_FLIGHT") -}}
```
mattinannt, line 530: Nit (non-blocking): `formbricks.hubEmbeddingEnvManaged` now treats `EMBEDDING_BATCH_SIZE`/`EMBEDDING_BATCH_MAX_WAIT_MS`/`EMBEDDING_BATCH_MAX_IN_FLIGHT` as globally managed, but `formbricks.hubEmbeddingEnv` only actually emits them when called with `worker=true` (guarded by `{{- if $worker }}` a few lines up). `hub-deployment.yaml`'s call (line ~383) doesn't pass `worker: true`, so if an operator ever sets `hub.env.EMBEDDING_BATCH_SIZE` (etc.), the foreground-env loop at line ~386 will now skip it as "managed" while `hubEmbeddingEnv` never renders it for that call — the var silently disappears instead of passing through or applying. Narrow edge case since these keys aren't documented for `hub.env` directly, but worth a `$worker`-gated managed check (or a comment) so it doesn't surprise someone later.

---

charts/formbricks/templates/_helpers.tpl
```diff
@@ -483,11 +527,17 @@ Returns true when an env var is managed by hub.embeddings and should not be rend
...
 true
 {{- end -}}
 {{- end }}
 
+{{/* Name of one deliberate embedding backfill run. */}}
+{{- define "formbricks.hubEmbeddingBackfillName" -}}
+{{- $runID := regexReplaceAll "[^a-z0-9-]+" (.Values.hub.embeddingBackfill.runId | lower) "-" | trunc 20 | trimAll "-" -}}
```
mattinannt, line 537: `formbricks.hubEmbeddingBackfillName` truncates the sanitized `runId` to its first 20 characters before building the Job name. Two different `runId`s that share the same first-20-char sanitized prefix (e.g. `eu-canary-2026-08-13-wave1` vs. `...-wave2`) collide into the same Job name, which defeats the README's stated safety guarantee that "Each deliberate run requires a new `runId`" — the second `helm upgrade` would either fail on immutable Job fields or reuse the first run's Job instead of creating a distinct one. Worth truncating on a hash suffix or widening the budget (e.g. keep more of the runId, or append a short hash of the full runId) so distinct runIds can't collapse to the same name.

---

mattinannt (commented):
Reviewed for correctness and reuse/simplification issues (this is on top of the existing human security review — no new security concerns found). Two non-blocking findings inline:

1. `_helpers.tpl`: the shared "managed env var" list now includes the batch-size keys globally, but they're only ever emitted for `worker=true` callers — an edge case where `hub.env` overrides for those keys silently vanish on the foreground Hub deployment.
2. `_helpers.tpl`: the embedding-backfill Job name truncates the sanitized `runId` to 20 chars, so two different runIds sharing that prefix collide into the same Job name — undermines the "new runId per run" safety guarantee documented in the README.

Both are narrow edge cases, not blockers.

---

xernobyl (approved):
**no findings** — reviewed at `936912292c`

Re-review after `936912292` ("fix: address embedding chart review feedback"). My previous approval-blocking pass was dismissed by the new push, so this is a fresh read of the whole diff rather than just the delta. Both nits from the prior round are fixed, and fixed better than I asked:

- **`_helpers.tpl:518`** — `EMBEDDING_BATCH_MAX_WAIT_MS` now uses the same `ternary … background.enabled` guard as its two siblings, falling back to `"25"` when the background pool is off. Bhagya also added a CI assertion for it in the legacy render (`--set-string hub.embeddings.background.batchMaxWaitMs=37` → asserts `value: "25"`), so the ternary is now pinned rather than just consistent-looking. That's the part I'd have skipped.
- **HF-token/existingSecret conflict guard** — extracted to `formbricks.validateHubEmbeddingsHuggingFaceSecret` and included from both the foreground Deployment and the background StatefulSet, so the two copies can no longer drift. Exactly the fix.

Fresh pass on the rest of the chart, with the security-relevant surfaces first:

- **Backfill Job** — still well-guarded. Disabled by default and `countOnly: true` by default; `runId` is required, must contain at least one alphanumeric, and is regex-sanitized into the resource name; `maxRecords` rejects negatives; `-prune-stale-models` isn't exposed at all. Args go through `args:` as argv, never a shell, and `tenantId` is quoted, so no injection path from values into the container command. `runAsNonRoot` / `runAsUser: 1000` / `readOnlyRootFilesystem: true`. Because the Job name embeds `runId`, an unchanged `runId` renders an identical Job and re-running `helm upgrade` cannot silently re-enqueue — which is the property the design is after.
- **Background pool exposure** — both new Services are cluster-internal (`ClusterIP` and a headless `clusterIP: None`); no Ingress, no NodePort, no `hostNetwork`. `publishNotReadyAddresses: true` is set only on the headless identity service, not on the load-balancing one, so traffic still lands only on ready replicas. API key comes from the existing secret ref; HF token only via `secretKeyRef`.
- **Endpoint separation** — `hubEmbeddingEnv` now takes `worker`, and only the worker/backfill call sites pass `worker: true`. Hub API keeps rendering the foreground base URL, which is the whole point of the change and is asserted in CI both ways.
- **Compatibility with a pre-#121 Hub** — worth stating since this chart can land before the Hub release: with the background pool disabled, hub-worker still receives `EMBEDDING_BATCH_SIZE=1`, `EMBEDDING_BATCH_MAX_WAIT_MS=25`, and `EMBEDDING_BATCH_MAX_IN_FLIGHT=1`. An older Hub image simply ignores env vars its config struct doesn't declare, so a chart bump ahead of the Hub release is inert rather than breaking. Matches the "no action when left disabled" row in the breaking-changes table.
- **PVC retention** — `whenDeleted: Retain` / `whenScaled: Retain` on the volume claim templates means an HPA scale-down keeps the warm model cache for the next scale-up, which is the right call for a 10Gi TEI cache and avoids a cold re-download storm.

CI is green across all 15 checks, including the three new render assertions this PR adds to the Helm chart validation workflow. Approving.

PR #8844, opened 2026-08-12
feat(unify-feedback): purge a feedback dataset's records from dataset settings (ENG-2129)

xernobyl:
## What & why

Today the only way to clear a feedback dataset is to delete the whole organization. This adds a **"Delete all records"** control to the dataset settings modal, which empties a dataset — its records, their embeddings, their AI enrichment and the topics generated from them — while keeping the dataset itself, its sources and its workspace assignments.

The Hub's existing `DELETE /v1/tenants/{tenant_id}/data` was not usable for this: it is an *offboarding* primitive that also destroys Hub webhooks and per-dataset enrichment settings — integrator configuration this operation has no business touching — and it is synchronous. The narrower Hub endpoint this calls is [formbricks/hub#122](https://github.com/formbricks/hub/pull/122), now **merged and shipped in Hub 0.8.5** — which this PR also pins (see Migrations).

Three things worth a reviewer's attention:

**This is the first route under `/api/internal`.** It follows the shape settled in ENG-1668 / the Internal API RFC: no OpenAPI entry and no stability promise, but every other v3 convention — `withV3ApiWrapper`, session-only auth, strict Zod, RFC 9457 problems, audit logging, rate limit. Nothing was added to the existing `app/api/(internal)` route group, which is the Next.js grouping the RFC explicitly rejects (it doesn't change the URL). Worth coordinating with the RFC's Wave 0 relocation so the two land consistently. No server action — `modules/ee/feedback-directory/actions.ts` is untouched.

**Authorization is org-scoped, and deliberately not workspace-scoped.** The obvious move was to reuse `requireUnifyDirectoryMutationAccess`, but that guard reaches a dataset *through* a workspace, and `workspaceIds` is optional on a dataset — so it could never purge a dataset with no workspace assigned, which is exactly the kind you'd want to empty. `requireFeedbackDatasetMutationAccess` instead derives the organization *from* the dataset and checks the caller's role against it, so a dataset id is never trusted as a claim about who owns it. Same checks otherwise, but ordered the other way round from the sibling guard: **role first, entitlement second**. Checking entitlement first would tell a stranger whether another organization holds an Enterprise licence. For the same reason a dataset that does not exist and one the caller cannot reach return an identical 403 — never a 404 — so the endpoint is not an existence oracle. Owner/manager only, the ENG-1770 reason, which is strongest here since this is the only operation that removes every record at once. The request body is empty; the path id is the only input.

**"Started", not "done".** The Hub runs the purge as a background job (it has to — `feedback_records` has no per-tenant cap and a synchronous delete would outlive the API's write timeout), so the route returns 202 and the toast says deletion has started. Claiming completion would make the next screen look broken. There is deliberately no progress polling: this screen lists datasets, not records, so there is nothing here for a live count to update.

Also fixes two stale things in `modules/hub/service.ts` while adding the sibling wrapper: `deleteHubTenantData` now maps errors through `createHubResultFromError` like its siblings (so callers get the Hub's problem code rather than an opaque message), and its comment no longer claims the SDK lacks a typed method — it has had one since `@formbricks/hub@0.9.0`. Its doc now states plainly that it destroys taxonomy, webhooks and settings, so the two purges can't be confused.

## Linear ticket

Fixes ENG-2129

## How this was tested

- `pnpm vitest run` (full `apps/web` suite) ✅ — 606 files / 7880 tests. The full run is what caught the rate-limit config test, which pins the exact API bucket key set.
- `pnpm typecheck` ✅. (First time round I ran bare `tsc -p tsconfig.json`, which OOM-crashes on this project — I read its zero `error TS` lines as a pass, and a real arity error in the new component reached CI as a failed `@formbricks/web#build`. Fixed in 021aca2a2; `.tsx` is only type-checked by `next build` / `tsconfig.typecheck.json`.)
- `npx eslint` + `prettier --check` on all touched paths ✅.
- `pnpm scan-translations` — confirms the 6 new `en-US` keys are all used, none are missing, and the lockfile is in sync.

**⚠️ The 14 non-English locales are hand-written, not Lingo.dev-generated — and they are permanent until someone deliberately re-runs Lingo.** No API key was available, and `Validate Translation Keys` blocks on missing keys. AGENTS.md says these files are machine-generated, so this is a deliberate exception — **if you read one of these languages, please review your locale rather than assuming it's Lingo output.**

An earlier version of this description claimed "a later `pnpm i18n` will regenerate them all anyway". **That was wrong**, and it is worth being explicit about why. `i18n.lock` keys checksums by *source* string, and the entries added here match the current `en-US` text, so `lingo.dev run` treats these keys as already-translated and skips them — for good. Nothing in CI will ever flag it, because the keys are present in every locale.

Removing the lock entries is not an option on its own: `scan-translations` validates lockfile sync, not just key presence, and fails with `LOCKFILE OUT OF SYNC` if they are missing. To actually regenerate these strings, someone with a Lingo key must delete the `workspace/settings/feedback_directories/purge_*` lines from `apps/web/i18n.lock` **and** run `pnpm i18n` in the same change.

Each locale reuses the vocabulary and register its own `feedback_directories` / `unify` blocks already use rather than translating the English literally: the established word for "dataset" (*Datensatz*, *conjunto de datos*, *ensemble de données*, *adatkészlet*, データセット, *veri kümesi*, 資料集 …), that locale's existing wording for feedback records, its formality (informal *du/tu/je/tú*, formal *Ön* in Hungarian, *вы* in Russian, *siz* in Turkish).

`i18n.lock` carries the six new entries too, or the lockfile check stays red. The checksum is object-hash's MD5 over `string:<character-length>:<value>` — note **characters, not bytes**: the two coincide for ASCII, which is ~96% of the corpus, so a byte-length assumption looks correct until it silently isn't. Measured across the 160 non-ASCII `en-US` keys in the lockfile, 139 match on character length and 1 on byte length. Quotation marks follow each locale's own established convention, checked by counting existing usage in that file (de `„…“`, sv `”…”`, hu/ro `„…”`, fr/ru `« »`, ja/zh-Hant `「」`). `pnpm scan-translations` now passes: 4167 keys used, 4167 in translations, 0 missing, all locales complete.
- Tests added: the org-scoped guard (8 cases, including "checks the caller's role against the dataset's *own* organization", unauthenticated-before-lookup, that a missing dataset is indistinguishable from an unreachable one, that the licence state is only revealed after the role check passes, and that an unexpected error rethrows rather than reading as allowed), the operation (9 cases, including that a Hub failure never relays the Hub's error text and that a failed purge leaves no audit entry claiming one happened), the confirmation helper, and the new Hub service wrapper.
- I fault-injected the error-relay guard to confirm that test actually fails rather than passing vacuously.

**Verified in a browser**, most recently against the published **`ghcr.io/formbricks/hub:0.8.5`** release image (API + worker) rather than a local build, on a real Hub database, as an org **owner**.

Dataset settings modal showing the Delete all records button

Purge confirmation dialog with the confirm button disabled

| | before | after |
|---|---|---|
| feedback records | 421 | **0** |
| embeddings | 47 | **0** |
| taxonomy cluster memberships | 25 | **0** |
| taxonomy runs / nodes / clusters | 1 / 2 / 1 | **0 / 0 / 0** |
| **Hub webhook** | 1 | **1** |
| **`tenant_settings`** | 1 | **1**, `{"sentiment_enabled": false}` intact |

The webhook, the settings row and the taxonomy tree were seeded first — that tenant had none, so without them the "configuration survives, taxonomy goes" assertions would have passed vacuously.

Three things this caught that the unit tests structurally could not:

- **The ICU placeholder interpolates.** The dataset name renders in both the body and `Type "…" to confirm`. A literal `{{directoryName}}` would leave the confirm gate unsatisfiable, since the typed name could never match — this is the bug fixed in 30b9149, now confirmed at runtime rather than by corpus inference.
- **The type-to-confirm gate behaves.** A partial name keeps the button disabled; `"  qa53 directory  "` (wrong case, padded) enables it, so trim + case-insensitive matching work. Reopening the dialog resets the field to empty and the button to disabled.
- **A second purge is not swallowed.** Purging again on the now-empty dataset enqueued and ran a fresh job (all counts zero, no error). This is the regression that matters most on the Hub side: River's default unique states include `completed`, which made every purge after the first a silent no-op returning 202. Exercised through the button, not a harness.

Both requests returned **202** from `/api/internal/feedback-datasets/{datasetId}/purge`, and neither Hub process logged an error.

**Still not exercised:** audit logging (no audit table or `AUDIT_*` configuration in a local environment, so that path remains unit-test-only) and the member-role disabled state. Both are still worth a pass during QA.

### Fixed during review

**Enter in the confirmation input saved the dataset instead of purging it.** Found in manual testing: typing anything and pressing Enter produced a "dataset updated" toast and closed both dialogs. The confirmation `<form>` is rendered, via `DeleteDialog`, inside the settings modal's own `<form onSubmit={handleSubmit(handleSubmitForm)}>`. Radix portals the dialog's DOM out to `document.body`, but React's synthetic `onSubmit` bubbles along the **JSX tree**, not the DOM tree — so the event still reached the outer form. `preventDefault()` only stopped the browser's native submit; the React handler above it kept running. Fixed with `stopPropagation()` in b285af9c9.

Verified at the database layer rather than the UI, because dev-mode Fast Refresh made the visible behaviour unreliable: before the fix a wrong-name submit moved `FeedbackDirectory.updated_at`; after it the timestamp does not move, no purge job is enqueued, and a correct-name submit still purges normally.

**Dropped the `router.refresh()` after a successful purge** (CodeRabbit). Nothing on that screen changes — the dataset table renders name, workspaces and archived status, none of which a purge touches — and the purge is async, so the records are still present when the call resolves. `ArchiveFeedbackDirectory` keeps its refresh, correctly, because archiving does flip the status column.

**Note for the reviewer:** neither fix is unit-testable here. Both are component behaviour, and `.tsx` unit tests are out per AGENTS.md, so the purge journey has no automated coverage — the Enter bug is a concrete example of what that gap lets through. Worth weighing whether this should ship with a Playwright spec.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] As an org **owner or manager**, go to Organization settings → Feedback datasets → open a dataset's settings → the footer shows a red **"Delete all records"** next to Archive.
- [ ] Click it → a confirmation dialog appears naming the dataset, listing what is deleted (records, AI enrichment and generated topics, across every workspace it's shared with) and what is kept (dataset, sources), and warning that charts on this dataset will be empty until new feedback arrives.
- [ ] The confirm button is **disabled** until you type the dataset's exact name → typing a wrong or partial name keeps it disabled; correct name (any casing, surrounding spaces ignored) enables it.
- [ ] Confirm → toast "Deleting records…", the dialog closes. Within a minute the dataset's feedback records are gone from the workspace Unify view.
- [ ] **The point of the feature**: after the purge, the dataset still exists and its feedback sources are still listed. Topics/Subtopics shows the "collect more feedback" empty state rather than a stale tree, and topics can be generated again once the dataset has enough feedback.
- [ ] **Configuration is untouched**: any Hub webhooks and per-dataset enrichment settings (e.g. a sentiment opt-out) are unchanged. This is the line between this action and deleting the organization.
- [ ] Close and reopen the dialog → the confirmation box is empty again (it must not stay pre-filled).
- [ ] As an org **member** (not owner/manager), the button is disabled with a tooltip explaining why.
- [ ] Purge a dataset that has **no workspaces assigned** → still works. (This is the case the workspace-scoped guard could not have handled.)
- [ ] A dataset id from another organization, or one that does not exist → both return the same "not authorized" error, with no hint that one of them exists.
- [ ] Purge an **already-empty** dataset → succeeds, no error.
- [ ] Double-click confirm / purge twice quickly → no error, no duplicate work.
- [ ] With `HUB_API_URL`/`HUB_API_KEY` unset → the action fails with a clear error rather than silently succeeding.
- [ ] Rate limit: more than 5 purges in an hour from one user → 429.

**Preconditions / test data**

- Enterprise plan or license (`feedbackDirectories` entitlement) — without it the endpoint returns 403.
- A configured Hub running **hub#122 or later** (the endpoint is `DELETE /v1/tenants/{tenant_id}/feedback-records`); against an older Hub the purge returns a 502.
- A feedback dataset with records, ideally with a generated taxonomy so the "topics survive" check is meaningful.

**Risks & regressions**

- `modules/hub/service.ts` — `deleteHubTenantData`'s error mapping changed. Its only caller is the organization-delete cascade, which treats failures as best-effort; re-check that deleting an org with feedback datasets still works.
- `ZAuditAction` gained `"purged"` — additive, but audit-log consumers that switch exhaustively on the action should be checked.
- `rateLimitConfigs.api` gained a key — the config test pins the key set (updated here).
- The dataset settings modal footer now holds two destructive buttons; check they don't crowd on narrow viewports.
- The Hub image pin moved 0.8.4 → 0.8.5 for both Helm and dev compose; anyone running the dev stack picks up the new image on their next pull.

- No DB migration, no new env vars.
- **Hub image pinned to 0.8.5** here — `charts/formbricks/values.yaml` (digest + tag), the `charts/formbricks/README.md` rows, and all three hub services in `docker-compose.dev.yml`. 0.8.5 is the first release containing the endpoint. The digest is the multi-arch index digest, and the fetch method was checked by reproducing the committed 0.8.4 digest before trusting it. `Chart.yaml` stays `0.0.0-dev`; chart releases are cut via tags, so a deployment only picks this up on the next chart release.
- **Ordering**: the running Hub must actually be 0.8.5 or later, or the purge returns 502. No SDK bump is required — the web side calls the raw Hub path, as `deleteHubTenantData` already does. `@formbricks/hub@0.12.0` predates hub#122 and has no typed method for it yet; switch once Stainless publishes one.


apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
```diff
@@ -0,0 +1,131 @@
...
+            }}>
+            <label htmlFor="purgeDatasetConfirmation">
+              {t("workspace.settings.feedback_directories.purge_confirmation_name", {
+                directoryName: displayName,
+              })}
+            </label>
+            <Input
+              value={confirmationName}
+              onChange={(e) => setConfirmationName(e.target.value)}
+              placeholder={displayName}
```
mattinannt: **Confirmation can never succeed for dataset names longer than 30 characters.**

The label (line 113-115) and the `Input`'s `placeholder` (line 120) both show `displayName`, which is `truncate(directoryName, 30)` (adds a literal `...` if the name is longer than 30 chars). But `hasValidConfirmation` (line 42) compares the typed text against the full, untruncated `directoryName`. A user who types exactly what the UI tells them to type — the truncated string ending in `...` — will never satisfy the check, and the purge button stays disabled with no way to proceed.

This diverges from the sibling pattern this component's own comment says it mirrors: in `delete-workspace-render.tsx`, only the descriptive warning text is truncated (`workspaceName: truncate(currentWorkspace.name, 30)`), while the confirmation label and `Input` placeholder use the full, untruncated `currentWorkspace.name` (lines 136 and 142 there). This component should do the same — keep `displayName` for the warning copy, but use the full `directoryName` for the confirmation label and placeholder.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
```diff
@@ -0,0 +1,131 @@
...
+        title={t("workspace.settings.feedback_directories.purge_all_data")}
+        buttonLabel={t("workspace.settings.feedback_directories.purge_all_data")}
+        onDelete={handlePurge}
+        isDeleting={isPending}
+        disabled={!hasValidConfirmation}
+        text={t("workspace.settings.feedback_directories.purge_all_data_warning", {
+          directoryName: displayName,
+        })}>
+        <div className="py-5">
+          <form
```
mattinannt, line 129: The destructive button in `DeleteDialog`'s footer is correctly disabled while `isPending` (via `isDeleting={isPending}`), but this `<form onSubmit>` around the confirmation input calls `handlePurge()` unconditionally on Enter without checking `isPending`. Holding/double-pressing Enter before the first `mutateAsync` resolves can fire a second concurrent purge request through this entry point, bypassing the guard the button has. Likely low-impact given the rate limit and the Hub-side "join the in-flight purge" behavior, but worth an `isPending` check here too for consistency with the button.

---

apps/web/app/api/internal/feedback-datasets/lib/access.ts
```diff
@@ -0,0 +1,88 @@
...
+      "Feedback datasets are not enabled for this organization. It requires an Enterprise plan or license.",
+      instance
+    );
+  }
+
+  return { organizationId };
+}
+
+/** Extract the session user id. Present because this route is session-auth. */
+export function getSessionUserId(authentication: TV3Authentication): string | null {
```
mattinannt, line 83: Minor: this `getSessionUserId` is a verbatim third copy of the same function (identical `authentication && "user" in authentication && authentication.user?.id` logic already exists in `apps/web/app/api/v3/surveys/lib/operations.ts:112` and `apps/web/app/api/v3/unify-feedback/taxonomy/lib/access.ts:105`). Not a new pattern introduced by this PR, but it'd be worth extracting to a shared helper so a future fix to the session-shape extraction doesn't need to be applied in three places to stay in sync.

---

mattinannt (commented):
Reviewed the feedback-dataset purge feature (internal purge route + access guard, Hub service additions, settings-modal UI, tests, i18n).

One real bug found: the purge confirmation input compares against the full dataset name while the label/placeholder show a truncated one, so datasets with names over 30 characters can never be confirmed. Also flagged a missing in-flight guard on Enter-to-submit, and a pre-existing `getSessionUserId` duplication pattern this PR adds a third copy of. Details inline.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
xernobyl: Confirmed and fixed in `80223b8b9` — including the sibling pattern you cite, which I checked rather than took on trust: `delete-workspace-render.tsx` truncates only the descriptive copy (`:95`, `:124`) and uses the full name for both the confirmation label (`:136`) and the placeholder (`:142`).

The purge dialog was using `displayName` in all three places while `hasValidConfirmation` compared against the full `directoryName`, so any dataset with a name over 30 characters was unpurgeable: `truncate` appends a literal `...`, the matcher rejects it, and the button stays disabled with nothing the user can type to proceed. Label and placeholder now use the full name; the warning copy stays truncated.

Pinned in `purge-confirmation.test.ts` rather than in the component, since `.tsx` is Playwright's job by policy and there is no Playwright coverage for purge yet — the assertion is that the truncated form of a long name is rejected while the full name is accepted, which is exactly the property that forces the dialog to show it in full. Mutation-checked: teaching the matcher to accept a truncated prefix fails it.

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
xernobyl: Fixed in `80223b8b9` — `onSubmit` now returns early while `isPending`.

Your scoping was right, and narrower than it first looked to me. I checked whether Enter could also bypass the *confirmation*, since the footer button carries `disabled={!hasValidConfirmation}` and the form does not: it cannot, because `handlePurge` opens with `if (!hasValidConfirmation) return`. So the only thing Enter escaped was the `isDeleting` guard, exactly as you described — a held or double-pressed Enter could start a second purge while the first `mutateAsync` was still in flight.

Agreed on the impact too: the rate limit and Hub's join-the-in-flight-purge behaviour make it mild. Worth closing for consistency rather than for damage.

apps/web/app/api/internal/feedback-datasets/lib/access.ts
xernobyl: Agreed on the observation, declining the extraction — and flagging that as a judgement call rather than a disagreement about the facts.

You are right that it is a verbatim third copy: `app/api/v3/surveys/lib/operations.ts:112` and `app/api/v3/unify-feedback/taxonomy/lib/access.ts:105` carry the identical `authentication && "user" in authentication && authentication.user?.id` shape.

The reason I would leave it: it is a single boolean-and expression over a type union, with no branching and nothing to get wrong independently in each copy. A shared helper buys one edit site for a line that has not changed since it was written, and costs an import from three modules that are otherwise unrelated to each other. This repo's steer on trivial one-liner helpers is to leave module-local copies alone and reserve consolidation for logic where drift actually causes bugs.

If the session shape does change — which is the case you are guarding against — it changes in `TV3Authentication`, and the type checker points at all three call sites. That is the mechanism that keeps them in sync, and it works whether or not the expression is shared.

Happy to be overruled if you would rather have it extracted; it is a small change and I have no strong attachment to the current shape. It just is not something I would widen a purge PR to do.

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
xernobyl: Good catch, and a real one — fixed in 80223b8b9.

Worse than it looks, too: it isn't just that the check fails, it's that the UI *instructs* the user to type the string that can never match. The button stays disabled with no way forward and nothing on screen explaining why.

Fixed exactly as you describe, and for the reason you give — `delete-workspace-render.tsx` truncates only the descriptive copy and uses the full name at the label and placeholder. The label, the placeholder and the comparison now all read the same `directoryName`, so they can't drift apart again, and the comment says why truncation stays out of those two spots.

Pinned in `purge-confirmation.test.ts` rather than the component, since `.tsx` unit tests are out per AGENTS.md: the test asserts the truncated form of a long name is rejected — which is precisely the reason the dialog has to show it in full.

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
xernobyl: Agreed — fixed in 80223b8b9 with an `isPending` check in `onSubmit`.

Your read of the impact is right: it can't bypass the confirmation itself, since `handlePurge` opens with `if (!hasValidConfirmation) return`, so the only gap was the in-flight guard. Low impact given the rate limit and that the Hub collapses a repeat request into the running purge — but the footer button and the Enter path should not disagree about when the dialog is busy, and that inconsistency is the kind of thing that outlives the mitigations that make it harmless today.

apps/web/app/api/internal/feedback-datasets/lib/access.ts
xernobyl: Fair, and you're right that it's now three copies.

I'd rather not fold it into this PR though: extracting it means touching `app/api/v3/surveys/lib/operations.ts` and `app/api/v3/unify-feedback/taxonomy/lib/access.ts`, neither of which this change otherwise goes near, and the repo's own rule is that adjacent cleanups get their own ticket rather than widening the PR. A shared-helper change across three route modules also wants its own review, not a paragraph at the bottom of a purge PR.

Happy to file it — and it's a better ticket than it looks, since the thing worth fixing is arguably the underlying shape: `TV3Authentication` being loose enough that every consumer has to re-narrow it by hand is why the helper keeps getting copied. Want me to raise it that way?

xernobyl:
@mattinannt all three addressed — replies on each thread.

**Fixed in 80223b8b9**

- **The confirmation could never be satisfied for names over 30 characters.** Real bug, and worse than a failed check: the dialog *instructed* the user to type the truncated string, so the button stayed disabled with nothing on screen explaining why. Truncation now applies only to the warning copy, matching `delete-workspace-render.tsx`; the label, the placeholder and the comparison all read the same value so they can't drift again. Pinned in `purge-confirmation.test.ts` rather than the component, since `.tsx` unit tests are out by policy — the test asserts the truncated form is rejected, which is the reason the dialog has to show the name in full.
- **Enter bypassed the in-flight guard.** `onSubmit` now checks `isPending`. It could never bypass the confirmation itself (`handlePurge` opens with `!hasValidConfirmation`), so the gap was only the double-fire.

**Not changed:** the third `getSessionUserId` copy. Extracting it means touching two v3 route modules this PR doesn't otherwise go near, and adjacent cleanups get their own ticket here rather than widening a PR. Happy to file it — arguably as the underlying issue, that `TV3Authentication` is loose enough that every consumer re-narrows it by hand, which is why the helper keeps getting copied.

**Note on CI:** `Run Typecheck` is currently red on this branch, and it isn't from this diff — every error is `Cannot find module '@formbricks/jobs'` and cascades in `lib/jobs/*` and `modules/ee/workflows/lib/runner/*`, none of which this PR touches. The branch is 36 commits behind `main`, and `packages/jobs/package.json` gained a `typecheck` script in the monorepo config audit (#8898) that this branch predates. Updating the branch from `main` should clear it; other open PRs pass the same check.

Still a draft on purpose — it needs hub#122 merged and a Hub release deployed before it can land.

apps/web/app/api/internal/feedback-datasets/lib/access.ts
xernobyl: Filed as ENG-2487 — https://linear.app/formbricks/issue/ENG-2487

Raised it as the underlying issue rather than just the extraction: the reason the helper keeps getting copied is that `TV3Authentication` is loose enough that every session-auth consumer has to re-narrow it by hand, so a shared `getSessionUserId` would remove three copies without stopping a fourth. Linked to ENG-2371, which is the same theme for workspace-scoped resource auth.

kiln-agent:
## Review — feedback-dataset purge (ENG-2129)

Reviewed at `5d10c75`. The security reasoning in the guard is the strongest part of this diff and it holds up under checking. One finding matters enough to act on before merge, plus one plausible UX inversion and two nits.

---

### 1. The `i18n.lock` entries permanently freeze the 14 hand-written locales

`apps/web/i18n.lock`

The description says the hand-written translations are a temporary exception because *"A later `pnpm i18n` will regenerate them all anyway."* It won't — the lockfile entries added here are what prevents that.

`i18n.json` declares one source (`en-US`) and 14 targets, and `i18n.lock` has a **single** source-keyed checksum group (line 3), so the checksums are of the *source* string, not per-target. `lingo.dev run` re-translates a key when its source checksum is missing or differs, and skips it when it matches. The five entries added here match the current en-US text exactly:

```
workspace/settings/feedback_directories/purge_all_data          ✓ matches en-US
workspace/settings/feedback_directories/purge_all_data_warning  ✓ matches en-US
workspace/settings/feedback_directories/purge_confirmation_name ✓ matches en-US
workspace/settings/feedback_directories/purge_not_allowed       ✓ matches en-US
workspace/settings/feedback_directories/purge_started           ✓ matches en-US
```

I calibrated the scheme against the existing lockfile rather than trusting the description: it is `md5("string:<len>:<value>")` (object-hash's string encoding — the description's `string::` is a slightly loose rendering of it, though the computed values are right). Across the whole file, **4076 of 4095** entries match the current en-US text and **19 do not** — and those 19 are precisely the keys whose English changed since the last Lingo run, i.e. the "needs re-translation" state. The five new keys land in the *matching* bucket.

So the outcome is: the hand-written de/es/fr/hu/ja/nl/pt-BR/pt-PT/ro/ru/sv/tr/zh-Hans/zh-Hant strings are permanent. `Validate Translation Keys` will never flag it, because it checks key presence and the keys are filled. This is the exact failure mode AGENTS.md's "never hand-write the non-English locales" rule exists to prevent, and it's invisible to every check in CI.

**Suggested fix:** drop the five `workspace/settings/feedback_directories/purge_*` lines from `i18n.lock` while keeping the `en-US.json` additions. The next `pnpm i18n` then sees them as new and overwrites all 14 targets with Lingo output. The keys are absent from the targets until that run, which is the state AGENTS.md expects and the reason it says not to fill them by hand.

If instead the intent is to *keep* the hand-written translations, that's a reasonable call for a native-reviewed set — but it should be a deliberate decision recorded on the PR, not a side effect of the lockfile, because nothing downstream will ever revisit them.

---

### 2. A client-side timeout reports failure for a purge that has already started (plausible)

`apps/web/modules/ee/feedback-directory/lib/api-client.ts` · `purge-feedback-directory-data.tsx`

`AbortSignal.timeout(15_000)` bounds the fetch, but aborting a fetch doesn't cancel the server-side request — the route still runs and still enqueues the Hub job. The `DOMException` branch then shows `common.something_went_wrong_please_try_again`, so the user is told the purge failed while it is in fact running.

That's the one direction it's worst to be wrong in for an irreversible dataset-wide delete, and it inverts the care taken everywhere else in this diff about "started, not done". 15s is reachable: the handler does two DB round-trips plus a licence check before the Hub call, and a slow Hub connect can exceed the client budget on its own.

Retrying is harmless (the purge is idempotent, and hub#122's unique-state fix is what makes that true), so this is about the message, not the mechanism — consider wording the abort case as indeterminate ("we couldn't confirm this finished — check back in a few minutes") rather than as a failure.

---

### 3. Nit — line-number references in a code comment will rot

`purge-feedback-directory-data.tsx`

> `delete-workspace-render.tsx does the same, full name at :136 and :142`

Both are correct right now (136 is `workspaceName: currentWorkspace.name`, 142 is `placeholder={currentWorkspace.name}`) — I checked, because the comment is load-bearing for the truncation decision. They'll go stale on the next edit to that file. Naming the two props instead of the lines keeps the point and survives.

### 4. Nit — `deleteWhat` is dead here

`DeleteDialog` uses `deleteWhat` only for its fallback title, and `title` is passed, so `deleteWhat={displayName}` never renders. Required prop, so it has to be *something*, but as written it reads as if it drives the dialog.

---

### Checked and clear

Recording these so the negatives are visible, since several are the kind of thing worth not re-deriving:

- **Tenant scoping.** Org is derived from the dataset and the role checked against *that* org; entitlement strictly after role; missing and unreachable both return an identical 403. `getOrganizationIdFromDirectoryId` does throw `ResourceNotFoundError`, so the `catch` narrows on the right type and anything else rethrows rather than reading as allowed. The ordering rationale in the doc comment is correct.
- **CSRF on a session-only destructive POST.** A cross-site POST with no body is a simple request (no preflight), so the cookie policy is the whole defence — `modules/auth/lib/auth.ts:341` sets `sameSite: "lax"`, which withholds it. Fine.
- **`/api/internal` exposure.** Not in the public-domain allowlist, so `isRouteAllowedForDomain` blocks it there; `isAdminDomainRoute` allows unknown routes only on the admin domain. Correct for a dashboard-only surface.
- **`"purged"` on `ZAuditAction`.** No Prisma enum backing it and no exhaustive consumer (`Record<TAuditAction, …>` / switch) anywhere, so additive is safe.
- **Audit on the denied path.** `buildAuditLogBaseObject` defaults `status: "failure"` and `organizationId: UNKNOWN_DATA`, and `queueV3AuditLog` still emits, so setting `targetId` before the guard does what its comment claims. Worth knowing: a cross-tenant attempt lands with no `organizationId`, so the targeted org can't see it — but every sibling v3 operation sets both fields only *after* authorization, so this route is already more attributable than its neighbours. Repo-wide gap, not this PR's.
- **SDK prose.** `client.tenants.deleteData()` does exist in the pinned `@formbricks/hub@0.12.0`, and there is no typed feedback-records purge in it. Both revised comments are accurate.
- **`problemNotFound` guidance.** Its doc really does warn against existence-sensitive resources and point at `problemForbidden`, as the guard's comment claims.
- **Interpolation.** Single-brace `{directoryName}` matches the repo convention (255 single-brace vs 7 double-brace in `en-US.json`), and both interpolated keys carry the placeholder in all 15 locale files — so no locale renders an unsatisfiable confirm gate. This was the 30b9149 bug class; it's closed.
- **`z.cuid2()`.** Consistent with every other v3 id schema despite ids being Prisma `cuid()`; zod's cuid2 pattern accepts them. Not new here.
- **Component patterns.** The nested `DeleteDialog`-inside-settings-modal, `TooltipRenderer` around a disabled button, and the wrapper `div` are all copied from `ArchiveFeedbackDirectory`, so the footer stays internally consistent.
- **Tests.** The confirmation regression is pinned where it can actually fail (truncated form rejected, full form accepted), the audit test asserts on a real object rather than a mock's call args, and the abort-signal and path-escaping assertions are both substantive. No vacuous passes found.

All 18 checks are green at this head; `mergeable_state: blocked` is just the missing approval.

Finding 1 is the one I'd want resolved before this merges — everything else is optional.

---

kiln-agent (commented):
Reviewed at `5d10c75` (internal purge route + org-scoped guard, Hub service wrapper, settings-modal UI, tests, i18n, chart bump). The three items from the earlier review are fixed and I did not re-raise them. Nothing here blocks on the route or the guard — the security-relevant properties hold under checking (details at the end). Four things worth acting on before this lands, one of which contradicts a mitigation the description relies on.

---

### 1. The hand-written locales will **not** be regenerated by a later `pnpm i18n` — the lockfile entries freeze them

`apps/web/i18n.lock` (+5 entries) is doing the opposite of what the description assumes. The description says *"A later `pnpm i18n` will regenerate them all anyway"*, which is the stated safety net for knowingly writing 14 non-English locales by hand against the AGENTS.md rule. That net isn't there.

I reverse-engineered the checksum format to confirm it, rather than taking the description's word for it: it is `md5("string:" + byteLength + ":" + value)`. Verified against an existing entry —

```
workspace/settings/feedback_directories/no_feedback_sources
  md5("string:47:No feedback sources linked to this dataset yet.") = 409051191687e1ba65eeda82aa68fa06  ✓ matches the lock
```

— and then against all five new entries: `purge_all_data`, `purge_all_data_warning`, `purge_confirmation_name`, `purge_not_allowed`, `purge_started` **all match** their `en-US` values exactly.

That is precisely the state that makes lingo.dev skip a key: the lock records the source string as already translated, and every target locale already has a value, so there is nothing for the next run to do. The five keys are now permanently pinned to hand-written output. The top-level group hash in the lock (`55dc377b…`) is a stable bucket identity — unchanged since it was introduced in `de993bb`, and unchanged by this diff — so it won't invalidate the group either.

CI cannot see this: `Validate Translation Keys` only checks key presence and lock sync, both of which pass *because* the entries were added.

The fix has to happen before merge, and the PR can't ship until hub#122 deploys anyway, so there is room: someone with a Lingo key drops those five lock lines and runs `pnpm i18n`, so real Lingo output replaces the 14 hand-written strings. Dropping the lock lines alone isn't viable — `scan-translations` fails them as "Keys missing from lockfile". If the decision is instead to keep the hand-written strings, that is a legitimate call, but then it needs saying explicitly in the description (they're permanent, not provisional) and the request for native-speaker review becomes a merge requirement rather than a nice-to-have.

### 2. Feedback arriving during the purge window may be silently lost — verify hub#122's write serialization

The SDK's own doc for the sibling offboarding purge (`node_modules/@formbricks/hub/resources/tenants/tenants.d.ts:19-28`) states that it *serializes against tenant-owned writes*, and that writes during the purge are **rejected with HTTP 409**. If hub#122's `DELETE /v1/tenants/{id}/feedback-records` behaves the same way, this matters here in a way it doesn't for offboarding — an offboarded tenant has no live ingestion, an emptied dataset does.

The web-side pipeline has no retry and no dead-letter path: `apps/web/lib/feedback-source/pipeline-handler.ts:63-77` logs failures at `warn` (per-record detail at `debug`, off in production) and drops them. `lastSyncAt` isn't advanced, and `responseFinished` doesn't come round again. So any survey response completing inside the purge window is gone with only a log line — and the window is explicitly minutes: the toast says "This can take a few minutes for a large dataset."

Two outcomes, both worth knowing before ship:
- Hub rejects writes during the purge → responses finishing mid-purge are dropped silently.
- Hub accepts them → newly arrived records may be swept by the purge job itself, which reads the same way to the user.

The browser verification (421 records, no concurrent traffic) structurally couldn't surface either. Worth a QA line — purge a dataset that has a live survey source while submitting a response — and, if writes are rejected, either a note in the dialog copy or an issue against the ingestion path.

### 3. The new rate-limit quota is pinned by key only, not by value

`apps/web/modules/core/rate-limit/rate-limit-configs.test.ts:78-88` adds `internalDatasetPurge` to the key-set assertion. But this file already has a convention for quotas that are the *only* thing bounding a sensitive operation — `auth.emailToken` and `actions.unsplash` both assert exact `{interval, allowedPerInterval, namespace}`, with comments saying a loosened quota is a security regression that should fail here rather than in production.

Your own justification in `rate-limit-configs.ts:22-27` ("irreversible and dataset-wide; a burst is a bug or an attack") is that same argument. As it stands, someone raising `allowedPerInterval` to 500 or shortening the interval keeps the suite green. One assertion closes it.

### 4. The Hub image bump isn't in `Migrations / env / cutover`

`charts/formbricks/values.yaml`, `charts/formbricks/README.md` and `docker-compose.dev.yml` move Hub `0.8.4` → `0.8.5` (digest re-pinned). The Migrations section says only "No DB migration, no new env vars" plus the hub#122 ordering constraint — but this bump *is* the self-hosted half of that constraint, and with a digest pin the chart is what enforces it. AGENTS.md asks for env/migration/cutover changes recorded there, breaking or not, and it's the line a self-hoster reads. Please name it, and confirm `0.8.5` actually contains hub#122 — if it doesn't, the chart advertises an ordering guarantee it can't keep. (Also worth a word on whether the bump carries Hub-side DB migrations, since `hub-migrate` runs on the new image.)

### Minor

- **No Playwright coverage for a destructive, role-gated flow.** With `.tsx` unit tests out by policy, the confirmation gate, the dialog reset on reopen and the member-disabled state have no automated coverage at all — the repo's own heuristic ("user journey risk: E2E") points straight at this one. The description's `Open gaps` lists audit logging and the member-role state; adding "no E2E spec exists for purge" makes the gap visible rather than implied.
- **Purge is undoable for survey-backed datasets, which the copy doesn't reflect.** `importHistoricalResponses` (`apps/web/lib/feedback-source/import.ts:67`) re-imports *all* of a survey's responses and runs on source creation, so deleting and re-adding a source after a purge repopulates the dataset. The Formbricks-side responses are never touched either. That's fine as behaviour, but "This permanently deletes every feedback record" reads as an erasure primitive, and it isn't one — worth a QA line so nobody reaches for this to satisfy a data-deletion request.
- `operations.ts:63` relays the Hub's `status` string verbatim into the 202 body, and the only client (`api-client.ts`) ignores it. Harmless, but it's an unvalidated upstream string in a response shape you own.

### Checked and holds

Recording these so they don't get re-litigated: the guard cannot 500 on a non-member (`getMembershipRole` → `AuthorizationError` → the same `denied` 403, so the no-existence-oracle property is real, not just intended); tenant id ↔ dataset id is consistent with the org-delete cascade (`lib/organization/service.ts:317` also passes `directory.id`); `z.cuid2()` accepts Prisma `cuid()` v1 ids (Zod 4's pattern is `/^[0-9a-z]+$/`, and it matches repo-wide `ZId`); session cookies are `sameSite: "lax"` (`modules/auth/lib/auth.ts:341`) so the bodyless session POST has no CSRF gap; `/api/internal` resolves correctly under domain-aware routing (allowed on the admin domain, blocked on the public one) with no allowlist edit needed; 202 satisfies `response.ok`, so the audit entry is marked `success` rather than `failure`; the single-brace ICU placeholders match repo convention; `deleteData()` does exist in the installed SDK `0.12.0` and there is still no typed method for `/feedback-records`, so both new comments are accurate; and no `route.test.ts` exists for any of the 29 v3 routes, so the absence of one here isn't a convention break.

---

apps/web/modules/ee/feedback-directory/components/feedback-directory-settings/purge-feedback-directory-data.tsx
```diff
@@ -0,0 +1,137 @@
...
+    if (!hasValidConfirmation) return;
+
+    try {
+      await purgeDataset({ datasetId: directoryId });
+      // "Started", not "done": the purge runs in the background, so the records are still there for
+      // a moment after this resolves. Promising completion here would make the next screen look broken.
+      toast.success(t("workspace.settings.feedback_directories.purge_started"));
+      handleDialogOpenChange(false);
+      onPurge?.();
+      router.refresh();
```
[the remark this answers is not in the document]
xernobyl: Agreed, removed in b285af9c9 along with the now-unused `useRouter`.

Checked what the refresh could actually update before dropping it: the dataset table renders name, workspaces and archived status (`feedback-directory-table.tsx`), and a purge changes none of them. There's no record count on this screen. On top of that the endpoint returns 202, so the records are still present when `mutateAsync` resolves — the refresh couldn't have shown a change even if the table had one.

Worth noting the sibling `ArchiveFeedbackDirectory` keeps its `router.refresh()`, and correctly so: archiving flips `isArchived`, which *is* the status column. I left a comment saying that, so the asymmetry doesn't read as an oversight later.

apps/web/modules/ee/feedback-directory/lib/api-client.ts
```diff
@@ -0,0 +1,33 @@
...
+ * every call throws a parsed `V3ApiError` on a non-2xx response.
+ */
+
+const BASE_PATH = "/api/internal/feedback-datasets";
+const MUTATION_TIMEOUT_MS = 15_000;
+
+export type TPurgeFeedbackDatasetResponse = {
+  datasetId: string;
+  status: string;
+};
```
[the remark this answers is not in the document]
xernobyl: Not taking this one, because it would make this file the only exception to a consistent pattern.

Every `api-client.ts` in the codebase unwraps the envelope with the same cast, none validates:

- `modules/ee/unify-feedback/topics-subtopics/lib/api-client.ts` — 4 sites, including the one this client was modelled on (`((await response.json()) as { data: T }).data`)
- `modules/ee/workflows/lib/api-client.ts` — 4 sites
- this file — 1 site

The body isn't untrusted input either: it comes from `successResponse` in `app/api/v3/lib/response.ts`, a first-party route in the same repo and same deploy, and the response is two fields the route builds literally (`{ datasetId, status }`). A schema here would assert against a shape that cannot drift independently.

If we do want runtime validation of v3 envelopes on the client, I'd rather it be one change across all three clients — consistent, reviewed together — than a single divergent file. Happy to file that if you think it's worth it.

xernobyl:
image
image

xernobyl:
Thanks — worked through all four. Findings 2, 3 and 4 are fixed in 2f4bbba34. Finding 1 is confirmed as a diagnosis but **its suggested fix doesn't work in this repo**, so it's resolved a different way. Detail below, including two places the analysis needs correcting.

### 1. `i18n.lock` freezing the hand-written locales — confirmed, fixed differently

The diagnosis is right, and I verified it independently rather than taking it on trust: `i18n.lock` has a single source-keyed checksum group, all five purge keys' checksums match the current `en-US` text, so `lingo.dev run` skips them and the hand-written translations are permanent. My PR description's claim that "a later `pnpm i18n` will regenerate them all anyway" was false.

**But the suggested fix turns CI red.** `scan-translations` does not only check key presence — it validates lockfile sync, and dropping those five lines fails it outright:

```
❌ LOCKFILE OUT OF SYNC:
   • Keys missing from lockfile (5):
      - workspace/settings/feedback_directories/purge_all_data
      ...
```

I tried it before concluding that. So with no Lingo API key available, option (a) isn't reachable — which leaves your option (b): keep them, and make it a deliberate decision recorded on the PR rather than a silent side effect. I've rewritten that section of the description to say the translations are **permanent until someone re-runs Lingo**, and to give the exact procedure (delete the `purge_*` lock lines *and* run `pnpm i18n` in the same change — either alone leaves the tree inconsistent).

Also worth flagging: **the checksum is over the character length, not the byte length.** My description said `string:<byte-length>:<value>` and that's wrong — it only appeared to hold because ~96% of the corpus is ASCII, where the two coincide. Measured across the 160 non-ASCII `en-US` keys in the lockfile: 139 match on character length, 1 on byte length. Your `string:<len>:<value>` was the more careful phrasing. This mattered in practice — my first hand-computed checksum for the new key was wrong because of it, and `scan-translations` caught it as "out of sync/stale".

### 2. Timeout reported as failure — fixed

Agreed, and this was the one I'd have regretted shipping. A `DOMException` from `AbortSignal.timeout` now shows a dedicated "we couldn't confirm this started, check back in a few minutes" message and closes the dialog, instead of `something_went_wrong`. Re-running stays safe: the purge is idempotent, and the Hub collapses a repeat request into the in-flight one.

That did mean a sixteenth hand-written key (`purge_outcome_unknown`), which deepens finding 1 by one string — a deliberate trade, since mis-reporting an irreversible delete as failed is worse than one more line of translation debt.

### 3 & 4. Both nits — fixed

Line numbers replaced with prop names, and `deleteWhat` now carries a comment saying it only feeds the fallback title that `title` overrides.

### On the "checked and clear" list

Useful, and it matches what I found independently on the same points — the guard ordering, the `SameSite=Lax` reasoning for a bodyless cross-site POST, and the audit-on-denied-path behaviour. The note that a cross-tenant attempt lands without an `organizationId`, so the targeted org can't see it, is a genuinely good catch that I'd missed; agreed it's a repo-wide gap rather than this PR's.

One correction there too: the review says the confirmation regression is pinned "where it can actually fail". True, but only for the *helper*. The two bugs found since — Enter submitting the outer settings form, and this timeout inversion — are both component behaviour, and `.tsx` unit tests are barred by AGENTS.md, so neither is covered by anything. That gap is now called out explicitly in the PR description.

Dhruwang (approved):

PR #8846, opened 2026-08-13
fix: support internal MCP OAuth JWKS URL

BhagyaAmarasinghe:
### How this was found

During AuthZed shadow-mode QA in `authzed-sandbox`, MCP requests authenticated with API keys reached the authorization layer, but MCP requests authenticated as users failed earlier during OAuth bearer-token verification. The Formbricks pod could not retrieve Better Auth's JWKS from the configured public issuer URL.

We temporarily set `BETTER_AUTH_URL` to the in-cluster Formbricks service to prove that JWKS self-resolution was the blocker. That made MCP user authentication work, but it was not a valid fix: `BETTER_AUTH_URL` also defines the externally visible OAuth issuer and affects discovery metadata, redirects, cookies, and token validation. Pointing it at an internal Kubernetes address would therefore break the public OAuth contract.

### Why this is needed

Although the issue was discovered while testing AuthZed, it is not an AuthZed requirement. It is an MCP OAuth deployment and self-hosting concern.

An application instance cannot always reach its own public URL from inside its container or Kubernetes pod. This can happen with DNS hairpinning limitations, split-horizon DNS, restricted egress, private clusters, or public ingress protected by Cloudflare Access, a WAF, or another authentication proxy. In those environments, MCP user authentication fails before product authorization runs because the server cannot fetch the signing keys needed to verify the bearer token.

### Problem solved

This PR separates the two URL responsibilities:

- `WEBAPP_URL`, `NEXTAUTH_URL`, and `BETTER_AUTH_URL` continue defining the public OAuth issuer, discovery metadata, redirects, cookies, and MCP audience.
- The new optional server-only `MCP_OAUTH_JWKS_URL` controls only the JWKS fetch used to verify MCP OAuth bearer tokens.

The public issuer's HTTPS `/jwks` endpoint remains the default everywhere. Docker Compose and Helm leave the override unset unless an operator explicitly opts into a complete internal URL on a trusted application network. This avoids silently moving the token-signature trust anchor to plaintext HTTP and avoids guessing a Helm service port or custom subpath.

OAuth verification failures keep the same opaque 401 response, while bounded logs now distinguish a JWKS/network failure from token verification using only `failureSource`, `errorName`, and a structural `errorCode`. Tokens, raw errors, error messages, and endpoint URLs are not logged.

This change does not alter AuthZed, SpiceDB permissions, OAuth token claims, or authorization decisions. It only ensures that deployments which need an internal JWKS route can configure one without changing the public OAuth contract.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2360/make-mcp-user-authentication-work-in-authzed-sandbox-without-a

## How this was tested

- `pnpm --filter @formbricks/web exec vitest run lib/env.test.ts modules/mcp/auth.test.ts modules/auth/lib/oauth-urls.test.ts` ✅ — 3 files, 137 tests on #8846; 147 tests after integration with the synchronized epic.
- Focused Prettier and ESLint on the changed TypeScript files ✅
- Helm default render omits `MCP_OAUTH_JWKS_URL`; explicit `formbricks.mcpOauthJwksUrl` and `deployment.env` values render exactly once ✅
- Docker Compose default render omits `MCP_OAUTH_JWKS_URL`; `.env` remains the opt-in configuration path ✅
- Deployed the earlier opt-in sandbox configuration and verified public OAuth discovery and issuer stayed on the public origin while the pod fetched JWKS from the internal service with HTTP 200 ✅
- Verified an invalid MCP bearer token returns 401; a valid Better Auth user token initializes MCP and executes `list_surveys` with HTTP 200 ✅
- Verified the request produced two `mcp:user` shadow comparisons, both `match`, with no mismatch or operational-error outcome ✅
- Restored the original eight-target/two-organization sandbox shadow cohort and removed the disposable PostgreSQL and SpiceDB fixture ✅

## Breaking changes

| Optional server-only `MCP_OAUTH_JWKS_URL` and Helm `formbricks.mcpOauthJwksUrl` | MCP token verification fetched JWKS from the public issuer URL | The public issuer remains the default; operators may opt into a separate internal JWKS URL | Self-hosters whose runtime cannot reach its public origin | Set the complete trusted internal URL, including the service port and any custom application subpath |

## QA / Test Plan

**How to test**

- [x] Deploy with public `WEBAPP_URL` / `NEXTAUTH_URL` and an explicitly configured internal `MCP_OAUTH_JWKS_URL` → OAuth discovery and issuer metadata still use the public origin.
- [x] From the Formbricks pod, request the configured internal JWKS URL → the endpoint returns a JWKS document.
- [x] Authenticate an MCP client as a user and call the MCP endpoint → bearer authentication succeeds without changing `BETTER_AUTH_URL`.
- [x] Use an invalid bearer token → the MCP endpoint returns the existing unauthorized response and logs a bounded token-verification classification.
- [x] Make the internal JWKS endpoint unavailable → the response remains the same opaque 401 while logs identify `failureSource: jwks_fetch`.
- [x] Configure an invalid URL scheme, credentials, or fragment → environment validation rejects startup with the sanitized variable name.
- [x] Render Docker Compose and Helm without an override → no internal plaintext JWKS URL is injected.

**Preconditions / test data**

- A deployment with MCP OAuth enabled, one active user, and a public application URL.
- Kubernetes or Docker internal service networking available to the Formbricks process when testing the opt-in override.

**Risks & regressions**

- Re-check OAuth discovery, consent redirects, cookie behavior, token issuer, and MCP token audience; none should use the internal URL.
- Internal HTTP is appropriate only on a trusted application network.
- Custom-subpath deployments must configure the full internal JWKS path.
- The existing sandbox PostgreSQL installation does not have the chart's expected `postgres` bootstrap role, so the already-initialized sandbox was upgraded with bundled AuthZed database bootstrap disabled. This is an existing-install chart compatibility issue, not an MCP/JWKS failure.

- No database migration.
- New optional `MCP_OAUTH_JWKS_URL`; Docker and Helm do not enable it automatically.

### DIFF
--- a/apps/web/app/api/(internal)/unify-feedback/sources/csv/import/route.test.ts
+++ b/apps/web/app/api/(internal)/unify-feedback/sources/csv/import/route.test.ts
@@ -1,0 +1,98 @@
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { logger } from "@formbricks/logger";
+import { AuthorizationError } from "@formbricks/types/errors";
+import { assertFeedbackSourceDirectoryAccess } from "@/lib/feedback-source/access";
+import { importCsvFile } from "@/lib/feedback-source/csv-file-import";
+import { getFeedbackSourceWithMappingsById } from "@/lib/feedback-source/service";
+import { getUser } from "@/lib/user/service";
+import { checkAuthorizationUpdated } from "@/lib/utils/action-client/action-client-middleware";
+import { getOrganizationIdFromFeedbackSourceId } from "@/lib/utils/helper";
+import { getSession } from "@/modules/auth/lib/session";
+import { POST } from "./route";
+
+vi.mock("@formbricks/logger", () => ({
+  logger: { error: vi.fn() },
+}));
+vi.mock("@/lib/feedback-source/access", () => ({
+  assertFeedbackSourceDirectoryAccess: vi.fn(),
+}));
+vi.mock("@/lib/feedback-source/csv-file-import", () => ({
+  CsvImportValidationError: class CsvImportValidationError extends Error {},
+  importCsvFile: vi.fn(),
+}));
+vi.mock("@/lib/feedback-source/service", () => ({
+  getFeedbackSourceWithMappingsById: vi.fn(),
+}));
+vi.mock("@/lib/user/service", () => ({ getUser: vi.fn() }));
+vi.mock("@/lib/utils/action-client/action-client-middleware", () => ({
+  checkAuthorizationUpdated: vi.fn(),
+}));
+vi.mock("@/lib/utils/helper", () => ({
+  getOrganizationIdFromFeedbackSourceId: vi.fn(),
+}));
+vi.mock("@/modules/auth/lib/session", () => ({ getSession: vi.fn() }));
+
+const userId = "user_1";
+const organizationId = "organization_1";
+const workspaceId = "workspace_1";
+const feedbackSourceId = "source_1";
+const feedbackDirectoryId = "directory_1";
+
+const makeRequest = (): Request => {
+  const body = new FormData();
+  body.set("workspaceId", workspaceId);
+  body.set("feedbackSourceId", feedbackSourceId);
+  body.set("file", new File(["value\nexample"], "feedback.csv", { type: "text/csv" }));
+  return new Request("http://localhost/api/unify-feedback/sources/csv/import", { method: "POST", body });
+};
+
+describe("CSV feedback source import authorization", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.mocked(getSession).mockResolvedValue({ user: { id: userId } } as never);
+    vi.mocked(getUser).mockResolvedValue({ id: userId } as never);
+    vi.mocked(getOrganizationIdFromFeedbackSourceId).mockResolvedValue(organizationId);
+    vi.mocked(checkAuthorizationUpdated).mockResolvedValue(true);
+    vi.mocked(getFeedbackSourceWithMappingsById).mockResolvedValue({ feedbackDirectoryId } as never);
+    vi.mocked(assertFeedbackSourceDirectoryAccess).mockResolvedValue(undefined);
+    vi.mocked(importCsvFile).mockResolvedValue({ imported: 1 } as never);
+  });
+
+  test("requires exact assignment write access before importing", async () => {
+    const response = await POST(makeRequest());
+
+    expect(response.status).toBe(200);
+    expect(assertFeedbackSourceDirectoryAccess).toHaveBeenCalledWith(
+      userId,
+      feedbackDirectoryId,
+      workspaceId,
+      "write"
+    );
+    expect(importCsvFile).toHaveBeenCalledWith({ feedbackSourceId, workspaceId, file: expect.any(File) });
+  });
+
+  test("preserves the existing forbidden response for a denied assignment", async () => {
+    vi.mocked(assertFeedbackSourceDirectoryAccess).mockRejectedValue(
+      new AuthorizationError("Not authorized")
+    );
+
+    const response = await POST(makeRequest());
+
+    expect(response.status).toBe(403);
+    expect(importCsvFile).not.toHaveBeenCalled();
+  });
+
+  test("does not log identifiers or raw operational errors", async () => {
+    vi.mocked(assertFeedbackSourceDirectoryAccess).mockRejectedValue(
+      new Error(`evaluator failed for ${feedbackDirectoryId} and ${feedbackSourceId}`)
+    );
+
+    const response = await POST(makeRequest());
+
+    expect(response.status).toBe(500);
+    const logOutput = JSON.stringify(vi.mocked(logger.error).mock.calls);
+    expect(logOutput).not.toContain(feedbackDirectoryId);
+    expect(logOutput).not.toContain(feedbackSourceId);
+    expect(logOutput).not.toContain("evaluator failed");
+  });
+});
--- a/apps/web/app/api/(internal)/unify-feedback/sources/csv/import/route.ts
+++ b/apps/web/app/api/(internal)/unify-feedback/sources/csv/import/route.ts
@@ -6,7 +6,9 @@ import {
   InvalidInputError,
   ResourceNotFoundError,
 } from "@formbricks/types/errors";
+import { assertFeedbackSourceDirectoryAccess } from "@/lib/feedback-source/access";
 import { CsvImportValidationError, importCsvFile } from "@/lib/feedback-source/csv-file-import";
+import { getFeedbackSourceWithMappingsById } from "@/lib/feedback-source/service";
 import { getUser } from "@/lib/user/service";
 import { checkAuthorizationUpdated } from "@/lib/utils/action-client/action-client-middleware";
 import { getOrganizationIdFromFeedbackSourceId } from "@/lib/utils/helper";
@@ -94,6 +96,17 @@ export const POST = async (request: Request) => {
       ],
     });
 
+    const feedbackSource = await getFeedbackSourceWithMappingsById(feedbackSourceId, workspaceId);
+    if (!feedbackSource) {
+      throw new ResourceNotFoundError("FeedbackSource", feedbackSourceId);
+    }
+    await assertFeedbackSourceDirectoryAccess(
+      user.id,
+      feedbackSource.feedbackDirectoryId,
+      workspaceId,
+      "write"
+    );
+
     const result = await importCsvFile({ feedbackSourceId, workspaceId, file });
 
     return NextResponse.json(result);
@@ -121,7 +134,10 @@ export const POST = async (request: Request) => {
       return buildCsvImportErrorResponse(error.message, 400);
     }
 
-    logger.error({ error }, "Failed to import CSV feedback source data");
+    logger.error(
+      { errorName: error instanceof Error ? error.name : "unknown" },
+      "Failed to import CSV feedback source data"
+    );
     return buildCsvImportErrorResponse(CSV_IMPORT_FAILED_ERROR_CODE, 500);
   }
 };
--- a/apps/web/app/api/v3/feedbackRecords/lib/access.ts
+++ b/apps/web/app/api/v3/feedbackRecords/lib/access.ts
@@ -1,11 +1,12 @@
 import "server-only";
 import type { logger } from "@formbricks/logger";
 import type { TAuthenticationApiKey } from "@formbricks/types/auth";
-import { AuthorizationError } from "@formbricks/types/errors";
+import { getV3AuthorizationActor } from "@/app/api/v3/lib/auth";
 import { requireUnifyFeedbackWorkspaceAccess } from "@/app/api/v3/lib/feedback-access";
 import { problemBadRequest, problemForbidden, problemUnprocessableContent } from "@/app/api/v3/lib/response";
 import type { TV3Authentication } from "@/app/api/v3/lib/types";
-import { checkAuthorizationUpdated } from "@/lib/utils/action-client/action-client-middleware";
+import { can } from "@/lib/authorization";
+import { getFeedbackDirectoryAssignmentActionForPermission } from "@/lib/authorization/compatibility";
 import {
   getFeedbackDirectoriesByWorkspaceId,
   getFeedbackDirectoryAuthContext,
@@ -48,6 +49,22 @@ export type TResolvedFeedbackTenant = {
 };
 
 type TResolveResult = ({ ok: true } & TResolvedFeedbackTenant) | { ok: false; response: Response };
+
+const canAccessFeedbackDirectoryAssignment = async (
+  authentication: TV3Authentication,
+  feedbackDirectoryId: string,
+  workspaceId: string,
+  minPermission: TTeamPermission
+): Promise<boolean> => {
+  const actor = getV3AuthorizationActor(authentication);
+  if (!actor) return false;
+
+  return can(actor, getFeedbackDirectoryAssignmentActionForPermission(minPermission), {
+    type: "feedbackDirectoryAssignment",
+    feedbackDirectoryId,
+    workspaceId,
+  });
+};
 
 /**
  * Resolve (and authorize) the Hub tenant for a feedback-records request. This is the single tenant-
@@ -97,6 +114,23 @@ export async function resolveWorkspaceFeedbackTenant({
         ),
       };
     }
+    if (
+      !(await canAccessFeedbackDirectoryAssignment(
+        authentication,
+        requested.id.trim(),
+        resolvedWorkspaceId,
+        minPermission
+      ))
+    ) {
+      return {
+        ok: false,
+        response: problemForbidden(
+          requestId,
+          "You are not authorized to access this feedback dataset",
+          instance
+        ),
+      };
+    }
     return {
       ok: true,
       workspaceId: resolvedWorkspaceId,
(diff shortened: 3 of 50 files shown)


charts/formbricks/templates/deployment.yaml
```diff
@@ -131,6 +131,10 @@ spec:
             - name: SKIP_STARTUP_MIGRATION
               value: "true"
             {{- end }}
+            {{- if not (hasKey .Values.deployment.env "MCP_OAUTH_JWKS_URL") }}
+            - name: MCP_OAUTH_JWKS_URL
+              value: {{ .Values.formbricks.mcpOauthJwksUrl | default (printf "http://%s:3000/api/auth/jwks" (include "formbricks.name" .)) | quote }}
```
mattinannt: The default `MCP_OAUTH_JWKS_URL` hardcodes port `3000` instead of deriving it from the configurable container port.

If an operator sets `deployment.ports.http.containerPort` to something other than 3000 (already supported by the chart, and what `templates/service.yaml` derives its port from via `default $config.port $config.containerPort`), the Service listens on the new port but this default still renders `http://<formbricks.name>:3000/api/auth/jwks`. The pod's internal JWKS fetch then fails (connection refused), so every MCP OAuth bearer-token request fails verification while API-key based MCP auth keeps working — a confusing failure mode to diagnose.

Worth deriving the port from the same source `service.yaml` uses, rather than hardcoding `3000`.

---

apps/web/lib/env.ts
```diff
@@ -19,6 +19,26 @@ const ZOpenAICompatibleBaseUrl = z.url().refine(isHttpUrl, {
   message: "AI_OPENAI_COMPATIBLE_BASE_URL must be a valid http(s) URL",
 });
 
+const isValidMcpOauthJwksUrl = (value: string): boolean => {
+  try {
+    const url = new URL(value);
+
+    return (
+      (url.protocol === "http:" || url.protocol === "https:") &&
+      url.hostname.length > 0 &&
```
mattinannt: `isValidMcpOauthJwksUrl` re-implements the http/https protocol check that `isHttpUrl` (defined a few lines above, used by `ZOpenAICompatibleBaseUrl`) already provides.

Not a bug today, but it duplicates the same `url.protocol === "http:" || url.protocol === "https:"` condition in two places in this file. A future change to what counts as an acceptable scheme is likely to be applied to only one of the two functions, silently reintroducing inconsistency between `AI_OPENAI_COMPATIBLE_BASE_URL` validation and `MCP_OAUTH_JWKS_URL` validation.

Consider having `isValidMcpOauthJwksUrl` call `isHttpUrl(value)` and then layer on the extra hostname/credentials/hash checks, instead of duplicating the try/catch and protocol check.

---

mattinannt (commented):
Reviewed the JWKS URL separation change. The core approach is sound — keeping `BETTER_AUTH_URL` as the public issuer while adding an internal-only `MCP_OAUTH_JWKS_URL` for the signing-key fetch is the right fix for the hairpinning problem described. Two minor findings inline, neither blocking:

1. The Helm chart's default JWKS URL hardcodes port 3000 instead of deriving it from the configurable container port.
2. `isValidMcpOauthJwksUrl` duplicates the protocol check already in `isHttpUrl`.

---

charts/formbricks/values.yaml
```diff
@@ -21,6 +21,11 @@ formbricks:
   # Optional: Public URL for surveys (defaults to webappUrl if not set)
   publicUrl: ""
 
+  # Optional: Server-side Better Auth JWKS endpoint for MCP OAuth verification.
+  # Defaults to the in-cluster Formbricks service so the public issuer does not need DNS hairpinning.
+  # Set the complete URL when the app is served below a custom subpath.
+  mcpOauthJwksUrl: ""
```
xernobyl, line 28: **[should] The default moves the token-signature trust anchor to unauthenticated HTTP for every Helm and Docker install — including the ones that don't need it.**

Whatever answers this URL supplies the public keys that authenticate MCP bearer tokens, so it is the trust anchor for the entire `mcp:user` auth path. The `iss`/`aud` checks don't constrain that, since a forged token carries whatever claims the forger chooses. Before this PR the fetch went to the public issuer over authenticated HTTPS; with the chart and compose defaults it goes to plaintext `http://` on the cluster/bridge network.

To be fair on severity: exploiting it needs a pre-existing foothold (occupying the service name, CoreDNS compromise, or pod-to-pod MITM), so this is defense-in-depth rather than a live hole, and treating ClusterIP traffic as trusted is a conventional assumption.

What makes it worth a second look is that it's **opt-out**. It applies to deployments that can reach their public issuer perfectly well, and the trust caveat you wrote ("Use an internal HTTP URL only on a trusted application network") lives in the env-var reference — which isn't what someone installing the chart reads.

Either resolution works:

- leave `mcpOauthJwksUrl: ""` meaning "use the public issuer" and let affected deployments opt in — the code already falls back correctly; or
- keep the default and move the trust assumption into this comment block, next to the value operators actually see.

apps/web/modules/mcp/auth.ts
```diff
@@ -322,7 +323,7 @@ async function authenticateMcpOAuthBearer(
         audience: getMcpResourceUrl(),
         issuer: getAuthIssuerUrl(),
       },
-      jwksUrl: `${getAuthIssuerUrl()}/jwks`,
+      jwksUrl: getMcpOAuthJwksUrl(),
     });
   } catch {
```
xernobyl: **[should] A JWKS fetch failure is indistinguishable from a bad token — precisely where this PR adds a new way to misconfigure the fetch.**

`catch {` discards the error, and the only signal is `log.warn({ statusCode: 401 }, "MCP OAuth authentication failed")`. So an unreachable or 404 JWKS endpoint — wrong port, missing subpath, service not ready during a rollout, typo in the override — produces byte-identical output to an expired or forged token: same 401, same message, no cause recorded.

That's the same failure described in the PR body as taking a sandbox session to pin down, and this change makes it easier to hit: the JWKS URL is now an independently-settable knob that no longer moves in lockstep with the issuer. Previously a wrong issuer broke everything visibly; now it can break only `mcp:user` auth while API-key MCP auth keeps working.

There's a security-observability cost too: "our internal JWKS is unreachable or being tampered with" and "someone is probing with forged tokens" are the same line in the logs.

Binding the error and logging its type/message (never the token) would be enough:

```ts
} catch (error) {
  // ...
  log.warn({ statusCode: 401, err: error }, "MCP OAuth authentication failed");
```

Optionally, separating a fetch/network failure from a verification failure would let the 401 stay deliberately generic while the log tells the operator which one it actually was.

xernobyl (commented):
Reviewed at `96d1631c` against `epic/authzed`. The core design is right, and the part that's easiest to get wrong is the part that's correct: `getMcpOAuthJwksUrl()` has exactly one call site, so the internal URL can't reach discovery metadata, the audience, or the `WWW-Authenticate` challenge — and `oauth-urls.test.ts` pins that invariant directly ("uses an internal JWKS URL without changing the public issuer"). That's the right assertion to have written.

**Verified and clean**, listed so nobody re-derives it:

- `getMcpOAuthJwksUrl()` is used only at `modules/mcp/auth.ts:326`. The 401 challenge emits `getMcpProtectedResourceMetadataUrl()` only, so no internal hostname is exposed to unauthenticated clients.
- The `turbo.json` entry lands on the root `build` task and `apps/web` has no package-level `turbo.json` to replace the list, so the var genuinely participates in the build hash.
- Env validation rejects embedded credentials and fragments, and the raised message carries only the variable name — a URL with credentials can't leak into logs.
- The default host uses `formbricks.name`, the same helper `service.yaml` names the Service with, so it resolves.
- Not SSRF: `MCP_OAUTH_JWKS_URL` is operator-set env and never attacker-influenced.

**Two findings inline**, both `should`, neither blocking.

Not repeating what's already covered: the hardcoded `3000` (Matti's thread) and the custom-subpath docs (CodeRabbit).

**Considered and dropped:** an unreachable JWKS endpoint means every bearer request attempts an outbound fetch before failing, but `createRemoteJWKSet` caches/cools down and the unauthenticated rate limit already applies — below the bar as a DoS vector.

**Merge-conflict heads-up, not a review point:** ENG-2343 (#8835, still draft, targets `main`) rewrites this exact call site — `verifyAccessToken` → `verifyBearerToken` plus a `typ: at+jwt` verify option. Whichever lands second will need to reapply the other's change on that one line. Flagging it early since this is on `epic/authzed` and that one is on `main`.

charts/formbricks/templates/deployment.yaml
BhagyaAmarasinghe: Addressed in 35df054d7 by removing the unsafe chart-generated default entirely. Helm now renders `MCP_OAUTH_JWKS_URL` only when the operator supplies the complete URL, so it no longer hardcodes or guesses a service port.

apps/web/lib/env.ts
BhagyaAmarasinghe: Addressed in 35df054d7: `isValidMcpOauthJwksUrl` now delegates the scheme/parse check to the existing `isHttpUrl` helper and layers only the credentials/fragment constraints.

docs/platform/mcp/overview.mdx
```diff
@@ -96,6 +96,10 @@ searchable a short while after they arrive, and records without text are never s
   / `BETTER_AUTH_URL`). The discovery metadata and browser redirects are built from that URL, so an
   `http://localhost` or misconfigured origin will break the OAuth flow for remote clients. See the
   [self-hosting configuration](/self-hosting/configuration/environment-variables).
+
+  Kubernetes deployments that cannot reach their own public origin can set `MCP_OAUTH_JWKS_URL` to
+  the internal Formbricks `/api/auth/jwks` endpoint. This changes only the server-side key fetch;
+  OAuth discovery, issuers, redirects, cookies, and audiences remain on the public URL.
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: Addressed in 35df054d7: both MCP guides now state that `/api/auth/jwks` applies only without a custom application subpath and include a complete `/formbricks/api/auth/jwks` internal URL example.

docs/self-hosting/configuration/environment-variables.mdx
```diff
@@ -25,6 +25,7 @@ For `AI_PROVIDER=openai-compatible`, the LLM GA v1 self-hosted path is Qwen serv
...
 | NEXTAUTH_URL                            | Legacy-compatible auth base URL. This should normally be the same as WEBAPP_URL; Formbricks derives the Better Auth issuer under `/api/auth`.                                                   | required                                                                                 | http://localhost:3000                                                     |
 | BETTER_AUTH_URL                         | Optional Better Auth base URL override. Set this only when the auth server base differs from NEXTAUTH_URL; preserve any custom subpath.                                                         | optional                                                                                 | NEXTAUTH_URL                                                              |
+| MCP_OAUTH_JWKS_URL                      | Optional server-only URL used to fetch Better Auth signing keys for MCP OAuth. It does not change the public issuer, redirects, cookies, or token audience. Use an internal HTTP URL only on a trusted application network. | optional | `{BETTER_AUTH_URL or NEXTAUTH_URL}/jwks`; Helm uses the internal Formbricks service |
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: Addressed in 35df054d7: the environment reference now distinguishes the application fallback from deployment behavior. Docker Compose and Helm leave the override unset unless an operator explicitly configures it.

charts/formbricks/values.yaml
BhagyaAmarasinghe: Addressed in 35df054d7 with the opt-in resolution: Helm and Docker no longer redirect JWKS fetching to plaintext internal HTTP by default. The public HTTPS issuer remains the fallback; operators must explicitly choose the internal endpoint and the docs place the trusted-network caveat beside the chart value.

apps/web/modules/mcp/auth.ts
BhagyaAmarasinghe: Addressed in 35df054d7: failures keep the same opaque 401 response but logs now include bounded `failureSource`, `errorName`, and structural `errorCode` fields. JWKS/network failures are distinguishable from token-verification failures without logging the token, raw error, URL, or error message.

xernobyl (approved):
Re-reviewed at `35df054d7`, including a fresh security pass on the new error-handling and config-default code.

All four findings from the last round are resolved, cleanly:

- **Trust anchor / opt-out default (mine)** and **hardcoded port (Matti's)** — both resolved the same way: the Helm and Compose defaults are gone. `MCP_OAUTH_JWKS_URL` is now injected only when `formbricks.mcpOauthJwksUrl` is explicitly set, so no deployment gets silently pointed at plaintext HTTP or a wrong port. The Helm validation workflow was flipped to *assert* the var is absent by default — a real regression test, not just a removed line.
- **`isHttpUrl` duplication (Matti's)** — `isValidMcpOauthJwksUrl` now calls `isHttpUrl` instead of re-implementing the protocol check.
- **Fetch failure indistinguishable from a bad token (mine)** — `getMcpOAuthFailureDetails` categorizes `jwks_fetch` vs `token_verification` and logs only `errorCode`/`errorName`. Verified by reading the function: it never touches `.message` or `.stack`, and there's a test (`"distinguishes an unavailable JWKS endpoint without logging the raw error"`) that explicitly asserts the raw error text (`"connection refused"`) never reaches the log output. That's the right way to close an observability gap without opening a new information-disclosure one.

Also closes the basePath variant of the port issue that I hadn't posted yet — same root cause, same fix (no default to go wrong).

**Security review:** no P1/P2. `issuer`/`audience` still come from the public URLs — only key retrieval moves. The new default-removal is strictly safer than what shipped before. Nothing here weakens the OAuth contract.

Docs now cover the custom-subpath case explicitly with a worked example, and `charts/formbricks/README.md` / the env-var reference both describe the opt-in-only behavior accurately.

All 6 threads resolved, CI green (14/14). Approving.

xernobyl:
@BhagyaAmarasinghe heads up — I pushed a merge commit (`73a6b6194`) to this branch to clear the conflicts that appeared when #8863 landed. Shout if you'd rather I hadn't touched it and I'll back it out.

Four conflicts, all from epic moving under the branch rather than from any disagreement with what this PR does. **Everything this PR changes is preserved** — I checked each one rather than resolving mechanically:

**`turbo.json`** — took epic's version wholesale. This branch carried the pre-sync structure (no `globalDependencies`, no `build.inputs`, no `database#generate` task), while its own contribution to that file is a single `passThroughEnv` entry, `MCP_OAUTH_JWKS_URL` — which epic already contains. So there was nothing to port forward.

**`apps/web/modules/mcp/auth.ts`** — the one worth a second look. Epic extracted the inline reject path into `rejectUnauthenticatedMcpRequest()`. I read that helper before taking it: it does exactly what your inline block did — rate limit first, then `log.warn` with `statusCode` last, then the 401 with the OAuth challenge — so no behaviour is dropped. I resolved **only the conflicting hunk**, not the whole file, so your `getMcpOAuthJwksUrl()` wiring at the `verifyAccessToken` call is untouched.

**`oauth-urls.test.ts`** and **`.github/workflows/helm-chart-validation.yml`** — additive on epic's side only (a UserInfo-audience test; deployment lifecycle/volume render assertions). Kept both; your own tests in that spec are unaffected.

Verified on the merged tree: typecheck 12/12, `format:check` clean, **126 tests green** across `modules/mcp` and `oauth-urls`, full web suite **8662 passed** with the 6 known pre-existing `authzed-entrypoints` failures (Node 26 only — green on CI's Node 24). Confirmed `values.yaml`, `deployment.yaml` and the docs changes all survived.

My approval stands.

BhagyaAmarasinghe:
Addressed by repairing the PR branch rather than editing historical migrations.

The unresolved CodeRabbit findings were all attached to an accidental 162-file diff that reintroduced the retired `packages/database/migrations` tree after merge-based branch synchronization. They were not part of the ENG-2360 JWKS change.

The current `epic/authzed` branch already contains the complete internal MCP JWKS implementation through #8863. I realigned this branch to the exact current epic head (`137663366`). The PR now has zero files and zero commits, and GitHub closed it automatically as already integrated.

Validation on the current epic implementation:
- `lib/env.test.ts`
- `modules/auth/lib/oauth-urls.test.ts`
- `modules/mcp/auth.test.ts`
- `app/api/mcp/route.test.ts`

Result: 4 test files passed, 159 tests passed.

PR #8848, opened 2026-08-13
fix(security): bind email-change tokens to credential state [ENG-2106]

Dhruwang:
## What & why

The email-change verification link is consumable **without a session** on purpose — the user clicks it from the new mailbox, often in a different browser. That means for its whole 24-hour life the token *is* a credential that can move the account's login address, and with it where every future password-reset link is delivered.

The token was a bare `{ id, email }` JWT, so it outlived the events that are supposed to end an attacker's hold on an account. Reported flow: given temporary possession of a session **and** the current password, an attacker requests a change to their own address and sits on the token without consuming it. The victim's password recovery revokes every session and rotates the password — but the stale token stays valid, long enough to repoint the account at the attacker's mailbox, receive a real password-reset email there, and reset the password back. (CWE-613 / CWE-284.)

The fix binds the token to the credential state it was issued under:

- `createEmailChangeToken` (`apps/web/lib/jwt.ts`) now mints a `purpose: "email_change"` claim plus a `fingerprint` — an HMAC-SHA256 over the user's current login email and the `updatedAt` of their `credential` `Account` row, looked up by the full `(provider, providerAccountId)` unique tuple so it resolves to the same row `getCredentialPasswordHash` reads. Prisma applies `@updatedAt` on every write to that row, and every password write goes through it (Better Auth's `prismaAdapter`; no raw-SQL password writes exist in the app), so the timestamp moves on every reset or change. HMAC so the claim, which travels in a readable JWT payload, can't be recomputed or probed by whoever holds the token.
- `verifyEmailChangeToken` requires both claims and recomputes the fingerprint from live DB state, rejecting the token on any mismatch (constant-time compare). A password reset or change — or any other email change — re-keys the fingerprint and kills every token issued before it. Consuming a token changes the email, which makes the link single-use for free.
- Fail closed: a user with no `credential` account can neither be issued a token nor verify one.

Binding to `updatedAt` rather than the password hash is deliberate — it keeps password-derived material out of the token entirely, and it invalidates a reset that happens to set the *same* password, which a hash binding would silently have let through. (It also clears the CodeQL `js/insufficient-password-hash` alert the first revision tripped; see the review thread.) The tradeoff is called out in the doc comment: the binding depends on writes to that row going through Prisma, so a future raw-SQL password write must not quietly leave stale tokens live.

Both claims are mandatory, with **no legacy allowance** — accepting a token that carries neither would leave exactly the unbound token this check exists to reject. Links already in flight when this ships stop working; re-requesting the change issues a bound one.

The TTL stays at 24 hours (matching the email copy) and the endpoint stays unauthenticated — shortening the window doesn't fix invalidation, and requiring a session would break clicking the link from another device.

### From review

- **Row scoping** (@mattinannt, @xernobyl): the fingerprint originally matched on `provider: "credential"` alone and took `accounts[0]` unordered. Only `(provider, providerAccountId)` is unique, so a user holding a second credential row could have bound to a timestamp that never moves on a reset — silently reopening the issue for that user, since mint and verify would be wrong identically and agree. Now scoped to the full tuple, with a test pinning it (the failure mode has no other observable symptom).
- **Shared constant-time compare** (@mattinannt): extracted `constantTimeEqual(a, b, encoding)` into `apps/web/lib/crypto.ts` and routed all three copies through it — `jwt.ts`, `single-use-surveys.ts`, and `session-cookie.ts` (which had the same shape inverted inside an `if`). The helper also guards a footgun all three shared: `Buffer.from` silently drops input invalid for its encoding, so two malformed hex values decoded to zero bytes and compared **equal**. Unreachable from current callers, since one side is always a freshly computed digest.

That refactor also required two test-mock fixes, both the same shape: `single-use-surveys.test.ts` and `app/api/v2/client/[workspaceId]/responses/lib/utils.test.ts` replaced `@/lib/crypto` wholesale, so the new export was `undefined` and the resulting TypeError surfaced as a *rejected valid signature*. Both now spread the real module and stub only what they assert on — keeping the signature comparison real, since a stub would let validation pass without comparing anything. The second of those is what turned CI red on `5a02ab6` and is fixed in `46d0f8b`.

## Linear ticket

Fixes ENG-2106

## How this was tested

- Targeted: `pnpm vitest run lib/jwt.test.ts lib/crypto.test.ts lib/utils/single-use-surveys.test.ts modules/auth/lib/session-cookie.test.ts "app/api/v2/client/[workspaceId]/responses/lib/utils.test.ts"` → all green.
- New unit tests in `apps/web/lib/jwt.test.ts` under `verifyEmailChangeToken › credential-state binding`, written behaviorally (mint under one credential state, mutate the state, verify):
  - token minted before a password reset → rejected (**the reported attack**)
  - token replayed after the email already changed → first use succeeds, second rejected
  - credential account gone → rejected
  - unbound token in the pre-fix `{ id, email }` shape → rejected
  - well-formed but non-matching fingerprint → rejected
  - token minted for a different flow (`purpose: "email_verification"`) → rejected
  - the credential lookup stays scoped to this user's own credential row
  - `createEmailChangeToken` refuses to mint for a user with no credential account
- New `constantTimeEqual` tests in `apps/web/lib/crypto.test.ts` (equal, differing, length mismatch, hex decoding, empty, invalid-for-encoding).
- Retargeted one legacy-compatibility test (per-field decryption fallback) from the email-change token — which no longer has a legacy path — to the invite token, which still does.
- **Whole-suite differential.** After the first CI failure I built a CI-equivalent env locally (`bash scripts/setup-dev-env.sh`, `REDIS_URL=` as the workflow does) and ran the full `apps/web` suite (602 files) at this branch and at the previous head, diffing the set of failing files. That's what identified the one regression, and the set is now **identical to the previous head's** — no new failures. (This container still can't run ~81 files for unrelated env/service reasons, which is why the differential rather than an absolute pass count.)
- `eslint`, `prettier --check`, and `tsc --noEmit` clean on all eight changed files.
- CI on `46d0f8b`: Unit Tests, Linters, Better Auth integration tests, both CodeQL analyses, Helm, PR size, translation keys all green.
- Not manually exercised end-to-end: the flow needs SMTP and a second mailbox, which this environment has neither of. See the QA plan below.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] **Happy path.** As a credential (email/password) user, go to Account Settings → Profile, change your email to a second address you control, enter your current password, save. → A verification email arrives at the **new** address. Click its link → `/verify-email-change` shows the success message, the browser is signed out, and you can sign in with the new address and the same password.
- [ ] **The reported attack (must now fail).** Sign in as the victim, request an email change to a second mailbox, and do **not** click the link. Sign out. Run a password recovery for the victim's original address (`/auth/forgot-password`) and set a new password. Now open the still-unclicked verification link from the second mailbox → the page must show **"invalid or expired token"** and the account's email must remain unchanged. Confirm by signing in with the original address and the new password.
- [ ] **Same-password reset.** Repeat the step above but set the recovery password to the *same* value as before. The pending link must still show **"invalid or expired token"**.
- [ ] **Password change, same expectation.** Request an email change, don't click the link, then change the password from Account Settings → Security instead of via recovery. The pending link → **"invalid or expired token"**.
- [ ] **Single use.** Complete a happy-path email change, then click the same verification link a second time → **"invalid or expired token"**; the email stays as the newly-verified address.
- [ ] **Superseded request.** Request a change to address A, then (before clicking) request a change to address B and complete B's link. A's link → **"invalid or expired token"**.
- [ ] **Unchanged rejections.** A malformed/truncated token and an expired one (older than 24h) both → **"invalid or expired token"**.
- [ ] **SSO users unaffected.** As an SSO-provisioned user, the email field in Account Settings → Profile is still not changeable (the action rejects with "Email update is not allowed for non-credential users").
- [ ] **Long-standing pre-Better-Auth account.** On an upgraded self-host, take a credential account created before the Better Auth cutover that has never reset its password since, and run the happy path → the email change still works end to end.
- [ ] **Regression sweep on the shared compare** (three call sites now route through `constantTimeEqual`): sign in and confirm the session sticks across a page reload (signed session cookie), open a single-use survey link and confirm it's accepted once and rejected on reuse, and tamper with a single character of a single-use link's signature → rejected. **Give this one real attention** — a mocked-out version of exactly this comparison is what turned CI red mid-review.

**Preconditions / test data**

- A self-host or staging deployment with SMTP configured (Mailpit is enough) and `EMAIL_VERIFICATION_DISABLED` **unset** — with verification disabled the email changes immediately and no token is involved.
- Two mailboxes you can read: the account's current address and the new one.
- A credential (email/password) account, plus an SSO account for the last check.
- `PASSWORD_RESET_DISABLED` unset, so the recovery step in the attack scenario is available.
- A survey with single-use links enabled, for the shared-compare sweep.

**Risks & regressions**

- Verification links **issued before this deploys** (up to 24h old) are rejected — expected, and users recover by re-requesting the change. Worth watching for support reports in the first day after release.
- `constantTimeEqual` is now on three auth-adjacent paths: session-cookie signature verification (forward-auth / proxy session), single-use survey link signatures, and the email-change fingerprint. Behavior is unchanged for every current caller — same length check, same `timingSafeEqual`, plus a stricter empty/invalid-encoding rejection — but a break here presents as "sessions don't stick" or "valid single-use links rejected", hence the sweep above.
- The fingerprint reads the user's `credential` `Account` row. That is not a new dependency for this flow — requesting an email change already re-verifies the password through the same row (`verifyUserPassword` → `getCredentialPasswordHash`), and the ENG-1054 `credential_account_backfill` migration created it for every pre-cutover user — so no upgraded account should be affected. The pre-Better-Auth QA step above confirms that on real upgraded data.
- Nearby flows sharing `lib/jwt.ts` — signup email verification, invite links, link-survey email verification, SSO recovery — are untouched; a smoke test of signup verification and an org invite is cheap insurance.

- none

## Credit

Reported by Jaejun Cho.

### DIFF
--- a/apps/web/lib/jwt.ts
+++ b/apps/web/lib/jwt.ts
@@ -1,9 +1,9 @@
 import jwt, { JwtPayload, SignOptions } from "jsonwebtoken";
-import { createHmac, timingSafeEqual } from "node:crypto";
+import { createHmac } from "node:crypto";
 import { prisma } from "@formbricks/database";
 import { logger } from "@formbricks/logger";
 import { ENCRYPTION_KEY, NEXTAUTH_SECRET } from "@/lib/constants";
-import { symmetricDecrypt, symmetricEncrypt } from "@/lib/crypto";
+import { constantTimeEqual, symmetricDecrypt, symmetricEncrypt } from "@/lib/crypto";
 import { TGatewayAuthService, getGatewayAuthServiceTokenPurpose } from "@/modules/gateway-auth/lib/service";
 
 const FEEDBACK_RECORDS_GATEWAY_TOKEN_TTL_SECONDS = 60 * 10;
@@ -166,7 +166,15 @@ const getEmailChangeCredentialFingerprint = async (userId: string): Promise<stri
     where: { id: userId },
     select: {
       email: true,
-      accounts: { where: { provider: "credential" }, select: { updatedAt: true } },
+      // Scoped to the full `(provider, providerAccountId)` unique tuple, not just the provider: this
+      // must resolve to the one row Better Auth bumps `updatedAt` on for a password write — the same
+      // row getCredentialPasswordHash reads. Matching on `provider` alone would take an arbitrary
+      // first row if a user ever ended up with a second `credential` row under a different
+      // providerAccountId, binding the token to a timestamp that never moves on a reset.
+      accounts: {
+        where: { provider: "credential", providerAccountId: userId },
+        select: { updatedAt: true },
+      },
     },
   });
 
@@ -181,13 +189,6 @@ const getEmailChangeCredentialFingerprint = async (userId: string): Promise<stri
   return createHmac("sha256", NEXTAUTH_SECRET)
     .update(`${userId}:${user.email}:${credentialUpdatedAt.toISOString()}`)
     .digest("hex");
-};
-
-const fingerprintsMatch = (tokenFingerprint: string, currentFingerprint: string): boolean => {
-  const tokenBytes = Buffer.from(tokenFingerprint, "hex");
-  const currentBytes = Buffer.from(currentFingerprint, "hex");
-
-  return tokenBytes.length === currentBytes.length && timingSafeEqual(tokenBytes, currentBytes);
 };
 
 export const verifyEmailChangeToken = async (token: string): Promise<{ id: string; email: string }> => {
@@ -223,7 +224,9 @@ export const verifyEmailChangeToken = async (token: string): Promise<{ id: strin
   const decryptedId = decryptWithFallback(payload.id, ENCRYPTION_KEY);
   const decryptedEmail = decryptWithFallback(payload.email, ENCRYPTION_KEY);
 
-  if (!fingerprintsMatch(payload.fingerprint, await getEmailChangeCredentialFingerprint(decryptedId))) {
+  const currentFingerprint = await getEmailChangeCredentialFingerprint(decryptedId);
+
+  if (!constantTimeEqual(payload.fingerprint, currentFingerprint, "hex")) {
     throw new Error("Email change token is no longer valid");
   }
 
--- a/apps/web/lib/jwt.ts
+++ b/apps/web/lib/jwt.ts
@@ -144,12 +144,18 @@ export const createTokenForLinkSurvey = (surveyId: string, userEmail: string): s
  * before it stayed usable for the rest of its 24 hours, long enough to point the account at an
  * attacker-controlled mailbox and reset the password back (ENG-2106, CWE-613).
  *
- * Binding the token to the login email plus the credential password hash ties it to that state, so a
- * password reset or change — or any other email change — re-keys the fingerprint and kills every token
- * issued before it. It also makes the link single-use for free: consuming it changes the email.
+ * Binding the token to the login email plus the last write to the user's `credential` account ties it
+ * to that state, so a password reset or change — or any other email change — re-keys the fingerprint and
+ * kills every token issued before it. It also makes the link single-use for free: consuming it changes
+ * the email.
  *
- * HMAC rather than a bare hash so the claim, which travels in a readable JWT payload, discloses nothing
- * about the password hash it covers.
+ * The credential's `updatedAt` is the binding, deliberately not its password hash: no password-derived
+ * material then travels in the token at all, and it still invalidates a reset that happens to set the
+ * *same* password, which a hash would not. This relies on every write to that row going through Prisma,
+ * which applies `@updatedAt` — keep it that way; a raw-SQL password write would leave stale tokens live.
+ *
+ * HMAC rather than a bare hash so the claim, which travels in a readable JWT payload, cannot be
+ * recomputed or probed by whoever holds the token.
  */
 const getEmailChangeCredentialFingerprint = async (userId: string): Promise<string> => {
   if (!NEXTAUTH_SECRET) {
@@ -160,20 +166,20 @@ const getEmailChangeCredentialFingerprint = async (userId: string): Promise<stri
     where: { id: userId },
     select: {
       email: true,
-      accounts: { where: { provider: "credential" }, select: { password: true } },
+      accounts: { where: { provider: "credential" }, select: { updatedAt: true } },
     },
   });
 
-  const passwordHash = user?.accounts?.[0]?.password;
+  const credentialUpdatedAt = user?.accounts?.[0]?.updatedAt;
 
-  // Fail closed: with no credential password there is nothing to bind to, and an email change is only
+  // Fail closed: with no credential account there is nothing to bind to, and an email change is only
   // ever offered to credential users in the first place.
-  if (!user || !passwordHash) {
-    throw new Error("Email change token cannot be bound: user has no credential password");
+  if (!user || !credentialUpdatedAt) {
+    throw new Error("Email change token cannot be bound: user has no credential account");
   }
 
   return createHmac("sha256", NEXTAUTH_SECRET)
-    .update(`${userId}:${user.email}:${passwordHash}`)
+    .update(`${userId}:${user.email}:${credentialUpdatedAt.toISOString()}`)
     .digest("hex");
 };
 
--- a/apps/web/lib/jwt.ts
+++ b/apps/web/lib/jwt.ts
@@ -1,4 +1,5 @@
 import jwt, { JwtPayload, SignOptions } from "jsonwebtoken";
+import { createHmac, timingSafeEqual } from "node:crypto";
 import { prisma } from "@formbricks/database";
 import { logger } from "@formbricks/logger";
 import { ENCRYPTION_KEY, NEXTAUTH_SECRET } from "@/lib/constants";
@@ -25,9 +26,11 @@ const decryptWithFallback = (encryptedText: string, key: string): string => {
  */
 export const LINK_SURVEY_EMAIL_VERIFICATION_PURPOSE = "link_survey_email_verification";
 export const EMAIL_DISPLAY_TOKEN_PURPOSE = "email_display";
+export const EMAIL_CHANGE_TOKEN_PURPOSE = "email_change";
 
 const LINK_SURVEY_TOKEN_TTL = "7d";
 const EMAIL_DISPLAY_TOKEN_TTL = "1d";
+const EMAIL_CHANGE_TOKEN_TTL = "1d";
 
 export const VERIFICATION_TOKEN_PURPOSES = ["email_verification", "sso_recovery"] as const;
 
@@ -130,6 +133,57 @@ export const createTokenForLinkSurvey = (surveyId: string, userEmail: string): s
   );
 };
 
+/**
+ * Fingerprint of the credential state an email-change token was minted against.
+ *
+ * The email-change link is deliberately consumable without a session — the user clicks it from the new
+ * mailbox, often in another browser — so for its whole lifetime the token *is* a credential that can
+ * move the account's login address, and with it where every future password-reset link is delivered. A
+ * bare `{ id, email }` JWT therefore outlived the events that are supposed to end an attacker's hold on
+ * the account: a password recovery revoked every session and rotated the password, yet a token minted
+ * before it stayed usable for the rest of its 24 hours, long enough to point the account at an
+ * attacker-controlled mailbox and reset the password back (ENG-2106, CWE-613).
+ *
+ * Binding the token to the login email plus the credential password hash ties it to that state, so a
+ * password reset or change — or any other email change — re-keys the fingerprint and kills every token
+ * issued before it. It also makes the link single-use for free: consuming it changes the email.
+ *
+ * HMAC rather than a bare hash so the claim, which travels in a readable JWT payload, discloses nothing
+ * about the password hash it covers.
+ */
+const getEmailChangeCredentialFingerprint = async (userId: string): Promise<string> => {
+  if (!NEXTAUTH_SECRET) {
+    throw new Error("NEXTAUTH_SECRET is not set");
+  }
+
+  const user = await prisma.user.findUnique({
+    where: { id: userId },
+    select: {
+      email: true,
+      accounts: { where: { provider: "credential" }, select: { password: true } },
+    },
+  });
+
+  const passwordHash = user?.accounts?.[0]?.password;
+
+  // Fail closed: with no credential password there is nothing to bind to, and an email change is only
+  // ever offered to credential users in the first place.
+  if (!user || !passwordHash) {
+    throw new Error("Email change token cannot be bound: user has no credential password");
+  }
+
+  return createHmac("sha256", NEXTAUTH_SECRET)
+    .update(`${userId}:${user.email}:${passwordHash}`)
+    .digest("hex");
+};
+
+const fingerprintsMatch = (tokenFingerprint: string, currentFingerprint: string): boolean => {
+  const tokenBytes = Buffer.from(tokenFingerprint, "hex");
+  const currentBytes = Buffer.from(currentFingerprint, "hex");
+
+  return tokenBytes.length === currentBytes.length && timingSafeEqual(tokenBytes, currentBytes);
+};
+
 export const verifyEmailChangeToken = async (token: string): Promise<{ id: string; email: string }> => {
   if (!NEXTAUTH_SECRET) {
     throw new Error("NEXTAUTH_SECRET is not set");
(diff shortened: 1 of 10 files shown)


apps/web/lib/jwt.ts
```diff
@@ -130,6 +133,57 @@
...
+  const passwordHash = user?.accounts?.[0]?.password;
+
+  // Fail closed: with no credential password there is nothing to bind to, and an email change is only
+  // ever offered to credential users in the first place.
+  if (!user || !passwordHash) {
+    throw new Error("Email change token cannot be bound: user has no credential password");
+  }
+
+  return createHmac("sha256", NEXTAUTH_SECRET)
+    .update(`${userId}:${user.email}:${passwordHash}`)
```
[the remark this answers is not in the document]
Dhruwang: Fixed in d727184 rather than dismissed — the alert was pointing at something worth changing.

The flagged HMAC wasn't storing a password (the input was already a bcrypt hash, and the output is a token-binding claim, not a credential at rest), but "credential's bcrypt hash flows into SHA-256" is a bad shape to have in the codebase regardless of intent, and the binding doesn't need it.

It now binds to the `credential` account row's `updatedAt` instead of its `password`. Prisma applies `@updatedAt` on every write to that row, and every password write goes through it — Better Auth's `prismaAdapter` for reset/change, and there are no raw-SQL password writes in the app — so the security invariant is unchanged: a reset or change re-keys the fingerprint and invalidates every token issued before it.

Two things get better as a side effect:

- No password-derived material travels in the token at all now, hashed or otherwise.
- A reset that happens to set the **same** password still invalidates pending tokens. Binding to the hash would silently have let them survive, so this is a strictly stronger invariant than what I opened the PR with.

The dependency on writes going through Prisma is called out in the doc comment so a future raw-SQL password write doesn't quietly leave stale tokens live. Tests updated to simulate the reset by moving `updatedAt`; `lib/jwt.test.ts` → 99 passed.

---

apps/web/lib/jwt.ts
```diff
@@ -130,6 +133,63 @@ export const createTokenForLinkSurvey = (surveyId: string, userEmail: string): s
...
+const getEmailChangeCredentialFingerprint = async (userId: string): Promise<string> => {
+  if (!NEXTAUTH_SECRET) {
+    throw new Error("NEXTAUTH_SECRET is not set");
+  }
+
+  const user = await prisma.user.findUnique({
+    where: { id: userId },
+    select: {
+      email: true,
+      accounts: { where: { provider: "credential" }, select: { updatedAt: true } },
```
mattinannt: The fingerprint looks up the credential account only by `provider: "credential"`, taking `accounts[0]` with no ordering and no `providerAccountId` filter. Compare with `getCredentialPasswordHash` in `apps/web/lib/user/password.ts`, which scopes strictly to `provider_providerAccountId: { provider: "credential", providerAccountId: userId }` — precisely because that's the row Better Auth actually bumps `updatedAt` on for every password write.

`Account.provider` has only a `(provider, providerAccountId)` unique constraint, so nothing in the schema prevents a user from ever having a second `provider: "credential"` row with a different `providerAccountId` (account-merge feature, bad migration, manual data fix). If that ever happens and the extra row sorts first, the fingerprint binds to a row whose `updatedAt` never moves on a real password reset — silently reopening the exact ENG-2106 attack this PR closes, for that one user, with no error or test to catch it.

Worth filtering by `providerAccountId: userId` (or ordering deterministically) to guarantee the same row `getCredentialPasswordHash` uses.

---

apps/web/lib/jwt.ts
```diff
@@ -130,6 +133,63 @@ export const createTokenForLinkSurvey = (surveyId: string, userEmail: string): s
...
+  if (!user || !credentialUpdatedAt) {
+    throw new Error("Email change token cannot be bound: user has no credential account");
+  }
+
+  return createHmac("sha256", NEXTAUTH_SECRET)
+    .update(`${userId}:${user.email}:${credentialUpdatedAt.toISOString()}`)
+    .digest("hex");
+};
+
+const fingerprintsMatch = (tokenFingerprint: string, currentFingerprint: string): boolean => {
```
mattinannt: Minor: `fingerprintsMatch` (compare buffer lengths, then `timingSafeEqual`) duplicates the same shape already in `apps/web/lib/utils/single-use-surveys.ts` (lines ~61-65: `expected.length === received.length && timingSafeEqual(expected, received)`). Not a bug, but a future fix to one constant-time-compare copy (e.g. an encoding tweak) is easy to make in one file and miss in the other. Worth extracting a shared `constantTimeEqual(a, b, encoding)` helper.

---

mattinannt (commented):
Reviewed the HMAC-fingerprint binding fix. The core design is sound: fail-closed on a missing credential account, constant-time fingerprint compare, mandatory `purpose`+`fingerprint` claims with no legacy allowance, and binding to `updatedAt` rather than the password hash (avoids putting password-derived material in the token and catches same-password resets that a hash binding would miss). Test coverage for the reported attack path, replay, and malformed/wrong-purpose tokens looks thorough.

Two lower-severity items left as inline comments — neither blocks this fix, but the first is worth a look before merge since it's a latent gap in the same invariant this PR is closing.

---

xernobyl (commented):
Reviewed at head 8b9b144, including a security pass on the token-binding logic. CI is fully green and CodeQL's earlier `js/insufficient-password-hash` alert is already resolved (the binding moved from the credential's password hash to its `updatedAt` in this same head). CodeRabbit skipped this PR (path filters excluded all 4 changed files), so no bot findings to reconcile there.

**Seconding Matti's still-open finding — not yet addressed as of this head:**

**[P2] `getEmailChangeCredentialFingerprint` can bind to the wrong `Account` row** — `apps/web/lib/jwt.ts:169,173`

```ts
accounts: { where: { provider: "credential" }, select: { updatedAt: true } },
...
const credentialUpdatedAt = user?.accounts?.[0]?.updatedAt;
```

This scopes only by `provider: "credential"` and takes `accounts[0]` with no ordering and no `providerAccountId` filter. `apps/web/lib/user/password.ts:16` (`getCredentialPasswordHash`) reads this same row for password verification and scopes strictly to `provider_providerAccountId: { provider: "credential", providerAccountId: userId }` — the compound unique key — specifically because that's the one row Better Auth bumps `updatedAt` on for every password write. Nothing in the schema stops a user from ever having a second `provider: "credential"` row with a different `providerAccountId` (bad migration, manual fix, future account-merge). If one exists and sorts first, the fingerprint binds to a row whose `updatedAt` never moves on a real reset — silently reopening the ENG-2106 attack for that user, with no test or error to catch it, since mint and verify are both wrong the same way.

Not exploitable through any live signup/reset path today (those only ever create one correctly-keyed row), so this is a latent robustness gap rather than an active hole today — but it's the exact invariant this PR exists to guarantee, and the fix is a one-line mirror of the lookup `getCredentialPasswordHash` already uses two files over.

**[P3] `fingerprintsMatch` duplicates the constant-time-compare shape in `apps/web/lib/utils/single-use-surveys.ts` (~61-65)** — also Matti's, still open, cosmetic. Worth a shared `constantTimeEqual` helper at some point.

Everything else checked out on my pass: the HMAC design, fail-closed on no credential account, mandatory `purpose`+`fingerprint` with no legacy allowance, the single-use-for-free property from binding the login email into the fingerprint, and the attack/replay/malformed-token test coverage.

apps/web/lib/jwt.ts
Dhruwang: Good catch — fixed in 5a02ab6. You're right, and the reason it's worth fixing even though no live path creates a second row is that the failure is *silent*: mint and verify would both read the same wrong row, so the fingerprints would still agree and the token would verify happily against a timestamp that never moves. No error, no failing test — just that user's tokens quietly surviving a password reset again.

Now scoped to the full unique tuple, so it resolves to the same row `getCredentialPasswordHash` reads:

```ts
accounts: {
  where: { provider: "credential", providerAccountId: userId },
  select: { updatedAt: true },
},
```

I also added a test asserting the lookup stays scoped to the user id. Normally I'd avoid asserting query shape, but here the invariant has no observable behavior to test against — a widened filter wouldn't fail any other test in the file, which is exactly what makes it worth pinning.

---

apps/web/lib/jwt.ts
Dhruwang: Done in 5a02ab6 — `constantTimeEqual(a, b, encoding = "utf8")` now lives in `apps/web/lib/crypto.ts`.

There were **three** copies, not two: `session-cookie.ts:56-60` had the same shape inverted inside an `if`. All three now route through the helper.

Extracting it also surfaced a footgun all three copies shared. `Buffer.from` silently drops input that's invalid for its encoding, so `Buffer.from("zz", "hex")` is an empty buffer — meaning two malformed values decode to zero bytes, pass the length check, and `timingSafeEqual` reports them **equal**. Not reachable in any current caller (one side is always a freshly computed digest), but it's precisely the sort of thing that bites when a future caller compares two attacker-influenced values. The helper returns `false` when either side decodes to nothing.

One test change worth flagging: `single-use-surveys.test.ts` replaced its full-module `@/lib/crypto` mock with a spread of the real module, keeping `constantTimeEqual` real and stubbing only the two functions it asserts on. Adding a stubbed `constantTimeEqual` to that mock would have made signature validation pass without comparing anything — and re-implementing it in the mock would have recreated the exact duplication you flagged, one layer down.

Note on verification: `lib/crypto.test.ts` can't run in my container (pre-existing — it dies importing `@/lib/constants` on env validation, collecting zero tests), so I exercised the real helper's nine cases through a throwaway spec that mocks constants the way `jwt.test.ts` does, confirmed them, and deleted it. CI runs the real file with proper env.

---

xernobyl (approved):
Tested, looks good from my side — approving :slightly_smiling_face:

I took the QA plan for a spin on a local deploy with SMTP and two mailboxes, since that was the part called out as not exercised end to end:

- **The reported attack.** Requested the change, left the link unclicked, ran a real password recovery, then opened the link → "invalid or expired token", and the account email was untouched. The credential row's `updated_at` did move on the reset (and the hash rotated), so the binding is doing what the description says.
- **Happy path** → success page, signed out, login with the new address and the same password. **Second click** on the same link → rejected. **Re-request after a reset** → the new token carries a different fingerprint.

A few things I noticed that I would *not* hold the merge for — follow-up material, your call on what's worth a ticket:

**[P2] The invariant everything rests on isn't covered by a test.** `getEmailChangeCredentialFingerprint` is only as good as "every password write bumps `Account.updatedAt`", and the unit tests mock prisma, so nothing goes red if that stops being true. It *is* true today — I checked against real Better Auth + real Postgres — but ENG-2343 has the 1.7 upgrade queued, and a change in how the adapter writes that row would land on a green suite. We already have the harness for this (`*.integration.test.ts`); two cases would cover it: a real reset kills a pending token, and a sign-in does not.

**[P3] The happy path reads as broken on any dev server.** `apps/web/modules/auth/verify-email-change/components/email-change-sign-in.tsx:17` — dev StrictMode double-invokes the effect, so the action fires twice concurrently. The first call succeeds and moves the email, the second is then correctly rejected (the email just changed, which re-keys the fingerprint), and its `setStatus("error")` overwrites the success. The email really does change; only the page lies. Confirmed by flipping `reactStrictMode: false` and re-running — same flow, success page. Production is fine, but this cost me a while before I spotted it, and whoever runs QA item 1 locally will hit it too. An in-flight `useRef` guard would settle it, or just a note in the plan. Worth mentioning that the action used to be idempotent, so a duplicate call was harmless before this change.

**[P3] Rejected tokens land in Sentry as unexpected server errors.** `apps/web/modules/auth/verify-email-change/actions.ts:15` — the throws are plain `Error`, so `handleServerError` treats them as unexpected and does `captureException` plus a `SERVER ERROR` log. Pre-existing shape, but this PR adds two rejection classes that will actually be common: every link in flight on deploy day (which the description already flags), and every second click of a consumed link, which used to just succeed. `InvalidPasswordResetTokenError` in `packages/types/errors.ts` is the same shape if we want these to stay quiet — the page copy doesn't change either way.

**[P3] `sendVerificationNewEmail` swallows the new throw.** `apps/web/modules/email/index.tsx:139` — `createEmailChangeToken` can now reject with "cannot be bound", the catch turns that into `false`, and `handleEmailUpdate` ignores the return, so the user would see success with no email sent. Unreachable today (`verifyUserPassword` gates on the same row), so this is really just noting that a new failure mode now sits behind an existing silent catch.

Nothing here blocks — the fix itself holds up and the row-scoping and shared-compare items from the earlier round are properly addressed.

Dhruwang:
Thanks for actually running the QA plan end to end — the `updated_at`-moved-on-reset confirmation against real Better Auth + real Postgres is the part I couldn't do from my environment, so that's the most useful thing anyone could have checked here.

**Nothing pushed to this PR** — it's in the merge queue at `46d0f8b`, and none of the four items block, so I'm not going to eject it from the queue to merge a SHA nobody reviewed. Disposition below; the P3 code fixes are written and locally verified, held for a follow-up PR off merged `main`.

**[P2] Integration coverage for the `updatedAt` invariant — agreed, and it's the one I'd prioritize.** You've put your finger on the real weakness: the unit tests mock prisma, so the assumption the whole binding rests on is asserted nowhere, and ENG-2343 is exactly the change that could quietly falsify it. I'm deliberately *not* pushing this one: the harness needs Docker/Postgres, which my environment doesn't have, so CI would be its first execution — and I already turned this PR's CI red once. An integration test whose first run is on someone else's machine isn't worth much. Worth a ticket, tied to ENG-2343 so it lands before the upgrade rather than after; the two cases you named (a real reset kills a pending token, a sign-in does not) are the right shape.

**[P3] StrictMode double-invoke — confirmed, fixed.** Good catch, and thank you for chasing it to `reactStrictMode: false` rather than filing it as "flaky". You're right that this is new: the action was idempotent before, and the binding is what made a duplicate call fail. Guarded with a ref keyed on the token so the effect submits each token once; the rejected second call can no longer overwrite the first one's success. Per AGENTS.md I haven't added a unit test (`.tsx`), and there's no E2E for this flow, so it's verified by lint/types and reasoning only — flagging that rather than implying more.

**[P3] Sentry noise — confirmed, fixed.** `handleServerError` treats anything not in `EXPECTED_ERROR_NAMES` as unexpected, so you're right that both new rejection classes would page. The routine rejections now throw `InvalidInputError` — including a wrap around `jwt.verify`, since `TokenExpiredError`/`JsonWebTokenError` come out of the library as plain errors and expiry is the single most common case. Deliberately left as plain `Error`: unset secrets, and a user with no credential row. Those mean state that shouldn't exist, and should still page. Added tests pinning both sides of that line, since the distinction is easy to erase later. No change to the page copy.

**[P3] `sendVerificationNewEmail` — the mechanism here isn't right, and the real gap is a different one.** That catch rethrows (`logger.error(...)` then `throw error`, `modules/email/index.tsx:138-141`), so the new "cannot be bound" rejection does propagate rather than becoming `false`.

The genuine hole is the one next to it: `sendEmail` returns `false` **without throwing** when SMTP isn't configured, `sendVerificationNewEmail` passes that boolean straight through, and `handleEmailUpdate` ignores the return value — so "success, no email sent" is reachable today, independent of this PR. That's precisely the trap ENG-2091 called out and fixed for `sendResetPassword` and `sendVerificationEmail` by throwing on a falsy return; this third mailer never got the same treatment. Left out of the follow-up as unrelated to this diff, but it deserves its own ticket and it's a three-line fix.

---

PR #8850, opened 2026-08-13
chore(eslint): restore type-aware rules dropped in the flat-config migration

pandeymangg:
## What & why

The ESLint 9 flat-config migration (#8505) replaced the `@vercel/style-guide` tiers, which set `parserOptions.project`, with `tseslint.configs.recommended` — which has no type information. A whole class of type-aware rules (`no-floating-promises`, `no-unsafe-*`, `await-thenable`, `only-throw-error`, `unbound-method`, …) silently stopped running in 12 of 15 packages, including `js-core`, which ships to customers.

This PR restores type-aware linting where it existed before:

- **`library` and `react` tiers** now build on `tseslint.configs.recommendedTypeChecked` with `projectService`. Each tier is exported as a factory taking `{ tsconfigRootDir }`, and all 12 consuming packages pass `import.meta.dirname`, so the project resolves per-package. Non-TS files get `disableTypeChecked` so they don't hard-error.
- **`@typescript-eslint/no-unnecessary-condition`** re-added individually on both tiers (it lives in `strictTypeChecked`, which we deliberately do not adopt wholesale).
- **112 violations fixed** across the packages, mostly real fixes: 39 auto-fixed redundant type assertions, storage's async-generator test mocks made sync, js-core's `setUserId`/`setAttributes` keep their declared `Promise<Result<…>>` public types. A handful of narrow, justified inline disables remain (e.g. vitest asymmetric matchers, a transaction-callback closure TS can't see through).
- **Review caught a real bug the new rules surfaced**: js-core's legacy localStorage migration dereferenced `newState.user` unguarded — a user-less legacy config (an old first-migration format really did persist those) would throw before the graceful reset+resync path, leaving the SDK permanently dead for that visitor. Fixed with honest typing: `TLegacyConfig.user` is `Partial<TUserState>` (the blob is unchecked JSON, so inner fields can't be trusted either), the guard covers both levels (`user?.data?.userId`), and a shapeless user is replaced by a default state (or rebuilt into a complete one when its `data` survived). Regression tests cover the user-less and data-less shapes; both fail on the unguarded version.
- **Stale directives and markers cleaned**: the stale `eslint-disable` directives left by #8505 are gone (ones that became live again under the restored rules were kept), and no `ENG-1677` references remain in code — actionable markers now point at the live tickets ENG-2366 (react-hooks compiler-era rules) and ENG-2264 (apps/web typescript-eslint baseline).
- **`apps/web` lint behaviour is unchanged** — its tier gains no rules; the apps/web diff is comment text plus removal of 6 already-dead disable directives.

## Linear ticket

Fixes ENG-2046

## How this was tested

- `pnpm lint` — 20/20 tasks ✅, zero unused-disable-directive warnings workspace-wide
- `pnpm typecheck` — 25/25 tasks ✅
- `pnpm test` — 25/25 tasks ✅ (js-core 294 tests incl. three legacy-config regression tests; apps/web 602 files / 7850 tests)
- `pnpm format:check` — clean ✅
- `eslint --print-config` on a `.ts` file in js-core, storage, database and survey-ui shows `projectService: true` with the correct `tsconfigRootDir`, and `no-floating-promises` / `no-unnecessary-condition` at error; the same check on `apps/web/app/layout.tsx` shows none of the type-checked rules (tier untouched)
- Regression tests mutation-verified: re-introducing the unguarded `newState.user` access (or the unguarded `.data.userId` flagged in review) makes them fail with the original TypeError path
- `grep -rn "ENG-1677" apps packages` (excluding build dirs) — zero hits

## Breaking changes

None

## QA / Test Plan

Almost all of this diff is dev-tooling (lint config + mechanical code fixes) with no user-facing surface. Two runtime-adjacent spots are worth a manual pass:

**How to test**

- [ ] On a site embedding the Formbricks JS SDK (app surveys set up), open the site with a fresh browser profile → SDK initializes, app surveys display and submit as before
- [ ] Edge case (legacy config migration): before loading the site, seed localStorage key `formbricks-js` with a pre-2.0-style config that has `environmentId` but **no `user` object**, then reload → the SDK must still initialize cleanly (it discards the legacy config and re-syncs); previously-broken behavior would be the SDK silently doing nothing on every reload
- [ ] Same edge case, second shape: seed the legacy config with `"user": {}` (present but empty) → same expected result: SDK initializes cleanly and re-syncs
- [ ] Render a link survey containing a matrix/rating question and answer it → values save and the summary shows them (covers the `survey-ui` form-field cleanup)

**Preconditions / test data**

- A workspace with one app survey (always-on trigger) and one link survey containing a matrix question; a page embedding the js-core SDK

**Risks & regressions**

- js-core SDK boot/setup path (localStorage config migration) — re-check that returning visitors with a valid existing config are untouched
- Survey rendering via the prebuilt surveys bundle (survey-ui feeds it) — a fresh build is part of CI, but self-built setups need a rebuild of `@formbricks/surveys`

- none


packages/js-core/src/lib/common/setup.ts
```diff
@@ -89,8 +89,10 @@ export const setup = async (
     // If the js sdk is being used for non identified users, and we have a new state to update to after migrating, we update the state
     // otherwise, we just sync again!
 
-    if (newState && !newState.user?.data?.userId) {
-      config.update(newState);
+    if (newState && !newState.user?.data.userId) {
```
mattinannt: This line dropped a null-safety check the original code had. Before this PR: `newState.user?.data?.userId`. After: `newState.user?.data.userId` — the `?.` before `.userId` is gone, so `data` is now dereferenced unconditionally once `user` is truthy.

`TUserState.data` is typed as required, which is presumably why `no-unnecessary-condition` flagged the old `?.` as redundant. But `newState` comes from `JSON.parse(existingConfig) as TLegacyConfig` — an unchecked cast over arbitrary localStorage content, the same class of gap this PR just fixed one level up (a user-less legacy blob). If a legacy blob has `user` present but without a `data` key (a plausible intermediate format given the PR description references multiple prior migration formats), `newState.user?.data` short-circuits nothing (since `user` is truthy) and `.userId` throws on `undefined` — reproducing the exact "SDK permanently dead for that visitor" failure this PR set out to fix, just one field deeper. The new regression test only covers a fully-missing `user`, not this shape.

Worth restoring the second `?.` (or adding an inline disable if the type guarantee is trusted) so this stays as defensive as the migration path it lives in.

---

mattinannt (commented):
Reviewed. Nice catch fixing the user-less legacy config crash and adding type-aware linting back — one follow-up worth a look inline: the fix for that exact bug class removed a layer of null-safety (`?.data.userId` vs the original `?.data?.userId`) that could let the same crash resurface one field deeper for a legacy blob with a shapeless `user`.

---

packages/js-core/src/lib/common/setup.ts
pandeymangg: Good catch — fixed in 0d5cd2fb5. `TLegacyConfig.user` is now `Partial<TUserState>` (the blob is unchecked JSON, so inner fields can't be trusted either), the guard is back to `user?.data?.userId`, and a user with surviving `data` gets rebuilt into a complete state instead of being stored shapeless. Added regression tests for both shapes — the `user: {}` one fails on the previous version.

mattinannt (approved):

PR #8851, opened 2026-08-13
chore: add CODEOWNERS with CI validation

mattinannt:
## What & why

The repo has no CODEOWNERS file, so nothing tells an automated reviewer-assignment step who to ask about a given path. This adds one whose sole job is to answer "who knows this part of the app best?" per path.

It is **advisory**. It is not wired into branch protection and cannot block a merge.

Design decisions worth knowing before reading the diff:

- **Named individuals, not teams.** For a blocking gate a team slug is right, but here it destroys the signal — `@formbricks/engineering` on `packages/database/migration/` says nothing useful. "Who knows this best" resolves to a person.
- **Owners are listed primary-first.** GitHub ignores that order; it is the signal for humans and routing tools. Documented in the file header.
- **Two to three owners per path**, so no path is orphaned by a holiday or a departure. GitHub treats multiple owners as "any of", so extra names widen the pool rather than adding review burden. All 98 rules carry two or three.
- **No catch-all `*` rule.** An unmatched path means "no designated expert" and falls through to normal reviewer selection. 95% of tracked files resolve to an owner; the rest are static assets, machine-generated locales and root dotfiles.
- **14 commented sections ordered broad → specific**, with high-risk overrides (auth, billing, licensing, migrations, env/secrets) strictly last, because the last matching pattern wins. Precedence comes from section order; ordering *within* a section is alphabetical purely for scannability and carries no meaning.

Assignments were derived from each engineer's specialism, using 18 months of per-path commit history as the tiebreaker rather than the primary signal. Ordering strictly by commit volume would have made one person the first-named owner on most of the repo, which defeats the purpose.

The second file is a standalone validation workflow. CODEOWNERS fails **silently**: an unparseable line is skipped, an owner without write access is dropped, and a pattern whose directory got renamed simply stops matching. Nothing surfaces any of that. The workflow runs `syntax`, `duppatterns`, `files` and `owners`, plus the experimental `avoid-shadowing` check that enforces the broad→specific ordering the file depends on.

It is deliberately **outside** `pr.yml`'s `required` gate, so ownership metadata going stale never blocks unrelated work. The weekly cron matters as much as the PR trigger: renaming a directory elsewhere breaks a pattern without touching CODEOWNERS, so the `paths` filter would never fire for it.

`owners` is dropped for pull requests **from a fork**, because GitHub does not forward repository secrets to those runs — the token would resolve empty and the check would fail on authentication, putting a red X on an outside contributor's PR for a reason unrelated to their change. Fork PRs keep the other four checks, a log notice explains the reduced set, and `owners` stays covered by the weekly run against `main`.

## Linear ticket

No ticket — repo tooling/config change, raised directly rather than through planning.

## How this was tested

- **The validation workflow passes on this PR**, with all five checks executing against the real file:
  ```
  ==> Executing Duplicated Pattern Checker (78.056µs)            Check OK
  ==> Executing Valid Syntax Checker (313.727µs)                 Check OK
  ==> Executing [Experimental] Avoid Shadowing Checker (1.51ms)  Check OK
  ==> Executing File Exist Checker (34.11ms)                     Check OK
  ==> Executing Valid Owner Checker (868.75ms)                   Check OK
  5 check(s) executed, no failure(s)
  ```
  `Valid Owner Checker` confirms every named owner is a valid org member holding write access — the sub-second timing is live API calls, versus microseconds for the offline checks. `avoid-shadowing` is the meaningful one for the file's structure: it proves no later rule silently voids an earlier override.
- **Full CI green**: 16 checks, no failures.
- **GitHub's own CODEOWNERS parser: 0 errors** (`GET /repos/formbricks/formbricks/codeowners/errors?ref=…`). This is the authoritative check that no line will be silently skipped.
- **Independent resolution** with `hmarr/codeowners` across 26 representative paths, confirming intent rather than just validity: `modules/auth/` beats the `modules/` default, `modules/ee/ai-translation/` correctly falls through to that default, `locales/de-DE.json` is unowned while `locales/en-US.json` is owned, `charts/formbricks/values.yaml` routes to the infra owner.
- **The alphabetical re-sort was verified behaviour-preserving, not assumed**: resolving all 4,566 tracked files against the file before and after the sort produces byte-identical owners.
- **Every owner verified to hold write access** via the collaborators API, since GitHub silently drops owners who don't.
- `prettier --check` clean; the pre-commit hook passed. No unit or E2E tests apply — this is configuration, and `AGENTS.md` discourages low-signal coverage.

## Breaking changes

None

## QA / Test Plan

Not user-facing: nothing in the product changes, so there is nothing to exercise in the running app. The observable checks are all in the GitHub UI.

**How to test**

- [ ] Open the **Checks** tab on this PR → "Validate CODEOWNERS" has run and passed, and its log shows `5 check(s) executed, no failure(s)`.
- [ ] Browse to `.github/CODEOWNERS` on this branch → GitHub renders it with no syntax-error banner. A banner would name the offending line.
- [ ] After merge, open any file under `charts/formbricks/` in the GitHub UI and hover the shield icon near the file header → tooltip reads `Owned by BhagyaAmarasinghe (from CODEOWNERS line N)`. This is also how you debug which rule won for any path.
- [ ] After merge, open a PR touching only `apps/web/modules/ui/` → the UI/design-system owner is auto-requested, not the whole team.
- [ ] Edge case, after merge: a PR touching only `apps/web/locales/de-DE.json` → **no** owner is requested, because machine-translated output is deliberately unowned. A PR touching `apps/web/locales/en-US.json` **does** request owners.
- [ ] Edge case, fork PR: open a PR from a fork that edits `.github/CODEOWNERS` → the workflow runs `4 check(s)` instead of 5, and the log carries a notice explaining that `owners` was skipped because repository secrets are unavailable. It should **not** fail.

**Preconditions / test data**

- A `CODEOWNERS_VALIDATOR_TOKEN` repository secret must exist for the `owners` check: a **classic** PAT with the `read:org` scope, belonging to an organization member. Fine-grained tokens do not send the `X-OAuth-Scopes` header the validator reads and fail with `missing scopes: "read:org"`. `read:org` is the only scope needed — the validator's own docs also suggest `public_repo`, which over-specifies for a public repo. Note this diagnosis applies only to same-repo runs; on a fork the secret is never forwarded at all, which is why `owners` is skipped there rather than attempted.
- No app, database, seed data, or feature flag needed.
- Note the chicken-and-egg: GitHub always reads CODEOWNERS from the pull request's **base** branch, so the routing behaviour above cannot be observed on this PR itself — only once merged to `main`. The two reviewers on this PR were resolved from the branch version of the file by hand for that reason.

**Risks & regressions**

- Reviewer routing is the only behaviour introduced. Because nothing is wired into branch protection, a wrong or missing owner cannot block a merge — the worst case is a review request going to the wrong person.
- Watch for review-request noise on the broadest rules (`/apps/web/app/`, `/apps/web/lib/`, `/apps/web/modules/`). If a section proves noisy, narrow that rule rather than deleting owners.
- Coverage gap to be aware of: an ownership defect introduced by a fork PR is not caught until the weekly scheduled run, since `owners` cannot run without secrets there.
- The validator workflow can fail for a token reason rather than an ownership one: classic PATs expire, and on expiry the weekly run goes red. That is expected and is a credential renewal, not a CODEOWNERS defect.
- If rules are ever appended to the **end** of the file, they will shadow the high-risk overrides. The `avoid-shadowing` check exists to catch exactly that, and the file header says so.

- No DB migrations, no application env vars, no `.env.example` change.
- One repository **secret** is required, as described above: `CODEOWNERS_VALIDATOR_TOKEN`.

https://claude.ai/code/session_01Lfb7QA7A5V5KCvnWmmS43w

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -42,7 +42,11 @@
 #     route segment such as `[workspaceId]` in a pattern — the brackets are
 #     unsupported. Own the `modules/` path instead of the `app/` route path.
 #  6. Owners must have WRITE access to this repo, or they are silently dropped.
-#  7. Within a section, keep rules ordered by depth, then alphabetically.
+#  7. Within a section — or within a `# --- labelled group ---` inside one —
+#     keep rules in alphabetical order, with no exceptions, so there is exactly
+#     one obvious place to insert a new rule. Ordering *between* sections is
+#     what controls precedence (rule 1); ordering *within* one is purely for
+#     scannability, and no rule in this file relies on it.
 #
 #  Changes to this file only take effect once merged — GitHub always reads the
 #  version on the pull request's BASE branch.
@@ -59,15 +63,15 @@
 #  1. Repo, build & tooling
 # ------------------------------------------------------------------------------
 /package.json                              @xernobyl @BhagyaAmarasinghe
-/pnpm-workspace.yaml                       @xernobyl @BhagyaAmarasinghe
-/prisma.config.mjs                         @Dhruwang @xernobyl
-/sonar-project.properties                  @BhagyaAmarasinghe @xernobyl
-/turbo.json                                @xernobyl @BhagyaAmarasinghe
-/scripts/                                  @BhagyaAmarasinghe @xernobyl
 /packages/config-eslint/                   @xernobyl @itsjavi
 /packages/config-prettier/                 @xernobyl @itsjavi
 /packages/config-typescript/               @xernobyl @itsjavi
 /packages/vite-plugins/                    @xernobyl @itsjavi
+/pnpm-workspace.yaml                       @xernobyl @BhagyaAmarasinghe
+/prisma.config.mjs                         @Dhruwang @xernobyl
+/scripts/                                  @BhagyaAmarasinghe @xernobyl
+/sonar-project.properties                  @BhagyaAmarasinghe @xernobyl
+/turbo.json                                @xernobyl @BhagyaAmarasinghe
 
 
 # ------------------------------------------------------------------------------
@@ -114,30 +118,32 @@
 # ------------------------------------------------------------------------------
 #  6. Testing & QA
 # ------------------------------------------------------------------------------
+/apps/web/playwright/                      @itsjavi @Dhruwang @pandeymangg
 /playwright.config.ts                      @itsjavi @Dhruwang
 /playwright.service.config.ts              @itsjavi @Dhruwang
-/apps/web/playwright/                      @itsjavi @Dhruwang @pandeymangg
 
 
 # ------------------------------------------------------------------------------
 #  7. apps/web — broad defaults, app config & instrumentation
 #
-#  The first three rules cover the whole Next.js app. Everything in sections
-#  8-14 overrides them for a narrower path, so keep them here at the top.
+#  `app/`, `lib/` and `modules/` are the broad defaults covering the whole
+#  Next.js app; sections 8-14 override them for narrower paths. Their position
+#  inside this section does not matter — precedence comes from section order —
+#  so they sit in the same alphabetical run as everything else here.
 #
 #  `scripts/` reads deployment secrets and validates env at container start;
 #  `proxy*` is the request-routing edge. Both are more sensitive than their
 #  location suggests.
 # ------------------------------------------------------------------------------
 /apps/web/app/                             @Dhruwang @pandeymangg @xernobyl
+/apps/web/instrumentation*                 @BhagyaAmarasinghe @xernobyl
+/apps/web/integration/                     @xernobyl @itsjavi
 /apps/web/lib/                             @Dhruwang @xernobyl @pandeymangg
 /apps/web/modules/                         @Dhruwang @pandeymangg @xernobyl
-/apps/web/integration/                     @xernobyl @itsjavi
-/apps/web/scripts/                         @BhagyaAmarasinghe @xernobyl
-/apps/web/instrumentation*                 @BhagyaAmarasinghe @xernobyl
 /apps/web/next.config.mjs                  @xernobyl @BhagyaAmarasinghe
 /apps/web/prometheus.yml                   @BhagyaAmarasinghe @xernobyl
 /apps/web/proxy*                           @xernobyl @BhagyaAmarasinghe
+/apps/web/scripts/                         @BhagyaAmarasinghe @xernobyl
 /apps/web/sentry.*                         @BhagyaAmarasinghe @xernobyl
 /apps/web/vite*                            @itsjavi @xernobyl
 
@@ -151,6 +157,7 @@
 /apps/web/modules/analysis/                @Dhruwang @pandeymangg
 /apps/web/modules/api/                     @pandeymangg @xernobyl @BhagyaAmarasinghe
 /apps/web/modules/core/                    @xernobyl @Dhruwang
+/apps/web/modules/ee/contacts/             @pandeymangg @Dhruwang
 /apps/web/modules/hub/                     @xernobyl @Dhruwang
 /apps/web/modules/mcp/                     @xernobyl @pandeymangg
 /apps/web/modules/organization/            @pandeymangg @Dhruwang
@@ -159,14 +166,13 @@
 /apps/web/modules/survey/                  @Dhruwang @itsjavi @pandeymangg
 /apps/web/modules/ui/                      @itsjavi @Dhruwang
 /apps/web/modules/workspaces/              @pandeymangg @Dhruwang
-/apps/web/modules/ee/contacts/             @pandeymangg @Dhruwang
 
 
 # ------------------------------------------------------------------------------
 #  9. Workflows
 # ------------------------------------------------------------------------------
-/packages/workflows/                       @itsjavi @xernobyl
 /apps/web/modules/ee/workflows/            @itsjavi @xernobyl
+/packages/workflows/                       @itsjavi @xernobyl
 
 
 # ------------------------------------------------------------------------------
@@ -176,12 +182,12 @@
 #  is a release event. v3 is the current surface for new work. `(internal)`
 #  carries no compatibility promise.
 # ------------------------------------------------------------------------------
-/openapi.yml                               @pandeymangg @xernobyl
 /apps/web/app/api/(internal)/              @xernobyl @BhagyaAmarasinghe
 /apps/web/app/api/v1/                      @pandeymangg @Dhruwang
 /apps/web/app/api/v2/                      @pandeymangg @xernobyl
 /apps/web/app/api/v3/                      @xernobyl @BhagyaAmarasinghe @pandeymangg
 /apps/web/modules/api/v2/                  @pandeymangg @xernobyl
+/openapi.yml                               @pandeymangg @xernobyl
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,276 @@
...
+/docker/                                   @BhagyaAmarasinghe @xernobyl
+/apps/web/Dockerfile                       @BhagyaAmarasinghe @xernobyl
+
+
+# ------------------------------------------------------------------------------
+#  12. Documentation
+# ------------------------------------------------------------------------------
+/CONTRIBUTING.md                           @jobenjada @BhagyaAmarasinghe
+/README.md                                 @jobenjada
+/docs/                                     @jobenjada
```
mattinannt: These two lines are the only ones in the file with a single owner. Every other rule (including `/docs/api-v3-reference/` and `/docs/self-hosting/` right below) follows the file's own stated design principle of 2-3 owners "so no path is orphaned by a holiday or a departure." If `@jobenjada` is on leave or leaves the org, `/README.md` and any `/docs/` file not covered by the `api-v3-reference/`/`self-hosting/` overrides resolve to zero designated owners — the exact single-point-of-failure scenario the design says it's preventing everywhere else.

Worth adding a second owner here, or documenting explicitly why docs are the intentional exception.

---

mattinannt (commented):
Reviewed this CODEOWNERS + validator PR. Verified independently: all listed patterns exist against `origin/main`, all owners hold write access (checked via the collaborators API), no duplicate patterns, no split-owner lines across two lines, no `dir/*` misuse, and the validator action's inputs/checks match the real `codeowners-validator` v0.7.4 README. Workflow permissions and harden-runner placement follow this repo's conventions.

One finding worth a look (inline below): `/README.md` and `/docs/` (lines 204-205) are the only two rules in the file with a single owner, which cuts against the file's own "2-3 owners per path" design principle. Not a blocker for an advisory, non-gating file — flagging as a suggestion.

---

.github/workflows/codeowners-validation.yml
```diff
@@ -0,0 +1,55 @@
...
+          # duppatterns  - the same pattern declared twice
+          # files        - patterns that match nothing, i.e. a renamed directory
+          # owners       - owners who do not exist or have lost write access
+          checks: "syntax,duppatterns,files,owners"
+          # avoid-shadowing enforces the general-to-specific ordering the whole
+          # file depends on, since the last matching pattern wins.
+          experimental_checks: "avoid-shadowing"
+          # The `owners` check reads organization membership, which the default
+          # GITHUB_TOKEN cannot do. Needs a PAT with read:org.
+          github_access_token: ${{ secrets.CODEOWNERS_VALIDATOR_TOKEN }}
```
mattinannt, line 70: The `owners` check depends on `secrets.CODEOWNERS_VALIDATOR_TOKEN`, but GitHub does not forward repository secrets to `pull_request`-triggered workflow runs originating from a fork. On this public OSS repo, an external contributor forking and opening a PR that edits `.github/CODEOWNERS` will hit `secrets.CODEOWNERS_VALIDATOR_TOKEN` resolving empty, so the `owners` check errors on missing/invalid auth — unrelated to any real ownership defect. Per the PR's own troubleshooting note, this reads like a "classic vs fine-grained token" or "expired PAT" problem, but no token value could ever have worked from a fork context. Worth gating the `owners` step on `github.event.pull_request.head.repo.full_name == github.repository` (or running it only on `push`/schedule) so fork PRs don't get a spurious failure.

---

.github/CODEOWNERS
```diff
@@ -0,0 +1,276 @@
...
+#  .github/workflows/codeowners-validation.yml
+#
+#  Reference:
+#  https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+# ==============================================================================
+
+
+# ------------------------------------------------------------------------------
+#  1. Repo, build & tooling
+# ------------------------------------------------------------------------------
```
mattinannt, line 64: This section doesn't follow the file's own documented rule 7 ("Within a section, keep rules ordered by depth, then alphabetically"): `/scripts/` (depth 1) sits after `/turbo.json` and the depth-2 `/packages/config-*/` rules instead of alphabetically between `/prisma.config.mjs` and `/sonar-project.properties`. Section 7 drifts further beyond its commented app/lib/modules exception. Since none of these patterns overlap, matching isn't affected today, but a future editor trusting rule 7 when inserting a new same-depth rule will compound the drift rather than notice it — which undermines the scannability the rule exists to guarantee.

---

mattinannt (commented):
Follow-up findings from the same review pass (in addition to the earlier review): two more issues worth a look, both flagged inline.

---

.github/CODEOWNERS
mattinannt: Good catch — fixed in f7b5279.

I verified the claim before acting on it: those were indeed the only two single-owner rules out of 98 (85 had two owners, 11 had three). And the blast radius was larger than the two lines suggest, because `/docs/` covers the entire tree apart from the `api-v3-reference/` and `self-hosting/` overrides — so most of the docs would have resolved to zero owners whenever that one person was unavailable.

Added `@pandeymangg` as second owner on both. He is the highest-volume eligible contributor to `docs/` over the last 18 months (29 commits, ahead of xernobyl 26, Dhruwang 23, BhagyaAmarasinghe 21), so this follows the same specialism-then-history rule as the rest of the file rather than being an arbitrary pick. `@jobenjada` stays primary.

Now every one of the 98 rules carries two or three owners. Re-verified after the change: validator still passes all checks, and the subject overrides still win where they should —

```
README.md                              @jobenjada @pandeymangg
docs/surveys/link-surveys.mdx          @jobenjada @pandeymangg
docs/self-hosting/setup/kubernetes.mdx @BhagyaAmarasinghe @jobenjada
docs/api-v3-reference/openapi.json     @xernobyl @jobenjada
```


---

.github/workflows/codeowners-validation.yml
mattinannt: You're right, and this was a real bug rather than a nit — fixed in b1238a2.

The part I'd got wrong is worse than the failure itself: the troubleshooting note I wrote in the PR body would have actively misled the contributor. It says a failure here means "the token is fine-grained rather than classic", when from a fork context no token value could ever have worked, because the secret is never forwarded in the first place. So the check would have failed, the documented diagnosis would have been wrong, and the contributor had no way to fix it.

The `checks` input is now conditional:

```yaml
checks: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && 'syntax,duppatterns,files,owners' || 'syntax,duppatterns,files' }}
```

Same-repo PRs, the weekly schedule and `workflow_dispatch` keep all four checks; fork PRs drop `owners` and keep the other three. I went with narrowing the check set rather than skipping the step or the workflow, so a fork PR editing CODEOWNERS still gets `syntax`, `duppatterns`, `files` and `avoid-shadowing` — those need no credentials and catch most of what can go wrong in this file.

I also added a step that emits a `::notice::` on fork runs explaining the reduced set, so the coverage gap shows up in the log instead of being silent. `owners` remains covered by the weekly run against the default branch, which is where a departed member would surface anyway.

While there I folded the classic-vs-fine-grained and org-member requirements into the comment above the token, so the next person doesn't reach for a fine-grained PAT — that detail was only in the PR body before.


---

.github/CODEOWNERS
mattinannt: Fixed in b1238a2, and the drift was wider than section 1 and 7. I wrote an audit over every section and labelled group rather than eyeballing it, and found six groups out of compliance:

| Group | Problem |
|---|---|
| 1. Repo, build & tooling | `/scripts/` after `/turbo.json` |
| 7. apps/web | `integration/`, `scripts/`, `instrumentation*` out of order |
| 11. Infrastructure & CI | `/docker-compose.dev.yml` first, ahead of `/.github/` and `/charts/` |
| 12. Documentation | `/README.md` before `/docs/` |
| 14 / Auth & identity | `modules/api/v2/auth/` after `modules/auth/` |
| 14 / Environment & secrets | `env.ts` before `env-client.ts` |

Rather than only fix the two you flagged, I changed the rule and enforced it everywhere, because the depth-then-alphabetical form was the deeper problem: applied strictly it reads *worse* at this size, pushing `/apps/web/app/api/auth/` below `/apps/web/modules/auth/` in the auth group purely on slash count. Rule 7 is now plain alphabetical within each section or labelled group, **with no exceptions**.

That let me delete the exception I'd carved out for section 7's three broad defaults, too. Their position inside the section never mattered — precedence comes from the order of the sections, not from ordering within one — so they now sit in the same alphabetical run as everything else, and the section comment says so instead of claiming they lead.

Since ordering within a section carries no meaning, this should be provably cosmetic, and I checked that rather than asserting it: resolving all 4,566 tracked files against the file before and after the sort produces byte-identical owners.

```
resolution: baseline vs fully-sorted
  IDENTICAL across all 4,566 files
4 check(s) executed, no failure(s)
```

Rule count and owner spread also unchanged — 98 rules, 87 with two owners, 11 with three. The audit script is easy to re-run if you'd ever want it wired into the validation workflow as a repo-specific check; the upstream validator has no opinion on intra-section ordering.


---

xernobyl (approved):
Reviewed the full diff against the current head (b1238a2). The three issues raised earlier (single-owner docs rules, fork-PR token gating, alphabetical ordering drift) are all fixed and verified in place. Independently spot-checked pattern-to-path matches, action pin SHAs, and override precedence across sections — no issues found. Advisory-only nature confirmed: not part of pr.yml's required gate.

PR #8853, opened 2026-08-13
chore(ci): run pnpm typecheck on every PR [ENG-2261]

pandeymangg:
## What & why

`pnpm typecheck` was invoked by no GitHub Actions workflow. The only type checking in CI was a side effect of building the web app, which covers only files in the Next build graph — `packages/*` were never type-checked, `apps/web/tsconfig.typecheck.json` (tests/scripts coverage) was unenforced, and a type error in a package file could merge green.

This PR makes `pnpm typecheck` a required PR check:

- **New `.github/workflows/typecheck.yml`** (called from `pr.yml`'s fan-out, added to the `required` summary gate). Same setup as `lint.yml`; the `pnpm dev:setup` step matters because apps/web's typegen and `@formbricks/database`'s build read the root `.env`.
- **`packages/jobs` gets a `typecheck` script** — it was the only TS package without one, so the turbo run silently skipped it.
- **Three `turbo.json` graph fixes** the first cold run surfaced (all latent on `main`, only visible once the full typecheck graph runs in one invocation):
  1. `database#build` now depends on `database#generate` — both write `generated/prisma` (build's script embeds a clean+generate), and they raced when both were in the same graph, producing flaky `TS2307`/stale-type failures.
  2. `database#typecheck` now runs after `database#build` instead of `#generate` — the reader follows the last writer (same pattern `database#lint` already uses).
  3. `web#typecheck` now depends on `jobs#build` — apps/web imports `@formbricks/jobs` from its dist; `web#test`/`web#dev` had this edge, typecheck didn't, so a cold run failed deterministically with 34 errors.

**Recorded decision on test files:** `**/*.test.ts` stays excluded from `apps/web/tsconfig.typecheck.json` in this PR (the ratchet option). ENG-2265, which this ticket blocks, owns including test files. The baseline backlog is zero — `pnpm typecheck` passes with 0 pre-existing errors, so the gate lands green.

## Linear ticket

Fixes ENG-2261

## How this was tested

- Cold-cache run (wiped turbo cache, all workspace `dist/`, `generated/prisma`): `pnpm typecheck` → **27/27 tasks, exit 0, ~37s** — validated twice (the `turbo.json` edit invalidated all hashes, so the second run was also fully cold).
- Before fix 3 above, the same cold run failed deterministically (`Cannot find module '@formbricks/jobs'` + 34 cascading errors in `@formbricks/web#typecheck`) — confirming the missing edge was real.
- Deliberate-error probe (per the ticket's done-when): added `const _probe: number = "boom"` to `packages/jobs/src/connection.ts` → `pnpm typecheck` fails with TS2322; reverted → green. This exercises the exact command CI runs, in the newly covered package.
- `prettier --check` passes on all four changed files.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open any PR against `main` → the checks list shows **Typecheck** alongside Lint/Unit Tests, and **PR Check Summary** waits for it.
- [ ] On a draft PR, push a commit introducing a type error in any `packages/*` file (e.g. assign a string to a `number` in `packages/jobs/src/connection.ts`) → the Typecheck check fails and PR Check Summary fails; revert the commit → both go green.

**Preconditions / test data**

- A branch/PR against `main`; no special accounts or env needed.

**Risks & regressions**

- PRs with pre-existing type errors in files that were never checked before (package sources, apps/web scripts) will now be blocked until fixed. Baseline on `main` is 0 errors, so only newly introduced errors should fail.
- Build-order changes in `turbo.json` (`database#generate` → `database#build` serialization) also apply to `pnpm build`/`pnpm go` locally — verified both still work.

- none


turbo.json
```diff
@@ -46,15 +46,19 @@
...
+    "@formbricks/database#generate": {
+      "dependsOn": ["^generate"],
+      "env": ["DATABASE_URL"]
+    },
     "@formbricks/database#lint": {
       "dependsOn": ["@formbricks/logger#build", "@formbricks/database#build"]
     },
     "@formbricks/database#typecheck": {
-      "dependsOn": ["@formbricks/logger#build", "@formbricks/database#generate"]
+      "dependsOn": ["@formbricks/logger#build", "@formbricks/database#build"]
```
mattinannt, line 62: **Efficiency/altitude:** `@formbricks/database#typecheck` was switched from depending on `@formbricks/database#generate` to depending on `@formbricks/database#build`. `build`'s script is `pnpm generate && vite build && tsc -p tsconfig.generated-prisma.json`, but `typecheck`'s own script is just `tsc --noEmit` over `src/**`, `types/**`, `zod/**`, `migration/**`, which only needs the generated Prisma client (produced by `generate`), not the bundled `dist/` output. This PR introduces `@formbricks/database#generate` as a properly-scoped, lightweight task (with the `DATABASE_URL` env fix) specifically to avoid needing a full build — but then routes the closest consumer (database's own typecheck) through the heavier `build` path anyway, adding a full vite bundle + extra `tsc` pass to every `pnpm typecheck` run, which now gates every PR via the new CI workflow.

Suggest keeping `"@formbricks/database#typecheck": { "dependsOn": ["@formbricks/logger#build", "@formbricks/database#generate"] }` as it was, now that `@formbricks/database#generate` correctly declares `DATABASE_URL`.

---

turbo.json
```diff
@@ -46,15 +46,19 @@
...
     "@formbricks/database#build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "@formbricks/database#generate"],
       "env": ["DATABASE_URL"],
       "outputs": ["dist/**", "generated/prisma/**"]
     },
+    "@formbricks/database#generate": {
+      "dependsOn": ["^generate"],
+      "env": ["DATABASE_URL"]
+    },
```
mattinannt, line 57: **Correctness/config gap:** the new `@formbricks/database#generate` task declares no `outputs`, unlike `@formbricks/database#build` which explicitly caches the same `generated/prisma/**` path this task produces (confirmed via `turbo run typecheck --dry-run=json`: this task's `outputs` field is `null`, while `build`'s is `["dist/**", "generated/prisma/**"]`).

Because `generated/prisma/**` isn't declared as an output, a turbo cache *hit* on this task restores nothing — it just replays the cached log and reports success without writing the Prisma client to disk. Per the same dry-run, this task is now a direct dependency of typecheck for essentially every package in the monorepo (`config-prettier#typecheck`, `email#typecheck`, `js-core#typecheck`, `logger#typecheck`, etc., via the generic root `typecheck` rule). If a hash-matching cache entry exists (e.g. a warm local `.turbo`/`node_modules/.cache/turbo`, which this repo's own AGENTS.md already flags as a source of stale-output pitfalls for `packages/surveys`) but `packages/database/generated/prisma` doesn't currently exist on disk (fresh clone/worktree, `git clean`, branch switch), the generate step will silently no-op and every dependent typecheck will fail with "cannot find module" errors for the generated Prisma types.

Suggest adding `"outputs": ["generated/prisma/**"]` to this task, matching `@formbricks/database#build`.

---

mattinannt (commented):
Automated single-pass code review (high effort, Agent tool unavailable — no multi-agent fan-out or subagent verify pass ran). 2 findings posted inline, both about the new `turbo.json` task graph for `@formbricks/database#generate`/`#typecheck`.

---

turbo.json
pandeymangg: Good catch — I hit this exact failure while testing locally (cache hit replayed logs, restored nothing, `database#typecheck` failed with `Cannot find module '../generated/prisma/browser'`). Added `"outputs": ["generated/prisma/**"]` in f722431e7 and verified the scenario directly: populated the cache, deleted `generated/`, re-ran — the cache hit now restores the client and all 27 tasks pass.

turbo.json
pandeymangg: Deliberate — reverting this reintroduces a race I hit while validating. `database#build`'s script embeds `pnpm generate`, whose first step deletes `generated/prisma`. With typecheck depending only on `generate`, typecheck and build become ready at the same time and run concurrently — build's embedded regenerate wipes `generated/prisma` while typecheck's tsc reads it (saw both TS2307 and stale-member errors locally). The reader has to follow the last writer, which is `build` whenever it's in the graph — and `web#typecheck` pulls it into every full run, so this edge adds no wall-clock to the CI gate; it only pulls `build` into narrow `--filter=@formbricks/database` runs. Same pattern `database#lint` already uses.

Truly decoupling them would mean dropping the embedded generate from build's script, but `db:migrate:dev` calls `pnpm build` directly (no turbo) and relies on it — out of scope for this PR.

.github/workflows/pr.yml
```diff
@@ -28,6 +28,11 @@ jobs:
     uses: ./.github/workflows/lint.yml
     secrets: inherit
 
+  typecheck:
+    name: Run Typecheck
+    uses: ./.github/workflows/typecheck.yml
+    secrets: inherit
```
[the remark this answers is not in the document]
pandeymangg: Removed in 3fa4f5439 — the workflow declares no secrets, so `inherit` was only there for symmetry with the sibling jobs. (Practical exposure was limited anyway: this runs on `pull_request`, where fork runs get no secrets at all.)

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```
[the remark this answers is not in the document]
pandeymangg: Done in 3fa4f5439, both spots — this workflow's own checkout and `dangerous-git-checkout`. Before touching the shared action I checked all five consumers (lint, typecheck, test, integration-tests, e2e): none run authenticated git operations after checkout, so nothing depends on the persisted token.

Dhruwang (approved):

PR #8854, opened 2026-08-13
fix(integrations): read Google Sheets integration server-side when linking a sheet

Dhruwang:
## What & why

"Link Sheet" in the Google Sheets **Add Integration** modal fails for everyone on `5.3.2` with a raw `No refresh token is set.` toast.

`google-sheets/page.tsx` blanks `access_token` / `refresh_token` before the integration crosses to the client (ENG-2078's redaction fix). `AddIntegrationModal` then passed that same object straight back into `getSpreadsheetNameByIdAction`, and `ZGoogleCredential` types both tokens as `z.string()` — so `""` is schema-valid and `authorize()` reached `refreshAccessToken()` with an empty refresh token. googleapis threw, and the message went to the user verbatim.

Two changes:

- **`getSpreadsheetNameByIdAction` now reads the stored integration** via `getIntegrationByType(workspaceId, "googleSheets")` instead of accepting one from the request, so it always has the real tokens. The `googleSheetIntegration` input is gone from the schema and from the modal's call. This also removes the explicit cross-workspace guard the action needed while the integration was client-supplied (ENG-1921) — the integration can now only come from the authorized workspace — and the manual `createdAt` date coercion, since `getIntegrationByType` already returns `Date`s.
- **`authorize()` throws the `invalid_grant` `AuthenticationError` when no refresh token is stored**, instead of letting googleapis surface a bare message. The modal already maps that error to a proper "reconnect your account" toast, which is the only real way out of that state.

The write path (`createOrUpdateIntegrationAction`) is unaffected — it was already guarded by `withStoredIntegrationKey`; this PR wires the equivalent protection into the read path it was missing from.

## Linear ticket

Fixes ENG-2303

## How this was tested

- `pnpm vitest run lib/googleSheet/ app/(app)/.../google-sheets/` ✅ — 4 files, 12 tests passing.
- **Tests added/updated:**
  - `apps/web/lib/googleSheet/service.test.ts` (new) — `getSpreadsheetNameById` with a redacted (blank) `config.key` rejects with `AuthenticationError("invalid_grant")`. Hermetic: the guard short-circuits before any Google call.
  - `apps/web/app/(app)/.../google-sheets/actions.test.ts` — rewritten for the new signature: asserts the action reads the stored integration and forwards *that* object to `getSpreadsheetNameById`, plus the no-integration case. The old ENG-1921 client-input tests are dropped because the client input they guarded no longer exists.
- **Revert check:** with both production files stashed, all 3 tests go red, and the service test fails with the exact Sentry message — `expected AuthenticationError: invalid_grant, got "No refresh token is set."` So they fail for the right reason, not incidentally.
- `tsc --noEmit` on `apps/web` ✅, `eslint` on the 5 changed files ✅, `prettier` ✅.

No UI change — same modal, same controls; only the error text in the failure case changes (raw googleapis string → existing translated toast).

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Connect Google Sheets: *Workspace → Settings → Integrations → Google Sheets → Connect*, complete the Google OAuth consent → returns to the integration page showing the connected account email.
- [ ] Reload the integration page (this is the case that used to fail — after a reload the page-supplied integration is redacted). Click **Link New Sheet**, paste a Google Sheets URL you have access to, pick a survey, select at least one question → **Link Sheet** succeeds, the modal closes, a success toast appears, and the new link is listed with the sheet's real title.
- [ ] Submit a response to that survey → the response row appears in the linked spreadsheet.
- [ ] Edit an existing link (click a row in the list, change the selected questions, save) → saves successfully and the integration stays connected after a reload.
- [ ] Delete a link from the same modal → row disappears, integration remains connected.
- [ ] Edge — sheet you don't have access to: paste a Google Sheets URL owned by another account → toast reads the spreadsheet-permission message, not a raw Google error.
- [ ] Edge — invalid URL: paste a non-Sheets URL → "enter a valid spreadsheet URL" validation error, no request made.
- [ ] Edge — revoked access: revoke Formbricks' access at https://myaccount.google.com/permissions, then try to link a sheet → toast reads the "token expired, please reconnect" message (translated), **not** `No refresh token is set.`. Reconnecting the integration restores linking.
- [ ] Edge — permissions: as a workspace member with **read-only** access, open the Google Sheets integration page → redirected away from the page (no access to the modal).

**Preconditions / test data**

- A Google account with at least one spreadsheet, plus a second spreadsheet the test account cannot access (for the permission case).
- `GOOGLE_SHEETS_CLIENT_ID`, `GOOGLE_SHEETS_CLIENT_SECRET`, `GOOGLE_SHEETS_REDIRECT_URL` set — without them the integration page renders as disabled.
- Org owner/manager, or a workspace team member with `readWrite`; plus one read-only member for the permissions case.
- At least one survey with responses in the workspace.
- Works the same on Cloud and self-hosted; no feature flag.

**Risks & regressions**

- Sheet linking is the changed path — re-check link / edit / delete of a mapping, and that response data still writes to the sheet afterwards.
- The action no longer trusts a client-supplied integration, so verify it still resolves correctly for a workspace with **multiple** integrations connected (Airtable, Notion, Slack alongside Google Sheets) — it must pick the Google Sheets one.
- The new `authorize()` guard sits on the shared auth path used by the response-write path too: confirm a healthy integration with a valid token still writes responses, i.e. the guard doesn't fire on connected accounts.
- Error-toast wording changed only in the missing-token case; confirm the invalid-grant and insufficient-permission toasts still show their own distinct messages.

- none


apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/google-sheets/actions.ts
```diff
@@ -70,18 +66,16 @@ export const getSpreadsheetNameByIdAction = authenticatedActionClient
...
-      throw new OperationNotAllowedError("Integration does not belong to the specified workspace");
+    // The integration is read from the database rather than accepted from the client. The settings page
+    // redacts `config.key` before handing the integration to the client (ENG-2078), so a client-supplied
+    // object arrives with blank OAuth tokens and `authorize()` fails with "No refresh token is set."
+    // (ENG-2303). Reading it server-side also removes the cross-workspace hijack this action had to
+    // guard against explicitly (ENG-1921), since the integration can only come from the authorized
+    // workspace.
+    const integration = await getIntegrationByType(parsedInput.workspaceId, "googleSheets");
+    if (!integration) {
+      throw new ResourceNotFoundError("Integration", "googleSheets");
```
Comment (mattinannt) [inline apps/web/app/(app)/workspaces/[workspaceId]/settings/workspace/integrations/google-sheets/actions.ts]: Nit: `ResourceNotFoundError`'s second argument is documented/used elsewhere as the entity's ID (see `apps/web/lib/utils/helper.ts` / `services.ts` call sites), but here it's the integration *type* (`"googleSheets"`). The constructor builds `${resource} with ID ${id} not found`, so this renders as "Integration with ID googleSheets not found" — readable enough for a log, but slightly misleading if it ever surfaces raw (e.g. via `getFormattedErrorMessage`'s fallback branch in `AddIntegrationModal.tsx`). Not blocking, just flagging for consistency with other call sites.

---

mattinannt (commented):
Reviewed the fix for ENG-2303. The core change is sound: reading the Google Sheets integration server-side (`getIntegrationByType`) instead of trusting the client-supplied object correctly closes the "No refresh token is set." failure and, as a side effect, subsumes the ENG-1921 cross-workspace guard since the integration can now only come from the authorized workspace. Traced `authorize()`, `redactIntegrationCredentials`, `transformIntegration`'s existing `Date` coercion, and the new/updated tests — didn't find anything that regresses. One minor non-blocking nit left inline on the `ResourceNotFoundError` call.

---

pandeymangg (approved):

PR #8017, opened 2026-05-15
feat: Formbricks 5

Dhruwang:
## What does this PR do?

Introduces **Formbricks 5** 🎉

A major release that flattens the platform hierarchy, ships the new **Unify Feedback** experience, adds an analytics layer powered by Cube.js, introduces background job processing, and externalises auth/rate-limiting to a gateway.

---

### 🏗️ Architecture

- **Environments are gone.** The `Organization → Workspace → Environment → Resources` hierarchy collapses to `Organization → Workspace → Resources`. Surveys, contacts, action classes, webhooks, segments, tags, integrations, and API keys now live directly under a workspace. (12-phase migration; see [`docs/deprecate-environments-plan.md`](docs/deprecate-environments-plan.md))
- **Project → Workspace rename** across the entire codebase, routes, and translations (#7620). Legacy `/environments/[envId]` and `/projects/[projectId]` URLs redirect via a `legacyEnvironmentId` lookup on Workspace so existing links and SDK installations keep working.
- **Unified settings UI** under `/workspaces/[id]/settings` with a shared sidebar covering account, workspace, organization, and integrations (#7904).
- **All Vercel deployment references removed** in favour of Docker + Helm as the supported self-host paths (#7997).

### 🧠 Unify Feedback (new product surface)

- **Feedback Record Directories (FRDs)** — first-class object for organising feedback across sources, with API-key auth, license gating, and workspace-level wiring (#7592, #7804, #7910, #7924).
- **CSV connector** with a revamped mapping UI for importing external feedback (#7951, #7825).
- **Similar feedback preview** powered by embeddings (#7917).
- **Delete feedback records** from the directory (#7991, ENG-938).
- **Create Survey CTA** when a directory has no sources yet (#7943).
- Beta badge in the navigation (#7968).

### 📊 Analytics, Charts & Dashboards

- **Cube.js is now mandatory** for the XM suite in v5 (#7913), with an internal Cube Helm deployment (#7955).
- **CSAT & CES measures** augment the Cube schema with grouping fixes (#7985, ENG-804).
- **AI-powered chart generation** gated behind a full set of AI checks (#7937).
- Dashboard polish: chart removal flow, add-charts dialog focus ring, pie tooltip spacing, empty-name validation, time-range crash and column-mismatch fixes (#7981, #7970, #7973, #8002).

### 🤖 AI

- **AI-powered survey translation** via BullMQ background jobs (#7793).
- **Google Vertex AI ADC credentials** supported alongside the existing providers (#7938).

[...]

PR #8855, opened 2026-08-13
fix(api): restore legacy environmentId on API v1 surveys and webhooks

Dhruwang:
## What & why

The Zapier "Response Finished" trigger cannot be activated on Formbricks Cloud. Survey options in the trigger dropdown render with an `-undefined` suffix, and activating the Zap fails with `400 workspaceId must be provided`.

Formbricks 5 (#8017) renamed environments to workspaces. The API v1 **request** side kept backwards compatibility — `resolveBodyIds` still accepts `environmentId` as an alias for `workspaceId` — but the **response** side did not: `selectSurvey` swapped `environmentId` for `workspaceId`, and webhook responses did the same. So a v1 client can still send an environment id, but can no longer obtain one.

That single gap explains all three reported symptoms:

- **`-undefined` suffix** — Zapier builds each survey dropdown option from the survey id plus its environment id, which `GET /api/v1/management/surveys` no longer returns.
- **`workspaceId must be provided`** — the subscribe call posts an undefined id, `JSON.stringify` drops the key, and the request is rejected by `resolveBodyIds`. That error string exists only in the v1 resolver, which confirms Zapier is on v1.
- **Trigger test succeeds, activation fails** — auth goes through `GET /api/v1/management/me`, which *did* keep its compat mapping (`id: legacyEnvironmentId ?? workspace.id`). Only the subscribe step breaks.

New helper `apps/web/app/lib/api/legacy-environment-id.ts` emits `environmentId` again on v1 surveys and webhooks as `legacyEnvironmentId ?? workspaceId` — the same value `/api/v1/management/me` already returns, and the one `resolveBodyIds` accepts back on writes, so a client can round-trip it without knowing about workspaces. Workspaces created after the migration have no legacy id, so their own id serves as the v1 environment id.

This mirrors the compat layer already in place for `projectOverwrites`, `/api/v2/me`, and the client environment route. The canonical `workspaceId` is unchanged and remains the field to build against; `environmentId` is a legacy alias for existing integrations. It costs one batched workspace lookup per v1 request.

Applied to v1 surveys (list, get, create, update, delete) and v1 webhooks (list, get, create, delete).

On the two DELETE handlers the destructive call owns its own `try`/`catch` and the enrichment runs after it, through `addLegacyEnvironmentIdBestEffort`. The row is already gone by then, so a failed workspace lookup degrades to the un-enriched entity rather than reporting a failed delete — otherwise the webhook route would flip its audit log to `"failure"` and return a false 404, prompting a caller to retry a delete that had already succeeded. Read paths keep the throwing variant on purpose: a fallback `environmentId` there would be stored by the client and posted back on the next write, where it would fail workspace resolution.

Two things deliberately left out of scope:

- The v1 OpenAPI examples still show `workspaceId` only. `environmentId` is a deprecated alias and documenting it would advertise it to new integrations.
- v1 action-classes and contact-attribute-keys lost the same field in the same rename. They are unrelated to this ticket and are tracked separately.

## Linear ticket

Fixes ENG-2270

## How this was tested

- `pnpm test` on the touched areas — 24 files / 205 tests pass (`app/api/v1` plus the helper), including 9 unit tests for the helper in `apps/web/app/lib/api/legacy-environment-id.test.ts` ✅
- `tsc --noEmit` (apps/web typecheck config) and `eslint` on all touched files — clean ✅
- New E2E spec `apps/web/playwright/api/management/legacy-environment-id.spec.ts` walks the exact Zapier subscribe path against a running stack: create survey → list surveys → read `environmentId` → create a v1 webhook posting **only** `environmentId` → verify list/detail/delete all echo it. Passes ✅, and re-run green after the DELETE restructure.
- Red/green check: with the four route files stashed and the same stack running, that spec fails at `expect(responseBody.data.environmentId).toBeTruthy()` → `Received: undefined`. Restoring the routes turns it green again, so the new test genuinely covers the regression.
- The best-effort DELETE path is unit-tested in both directions: enriched on a successful lookup, un-enriched (no thrown error) when the lookup fails, with the read-path variant asserted to still propagate the failure.

No UI changes.

## Breaking changes

None — this restores a response field that API v1 used to return.

## QA / Test Plan

**How to test**

- [ ] With a workspace-scoped API key, `GET /api/v1/management/surveys` → 200, and every survey in `data` has both `workspaceId` and a non-empty `environmentId` string.
- [ ] `GET /api/v1/management/surveys/{surveyId}` → 200 and the survey carries `environmentId`.
- [ ] `POST /api/v1/webhooks` with a body containing `environmentId` (the value from the step above) instead of `workspaceId`, plus `url`, `source: "zapier"`, `triggers: ["responseFinished"]` → 201/200, and the returned webhook has both `workspaceId` and the same `environmentId`. Before this fix, this returned `400 workspaceId must be provided` when the client had no way to read the id.
- [ ] `GET /api/v1/webhooks` and `GET /api/v1/webhooks/{webhookId}` → the webhook carries `environmentId`.
- [ ] `DELETE /api/v1/webhooks/{webhookId}` → 200 and the deleted webhook body carries `environmentId`.
- [ ] End-to-end on Zapier: create a Zap with the Formbricks "Response Finished" trigger, connect a Cloud account, open the survey dropdown → survey options no longer show an `-undefined` suffix. Test the trigger, then **publish/activate** the Zap → activation succeeds with no 400.
- [ ] Submit a response to the selected survey → the Zap fires with the response payload.
- [ ] Sanity check that `POST /api/v1/webhooks` with `workspaceId` (no `environmentId`) still works — the canonical path must be unaffected.
- [ ] Negative case: `POST /api/v1/webhooks` with neither `workspaceId` nor `environmentId` → still `400 workspaceId must be provided`.

**Preconditions / test data**

- A Formbricks Cloud (or self-hosted) account with a workspace-scoped API key that has write permission.
- At least one survey in that workspace.
- For the Zapier steps: a Zapier account and the published Formbricks Zapier app.
- Worth covering both workspace types if available: one migrated from a pre-Formbricks-5 environment (it has a `legacyEnvironmentId`, so `environmentId` differs from `workspaceId`) and one created after the migration (`environmentId` equals `workspaceId`). Both must work.

**Risks & regressions**

- v1 survey and webhook responses now carry one additional field. Clients that assert on an exact response shape would see the extra key, though v1 clients historically expected it.
- The added workspace lookup runs per v1 surveys/webhooks request — re-check that large survey lists still respond normally.
- Verify the v2 management API (`/api/v2/management/webhooks`) is untouched and still returns `workspaceId` only.
- Verify webhook delivery itself is unchanged: the outbound "Response Finished" payload was not modified by this PR.
- Both DELETE handlers were restructured so the delete owns its own error handling. Re-check that deleting a non-existent webhook still returns 404 with a `"failure"` audit entry, and that a normal delete still returns 200 with the deleted object.

- none


apps/web/app/api/v1/webhooks/[webhookId]/route.ts
```diff
@@ -66,7 +67,7 @@ export const DELETE = withV1ApiWrapper({
     try {
       const deletedWebhook = await deleteWebhook(params.webhookId);
       return {
-        response: responses.successResponse(deletedWebhook),
+        response: responses.successResponse(await addLegacyEnvironmentId(deletedWebhook)),
```
Comment (mattinannt) [inline apps/web/app/api/v1/webhooks/[webhookId]/route.ts]: `await addLegacyEnvironmentId(deletedWebhook)` now runs inside the same `try` that wraps `deleteWebhook`. If the delete succeeds but this enrichment call throws afterward (it does its own `prisma.workspace.findMany` internally, e.g. on a transient DB error), the `catch` below treats it exactly like a failed delete: it sets `auditLog.status = "failure"` and returns a `404 Webhook not found`, even though the webhook was actually deleted. That's a false negative for the caller (who may retry a delete that already succeeded) and a false failure in the audit trail.

Consider moving the `deleteWebhook` call and its `try/catch` above the `addLegacyEnvironmentId` call, so an enrichment failure can't be mistaken for a delete failure — e.g.:

```js
let deletedWebhook;
try {
  deletedWebhook = await deleteWebhook(params.webhookId);
} catch (e) {
  if (auditLog) {
    auditLog.status = "failure";
  }
  logger.error({ error: e, url: req.url }, "Error deleting webhook");
  return {
    response: responses.notFoundResponse("Webhook", params.webhookId),
  };
}
return {
  response: responses.successResponse(await addLegacyEnvironmentId(deletedWebhook)),
};
```

---

apps/web/app/api/v1/management/surveys/[surveyId]/route.ts
```diff
@@ -107,7 +110,7 @@ export const DELETE = withV1ApiWrapper({
 
       const deletedSurvey = await deleteSurvey(params.surveyId);
       return {
-        response: responses.successResponse(deletedSurvey),
+        response: responses.successResponse(await addLegacyEnvironmentId(deletedSurvey)),
```
Comment (mattinannt) [inline apps/web/app/api/v1/management/surveys/[surveyId]/route.ts]: Same shape of issue as the webhook DELETE route: `deleteSurvey` has already committed by the time `addLegacyEnvironmentId(deletedSurvey)` runs, but both are inside the same `try`. If the enrichment call throws, the surrounding `catch (error) { return handleErrorResponse(error); }` returns a generic error response even though the survey was in fact deleted — the client has no way to tell the destructive operation already succeeded. Lower severity than the webhook case (no false 404 / no audit misreport here), but same underlying fix applies: fetch/enrich before returning, or fall back to returning the un-enriched `deletedSurvey` if enrichment fails, rather than letting a post-delete failure mask a successful delete.

---

mattinannt (commented):
The `environmentId ?? workspaceId` backwards-compat approach and the deduped `findMany` batching in `addLegacyEnvironmentIdToList` look solid, and the new unit + Playwright coverage exercises the actual Zapier round-trip this fixes.

Left two comments on a shared pattern: in both the webhook and survey `DELETE` handlers, the new `addLegacyEnvironmentId(...)` call was added *inside* the `try` that wraps the delete itself, so a post-delete enrichment failure can be mistaken for a failed delete (worse on the webhook route, where it also flips the audit log to `"failure"` and returns a false 404). Not a blocker for the core fix, but worth tightening before merge.

---

apps/web/app/api/v1/webhooks/[webhookId]/route.ts
Comment (Dhruwang) [reply apps/web/app/api/v1/webhooks/[webhookId]/route.ts]: Good catch — fixed in 206d458207.

You're right that the webhook route was the worse of the two: a transient failure in the workspace lookup, after the row was already gone, would set `auditLog.status = "failure"` and return a 404, so a caller would retry a delete that had already succeeded.

Took your restructure, plus one addition. `deleteWebhook` now owns its own `try/catch` and the enrichment happens after it. But moving the call outside that block alone would still turn a lookup failure into a 500 for an already-deleted webhook, so the post-delete call goes through a new `addLegacyEnvironmentIdBestEffort`, which logs and degrades to the un-enriched entity. An irreversible write now always reports the success it was.

Read paths deliberately keep the throwing variant: returning a fallback `environmentId` there would hand the client a value it stores and posts back on the next write, which would then 404 on workspace resolution. Failing loudly is the right behavior for those. Both branches are covered by unit tests.

apps/web/app/api/v1/management/surveys/[surveyId]/route.ts
Comment (Dhruwang) [reply apps/web/app/api/v1/management/surveys/[surveyId]/route.ts]: Agreed, and fixed in 206d458207 alongside the webhook route.

`deleteSurvey` now runs in its own `try/catch` and the enrichment happens after it, through `addLegacyEnvironmentIdBestEffort` — which is the fallback option you suggested: it logs and returns the un-enriched `deletedSurvey` rather than letting a post-delete lookup failure surface as an error for a survey that is genuinely gone.

Worth noting these two DELETE enrichments were not needed for the Zapier fix at all — I added them so a v1 consumer sees `environmentId` consistently across the resource, since pre-Formbricks-5 DELETE responses carried it too. Keeping them, now that they can't mask the write.

apps/web/app/api/v1/management/surveys/route.ts
```diff
@@ -147,7 +150,9 @@ export const POST = withV1ApiWrapper({
 
       return {
         response: responses.successResponse(
-          addLegacyProjectOverwrites(resolveStorageUrlsInObject(withDerivedQuestions(survey)))
+          await addLegacyEnvironmentId(
```
xernobyl: Same shape of issue as the webhook create fix in `206d458`/`16f73ce2`: `createSurvey` (line 145) has already committed by the time `addLegacyEnvironmentId` runs, but both are inside the same `try`. If the workspace lookup throws (e.g. a transient DB error), the `catch` below returns an error response for a survey that was in fact created — and since `Survey` has no natural dedup key, a client retrying on that false error would create a duplicate.

Suggest swapping to `addLegacyEnvironmentIdBestEffort` here, consistent with the other post-commit call sites.

apps/web/app/api/v1/management/surveys/[surveyId]/route.ts
```diff
@@ -219,7 +230,9 @@ export const PUT = withV1ApiWrapper({
 
         return {
           response: responses.successResponse(
-            addLegacyProjectOverwrites(resolveStorageUrlsInObject(withDerivedQuestions(updatedSurvey)))
+            await addLegacyEnvironmentId(
```
Comment (xernobyl) [inline apps/web/app/api/v1/management/surveys/[surveyId]/route.ts]: Same pattern here: `updateSurvey` (line 226) has already committed, but `addLegacyEnvironmentId` runs inside the same `try`/`catch`. A lookup failure would report an error for an update that succeeded (and leave the audit log as `"failure"`). Less severe than the create case since a retry is idempotent-ish, but still a false negative.

Suggest `addLegacyEnvironmentIdBestEffort` here too, for consistency with the create/delete paths.

xernobyl (commented):
One more spot with the same shape of bug the last two rounds fixed for DELETE and webhook-create: survey `POST` and `PUT` also enrich a committed write with the throwing `addLegacyEnvironmentId`, inside the same `try` as the write itself.

xernobyl (approved):
Approving — verified end-to-end against a live stack (isolated DB/redis, seeded a migrated workspace with a `legacyEnvironmentId` and a post-migration one, plus a second org). Survey/webhook create, get, list, and delete all round-trip `environmentId` correctly; the actual Zapier bug (`POST /api/v1/webhooks` with only `environmentId`) now succeeds instead of `400 workspaceId must be provided`; the canonical `workspaceId`-only path and the missing-both-ids 400 both still hold. Cross-org checks (including trying another org's `legacyEnvironmentId` as the alias) correctly 401 — no BOLA from the new lookup. Unit tests, typecheck, and lint are clean locally; CI is green.

One minor, non-blocking thing I don't think the bots flagged: all four GET handlers (`apps/web/app/api/v1/management/surveys/route.ts:54`, `apps/web/app/api/v1/management/surveys/[surveyId]/route.ts:72`, `apps/web/app/api/v1/webhooks/route.ts:27`, `apps/web/app/api/v1/webhooks/[webhookId]/route.ts:31`) use the throwing `addLegacyEnvironmentId`/`addLegacyEnvironmentIdToList`, so a transient workspace-lookup failure now 500s a read that would otherwise have succeeded — a failure mode these endpoints didn't have before. The strict-on-reads reasoning (never hand back a wrong id a client could post back on a write) makes sense for writes, but on a read `environmentId` is never "wrong" here — the lookup either resolves or it doesn't — so best-effort would be just as safe and slightly more resilient. Not asking for a change, just flagging in case it's worth a follow-up.

PR #8857, opened 2026-08-13
feat(embedded-data): repoint readers to the tables and inline definitions on survey load [ENG-1837]

itsjavi:
## What & why

ENG-1833 added the tables, ENG-1978 keeps them current on editor saves, ENG-1836 built the read seam
and ENG-1835 backfilled every existing survey. Nothing read any of it yet. This makes the tables the
read source of truth: survey loads join the linked `EmbeddedData` rows and inline them onto the survey
object as `embeddedFields`, and every reader resolves definitions through that instead of
`survey.variables` / `survey.hiddenFields`.

**Definitions move; values do not.** That distinction is the whole PR. The resolver's own ENG-1837 swap
checklist (`embedded-data-resolver.ts`) names four places where `resolveEmbeddedValue`'s semantics
deliberately differ from today's consumers — a number variable holding a non-numeric string, a response
missing a variable's key, recall rendering uncoerced values, and the two recall paths disagreeing on a
key present in both maps. Adopting it for value lookup would have satisfied "no direct reads" while
silently changing exported data, so **every call site keeps its value expression verbatim**:
`Number(v) || 0`, `v || ""`, `response.variables[storageKey]` and `response.data[storageKey]` are
untouched. `resolveEmbeddedValue` is called from **zero** call sites; `coerceToEmbeddedDataType` from
exactly one — seeding a computed field's initial value, where it provably equals the declared value for
every legacy variable shape (`ZSurveyVariable.value` is always non-null and type-matched, so the coerce
is a pass-through).

### Two accessors, two questions

| Accessor | Answers | Callers |
| --- | --- | --- |
| `getSurveyEmbeddedFields` | *what is stored* — rows win, cards as fallback | export, response filters, response table, summary, single-response card, emails, integrations, both logic engines, Notion mapping |
| `getDeclaredEmbeddedFields` | *what the author declares right now* — cards only | recall picker, logic operand picker, calculate widget, follow-up recipients, editor preview, recall labelling |

The editor needs the second because `localSurvey` is cloned once at mount and the rows are only written
on save. An earlier attempt stripped `embeddedFields` from that clone instead — which broke change
detection for **every** survey, since `isDeepEqual` short-circuits on key count, leaving an idle draft
editor auto-saving every ten seconds. `localSurvey` is a plain `structuredClone` again and there is a
guard test pinning that.

Recall *labelling* sits on the declared side too, because the picker writes `@label` into the text and
`getRecallItems` reads it back by matching that label — split them and a field renamed since the last
save leaves a raw `#recall:…#` tag in the input. **This constrains ENG-1853:** the unified picker has to
move recall and the picker onto the tables together, not one at a time.

### Order

Neither table carries an ordinal column, so the legacy JSON columns stay the source of truth for
**order** while the rows are the source of truth for **definitions**. Joined rows are sorted by their
index in `toDesiredEmbeddedFields`, which reproduces today's CSV and picker order exactly. Revisit when
the legacy columns are dropped.

### One write path had to change

`patchV3Survey` wrote `variables` / `hiddenFields` and never reconciled the rows, so the reconciling
writes were three, not four. Dormant while readers used the legacy columns this write *does* update —
but pointing readers at the rows makes it a live regression: after `PATCH /api/v3/surveys/{id}` or MCP
`patch_survey`, a newly added variable would be invisible to the logic engine, export columns, the
response table, the summary and the response filters. The reconcile now runs in the same transaction as
the survey update, derived from the persisted survey rather than the patch document (a patch may omit
either key, and reading the payload would see them as absent and delete every row). The transaction is
unconditional now, where it previously existed only when targeting changed.

Both doc comments enumerate the four reconciling writes by name, so `grep reconcileEmbeddedData(` is the
audit rather than a claim to be trusted.

### Two unrelated fixes, in their own commits

Both were found while working on this and are independent of the read repoint — split out so they can be
dropped or cherry-picked without touching the rest.

- **`fix(playwright)`** — `login`, `signUpAndLogin` and `signupUsingInviteToken` queried
  `getByRole("button", { name: "Login with Email" })`, but the button renders "Log in with Email". Role
  name matching is substring-based, so none of the seven queries could ever match. Dead code today
  (every spec authenticates through the `users` fixture), which is why nobody noticed.
- **`fix(turbo)`** — the `generate` task declared no `outputs`, so turbo cached it with an empty output
  set. On a cache hit it restores nothing *and* skips the task, including the clean step that deletes
  `generated/prisma` — so once the client is missing it stays missing while turbo reports FULL TURBO,
  and every downstream typecheck and build fails on the absent client. Reproduced and fixed against that
  exact case. This bit me mid-PR.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1837/repoint-readers-to-the-tables-inline-definitions-on-survey-load

Milestone 1 of [Embedded Data](https://linear.app/formbricks/project/embedded-data-3d184a43d2f9).
Targets `epic/embedded-data-v1` — all of v1 ships as one release, so this does not reach production
until the epic merges. Ships together with ENG-1838 (SDK/API serialisation back-compat).

## Screenshots

**Identical output is the acceptance criterion here, not a coincidence** — v1 is invisible plumbing, so
a conventional before/after would just be the same image twice. Instead, this is the *same page* served
through the two code paths this PR puts side by side: on the left the survey's `SurveyEmbeddedData`
links are deleted, so definitions resolve through the legacy JSON columns (the fallback); on the right
they resolve through the tables.

| Before — legacy JSON path | After — EmbeddedData tables path |
| --- | --- |
|  |  |

Same survey, same responses, same scroll position; column headers and every cell match, verified
programmatically as well as visually. Note the second row: `utm_source` is `ads` but `plan` is **empty**,
because that response never captured it — a resolver-based implementation would have filled it with the
field's default. That empty cell is the point of the PR.

## How this was tested

**Full suite green, against a clean baseline.** `pnpm test` on the epic base before any change was
25/25 tasks with zero failures, so there is no pre-existing noise to hide behind — every number below is
directly comparable.

| | baseline `244afe6` | this branch |
| --- | --- | --- |
| turbo tasks | 25/25 | 25/25 |
| `@formbricks/web` | 7638 | 7667 |
| `@formbricks/surveys` | 665 | 675 |
| `@formbricks/types` | 203 | 222 |

`pnpm typecheck` 25/25 · `pnpm lint` 20/20, 0 errors (web warnings 43, all pre-existing, verified
against baseline blobs) · `prettier --check` clean on every changed file · `pnpm i18n:validate` passed
(no strings changed).

**Integration — 12 new against real Postgres.** `embedded-data-read.integration.test.ts` (6): the join
round-trips, the raw relation never leaks onto `TSurvey`, declared ordering is preserved, a survey
missing its backfill falls back, the partial-row case is pinned, and reads write nothing.
`embedded-data-v3-patch.integration.test.ts` (6): add / rename / remove a variable or hidden field via
the **real** `patchV3Survey` and assert the rows match the persisted columns, plus response integrity.

**Every claim was checked by sabotage, not assumed.** Each of these was run by breaking the code and
confirming the intended test — and only that test — fails:

| Sabotage | Result |
| --- | --- |
| delete the `reconcileEmbeddedData` call in the v3 patch | 4 of 6 integration tests fail (add / add / rename / remove) |
| make the reconcile migrate stored answers on rename | **only** the AC-4 response-integrity test fails, with the exact corruption it forbids |
| revert recall labelling to the stored accessor | 2 recall tests fail |
| remove the `storageKey` refinements | the blank/whitespace schema test fails |
| strip `embeddedFields` from the editor clone | the change-detection guard fails |

Three tests that *couldn't* fail were caught in review and rebuilt: one compared a pure function with
itself; one used an empty `embeddedFields`, which the stored accessor treats as "no join" and answers
from the declarations anyway; and the AC-4 test seeded no responses, so it could only have caught a
reconcile that *created* a row rather than one that re-keyed existing answers.

**End-to-end against a running app.** A fixture survey was built so the row order and the declared order
disagree in both groups (variables `zscore`→`atext`, hidden fields `utm_source`→`plan`; under
`storageKey ASC` the DB returns `plan, atext, zscore, utm_source`). Exported through the real UI:

```
… ,"1. How satisfied are you with our product?","zscore","atext","utm_source","plan"
1,"…0003", … ,"","abc","","","free"
2,"…0002", … ,"",13,,"ads",""
3,"…0001", … ,"",42,"alpha","newsletter","pro"
```

Declared order, and the three preserved deltas visible in real data: response `…0002` has no `atext`
and no `plan`, so both cells are **empty** rather than showing the declared defaults `"seed"` / a
fallback; response `…0003` holds `"abc"` in the *number* variable and exports it verbatim rather than
coercing it or substituting the declared `7`. Under a resolver-based implementation all three cells
would have changed.

**Playwright — 1 new spec**, `survey-editor-embedded-fields.spec.ts` (`@slow`), covering AC #3, the one
criterion only a browser can prove. It saves a survey, then **poisons the stored row behind the
editor's back** (renames it, flips its `dataType`) so the row and the card actively disagree, and
asserts every editor surface follows the card — recall picker via both the `@` keyboard path and the
toolbar button, logic operand picker, and the calculate widget's input `type`. Then it renames and
retypes the variable live, without reloading, and re-asserts all three. The poisoned row is what makes
the test able to fail: on a fresh survey an empty row list falls back to the cards anyway, so a naïve
implementation would pass.

`npx playwright test survey-editor-embedded-fields.spec.ts --project=chromium` → 1 passed;
`--repeat-each=3 --workers=1` → 3 passed; alongside `survey-editor-rich-text-tabs.spec.ts` → 3 passed.

⚠️ Under 2 parallel workers it flakes on the shared `waitForSurveyEditor` helper. **Pre-existing and
environmental** — it reproduces identically on `survey-editor-rich-text-tabs.spec.ts`, which this PR
does not touch, and is caused by 2 workers plus a `next dev --turbopack` server on 4 vCPUs with HMR
remounts mid-run. CI builds for production and runs `pnpm start`. No timeout was raised to paper over
it.

**Not tested here:** follow-up emails (needs SMTP) and the third-party integrations (need credentials) —
both are covered by unit tests only. Widget bundle grew 952,982 → 954,348 B (+0.14%); zod was already
present, so no new dependency.

## Breaking changes

None.

`embeddedFields` is explicitly stripped from the v1 management responses
(`withoutInternalSurveyProjections`, applied at all four survey-returning sites), because API
serialisation belongs to ENG-1838. The SDK workspace state and the link-survey payload deliberately
carry it — that is part 1 of this ticket. v2 has no survey-body endpoint and v3 builds an explicit
object.

## QA / Test Plan

**How to test**

- [ ] Open a survey with at least two variables and two hidden fields → Responses → Download → CSV →
      the variable and hidden-field columns appear in the order they were declared in the editor cards,
      variables first, exactly as before this change
- [ ] Same survey → Responses table → the same columns in the same order
- [ ] A response that never captured one of those fields → the cell is **empty**, not the field's
      default value
- [ ] A number variable whose stored value is not a number (e.g. set by a calculate action) → exports
      verbatim, not coerced and not replaced by its default
- [ ] Survey editor → add a hidden field in the Hidden Fields card → **without reloading**, type `@` in
      a question headline → the new field is offered in the recall dropdown
- [ ] Same, without reloading → open Logic → the new field appears in the operand picker
- [ ] Rename a variable and change its type Text → Number → without reloading, the calculate action's
      value input becomes numeric and the logic operand picker shows the new name
- [ ] Follow-ups tab → "Send to" dropdown → a hidden field added moments ago (unsaved) is offered
- [ ] Open a draft survey editor and leave it untouched for ~30 seconds → **no** auto-save fires (the
      survey's "last updated" does not advance); pressing Back navigates without a discard-changes
      dialog
- [ ] `PATCH /api/v3/surveys/{surveyId}` adding a variable → that variable then appears as an export
      column and can be used in a logic condition
- [ ] Rename a hidden field via the API on a survey with responses → previously collected responses keep
      their original values under the original key (renaming orphans history, exactly as before)
- [ ] Recall, logic branching and calculations on an existing survey behave exactly as before

**Preconditions / test data**

- A workspace with surveys that have both variables and hidden fields, and responses collected before
  this change — including at least one response predating a variable's creation, so the "missing key
  stays empty" case is exercised.
- Ideally a survey whose variables were reordered after creation, to confirm export header order.
- For the v3 patch case: an API key with survey write access, or the MCP `patch_survey` tool.

**Risks & regressions**

- **Highest:** anything reading Embedded Data now goes through a join. If a survey select were missed,
  that reader falls back to the legacy columns and stays correct — the failure mode is silent staleness,
  not an error. The place to look first is export column order and the response table.
- **v3 PATCH now always runs in a transaction** where targeting-free patches previously ran as a single
  statement. Watch for lock duration or connection-pool pressure under concurrent patching.
- Editor change detection is the subtle one: if the draft auto-save starts firing on an untouched
  editor, or Back always prompts to discard, that is this PR.
- Recall tokens for fields renamed since the last save are the second subtle one: a raw `#recall:…#`
  tag rendering in an input means the picker and the labeller have desynced.

- none. No schema change, no data migration, no env change. Reads only — no `Response` row is read or
  written by anything this PR adds.

### DIFF
--- a/apps/web/lib/embedded-data/survey-fields.ts
+++ b/apps/web/lib/embedded-data/survey-fields.ts
@@ -2,6 +2,7 @@ import "server-only";
 import { Prisma } from "@formbricks/database/prisma";
 import { toDesiredEmbeddedFields } from "@formbricks/types/embedded-data-mapping";
 import { type TLinkedEmbeddedField } from "@formbricks/types/embedded-data-resolver";
+import { type TSurveyVariable } from "@formbricks/types/surveys/types";
 
 /**
  * The join that makes the `EmbeddedData` / `SurveyEmbeddedData` tables the read source of truth
@@ -38,6 +39,48 @@ interface TSurveyWithEmbeddedDataLinks {
 }
 
 /**
+ * `variables` and `hiddenFields` are `Json` columns. Prisma types them from the schema annotation,
+ * but nothing enforces that shape in the database, so a row can hold an object where an array
+ * belongs. `toDesiredEmbeddedFields` maps both with `?? []`, which catches `null` and `undefined`
+ * but not a wrong type — one malformed row would throw `(variables ?? []).map is not a function`
+ * inside `transformPrismaSurvey` and fail the **entire survey read**, not just its ordering.
+ *
+ * Guarded here rather than inside `toDesiredEmbeddedFields`, because that mapper is shared with
+ * `reconcileEmbeddedData` on the WRITE path, where throwing is the correct behaviour: silently
+ * mapping a malformed survey to an empty desired set would delete every row it has. Reads must
+ * degrade, writes must fail loudly — so the leniency belongs at this boundary and nowhere else.
+ * (ENG-1835's `planSurveyBackfill` makes the same call from the other side: it checks these shapes
+ * and skips the survey.)
+ *
+ * The two groups are guarded independently on purpose. A survey whose `variables` are malformed
+ * still ranks its hidden fields correctly; only the malformed group loses its declared order and
+ * sorts last. A blanket `try`/`catch` would lose both — and would also swallow a genuine bug inside
+ * the mapper.
+ */
+
+/**
+ * Not "is a valid variable list" — only that mapping over it cannot throw. Entries that are objects
+ * but not variables yield a garbage `storageKey`, which simply matches no row and therefore ranks
+ * nothing; entries that are `null` would throw on the property read, so they disqualify the list.
+ */
+const toRankableVariables = (value: unknown): TSurveyVariable[] =>
+  Array.isArray(value) && value.every((entry) => typeof entry === "object" && entry !== null)
+    ? (value as TSurveyVariable[])
+    : [];
+
+/** Non-string ids cannot be storage keys, so they are dropped rather than disqualifying the list. */
+const toRankableFieldIds = (value: unknown): string[] =>
+  Array.isArray(value) ? value.filter((id): id is string => typeof id === "string") : [];
+
+/** Reads `hiddenFields.fieldIds` without assuming `hiddenFields` is an object at all. */
+const readFieldIds = (hiddenFields: unknown): string[] =>
+  toRankableFieldIds(
+    typeof hiddenFields === "object" && hiddenFields !== null
+      ? (hiddenFields as { fieldIds?: unknown }).fieldIds
+      : undefined
+  );
+
+/**
  * Reshapes the joined rows into the `{ field, link }` pairs the read seam consumes, or `undefined`
  * when the select omitted the join — which is exactly the input `getSurveyEmbeddedFields`' fallback
  * expects, so a survey read through a narrower select keeps resolving off its legacy columns.
@@ -56,12 +99,13 @@ export const inlineSurveyEmbeddedFields = (
   const links = surveyPrisma.embeddedDataLinks;
   if (!links) return undefined;
 
+  // Still routed through `toDesiredEmbeddedFields` rather than reading the ids here: which key a
+  // field is addressed by (a variable's cuid, a hidden field's name) is that function's rule, and
+  // re-deriving it at this boundary is how the ordering would silently drift from the rows.
   const legacyRankByStorageKey = new Map(
     toDesiredEmbeddedFields({
-      variables: surveyPrisma.variables as Parameters<typeof toDesiredEmbeddedFields>[0]["variables"],
-      hiddenFields: surveyPrisma.hiddenFields as Parameters<
-        typeof toDesiredEmbeddedFields
-      >[0]["hiddenFields"],
+      variables: toRankableVariables(surveyPrisma.variables),
+      hiddenFields: { enabled: false, fieldIds: readFieldIds(surveyPrisma.hiddenFields) },
     }).map((field, index) => [field.storageKey, index] as const)
   );
   const rank = (storageKey: string): number =>
--- a/apps/web/lib/embedded-data/survey-fields.ts
+++ b/apps/web/lib/embedded-data/survey-fields.ts
@@ -1,0 +1,87 @@
+import "server-only";
+import { Prisma } from "@formbricks/database/prisma";
+import { toDesiredEmbeddedFields } from "@formbricks/types/embedded-data-mapping";
+import { type TLinkedEmbeddedField } from "@formbricks/types/embedded-data-resolver";
+
+/**
+ * The join that makes the `EmbeddedData` / `SurveyEmbeddedData` tables the read source of truth
+ * (ENG-1837). Add it to a survey select and pass the row through {@link inlineSurveyEmbeddedFields};
+ * every reader then resolves definitions through `getSurveyEmbeddedFields` instead of reading
+ * `survey.variables` / `survey.hiddenFields`.
+ *
+ * Only the columns the read seam consumes are selected — this shape reaches public survey payloads
+ * (the SDK workspace state and the link-survey page), so the row's ids, ownership and timestamps
+ * stay server-side. It mirrors `SELECT_CURRENT_FIELDS` in reconcile.ts minus exactly those.
+ *
+ * `orderBy` is not optional: without it Postgres row order is unspecified, and this list drives
+ * user-visible column and picker order. It is the tie-break, not the ordering rule — see
+ * {@link inlineSurveyEmbeddedFields}.
+ */
+export const selectSurveyEmbeddedDataLinks = {
+  select: {
+    storageKey: true,
+    embeddedData: {
+      select: { name: true, source: true, dataType: true, defaultValue: true, locked: true },
+    },
+  },
+  orderBy: { storageKey: "asc" },
+} as const satisfies Prisma.SurveySelect["embeddedDataLinks"];
+
+/** The shape {@link selectSurveyEmbeddedDataLinks} produces, as much of it as the mapping needs. */
+interface TSurveyWithEmbeddedDataLinks {
+  variables?: unknown;
+  hiddenFields?: unknown;
+  embeddedDataLinks?: {
+    storageKey: string;
+    embeddedData: TLinkedEmbeddedField["field"];
+  }[];
+}
+
+/**
+ * Reshapes the joined rows into the `{ field, link }` pairs the read seam consumes, or `undefined`
+ * when the select omitted the join — which is exactly the input `getSurveyEmbeddedFields`' fallback
+ * expects, so a survey read through a narrower select keeps resolving off its legacy columns.
+ *
+ * **Ordering.** Neither table carries an ordinal column, so the rows cannot reproduce the order the
+ * author put their variables and hidden fields in — and that order is user-visible: it is the CSV
+ * and XLSX header order and the order of every picker. Until the legacy JSON columns are dropped
+ * they therefore remain the source of truth for *order* while the rows are the source of truth for
+ * *definitions*, which is what this sort expresses. A row whose storage key the columns don't know
+ * (impossible today — `reconcileEmbeddedData` writes exactly the derived set in one plan) sorts last
+ * rather than being dropped, in the select's `storageKey` order.
+ */
+export const inlineSurveyEmbeddedFields = (
+  surveyPrisma: TSurveyWithEmbeddedDataLinks
+): TLinkedEmbeddedField[] | undefined => {
+  const links = surveyPrisma.embeddedDataLinks;
+  if (!links) return undefined;
+
+  const legacyRankByStorageKey = new Map(
+    toDesiredEmbeddedFields({
+      variables: surveyPrisma.variables as Parameters<typeof toDesiredEmbeddedFields>[0]["variables"],
+      hiddenFields: surveyPrisma.hiddenFields as Parameters<
+        typeof toDesiredEmbeddedFields
+      >[0]["hiddenFields"],
+    }).map((field, index) => [field.storageKey, index] as const)
+  );
+  const rank = (storageKey: string): number =>
+    legacyRankByStorageKey.get(storageKey) ?? Number.MAX_SAFE_INTEGER;
+
+  // Array.prototype.sort is stable, so rows the columns don't rank keep the select's storageKey order.
+  return links
+    .map((link) => ({ field: link.embeddedData, link: { storageKey: link.storageKey } }))
+    .sort((a, b) => rank(a.link.storageKey) - rank(b.link.storageKey));
+};
+
+/**
+ * Replaces the raw `embeddedDataLinks` relation on a Prisma survey row with the inlined
+ * `embeddedFields` the read seam consumes, so the relation shape never leaks onto `TSurvey`.
+ * A no-op for rows read through a select without the join.
+ */
+export const withInlinedEmbeddedFields = <T extends TSurveyWithEmbeddedDataLinks>(
+  surveyPrisma: T
+): Omit<T, "embeddedDataLinks"> => {
+  const { embeddedDataLinks: _links, ...rest } = surveyPrisma;
+  const embeddedFields = inlineSurveyEmbeddedFields(surveyPrisma);
+  return embeddedFields ? { ...rest, embeddedFields } : rest;
+};
--- a/apps/web/lib/survey/service.ts
+++ b/apps/web/lib/survey/service.ts
@@ -637,6 +637,22 @@ export const updateSurveyInternal = async (
         // a partial update leaves `variables` / `hiddenFields` untouched in the column, and reading the
         // payload instead would see them as absent and delete every row. workspaceId comes from the
         // stored survey for the ENG-1749 reason above — never from the client.
+        //
+        // NOTE (ENG-1837): `survey` was read BEFORE this reconcile, so the `embeddedDataLinks` it
+        // carries — and the `embeddedFields` inlined from them by `transformPrismaSurvey` below —
+        // describe the PRE-reconcile rows. A save that renames or removes a field therefore returns a
+        // non-empty *stale* list. No consumer reads it today: the editor's save action feeds the
+        // return into `setLocalSurvey` / `surveyRef.current` and every editor surface resolves through
+        // `getDeclaredEmbeddedFields` (the cards); the v1 management route strips the key with
+        // `withoutInternalSurveyProjections`; the summary's single-use action discards the value and
+        // refreshes. The one surface that does carry it is the audit log's `newObject`.
+        //
+        // Deliberately NOT re-read here: it would put a second deep `selectSurvey` on the editor-save
+        // hot path for a value nothing consumes. If a future consumer needs it (ENG-1853 pointing a
+        // serializer at the rows), the fix must be a re-read through `selectSurvey` — which preserves
+        // the returned object's key shape. Do not strip or re-derive the key instead: this return
+        // value reaches `survey-menu-bar.tsx`, whose change detection deep-compares it against the
+        // editor's working copy and short-circuits on differing key counts.
         await reconcileEmbeddedData(tx, {
           surveyId,
           workspaceId: currentSurvey.workspaceId,
--- a/apps/web/lib/survey/service.ts
+++ b/apps/web/lib/survey/service.ts
@@ -15,6 +15,7 @@ import { TBaseFilters, ZSegmentFilters } from "@formbricks/types/segment";
 import { TSurveyBlock } from "@formbricks/types/surveys/blocks";
 import { TSurvey, TSurveyCreateInput, ZSurvey, ZSurveyCreateInput } from "@formbricks/types/surveys/types";
 import { reconcileEmbeddedData } from "@/lib/embedded-data/reconcile";
+import { selectSurveyEmbeddedDataLinks, withInlinedEmbeddedFields } from "@/lib/embedded-data/survey-fields";
 import {
   getOrganizationByWorkspaceId,
   subscribeOrganizationMembersToSurveyResponses,
@@ -124,6 +125,9 @@ export const selectSurvey = {
   },
   followUps: true,
   slug: true,
+  // ENG-1837: the definitions every reader resolves through, joined and inlined by
+  // `transformPrismaSurvey`. Read-only — the rows are written by `reconcileEmbeddedData`.
+  embeddedDataLinks: selectSurveyEmbeddedDataLinks,
 } satisfies Prisma.SurveySelect;
 
 const reconcilePersistedSurveySchedulingIfDue = async ({
@@ -329,6 +333,12 @@ export const updateSurveyInternal = async (
       id: _id,
       // archivedAt is owned exclusively by the archive/restore flows; never let a survey update touch it.
       archivedAt: _archivedAt,
+      // ENG-1837: `embeddedFields` is a read-only projection of the EmbeddedData tables, inlined by
+      // the join below. `surveyData` is spread straight into `tx.survey.update`'s `data`, and
+      // `Survey` owns relations named `embeddedData` / `embeddedDataLinks` — so leaving it in would
+      // turn a read projection into a nested relation write. The rows are written by
+      // `reconcileEmbeddedData` from the persisted legacy columns instead.
+      embeddedFields: _embeddedFields,
       ...surveyData
     } = updatedSurvey;
 
@@ -904,7 +914,11 @@ export const createSurvey = async (
     // TODO: Fix this, this happens because the survey type "web" is no longer in the zod types but its required in the schema for migration
     // @ts-expect-error
     const transformedSurvey: TSurvey = {
-      ...survey,
+      // ENG-1837: this result is hand-built rather than routed through `transformPrismaSurvey`, so
+      // the inlining has to happen here too — otherwise the raw relation leaks onto TSurvey. The
+      // rows are reconciled *after* the create above, so the list is empty here and the accessor
+      // falls back to the freshly written legacy columns, which carry the same definitions.
+      ...withInlinedEmbeddedFields(survey),
       ...(survey.segment && {
         segment: {
           ...survey.segment,
--- a/apps/web/lib/survey/service.ts
+++ b/apps/web/lib/survey/service.ts
@@ -906,8 +906,9 @@ export const createSurvey = async (
       },
       // This transaction predates ENG-1978, but the reconcile above adds a read plus two writes per
       // field inside it, and neither `variables` nor `hiddenFields` is bounded — so a large template or
-      // API create could now reach Prisma's 5s default where it used to fit. Matched to the other two
-      // reconcile call sites rather than left to inherit a ceiling this work made easier to hit.
+      // API create could now reach Prisma's 5s default where it used to fit. Matched to the other
+      // reconcile call sites (enumerated on `getDeclaredEmbeddedFields`) rather than left to inherit
+      // a ceiling this work made easier to hit.
       { timeout: 20_000, maxWait: 10_000 }
     );
 
--- a/apps/web/lib/utils/recall.ts
+++ b/apps/web/lib/utils/recall.ts
@@ -1,3 +1,4 @@
+import { type TLinkedEmbeddedField, getSurveyEmbeddedFields } from "@formbricks/types/embedded-data-resolver";
 import { type TI18nString } from "@formbricks/types/i18n";
 import { TResponseData, TResponseDataValue, TResponseVariables } from "@formbricks/types/responses";
 import { TSurveyElement } from "@formbricks/types/surveys/elements";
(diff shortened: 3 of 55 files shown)


packages/surveys/src/lib/logic.ts
```diff
@@ -251,7 +267,7 @@ const evaluateSingleCondition = (
 
     if (
       condition.leftOperand.type === "variable" &&
-      (leftField as TSurveyVariable).type === "number" &&
+      (leftField as TResolvableEmbeddedField).dataType === "number" &&
```
mattinannt: **Possible crash: `leftField` can be `undefined` here.**

`leftField` is set from `findComputedField(computedFields, condition.leftOperand.value)?.field as TResolvableEmbeddedField` a few lines up (the `variable` branch). If the condition's stored `storageKey` no longer matches any computed field (e.g. the variable was renamed/deleted after the condition was saved), that lookup resolves to `undefined`, and the `as` cast doesn't change that at runtime. Then `condition.leftOperand.type === "variable"` is still true, so `(leftField as TResolvableEmbeddedField).dataType` throws `TypeError: Cannot read properties of undefined (reading 'dataType')` inside the survey renderer, breaking logic evaluation for the respondent.

This diff rewrote this exact block (same shape reproduced at `apps/web/lib/surveyLogic/utils.ts:308`), so it's a good spot to add a guard for the not-found case, even though the underlying issue predates this PR (same gap existed against `TSurveyVariable`).

---

packages/types/surveys/types.ts
```diff
@@ -3856,6 +3869,11 @@ export const ZSurveyCreateInput = makeSchemaOptional(ZSurveyBase)
     // archivedAt is owned exclusively by the archive/restore flows; a create must never set it,
     // otherwise a caller could POST an already-archived, purge-eligible survey.
     archivedAt: true,
+    // Read-only projection of the EmbeddedData tables (ENG-1837). `createSurvey` spreads the parsed
+    // body straight into `Prisma.SurveyCreateInput`, and `Survey` owns relations named
+    // `embeddedData` / `embeddedDataLinks`, so admitting this key would turn a read projection into
+    // a nested relation write. The rows are written by `reconcileEmbeddedData` instead.
+    embeddedFields: true,
```
mattinannt, line 3887: `ZSurveyCreateInput` and `ZSurveyCreateInputWithWorkspaceId` both omit `embeddedFields` here with this explanation, but `ZSurveyUpdateInput` (a few hundred lines up) doesn't get the same treatment.

In practice it's currently harmless: `PUT /api/v1/management/surveys/{surveyId}` does `ZSurveyUpdateInput.safeParse({ ...result.survey, ...surveyUpdate })` (`result.survey.embeddedFields` being the joined rows from `getSurvey()`), and `updateSurveyInternal` (`apps/web/lib/survey/service.ts:341`) destructures `embeddedFields` back out before it ever reaches Prisma. But that safety lives entirely in the service layer now, not the schema — unlike the create path, which this PR correctly hardens at the validation layer. Worth adding the same `.omit({ embeddedFields: true })` to `ZSurveyUpdateInput` for consistency, so a future caller of that schema doesn't have to rediscover the destructure requirement.

---

mattinannt (commented):
Reviewed the diff between `epic/embedded-data-v1` and this branch. Two findings, both narrow and non-blocking:

1. `packages/surveys/src/lib/logic.ts:270` — a possible `undefined.dataType` crash when a condition's stored variable no longer resolves to a computed field. Pre-existing pattern, but this PR rewrote the exact lines.
2. `packages/types/surveys/types.ts:3876` — `ZSurveyUpdateInput` doesn't omit `embeddedFields` the way the sibling create-input schemas (rightly) do in this PR. Currently mitigated at the service layer, but worth closing at the schema level for consistency.

CodeRabbit's review already covers the missing-`else` numeric coercion bug in `logic.ts`'s `assign` handling (lines 498-506), so I skipped re-flagging that.

Everything else — the resolver/accessor split, the reconcile-on-patch transaction, the editor's declared-vs-stored distinction, and the test coverage described in the PR body — looks sound and well-reasoned for the stated acceptance criteria.

---

apps/web/lib/survey/service.ts
```diff
@@ -904,7 +914,11 @@ export const createSurvey = async (
     // TODO: Fix this, this happens because the survey type "web" is no longer in the zod types but its required in the schema for migration
     // @ts-expect-error
     const transformedSurvey: TSurvey = {
-      ...survey,
+      // ENG-1837: this result is hand-built rather than routed through `transformPrismaSurvey`, so
+      // the inlining has to happen here too — otherwise the raw relation leaks onto TSurvey. The
+      // rows are reconciled *after* the create above, so the list is empty here and the accessor
+      // falls back to the freshly written legacy columns, which carry the same definitions.
+      ...withInlinedEmbeddedFields(survey),
```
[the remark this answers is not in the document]
itsjavi: Accurate observation, and I've documented it — but I'm not adding the re-read. Reasoning, since this is the one finding where I'm knowingly leaving the code as-is:

I traced every consumer of the three return values.

- **`createSurvey`** — the links really are `[]` here, but that is the correct answer, not a stale one: `getSurveyEmbeddedFields` treats an empty list as "no join" and falls back to the legacy columns the same transaction just wrote. Already documented at `service.ts:917-921`.
- **`updateSurveyInternal`** — the only case that returns non-empty stale definitions (a rename). Consumers: `editor/actions.ts:290` → `setLocalSurvey` in the menu bar, and every editor surface reads through `getDeclaredEmbeddedFields`, which ignores rows by design; `v1/management/surveys/[surveyId]/route.ts:218`, where `withoutInternalSurveyProjections` strips the key; `summary/actions.ts:393`, whose caller discards the value and calls `router.refresh()`; and the audit log's `newObject`, which is the one surface that actually carries it.
- **`patchV3Survey`** — same shape, already carries an ENG-1853 note.

So the fix would add a second deep `selectSurvey` to the editor-save hot path for a value nothing currently reads. I've put a note at `updateSurveyInternal`'s reconcile call recording the staleness, the audit log as the affected surface, and the constraint on any future fix: it has to be a **re-read that preserves the key shape**, not a strip or a re-derive — `survey-menu-bar.tsx` feeds this return into both `setLocalSurvey` and `surveyRef.current`, and the editor's change detection short-circuits on key count. (A `selectSurvey` re-read does preserve the shape, so it stays a safe option — it is just unpaid-for today.)

I also deliberately skipped the suggested assertion in `embedded-data-v3-patch.integration.test.ts`: asserting the returned `embeddedFields` right now would pin the stale value as *expected* behaviour, which is worse than leaving it untested.

Worth reconsidering the moment anything reads definitions off a post-write return value — most likely when ENG-1853 moves the v3 serializer onto the tables, which is exactly what the `patch.ts` note flags.

---

apps/web/lib/utils/recall.ts
```diff
@@ -159,14 +203,15 @@ export const getRecallItems = (text: string, survey: TSurvey, languageCode: stri
...
   ids.forEach((recallItemId) => {
-    const isHiddenField = survey.hiddenFields.fieldIds?.includes(recallItemId);
+    const isHiddenField = findEmbeddedField(embeddedFields, recallItemId, "ingested");
     const questions = getElementsFromBlocks(survey.blocks);
     const isSurveyQuestion = questions.find((question) => question.id === recallItemId);
-    const isVariable = survey.variables.find((variable) => variable.id === recallItemId);
+    const isVariable = findEmbeddedField(embeddedFields, recallItemId, "computed");
 
-    const recallItemLabel = getRecallItemLabel(recallItemId, survey, languageCode);
+    const recallItemLabel = resolveRecallItemLabel(recallItemId, survey, languageCode, embeddedFields);
```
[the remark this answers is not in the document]
itsjavi: Applied in `397580c`, and widened slightly — the flagged line was one of three places flattening the same blocks per token.

`getElementsFromBlocks` ran once per recall id in `getRecallItems`, and `resolveRecallItemLabel` derived the same list again for the same id, so it was 2N flattenings for N tokens. `recallToHeadline`'s nested-recall `while` loop had the same problem. All three now hoist the list once, and `resolveRecallItemLabel` takes `elements: TSurveyElement[]` instead of the survey, since that is all it ever used — a narrower signature, and it makes the repeat structurally impossible rather than just fixed at one call site.

Same list, same order, same precedence. `recall.test.ts` 56/56.

---

apps/web/modules/response-pipeline/lib/handle-integrations.ts
```diff
@@ -329,9 +335,11 @@ const extractResponses = async (
...
 
+  const ingestedStorageKeys = getIngestedStorageKeys(survey);
+
   for (const elementId of elementIds) {
-    // Check for hidden field Ids
-    if (survey.hiddenFields.fieldIds?.includes(elementId)) {
+    // Check for ingested (hidden) field storage keys
+    if (ingestedStorageKeys.includes(elementId)) {
       responses.push(processResponseData(pipelineData.response.data[elementId]));
       elements.push(elementId);
```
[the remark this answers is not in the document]
itsjavi: Not taking this one, and I'd rather say why than quietly resolve it — the destination you're describing is right, but applying it here makes things worse, not better.

It is a **no-op today**: `toDesiredEmbeddedFields` sets `name: fieldId` and `storageKey: fieldId` for every ingested field, so `field.name === link.storageKey` for all of them. Nothing changes until names can diverge, which needs the v2 shared library (ENG-1851).

Two problems with landing it now:

1. **It desyncs the pipeline from the mapping modal.** `AddIntegrationModal.tsx:137-141` keys *and* labels ingested fields on the storage key. Moving only the pipeline to `field.name` recreates exactly the picker-vs-consumer mismatch this PR already had to fix once for this same pair of files — the recall picker writing one label while the consumer resolved another.
2. **`elements` is the destination column header.** Notion/Sheets mappings are keyed by it, so once names diverge, changing the label silently re-headers every already-configured integration. That is a migration, not a rename.

So it needs both files plus a mapping-migration story, moved together, in the ticket that makes divergence possible. Doing it inside a PR whose acceptance criterion is byte-identical behaviour buys nothing and takes on that risk early.

I've recorded the asymmetry at the ingested branch in `397580c` — why computed labels by `field.name` (its storage key is an opaque cuid) and ingested by storage key (that *is* its name today), that the modal keys on the storage key, that the label is the column header, and that the two files must move together at ENG-1851 — so whoever picks this up finds the constraint instead of rediscovering it.

---

apps/web/modules/survey/components/element-form-input/components/recall-item-select.tsx
```diff
@@ -82,36 +83,59 @@ export const RecallItemSelect = ({
...
-    return [];
-  }, [localSurvey.variables, recallItemIds]);
+    return embeddedFields.map(({ field, link }) => ({
+      key: link.storageKey,
+      // The enumerator's blank-name fallback is the key, so mirror it when a field is not listed.
+      label: labelByKey.get(link.storageKey) ?? link.storageKey,
+      source: field.source,
+      dataType: field.dataType,
+    }));
+  }, [embeddedFields]);
```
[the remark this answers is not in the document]
itsjavi: You're right that it is functionally redundant — the Map lookup reproduces `labelOrKey(field.name, link.storageKey)` exactly. Keeping it anyway, for two reasons:

- The blank-name fallback stays the enumerator's single responsibility. Inlining it here duplicates a rule that `listReadableFields` already owns for every other consumer.
- ENG-1853 extends this path onto the unified picker rather than replacing it, so inlining now is work that gets undone.

It runs once per `useMemo`, not per render, so the cost is a single pass over a list that is a handful of entries in practice.

The keying on `storageKey` rather than position is load-bearing though — that came out of an earlier review round, where the join was index-based and would have mislabelled every entry past the first drop if `listReadableFields` ever filtered. The existing comment covers it.

---

packages/types/embedded-data-resolver.test.ts
```diff
@@ -880,3 +888,178 @@ describe("deriveLegacyEmbeddedData", () => {
...
+
+    expect(z.array(ZLinkedEmbeddedField).parse(derived)).toStrictEqual(derived);
+  });
+
+  test("strips nothing the readers need, and rejects a blank name or storage key", () => {
+    expect(ZLinkedEmbeddedField.safeParse({ ...pair, field: { ...pair.field, name: "  " } }).success).toBe(
+      false
+    );
+    expect(ZLinkedEmbeddedField.safeParse({ ...pair, link: { storageKey: "" } }).success).toBe(false);
+  });
```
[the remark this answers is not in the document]
itsjavi: Applied in `397580c`, plus a valid-pair control alongside it so the rejections prove the rules are firing rather than the whole shape being refused.

I checked the assertion actually discriminates, since "add a test" is only worth it if the test can fail:

| Sabotage | Result |
| --- | --- |
| both `storageKey` refines removed (`storageKey: z.string()`) | **fails** — `AssertionError: expected true to be false` |
| only the blank refine removed, charset kept | still passes |

So whitespace is defended twice — by `key.trim().length > 0` *and* by `isLegacyIdCharset`, whose `/^[a-zA-Z0-9_-]+$/` rejects spaces — and removing either alone keeps the test green. The comment says exactly that, so nobody reads the test as pinning the blank check alone.

---

packages/surveys/src/lib/logic.ts
itsjavi: Good catch — fixed in `0d59c0f`. You were right that it predates this PR (`244afe6` did `localSurvey.variables.find(...) as TSurveyVariable` then read `.type`, failing identically), and right that this was the place to deal with it.

**One correction to the severity, and it is worse than a crash.** `evaluateSingleCondition` wraps its whole body in `try { … } catch { return false }` (`logic.ts:435`, `utils.ts` likewise), so the `TypeError` never surfaced. The condition just evaluated `false` — respondent sent down the wrong branch, any quota depending on it miscounted, nothing in any log. That also shaped the tests: a "does not throw" assertion would have passed *without* the fix, so they assert the evaluated result instead.

**Audit before fixing** — all 19 field accesses across both engines:

| Access | Guard in front of it | Reachable with `undefined`? |
| --- | --- | --- |
| `(leftField).dataType === "number"` | `leftOperand.type === "variable" &&` — the first clause, i.e. exactly the branch that yields `undefined` | **yes, the one site** |
| `(leftField).type` ×4 | all under `leftOperand.type === "element"`, whose branch is `?? ""` | no |
| `(rightField).type` ×4 | all under `rightOperand?.type === "element"`, same | no |

Exactly one reachable site per engine — the line you pointed at. `rightField`'s variable branch could also hold `undefined`, but nothing reads it under a variable guard: latent, not live.

**Fixed once, not twice.** SonarCloud was simultaneously failing this PR on new-code duplication, and the duplicated blocks it named were precisely the helpers this PR had added verbatim to both engines — so guarding both symmetrically would have made that worse. Instead the shared pieces moved into `@formbricks/types/embedded-data-resolver` (both engines already import it): the field lookup, a `getComputedFieldDataType` that answers `undefined` rather than throwing, and the value expression. One sabotage in the shared helper now fails **both** engines' tests, which is the property worth having given how often these two have drifted.

That also let it collapse further than a guard would: with the dataType lookup no longer needing the field object, the `variable` branch folds into the existing default and both unions narrow from `TSurveyElement | TResolvableEmbeddedField | string` to `TSurveyElement | string` — so the `undefined` can no longer be *constructed*, rather than being constructed and then guarded.

Semantics: a missing definition means the operand's type is unknown, so the number-coercion branch does not fire and evaluation falls through exactly as for a text variable. No-op for every condition whose variable still exists.

Falsifiability — dropping the optional chain in the shared helper:

```
packages/surveys  × evaluates on its own merits instead of collapsing to false   1 failed | 52 passed
apps/web          × evaluates on its own merits instead of collapsing to false   1 failed | 35 passed
```

No existing test needed editing. Two regression tests per engine (the stale operand, plus the paired `isSet` case so the passing direction can't be vacuous), eight in `packages/types` at the rule's new home, and I backfilled the number-vs-hiddenField coercion test into `apps/web`, which only `packages/surveys` had. Widget bundle is 97 B *smaller*.

---

packages/types/surveys/types.ts
itsjavi: Agreed — added in `0d59c0f`. `ZSurveyUpdateInput` now omits `embeddedFields` like both create schemas, so the guarantee lives at the validation layer on the update path too rather than only in the service destructure.

Verified `.omit()` strips rather than rejects before relying on it, with a probe against the real schema, since a reject would have broken the v1 PUT round-trip you describe:

```
SUCCESS: true                          ← parses fine, not rejected
HAS embeddedFields AFTER PARSE: false  ← key stripped from the output
```

**The service-layer destructure stays**, and I've noted why at the site: `updateSurveyAction` validates with `ZSurvey` (which retains the key), and `summary/actions.ts:393` spreads a raw `getSurvey()` result — neither goes through `ZSurveyUpdateInput`, so removing the destructure on the strength of this omit would reopen the Prisma write path. The omit is defence in depth for callers of the schema, not a replacement for it.

Adjacent suites green: 93 files / 1075 tests in `apps/web`, 230 in `packages/types`.

---

apps/web/lib/embedded-data/survey-fields.ts
```diff
@@ -0,0 +1,87 @@
...
+ * rather than being dropped, in the select's `storageKey` order.
+ */
+export const inlineSurveyEmbeddedFields = (
+  surveyPrisma: TSurveyWithEmbeddedDataLinks
+): TLinkedEmbeddedField[] | undefined => {
+  const links = surveyPrisma.embeddedDataLinks;
+  if (!links) return undefined;
+
+  const legacyRankByStorageKey = new Map(
+    toDesiredEmbeddedFields({
```
pandeymangg, line 106: This feeds raw Prisma JSON to `toDesiredEmbeddedFields` behind a cast, and that function maps with `?? []` — which catches null and undefined but not a wrong type. Confirmed on this branch:

```
TypeError: (variables ?? []).map is not a function
```

`inlineSurveyEmbeddedFields` runs inside `transformPrismaSurvey`, so a survey whose `variables` is an object rather than an array stops loading entirely — not a degraded ordering, a failed read on a live path. Same for a non-array `hiddenFields.fieldIds`.

This is the asymmetry you flagged on #8832, now on the read side: ENG-1835's `planSurveyBackfill` checks these shapes and skips the survey, and the very next consumer of the same columns casts. Production has zero such rows — I ran the crash-path query before that PR merged — so this is insurance, not a live bug.

Cheapest fix is the ordering doesn't need to be correct for a malformed survey, only survivable:

```ts
const rank = (storageKey: string): number => ...
```
wrapped so a throw falls back to the select's `storageKey` order, or an `Array.isArray` check before the call. Either keeps the survey loading and degrades only the column order — which for a survey with malformed declarations is already undefined.

pandeymangg (commented):
Reviewed the source rather than the diff stat — 2021 of the 3005 additions are tests, and the source is +984/−327 across 44 files with no file over 145 lines. Wide, not deep, which is what repointing every reader looks like. Ran `packages/types` (230) and `packages/surveys` (677) on `0d59c0f5`, both green.

One finding inline, not blocking. What I checked and can confirm:

- **"Exactly four reconciling writes" holds.** Four `reconcileEmbeddedData(` call sites, and the other survey writers (`segments.ts`, `process-response-pipeline-job.ts`, `slug.ts`, `surveys.ts`) touch only segment / status / slug / archivedAt — never the legacy columns. The grep really is the audit.
- **The value expressions are verbatim.** `getLogicVariableValue` is `dataType === "number" ? Number(v) || 0 : v || ""`, character for character what `getVariableValue` was. Given the resolver's own swap checklist, keeping value arithmetic off `resolveEmbeddedValue` was the right call — adopting it would have changed what stored responses evaluate to.
- **The stale post-reconcile `embeddedFields` really is inert.** `withoutInternalSurveyProjections` does strip it, and naming the audit log's `newObject` as the one surface that keeps it is correct — that assignment happens before the strip.
- **`ZLinkedEmbeddedField` earns its place.** Zod v4 stripping undeclared keys would have made the join vanish on `ZSurvey.parse`, and both create inputs plus the update input omit it so it can't become a nested relation write.
- The ordering solution — rank by the legacy columns, stable sort, unranked last — is the right shape given neither table carries an ordinal, and the comment says so where the next person will look.

The stale-variable-operand fix is a good catch and the reasoning for asserting the *evaluated result* rather than absence-of-throw is exactly right; a "does not throw" test would have passed against the bug.

Nit: the test table in the description reads 222 / 675, but the branch is at 230 / 677 — looks written before the last commit added its tests. Only worth fixing because the table invites a reviewer to re-run it.

pandeymangg:
Small hygiene thing: there are six `embedded-data*` files sitting flat at the root of `packages/types` now, and they'd read better folded into `packages/types/embedded-data/` the way `surveys/` and `integration/` already are.

apps/web/lib/embedded-data/survey-fields.ts
itsjavi: Good catch, and the asymmetry framing is the convincing part — the writer checks these shapes and the very next reader of the same columns casts. Fixed in `adfe420`.

**Guards at the read boundary, deliberately not in `toDesiredEmbeddedFields`.** That mapper is shared with `reconcileEmbeddedData` on the *write* path, where throwing on a malformed survey is correct: making it permissive would map to an empty desired set and **delete every row the survey has**. Leniency belongs on the read side only, and the comment says so — which is the same read/write split `planSurveyBackfill` makes from the other end.

**Explicit shape guards rather than the `try`/`catch` fallback.** A catch would also swallow a genuine bug inside the mapper, and it degrades worse: guarding each column independently means a survey with malformed `variables` still ranks its hidden fields correctly and only the variables sort last, where a catch loses all ranking.

I traced the actual throw surface rather than guarding by feel. `toDesiredEmbeddedFields` throws on the *container* shape (`(variables ?? []).map`, `(fieldIds ?? []).map`) and on a `null` entry, which throws at the property read. Entries of the wrong *kind* — a string in `variables`, a number in `fieldIds` — never throw; they yield a storage key matching no row, which already sorts last. So the guards disqualify a list only for `null` entries and drop non-strings rather than discarding the list: the narrowest change that closes the throw. `hiddenFields` needed one more than you named, incidentally — a non-object `hiddenFields` (a string, an array) reaches `.fieldIds` too, so that read is guarded separately.

The call still routes *through* `toDesiredEmbeddedFields` rather than deriving ids inline: which key addresses a field is that function's rule, and re-deriving it here is exactly how the ordering would drift from the rows it is ranking.

**Three regression cases, written with raw SQL** (`prisma.$executeRaw`) — the generated Prisma types make the malformed shape unrepresentable through the client, so a typed fixture could only *resemble* the row. Reverting to the old casts:

```
× still loads when `variables` holds an object instead of an array
    → TypeError: (variables ?? []).map is not a function
× still loads when `hiddenFields.fieldIds` holds a string instead of an array
    → TypeError: ((intermediate value) ?? []).map is not a function
× still loads when both columns are malformed
    → TypeError: (variables ?? []).map is not a function
Tests  3 failed | 6 passed (9)
```

Your exact reproduction, and only the three new cases discriminate. Each asserts the survey resolves **and** that the well-formed group keeps its declared order — with `variables` malformed, the hidden fields still come out `["utm_source", "plan", …]` with the variable rows trailing in storage-key order. Integration suite 26 files / 120 tests.

---

pandeymangg (approved):

PR #8859, opened 2026-08-13
refactor(mcp): migrate MCP to latest protocol (SDK v2) and reject undeclared tool arguments

xernobyl:
## What & why

The MCP server was on the 2025-era protocol via `@modelcontextprotocol/sdk` v1. This moves it to
revision **2026-07-28** and the v2 package split, and — because the migration is what finally makes it
possible — fixes ENG-2256 in the same PR.

**Not a breaking change for clients**, despite the revision being one. SDK v2 is a dual-era
implementation and `mcp-handler` serves with `legacy: "stateless"`, so the one endpoint answers
2026-07-28 and the legacy `initialize` handshake side by side. Both legs are covered by tests. Nothing was
needed for `server/discover`, the newly required `resultType`, or the `ttlMs`/`cacheScope` cache hints —
the SDK handles all three.

Two commits, deliberately separate:

**1. `refactor(mcp)` — the migration.** `@modelcontextprotocol/sdk` → `@modelcontextprotocol/server`
2.0.0, `mcp-handler` → 2.1.1, `@posthog/mcp` → 0.11.6. `createMcpHandler` takes one options object
instead of three arguments; the v1 handler options are gone rather than moved (sessions and the SSE
transport left the protocol, and the handler now serves whatever route it is mounted on). Tool input
schemas are registered as schema instances, which v2 requires in place of raw `.shape` shapes. The
handler context is `ctx` with auth at `ctx.http.authInfo`, reached through one new accessor
(`getMcpToolAuthInfo`) so the next such move is one edit rather than one per tool.

Our own auth layer is untouched: v2's `mcp-handler` still declares and reads `Request.auth`, and
`AuthInfo.extra` survives, so the API-key/OAuth wrapper and `TMcpAuthInfo` port as-is.

It also fixes an identity regression the migration would otherwise have introduced **silently**:
`identifyMcpUser` read the flat v1 `extra.authInfo`, which is `undefined` under v2, so every PostHog MCP
event would have fallen back to an anonymous `distinctId` with nothing logged. It now reads both
locations, and two new tests assert *which* one is used rather than mocking past it.

**2. `fix(mcp)` — ENG-2256.** Every tool input schema is now `.strict()`. This was not fixable at this
layer before: v1 was handed a raw `.shape` and rebuilt it through `objectFromShape()`, which discards any
unknown-keys policy, so `.strict()` would have been decorative. On a filter the old behaviour failed in
the dangerous direction — `count_feedback_records` with the pre-#8650 spelling `userId` had the key
stripped, ran unfiltered, and returned the count for every record in the dataset while reporting success.

Strictness applies at **every depth**, not just the outermost object — a nested `filter.status.include`
typo used to leave `filter.status` as `{}` and run the query unfiltered. Raised by @mattinannt in review;
`schemas.test.ts` now enforces it against the schema we actually advertise.

**One documented exception**, found by that same audit and deliberately not fixed here: everything below
`definition` on `create_workflow` / `patch_workflow` is still open (15 nodes each, versus 0 across the
other 24 tool schemas). That subtree is `ZWorkflowDefinition` from `packages/workflows`, which the v3
Workflows REST route parses and the workflow builder posts, so making it strict is a v3 API change with its
own blast radius — the same reason the batch feedback-record fix uses an MCP-only strict variant instead of
tightening the shared v3 body. Filed as ENG-2437; the new guard pins the hole to that subtree so it cannot
spread to a top-level argument.

## Linear ticket

Fixes ENG-2301
Fixes ENG-2256

## How this was tested

- `pnpm typecheck` ✅, `eslint` on the touched files ✅, `prettier --check` ✅
- Full web suite: **610 files / 8082 tests ✅** (`vitest run` in `apps/web`), clean exit. An earlier run on
  this branch showed a pre-existing unhandled rejection in
  `modules/ee/workflows/lib/runner/process-workflow-run-job.test.ts` (`logger.withContext is not a
  function`, from better-auth observability); it did not reproduce on the latest full run, so treat it as
  order-dependent rather than fixed here — nothing in this diff touches it.
- MCP suites specifically (incl. `app/api/mcp`): **371 tests ✅** (was 131 — the rest added here).
- **Tests added** — all driving a real `McpServer` and a real `tools/call`, because every pre-existing MCP
  test calls handlers directly and so skips the SDK's `validateToolInput`, which is exactly why ENG-2256
  was invisible from our side of the boundary:
  - `surveys.test.ts`: an undeclared argument on `list_surveys` is rejected and the v3 operation is never
    reached, in **both** the 2026-07-28 and the legacy leg. **Fault-injected to confirm it can fail:**
    removing `.strict()` from `ZMcpListSurveysInput` fails both era cases; restored and re-verified.
  - `feedback-records.test.ts`: the reported `userId` symptom, both eras. Documented in the test that
    this one passes either way — that schema extends an already-`.strict()` v3 filter schema and Zod 4's
    `.extend()` carries strictness over, so for the feedback-record tools the hole closes on the v2
    migration alone.
  - `server.test.ts`: two tests asserting which context location `identifyMcpUser` reads.
  - `schemas.test.ts` (new file): walks **every** advertised tool input schema to any depth — including
    nodes reachable only through a `$ref` — and fails on any structured object that is not strict. This
    exists because the strict change was wrong twice in the same way and both times a human found it, not
    a test: the first version left every nested `filter` open, the next still left the workflow
    `definition` subtree open. It also pins the set of fields allowed to be free-form, so *widening* a
    structured object to `additionalProperties: true` is as loud as leaving it open. **Fault-injected three
    ways:** dropping strictness from a nested filter, from a workflow tool's top-level object, and switching
    a strict sub-object to `z.looseObject` — each turns it red with the offending path named.
  - `app/api/mcp/route.test.ts`: `subscriptions/listen` is refused end to end through the real route, which
    is what proves `maxSubscriptions: 0` reaches the SDK rather than just being declared. Note for anyone
    writing 2026-era requests: the full `_meta` envelope (`protocolVersion`, `clientCapabilities`,
    `clientInfo`) is required on **every** request, since there is no `initialize` to carry it once, and the
    method must also appear in an `Mcp-Method` header agreeing with the body.

**On review findings.** @mattinannt found two (nested strictness, stale prose); CodeRabbit found three more,
two of them real. Every one of the real ones was the same defect: an assertion whose payload was already
invalid for unrelated reasons, so it passed regardless of the behaviour it claimed to pin — `create_workflow`
strictness asserted against a payload with a bad cuid2 *and* an empty `definition`, and my own first draft of
the subscription test passed with incomplete params because the cap is checked before param validation. Each
is now asserted against a fully valid payload and fault-injected to prove it can fail.
- The deployed request path **is** covered: `app/api/mcp/route.test.ts` drives the real `POST /api/mcp`
  → auth wrapper → `mcpHandler` → real SDK (only Prisma, JWT verification and the rate limiter are
  mocked), including a real `tools/list`, real `list_surveys`/`create_survey`/`validate_survey` calls,
  both auth methods, and scope gating. 12 tests ✅.
- **Verified with a real external client against a running server on this branch — both eras.**
  Claude Code over OAuth (full DCR → consent → token → tool calls):
  - **Legacy (2025) leg**, which is Claude Code's default: connects via `initialize` +
    `notifications/initialized` (visible as a `202` on `/api/mcp`), then lists and calls tools.
  - **2026-07-28 leg**, forced with `MCP_SDK_GENERATION=v2 MCP_PROTOCOL_NEGOTIATION=auto claude`.
    Server-side trace of that session, with the negotiated version read off each request's `_meta`:

    | time | method | protocol version |
    | --- | --- | --- |
    | 13:23:54.673 | `server/discover` | `2026-07-28` |
    | 13:23:54.690 | `subscriptions/listen` | `2026-07-28` (refused, see Risks) |
    | 13:23:54.711 | `tools/list` | `2026-07-28` |
    | 13:23:55.712 | `subscriptions/listen` | retry +1s |
    | 13:23:57.746 | `subscriptions/listen` | retry +2s |
    | 13:24:01.788 | `subscriptions/listen` | retry +4s, then stops |
    | 13:24:18.617 | `tools/call` | `2026-07-28` |
    | 13:24:20.334 | `tools/call` | `2026-07-28` |

    `server/discover` and `subscriptions/listen` exist **only** in 2026-07-28, and the version string is
    read from the request rather than inferred from the method names.

  So the "one endpoint serves both eras" claim is confirmed end to end by a third-party client, not only
  by tests. Both env vars are required — see the QA plan for why.

  What this trace does **not** show: which tools those two `tools/call`s were, or whether they returned
  `isError`. Our tools log only on failure, and a tool-level error rides inside an HTTP 200 (that is
  exactly how the refused `subscriptions/listen` appears above), so the trace evidences the protocol
  behaviour and not the per-tool outcomes. Those are covered by the tests instead.
- Also unverified locally: that PostHog still receives MCP tool-call events under the stateless protocol
  (see Risks).
- **Live smoke against a running server on this branch**, re-run on the final head: 401 unauthenticated with
  the correct `WWW-Authenticate` challenge, 200 protected-resource metadata, 401 bogus API key, 400 for a
  credential in the query string, 403 cross-origin. Plus two authenticated calls from a real client, which
  confirm the advertised strict schemas work end to end rather than only in tests:
  - `list_workspaces` — advertised as `{"type":"object","properties":{},"additionalProperties":false}`
    and called successfully with `{}`, so the strict empty schema does not break the discovery tool for the
    client we ship against.
  - `count_feedback_records` with `user_id: "nobody-at-all"` returns `count: 0`, not the whole-dataset
    total — ENG-2256's original symptom, now behaving. Worth noting the wrong spelling can no longer even be
    *sent* by a schema-aware client: `additionalProperties: false` makes it fail client-side before a
    request exists, which is why the server-side proof lives in the tests.

## Why this wasn't split

The size check flags it at ~1050 lines, so: the two commits are the smallest pair that each leave the tree
green, and neither works alone. ENG-2256 is only *fixable* after the migration — v1 rebuilt the raw
`.shape` and discarded the unknown-keys policy, so shipping `.strict()` first would have added 26 no-op
decorations and no behaviour change, with nothing to test. Splitting the other way is worse: the migration
alone changes 30 tool registrations from `.shape` to schema instances, which is the line that *activates*
strictness, so it would silently ship the client-visible behaviour change with no test, no docs and no
breaking-change note. Roughly half the diff is tests and the handbook.

## Breaking changes

| Undeclared keys in an MCP tool call's `arguments` | Silently dropped; the call ran without them | Rejected with an input-validation error; the operation is not reached | MCP clients that add their own keys to `arguments` | Send only the arguments a tool declares. A misspelled filter now errors instead of returning a wider result set |
| **Read-modify-write on `update_feedback_record`** | A record echoed back from `get_feedback_record` had its immutable fields silently ignored | Rejected — the tool accepts only the eight mutable fields | Any client that fetches a record, edits a value and sends the record back | Strip provenance (`source_*`, `field_*`, `submission_id`, `collected_at`) and the derived sentiment/emotions/translation before sending. The error names every offending key, and the tool description now says so |
| **`list_workspaces` with any argument** | Placeholder/extra keys dropped; the call succeeded | Rejected (the schema declares no properties) | A client that pads zero-argument calls with a dummy key | Send `{}`. Called out because this is the discovery tool every workspace-scoped tool takes its `workspaceId` from, so losing it looks like the whole server failing |

The protocol revision itself is **not** breaking here: the endpoint continues to serve 2025-era clients.

## QA / Test Plan

**How to test**

- [ ] Connect Claude Code (or the Claude app) to `/api/mcp` over OAuth as today → the connection
      completes and `list_workspaces` returns your workspaces. Nothing about the setup flow changes.
- [ ] Run a read tool (`list_surveys` with a `workspaceId`) → returns surveys as before.
- [ ] Run a mutation (`create_survey`, then `delete_survey`) → succeeds, and the audit log records the
      create and the delete.
- [ ] **Legacy era (Claude Code's default):** connect normally → works. Confirm it took the 2025 path
      by looking for a `202` response on `/api/mcp` (that is `notifications/initialized`, which only
      exists pre-2026-07-28).
- [ ] **2026-07-28 era:** start the client with `MCP_SDK_GENERATION=v2 MCP_PROTOCOL_NEGOTIATION=auto claude`
      → still connects and calls tools, and the server now sees `server/discover` and
      `subscriptions/listen`, neither of which exists in the 2025 era. Both env vars are needed: unset,
      `MCP_SDK_GENERATION` resolves to `v1` (a client that cannot speak 2026-07-28) and
      `MCP_PROTOCOL_NEGOTIATION` falls back to a per-transport flag that is off, so the client stays on
      the legacy leg. This is the compatibility claim worth checking by hand, in both directions.
- [ ] Feedback-record tools with the Enterprise entitlement → `count_feedback_records` with `user_id`
      returns that user's count.
- [ ] **Edge case, the ENG-2256 fix:** call `count_feedback_records` with `userId` (camelCase, the wrong
      key) → the call now **fails** with an input-validation error naming `userId`, instead of returning
      the whole-dataset count. Same for any invented key on any tool, e.g. `list_surveys` with
      `statuses`.
- [ ] Edge case: a tool call with only valid arguments plus one extra unrelated key → rejected. Confirm
      your client surfaces the error rather than hanging.

- [ ] **Read-modify-write:** `get_feedback_record`, then send the record straight back to
      `update_feedback_record` with one value changed → rejected, and the error names the provenance keys to
      remove. Strip those and retry → succeeds.
- [ ] **Another client, if one is to hand** (Cursor, VS Code, an SDK script): connect and call
      `list_workspaces`. Every step above uses Claude Code, so nothing else currently covers a client that
      formats zero-argument calls differently.

**Preconditions / test data**

- A workspace with at least one survey, and an account with write/manage permission for the mutation
  steps.
- Feedback-record steps need the `feedbackDirectories` Enterprise entitlement and a dataset assigned to
  the workspace.
- For the older-client step, an MCP client build that predates 2026-07-28 support.

**Risks & regressions**

- **PostHog MCP analytics.** `@posthog/mcp` documents `identify` as firing once per session "at
  `initialize`", and learns `$mcp_protocol_version` there. 2026-07-28 has neither an `initialize` nor a
  session, so session grouping and that property may degrade or fire per request for modern clients. The
  identity *location* fix is covered by tests; whether events still land and group correctly needs a look
  at PostHog after deploy. Follow-up ticket rather than widening this PR if it has degraded.
- **`subscriptions/listen` — newly reachable, now refused (`maxSubscriptions: 0`).** The revision
  replaced the GET SSE endpoint with a long-lived `subscriptions/listen` POST stream, which our POST
  route serves; under the v1 `disableSse: true` there was no long-lived path at all, so the migration
  opened one. A 2026-era client opens a stream on connect (observed, not hypothesised) and would then hold
  a connection plus a keepalive timer for nothing — we register no resources and emit no list-changed
  notifications, so it can never deliver anything. The SDK's default cap is 1024 *per process* and the
  handler is a module singleton, so any authenticated caller could hold that many.

  Closed in this PR rather than deferred, since the risk is introduced here. `0` disables rather than
  meaning unlimited (the router gates on `open.size >= maxSubscriptions`, read with `??`), verified
  against the SDK: `-32603 Subscription limit reached` on plain JSON with the connection closed.

  **Refusing it is safe for a real client** — this was the open question, since the client opens the
  stream during connect and a fatal reaction would have broken every 2026-era client without any test
  going red. Verified against Claude Code: it takes the refusal, proceeds straight to `tools/list`,
  retries three times with exponential backoff (1s/2s/4s), gives up, and calls tools normally (the trace
  above). If a future client does treat it as fatal, prefer `maxSubscriptions: 1` — accepted but capped —
  over reverting to the default.
- **Strictness is client-visible.** Any client adding keys to `arguments` starts failing. Nothing in the
  spec invites that, but it is a behaviour change (see Breaking changes). Two cases are sharp enough to have
  their own rows in that table: read-modify-write against `update_feedback_record`, and `list_workspaces`
  rejecting placeholder arguments.
- **`.strict()` on `list_workspaces` buys no ENG-2256 protection** — an empty object has no key to misspell,
  so it is uniformity only, and it is the one place the cost is not matched by a benefit. Kept deliberately
  (uniform strictness was an explicit choice on this PR, and no client that pads zero-argument calls is
  confirmed), but flagged: our QA only exercises Claude Code, so no listed step would catch a client-quirk
  regression here. Raised in review; **open for a second opinion** — dropping it from this one schema costs
  nothing in coverage.
- **The workflow `definition` subtree is still lenient** (ENG-2437). A misspelled key inside a trigger,
  node, edge or node `config` is dropped rather than rejected, so `create_workflow` can report success
  having quietly ignored part of what was asked for. Pre-existing, not introduced here, and bounded by the
  new guard — but it means the "every argument is validated" claim stops at `definition`.
- Nearby: the OAuth/API-key wrapper in `modules/mcp/auth.ts` is logically unchanged but every tool now
  reaches the token through the new accessor — worth re-checking that a token missing a scope still gets
  a 403 insufficient-scope result (covered by tests, cheap to confirm live).
- `mcp-handler` 2.1.1 and `@posthog/mcp` 0.11.6 were both inside the repo's 3-day
  `minimumReleaseAge` window when this branch opened, so it started on 2.1.0 / 0.11.4. Both now clear the
  gate and have been bumped. 2.1.1 is a single patch (forwards `maxSubscriptions`); we register no
  subscriptions and its internal `legacy: "stateless"` serving is unchanged, so the dual-era behaviour is
  unaffected.
- `@modelcontextprotocol/sdk` v1 is still in the lockfile as a transitive optional peer of
  `@posthog/mcp`. It is types-only there (that package has zero runtime imports of either SDK), so
  nothing bundles it.

- none


apps/web/modules/mcp/tools/schemas.ts
```diff
@@ -12,64 +12,89 @@ import {
...
+      .default(true),
+    filter: z
+      .object({
+        name: z
+          .object({
+            contains: z.string().max(512).optional().describe("Case-insensitive survey name substring."),
+          })
+          .describe("Filter by survey name.")
+          .optional(),
+        status: z
```
mattinannt, line 86: **Nested filter objects aren't `.strict()`, so ENG-2256's failure mode survives one level down.**

`ZMcpListSurveysInput` gets `.strict()` on the *outer* object only. `filter`, and its `name`/`status`/`type` sub-objects, are plain `z.object(...)` — not `.strict()`. Verified against the installed zod (4.3.6):

```js
const schema = z.object({
  workspaceId: z.string(),
  filter: z.object({
    status: z.object({ in: z.array(z.string()).optional() }).optional(),
  }).optional(),
}).strict();

schema.safeParse({ workspaceId: 'x', filter: { status: { in: ['draft'], bogusKey: 1 } } }).success
// => true — bogusKey silently dropped, "in" filter still applied but a typo'd key (e.g. `include` instead of `in`) would be dropped too, producing an unfiltered/wider result reported as success.
```

That's exactly the `count_feedback_records`/`userId` bug this PR fixes at the top level, reopened one level down: a client (or LLM) that misspells a nested filter key — e.g. `filter.status.include` instead of `filter.status.in` — gets the key silently stripped, the filter becomes a no-op, and `list_surveys` returns every survey while reporting success.

Same gap in `apps/web/modules/mcp/tools/workflow-schemas.ts`: `ZMcpListWorkflowsInput.filter.{name,status}` (lines ~31/44) and `ZMcpListWorkflowRunsInput.filter.status` (line ~89). Also `ZMcpSurveyLanguageInput` (used by `ZMcpCreateSurveyInput.languages`, schemas.ts ~line 99) is a structured (non-free-form) sub-schema that's likewise not strict.

The doc note added in this PR ("Strictness binds the top level only... the opaque `z.record` fields... still accept any nested shape") only calls out the deliberately-free-form fields (`blocks`, `metadata`, etc.) as exceptions — these nested filter/language objects are structured schemas that seem to have been missed rather than deliberately excluded.

Suggest adding `.strict()` to each nested filter/language sub-object as well.

---

apps/web/modules/mcp/tools/schemas.ts
```diff
@@ -139,72 +172,80 @@ const datasetIdField = ZId.optional().describe(
...
+export const ZMcpGetFeedbackRecordInput = z
+  .object({
+    workspaceId: ZId.describe("Workspace ID that owns the feedback record."),
+    feedbackRecordId: z.uuid().describe("Feedback record ID (UUID) to fetch."),
+    datasetId: datasetIdField,
+  })
+  .strict();
 
 // Extends the plain field object (not the refined body): `inputSchema` needs a raw shape, and the
 // value/field_type rule is enforced by the operations layer, which is also where MCP-stripped unknown
```
mattinannt: Minor: this comment ("...MCP-stripped unknown keys become visible as a missing value") describes the pre-ENG-2256 behavior. With `.strict()` now applied on `ZMcpCreateFeedbackRecordInput` below, an unknown/misspelled key is no longer silently *stripped down to a missing value* at the operations layer — the whole call is *rejected* by the SDK before this handler ever runs. Worth updating so the comment matches the new (stricter, and better) behavior it sits next to.

Same stale phrasing in `apps/web/modules/mcp/tools/feedback-records.ts` (~line 144): "Safe because MCP strips unknown keys and the operation allowlists what reaches the Hub" — also describes the old strip-and-fall-through behavior rather than the new reject-before-handler behavior this PR introduces.

---

mattinannt (commented):
Automated single-pass review (no multi-agent fan-out / subagent verify in this run). Verified the SDK v2 wiring (`http.authInfo` threading, `mcp-handler`'s `.fetch(req, {authInfo: req.auth})`, `maxSubscriptions` gating, zod `.strict()`/`.extend()` semantics) directly against the installed `@modelcontextprotocol/server`, `mcp-handler`, and `zod` packages in `node_modules` — that part of the migration checks out. Two findings below, most-severe first; the second is a documentation nit rather than a functional bug.

---

mattinannt (commented):
## Code review

Reviewed the MCP protocol v2 migration (ENG-2301) and the `.strict()` undeclared-argument fix (ENG-2256).

**Findings: none.**

What was checked, beyond a static read of the diff:

- No stale `@modelcontextprotocol/sdk` v1 imports, `extra.authInfo`, or `.shape`-as-inputSchema call sites remain anywhere in `apps/web` after the migration.
- Verified against the actually-installed `@modelcontextprotocol/server@2.0.0` / `mcp-handler@2.1.1` sources that `registerTool` accepts a schema instance (not `.shape`), that `ServerContext.http.authInfo` is the real v2 auth location, and that `maxSubscriptions` is forwarded and gated with `??`/`>=` exactly as described (so `0` truly disables rather than falling back to the SDK's 1024 default).
- Confirmed `@posthog/mcp`'s `identify` shape is unchanged between the old and new version, and traced its internal `extra` plumbing to validate the dual-read fallback in `identifyMcpUser`.
- Ran the MCP test suite: 144/144 passing. `pnpm typecheck` clean on the touched files.
- `.strict()` placement and the `.extend()`/already-`strictObject` interactions are internally consistent with what the added tests assert.
- No applicable CLAUDE.md violations (no new env vars, i18n strings, or React components touched).

Solid migration with unusually thorough test/verification coverage in the PR description itself. Nothing blocking from my side.

---

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Confirmed and fixed in 19d1f0d4d — thanks, this was a real gap and it undercut the PR's own claim.

Reproduced against zod 4.3.6 before changing anything, and it's slightly worse than described: the typo isn't just dropped, it leaves `filter.status` as `{}`, so the parsed filter is *present but empty* and `list_surveys` returns every survey while reporting success. Exactly the `count_feedback_records`/`userId` shape, one level down.

Now strict, with your list plus one you didn't have to find:

- `ZMcpListSurveysInput.filter` and its `name` / `status` / `type` children
- `ZMcpSurveyLanguageInput` (structured sub-schema of `create_survey.languages`)
- `ZMcpListWorkflowsInput.filter.{name,status}` and `ZMcpListWorkflowRunsInput.filter.status`

The feedback-record filters need nothing — checked, their fields are scalars/arrays with no nested objects, so the top-level `.strict()` already covers them.

One wrinkle worth flagging for anyone doing this again: the nested ones use `z.strictObject({...})` rather than appending `.strict()`, because `.strict()` returns a clone that **drops `.describe()`** — appending it after a `.describe()` silently deletes the description the model reads. That trap already cost this branch the `create_workflow` schema description (fixed in 9d12f2416). Verified the `filter` and `filter.status` descriptions still reach the advertised schema after this change.

Two new test cases cover a misspelled nested key and an unknown key on `filter` itself, both through a real `tools/call`; fault-injected to confirm they go red when the `filter.status` strictness alone is reverted.

You were also right that the doc note was the thing that let this through — it framed nested laxity as being only about the deliberately free-form `z.record` fields. Rewritten to say the opposite: structured sub-objects are strict too, and only `blocks`/`metadata`/`welcomeCard`/`data` accept an arbitrary nested shape.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Right on both counts — fixed in 19d1f0d4d. Both comments were mine, left in place by the earlier commits while the change underneath them inverted what they described, which is the stale-prose class rather than a nit as far as I'm concerned: the next reader would have trusted them.

- `schemas.ts` (create-record input): now says the operations layer enforces the value/field_type rule, and notes that the strip-to-missing-value path it used to describe is gone — a misspelled key is rejected by the SDK before the handler runs, so the operations layer only ever sees declared keys.
- `feedback-records.ts` (the spread filter): the spread is still safe, but for the opposite reason to the one written down. Now: safe because the schema is `.strict()`, so an undeclared key is rejected before the handler rather than spread onward, and the operation allowlists what reaches the Hub regardless.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Follow-up on this one, because your finding had a bigger tail than either of us checked.

Your point was that `.strict()` binds one object, so nested structured objects need it too. I fixed the filters and then asserted in my reply that structured sub-objects were strict. That claim was wrong, and I only found out by measuring instead of reading. Walking the JSON Schema each tool actually advertises (`z.toJSONSchema(schema, { io: "input" })`) rather than the Zod source:

| | strict nodes | free-form `z.record` | **open structured** |
| --- | --- | --- | --- |
| the other 23 tool schemas | 33 | 11 | **0** |
| `ZMcpCreateWorkflowInput` | 1 | 3 | **15** |
| `ZMcpPatchWorkflowInput` | 2 | 3 | **15** |

Everything below `definition` on the two workflow tools is open: `definition` itself, each trigger/node union member, every node's `config`, every `ui.position`, `edges[]`, and the if/else condition group — that last one reachable only through a `$ref`, so a walker that doesn't follow `$defs` never sees it. Same failure mode you described, in a mutation this time: a misspelled key inside a node's `config` is dropped and `create_workflow` reports success.

Not fixed in this PR, deliberately. `definition` is `ZWorkflowDefinition` in `packages/workflows`, which the v3 Workflows REST route parses and the workflow builder posts — the builder sends `ui` metadata and would have to be checked first, so strictness there is a v3 API decision, not an MCP migration one. That's the same reasoning behind the MCP-only strict variant for batch feedback records rather than tightening the shared v3 body. Filed as ENG-2437 with the measurements and a suggested order of work.

What is in this PR is the thing that should have caught both rounds: `apps/web/modules/mcp/tools/schemas.test.ts` walks every advertised tool schema to any depth, follows `$defs`, and distinguishes three states — strict, deliberately free-form `z.record`, and open structured object, which is the only one that's a bug. The two workflow schemas are excluded by name, and that exclusion asserts the hole stays inside the `definition` subtree, so a new lax object on a top-level workflow argument still fails. Fault-injected both ways before trusting it: reverting the `filter.status` strictness (your original case) and dropping `.strict()` from `ZMcpPatchWorkflowInput` each turn it red with the offending path named.

Two rounds of this were found by a human reading the diff. That's what the guard is for.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Reopening this briefly: your point generalised further than the two comments you pointed at, and I missed a third instance in the same sweep.

You flagged prose that described the strip-and-fall-through behaviour after the diff had inverted it. I fixed the two you named. Grepping for the claim rather than the file turned up more of it, all in the v3 feedbackRecords layer this PR touches:

- `apps/web/app/api/v3/feedbackRecords/lib/schemas.ts` (filter header) — said "the SDK validates against its own non-strict copy of the shape and strips unknown keys before an operation sees them". That was the whole justification for why the filter *names* mattered more than the `.strict()` there, and it's now false: the SDK validates the strict instance itself.
- Same file, create-body header — "MCP strips unknown keys before we ever see them, so `valueText` simply vanishes". The rule it justifies is still needed, but for the v3 REST routes, which parse that shape directly and do still ignore unknown keys. The MCP tools no longer reach it. Rewritten to attribute it correctly rather than deleted, since deleting it would have made the refinement look unmotivated.
- `apps/web/app/api/v3/feedbackRecords/lib/operations.test.ts` — the comment on the `field_type`/`value_*` test, plus the inline one calling `valueText` "what an agent typo looks like after MCP has stripped it". Not in this PR's diff, but made wrong by it, so same class.

`grep -rn "MCP strips\|MCP has stripped"` is now empty. The remaining mentions of stripping are past-tense history (why `.strict()` was decorative under v1), which is still accurate.

Worth saying plainly: this class was the more valuable of your two findings. The nested-filter gap was a bug in the code; this was a comment actively telling the next reader the opposite of what the code does, which is how the nested gap got written in the first place. On that thread I've now had to correct my own overclaim for the same reason — the docs page still said "strictness binds the top level only" after I'd told you I'd rewritten it, because I'd only fixed the code comment and not the handbook. Both are fixed, and there's a test now instead of a promise.

xernobyl:
Ready for another look — @mattinannt @Dhruwang.

One new commit since the last round (`c098bd5d6`), which came out of re-auditing my own answer to @mattinannt's nested-strictness finding. I'd replied that structured sub-objects were strict; measuring instead of reading showed they weren't everywhere. Details on that thread, short version here:

- **New guard**: `apps/web/modules/mcp/tools/schemas.test.ts` walks the JSON Schema every tool actually advertises, to any depth, following `$defs`, and fails on any open structured object. Fault-injected both ways (nested filter, and a workflow tool's top-level object) so it demonstrably can go red.
- **What the audit found**: everything below `definition` on `create_workflow`/`patch_workflow` is still open — 15 nodes each, versus 0 across the other 23 tool schemas. Not fixed here on purpose: that's `ZWorkflowDefinition` in `packages/workflows`, shared with the v3 Workflows REST route and the builder, so it's a v3 API change rather than an MCP one. Filed as **ENG-2437**; the guard pins the hole to that subtree.
- **Prose corrections**: the handbook still claimed "strictness binds the top level only" — the exact framing that let the nested gap through — and two comments in the v3 feedbackRecords layer still credited MCP with stripping unknown keys. Both fixed; `grep -rn "MCP strips"` is clean.
- **PR description** updated: the strictness claim is now scoped honestly, the size-check question is answered, and the suite numbers are current (610 files / 8056 tests).

Local: typecheck ✅, eslint ✅, prettier ✅, full web suite ✅. CI was green on the previous head; the two failures you may have seen earlier were GitHub codeload 429s in `Set up job`, not the diff.

Nothing else outstanding from my side.

xernobyl:
@coderabbitai review

apps/web/modules/mcp/tools/schemas.ts
```diff
@@ -139,72 +194,84 @@ const datasetIdField = ZId.optional().describe(
...
+      .describe(
+        `Feedback records to create, 1–${MAX_FEEDBACK_RECORDS_PER_BATCH} per call. Every record is validated before any is written, so an invalid record fails the whole call rather than storing part of the batch.`
+      ),
+  })
+  .strict();
 
 // The plain field object again (not the refined one): `inputSchema` needs a raw shape. The
 // at-least-one-field rule is enforced by the operations layer.
 export const ZMcpUpdateFeedbackRecordInput = ZV3FeedbackRecordUpdateBodyFields.extend({
   workspaceId: ZId.describe("Workspace ID that owns the feedback record."),
```
kiln-agent, line 277: **`update_feedback_record` now breaks the exact round-trip this PR declines to break at the v3 layer.**

`ZV3FeedbackRecordCreateBodyStrict` in `app/api/v3/feedbackRecords/lib/schemas.ts:419-425` is a separate export specifically so the shared body stays lenient, and the stated reason is:

> immutable provenance fields on an update are all silently dropped today, each with a test asserting exactly that. […] a client that round-trips a record it read would start getting 422s

That reasoning is sound, but `.strict()` here reintroduces precisely that outcome on the MCP side. `ZV3FeedbackRecordUpdateBodyFields` is a `.pick()` of eight mutable fields; a record returned by `get_feedback_record` carries `source_type`, `source_id`, `source_name`, `field_id`, `submission_id`, `collected_at` and the derived enrichment fields. So the natural agent loop — read a record, change `value_text`, send it back — went from "provenance silently ignored" to a hard input-validation failure.

That may well be the intent (it *is* the ENG-2256 semantics, and the breaking-change table covers undeclared keys generally). But the two decisions sitting one file apart reach opposite conclusions about the same client behaviour, and only one of them is written down. If it's deliberate, the breaking-change table is the place to say so — read-modify-write against `update_feedback_record` is a materially more likely client pattern than an invented key, and an agent hitting it has no way to know which of the fields it echoed back was the problem. If it isn't, `ZV3FeedbackRecordUpdateBodyFields.extend({...})` without `.strict()` (or a `.strip()` on the provenance subset) keeps the typo protection where it matters — `user_id` / `value_*` — without rejecting the record's own fields.

---

apps/web/modules/mcp/tools/schemas.ts
```diff
@@ -3,134 +3,189 @@ import { ZId } from "@formbricks/types/common";
...
-});
+export const ZMcpDeleteSurveyInput = z
+  .object({
+    surveyId: z.cuid2().describe("Survey ID to delete."),
+  })
+  .strict();
 
 // list_workspaces takes no arguments — it returns the workspaces the authenticated caller can access.
-export const ZMcpListWorkspacesInput = z.object({});
+export const ZMcpListWorkspacesInput = z.object({}).strict();
```
kiln-agent, line 198: **A strict empty object is the one place strictness has an outsized blast radius.**

`list_workspaces` now advertises `{"type":"object","properties":{},"additionalProperties":false}`. Some MCP clients synthesize a placeholder argument for zero-argument tools rather than sending `{}` — Cursor's well-known `random_string` dummy property is the case that gets reported most, and it's a client-side workaround for tools whose schemas have no properties, so it lands exactly here. Under v1 that placeholder was dropped and the call succeeded; now it fails validation.

The blast radius is what makes this worth a second look rather than the general breaking-change note: `list_workspaces` is the discovery tool, and every other workspace-scoped tool needs a `workspaceId` from it. A client that can't call it loses the whole server, not one tool — and the failure looks like "the Formbricks MCP server is broken", not "my client added a key".

Two cheap options: leave this one schema non-strict (there is nothing to typo-protect in an empty object, so `.strict()` buys no ENG-2256 coverage here — it's consistency only), or keep it and add a QA step that connects a non-Claude client. The 2026-07-28 / legacy matrix in the QA plan is thorough on protocol era but only ever exercises Claude Code, so a client-quirk regression like this one wouldn't be caught by any step currently listed.

---

apps/web/modules/mcp/tools/workflows.test.ts
```diff
@@ -548,8 +563,25 @@ describe("MCP scope enforcement (ENG-1967)", () => {
...
+   * that drops `.describe()`. Appending it for consistency with the sibling schemas would keep validation
+   * identical while silently deleting the contract's own description from `tools/list` — a loss nothing
+   * else here would catch, since a strictness assertion passes either way.
+   */
+  test("carries the contract description and still rejects unknown keys", () => {
+    const { tools } = createToolServer();
+    const schema = tools.get("create_workflow")!.config.inputSchema as ZodObjectLike;
+
+    expect(schema.description).toBe("Creates a draft workflow.");
+    expect(schema.safeParse({ workspaceId: "abc", name: "x", definition: {}, bogus: 1 }).success).toBe(false);
```
kiln-agent: **This assertion is vacuous — it passes without the strictness it claims to test.**

`definition: {}` is already invalid on its own, so the parse fails before `bogus` is ever considered:

```
ZCreateWorkflowInput.safeParse({ workspaceId: "abc", name: "x", definition: {} })
→ success: false
  issues: ["definition.trigger:invalid_type", "definition.entryNodeId:invalid_type"]
```

`bogus: 1` contributes nothing to the result, and a non-strict clone of the schema returns `success: false` for the same input. The test name promises "still rejects unknown keys" but nothing here is sensitive to unknown-key handling.

This is the same false-positive shape the PR fixes elsewhere, and the docblock above states the reason it matters: strictness on this schema is *not* asserted anywhere else (`ZMcpCreateWorkflowInput` deliberately skips `.strict()` to preserve `.describe()`, and `schemas.test.ts` excludes both workflow tools from the general case). So this is the only guard that `ZCreateWorkflowInput`'s `z.strictObject` base hasn't been swapped for a plain `z.object` — and it doesn't guard it.

Fix is a valid `definition` plus the extra key:

```ts
const definition = { trigger: /* … */, entryNodeId: "…", nodes: [], edges: [] };
expect(schema.safeParse({ workspaceId: "abc", name: "x", definition }).success).toBe(true);
expect(schema.safeParse({ workspaceId: "abc", name: "x", definition, bogus: 1 }).success).toBe(false);
```

The positive case is the part that makes the negative one mean something — without it, any future change that breaks `definition` parsing turns this back into a tautology silently. (There are valid definition fixtures in the workflows test suites to borrow from.)

---

apps/web/modules/mcp/tools/schemas.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  test.each(allSchemas.filter(([name]) => SCHEMAS_WITH_OPEN_WORKFLOW_DEFINITION.includes(name)))(
+    "%s is open only inside the shared workflow definition",
+    (_name, schema) => {
+      const open = openObjectPaths(_name, schema);
+
+      // The known gap is bounded: it must stay confined to the `definition` subtree (or a `$defs` entry,
+      // which exists only because a definition sub-schema is referenced twice). A new open object on a
+      // *top-level* tool argument is a new bug and fails here.
+      const outsideDefinition = open.filter(
+        (path) => !path.includes(".definition") && !path.includes(".__schema")
```
kiln-agent: **The `.__schema` exclusion is broader than the comment above it claims.**

The comment says a `$defs` path "exists only because a definition sub-schema is referenced twice", and that "a new open object on a *top-level* tool argument is a new bug and fails here". The first is true of the tree as it stands today; the second doesn't follow from it, because the filter drops every `__schema`-reaching path unconditionally rather than only those that also pass through `definition`. Anything hoisted into `$defs` — which is where the JSON Schema emitter puts any node referenced more than once, definition-related or not — is excluded regardless of where else it's reachable from.

I re-ran the walk with the exclusion narrowed to `path.includes(".definition") || path.startsWith("<name>.__schema.$defs.<workflow-def-ref>")` and got the same 15-per-tool result, so **there is no live gap** — this is about what the guard will catch later, not what it misses now.

Two consequences worth a line of code:

1. If a future shared sub-schema gets referenced twice and lands in `$defs` while non-strict, the guard stays green even though the open node is reachable from a top-level argument.
2. The `expect(open.length).toBeGreaterThan(0)` premise check weakens once ENG-2437 lands: `open` counts `__schema` paths that `outsideDefinition` filtered out, so it can stay non-zero after the definition subtree becomes strict — which is the exact signal the assertion exists to catch ("if this ever empties, delete this test"). Asserting on the *unfiltered-by-`__schema*` count, or on `open.some(p => p.includes(".definition"))`, keeps that tripwire live.

Both are cheap to tighten now and awkward to notice later, since neither shows up as a failure.

---

apps/web/modules/mcp/tools/schemas.ts
```diff
@@ -139,72 +194,84 @@ const datasetIdField = ZId.optional().describe(
...
+      .array(ZV3FeedbackRecordCreateBodyStrict)
+      .min(1)
+      .max(MAX_FEEDBACK_RECORDS_PER_BATCH)
+      .describe(
+        `Feedback records to create, 1–${MAX_FEEDBACK_RECORDS_PER_BATCH} per call. Every record is validated before any is written, so an invalid record fails the whole call rather than storing part of the batch.`
+      ),
+  })
+  .strict();
 
 // The plain field object again (not the refined one): `inputSchema` needs a raw shape. The
```
kiln-agent: Stale invariant: `inputSchema` no longer needs a raw shape — removing that requirement is what this PR does (tools now register schema instances, which is what makes `.strict()` load-bearing at all). The equivalent comment above `ZMcpCreateFeedbackRecordInput` was updated; this one and the batch note around line 230 still cite the v1 rule as the reason for the plain-vs-refined split.

The split is still correct, just for a different reason — the refinement is enforced in the operations layer — so it's the justification that's wrong, not the code. Same stale claim also survives at `app/api/v3/feedbackRecords/lib/schemas.ts:443` (`/** The plain object form first, because `inputSchema` needs a raw shape rather than a refined schema. */`), which is worth catching in the same pass since the PR already fixed the other two "MCP strips unknown keys" comments in that file.

---

kiln-agent (commented):
## Code review — MCP protocol v2 migration + ENG-2256

Reviewed at `c098bd5`, all 23 files. I verified the load-bearing claims empirically against the installed packages rather than by reading the diff, because most of them are assertions about SDK internals that a static read can't settle.

**Verified correct:**

- `ctx.http.authInfo` really is where SDK v2 puts auth (`@modelcontextprotocol/server` 2.0.0, `src-CX2iR2pK.mjs:6443`), and `@posthog/mcp` 0.11.6 forwards that same ctx to `identify`. Confirmed end-to-end with a live `tools/call` through a real handler: `identify` received `{sessionId, mcpReq, http:{authInfo}}`, and `$identify` / `$mcp_tool_call` both fired. The v2 `McpServer` also satisfies `isCompatibleServerType`, so the instrumentation isn't silently no-oping — worth stating, since the "does PostHog still receive events" risk in the description is partly answerable now: the events fire and carry the right identity. Whether they still *group* per session under a sessionless protocol is the part that genuinely needs a post-deploy look.
- `maxSubscriptions: 0` survives the SDK's option handling (`options.maxSubscriptions ?? DEFAULT_MAX_SUBSCRIPTIONS`, gate is `open.size >= maxSubscriptions`), and mcp-handler 2.1.1 forwards it. `0` disables, as claimed.
- `legacy` defaults to the stateless fallback inside the SDK, so the new tests calling `createMcpHandler(...)` **without** `legacy: "stateless"` still match production. That's a non-obvious way for a dual-era test to be quietly meaningless, and it holds.
- The `.strict()`-drops-`.describe()` claim is real (`.strict()` clones and loses the registry description), no schema in the diff applies them in the wrong order, and field-level descriptions survive.
- Strictness holds at every depth in the schemas **actually advertised**: I ran the SDK's own `~standard.jsonSchema.input()` path over all 26 schemas with an independent exhaustive walker and got exactly the 15 open nodes per workflow tool that the description documents. The new test's narrower walker misses nothing in the current tree.
- 346 tests across `modules/mcp`, `app/api/mcp` and `app/api/v3/feedbackRecords` pass locally.

**Five findings inline**, most significant first. None of them is a defect in the migration itself — that part is solid, and the ENG-2437 scoping is the right call. Summary:

1. `schemas.ts:252` — `.strict()` on `update_feedback_record` reintroduces exactly the read-modify-write breakage that `ZV3FeedbackRecordCreateBodyStrict` exists to avoid at the v3 layer. Two decisions one file apart, opposite conclusions about the same client behaviour, only one written down.
2. `schemas.ts:188` — `list_workspaces` advertises an empty strict object; clients that synthesize a placeholder argument for zero-arg tools (Cursor's `random_string`) lose the discovery tool the whole server depends on. Also the one gap in an otherwise very thorough QA matrix: every step exercises Claude Code.
3. `workflows.test.ts:585` — the "still rejects unknown keys" assertion is vacuous; `definition: {}` fails on its own and a non-strict clone returns `success: false` too. It's the only guard on `ZCreateWorkflowInput`'s strict base, and it isn't guarding it.
4. `schemas.test.ts:97` — the `.__schema` exclusion is unconditional, so the stated "a new open object on a top-level argument fails here" doesn't hold for `$defs`-hoisted nodes, and the `open.length > 0` tripwire stops firing correctly once ENG-2437 lands. No live gap; both are latent.
5. `schemas.ts:249` — stale "`inputSchema` needs a raw shape" comment (the invariant this PR removes), plus the twin at `app/api/v3/feedbackRecords/lib/schemas.ts:443`.

Nothing blocking. 1 and 2 are client-visible and worth a decision; 3–5 are cheap and awkward to notice later.

One note on the description, since it's unusually careful about what it does and doesn't claim: the "every argument is validated, except `definition`" framing now matches what I could measure, which was not true of the earlier revision. The self-correction on nested strictness — and the guard that pins it — is the reason the remaining findings are this small.

---

apps/web/modules/mcp/tools/workflows.test.ts
```diff
@@ -548,8 +563,25 @@ describe("MCP scope enforcement (ENG-1967)", () => {
...
+  });
```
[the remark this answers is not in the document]
xernobyl: Confirmed and fixed. This was a real defect, and the sharpest of the three.

`{ workspaceId: "abc", name: "x", definition: {} }` fails three ways before `bogus` is even considered — `abc` is not a cuid2, and `definition: {}` misses `schemaVersion`, `trigger`, `nodes`, `edges` and `entryNodeId`. So `.success === false` was guaranteed regardless of strictness, and the assertion proved nothing.

Fixed by asserting both directions against a fully valid create-workflow payload:

```ts
expect(schema.safeParse(VALID_CREATE_WORKFLOW_INPUT).success).toBe(true);
expect(schema.safeParse({ ...VALID_CREATE_WORKFLOW_INPUT, bogus: 1 }).success).toBe(false);
```

Fault-injected to confirm it can now fail, which the old version could not: patching `ZCreateWorkflowInput` from `z.strictObject` to `z.object` in the built `@formbricks/workflows` chunk turns it red (and independently turns the new depth guard red too). Restored afterwards.

Worth noting this is the third time this exact shape has appeared in this PR — a rejection assertion whose payload was invalid for unrelated reasons. It bit the batch feedback-record test earlier (an invalid `field_type` was doing the rejecting, not the unknown key), and I caught it there by the same method. Good catch.

apps/web/modules/mcp/tools/schemas.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+}
+
+function openObjectPaths(name: string, schema: z.ZodType): string[] {
+  const open: string[] = [];
+  const freeForm: string[] = [];
+  // `io: "input"` matches how the tool schemas are advertised — the same conversion the SDK performs.
+  collectObjectNodes(z.toJSONSchema(schema, { io: "input", unrepresentable: "any" }), name, open, freeForm);
+  return open;
```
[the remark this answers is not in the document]
xernobyl: Split verdict: the second half was a real hole, the first half was not — but it pointed at a better fix than the one suggested, so both are addressed.

**`freeForm` being discarded — real, fixed.** You're right that a structured object flipped to `additionalProperties: true` would have sailed through: it left `open`, so nothing looked at it. The free-form set is now pinned per schema in `EXPECTED_FREE_FORM_PATHS` and compared exactly, so widening a structured object is as loud as leaving it open. It earned its place immediately — my first version of the list wrongly included `ZMcpCreateSurveyInput.languages[]`, and the test failed on it. That path is `ZMcpSurveyLanguageInput`, a `z.strictObject`, and it does not widen. Also fault-injected deliberately: switching it to `z.looseObject` turns the test red with that path named.

**`if`/`then`/`else` — not reachable today.** I probed every advertised schema for `if`, `then`, `else`, `prefixItems`, `contains`, `patternProperties` and `not`. The only hit was `contains`, and that is a false positive: `filter.name.contains` is a *property named* `contains`, not the JSON Schema keyword. Zod's converter does not emit conditional keywords for any construct these schemas use, so adding traversal branches for them would be dead code — untestable, and the sort of thing that rots.

So rather than extend the keyword list, I inverted the walker: it now recurses into **every** value and only special-cases the keys whose names belong in the reported path (`properties`, `$defs`, `items`, `additionalProperties`). Keys that hold no schema (`required`, `type`, `description`) contain strings, so recursing into them finds nothing.

That fixes the general case instead of one instance of it. A keyword-driven walker is only ever as complete as its list — and the first version of this file proved that by missing `$defs`, which is exactly where the workflow if/else condition group was hiding. With generic recursion, anything the generator emits now or later is walked without the test needing to know it exists.

apps/web/modules/mcp/server.test.ts
```diff
@@ -29,6 +29,21 @@ vi.mock("@formbricks/logger", () => ({
...
+   * connection closed, where the default (1024 per process) accepts it and holds it open.
+   *
+   * Worth pinning even so, because dropping the option is invisible: nothing fails, no test goes red,
+   * the stream simply starts being accepted again.
+   */
+  test("refuses subscription streams", () => {
+    expect(MCP_HANDLER_OPTIONS.maxSubscriptions).toBe(0);
+  });
+});
+
```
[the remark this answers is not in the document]
xernobyl: Fair, and fixed — the critique holds: the test would have passed if the option stopped reaching the handler, which is precisely the regression worth catching.

There is now a behavioural test through the real route in `app/api/mcp/route.test.ts`, driving `subscriptions/listen` end to end and asserting `-32603 Subscription limit`. Building it turned up two things about the 2026 revision that are worth recording, since both are easy to get wrong:

1. The full `_meta` envelope is required on **every** request — `protocolVersion`, `clientCapabilities` *and* `clientInfo`. Sessionless means what `initialize` used to carry once now rides on each call; omit any part and you get `-32602 Invalid _meta envelope`.
2. The method must also appear in an `Mcp-Method` header, and the SDK rejects the call if header and body disagree.

More importantly, the first draft of this test was itself an instance of the problem you flagged on `workflows.test.ts`. With an incomplete `params`, it passed — but only because the subscription cap is checked *before* param validation, so the assertion held with the cap removed. Fixed by sending a real `notifications: { toolsListChanged: true }` filter. With a valid payload and `maxSubscriptions` removed, the stream is genuinely accepted and held open, and the test fails on a timeout — slower than an assertion failure, but red, and red for the right reason. Verified by injecting exactly that.

I kept the `MCP_HANDLER_OPTIONS.maxSubscriptions` assertion alongside it rather than replacing it. The two fail differently and both are cheap: the options pin names the intended value where it is declared and fails fast and legibly, while the route test is the one that proves the SDK honours it. Its comment now points at the route test instead of at a manual verification I did once by hand.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Real, and the best finding on this PR. Confirmed empirically rather than by reading — a record echoed back from `get_feedback_record` with `value_text` edited:

```
success=false
issues[0] = {
  code: "unrecognized_keys",
  keys: ["source_type","source_id","source_name","field_id",
         "field_type","field_label","submission_id","collected_at"]
}
```

Your framing was exactly right about the description too, and it's worse than you had it: the description doesn't merely fail to warn, it *steers* callers into the trap — "to add a key you must send the existing keys too (fetch the record first with get_feedback_record)". So the tool tells an agent to fetch the whole record and then rejects it for containing what it contains.

**Keeping strict, and here's the reasoning written down**, since your central point was that only one of the two decisions was.

The alternative fails in the ENG-2256 direction on the tool where it costs most. Drop `.strict()` here and a misspelled `value_text` is discarded, the "at least one field" rule sees the remaining declared fields, and the correction silently doesn't happen — a lost edit reported as success. That's the whole bug this PR exists to close, on a destructive tool.

It's only defensible because the failure is *actionable*, so that's now asserted instead of assumed:

- `update_feedback_record`'s description tells callers to send only the mutable fields and to strip `source_*`, `field_*`, `submission_id`, `collected_at` and the derived enrichment fields before sending, and says the error names every key to remove.
- A new test in `feedback-records.test.ts` pins both halves: the echoed record is rejected with all eight keys named, **and** the same edit with provenance stripped is accepted — so the rejection provably comes from those keys and not from something else. Fault-injected by removing `.strict()`; it goes red.
- The handbook note and the PR's breaking-change table now call out read-modify-write specifically rather than leaving it under "undeclared keys generally".

On the inconsistency you identified: it's deliberate, and now stated at both sites. The v3 REST body stays lenient because a REST integration round-tripping a record it read would start getting 422s with no way to discover why. An MCP client re-reads the advertised schema and the error text on every call, so the loud version costs it one retry. Different clients, different call — but you're right that an undocumented split between two files one directory apart is indistinguishable from an oversight.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Your factual core is right and I hadn't weighed it: `.strict()` on `z.object({})` buys **zero** ENG-2256 protection. With no declared keys there is nothing to misspell, so this one schema is uniformity only — cost without benefit, unlike every other schema here.

The blast-radius argument is also right. `list_workspaces` is where every workspace-scoped tool gets its `workspaceId`, so a client that can't call it sees a dead server, not a dead tool.

**Leaving it strict**, for two reasons, but recording the trade so it reads as a decision rather than an oversight:

1. Uniform strictness was an explicit call on this PR, not a default that fell out — "`.strict()` on every schema, no per-tool judgement to maintain". Reversing it for one schema reintroduces exactly the per-tool exception list that choice was meant to avoid, and the next person adding a zero-argument tool would have to rediscover the reasoning.
2. I can't confirm the client behaviour. The well-known Cursor `random_string` business is a *server-side* workaround — authors add a dummy property because some clients mishandle empty-parameter schemas — which is not the same claim as a client inventing a key the schema doesn't declare. I'm not willing to weaken a security-relevant default on a quirk I haven't reproduced.

What I have done instead:

- A comment at `ZMcpListWorkspacesInput` stating plainly that strictness buys nothing here, that it's kept for uniformity, and that a report of "the Formbricks MCP server won't connect" from a non-Claude client should look here first.
- The handbook note now calls out that `list_workspaces` takes no arguments at all, so a client padding zero-argument calls fails on the discovery tool.
- Your QA point stands and I've taken it: the matrix is thorough on protocol era but only ever exercises Claude Code, so no listed step would catch a client-quirk regression.

Flagging this to @xernobyl as the one genuinely open decision on this PR rather than closing it myself — if you'd rather not carry the risk on the discovery tool specifically, dropping `.strict()` from this single schema costs nothing in coverage and I'll make the change.

apps/web/modules/mcp/tools/workflows.test.ts
xernobyl: Correct, and already fixed in `c248bc918` — CodeRabbit raised the same thing independently, so you're both anchored on the previous head. Your diagnosis matches exactly, including the part that makes it matter rather than being a nit: this is the *only* guard on `ZCreateWorkflowInput`'s `z.strictObject` base, because `ZMcpCreateWorkflowInput` deliberately skips `.strict()` to keep `.describe()`, and `schemas.test.ts` excludes both workflow tools from the general case.

Fixed as you prescribed, positive case included:

```ts
expect(schema.safeParse(VALID_CREATE_WORKFLOW_INPUT).success).toBe(true);
expect(schema.safeParse({ ...VALID_CREATE_WORKFLOW_INPUT, bogus: 1 }).success).toBe(false);
```

The fixture is borrowed from `packages/workflows/src/contracts/index.test.ts` as you suggested — trigger, one `send_email` action node, an edge connecting them, `entryNodeId`.

Fault-injected to confirm it can now fail, which the old version could not: patching `ZCreateWorkflowInput` from `z.strictObject` to `z.object` in the built `@formbricks/workflows` chunk turns it red. Restored after.

Your point about the positive case is the one I'd underweighted — without it a future change that breaks `definition` parsing turns the test back into a tautology *silently*, which is exactly how it got into this state.

Marking this resolved as fixed. Same defect shape appeared three more times on this PR (the batch feedback-record test, and my own first draft of the `subscriptions/listen` test, which passed with incomplete params only because the subscription cap is checked before param validation) — all now assert against valid payloads.

apps/web/modules/mcp/tools/schemas.test.ts
xernobyl: Both points correct, and I appreciate you separating "no live gap" from "won't catch it later" — that's what made it obvious this was worth fixing properly rather than waving through.

**Point 1, the unconditional `$defs` exclusion.** My first instinct was to make the problem disappear rather than solve it: `z.toJSONSchema` takes `reused: "inline"`, which stops hoisting reused sub-schemas into `$defs` and gives every node its real path. That turned out to be only half a fix — with inlining on, `__schema0` *still* appears, because the if/else condition group is **recursive**, and a recursive schema has to be a `$ref` no matter what `reused` says. So the hoisted entry can't be inlined away and does have to be reasoned about.

So the guard now proves the premise instead of asserting it in a comment. The walker records every `$ref` and where it occurred, `$defs` entries are pathed as `$defs:<name>` rather than pretending to sit at the root, and a `$defs` node is excusable only if *every* reference to it is inside `definition`:

```ts
const reachedOnlyViaDefinition = (path: string): boolean => {
  if (!path.startsWith(DEFS_PREFIX)) return false;
  const def = path.slice(DEFS_PREFIX.length).split(/[.[|{]/)[0];
  const referencedFrom = refs.filter((ref) => ref.def === def).map((ref) => ref.at);
  return referencedFrom.length > 0 &&
    referencedFrom.every((at) => at.includes(".definition") || at.startsWith(DEFS_PREFIX));
};
```

The `referencedFrom.length > 0` clause keeps it non-vacuous: a `$defs` entry nothing references is not excused. So your scenario 1 — a future shared sub-schema hoisted while non-strict and reachable from a top-level argument — now fails, because at least one of its refs won't be under `definition`.

`reused: "inline"` is kept anyway: it removes non-recursive reuse from `$defs` entirely, so the ref-checking path only has to handle the genuinely recursive case.

**Point 2, the tripwire.** Right, and this was the more dangerous of the two, because it fails by going quiet. Now:

```ts
expect(open.filter((path) => path.includes(".definition"))).not.toEqual([]);
```

so the count is taken inside the definition subtree specifically. When ENG-2437 lands this empties and the test fails, which is the signal to delete it and the exclusion list — instead of being propped up by an unrelated `$defs` node and never firing.

Re-verified after the rework that the top-level case still catches a regression: dropping `.strict()` from `ZMcpPatchWorkflowInput` turns it red naming the path. Both points were, as you say, cheap now and awkward later — neither would have shown up as a failure.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Right on all three sites, and thanks for including the v3 one — that's the same file I'd already swept for the "MCP strips unknown keys" claims, so I'd clearly stopped looking once the greps I chose came back clean.

Fixed at all three: `schemas.ts` line ~231 (batch note), line ~249 (update input), and `app/api/v3/feedbackRecords/lib/schemas.ts` line ~443.

Worth reporting what happened when I tried to write the replacement, because my first attempt at the "real" reason was also wrong. I assumed the split existed because you can't extend a refined schema. Probed it instead of asserting it:

```
refined.extend = function
refined.extend({...}).strict = function
refined.extend({...}).strict() → valid=true, +bogus=false, refinement still enforced=true
```

So in Zod 4 `refined.extend({...}).strict()` works completely — extendable, strict-able, **and** the `superRefine` survives both calls. There is no constraint here at all. The plain-vs-refined split is a *choice*: the value/field_type rule lives in the operations layer because that's the one path both the MCP tools and the v3 REST routes pass through, so it exists once instead of twice. The comments now say that, and say explicitly that the in-schema version is possible — which is worth knowing, since the batch tool already does it that way via `ZV3FeedbackRecordCreateBodyStrict`, and the asymmetry between the two feedback-record tools would otherwise look accidental.

Left the code alone, per your read that it's the justification and not the behaviour that's wrong. Moving the single-record rule into the schema would change where the error surfaces and what it looks like, which is a behaviour change outside this PR.

This is the third round of this same class on this PR — @mattinannt found two instances, you found these three. A stale comment reads as fact to the next person, and in this case the stale one would have actively discouraged the simplification it claimed was impossible.

xernobyl:
Ready for re-review — @mattinannt @Dhruwang. Head `02ced8f5c`.

@kiln-agent reviewed and found five things. Four were real; the fifth (the vacuous `create_workflow` assertion) CodeRabbit had raised independently and was already fixed. All addressed in `02ced8f5c`.

**The one worth your attention** is a behaviour consequence I hadn't spelled out: `.strict()` on `update_feedback_record` breaks read-modify-write. The update set is eight mutable fields, so fetching a record and sending it back — which the tool's own description encouraged, since it tells you to fetch the record first for metadata — went from "provenance silently ignored" to a hard rejection.

Kept strict, because the alternative loses the fix on the tool where it costs most: a misspelled `value_text` would be dropped and the correction would silently not happen, on a destructive tool. It's only defensible because the failure is actionable, so that's now asserted rather than assumed — Zod names all eight offending keys, a new test pins that *and* that the trimmed payload is accepted, and the description tells callers to strip provenance instead of steering them into it. It's now a row in the breaking-change table and a step in the QA plan.

Also: the depth guard no longer excuses `$defs` paths blindly (it tracks where each `$ref` occurs and only excuses nodes reachable solely via `definition`), its tripwire counts inside the `definition` subtree so it still fires when ENG-2437 lands, and three comments claiming `inputSchema` needs a raw shape are gone — that requirement is what this PR removed. Probing that one showed my replacement reasoning was also wrong: `refined.extend({...}).strict()` works fine in Zod 4 and keeps the refinement, so the plain-vs-refined split is a choice, not a constraint. Comments now say which.

**One open decision I've deliberately not made** — `.strict()` on `list_workspaces` buys zero ENG-2256 protection, since an empty object has no key to misspell. It's the only place the cost isn't matched by a benefit, and the cost lands on the discovery tool every other tool depends on. I've kept it (uniform strictness was an explicit call, and no offending client is confirmed) and documented the risk, but if either of you would rather not carry it, dropping it from that one schema costs nothing in coverage. Thread is left open for that.

Verification on this head: full web suite **610 files / 8082 tests ✅**, typecheck ✅, eslint ✅, prettier ✅. Live smoke against a running server on this branch — 401 unauthenticated with the right `WWW-Authenticate`, 200 resource metadata, 401 bogus key, 400 credential-in-query, 403 cross-origin, `iss` advertised — all as before.

Resolved the threads that are objectively closed. Left open: the `list_workspaces` decision above, and the `update_feedback_record` trade, since that one is a judgement call rather than a fix and you may weigh it differently.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Correction to the table above: the "others" row should read **24 schemas / 35 strict nodes**, not 23 / 33. I mis-transcribed both figures from the audit output when writing this up. Re-counted programmatically:

```
others:                  { n: 24, strict: 35, free: 11, open: 0 }
ZMcpCreateWorkflowInput: { n:  1, strict:  1, free:  3, open: 15 }
ZMcpPatchWorkflowInput:  { n:  1, strict:  2, free:  3, open: 15 }
```

Nothing downstream changes — the load-bearing figures (0 open outside the workflow definition, 15 in each workflow tool, 11 deliberate free-form) were right, and they're what the argument rests on. ENG-2437 and the PR description are corrected.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Smoke-tested this against a real authenticated client on a running server, which settles the part I couldn't answer before.

The advertised schema is what you said it is:

```json
"list_workspaces": { "type": "object", "properties": {}, "additionalProperties": false }
```

And the call works — Claude Code sends `{}`, no placeholder key, and gets its workspace back:

```json
{"data":[{"id":"cmsx7zk860001zkz3d3d4cjkc","name":"My workspace",…}],"meta":{"totalCount":1}}
```

So the discovery tool is confirmed working for the client we actually ship against. That doesn't disprove your concern for other clients — it just means the blast radius is hypothetical rather than live, which is where I'm leaving it: strict, documented at the schema with the reasoning and the "look here first" note, plus a QA step for a non-Claude client and a Risks entry in the PR description flagging it as open.

Resolving the thread since there's nothing left for me to do on it, but the decision is genuinely still open for @mattinannt / @Dhruwang — dropping `.strict()` from this one schema costs nothing in ENG-2256 coverage, and I'll make the change on a word. It's recorded in the PR description rather than only here so it doesn't get lost with the thread.

apps/web/modules/mcp/tools/schemas.ts
xernobyl: Closing this out. Everything actionable from it has landed:

- `update_feedback_record`'s description now tells callers to send only the mutable fields and to strip `source_*`, `field_*`, `submission_id`, `collected_at` and the derived enrichment fields, and says the error names the keys to remove — so the tool no longer steers an agent into the trap it then rejects.
- A test pins both directions: the echoed record is rejected with all eight keys named, and the same edit with provenance stripped is accepted, so the rejection provably comes from those keys. Fault-injected by removing `.strict()`; it goes red.
- The schema comment records the trade and, per your central point, records that it is the *opposite* call to the shared v3 REST body one file over, and why.
- It's a row in the PR's breaking-change table and a step in the QA plan.

Also smoke-tested the filter path live while I was at it, since the ENG-2256 case runs through the same strict-extended schema: `count_feedback_records` with `user_id: "nobody-at-all"` returns `count: 0` rather than the whole-dataset total, which is the original bug's exact symptom now behaving.

One note for the record: the wrong spelling can no longer even be *sent* from a schema-aware client — the advertised `additionalProperties: false` means the client rejects `userId` locally before a request is made. That's the fix working a layer earlier than the server, and it's why the server-side proof lives in the tests rather than in a live call.

Resolving. The trade itself stays visible in the breaking-change table if @mattinannt or @Dhruwang wants to weigh it differently.

xernobyl:
Re-review please — @mattinannt @Dhruwang. Head `1c9b7893d`, **all 10 review threads resolved**, CI green (21/21).

No reviewer comments outstanding. Since the last round:

**Found one thing on a self-review, and it was my own.** The comment I'd just written to explain `reused: "inline"` claimed every schema node reports the path it is really reachable from. That's what I set out to achieve and then discovered is false — `reused` only inlines plain reuse, and a *recursive* schema must stay a `$ref` regardless. The workflow if/else condition group is exactly that, which is the entire reason the ref-tracking below it exists. So the comment contradicted the code directly beneath it. Fixed, along with a docblock my scripted edit had orphaned onto a constant instead of the function it describes. `1c9b7893d`, comments only — no assertion changed.

That's the **fifth** instance of the same class on this PR (two from @mattinannt, three from the bots, one self-caught), and the first one I introduced *while fixing* the others. Noting it because it's the thing to watch in review here: the code has been checked hard, the prose around it less so.

**Smoke test on the final head**, now in the PR description as evidence: the five auth paths (401 + challenge, 200 metadata, 401 bogus key, 400 credential-in-query, 403 cross-origin) plus two authenticated calls from a real client:

- `list_workspaces` advertises `{"type":"object","properties":{},"additionalProperties":false}` and works with `{}` — so the strict empty schema @kiln-agent flagged does not break the discovery tool for the client we ship against. Blast radius is hypothetical, not live.
- `count_feedback_records` with `user_id: "nobody-at-all"` returns `count: 0` rather than the whole-dataset total — ENG-2256's original symptom, behaving.

**Two decisions are recorded rather than settled**, both in the breaking-change table and Risks so they don't vanish with their resolved threads:

1. `.strict()` on `list_workspaces` buys zero ENG-2256 protection (nothing to misspell in an empty object) and the cost lands on the discovery tool. Kept; dropping it costs nothing in coverage if you'd rather not carry it.
2. `update_feedback_record` now rejects an echoed record instead of ignoring its provenance. Kept, with the failure made actionable and tested, and the opposite call for the v3 REST body documented alongside it.

Verification: full web suite **610 files / 8082 tests ✅**, typecheck ✅, eslint ✅, prettier ✅, CI 21/21 ✅.

Dhruwang (approved):

PR #8861, opened 2026-08-14
fix(security): scope response file deletion to the owning workspace [ENG-2291]

pandeymangg:
## What & why

When a response is deleted, its uploaded files are removed from storage. The storage key for
each file was rebuilt from the id in the **file URL** rather than re-resolved against the survey's
own workspace. Write-time validation pins uploaded URLs to the survey's workspace, but it only
inspects response keys that match a file-upload element present at write time — so a URL stored
outside that check could later be treated as a real answer and deleted. The only guard on the
delete path was a `..` traversal check on the filename.

This change makes the delete side authoritative instead of trusting the URL:

- New shared helper `deleteResponseFileUrls` re-resolves each file URL's storage id against the
  owning survey's workspace (via `findWorkspaceByIdOrLegacyEnvId`, so a legacy environment-id
  prefix that maps back to the same workspace still works) and **skips anything that resolves
  outside the survey's workspace**.
- Both response-deletion sites — the legacy `lib/response/service.ts` path and the v2 management
  path — now go through that one helper. No caller-supplied storage id is trusted anymore.

Recovers the delete-side scoping that was written in abandoned PR #8667 (closed when that
large review branch was split by severity, and this hunk was dropped on the assumption the
write-side binding already covered it).

Details are in the linked Linear ticket, kept internal per `SECURITY.md` §III.C.

## Linear ticket

Fixes ENG-2291

## How this was tested

- `pnpm test` (apps/web unit suite) ✅
- New unit test `modules/storage/lib/delete-response-files.test.ts` — covers same-workspace
  delete, cross-workspace refusal, legacy env-id prefix, unresolvable id, missing workspace id,
  unparseable URL, and a mixed batch.
- Updated `modules/api/v2/.../[responseId]/lib/tests/utils.test.ts` for the new scoped behaviour,
  plus a cross-tenant refusal case.
- `tsc --noEmit` ✅ · `eslint` ✅ · `prettier --check` ✅

## Breaking changes

None — same public API and delete behaviour for files that belong to the survey's own workspace;
only out-of-workspace deletions (which should never have happened) are now refused.

## QA / Test Plan

**How to test**

- [ ] Create a survey with a **File Upload** question, publish it, and submit a response that
      uploads a file → the file is stored and reachable at its storage URL.
- [ ] Delete that response (UI, or `DELETE /api/v1/management/responses/{id}` /
      `DELETE /api/v2/management/responses/{id}`) → the response is deleted **and** its uploaded
      file is removed from storage (the file URL no longer resolves).
- [ ] Delete a response that has **no** file-upload answers → response deletes normally, nothing
      else is touched.

**Preconditions / test data**

- A workspace with a published file-upload survey and at least one response with an uploaded file.
- A management API key for the API delete paths.

**Risks & regressions**

- Response deletion and its storage cleanup for a survey's own workspace must keep working — the
  fix only refuses deletions that resolve to a different workspace.
- Legacy uploads stored under an environment-id prefix (pre-workspace-rename) must still be
  cleaned up on delete.

- none


apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,63 @@
...
+        const storageWorkspace = await findWorkspaceByIdOrLegacyEnvId(storageFile.storageId);
+        if (storageWorkspace?.id !== surveyWorkspaceId) {
+          logger.error(
+            { fileUrl, surveyWorkspaceId, storageId: storageFile.storageId },
+            "Refusing to delete a response file stored outside the survey's workspace"
+          );
+          return;
+        }
+
+        await deleteFile(
```
mattinannt: `deleteFile`'s return value is discarded here. `deleteFile` returns a `Result` (`apps/web/modules/storage/service.ts:135-154`) that is `err({ code: ... })` on S3 failure — that's silently swallowed, so a storage-layer failure (e.g. `FileNotFoundError` with no fallback match, or a transient S3 error) makes this code path believe the file was cleaned up when it wasn't. The `catch` only covers thrown exceptions, not a returned error result. Suggest checking `result.ok` and logging/reporting the failure the same way the "outside workspace" case does, so failed deletions aren't silently treated as successes.

---

apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,63 @@
...
+  await Promise.all(
+    fileUrls.map(async (fileUrl) => {
+      try {
+        const storageFile = parseStorageFileUrl(fileUrl);
+
+        if (!storageFile) {
+          throw new Error(`Invalid storage file URL: ${fileUrl}`);
+        }
+
+        const storageWorkspace = await findWorkspaceByIdOrLegacyEnvId(storageFile.storageId);
```
mattinannt: Minor/non-blocking: `findWorkspaceByIdOrLegacyEnvId` is called once per URL with no de-duplication by `storageId` and no concurrency cap on the `Promise.all`. In the common case (several files from the same survey response) this fires N identical DB lookups instead of one reused result. It's also worth noting the PR's own threat model — a caller planting arbitrary URLs under a not-yet-file-upload key before flipping it to a file-upload element — means `fileUrls.length` isn't bounded by "real" file uploads; `response.data` values are only validated as `z.array(z.string())` (`packages/database/zod/responses.ts`) with no max length, so a large planted array turns one response deletion into an unbounded fan-out of DB queries. Not something this PR needs to solve, but worth a follow-up if response-data size isn't capped elsewhere.

---

mattinannt (commented):
Reviewed the core change: `deleteResponseFileUrls` re-resolves each file URL's storage id against the survey's actual workspace before deleting, closing the TOCTOU gap described in the PR (a URL planted under a not-yet-file-upload key, then turned into one). I traced both call sites, `parseStorageFileUrl`, `deleteFile`'s primary/fallback semantics, and `findWorkspaceByIdOrLegacyEnvId` — the authorization logic itself is sound and the tests match the implementation.

Two non-blocking findings inline:
- Deletion failures from `deleteFile` are silently discarded (its `Result` return value isn't checked), so a storage-layer failure can look like a successful cleanup.
- No de-duplication/concurrency bound on the per-URL workspace lookup — fine for typical response sizes, but worth a follow-up since `response.data` has no array-length cap.

Neither blocks this security fix from landing.

---

apps/web/modules/storage/lib/delete-response-files.ts
pandeymangg: Fixed in 95e11e445 — the helper now captures the `deleteFile` result and logs on `!result.ok`, the same way the out-of-workspace case logs, so a returned `err(...)` no longer reads as a clean success. Added a test that returns a failure result from `deleteFile`.

apps/web/modules/storage/lib/delete-response-files.ts
pandeymangg: Deduped in 95e11e445 — workspace resolution is now cached per `storageId`, so several files under one prefix do a single lookup (test asserts one resolver call for two URLs under the same id).

The unbounded-array point is real but separate, so I filed it as ENG-2397: `response.data` isn't length-capped, so a large planted array is an unbounded fan-out regardless of the dedup. That ticket covers capping the array size / bounding the concurrency.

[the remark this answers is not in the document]
pandeymangg: Addressed in 95e11e445 — per-`storageId` promise cache, with a test asserting the resolver is called once for two URLs under the same id. The remaining unbounded-array concern (response.data has no length cap) is tracked separately in ENG-2397.

apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,63 @@
...
+          );
+          return;
+        }
+
+        await deleteFile(
+          storageFile.storageId,
+          storageFile.accessType,
+          storageFile.fileName,
+          surveyWorkspaceId
+        );
```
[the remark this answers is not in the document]
pandeymangg: Addressed in 95e11e445 — the result is now checked and logged on `!result.ok` instead of being discarded, with a test that returns a failure result from `deleteFile`.

apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,83 @@
...
+          logger.error(
+            { fileUrl, surveyWorkspaceId, storageId: storageFile.storageId },
+            "Refusing to delete a response file stored outside the survey's workspace"
+          );
+          return;
+        }
+
+        // deleteFile returns an error result (it does not throw) on S3 failures, so a discarded result
+        // would treat a failed deletion as a success and leave the object behind unlogged.
+        const result = await deleteFile(
```
Dhruwang, line 72: `fileName` here is still percent-encoded. Upload puts `encodeURIComponent(name)` in the URL but writes the raw name as the S3 key, and both other readers decode first (`getFileStreamForDownload`, and the storage DELETE route). This one doesn't, so anything with a space or non-ASCII char misses the key and never gets deleted. With the new `!result.ok` logging it'll show up as a constant stream of delete failures.

Decoding here rather than inside `deleteFile` also gets `hasTraversalSegment` to run on the decoded segments:

```ts
const fileName = decodeURIComponent(storageFile.fileName);
```

apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts
```diff
@@ -23,20 +21,7 @@ export const findAndDeleteUploadedFilesInResponse = async (
     .filter(([questionId]) => fileUploadQuestions.has(questionId))
```
Comment (Dhruwang) [inline apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts]: This set is built from `questions` only, while `lib/response/service.ts` builds it from `blocks` only — and a survey can't have both, `surveyRefinement` rejects it. So this path deletes zero files for blocks-based surveys, silently.

`getSurveyQuestions` already selects `blocks`, `response.ts` just doesn't pass them through. `getSurveyFileUploadConfigs` already does the union and is what write-time validation uses.

Pre-existing, not from this PR.

apps/web/modules/storage/lib/delete-response-files.ts
pandeymangg: Good catch — fixed in 7330594b5. Decode `storageFile.fileName` before `deleteFile`, as you suggested, so `hasTraversalSegment` now runs on the decoded segments too. Since both delete paths share this helper, v1 and v2 are covered at once. Added a test with a space + parens in the name.

apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts
Comment (pandeymangg) [reply apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts]: Fixed here in 7330594b5 rather than deferring. Both delete paths now build the file-upload id set from `getSurveyFileUploadConfigs` — the same blocks+questions union write-time validation uses — so block-based surveys' uploads get cleaned up. The v2 path now takes the survey (blocks + questions) and `response.ts` passes both through. Added a block-based deletion test.

Dhruwang (approved):

PR #8862, opened 2026-08-14
ci: drive the breaking-change label from a checkbox (ENG-2332)

Dhruwang:
## What & why

`pr-label-sync.yml` inferred the `breaking-change` label from the *prose* in the PR body's
"## Breaking changes" section. A bare "None" cleared it, but so did only a short aside built entirely
from an allowlisted vocabulary (`docs`, `tests`, `internal`, `only`…) — at most six words. Anything
longer or with one unrecognised word counted as a declaration.

That is why PRs whose section says there is nothing to declare still get labelled: explaining *why* a
change is compatible trips the heuristic. The case that prompted the ticket:

> None — same public API and delete behaviour for files that belong to the survey's own workspace;
> only out-of-workspace deletions (which should never have happened) are now refused.

Correct answer, correctly written, wrong label. The section was carrying two jobs at once — the
machine-readable signal and the human explanation — and prose destroyed the signal.

This PR splits them. The section now holds a checkbox, and that checkbox is the whole signal:

```
```

Ticked → the label is applied. Unticked → it is removed. The prose or table underneath is for humans
and is never parsed, so the author can explain at whatever length the change deserves without
affecting the label. Deciding is explicitly the author's (or agent's) job.

The trade-off is real and deliberate: the old heuristic erred towards a spurious label (cheap — someone
reading the release notes catches it), while a forgotten tick under-reports (expensive — a self-hoster
misses a migration entry). Three things carry that weight instead of the regex:

- the template asks for a decision, with the full list of what counts as breaking right above the box;
- `AGENTS.md` states that the checkbox is a judgement the agent owns and must re-check as the diff grows;
- the CodeRabbit check "Breaking changes match the diff" now reads the same checkbox, so it flags a
  breaking diff with an unticked box (and a ticked box with nothing breaking in the diff). Previously it
  compared the diff against the prose, which would have drifted from the workflow the moment this landed.

One behaviour was added rather than replaced: if the section contains no checkbox at all, the label is
left untouched instead of being treated as "not breaking", so a restructured section or a deliberately
hand-applied label survives.

Each rule matches *its own* checkbox by the label text the template ships, rather than taking the first
checkbox in the section — otherwise an unrelated item above it (`- [x] I read the list above`) would flip
the label. That couples the workflow to the template's wording, so a section holding checkboxes but not
the declaration emits a `core.warning` rather than freezing the label silently: the drift shows up in the
run instead of hiding. The match is a prefix, so appending a clause to the checkbox label keeps working.

Net effect on the engine: the two vocabulary regexes and the word-counting predicate are gone, replaced
by one checkbox match per rule. `RULES` stays generic — `isActive` may now also return `null` for "no
signal".

## Linear ticket

Fixes ENG-2332

## How this was tested

- Extracted the real inline `script:` block from `pr-label-sync.yml` and ran it under Node with stubbed
  `github` / `context` / `core`, over 22 bodies: **22/22 as expected**. Covered the template exactly as
  shipped, ticked (`[x]`, `[X]`, `*` marker, indented, bolded label, trailing clause), unticked with a
  bare "None" and with the long prose from the ticket, label already correct, stale label removed on
  untick, checkbox deleted with and without a hand-applied label, missing section, a `[x]` appearing only
  inside the guidance comment, ticked QA-plan checkboxes in a neighbouring section, CRLF bodies both ways,
  an unrelated ticked checkbox above the declaration in all three label states, and a reworded
  declaration (warns, leaves the label alone). The screenshot case that motivated the ticket now produces
  no label.
- `node -e "yaml.parse(...)"` on both `pr-label-sync.yml` and `.coderabbit.yaml` — parse clean. Only
  string values inside `custom_checks[].instructions` changed, so the published CodeRabbit schema shape
  is untouched (`coderabbit config validate` runs in CI).
- `npx prettier --check` on all four files — clean.
- No unit test committed: the logic lives in an inline `actions/github-script` block with no harness in
  the repo, and adding a YAML-extracting test runner is more machinery than 6 lines of regex warrants.
- Not testable before merge: `pull_request_target` always runs the workflow definition from the base
  branch, so this PR is still graded by the old prose logic — which reads the unticked checkbox line as
  content and will label this PR `breaking-change`. Remove it by hand; it stops happening once merged.

## Breaking changes

None — repo tooling only. No product code, API, SDK, webhook, env var, or migration is touched.

## QA / Test Plan

**How to test**

- [ ] After this merges to `main`, open a PR from a branch with a trivial change and leave the
      "## Breaking changes" checkbox unticked → no `breaking-change` label appears on the PR.
- [ ] Edit that PR body to tick the box (`- [x] This PR contains breaking changes`) and save → the
      `breaking-change` label appears within seconds (the `edited` event fires on a body change).
- [ ] Untick the box and save → the label is removed.
- [ ] Under the unticked box, write a long explanation that starts with "None" and continues for several
      sentences about API shapes and endpoints → still no label. This is the regression the ticket filed:
      the same body applied the label before.
- [ ] Tick the box and replace "None" with the example table from the template → label present, table
      renders. Confirm the text feeds nothing but the release notes read by a human.
- [ ] Apply `breaking-change` by hand, then delete the checkbox line from the section and save → the
      label stays (no checkbox = no signal = no change).
- [ ] Add an unrelated ticked checkbox above the declaration (`- [x] I read the list above`) while the
      declaration stays unticked → still no label; the unrelated box is ignored.
- [ ] Reword the declaration text itself (e.g. to `- [x] Breaking!`) and save → the label is left as it
      was, and the `PR Label Sync` run shows a warning that the template and workflow have drifted.
- [ ] Post-merge sweep: `gh pr list --state open --label breaking-change` → for each PR whose Breaking
      changes section opens with "None", remove the label by hand (7 as of opening this PR: #8864, #8863,
      #8861, #8860, #8857, #8855, #8831). #8859, #8846 and #8835 keep it — they declare real changes.
- [ ] Check the Actions tab for the `PR Label Sync` run on each edit → green, with a log line naming the
      state it decided (`added` / `removed` / `already in the correct state` / `no checkbox … leaving
      unchanged`).

**Preconditions / test data**

- Write access to the repo, plus a throwaway PR. No app, database, or env setup needed.
- Cloud vs self-host is irrelevant — this is repository CI only.
- The `breaking-change` label already exists in the repo (created under ENG-2087).

**Risks & regressions**

- Under-reporting is the new failure mode: a PR with a genuinely breaking change whose author forgets to
  tick the box gets no label, and the release-notes / migration-guide compile (Release QA step 4) would
  miss it. The diff-scan backstop in that step and the CodeRabbit warning are what catch it — worth
  watching over the first release cycle.
- **The existing false positives are not cleared by this PR and need a sweep.** 10 open PRs currently
  carry `breaking-change`, and 7 of them open their section with a prose "None" — #8864, #8863, #8861
  (the case in the ticket), #8860, #8857, #8855, #8831. Only #8859, #8846 and #8835 declare a real one
  with a table. None of those bodies has a checkbox, so `isActive` returns `null` and the label stays put
  by design. If one of the 7 merges labelled, the Release QA step 4 compile puts a tooling chore in the
  migration guide. Editing the body to a bare "None" no longer clears it either — only removing the label
  by hand, or pasting in the new checkbox line unticked, does. Sweep them when this merges.
- The `## QA / Test Plan` section also contains checkboxes. Section extraction stops at the next `##`
  heading, so those are never read as the breaking-change signal — covered by a test case above.

- none. Takes effect on merge to `main`, because `pull_request_target` runs the base-branch definition.


mattinannt (commented):
Reviewed. This replaces the fragile "breaking changes" text-heuristic with an explicit checkbox — traced the full pipeline in `pr-label-sync.yml` (HTML-comment stripping → section extraction → checkbox regex) against the actual `pull_request_template.md`, including a manual run reproducing it end-to-end: correctly returns `false` for the default unticked box, `true` for `[x]`/`[X]`, and leaves the label alone (`null`) when no checkbox is found. Grepped the repo for other consumers of the `breaking-change` label/heuristic — none exist outside these three files, so nothing else depends on the old contract. `.coderabbit.yaml` prose stays consistent with the new template/workflow, and GitHub Actions conventions (minimal permissions, harden-runner) are untouched.

No blocking issues found.

---

mattinannt (commented):
Re-reviewed at `29bff47` (re-review after new commits, requested via review-request).

Diff since my last review (`5c7daca`) is just `pr-label-sync.yml`: `CHECKBOX` (single unnamed first-match regex) → `CHECKBOX_LINE` + `checkboxFor(declaration)`, which now matches the checkbox by its label text (`/^this pr contains breaking changes/i`, prefix, markdown-strip-then-trim) instead of taking the first checkbox in the section.

This resolves both open items from my prior pass:

- **CodeRabbit's 🟠 / my prior finding** — an unrelated ticked box above the declaration (`- [x] I checked the list above`) no longer flips the label; traced `checkboxFor` by hand against that case and the shipped template's exact line (`- [ ] This PR contains breaking changes`) — matches. A section with checkboxes but none matching the declaration now emits `core.warning` instead of silently misreading the wrong box, so template/workflow drift surfaces in the Action run.
- **Risks-section drift** — the PR body's Risks bullet and "How this was tested" sweep list have been corrected to state the false positives need a manual sweep post-merge, matching actual `isActive` behavior (`null` on no checkbox, label stays).

No new issues in this diff. CRLF handling, trim-then-match, and the "no checkbox at all → leave label alone" behavior are all unchanged from the already-reviewed base.

**Checked** correctness of the new match logic, drift warning path, consistency with prior findings · **Not re-verified** the full 22/18-case test matrix (no harness in-repo to re-run; relying on the PR author's stated coverage) · **Not verifiable here** live label behavior on a real PR edit → @Dhruwang

No blocking issues found.

---

mattinannt (commented):
**No new findings** — re-reviewed at `96d00af` (single-pass, no Agent tool fan-out available in this context).

Diff since my last review (`29bff47`) is a merge of `main` (`96d00af` = merge of `29bff47` + `9eac5fd`, first-parent log confirms no new commits were added to the PR branch itself — only the merge commit). Verified the two places where a conflict was possible:

- **`.github/workflows/pr-label-sync.yml`** — only a comment line changed (`The Release QA audit filters on this label…` → `Release notes and the self-hoster migration guide are compiled from the PRs carrying it.`, picked up from main's `bd832b8`). The `CHECKBOX_LINE` regex and `checkboxFor(declaration)` matching logic this PR introduced are untouched.
- **`.github/pull_request_template.md`** — main's `bd832b8` ("move all QA onto the PR and slim the PR template") rewrote most of the template concurrently with this PR's checkbox change. Confirmed the merge kept the exact checkbox line the workflow depends on: `- [ ] This PR contains breaking changes`, unformatted, still matching `/^this pr contains breaking changes/i` after the `[*_\`]` strip. No drift between template wording and the workflow's declaration regex.
- `.coderabbit.yaml` and `AGENTS.md` changes in this range are main's QA-template rewrite passing through untouched; nothing here interacts with this PR's own logic.

No functional lines from this PR's own commits (`5c7daca`, `29bff47`) were altered by the merge, and the merge didn't reintroduce the old prose-heuristic this PR replaced.

---

kiln-agent (commented):
**1 minor, 2 trivial** — reviewed at `fcaf850` (first pass at this head; earlier passes stopped at `96d00af`).

New since the last review is one docs-only commit, `fcaf850`, correcting the template and `AGENTS.md`: they said the prose under the checkbox is not read at all, when only `pr-label-sync.yml` ignores it — the CodeRabbit `Breaking changes match the diff` check still expects a ticked box to be explained. Checked the new wording against `.coderabbit.yaml:353-355` and it is now accurate in both files. The engine is untouched by this commit.

I re-extracted the inline `script:` at this head and ran it under Node over 19 bodies — the shipped template verbatim in both states, the ENG-2332 prose-"None" case, an unrelated ticked box above the declaration in both label states, no checkbox with and without a hand-applied label, a reworded declaration, QA-section checkboxes, CRLF both ways, bold/`*`-marker/indented/trailing-clause variants, a ticked box only inside the guidance comment, missing section, empty and null bodies — **19/19 as documented**. `prettier --check` clean on all four files. Grepped for `breaking-change` and `Breaking changes` repo-wide: still only these four files, so nothing else depends on the contract.

The diff is correct. What follows is about the rollout plan, not the code.

**Not anchored to a line**

- **🟡 Minor · drift** — the post-merge sweep only covers *open* PRs, and the false positives have been merging

  The `How this was tested` sweep step is `gh pr list --state open --label breaking-change`, over a 7-PR list captured on Aug 14. Since then **8 PRs have merged carrying `breaking-change`**, and I read the bodies of five: #8873 ("None. No schema change, no new env var, no migration, no default altered."), #8898 ("None — internal tooling only"), #8863 ("None — branch synchronization only"), #8864 ("None — adds an audit result field/counter to internal operator tooling"), #8868 ("None — authorization outcomes, external APIs, environment variables, and database schema are unchanged"). All five declare no breaking change and all five merged labelled.

  - The snapshot going stale is fine — the body says "as of opening this PR". The problem is the *method*: `--state open` cannot reach a merged PR, and merge is exactly when the label starts to matter. Release notes and the migration guide are compiled from the PRs carrying it (`.github/workflows/pr-label-sync.yml:9`), and that compile runs after merge.
  - Nothing in-repo consumes the label, so this misleads the human doing the release-notes/migration-guide pass rather than corrupting a generated artifact. Five tooling chores and refactors currently look like migration entries.
  - Fix: make the sweep step cover the merge window too — `gh pr list --state merged --label breaking-change --search 'merged:>=2026-08-14'` — and drop the fixed PR-number list, which is already 6/7 out of date.

- **🔵 Trivial · drift** — the template's breaking-change list is narrower than the list the check grades against

  `.github/pull_request_template.md:26-31` lists API/SDK shape, endpoint/route, env var or config key removal/rename, default or webhook payload, and manual migration action. `.coderabbit.yaml:335-345` adds two cases an author following the template would answer "no" to:

  - "drops or renames a column or table … or adds a NOT NULL column with no default and no backfill" — breaking per the check even with no manual action required, which is the only DB wording the template has;
  - adding an env var, which "counts only when it is required, or has no default that keeps existing installs working" — the template mentions only removes/renames.

  Both predate this PR (neither list changed here), so this is out of the diff's scope — but `fcaf850` exists to stop the docs contradicting the check, and the *list* is the half authors actually decide from. Worth folding in while you are here.

- **🔵 Trivial** — CodeRabbit's `pr-label-sync.yml:117` nit is still open and reproduces. On the drift path the run logs the warning "Found 1 checkbox(es) but none matching …" and then, immediately after, "no checkbox under "## Breaking changes"; leaving unchanged." The contradiction lands precisely where someone is diagnosing drift. Also still open: its three outside-diff comments from `96d00af` and the `.coderabbit.yaml:351-355` unedited-placeholder one.

Also worth a look: `How this was tested` was not refreshed for `fcaf850` — it still describes only the workflow work and "all four files", while the new commit touches two docs files for a different reason. Its claims are not wrong (I re-ran prettier myself), just silent about the newest commit, which `AGENTS.md` asks you to treat as a bug.

**Checked** correctness of the extracted engine at this head, doc↔check accuracy of the new commit, drift, conventions (`permissions`, pinned `harden-runner`, `pull_request_target` safety — the job reads `pr.body` only and never checks out PR code), other consumers of the label, CI (17/17 green) · **Skipped** tenancy and frontend (no product code, no `.tsx`, no surveys/survey-ui), simplicity (nothing simpler to propose) · **Not verifiable here** live label behaviour on a real PR edit, since `pull_request_target` runs the base definition until merge → @Dhruwang

No blocking issues found.

---

mattinannt (approved):

PR #8863, opened 2026-08-14
chore: sync main into AuthZed epic

BhagyaAmarasinghe:
## What & why

`epic/authzed` had diverged substantially from current `main`, while the next authorization work depends on both the current product/security behavior and the complete AuthZed foundation. This synchronizes `main` at `48b9a1719` into the epic at `ac06562b1`, where #8805 is now merged, and incorporates the refreshed open head of #8846 so subsequent ENG-2362/ENG-2398 branches have one coherent baseline.

Conflict resolution treats `main` as authoritative for product behavior and the epic as authoritative for AuthZed infrastructure. Existing feedback-record, feedback-source, Hub, chart, workflow, MCP, authorization, deployment, operations, performance, and settings behavior is retained. The old #8805 PR-head conflicts were resolved in favor of its reviewed merged epic version.

A post-merge audit found that the first synchronization had accidentally dropped the production Compose `authzed-db-bootstrap`, `spicedb-migrate`, and `spicedb` services and the Formbricks AuthZed client environment. That made the retained `authzed-ops` service depend on a missing service and broke the Docker contract tests. The corrected merge restores the complete private SpiceDB runtime, keeps Formbricks startup independent from SpiceDB health, retains #8846's opt-in MCP JWKS override without changing the public issuer or injecting plaintext defaults, and excludes generated operator artifacts from repository-wide Prettier checks instead of reformatting generated files.

The refreshed GitHub unit job then exposed three additional merge regressions that were hidden by narrower local validation: `apps/web/turbo.json` did not inherit the newly synchronized AuthZed/MCP environment allowlist, the workflow-runner test mock omitted Prisma enums imported by the AuthZed projector graph, and the team action test still used the pre-sync update payload. These are reconciled at the test/configuration boundary without changing runtime authorization behavior.

## Linear ticket

No Linear ticket — this is the prerequisite branch synchronization for ENG-2362 and ENG-2398.

## How this was tested

- `pnpm format:check` — passed.
- `pnpm --filter @formbricks/web typecheck` — passed.
- `pnpm --filter @formbricks/web exec vitest run lib/turbo-build-env.test.ts modules/ee/workflows/lib/runner/process-workflow-run-job.test.ts modules/ee/teams/team-list/actions.test.ts` — 3 files / 46 tests passed.
[...]

PR #8864, opened 2026-08-14
fix: detect mismatched AuthZed permission relations

BhagyaAmarasinghe:
## What & why

The relationship audit compared source-record existence rather than exact relation types. A PostgreSQL `read` workspace-team grant projected as `manager_team` therefore looked present and could survive a clean dry run, leaving an undetected privilege escalation.

This derives the complete expected managed relationship set from PostgreSQL using the same enum-to-relation maps as the projectors, compares it with observed SpiceDB tuples, and reports `mismatchedPermissions`. Missing sources, cross-tenant parent edges, invalid source rows, and unmanaged tuples keep their existing distinct classifications. Applying reconciliation heals permission mismatches; a follow-up dry run verifies they are gone.

Stacked on #8863. This must land before ENG-2398 adds feedback-dataset assignment relationships.

## Linear ticket

Fixes ENG-2362

## How this was tested

- `pnpm --filter @formbricks/web exec vitest run lib/authzed` ✅ (22 files, 421 tests)
- `pnpm authzed:validate` ✅ (27 relationships, 131 assertions, 2 expected relations)
- Added regression coverage for stale higher workspace grants, independent API-key flags, missing sources, parent edges, source enumeration, and result/status reporting.

## Breaking changes

None — adds an audit result field/counter to internal operator tooling; application authorization behavior is unchanged.

## QA / Test Plan

**How to test**

- [ ] Seed a workspace-team `read` source row but a `manager_team` SpiceDB relationship → dry run reports `mismatchedPermissions: 1` and `status: drifted`.
- [ ] Run apply reconciliation and repeat the dry run → `mismatchedPermissions: 0` and the relationship is `reader_team`.
- [ ] Remove the entire relationship for a live source row → it remains classified as `missing`, not duplicated as a permission mismatch.
- [ ] Add a wrong organization parent edge → it remains classified as `mismatchedParents`.

**Preconditions / test data**

- AuthZed-enabled sandbox with disposable organization/team/workspace data.

**Risks & regressions**

- Re-check large relationship audits and all role/permission mappings; exact-set derivation must stay aligned with projector maps.

- none

### Stack refresh after #8863 repair

The latest parent chain restores the production AuthZed Compose bootstrap/migration/server contract that was accidentally lost during the main-to-epic synchronization. The refreshed combined head passes repository formatting, Docker Compose contracts, Docker tests, Helm lint/render contracts, AuthZed schema validation (31 relationships / 144 assertions / 2 expected relations), 798 focused authorization/AuthZed tests, and web typecheck. GitHub CI is green on this exact head, including lint, typecheck, unit, E2E, AuthZed schema, Helm, and Sonar checks.

### Parent stack refresh

Merged refreshed #8863 head 815e02214 into this branch after epic/authzed advanced. The combined head preserves the resolved Turborepo schema cache input and latest ENG-2340 schema changes. Validation passed: 49 focused backfill/schema-input tests and AuthZed validation with 32 relationships, 136 assertions, and 2 expected relations.

### Refresh after #8863 merged

- Merged the current `epic/authzed` head after #8863 landed.
- Resolved the README and runbook conflict by preserving this PR’s exact-relation `mismatchedPermissions` contract; the merged base’s older record-only wording would contradict ENG-2362.
- Validation on `5104ff212`: 49 focused tests passed, web typecheck passed after building workspace dependencies, AuthZed validation passed with 32 relationships / 136 assertions / 2 expected relations, and both documentation files pass Prettier.

### Reviewer Sonar feedback

- Refactored `readOrganizationSource` into shared partition/target helpers and focused relationship construction, eliminating the reported cognitive-complexity hotspot while preserving the exact PostgreSQL-to-SpiceDB mapping.
- Split the local mismatch sort from the return mapping so the mutation is explicit and no longer triggers the chained mutating-sort issue.
- Added regression coverage for both independent API-key organization access relations.
- Validation on `58ac8d667`: 127 focused tests passed, web typecheck passed, focused ESLint passed, and AuthZed validation passed with 32 relationships / 136 assertions / 2 expected relations.


xernobyl (approved):
This is exactly the gap I'd flagged on the ticket, and the design closes it correctly.

**Verified the core logic by reading it, not just the tests:**

- `findMismatchedPermissionRelations` groups by source ref on both sides and skips a wholly-absent source (`if (!observedGroup) continue`) — confirmed it does not double-report something `findUnprojectedSourceRefs` already classifies as `missing`. Parent edges are explicitly excluded (`relationship.relation !== PARENT_RELATION`) since `findMismatchedParentEdges` already owns that distinction.
- `expectedRelationships` in `backfill-source.ts` is built from `relationship-map.ts`'s existing const maps (`ORGANIZATION_RELATIONS`, `TEAM_RELATIONS`, `WORKSPACE_TEAM_RELATIONS`, `WORKSPACE_API_KEY_RELATIONS`, `ORGANIZATION_ACCESS_RELATIONS`) — the same source of truth the projectors use, not a re-derived copy. That's the right call: a re-derived mapping is exactly what would let a future rename silently reclassify a correct tuple.
- API-key organization access is modeled correctly as up to two independent relations (`normalizeOrganizationAccess` + iterating both flags), not a single-value ladder — matches how `ApiKey.organizationAccess` actually works.
- Unmanaged/ignored relationship types (survey/dashboard/response) never enter `managedRelationships`, so they can't get miscategorized as a permission mismatch — confirmed in `summarizeObservation`, the branch only populates it when `toSourceRef` resolves.
- Gated to `dry_run` only in `backfill.ts` (`observeOrganization`/`observeWorkspace`), consistent with how `missingCount` is computed, and correctly folded into `toRunStatus`'s drift check.

The repro from the ticket — a `read` grant projected as `manager_team` — is covered directly in `backfill.test.ts` (`"reports a stale higher workspace grant even though the source pair is present"`), and asserts the exact `expectedRelations`/`observedRelations` output shape.

**Security review:** no P1/P2. This is detection-only, internal operator tooling, no application-facing surface. The property that matters — a stale *higher*-permission relation (privilege escalation) is now caught by dry-run rather than silently passing as "projected" — is real and tested.

CI green, no open threads. Removed the `breaking-change` label; the PR body already correctly says "None." Approving.

xernobyl:
Could you take a look at the two SonarCloud issues on this one? The quality gate **passes**, so nothing is blocking — but both are real and one has a wrinkle worth knowing before you spend time on it.

Verified against this PR's head (`b57d59fae`):

**1. `apps/web/lib/authzed/backfill-source.ts:226` — CRITICAL, cognitive complexity 24 vs 15 allowed** (`readOrganizationSource`)

I think you've **already fixed this in #8867**. Its body mentions "refactored organization backfill source construction into focused helpers, reducing the reported cognitive complexity," and I can see the extracted helpers there (`getApiKeyRelationships` and friends) — that file differs by ~660 lines between the two branches, and **#8867 reports 0 new Sonar issues**.

So this is really a question of stack mechanics rather than new work: the finding is attributed to #8864 because that's where the function first grows, and it clears by the time #8867 lands on top. If the stack merges as a unit, I'd leave it. Only worth back-porting the refactor into #8864 if these can land separately and you'd rather not have a CRITICAL attributed to the parent commit.

**2. `apps/web/lib/authzed/backfill-diff.ts:387` — MAJOR, sort mutation**

```ts
return mismatches.sort(([left], [right]) => byCodeUnit(left, right)).map(([, mismatch]) => mismatch);
```

Sonar wants `toSorted` or a separate statement because `.sort()` mutates in place. Worth saying that the underlying risk doesn't apply here — `mismatches` is a local array built a few lines above and never shared, so nothing observes the mutation. It's a readability/consistency fix, not a correctness one. A one-line change if you want the counter at zero.

Neither changes behaviour, and my approval stands either way.

PR #8866, opened 2026-08-14
fix(auth): fail closed on an unrecognized minPermission

Dhruwang:
## What & why

`checkWorkspaceTeamAccess` in the shared server-action middleware compared permission weights directly:

```ts
teamPermissionWeight[workspacePermission] < teamPermissionWeight[accessItem.minPermission]
```

If `minPermission` is anything outside `{read, readWrite, manage}` the right-hand lookup is `undefined`, `number < undefined` is `false`, the insufficient-permission guard never fires, and **every workspace member is admitted — read-only included**. The same holds if the *granted* permission is unrecognized. `accessItem` is typed `any` inside the function, so TypeScript does not protect the body; only `TAccess` at the call sites does.

This diff looks both sides up explicitly in a small `meetsMinimumWeight` helper and refuses an unknown value on either side. `checkTeamAccess` carried the identical pattern against `teamRoleWeight`, so it uses the same helper rather than leaving the same fail-open in the adjacent function.

**Latent — no current behaviour changes.** All 86 non-test call sites pass valid string literals (64 `readWrite`, 32 `read`, 9 `manage` on `workspaceTeam`; 4 `admin`, 3 `contributor` on `team`), and for recognized values the new `granted >= minimum` is exactly the old `!(granted < minimum)`. The value is removing an asymmetry: #8741 introduced two adjacent permission vocabularies — `TFeedbackRecordsGatewayPermission` (`read | write | manage`) and `TTeamPermission` (`read | readWrite | manage`) — so passing the gateway spelling `"write"` where the team spelling is expected is now a plausible mistake. It fails **closed** in `apiKeyPermissionAllows` (`app/api/v3/lib/auth.ts`) and, before this change, **open** here.

Tightening `accessItem: any` prevents the class outright, and is now done in this PR (it turned out not to need `checkOrganizationAccess`'s signature after all): `checkWorkspaceTeamAccess` and `checkTeamAccess` each take their own `Extract<TAccess<...>, ...>` variant, so a typo in `workspaceId`/`teamId`/`minPermission` — or passing the wrong variant — is now a compile error. `checkAuthorizationUpdated` already narrows on `type` before calling either, so the internal type guards were unreachable once the parameters were typed and have been removed.

## Linear ticket

Fixes ENG-2246

## How this was tested

- `pnpm test` in `apps/web` ✅ — full suite green:
  ```
  Test Files  608 passed (608)
       Tests  8001 passed (8001)
    Duration  172.03s
  ```
  The 8 pre-existing `checkAuthorizationUpdated` cases covering valid permissions and roles still pass, which is the evidence that recognized values are unaffected.
- Three cases added to `lib/utils/action-client/action-client-middleware.test.ts`, all driven through the real `checkAuthorizationUpdated` call path (not the guard expression in isolation):
  - a read-only member with `minPermission: "write"` (the gateway spelling, not a `TTeamPermission`) → `AuthorizationError`;
  - a workspace permission of `"write"` against `minPermission: "readWrite"` → `AuthorizationError`;
  - a `team` item with `minPermission: "manage"` (a `TTeamPermission`, not a `TTeamRole`) against a `contributor` → `AuthorizationError`.
- Confirmed all three can actually fail: with the source file stashed and only the test changes applied, they fail on `main`'s behaviour, each admitting the caller —
  ```
  × refuses workspaceTeam access when minPermission is an unrecognized value
  × refuses workspaceTeam access when the granted permission is an unrecognized value
  × refuses team access when minPermission is an unrecognized value
    AssertionError: promise resolved "true" instead of rejecting  (×3)
  Tests  3 failed | 49 passed (52)
  ```
  This also demonstrates the `team` variant of the fail-open end-to-end, which the ticket had only inferred.
- `pnpm lint --filter=@formbricks/web` ✅ — `21 successful, 21 total`, 0 errors (warnings are pre-existing and elsewhere; ESLint on both changed files is clean).
- `pnpm format:check` ✅ — `All matched files use Prettier code style!`
- No UI change, so no screenshots.

## Breaking changes

None

## QA / Test Plan

This is a latent hardening fix with no reachable behaviour change today, so the QA goal is confirming **nothing regressed** in permission enforcement across the app. Server actions are the surface — every permission-gated action in the product runs through this middleware.

**How to test**

- [ ] As a workspace member with **read** permission on a workspace: open a survey in that workspace → you can view it; attempt an edit-level action (rename the survey, change a question, publish) → blocked with "Not authorized". Unchanged.
- [ ] As a member with **readWrite**: create, edit and publish a survey in that workspace → all succeed. Deleting the workspace or managing its members → blocked. Unchanged.
- [ ] As a member with **manage**: workspace-level management (workspace settings, workspace deletion, member permissions) → succeeds. Unchanged.
- [ ] Team roles: as a team **contributor**, team-management actions (renaming the team, adding/removing members) → blocked; as a team **admin** → succeed. Unchanged.
- [ ] Organization roles: as an org **owner**/**manager**, actions gated on organization role still succeed; as a plain **member** with no workspace permission on the target workspace → blocked with "Not authorized".
- [ ] A user with **no membership** in the target workspace → every workspace-scoped action blocked. Unchanged.
- [ ] Access items that specify **no** minimum permission still admit any member of the workspace/team holding a recognized permission (e.g. a read-only member reaching a no-minimum action succeeds). Note the helper now validates the *granted* value even when no minimum is set, so an unrecognized grant is refused rather than admitted — not reachable through today's callers, whose grants come from Postgres enums.

**Preconditions / test data**

- One organization with at least two workspaces and a team.
- Four accounts: workspace `read`, workspace `readWrite`, workspace `manage`, and one with no membership in the target workspace.
- One team with a `contributor` and an `admin` member.
- Access control / RBAC must be available on the plan for per-workspace permissions to be assignable (Enterprise entitlement, or a self-hosted instance with an active license).

**Risks & regressions**

- Highest risk is an over-tightening regression: a permission that used to be accepted now being refused. The escalation path is the whole product, so spread the spot checks across surfaces (surveys, contacts, integrations, workspace and team settings) rather than one screen.
- The `team` code path is touched as well as `workspaceTeam` — re-check team-scoped actions specifically, not just workspace ones.
- Nothing about organization-role checks changed (`checkOrganizationAccess` is untouched); a failure there would point at something else.

- none

---
_Generated by [Claude Code](https://claude.ai/code/session_01LDiu2Devg6C2EFwDuE8GuS)_

### DIFF
--- a/apps/web/lib/utils/action-client/action-client-middleware.ts
+++ b/apps/web/lib/utils/action-client/action-client-middleware.ts
@@ -91,15 +91,22 @@ const meetsMinimumWeight = (
   return grantedWeight >= minimumWeight;
 };
 
-const checkWorkspaceTeamAccess = async (accessItem: any, userId: string) => {
-  if (accessItem.type !== "workspaceTeam") return false;
+/**
+ * The `workspaceTeam` and `team` variants of `TAccess` carry no schema, so neither depends on `T` —
+ * hence the concrete `z.ZodRawShape` here rather than threading the generic through these two helpers.
+ * `checkAuthorizationUpdated` narrows on `type` before calling either, so each receives exactly its
+ * own variant and `workspaceId` / `teamId` / `minPermission` are checked at compile time.
+ */
+type TWorkspaceTeamAccess = Extract<TAccess<z.ZodRawShape>, { type: "workspaceTeam" }>;
+type TTeamAccess = Extract<TAccess<z.ZodRawShape>, { type: "team" }>;
+
+const checkWorkspaceTeamAccess = async (accessItem: TWorkspaceTeamAccess, userId: string) => {
   const workspacePermission = await getWorkspacePermissionByUserId(userId, accessItem.workspaceId);
   if (!workspacePermission) return false;
   return meetsMinimumWeight(teamPermissionWeight, workspacePermission, accessItem.minPermission);
 };
 
-const checkTeamAccess = async (accessItem: any, userId: string) => {
-  if (accessItem.type !== "team") return false;
+const checkTeamAccess = async (accessItem: TTeamAccess, userId: string) => {
   const teamRole = await getTeamRoleByTeamIdUserId(accessItem.teamId, userId);
   if (!teamRole) return false;
   return meetsMinimumWeight(teamRoleWeight, teamRole, accessItem.minPermission);
--- a/apps/web/lib/utils/action-client/action-client-middleware.ts
+++ b/apps/web/lib/utils/action-client/action-client-middleware.ts
@@ -63,32 +63,39 @@ const checkOrganizationAccess = <T extends z.ZodRawShape>(
   return accessItem.roles.includes(role);
 };
 
+/**
+ * Compares a granted role/permission against the minimum required one.
+ *
+ * Both sides are looked up explicitly and an unrecognized value on either side is refused. Comparing
+ * the weights directly would fail open: an unknown value resolves to `undefined`, `number < undefined`
+ * is `false`, so the insufficient-permission guard would not fire and every member would be admitted.
+ */
+const meetsMinimumWeight = (
+  weights: Record<string, number>,
+  granted: string,
+  minimum: string | undefined
+): boolean => {
+  if (minimum === undefined) return true;
+
+  const grantedWeight = weights[granted];
+  const minimumWeight = weights[minimum];
+  if (grantedWeight === undefined || minimumWeight === undefined) return false;
+
+  return grantedWeight >= minimumWeight;
+};
+
 const checkWorkspaceTeamAccess = async (accessItem: any, userId: string) => {
   if (accessItem.type !== "workspaceTeam") return false;
   const workspacePermission = await getWorkspacePermissionByUserId(userId, accessItem.workspaceId);
   if (!workspacePermission) return false;
-  if (
-    accessItem.minPermission !== undefined &&
-    teamPermissionWeight[workspacePermission as keyof typeof teamPermissionWeight] <
-      teamPermissionWeight[accessItem.minPermission as keyof typeof teamPermissionWeight]
-  ) {
-    return false;
-  }
-  return true;
+  return meetsMinimumWeight(teamPermissionWeight, workspacePermission, accessItem.minPermission);
 };
 
 const checkTeamAccess = async (accessItem: any, userId: string) => {
   if (accessItem.type !== "team") return false;
   const teamRole = await getTeamRoleByTeamIdUserId(accessItem.teamId, userId);
   if (!teamRole) return false;
-  if (
-    accessItem.minPermission !== undefined &&
-    teamRoleWeight[teamRole as keyof typeof teamRoleWeight] <
-      teamRoleWeight[accessItem.minPermission as keyof typeof teamRoleWeight]
-  ) {
-    return false;
-  }
-  return true;
+  return meetsMinimumWeight(teamRoleWeight, teamRole, accessItem.minPermission);
 };
 
 export const checkAuthorizationUpdated = async <T extends z.ZodRawShape>({
--- a/apps/web/lib/utils/action-client/action-client-middleware.ts
+++ b/apps/web/lib/utils/action-client/action-client-middleware.ts
@@ -69,17 +69,24 @@ const checkOrganizationAccess = <T extends z.ZodRawShape>(
  * Both sides are looked up explicitly and an unrecognized value on either side is refused. Comparing
  * the weights directly would fail open: an unknown value resolves to `undefined`, `number < undefined`
  * is `false`, so the insufficient-permission guard would not fire and every member would be admitted.
+ *
+ * `granted` is validated before the no-minimum case, so an unrecognized grant is refused even when the
+ * caller asks for no minimum. Both current callers read `granted` from a native Postgres enum, so this
+ * is unreachable today; it is here so the helper stays fail-closed for a future caller whose grant is
+ * not enum-backed.
  */
 const meetsMinimumWeight = (
   weights: Record<string, number>,
   granted: string,
   minimum: string | undefined
 ): boolean => {
+  const grantedWeight = weights[granted];
+  if (grantedWeight === undefined) return false;
+
   if (minimum === undefined) return true;
 
-  const grantedWeight = weights[granted];
   const minimumWeight = weights[minimum];
-  if (grantedWeight === undefined || minimumWeight === undefined) return false;
+  if (minimumWeight === undefined) return false;
 
   return grantedWeight >= minimumWeight;
 };
--- a/apps/web/lib/utils/action-client/action-client-middleware.test.ts
+++ b/apps/web/lib/utils/action-client/action-client-middleware.test.ts
@@ -366,6 +366,62 @@ describe("action-client-middleware", () => {
       expect(result).toBe(true);
     });
 
+    test("refuses workspaceTeam access when minPermission is an unrecognized value", async () => {
+      vi.mocked(getMembershipRole).mockResolvedValue("member");
+
+      // "write" is the feedback-records gateway spelling, not a TTeamPermission ("readWrite").
+      // An unrecognized minimum must not admit a read-only member.
+      const access = [
+        {
+          type: "workspaceTeam" as const,
+          workspaceId,
+          minPermission: "write" as any,
+        },
+      ];
+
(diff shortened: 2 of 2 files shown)


mattinannt (dismissed):
Reviewed the diff in `action-client-middleware.ts` / `action-client-middleware.test.ts`.

- Traced `meetsMinimumWeight` against the old guard: for defined/valid inputs, `grantedWeight >= minimumWeight` is exactly `!(grantedWeight < minimumWeight)`, so recognized-value behavior is unchanged; unrecognized values on either side now correctly return `false` instead of `number < undefined` silently evaluating to `false` and admitting the caller.
- Grepped all `minPermission:` call sites (86+) — all pass valid literals, consistent with the PR's "latent, no behavior change" claim.
- Confirmed the one real `write`→`readWrite` vocabulary mismatch (`feedback-records-gateway.ts`) is already translated via `GATEWAY_PERMISSION_TO_TEAM_PERMISSION` before reaching this middleware, so the new tests are a regression guard rather than covering a live bug.
- Ran the test file (19/19 pass, including the 3 new cases) and `tsc --noEmit` — clean.
- `meetsMinimumWeight` is a reasonable dedup of the previously-duplicated pattern between `checkWorkspaceTeamAccess` and `checkTeamAccess`; no over-abstraction.

No issues found. LGTM.

---

apps/web/lib/utils/action-client/action-client-middleware.ts
```diff
@@ -63,32 +63,39 @@ const checkOrganizationAccess = <T extends z.ZodRawShape>(
...
+ * Both sides are looked up explicitly and an unrecognized value on either side is refused. Comparing
+ * the weights directly would fail open: an unknown value resolves to `undefined`, `number < undefined`
+ * is `false`, so the insufficient-permission guard would not fire and every member would be admitted.
+ */
+const meetsMinimumWeight = (
+  weights: Record<string, number>,
+  granted: string,
+  minimum: string | undefined
+): boolean => {
+  if (minimum === undefined) return true;
```
xernobyl, line 86: **[P3]** When `minimum` is `undefined`, `granted` is never checked against `weights` at all — an unrecognized granted value with no `minPermission` specified would still return `true`. CodeRabbit flagged this as a high merge risk.

Verified it's not currently reachable: the only two callers are `checkWorkspaceTeamAccess`/`checkTeamAccess` below, and `granted` there always comes from `getWorkspacePermissionByUserId`/`getTeamRoleByTeamIdUserId`, which read `WorkspaceTeamPermission`/`TeamUserRole` — native Postgres enums in `packages/database/schema/main.prisma` — so the DB guarantees a recognized value today.

Still, worth considering `if (weights[granted] === undefined) return false;` before the `minimum === undefined` short-circuit, purely as defense-in-depth if this helper ever gets a caller whose `granted` isn't enum-backed. Not a blocker.

apps/web/lib/utils/action-client/action-client-middleware.ts
```diff
@@ -63,32 +63,39 @@ const checkOrganizationAccess = <T extends z.ZodRawShape>(
...
+
+  const grantedWeight = weights[granted];
+  const minimumWeight = weights[minimum];
+  if (grantedWeight === undefined || minimumWeight === undefined) return false;
+
+  return grantedWeight >= minimumWeight;
+};
+
 const checkWorkspaceTeamAccess = async (accessItem: any, userId: string) => {
```
xernobyl: **[P3]** `accessItem: any` (also on `checkTeamAccess` below) means the compiler doesn't protect either function body — the PR description already calls this out and defers it as separate work since it'd touch `checkOrganizationAccess`'s signature too via `TAccess<T>`. Agreed that's reasonable to scope out of this fix; flagging only because CodeRabbit raised the same point and it's worth a follow-up ticket rather than silently dropping it.

xernobyl (dismissed):
Reviewed for correctness and security (ENG-2246). No P1/P2 findings — approving. Two P3 nitpicks left as inline comments.

**Correctness/security walkthrough**
- `meetsMinimumWeight` correctly closes the fail-open hole: for recognized values, `granted >= minimum` is exactly `!(granted < minimum)` (behavior-preserving), and it now returns `false` when either side isn't a key in `weights` instead of `number < undefined` silently evaluating `false`.
- Grepped every `minPermission:` call site in the repo (86+, across `app/`, `modules/`, `packages` callers) — all pass valid `TTeamPermission`/`TTeamRole` literals, confirming the PR's "latent, no behavior change" claim.
- The one real vocabulary mismatch this guards against — `feedback-records-gateway.ts` using the gateway spelling (`read`/`write`/`manage`) — is already translated to team spelling via `GATEWAY_PERMISSION_TO_TEAM_PERMISSION` before it reaches this middleware, so today it's a regression guard, not a live exploit path.
- `checkOrganizationAccess` (untouched) and the `checkAuthorizationUpdated` loop are unaffected by this diff.
- Tests: the 3 new cases each drive `checkAuthorizationUpdated` end-to-end (not the guard expression in isolation), and the PR's own repro (stash the source, keep the tests) shows they fail against `main`'s current behavior — solid evidence the fix is real.

**On CodeRabbit's "High" merge-risk flag** — it claims an unrecognized *granted* permission with no `minPermission` specified would still pass. True by construction, but not reachable today — see the inline comment for why. I'd treat that as a P3/optional hardening rather than a blocker.

**Non-security nit** — the PR description still carries the `_Generated by Claude Code_` footer, which CodeRabbit's own pre-merge check also flagged. Worth stripping per this repo's convention against AI-tooling references in PR descriptions/commits.

apps/web/lib/utils/action-client/action-client-middleware.ts
Dhruwang: Done in `ee1486a3c0` — took the defence-in-depth option you suggested. `granted` is now looked up in `weights` before the `minimum === undefined` short-circuit, so omitting `minPermission` means "any *recognized* grant" rather than "no check".

Your reachability analysis matched what I found: both callers get `granted` from `getWorkspacePermissionByUserId`/`getTeamRoleByTeamIdUserId`, which read the `WorkspaceTeamPermission`/`TeamUserRole` Postgres enums, so nothing can reach it today. Added two regression tests anyway (unrecognized grant + omitted `minPermission`, one per helper) and confirmed they fail against the old ordering while the other 19 stay green — so the guard is pinned rather than just asserted.

apps/web/lib/utils/action-client/action-client-middleware.ts
Dhruwang: Rather than file a follow-up, I did it here — `30cb100410`. It turned out not to need `checkOrganizationAccess`'s signature after all, which was the reason for deferring.

The `workspaceTeam` and `team` variants carry no schema, so neither depends on `T`; each helper now takes its own `Extract<TAccess<z.ZodRawShape>, ...>` variant. `checkAuthorizationUpdated` already narrows on `type` before calling either, so the internal type guards became unreachable and are gone — the compiler now rejects at the call site what they caught at runtime.

Verified it actually bites by reintroducing both mistakes the `any` allowed: renaming `workspaceId` to `workspaceID`, and passing a `workspaceTeam` item to `checkTeamAccess`. Both are now tsc errors. PR description updated, since it still described this as deferred.

apps/web/lib/utils/action-client/action-client-middleware.ts
```diff
@@ -63,32 +63,39 @@ const checkOrganizationAccess = <T extends z.ZodRawShape>(
...
   if (!teamRole) return false;
-  if (
-    accessItem.minPermission !== undefined &&
-    teamRoleWeight[teamRole as keyof typeof teamRoleWeight] <
-      teamRoleWeight[accessItem.minPermission as keyof typeof teamRoleWeight]
-  ) {
-    return false;
-  }
-  return true;
+  return meetsMinimumWeight(teamRoleWeight, teamRole, accessItem.minPermission);
```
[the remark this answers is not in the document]
Dhruwang: Done in `30cb100410`, taking the `Extract<TAccess<...>, ...>` approach you suggested.

One deviation: the two variants carry no schema and so don't depend on `T`, so they resolve against a concrete `z.ZodRawShape` rather than threading the generic through both helpers. Same compile-time protection for `workspaceId` / `teamId` / `minPermission`, without widening the signatures.

Since `checkAuthorizationUpdated` narrows on `type` before calling either helper, the internal `if (accessItem.type !== ...) return false` guards were unreachable once the parameters were typed, so I removed them rather than leave untestable branches behind.

xernobyl (approved):
Thank you for the PR. LGTM.

PR #8871, opened 2026-08-14
refactor(settings): move the api keys table onto SettingsTable

itsjavi:
## What & why

Seventh of the ENG-762 series, and the **first CSS-grid conversion** — so this is the proof case for that
half of the work, the way #8839 was for the semantic-table half. `hideBelow`, `stopRowClick` and the row
activator all get their first real use here.

**Branches off `main` directly.** #8837–#8840 have merged, so PRs 7–12 no longer stack: each converts one
independent table and can land in any order.

### What the grid was hiding

- **A real header/row misalignment.** The header declared `grid-cols-10` with **four** cells; the rows
  declared `grid-cols-10` with **three or four** depending on `isReadOnly`. The trailing empty header cell
  was unconditional while the actions cell was not, so in the read-only case the header had a column the
  rows didn't. Both are now one column object, so they cannot disagree.
- **Dead code the grid made invisible:** a `grid-cols-9` wrapper around the row list (a *third* column
  count, different from both the header and the rows), a `rounded-t-lg` on a header whose top corners have
  never been visible because the card's header block owns them, and a `rounded-lg` on every row.
- **A row that claimed to be a button.** `role="button"` plus `tabIndex={0}` plus a hand-rolled
  `onKeyDown` on a `<div>`. A `row` cannot also be a `button` — the role override drops the row out of the
  table's accessibility tree entirely, and the six cell divs inside had no column association. The
  primitive wraps the first cell's content in a real `<button>` instead, so the row keeps its `row` role,
  the button gets a proper accessible name, and there is still exactly one tab stop.

### The skeleton is the payoff

`loading.tsx` hand-rolled the header markup and **had already drifted: three columns against the table's
four.** Nothing could catch that — it's two files agreeing by convention. It now renders
`SettingsTableSkeleton` from the table's own exported `getApiKeyColumns`, so the two cannot diverge.

This is the first table in the series where the factory-export pattern earns its keep rather than just
being available — which is why I un-exported the equivalent factories on #8839 and #8841 when a reviewer
pointed out they had no consumer.

## Changes from code review

Three findings, all valid, all fixed in `e9b5c0d`:

- **The actions cell is a real table cell again.** `cellClassName: "flex justify-end"` lands on the `<td>`
  itself, and overriding its display away from `table-cell` makes the shared `align-middle` a no-op
  (`vertical-align` applies only to inline-level and table-cell boxes) and forces the browser to wrap the
  cell in an anonymous table-cell defaulting to baseline — so the delete button would sit high in the row
  on every row. The flex moved to a wrapper inside `cell`. **This is the same reasoning the `hideBelow`
  comment two columns above already spells out, and I failed to apply it here.** Fixed in the three
  sibling tables too (#8841, #8842) and documented on the `cellClassName` prop, since three call sites
  making the same mistake is an API problem rather than three coding slips.
- **The icon-only delete button gained `aria-label={t("common.delete")}`.** It had no accessible name at
  all, before this PR or after it — only a `TrashIcon` inside.
- **`isReadOnly` is removed rather than worked around.** It was dead through three files: `page.tsx`
  computes `canAccessApiKeys`, **throws** `not_authorized` when false, and *then* passes
  `isReadOnly={!canAccessApiKeys}`. My first version kept the prop and hardcoded `isReadOnly: false` in
  the skeleton — which left a skeleton/table column contract with no compiler signal, since the four
  widths only sum to 100% with all four columns present. Removing the prop makes that failure mode
  unrepresentable instead of merely unreachable, and made the diff smaller.

## Preserved deliberately (E2E-critical)

**`.copyApiKeyIcon` must not move.** `playwright/lib/utils.ts:141,147` waits for it and clicks it to read a
freshly created key, and **eight API specs take their bearer token from that helper** — webhook, contacts,
response, survey, role, user, workspace-team, team. Breaking it breaks all eight, in a way that looks like
an auth failure rather than a selector failure.

| Kept | Where |
| --- | --- |
| `.copyApiKeyIcon` class | on the copy affordance, with a comment saying why it can't be renamed |
| `data-testid="copy-button"` | on the icon |
| `data-testid="api-key-row"` | via `getRowProps` |

`hideBelow: "sm"` reproduces the old `hidden sm:block` breakpoint exactly, so the copy icon is visible at
the same viewports as before — which matters, because that helper waits on its *visibility*.

One subtlety worth stating: **the api key column is deliberately not `stopRowClick`.** Clicking the key
*text* has always opened the permissions modal; only the copy icon is an exception, so it keeps its own
`stopPropagation` rather than the whole column opting out. Marking the column would have silently removed
a click target.

## Behaviour and visual deltas

| Change | Why |
| --- | --- |
| **"Add API key" moves from below-left to above-right** | Matches every other converted settings table, and is required: a flush card body means the table must be the last thing in the card. This is the one deliberate UX change — say so in review if you'd rather it stayed below. |
| Header type goes **bold slate-900 → medium slate-500** | The shared settings-table treatment. |
| Actions button goes **centred → right-aligned** | Consistency with the other actions columns. In a 10%-wide column it's a few pixels. |
| Empty state row gets taller (`h-12` → `h-24`) | The primitive's empty cell. Same copy. |

`hidden sm:block` → `hideBelow: "sm"` is not cosmetic: `display: block` on a `<td>` drops it out of the
table's column layout and misaligns the whole row, so the primitive uses `table-cell`.

## Linear ticket

Ref ENG-762

## How this was tested

- `npx turbo run typecheck --filter=@formbricks/web` → **12/12 tasks successful, 0 errors.**
- `npx vitest run modules/ui/components/settings-table` → **25/25 pass.**
- `npx eslint` over the whole `api-keys` module and `npx prettier --check` → clean.
- **`pnpm scan-translations` → all validations pass.** This needed care: the row activator's accessible
  name is a new key, and the scanner fails on a key present in `en-US` but missing from the other 14
  locales *and* on a stale `i18n.lock`. Hand-adding to the other locales is forbidden by AGENTS.md, so I
  added it to `en-US.json` only and ran `pnpm i18n:web:generate`, which regenerated all 14 locales and
  rebuilt the lockfile. Diff is exactly **one key per locale plus the lockfile line** — no incidental
  reformatting.
- **Traced every consumer of the E2E selectors before touching them**, rather than assuming
  `data-testid="api-key-row"` was the only coupling. `.copyApiKeyIcon` is a bare CSS class with no testid,
  which is exactly what a refactor deletes by accident.
- **Swept the whole app for the `cellClassName` display bug** after review, not just this file:
  `grep 'cellClassName: "[^"]*\(flex\|grid\|block\)'` now returns nothing across `modules/` and `app/`.

**Still owed before merge:** I can't run Playwright here — no Docker, so no database and no app. Nothing in
this PR edits a spec, so the whole suite should pass **unchanged**; the risk concentrates on
`.copyApiKeyIcon` still being clickable, which `pnpm test:e2e -- playwright/api/` covers. The vertical
centring of the delete button is also reasoned from CSS box semantics rather than observed, so it needs a
real look.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open Organization settings → API Keys → the table fills the card edge to edge, header band flush
      against the card's title divider, last row meeting the card's bottom border
- [ ] On that page, find the "Add API key" button → it is now **above** the table, right-aligned, and inset
      from the card edges → a button touching the card edge is the regression
- [ ] On that page, read across a row → label (semibold), the key, created-at, and a trash button
- [ ] On that page, look down the trash-button column → each button is **vertically centred** in its row,
      level with the label text beside it → a button sitting high or on the text baseline is the bug the
      review round fixed, and it shows up most on rows where the key text wraps to two lines
- [ ] Click "Add API key", fill the form and submit → the new row appears with the plaintext key visible
- [ ] Click the copy icon on that new row → toast confirms, and pasting gives the key → **highest-risk
      check: eight API specs read their bearer token through this exact icon**
- [ ] Click the copy icon again and confirm the permissions modal does **not** open → the icon stops the
      click from reaching the row
- [ ] Click anywhere else on a row — the label, the key text, the created-at cell → the permissions modal
      opens for that key
- [ ] Tab to a row → focus lands on one visible ring around the label cell, not the whole row; press Enter
      → the permissions modal opens
- [ ] With a screen reader, move onto a row's activator → it announces "View permissions for &lt;label&gt;";
      move onto the trash button → it announces "Delete"
- [ ] Click the trash button → the delete dialog opens and the permissions modal does **not**
- [ ] Confirm the delete → the row disappears and a toast confirms
- [ ] In the permissions modal, rename the key and click Update → the row's label updates in place
- [ ] Narrow the browser below 640px, then reload → the API-key column disappears and the remaining three
      columns rescale with no empty gap and no sideways page scroll
- [ ] Widen back above 640px → the API-key column returns
- [ ] Throttle the network and reload → the loading skeleton shows **four** column headers matching the
      real table, plus a placeholder button above it → three headers is the drift this PR fixes
- [ ] Open an organization with no API keys → "You do not have any API keys yet" centred in a full-width
      row, headers still visible above it

**Preconditions / test data**

- An org owner or manager account — the page throws for anyone else
- At least three API keys, one created during the session so its plaintext value is on screen
- A key whose label is long, and one row where the key text wraps, to make the vertical-centring check
  meaningful
- A browser where clipboard read is permitted, for the copy check
- A throttled connection or cold cache to see the loading skeleton at all

**Risks & regressions**

- **`.copyApiKeyIcon` is the single highest risk in this PR.** A bare CSS class, referenced only from
  `playwright/lib/utils.ts`, with eight API specs depending on the helper that clicks it. If those specs
  fail with auth errors rather than selector errors, this is why.
- The row activator replaces a hand-rolled `role="button"`/`onKeyDown` div. Keyboard activation is the
  thing to check — mouse clicking will work regardless.
- `hideBelow: "sm"` must hide and show at the same width as the old `hidden sm:block`, since the copy-icon
  helper waits on visibility.
- The "Add API key" reposition is deliberate but is a real UX change; anyone used to the old position will
  notice.
- Vertical alignment of the delete button changed twice in this PR (broken, then fixed). It is reasoned
  from CSS box semantics, not screenshotted, so it deserves an actual look.

- One new i18n key (`workspace.api_keys.view_permissions_for`) with all 14 locales generated. No env, no
  migration.

## Follow-ups this PR deliberately does not do

- **The page throws instead of rendering read-only.** `page.tsx` throws `not_authorized` for a
  non-owner/manager, which is what made `isReadOnly` dead. Fixing that is a behaviour change well outside a
  restyle; whoever does it should reintroduce the flag deliberately and reconsider the column widths at the
  same time.
- **`MemberActions`' icon-only share-invite and resend-invite buttons** have the same missing accessible
  name this PR fixed for the trash button. Different module, so not touched here.
- **The skeleton's actions bar is left-aligned** while the real button is right-aligned. It can't be
  addressed through column config — the skeleton renders a block-level bar, so neither `align` nor a `cell`
  wrapper reaches it. A few pixels of grey rectangle in a 10% column; not worth new primitive API.

### DIFF
--- a/apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
+++ b/apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
@@ -73,69 +73,67 @@ const ApiKeyDisplay = ({ apiKey }: Readonly<{ apiKey: string }>) => {
 export const getApiKeyColumns = ({
   t,
   locale,
-  isReadOnly,
   onDelete,
 }: Readonly<{
   t: TFunction;
   locale: TUserLocale;
-  isReadOnly: boolean;
   onDelete: (event: React.MouseEvent, apiKey: TApiKeyRow) => void;
-}>): TSettingsTableColumn<TApiKeyRow>[] => {
-  const columns: TSettingsTableColumn<TApiKeyRow>[] = [
-    {
-      id: "label",
-      header: t("common.label"),
-      headerClassName: "w-[25%]",
-      cellClassName: "font-semibold",
-      skeletonWidth: "w-32",
-      cell: (apiKey) => apiKey.label,
-    },
-    {
-      id: "apiKey",
-      header: t("workspace.api_keys.api_key"),
-      headerClassName: "w-[45%]",
-      // Replaces `hidden sm:block`, which would have restored a `<td>` to `display: block` and dropped it
-      // out of the table's column layout. `hideBelow` uses `table-cell` for exactly that reason.
-      hideBelow: "sm",
-      skeletonWidth: "w-64",
-      cell: (apiKey) => <ApiKeyDisplay apiKey={apiKey.actualKey ?? ""} />,
-    },
-    {
-      id: "createdAt",
-      header: t("common.created_at"),
-      headerClassName: "w-[20%]",
-      skeletonWidth: "w-20",
-      cell: (apiKey) => timeSince(apiKey.createdAt.toString(), locale),
-    },
-  ];
-
-  if (!isReadOnly) {
-    columns.push({
-      id: "actions",
-      header: null,
-      srLabel: t("common.actions"),
-      headerClassName: "w-[10%]",
-      // The flex aligns the button; `align` would be inert here, since the button is block-level and the
-      // header has no visible text.
-      cellClassName: "flex justify-end",
-      stopRowClick: true,
-      skeletonWidth: "w-8",
-      cell: (apiKey) => (
-        <Button size="icon" variant="ghost" onClick={(event) => onDelete(event, apiKey)}>
+}>): TSettingsTableColumn<TApiKeyRow>[] => [
+  {
+    id: "label",
+    header: t("common.label"),
+    headerClassName: "w-[25%]",
+    cellClassName: "font-semibold",
+    skeletonWidth: "w-32",
+    cell: (apiKey) => apiKey.label,
+  },
+  {
+    id: "apiKey",
+    header: t("workspace.api_keys.api_key"),
+    headerClassName: "w-[45%]",
+    // Replaces `hidden sm:block`, which would have restored a `<td>` to `display: block` and dropped it
+    // out of the table's column layout. `hideBelow` uses `table-cell` for exactly that reason.
+    hideBelow: "sm",
+    skeletonWidth: "w-64",
+    cell: (apiKey) => <ApiKeyDisplay apiKey={apiKey.actualKey ?? ""} />,
+  },
+  {
+    id: "createdAt",
+    header: t("common.created_at"),
+    headerClassName: "w-[20%]",
+    skeletonWidth: "w-20",
+    cell: (apiKey) => timeSince(apiKey.createdAt.toString(), locale),
+  },
+  {
+    id: "actions",
+    header: null,
+    srLabel: t("common.actions"),
+    headerClassName: "w-[10%]",
+    stopRowClick: true,
+    skeletonWidth: "w-8",
+    // The flex goes on a wrapper inside the cell, never on `cellClassName`: that class lands on the `<td>`
+    // itself, and `display: flex` there stops it being a table-cell — which silently kills the shared
+    // `align-middle` (`vertical-align` applies only to inline-level and table-cell boxes) and makes the
+    // browser wrap the cell in an anonymous table-cell that defaults to baseline alignment. The button
+    // would sit high in the row instead of centred. Same reasoning as `hideBelow` above.
+    cell: (apiKey) => (
+      <div className="flex justify-end">
+        <Button
+          size="icon"
+          variant="ghost"
+          aria-label={t("common.delete")}
+          onClick={(event) => onDelete(event, apiKey)}>
           <TrashIcon />
         </Button>
-      ),
-    });
-  }
-
-  return columns;
-};
+      </div>
+    ),
+  },
+];
 
 interface EditAPIKeysProps {
   organizationId: string;
   apiKeys: TApiKeyWithEnvironmentPermission[];
   locale: TUserLocale;
-  isReadOnly: boolean;
   workspaces: TOrganizationWorkspace[];
   isFormbricksCloud: boolean;
 }
@@ -144,7 +142,6 @@ export const EditAPIKeys = ({
   organizationId,
   apiKeys,
   locale,
-  isReadOnly,
   workspaces,
   isFormbricksCloud,
 }: Readonly<EditAPIKeysProps>) => {
@@ -245,23 +242,21 @@ export const EditAPIKeys = ({
 
   return (
     <>
-      {!isReadOnly && (
-        // Moved above the table, matching every other settings table in this series — and required, since
-        // a flush card body means the table has to be the last thing in it. The control carries the card's
-        // gutter itself.
-        <div className="mb-4 flex justify-end px-4 pt-4">
-          <Button
-            size="sm"
-            onClick={() => {
-              setIsAddAPIKeyModalOpen(true);
-            }}>
-            {t("workspace.settings.api_keys.add_api_key")}
-          </Button>
-        </div>
-      )}
+      {/* Moved above the table, matching every other settings table in this series — and required, since
+          a flush card body means the table has to be the last thing in it. The control carries the card's
+          gutter itself. */}
+      <div className="mb-4 flex justify-end px-4 pt-4">
+        <Button
+          size="sm"
+          onClick={() => {
+            setIsAddAPIKeyModalOpen(true);
+          }}>
+          {t("workspace.settings.api_keys.add_api_key")}
+        </Button>
+      </div>
 
       <SettingsTable
-        columns={getApiKeyColumns({ t, locale, isReadOnly, onDelete: handleOpenDeleteKeyModal })}
+        columns={getApiKeyColumns({ t, locale, onDelete: handleOpenDeleteKeyModal })}
         rows={apiKeysLocal}
         getRowId={(apiKey) => apiKey.id}
         emptyMessage={t("workspace.api_keys.no_api_keys_yet")}
--- a/apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
+++ b/apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import type { TFunction } from "i18next";
 import { FilesIcon, TrashIcon } from "lucide-react";
 import { useState } from "react";
 import toast from "react-hot-toast";
@@ -17,8 +18,12 @@ import {
 } from "@/modules/organization/settings/api-keys/types/api-keys";
 import { Button } from "@/modules/ui/components/button";
 import { DeleteDialog } from "@/modules/ui/components/delete-dialog";
+import { SettingsTable, type TSettingsTableColumn } from "@/modules/ui/components/settings-table";
 import { createApiKeyAction, deleteApiKeyAction, updateApiKeyAction } from "../actions";
 import { AddApiKeyModal } from "./add-api-key-modal";
+
+/** A stored key, plus the plaintext value the create response returns once and only once. */
+type TApiKeyRow = TApiKeyWithEnvironmentPermission & { actualKey?: string };
 
 const ApiKeyDisplay = ({ apiKey }: Readonly<{ apiKey: string }>) => {
   const { t } = useTranslation();
@@ -38,10 +43,15 @@ const ApiKeyDisplay = ({ apiKey }: Readonly<{ apiKey: string }>) => {
   return (
     <div className="flex items-center justify-between gap-2">
       <span className="break-all whitespace-pre-line">{apiKey}</span>
+      {/* `copyApiKeyIcon` is load-bearing: `playwright/lib/utils.ts` waits for and clicks it to read a
+          freshly created key, which eight API specs depend on for their bearer token. */}
       <div className="copyApiKeyIcon shrink-0">
         <FilesIcon
           className="size-4 cursor-pointer"
           onClick={(e) => {
+            // Stops the click reaching the row, which would open the permissions modal. The column is
+            // deliberately not `stopRowClick`: clicking the key *text* has always opened the modal, and
+            // only the copy affordance is an exception.
             e.stopPropagation();
             void copyToClipboard();
           }}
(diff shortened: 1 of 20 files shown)


apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
```diff
@@ -52,6 +62,75 @@ const ApiKeyDisplay = ({ apiKey }: Readonly<{ apiKey: string }>) => {
...
+
+  if (!isReadOnly) {
+    columns.push({
+      id: "actions",
+      header: null,
+      srLabel: t("common.actions"),
+      headerClassName: "w-[10%]",
+      // The flex aligns the button; `align` would be inert here, since the button is block-level and the
+      // header has no visible text.
+      cellClassName: "flex justify-end",
```
mattinannt: `cellClassName: "flex justify-end"` sets `display: flex` directly on the `<td>` (the shared `TableCell` primitive applies this class to the cell element itself, alongside its own `align-middle`/`p-4`). Once the `<td>`'s display is overridden away from `table-cell`, `vertical-align: middle` (from `align-middle`) becomes a no-op — `vertical-align` only affects inline-level/table-cell boxes — and per the CSS table anonymous-box rules the browser has to wrap this cell in an anonymous table-cell box to keep the row valid, with that wrapper defaulting to `vertical-align: baseline`. Net effect: the delete button in this column will render vertically top/baseline-aligned instead of centered like the "Label"/"Created At" text in the same row, on every row for every non-read-only user.

This is exactly the failure mode the adjacent `hideBelow` comment in this same file warns about ("`hidden sm:block` would have restored a `<td>` to `display: block` and dropped it out of the table's column layout... `hideBelow` uses `table-cell` for exactly that reason") — the same reasoning wasn't applied here. Suggest wrapping the button in an inner `<div className="flex justify-end">` instead of putting `flex` on `cellClassName`, leaving the `<td>` itself as `table-cell`.

(Single-pass review without the Agent tool / subagent verification — flagging with high confidence based on CSS box-type semantics, not empirically screenshotted.)

---

apps/web/modules/organization/settings/api-keys/loading.tsx
```diff
@@ -16,28 +18,30 @@ const LoadingCard = () => {
...
+          four — the exact failure the shared factory makes unrepresentable.
+
+          `locale` and `onDelete` are never reached: the skeleton renders each column's header and a
+          placeholder bar, and never calls `cell`.
+        */}
+        <SettingsTableSkeleton
+          columns={getApiKeyColumns({
+            t,
+            locale: "en-US",
+            isReadOnly: false,
```
mattinannt: The skeleton always calls `getApiKeyColumns` with `isReadOnly: false`, so it always renders 4 columns (including the "actions" skeleton bar). `getApiKeyColumns` itself only includes the "actions" column `!isReadOnly`, and the four real column widths (`w-[25%]+w-[45%]+w-[20%]+w-[10%]`) are chosen to sum to 100% only when all four are present — for a read-only viewer the real table renders 3 columns totaling 90% of declared width, and the loading skeleton no longer matches it (extra column + different width distribution), which is a visible layout shift the moment `loading.tsx` resolves to the real table for a read-only user. (Today this is masked because `apps/web/modules/organization/settings/api-keys/page.tsx` computes `isReadOnly={!canAccessApiKeys}` after already throwing when `!canAccessApiKeys`, so `isReadOnly` is currently always `false` in production — but that makes this hardcoding a landmine rather than a fix: if that unrelated bug is ever corrected, or another caller passes `isReadOnly=true`, the skeleton/table column-count contract silently breaks with no compiler signal.)

(Single-pass review without the Agent tool / subagent verification; medium confidence — currently unobservable given the separate always-false `isReadOnly` condition upstream.)

---

mattinannt (commented):
Single-pass review at high effort (Agent tool unavailable, no subagent fan-out/verify). Two findings posted inline. Overall this is a careful, well-commented refactor onto the shared `SettingsTable` primitive — most of the angles I worked through (removed-behavior audit, cross-file tracing of `getApiKeyColumns`/`EditAPIKeys` callers, reuse/simplification/efficiency/altitude, CLAUDE.md conventions) turned up nothing else worth flagging with a concrete failure scenario.

---

apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
itsjavi: Correct, and this is the most useful finding on the series so far — fixed in `e9b5c0d`. Your mechanism holds up under checking, and the part I want to acknowledge is the second paragraph: I wrote the `hideBelow` comment two columns above and then did the exact thing it warns about.

The flex now goes on a wrapper inside `cell`, leaving the `<td>` a real table-cell.

**One correction to the reasoning, which matters because it changes a fix I made earlier today.** You wrote "block-level Button/Link children" — they aren't. `Button`'s base class is `inline-flex` (`button/index.tsx:8`), so it's inline-level, and `text-align` *does* move it. What actually made `text-align` inert on that cell was the `display: flex` on the `<td>` blockifying its children — not the children's own display.

That matters because your earlier finding on #8841 — that `align: "right"` was dead config next to `cellClassName: "flex justify-end gap-2"` — was **right about the conclusion but for this same wrong reason**. I removed `align` there and repeated the "buttons are block-level" claim in a comment and in the `align` JSDoc. Both are now corrected on that branch (`091796ca`): `align` moves inline-level content including Buttons and Links, and what makes it inert is a flex on the cell or a block-level wrapper.

**Fixed everywhere the pattern appears, not just here.** I swept for it — `grep 'cellClassName: "[^"]*\(flex\|grid\|block\)'` now returns nothing across `modules/` and `app/`:

| Table | PR | Fix |
| --- | --- | --- |
| api keys | #8871 | flex → wrapper in `cell` |
| feedback directories | #8841 | flex → wrapper in `cell` |
| org teams | #8842 | flex → wrapper in `cell` |
| members | #8842 | flex **removed** — `MemberActions` already renders its own `flex justify-end` wrapper |

The members one is worth calling out: I added that class **an hour ago**, while fixing your `align` remark on #8841. So the previous round of review directly caused a second instance of this bug. It was redundant as well as harmful.

**And documented on the prop**, since three call sites making the same mistake is an API problem, not three coding mistakes:

```ts
/**
 * Extra classes merged onto every `<td>` in this column, e.g. `font-medium text-slate-900`.
 *
 * **Do not change the cell's `display` here** — no `flex`, `grid` or `block`. These classes land on the
 * `<td>` itself, so overriding `display` stops it being a table-cell: the shared `align-middle` becomes
 * a no-op (`vertical-align` applies only to inline-level and table-cell boxes) and the browser wraps the
 * cell in an anonymous table-cell that defaults to baseline, so the content sits high in the row. To lay
 * cell content out, put the flex on a wrapper element **inside** `cell` instead. (`hideBelow` exists for
 * the same reason — it uses `table-cell`, not `block`.)
 */
```

That lands on #8841, which already touches `types.ts` for the `align` doc, so the primitive's docs stay in one PR rather than conflicting across three.

Noted for honesty since you flagged that yours wasn't empirically screenshotted: mine isn't either. No Docker here, so no app to render against. I verified the box-type semantics and `Button`'s actual display class, but the "button sits high in the row" symptom is reasoned, not observed. It's in the QA plan.

---

apps/web/modules/organization/settings/api-keys/loading.tsx
itsjavi: Valid, and I took the root-cause option rather than the workaround — fixed in `e9b5c0d`. **`isReadOnly` is gone entirely.**

Your framing is what decided it: *"that makes this hardcoding a landmine rather than a fix."* I had the same observation in the PR description but drew the weaker conclusion — I noted the prop was dead, kept it "rather than deleting a prop mid-refactor," then hardcoded around it in the skeleton. That leaves a contract with no compiler signal, which is strictly worse than either honest option.

Traced it: `page.tsx:25-27` computes `canAccessApiKeys`, throws `not_authorized` when false, and *then* passes `isReadOnly={!canAccessApiKeys}` — so it is unconditionally `false`. `api-key-list.tsx` is the only other link in the chain and just forwards it. Nothing else in the repo constructs `EditAPIKeys`.

So the prop was dead through three files. Removed from all three, plus the skeleton call:

- `getApiKeyColumns` no longer takes `isReadOnly`, and the actions column is unconditional — the four widths now always sum to 100%, so the skeleton/table column contract can't diverge because there's only one shape.
- `loading.tsx` no longer passes a hardcoded value it had no business knowing (a route-level `loading.tsx` has no auth context, which is the real reason it couldn't be correct).
- `EditAPIKeys`, `ApiKeyList` and `page.tsx` each lose the prop.

Net effect on the diff: it got **smaller**, and the "extra column + different width distribution" failure mode you described is now unrepresentable rather than merely unreachable.

What I did *not* do is fix the underlying oddity — that the page **throws** for a non-owner/manager instead of rendering read-only. That's a behaviour change well outside a restyle, and it's the thing that made `isReadOnly` dead in the first place. Left as a follow-up in the PR description: if someone later wants a read-only view, they should fix the throw and reintroduce the flag deliberately, with the column widths reconsidered at the same time.

One residual cosmetic point I should name rather than let you find: the skeleton's placeholder bar in the actions column is left-aligned while the real button is right-aligned. It can't be fixed through column config — the skeleton renders a block-level bar, so neither `align` nor a `cell` wrapper reaches it. In a 10%-wide column it's a few pixels of a grey rectangle, so I've left it rather than adding primitive API for it.

---

apps/web/modules/organization/settings/api-keys/components/edit-api-keys.tsx
```diff
@@ -52,6 +62,75 @@ const ApiKeyDisplay = ({ apiKey }: Readonly<{ apiKey: string }>) => {
...
+      stopRowClick: true,
+      skeletonWidth: "w-8",
+      cell: (apiKey) => (
+        <Button size="icon" variant="ghost" onClick={(event) => onDelete(event, apiKey)}>
+          <TrashIcon />
+        </Button>
+      ),
```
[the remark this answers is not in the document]
itsjavi: Both findings valid — fixed in `e9b5c0d`.

**The `flex` on `cellClassName`:** confirmed. That class lands on the `<td>` itself, so `display: flex` stops it being a table-cell — which kills the shared `align-middle` (`vertical-align` applies only to inline-level and table-cell boxes) and makes the browser wrap the cell in an anonymous table-cell defaulting to baseline. The flex now goes on a wrapper inside `cell`. Fixed in the three sibling tables too (#8841, #8842) and documented on the `cellClassName` prop, since three call sites making the same mistake is an API problem — detail in [this thread](https://github.com/formbricks/formbricks/pull/8871#discussion_r3783780718).

**The `aria-label`:** correct, and it was a pre-existing gap rather than one I introduced — the button has only a `TrashIcon` inside, so it has had no accessible name at all. Added `aria-label={t("common.delete")}`, reusing the existing key.

Two notes on scope, since you flagged both sites as the same fix:

- The two line ranges you list are the same column definition, so it's one change, not two.
- The suggested `aria-label` value was already the right one — `common.delete` exists and is what `MemberActions` uses for its own trash tooltip, so this stays consistent with the neighbouring table.

While in there: the sibling icon-only buttons in `MemberActions` (share-invite, resend-invite) have the same missing-accessible-name problem, with tooltips but no `aria-label`. Out of scope here — different module, different PR — but worth a follow-up if you want it tracked.

Verification: typecheck 12/12 tasks 0 errors, 25/25 unit tests, eslint clean, `prettier --check` clean, `pnpm scan-translations` passes.

---

Dhruwang (approved):

PR #8868, opened 2026-08-14
refactor: route feedback dataset authorization through central interface

BhagyaAmarasinghe:
## What & why

Feedback Dataset authorization still bypassed the central `can(actor, action, resource)` interface on several product surfaces, which meant the new SpiceDB model could not be compared safely in shadow mode. This PR routes the current Feedback Directory rules through exact directory or directory/workspace-assignment resources while preserving PostgreSQL and the legacy evaluator as the authority.

The routing covers Feedback Dataset administration, server-rendered Unify pages, feedback records, sources and CSV imports, taxonomy, chart queries and creation, V3 operations, MCP-backed paths, and the Feedback Records gateway. It also introduces the bounded `feedback_gateway:user` and `feedback_gateway:apiKey` shadow targets. Existing entitlement, archive, OAuth-scope, Hub tenant, ownership/integrity, HTTP-status, and API-key single-workspace safeguards remain in place.

The ENG-2398 foundation from #8867 is now merged into `epic/authzed`; this routing PR targets that branch directly. The earlier #8863, #8864, #8805, and #8846 prerequisites are also present in the epic branch.

## Linear ticket

Fixes ENG-2398

## How this was tested

- `CUBEJS_API_URL=http://localhost:4000 HUB_API_URL=http://localhost:3003 HUB_API_KEY=test pnpm --filter @formbricks/web exec vitest run 'app/api/(internal)/unify-feedback/sources/csv/import/route.test.ts' lib/authorization lib/authzed app/api/v3/unify-feedback app/api/v3/feedbackRecords modules/hub modules/ee/feedback-directory modules/ee/analysis lib/feedback-source modules/mcp` — 102 files and 1,795 tests passed.
- Focused routing regression suite — 10 files and 222 tests passed.
- `pnpm --filter @formbricks/web typecheck` passed.
- Changed-file lint completed with 0 errors; 46 pre-existing warnings remain outside this change.
- `pnpm authzed:validate` passed with 31 relationships, 144 assertions, and 2 expected relations.
- `pnpm authzed:smoke` passed the isolated schema, relationship, persistence, health, outage, and secret-safety lifecycle.
- `pnpm build --filter=@formbricks/web... --force` passed with AuthZed disabled and credentials absent.
- The same build passed with AuthZed enabled and `127.0.0.1:1` as an unreachable endpoint, confirming no build-time RPC.
- A temporary Client Component importing the new direct runtime failed `next build` at the expected `server-only` boundary; the fixture was removed and the clean build rerun.
[...]

PR #8872, opened 2026-08-14
refactor(authorization): add a page surface so server-rendered route checks are comparable

xernobyl:
## What & why

Server-rendered route checks were correct and simultaneously invisible.

`can()` resolves a rollout target from the authorization surface on the async stack. Five surfaces existed — `server_action`, `api_v1`, `api_v2`, `api_v3`, `mcp` — and none of them covered React Server Components. So when a page or layout called `can()`, `getAuthorizationRolloutTarget()` returned `null`, the coordinator short-circuited straight to the legacy evaluator (`coordinator.ts:236`), and **no shadow comparison was ever scheduled**. Every product page in the app renders through this path, which means the largest single class of authorization decisions in the product contributed nothing to the parity evidence the epic's cutover depends on.

This adds a sixth surface, `page`, plus the matching `page:user` rollout target, and opens it at the choke points product routes already funnel through — `getWorkspaceAuth`, `workspaceIdLayoutChecks`, and `getWorkspaceLayoutData`. Between them those three carry the organization, workspace, team, survey, dashboard and response paths, so most of the inventory is covered by three wrappers rather than dozens of call sites: the surface opens *inside* each helper, so its many callers need no change.

Three route gates additionally needed touching at the route level:

| File | Before | After |
| --- | --- | --- |
| `landing/layout.tsx` | `getMembershipByUserIdOrganizationId` truthiness | `organization.read` |
| `workspaces/new/layout.tsx` | denylist: `if (isMember \|\| isBilling) notFound()` | allowlist: `organization.manage` |
| `(onboarding)/organizations/[organizationId]/layout.tsx` | already `can()` via `canUserAccessOrganization` | unchanged; surface only |

The onboarding parent layout matters more than "surface only" suggests: it is the parent of the two child layouts above, and a parent layout renders in its own async context, so a child's surface does not extend upward.

**The `landing` swap is set-equivalent.** The schema grants `organization.read` to `owner + manager + member + billing` (`schema.zed:69`) — every membership role and nobody else — so it admits exactly the principals the truthiness check admitted.

**The `workspaces/new` swap, stated accurately.** The old check was a denylist naming only `member` and `billing`, so a user with *no membership in the organization at all* produced all-false flags and fell through it. `organization.manage` (`owner + manager`, `schema.zed:82`) is the allowlist that check was reaching for.

I originally described the non-member case as a deliberate narrowing. **That was overstated, and I corrected it after standing the app up and driving a real non-member at the route** — the parent onboarding layout already refuses them via `canUserAccessOrganization`, throwing `AuthorizationError` before this subtree completes. So no principal changes side here either; all three route-level swaps are behaviour-preserving.

What the allowlist does buy is that this layout no longer *depends* on the parent. RSC renders a parent and its child concurrently, so a child that admits everyone and then invalidates a billing cache is relying on render interleaving to stay correct. The roles this line actually decides are `member` and `billing`, exactly as before — the intent is now stated directly rather than inferred from which roles were named for rejection.

### Why the surface is opened per choke point, not per request

Next.js gives no RSC equivalent of the action-client or API wrapper, and a layout's render and its page's render are separate async contexts — there is no single boundary to wrap. `withAuthorizationSurface` returns early when a surface is already open, so opening it at several choke points is idempotent rather than nesting.

One consequence worth knowing when reading the checks-per-request histogram: a navigation that runs both a layout check and a page check records **two observations rather than one**, splitting that request's count. It weakens the N+1 signal slightly; it does not make it wrong. That is why the wrapper sits at as few and as high a point as possible. Documented in the `TAuthorizationSurface` docstring.

### Deliberate asymmetry

There is no `page:apiKey` target. Pages are session-authenticated; an API-key actor on this surface stays on the legacy path rather than silently acquiring a rollout cohort it was never meant to belong to. Pinned by a test.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2388

## How this was tested

- `pnpm --dir apps/web exec vitest run lib/authorization lib/authzed modules/workspaces modules/organization` → **60 files, 819 tests, all passing** ✅
- Full web unit suite → `6 failed | 8235 passed (8241)`. All 6 are pre-existing `scripts/authzed-entrypoints.test.ts` failures caused by a Node 26 deprecation warning breaking a `stderr === ""` assertion — that file is absent from this diff (`git diff --name-only` → 0 matches) and CI runs Node 24, where it is green.
- `pnpm turbo run typecheck --filter=@formbricks/web` → 12/12 tasks ✅
- `eslint` + `prettier --check` on every changed file → clean ✅
- `pnpm authzed:validate` → `36 relationships loaded, 149 assertions run, 2 expected relations validated` ✅ (confirms the ENG-2340 schema fix this branch builds on is intact)

**End-to-end proof against a real database.** `lib/authorization/checks-per-request.integration.test.ts` → **4 passed** ✅ with the `page` surface, confirming it resolves a rollout target through the real `can()` and coordinator, and that the O(1) claim still holds at 50 vs 3,000 surveys.

Those tests previously declared `server_action` as a stand-in — the scenario they model is a page render, but no page surface existed to name. That substitution was itself the bug: the checks they counted were attributed to the wrong surface, so the histogram this file exists to defend reported page renders as server-action traffic. They now declare `page`.

**Tests added:** one in `context.test.ts` pinning `page:user` resolution and the absence of `page:apiKey`.

**Known limitation, deliberately not folded in.** `getOrganizationAuth` (`modules/organization/lib/utils.ts:16`) gates 13 pages and throws on a missing membership — a real authorization decision — but it never calls `can()`; it reads the membership directly and derives role flags from it. A surface there would wrap nothing, and routing it through `can()` would be *additive* (the membership row is still needed for the role flags those pages consume), costing an extra check per render against this ticket's no-amplification criterion. Filed separately rather than widened into this PR.

**Coverage gap, stated plainly — including a spec I wrote, tested, and then threw away.** `playwright/billing-role-access.spec.ts` (ENG-1763) already covers the `getWorkspaceAuth` path, which is the bulk of this diff. The two onboarding layouts are **not** covered by any existing spec (`onboarding.spec.ts` visits neither `workspaces/new` nor `landing`).

I tried to close that. I stood the app up locally and wrote a four-case spec, and it went green — but mutation testing showed it was worthless: with the `organization.manage` gate **deleted entirely**, every assertion still passed. Two reasons, both worth knowing before anyone else attempts this:

1. `/organizations/<id>/workspaces/new` has **no `page.tsx`** — it is a layout over four child routes (`plan`, `ai`, `survey`, `templates`). Any assertion against that bare segment 404s through Next.js routing regardless of what the layout decided.
2. Retargeting a real child route did not fix it, because these routes sit behind a stack of gates and redirects — the parent layout's `organization.read`, this gate, the child page's `IS_FORMBRICKS_CLOUD` redirect, `redirectIfOnboardingComplete`, and `getOrganizationAuth` — so a refusal at the browser tells you almost nothing about *which* gate produced it.

So the spec is not in this PR: a test that passes with the feature removed is worse than no test. The debugging was not wasted — it is what caught the overstated narrowing claim above. Flagging the gap rather than papering it with a green tick; if we want real coverage here it needs a route with fewer confounds, which I'd rather scope properly than bolt on.

## Breaking changes

None.

`page:user` is a new *permitted value* for the existing `AUTHZED_AUTHORIZATION_ROLLOUT_TARGETS` contract, not a new env var, and no default enables it — it behaves like every other target: opt-in via `AUTHZED_ENFORCEMENT_TARGETS` / shadow config. With nothing configured, behavior is byte-identical to today.

## QA / Test Plan

**How to test**

- [ ] Sign in as an org **owner** → navigate to any workspace (`/workspaces/<id>/`) → the workspace loads exactly as before, no redirect, no error.
- [ ] Sign in as a **billing-role** member → navigate to `/workspaces/<id>/` → still redirected to the billing/enterprise screen, unchanged.
- [ ] Sign in as an org **member with no team grant** → navigate to `/workspaces/<id>/` → still rejected, unchanged.
- [ ] As an **owner or manager** → visit `/organizations/<orgId>/workspaces/new` → the create-workspace screen renders.
- [ ] As a plain **member** or **billing** user → visit `/organizations/<orgId>/workspaces/new` → **404**, unchanged from today.
- [ ] Signed in as a user who is *not a member of that organization at all* → visit `/organizations/<otherOrgId>/workspaces/new/plan` → refused, same as before. (The parent onboarding layout already refuses non-members, so this is unchanged — verified against a running app.)
- [ ] As any org member → visit `/organizations/<orgId>/landing` → behaves as before (redirects to your first workspace if you have one; otherwise renders the landing screen).
- [ ] As a **non-member** → visit `/organizations/<otherOrgId>/landing` → **404**, unchanged.

**Preconditions / test data**

- One organization with an owner, a manager, a plain member, and a billing-role member.
- A second organization the first org's users do not belong to (needed for the non-member case).
- At least one workspace, and one user with no workspace at all (to exercise the landing screen rather than its redirect).
- No feature flag required — this PR changes no defaults. To exercise the new surface under shadow comparison, add `page:user` to the authorization shadow targets.

**Risks & regressions**

- Verify owners/managers still reach the create-workspace flow, and that `member`/`billing` are still refused — those are the roles this gate decides.
- Every product page now opens an authorization surface. Worth confirming pages still render normally under load and that no page throws where it previously rendered.
- The checks-per-request histogram now attributes page renders to `page` rather than `server_action`. Any dashboard or alert filtering on `server_action` will see its page-render traffic move to the new label — a reporting change, not a behavior change, but it will look like a drop if you are watching that series.

- none. No DB migration, no new env var, no `.env.example` change.

## Stack maintenance (2026-08-16)

Rebuilt the branch onto the current `epic/authzed` head (`137663366`) instead of carrying its previous merge commit. The only textual conflict was additive: the base added `feedback_gateway`, while this PR adds `page`; both rollout surfaces and both contract tests are preserved. The obsolete setup-invite merge note is no longer part of this parent PR because ENG-2409/#8873 now owns that centralization.

The repaired head is `e56d27dde`, with an eight-file ENG-2388 diff and no duplicated feedback-dataset changes. Exact-head validation: 60 focused suites / 819 tests, 12/12 web typecheck tasks, changed-file ESLint and Prettier, and AuthZed validation with 36 relationships, 149 assertions, and 2 expected relations.

### DIFF
--- a/apps/web/lib/authorization/context.ts
+++ b/apps/web/lib/authorization/context.ts
@@ -17,10 +17,25 @@ import { recordAuthorizationChecksPerRequest } from "./metrics";
  * route already funnels through. `withAuthorizationSurface` returns early when a surface is already
  * open, so opening it at more than one choke point is idempotent rather than nesting.
  *
- * The practical consequence, stated because it affects how the checks-per-request histogram reads:
- * one navigation that runs both a layout check and a page check records two observations rather
- * than one. That splits the count for that request — it weakens the N+1 signal slightly, it does
- * not make it wrong — which is why the wrapper is applied at as few and as high a point as possible.
+ * Two consequences follow, and the second is a real coverage limit rather than a reporting quirk.
+ *
+ * The histogram splits: one navigation that runs both a layout check and a page check records two
+ * observations rather than one. That weakens the N+1 signal slightly; it does not make it wrong.
+ *
+ * **The surface does not span the whole page.** `AsyncLocalStorage` scopes to the awaited callback,
+ * so it closes when the choke-point helper returns — a page that calls `getWorkspaceAuth()` and then
+ * issues further central checks (`feedbackDirectoryAssignment.read` via
+ * `getAuthorizedWorkspaceFeedbackDirectories`, `organization.manage` on the taxonomy page) makes
+ * those checks outside any surface. `can()` then answers from the legacy evaluator whatever the
+ * rollout selects, so they produce no shadow evidence *and* would bypass enforcement. Those checks
+ * are counted by `formbricks_authzed_authorization_unscoped_checks_total`, because otherwise the
+ * gap is indistinguishable from a clean cutover: nothing compares, so nothing ever mismatches.
+ *
+ * Closing it needs a boundary that survives past the helper — a request-scoped store (React `cache`)
+ * rather than an async-scoped one. That is deliberately out of scope here, because it replaces this
+ * module's context mechanism and cannot be covered by a unit test: `cache` does not dedupe outside a
+ * render, so proving it needs a render harness or an E2E. Until that lands, `page:user` belongs in
+ * shadow only and must not be added to an enforcement list.
  */
 export type TAuthorizationSurface =
   | "server_action"
--- a/apps/web/lib/authorization/context.ts
+++ b/apps/web/lib/authorization/context.ts
@@ -11,7 +11,7 @@ import { recordAuthorizationChecksPerRequest } from "./metrics";
 /**
  * `page` is the server-rendered route surface (React Server Components), added for ENG-2388.
  *
- * Unlike the other five, it is not established at a single request boundary: Next.js gives no RSC
+ * Unlike every other surface, it is not established at a single request boundary: Next.js gives no RSC
  * equivalent of the action-client or API wrapper, and a layout's render and its page's render are
  * separate async contexts. It is therefore opened at the authorization choke points every product
  * route already funnels through. `withAuthorizationSurface` returns early when a surface is already
--- a/apps/web/lib/authorization/context.ts
+++ b/apps/web/lib/authorization/context.ts
@@ -8,8 +8,23 @@ import {
 import type { TAuthorizationActor } from "./contract";
 import { recordAuthorizationChecksPerRequest } from "./metrics";
 
+/**
+ * `page` is the server-rendered route surface (React Server Components), added for ENG-2388.
+ *
+ * Unlike the other five, it is not established at a single request boundary: Next.js gives no RSC
+ * equivalent of the action-client or API wrapper, and a layout's render and its page's render are
+ * separate async contexts. It is therefore opened at the authorization choke points every product
+ * route already funnels through. `withAuthorizationSurface` returns early when a surface is already
+ * open, so opening it at more than one choke point is idempotent rather than nesting.
+ *
+ * The practical consequence, stated because it affects how the checks-per-request histogram reads:
+ * one navigation that runs both a layout check and a page check records two observations rather
+ * than one. That splits the count for that request — it weakens the N+1 signal slightly, it does
+ * not make it wrong — which is why the wrapper is applied at as few and as high a point as possible.
+ */
 export type TAuthorizationSurface =
   | "server_action"
+  | "page"
   | "api_v1"
   | "api_v2"
   | "api_v3"
--- a/apps/web/lib/authzed/rollout-contract.ts
+++ b/apps/web/lib/authzed/rollout-contract.ts
@@ -7,6 +7,7 @@
  */
 export const AUTHZED_AUTHORIZATION_ROLLOUT_TARGETS = [
   "server_action:user",
+  "page:user",
   "api_v1:user",
   "api_v1:apiKey",
   "api_v2:apiKey",
--- a/apps/web/modules/workspaces/lib/utils.ts
+++ b/apps/web/modules/workspaces/lib/utils.ts
@@ -306,79 +306,91 @@ export const getWorkspaceWithRelations = reactCache(async (workspaceId: string, 
 /**
  * Fetches all data required for workspace layout rendering.
  * Resolves the production environment automatically.
+ *
+ * Opened as the `page` surface for the same reason as its two siblings above (ENG-2388). This one
+ * backs the top-level `/workspaces/[workspaceId]` layout, so it is the highest-traffic authorization
+ * gate of the three — its `canUserNavigateWorkspace` call reaches `can()` on essentially every
+ * product navigation, and without a surface on the stack every one of those decisions resolved no
+ * rollout target and scheduled no shadow comparison.
  */
 export const getWorkspaceLayoutData = reactCache(
-  async (workspaceId: string, userId: string): Promise<TWorkspaceLayoutData> => {
-    validateInputs([workspaceId, ZId]);
-    validateInputs([userId, ZId]);
-
-    const t = await getTranslate();
-    const session = await getSession();
-
-    if (!session?.user) {
-      throw new AuthenticationError(t("common.not_authenticated"));
-    }
-
-    if (session.user.id !== userId) {
-      throw new AuthenticationError("User ID mismatch with session");
-    }
-
-    const user = await getUser(userId);
-    if (!user) {
-      throw new AuthenticationError(t("common.not_authenticated"));
-    }
-
-    // Resolved first so the navigation gate below can name the owning organization. This is
-    // the same request-memoized read the rest of this function already relied on, so it costs
-    // nothing extra; it only moves.
-    const relationData = await getWorkspaceWithRelations(workspaceId, userId);
-    if (!relationData) {
-      throw new ResourceNotFoundError(t("common.workspace"), workspaceId);
-    }
-
-    const { workspace, organization, membership } = relationData;
-
-    // Navigation, not data — the layout shell a billing-role member passes through on the way
-    // to billing. Product data on the pages inside stays gated on workspace.read.
-    const hasAccess = await canUserNavigateWorkspace(userId, {
-      id: workspace.id,
-      organizationId: organization.id,
-    });
-    if (!hasAccess) {
-      throw new AuthorizationError(t("common.not_authorized"));
-    }
-
-    if (!membership) {
-      throw new AuthorizationError(t("common.membership_not_found"));
-    }
-
-    const [isAccessControlAllowed, workspacePermission, license] = await Promise.all([
-      getAccessControlPermission(organization.id),
-      getWorkspacePermissionByUserId(userId, workspace.id),
-      getEnterpriseLicense(),
-    ]);
-
-    let responseCount = 0;
-    if (IS_FORMBRICKS_CLOUD) {
-      responseCount = await getMonthlyOrganizationResponseCount(organization.id);
-    }
-
-    return {
-      session,
-      user,
-      workspace,
-      organization: {
-        ...organization,
-        billing: {
-          ...organization.billing,
-          stripe: organization.billing.stripe ?? undefined,
-        },
-      },
-      membership,
-      isAccessControlAllowed,
-      workspacePermission,
-      license,
-      responseCount,
-    };
-  }
+  async (workspaceId: string, userId: string): Promise<TWorkspaceLayoutData> =>
+    withAuthorizationSurface("page", () => resolveWorkspaceLayoutData(workspaceId, userId))
 );
+
+const resolveWorkspaceLayoutData = async (
+  workspaceId: string,
+  userId: string
+): Promise<TWorkspaceLayoutData> => {
+  validateInputs([workspaceId, ZId]);
+  validateInputs([userId, ZId]);
+
+  const t = await getTranslate();
+  const session = await getSession();
+
+  if (!session?.user) {
+    throw new AuthenticationError(t("common.not_authenticated"));
+  }
+
+  if (session.user.id !== userId) {
+    throw new AuthenticationError("User ID mismatch with session");
+  }
+
+  const user = await getUser(userId);
+  if (!user) {
+    throw new AuthenticationError(t("common.not_authenticated"));
+  }
+
+  // Resolved first so the navigation gate below can name the owning organization. This is
+  // the same request-memoized read the rest of this function already relied on, so it costs
+  // nothing extra; it only moves.
+  const relationData = await getWorkspaceWithRelations(workspaceId, userId);
+  if (!relationData) {
+    throw new ResourceNotFoundError(t("common.workspace"), workspaceId);
+  }
+
+  const { workspace, organization, membership } = relationData;
+
+  // Navigation, not data — the layout shell a billing-role member passes through on the way
+  // to billing. Product data on the pages inside stays gated on workspace.read.
+  const hasAccess = await canUserNavigateWorkspace(userId, {
+    id: workspace.id,
+    organizationId: organization.id,
+  });
+  if (!hasAccess) {
+    throw new AuthorizationError(t("common.not_authorized"));
+  }
+
+  if (!membership) {
+    throw new AuthorizationError(t("common.membership_not_found"));
+  }
+
+  const [isAccessControlAllowed, workspacePermission, license] = await Promise.all([
+    getAccessControlPermission(organization.id),
+    getWorkspacePermissionByUserId(userId, workspace.id),
+    getEnterpriseLicense(),
+  ]);
+
+  let responseCount = 0;
+  if (IS_FORMBRICKS_CLOUD) {
+    responseCount = await getMonthlyOrganizationResponseCount(organization.id);
+  }
+
+  return {
+    session,
+    user,
+    workspace,
+    organization: {
+      ...organization,
+      billing: {
+        ...organization.billing,
+        stripe: organization.billing.stripe ?? undefined,
+      },
+    },
+    membership,
+    isAccessControlAllowed,
+    workspacePermission,
+    license,
+    responseCount,
+  };
+};
--- a/apps/web/modules/workspaces/lib/utils.ts
+++ b/apps/web/modules/workspaces/lib/utils.ts
@@ -11,6 +11,7 @@ import {
   ResourceNotFoundError,
 } from "@formbricks/types/errors";
 import { can } from "@/lib/authorization";
+import { withAuthorizationSurface } from "@/lib/authorization/context";
 import { IS_FORMBRICKS_CLOUD } from "@/lib/constants";
 import { getBillingFallbackPath } from "@/lib/membership/navigation";
 import { getMembershipByUserIdOrganizationId } from "@/lib/membership/service";
@@ -36,8 +37,19 @@ import { TWorkspaceAuth, TWorkspaceLayoutData } from "@/modules/workspaces/types
  * route. Billing-role members are redirected to billing/enterprise screens; any org
  * member without a WorkspaceTeam grant (and who is not an owner/manager) is rejected
  * with an AuthorizationError instead of being silently admitted as a writer.
+ *
+ * Opened as the `page` authorization surface (ENG-2388). Every product page funnels through here,
+ * and the `can()` calls below were already central — but without a surface on the stack the
+ * coordinator has no rollout target to match, short-circuits to the legacy evaluator, and schedules
+ * no shadow comparison. So these decisions were correct and simultaneously invisible to parity
+ * evidence. Wrapping here is what makes them comparable; it changes no decision.
  */
-export const getWorkspaceAuth = reactCache(async (workspaceId: string): Promise<TWorkspaceAuth> => {
+export const getWorkspaceAuth = reactCache(
+  async (workspaceId: string): Promise<TWorkspaceAuth> =>
+    withAuthorizationSurface("page", () => resolveWorkspaceAuth(workspaceId))
+);
+
+const resolveWorkspaceAuth = async (workspaceId: string): Promise<TWorkspaceAuth> => {
   const t = await getTranslate();
 
   const [workspace, session] = await Promise.all([getWorkspace(workspaceId), getSession()]);
@@ -118,12 +130,19 @@ export const getWorkspaceAuth = reactCache(async (workspaceId: string): Promise<
     hasManageAccess,
     isReadOnly,
   };
-});
+};
 
 /**
  * Lightweight layout checks for workspace routes (survey editor, onboarding).
+ *
+ * Opened as the `page` surface for the same reason as `getWorkspaceAuth` (ENG-2388): the navigation
+ * gate below already routes through `can()` via `canUserNavigateWorkspace`, but a layout render is
+ * its own async context, so it needs its own surface to be comparable.
  */
-export const workspaceIdLayoutChecks = async (workspaceId: string) => {
+export const workspaceIdLayoutChecks = async (workspaceId: string) =>
+  withAuthorizationSurface("page", () => resolveWorkspaceIdLayoutChecks(workspaceId));
+
+const resolveWorkspaceIdLayoutChecks = async (workspaceId: string) => {
   const t = await getTranslate();
   const session = await getSession();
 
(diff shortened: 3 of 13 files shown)


mattinannt (commented):
Nice write-up and testing on the intended surfaces. Two sibling gaps that look like the same bug this PR is fixing, just missed:

**1. `apps/web/modules/workspaces/lib/utils.ts:310` — `getWorkspaceLayoutData` isn't wrapped in `withAuthorizationSurface("page", ...)`.**

It's the auth check backing `apps/web/app/(app)/workspaces/[workspaceId]/layout.tsx` — the top-level layout for essentially every `/workspaces/[workspaceId]/...` route. Its `canUserNavigateWorkspace` → `can()` call still runs with no surface on the stack, so `getAuthorizationRolloutTarget` returns `null` and it silently falls back to the legacy evaluator (no checks-per-request metric, no shadow comparison) even once `page:user` is enabled. Its sibling `workspaceIdLayoutChecks`, a few lines down in this same file, gets exactly this fix — this one looks like it was just missed, and it's the highest-traffic gate of the bunch.

**2. `apps/web/app/(app)/(onboarding)/organizations/[organizationId]/layout.tsx` — the onboarding parent layout isn't wrapped either.**

It calls `canUserAccessOrganization` (→ `can()` with `organization.read`) but this file isn't touched by the diff at all, even though this PR wraps both of its child layouts (`landing/layout.tsx`, `workspaces/new/layout.tsx`) for exactly this reason. A request to either child route renders this parent first, so that `organization.read` decision runs outside any surface and stays invisible to the rollout, one layer above two siblings this PR did fix.

Neither line lands inside this diff's hunks, so posting as a general comment rather than forcing an inline anchor. Everything else — the `landing`/`workspaces/new` behavior analysis, the `page:apiKey` asymmetry test, the integration test surface fix — looks solid.

---

xernobyl:
Both confirmed and fixed in `75a4be04e`. Thanks — these were genuine misses, not judgement calls, and the second one is the more embarrassing of the two.

**1. `getWorkspaceLayoutData`** — extracted to `resolveWorkspaceLayoutData` and wrapped, matching the shape of its two siblings in the same file. You're right that it's the highest-traffic gate of the three: it backs `app/(app)/workspaces/[workspaceId]/layout.tsx`, so it runs on essentially every product navigation. I wrapped the two functions I had gone looking for and never re-read the file for a third.

**2. The onboarding parent layout** — wrapped. This is the one that should have been obvious: I wrapped both children *because* the parent's surface would not reach them, which is exactly the reasoning that should have sent me one directory up. A parent layout renders in its own async context, so the org-level `organization.read` decision guarding both children was invisible while the children themselves were comparable.

**I swept for more rather than just fixing the two.** Enumerated every helper that reaches `can()` and cross-referenced against every `page.tsx` / `layout.tsx` calling one. The result is clean now — the many apparent "unwrapped" callers of `getWorkspaceAuth` are fine, since the surface opens *inside* the helper rather than at each call site.

**One thing the sweep turned up that I deliberately did not fold in.** `getOrganizationAuth` (`modules/organization/lib/utils.ts:16`) gates 13 pages and throws `ResourceNotFoundError` on a missing membership — a real authorization decision — but it never calls `can()`. It reads `getMembershipByUserIdOrganizationId` directly and derives role flags from `membership.role`. So a surface there would wrap nothing; making it comparable means routing it through `can()`, which is a different and larger change:

- The membership row is still needed for the `isMember`/`isOwner`/`isManager`/`isBilling` flags those 13 pages consume, so a `can()` call would be **additive** rather than a replacement — one extra check per page render, which rubs against this ticket's "avoid per-item request amplification" criterion.
- It changes what those pages are gated on, which is not the exact-preservation this ticket asks for.

That reads to me as its own ticket rather than scope creep on this one — filed as [ENG-2409](https://linear.app/formbricks/issue/ENG-2409) with the three options laid out (accept the extra check, collapse the flags into the permission path, or accept the gap and document those pages as legacy-only for the cutover). Happy to be told otherwise if you would rather see it here.

CI was fully green on the previous head (unit, E2E, Sonar 0 new issues, CodeRabbit 0 comments); re-running now. Locally: 758 targeted tests green, typecheck 10/10, and the checks-per-request integration tests still pass 4/4 against a live database.

xernobyl:
## ⚠️ Merge-order note: this PR conflicts with #8863, and resolving it the obvious way loses something

Simulated the merge against `bhagya/sync-main-into-epic-authzed` (#8863). One conflict, and it is **semantic rather than textual** — worth flagging before anyone resolves it by picking a side.

**File:** `apps/web/modules/setup/organization/[organizationId]/invite/page.tsx`

Main replaced the gate I wrapped. Where the epic has `verifyUserRoleAccess`, main now calls a new `hasSetupInviteAccess` (`.../invite/lib/authorization.ts`), added by ENG-2169 to narrow the onboarding invite path to **owners only** — because `inviteUser` always persists an owner invite, so allowing managers there lets a manager mint an owner without an existing owner's approval.

**That change is correct and must survive the merge.** It is a real privilege-escalation narrowing. Do not resolve this conflict by keeping the epic's `verifyUserRoleAccess` side.

**But it also silently undoes what this PR did to that file.** `hasSetupInviteAccess` reads the membership directly and checks a role list:

```ts
const membership = await getMembershipByUserIdOrganizationId(userId, organizationId);
return membership ? SETUP_INVITE_ROLES.includes(membership.role) : false;
```

No `can()`. So once #8863 lands, wrapping that call site in `withAuthorizationSurface("page", …)` wraps nothing, and the gate drops back out of shadow comparison — precisely the blind spot this PR exists to close, reintroduced from the main side. Nobody did anything wrong; the two changes were authored against different branches and are individually right.

**Resolution when #8863 merges** (I'll do this, it's my conflict to resolve):

1. Take **main's** side wholesale — `hasSetupInviteAccess`, owner-only.
2. Drop the surface wrapper from this file, since there is no longer a `can()` call under it to make comparable.
3. Log the gate as legacy-only against [ENG-2409](https://linear.app/formbricks/issue/ENG-2409), which already tracks exactly this class — page gates that decide access without going through `can()` at all, and so need *routing* rather than a surface. It now covers two instances instead of one.

Net effect on this PR: one of the four route-level gates goes away after the sync. The other three, and the three choke-point wrappers that carry the bulk of the traffic, are unaffected.

**Suggested order:** merge #8863 first (it is the base of your stack anyway), then I rebase this and resolve as above. Flagging rather than pre-resolving, since #8863 hasn't landed and I'd only be guessing at its final head.

xernobyl:
**Correction to the PR description, from standing the app up rather than reading it.**

I claimed the `workspaces/new` allowlist was a deliberate narrowing that denies a non-member the old denylist let through. **That was wrong.** Running a real non-member at the route shows the parent onboarding layout refuses them first:

```
Error [AuthorizationError]: Not authorized
    at WorkspaceOnboardingLayout (app/(app)/(onboarding)/organizations/[organizationId]/layout.tsx:41:11)
```

So no principal changes side — all three route-level swaps are behaviour-preserving, which is what the ticket asked for. What the allowlist still buys is that this layout no longer *depends* on the parent: RSC renders parent and child concurrently, so a child that admits everyone and then invalidates a billing cache is relying on render interleaving to stay correct. Description and the source comment are both fixed (`16fb7c883`); code is unchanged.

**Also worth recording: I wrote an E2E spec for these two layouts, got it green, and deleted it.** Mutation testing killed it — with the `organization.manage` gate *deleted entirely*, all four assertions still passed. Two traps for whoever tries next:

1. `/organizations/<id>/workspaces/new` has **no `page.tsx`**; it is a layout over four child routes. Asserting against that bare segment 404s through Next.js routing no matter what the layout decided.
2. Retargeting a real child route does not rescue it — the parent's `organization.read`, this gate, the child's `IS_FORMBRICKS_CLOUD` redirect, `redirectIfOnboardingComplete` and `getOrganizationAuth` all sit in the same path, so a browser-level refusal does not tell you which gate produced it.

A test that passes with the feature removed is worse than no test, so it is not in the PR. The gap is called out in the description instead.

xernobyl:
**#8863 landed, and the conflict I flagged above materialised exactly as described. Resolved in `c57ab2ac6`; this PR is `MERGEABLE` again.**

One conflict, `modules/setup/.../invite/page.tsx`, resolved as **neither side wholesale**:

- **Main's semantics win.** `hasSetupInviteAccess` is owner-only because `inviteUser` always persists an *owner* invite, so admitting managers would let a manager mint an owner without an existing owner's approval (ENG-2169). Real privilege narrowing — not reverted.
- **This branch's surface wrapper is dropped**, not reapplied on top. `hasSetupInviteAccess` reads the membership row and tests a role list, so there is no `can()` beneath the wrapper for it to make comparable — keeping it would emit a zero-check observation and nothing else.

Net effect: one of the four route-level gates in this PR goes away. The other three, and the three choke-point wrappers carrying the bulk of the traffic, are untouched. Restoring coverage on that gate means routing the helper itself — tracked as instance 2 of [ENG-2409](https://linear.app/formbricks/issue/ENG-2409). The reasoning is recorded in the file so the next reader doesn't have to reconstruct it.

Re-verified on the merged tree: typecheck 12/12, `authzed:validate` 143 assertions, full web suite `6 failed | 8670 passed` (the 6 are the known pre-existing Node-26 `authzed-entrypoints` failures, absent from this diff, green on CI's Node 24).

The sync also brought `pnpm format:check` into CI, which did not exist when this branch was written — it flagged one file of mine, now fixed.

apps/web/lib/authorization/context.ts
```diff
@@ -8,8 +8,23 @@ import {
 import type { TAuthorizationActor } from "./contract";
 import { recordAuthorizationChecksPerRequest } from "./metrics";
 
+/**
+ * `page` is the server-rendered route surface (React Server Components), added for ENG-2388.
+ *
+ * Unlike the other five, it is not established at a single request boundary: Next.js gives no RSC
```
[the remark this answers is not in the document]
xernobyl: Correct, and worth recording *why* it drifted. Fixed in `113d69b6e`.

The count was accurate when written — six surfaces, five others. It went stale because this branch was rebased onto an epic that had gained `feedback_gateway` from #8868, so the union is now seven and there are six others.

Rather than changing five to six, I dropped the count: "Unlike **every other surface**, it is not established at a single request boundary…". The number is incidental to the point, and correcting it to six would just queue up the same rot the next time a surface lands.

apps/web/modules/workspaces/lib/utils.ts
```diff
@@ -36,8 +37,19 @@ import { TWorkspaceAuth, TWorkspaceLayoutData } from "@/modules/workspaces/types
...
+ * Opened as the `page` authorization surface (ENG-2388). Every product page funnels through here,
+ * and the `can()` calls below were already central — but without a surface on the stack the
+ * coordinator has no rollout target to match, short-circuits to the legacy evaluator, and schedules
+ * no shadow comparison. So these decisions were correct and simultaneously invisible to parity
+ * evidence. Wrapping here is what makes them comparable; it changes no decision.
  */
-export const getWorkspaceAuth = reactCache(async (workspaceId: string): Promise<TWorkspaceAuth> => {
+export const getWorkspaceAuth = reactCache(
+  async (workspaceId: string): Promise<TWorkspaceAuth> =>
+    withAuthorizationSurface("page", () => resolveWorkspaceAuth(workspaceId))
```
BhagyaAmarasinghe, line 49: [P1] Keep the page surface open for the whole page decision chain

This AsyncLocalStorage context ends when `resolveWorkspaceAuth()` returns; the caller continuation after `await getWorkspaceAuth()` runs outside it. Several pages then issue central checks, including `feedbackDirectoryAssignment.read` through `getAuthorizedWorkspaceFeedbackDirectories()` and `organization.manage` in the taxonomy page. With `page:user` selected, those decisions therefore remain legacy-only and produce no shadow evidence, while enforcement is silently partial. Establish a boundary that spans the complete page/layout authorization chain, and add an integration test that performs a central check after the workspace helper returns.

apps/web/lib/authzed/rollout-contract.ts
```diff
@@ -7,6 +7,7 @@
  */
 export const AUTHZED_AUTHORIZATION_ROLLOUT_TARGETS = [
   "server_action:user",
+  "page:user",
```
BhagyaAmarasinghe, line 10: [P2] Document the new rollout target

This extends the accepted rollout configuration with `page:user`, but the canonical list under “Valid targets” in `authzed/RUNBOOK.md` still omits it. Update that runbook section so operators can discover and configure the page shadow/enforcement surface from the documented contract.

BhagyaAmarasinghe (changes requested):
Requesting changes for two rollout-contract gaps. The page surface currently ends with the workspace helper, leaving later page authorization decisions outside shadow and enforcement, and the operator runbook does not list the newly accepted target. Details inline.

apps/web/lib/authzed/rollout-contract.ts
xernobyl: Fixed in `dff3fd313`, and the gap was wider than the one target.

The "Valid targets" block was missing **three** entries against `AUTHZED_AUTHORIZATION_ROLLOUT_TARGETS`: `page:user` from this PR, and `feedback_gateway:user` / `feedback_gateway:apiKey`, which landed before it. Adding only mine would have shipped a knowingly-incomplete canonical list, so all three are in and the block now equals the contract exactly, in order.

Since it has now drifted silently three times, it is asserted rather than remembered — `apps/web/lib/authzed/rollout-runbook.test.ts` parses the block out of `RUNBOOK.md` and compares it to the constant. It also asserts the parse found something, so a renamed heading or fence fails loudly instead of comparing against an empty list and passing.

apps/web/modules/workspaces/lib/utils.ts
xernobyl: Confirmed, both halves, and the second one is worse than the first — but I would like your call on the fix before I land it, because it changes the module's context mechanism rather than a call site.

**The scoping is exactly as you describe.** `withAuthorizationSurface` runs the callback inside `AsyncLocalStorage.run`, so the store is gone once `resolveWorkspaceAuth()` resolves; the caller's continuation after `await getWorkspaceAuth()` is a different async context. Every later central check in that page body — `feedbackDirectoryAssignment.read` via `getAuthorizedWorkspaceFeedbackDirectories()`, `organization.manage` on the taxonomy page — sees no store.

**And enforcement really is silently partial, not just the evidence.** `coordinator.ts:235-237`:

```ts
const target = getAuthorizationRolloutTarget(actor.type);
if (!target) return legacyEvaluator.can(actor, action, resource);
```

`getAuthorizationRolloutTarget` returns `null` outside a surface, so those checks fall back to legacy *unconditionally* — including under enforcement, where the whole point is that SpiceDB decides. That is the sharper half of your finding and I had not stated it anywhere.

**Why I have not just pushed a fix.** There is only one mechanism that actually spans the chain, and it has a real caveat:

- **`enterWith`** — rejected. It mutates the current async resource's store and leaks unpredictably into sibling work; leaking an authorization surface across concurrent requests is not a trade I will make.
- **Wrapping each page body** — the same bug returns the first time someone adds a page and forgets.
- **A React `cache()`-backed store** — this is the one that works. `cache()` is request-scoped for the whole RSC render, so a layout and its page share it. `withAuthorizationSurface("page", …)` would publish its context into a `reactCache`d holder, and the context lookups would read the ALS store first and fall back to that holder. Outside a render the holder is a fresh `{context: null}` per call, so scripts, route handlers and server actions behave exactly as they do today. It also collapses the "one navigation records two observations" wrinkle currently documented on `TAuthorizationSurface`, since the layout and page choke points would share one context.

The caveat is testability. That fix only works inside a React render, and `cache()` does not dedupe outside one — so a plain Vitest test cannot cover it, and mocking `cache` would only test the mock. Real coverage means either a render harness or an E2E that asserts a post-helper check is attributed to `page:user` in the shadow comparisons. Given this is your epic and the change is to the core context mechanism, I would rather have your agreement on the approach than land an untested rework of it.

So: happy to implement the `cache()`-backed holder immediately, with the E2E as its proof, if you agree that is the right shape. If you would rather this PR stay narrow, the alternative is to land it as-is with the limitation stated explicitly on `TAuthorizationSurface` — it is currently understated there — and take the boundary work as its own ticket before `page:user` goes anywhere near enforcement. I do not think it should reach enforcement in the current shape either way.

xernobyl: Revising my own recommendation from above — I think the boundary rework is the wrong thing to do in this PR, and I was aiming at the wrong target.

**The dangerous behaviour is not specific to this surface.** `coordinator.ts` falls back to legacy for any check that resolves no surface, on any surface. For `api_v3`/`server_action` the wrapper spans the whole handler so nothing escapes, which is why it has never bitten — but the fallback itself is general, and it is silent everywhere. Reworking the `page` context mechanism fixes one instance and leaves the mechanism just as quiet.

It is also not live: `page:user` is in no enforcement list, so today's only real cost is missing shadow evidence on a few downstream checks. Evidence from the choke points every product route funnels through is still useful evidence.

So instead of the `cache()` rework, pushed in `<pending>`:

1. **The unscoped fallback is now counted** — `formbricks_authzed_authorization_unscoped_checks_total`, tagged `rollout_enabled`. That is the metric that makes your finding visible rather than theoretical: a surface whose boundary does not cover every check it should looks *identical* to a clean cutover, because nothing compares and so nothing ever mismatches. Cheap, unit-testable, and it covers every surface rather than just this one. Tests assert it fires for an unresolvable surface, and specifically that it does **not** fire when a surface resolved but the organization is outside the cohort — otherwise the gap would vanish into the noise of normal operation.
2. **`TAuthorizationSurface` now states the limit plainly.** It previously only mentioned the split histogram count, which undersold it: it now says the surface does not span the whole page, names the two checks you identified, and records that this means no shadow evidence *and* an enforcement bypass.
3. **The boundary rework wants its own ticket**, gated so `page:user` cannot reach an enforcement list before it lands. That is where the request-scoped store and its E2E belong, with room to do it properly — rather than on the critical path of this PR, untested.

I have not filed the ticket yet: the Linear connector is not authorised in my session, so @xernobyl needs to open it (or I will once it is). The docstring deliberately carries no ticket ID rather than a fabricated one.

If you would rather the rework land here regardless, say so and I will do it with an E2E as its proof — but I do not think an untestable change to the context mechanism belongs in a PR that is otherwise about adding the surface.

xernobyl: Pushed as `6895ec888` — replacing the `<pending>` in my note above.

Verified rather than assumed, since a metric that is never recorded is exactly the kind of thing that ships green:

- 39 tests pass across `coordinator`, `metrics`, `context` and the new runbook assertion.
- **Mutation-tested both directions.** Dropping the `recordUnscopedAuthorizationCheck` call fails the two positive cases; making it unconditional (removing the `!target` guard) fails the rollout-disabled case. So neither the call nor the guard can be removed silently.
- `prettier --check` clean on all five touched files.

One note on the local run for anyone reproducing it: three unrelated suites fail in a fresh worktree — `lib/authzed/backfill-boundary.test.ts` and `backfill-cli.test.ts` on a broken `apps/web/.env` symlink, and the setup-invite authorization test on missing `CUBEJS_*` env vars. I confirmed they fail identically with my changes stashed, so they are worktree environment gaps rather than anything in this diff; they are green on this PR's CI.

BhagyaAmarasinghe (approved):

PR #8873, opened 2026-08-14
refactor(authorization): route the organization page gates through can()

xernobyl:
> **#8872 (ENG-2388) has merged**, so this now targets `epic/authzed` directly and no longer depends on an open PR. It was rebased onto the post-merge epic head — see *Stack maintenance* at the bottom.

## What & why

ENG-2388 made server-rendered route checks comparable, but it could only cover gates that already called `can()`. Sweeping for the rest turned up gates that decide access **without going through `can()` at all** — they read a membership row and test a role flag. A surface there would wrap nothing; they need *routing*.

That was the whole organization settings area: 13 pages behind `getOrganizationAuth`. Legacy stays authoritative so nothing was exploitable, but these decisions produced no parity evidence, and the cutover depends on that evidence.

| Gate | Was | Now |
| --- | --- | --- |
| `getOrganizationAuth` tenancy check (13 pages) | `if (!membership) throw` | `organization.read` |
| `redirectBillingRoleFromRestrictedOrgSettings` (5 pages) | `isBilling` | `organization.read_access` |
| Enterprise settings page | `isMember` | `organization.manage_billing` |
| Feedback directories page | `isOwner \|\| isManager` | `organization.manage` |
| API keys page | `role === "owner" \|\| "manager"` | `organization.manage_api_keys` |
| Onboarding setup-invite (page **and** action) | `SETUP_INVITE_ROLES = ["owner"]` | `organization.write` |

**Every mapping is set-equivalent** — no principal changes side. Each gate sits after `getOrganizationAuth`, which throws for a caller with no membership, so the reachable set is always the four membership roles and they partition cleanly.

**No schema change.** Every gate landed on an existing permission whose expansion already matched.

### The ticket's stated objection was wrong

ENG-2409 was filed saying this would be *additive* — an extra check per render, against the epic's own no-amplification criterion. It is not. The legacy evaluator's `canOrganization` resolves organization actions through `getMembershipByUserIdOrganizationId`, the same `reactCache`-memoized function `getOrganizationAuth` already calls. **Net added DB queries: zero.** Under enforcement `can()` takes the scope-resolver path and issues no membership query at all. I have corrected the ticket.

### Two mappings where the obvious choice is wrong

**`manage_billing`, not `manage`, for the enterprise page.** This is the one that would have shipped a real regression. `getOrganizationBillingPath(orgId, false)` resolves to `settings/enterprise` — so on self-hosted that page *is* where the billing role gets redirected **to**. Gating it on `organization.manage` (owner+manager) would 404 that role on its own landing page. `manage_billing` is owner+manager+billing, exactly `!isMember`.

Worth noting how nearly this slipped through: `playwright/billing-role-access.spec.ts` asserts on URL, and a 404 keeps the URL, so the existing E2E would have stayed green.

**`read_access`, not `read`, for the billing redirect.** `read` admits billing — the exact role being excluded. `read_access` is the only permission whose expansion is "holds a product-eligible membership role". The name is a stretch (it is documented around access-control resources) and the call site says so. `product_member` has the right expansion but is deliberately absent from the permission map: it exists to intersect into `team#member`, and giving it a second job would mean a future edit silently changed team-derived workspace access.

### The setup-invite gate (both halves)

Filed as instance 2 of the ticket and originally deferred, because the file did not exist on the epic until #8863 landed. It does now, so it is in.

That path decided access from a `SETUP_INVITE_ROLES = ["owner"]` list **two different ways**: the page read a membership row and tested the list; the action passed the list to the deprecated `checkAuthorizationUpdated` adapter. Both now ask `organization.write`.

The action's decision is unchanged *by construction* — the adapter's `ORGANIZATION_ACTION_BY_ROLE_SET` already maps `["owner"]` onto `organization.write`, so it was resolving to this exact question already. What changes is the page: its decision now reaches the coordinator. It also retires a `checkAuthorizationUpdated` caller, which the module README asks for.

**Both were converted together deliberately.** They agreed only because of that adapter mapping, so converting one and leaving the other would have made the file's own "cannot drift" comment false.

The owner-only restriction is load-bearing and preserved exactly: `inviteUser` always persists an *owner* invite, so admitting a manager would let them mint an owner without an existing owner's approval (ENG-2169). `organization.write` is `permission write: user = owner` — the vocabulary's owner-only capability.

This also restores the `page` surface that the #8863 merge dropped from the invite page, now that there is a `can()` beneath it to make comparable.

### Deliberate non-changes

- The enterprise page's Cloud-only `isBilling → redirect` is **not** routed. It decides 302-vs-404 for a principal the next line refuses anyway; expressing it centrally would need a "billing role only" capability, and inventing one would encode a role name as a permission.
- `getOrganizationAuth` still gates on membership only, never on a product permission. Unlike `getWorkspaceAuth` it must not exclude billing — the billing settings page is that role's own page. Recorded in the README so the symmetry isn't "fixed" later.
- The role flags stay on the return value for UI props (`isReadOnly`, `isDeleteDisabled`, `membershipRole`). Retained by design; only gates moved.

### Two small behavior changes, both deliberate

1. **API keys page throws `AuthorizationError` instead of a bare `Error`.** The bare `Error` is not in `EXPECTED_ERROR_NAMES`, so every unauthorized load of that page was being reported to Sentry as an unexpected exception. `app/error.tsx` renders both identically — nothing user-visible moves, the noise stops.
2. **That gate now runs before the workspace/locale fetch it protects**, rather than after. Harmless before (nothing was rendered), but the wrong order to leave in place.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2409

## How this was tested

**The schema suite had a hole this change would have depended on.** `organization#read_access` appeared three times in `schema-validation.yaml`, all in `assertTrue`. There was **no `assertFalse` for it anywhere** — so the billing-exclusion property that five settings pages now rest on was asserted nowhere, and adding `billing` to `read_access` in the schema left all 149 assertions green. Same shape as the ENG-2340 gap. Closed here: 149 → **156 assertions**, 2 → 4 expected relations, and `read_access`/`manage_api_keys` are now enumerated in the expected-relations block so a widening shows as a diff.

- `pnpm authzed:validate` → **36 relationships, 156 assertions, 4 expected relations** ✅. Verified by mutation: adding `billing` to `read_access` now **fails** the suite.
- **New** `modules/settings/lib/redirect-billing-role.test.ts` → 5 passing ✅. This guard fronts five pages and had no test at all. `can()` is left **real** — the rollout is off in the unit env, so the legacy evaluator answers from the mocked membership service and the role is the only input. Verified by mutation: swapping `read_access` for `read`, which billing does hold, fails the billing case.
- `modules/organization/lib/utils.test.ts` → the four existing tests pass **unmodified**, which is the equivalence evidence; two added (the authorization arm of the throw, isolated from the row; and the actor/action/resource). Verified by mutation: dropping `hasOrganizationAccess` from the condition fails the first.
- `modules/setup/.../invite/lib/authorization.test.ts` → **16 passing with no assertion touched** ✅ — only its header comment changed, which named the adapter that no longer runs. That equivalence is the strongest evidence in this PR. Verified by mutation: widening the action to `organization.manage` fails exactly the two manager cases, which is the escalation ENG-2169 exists to prevent.
- **New** `lib/authorization/org-page-gates.integration.test.ts` → **11 passing against a real database** ✅. Asserts the full four-role partition for all five permissions through the real coordinator and legacy evaluator, the no-membership case for each, and that a page's gates cost one check each.
- Full web unit suite → `6 failed | 8242 passed`. All 6 are the pre-existing `scripts/authzed-entrypoints.test.ts` failures from a Node 26 deprecation warning breaking a `stderr === ""` assertion — absent from this diff, and CI runs Node 24 where they are green.
- `pnpm turbo run typecheck --filter=@formbricks/web` → 10/10 ✅; eslint + prettier clean.

**Smoke-tested against the running app.** Started the dev server, seeded an owner and a plain member, and hit `/organizations/<id>/settings/api-keys` as each: the owner renders the page, the member is refused, and the server log shows exactly **one** `AuthorizationError` from `assertCan` across the two loads — so the gate fires for the member and not for the owner.

**Security review.** gitleaks clean over the branch range. No new data access, logging, or outbound calls. All seven new `can`/`assertCan` call sites take the actor from `session.user.id` — never from a parameter — and the resource is the organization the caller already resolved through `getOrganizationAuth`.

## Breaking changes

None. No schema change, no new env var, no migration, no default altered.

## QA / Test Plan

**How to test**

- [ ] As an **owner** → `/organizations/<id>/settings/{general,domain,api-keys,teams,feedback-directories}` → all render as before.
- [ ] As a **manager** → same five → all render as before.
- [ ] As a **plain member** → same five → `general`, `domain` and `teams` render (role only controls affordances there); `api-keys` shows the error page; `feedback-directories` shows its no-access message.
- [ ] As a **billing-role** user → same five → redirected to the billing home (Cloud) or the enterprise page (self-hosted). This is the widest gate; check all five.
- [ ] **Self-hosted only:** as a **billing-role** user → `/organizations/<id>/settings/enterprise` → the page **renders**. It is that role's landing target, so a 404 here is the regression to watch for.
- [ ] Self-hosted, as a **plain member** → `/organizations/<id>/settings/enterprise` → 404, unchanged.
- [ ] As a user who is **not a member of that organization** → any of the above → refused, unchanged.
- [ ] Sanity: creating and revoking an API key as an owner still works (the gate moved ahead of the data fetch).
- [ ] **Onboarding invite path.** Create a new organization → you land on the invite screen as its owner → it renders, and sending an invite still works.
- [ ] As a **manager/member/billing** user of an existing org, open that org's setup invite URL directly → **404**, unchanged. (Owner-only is intentional here — the invite always creates an *owner* invite, so a manager must not be able to send one.)

**Preconditions / test data**

- One organization with an owner, a manager, a plain member, and a billing-role member; a second organization the first org's users do not belong to.
- The enterprise-page case needs a **self-hosted** instance (`IS_FORMBRICKS_CLOUD` unset) — on Cloud that page 404s for everyone by design.
- No feature flag required; nothing changes by default. To exercise the new decisions under shadow comparison, add `page:user` to the authorization shadow targets.

**Risks & regressions**

- The billing-role redirect covers five pages from one helper — if it regresses, it regresses everywhere at once.
- The setup-invite path changed on **both** sides (page gate and action). The action's decision is unchanged by construction, but it is worth confirming an owner can still actually send an invite end to end, not just load the screen.
- The self-hosted enterprise page for the billing role is the specific case a wrong mapping would break, and URL-based checks cannot see it. Confirm the page content renders.
- Organization settings renders now report under the `page` metric surface. A dashboard filtering on `server_action` will look like it dropped — reporting change, not behavior.

- none.

## Stack maintenance (2026-08-16)

After repairing #8872 onto the current `epic/authzed` head, this PR's single ENG-2409 commit was rebased onto the repaired parent `e56d27dde`. The two overlaps were resolved semantically: the feedback-dataset page keeps the newly merged central authorization and adds the `page` shadow surface without serializing the license lookup; the setup-invite path keeps owner-only `organization.write` semantics and now emits a valid page comparison.

The repaired child head is `e60d524c2`. Its review diff remains only ENG-2409: 14 files, 478 additions, and 52 deletions. Exact-head validation: 21 focused unit tests, 11 organization-gate PostgreSQL integration tests, 4 page-surface/check-amplification PostgreSQL integration tests, 12/12 web typecheck tasks, changed-file ESLint and Prettier, and AuthZed validation with 36 relationships, 156 assertions, and 4 expected relations.

## Stack maintenance (2026-08-18) — rebased onto the merged epic

#8872 merged as a squash, which left this branch based on the *pre-squash* copies of that work and of Bhagya's feedback-dataset stack. Merging `epic/authzed` produced **33 conflicting files**, nearly all of them other people's already-merged code — resolving those by hand would have meant reconstructing a state that already exists in epic, with a real chance of silently reverting a merged guard.

So the six real ENG-2409 commits were replayed onto the current epic head instead. The result is **14 files, +478/−52** across 7 commits — the other 8 files the PR used to show were #8872's, now upstream.

- Two semantic conflicts, resolved the way this PR always said they would be. The feedback-dataset page keeps epic's central authorization and takes this PR's earlier, surface-wrapped gate — the license check still runs first, so an unlicensed organization still gets the upgrade prompt rather than a no-access message. The setup-invite path keeps owner-only `organization.write` and re-applies the `page` wrapper.
- The 162-file accidental Prisma-migration commit and its revert are gone. They did carry a genuine README reformat, so that was re-derived by running Prettier rather than dropped.
- Verified before force-pushing: the remote head was a commit this session had not seen (`e60d524c2`), and `--force-with-lease` refused the first push because of it. Diffed against it — **12 of the 14 files are byte-identical**, and the only two that differ are the review fix below plus epic's newer README content flowing in. Nothing was lost.

### Review fix — Bhagya's P2

The expected-relations block covered `read_access` and `manage_api_keys`, but this PR also turns `organization.read` (the `getOrganizationAuth` tenancy gate) and `organization.manage_billing` (the enterprise settings page) into page-gate permissions, and neither was enumerated. Both are now. An assertion proves a role *holds* a permission; only this block makes a **widening** show up as a diff.

Adding them surfaced a gap that was not raised: **`manage_billing` had no `assertFalse` anywhere** — both mentions were `assertTrue`. The enterprise page maps `isMember -> notFound` onto `manage_billing`, and that mapping is only correct because a plain member does not hold it, so adding `+ member` to it in the schema would have kept the suite green while opening the billing surface to every member. That is the same shape as the `read_access` gap this PR already closed, and as ENG-2340 before it. `assertFalse: organization:acme#manage_billing@user:mia` now pins it.

Suite is 83 `assertTrue` / 74 `assertFalse` with 6 expected-relations blocks; the YAML parses. `pnpm authzed:validate` needs the `zed` CLI, which is not on this machine, so that assertion set is CI-verified only.

### DIFF
--- a/apps/web/modules/ee/feedback-directory/page.tsx
+++ b/apps/web/modules/ee/feedback-directory/page.tsx
@@ -23,14 +23,22 @@ export const FeedbackDirectoriesPage = async (props: { params: Promise<{ organiz
   // ENG-2409: was a second `getAccessFlags(currentUserMembership.role)` on a role this page had
   // already been handed, then `!isOwner && !isManager`. `organization.manage` is the same set.
   // `membershipRole` below still comes from the row — that is a rendering prop, retained by design.
-  const canManageOrganization = await withAuthorizationSurface("page", () =>
-    can({ type: "user", id: session.user.id }, "organization.manage", {
-      type: "organization",
-      id: organization.id,
-    })
-  );
+  //
+  // Run alongside the license lookup rather than before it: the two are independent (one asks about
+  // the caller's role, the other about the organization's plan), and the flag test this replaced was
+  // synchronous, so awaiting them in series would have made every load of this page pay for both
+  // round trips end to end. The *branches* below keep their original order — an unlicensed
+  // organization must still see the upgrade prompt rather than a no-access message.
+  const [canManageOrganization, isFeedbackDirectoriesAllowed] = await Promise.all([
+    withAuthorizationSurface("page", () =>
+      can({ type: "user", id: session.user.id }, "organization.manage", {
+        type: "organization",
+        id: organization.id,
+      })
+    ),
+    getIsFeedbackDirectoriesEnabled(organization.id),
+  ]);
 
-  const isFeedbackDirectoriesAllowed = await getIsFeedbackDirectoriesEnabled(organization.id);
   const pageTitle = t("workspace.settings.feedback_directories.title");
 
   if (!isFeedbackDirectoriesAllowed) {
--- a/apps/web/modules/ee/feedback-directory/page.tsx
+++ b/apps/web/modules/ee/feedback-directory/page.tsx
@@ -1,5 +1,6 @@
 import { SettingsCard } from "@/app/(app)/workspaces/[workspaceId]/settings/components/SettingsCard";
 import { can } from "@/lib/authorization";
+import { withAuthorizationSurface } from "@/lib/authorization/context";
 import { ENTERPRISE_LICENSE_REQUEST_FORM_URL, IS_FORMBRICKS_CLOUD } from "@/lib/constants";
 import { getTranslate } from "@/lingodotdev/server";
 import { FeedbackDirectoryView } from "@/modules/ee/feedback-directory/components/feedback-directory-view";
@@ -18,6 +19,16 @@ export const FeedbackDirectoriesPage = async (props: { params: Promise<{ organiz
   await redirectBillingRoleFromRestrictedOrgSettings(params.organizationId);
 
   const { currentUserMembership, organization, session } = await getOrganizationAuth(params.organizationId);
+
+  // ENG-2409: was a second `getAccessFlags(currentUserMembership.role)` on a role this page had
+  // already been handed, then `!isOwner && !isManager`. `organization.manage` is the same set.
+  // `membershipRole` below still comes from the row — that is a rendering prop, retained by design.
+  const canManageOrganization = await withAuthorizationSurface("page", () =>
+    can({ type: "user", id: session.user.id }, "organization.manage", {
+      type: "organization",
+      id: organization.id,
+    })
+  );
 
   const isFeedbackDirectoriesAllowed = await getIsFeedbackDirectoriesEnabled(organization.id);
   const pageTitle = t("workspace.settings.feedback_directories.title");
@@ -53,12 +64,7 @@ export const FeedbackDirectoriesPage = async (props: { params: Promise<{ organiz
     );
   }
 
-  const canManageFeedbackDirectories = await can(
-    { type: "user", id: session.user.id },
-    "organization.manage",
-    { type: "organization", id: organization.id }
-  );
-  if (!canManageFeedbackDirectories) {
+  if (!canManageOrganization) {
     return (
       <PageContentWrapper>
         <PageHeader pageTitle={pageTitle} />
--- a/apps/web/modules/settings/lib/redirect-billing-role.ts
+++ b/apps/web/modules/settings/lib/redirect-billing-role.ts
@@ -26,7 +26,7 @@ import { getOrganizationBillingPath } from "@/modules/settings/lib/routes";
 export const redirectBillingRoleFromRestrictedOrgSettings = async (organizationId: string): Promise<void> => {
   const { session } = await getOrganizationAuth(organizationId);
 
-  const hasOrganizationAccess = await withAuthorizationSurface("page", () =>
+  const hasOrganizationReadAccess = await withAuthorizationSurface("page", () =>
     can({ type: "user", id: session.user.id }, "organization.read_access", {
       type: "organization",
       id: organizationId,
@@ -36,7 +36,7 @@ export const redirectBillingRoleFromRestrictedOrgSettings = async (organizationI
   // Deliberately outside the surface callback: redirect() throws a Next control-flow error, and
   // keeping it out here means the drain scheduled by withAuthorizationSurface never has to survive
   // a throw from inside its own callback.
-  if (!hasOrganizationAccess) {
+  if (!hasOrganizationReadAccess) {
     redirect(getOrganizationBillingPath(organizationId, IS_FORMBRICKS_CLOUD));
   }
 };
--- a/apps/web/modules/settings/lib/redirect-billing-role.ts
+++ b/apps/web/modules/settings/lib/redirect-billing-role.ts
@@ -1,4 +1,6 @@
 import { redirect } from "next/navigation";
+import { can } from "@/lib/authorization";
+import { withAuthorizationSurface } from "@/lib/authorization/context";
 import { IS_FORMBRICKS_CLOUD } from "@/lib/constants";
 import { getOrganizationAuth } from "@/modules/organization/lib/utils";
 import { getOrganizationBillingPath } from "@/modules/settings/lib/routes";
@@ -6,10 +8,35 @@ import { getOrganizationBillingPath } from "@/modules/settings/lib/routes";
 // Org-scoped equivalent of redirectBillingRoleFromRestrictedSettings (the workspace-scoped guard).
 // Bounces a billing-role member away from a restricted org settings page to their billing/enterprise
 // home. getOrganizationAuth is React-cached, so calling it here in addition to the page is free.
+//
+// ENG-2409: was `isBilling -> redirect`, a role-name test that never reached the central interface,
+// so the widest gate in the organization settings area (five pages) produced no parity evidence.
+//
+// `organization.read_access` is the same set. It expands to owner + manager + member — every
+// membership role except billing — and getOrganizationAuth has already thrown for a caller with no
+// membership at all, so `!read_access` and `isBilling` select exactly the same principals here.
+//
+// The name is admittedly a stretch: read_access is documented around access-control resources, and
+// four of the five pages it guards are not that. It is used anyway because it is the only permission
+// whose expansion is "holds a product-eligible membership role", which is what this guard means, and
+// because ORGANIZATION_ACTION_BY_ROLE_SET already maps that role set to it. The alternative with the
+// right expansion, `product_member`, is deliberately absent from the permission map — it exists only
+// to intersect into `team#member`, and giving it a second job would mean a future edit to it
+// silently changed team-derived workspace access.
 export const redirectBillingRoleFromRestrictedOrgSettings = async (organizationId: string): Promise<void> => {
-  const { isBilling } = await getOrganizationAuth(organizationId);
+  const { session } = await getOrganizationAuth(organizationId);
 
-  if (isBilling) {
+  const hasOrganizationAccess = await withAuthorizationSurface("page", () =>
+    can({ type: "user", id: session.user.id }, "organization.read_access", {
+      type: "organization",
+      id: organizationId,
+    })
+  );
+
+  // Deliberately outside the surface callback: redirect() throws a Next control-flow error, and
+  // keeping it out here means the drain scheduled by withAuthorizationSurface never has to survive
+  // a throw from inside its own callback.
+  if (!hasOrganizationAccess) {
     redirect(getOrganizationBillingPath(organizationId, IS_FORMBRICKS_CLOUD));
   }
 };
--- a/authzed/schema-validation.yaml
+++ b/authzed/schema-validation.yaml
@@ -216,6 +216,12 @@ assertions:
     - organization:acme#manage@user:mia
     - organization:acme#manage_access@user:mia
     - organization:acme#manage_api_keys@user:mia
+    # ENG-2409: the enterprise settings page maps `isMember -> notFound` onto `manage_billing`, and
+    # that mapping is only correct because a plain member does not hold it. `manage_billing` had no
+    # assertFalse anywhere — both mentions were assertTrue — so adding `+ member` to it in the schema
+    # would have kept this suite green while opening the billing surface to every member. Same shape
+    # as the read_access gap closed above.
+    - organization:acme#manage_billing@user:mia
     - workspace:main#read@user:mia
     - survey:onboarding#read@user:mia
     - dashboard:insights#read@user:mia
@@ -346,6 +352,24 @@ validation:
     - "[user:tom] is <organization:acme#member>"
     - "[user:uma] is <organization:acme#member>"
   organization:acme#manage_api_keys:
+    - "[user:mark] is <organization:acme#manager>"
+    - "[user:olivia] is <organization:acme#owner>"
+  # ENG-2409: `read` backs the getOrganizationAuth tenancy gate and `manage_billing` backs the
+  # enterprise settings page, so both are page-gate permissions now and both are enumerated for the
+  # same reason as the two above — an assertion proves a role holds it, only this block makes a
+  # *widening* show up as a diff.
+  organization:acme#read:
+    - "[api_key:org_reader_key] is <organization:acme#api_key_reader>"
+    - "[api_key:org_writer_key] is <organization:acme#api_key_writer>"
+    - "[user:bella] is <organization:acme#billing>"
+    - "[user:mark] is <organization:acme#manager>"
+    - "[user:mia] is <organization:acme#member>"
+    - "[user:olivia] is <organization:acme#owner>"
+    - "[user:tina] is <organization:acme#member>"
+    - "[user:tom] is <organization:acme#member>"
+    - "[user:uma] is <organization:acme#member>"
+  organization:acme#manage_billing:
+    - "[user:bella] is <organization:acme#billing>"
     - "[user:mark] is <organization:acme#manager>"
     - "[user:olivia] is <organization:acme#owner>"
   response:r1#export:
--- a/authzed/schema-validation.yaml
+++ b/authzed/schema-validation.yaml
@@ -93,7 +93,13 @@ assertions:
     - organization:acme#manage_access@user:olivia
     - organization:acme#manage_access@user:mark
     - organization:acme#manage_api_keys@user:mark
+    - organization:acme#manage_api_keys@user:olivia
     - organization:acme#manage_billing@user:mark
+    # ENG-2409: read_access is the "holds a product-eligible membership role" permission, and it
+    # now gates five organization settings pages. Its owner/manager arms were never asserted —
+    # only the member arm below was — so the permission could have narrowed without a red suite.
+    - organization:acme#read_access@user:olivia
+    - organization:acme#read_access@user:mark
     - workspace:main#manage@user:olivia
     - workspace:main#manage@user:mark
     - workspace:other#read@user:olivia
@@ -206,8 +212,10 @@ assertions:
     # Scenario 1 — managers must not update/delete the organization itself;
     # a teamless member has organization visibility but zero product access.
     - organization:acme#write@user:mark
+    - organization:acme#write@user:mia
     - organization:acme#manage@user:mia
     - organization:acme#manage_access@user:mia
+    - organization:acme#manage_api_keys@user:mia
     - workspace:main#read@user:mia
     - survey:onboarding#read@user:mia
     - dashboard:insights#read@user:mia
@@ -231,6 +239,13 @@ assertions:
     - organization:acme#write@user:bella
     - organization:acme#manage_access@user:bella
     - organization:acme#manage_api_keys@user:bella
+    # ENG-2409: the load-bearing one. `read_access` is the only permission whose expansion is
+    # "holds a product-eligible membership role", which is what makes it the central expression of
+    # "not the billing role" — and it is what five organization settings pages now gate on. Until
+    # this line existed the suite had NO assertFalse for read_access at all, so adding `billing` to
+    # it in the schema kept every assertion green. Exactly the shape of the ENG-2340 gap: a
+    # guarantee written in prose and asserted nowhere.
+    - organization:acme#read_access@user:bella
     - workspace:main#read@user:bella
     - survey:onboarding#read@user:bella
     - survey:onboarding#response_read@user:bella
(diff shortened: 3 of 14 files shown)


apps/web/modules/ee/feedback-directory/page.tsx
```diff
@@ -17,9 +18,17 @@ export const FeedbackDirectoriesPage = async (props: { params: Promise<{ organiz
...
   await redirectBillingRoleFromRestrictedOrgSettings(params.organizationId);
 
-  const { currentUserMembership, organization } = await getOrganizationAuth(params.organizationId);
+  const { currentUserMembership, organization, session } = await getOrganizationAuth(params.organizationId);
 
-  const { isOwner, isManager } = getAccessFlags(currentUserMembership.role);
+  // ENG-2409: was a second `getAccessFlags(currentUserMembership.role)` on a role this page had
+  // already been handed, then `!isOwner && !isManager`. `organization.manage` is the same set.
+  // `membershipRole` below still comes from the row — that is a rendering prop, retained by design.
+  const canManageOrganization = await withAuthorizationSurface("page", () =>
```
mattinannt: Efficiency regression: before this change, `getAccessFlags(currentUserMembership.role)` was synchronous, so the only awaited call on this page was `getIsFeedbackDirectoriesEnabled(organization.id)`. Now `canManageOrganization` is a real async `can()` round trip, and it's awaited sequentially before the independent license check starts — even though the two checks don't depend on each other (one depends only on membership role, the other only on the org's plan), and the license check will short-circuit and discard `canManageOrganization` anyway when the feature is disabled.

Consider running them concurrently, e.g. with `Promise.all`, so every page load doesn't pay for both round trips in series.

---

apps/web/modules/settings/lib/redirect-billing-role.ts
```diff
@@ -1,15 +1,42 @@
...
+// because ORGANIZATION_ACTION_BY_ROLE_SET already maps that role set to it. The alternative with the
+// right expansion, `product_member`, is deliberately absent from the permission map — it exists only
+// to intersect into `team#member`, and giving it a second job would mean a future edit to it
+// silently changed team-derived workspace access.
 export const redirectBillingRoleFromRestrictedOrgSettings = async (organizationId: string): Promise<void> => {
-  const { isBilling } = await getOrganizationAuth(organizationId);
+  const { session } = await getOrganizationAuth(organizationId);
 
-  if (isBilling) {
+  const hasOrganizationAccess = await withAuthorizationSurface("page", () =>
```
mattinannt: Minor naming nit: `hasOrganizationAccess` here means "holds `organization.read_access`", but the same identifier is used in `apps/web/modules/organization/lib/utils.ts` for a different permission (`organization.read`). Since both are plain booleans with no type distinction, a future edit that copies logic between these two call sites by pattern-matching on the variable name (rather than re-reading the permission string) could wire the wrong permission into a new page gate. Consider a more specific name like `hasBillingReadAccess` to avoid the collision.

---

mattinannt (commented):
Reviewed the migration of the organization page gates to `can()`. The permission mappings for `organization.read`, `read_access`, `manage`, `manage_billing`, and `manage_api_keys` all check out exactly against `canOrganization` in `legacy-evaluator.ts`, and the new tests correctly exercise the four-role partition. No security or correctness regressions found.

Two minor findings left as inline comments:
1. An efficiency regression in `feedback-directory/page.tsx` — an independent async permission check is now awaited sequentially before an independent license check that would discard its result anyway.
2. A naming nit in `redirect-billing-role.ts` — a locally-scoped `hasOrganizationAccess` variable shares a name with a different permission check elsewhere in the codebase.

---

xernobyl:
Both addressed in `b7be9502d`. Thanks — the second one was sharper than it looked.

**1. Sequential await on the feedback-directories page.** Fair, and it was a regression I introduced: the flag test I replaced was synchronous, so before this change the page had exactly one awaited call. Now under `Promise.all`.

I kept the check unconditional rather than making it lazy behind the license branch. Skipping it for unlicensed organizations would save a round trip, but it would also mean those page loads emit no shadow comparison — and comparison coverage is the entire point of the ticket. Concurrency gets the latency back without trading that away. The **branch** order is unchanged, so an unlicensed org still sees the upgrade prompt rather than the no-access message.

**2. The `hasOrganizationAccess` collision.** Worse than a nit — `hasOrganizationAccess` is an **exported helper** in `apps/web/lib/auth.ts:16`, so my locals were shadowing a real function name, not just each other. Exactly the setup your failure mode needs.

Renamed both to mirror the permission they hold, so the name cannot drift from the string:

| File | Permission | Was | Now |
|---|---|---|---|
| `redirect-billing-role.ts` | `organization.read_access` | `hasOrganizationAccess` | `hasOrganizationReadAccess` |
| `organization/lib/utils.ts` | `organization.read` | `hasOrganizationAccess` | `hasOrganizationRead` |

I renamed the second one too even though you only flagged the first — leaving it would have kept the ambiguity alive in the other direction.

Verified: typecheck 10/10, 274 tests green across `modules/organization`, `modules/settings` and `lib/authorization`, prettier/eslint clean.

xernobyl:
**Instance 2 is now in** (`a02874133`) — #8863 landing made it possible, so the ticket ships whole rather than leaving half of it as a follow-up.

The onboarding invite path decided access from a `SETUP_INVITE_ROLES = ["owner"]` list in two different ways: the page read a membership row and tested the list, the action passed the list to the deprecated `checkAuthorizationUpdated` adapter. Both now ask `organization.write`.

The action's decision is **unchanged by construction** — the adapter's `ORGANIZATION_ACTION_BY_ROLE_SET` already maps `["owner"] → organization.write`, so it was resolving to this exact question already. What changes is the page: its decision now reaches the coordinator. It also retires a `checkAuthorizationUpdated` caller, which the module README asks for.

I converted **both halves in the same commit on purpose.** They agreed only because of that adapter mapping — converting one and leaving the other would have made the file's own "the two cannot drift" comment false.

**The owner-only restriction is untouched**, and that matters: `inviteUser` always persists an *owner* invite, so admitting a manager would let them mint an owner without an existing owner's approval (ENG-2169). `organization.write` is `permission write: user = owner` — the vocabulary's owner-only capability.

Evidence: `authorization.test.ts` passes with **no assertion touched** — only its header comment, which named the adapter that no longer runs. Verified by mutation, widening to `organization.manage` fails exactly the two manager cases, which is the escalation the restriction exists to prevent.

This also restores the `page` surface I dropped from the invite page during the #8863 merge resolution, now that there is a `can()` beneath it.

typecheck 12/12, `format:check` clean, 290 tests green across setup/authorization/organization/settings, full suite 8662 passed with the 6 known pre-existing `authzed-entrypoints` failures.

xernobyl:
## ⚠️ Rebased onto #8868 — read this before merging

Base changed from `tiago/eng-2388-page-layout-shadow-comparison` to `bhagya/eng-2398-route-feedback-dataset-authorization` (#8868), and the conflict between the two branches is resolved here in `50af5f502` rather than left to surface at merge time.

### The one that would have bitten

Both branches add a surface to the same union in `lib/authorization/context.ts`:

```ts
export type TAuthorizationSurface =
  | "server_action"
  | "page"              // this PR (ENG-2388)
  | "api_v1" | "api_v2" | "api_v3" | "mcp"
  | "feedback_gateway"; // #8868
```

**Taking either side wholesale drops a surface while its rollout targets stay in `rollout-contract.ts`** — `page:user` or `feedback_gateway:user`/`:apiKey` would remain valid config strings for a surface the type no longer knows about. Resolved as a union; `rollout-contract.ts` auto-merged with all three target sets intact, and I verified that.

The other two: both surface tests kept in `context.test.ts` (neither subsumes the other — `page` deliberately has no apiKey target, `feedback_gateway` has both), and `feedback-directory/page.tsx` where **we independently made the same change** — both routed `isOwner || isManager` to `can(organization.manage)`. Kept this branch's version as a strict superset: same decision, plus the page surface that makes it comparable, plus the `Promise.all` so the check and the license lookup do not run in series.

### What still needs doing after #8868 merges

1. **Retarget this PR back to `epic/authzed`.** GitHub does that automatically when the base branch merges — confirm it landed on `epic/authzed` and not somewhere else.
2. **Re-check the surface union survived the epic merge.** The failure mode is silent: a dropped member is a type error only if something references it, and `rollout-contract.ts` would still accept the orphaned target strings. Quickest check:
   ```bash
   grep -A8 'export type TAuthorizationSurface' apps/web/lib/authorization/context.ts
   ```
   Expect **seven** members including both `page` and `feedback_gateway`.
3. **Re-run `pnpm authzed:validate`.** Currently **156 assertions** on the merged tree (143 from this branch plus #8868's feedback-dataset ones). My merge commit message says "still at 143" — that was the pre-merge number and is wrong; 156 is correct.
4. **This PR's diff currently includes #8872's changes**, because #8868's branch does not contain ENG-2388. That is expected while the base is Bhagya's branch, and resolves itself once both land on the epic. Review #8872 on its own PR rather than here.

### Verified on the merged tree

typecheck 12/12 · 363 tests green across `lib/authorization`, `feedback-directory`, `organization`, `settings`, `setup` · `authzed:validate` 36 relationships / 156 assertions · `format:check` clean.

apps/web/modules/ee/feedback-directory/page.tsx
xernobyl: Fixed in `b7be9502d` — replying inline as well, since I answered this in a top-level comment at the time and the thread still reads as open.

Under `Promise.all` now. You were right that it was a regression I introduced: the flag test this replaced was synchronous, so before the change the page had exactly one awaited call.

I kept the check unconditional rather than lazily skipping it behind the license branch. Skipping would save a round trip for unlicensed organizations, but those page loads would then emit no shadow comparison — which is the entire point of the ticket. Concurrency recovers the latency without trading that away, and the branch order is unchanged so an unlicensed org still sees the upgrade prompt.

apps/web/modules/settings/lib/redirect-billing-role.ts
xernobyl: Fixed in `b7be9502d` — replying inline for the record, since this was answered in a top-level comment at the time.

Worse than a nit, as it turned out: `hasOrganizationAccess` is an **exported helper** in `apps/web/lib/auth.ts:16`, so my locals were shadowing a real function name, not merely each other — exactly the setup your failure mode needs.

Renamed both to mirror the permission they hold, so the name cannot drift from the string:

| File | Permission | Now |
|---|---|---|
| `redirect-billing-role.ts` | `organization.read_access` | `hasOrganizationReadAccess` |
| `organization/lib/utils.ts` | `organization.read` | `hasOrganizationRead` |

I renamed the second even though you only flagged the first — leaving it would have kept the ambiguity alive in the other direction.

authzed/schema-validation.yaml
```diff
@@ -314,6 +333,21 @@ validation:
     - "[user:tina] is <organization:acme#member>/<team:product#contributor>"
     - "[user:tom] is <organization:acme#member>/<team:product#admin>"
     - "[user:uma] is <organization:acme#member>/<team:analysts#contributor>"
+  # ENG-2409: both of these became organization page gates, so enumerate their subjects here as
+  # well. The assertions above prove the roles that must and must not hold them; this block is what
```
BhagyaAmarasinghe, line 343: [P2] Lock every new page-gate permission

This expected-relations expansion covers `read_access` and `manage_api_keys`, but the PR also makes `organization.read` and `organization.manage_billing` page-security dependencies. The validation suite still passes if owner/manager are removed from `read`, if owner is removed from `manage_billing`, or if member is added to `manage_billing`. Add exact expected relations or complete allow/deny assertions for both permissions so a SpiceDB narrowing or widening cannot pass validation and change page access at cutover.

BhagyaAmarasinghe (changes requested):
Requesting changes for one schema-validation gap. The current permission definitions are correct, but the added parity suite does not lock organization.read and organization.manage_billing, both now used as page gates. Mutation testing confirmed that narrowing or widening these permissions still passes all 156 assertions and 4 expected relations. Details inline.

authzed/schema-validation.yaml
xernobyl: Fixed in `09f58ea1e`, and enumerating them turned up a second gap in the same block.

**Your point, applied.** `organization.read` (the `getOrganizationAuth` tenancy gate) and `organization.manage_billing` (the enterprise settings page) both became page-gate permissions in this PR and neither was in `validation:`. Both are enumerated now — `read` across all nine subjects (four membership roles plus the two org-scoped API keys), `manage_billing` across owner/manager/billing. And you are right about *why* it matters: the assertions prove a role **holds** a permission, but only the expected-relations block makes a **widening** show up as a diff rather than a still-green suite.

**The one hiding behind it.** While adding `manage_billing` I checked its assertion coverage: it had **no `assertFalse` anywhere** — both mentions in the file were `assertTrue`. That matters more than the enumeration. This PR maps the enterprise page's `isMember -> notFound()` onto `manage_billing`, and that mapping is only sound because a plain member does not hold it. Adding `+ member` to `manage_billing` in the schema would have kept every assertion green while opening the billing surface to every member of the organization. `assertFalse: organization:acme#manage_billing@user:mia` now pins it.

That is the third instance of this shape on the epic — a guarantee written in prose and asserted nowhere — after ENG-2340 and the `read_access` gap this PR already closed.

**Also on this push: the branch is rebased onto the post-merge epic.** #8872 squash-merged, which left this branch sitting on the pre-squash copies of it and of your feedback-dataset stack; merging `epic/authzed` produced 33 conflicting files, nearly all already-merged code that would have had to be reconstructed by hand. The six real ENG-2409 commits were replayed onto the current epic head instead, so the PR is 14 files rather than 22. The two semantic overlaps resolved as the body always said they would: the feedback-dataset page keeps epic's central authorization and takes the earlier surface-wrapped gate with the license check still first, and the setup-invite path keeps owner-only `organization.write`.

Worth flagging about the force-push: the remote head was a commit my session had not seen, and `--force-with-lease` refused the first attempt because of it. I diffed against that commit before overwriting — 12 of the 14 files are byte-identical, and the only two that differ are this review fix plus epic's newer README content flowing in. Nothing was dropped.

One caveat stated plainly: `pnpm authzed:validate` needs the `zed` CLI, which is not on this machine, so the new assertions are CI-verified only. The YAML parse (83 assertTrue / 74 assertFalse, 6 expected-relations blocks), the 12-file identity check and the conflict resolutions I verified locally.

BhagyaAmarasinghe (approved):

PR #8875, opened 2026-08-15
fix(helm): let the AuthZed bootstrap use an existing admin role

xernobyl:
## What & why

The bundled AuthZed database bootstrap connected as a hardcoded `postgres` superuser:

```yaml
- name: PGUSER
  value: postgres
```

An existing PostgreSQL installed with `postgresql.auth.enablePostgresUser=false` has no such role, so the Job could not authenticate. The only escape was `authzed.bundledPostgresqlBootstrap.enabled=false` — which doesn't just skip a step, it leaves the `spicedb` role and database **uncreated**. That is exactly what forced the values override during the `authzed-sandbox` upgrade.

The connection identity is now configurable:

```yaml
authzed:
  bundledPostgresqlBootstrap:
    adminUsername: fbadmin
    adminDatabase: formbricks # maintenance database to attach to
    adminPasswordSecretName: existing-pg-admin
    adminPasswordKey: password
```

The role needs `CREATEROLE` and `CREATEDB` — **not** superuser.

### Defaults are unchanged

`adminUsername` defaults to `postgres`, `adminDatabase` to `postgres`, and the password falls back to the bundled admin Secret and key. Existing releases render byte-identically; the render test asserts this.

### The precondition is now scoped, not removed

`postgresql.auth.enablePostgresUser` is still required — but only when the admin actually *is* the bundled superuser. An operator who pointed `adminUsername` at their own role has no reason to also materialise a superuser the subchart was told not to create; requiring it is what pushed people to disable bootstrap wholesale.

When it does fire it now names all three ways out rather than asserting one:

> Set `authzed.bundledPostgresqlBootstrap.adminUsername` (with `adminPasswordSecretName` and `adminPasswordKey`) to an existing role with CREATEROLE and CREATEDB, or enable `postgresql.auth.enablePostgresUser`, or disable `authzed.bundledPostgresqlBootstrap.enabled`

Two new guards close gaps this opened: a custom `adminUsername` **without** `adminPasswordSecretName` is refused rather than silently falling back to the bundled superuser's password (it would fail to authenticate at runtime instead of at render), and the admin password key is validated against the effective value rather than only the subchart's.

### What deliberately did not change

Only the *connection* moved. The SQL already detects existing state — role and database are created only when absent (`WHERE NOT EXISTS`) — which is the ticket's "safely detect that the target role/database are already initialized".

**Correction after review:** I first described a re-run as a no-op. It isn't. `ALTER ROLE … PASSWORD` is unconditional, so every run reconciles the `spicedb` role's password to the chart Secret. Rotating that password outside Helm is silently undone by the next upgrade. Both documents now say so. `bundledPostgresqlBootstrap.enabled=false` is untouched and remains correct when both are provisioned by hand. External bootstrap is untouched.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2390

## How this was tested

**The render suite could not have caught the bug this PR was for.** `charts/formbricks/tests/authzed-operations.sh` covered NOTES.txt only when I started; I added render assertions, and they proved the manifest said the right thing while the SQL underneath did not work at all. That gap is now closed from both ends.

### Render assertions (`authzed-operations.sh`, run by Helm Chart Validation)

| Assertion | Why |
| --- | --- |
| Default renders `postgres`/`postgres` | The no-op guarantee for existing releases |
| `enablePostgresUser=false` alone still refuses, and the message names `adminUsername` | The regression, and that the way out is discoverable |
| Configuring an existing admin role renders `fbadmin` / `formbricks` / `existing-pg-admin` | The ENG-2390 scenario itself |
| Custom `adminUsername` without a Secret is refused | Prevents the silent-wrong-password path |
| Custom `adminUsername` without a key is refused | The key falls back to the subchart's name inside a foreign Secret |
| `adminUsername=postgres` + explicit Secret + `enablePostgresUser=false` renders | An existing privileged `postgres` role is a configured admin, not the bundled superuser |
| ...and the same, **without** a key, is refused | The guard keys off the credential source, not the username |
| `PGPASSWORD`, `SPICEDB_DATABASE_PASSWORD`, `ADMIN_DATABASE_URL` are supplied by `secretKeyRef` | Secret-safety, asserted structurally, in both modes |

### Runtime assertions (`authzed-bootstrap-runtime.sh`, new)

Render tests cannot execute SQL, and the SQL was the actual defect. This extracts the script from the **rendered Job** and runs it against real PostgreSQL — so it cannot drift from the template, and cannot pass against SQL the chart does not ship. It runs on **15 and 17**, either side of the 16 `CREATEROLE` change, and skips itself when Docker is unavailable.

| Case | Asserted |
| --- | --- |
| Fresh server, `CREATEROLE`+`CREATEDB` admin | Succeeds; `spicedb` database exists, is owned by `spicedb`, and that role can actually log in with the configured password |
| Re-run | Succeeds — the Job is a post-upgrade hook, so this is the common path |
| Bundled superuser | Succeeds, and is *not* granted the `spicedb` role |
| Pre-existing `spicedb` role created by someone else | Fails loudly on 16+ (documented limitation), still succeeds on 15 |

**Mutation-tested, not just green:**

- removing the `GRANT` → runtime suite red (`must be member of role` on 15, `must be able to SET ROLE` on 17)
- restoring the username-keyed `adminPasswordKey` guard → render suite red
- the structural secret assertion checked directly against four manifests (`reference` / `literal` / literal-as-last-entry / absent), because the in-situ mutation trips an earlier assertion first and would not have exercised it

`helm lint`, both suites, `bash -n`, and `prettier --check` are clean.

**Not verified by me:** the ticket's final criterion — "verified in `authzed-sandbox` without the temporary values override". That needs a real cluster upgrade against the existing database, which I can't do from here. @BhagyaAmarasinghe you hit the original failure, so you're best placed to confirm; the values to use are in the QA plan below.

## Breaking changes

None.

## QA / Test Plan

**How to test**

- [ ] Fresh install with defaults (`authzed.enabled=true`, bundled PostgreSQL) → bootstrap Job succeeds, `spicedb` role and database exist.
- [ ] Re-run `helm upgrade` with no changes → Job runs again and succeeds (idempotent, nothing recreated).
- [ ] **The regression:** upgrade an existing release whose PostgreSQL has no `postgres` role, with no overrides → render fails with a message naming the three remedies (previously: an obscure runtime auth failure).
- [ ] **The acceptance case:** same upgrade with `adminUsername`/`adminDatabase`/`adminPasswordSecretName`/`adminPasswordKey` pointing at an existing role that holds only `CREATEROLE` and `CREATEDB` → succeeds without `bundledPostgresqlBootstrap.enabled=false`, and the `spicedb` database is owned by `spicedb`. Worth doing on a server that is PostgreSQL 16 or newer, since that is where the privilege rules are strictest.
- [ ] With that same non-superuser admin, run the upgrade a second time → still succeeds.
- [ ] Pre-create a `spicedb` role as a *different* role, then run the bootstrap as the non-superuser admin on PostgreSQL 16+ → fails with a privilege error rather than appearing to succeed, and the README's remedy (`GRANT spicedb TO <admin> WITH ADMIN OPTION`) clears it.
- [ ] Set `adminUsername` but omit `adminPasswordSecretName`, then omit `adminPasswordKey` → each refused at render.
- [ ] `adminUsername=postgres` with an explicit Secret but no `adminPasswordKey` → refused at render (previously rendered a dangling `secretKeyRef`).
- [ ] Externally managed mode (`externalPostgresqlBootstrap.enabled=true`) → unchanged.
- [ ] `helm template ... --show-only templates/authzed-postgresql-bootstrap.yaml` → no credential values, only `secretKeyRef`.

**Preconditions / test data**

- A cluster with an existing Formbricks PostgreSQL release, ideally one installed with `postgresql.auth.enablePostgresUser=false` (that is the failing shape).
- A Secret holding the administrative role's password, and a role with `CREATEROLE` and `CREATEDB`.

**Risks & regressions**

- The blast radius is the bootstrap Job's connection identity. Defaults are unchanged, so an existing release should see no diff — worth confirming with `helm diff` before upgrading.
- The Job now grants its administrator membership of the `spicedb` role. That is not an escalation — the administrator holds `CREATEROLE` and could grant itself anything — but it is a visible change in `pg_auth_members`, and it persists after the Job. It is skipped entirely for superusers.
- A misconfigured `adminUsername` still fails at Job runtime rather than at render, and that is where it belongs: the chart cannot know which roles exist on the server. The render guards now cover both credential mistakes (missing Secret name, missing key) that would otherwise have failed at Pod creation instead of in Job logs.
- One case is documented rather than fixed: a `spicedb` role that already exists and was created by someone else cannot be adopted by a non-superuser administrator on PostgreSQL 16+. The Job fails loudly; the remedy is in the README.

- none. Existing values files keep working unchanged.

## Review fixes

Bhagya's three findings and CodeRabbit's open one, each reproduced before acting:

- **[P1] The documented non-superuser path did not work.** Confirmed against `postgres:17-alpine`, and it was broader than reported: PostgreSQL 15 fails too, with `must be member of role` instead of `must be able to SET ROLE`, because pre-16 a creator gets no membership at all. That also killed the obvious portable guard — `MEMBER` is true on 17 while the operation still fails, and `SET` is not a valid `pg_has_role` privilege type before 16. Fixed by granting the role explicitly, skipped for superusers, plus the runtime suite above.
- **[P2] The `adminPasswordKey` guard keyed off the username**, so an existing administrator still named `postgres` could supply a Secret without a key and render a dangling `secretKeyRef`. Now keyed off the credential source. This was also CodeRabbit's open finding on the same lines, which I had half-answered on a sibling thread and left open.
- **[P3] The secret-safety assertion could not fail.** It grepped for `PGPASSWORD: `, a shape the renderer never emits, so a literal `value:` under `- name: PGPASSWORD` passed the exact regression it existed to catch — verified. Replaced with a structural check over the env list, extended to the external mode, which had no such assertion at all.

### DIFF
--- a/charts/formbricks/README.md
+++ b/charts/formbricks/README.md
@@ -147,8 +147,11 @@ authzed:
 ```
 
 `adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
-would silently fall back to the bundled admin password and fail to authenticate. The Job creates the role and
-database only when they are absent, so re-running it against an already initialised database is a no-op, and
+would silently fall back to the bundled admin password and fail to authenticate.
+
+Re-running is safe but not inert: the role and the database are created only when absent, while the `spicedb`
+role's password is reconciled to the chart's Secret on **every** run. If you rotate that password outside Helm,
+update the Secret too, or the next upgrade will set it back.
 `authzed.bundledPostgresqlBootstrap.enabled=false` remains available for operators who provision both by hand.
 
 Then reference it from the release:
--- a/charts/formbricks/README.md
+++ b/charts/formbricks/README.md
@@ -131,6 +131,26 @@ The administrator URL must explicitly set `sslmode=require`, `sslmode=verify-ca`
 The bootstrap Job validates the TLS mode without logging the URL and passes the URL through unchanged, so other
 connection parameters and certificate settings are preserved.
 
+### Bootstrapping against an existing PostgreSQL without a `postgres` role
+
+The bundled bootstrap connects as the `postgres` superuser the subchart normally creates. An existing
+installation deployed with `postgresql.auth.enablePostgresUser=false` has no such role, so point the Job at a
+role that does hold `CREATEROLE` and `CREATEDB` instead:
+
+```yaml
+authzed:
+  bundledPostgresqlBootstrap:
+    adminUsername: fbadmin
+    adminDatabase: formbricks # the maintenance database to attach to
+    adminPasswordSecretName: existing-pg-admin
+    adminPasswordKey: password
+```
+
+`adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
+would silently fall back to the bundled admin password and fail to authenticate. The Job creates the role and
+database only when they are absent, so re-running it against an already initialised database is a no-op, and
+`authzed.bundledPostgresqlBootstrap.enabled=false` remains available for operators who provision both by hand.
+
 Then reference it from the release:
 
 ```yaml
--- a/charts/formbricks/README.md
+++ b/charts/formbricks/README.md
@@ -147,7 +147,17 @@ authzed:
 ```
 
 `adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
-would silently fall back to the bundled admin password and fail to authenticate.
+would silently fall back to the bundled admin password and fail to authenticate. `adminPasswordKey` is required
+whenever that Secret is configured explicitly, since it otherwise defaults to the bundled subchart's key name
+looked up inside your own Secret.
+
+`CREATEROLE` and `CREATEDB` cover the normal case, in which this administrator also creates the `spicedb` role.
+`CREATE DATABASE ... OWNER spicedb` additionally requires being able to `SET ROLE` to that owner, so the Job
+grants itself the `spicedb` role first; from PostgreSQL 16 that is only possible for a role it holds
+`ADMIN OPTION` on, which creating the role confers. A `spicedb` role that already exists and was created by
+someone else is therefore the one case the Job cannot adopt on PostgreSQL 16+ — grant it explicitly
+(`GRANT spicedb TO fbadmin WITH ADMIN OPTION`) or run the bootstrap once as a superuser. PostgreSQL 15 and
+older are unaffected.
 
 Re-running is safe but not inert: the role and the database are created only when absent, while the `spicedb`
 role's password is reconciled to the chart's Secret on **every** run. If you rotate that password outside Helm,
--- a/charts/formbricks/README.md
+++ b/charts/formbricks/README.md
@@ -146,10 +146,12 @@ authzed:
     adminPasswordKey: password
 ```
 
-`adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
-would silently fall back to the bundled admin password and fail to authenticate. `adminPasswordKey` is required
-whenever that Secret is configured explicitly, since it otherwise defaults to the bundled subchart's key name
-looked up inside your own Secret.
+`adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser, and
+`adminPasswordKey` whenever that Secret is configured explicitly. **Both are enforced when the chart renders**,
+so a missing one fails `helm template`/`upgrade` with a named value rather than producing a Job. That is the
+point of the guards: without them the first would silently fall back to the bundled admin password and the
+second would look up the subchart's key name inside your own Secret — neither of which surfaces until the Pod
+is created in the cluster.
 
 `CREATEROLE` and `CREATEDB` cover the normal case, in which this administrator also creates the `spicedb` role.
 `CREATE DATABASE ... OWNER spicedb` additionally requires being able to `SET ROLE` to that owner, so the Job
--- a/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
+++ b/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
@@ -8,11 +8,25 @@
 {{- if and $externalBootstrap (not .Values.authzed.externalPostgresqlBootstrap.adminSecretName) -}}
 {{- fail "authzed.externalPostgresqlBootstrap.adminSecretName is required when external bootstrap is enabled" -}}
 {{- end -}}
-{{- if and $bundledBootstrap (not .Values.postgresql.auth.enablePostgresUser) -}}
-{{- fail "postgresql.auth.enablePostgresUser must be true when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
+{{- $bundledAdminUsername := .Values.authzed.bundledPostgresqlBootstrap.adminUsername | default "postgres" -}}
+{{- $bundledAdminDatabase := .Values.authzed.bundledPostgresqlBootstrap.adminDatabase | default "postgres" -}}
+{{- $bundledAdminSecret := .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName | default $postgresSecretName -}}
+{{- $bundledAdminKey := .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey | default .Values.postgresql.auth.secretKeys.adminPasswordKey -}}
+{{- $usesBundledPostgresSuperuser := eq $bundledAdminUsername "postgres" -}}
+{{- /*
+  Only the bundled `postgres` superuser depends on enablePostgresUser. An operator who pointed
+  adminUsername at their own CREATEROLE/CREATEDB role has no reason to also materialise a superuser
+  the subchart was told not to create — requiring it is what forced existing installations to turn
+  bootstrap off entirely (ENG-2390).
+*/ -}}
+{{- if and $bundledBootstrap $usesBundledPostgresSuperuser (not .Values.postgresql.auth.enablePostgresUser) -}}
+{{- fail "Set authzed.bundledPostgresqlBootstrap.adminUsername (with adminPasswordSecretName and adminPasswordKey) to an existing role with CREATEROLE and CREATEDB, or enable postgresql.auth.enablePostgresUser, or disable authzed.bundledPostgresqlBootstrap.enabled" -}}
 {{- end -}}
-{{- if and $bundledBootstrap (not .Values.postgresql.auth.secretKeys.adminPasswordKey) -}}
-{{- fail "postgresql.auth.secretKeys.adminPasswordKey is required when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
+{{- if and $bundledBootstrap (not $bundledAdminKey) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey (or postgresql.auth.secretKeys.adminPasswordKey) is required when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
+{{- end -}}
+{{- if and $bundledBootstrap (not $usesBundledPostgresSuperuser) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordSecretName is required when adminUsername is not the bundled postgres superuser" -}}
 {{- end -}}
 {{- $bootstrap := ternary .Values.authzed.bundledPostgresqlBootstrap .Values.authzed.externalPostgresqlBootstrap $bundledBootstrap -}}
 apiVersion: batch/v1
@@ -58,12 +72,14 @@ spec:
             - name: PGHOST
               value: formbricks-postgresql
             - name: PGUSER
-              value: postgres
+              value: {{ $bundledAdminUsername | quote }}
             - name: PGPASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ $postgresSecretName }}
-                  key: {{ .Values.postgresql.auth.secretKeys.adminPasswordKey }}
+                  name: {{ $bundledAdminSecret }}
+                  key: {{ $bundledAdminKey }}
+            - name: ADMIN_DATABASE_NAME
+              value: {{ $bundledAdminDatabase | quote }}
             - name: SPICEDB_DATABASE_NAME
               value: spicedb
             - name: SPICEDB_DATABASE_USERNAME
@@ -130,7 +146,10 @@ spec:
                 esac
                 admin_database_url="$ADMIN_DATABASE_URL"
               else
-                admin_database_url=postgres
+                # Bundled mode connects through PGHOST/PGUSER/PGPASSWORD; this is the maintenance
+                # database to attach to, which is not necessarily named `postgres` on an existing
+                # server (ENG-2390).
+                admin_database_url="${ADMIN_DATABASE_NAME:-postgres}"
               fi
               readiness_attempt=1
               until pg_isready --dbname "$admin_database_url"; do
--- a/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
+++ b/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
@@ -30,6 +30,17 @@
 {{- if and $bundledBootstrap (not $isBundledSuperuserName) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
 {{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordSecretName is required when adminUsername is not the bundled postgres superuser" -}}
 {{- end -}}
+{{- /*
+  The key needs its own explicit requirement, not just the non-empty check above. `$bundledAdminKey`
+  falls back to `postgresql.auth.secretKeys.adminPasswordKey`, which the subchart defaults to
+  `POSTGRES_ADMIN_PASSWORD` — never empty — so that check cannot fire for a custom role. Without this,
+  setting adminUsername and adminPasswordSecretName but forgetting the key renders happily and looks
+  up the *subchart's* key name inside the *operator's own* Secret, which fails in-cluster as
+  CreateContainerConfigError rather than at render. Fail-fast is the whole point of these guards.
+*/ -}}
+{{- if and $bundledBootstrap (not $isBundledSuperuserName) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey is required when adminUsername is not the bundled postgres superuser" -}}
+{{- end -}}
 {{- $bootstrap := ternary .Values.authzed.bundledPostgresqlBootstrap .Values.authzed.externalPostgresqlBootstrap $bundledBootstrap -}}
 apiVersion: batch/v1
 kind: Job
@@ -150,8 +161,10 @@ spec:
               else
                 # Bundled mode connects through PGHOST/PGUSER/PGPASSWORD; this is the maintenance
                 # database to attach to, which is not necessarily named `postgres` on an existing
-                # server (ENG-2390).
-                admin_database_url="${ADMIN_DATABASE_NAME:-postgres}"
+                # server (ENG-2390). No shell-level default: this branch only runs for bundled
+                # bootstrap, where the template always sets ADMIN_DATABASE_NAME, so a second
+                # fallback here would just be a competing source of truth for the same default.
+                admin_database_url="$ADMIN_DATABASE_NAME"
               fi
               readiness_attempt=1
               until pg_isready --dbname "$admin_database_url"; do
--- a/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
+++ b/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
@@ -37,9 +37,14 @@
   setting adminUsername and adminPasswordSecretName but forgetting the key renders happily and looks
   up the *subchart's* key name inside the *operator's own* Secret, which fails in-cluster as
   CreateContainerConfigError rather than at render. Fail-fast is the whole point of these guards.
+
+  Keyed off the *credential source*, not the username: an existing administrator that happens to be
+  called `postgres` still supplies its own Secret, and that Secret is no more likely to carry the
+  subchart's key name than any other. Keying off the name left exactly that configuration rendering
+  a dangling secretKeyRef.
 */ -}}
-{{- if and $bundledBootstrap (not $isBundledSuperuserName) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey) -}}
-{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey is required when adminUsername is not the bundled postgres superuser" -}}
+{{- if and $bundledBootstrap (not $reliesOnBundledAdminSecret) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey is required when the administrator password comes from an explicitly configured Secret" -}}
 {{- end -}}
 {{- $bootstrap := ternary .Values.authzed.bundledPostgresqlBootstrap .Values.authzed.externalPostgresqlBootstrap $bundledBootstrap -}}
 apiVersion: batch/v1
@@ -181,6 +186,14 @@ spec:
                 --set database_password="$SPICEDB_DATABASE_PASSWORD" <<'SQL'
               SELECT format('CREATE ROLE %I LOGIN', :'database_username')
                 WHERE NOT EXISTS (SELECT FROM pg_roles WHERE rolname = :'database_username') \gexec
+              -- CREATE DATABASE ... OWNER requires the administrator to be able to SET ROLE to the
+              -- owner, and merely holding CREATEROLE does not give that. PostgreSQL 16+ grants a
+              -- role's creator ADMIN but SET FALSE; PostgreSQL 15 and older grant nothing at all. So
+              -- a non-superuser administrator has to take the membership explicitly or the next
+              -- statement fails with "must be able to SET ROLE" (16+) / "must be member of role"
+              -- (15). Superusers already qualify and are skipped, so the bundled path is unchanged.
+              SELECT format('GRANT %I TO CURRENT_USER', :'database_username')
+                WHERE NOT (SELECT rolsuper FROM pg_roles WHERE rolname = CURRENT_USER) \gexec
               SELECT format('ALTER ROLE %I WITH LOGIN PASSWORD %L', :'database_username', :'database_password') \gexec
               SELECT format('CREATE DATABASE %I OWNER %I', :'database_name', :'database_username')
                 WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = :'database_name') \gexec
--- a/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
+++ b/charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
@@ -12,20 +12,22 @@
 {{- $bundledAdminDatabase := .Values.authzed.bundledPostgresqlBootstrap.adminDatabase | default "postgres" -}}
 {{- $bundledAdminSecret := .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName | default $postgresSecretName -}}
 {{- $bundledAdminKey := .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey | default .Values.postgresql.auth.secretKeys.adminPasswordKey -}}
-{{- $usesBundledPostgresSuperuser := eq $bundledAdminUsername "postgres" -}}
+{{- $isBundledSuperuserName := eq $bundledAdminUsername "postgres" -}}
+{{- $reliesOnBundledAdminSecret := and $isBundledSuperuserName (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
 {{- /*
-  Only the bundled `postgres` superuser depends on enablePostgresUser. An operator who pointed
-  adminUsername at their own CREATEROLE/CREATEDB role has no reason to also materialise a superuser
-  the subchart was told not to create — requiring it is what forced existing installations to turn
-  bootstrap off entirely (ENG-2390).
+  enablePostgresUser only matters when this Job is relying on the subchart to *create* the superuser
+  and hand over its password — that is, the default name AND the default Secret. Keying off the name
+  alone would refuse an existing server whose privileged role happens to be called `postgres` and
+  whose credentials the operator supplied explicitly, which is the same over-strict refusal that
+  forced installations to disable bootstrap entirely (ENG-2390).
 */ -}}
-{{- if and $bundledBootstrap $usesBundledPostgresSuperuser (not .Values.postgresql.auth.enablePostgresUser) -}}
+{{- if and $bundledBootstrap $reliesOnBundledAdminSecret (not .Values.postgresql.auth.enablePostgresUser) -}}
 {{- fail "Set authzed.bundledPostgresqlBootstrap.adminUsername (with adminPasswordSecretName and adminPasswordKey) to an existing role with CREATEROLE and CREATEDB, or enable postgresql.auth.enablePostgresUser, or disable authzed.bundledPostgresqlBootstrap.enabled" -}}
 {{- end -}}
 {{- if and $bundledBootstrap (not $bundledAdminKey) -}}
 {{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey (or postgresql.auth.secretKeys.adminPasswordKey) is required when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
 {{- end -}}
-{{- if and $bundledBootstrap (not $usesBundledPostgresSuperuser) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
+{{- if and $bundledBootstrap (not $isBundledSuperuserName) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
 {{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordSecretName is required when adminUsername is not the bundled postgres superuser" -}}
 {{- end -}}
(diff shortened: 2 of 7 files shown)


mattinannt (commented):
Reviewed the diff. Overall this is a well-scoped fix with strong new render-test coverage (default no-op, the regression itself, the ENG-2390 unblock path, and a credential-leak guard) — the byte-identical-defaults claim checks out and the mutation-testing note is a nice touch for a chart that had zero coverage before.

One gap in the new guards, `charts/formbricks/templates/authzed-postgresql-bootstrap.yaml:25-30`:

The PR adds a guard for a custom `adminUsername` with no `adminPasswordSecretName` (so it can't silently fall back to the bundled superuser's password) — but there's no equivalent guard for `adminPasswordKey`. If an operator sets `adminUsername` **and** `adminPasswordSecretName` but leaves `adminPasswordKey` unset (as easy to do as omitting the secret name — every example in the README/docs pairs the two, but only one is enforced), `$bundledAdminKey` silently falls back to `.Values.postgresql.auth.secretKeys.adminPasswordKey` (default `POSTGRES_ADMIN_PASSWORD`, line 14). That key almost certainly doesn't exist in the operator's own secret (the docs' own example uses `adminPasswordKey: password`), so the Job renders fine but the pod never starts — `CreateContainerConfigError: key POSTGRES_ADMIN_PASSWORD not found` — instead of getting the same crisp render-time message this PR otherwise took care to add for the sibling case.

Concretely: `--set authzed.bundledPostgresqlBootstrap.adminUsername=fbadmin --set authzed.bundledPostgresqlBootstrap.adminPasswordSecretName=existing-pg-admin` (key omitted) renders successfully today, but will fail at pod scheduling against a real `existing-pg-admin` secret whose key is anything other than `POSTGRES_ADMIN_PASSWORD`.

Worth adding a fourth guard mirroring line 28-30 (`and $bundledBootstrap (not $usesBundledPostgresSuperuser) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey)`) so this fails at `helm template`/`helm upgrade --dry-run` time with a clear message rather than at Job scheduling. Low severity since it's k8s-guarded either way (no auth ever happens with the wrong key), but it's the one asymmetry left in an otherwise thorough set of guards.

Nothing else stood out — the TLS/sslmode handling, the idempotent SQL, and the external-bootstrap path are all untouched and the new tests exercise the right scenarios.

---

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,25 @@
...
+  bootstrap off entirely (ENG-2390).
+*/ -}}
+{{- if and $bundledBootstrap $usesBundledPostgresSuperuser (not .Values.postgresql.auth.enablePostgresUser) -}}
+{{- fail "Set authzed.bundledPostgresqlBootstrap.adminUsername (with adminPasswordSecretName and adminPasswordKey) to an existing role with CREATEROLE and CREATEDB, or enable postgresql.auth.enablePostgresUser, or disable authzed.bundledPostgresqlBootstrap.enabled" -}}
 {{- end -}}
-{{- if and $bundledBootstrap (not .Values.postgresql.auth.secretKeys.adminPasswordKey) -}}
-{{- fail "postgresql.auth.secretKeys.adminPasswordKey is required when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
+{{- if and $bundledBootstrap (not $bundledAdminKey) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey (or postgresql.auth.secretKeys.adminPasswordKey) is required when bundled AuthZed PostgreSQL bootstrap is enabled" -}}
+{{- end -}}
```
mattinannt, line 29: **`adminPasswordKey` isn't actually required together with `adminPasswordSecretName` — only the secret name is.**

Lines 28-29 fail the render when `adminUsername` is overridden but `adminPasswordSecretName` is left empty. But the parallel guard for the key (lines 25-27) only checks that `$bundledAdminKey` resolves to *something*, and `$bundledAdminKey` (line 14) falls back to `.Values.postgresql.auth.secretKeys.adminPasswordKey`, which defaults to a non-empty value (`POSTGRES_ADMIN_PASSWORD` in `values.yaml`). So this check can never fire for a custom-role setup unless the operator also blanks out that subchart default.

Concrete scenario: operator sets
```yaml
authzed:
  bundledPostgresqlBootstrap:
    adminUsername: fbadmin
    adminPasswordSecretName: existing-pg-admin
```
but forgets `adminPasswordKey` (their own secret's key is e.g. `password`, not `POSTGRES_ADMIN_PASSWORD`). Render succeeds — `PGPASSWORD` ends up as `secretKeyRef{name: existing-pg-admin, key: POSTGRES_ADMIN_PASSWORD}`. That key doesn't exist in their secret, so the bootstrap Job pod fails at `CreateContainerConfigError` in the cluster, discovered only after deploying — exactly the render-time-vs-runtime failure class this PR otherwise goes out of its way to close for the secret name (see the `values.yaml` comment at lines 661-663, which explicitly says "overriding `adminUsername` alone is not enough — set **these too** [secretName **and** key] whenever the role is not the bundled `postgres` superuser" — that promise isn't enforced for the key).

Suggest gating the key check the same way the secret-name check is gated: require `authzed.bundledPostgresqlBootstrap.adminPasswordKey` to be explicitly set (not just the defaulted `$bundledAdminKey`) whenever `not $usesBundledPostgresSuperuser`, mirroring the `fail` at line 28-29.

---
Single-pass review (no Agent-tool fan-out / subagent verification available this run) — treat as one careful reviewer's pass, not the full multi-agent process.

---

mattinannt (commented):
Single-pass review of the AuthZed bootstrap admin-role changes (charts/formbricks/templates/authzed-postgresql-bootstrap.yaml, values.yaml, README.md, docs/self-hosting/advanced/authzed-operations.mdx, tests/authzed-operations.sh). No Agent-tool fan-out/subagent verify pass was available this run, so this is one reviewer's single pass rather than the full multi-agent process — treat findings accordingly.

One inline finding: the new `adminPasswordSecretName` guard (required whenever `adminUsername` is overridden) doesn't have a matching guard for `adminPasswordKey`, even though the `values.yaml` comment says both are required together. See the inline comment for the concrete failure scenario.

Everything else checked out: the `$usesBundledPostgresSuperuser` gating logic, the `ADMIN_DATABASE_NAME` plumbing into the bash script, the new test assertions, and the docs/README additions are all internally consistent with each other and with the template.

---

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,25 @@
...
+{{- if and $bundledBootstrap (not $usesBundledPostgresSuperuser) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordSecretName is required when adminUsername is not the bundled postgres superuser" -}}
```
mattinannt, line 31: **Missing guard for `adminPasswordKey` when a custom admin role is configured.**

This guard requires `adminPasswordSecretName` whenever `adminUsername` isn't the bundled `postgres` superuser (correctly closing the "silently uses the bundled superuser's password" gap called out in the PR description). But there's no equivalent requirement for `adminPasswordKey`.

If an operator sets `adminUsername` + `adminPasswordSecretName` (their own Secret) but leaves `adminPasswordKey` unset, `$bundledAdminKey` silently falls back to `.Values.postgresql.auth.secretKeys.adminPasswordKey` (default `"POSTGRES_ADMIN_PASSWORD"`) — the *bundled subchart's* key name, looked up inside the *operator's own* Secret. Since that Secret is unlikely to have a key with that exact name, the render succeeds but the Job's Pod then fails at scheduling with a missing-secret-key error (`CreateContainerConfigError`), not inside the container/Job logs as the PR's "Risks & regressions" section implies ("a wrong username or an under-privileged role still surfaces in Job logs").

Given the design intent here is explicitly to fail fast at render for exactly this class of misconfiguration, consider also requiring `adminPasswordKey` be explicit whenever `adminUsername` is overridden, e.g.:
```
{{- if and $bundledBootstrap (not $usesBundledPostgresSuperuser) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey) -}}
{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey is required when adminUsername is not the bundled postgres superuser" -}}
{{- end -}}
```

(Single-pass review without the Agent tool fan-out — flagging for a second look rather than treating this as certain.)

---

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -130,7 +146,10 @@ spec:
                 esac
                 admin_database_url="$ADMIN_DATABASE_URL"
               else
-                admin_database_url=postgres
+                # Bundled mode connects through PGHOST/PGUSER/PGPASSWORD; this is the maintenance
+                # database to attach to, which is not necessarily named `postgres` on an existing
+                # server (ENG-2390).
+                admin_database_url="${ADMIN_DATABASE_NAME:-postgres}"
```
mattinannt: Minor: `${ADMIN_DATABASE_NAME:-postgres}` is dead-code redundancy. This `else` branch only runs when `$bundledBootstrap` is true, and in that case the template always sets `ADMIN_DATABASE_NAME` via `$bundledAdminDatabase := ... | default "postgres"` (line ~12), which is never empty. The shell-level `:-postgres` fallback can never actually trigger — the default already lives one layer up. Not a bug, just two places doing the same defaulting; could drop the shell fallback (`admin_database_url="$ADMIN_DATABASE_NAME"`) for one source of truth.

---

mattinannt (commented):
Single-pass review (high effort, recall-oriented, no Agent-tool fan-out/verify pass — treat findings as candidates for a second look, not pre-verified). This is a small, well-scoped, well-tested chart change (5 files, +143/-8); most angles (removed-behavior, cross-file callers, test-script logic, i18n/env/Prisma conventions — N/A here) came back clean. Two findings posted inline, both in `charts/formbricks/templates/authzed-postgresql-bootstrap.yaml`:

1. The new `adminPasswordSecretName` guard isn't mirrored for `adminPasswordKey` — a custom `adminUsername` + custom Secret without an explicit key silently defaults to the bundled subchart's key name against the operator's own Secret, failing at Pod scheduling rather than at render (the failure mode this PR otherwise designed to catch early).
2. Minor: the shell fallback `${ADMIN_DATABASE_NAME:-postgres}` is unreachable dead code since the Helm template already guarantees a non-empty default one layer up.

No other correctness issues found (guard ordering, `eq $bundledAdminUsername "postgres"` heuristic, quoting, SQL injection surface, test assertions, docs) checked out under the codebase's stated invariant that a `postgres` role only exists via the bundled subchart's `enablePostgresUser`.

---

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,25 @@
...
+{{- $bundledAdminKey := .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey | default .Values.postgresql.auth.secretKeys.adminPasswordKey -}}
+{{- $usesBundledPostgresSuperuser := eq $bundledAdminUsername "postgres" -}}
+{{- /*
+  Only the bundled `postgres` superuser depends on enablePostgresUser. An operator who pointed
+  adminUsername at their own CREATEROLE/CREATEDB role has no reason to also materialise a superuser
+  the subchart was told not to create — requiring it is what forced existing installations to turn
+  bootstrap off entirely (ENG-2390).
+*/ -}}
+{{- if and $bundledBootstrap $usesBundledPostgresSuperuser (not .Values.postgresql.auth.enablePostgresUser) -}}
+{{- fail "Set authzed.bundledPostgresqlBootstrap.adminUsername (with adminPasswordSecretName and adminPasswordKey) to an existing role with CREATEROLE and CREATEDB, or enable postgresql.auth.enablePostgresUser, or disable authzed.bundledPostgresqlBootstrap.enabled" -}}
```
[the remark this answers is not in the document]
xernobyl: Valid, and it's the same over-strict refusal this PR set out to remove — just narrower. Fixed in `db696c9c3`.

The predicate keyed off the role **name**, so an existing server whose privileged role happens to be called `postgres`, with credentials supplied explicitly via `adminPasswordSecretName`, was still refused under `enablePostgresUser=false`.

It now keys off the default *credential configuration* instead:

```
$reliesOnBundledAdminSecret := and $isBundledSuperuserName (not adminPasswordSecretName)
```

The precondition applies only when this Job is relying on the subchart both to create the superuser and to hand over its password. Supplying a Secret is the operator asserting the role already exists, which is exactly the configurable-administrator contract.

The separate "custom username requires a Secret" guard still keys off the name, deliberately — that one exists to stop a non-`postgres` role silently reusing the bundled password, so the name is the right signal there.

Added your suggested render case (`adminUsername` left at `postgres`, explicit Secret, `enablePostgresUser=false` → must render), and mutation-tested it: restoring the name-only predicate makes the harness exit 1.

charts/formbricks/README.md
```diff
@@ -131,6 +131,26 @@ The administrator URL must explicitly set `sslmode=require`, `sslmode=verify-ca`
...
+    adminUsername: fbadmin
+    adminDatabase: formbricks # the maintenance database to attach to
+    adminPasswordSecretName: existing-pg-admin
+    adminPasswordKey: password
+```
+
+`adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
+would silently fall back to the bundled admin password and fail to authenticate. The Job creates the role and
+database only when they are absent, so re-running it against an already initialised database is a no-op, and
+`authzed.bundledPostgresqlBootstrap.enabled=false` remains available for operators who provision both by hand.
```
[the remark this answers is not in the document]
xernobyl: Correct, and I'd asserted the opposite in two docs, the commit message and the PR body. Fixed in `db696c9c3`.

Checked the SQL rather than trusting my own summary:

```sql
SELECT format('CREATE ROLE %I LOGIN', …)    WHERE NOT EXISTS (…) \gexec
SELECT format('ALTER ROLE %I WITH LOGIN PASSWORD %L', …)         \gexec   -- unconditional
SELECT format('CREATE DATABASE %I OWNER %I', …) WHERE NOT EXISTS (…) \gexec
```

Creation is guarded; the password reconcile is not. So a re-run resets the `spicedb` role's password to whatever the chart Secret holds.

That has an operational consequence worth stating rather than just removing the word "no-op": **rotating that password outside Helm is silently undone by the next upgrade.** Both documents now say the run is safe but not inert, and to update the Secret alongside any out-of-band rotation. The PR body carried the same wrong claim and is corrected too.

xernobyl:
Both CodeRabbit findings addressed in `db696c9c3` — replies inline. Summary:

1. **Precondition keyed off the role name** — so an existing privileged role that happens to be called `postgres`, with an explicitly supplied Secret, was still refused. That is the same over-strict refusal this PR exists to remove. Now keyed off the default *credential configuration*, with your suggested render case added and mutation-tested.
2. **"Re-running is a no-op" was wrong** — `ALTER ROLE … PASSWORD` is unconditional, so every run reconciles the `spicedb` role's password to the chart Secret. Corrected in both docs and in the PR body, with the operational consequence spelled out: rotate that password outside Helm and the next upgrade silently reverts it.

Second one is a good catch on a claim I'd repeated in four places without re-reading the SQL.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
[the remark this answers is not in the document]
xernobyl: Confirmed and fixed in `02eebe14c`. You're right that the guard is unreachable, and I reproduced it before changing anything: `$bundledAdminKey` falls back to `postgresql.auth.secretKeys.adminPasswordKey`, which `values.yaml:1479` defaults to `POSTGRES_ADMIN_PASSWORD` — never empty — so `not $bundledAdminKey` can't fire for a custom role.

Your scenario is exactly the failure: `adminUsername` + `adminPasswordSecretName` without the key renders happily with `secretKeyRef{name: existing-pg-admin, key: POSTGRES_ADMIN_PASSWORD}` — the subchart's key name looked up inside the operator's own Secret — and dies as `CreateContainerConfigError` after deploy. That is the render-time-vs-runtime class this PR closes for the secret name, left open for the key.

Worth adding: it also broke the promise in my own `values.yaml` comment, which tells operators to set **both**. The comment described a contract nothing enforced.

Added the parallel guard you suggested, and a render case for it, mutation-tested — removing the guard makes the harness exit 1.

charts/formbricks/README.md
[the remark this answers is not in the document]
xernobyl: Same root cause as the sibling comment on line 29 — fixed together in `02eebe14c` with an explicit `adminPasswordKey` requirement mirroring the secret-name guard, plus a mutation-tested render case.

Also correcting the PR's "Risks & regressions" wording you quoted: I wrote that a misconfiguration "still surfaces in Job logs", but as you note a missing Secret key never reaches the container — it fails at Pod creation. That was wrong and the guard now removes the case entirely.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
xernobyl: Agreed, dropped in `02eebe14c` — now `admin_database_url="$ADMIN_DATABASE_NAME"`.

You're right it can't trigger: that branch only runs for bundled bootstrap, and there the template always sets `ADMIN_DATABASE_NAME` from `$bundledAdminDatabase`, which already carries the `postgres` default. Two layers defaulting the same value with the lower one unreachable — the comment now says why there is no shell fallback, so it doesn't get re-added.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
xernobyl: Fixed in `02eebe14c` (answered on the sibling thread too). Reproduced first: `$bundledAdminKey` falls back to `postgresql.auth.secretKeys.adminPasswordKey`, which `values.yaml:1479` defaults to `POSTGRES_ADMIN_PASSWORD`, so the existing check can never fire for a custom role. Added the parallel `adminPasswordKey` guard, with a mutation-tested render case.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
xernobyl: Same fix, `02eebe14c` — explicit `adminPasswordKey` requirement whenever `adminUsername` is overridden, mirroring the secret-name guard.

Also corrected the PR's "Risks & regressions" claim you quoted: I wrote that a misconfiguration "still surfaces in Job logs", but a missing Secret key never reaches the container — it fails at Pod creation. Wrong as written, and the guard removes the case.

charts/formbricks/README.md
```diff
@@ -131,6 +131,29 @@ The administrator URL must explicitly set `sslmode=require`, `sslmode=verify-ca`
...
+authzed:
+  bundledPostgresqlBootstrap:
+    adminUsername: fbadmin
+    adminDatabase: formbricks # the maintenance database to attach to
+    adminPasswordSecretName: existing-pg-admin
+    adminPasswordKey: password
+```
+
+`adminPasswordSecretName` is required whenever `adminUsername` is not the bundled superuser — otherwise the Job
+would silently fall back to the bundled admin password and fail to authenticate.
```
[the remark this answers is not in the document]
xernobyl: Fixed in `db696c9c3`. Checked the SQL rather than trusting my own summary: `CREATE ROLE` and `CREATE DATABASE` are guarded by `WHERE NOT EXISTS`, but `ALTER ROLE … PASSWORD` is unconditional — so a re-run resets the `spicedb` role's password to the chart Secret.

Both documents now state the consequence rather than just dropping the word "no-op": rotating that password outside Helm is silently undone by the next upgrade, so the Secret has to be updated alongside it.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,38 @@
...
+{{- $reliesOnBundledAdminSecret := and $isBundledSuperuserName (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordSecretName) -}}
+{{- /*
+  enablePostgresUser only matters when this Job is relying on the subchart to *create* the superuser
+  and hand over its password — that is, the default name AND the default Secret. Keying off the name
+  alone would refuse an existing server whose privileged role happens to be called `postgres` and
+  whose credentials the operator supplied explicitly, which is the same over-strict refusal that
+  forced installations to disable bootstrap entirely (ENG-2390).
+*/ -}}
+{{- if and $bundledBootstrap $reliesOnBundledAdminSecret (not .Values.postgresql.auth.enablePostgresUser) -}}
+{{- fail "Set authzed.bundledPostgresqlBootstrap.adminUsername (with adminPasswordSecretName and adminPasswordKey) to an existing role with CREATEROLE and CREATEDB, or enable postgresql.auth.enablePostgresUser, or disable authzed.bundledPostgresqlBootstrap.enabled" -}}
```
BhagyaAmarasinghe, line 25: [P1] The documented non-superuser path does not work with only `CREATEROLE` and `CREATEDB`. I reproduced the rendered SQL against `postgres:17-alpine`: after this administrator creates the `spicedb` role, `CREATE DATABASE ... OWNER spicedb` fails with `ERROR: must be able to SET ROLE "spicedb"`, because PostgreSQL grants the creator `ADMIN` but `SET FALSE` by default. When `spicedb` already exists, the unconditional `ALTER ROLE` also fails unless the administrator has `ADMIN OPTION` on that role. Establish the required membership in the SQL or require and document the additional privileges, and cover both fresh and existing-role cases with a real PostgreSQL runtime test.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,38 @@
...
+{{- end -}}
+{{- /*
+  The key needs its own explicit requirement, not just the non-empty check above. `$bundledAdminKey`
+  falls back to `postgresql.auth.secretKeys.adminPasswordKey`, which the subchart defaults to
+  `POSTGRES_ADMIN_PASSWORD` — never empty — so that check cannot fire for a custom role. Without this,
+  setting adminUsername and adminPasswordSecretName but forgetting the key renders happily and looks
+  up the *subchart's* key name inside the *operator's own* Secret, which fails in-cluster as
+  CreateContainerConfigError rather than at render. Fail-fast is the whole point of these guards.
+*/ -}}
+{{- if and $bundledBootstrap (not $isBundledSuperuserName) (not .Values.authzed.bundledPostgresqlBootstrap.adminPasswordKey) -}}
```
BhagyaAmarasinghe: [P2] This guard is based on the username, so an existing administrator still named `postgres` can provide `adminPasswordSecretName` without `adminPasswordKey`. That configuration currently renders and points at the fallback `POSTGRES_ADMIN_PASSWORD` key inside the custom Secret, causing `CreateContainerConfigError` when the Secret uses another key. Require an explicit key whenever administrator credentials come from an explicit Secret, as the unresolved CodeRabbit finding suggests.

charts/formbricks/tests/authzed-operations.sh
```diff
@@ -126,4 +126,72 @@ helm template authzed-external "${CHART_DIR}" "${COMMON_ARGS[@]}" \
...
+
+# A custom admin role with no Secret would silently fall back to the bundled superuser's password.
+if render_bootstrap authzed-bootstrap-admin-without-secret \
+  --set authzed.bundledPostgresqlBootstrap.adminUsername=fbadmin >/dev/null 2>&1; then
+  printf '%s\n' "Bootstrap must require adminPasswordSecretName when adminUsername is overridden." >&2
+  exit 1
+fi
+
+# Credentials reach the Job only by reference, in every mode.
+if grep --extended-regexp 'PGPASSWORD: |password: [^{]' <<<"${existing_admin_bootstrap}" >/dev/null; then
```
BhagyaAmarasinghe: [P3] This regex does not inspect the `PGPASSWORD` environment entry and misses a literal leak such as `- name: PGPASSWORD` followed by `value: leaked-admin-password`. Assert structurally that this env entry uses `valueFrom.secretKeyRef` and has no literal `value`; otherwise the stated secret-safety test can pass after the exact regression it is intended to detect.

BhagyaAmarasinghe (changes requested):
Requesting changes for two runtime correctness blockers and one test gap. The primary non-superuser bootstrap path fails against PostgreSQL 17 with the documented CREATEROLE plus CREATEDB privileges, and explicitly supplied postgres credentials can still render with the wrong password-key fallback. The Helm render and lint checks are green, but they do not execute the bootstrap SQL.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
xernobyl: Reproduced exactly as you describe, and it turned out to be broader than reported. Fixed in `8baaff5c1`.

Against `postgres:17-alpine`, a role with only `CREATEROLE CREATEDB` running the rendered SQL:

```
CREATE ROLE
ALTER ROLE
ERROR:  must be able to SET ROLE "spicedb"
```

and its membership row is `spicedb admin=true set=false` — the PostgreSQL 16 implicit grant, precisely as you said.

**It also fails on PostgreSQL 15, differently.** There the creator gets *no* membership at all (`pg_has_role(..., 'MEMBER') = false`) and the same statement fails with `must be member of role "spicedb"`. So the documented non-superuser path was broken on every supported server, not only 16+. That also ruled out the obvious portable guard: `MEMBER` is true on 17 while the operation still fails, and `SET` is not a valid privilege type before 16 (`ERROR: unrecognized privilege type: "SET"`), so neither works across the split.

**Fix** — take the membership explicitly, before `CREATE DATABASE`:

```sql
SELECT format('GRANT %I TO CURRENT_USER', :'database_username')
  WHERE NOT (SELECT rolsuper FROM pg_roles WHERE rolname = CURRENT_USER) \gexec
```

Skipped for superusers, so the bundled path is byte-identical and does not collect a membership it cannot use — asserted, because granting unconditionally was the simpler option and I did not want it creeping back.

**Runtime test**, as you asked — `charts/formbricks/tests/authzed-bootstrap-runtime.sh`, wired into Helm Chart Validation. It runs on 15 and 17 and covers fresh, rerun, bundled-superuser, and pre-existing-role. Two things about how it is built:

- It extracts the script from the **rendered Job** rather than restating the SQL, so it cannot drift from the template and cannot pass against SQL the chart does not ship.
- Mutation-verified: removing the `GRANT` turns it red (`must be member of role "spicedb"` on 15, `must be able to SET ROLE` on 17). It skips itself when Docker is absent so the render suite still runs anywhere.

**On the existing-role case** — you are right that `ALTER ROLE` needs `ADMIN OPTION`, and this is version-split too. Before 16, `CREATEROLE` carried authority over every non-superuser role, so it just works. From 16 it does not, and a `spicedb` role created by *someone else* is genuinely unfixable from inside the Job: it cannot grant itself a role it has no `ADMIN OPTION` on. So that one is documented rather than papered over — README, values.yaml and the ops guide now state the requirement and the remedy (`GRANT spicedb TO fbadmin WITH ADMIN OPTION`, or one superuser run) — and the test asserts it fails *loudly* on 16+ while still succeeding on 15, so neither behaviour can regress silently.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
xernobyl: Correct, and it was CodeRabbit's open finding on this file too — I had replied to the sibling thread about the *custom-role* case and never came back to this one. Fixed in `8baaff5c1`.

The guard now keys off the credential source rather than the username:

```
{{- if and $bundledBootstrap (not $reliesOnBundledAdminSecret) (not ...adminPasswordKey) -}}
```

`$reliesOnBundledAdminSecret` is already `and $isBundledSuperuserName (not adminPasswordSecretName)`, so `not` of it is exactly the set you describe: any custom username, **or** `postgres` with an explicitly supplied Secret. Both are cases where the key would otherwise fall back to the subchart's name inside an operator-managed Secret.

Added the render case (`adminUsername` left at `postgres`, explicit Secret, no key → must be refused) and mutation-tested it: restoring the username-keyed predicate makes the suite exit 1.

charts/formbricks/tests/authzed-operations.sh
xernobyl: Confirmed, and the test was worse than not existing — it read as coverage of the credential boundary while being unable to fail. Fixed in `8baaff5c1`.

Proved it first, against the exact shape you named:

```
$ printf '            - name: PGPASSWORD\n              value: leaked-admin-password\n' > leak.yaml
$ grep -E 'PGPASSWORD: |password: [^{]' leak.yaml
(no match)
```

`PGPASSWORD: ` is a shape the renderer never emits — env entries are `- name: PGPASSWORD` followed by `value:` or `valueFrom:` — so the pattern could only ever match nothing.

A better regex is not available here: over the whole manifest, `value:` under `PGPASSWORD` is indistinguishable from the legitimate one under `PGHOST`. So it now walks the env list and reports how each entry is supplied, asserted for `PGPASSWORD` and `SPICEDB_DATABASE_PASSWORD` in bundled mode and `ADMIN_DATABASE_URL` / `SPICEDB_DATABASE_PASSWORD` in external mode — the external path had no such assertion at all before.

Verified directly against four shapes rather than only in situ, since the in-situ mutation trips an earlier assertion first and would not have exercised this one:

| manifest | result | exit |
| --- | --- | --- |
| `valueFrom: secretKeyRef` | `reference` | 0 |
| `value: leaked-admin-password` | `literal` | 1 |
| literal as the **last** env entry | `literal` | 1 |
| variable absent entirely | `absent` | 2 |

The last-entry and absent cases matter: the first exercises the `END` block, and the second stops the assertion passing vacuously if the variable is ever renamed.

charts/formbricks/templates/authzed-postgresql-bootstrap.yaml
```diff
@@ -8,11 +8,38 @@
...
+{{- fail "authzed.bundledPostgresqlBootstrap.adminPasswordKey is required when adminUsername is not the bundled postgres superuser" -}}
 {{- end -}}
```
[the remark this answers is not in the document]
xernobyl: You were right and this stayed open — I answered the custom-role half on the sibling thread and never addressed the `postgres`-with-explicit-Secret half you raised here. Bhagya independently hit the same gap in review. Fixed in `8baaff5c1`.

The guard is now keyed off `$reliesOnBundledAdminSecret` exactly as you proposed, and the render case you asked for (`adminUsername=postgres`, explicit Secret, no `adminPasswordKey` → must be refused) is in `authzed-operations.sh`, mutation-tested against the old username-keyed predicate.

BhagyaAmarasinghe (approved):

PR #8876, opened 2026-08-16
refactor: route MCP workspace discovery through authorization

BhagyaAmarasinghe:
## What & why

MCP `list_workspaces` currently resolves its complete authoritative workspace set from PostgreSQL, but scalar `can()` calls cannot compare a list without causing one SpiceDB request per workspace. This PR adds a narrow Phase 1 list observation: it returns the unchanged PostgreSQL result immediately, then compares the selected legacy workspace set with one post-response SpiceDB `LookupResources` stream.

The change adds a private, server-only lookup operation with the existing AuthZed deadline/retry/consistency/error contract; a bounded workspace-to-organization resolver; set-level shadow telemetry; and MCP user/API-key integration through the existing authorization request context. It emits only bounded counts and stable error metadata. The public AuthZed barrel, the 35-action vocabulary, response DTOs, permission schema, projections, and enforcement behavior remain unchanged. Generic or enforcement-authoritative list authorization remains ENG-1713.

Cross-organization API-key grants remain filtered at the shared authentication boundary. This PR adds a bounded, identifier-free warning at that real filter, scopes the workspace-list database read by both organization and permitted IDs, and retains the result filter as defense in depth. It also restores the three explicit organization memberships required after #8869 changed team-derived membership semantics. That repair is isolated in commit `cd99273fa` so it can be cherry-picked independently.

## Linear ticket

Fixes ENG-2389

## How this was tested

- `pnpm --filter @formbricks/web exec vitest run lib/authzed lib/authorization lib/workspace/service.test.ts app/api/v3/workspaces modules/mcp modules/api/lib/api-key-auth.test.ts` ✅ — 55 files, 810 tests
- Real PostgreSQL integration test ✅ — 1- and 100-workspace lists each recorded exactly one positive central authorization operation
- `pnpm authzed:validate` ✅ — 36 relationships, 149 assertions, 2 expected relations
- `pnpm authzed:smoke` ✅ — real user/API-key `LookupResources`, empty and multi-workspace results, consistency, outage recovery, persistence, and secret scanning
- Focused ESLint, Prettier, `bash -n docker/authzed-smoke.sh`, and `git diff --check` ✅
- `pnpm build --filter=@formbricks/web... --force` ✅ — 13/13 tasks
- Live sandbox/browser/API/MCP QA was intentionally deferred to the final consolidated ENG-2364 run, per the ticket plan. That operator-owned run will query `formbricks_authzed_authorization_comparisons_total` in the internal observability platform; this PR deliberately adds no public metrics or browser surface.

## Breaking changes

None.

## QA / Test Plan

**How to test**

- [ ] Connect an MCP OAuth user with a supported resource-read scope and call `list_workspaces` → the tool returns every currently accessible workspace once, ordered by name and then ID, with the existing response shape.
- [ ] Use a user that belongs to multiple selected organizations, including owner/manager and team-derived access → only the same PostgreSQL-authorized workspace set is returned; the shadow comparison count increases by one and reports a match.
- [ ] Use a user with no accessible workspaces → the tool returns an empty list and one matching deny/deny comparison.
- [ ] Connect through an API key with duplicate read/write/manage grants → the tool returns each existing same-organization workspace once, with no foreign-organization workspace.
- [ ] Use a write-only MCP OAuth token → scope enforcement denies the tool before `list_workspaces` runs and no comparison is emitted.
- [ ] Revoke or deactivate the credential and retry → authentication fails with the existing status/error body and no comparison is emitted.
- [ ] Add a disposable cross-organization API-key grant in test data → the foreign workspace is absent from the legacy response; if represented in SpiceDB it is counted as an AuthZed-extra mismatch without exposing any identifier.
- [ ] In the internal observability platform, replace `$cohort` with the configured cohort label and run:

  ```promql
  sum by (outcome) (
    increase(formbricks_authzed_authorization_comparisons_total{
      surface="mcp",
      action="workspace.read",
      resource_type="workspace",
      cohort="$cohort"
    }[30m])
  )
  ```

  A matching list emits one `match` sample. Bidirectional drift emits one sample for each mismatch direction, so both mismatch series must remain zero and the total counter samples must not be treated as a request count.

**Preconditions / test data**

- Deploy the merged image to `authzed-sandbox`, complete schema validation and relationship repair, and keep all enforcement target/organization lists empty.
- Enable `mcp:user` and `mcp:apiKey` shadow targets for one disposable organization cohort with a fresh cohort label and snapshot floor.
- Prepare OAuth users and API keys covering allowed, empty, insufficient-scope, revoked, duplicate-grant, and cross-tenant cases.

**Risks & regressions**

- Set comparison must attribute looked-up workspace IDs to the selected organization cohort without leaking identifiers.
- API-key discovery must remain restricted to its owning organization even if stale or invalid cross-tenant relationships exist.
- Queue saturation or AuthZed/PostgreSQL resolver failures must remain post-response operational telemetry and never alter the authoritative MCP result.
- The number of AuthZed lookups and central operation counts must remain one regardless of list size.
- One list observation may emit two directional mismatch samples; checks-per-request remains the request-amplification signal.

- none — no schema, database, environment, projection, permission, or enforcement change

## CodeRabbit review fixes

- Tests now exercise the production rollout matching helpers and V3 actor conversion instead of copying their implementations.
- The AuthZed public-barrel type assertion is included in the real web typecheck configuration.
- The Docker smoke cleanup removes team relationships that reference the disposable organization as a subject before recreating the graph.
- API-key workspace discovery now uses one organization-and-ID-scoped batch query instead of ID-only reads, with the scoped Prisma filter covered by a service test.
- Empty-user and API-key list tests require exactly one shadow observation.
- Latest review-fix validation: 55 focused files / 810 tests and web typecheck pass; the earlier 149 schema assertions and isolated Docker smoke suite remain green.

## Matti review fixes

- Detects user and API-key cross-tenant AuthZed results while ignoring real actor organizations outside the rollout cohort.
- Pages LookupResources in 250-result requests, rejects stalled cursors, and fails safely above 20,000 accumulated results.
- Resolves independent workspace batches concurrently.
- Centralizes Formbricks-to-SpiceDB object-type mapping and sanitized comparison-error normalization.
- Emits a bounded warning count at the shared API-key authentication filter where invalid cross-organization grants are actually dropped; the list path retains defense-in-depth filtering.
- Documents that bidirectional workspace-list drift emits two directional comparison samples and provides the exact cohort-scoped PromQL used for final QA.
- Isolates the three #8869 smoke-membership repairs in cherry-pickable commit `cd99273fa`.
- Review validation: 55 focused files / 810 tests, web typecheck, 149 schema assertions, and the isolated Docker smoke suite all pass.


apps/web/lib/authorization/workspace-list-observer.ts
```diff
@@ -0,0 +1,261 @@
...
+    for (const workspaceId of lookup.resourceIds) {
+      const organizationId = organizationByWorkspaceId.get(workspaceId);
+      if (!organizationId) {
+        unattributedAuthzedCount += 1;
+        continue;
+      }
+
+      if (context.selectedOrganizationIds.has(organizationId)) {
+        authzedWorkspaceIds.add(workspaceId);
+      } else if (context.actor.type === "apiKey") {
```
mattinannt: **[correctness, single-pass review without Agent fan-out]** Cross-organization drift is only detected for `apiKey` actors, not `user` actors, creating a blind spot in the exact telemetry meant to catch it.

For a `user` actor, when an AuthZed-returned `workspaceId` resolves to an `organizationId` that is not in `context.selectedOrganizationIds`, the loop does nothing (no `unattributedAuthzedCount` increment, no comparison, no log). `selectedOrganizationIds` is derived only from `observation.organizationIds` (the orgs the user actually belongs to per Postgres) filtered by rollout cohort — so this same branch silently swallows two very different situations: (1) a legitimate org the user belongs to that's simply out of the shadow rollout cohort (fine to skip), and (2) SpiceDB granting the user `read` on a workspace in an organization they have **zero membership in** per Postgres — a genuine cross-tenant relationship-drift bug, exactly the class of issue the `apiKey` branch right below explicitly calls out as "security-relevant drift even when the foreign organization is outside this rollout cohort."

Concrete scenario: user Alice belongs only to `org-A` (in the shadow cohort). A stale/erroneous SpiceDB relationship grants her `read` on `workspace-X` in `org-B`, an organization she is not a member of at all. Because `org-B` is absent from `observation.organizationIds`, it's also absent from `selectedOrganizationIds`; the `else if (context.actor.type === "apiKey")` guard means this drift is never flagged for a `user` actor — `unattributedAuthzedCount` stays 0, the comparison silently reports `match`/no comparison at all, and this exact class of leak (which the code demonstrably considers worth catching for API keys) goes undetected for session users.

---

apps/web/lib/authorization/resolvers.ts
```diff
@@ -84,6 +91,39 @@ export const getWorkspaceOrganizationId = reactCache(async (workspaceId: string)
...
+export const getWorkspaceOrganizationReferences = async (
+  workspaceIds: ReadonlyArray<string>
+): Promise<ReadonlyArray<TWorkspaceOrganizationReference>> => {
+  const uniqueWorkspaceIds = [...new Set(workspaceIds)];
+  if (uniqueWorkspaceIds.length === 0) return [];
+
+  const references: TWorkspaceOrganizationReference[] = [];
+  try {
+    for (
+      let offset = 0;
```
mattinannt, line 110: **[efficiency, single-pass review without Agent fan-out]** `getWorkspaceOrganizationReferences` awaits each 500-id batch sequentially in a `for` loop even though the batches are independent `IN` queries with no data dependency between them. For a `LookupResources` result large enough to span multiple batches (e.g. 1500 workspace IDs → 3 batches), this serializes 3 round trips where `Promise.all` over the batch queries would let them run concurrently, cutting the wall-clock cost roughly 3x. Since this runs in the post-response shadow path it's not user-facing latency, but it does needlessly extend how long the comparison job (and thus the bounded `MAX_COMPARISON_JOBS` queue) stays occupied.

---

apps/web/lib/authorization/workspace-list-observer.ts
```diff
@@ -0,0 +1,261 @@
...
+const runWorkspaceListComparison = async (context: TWorkspaceListComparisonContext): Promise<void> => {
+  const startedAt = performance.now();
+
+  try {
+    const lookup = await getAuthzedClient().lookupResources({
+      permission: "read",
+      resourceType: "workspace",
+      subject: {
+        objectId: context.actor.id,
+        objectType: context.actor.type === "apiKey" ? "api_key" : "user",
```
mattinannt: **[reuse, single-pass review without Agent fan-out]** This inline ternary re-implements the actor-type → SpiceDB object-type mapping (`"user"` → `"user"`, `"apiKey"` → `"api_key"`) that already exists as `getSpicedbObjectType` in `apps/web/lib/authorization/spicedb-evaluator.ts` (used there for both actor and resource-type mapping). It's currently private to that module, so reusing it here would mean exporting it, but duplicating the mapping inline instead of centralizing it means the two call sites can silently drift if the SpiceDB schema's object-type naming ever changes.

---

mattinannt (commented):
Single-pass code review (Agent tool unavailable in this context — no multi-agent fan-out or subagent verify pass ran). 3 findings posted inline, ranked by severity. Full diff reviewed: `1376633...pr-8876` (base `epic/authzed` @ 1376633, head e75de91), 19 files, +1106/-50.

---

apps/web/lib/authorization/workspace-list-observer.ts
mattinannt: For a `user` actor this `else if` is skipped entirely, so a workspace that AuthZed attributes to an organization the user isn't a Postgres member of at all (not merely one outside the current rollout cohort) is silently dropped instead of counted in `unattributedAuthzedCount`.

`selectedOrganizationIds` is a rollout-filtered subset of the user's real orgs, so `!selectedOrganizationIds.has(organizationId)` conflates two very different cases for users: (a) the org is one of the user's real orgs but just outside this cohort — expected, not drift — and (b) the org isn't among the user's real orgs at all — a genuine cross-tenant AuthZed grant. Only case (b) should count as drift, the same way it does for `apiKey` actors just below.

Given the PR description explicitly calls out "stale or invalid cross-tenant relationships" as a risk to guard against, and today that guard only fires for `apiKey`, should this branch also flag case (b) for `user` actors (e.g. by checking membership in `observation.organizationIds` rather than only `selectedOrganizationIds`)?

---

apps/web/lib/authzed/client.ts
```diff
@@ -546,6 +578,50 @@ const createAuthzedClient = (requestTimeoutMs: number): TAuthzedClientSingleton
...
       }),
+    lookupResources: async (lookup) => {
+      validateResourceLookup(lookup);
+
+      return executeAuthzedOperation("lookup_resources", async () => {
+        const responses = await sdkClient.promises.lookupResources({
+          consistency: getPermissionCheckConsistency(config),
+          context: undefined,
+          optionalCursor: undefined,
+          optionalLimit: 0,
```
mattinannt: `lookupResources` sends `optionalLimit: 0` with no cursor loop. The comment on `validateRelationshipQuery` a few dozen lines up (around line 381) spells out exactly why this is risky for `readRelationships`: "SpiceDB treats `optionalLimit: 0` as unlimited, which under the channel-wide deadline is a guaranteed timeout and an unbounded allocation, because the promisified streaming call buffers every message before it resolves" — and that call enforces a mandatory positive, capped limit for this reason.

`lookupResources` uses the identical `sdkClient.promises.*` buffer-everything pattern but doesn't apply that same guard. For a principal with an unexpectedly large AuthZed-visible workspace set (stale/looping relationships, a large org), this call could hit the same deadline/memory failure mode this file otherwise guards against. Since results also feed the post-response shadow comparison rather than the authoritative response, a failure here is contained — but is an unbounded page/cursor loop (mirroring `readRelationships`) intentionally out of scope for this Phase 1 cut, or worth adding now?

---

apps/web/lib/authorization/workspace-list-observer.ts
```diff
@@ -0,0 +1,261 @@
...
+  target: NonNullable<ReturnType<typeof getAuthorizationRolloutTarget>>;
+  workspaceIds: ReadonlySet<string>;
+}>;
+
+const toDecisionLabel = (allowed: boolean | undefined): TAuthorizationDecisionLabel => {
+  if (allowed === undefined) return "unknown";
+  return allowed ? "allow" : "deny";
+};
+
+const normalizeOperationalError = (error: unknown, operation: string): AuthzedError => {
```
mattinannt: `normalizeOperationalError` here is byte-for-byte identical to the existing helper of the same name in `coordinator.ts` (lines 43-58), and the `decisionLabel`/`recordComparison`/`logger.warn` shape around it (this file's `toDecisionLabel` + `recordComparison`) is structurally the same pattern as `coordinator.ts`'s `decisionLabel` + `recordComparison`. Non-blocking, but worth considering extracting the error-normalization helper (and maybe the comparison-logging shape) into a shared module so a future change to retry/error semantics doesn't need to be made in both places to stay in sync.

---

apps/web/app/api/v3/workspaces/lib/operations.ts
```diff
@@ -35,21 +42,30 @@ const serializeV3WorkspaceListItem = (workspace: {
...
   const workspaceIds = Array.from(new Set(keyAuth.workspacePermissions.map((p) => p.workspaceId)));
   const workspaces = await Promise.all(workspaceIds.map((id) => getWorkspace(id)));
-  return workspaces
-    .filter((workspace): workspace is NonNullable<typeof workspace> => workspace !== null)
-    .map(serializeV3WorkspaceListItem);
+  return {
+    items: workspaces
+      .filter(
+        (workspace): workspace is NonNullable<typeof workspace> =>
+          workspace !== null && workspace.organizationId === keyAuth.organizationId
```
mattinannt: This new `workspace.organizationId === keyAuth.organizationId` check is a good defense-in-depth addition, but a mismatch is now silently filtered out of the response with no logging/metric. If the documented invariant that a key's `workspacePermissions` only ever contains workspaces in its own org (referenced elsewhere as an assumption) were ever violated upstream, this would quietly hide it rather than surface it — the exact kind of drift this PR is otherwise instrumenting for. Worth a `log.warn` (or feeding into the same shadow-comparison telemetry) when this filter actually drops something, so a real upstream bug wouldn't go unnoticed?

---

mattinannt (commented):
Automated review (high-effort pass, single-context reviewer). Overall this is a well-scoped, careful Phase 1 shadow-observation change — the guardrails around fail-safe telemetry (never altering the authoritative response), the explicit unbounded-limit tripwire comment in `client.ts`, and the integration test coverage are all solid. Left 4 inline notes, most significant is on `workspace-list-observer.ts` — a possible gap in cross-tenant drift detection for `user` actors that the PR description's own stated risk ("stale or invalid cross-tenant relationships") calls out but currently only guards for `apiKey` actors. The rest are lower-severity: an unbounded `lookupResources` call mirroring a pattern this file otherwise explicitly avoids, some duplicated helper logic with `coordinator.ts`, and a silently-dropped cross-org filter with no observability.

---

apps/web/lib/authorization/workspace-list-observer.ts
mattinannt: The cross-organization drift check here only applies `if (context.actor.type === "apiKey")`. For session-user actors, a workspace that AuthZed grants `read` on but whose resolved `organizationId` isn't in `selectedOrganizationIds` is silently dropped: it's not added to `authzedWorkspaceIds`, and (since only the `apiKey` branch increments it) it doesn't bump `unattributedAuthzedCount` either.

Concretely: a user is a real member of orgs A and B (both in the shadow cohort). A stale/incorrect SpiceDB relationship grants them `read` on a workspace belonging to org C, which they have no membership in at all. `organizationByWorkspaceId` resolves that workspace to org C; `selectedOrganizationIds` (built from the user's real org memberships in `observeWorkspaceListAuthorization`) never contains C. Because the actor isn't an API key, this workspace is just skipped — no `legacyOnly`/`authzedOnly` diff, no `unattributedAuthzedCount`. If the rest of the sets match, the comparison logs `outcome: "match"` even though AuthZed just proved it would grant cross-org access.

That's exactly the class of bug this shadow comparison exists to catch ahead of the enforcement cutover, but it's currently blind for session users — the majority path — and only instrumented for API keys. Worth extending the `else if` to cover both actor types (or explicitly documenting why session users are exempt, if that's intentional).

---

apps/web/lib/authorization/resolvers.ts
```diff
@@ -84,6 +91,39 @@ export const getWorkspaceOrganizationId = reactCache(async (workspaceId: string)
...
+      offset < uniqueWorkspaceIds.length;
+      offset += WORKSPACE_ORGANIZATION_RESOLUTION_BATCH_SIZE
+    ) {
+      const batch = uniqueWorkspaceIds.slice(offset, offset + WORKSPACE_ORGANIZATION_RESOLUTION_BATCH_SIZE);
```
mattinannt: Minor efficiency note: each 500-id batch is awaited sequentially in this `for` loop, but the batches are independent `findMany` queries — nothing here depends on a prior batch's result. If `LookupResources` (called with `optionalLimit: 0`, i.e. unbounded) returns a large resource set — e.g. during an actual drift incident where AuthZed over-grants broadly — this turns into N/500 fully sequential DB round trips instead of running them concurrently (e.g. via `Promise.all`), which stretches out the background comparison job and, given `MAX_COMPARISON_JOBS`/`COMPARISON_CONCURRENCY` in `context.ts`, could delay other queued comparisons behind it. Not correctness-affecting, just worth batching concurrently given the doc comment above already calls out avoiding N+1.

---

mattinannt (commented):
Automated review pass. Two findings, both in the new shadow-comparison authorization code — one a real detection gap (session-user actors bypass the cross-org drift check that API keys get), one a minor efficiency note (sequential batch resolution). Rest of the diff (API-key org filter in `operations.ts`, AuthZed client's `lookupResources`, barrel split, docs/scripts) looked consistent with existing invariants.

---

xernobyl:
Reviewed `47081a91` end to end against `epic/authzed`. CodeRabbit's three and Matti's six all look genuinely addressed at this head, so I've only written up what isn't already in a thread.

### [medium] The smoke-seed change here is repairing a break I introduced in #8869, and nothing in CI would have caught it

`seedTeamWorkspaceProjection` now seeds `organization:application-graph-smoke#member` for alice and bob (`apps/web/scripts/authzed-relationships-smoke.ts:173`) and `organization:application-feedback-organization#member` for the feedback user (`:345`). That isn't scaffolding for the new lookup commands — it's load-bearing for assertions that already existed.

ENG-2340 (#8869, merged Aug 14) changed `team#member` to `(admin + contributor) & organization->product_member`. On `epic/authzed` today alice holds `team:application-graph-manager#admin` but has no membership relation on `organization:application-graph-smoke` — the base script only ever writes that org as a *subject* (team→org, workspace→org edges), never as a resource with a user subject. So the intersection is empty, and `workspace:application-graph-smoke#manage@user:alice` resolves false. `docker/authzed-smoke.sh:410` asserts it's true; bob's `read` at `:418` and the feedback-directory checks fail for the same reason. #8869 touched only `schema.zed`, `schema-validation.yaml` and `turbo.json` — it never updated the smoke seeds.

The reason nobody noticed: Docker Build Validation only triggers on `pull_request`/`merge_group` into `main`, so none of the epic-stacked PRs run it. It'll first go red on the epic→main PR.

Two things worth deciding:

- If this PR slips or gets reordered behind the other stacked ones, the base stays broken. Might be worth isolating the three seed writes in their own commit so it can be cherry-picked ahead.
- Worth a line in the PR body either way, so nobody later reads those seeds as lookup-test scaffolding and strips them.

That one's mine to own — I should have updated the seeds in #8869. Happy to file it and take the split if you'd rather keep this PR to the lookup work.

### [low] `crossOrganizationGrantCount` can never be non-zero, so the warning Matti asked for won't fire

`apps/web/app/api/v3/workspaces/lib/operations.ts:114` warns when an API key's grants include a foreign-organization workspace. But every path that produces a `TAuthenticationApiKey` for `listV3Workspaces` goes through `authenticateApiKeyFromHeaders` — MCP via `authenticateMcpApiKey`, v3 via `app/api/v1/auth.ts`'s `authenticateRequest` — and `apps/web/modules/api/lib/api-key-auth.ts:60-63` already applies the identical `workspace.organizationId === apiKey.organizationId` filter (ENG-1749, untouched on this branch). By the time `keyAuth.workspacePermissions` reaches `fetchApiKeyWorkspaces`, foreign grants are already gone.

The filter at `operations.ts:62-66` is fine as defense in depth. The telemetry isn't: `operations.test.ts` hand-builds an authentication object carrying a cross-org permission, which is a shape the request path cannot produce, so the test passes against a branch that is unreachable in production. If the goal is to notice a real upstream invariant violation — which is what Matti asked for — the warning belongs on the ENG-1749 filter itself, which is where the drop actually happens and which is currently silent.

Same reason the "Breaking changes" note overstates it: nothing reaches this filter that wasn't already filtered upstream, so there's no observable change in what a key can discover.

### [nit] One list observation can emit two comparison samples

In `runWorkspaceListComparison`, a match records once, but a list differing in both directions records `legacy_allow_authzed_deny` and `legacy_deny_authzed_allow` as separate calls. So `formbricks_authzed_authorization_comparisons_total` isn't 1:1 with observations on this surface the way it is for scalar `can()`. The QA plan has the operator reading match rate off that counter — worth knowing the denominator inflates on mismatches, or scoping the query to `action="workspace.read"` when reading it.

---

Nothing above blocks the lookup work itself — the pagination guard, the cursor-stall check, the cohort-aware cross-tenant attribution and the `getSpicedbObjectType` centralisation all read correctly to me, and the one-operation-per-list claim holds up in the integration test.

xernobyl:
One follow-up to the second finding above, since I've just taken the `breaking-change` label off by hand and that won't hold.

The label isn't set manually — `.github/workflows/pr-label-sync.yml` derives it from the `## Breaking changes` section and re-syncs on `opened`, `edited`, `reopened`. `isDeclared` sees the `None` opener, then only clears the section if the remainder is at most six words and every one of them is a scope word (`docs`, `tests`, `internal`, `only`…). The current aside is ~35 words and includes `filters`, `grants`, `boundary`, so it evaluates as declared and the label comes straight back on the next body edit. That asymmetry is deliberate — the workflow errs toward labelling because a missed breaking change silently drops a migration-guide entry self-hosters need — so the durable fix is always the body, never the label.

Which means it folds into the same edit: the section should just read `None.`

That's also the accurate version. Per the finding above, nothing reaches the new filter that `api-key-auth.ts:60-63` hasn't already dropped, so there's no tenant-boundary or response-contract change to declare — and this section feeds both the release notes and the self-hoster migration guide, which is where a spurious entry would surface.

xernobyl (changes requested):
The lookup work itself reads well and I don't have a correctness objection to it — pagination guard, cursor-stall detection, cohort-aware cross-tenant attribution and the `getSpicedbObjectType` centralisation all hold up, and the one-operation-per-list claim is properly pinned by the integration test. All three items below are in this diff rather than adjacent to it, so I'd rather they land here than as follow-ups.

Detail and verification in my two comments above; the asks:

**1. Drop or relocate the `crossOrganizationGrantCount` plumbing.**
`listV3Workspaces` has exactly one caller (`modules/mcp/tools/workspaces.ts:34`), and that path's API-key auth goes through `authenticateApiKeyFromHeaders`, which already applies the identical org filter at `modules/api/lib/api-key-auth.ts:60-63` (ENG-1749, untouched on this branch). So the warn branch at `operations.ts:114` is unreachable, not merely unlikely.

The filter at `:62-66` is fine as defense in depth — keep it. It's the counter, the `TResolvedWorkspaceList` field and the test I'd like gone or moved. The test is the part I'd push hardest on: it hand-builds an authentication shape the request path cannot produce, so it reads as coverage of a tenant-boundary guard and isn't. Either delete both, or move the warning onto the ENG-1749 filter where the drop actually happens and is currently silent — the latter is what Matti was asking for, and I'd slightly prefer it.

**2. `## Breaking changes` → `None.`**
Both because the current claim isn't accurate (nothing reaches that filter unfiltered) and because it's what re-adds the `breaking-change` label on the next body edit. That section feeds the release notes and the self-hoster migration guide, and it becomes the squash commit message, so it's worth getting exactly right.

**3. Split the three smoke seeds into their own commit.**
The `organization#member` writes at `authzed-relationships-smoke.ts:173` and `:345` are repairing a break I introduced in #8869, not scaffolding for the new lookup commands — `docker/authzed-smoke.sh` has been failing on `epic/authzed` since Aug 14, invisibly, because Docker Build Validation only triggers on PRs into `main`.

Isolating them means we can cherry-pick the repair onto the base if this PR slips or gets reordered behind the other stacked ones. A line in the commit message saying they fix #8869's schema change would also stop a later reader from removing them as unrelated to workspace discovery. If you'd rather keep this PR whole, say so and I'll take the split on a separate branch — that break is mine to own either way.

None of this is a redesign; happy to re-review quickly once it's in.

pandeymangg (approved):

xernobyl (approved):
Approving — all three items from my change request are addressed, and two of them better than I asked.

**Dead telemetry.** You took the option I preferred: the warning moved to `api-key-auth.ts:68`, where the ENG-1749 filter actually drops the rows, and `operations.ts` lost the counter and the `TResolvedWorkspaceList` field entirely. Better than the ask on the test side too — the relocated warning is asserted to fire exactly once with the right message *and* to carry no identifiers in its payload, and `operations.test.ts` now asserts `getWorkspacesByIds` was called with `("org_1", ["w1", "w9"])`. That tests the organization scoping rather than a shape the request path cannot produce, which was my actual objection to the old case.

**Smoke-seed repair.** `cd99273fa` is exactly the shape I hoped for: first commit on the branch, smoke script only, cherry-pickable, with the reason in the message. I checked the contents against my own diagnosis rather than the description — alice and bob on `GRAPH_ORGANIZATION_ID`, the feedback user on `FEEDBACK_ORGANIZATION_ID`, which is precisely the three memberships `organization->product_member` needs after #8869.

**Breaking changes** reads `None.` and the label has cleared, so the body-driven rule is satisfied at the source rather than by removing the label.

The counter nit is handled more thoroughly than I raised it: the runbook now states that a both-directions list mismatch emits two samples, warns against reading the total as a request count, and adds a targeted query for the MCP list outcomes.

And `f18debbca` is a genuine improvement nobody asked for — the per-id `Promise.all(getWorkspace)` becomes one `getWorkspacesByIds(organizationId, ids)` with the tenant constraint in the `WHERE` clause and `validateInputs` on both arguments.

Two non-blocking nits, both cosmetic:

- `sameOrganizationWorkspaces` in `fetchApiKeyWorkspaces` is now provably dead — `getWorkspacesByIds` already constrains `organizationId`, so every returned row satisfies the predicate. Harmless, and mildly ironic given the previous round removed a dead counter for the same reason.
- `cd99273fa`'s message says #8869 "removed team-derived organization membership from the schema". It actually *added an intersection requiring* organization membership. Worth correcting only because that commit is meant to travel on its own.

One thing I did **not** verify, stated so it is not mistaken for checked: I have not executed `docker/authzed-smoke.sh` against this branch. The seeds match my diagnosis statically, but that workflow only triggers on PRs into `main`, so nothing has actually run it — the cherry-pick claim is reasoned, not proven. Happy to run it locally against a SpiceDB container if you would rather have that before the epic goes up.

PR #8877, opened 2026-08-17
feat(embedded-data): give SurveyEmbeddedData an explicit order column [ENG-2401]

pandeymangg:
## What & why

`SurveyEmbeddedData` had no ordinal, so the order a survey's fields render in came from their index in the legacy `survey.variables` / `survey.hiddenFields` JSON arrays — `inlineSurveyEmbeddedFields` built a rank map from `toDesiredEmbeddedFields(<legacy columns>)` and stable-sorted the joined rows by it.

That order is user-visible, and the biggest consumer isn't the CSV. `handle-integrations.ts` builds `elements` — the live spreadsheet column header array for Google Sheets / Airtable / Notion — from `getIngestedStorageKeys()` / `getComputedEmbeddedFields()`, both of which resolve through the rows. A wrong order writes response values under shifted headers, per response, with no error. The column has to exist before the legacy JSON can ever be dropped, and it can only be derived while that JSON is still authoritative.

- **Schema** — `order Int` on `SurveyEmbeddedData`, no default, plus `@@index([surveyId, order])`. No default because this is a fresh `CREATE TABLE`, so a writer that forgets the position is a type error rather than a row that silently sorts to the front. The index is deliberately not unique — a two-field swap would otherwise collide mid-transaction, since reconcile writes rows one at a time.
- **Write bridge** — new `toReorder` bucket on the reconcile plan. The link table had no update path at all before this: every write to it was a create or a delete, and the only update targeted `EmbeddedData`. The check sits *above* the ownership guard, because position belongs to the link, not to the definition — a shared field this survey doesn't own still moves when the fields around it change.
- **Read seam** — `orderBy: [{ order: "asc" }, { storageKey: "asc" }]`. Ordering now lives entirely in the query (a JS sort can't recover an order the query never imposed), so `inlineSurveyEmbeddedFields` drops its rank map and the three JSON guards that existed only to stop a malformed column failing an entire survey read. `storageKey` is the tiebreak because `order` alone isn't a total order, and `@@unique([surveyId, storageKey])` is what makes the pair one.

**"Moved" is `storedOrder !== desiredIndex`**, not a relative rearrangement. That makes every save self-healing — a link left at the wrong position by any route repairs itself on the next save. It matters because on a fresh database data migrations are baselined as applied *without running*, so `order` there comes only from reconcile.

**The column is added to the original `20260806000000` migration and to the ENG-1835 backfill**, rather than layered on as an `ALTER` plus a second data migration. Neither has reached `main`, so no deployed database has run them. This keeps order and row creation in one pass — they can't disagree — and means a survey the backfill skips has no rows at all rather than rows with no order.

> [!IMPORTANT]
> **Anyone with this epic checked out needs to reset their local database.** Editing an applied migration breaks its Prisma checksum, so `migrate deploy` will fail loudly on a database that already ran `20260806000000_add_embedded_data_tables`. Nothing to do for CI or any deployed environment — the migration has never run outside dev.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2401/give-surveyembeddeddata-an-explicit-order-column

## How this was tested

- `pnpm vitest run` in `apps/web` — 593 files, **7673 passed** ✅
- `pnpm test:integration` in `apps/web` — 26 files, **123 passed** ✅ (against real Postgres)
- `pnpm --filter @formbricks/database test` — **50 passed** ✅
- `pnpm turbo run typecheck` (web + database) — 12/12 ✅, `eslint` on the changed files clean ✅
- Migrations applied from scratch against a real database; verified `order integer not null` with no default and both indexes present via `\d "SurveyEmbeddedData"`.
- The hand-edited `migration.sql` was diffed against `prisma migrate diff --from-empty --to-schema` — byte-identical, including column position, reserved-word quoting and the generated index name.

Tests added:

- `reconcile.test.ts` — a **middle delete re-numbers the tail** (the only thing that moves a position in v1: there's no drag-to-reorder UI, the cards only append and filter, so `toReorder` almost always fires alongside `toUnlink` and a two-field swap wouldn't cover it); an insert pushes the rest down; a shared link still moves; a link on its way out isn't also reordered; all-zero links repair themselves.
- `embedded-data-reconcile.integration.test.ts` — the same three against real Postgres, asserting the stored `order` values.
- `embedded-data-read.integration.test.ts` — the malformed-legacy-JSON cases now assert **declared order regardless**, since the read no longer touches those columns at all.
- backfill `utils.test.ts` — links come back numbered by declaration, variables then hidden fields.
- `modules/survey/list/lib/survey.test.ts` — a copy to another workspace carries the source's positions.

One thing worth calling out: the required (no-default) `order` earned its keep during this work — it caught a `surveyEmbeddedData.create` I'd missed in a tenancy test, loudly, instead of letting that row default to 0 and jump to the front of the list.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a survey, add two hidden fields (`h1`, `h2`), then a variable (`score`), then a third hidden field (`h3`). Save. → Export responses to CSV: the columns read `score`, `h1`, `h2`, `h3`. Variables always come before hidden fields, regardless of the order they were added in — the two editor cards write to separate lists, so the order you typed them in was never stored.
- [ ] In the same survey, delete `h2` and save. → CSV columns read `score`, `h1`, `h3`, with no gap or reshuffle.
- [ ] Add a new variable to that survey and save. → It appears immediately after `score` and before the hidden fields; nothing else changes position.
- [ ] Open a survey that already had variables and hidden fields *before* this change. → Its export column order and the field order in every picker (recall `@`-menu, logic builder operands, follow-up recipients) are unchanged from before.
- [ ] Duplicate a survey with several fields. → The copy's export columns are in the same order as the original's.
- [ ] Copy a survey to a different workspace. → Same, in the target workspace.
- [ ] **Integrations (highest risk).** Configure a Google Sheets or Notion integration on a survey with hidden fields, submit a response, then add a hidden field, submit another. → Both responses land under the correct headers; no column is shifted or re-labelled.
- [ ] Response table, response summary, single-response card, and both notification emails list fields in the same order as the export.

**Preconditions / test data**

- A survey with at least two variables and two hidden fields, ideally one created before this branch so the backfill path is exercised.
- Some existing responses on it, so export and integration column alignment can be checked against real values.
- For the integration step: a connected Google Sheets, Airtable or Notion integration with hidden fields mapped.

**Risks & regressions**

- **Field ordering everywhere it's user-visible** — CSV/XLSX headers, response table columns, response filters, the recall picker, logic operand lists, the calculate widget, follow-up recipients, both notification-email templates, and the Notion field-mapping modal. Nothing there changed in code; they all read the same accessor, but they're what a wrong `order` would surface in.
- **Live integration column alignment** is the one that fails silently rather than visibly — a shifted header writes values into the wrong spreadsheet column with no error. Worth an explicit check rather than assuming.
- Survey duplication and cross-workspace copy, since both create fresh links.

- Two existing migrations are modified rather than added: `20260806000000_add_embedded_data_tables` (the column and index move into the original `CREATE TABLE`) and `20260812121944_backfill_embedded_data` (its links now carry a position).
- No action for CI or any deployed environment — neither migration has ever run outside local development.
- **Local dev databases on this epic must be reset**, or `migrate deploy` fails on the changed checksum.
- No env changes.

### DIFF
--- a/apps/web/app/api/v3/surveys/patch.test.ts
+++ b/apps/web/app/api/v3/surveys/patch.test.ts
@@ -41,6 +41,7 @@ vi.mock("@formbricks/database", () => {
       findMany: vi.fn(),
       deleteMany: vi.fn(),
       create: vi.fn(),
+      updateMany: vi.fn(),
     },
     embeddedData: {
       create: vi.fn(),
@@ -857,6 +858,7 @@ describe("patchV3Survey", () => {
           findMany: vi.fn().mockResolvedValue([]),
           deleteMany: vi.fn().mockResolvedValue({ count: 0 }),
           create: vi.fn().mockResolvedValue({}),
+          updateMany: vi.fn().mockResolvedValue({ count: 0 }),
         },
         embeddedData: {
           create: vi.fn().mockResolvedValue({ id: "ed_1" }),
--- a/apps/web/integration/embedded-data-backfill.integration.test.ts
+++ b/apps/web/integration/embedded-data-backfill.integration.test.ts
@@ -117,7 +117,7 @@ describe("Embedded Data backfill (real Postgres)", () => {
       data: { workspaceId, surveyId, name: "plan", source: "ingested" },
     });
     await prisma.surveyEmbeddedData.create({
-      data: { workspaceId, surveyId, embeddedDataId: existing.id, storageKey: "plan" },
+      data: { workspaceId, surveyId, embeddedDataId: existing.id, storageKey: "plan", order: 0 },
     });
 
     const stats = await backfillEmbeddedDataRows(prisma);
--- a/apps/web/integration/embedded-data-read.integration.test.ts
+++ b/apps/web/integration/embedded-data-read.integration.test.ts
@@ -79,13 +79,14 @@ describe("Embedded Data read seam (real Postgres)", () => {
     expect(await loadSurvey(surveyId)).not.toHaveProperty("embeddedDataLinks");
   });
 
-  test("the inlined order is the declared order, not the database's", async () => {
+  test("the inlined order is the declared order, not the storage keys'", async () => {
     const { surveyId } = await seedSurvey();
 
     const survey = await loadSurvey(surveyId);
 
     // Declared order is variables-then-hidden-fields, and `tier` is declared before `score` even
-    // though `clx…001` sorts before `clx…002`. This is CSV/XLSX header order and picker order.
+    // though `clx…001` sorts before `clx…002`. Storage key is only the tie-break, so this passing
+    // means the `order` column is what decided it — and this is CSV/XLSX header and picker order.
     expect(survey.embeddedFields?.map(({ link }) => link.storageKey)).toEqual([
       "clx000000000000000000002",
       "clx000000000000000000001",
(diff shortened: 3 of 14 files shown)


mattinannt (commented):
**Single-pass review (no Agent tool / no multi-agent fan-out — one context, all 8 angles worked manually).**

Traced the full ordering mechanism end-to-end: `planEmbeddedDataReconcile`'s `toReorder` derivation (index-in-`desired` vs stored `entry.order`, ownership guard correctly placed *after* the reorder push so shared/foreign links still reposition but don't get renamed/deleted), the `reconcileEmbeddedData` write sequence (unlink → update → reorder → create — no unique-constraint or ordering hazard between them), the backfill's `desired.entries()` → `order` assignment (no gaps since `reserved` throws rather than being skipped), and the read seam's `orderBy: [{order:"asc"},{storageKey:"asc"}]`.

Checked for drift risk explicitly:
- Every production `surveyEmbeddedData.create`/`createMany` call site (`reconcile.ts`, the backfill `migration.ts`) now supplies `order`; no call site was missed.
- Every survey select that embeds `embeddedDataLinks` (`survey/service.ts`, `survey/lib/survey.ts`, `survey/link/lib/data.ts`, `response-pipeline`, `api/v1/client/.../data.ts`) goes through the shared `selectSurveyEmbeddedDataLinks` constant rather than redeclaring its own `orderBy`, so there's no reader left sorting only by `storageKey`.
- `planEmbeddedDataReconcile` has a single call site (`reconcileEmbeddedData`), so the new `order`/`toReorder` shape can't be out of sync with a second caller.
- Migration SQL is well-formed and the reserved word `"order"` is quoted everywhere it's referenced in raw SQL/DDL.

No correctness bugs, removed-invariant gaps, or reuse/simplification/efficiency/altitude/CLAUDE.md issues survived scrutiny in this pass. This was a clean implementation — nothing to flag as an inline comment.

---

itsjavi (changes requested):
thanks lgtm, there is only a Sonar issue that needs fixing

itsjavi (approved):

PR #8878, opened 2026-08-17
feat(embedded-data): write rows from the payload rather than the persisted columns [ENG-2412]

pandeymangg:
> [!NOTE]
> Stacked on #8877 (ENG-2401). Base is that branch, so this diff shows only ENG-2412. Re-target to the epic once #8877 merges.

## What & why

`reconcileEmbeddedData` derived its target from the survey row it had **just written** — so `survey.variables` / `survey.hiddenFields` were the write source of truth and the `EmbeddedData` tables were a copy of them:

```
before:  payload → survey.variables → rows
after:   payload → rows              (columns still written, but nothing derives from them)
```

Reversing that is what M3's new write paths need — ENG-1844 (`setEmbeddedData()`) and ENG-1843 (URL params) both write Embedded Data from outside the editor, and would otherwise have to speak the legacy shape too. It is also what the eventual column drop (ENG-2404) is blocked on. And it is the cheapest moment to do it: nothing has reached production, so the blast radius is the epic branch.

**New `resolveDesiredEmbeddedFields`** — the payload patches the survey's current rows. The two legacy groups merge independently, so a payload carrying `variables` alone leaves the ingested rows exactly where they are, and a group the payload does not carry is left untouched rather than read as "delete these".

**Presence is `!== undefined`, not the `in` operator.** This one is load-bearing and easy to get wrong: every write seam builds a single object with both keys spelled out and lets Prisma ignore the undefined ones, so `"variables" in patch` is true even for a payload that never mentioned variables — and would clear its rows. There is a test pinning it.

**Reconcile reads its links ordered.** The carried-over group keeps its stored positions, and those become indexes in the desired list, so the rows have to arrive in the order they are stored in.

**`getSurveyEmbeddedFields` loses its legacy fallback.** Deleting a survey's rows now makes its fields disappear rather than reappear — the behaviour that made the previous model hard to reason about, and the thing this whole change is for. Audited before removing it: every reader gets its survey through `selectSurvey` or a select embedding the same `selectSurveyEmbeddedDataLinks` constant, including the ones that looked risky (`surveyLogic/utils.ts`'s `localSurvey`, the Notion integration modal's `selectedSurvey`).

## No behaviour change on any existing path

Worth stating explicitly, since M1 is the "nothing visible changes" milestone.

Every caller already merges over the loaded survey before calling in — `{ ...survey, singleUse }` in the summary action, `{ ...result.survey, ...surveyUpdate }` in the v1 management route, `mergeV3SurveyPatch(currentDocument, patch)` in v3. So both keys are always present, the payload and the persisted columns always agree, and `resolveDesiredEmbeddedFields` reduces to exactly what the old code computed. The duplicate-address guard is untouched and fires on the same inputs it always did.

The merge is therefore defensive today. It stops being defensive the moment ENG-1844 lands, which is the next milestone.

`hiddenFields.enabled` is untouched: it rides in the column as before, and `toDesiredEmbeddedFields` reads `fieldIds` alone by design. Audited it while here and recorded the findings on ENG-2404 — it is close to vestigial (schema default `false`, the card only ever writes `true`, never set back, one reader).

## Linear ticket

https://linear.app/formbricks/issue/ENG-2412/write-rows-from-the-payload-and-derive-the-legacy-columns-from-them

## How this was tested

- `pnpm vitest run` in `apps/web` — 593 files, **7679 passed** ✅
- `pnpm test:integration` in `apps/web` — 26 files, **123 passed** ✅ (real Postgres)
- `pnpm turbo run typecheck` (web + types + database) — 13/13 ✅, `eslint` on the changed files clean ✅

Tests added:

- `reconcile.test.ts` — six cases on the resolver: an omitted group carries over, an explicit `[]` clears, one group replaced leaves the other alone, `undefined` is an omission not a clear, and computed always precedes ingested (which is what `order` means).
- Integration — the reconcile and read seams re-pointed at the new signature; the read seam now asserts that a survey with no rows has no fields, where it previously asserted the fallback.
- Six test files had fixtures declaring `variables` / `hiddenFields` with no `embeddedFields`. They now carry the inlined rows, because that is what a real survey read returns — a fixture with columns alone describes a row nobody reads.

Not tested here: `hiddenFields.enabled` behaviour, since nothing in this change touches it.

<details>
<summary>Pre-existing failure, unrelated to this change</summary>

`modules/auth/actions.test.ts` has two failing cases (`TooManyRequestsError` from the real rate limiter). Verified by stashing this branch's changes and re-running — they fail without it too.
</details>

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a survey with variables and hidden fields, add one of each, save, reopen. → Both appear, in the same order as before; recall tokens, logic conditions and follow-up recipients still resolve.
- [ ] Rename the survey (nothing else) and save. → Its variables and hidden fields are all still there. This is the case that would break if the payload were read naively.
- [ ] Delete a hidden field, save, export responses to CSV. → The column is gone; the remaining columns keep their order.
- [ ] Update a survey through the **v1 management API** (`PUT /api/v1/management/surveys/:id`) with a body carrying `variables` and `hiddenFields`. → Both persist; a follow-up `GET` returns the same shape as before.
- [ ] Update one through the **v3 API** (`PATCH /api/v3/surveys/:id`) sending only `{ "name": "..." }`. → 200, and the survey's fields are unchanged.
- [ ] Send a v3 `PATCH` that puts the same name in both `variables` and `hiddenFields`. → 400, "Duplicate embedded data field" (unchanged from before this PR).
- [ ] Duplicate a survey, and copy one to another workspace. → Both carry the same fields in the same order.
- [ ] **The changed behaviour:** delete a survey's rows directly (`DELETE FROM "SurveyEmbeddedData" WHERE "surveyId" = '…'`), then reload it. → Its variables and hidden fields are now **gone** from the editor, pickers and exports. Before this PR they reappeared from the legacy columns. Save the survey and they come back, written from the editor's cards.

**Preconditions / test data**

- A survey with at least two variables and two hidden fields, with responses, so export and picker order can be checked.
- API credentials for the v1 and v3 checks.
- A follow-up configured with a hidden field as its recipient, to confirm `hiddenFields.enabled` still gates that label.

**Risks & regressions**

- Anything that reads a survey's Embedded Data: recall picker, logic builder operands, calculate widget, follow-up recipients, response table columns, response filters, summary, CSV/XLSX export, notification emails, and the Google Sheets / Airtable / Notion integrations.
- Survey creation from a template, duplication, and cross-workspace copy — all three write rows through this path.
- A survey read through a select that omits the Embedded Data join would now show no fields at all. Audited as none today; worth re-checking if a new survey select is added.

- none

### DIFF
--- a/apps/web/app/api/v3/surveys/patch.ts
+++ b/apps/web/app/api/v3/surveys/patch.ts
@@ -1,7 +1,6 @@
 import "server-only";
 import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
-import { toDesiredEmbeddedFields } from "@formbricks/types/embedded-data-mapping";
 import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
 import type { TSurvey } from "@formbricks/types/surveys/types";
 import { getActionClasses } from "@/lib/actionClass/service";
@@ -258,7 +257,7 @@ export async function executeV3SurveyPatch(params: {
         await reconcileEmbeddedData(tx, {
           surveyId: currentSurvey.id,
           workspaceId: currentSurvey.workspaceId,
-          desired: toDesiredEmbeddedFields(survey),
+          patch: { variables: document.variables, hiddenFields: document.hiddenFields },
         });
 
         return survey;
--- a/apps/web/app/api/v3/surveys/patch.ts
+++ b/apps/web/app/api/v3/surveys/patch.ts
@@ -244,9 +244,12 @@ export async function executeV3SurveyPatch(params: {
 
         const survey = await runSurveyUpdate(tx);
 
-        // Derived from the PERSISTED survey, not the patch document: a patch may omit `variables` or
-        // `hiddenFields` entirely, and reading the payload would see them as absent and delete every
-        // row. `workspaceId` comes from the stored survey, never the client (ENG-1749).
+        // ENG-2412: from the patch document, which is what makes the rows the write source of
+        // truth rather than a copy of the columns `data` just wrote. Safe on a partial patch for two
+        // reasons: `prepareV3SurveyPatchInput` merges the body over the current survey first, so both
+        // keys arrive populated; and `resolveDesiredEmbeddedFields` carries a group's current rows
+        // over untouched if its key is absent anyway. `workspaceId` comes from the stored survey,
+        // never the client (ENG-1749).
         //
         // NOTE for whoever moves the v3 serializer onto the tables (ENG-1853): `survey` was read
         // BEFORE this reconcile, so the `embeddedDataLinks` it carries — and the `embeddedFields`
--- a/apps/web/lib/embedded-data/reconcile.ts
+++ b/apps/web/lib/embedded-data/reconcile.ts
@@ -68,6 +68,12 @@ const SELECT_CURRENT_FIELDS = {
  *
  * The two groups are merged independently because they arrive independently: a payload carrying
  * `variables` alone must leave the ingested rows exactly where they are.
+ *
+ * `computed` and `ingested` are the only two sources carried over, and that is the whole set a row
+ * can have: `ZEmbeddedData` rejects `source: "reserved"` outright, because reserved fields are a
+ * code catalog projected at read time rather than anything stored. A `reserved` row is therefore
+ * unrepresentable through every write path — and if one ever appeared through raw SQL, this would
+ * drop it, which is the correct outcome for a row the schema says cannot exist.
  */
 export const resolveDesiredEmbeddedFields = (
   current: TDesiredEmbeddedField[],
--- a/apps/web/lib/embedded-data/reconcile.ts
+++ b/apps/web/lib/embedded-data/reconcile.ts
@@ -6,7 +6,11 @@ import {
   type TEmbeddedDataType,
   isLocalEmbeddedData,
 } from "@formbricks/types/embedded-data";
-import { type TDesiredEmbeddedField } from "@formbricks/types/embedded-data-mapping";
+import {
+  type TDesiredEmbeddedField,
+  type TLegacyEmbeddedFields,
+  toDesiredEmbeddedFields,
+} from "@formbricks/types/embedded-data-mapping";
 import { InvalidInputError } from "@formbricks/types/errors";
 
 /** One field a survey currently has, as the link plus the definition it points at. */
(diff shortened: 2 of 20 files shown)


apps/web/lib/embedded-data/reconcile.ts
```diff
@@ -49,6 +53,41 @@ const SELECT_CURRENT_FIELDS = {
...
+ * true even for a payload that never mentioned variables — and would clear them.
+ *
+ * The two groups are merged independently because they arrive independently: a payload carrying
+ * `variables` alone must leave the ingested rows exactly where they are.
+ */
+export const resolveDesiredEmbeddedFields = (
+  current: TDesiredEmbeddedField[],
+  patch: Partial<TLegacyEmbeddedFields>
+): TDesiredEmbeddedField[] => {
+  const carriedOver = (source: TEmbeddedDataSource): TDesiredEmbeddedField[] =>
```
mattinannt, line 82: `resolveDesiredEmbeddedFields` only carries forward rows with source `"computed"` or `"ingested"`. `TEmbeddedDataSource` also has a `"reserved"` value — once anything starts writing `reserved` rows (e.g. the planned `setEmbeddedData()`/URL-params work), an ordinary save through this merge would silently drop and unlink them, since they appear in neither carry-over branch. Not a regression from this PR (the old `toDesiredEmbeddedFields` had the same gap), but worth a `// TODO` or explicit note here so it isn't forgotten when that source starts being written.

---

apps/web/app/api/v3/surveys/patch.ts
mattinannt, line 1: Stale comment (not touched by this diff), lines 247-249:

> `// Derived from the PERSISTED survey, not the patch document: a patch may omit \`variables\` or \`hiddenFields\` entirely, and reading the payload would see them as absent and delete every row.`

This now contradicts the code two lines below it, which was changed by this PR to read `document.variables`/`document.hiddenFields` (the patch document) — the exact thing the comment says would be wrong. A future reader could conclude the code is buggy, or revert it, based on this stale note. Worth updating to describe the new `resolveDesiredEmbeddedFields` merge behavior instead.

---

apps/web/lib/survey/service.ts
mattinannt, line 1: Stale comment (not touched by this diff), lines 338-339, inside the destructuring of `embeddedFields` from `updatedSurvey`:

> `// The rows are written by \`reconcileEmbeddedData\` from the persisted legacy columns instead.`

This is now inaccurate: the `reconcileEmbeddedData` call further down in this same function (this PR's change) reads from `updatedSurvey.variables`/`updatedSurvey.hiddenFields` — the payload — not the persisted `currentSurvey`/`data`. A reader who only sees this earlier comment (rather than the updated one near the actual call) gets the pre-ENG-2412 mental model.

---

packages/types/embedded-data-resolver.ts
mattinannt, line 1: Stale comment (not touched by this diff), around lines 548-553, on `getSurveyEmbeddedFields`'s docstring:

> `... a saved survey's rows and declarations agree element for element, because every write path that persists those columns calls \`reconcileEmbeddedData\` with \`toDesiredEmbeddedFields(<the persisted survey>)\` in the same transaction... a \`reconcileEmbeddedData(\` grep is the audit...`

This is now inaccurate: after this PR, `reconcileEmbeddedData` takes a `patch: Partial<TLegacyEmbeddedFields>` derived from the payload (not a direct `toDesiredEmbeddedFields(survey)` call), and the merge with existing rows happens inside `resolveDesiredEmbeddedFields`. The stated audit method ("grep for `reconcileEmbeddedData(`") still finds call sites, but the claim about what each call passes is now wrong, and a reviewer relying on this description to verify the "exactly four write paths" invariant could be misled.

---

mattinannt (commented):
Automated review (`/pr-review`). Traced the core merge logic in `reconcile.ts` (`resolveDesiredEmbeddedFields`/`reconcileEmbeddedData`), the `getSurveyEmbeddedFields` fallback removal, and all call-site updates — the logic checks out against its own test suite and every call site found. Left 4 comments: 3 stale/self-contradicting doc comments left behind in functions this diff touched, and one un-generalized edge (the `"reserved"` embedded-data source isn't carried over by the new merge — pre-existing gap, not a regression, but worth a note given the diff rewrites this exact function's docs).

---

packages/types/embedded-data-resolver.ts
pandeymangg: fixed 🙏

apps/web/lib/survey/service.ts
pandeymangg: fixed 🙏

apps/web/app/api/v3/surveys/patch.ts
pandeymangg: fixed 🙏

itsjavi (approved):

itsjavi (approved):

PR #8885, opened 2026-08-17
fix(surveys): reject script-capable schemes in survey link URLs [ENG-2411]

Dhruwang:
## What & why

A CTA element's `buttonUrl` was validated with `ZUrl` (`z.url()`), which only checks that the value parses as a URL. `new URL()` accepts `javascript:`, `data:` and `vbscript:`, so `z.url().safeParse("javascript:alert(1)")` returns `success: true`. The renderer then hands that value straight to `window.open()` (`packages/survey-ui/src/components/elements/cta.tsx`) — stored XSS executing on the survey's own origin.

The inbound report describes this as reachable only through the draft skip-validation path (`updateSurveyInternal(survey, true)`), with a collaborator clicking the CTA in the editor preview. That path is real, but it is not the only one: because the scheme was never constrained, the payload also survives the **fully validated** path and `/api/v3/surveys`, which reuses `ZSurveyBlocks` from `@formbricks/types`. A published **link survey** can therefore carry the payload for every respondent, which is a wider blast radius than the editor preview.

Screenshot 2026-08-17 at 4 56 30 PM

Changes:

- `packages/types/common.ts` — new `isSafeLinkUrl()`: parses with `new URL()` and allowlists the scheme (`http`, `https`, `mailto`, `tel`). Parsing rather than a prefix test normalizes obfuscated schemes such as `java\tscript:`, which a `startsWith` check waves through. `mailto:`/`tel:` are kept because they are legitimate CTA targets that cannot execute script — a shipped template (`mailto:ceo@company.com`) relies on `mailto:`.
- `packages/types/surveys/elements.ts` — `ZSurveyCTAElement.buttonUrl` uses `isSafeLinkUrl` instead of `ZUrl`. This covers the editor's validated save path and every `/api/v3/surveys` write, since both parse through `ZSurveyBlocks`.
- `packages/survey-ui/src/lib/url.ts` (new) — the renderer-side twin of the same check. Kept local to the package rather than imported from `@formbricks/types` because `survey-ui` deliberately carries no dependency on it — the same reason `isSafeMediaUrl` lives there. Needed as defence in depth: the renderer is handed survey JSON from the API, so rows written before this validation landed, or through the draft path that skips validation, must not execute either.
- `packages/survey-ui/src/components/elements/cta.tsx` — gate `window.open()` on `isSafeLinkUrl`, and pass `noopener,noreferrer`. `noopener` is what removes the `window.opener` handle the reporter used to reach back into the authenticated page; it also makes `window.open` return `null`, so the now-pointless `?.focus()` is dropped.
- `packages/surveys/src/components/general/ending-card.tsx` — the sibling sink: `processAndRedirect` only checked `new URL(url)` was constructible before calling `window.top.location.replace(url)`, so a `javascript:` ending link executed the same way. Gated on the same check, applied *after* recall substitution so the final string is what gets validated.
- `apps/web/modules/email/components/preview-email-template.tsx` — the CTA button is rendered with `href={ctaElement.buttonUrl}`; skip the button entirely when the stored URL is unsafe.

## Linear ticket

Fixes ENG-2411

### Regression proof

| Test | Tier | On main | Here |
| --- | --- | --- | --- |
| `rejects a cta buttonUrl with the script-capable scheme %j` — 5 cases: `javascript:`, `JavaScript:`, `java\tscript:`, `data:`, `vbscript:` (`apps/web/app/api/v3/surveys/schemas.test.ts`) | red on main | ❌ all 5 accepted — `ZUrl = z.url()` parses every one of them (checked against the repo's zod 4.3.6) | ✅ rejected, issue on `blocks.0.elements.0.buttonUrl` |
| `accepts a cta buttonUrl with the safe scheme %j` — 3 cases: `https:`, `http://localhost`, `mailto:` | guard | ✅ | ✅ |
| `rejects %j` — 12 cases over `isSafeLinkUrl` (`packages/survey-ui/src/lib/url.test.ts`) | mutation | ❌ all 12 accepted | ✅ |
| `accepts %j` — 6 cases incl. `mailto:`, `tel:`, whitespace-padded | guard | ✅ | ✅ |

**How the tiers were proven.** *Red on main:* `schemas.test.ts` runs unchanged against `main`, where
`ZSurveyCTAElement.buttonUrl` is validated with `ZUrl` (`z.url()`) — so the payload parses as valid and
the five rejection cases fail.

```bash
git stash push -- packages/types/surveys/elements.ts packages/types/common.ts
pnpm test apps/web/app/api/v3/surveys/schemas.test.ts                    # 5 failed
git stash pop && pnpm test apps/web/app/api/v3/surveys/schemas.test.ts   # 8 passed
```

*Mutation:* `packages/survey-ui/src/lib/url.ts` is a new file, so `url.test.ts` cannot run against
`main` at all — it fails on a missing import, which proves nothing about the bug. Proven instead by
relaxing the allowlist in `url.ts` to accept anything `new URL()` parses, i.e. the pre-fix behaviour:
all 12 rejection cases go red.

*Guard:* passes either way. Pins the schemes that must keep working — `mailto:` is used by a shipped
template — rather than evidencing the bug.

### Coverage

| Behaviour | How | Outcome |
| --- | --- | --- |
| Script schemes refused at the API boundary | unit — red on main | `400`, issue on `blocks.0.elements.0.buttonUrl` |
| Script schemes refused in the renderer | unit (mutation) + manual | `window.open` never called; no alert in editor preview |
| Obfuscated scheme (`java\tscript:`) normalized, not prefix-matched | unit — red on main | rejected |
| `mailto:` / `tel:` still accepted | unit + manual | mail client and dialler open |
| Ending-card redirect gated after recall substitution | manual | no navigation, console logs the refusal |
| Email preview drops an unsafe CTA button | manual | question renders without the button |
| `window.opener` removed on a valid CTA | manual | `null` in the opened tab |

Full suite: `pnpm test` → 26/26 tasks, apps/web 609 files / 8021 tests passed; survey-ui 108, surveys
668. `pnpm lint` 21/21 (0 errors, 46 pre-existing warnings, none in touched files), `pnpm typecheck`
27/27, `pnpm format:check` clean, `pnpm install` lockfile unchanged.

Per `AGENTS.md` (no component unit tests for `.tsx`), the three renderer sinks are not unit tested.

No UI change: the fix removes a link target that should never have been accepted. The CTA button
renders identically for every valid URL; only its `window.open` call gained `noopener,noreferrer`.

### Open gaps

- **The two copies of `isSafeLinkUrl` are not pinned to each other.** `packages/types/common.ts` and
  `packages/survey-ui/src/lib/url.ts` carry identical bodies for a deliberate reason (survey-ui has no
  dependency on types), but only the survey-ui copy has direct tests, and `packages/types` has no test
  runner at all — no `*.test.ts`, no vitest config. The types-side helper is covered only indirectly,
  through `schemas.test.ts`. If one copy is later relaxed, nothing goes red for the other.
- **No automated coverage of the three renderer sinks.** No Playwright spec touches `buttonUrl` or the
  redirect ending, so the editor-preview, ending-card and email paths rest on the manual runs above.
- **Existing rows with unsafe schemes were not surveyed.** The fix refuses them going forward and the
  renderer refuses to navigate to them, but no query was run against production data to see whether any
  already exist.

## Breaking changes

| CTA element `buttonUrl` scheme validation | Any value `new URL()` could parse was accepted, including `javascript:`, `data:`, `ftp:` | Only `http:`, `https:`, `mailto:` and `tel:` are accepted; anything else is a `400` with the issue on the element's `buttonUrl` | API v3 (`POST`/`PATCH /api/v3/surveys`) and API v1 management consumers that set a CTA button URL with a non-web scheme | Send an `http(s)`, `mailto:` or `tel:` URL. Existing surveys are untouched until their next write; if one stores an unsupported scheme, the CTA button now renders without opening it |

## QA / Test Plan

**How to test**

- [ ] Create a survey, add a **CTA** element, enable the external-button toggle, set Button URL to `https://formbricks.com/docs` and a button label → survey saves and publishes as before; opening the survey and clicking the CTA opens the docs in a new tab.
- [ ] In the same CTA, replace the Button URL with `javascript:alert(document.domain)` and try to publish the survey → publishing is refused with a validation error naming the button URL. Before this change it published successfully.
- [ ] Save that same `javascript:` URL while the survey is still a **draft** (drafts are saved without validation, so this still persists), then open the survey editor **preview** and click the CTA button → nothing happens; no alert dialog appears and no new tab opens. Before this change the alert fired inside the authenticated editor.
- [ ] With a valid `https://` CTA, click the button in the live survey and inspect the opened tab's console: `window.opener` is `null` → the opened page cannot reach back into the survey page.
- [ ] Set a CTA Button URL to `mailto:support@example.com`, publish, and click the button → the mail client opens. Repeat with `tel:+123456789` → the dialler opens. Both must keep working.
- [ ] Add an **ending card** of type "Redirect to URL", save it as a draft with `javascript:alert(1)` as the URL, then complete the survey → the survey does not navigate and no alert appears (a "Refusing to redirect to an unsafe URL" line is logged to the browser console). With a normal `https://` URL, the redirect still happens on completion.
- [ ] Send the "email embed" survey preview for a survey whose CTA has a valid `https://` URL → the email shows the CTA button and it links to that URL. For a draft-stored `javascript:` URL, the email renders the question without the CTA button rather than a clickable payload.
- [ ] `POST /api/v3/surveys` with a `cta` element whose `buttonUrl` is `data:text/html,<script>alert(1)</script>` → `400`, with the invalid param pointing at `blocks.0.elements.0.buttonUrl`. The same request with `https://example.com` → `201`.

**Preconditions / test data**

- Two accounts in one workspace with `readWrite` (organization) or Editor (project) access, to reproduce the reporter's collaborator-to-collaborator scenario on a shared draft.
- External CTA URLs are plan-gated (`checkExternalUrlsPermission`): use an organization entitled to external URLs, or the enterprise licence/Cloud plan that grants it, otherwise the save is refused for billing reasons before validation is reached.
- Works on both Cloud and self-hosted; no feature flag involved.
- An API key with survey write scope for the `/api/v3/surveys` step.

**Risks & regressions**

- Narrowed input validation is the main risk: any pre-existing survey whose CTA URL uses a scheme outside `http`/`https`/`mailto`/`tel` will now fail validation on its next non-draft save. Worth a check for stored `buttonUrl` values with unusual schemes before release. The `mailto:ceo@company.com` shipped template was verified as still valid.
- `window.open` gained `noopener`, so it now returns `null`. Nothing else in the CTA path used that return value, but the button's "opens a new tab and focuses it" behavior is worth a click-through — the tab still opens, the browser simply owns focus.
- The ending-card guard runs after recall substitution, so a redirect URL built from recall/hidden-field values is re-checked at click time. Verify a legitimate ending redirect with a recall value in the query string still navigates.
- Both `packages/survey-ui` and `packages/surveys` changed, and `@formbricks/surveys` ships as a prebuilt bundle served from `apps/web/public/js/`. QA needs a build that includes the rebuilt bundle and a hard refresh, otherwise a cached `surveys.umd.cjs` will still show the old behavior.

- none

### DIFF
--- a/apps/web/app/api/v3/surveys/schemas.test.ts
+++ b/apps/web/app/api/v3/surveys/schemas.test.ts
@@ -219,6 +219,51 @@ describe("ZV3CreateSurveyBody", () => {
     });
   });
 
+  // Regression (ENG-2411): `buttonUrl` used to be validated with `z.url()`, which accepts any value
+  // `new URL()` parses — `javascript:` included — so a script-capable scheme reached the renderer, where
+  // it is handed to `window.open()`.
+  const ctaCreateBody = (buttonUrl: string) => ({
+    ...validCreateBody,
+    blocks: [
+      {
+        ...validCreateBody.blocks[0],
+        elements: [
+          {
+            id: "cta",
+            type: "cta",
+            headline: { "en-US": "Read the docs" },
+            required: false,
+            buttonExternal: true,
+            ctaButtonLabel: { "en-US": "Open" },
+            buttonUrl,
+          },
+        ],
+      },
+    ],
+  });
+
+  test.each([
+    "javascript:alert(document.domain)",
+    "JavaScript:alert(1)",
+    "java\tscript:alert(1)",
+    "data:text/html,<script>alert(1)</script>",
+    "vbscript:msgbox(1)",
+  ])("rejects a cta buttonUrl with the script-capable scheme %j", (buttonUrl) => {
+    const result = ZV3CreateSurveyBody.safeParse(ctaCreateBody(buttonUrl));
+
+    expect(result.success).toBe(false);
+    expect(result.error?.issues.map((issue) => issue.path.join("."))).toContain(
+      "blocks.0.elements.0.buttonUrl"
+    );
+  });
+
+  test.each(["https://formbricks.com/docs", "http://localhost:3000/docs", "mailto:support@example.com"])(
+    "accepts a cta buttonUrl with the safe scheme %j",
+    (buttonUrl) => {
+      expect(ZV3CreateSurveyBody.safeParse(ctaCreateBody(buttonUrl)).success).toBe(true);
+    }
+  );
+
   test("rejects choice fields that do not belong to the selected element type", () => {
     const result = ZV3CreateSurveyBody.safeParse({
       ...validCreateBody,
--- a/apps/web/modules/email/components/preview-email-template.tsx
+++ b/apps/web/modules/email/components/preview-email-template.tsx
@@ -14,6 +14,7 @@ import {
   Text,
   render,
 } from "@formbricks/email";
+import { isSafeLinkUrl } from "@formbricks/types/common";
 import {
   type TSurveyAddressElement,
   type TSurveyCTAElement,
@@ -448,27 +449,33 @@ export async function PreviewEmailTemplate({
       const ctaElement = firstQuestion as TSurveyCTAElement;
       return (
         <PreviewQuestionCard headline={headline} styleTokens={styleTokens} subheader={subheader} t={t}>
-          {ctaElement.buttonExternal && ctaElement.ctaButtonLabel && ctaElement.buttonUrl && (
-            <Section className="mt-4 text-left" style={{ textAlign: "left" }}>
-              <EmailButton
-                className={SECONDARY_BUTTON_CLASSNAME}
-                style={getPrimaryButtonStyle(styleTokens)}
-                href={ctaElement.buttonUrl}
-                target={PREVIEW_LINK_TARGET}>
-                {getLocalizedValue(ctaElement.ctaButtonLabel, defaultLanguageCode)}
-                <SquareArrowOutUpRightIcon
-                  color={styleTokens.buttonTextColor}
-                  size={16}
-                  strokeWidth={2}
-                  style={{
-                    display: "inline-block",
-                    marginLeft: "8px",
-                    verticalAlign: "text-bottom",
-                  }}
-                />
-              </EmailButton>
-            </Section>
-          )}
+          {/* `isSafeLinkUrl`: the button URL is an editable survey field, and drafts are persisted
+              without schema validation, so a stored value can carry a `javascript:` scheme. Rendering
+              it into an `href` would ship that payload inside the email preview. */}
+          {ctaElement.buttonExternal &&
+            ctaElement.ctaButtonLabel &&
+            ctaElement.buttonUrl &&
+            isSafeLinkUrl(ctaElement.buttonUrl) && (
+              <Section className="mt-4 text-left" style={{ textAlign: "left" }}>
+                <EmailButton
+                  className={SECONDARY_BUTTON_CLASSNAME}
+                  style={getPrimaryButtonStyle(styleTokens)}
+                  href={ctaElement.buttonUrl}
+                  target={PREVIEW_LINK_TARGET}>
+                  {getLocalizedValue(ctaElement.ctaButtonLabel, defaultLanguageCode)}
+                  <SquareArrowOutUpRightIcon
+                    color={styleTokens.buttonTextColor}
+                    size={16}
+                    strokeWidth={2}
+                    style={{
+                      display: "inline-block",
+                      marginLeft: "8px",
+                      verticalAlign: "text-bottom",
+                    }}
+                  />
+                </EmailButton>
+              </Section>
+            )}
         </PreviewQuestionCard>
       );
     }
--- a/packages/survey-ui/src/components/elements/cta.tsx
+++ b/packages/survey-ui/src/components/elements/cta.tsx
@@ -3,6 +3,7 @@ import * as React from "react";
 import { Button } from "@/components/general/button";
 import { ElementError, getElementErrorAria } from "@/components/general/element-error";
 import { ElementHeader } from "@/components/general/element-header";
+import { isSafeLinkUrl } from "@/lib/url";
 
 /**
  * Props for the CTA (Call to Action) element component
@@ -66,8 +67,12 @@ function CTA({
     if (disabled) return;
     onClick();
 
-    if (buttonExternal && buttonUrl) {
-      window.open(buttonUrl, "_blank")?.focus();
+    // `isSafeLinkUrl` gate: `buttonUrl` is an editable survey field, and a `javascript:` value reaching
+    // `window.open()` executes on the survey's own origin. `noopener` is what keeps the opened tab from
+    // reaching back through `window.opener` into the authenticated page that opened it; it also makes
+    // `window.open` return null, so there is no handle left to `focus()`.
+    if (buttonExternal && buttonUrl && isSafeLinkUrl(buttonUrl)) {
+      window.open(buttonUrl, "_blank", "noopener,noreferrer");
     }
   };
 
--- a/packages/survey-ui/src/index.ts
+++ b/packages/survey-ui/src/index.ts
@@ -37,6 +37,7 @@ export { DateElement, type DateElementProps } from "@/components/elements/date";
 export { getDateFnsLocale } from "@/lib/locale";
 export { sanitizeSurveyHtml } from "@/lib/utils";
 export { isSafeMediaUrl } from "@/lib/video";
+export { isSafeLinkUrl } from "@/lib/url";
 export {
   PictureSelect,
   type PictureSelectProps,
--- a/packages/survey-ui/src/lib/url.test.ts
+++ b/packages/survey-ui/src/lib/url.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, test } from "vitest";
+import { isSafeLinkUrl } from "./url";
+
+describe("isSafeLinkUrl", () => {
+  // Regression (ENG-2411): a CTA `buttonUrl` comes from an editable survey field and used to be handed
+  // straight to `window.open()`, where a `javascript:` URL executes on the survey's own origin.
+  test.each([
+    "javascript:alert(document.domain)",
+    "JavaScript:alert(1)",
+    // Obfuscated scheme: browsers strip the control character, so a `startsWith("javascript:")` check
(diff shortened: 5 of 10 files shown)


mattinannt (commented):
Reviewed the fix for the CTA `buttonUrl` scheme-injection XSS (ENG-2411). The core patch is solid: `isSafeLinkUrl` correctly allowlists by parsed `protocol` (not a prefix test, so `java\tscript:`/`JavaScript:` normalize and get caught), and I traced every `buttonUrl` sink I could find — the schema (`ZSurveyCTAElement`), the `survey-ui` `Cta` component's `window.open`, the email preview's `href`, and the ending-card's `location.replace` (gated post-recall-substitution, correctly). I also checked the two *other* pre-existing CTA validation paths in `packages/types/surveys/types.ts` (legacy `questions` array and the `ZSurvey`-level element check) — both already used `getZSafeUrl`/prefix-based validation that rejects `javascript:`, so they were never part of the vulnerability and don't need touching here.

**One real gap found**, left as an inline comment on `packages/surveys/src/components/elements/cta-element.tsx` (couldn't anchor it there directly — that file isn't part of this PR's diff, so GitHub won't let me attach a line comment to it):

`CTAElement.handleClick` calls `onOpenExternalURL(element.buttonUrl)` with the **raw, unvalidated** URL, unconditionally — no `isSafeLinkUrl` gate. This is the exact same external-callback pattern this PR *does* gate in `ending-card.tsx` (`if (url && isSafeLinkUrl(url)) { onOpenExternalURL?.(url) ... }`), but here it's missing:

```js
// packages/surveys/src/components/elements/cta-element.tsx:40-45
if (element.buttonExternal && element.buttonUrl) {
  if (onOpenExternalURL) {
    onOpenExternalURL(element.buttonUrl);
  }
  // Note: The survey-ui CTA component handles external URL opening itself
}
```

That trailing comment is misleading after this PR: `Cta.handleButtonClick` (survey-ui) calls `onClick()` (i.e. this `handleClick`) *and, independently*, does its own now-gated `window.open()`. Both fire when `onOpenExternalURL` is supplied by a host — the in-package `window.open` is safe, but the raw URL still reaches the external callback unfiltered.

`onOpenExternalURL` is a public prop on `SurveyContainerProps` (`packages/types/formbricks-surveys.ts`), threaded through but not implemented anywhere in this repo — it exists for embedders/host apps to hook into (e.g. opening links via a native bridge). There's no documented contract that the URL handed to it is pre-validated. The PR's own stated rationale for defense-in-depth on `buttonUrl` — *"the renderer is handed survey JSON from the API… rows written before this validation landed, or through the draft path that skips validation, must not execute either"* — applies identically here. Any embedder whose `onOpenExternalURL` does something as ordinary as `(url) => window.open(url)` gets the same `javascript:` payload this PR closes elsewhere, through a side door.

Suggested fix, mirroring `ending-card.tsx`:
```js
if (element.buttonExternal && element.buttonUrl && isSafeLinkUrl(element.buttonUrl)) {
  onOpenExternalURL?.(element.buttonUrl);
}
```
`isSafeLinkUrl` is already re-exported from `@formbricks/survey-ui`, which this file's sibling `ending-card.tsx` already imports it from — a one-line import away.

Everything else (schema tests, `isSafeLinkUrl` unit tests, the `noopener,noreferrer` addition, the duplicate-helper and no-test-runner gaps the PR description already discloses) looks correct and honestly scoped. This is the one sink I'd want closed before merge, since it's the same vulnerability class the PR exists to fix.

---

xernobyl (dismissed):
Verified the fix closes the reported CTA `buttonUrl` XSS (`javascript:`/`data:`/`vbscript:`) across all three live renderer sinks — survey-ui `Cta` (`window.open` + `noopener,noreferrer`), `ending-card.tsx`'s redirect (checked after recall substitution), and the email preview (button omitted when unsafe) — plus the `ZSurveyCTAElement` schema backing `/api/v3/surveys`. CI green, CodeRabbit/Sonar clean.

Non-blocking: `packages/surveys/src/components/elements/cta-element.tsx:41-42` forwards the raw `buttonUrl` through `onOpenExternalURL` without the `isSafeLinkUrl` gate used everywhere else. It's inert today — `onOpenExternalURL` isn't wired to any implementation in this repo — but worth gating for consistency in case it's ever wired up later.

Dhruwang:
Confirmed and fixed in a7d6b4d135 — thanks, this was a real miss and the reasoning is exactly right.

Traced it before changing anything: `Cta.handleButtonClick` calls `onClick()` (this `handleClick`) and then does its own `isSafeLinkUrl`-gated `window.open` — two independent paths, and only one of them was gated. When a host supplies `onOpenExternalURL`, the raw value still reached it. `ending-card.tsx` in this same PR gates the identical callback, so the inconsistency was ours, not a deliberate scoping call.

Applied the suggested shape, with the refusal logged the way the ending card logs its own:

```ts
if (element.buttonExternal && element.buttonUrl && onOpenExternalURL) {
  if (isSafeLinkUrl(element.buttonUrl)) {
    onOpenExternalURL(element.buttonUrl);
  } else {
    console.error("Refusing to open an unsafe CTA button URL");
  }
}
```

`isSafeLinkUrl` comes from `@formbricks/survey-ui` (already re-exported at `index.ts:40`), same import the sibling ending card uses. The stale "survey-ui handles external URL opening itself" comment is replaced with why both paths need the check.

No unit test: `.tsx`, so `AGENTS.md` rules it out and there is no non-component seam here — the gated helper itself is covered by the 18 cases in `url.test.ts`. That leaves this sink in the same manual-only bucket as the other two renderer paths, which the description's Open gaps already names. `packages/surveys` 27 files / 668 tests and `survey-ui` 108 pass; lint clean.

On the two other CTA validation paths in `packages/types/surveys/types.ts` — agreed, `getZSafeUrl` already rejects `javascript:` there, so they were never part of this.

xernobyl (approved):
Confirmed: `a7d6b4d` gates `onOpenExternalURL` on `isSafeLinkUrl` (matching the ending-card pattern) and closes the P3 from the earlier pass. Diff otherwise unchanged from what I already verified. CI green, no other findings.

PR #8886, opened 2026-08-17
fix(security): scope response file deletion to the owning workspace [ENG-2291]

mattinannt:
Ref ENG-2291 — [[Security] Cross-tenant storage object deletion via planted response data key](https://linear.app/formbricks/issue/ENG-2291/security-cross-tenant-storage-object-deletion-via-planted-response)

> [!IMPORTANT]
> Showcase duplicate of PR 8861 — do not merge. Same diff, kept open to test the PR template.

## What & why

Deleting a response could delete a storage object belonging to a **different workspace**. The storage
key for each file was rebuilt from the id in the **file URL**, and nothing on the delete path tied
that file to the survey's own workspace. Write-time validation does pin uploaded URLs to the
workspace, but only for response keys matching a file-upload element present at write time — so a URL
stored outside that check could later be treated as a real answer and deleted.

**Fix.** New helper `deleteResponseFileUrls` re-resolves each URL's storage id against the owning
survey's workspace and skips anything that resolves outside it. Both deletion sites — legacy
`lib/response/service.ts` (UI and v1) and the v2 management path — now go through it, so no
caller-supplied storage id is trusted. Two follow-ups from review: a failed storage delete is logged
instead of counted as a success, and workspace lookups are cached per storage id.

<details>
<summary>Why this wasn't caught earlier</summary>

Restores the delete-side scoping written in abandoned PR #8667 — closed when that large review branch
was split by severity, and this hunk was dropped on the assumption that the write-side binding
already covered it. Exploit path and severity stay in the Linear ticket, per `SECURITY.md` §III.C.

</details>

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Planted cross-workspace URL refused, v2 management path | `unit (red on main)` — `refuse to delete a file whose storage id belongs to a different workspace` | 1 failed / 4 passed against `main`'s `utils.ts`: the foreign file was deleted before the fix |
| Planted cross-workspace URL refused, shared helper | `unit (mutation)` — `refuses to delete a planted URL that resolves to a different workspace`, `refuses when the storage id resolves to no workspace`, `in a mixed batch, deletes the owned file and refuses the foreign one` | 3 failed / 6 passed with the guard removed; all 9 pass with it |
| Own-workspace file deleted — UI, v1 and v2 paths | `manual`, local instance with S3-compatible storage | response gone and object gone from storage |
| Legacy env-id prefix mapping back to the same workspace | `unit (guard)`, `findWorkspaceByIdOrLegacyEnvId` mocked | file still deleted |
| Bad input — unparseable URL, unresolvable id, no workspace id | `unit (guard)`, both suites | skipped and logged, owned files still deleted |
| Failed storage delete (`deleteFile` returns an error result) | `unit (guard)` — `logs when deleteFile returns a failure result` | logged instead of counted as a success |
| Several files sharing one storage id | `unit (guard)` — `resolves each distinct storage id only once for multiple files under it` | one workspace lookup per distinct id |
| Response with no file-upload answers | `unit (guard)` + `manual` | no storage call at all |

14 cases total across `apps/web/modules/storage/lib/delete-response-files.test.ts` (9, new) and
`apps/web/modules/api/v2/management/responses/[responseId]/lib/tests/utils.test.ts` (5, reworked) —
**1 red on main, 3 mutation, 10 guards**. Only one case can fail against `main`, because the other
suite tests a file that does not exist there; that is why the helper's own cases are tiered as
mutation rather than presented as proof the bug existed.

Rerun either tier:

```bash
# red on main — 1 failed | 4 passed
git checkout origin/main -- "apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts"
cd apps/web && ../../node_modules/.bin/vitest run "modules/api/v2/management/responses/[responseId]/lib/tests/utils.test.ts"
git checkout HEAD -- "apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts"

# mutation — delete the workspace check at delete-response-files.ts:55-62, then: 3 failed | 6 passed
cd apps/web && ../../node_modules/.bin/vitest run modules/storage/lib/delete-response-files.test.ts
```

Workspace packages must be built first (`pnpm turbo run build --filter=@formbricks/web^...`), or the
suites fail to resolve `@formbricks/logger` rather than reporting test results.

**Open gaps**

- The cross-workspace refusal has not been reproduced against live storage. It is proven at the
  helper boundary (`deleteFile` is never reached), but an end-to-end run needs two workspaces and a
  file URL planted directly into response data.
- The legacy environment-id prefix path has not been exercised against real keys — a fresh instance
  has no pre-workspace-rename objects, so that branch is proven only with the workspace lookup mocked,
  and it sits in the guard tier for that reason.

**Risks**

- Own-workspace cleanup must keep working; only files resolving to another workspace are refused.
- Legacy env-prefixed uploads must still be cleaned up on delete (see gap 2).
- A normal delete should log nothing — failed deletes are now logged rather than silent.
- Shared surface for cross-PR interplay: both delete paths now depend on `deleteResponseFileUrls`,
  `findWorkspaceByIdOrLegacyEnvId`, and the storage key format.

<details>
<summary>Call-site audit and manual-run setup</summary>

**Call sites.** Response-file deletion exists in exactly two places, both routed through the new
helper: `apps/web/lib/response/service.ts:645` (UI and v1 management delete) and
`apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts:24` (v2 management delete).
The only other `deleteFile(` caller is
`apps/web/app/storage/[workspaceId]/[accessType]/[...filePath]/route.ts:142`, the direct file-delete
endpoint, which takes its workspace from the route path and is untouched by this PR.

**Manual runs.** UI delete from the Responses table; `DELETE /api/v1/management/responses/{id}`;
`DELETE /api/v2/management/responses/{id}`; and a response with no file-upload answers. In each of
the first three the response and its stored object were gone; in the fourth nothing else was touched.

**To reproduce.** A published file-upload survey with a response carrying an uploaded file, a
management API key for the API paths, and storage configured (S3 on Cloud, RustFS/MinIO on
self-host).

</details>

---

> [!NOTE]
> **AI model used** — `…`, reasoning effort `high`.

### DIFF
--- a/apps/web/modules/storage/lib/delete-response-files.ts
+++ b/apps/web/modules/storage/lib/delete-response-files.ts
@@ -31,6 +31,18 @@ export const deleteResponseFileUrls = async (
     return;
   }
 
+  // Several files in one response usually share a storage id (same survey/workspace prefix). Cache the
+  // resolution promise per id so the batch does one lookup per distinct id instead of one per file.
+  const workspaceByStorageId = new Map<string, ReturnType<typeof findWorkspaceByIdOrLegacyEnvId>>();
+  const resolveStorageWorkspace = (storageId: string) => {
+    const cached = workspaceByStorageId.get(storageId);
+    if (cached) return cached;
+
+    const pending = findWorkspaceByIdOrLegacyEnvId(storageId);
+    workspaceByStorageId.set(storageId, pending);
+    return pending;
+  };
+
   await Promise.all(
     fileUrls.map(async (fileUrl) => {
       try {
@@ -40,7 +52,7 @@ export const deleteResponseFileUrls = async (
           throw new Error(`Invalid storage file URL: ${fileUrl}`);
         }
 
-        const storageWorkspace = await findWorkspaceByIdOrLegacyEnvId(storageFile.storageId);
+        const storageWorkspace = await resolveStorageWorkspace(storageFile.storageId);
         if (storageWorkspace?.id !== surveyWorkspaceId) {
           logger.error(
             { fileUrl, surveyWorkspaceId, storageId: storageFile.storageId },
@@ -49,12 +61,20 @@ export const deleteResponseFileUrls = async (
           return;
         }
 
-        await deleteFile(
+        // deleteFile returns an error result (it does not throw) on S3 failures, so a discarded result
+        // would treat a failed deletion as a success and leave the object behind unlogged.
+        const result = await deleteFile(
           storageFile.storageId,
           storageFile.accessType,
           storageFile.fileName,
           surveyWorkspaceId
         );
+        if (!result.ok) {
+          logger.error(
+            { fileUrl, surveyWorkspaceId, error: result.error },
+            "Failed to delete a response file from storage"
+          );
+        }
       } catch (error) {
         logger.error({ error, fileUrl }, "Failed to delete file");
       }
--- a/apps/web/modules/storage/lib/delete-response-files.ts
+++ b/apps/web/modules/storage/lib/delete-response-files.ts
@@ -1,0 +1,63 @@
+import "server-only";
+import { logger } from "@formbricks/logger";
+import { findWorkspaceByIdOrLegacyEnvId } from "@/lib/utils/resolve-client-id";
+import { deleteFile } from "@/modules/storage/service";
+import { parseStorageFileUrl } from "@/modules/storage/utils";
+
+/**
+ * Deletes the storage objects a response's file-upload answers point at, restricted to the survey's own
+ * workspace.
+ *
+ * The file URLs come out of `response.data`, i.e. from whoever wrote the response, and the S3 key is
+ * built from the id in the *URL* rather than the survey's workspace. Write-time validation
+ * (`validateClientFileUploads` -> `isScopedPrivateUploadUrl`) does pin uploaded URLs to the survey's
+ * workspace, but it only inspects keys that match a file-upload element that exists *at write time*. A
+ * caller can therefore plant a foreign URL under a key that is not yet an element, then edit the survey
+ * to turn that key into a file-upload element — a time-of-check/time-of-use gap that makes the planted,
+ * unvalidated URL look like a real answer at delete time.
+ *
+ * So the delete side cannot trust the URL's id. Re-resolve each URL's storage id here and drop anything
+ * that does not belong to this survey's workspace, regardless of which write path produced the data. The
+ * id may be a workspace id or a legacy environment id (older uploads were prefixed with the environment
+ * id), which is why it goes through `findWorkspaceByIdOrLegacyEnvId` rather than a plain string compare.
+ */
+export const deleteResponseFileUrls = async (
+  fileUrls: string[],
+  surveyWorkspaceId: string | undefined
+): Promise<void> => {
+  if (!surveyWorkspaceId) {
+    // Without the owning workspace there is nothing to authorize against, so delete nothing.
+    logger.error({ fileCount: fileUrls.length }, "Skipping response file deletion: no workspace id given");
+    return;
+  }
+
+  await Promise.all(
+    fileUrls.map(async (fileUrl) => {
+      try {
+        const storageFile = parseStorageFileUrl(fileUrl);
+
+        if (!storageFile) {
+          throw new Error(`Invalid storage file URL: ${fileUrl}`);
+        }
+
+        const storageWorkspace = await findWorkspaceByIdOrLegacyEnvId(storageFile.storageId);
+        if (storageWorkspace?.id !== surveyWorkspaceId) {
+          logger.error(
+            { fileUrl, surveyWorkspaceId, storageId: storageFile.storageId },
+            "Refusing to delete a response file stored outside the survey's workspace"
+          );
+          return;
+        }
+
+        await deleteFile(
+          storageFile.storageId,
+          storageFile.accessType,
+          storageFile.fileName,
+          surveyWorkspaceId
+        );
+      } catch (error) {
+        logger.error({ error, fileUrl }, "Failed to delete file");
+      }
+    })
+  );
+};
--- a/apps/web/lib/response/service.ts
+++ b/apps/web/lib/response/service.ts
@@ -4,7 +4,6 @@ import { z } from "zod";
 import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
 import { PrismaErrorType } from "@formbricks/database/types/error";
-import { logger } from "@formbricks/logger";
 import { ZId, ZOptionalNumber, ZString } from "@formbricks/types/common";
 import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
 import {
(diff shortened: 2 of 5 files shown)


apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export const deleteResponseFileUrls = async (
+  fileUrls: string[],
+  surveyWorkspaceId: string | undefined
+): Promise<void> => {
+  if (!surveyWorkspaceId) {
+    // Without the owning workspace there is nothing to authorize against, so delete nothing.
+    logger.error({ fileCount: fileUrls.length }, "Skipping response file deletion: no workspace id given");
+    return;
+  }
```
mattinannt, line 32: **Simplification:** This hand-rolled `Map`-based memoization for `findWorkspaceByIdOrLegacyEnvId` duplicates the request-level dedupe pattern already used for the same lookup elsewhere (a `reactCache()`-wrapped sibling: `resolveClientApiIds` in `@/lib/utils/resolve-client-id.ts`). AGENTS.md's Caching section calls for `React cache()` for request-level dedupe rather than a bespoke `Map<string, Promise>` cache that only lives for the duration of `deleteResponseFileUrls`. Not a runtime bug, but worth reusing (or extending) the established helper instead of introducing separate caching logic to maintain.

---

mattinannt (commented):
Reviewed the security fix (scoping response-file deletion to the owning workspace, TOCTOU guard) and the follow-up commit fixing swallowed delete errors + deduping workspace lookups. The core logic is correct and also fixes a latent bug (an unawaited `deleteFile()` whose rejection previously wasn't caught). One minor simplification suggestion left inline; no correctness issues found.

---

apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // Several files in one response usually share a storage id (same survey/workspace prefix). Cache the
+  // resolution promise per id so the batch does one lookup per distinct id instead of one per file.
+  const workspaceByStorageId = new Map<string, ReturnType<typeof findWorkspaceByIdOrLegacyEnvId>>();
+  const resolveStorageWorkspace = (storageId: string) => {
+    const cached = workspaceByStorageId.get(storageId);
+    if (cached) return cached;
+
+    const pending = findWorkspaceByIdOrLegacyEnvId(storageId);
```
mattinannt, line 41: **Efficiency:** for N distinct storage ids in a response, this fires N separate `prisma.workspace.findFirst()` round trips concurrently via `Promise.all` (deduped per id, but not batched). A response with many file-upload elements pointing at different storage-id prefixes (legitimate multi-source data, or an attacker deliberately planting many distinct bogus/foreign ids to probe the new check) causes one DB query per distinct id instead of a single batched lookup, e.g. `prisma.workspace.findMany({ where: { OR: [{ id: { in: ids } }, { legacyEnvironmentId: { in: ids } }] } })` resolved once and then indexed in the local map. Worth batching, especially since this now runs on every response delete where it previously did zero DB queries.

---

apps/web/modules/storage/lib/delete-response-files.ts
```diff
@@ -0,0 +1,83 @@
...
+): Promise<void> => {
+  if (!surveyWorkspaceId) {
+    // Without the owning workspace there is nothing to authorize against, so delete nothing.
+    logger.error({ fileCount: fileUrls.length }, "Skipping response file deletion: no workspace id given");
+    return;
+  }
+
+  // Several files in one response usually share a storage id (same survey/workspace prefix). Cache the
+  // resolution promise per id so the batch does one lookup per distinct id instead of one per file.
+  const workspaceByStorageId = new Map<string, ReturnType<typeof findWorkspaceByIdOrLegacyEnvId>>();
```
mattinannt, line 36: **Convention (AGENTS.md — Caching):** "Use React `cache()` for request-level dedupe" is the established pattern in this exact module — `resolveClientApiIds` in `apps/web/lib/utils/resolve-client-id.ts` already wraps `findWorkspaceByIdOrLegacyEnvId` with `reactCache` for this purpose. This hand-rolled `Map<string, Promise<...>>` reimplements the same dedupe manually (and only within this one call, not across the request), instead of reusing/extending the existing `reactCache`-based helper. Minor, but it's the kind of duplicate caching layer the guideline is meant to avoid.

---

mattinannt (commented):
Single-pass code review (high effort, Agent tool unavailable — no multi-agent fan-out or subagent verify pass ran; one context worked through all 8 angles and re-checked each candidate against the diff/tests before posting). Findings summary below; only 2 survived verification, both cleanup/efficiency-level — no correctness bug was found in the security fix itself, and the new/updated unit tests (`delete-response-files.test.ts`, the v2 `utils.test.ts`) were run and pass after building the `@formbricks/logger` and `@formbricks/storage` workspace packages.

---

xernobyl (approved):
Verified the fix closes the cross-tenant deletion bug: both response-delete paths (legacy `lib/response/service.ts` and v2 management) now go through the shared `deleteResponseFileUrls` helper, which re-resolves each file URL's storage id against the DB at delete time and refuses anything that doesn't match the survey's own workspace — closing the TOCTOU where a planted URL under a not-yet-existing element id could later be treated as a real answer. The trusted workspace id at both call sites comes from the response's own `surveyId`, not client input. No other call site bypasses the new helper.

Test coverage is solid (same-workspace, cross-workspace refusal, legacy env-id prefix, unresolvable id, unparseable URL, mixed batch, dedup caching, failed-delete logging), and CI is green across the board (typecheck/lint/unit/e2e, Sonar quality gate passed).

Non-blocking, for a follow-up if you want it: the open CodeRabbit/self-review comments on `delete-response-files.ts` (batching the per-storage-id workspace lookups into one `findMany`, reusing the existing `reactCache`-wrapped resolver instead of a local `Map`, and skipping the redundant fallback id when it equals the primary) are all still valid but don't affect correctness or security.

kiln-agent (changes requested):
The workspace ownership check closes the cross-tenant deletion path, but the legacy-prefix deletion call can redirect deletion to a different key under the current workspace prefix. The parsed URL already identifies the exact authorized storage prefix, so response cleanup should delete only from that prefix and should not supply the current workspace as a fallback.

kiln-agent (commented):
Reviewed the full diff at `108f9cbf56aedaf65c2e13d69a8d421c7c2ddbe3`. I found no blocking correctness, security, or maintainability issues.

The delete-side authorization is applied in the shared helper before `deleteFile` is reached, both response-deletion paths now use that helper, and the legacy environment-id case remains scoped by resolving it back to the owning workspace. The focused tests cover the cross-workspace refusal, unresolved and malformed inputs, same-workspace and legacy deletion, mixed batches, failed storage results, and lookup deduplication.

I also verified the patch with `git diff --check`. I could not rerun the focused Vitest files locally because this checkout has no installed `node_modules`; the PR's unit-test, typecheck, lint, E2E, CodeQL, and Sonar checks are green.

PR #8888, opened 2026-08-17
chore: move all QA onto the PR and slim the PR template

mattinannt:
Fixes ENG-2007 — [Retarget PR template for agent-authored, Linear-driven PRs](https://linear.app/formbricks/issue/ENG-2007/retarget-pr-template-for-agent-authored-linear-driven-prs)
Ref ENG-2315 — [Define the PR evidence contract](https://linear.app/formbricks/issue/ENG-2315/define-the-pr-evidence-contract)

## What & why

`## QA / Test Plan` existed so a release-QA pass could harvest tester-facing steps weeks after
merge. That pass is going away: all QA for a change now happens on its own PR before review, and
release review only looks for problems arising from the interplay of several changes.

So the section merges into `## How this was tested`, which becomes **Coverage** (one row per changed
behaviour: the test that covers it, how strong that check is, what was observed) plus **Open gaps**
(what is not verified, and why). The creator shows the coverage; the reviewer challenges the gaps.
For UI work the proof is a screenshot or screencast attached as coverage, not a click-by-click script.

Also in this PR:

- Guidance cut to one or two sentences per section, and one instruction at the top that applies to
  every PR: bullets and `<details>` folds for readability, with the Coverage table, Open gaps and
  Breaking changes never folded.
- `## Linear ticket` heading dropped — the ticket line stands on its own, and Linear's integration
  scans the whole body for the magic word.
- **`## Migrations & env`** kept as its own small section, covering non-breaking env vars and additive
  migrations too. It cannot live inside Breaking changes: `pr-label-sync.yml` reads that whole span
  and would mislabel the PR.
- Authors told not to restate what CI already reports; the description carries what checks cannot
  show.
- **Coverage records evidence strength, not test counts.** `How` takes a tier — `unit (red on main)`,
  `unit (mutation)`, `unit (guard)`, `e2e`, `manual` — and a red-on-main or mutation row must carry
  the command or the mutated `file:line` so a reviewer can rerun it. A suite of 14 green cases where
  only one fails against the old code now reads differently from one where all 14 do.
- Optional model note at the bottom, for PRs where an agent wrote code or the description.
- **Magic words are confined to the ticket lines.** Every ticket id in the template and CodeRabbit
  guidance is an unresolvable placeholder, `AGENTS.md` states the rule, and an advisory check flags a
  closing keyword next to a ticket id anywhere else. This bug bit this very PR — see the fold below.

**How the template covers the author's side of the evidence contract:** intent → the ticket line; what
changed and why → `What & why`; what was tested and how → `Coverage`; self-verification → `Coverage`
plus `Open gaps`; risk areas → `Risks`; which model produced it → the model note.

**Two bullets of ENG-2007 are superseded rather than delivered**, so that ticket's description needs a
pass before it closes: it asks to *keep* `QA / Test Plan` and its subsection headings as the release-QA
source of truth, and to add a prominent `Linear ticket` section. Both are deliberately reversed here —
ENG-2315 anticipated exactly this ("ENG-2007 sits in M1 but its content depends on this contract").
Still outstanding on ENG-2315 itself: the reviewer's half of the contract lives in the Notion PR
guidelines rather than the template, and one of three retro-fits is done (#8886).

<details>
<summary>The bug that prompted the magic-word guard</summary>

An earlier revision of this description explained the dropped heading by writing the example ticket
line out in prose, inside backticks. Linear read it, attached this PR to that unrelated ticket, and
its workflow automation moved the ticket out of Done — and on merge the closing keyword would have
closed it. Inline code is not an exemption: the scanners read the raw body.

The attachment and the ticket's state have been restored by hand. The guard in this PR is so the
next author doesn't repeat it while documenting the convention.

</details>

<details>
<summary>Where the old QA approach was written down</summary>

- `.github/pull_request_template.md` — the section itself and its guidance.
- `AGENTS.md` — called the template "the source of truth for PR structure, including the
  `## QA / Test Plan` section used to generate release QA", and asked authors to paste `pnpm test`
  output. Both updated.
- `.coderabbit.yaml` — the `QA plan is executable by a tester` check keyed on the old heading and
  **failed** any step referencing a source file or function name. Its premise was a tester with no
  repo access, so it is replaced by `Coverage accounts for the whole diff`.
- `.github/workflows/pr-label-sync.yml` — described a separate "Release QA audit"; reworded to what
  still happens, which is that release notes and the migration guide are compiled from PRs carrying
  the `breaking-change` label. The label mechanism is unchanged.
- Outside the repo: the Notion Release QA process and its QA Project Generator Prompt still key on
  the old heading — tracked in ENG-2417.

</details>

## Breaking changes

None

## Migrations & env

- none

## How this was tested

No product behaviour changes — this PR touches a PR template, an agent instruction file, a CodeRabbit
config and one workflow comment.

| Behaviour | How | Outcome |
| --- | --- | --- |
| Template, AGENTS.md, CodeRabbit config and workflow formatting | `manual` — `prettier --check` on the four files | passes |
| CodeRabbit config still parses and keeps its checks | `manual` — `yaml.safe_load`, listed `custom_checks` | three checks: breaking changes, magic word, coverage |
| Label workflow reads the new template correctly | `manual` — replayed `pr-label-sync.yml`'s section split and scope-word check against the template | `Breaking changes` span is exactly `None`, no label applied; `Migrations & env` is a separate span |
| Config still valid against the published schema | `manual` — `Validate CodeRabbit Config` job in CI | see checks below |
| No resolvable ticket id left in PR guidance | `manual` — grep for `ENG-<digits>` and `#<digits>` across the files | none remain |
| The tier vocabulary is usable on a real PR | `manual` — applied it to #8886, measuring each tier by running the suites | 14 cases resolved to 1 red-on-main, 3 mutation, 10 guards; both commands documented there |

**Open gaps**

- The new CodeRabbit checks have not run against a real PR yet — they take effect once merged, so
  their first real signal is the next PR opened after that.
- Three custom checks may exceed what the CodeRabbit plan allows; if so, the magic-word check is the
  one to drop, since the placeholders already remove the hazard from the template itself.
- `pull_request_target` workflows run from the base branch, so the label-workflow behaviour above was
  verified by replaying its logic locally rather than by a live run on this PR.

**Risks**

- The template is the first thing every contributor sees; wording that reads as optional will be
  skipped. Worth a round of team feedback before merge rather than after.
- The tier vocabulary asks authors to classify their own evidence, which is a judgement call. If it
  gets applied loosely, `unit (guard)` becomes the default and the distinction stops meaning anything.
- Dropping `## QA / Test Plan` removes the anchor anything outside this repo might key on — the
  Notion process is tracked in ENG-2417, but saved queries or personal checklists may also need it.

---

> [!NOTE]
> **AI model used** — `…`, reasoning effort `high`.


.coderabbit.yaml
```diff
@@ -355,27 +355,28 @@ reviews:
...
           this PR does not touch.
 
-      - name: QA plan is executable by a tester
+      - name: Coverage accounts for the whole diff
         mode: warning
         instructions: |
-          Check the PR description's "## QA / Test Plan" section. Its audience is a manual tester
-          with no repository or CLI access, working only from the running app or the live API.
+          Check the PR description's "## How this was tested" section. It holds all QA for the
+          change — there is no separate release QA pass — so it must account for every behaviour
```
mattinannt, line 376: This PR asserts in three places (here, AGENTS.md, and the PR template) that "there is no separate release QA pass" anymore. But `.github/workflows/pr-label-sync.yml` (untouched by this diff) still has a header comment saying "The Release QA audit filters on this label to compile release notes + the migration guide." That leftover description of a distinct "Release QA audit" process contradicts the policy this PR just wrote — worth updating that comment too so the repo doesn't describe two different processes for the same thing.

---

.github/pull_request_template.md
```diff
@@ -49,29 +35,37 @@ None
...
+to challenge this list, so an empty one is a claim rather than a formality. -->
+
+- none
+
+**Risks**
 
-**Risks & regressions** <!-- what nearby behavior could break; what to re-check -->
+<!-- What nearby behaviour could break, and what to re-check if it does. -->
 
 -
```
mattinannt, line 81: The old template's `**Migrations / env / cutover**` subsection (new/changed env vars, DB migrations, cutover steps — mandatory "none" if empty) has no replacement here. The new "Breaking changes" section only forces disclosure when an env/config change is required or has no safe default; a PR that adds an optional env var or a backward-compatible migration is no longer required to mention it anywhere. Deployers/self-hosters lose the guaranteed "here's what changed operationally" line that used to be enforced even for non-breaking changes.

---

.github/pull_request_template.md
```diff
@@ -49,29 +35,37 @@ None
...
```
mattinannt: The old template's `**Preconditions / test data**` subsection (accounts, seed data, feature flags, plan/entitlement, Cloud vs self-host) was dropped with no equivalent field. The new "Coverage" table's "How" column just says "unit / e2e / manual" — a reviewer challenging an "Outcome" claim (e.g. "verified manually") has no required place to learn what account/plan/flag state made that verification meaningful, which was previously mandatory disclosure.

---

mattinannt (commented):
Single-pass review (Agent tool unavailable, no subagent fan-out/verify — high-effort recall pass done inline). This is a docs/config-only PR (`.coderabbit.yaml`, `.github/pull_request_template.md`, `AGENTS.md`); findings below are process/consistency issues rather than code bugs. Full JSON summary in the top-level reply.

---

Dhruwang:
Suggestion on the Coverage table 

Coverage rows say *how* something was verified ("unit ×2") but not *how strong* that verification is. Three quite different things end up looking the same:

- a test that fails on the old code — proves the bug existed
- a test that only fails under a deliberate mutation, because the fix lives in a new file and the test can't run against main at all
- a test that passes either way — a guard against future breakage, not evidence

#8886 is the worked example this template came from, and it's exactly where this bites: 14 cases, of which 1 actually goes red on main. The description says "reverting the helper to the URL-derived key turns the cross-workspace case red" — true, but a reviewer can't rerun that sentence.

What it could look like as a sub-table under Coverage:

| Test | Tier | On old code | Here |
| --- | --- | --- | --- |
| `refuse to delete a file whose storage id belongs to a different workspace` | fails on main | ❌ foreign file deleted | ✅ |
| `refuses to delete a planted URL that resolves to a different workspace` | mutation | ❌ deleted | ✅ |
| `in a mixed batch, deletes the owned file and refuses the foreign one` | mutation | ❌ both deleted | ✅ |
| 9 others — legacy prefixes, unparseable URLs, failed deletes | guard | ✅ | ✅ |

with one line naming the mutation (`delete-response-files.ts:406`, workspace check deleted) and the revert command for the red-on-main row:

```bash
git stash push -- .../[responseId]/lib/utils.ts
pnpm test .../lib/tests/utils.test.ts     # 1 failed
git stash pop && pnpm test ...            # 5 passed
```

Two changes in substance: name the tests rather than count them, and tier them. No new tooling or extra tests — `pr-tests` already runs this revert gate when writing them, the output just has nowhere to land today.

Only worth it where a behaviour change is claimed; refactors and copy changes wouldn't carry the table. Could also be a `<details>` fold under Coverage rather than always-visible, if the concern is template weight.

AGENTS.md
```diff
@@ -168,9 +168,9 @@ Heuristic:
 
 ## Commit & Pull Request Guidelines
 
-Commits follow a lightweight Conventional Commit format (`fix:`, `chore:`, `feat:`) and usually append the PR number, e.g. `fix: update OpenAPI schema (#6617)`. Keep commits scoped and lint-clean. Pull requests should outline the problem, summarize the solution, and link to issues or product specs. Attach screenshots or gifs for UI-facing work, list any migrations or env changes, and paste the output of relevant commands (`pnpm test`, `pnpm lint`, `pnpm db:migrate:dev`) so reviewers can verify readiness.
+Commits follow a lightweight Conventional Commit format (`fix:`, `chore:`, `feat:`) and usually append the PR number, e.g. `fix: update OpenAPI schema (#6617)`. Keep commits scoped and lint-clean. Pull requests should outline the problem, summarize the solution, and link to issues or product specs. Attach screenshots or gifs for UI-facing work and list any migrations or env changes. Don't restate what CI already reports (lint, typecheck, unit tests, build, Sonar) — the description carries what those checks cannot show.
```
mattinannt: This still says to "list any migrations or env changes" in the PR description, but this same PR removes the template's dedicated "Migrations / env / cutover" subsection that was the place to do that (see the `.github/pull_request_template.md` diff). A non-breaking env var addition or an additive migration doesn't trigger the "Breaking changes" section (per the updated `.coderabbit.yaml` criteria), and "Coverage"/"Open gaps" only track test verification — so there's no longer a required field where that fact gets recorded. Either point this instruction at "Breaking changes" (if that's now the intended home) or restore a slot for non-breaking migration/env notes.

---

.github/pull_request_template.md
```diff
@@ -49,29 +35,37 @@ None
...
+pass. Don't restate the checks below (lint, typecheck, unit tests, build, Sonar); give the reviewer
+what those cannot show. -->
 
-**How to test** <!-- happy path first, then edge cases (boundaries, negative input, permissions) -->
+**Coverage**
 
-- [ ] Step → expected result (include route, error code, and any env flag/var)
-- [ ] Edge case → expected result
+<!-- One row per behaviour the diff changes: how you verified it and what you observed. For UI work,
+attach the screenshot or screencast that proves it — fold it if long. -->
```
mattinannt: This says to fold the Coverage screenshot/screencast "if long," but the header comment a few lines up (lines 4-5) says "Never fold Coverage, Open gaps or Breaking changes." An author (or an AI agent following this literally) can't satisfy both at once, and it's ambiguous whether "never fold Coverage" means the section itself or extends to media attached inside it. Worth clarifying — e.g. "never fold the Coverage table itself, but long media inside it may still be folded."

---

mattinannt (commented):
Reviewed. This PR reworks the PR template and its supporting docs/config (docs-only diff, no executable code). Two self-consistency issues found and left as inline comments — both are documentation gaps rather than functional bugs.

---

.github/pull_request_template.md
```diff
@@ -49,29 +35,37 @@ None
...
-<!--
-Source of truth for release QA. Audience: a manual tester with no repo/CLI access, using only the
-running app (UI) or live API (HTTP). Write each step as an observable action → expected result. No
-automated verification here (`pnpm test`, `grep`, `pnpm build`, code references) — that belongs in
-"How this was tested". Describe the ACTUAL behavior click-by-click; omit a subsection only when it
-truly doesn't apply. Flag inferred specifics (thresholds, limits, timeouts) for human verification.
--->
+<!-- REQUIRED. All QA for this change happens here, before review — there is no separate release QA
+pass. Don't restate the checks below (lint, typecheck, unit tests, build, Sonar); give the reviewer
+what those cannot show. -->
```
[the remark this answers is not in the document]
mattinannt: This PR changes this approach and requirement. This has been discussed with the team already. No need to fix this.

.github/pull_request_template.md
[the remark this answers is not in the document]
mattinannt: Fixed in 19c5f8a — the placeholder is now `- none`, matching the `Open gaps` default just above it. Consistent defaults also make an unedited section obvious rather than ambiguous.

---

.github/pull_request_template.md
mattinannt: Fixed in 19c5f8a. The top rule now reads:

> Never fold the Coverage table, Open gaps or Breaking changes — long media inside Coverage may still be folded.

That resolves the ambiguity in the direction intended: the table itself must stay visible because it carries the contract, while a long screencast inside it may be folded. "long media" also came out of the fold-candidates list in the same comment, since it is now covered explicitly.

---

.github/pull_request_template.md
mattinannt: Agreed, and fixed in 19c5f8a with a dedicated `## Migrations & env` section that defaults to `none` and explicitly covers non-breaking cases:

> New or changed env vars, DB migrations and cutover steps — including non-breaking ones, since this is what a deployer or self-hoster acts on. Write "none".

One constraint worth recording: this could not be folded into `## Breaking changes`, which was the obvious cheaper option. `pr-label-sync.yml` reads everything between that heading and the next `##`, and its scope-word vocabulary (lines 59-61) treats any unrecognised word as a declaration — "env", "migration" and "API" all qualify — so operational notes there would falsely apply the `breaking-change` label, which feeds the release notes and the migration guide.

Verified by replaying the workflow's extraction against the new template: the `Breaking changes` span is exactly `None`, and the new section is a separate span.

---

.github/pull_request_template.md
mattinannt: Valid — fixed in 19c5f8a, though as a clause in the Coverage guidance rather than a restored subsection:

> Name the account, plan or flag state where an outcome depends on it, and for a behaviour fix name the test that fails without the fix.

Reasoning for not bringing back `**Preconditions / test data**`: as its own subsection it was filled in unconditionally, including for the many PRs where no account or flag state matters, which is how it became boilerplate. As a qualifier on the Coverage row it sits next to the claim it makes challengeable, and only appears when an outcome actually depends on it. If it turns out reviewers still can't challenge manual rows, a subsection is the fallback.

---

AGENTS.md
mattinannt: Fixed in 19c5f8a by taking the second option — a slot now exists, and this line points at it:

> …and record any migrations or env changes under `Migrations & env`, breaking or not.

Pointing it at `Breaking changes` instead would have been wrong for the reason the sibling thread covers: prose in that section trips the label workflow's scope-word check and mislabels the PR.

---

.coderabbit.yaml
mattinannt: Half right, and fixed in 19c5f8a. The comment now reads:

> Release notes and the self-hoster migration guide are compiled from the PRs carrying it.

The stale part was the phrase "Release QA audit", which named a distinct process this PR removes. What that comment described underneath is still true and unchanged: the `breaking-change` label is what release notes and the migration guide are compiled from. So the fix is the wording, not the mechanism — nothing about label syncing changes here.

The same heading is still referenced outside this repo, in the Notion Release QA process and its QA Project Generator Prompt. That's tracked in ENG-2417, since it needs a decision from the QA Lead rather than a find-and-replace.

---

mattinannt:
Your analysis of #8886 is right, and it's a better catch than the template deserved. Checked it: `delete-response-files.test.ts` is a new file testing a new module, so none of its 9 cases can run against `main` at all. Only the cross-workspace case added to `utils.test.ts` goes red there. So "the suite can fail" was doing work it hadn't earned — the honest statement is 1 red-on-main, 2 mutation-only, 11 guards.

Adopted the substance in 19c5f8a as one clause in the Coverage guidance:

> …and for a behaviour fix name the test that fails without the fix.

That forces the distinction you're drawing — a named test whose failure is attributable beats a count — without a per-test table on every PR. `Coverage` rows still say *how*, and the named test says *how strong*.

Left out of the template, deliberately:

- **The Tier / On-old-code sub-table.** It's the right instrument for a security or data-integrity fix, and the wrong one for the copy change that also has to fill this in. I'd rather it live in the Notion PR guidelines as an optional pattern you reach for on high-risk PRs than as template weight everyone pays. If you'd rather it be structural, say so — it's a small change to add it as a folded block, and this PR is the moment to decide.
- **The revert-command block.** Same reasoning, plus it's per-PR bespoke.

One thing I couldn't verify: `pr-tests` isn't in this checkout — `.claude/skills` and `.agents/skills` hold only `nextjs-docs`. So the revert gate you're describing has no home the rest of us can run yet. Where does that skill live? If it's in the robots clone it can be installed here, and then the tier column becomes something a skill fills in rather than a thing authors hand-write, which is a much better version of this idea.

---

mattinannt:
Went back over this — my first pass adopted your naming point but left the part you actually objected to. You wrote that a reviewer can't rerun a sentence; I'd replaced one unrerunnable sentence with another. Fixed properly in bec8fda, and I ran the gate rather than reasoning about it.

**Measured on #8886** — 14 cases, and the split is not what either of us guessed:

| Tier | Count | Observed |
| --- | --- | --- |
| red on main | 1 | `refuse to delete a file whose storage id belongs to a different workspace` → 1 failed / 4 passed against `main`'s `utils.ts` |
| mutation | 3 | guard deleted at `delete-response-files.ts:55-62` → 3 failed / 6 passed |
| guard | 10 | pass either way |

You estimated 2 mutation and 9 guards. The third mutation case is `refuses when the storage id resolves to no workspace` — it goes red too, because deleting the guard removes the no-workspace branch along with the mismatch branch. Baseline is 14/14 green, and the tree is back to clean.

**Template changes:**

1. `How` now takes a tier instead of a count — `unit (red on main)` / `unit (mutation)` / `unit (guard)` / `e2e` / `manual`. No new column; the existing column just gets a vocabulary, with one line defining each.
2. Any red-on-main or mutation row must carry the command or the mutated `file:line`. `.coderabbit.yaml`'s coverage check now fails a row claiming either tier without something rerunnable.

#8886 is updated to comply — tiered rows, named tests, and both commands pasted, including the detail that cost me twenty minutes: the suites fail to resolve `@formbricks/logger` unless workspace packages are built first.

What I did **not** adopt is the separate sub-table. With the tier in the `How` column and the command below the table, it would be a second place to say the same thing. If you still want the standalone table for high-risk PRs, easiest next step is your `pr-tests` skill emitting exactly this shape — which needs answering where that skill lives, since `.claude/skills` and `.agents/skills` here hold only `nextjs-docs`.

---

.coderabbit.yaml
```diff
@@ -355,27 +355,44 @@ reviews:
...
-          PASS also when the PR changes no user-facing or API behaviour — for example it only
-          touches CI configuration, repository tooling, tests, or documentation — and the section
-          says so explicitly rather than being left empty.
+          - The "Coverage" table is empty, or holds only the unedited template placeholder row.
+          - A behaviour the diff changes has neither a row in "Coverage" nor an entry under
+            "Open gaps".
+          - A row names a behaviour with no verification method, or a method with no observed
+            outcome.
+          - A row claims `red on main` or `mutation` without naming the test, and without a command
+            or a mutated `file:line` a reviewer could rerun.
```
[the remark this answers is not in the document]
mattinannt: Correct on the `and`, and it's the worst place to have got it wrong — fixed in 9014116. The joined condition meant a row satisfying either half passed, so a `red on main` claim with no rerunnable command would have sailed through, which is precisely the case the tier vocabulary was added to catch. Now two independent conditions:

```
- A `unit` or `e2e` row does not name the test or spec it rests on. A `manual` row has no
  test to name and passes without one.
- A row claims `red on main` or `mutation` without a command or a mutated `file:line` a
  reviewer could rerun.
```

Partly disagreed on the second half. Requiring a test name from **every** row is what the template said, but that was the template's error rather than the check's: a `manual` row has no test to name. This PR's own Coverage table is all `manual` rows, and #8886 has one — under a blanket rule both would fail for omitting something that does not exist. So naming is now scoped to `unit` and `e2e` rows, and the template wording changed to match:

> for an automated row name the test or spec (not a count)

Not taken from the proposed diff: it kept naming universal, which would have made the rule unsatisfiable for manual verification.

---

itsjavi (approved):
LGTM, let's try this

Dhruwang (approved):

PR #8889, opened 2026-08-17
fix(email): improve Outlook preview markup

mattinannt:
this is a test

### DIFF
--- a/apps/web/modules/email/components/preview-email-template.tsx
+++ b/apps/web/modules/email/components/preview-email-template.tsx
@@ -442,7 +442,7 @@ export async function PreviewEmailTemplate({
 
 function EmailTemplateWrapper({
   children,
-  surveyUrl,
+  surveyUrl: _surveyUrl,
   styling,
 }: {
   children: React.ReactNode;
@@ -480,12 +480,9 @@ function EmailTemplateWrapper({
           },
         },
       }}>
-      <Link
-        className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 block overflow-auto border border-solid p-8 font-sans text-inherit"
-        href={surveyUrl}
-        target="_blank">
+      <Section className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 border border-solid p-8 font-sans text-inherit">
         {children}
-      </Link>
+      </Section>
     </Tailwind>
   );
 }
--- a/apps/web/app/(app)/environments/[environmentId]/surveys/[surveyId]/(analysis)/summary/lib/emailTemplate.test.ts
+++ b/apps/web/app/(app)/environments/[environmentId]/surveys/[surveyId]/(analysis)/summary/lib/emailTemplate.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, test } from "vitest";
+import { extractEmailBodyFragment } from "./emailTemplate";
+
+describe("extractEmailBodyFragment", () => {
+  test("returns the body contents for rendered email documents", () => {
+    const html = `
+      <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+      <html>
+        <head>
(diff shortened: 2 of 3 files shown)


apps/web/modules/email/components/preview-email-template.tsx
```diff
@@ -480,12 +480,9 @@ function EmailTemplateWrapper({
...
         },
       }}>
-      <Link
-        className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 block overflow-auto border border-solid p-8 font-sans text-inherit"
-        href={surveyUrl}
-        target="_blank">
+      <Section className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 border border-solid p-8 font-sans text-inherit">
         {children}
-      </Link>
+      </Section>
```
mattinannt, line 485: **Regression: several survey-preview email types lose their only click-through to the survey.**

`EmailTemplateWrapper` used to wrap its `children` in `<Link href={surveyUrl} target="_blank">`, making the whole card clickable. Here it's replaced with a plain `<Section>` and `surveyUrl` is now unused (renamed to `_surveyUrl`).

That fixes the invalid nested-`<a>` issue for question types whose body already contains an `EmailButton`/`Link` (NPS, Rating, Consent, MultipleChoiceSingle, PictureSelection single-choice, CTA-with-external-button). But for the question types whose body has **no** interactive element at all — `OpenText`, `MultipleChoiceMulti`, `Ranking`, `Date`, `Address`/`ContactInfo`, `FileUpload`, `PictureSelection` with `allowMulti`, and `CTA` when `buttonExternal`/`ctaButtonLabel`/`buttonUrl` aren't all set — there is now **no** clickable element pointing at the survey anywhere in the rendered email. `ElementHeader` (the headline/subheader) contains no link either.

This html isn't just an internal preview: `getEmailTemplateHtml` → `sendEmbedSurveyPreviewEmailAction` sends it as a real email, and the "Embed" tab in `email-tab.tsx` lets admins copy this exact markup into their own outbound campaigns to real respondents. For any of the question types above, recipients would have no way to reach the survey from the email.

Consider keeping a click target (e.g. wrap only the `ElementHeader`/non-interactive fallback content in the survey `Link`, or fall back to wrapping the whole card in the link only when `children` contains no other anchor) instead of removing the affordance outright for every question type.

---

apps/web/modules/email/components/preview-email-template.tsx
```diff
@@ -480,12 +480,9 @@ function EmailTemplateWrapper({
           },
         },
       }}>
-      <Link
-        className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 block overflow-auto border border-solid p-8 font-sans text-inherit"
-        href={surveyUrl}
-        target="_blank">
+      <Section className="bg-card-bg-color border-card-border-color rounded-custom mx-0 my-2 border border-solid p-8 font-sans text-inherit">
```
mattinannt, line 483: Minor: the new `Section` className dropped `block overflow-auto` compared to the old `Link`'s className. That utility was presumably there so wide content (the 11-button NPS row, a Matrix table with many columns) scrolls/clips inside the fixed-width card instead of overflowing it. Worth double-checking rendering for those question types in email clients that don't otherwise constrain width.

---

apps/web/modules/email/components/preview-email-template.tsx
```diff
@@ -442,7 +442,7 @@ export async function PreviewEmailTemplate({
 
 function EmailTemplateWrapper({
   children,
-  surveyUrl,
+  surveyUrl: _surveyUrl,
```
mattinannt, line 445: Cleanup: since `surveyUrl` is no longer used inside `EmailTemplateWrapper`, the prop (and the `url`/`urlWithPrefilling` plumbing passed as `surveyUrl={url}` at every call site in the switch above) is now dead weight — either restore its use (see the click-through regression noted above) or remove the prop entirely instead of prefixing it with `_`.

---

mattinannt (commented):
Automated single-pass review (high effort, no multi-agent fan-out/verify subagent available in this run). Findings posted inline; ranked most-severe first.

---

PR #8890, opened 2026-08-17
chore(typecheck): include test files in the typecheck graph

pandeymangg:
## What & why

`apps/web/tsconfig.typecheck.json` excluded every test file, so no test fixture was type-checked in CI — a fixture could claim to model a production type and silently diverge from it (the mechanism behind the formbricks/formbricks#8743 audit-filter finding). This PR closes that hole (phase two of ENG-2261, which wired `pnpm typecheck` into CI):

- **Removed all five test exclusions** from `apps/web/tsconfig.typecheck.json` (`*.test.ts`, `*.test.tsx`, `tests/`, `__mocks__/`, `__tests__/`) and the same `src/**/*.test.ts` exclusion from `packages/survey-ui/tsconfig.json` (the only other package that had one — the rest already typecheck their tests; `tsconfig.build.json` exclusions are build-only). survey-ui passed with zero fixes.
- **Measured error count: 108 errors across 31 test files** — small enough to land in one pass, no directory scoping needed. (A raw first run showed 268; 160 were a stale local `packages/workflows/dist`, gone after a clean rebuild.)
- **Fixed all 108.** Most were genuine drift, exactly what the ticket predicted:
  - Survey-list fixtures missing `archivedAt`/`hasArchived` — fields production selects and returns (5 files). One test asserted an expected object production could never produce; once the fixture was honest it failed at runtime and the expectation was corrected.
  - AI service inputs (`generate`, `translate-fields`, `ai-chart-query`): 33 call sites missing the top-level `workspaceId`/`userId` production requires for AI tracing.
  - `Prisma.PrismaClientKnownRequestError` used in type position (it's value-only there) — switched 5 files to the named type import, matching the production convention in `lib/utils/prisma-error.ts`.
  - Removed the `as never` casts in `redact-credentials.test.ts` — the fixtures satisfy the real type; the casts were hiding that and collapsed results to `never`.
  - Assorted fixture gaps (`workspaceName`, `displayTimeZone`, `scopeType`, `legacyEnvironmentId`), an invalid `null as const`, unused imports, tuple annotations on `mock.calls`.
  - Only 4 casts survived, each with a justification comment: two i18next `TFunction` stubs (branded overloads can't be satisfied structurally), better-auth passing an `Error` where its own type says `string` (the test exercises that documented runtime gap), and the charts actions test where the mocked action client makes the export the raw handler.
- **One production file, type-only:** `modules/mcp/server.ts` — `identifyMcpUser` was annotated with PostHog's `identify` union (callback | static identity), which isn't callable; narrowed to the callback member. Zero runtime impact.
- **Include gaps closed (review follow-up):** the include globs missed apps/web's root-level and `lingodotdev/` test files (`proxy.test.ts`, `instrumentation*.test.ts`, `lingodotdev/*.test.ts`) — added `*.ts` and `lingodotdev/**/*.ts`, which surfaced and fixed 4 real drift errors in `proxy.test.ts` (fixed by narrowing two over-wide `Response | null` annotations in `proxy.ts` to `NextResponse | null` — type-only). Also removed `**/*.integration.test.ts` from the base `tsconfig.json` exclude so the IDE and CI typecheck agree on the same file set (`next build` does not gate on either).
- **Heap bump:** with tests in the graph, `tsc` OOMs at Node's default ~4GB heap. The typecheck script now uses `cross-env NODE_OPTIONS=--max-old-space-size=8192`, the same pattern as the build script. Verified a cold run fits in 8GB.

## Linear ticket

Fixes ENG-2265
Fixes ENG-2229 (earlier report of the same exclusion, filed from the ENG-1718/1719 fixture-drift incidents)

## How this was tested

- `pnpm typecheck` from the root: 27 tasks green with test files in the graph ✅
- `npx vitest run` on all 32 touched test files: 394 tests passed ✅ (includes the one runtime-visible fixture correction in `modules/survey/list/lib/survey.test.ts`)
- Done-when check: deliberately misspelled `archivedAt` in a fixture → `tsc` failed with TS2741; reverted → clean again ✅
- Cold `tsc` run (deleted `tsconfig.typecheck.tsbuildinfo`) passes within the new 8GB heap ✅
- Prettier + ESLint clean on every changed file ✅

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] No user-facing behavior changes — this PR only widens what CI type-checks and fixes test fixtures. The single production edit (`modules/mcp/server.ts`) is a type annotation that compiles away.
- [ ] Smoke: open the surveys list page → surveys render with status/archive state as before (its fixtures changed, the production code did not).

**Preconditions / test data**

- none

**Risks & regressions**

- CI-level only: the Typecheck job now covers ~2x the files and needs the larger Node heap; if it OOMs or slows past the 15-min job timeout, the heap value in `apps/web/package.json` is the knob.

- none


apps/web/lib/survey/service.test.ts
```diff
@@ -61,8 +61,7 @@ beforeEach(() => {
   prisma.survey.count.mockResolvedValue(1);
   // createSurvey now wraps its core writes in prisma.$transaction; run the callback with the same
   // mocked client so per-test prisma.survey/segment mocks still apply inside the transaction.
-  vi.mocked(prisma.$transaction).mockImplementation(((callback: (tx: typeof prisma) => Promise<unknown>) =>
-    callback(prisma)) as typeof prisma.$transaction);
+  vi.mocked(prisma.$transaction).mockImplementation(async (callback: any) => callback(prisma));
```
mattinannt: Single-pass review (no Agent-tool fan-out / subagent verify), flagging for recall.

**conventions**: `callback: any` re-introduces an `any` that AGENTS.md's Quality Checklist explicitly asks to avoid ("Prefer type inference, avoid `any`"), and it's duplicated verbatim in `apps/web/app/api/v3/surveys/patch.test.ts:229` and `apps/web/lib/survey/service.scheduling.test.ts:64`.

It also isn't needed: `apps/web/modules/auth/lib/user.test.ts` mocks the exact same `prisma.$transaction` call with a bare, un-annotated `callback` (e.g. `vi.mocked(prisma.$transaction).mockImplementation(async (callback) => callback({...}))`) and that already satisfies the mocked overload without `any` or a cast. Suggest dropping the `: any` here (and in the two sibling files) in favor of the untyped-parameter form already used elsewhere in the codebase, to keep the callback's argument type-checked.

```suggestion
    vi.mocked(prisma.$transaction).mockImplementation(async (callback) => callback(prisma));
```

---

mattinannt (commented):
Single-pass review (no Agent-tool fan-out / subagent verify pass available in this context) — most of the diff is mechanical type fixes to bring test files under `tsc` after their previous blanket exclusion was removed, and I traced each one back to a pre-existing production signature/type to confirm they weren't masking real regressions. One conventions finding survived; details inline.

---

mattinannt (dismissed):

apps/web/lib/feedback-source/service.test.ts
```diff
@@ -881,7 +881,7 @@ describe("updateFeedbackSourceWithMappings", () => {
   test("replaces field mappings when provided", async () => {
     // Starts errored: a csv save must not clear a flag only the formbricks reconciler can set.
     const tx = setupTransaction("error");
-    tx.feedbackSource.update.mockResolvedValue(undefined);
+    tx.feedbackSource.update.mockResolvedValue({ count: 0 });
```
kiln-agent: This override defeats the status tracker the assertion 30 lines below depends on.

`setupTransaction("error")` installs `update: vi.fn(applyStatus)` precisely so `trackedStatus` follows **both** `update` and `updateMany` — and the comment at the `persistedStatus()` assertion says so ("`persistedStatus` follows both `update` and `updateMany`, so this still holds if the clear moves between them"). But `mockResolvedValue(v)` is `mockImplementation(() => Promise.resolve(v))`, so `applyStatus` is gone for `update` and only `updateMany` still moves the tracker.

Concretely: if the error→active clear were moved out of `updateMany` into the main `tx.feedbackSource.update({ data: { status: "active" } })`, `trackedStatus` would stay `"error"` and `expect(persistedStatus()).toBe("error")` would still pass — the exact regression this test exists to catch slips through.

The override is also redundant today: production ignores the return value, and `applyStatus` already returns `{ count: 0 }` for an update that carries no `status`. Dropping the line restores the tracking with no other change:

```suggestion
```

(Same override at lines 777, 795 and 828 — harmless there today, but it makes `{ count: 0 }` read as an intentional model of `update`'s result, which actually returns the row, not a count.)

---

apps/web/tsconfig.typecheck.json
```diff
@@ -1,15 +1,5 @@
...
-    ".next",
-    "node_modules",
-    "playwright",
-    "**/*.test.ts",
-    "**/*.test.tsx",
-    "**/tests/**",
-    "**/__mocks__/**",
-    "**/__tests__/**"
-  ],
+  "exclude": ["../../.env", ".next", "node_modules", "playwright"],
```
kiln-agent, line 2: The hole isn't fully closed: dropping the exclusions only helps files the `include` globs reach. `include` here replaces the base config's and lists just `next-env.d.ts`, `**/*.d.ts`, `app/**`, `lib/**`, `modules/**`, `scripts/**` and `../../packages/types/*.d.ts` — so these test files are still never type-checked:

- `apps/web/proxy.test.ts`
- `apps/web/instrumentation.test.ts`
- `apps/web/instrumentation-jobs.test.ts`
- `apps/web/instrumentation-node-config.test.ts`
- `apps/web/lingodotdev/server.test.ts`
- `apps/web/lingodotdev/language.test.ts`

Their subjects are outside the program too (`proxy.ts`, `instrumentation*.ts`, `sentry.*.config.ts`): they are Next.js/Sentry entry points that no in-program file imports, so tsc never pulls them in transitively either. A fixture in `proxy.test.ts` drifting from the types in `proxy.ts` — the ENG-2265 failure mode — still passes CI after this PR.

Adding the root-level and `lingodotdev` globs to `include` (root `*.ts` and `lingodotdev/**/*.ts`) would cover them.

---

kiln-agent, line 2: Side effect worth making deliberate: the old `**/*.test.ts` pattern also matched `*.integration.test.ts`, so removing it pulls ~50 integration test files into this program — while the base `apps/web/tsconfig.json` still excludes `integration` and `**/*.integration.test.ts`.

That leaves the two projects disagreeing about the same files. Because a file in no project falls back to an inferred project (without this repo's `paths` and strict options), a contributor editing `modules/auth/lib/better-auth-hibp.integration.test.ts` gets no reliable local diagnostics and `next build` never sees it, yet `pnpm typecheck` in CI now gates on it — and those files are a sizeable chunk of the file-count growth behind the new 8GB heap requirement.

Either re-add `**/*.integration.test.ts` to this exclude list, or drop the exclusion from the base tsconfig so both agree that they are in scope.

---

kiln-agent (commented):
Reviewed the full diff for correctness. The fixture fixes hold up: I verified `archivedAt` really is in `surveySelect` and flows through `mapSurveyRowToSurvey`/`serializeV3SurveyListItem`, the removed imports (`prisma` in `operations.test.ts`, `TooManyRequestsError` in `ai/service.test.ts`) are genuinely unused, `PrismaClientKnownRequestError` is exported as a type from `@formbricks/database/prisma`, the `identify` narrowing in `modules/mcp/server.ts` resolves to the callback member and stays runtime-neutral, and `packages/survey-ui` typechecks clean with its tests included (`tsc --noEmit`, zero output). The `translate-fields`/`ai-chart-query` inputs that gained `userId` only assert with `objectContaining`, so newly enabling the `aiTracing` branch doesn't change what those tests check.

Three findings, all non-blocking; the first is the only one about test integrity.

---

mattinannt (approved):

PR #8896, opened 2026-08-17
feat(embedded-data): make reserved fields usable in recall and logic [ENG-1840]

itsjavi:
> **Stacked on #8892** (`javier/eng-1839-reserved-field-catalog-tier-1-name-collision-rules`), which carries the populated reserved-field catalog, `projectClientReservedValues`, and the `availability` field this PR depends on. **#8892 has merged**, so GitHub has retargeted this PR to `epic/embedded-data-v1` and the diff below is this PR's own changes only.

## What & why

Reserved fields (`url`, `source`, `country`, `durationSeconds`, …) are auto-captured metadata every survey can reference without declaring anything. The catalog existed, and `projectReservedValues()` existed — but with **zero production call sites**, so none of it was reachable from recall tokens, logic conditions, redirect URLs, or the field pickers. This wires it up.

**The ordering rule is the whole "nothing breaks" guarantee.** Every recall/logic value map is now built as `{ ...reservedValues, ...responseData }` — reserved first, so response data **shadows** it. A survey in production may legitimately declare a hidden field named `country`; its value lives at `response.data["country"]` and `#recall:country#` must keep resolving from there, unchanged. `FORBIDDEN_IDS` never guarded these names, so that collision set is frozen but non-empty. That ordering lives in exactly one expression:

**`packages/types/embedded-data-resolver.ts` → `mergeReservedValues()` — one line, called from every consumer.** That is the line to review.

What changed:

- **Types.** `ZDynamicReservedField` (`{ type: "reserved", value }`) added to both `ZDynamicLogicFieldValue` and the API back-compat `ZDynamicLogicFieldValueDeprecated`; `"reserved"` added to `ZSurveyRecallItem.type`. Widening only. The operand name is validated non-empty, deliberately *not* against the catalog, so a survey saved against a newer catalog still parses on an older deployment.
- **Both logic engines** get a `reserved` arm. The renderer (`packages/surveys`) resolves against the client-available projection; the server engine (`apps/web/lib/surveyLogic/utils.ts`) resolves against the **full** catalog via the new `buildServerEmbeddedValues()`. Only the `reserved` arm reads the merged map — `element` and `hiddenField` still read `data` alone, so a declared field can never pick up reserved metadata just because the respondent left it blank.
- **Renderer wiring** in `survey.tsx`: one memo projects the client-available values, one merges them under `responseData`. That single merged map feeds `parseRecallInformation`, `WelcomeCard`, `EndingCard` (which covers redirect-URL interpolation at `ending-card.tsx:87` for free), and `evaluateLogic`.
- **Mid-survey availability, enforced in the picker.** `listMidSurveyReservedEntries()` applies both gates: `availability !== "server"`, and drop any entry the survey already declares under the same name (grandfathering in the picker, so the author never sees two identically-named rows). The shadow list counts all three declaration kinds — ingested fields, variables **and element ids** — because all three win at value time.
- **Picker data sources**: the recall picker, the block-logic operand picker, and an operator set per `dataType` (string / number / boolean / date).
- **Quota evaluation resolves reserved operands.** `evaluateResponseQuotas` builds `buildServerEmbeddedValues(response)` from the row its callers already hold inside the ingest transaction (no extra query) and threads it through `evaluateQuotas` to `evaluateLogic`.
- **`privacy: "redactQuery"` is honoured at read time.** The projection narrows `url` to origin + path, so a recall token can never forward a query string. See Security below.

### Security: reserved `url` is redacted in the projection

A reserved value is interpolable — an author can put `#recall:url#` in an ending card's redirect URL (`redirect-url-form.tsx` wraps that field in `RecallWrapper`, and `ending-card.tsx` runs `replaceRecallInfo` over it before navigating) or in a follow-up email body. Projecting `window.location.href` verbatim therefore handed a third-party host whatever the survey link carried — on a single-use link that includes `suToken`, described in `validation.ts` as a credential, plus any prefill params. `LINK_SURVEY_SYSTEM_PARAMS` deliberately refuses to *capture* those into hidden fields; the reserved read routed around that refusal.

`projectEntries` — the shared body of `projectReservedValues` and `projectClientReservedValues` — now applies `redactQuery` to matching entries, so both engines narrow to origin + path from one place, and a future `redactQuery` entry is covered by declaring its policy.

Two deliberate choices:

1. **Unconditional, not gated on the Anonymize toggle.** A projection is a forwarding surface in a way storage is not. It costs an author nothing that was ever a documented capability — query params surface as their own `utm*` entries once ENG-1841 lands them.
2. **Storage is untouched.** `response.meta.url` still holds the full href unless the Anonymize toggle redacts it at ingest. This is a read-side narrowing; capture and history are unchanged.

**Follow-up for the epic branch, not this PR:** ENG-1842 (#8895) has its own `redactUrlQueryParams` for the ingest side — same behaviour, fragment included — but it is a sibling stacked on the same parent and cannot import this one's helper. Once both are merged, the ingest side should point at `redactUrlQuery` in `packages/types` and the local copy should go away.

### Correction to the ticket

The ticket description lists `browser`, `os` and `deviceType` as available during the survey. **That is wrong on this branch.** All three are parsed server-side by `UAParser` from the `user-agent` request header in the ingest routes; no client-side UA read exists in `packages/surveys/src` or `packages/js-core/src`. They are `availability: "server"` in the catalog and are correctly absent from mid-survey pickers. The pickers filter on the catalog's `availability` field rather than a hand-maintained list, so the ENG-1841 entries (`pageUrl`, `utm*`, `viewport*`, …) light up automatically when that branch merges — nothing here needs to change.

### Deliberately out of scope

Exports and response filters are untouched (ENG-1847 / ENG-1848) — an acceptance criterion is that their behaviour does not change. Picker *visual* treatment is ENG-1853.

**The quota condition *picker* stays opted out, while quota *evaluation* now works.** `includeReservedFields` still defaults to `false`, so only survey block logic offers reserved operands in the editor. The reason is no longer "evaluation would silently never match" — that is fixed — but that quota conditions evaluate server-side against a persisted response, where the **full** catalog is knowable. Offering them through the mid-survey filter would expose only the client-available subset and hide `country`, `browser` and `durationSeconds`, which are the ones an author would actually want in a quota. A server-availability picker variant is presentation work owned by ENG-1853. A reserved quota condition created through the API evaluates correctly today.

## Linear ticket

Fixes ENG-1840 — https://linear.app/formbricks/issue/ENG-1840/make-reserved-fields-usable-in-recall-and-logic

## On PR size

The size bot flags ~826 lines. This ticket is already the recall/logic slice of a five-way split of milestone 2 (#8892 catalog, #8894 capture, #8895 anonymize, #8897 collision guard, this one), and it cannot be split further without shipping something that does nothing: the operand schema, the two engine arms, the renderer wiring and the picker data sources are one behaviour, and half of them would be dead code on their own. Roughly a third of the diff is the review round documented above — the element-id shadow fix, the label formatter, the projection redaction and the quota threading — which touch many small call sites rather than adding surface.

## How this was tested

Unit tests and static checks all run locally and green. **Playwright was NOT run** — see Open gaps.

| Command | Result |
| --- | --- |
| `pnpm --filter=@formbricks/types test` | ✅ 258 passed / 5 files |
| `pnpm --filter=@formbricks/surveys test` | ✅ 683 passed / 26 files |
| `pnpm --filter=@formbricks/web test` (full suite) | ⚠️ 7656 passed, 8 failed / 589 passed, 4 failed files. All four are the same pre-existing sandbox problem: `@formbricks/jobs` does not resolve here, so `instrumentation-jobs.test.ts` fails 8 tests and `pipelines.test.ts` / `process-response-pipeline-job.test.ts` / `workflows/lib/runner/dispatch.test.ts` collect 0 tests. **Verified pre-existing:** re-run with every change in this PR stashed, `instrumentation-jobs.test.ts` fails identically. None of the four is in a path this PR touches |
| `npx tsc --noEmit` (types, surveys, apps/web) | ✅ no errors in any touched file |
| `npx eslint` (all touched files) + lint-staged on commit | ✅ clean |
| `npx prettier --write` (all touched files) | ✅ clean |
| `pnpm i18n` | ✅ "All translation validations passed", one key generated into every locale |
| `pnpm build --filter=@formbricks/surveys... --force` (after `rm -rf packages/surveys/dist apps/web/public/js/surveys.* node_modules/.cache/turbo`) | ✅ 3/3 tasks, bundle re-emitted to `apps/web/public/js/` |

Coverage per acceptance criterion:

| Acceptance criterion | Coverage |
| --- | --- |
| `#recall:country#` and a logic condition on a reserved field resolve on a real response | unit (`recall.test.ts`, `surveyLogic/utils.test.ts`) |
| Server-derived fields absent from mid-survey pickers | unit (`embedded-data-resolver.test.ts`, asserts all 10 server-only names) |
| Server-derived fields resolve as **unset** if referenced mid-survey — not `0`, not `""` | unit (`packages/surveys/src/lib/logic.test.ts`) |
| **A survey with a declared field of the same name still resolves to the declared value (grandfather)** | **unit (red on main)** |
| An element id colliding with a reserved name shadows it in both pickers | unit (`embedded-data-resolver.test.ts` → `listMidSurveyReservedEntries`) |
| A `redactQuery` field projects without its query string, while storage keeps it | unit (`embedded-data-resolver.test.ts`, both projections) |
| A reserved quota operand resolves from the response | unit (`quotas/lib/evaluation-service.test.ts`, real `buildServerEmbeddedValues`) |
| Reserved labels are readable (`Duration Seconds`, not `DurationSeconds`) | unit (`safe-identifier.test.ts`, `embedded-data-resolver.test.ts`) |
| No change to export or filter behaviour | no diff in those paths |
| Reserved recall token renders a human label, not a raw `#recall:…#` tag | unit (`recall.test.ts`) |

**Proving the grandfather row is red on main.** Flip the two spreads in `mergeReservedValues` (`packages/types/embedded-data-resolver.ts`) to `{ ...responseData, ...reservedValues }`, then:

```
pnpm --filter=@formbricks/web test lib/utils/recall.test.ts
```

Observed with the order flipped: **3 failed | 60 passed**, failing exactly —

- `a survey declaring its own 'country' resolves from response.data instead`
- `shadowing is per-key: other reserved values still resolve`
- `a declared field that is present but empty still wins`

Restored, the same command is 63/63 green. `packages/types/embedded-data-resolver.test.ts` carries the same guarantee at the unit level (`THE GRANDFATHER RULE: response data wins on a colliding key`), including the falsy cases (`""` and `0`), which a `??`-style merge would get wrong.

Tests added: reserved-operand coverage in both logic engines (including an explicit assertion that an unset reserved field behaves **identically** to an unset hidden field); `mergeReservedValues` and `listMidSurveyReservedEntries` in the resolver; reserved rule sets in `logic-rule-engine.test.ts`; recall label/type/shadowing in `recall.test.ts`; and from the review round — the camelCase label formatter (including a guard that the shared snake_case helper is unchanged, since it also names contact attribute keys), the projection redaction on both sides, and the quota threading through all six `evaluateResponseQuotas` call sites.

> **UI change without a screenshot.** The picker gains a "Survey data" group. I could not capture it — the sandbox has no Postgres and no Docker, so the app never booted (see Open gaps). `apps/web/playwright/survey-reserved-fields.spec.ts` asserts exactly what the screenshot would have shown: `Url`/`Source`/`Language` present, and `Country`/`Duration Seconds`/`Ip Address`/`Browser`/`Os`/`Finished` absent. **A reviewer with a running stack should run that spec, or attach the screenshot, before this is merged.**

### Open gaps

- **Playwright was not executed.** `pg_isready` finds no server and Docker is unavailable in this sandbox, so migrations/seed/dev-server never came up. The new spec is written and typechecks, but it is **unverified** — treat it as unrun until CI or a reviewer runs it.
- **The "already-deployed SDK" claim under Risks is reasoning, not observation.** I read the `default: return undefined` arm in the current source and confirmed the pre-change engine has no `reserved` case; I did **not** execute a stored `reserved` operand against a published `surveys.umd.cjs` from npm.
- Reserved recall in **server-rendered** surfaces (response summaries, follow-up emails) is not wired: those call `parseRecallInfo` with raw `response.data`. They would only need to pass `buildServerEmbeddedValues(response)` instead, but that touches summary/export paths this ticket is scoped out of.
- **Follow-up conditions do not exist on this branch.** The ticket brief assumed follow-ups evaluate logic conditions through `apps/web/lib/surveyLogic/utils.ts`; `evaluateFollowUp` actually only looks up a `to` address in `response.data`. Nothing to wire there today.

## Breaking changes

None — both widened schemas accept everything they accepted before. `ZDynamicLogicFieldValue` and `ZDynamicLogicFieldValueDeprecated` gained a union member; no member was narrowed or removed, and `ZSurveyRecallItem.type` gained an enum value. Existing surveys, stored logic, and API payloads parse exactly as before. Both new engine parameters default (`embeddedValues = {}`), so every existing caller compiles and behaves identically.

One behaviour change worth naming even though nothing depended on it: a recall of `url` now renders origin + path rather than the full href. Nothing shipped could have relied on the previous form — this PR is what first makes the value reachable.

## QA / Test Plan

**How to test**

- [ ] Open a survey in the editor → in a question headline, type `@` → the recall dropdown lists a reserved group containing **Url**, **Source**, **Action**, **Language**.
- [ ] In that same dropdown → **Country**, **Duration Seconds**, **Ip Address**, **Browser**, **Os** and **Finished** are **not** offered. (They are server-derived; offering them mid-survey would be the bug this ticket fixes.)
- [ ] Pick **Url** → it inserts as a labelled chip reading `Url`, **not** as raw text `#recall:url/fallback:#`.
- [ ] Publish as a **link survey**, open the link → the headline renders the actual survey URL in place of the chip, **without its query string** (open the link with `?foo=bar` to see the redaction).
- [ ] Add a hidden field named exactly `country` → reopen the recall dropdown → **Country** is no longer offered in the reserved group (the declared field is offered in the Hidden fields group instead), and `#recall:country#` renders the hidden field's value, not the geo-IP country. **This is the critical regression check.**
- [ ] Rename a question's id to exactly `url` (Advanced settings on the question card) → reopen the recall dropdown and the logic left-operand dropdown → **Url** is no longer offered in the reserved group. Before the review fix, both lists offered two rows that stored the same value.
- [ ] Survey editor → block logic → **Add condition** → the left-operand dropdown has a **Survey data** group with the same client-available fields. Choose `Url`, operator `contains`, value a fragment of your link **path** → the survey routes down that branch when opened. (A fragment of the *query* will not match — see the redaction above.)
- [ ] Ending card → set a **redirect URL** containing a recall of `url` or `source` → finish the survey → the redirect target has the value interpolated, and the interpolated `url` carries no query string. **Security check:** open the survey via a single-use link and confirm `suToken` does not reach the redirect target.
- [ ] **Quotas** (Enterprise): survey → Quotas → add a quota condition → the left-operand dropdown has **no** Survey data group (picker opt-out is intentional). If you can post a quota through the v3 API with `{ "type": "reserved", "value": "country" }`, submit a response from a known country and confirm the quota now matches — that path was silently dead before this PR.
- [ ] **Regression, exports:** download a response export → column set is unchanged; no reserved columns appear.
- [ ] **Regression, filters:** response filters offer the same fields as before; no reserved fields appear.

**Preconditions / test data**

- Any workspace; a link survey is the easiest way to see `url` resolve.
- The quota check needs an Enterprise plan/entitlement.
- The grandfather check needs a survey that declares a hidden field named `country` (create it explicitly — new declarations of reserved names are refused on the strict path by ENG-1839 part 2, so use a survey where it already exists, or a hidden field added before that guard applies).
- No feature flag. Cloud and self-host behave identically.

**Risks & regressions**

- **Highest risk: the merge order.** If reserved values ever spread *over* response data, every survey with a declared field named `country`/`url`/`source`/`action`/`language` starts showing metadata instead of the respondent's answer. Re-check recall and logic on any survey declaring one of those names.
- Recall in the **welcome card**, **ending cards** and **redirect URLs** now receives a merged map through the existing `responseData` prop. Both components use that prop *only* for recall (verified), and the prop is documented as such — but any future code reading it for the respondent's actual answers must take its own prop.
- A stored `reserved` operand reaching an **already-deployed SDK bundle** hits that engine's existing `default: return undefined` arm and reads as **unset** — no crash, no exception, the condition simply does not match. Reasoned from the source, not observed against a published `surveys.umd.cjs` (see Open gaps).
- **Quota evaluation now receives an extra argument on every response.** It is the row the caller already holds, so there is no new query, but every `evaluateResponseQuotas` call site was touched — worth a glance that none dropped it. The parameter is optional and defaults to `{}`, so a missed site degrades to "reserved operands read as unset" rather than failing.
- A recall of `url` renders origin + path. If a reviewer expects the full href, that is the redaction described under Security, not a bug.
- Picking a reserved left operand sets the default operator via the survey's first *element*, which may briefly not be a member of the reserved operator set until the author picks one. This is pre-existing behaviour shared with the `hiddenField` path, not introduced here.

- none. No schema change, no env var, no data migration. `packages/surveys` must be rebuilt for the renderer change to ship — CI does this; local devs need `pnpm build --filter=@formbricks/surveys... --force` plus a hard refresh.

---

### DIFF
--- a/apps/web/lib/surveyLogic/utils.ts
+++ b/apps/web/lib/surveyLogic/utils.ts
@@ -18,6 +18,7 @@ import {
 } from "@formbricks/types/surveys/blocks";
 import { TSurveyElement, TSurveyElementTypeEnum } from "@formbricks/types/surveys/elements";
 import { TConditionGroup, TSingleCondition } from "@formbricks/types/surveys/logic";
+import { evaluateConditionGroup } from "@formbricks/types/surveys/logic-evaluation";
 import { TActionCalculate, TSurveyLogicAction } from "@formbricks/types/surveys/types";
 import { getLocalizedValue } from "@/lib/i18n/utils";
 import { getElementsFromBlocks } from "@/modules/survey/lib/client-utils";
@@ -263,28 +264,10 @@ export const evaluateLogic = (
   conditions: TConditionGroup,
   selectedLanguage: string,
   embeddedValues: TResponseData = {}
-): boolean => {
-  const evaluateConditionGroup = (group: TConditionGroup): boolean => {
-    const results = group.conditions.map((condition) => {
-      if (isConditionGroup(condition)) {
-        return evaluateConditionGroup(condition);
-      } else {
-        return evaluateSingleCondition(
-          localSurvey,
-          data,
-          variablesData,
-          condition,
-          selectedLanguage,
-          embeddedValues
-        );
-      }
-    });
-
-    return group.connector === "or" ? results.some((r) => r) : results.every((r) => r);
-  };
-
-  return evaluateConditionGroup(conditions);
-};
+): boolean =>
+  evaluateConditionGroup(conditions, (condition) =>
+    evaluateSingleCondition(localSurvey, data, variablesData, condition, selectedLanguage, embeddedValues)
+  );
 
 const evaluateSingleCondition = (
   localSurvey: TJsWorkspaceStateSurvey,
--- a/apps/web/lib/surveyLogic/utils.ts
+++ b/apps/web/lib/surveyLogic/utils.ts
@@ -1,9 +1,13 @@
 import { createId } from "@paralleldrive/cuid2";
 import {
+  RESERVED_FIELD_CATALOG,
+  type TEmbeddedValueResponse,
   findComputedEmbeddedField,
   getComputedEmbeddedFields,
   getComputedFieldDataType,
   getLogicVariableValue,
+  mergeReservedValues,
+  projectReservedValues,
 } from "@formbricks/types/embedded-data-resolver";
 import { TJsWorkspaceStateSurvey } from "@formbricks/types/js";
 import { TResponseData, TResponseVariables } from "@formbricks/types/responses";
@@ -232,19 +236,47 @@ export const getUpdatedActionBody = (
   }
 };
 
+/**
+ * The reserved-field lookup map for a **persisted** response: the full catalog projected, then
+ * shadowed by the response's own data so a declared field of the same name still wins
+ * (`mergeReservedValues` is the single expression carrying that guarantee).
+ *
+ * Server-side every reserved field is knowable, so this reads `RESERVED_FIELD_CATALOG` whole rather
+ * than the mid-survey subset — the renderer's `projectClientReservedValues` exists precisely because
+ * that is *not* true in the browser.
+ */
+export const buildServerEmbeddedValues = (response: TEmbeddedValueResponse): TResponseData =>
+  mergeReservedValues(projectReservedValues(RESERVED_FIELD_CATALOG, response), response.data);
+
+/**
+ * @param embeddedValues Reserved-field values merged UNDER `data` by `mergeReservedValues` — what a
+ *   `reserved` operand reads. Server-side callers holding a real `TResponse` should build this with
+ *   {@link buildServerEmbeddedValues} so the **full** catalog resolves (country, durationSeconds,
+ *   finished, …), not just the mid-survey subset the renderer can see. Defaults to `{}` for callers
+ *   with no response in hand — quota screening evaluates before the row exists — where every
+ *   reserved operand then reads as unset rather than throwing.
+ */
 export const evaluateLogic = (
   localSurvey: TJsWorkspaceStateSurvey,
   data: TResponseData,
   variablesData: TResponseVariables,
   conditions: TConditionGroup,
-  selectedLanguage: string
+  selectedLanguage: string,
+  embeddedValues: TResponseData = {}
 ): boolean => {
   const evaluateConditionGroup = (group: TConditionGroup): boolean => {
     const results = group.conditions.map((condition) => {
       if (isConditionGroup(condition)) {
         return evaluateConditionGroup(condition);
       } else {
-        return evaluateSingleCondition(localSurvey, data, variablesData, condition, selectedLanguage);
+        return evaluateSingleCondition(
+          localSurvey,
+          data,
+          variablesData,
+          condition,
+          selectedLanguage,
+          embeddedValues
+        );
       }
     });
 
@@ -259,7 +291,8 @@ const evaluateSingleCondition = (
   data: TResponseData,
   variablesData: TResponseVariables,
   condition: TSingleCondition,
-  selectedLanguage: string
+  selectedLanguage: string,
+  embeddedValues: TResponseData
 ): boolean => {
   try {
     let leftValue = getLeftOperandValue(
@@ -267,11 +300,12 @@ const evaluateSingleCondition = (
       data,
       variablesData,
       condition.leftOperand,
-      selectedLanguage
+      selectedLanguage,
+      embeddedValues
     );
 
     let rightValue = condition.rightOperand
-      ? getRightOperandValue(localSurvey, data, variablesData, condition.rightOperand)
+      ? getRightOperandValue(localSurvey, data, variablesData, condition.rightOperand, embeddedValues)
       : undefined;
 
     const elements = getElementsFromBlocks(localSurvey.blocks);
@@ -502,7 +536,8 @@ const getLeftOperandValue = (
   data: TResponseData,
   variablesData: TResponseVariables,
   leftOperand: TSingleCondition["leftOperand"],
-  selectedLanguage: string
+  selectedLanguage: string,
+  embeddedValues: TResponseData
 ) => {
   switch (leftOperand.type) {
     case "element":
@@ -588,6 +623,10 @@ const getLeftOperandValue = (
       return getLogicVariableValue(getComputedEmbeddedFields(localSurvey), leftOperand.value, variablesData);
     case "hiddenField":
       return data[leftOperand.value];
+    // Server-side the caller projects the FULL catalog, so country/durationSeconds/finished all
+    // resolve here — unlike the renderer, which only ever sees the client-available subset.
+    case "reserved":
+      return embeddedValues[leftOperand.value];
     default:
       return undefined;
   }
@@ -597,7 +636,8 @@ const getRightOperandValue = (
   localSurvey: TJsWorkspaceStateSurvey,
   data: TResponseData,
   variablesData: TResponseVariables,
-  rightOperand: TSingleCondition["rightOperand"]
+  rightOperand: TSingleCondition["rightOperand"],
+  embeddedValues: TResponseData
 ) => {
   if (!rightOperand) return undefined;
 
@@ -608,6 +648,8 @@ const getRightOperandValue = (
       return getLogicVariableValue(getComputedEmbeddedFields(localSurvey), rightOperand.value, variablesData);
     case "hiddenField":
       return data[rightOperand.value];
+    case "reserved":
+      return embeddedValues[rightOperand.value];
     case "static":
       return rightOperand.value;
     default:
--- a/apps/web/lib/utils/recall.test.ts
+++ b/apps/web/lib/utils/recall.test.ts
@@ -572,7 +572,9 @@ describe("recall utility functions", () => {
         mergeReservedValues(reserved, { country: "" })
       );
 
-      expect(result).not.toContain("DE");
+      // Asserted exactly, not as "does not contain DE": that weaker form also passes on an empty
+      // string or an unrendered raw token, neither of which would prove the declared field won.
+      expect(result).toBe("You are in your-country");
     });
   });
 
--- a/apps/web/lib/utils/recall.test.ts
+++ b/apps/web/lib/utils/recall.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test, vi } from "vitest";
+import { mergeReservedValues } from "@formbricks/types/embedded-data-resolver";
 import { TResponseData, TResponseVariables } from "@formbricks/types/responses";
 import { TSurvey, TSurveyRecallItem } from "@formbricks/types/surveys/types";
 import { structuredClone } from "@/lib/pollyfills/structuredClone";
@@ -483,6 +484,95 @@ describe("recall utility functions", () => {
       const result = getRecallItems("Text with #recall:shared/fallback:x#", survey, "en");
 
       expect(result).toEqual([{ id: "shared", label: "Question headline", type: "element" }]);
+    });
+  });
+
+  describe("reserved fields in recall (ENG-1840)", () => {
+    test("a reserved token is labelled and typed, not left as a raw tag", () => {
+      // `getRecallItems` drops any id it cannot label, and the editor then renders the untouched
+      // `#recall:country/fallback:x#` as literal text. This is the test that catches that.
+      const survey = {
+        blocks: [],
+        hiddenFields: { fieldIds: [] },
+        variables: [],
+      } as unknown as TSurvey;
+
+      const result = getRecallItems("You are in #recall:country/fallback:your-country#", survey, "en");
+
+      expect(result).toEqual([{ id: "country", label: "Country", type: "reserved" }]);
+      expect(result[0].label).not.toContain("#recall:");
+    });
+
+    test("a declared field of the same name shadows the reserved entry", () => {
+      // The grandfather rule, label side: a survey that already declares `country` keeps showing its
+      // own field, so the author never sees two identically-named rows.
+      const survey = {
+        blocks: [],
+        hiddenFields: { fieldIds: ["country"] },
+        variables: [],
+      } as unknown as TSurvey;
+
+      const result = getRecallItems("You are in #recall:country/fallback:x#", survey, "en");
+
+      expect(result).toEqual([{ id: "country", label: "country", type: "hiddenField" }]);
+    });
+
+    test("an unknown reserved-looking name stays unresolved", () => {
+      const survey = {
+        blocks: [],
+        hiddenFields: { fieldIds: [] },
+        variables: [],
+      } as unknown as TSurvey;
+
+      expect(getRecallItems("#recall:notACatalogEntry/fallback:x#", survey, "en")).toEqual([]);
+    });
+  });
+
+  describe("the grandfather rule at value-resolution time (ENG-1840)", () => {
+    // These exercise `mergeReservedValues` — the one expression the guarantee rests on. Flip the two
+    // spreads there and the second test goes red:
+    //   pnpm --filter=@formbricks/web test lib/utils/recall.test.ts
+    const reserved = { country: "DE", url: "https://app.test/s/abc" };
+
+    test("a reserved token resolves from the projected value", () => {
+      const result = parseRecallInfo(
+        "You are in #recall:country/fallback:your-country#",
+        mergeReservedValues(reserved, {})
+      );
+
+      expect(result).toBe("You are in DE");
+    });
+
+    test("a survey declaring its own `country` resolves from response.data instead", () => {
+      const responseData: TResponseData = { country: "Declared answer" };
+
+      const result = parseRecallInfo(
+        "You are in #recall:country/fallback:your-country#",
+        mergeReservedValues(reserved, responseData)
+      );
+
+      expect(result).toBe("You are in Declared answer");
+      expect(result).not.toContain("DE");
+    });
+
+    test("shadowing is per-key: other reserved values still resolve", () => {
+      const result = parseRecallInfo(
+        "#recall:country/fallback:a# at #recall:url/fallback:b#",
+        mergeReservedValues(reserved, { country: "Declared answer" })
+      );
+
+      expect(result).toBe("Declared answer at https://app.test/s/abc");
+    });
+
+    test("a declared field that is present but empty still wins", () => {
+      // The respondent left the declared hidden field blank. That is an answer, and it must not fall
+      // back to reserved metadata just because the string is empty.
+      const result = parseRecallInfo(
+        "You are in #recall:country/fallback:your-country#",
+        mergeReservedValues(reserved, { country: "" })
+      );
+
+      expect(result).not.toContain("DE");
     });
   });
 
--- a/apps/web/modules/survey/components/element-form-input/components/recall-item-select.tsx
+++ b/apps/web/modules/survey/components/element-form-input/components/recall-item-select.tsx
@@ -5,6 +5,7 @@ import {
   FileDigitIcon,
   FileTextIcon,
   GaugeIcon,
+  GlobeIcon,
   HomeIcon,
   ListIcon,
   ListOrderedIcon,
@@ -17,7 +18,12 @@ import {
 } from "lucide-react";
 import { useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
-import { getDeclaredEmbeddedFields, listReadableFields } from "@formbricks/types/embedded-data-resolver";
+import {
+  RESERVED_FIELD_CATALOG,
+  getDeclaredEmbeddedFields,
+  listMidSurveyReservedEntries,
+  listReadableFields,
+} from "@formbricks/types/embedded-data-resolver";
 import { TSurveyElement, TSurveyElementId, TSurveyElementTypeEnum } from "@formbricks/types/surveys/elements";
 import { TSurvey, TSurveyHiddenFields, TSurveyRecallItem } from "@formbricks/types/surveys/types";
 import { getTextContent } from "@formbricks/types/surveys/validation";
@@ -63,7 +69,7 @@ export const RecallItemSelect = ({
   setShowRecallItemSelect,
   recallItems,
   selectedLanguageCode,
-}: RecallItemSelectProps) => {
+}: Readonly<RecallItemSelectProps>) => {
   const [searchValue, setSearchValue] = useState("");
   const { t } = useTranslation();
   const isNotAllowedElementType = (element: TSurveyElement): boolean => {
@@ -137,6 +143,30 @@ export const RecallItemSelect = ({
     [embeddedFieldEntries, recallItemIds]
   );
 
+  /**
+   * Reserved fields (ENG-1840). `listMidSurveyReservedEntries` applies both gates: it drops
+   * server-derived entries — recall renders while the respondent is still answering, so `country` or
+   * `durationSeconds` could only ever render as their fallback text — and it drops any entry this
+   * survey already declares under the same name, which would otherwise show twice with no way to
+   * tell the rows apart. Labels come from `listReadableFields`, the same enumerator the other groups
+   * use, so reserved rows are title-cased consistently rather than by a rule local to this file.
+   */
+  const reservedRecallItems = useMemo(() => {
+    const entries = listMidSurveyReservedEntries(
+      RESERVED_FIELD_CATALOG,
+      embeddedFieldEntries.map(({ key }) => key)
+    );
+
+    return listReadableFields({
+      blocks: [],
+      embeddedData: [],
+      reservedEntries: entries,
+      contactAttributeKeys: [],
+    })
+      .reserved.filter(({ key }) => !recallItemIds.includes(key))
+      .map(({ key, label }) => ({ id: key, label, type: "reserved" as const }));
+  }, [embeddedFieldEntries, recallItemIds]);
+
   const surveyElementRecallItems = useMemo(() => {
     const isWelcomeCard = elementId === "start";
     if (isWelcomeCard) return [];
@@ -162,17 +192,27 @@ export const RecallItemSelect = ({
   }, [elementId, elements, recallItemIds, selectedLanguageCode]);
 
   const filteredRecallItems: TSurveyRecallItem[] = useMemo(() => {
+    const allItems = [
+      ...surveyElementRecallItems,
+      ...hiddenFieldRecallItems,
+      ...variableRecallItems,
+      ...reservedRecallItems,
+    ];
     const query = searchValue.trim().toLowerCase();
-    if (!query) return [...surveyElementRecallItems, ...hiddenFieldRecallItems, ...variableRecallItems];
+    if (!query) return allItems;
 
     // Match the label's text content, not the label itself: an element's label is its raw headline HTML
     // (`<p class="fb-editor-paragraph">…`), so comparing against it made every query for question text
     // miss. `includes` rather than `startsWith` so a query also matches mid-headline words, and so it
     // still matches items whose displayed label is elided by the truncation below.
-    return [...surveyElementRecallItems, ...hiddenFieldRecallItems, ...variableRecallItems].filter(
-      (recallItem) => getTextContent(recallItem.label).toLowerCase().includes(query)
-    );
-  }, [surveyElementRecallItems, hiddenFieldRecallItems, variableRecallItems, searchValue]);
+    return allItems.filter((recallItem) => getTextContent(recallItem.label).toLowerCase().includes(query));
+  }, [
+    surveyElementRecallItems,
+    hiddenFieldRecallItems,
+    variableRecallItems,
+    reservedRecallItems,
+    searchValue,
+  ]);
 
   const getRecallItemIcon = (recallItem: TSurveyRecallItem) => {
     switch (recallItem.type) {
@@ -185,6 +225,8 @@ export const RecallItemSelect = ({
       }
       case "hiddenField":
         return EyeOffIcon;
+      case "reserved":
+        return GlobeIcon;
       case "variable": {
         const dataType = embeddedFieldEntries.find(({ key }) => key === recallItem.id)?.dataType;
         return dataType === "number" ? FileDigitIcon : FileTextIcon;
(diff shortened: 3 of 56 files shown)


apps/web/modules/survey/editor/lib/utils.tsx
```diff
@@ -1000,6 +1052,80 @@ export const getMatchValueProps = (
...
+        options: reservedOptions,
+      });
+    }
+
+    return {
+      show: true,
+      showInput: true,
+      inputType,
+      options: groupedOptions,
+    };
```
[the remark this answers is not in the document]
itsjavi: Fixed in 95f2378d6 — valid finding.

Right-hand candidates are now filtered by the reserved field's `dataType`, reusing the rules the element and variable branches already apply rather than inventing new ones:

- `number` → Rating / NPS / CSAT / CES, plus OpenText with `inputType: "number"`
- `date` → Date elements
- `boolean` → no elements (no element type answers with a boolean)
- variables narrow to `text` / `number`, so date and boolean fields offer none
- hidden fields stay in every list, matching how the number-variable branch already treats them as untyped strings

It also caught a case not in the finding: `reservedOptions` wasn't type-filtered either, so a `reserved.number` field could be compared against a `reserved.date` one. Same filter applied.

On the requested unit tests for number/date/boolean operands — not added, deliberately. `utils.tsx` has no test file, and `AGENTS.md` excludes `.tsx` from unit testing ("React components are covered by Playwright E2E tests instead"). The reserved picker is covered by `survey-reserved-fields.spec.ts`, which this PR adds.

---

apps/web/lib/utils/recall.test.ts
```diff
@@ -486,6 +487,95 @@ describe("recall utility functions", () => {
...
+    test("a declared field that is present but empty still wins", () => {
+      // The respondent left the declared hidden field blank. That is an answer, and it must not fall
+      // back to reserved metadata just because the string is empty.
+      const result = parseRecallInfo(
+        "You are in #recall:country/fallback:your-country#",
+        mergeReservedValues(reserved, { country: "" })
+      );
+
+      expect(result).not.toContain("DE");
+    });
```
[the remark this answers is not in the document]
itsjavi: Fixed in 95f2378d6 — the assertion was genuinely too weak; it passed just as well on an empty string or an unrendered raw token.

Applied the suggested equality assertion and ran it to confirm the expected output is real rather than assumed: `parseRecallInfo` renders `"You are in your-country"`, i.e. the declared-but-empty field wins by falling back to its own fallback text instead of leaking the reserved `DE`. That is exactly the property this test exists to pin, and it now fails if recall returns anything else. 63 tests pass in the file.

---

apps/web/modules/survey/editor/lib/utils.tsx
```diff
@@ -160,13 +164,45 @@ const getComputedFieldOptions = (localSurvey: TSurvey): TComputedFieldOption[] =
...
+  listMidSurveyReservedEntries(RESERVED_FIELD_CATALOG, getDeclaredFieldNames(localSurvey));
+
+const toReservedOption = (entry: TReservedFieldCatalogEntry): TComboboxOption => ({
+  icon: GlobeIcon,
+  label: formatSnakeCaseToTitleCase(entry.name),
+  value: entry.name,
+  meta: {
+    type: "reserved",
+  },
+});
```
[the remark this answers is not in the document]
itsjavi: Not changing this one — declining deliberately, with reasoning.

These labels are field **identifiers**, not UI copy, and the repo already establishes that identifiers are shown as-authored rather than translated:

- a variable named `score` renders as `score` in the same pickers
- a hidden field named `plan` renders as `plan`
- contact-attribute keys are labelled from the key itself (`listReadableFields` in `embedded-data-resolver.ts`)

Reserved fields are the same kind of thing — they are addressed by name in recall tokens (`#recall:country#`) and logic operands, and the label is a title-cased rendering of that address. Translating it would make the visible label diverge from the token the author must type, and would put `utm_source` / `pageUrl` through a translation pipeline where there is no meaningful translation to make.

So `formatSnakeCaseToTitleCase` is the intended treatment here, consistent with how `listReadableFields` already labels every other identifier group. `AGENTS.md`'s `t()` rule is about user-facing prose, which the surrounding group headings (`common.survey_data`, `common.questions`, `common.variables`) do go through.

Worth noting the presentation of this picker is owned by ENG-1853 in a later milestone; if a decision is made there to give reserved fields human-authored display names, that is the right place for it, and it would want real labels rather than translations of identifiers.

---

apps/web/modules/survey/editor/lib/utils.tsx
```diff
@@ -160,13 +165,58 @@ const getComputedFieldOptions = (localSurvey: TSurvey): TComputedFieldOption[] =
...
 
+/**
+ * Every name a survey already declares — the shadow list the grandfather rule filters against. Both
+ * declaration kinds count: an ingested field and a variable are equally capable of being named
+ * `country`, and either one resolves ahead of the reserved entry at read time.
+ */
+const getDeclaredFieldNames = (localSurvey: TSurvey): string[] => [
+  ...getDeclaredIngestedStorageKeys(localSurvey),
+  ...getComputedFieldOptions(localSurvey).map((variable) => variable.id),
+];
```
kiln-agent, line 179: **The shadow list is missing element ids.**

Element ids share the `response.data` namespace with hidden fields, and they are user-editable: `ZSurveyElementId` (packages/types/surveys/elements.ts:21) and `validateId`'s lenient path (packages/types/surveys/validation.ts:400) check only `FORBIDDEN_IDS`, which contains `source` but **not** `url`, `country`, `action` or `language`. `RESERVED_FIELD_NAMES` guards `ZEmbeddedData` names only, so it does not apply here either.

Concrete failure: an author renames an OpenText element's id to `url` in update-element-id.tsx. `getPickerReservedEntries` still returns the `url` entry, so the block-logic left-operand dropdown lists two options carrying the **identical** option value `url` — the question and the reserved "Url". Whichever is picked, the stored condition is `{ value: "url" }` and indistinguishable on re-render. At evaluation time `mergeReservedValues(reserved, localResponseData)` puts the answer over the reserved value, so a condition the author built as "survey URL contains X" silently tests the respondent's free-text answer.

Suggest including element ids in the shadow list (they are exactly the other keys `mergeReservedValues` lets win):

```suggestion
const getDeclaredFieldNames = (localSurvey: TSurvey): string[] => [
  ...getDeclaredIngestedStorageKeys(localSurvey),
  ...getComputedFieldOptions(localSurvey).map((variable) => variable.id),
  ...getElementsFromBlocks(localSurvey.blocks).map((element) => element.id),
];
```


---

apps/web/modules/survey/components/element-form-input/components/recall-item-select.tsx
```diff
@@ -137,6 +143,30 @@ export const RecallItemSelect = ({
...
+   * `durationSeconds` could only ever render as their fallback text — and it drops any entry this
+   * survey already declares under the same name, which would otherwise show twice with no way to
+   * tell the rows apart. Labels come from `listReadableFields`, the same enumerator the other groups
+   * use, so reserved rows are title-cased consistently rather than by a rule local to this file.
+   */
+  const reservedRecallItems = useMemo(() => {
+    const entries = listMidSurveyReservedEntries(
+      RESERVED_FIELD_CATALOG,
+      embeddedFieldEntries.map(({ key }) => key)
+    );
```
kiln-agent: Same gap as in `editor/lib/utils.tsx`: the shadow list passed here is `embeddedFieldEntries` (ingested fields + variables) only, so an **element id** colliding with a reserved name does not remove the reserved row — even though the element wins at both label time and value time.

Failure: a survey has an element whose id was set to `url` (legal — element renames check only `FORBIDDEN_IDS`). The dropdown still offers the reserved row "Url"; picking it writes `#recall:url/fallback:…#`. `resolveRecallItemLabel` (apps/web/lib/utils/recall.ts:100) checks elements *before* reserved, so the chip immediately re-renders as that question's headline rather than "Url", and at runtime `mergeReservedValues` resolves the token to the respondent's answer. The author picked a metadata field and silently got a question reference.

Passing the element ids alongside `embeddedFieldEntries.map(({ key }) => key)` fixes it in the same shape as the picker fix.

---

apps/web/modules/survey/editor/lib/utils.tsx
```diff
@@ -160,13 +165,58 @@ const getComputedFieldOptions = (localSurvey: TSurvey): TComputedFieldOption[] =
...
+const INPUT_TYPE_BY_DATA_TYPE: Record<TEmbeddedDataType, HTMLInputTypeAttribute> = {
+  string: "text",
+  number: "number",
+  boolean: "text",
+  date: "date",
+};
+
+const toReservedOption = (entry: TReservedFieldCatalogEntry): TComboboxOption => ({
+  icon: GlobeIcon,
+  label: formatSnakeCaseToTitleCase(entry.name),
```
kiln-agent: `formatSnakeCaseToTitleCase` splits on `_` only (packages/types/safe-identifier.ts:76), but every catalog name is **camelCase**. So multi-word entries label as `DurationSeconds` / `IpAddress` / `DeviceType`, never `Duration Seconds` / `Ip Address`.

Latent today (the four client-available entries are single words), but two consequences now:

1. The same expression labels recall chips (apps/web/lib/utils/recall.ts:115), and the ENG-1841 entries this PR says "light up automatically" — `pageUrl`, `utmSource`, `viewportWidth` — will surface to authors as `PageUrl`, `UtmSource`, `ViewportWidth`.
2. It makes the new Playwright guard vacuous. `apps/web/playwright/survey-reserved-fields.spec.ts:97` asserts `toHaveCount(0)` for the literals `"Duration Seconds"` and `"Ip Address"`, which no code path can ever emit — those two assertions would pass even if `durationSeconds`/`ipAddress` leaked into the mid-survey picker, which is the exact regression the spec exists to catch. (The PR body's QA checklist has the same wrong strings.)

Either add a camelCase split to the label helper, or give catalog entries an explicit `label`.

---

packages/surveys/src/components/general/survey.tsx
```diff
@@ -951,6 +963,39 @@ export function Survey({
...
+  const reservedValues = useMemo(
+    () =>
+      projectClientReservedValues(RESERVED_FIELD_CATALOG, {
+        surveyId: localSurvey.id,
+        language:
+          selectedLanguage === "default" ? (getDefaultLanguageCode(survey) ?? null) : selectedLanguage,
+        data: responseData,
+        variables: currentVariables,
+        ttc,
+        meta: { ...getWebSurveyMeta(), action },
```
kiln-agent, line 1009: **Security: the projected `url` is the raw `window.location.href`, query string included** — while the catalog entry declares `privacy: "redactQuery"` precisely because "a query string is where an identifier rides along". Nothing consumes `privacy` yet, and this PR is what first makes the value interpolable.

Path: `redirect-url-form.tsx` wraps the redirect URL field in `RecallWrapper`, so an author can insert `#recall:url#` there; `ending-card.tsx:94` then runs `replaceRecallInfo(urlString, recallValues, …)` before navigating. On a single-use / personalized link the href carries `suToken` — described in packages/types/surveys/validation.ts:70 as a credential — plus any prefill params. `LINK_SURVEY_SYSTEM_PARAMS` deliberately refuses to *capture* those into hidden fields; the reserved `url` read routes around that refusal and forwards the whole query string to a third-party redirect host (and its access logs / Referer chain).

Suggest either honouring `redactQuery` inside `projectClientReservedValues`, or stripping the query here before projecting (`url.origin + url.pathname`) — note `getWebSurveyMeta()`'s value is also what gets persisted, so a shared helper is probably cleanest.

---

apps/web/lib/surveyLogic/utils.ts
```diff
@@ -232,46 +237,58 @@ export const getUpdatedActionBody = (
...
+ * The reserved-field lookup map for a **persisted** response: the full catalog projected, then
+ * shadowed by the response's own data so a declared field of the same name still wins
+ * (`mergeReservedValues` is the single expression carrying that guarantee).
+ *
+ * Server-side every reserved field is knowable, so this reads `RESERVED_FIELD_CATALOG` whole rather
+ * than the mid-survey subset — the renderer's `projectClientReservedValues` exists precisely because
+ * that is *not* true in the browser.
+ */
+export const buildServerEmbeddedValues = (response: TEmbeddedValueResponse): TResponseData =>
+  mergeReservedValues(projectReservedValues(RESERVED_FIELD_CATALOG, response), response.data);
```
kiln-agent, line 250: `buildServerEmbeddedValues` has **no production call site** — `grep` finds only this definition and the tests. `evaluateLogic` from apps/web has exactly one production caller, `evaluateQuotas` (apps/web/modules/ee/quotas/lib/utils.ts:37), which passes five arguments, so `embeddedValues` is always `{}` and the new server-side `reserved` arm can never resolve anything.

That contradicts the PR body ("the server engine arm is still used by quotas and summaries") — quotas are the deliberate opt-out, and follow-ups don't evaluate conditions on this branch.

The gap is not purely cosmetic: `ZDynamicLogicFieldValue` was widened for *all* consumers, quota conditions included. A quota posted through the v3 API with `{ "type": "reserved", "value": "country" }` validates, saves, and then silently never matches on any response — exactly the silent-mismatch failure the `includeReservedFields` picker opt-out was introduced to prevent, just reachable via API instead of the UI.

Either thread `buildServerEmbeddedValues(response)` into `evaluateQuotas` (the response is in hand there), or reject `reserved` operands in the quota condition schema so the API can't accept one.

---

kiln-agent (commented):
Careful pass over the full diff (38 files). The core of the change is sound — `mergeReservedValues` has the spread order right, the `element`/`hiddenField` arms are correctly left reading `data` alone, both engines default `embeddedValues` to `{}` so existing callers are byte-identical, and extracting `evaluateConditionGroup` is behaviour-preserving (the `"conditions" in condition` check is strictly safer than the old `connector !== undefined` one on malformed groups).

Five findings, most-severe first — see inline comments:

1. **`getDeclaredFieldNames` omits element ids** (editor/lib/utils.tsx) — element ids are user-editable and only `FORBIDDEN_IDS`-guarded, so `url`/`country`/`action`/`language` are all legal element ids. The reserved entry is still offered, producing two dropdown options with the same option value, and the reserved operand silently evaluates the respondent's answer.
2. **Same omission in the recall picker** (recall-item-select.tsx) — picking reserved "Url" on such a survey renders the *question's* headline as the chip, because `resolveRecallItemLabel` checks elements before reserved.
3. **`formatSnakeCaseToTitleCase` doesn't split camelCase** — labels come out as `DurationSeconds`/`IpAddress`; the new Playwright spec's negative assertions on `"Duration Seconds"`/`"Ip Address"` can never match anything, so that guard would pass even if server-only fields leaked into the picker.
4. **Reserved `url` projects the full href incl. query** despite the entry's declared `privacy: "redactQuery"` — recalled into an ending-card redirect URL it forwards `suToken` and prefill params to a third-party host.
5. **`buildServerEmbeddedValues` has no production caller** — the server engine's `reserved` arm is unreachable, and an API-created quota with a reserved operand silently never matches.

Nit (no inline comment): logic-rule-engine.ts has two near-identical block comments introducing the reserved rule sets, and the boolean note is then repeated a third time above `reserved.boolean` — looks like a leftover from an edit.

---

apps/web/modules/survey/editor/lib/utils.tsx
itsjavi: Fixed — real gap, and the reasoning about *why* it is a gap is right: element ids are exactly the other keys `mergeReservedValues` lets win, so anything that can win at value time has to shadow at picker time.

Applied the suggested addition. `getElementsFromBlocks` was already imported in the file, so it is a three-line change, and it covers both operand sides at once: `getPickerReservedEntries` feeds the left-operand list and the right-operand list (`utils.tsx:1139`), so the duplicate-`url`-option case you describe is closed on both.

I widened the comment above it to say all three declaration kinds count, and why — the previous wording named only the two, which is how the third went missing.

---

apps/web/modules/survey/components/element-form-input/components/recall-item-select.tsx
itsjavi: Fixed — same shape as the picker fix. The shadow list here is now `embeddedFieldEntries` keys plus `elements.map(({ id }) => id)`, with `elements` added to the memo's dependency list (it was already computed at the top of the component from `localSurvey.blocks`).

The label-ordering detail you point out is the part that makes this more than cosmetic: `resolveRecallItemLabel` checks elements before reserved, so picking the reserved row produced a chip that immediately re-rendered as the question's headline. The author would have had no way to tell they had not got what they clicked.

---

apps/web/modules/survey/editor/lib/utils.tsx
itsjavi: Fixed, and the second consequence is the important one — those two `toHaveCount(0)` assertions were passing against strings no code path could emit, so they would not have caught the regression the step exists for. Verified before fixing: the labels really were `DurationSeconds` / `IpAddress`.

I did **not** add the camelCase split to `formatSnakeCaseToTitleCase`, which was my first instinct and would have been wrong. That helper also derives the stored display name for newly created contact attribute keys, and those genuinely arrive camelCase — `contacts.ts:427` dedupes `firstName` against `firstname` precisely because both show up in the wild. Splitting there would have renamed attribute keys as a side effect of a survey-editor fix.

So: a new `formatFieldNameToTitleCase` in `packages/types/safe-identifier.ts` that splits camelCase as well as `_`, used at the three reserved-label sites (`listReadableFields`, `toReservedOption`, `resolveRecallItemLabel`). The shared helper is untouched, and a test pins that it still returns `"FirstName"` so this cannot quietly leak into the contacts path later.

Also updated: the spec's strings and comments, and the PR body's QA checklist. Note the split has no *visible* effect on this branch — every client-available Group A entry is a single word — so it only surfaces once ENG-1841's `utmSource`/`pagePath` land, which is exactly the case you flagged. New tests: the formatter directly (including the digit-to-uppercase boundary) and a `listReadableFields` assertion that the catalog's camelCase names label as `"Duration Seconds"` / `"Ip Address"` / `"Response Id"`.

---

packages/surveys/src/components/general/survey.tsx
itsjavi: Fixed, and this was the finding worth the whole review. Confirmed the path end to end before changing anything: `redirect-url-form.tsx` does wrap the URL field in `RecallWrapper`, `ending-card.tsx` does run `replaceRecallInfo` over the URL before navigating, and the projected value was `window.location.href` verbatim — so on a single-use link the `suToken` would have been forwarded to the redirect host and into its access logs and `Referer` chain.

Took your first option: `redactQuery` is now honoured **inside the projection** — in `projectEntries`, the shared body of both `projectReservedValues` and `projectClientReservedValues` — so it applies to the server engine (follow-up email bodies interpolate recall too) and the renderer from one place, and a future `redactQuery` entry is covered by declaring its policy.

Two deliberate choices, both documented at the seam:

1. **Unconditional, not gated on the Anonymize toggle.** A projection is a forwarding surface in a way storage is not, so recall and logic only ever see origin + path regardless of the survey's settings. It costs an author nothing that was ever a documented capability — query params surface as their own `utm*` entries.
2. **Storage is untouched.** `response.meta.url` still holds the full href unless the Anonymize toggle redacts it at ingest. This is a read-side narrowing, so nothing about capture or history changes.

Tests: the client projection redacts `https://app.test/s/abc?suToken=secret&country=DE` to `https://app.test/s/abc` (the interpolable path you identified); the server projection does the same while the stored meta is asserted to still carry the query. The pinned catalog table needed updating too — its fixture URL was `?utm_source=news&email=a@b.co`, which is a fair illustration of the problem.

One follow-up I am **not** doing here: ENG-1842 (#8895) has its own `redactUrlQueryParams` for the ingest side. Same behaviour, including dropping the fragment, but it is a sibling PR stacked on the same parent, so it cannot import this one's helper. Once both are on the epic branch, the ingest side should point at `redactUrlQuery` in `packages/types` and the local copy go away. Noted in both PR bodies rather than left to be discovered.

---

apps/web/lib/surveyLogic/utils.ts
itsjavi: Fixed by threading it, and the correction to the PR body is fair — quotas were the one production caller and they were passing five arguments, so the server `reserved` arm could never resolve anything. I have removed the "still used by quotas and summaries" claim.

What landed:

- `evaluateQuotas` takes an optional sixth `embeddedValues` argument, defaulting to `{}`, and forwards it to `evaluateLogic`.
- `QuotaEvaluationInput` takes an optional `response`, and `evaluateResponseQuotas` builds `buildServerEmbeddedValues(response)` from it.
- All six callers of `evaluateResponseQuotas` pass it. Each already holds the row it just wrote inside the ingest transaction, so nothing is re-fetched — worth stating because an extra query there would be paid on every response.

Optional rather than required, so a caller without a row degrades to "reserved operands read as unset" instead of throwing — the same behaviour as an absent hidden field, which is what the rest of the engine does.

I did not take the reject-in-the-schema option: the API accepting a `reserved` quota condition is the desirable end state, and the operand was widened for all consumers deliberately. The gap was the wiring, not the schema.

Tests: `evaluateQuotas` forwards the argument; `evaluateResponseQuotas` with a response produces a real catalog projection (`country: "DE"`, `browser: "Chrome"`, `finished: "true"`, `durationSeconds: 120` — only `./utils` and the data loaders are mocked, so `buildServerEmbeddedValues` genuinely runs); and the existing call-shape assertions across the six call sites now pin that `response` is passed rather than silently dropped.

---

pandeymangg (approved):

PR #8897, opened 2026-08-17
feat(surveys): refuse reserved names for newly declared fields [ENG-1839]

itsjavi:
> **Stacked on #8892** (the Tier-1 reserved field catalog) — please merge that one first. This PR's base is `javier/eng-1839-reserved-field-catalog-tier-1-name-collision-rules`; GitHub will retarget it to `epic/embedded-data-v1` automatically once #8892 merges. The repo merges with merge commits, so no rebase is needed.

## What & why

Embedded Data **reserved** fields (`country`, `url`, `browser`, `duration…`) are system metadata readable by name on every response. If an author declares a field with one of those names, the two collide in the recall/logic namespace and the new field can never receive a value — `getHiddenFieldsFromSearchParams` refuses to capture a reserved param, so the field stays silently empty forever.

That guard was missing at the create path. ENG-1834 gave `validateId` an opt-in `requireSafeIdentifier` that rejects reserved names, but it is passed from **exactly one** place — the hidden-fields editor card. `ZSurveyHiddenFields` stays lenient by design, because the same schema parses surveys loaded from the database. So `PUT /api/v1/management/surveys/{id}` could still create a hidden field named `lang` or `country`, bypassing the editor guard entirely.

This adds the one create-time validation layer that the lenient load-time schema cannot be:

- **New** `packages/types/surveys/declared-field-guard.ts` — `validateNewDeclaredFieldNames({ existing, incoming })`. It delegates each new name to `validateId`'s strict branch, so the guard and the editor card cannot drift apart.
- **`validateId`'s strict branch now also consults `RESERVED_FIELD_NAMES`** (the Tier-1 catalog from #8892). The lenient branch is untouched.
- **Called at exactly three input boundaries, before the transaction in each case:**

  | Boundary | Covers | `existing` |
  | --- | --- | --- |
  | `updateSurveyInternal` | editor save, `PUT /api/v1/management/surveys/[surveyId]` | the survey's current variable names + `hiddenFields.fieldIds` |
  | `createSurvey` | templates, `POST /api/v1/management/surveys`, v3 create | `[]` — a create authors every name fresh |
  | `executeV3SurveyPatch` | `PATCH /api/v3/surveys/*`, the MCP write path | the loaded `currentSurvey`'s declared names |

### Grandfathering is the whole point

Surveys in production already have hidden fields named `country`, `url`, `source`, `browser`. Their values live at `response.data["country"]`, `#recall:country#` resolves from there, and **nothing is renamed**. A name already present in `existing` returns no error, whatever it is — the blocklist applies to names a write is *authoring*. Matching is case-insensitive in both directions.

Two deliberate non-changes, both load-bearing:

- `RESERVED_DECLARED_FIELD_NAMES` is **not** grown. It is also the capture-refusal list read by `getHiddenFieldsFromSearchParams`, so adding `country` there would stop `?country=DE` from filling the hidden field of a survey that legitimately declares `country` today.
- The guard is **not** inside `reconcileEmbeddedData`. It must return a validation error to the client rather than fail inside an interactive transaction, and the survey copy flow feeds a whole survey's fields in as "new" against zero existing rows — a guard there would make duplicating a grandfathered survey fail.

Payload presence is tested with `!== undefined`, not the `in` operator, mirroring `resolveDesiredEmbeddedFields`: every write seam builds one object literal with both keys spelled out, so a patch that never mentioned `hiddenFields` must not read as declaring the empty set.

## Linear ticket

Fixes ENG-1839 — https://linear.app/formbricks/issue/ENG-1839/reserved-field-catalog-tier-1-name-collision-rules

## How this was tested

| Behavior | Coverage | Result |
| --- | --- | --- |
| Reserved name refused for a **new** hidden field / variable | unit — `declared-field-guard.test.ts`, `service.test.ts`, `patch.test.ts` | ✅ |
| **Grandfather**: a survey that already declares `country` still saves, field left in place | **unit (red on main)** — see command below | ✅ |
| Grandfathered survey may still not add a *second* reserved name | unit — `service.test.ts`, `patch.test.ts` | ✅ |
| Case-insensitive matching (`Country`, `COUNTRY`, `deviceType`) | unit — `declared-field-guard.test.ts`, `validation.test.ts` | ✅ |
| `!== undefined` presence rule (payload omitting `hiddenFields` validates nothing) | unit — `declared-field-guard.test.ts` | ✅ |
| Lenient path unchanged, so stored surveys keep loading | unit — `validation.test.ts` | ✅ |
| **Duplicating** a survey that declares `country` still succeeds | integration (real Postgres) — `embedded-data-reconcile.integration.test.ts` | ✅ |
| Editor shows an error and does not save `country` | e2e — spec added, **not executed** (see Open gaps) | ⚠️ |
| **Variables card** refuses `country` before the save, with translated copy | e2e — spec added, **not executed** (see Open gaps) | ⚠️ |
| The refusal reason matches the list that refused the name | unit — `declared-field-guard.test.ts` | ✅ |
| Grandfathering is per-save: deleting the field makes the name refusable again | unit — `declared-field-guard.test.ts` | ✅ |

Commands actually run, with results:

```
# packages/types — 6 files, 263 tests
cd packages/types && npx vitest run                                   ✅ 263 passed

# apps/web unit — service, v3 patch, v3 create
cd apps/web && npx vitest run lib/survey/service.test.ts \
  app/api/v3/surveys/patch.test.ts app/api/v3/surveys/create.test.ts  ✅ 130 passed

# integration against real Postgres (18 tests, incl. the 3 new duplicate/grandfather cases)
cd apps/web && TEST_DB_PROVISIONED=1 TEST_DATABASE_URL={test db} \
  npx dotenv -e ../../.env -- npx vitest run \
  --config vitest.integration.config.mts \
  integration/embedded-data-reconcile.integration.test.ts             ✅ 18 passed

npx tsc --noEmit -p apps/web/tsconfig.json                            ✅ exit 0
npx tsc --noEmit -p packages/types/tsconfig.json                      ✅ exit 0
npx eslint {the 9 touched .ts files}                                  ✅ 0 errors
npx prettier --check {the 10 touched files}                           ✅ clean
```

**The grandfather row is red on main.** Delete the `existing` short-circuit in `packages/types/surveys/declared-field-guard.ts` — the line

```ts
if (grandfathered.has(lowered) || seen.has(lowered)) continue;
```

changed to `if (seen.has(lowered)) continue;` — then run:

```
cd packages/types && npx vitest run surveys/declared-field-guard.test.ts
```

Verified: **4 failed | 16 passed** — `a reserved name already in "existing" passes`, `a grandfathered survey may still not add a *different* reserved name`, `grandfathering is per name, not a blanket exemption for the payload`, and `"existing" grandfathers across casing too`. Restored, it is 20/20 green.

### Review round (pandeymangg)

Three findings, all addressed in `d5fdf0826`:

1. **The refusal reason was wrong for half the names it covered.** It said a reserved name "could never receive a value", which is true only of `RESERVED_DECLARED_FIELD_NAMES` — `getHiddenFieldsFromSearchParams` skips those. `RESERVED_FIELD_NAMES` is deliberately **not** in that list, so `?country=DE` does still fill a survey's declared `country`, and an integrator given the wrong reason could have removed URL params that work. `describeReservedReason` now branches on which set matched, with the capture-refusal reason winning for `source` (the one name in both).
2. **Variable names were not gated in the editor.** `isSafeIdentifier("country")` passes, so the variables card accepted the name and the author's first signal was the untranslated server message. The card now calls `validateId(..., { requireSafeIdentifier: true })` and reports through `getValidateIdErrorMessage(error, "variable", t)`, reusing `validate_id_reserved`. The check is skipped when an edit leaves the name unchanged — without that, a survey already declaring a variable called `country` would have become uneditable, breaking the grandfather rule in the editor by way of the change meant to enforce it.
3. **Grandfathering is per-save, and that was undocumented.** Deleting a grandfathered field spends the reprieve: the name is refusable again on the next write. Intended (remembering every name a survey ever had needs storage this layer has none of, and the survey gains the auto-captured field in exchange), now stated in the doc comment and walked by a test.

`packages/types` is 280 passed / 7 files after the round; `apps/web` `modules/survey/editor/lib/validation.test.ts` is 105 passed, covering the widened `getValidateIdErrorMessage` union.

**Blast radius, measured in production** (from the review, not by me): **8 surveys** declare a colliding name. That closes the optional read-only report ENG-1839 asked for, which I could not run from this sandbox.

One pre-existing test changed meaning rather than being deleted: `validation.test.ts` asserted `language` was accepted in strict mode. `language` is a Tier-1 reserved name, so it moved from the accepted list into the new rejected-in-strict-mode case, and the lenient path gained explicit assertions that `country` / `browser` / `language` still parse (which is what keeps stored surveys loading).

### Open gaps

- **The Playwright spec was not executed to completion.** `apps/web/playwright/survey-editor-embedded-fields.spec.ts` gained a `refuses a hidden field named after a reserved field` test (written, formatted, lint-clean). In this sandbox it reached the browser — Chromium launched, the user fixture was created, login succeeded — but the run then stalled on environment plumbing (the box exports a `DATABASE_URL` that overrides the repo `.env`, and the shared dev database is stuck on a failed migration unrelated to this branch). **No editor screenshot is attached for that row**, and the e2e row above is marked ⚠️ rather than ✅ for that reason. The equivalent behaviour is covered below the browser boundary by the service, v3-patch and integration suites; the editor toast copy asserted by the spec is the existing `validate_id_reserved` string, which this PR routes `country` into. Please run that spec (or the manual QA step below) before this leaves draft.
- **The edit-mode grandfather skip in the variables card is not covered.** Reaching that state through the editor is impossible by construction now (the card refuses to create the name), and seeding a survey that declares `country` means hand-building its `SurveyEmbeddedData` rows through Prisma — a fixture likely to fail for reasons unrelated to what it would test. Same boundary the hidden-field grandfather case documents. The skip fails *open* toward the author keeping their field, so the worst case if it were wrong is the refusal it prevents.
- No new i18n string was added. The editor refusal reuses the existing translated `workspace.surveys.edit.validate_id_reserved` copy (`Hidden field ID "country" is not allowed. It is a reserved keyword.`, and `Variable ID "country" …` for the variables card), which now covers reserved catalog names too; the server-side messages follow the existing non-localized `APP_SURVEY_TRIGGER_REQUIRED_MESSAGE` precedent for this layer. `pnpm i18n` was therefore not run — adding a key nothing reads would be dead copy. Happy to add a dedicated string if reviewers prefer distinct wording for system fields vs. other reserved keywords.

## Breaking changes

None for API/SDK request or response **shapes** — no field renamed, removed, retyped, or revalued; no endpoint or webhook payload changed.

One **behavioral** change worth calling out for integrators, which is the point of the ticket: a create or update that previously **accepted** a reserved field name is now **rejected**.

| Declaring a hidden field or variable named after a reserved field (`country`, `url`, `source`, `browser`, `os`, `deviceType`, `ipAddress`, `finished`, `language`, `responseId`, `surveyId`, `durationSeconds`, `startedAt`, `finishedAt`, `action`) | Accepted; the field was created and then never received a value | `POST`/`PUT /api/v1/management/surveys` → **400**; `POST`/`PATCH /api/v3/surveys` → **422** with a `forbidden_identifier` invalid param | API/MCP clients that create surveys with one of those names | Rename the field to a non-reserved name. **Surveys that already declare one keep working unchanged** — the refusal applies only to names being newly added |

## QA / Test Plan

**How to test**

- [ ] Editor → open a survey → **Hidden fields** card → type `country` → **Add hidden field ID** → an error toast reads `Hidden field ID "country" is not allowed. It is a reserved keyword.` and the field is **not** added to the card.
- [ ] Same card, type `Country` (capital C) → same refusal — matching is case-insensitive.
- [ ] Same card, type `team_size` → the field is added normally, and **Save** persists it. Confirms the card still works.
- [ ] Repeat the first step in the **Variables** card with the name `browser` → refused the same way.
- [ ] API: `PUT /api/v1/management/surveys/{id}` with `"hiddenFields": {"enabled": true, "fieldIds": ["country"]}` → **400**, body message names the field, and the survey is unchanged when you re-fetch it.
- [ ] API: `POST /api/v1/management/surveys` with the same `hiddenFields` → **400**, and no survey is created.
- [ ] API: `PATCH /api/v3/surveys/{id}` with the same `hiddenFields` → **422**, `invalid_params` contains a `country` entry with code `forbidden_identifier`, and the survey is unchanged.
- [ ] **Grandfather (the critical one)**: take a survey that ALREADY has a hidden field named `country` (create it on `epic/embedded-data-v1` before this PR, or insert one directly). Open it in the editor, change something unrelated (e.g. the survey name), **Save** → saves successfully, and the `country` field is still listed in the Hidden fields card afterwards.
- [ ] On that same grandfathered survey, submit a response with `?country=DE` on the link → the response still stores `country: "DE"` in `response.data`, and a `#recall:country#` token in a question still resolves from it.
- [ ] On that same grandfathered survey, try to add a **second** reserved name (`browser`) → refused, while `country` stays untouched.
- [ ] **Duplicate** a survey that declares `country` (survey list → ⋯ → Duplicate) → the copy is created successfully and also has the `country` field.
- [ ] Create a survey from a **template** (any template) → still creates normally, no name is refused.

**Preconditions / test data**

- A workspace with at least one survey, and one survey that already declares a hidden field named `country` (this is the grandfathered case — it has to be created before this PR's guard exists, or inserted directly).
- An API key for the v1 management endpoints, and one for v3.
- No feature flag, plan, or entitlement gates this. Applies equally to Cloud and self-hosted.

**Risks & regressions**

- **Grandfathered surveys must not break.** The blocklist applies to newly added names only; nothing is renamed and no stored value moves. Re-check that an existing survey with `country`/`url`/`source`/`browser` still saves, still captures `?country=…`, and still resolves `#recall:country#`.
- **Link-survey capture must not narrow.** `RESERVED_DECLARED_FIELD_NAMES` was deliberately **not** grown with the catalog names — it is also the capture-refusal list in `apps/web/modules/survey/link/lib/hidden-fields.ts`, so adding `country` there would stop `?country=DE` from filling a grandfathered survey's hidden field. Re-check link-survey URL param capture.
- **Survey duplication must not fail.** The guard is deliberately **not** inside `reconcileEmbeddedData` / `resolveDesiredEmbeddedFields`, and the copy flow at `apps/web/modules/survey/list/lib/survey.ts` is deliberately **not** guarded (it does its own `tx.survey.create` and never calls `createSurvey`). A duplicate of a grandfathered survey must keep succeeding — covered by the integration test.
- Templates: `apps/web/app/lib/templates.ts` declares `fieldIds: []` / `variables: []` and `minimal-survey.ts` declares `variables: []`, so no template carries a reserved name and guarding `createSurvey` cannot break template creation. Re-verify if templates change.
- Element and question **ids** are unaffected — only the strict (`requireSafeIdentifier`) path changed, so renaming a question to `Q1` still works.

- none

---

### DIFF
--- a/apps/web/app/api/v3/surveys/patch.ts
+++ b/apps/web/app/api/v3/surveys/patch.ts
@@ -2,6 +2,11 @@ import "server-only";
 import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
 import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
+import {
+  collectDeclaredFieldNames,
+  describeDeclaredFieldNameError,
+  validateNewDeclaredFieldNames,
+} from "@formbricks/types/surveys/declared-field-guard";
 import type { TSurvey } from "@formbricks/types/surveys/types";
 import { getActionClasses } from "@/lib/actionClass/service";
 import { reconcileEmbeddedData } from "@/lib/embedded-data/reconcile";
@@ -188,6 +193,25 @@ export async function executeV3SurveyPatch(params: {
     throw new V3SurveyReferenceValidationError([
       { name: "triggers", reason: APP_SURVEY_TRIGGER_REQUIRED_MESSAGE },
     ]);
+  }
+
+  // ENG-1839: a newly declared field may not take a reserved name. Before the transaction, so a
+  // refusal is a validation response rather than a rollback, and before `ensureV3WorkspaceLanguages`
+  // — which writes workspace languages — so a rejected patch creates nothing. Names `currentSurvey`
+  // already declares are grandfathered and pass untouched.
+  const declaredFieldNameErrors = validateNewDeclaredFieldNames({
+    existing: collectDeclaredFieldNames(currentSurvey),
+    incoming: collectDeclaredFieldNames(document),
+  });
+  if (declaredFieldNameErrors.length > 0) {
+    throw new V3SurveyReferenceValidationError(
+      declaredFieldNameErrors.map((error) => ({
+        name: error.field,
+        reason: describeDeclaredFieldNameError(error),
+        code: "forbidden_identifier" as const,
+        identifier: error.field,
+      }))
+    );
   }
 
   const languages = await ensureV3WorkspaceLanguages(currentSurvey.workspaceId, languageRequests, requestId);
--- a/packages/types/surveys/declared-field-guard.ts
+++ b/packages/types/surveys/declared-field-guard.ts
@@ -1,4 +1,5 @@
-import type { TSurveyHiddenFields, TSurveyVariables } from "./types";
+import type { z } from "zod";
+import { type TSurveyHiddenFields, ZSurveyVariables } from "./types";
 import { type TValidateIdError, TValidateIdErrorCode, validateId } from "./validation";
 
 /**
@@ -7,7 +8,15 @@ import { type TValidateIdError, TValidateIdErrorCode, validateId } from "./valid
  */
 export interface TDeclaredFieldSource {
   hiddenFields?: Pick<TSurveyHiddenFields, "fieldIds"> | null;
-  variables?: TSurveyVariables | null;
+  /**
+   * Only `name` is required, deliberately: this reads nothing else, and the write seams hand it both
+   * sides of the survey schema. `ZSurveyVariable` gives `value` a `prefault`, so a *create* payload
+   * (`ZSurveyCreateInput`) types `value` as optional while a parsed survey types it as required -
+   * demanding the *parsed* `TSurveyVariables` here made `createSurvey` fail to type check against
+   * its own input, so this is the schema's input side: `value` optional, which a parsed survey
+   * (value required) also satisfies.
+   */
+  variables?: z.input<typeof ZSurveyVariables> | null;
 }
 
 /**
--- a/packages/types/surveys/declared-field-guard.ts
+++ b/packages/types/surveys/declared-field-guard.ts
@@ -1,6 +1,12 @@
 import type { z } from "zod";
+import { RESERVED_FIELD_NAMES } from "../reserved-field-names";
 import { type TSurveyHiddenFields, ZSurveyVariables } from "./types";
-import { type TValidateIdError, TValidateIdErrorCode, validateId } from "./validation";
+import {
+  RESERVED_DECLARED_FIELD_NAMES,
+  type TValidateIdError,
+  TValidateIdErrorCode,
+  validateId,
+} from "./validation";
 
 /**
  * The legacy declared-field carriers of a survey write payload, as every write seam spells them:
@@ -61,6 +67,25 @@ export const collectDeclaredFieldNames = (source: TDeclaredFieldSource): string[
  *
  * Duplicate incoming names yield one error each at most — the caller sees one error per bad name.
  *
+ * ## Grandfathering is per-save, not permanent — and that is deliberate
+ *
+ * `existing` is the survey's *current* names, so the reprieve lasts exactly as long as the field does.
+ * Delete a grandfathered `country` field, save, and it can no longer be added back: the next write
+ * sees an `existing` set without it and refuses the name like any other new declaration. Editing the
+ * field, renaming other fields around it, or saving the survey untouched all keep it — only removing
+ * it spends the reprieve.
+ *
+ * That is the intended reading of "already declared". The alternative — remembering every name a
+ * survey ever had — would need storage this layer does not have, and would keep a name reserved for a
+ * survey that no longer uses it. A survey that gives up its declared `country` gains the auto-captured
+ * one in exchange, which is the field the name is supposed to mean from here on.
+ *
+ * The client-facing message says "fields a survey already has keep working", which is true of the
+ * survey as it stands and is what an integrator hitting this on a *different* survey needs to hear.
+ * Someone who deleted the field yesterday and is re-adding it today is the one case where that
+ * sentence reads as contradicting them; pinned by a test so the behaviour is a decision rather than an
+ * accident.
+ *
  * @param existing - Every declared field name the survey already has (variables + hidden fields).
  *   Pass `[]` on a create, which authors every name fresh.
  * @param incoming - The declared field names the payload carries, from
@@ -101,10 +126,32 @@ export const validateNewDeclaredFieldNames = ({
  * of this layer; the editor refuses these names client-side with a translated toast long before a
  * request is made.
  */
+const describeReservedReason = (field: string): string => {
+  // Order matters: a name in BOTH lists (`source` is the one Tier-1 field that is also a link-survey
+  // system param) gets the capture-refusal reason, which is the stronger and still-true statement.
+  //
+  // The two halves fail for genuinely different reasons, and saying "could never receive a value" for
+  // the catalog half would be actively misleading: `RESERVED_FIELD_NAMES` is deliberately kept OUT of
+  // the capture-refusal list read by `getHiddenFieldsFromSearchParams`, precisely so `?country=DE`
+  // keeps filling the field of a survey that already declares `country`. An integrator told the wrong
+  // reason here could go and remove URL params that work.
+  if (RESERVED_DECLARED_FIELD_NAMES.has(field.toLowerCase())) {
+    return "it is reserved by the link-survey URL contract, so a field declared under it is never filled from the URL and would stay empty";
+  }
+
+  if (RESERVED_FIELD_NAMES.has(field.toLowerCase())) {
+    return "it names an auto-captured system field that every survey can already read by name, so a second field under that name would be ambiguous in recall and logic";
+  }
+
+  // `validateId` classified this Reserved, so one of the two sets matched at the time. Reaching here
+  // means the sets and this description have drifted; say something true rather than guess which.
+  return "it is a reserved name";
+};
+
 export const describeDeclaredFieldNameError = (error: TValidateIdError): string => {
   const reason =
     error.code === TValidateIdErrorCode.Reserved
-      ? "it is a reserved name, read from every response, and a field declared under it could never receive a value"
+      ? describeReservedReason(error.field)
       : "it must start with a lowercase letter and contain only lowercase letters, numbers and underscores";
 
   return `Field name "${error.field}" cannot be used: ${reason}.`;
--- a/packages/types/surveys/declared-field-guard.ts
+++ b/packages/types/surveys/declared-field-guard.ts
@@ -1,0 +1,110 @@
+import type { TSurveyHiddenFields, TSurveyVariables } from "./types";
+import { type TValidateIdError, TValidateIdErrorCode, validateId } from "./validation";
+
+/**
+ * The legacy declared-field carriers of a survey write payload, as every write seam spells them:
+ * one object literal with both keys present, each either populated or `undefined`.
+ */
+export interface TDeclaredFieldSource {
+  hiddenFields?: Pick<TSurveyHiddenFields, "fieldIds"> | null;
+  variables?: TSurveyVariables | null;
+}
+
+/**
+ * The declared field names a payload actually *declares*.
+ *
+ * Presence is `!== undefined`, deliberately **not** the `in` operator — the same test
+ * `resolveDesiredEmbeddedFields` (apps/web/lib/embedded-data/reconcile.ts) uses, and for the same
+ * reason: every write seam builds one object literal with both keys spelled out and lets Prisma (or
+ * the reconcile) ignore the undefined ones. Under `in`, a patch that never mentioned `hiddenFields`
+ * would read as declaring the empty set, and the two modules would disagree about what a payload
+ * said — the reconcile carrying the current rows over while the guard treated them as absent.
+ *
+ * A `null` carrier is treated as absent for the same reason: it declares nothing to validate.
+ */
+export const collectDeclaredFieldNames = (source: TDeclaredFieldSource): string[] => [
+  ...(source.variables !== undefined && source.variables !== null
+    ? source.variables.map((variable) => variable.name)
+    : []),
+  ...(source.hiddenFields !== undefined && source.hiddenFields !== null
+    ? (source.hiddenFields.fieldIds ?? [])
+    : []),
+];
+
+/**
+ * Refuses reserved names for **newly declared** field names, and only for those.
+ *
+ * This is the create-time layer the lenient load-time schemas cannot be: `ZSurveyHiddenFields` and
+ * `ZSurveyVariables` also parse surveys read back out of the database, so they must keep accepting
+ * whatever is already stored. `validateId`'s strict mode is the same rule, but until now it was
+ * passed from exactly one place (the hidden-fields editor card) — so
+ * `PUT /api/v1/management/surveys/<id>` could still create a hidden field named `lang` or `country`
+ * that `getHiddenFieldsFromSearchParams` then refuses to fill, leaving it silently empty forever.
+ *
+ * ## Grandfathering is the whole point
+ *
+ * Surveys in production already declare `country`, `url`, `source`, `browser`. Their values live at
+ * `response.data["country"]`, `#recall:country#` resolves from there, and nothing may be renamed. So
+ * a name already in `existing` returns **no error**, whatever it is — the blocklist applies to names
+ * this write is *authoring*. Matching is case-insensitive in both directions, because
+ * `getHiddenFieldsFromSearchParams` refuses to capture a reserved param under any casing, and
+ * because `Country` and `country` would collide in the recall namespace all the same.
+ *
+ * Duplicate incoming names yield one error each at most — the caller sees one error per bad name.
+ *
+ * @param existing - Every declared field name the survey already has (variables + hidden fields).
+ *   Pass `[]` on a create, which authors every name fresh.
+ * @param incoming - The declared field names the payload carries, from
+ *   {@link collectDeclaredFieldNames}.
+ * @returns One error per refused new name, or `[]` when everything is allowed.
+ */
+export const validateNewDeclaredFieldNames = ({
+  existing,
+  incoming,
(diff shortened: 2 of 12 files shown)


apps/web/app/api/v3/surveys/patch.ts
```diff
@@ -190,6 +195,25 @@ export async function executeV3SurveyPatch(params: {
     ]);
   }
 
+  // ENG-1839: a newly declared field may not take a reserved name. Before the transaction, so a
+  // refusal is a validation response rather than a rollback, and before `ensureV3WorkspaceLanguages`
+  // — which writes workspace languages — so a rejected patch creates nothing. Names `currentSurvey`
+  // already declares are grandfathered and pass untouched.
+  const declaredFieldNameErrors = validateNewDeclaredFieldNames({
```
kiln-agent: The dry-run validator does not share this guard, so `POST /api/v3/surveys/validate` — and the MCP `validate_survey` tool, on the same write path this PR lists as covered — reports a document as valid that the write then refuses.

`validateV3Survey` (`apps/web/app/api/v3/surveys/lib/operations.ts:735` / `:765`) only calls `prepareV3SurveyCreateInput` / `prepareV3SurveyPatchInput` → `validateV3SurveyDocument`, none of which know about `validateNewDeclaredFieldNames`.

Scenario: an agent calls `validate_survey` with `{operation:"patch", surveyId, data:{hiddenFields:{enabled:true,fieldIds:["country"]}}}` → `valid: true`, no `invalid_params`. The identical `PATCH /api/v3/surveys/{id}` → 422 `forbidden_identifier`. That is exactly the round trip the dry-run exists to prevent, and an agent that trusts `validate_survey` will now happily submit a payload that cannot land.

`prepareV3SurveyPatchInput` already has both `survey` (for `existing`) and the merged document, so doing the check there (and in `prepareV3SurveyCreateInput` with `existing: []`) keeps it before the transaction while making the dry-run agree with the write.

---

packages/types/surveys/declared-field-guard.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (error) errors.push(error);
+  }
+
+  return errors;
+};
+
+/**
+ * One human-readable sentence for a set of refusals, for the error a write path returns to its
+ * client. Server-side and non-localized, matching `APP_SURVEY_TRIGGER_REQUIRED_MESSAGE` and the rest
+ * of this layer; the editor refuses these names client-side with a translated toast long before a
```
pandeymangg, line 126: This holds for hidden fields but not variables — `survey-variables-card-item.tsx` gates the name with `isSafeIdentifier` alone, and `isSafeIdentifier("country")` passes, so the editor accepts it and the author only finds out at save time via the untranslated server message. Pointing that card at `validateId(..., { requireSafeIdentifier: true })` + `getValidateIdErrorMessage` would close it and reuse the `validate_id_reserved` key that's already there.

packages/types/surveys/declared-field-guard.ts
```diff
@@ -0,0 +1,119 @@
...
+ * request is made.
+ */
+export const describeDeclaredFieldNameError = (error: TValidateIdError): string => {
+  const reason =
+    error.code === TValidateIdErrorCode.Reserved
+      ? "it is a reserved name, read from every response, and a field declared under it could never receive a value"
```
pandeymangg: "could never receive a value" is only true for the `RESERVED_DECLARED_FIELD_NAMES` half — `getHiddenFieldsFromSearchParams` skips those. The catalog names are deliberately not in that list, so `?country=DE` does still fill a `country` field, and an integrator who hits this could go remove URL params that actually work. Worth splitting the reason by which list matched.

packages/types/surveys/declared-field-guard.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Surveys in production already declare `country`, `url`, `source`, `browser`. Their values live at
+ * `response.data["country"]`, `#recall:country#` resolves from there, and nothing may be renamed. So
+ * a name already in `existing` returns **no error**, whatever it is — the blocklist applies to names
+ * this write is *authoring*. Matching is case-insensitive in both directions, because
+ * `getHiddenFieldsFromSearchParams` refuses to capture a reserved param under any casing, and
+ * because `Country` and `country` would collide in the recall namespace all the same.
+ *
+ * Duplicate incoming names yield one error each at most — the caller sees one error per bad name.
+ *
+ * @param existing - Every declared field name the survey already has (variables + hidden fields).
```
pandeymangg, line 89: Since `existing` is the survey's current names, deleting a grandfathered `country` field and saving makes it unaddable again — and the refusal then tells someone who had it yesterday that "fields a survey already has keep working". Reasonable if that's intended (they do pick up the reserved value in exchange), just worth a test and a note here either way.

pandeymangg (commented):
Went through the grandfathering path end to end and it holds up: `existing` comes off `selectSurvey` (which carries both `variables` and `hiddenFields`) at every call site, the guard runs before each transaction, the copy flow deliberately skips it, and `embeddedFields` is omitted from all three input schemas so there's no way around it. Good call keeping `RESERVED_FIELD_NAMES` out of `RESERVED_DECLARED_FIELD_NAMES` — that's what keeps `?country=DE` filling the field for the 8 surveys in prod that declare it. Three small things below, none blocking.

packages/types/surveys/declared-field-guard.ts
itsjavi: Right, and the consequence you name is the one that matters — an integrator told "could never receive a value" about `country` could go and strip URL params that are working today. The message was written for the `RESERVED_DECLARED_FIELD_NAMES` half and then applied to both, which is exactly the mistake the two-list split exists to avoid.

Split by which set matched, in `describeReservedReason`:

- `RESERVED_DECLARED_FIELD_NAMES` → "it is reserved by the link-survey URL contract, so a field declared under it is never filled from the URL and would stay empty" — true, since `getHiddenFieldsFromSearchParams` skips these.
- `RESERVED_FIELD_NAMES` → "it names an auto-captured system field that every survey can already read by name, so a second field under that name would be ambiguous in recall and logic" — the actual reason, with no claim about URL filling.

Checked first is the declared/system-param list, because `source` is in both and the capture-refusal statement is the stronger one that is still true of it. There is a third branch for "classified Reserved but in neither set", which can only mean the sets and this description have drifted — it says "it is a reserved name" rather than guessing.

Three tests pin the split, including the `source` overlap and a check that the catalog reason does **not** contain "never filled from the URL".

One thing I did not do: add a distinct `TValidateIdErrorCode` for the catalog half. That enum drives the translated editor messages through `getValidateIdErrorMessage`, so a new member means a new locale key and a `pnpm i18n` run for a distinction the editor does not currently draw. The server sentence was where the inaccuracy was.

---

packages/types/surveys/declared-field-guard.ts
itsjavi: Confirmed and intended, now documented and pinned rather than left implicit.

The behaviour: `existing` is the survey's current names, so the reprieve lasts exactly as long as the field does. Keeping it, editing it, renaming other fields around it, saving untouched — all fine. Delete it and save, and the name is new again on the next write.

Why that is the right reading rather than a gap: remembering every name a survey ever had needs storage this layer has none of, and it would keep a name reserved for a survey that no longer uses it. As you say, the survey gets the auto-captured `country` in exchange, which is what the name is supposed to mean from that point on.

Added a test that walks the whole sequence — the save that keeps it passes, the save that drops it passes, and the save that re-adds it is refused — so it reads as a decision in the test names rather than something a future change could quietly flip.

On the message: you are right that it reads badly for exactly this person. `describeDeclaredFieldNameErrors` still ends with "Fields a survey already has keep working; this applies to newly added names only", which is true of the survey as it stands and is what someone hitting this on a *different* survey needs. For the delete-then-re-add case it reads as contradicting them. I have left the sentence as-is and named the case in the doc comment instead: making it precise enough to cover both would need the message to know whether the name was just removed in this same payload, which the guard cannot see — it gets `existing` (post-read, pre-write) and `incoming`, not a diff of intent. Happy to add that plumbing if you think the copy matters more than the simplicity.

---

packages/types/surveys/declared-field-guard.ts
itsjavi: Verified and fixed — this was the real bug of the three. `isSafeIdentifier("country")` is `true` (lowercase, no separators), so the variables card accepted the name and the author's first signal was `describeDeclaredFieldNameErrors` at save time, untranslated. The doc comment claiming the editor refuses these client-side was true only for hidden fields.

Took your suggestion: the card's `validate` now calls `validateId(value, [], [], [], [], { requireSafeIdentifier: true })` and reports through `getValidateIdErrorMessage(error, "variable", t)`, reusing `validate_id_reserved`. `getValidateIdErrorMessage` gained a `"variable"` type mapping to `common.variable`, which is the string the card was already passing to `validate_id_not_safe_identifier` — so no new locale keys and no `pnpm i18n` run.

Two details worth flagging, since neither is in your comment:

**Empty id lists, deliberately.** `validateId` checks duplicates before reserved, so passing the survey's real ids would pre-empt this card's own duplicate messages — "already taken" and "conflicts with a hidden field" — which are more specific than the generic one. The card's duplicate checks stay where they are, just below. Same call shape the guard uses, for the same reason.

**Edit mode needed grandfathering, or the fix would have broken it.** The name field validates on blur in edit mode, so applying the gate unconditionally would refuse a survey's *existing* variable named `country` and block the author from changing its type or value — the grandfather rule violated in the editor by the very change meant to enforce it. The check is skipped when an edit leaves the name unchanged (case-insensitively, matching the server), so only a rename or a create is gated.

Coverage: a new Playwright test in `survey-editor-embedded-fields.spec.ts` — `country` refused inline, `Country` refused too, an ordinary name accepted, and the save writes one field.

The edit-mode skip is **not** covered: reaching that state through the editor is now impossible by construction, and seeding a survey that declares `country` through Prisma means hand-building the `SurveyEmbeddedData` rows, which risks a fixture that fails for unrelated reasons. Same boundary the hidden-field test documents for grandfathering. Noted under Open gaps rather than papered over — if you would rather have it, the seeding is a contained piece of work.

---

pandeymangg (approved):

PR #8898, opened 2026-08-18
chore: monorepo config audit cleanups [ENG-1695] [ENG-1696] [ENG-1697]

pandeymangg:
Fixes ENG-1695, fixes ENG-1696, fixes ENG-1697

## What & why

Three small cleanups from the monorepo configuration audit, one commit per ticket:

**Script matrix (ENG-1695)** — turbo runs a task only where the matching script exists and silently skips the rest, so packages with tests but no `test:coverage` script were invisible to the coverage pipeline.

- Added `test:coverage` + `@vitest/coverage-v8` + lcov reporter config to `database`, `email`, `logger` — the three packages that had tests but no coverage script.
- Documented the standard script set and the deliberate exceptions in `AGENTS.md` (config packages are script-less; `types` has no runtime logic to test; `email`/`types`/`vite-plugins` are consumed from source, so no `build`; storybook has no unit tests by policy).

**Shared tsconfig oddities (ENG-1696)**

- `base.json`: legacy `moduleResolution: node` → `bundler` (+ explicit `module: ESNext`); every consumer is bundled by Vite/Next, and the one CJS preset now pins its own resolution.
- `js-library.json`: display label was **"Node 16"** (a copy of node16.json's) → "JS Library"; dropped its own legacy resolution override and options duplicating base.
- `node16.json`: keeps classic resolution, now explicitly (`node10`), with the why documented in the package README (CJS consumers: `database`, `types`); the file itself stays strict JSON.
- Deleted `react-native-library.json` — nothing extends it, and the package is workspace-private so nothing outside the repo can.
- `apps/storybook/tsconfig.json` now extends the shared `react-library.json` instead of hand-rolling everything; only its genuine overrides remain.

**Config-package naming + READMEs (ENG-1697)**

- Renamed `@formbricks/eslint-config` → `@formbricks/config-eslint`: folders were already all `config-*` and 2 of 3 package names matched; this was the odd one out. One-line change in each of the 17 consumers.
- The stale Turborepo starter README the ticket mentions was already replaced during the eslint-tiers work; instead, the two config packages that had **no** README (`config-prettier`, `config-typescript`) got accurate ones.
- Review fix: the ESLint example in `docs/development/standards/practices/code-formatting.mdx` now calls the `library` factory instead of exporting it uncalled (pre-existing docs bug the rename surfaced).

## Breaking changes

None — internal tooling only; the renamed package is private to the workspace.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Rename resolves in all 17 consumers | manual | fresh `pnpm install`, then `turbo lint`: 21/21 green |
| `bundler` resolution compiles every consumer | manual | `turbo typecheck` 27/27; `turbo build --filter='./packages/*' --force` 12/12 — forced, since shared-tsconfig edits don't invalidate turbo's cache |
| New `test:coverage` scripts emit lcov | manual | ran for database/email/logger: 92 tests pass, `coverage/lcov.info` present in each |
| Storybook on the shared tsconfig | manual | `tsc --noEmit` and `storybook build` both pass |
| No wider regression | manual | full `pnpm test` (web: 610 files / 8,024 tests) and the web production build pass locally |

**Open gaps**

- The three packages now emit lcov but are **not** added to `sonar.sources` / `sonar.javascript.lcov.reportPaths` — that expands Sonar's quality-gate scope for everyone, so it's left as a team decision (happy to do it as a follow-up).

**Risks**

- Open branches touching a `package.json` that dev-depends on `@formbricks/eslint-config` will need the new name after rebasing; everyone needs a `pnpm install` after pulling.
- If anything downstream did rely on `node`-style resolution in the shared base, typecheck would be the failure mode — all 27 typecheck tasks and all forced builds are green.

---

> [!NOTE]
> **AI model used** — `claude-fable-5`, reasoning effort `default`.


mattinannt (approved):
Thanks for the PR :-) 

I also addressed the open gaps and created a ticket for updating the sonarconfig: https://linear.app/formbricks/issue/ENG-2432/decide-whether-to-include-package-coverage-in-sonar

PR #8899, opened 2026-08-18
feat(charts): add a breakdown-bars display option for pie charts [ENG-1558]

Dhruwang:
## What & why

A sentiment breakdown could only be drawn as a bar or a pie, both of which spend a lot of height on what is really a single distribution.

This adds a second rendering for pie charts: **Pie chart** or **Breakdown bars** — one horizontal bar split into a section per group, sized by each group's share of the total, with the count and share on hover. Same data, same query, a fraction of the height.

A pie and a split bar answer the same question — the share each group takes of the whole — so this is **one chart type with two renderings, not two chart types**. That is the change from the first version of this PR, which added a `sentiment` chart type; per review the type is gone and the rendering arrives as a display setting instead.

**Stacked on #8884**, which owns the Chart display settings section this option lives in. Review that one first; this PR's diff is only what sits on top.

Screenshot 2026-08-20 at 4 07 12 PM

Screenshot 2026-08-20 at 4 07 23 PM

https://github.com/user-attachments/assets/a8bc239b-b772-4965-b0fc-e959d2783dc0

### What the type removal takes with it

- The `ChartType` enum value and its migration.
- The entry in the chart type picker.
- **The query prefill.** Picking "Sentiment" used to populate the builder with response count grouped by sentiment, filtered to enriched records. A *display* setting that rewrote the query would repeat the confusion the type was faulted for — a control that quietly changes what the chart measures — so the prefill is dropped rather than moved. Building a sentiment breakdown now means choosing the count measure and the sentiment grouping, the same as any other chart. If that discovery step is worth keeping, it wants its own control (a template or a "suggested charts" entry), not this setting.

### What it keeps

The renderer, now `breakdown-bars.tsx`, unchanged in behaviour. It was already written for any single-measure grouped query, or a measure-only query where each measure becomes a section, so it needed no sentiment-specific logic — sections still pick up the sentiment scale colours when the query reads sentiment, via `getSemanticDimensionColor`.

Because the two displays are one chart type, they now agree on more than the data: `buildDistributionSegments` orders sections largest share first — the order `preparePieData` sorts slices into, and therefore the same palette handout — and shares print to one decimal, the precision `renderPieLabel` already used. Before that, toggling **Display as** moved and recoloured every group, and whole percents printed a real 0.4% section as `0%` while three equal groups added up to 99%.

`ZChartConfig` gains one optional field, `pieDisplay`, resolved and sanitized alongside `barOrientation`: the setting is dropped on save for any type that can't use it, so switching a breakdown pie to a bar chart doesn't leave it behind.

## Linear ticket

Fixes ENG-1558

## How this was tested

- `npx vitest run modules/ee/analysis` → **439 passed / 27 files** ✅
- `chart-utils.test.ts` covers the ordering directly: sections come out largest-first, and equal shares keep the caller's order so the palette handout stays stable.
- `chart-display.test.ts` gains 6 cases for the new setting: the default for charts saved before it existed, the saved value, that only a pie supports it, that each chart type keeps only its own setting, and that unrelated config survives sanitize.
- Tests for the removed machinery are gone with it — the `getChartTypePrefillQuery` suite and the prefill cases in `use-chart-dialog.test.ts`. The chart-type list tests are back to five types.
- `npx tsc --noEmit -p tsconfig.typecheck.json` → no errors in the touched modules ✅
- `npx eslint modules/ee/analysis` → clean, apart from one pre-existing `react-hooks/exhaustive-deps` warning in `chart-dropdown-menu.tsx`, which this diff doesn't touch ✅
- `pnpm i18n` ✅ — `chart_type_sentiment` removed, the three display strings added, all locales regenerated (0 missing, 0 unused)

**Screenshots pending.** The previous ones showed the type picker with a "Sentiment" tile, which no longer exists, so they have been removed rather than left to mislead. Replacements need the display setting exercised against a Cube-backed workspace: the setting itself, a pie and the same chart as breakdown bars, and a breakdown pie on a dashboard.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Analysis → Charts → **Create chart** → build a distribution (e.g. measure **Response count**, group by **Sentiment**) → pick chart type **Pie** → the **Chart display settings** card shows **Display as** with **Pie chart** selected.
- [ ] Switch to **Breakdown bars** → the preview becomes a single horizontal bar split into one section per group, sections in the same order and colours as the pie's slices.
- [ ] Hover a section → its group name, count and share of the total.
- [ ] Save, reopen the chart, and add it to a dashboard → still breakdown bars in the preview, the chart list and the widget.
- [ ] Switch the type from **Pie** to **Bar** → **Display as** disappears and **Bar direction** takes its place; save, reopen, switch back to **Pie** → back to **Pie chart** (the setting was not kept while it did not apply).
- [ ] A pie chart saved before this change → opens as a pie, unchanged.
- [ ] Edge case: measure-only query (e.g. the six per-sentiment count measures, no grouping) → each measure becomes its own section.
- [ ] Edge case: a query returning one group → a single full-width section, no crash.
- [ ] Edge case: a query returning no rows → "No data returned", no crash.
- [ ] Confirm the chart type picker has **five** tiles in a **single row**, and no "Sentiment" entry.
- [ ] Toggle **Display as** back and forth on a pie grouped by something without semantic colours (e.g. plan or survey) → every group keeps its position and its colour across the toggle.
- [ ] A group with a tiny share (e.g. 2 of 500) → reads `0.4%`, not `0%`, in both the tooltip and the legend, matching what the pie's slice label prints.

**Preconditions / test data**

- Dashboards + feedback directories entitlement enabled for the organization.
- A feedback directory with sentiment-enriched records, so a sentiment grouping has more than one section. Records are enriched asynchronously, so an unfiltered breakdown includes a "Not enriched" section — filter sentiment to *is set* to see only enriched ones.
- At least one dashboard to add a widget to.

**Risks & regressions**

- Pie charts generally — the pie branch of the renderer now chooses between two components. Existing pie charts must render exactly as before, including the centre total, slice labels and legend.
- `sanitizeChartDisplay` now handles two settings; check that saving a bar chart keeps its orientation and saving a pie keeps its display, and that neither leaks into the other.
- Dashboard widget sizing: a breakdown pie is short and wide, and the default widget layout no longer has a per-type entry for it (that went with the chart type). A newly added breakdown widget gets the standard 4×4, and since the renderer also dropped its `min-h-64` height floor, that widget is mostly empty space around one bar — the opposite of the overflow the height fix addressed. Resizing by hand is the workaround; a per-display default layout would need the display to be known at add-to-dashboard time.
- The chart type picker is back to one row of five tiles (it was three-per-row while a sixth type existed), so the builder's first card changes shape as well as content.
- The AI chart builder's schema context lists the chart types; it is back to five, so a prompt asking for a sentiment chart now returns a pie or bar rather than a dedicated type.

- none — the enum migration added earlier on this branch is removed, so there is nothing to apply or roll back.


packages/database/migration/20260818000000_add_sentiment_chart_type/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterEnum
+ALTER TYPE "public"."ChartType" ADD VALUE 'sentiment';
```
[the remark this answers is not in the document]
Dhruwang: Leaving this as a plain `ADD VALUE`, for consistency rather than disagreement about the semantics.

Every enum addition in `packages/database/migration` uses the bare form — 11 of them, `IF NOT EXISTS` in none — because that is what `prisma migrate diff` emits, so the next generated migration would drop the guard again. Replay is also handled a level up: the migration runner records applied migrations and skips them, so this statement does not run twice against the same database.

On the Squawk note about `BEFORE`/`AFTER`: `ChartType` carries no ordering semantics. Nothing reads the enum's declaration order — the chart type list and its display order live in `getChartTypes`, so appending is correct here.

Happy to add the guard if the team would rather have every migration idempotent regardless of the runner, but that feels like a repo-wide convention change rather than something to start in this PR.

apps/web/modules/ee/analysis/charts/components/sentiment-bar-chart.tsx
```diff
@@ -0,0 +1,137 @@
...
+                />
+                <TooltipContent>
+                  <div className="flex items-center gap-2">
+                    <div
+                      className="size-2.5 shrink-0 rounded-full"
+                      style={{ backgroundColor: segment.color }}
+                    />
+                    <span className="text-foreground text-sm font-medium">{segment.label}</span>
+                    <span className="text-muted-foreground text-sm tabular-nums">
+                      {valueText} ({percentText})
```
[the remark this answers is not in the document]
Dhruwang: Fixed in b310bd4936. All three templates — the tooltip, the legend and the section `aria-label` — now come from two keys, formatted once per section:

```json
"distribution_segment_label": "{label}: {value} ({percent})",
"distribution_value_share": "{value} ({percent})"
```

One correction worth recording, since it is a trap for the next person: the placeholders are **ICU single braces**, not i18next's `{{name}}`. `apps/web/lingodotdev/client.tsx` registers `i18next-icu`, so the double-brace form never interpolates. I wrote the keys that way first and the chart rendered a literal `{{value}} ({{percent}})` in the legend — caught it on a local run, not in review. (There are ~12 double-brace strings elsewhere in `en-US.json` that are inert for the same reason; out of scope here.)

Verified after the fix: the legend reads "Very negative 30 (12%)", "Negative 90 (37%)" and so on, and the sections expose accessible names like "Very negative: 30 (12%)". All 15 locales regenerated through `pnpm i18n`.

mattinannt (changes requested):
Implementation approach looks good to me, but I'm skeptical if this is the right solution from the product side @jobenjada 

it's a chart that is specifically only for sentiment; all the other charts that we have work with different kind of data and can adapt (labels, colors, values)

e.g. Qualtrics calls it [Breakdown Bar](https://www.qualtrics.com/support/vocalize/widgets/chart-widgets/breakdown-bar-widget/):

apps/web/modules/ee/analysis/charts/components/breakdown-bars.tsx
```diff
@@ -0,0 +1,145 @@
...
+  hasCategoryAxis,
+  xAxisKey,
+  formatDimensionValue,
+}: Readonly<BreakdownBarsProps>) {
+  const { t } = useTranslation();
+
+  let entries: TDistributionEntry[];
+  if (hasCategoryAxis) {
+    // Grouped query: one section per row, the first measure supplying the size.
+    entries = sortedData.map((row, index) => ({
```
kiln-agent, line 58: **Toggling "Display as" recolours and reorders every group.**

`preparePieData` sorts its rows **descending by value** before handing out `CHART_MEASURE_COLORS`, while these entries stay in row order and `buildDistributionSegments` hands out the same palette by *that* order. So for any dimension without semantic colours, the two renderings of one chart type disagree.

Concrete: a pie grouped by question type with rows `[Rating 10, NPS 30, CSAT 20]`.
- **Pie** → slices `[NPS, CSAT, Rating]` → NPS = teal, CSAT = blue, Rating = yellow.
- **Breakdown bars** → sections `[Rating, NPS, CSAT]` → Rating = teal, NPS = blue, CSAT = yellow.

Same query, same chart, every group changes both colour and position. That also makes the PR's own QA step — *"sections in the same order and colours as the pie's slices"* — fail, including for a sentiment pie (semantic colours match, but the pie sorts by count while this keeps scale order).

Worth picking one: either sort these entries the way `preparePieData` does, or factor the shared "drop non-positive, keep semantic colour, else next palette hue" step into one helper — that loop is currently duplicated verbatim in `preparePieData` and `buildDistributionSegments`.

---

apps/web/modules/ee/analysis/charts/components/manual-chart-builder.tsx
```diff
@@ -19,7 +19,7 @@ export function ManualChartBuilder({
   return (
     <div className="space-y-2">
       <div className="rounded-lg border border-gray-200 bg-white p-4">
-        <div className="grid grid-cols-5 gap-4">
+        <div className="grid grid-cols-3 gap-4">
```
kiln-agent: **Leftover from the removed sixth chart type.** The three-column grid arrived with the `Sentiment` type ("The chart type picker moves to three per row", e9aac54) when the list had six entries. The type is gone again but the grid stayed, so `getChartTypes` — back to five entries — now renders 3 tiles on the first row and 2 on the second with an empty third cell, where the base branch shows a single row of five.

The PR description says the picker is back to five tiles, but doesn't mention it also changes shape. Looks unintended:

```suggestion
        <div className="grid grid-cols-5 gap-4">
```

---

apps/web/modules/ee/analysis/charts/components/breakdown-bars.tsx
```diff
@@ -0,0 +1,145 @@
...
+  getSentimentMeasureColor,
+} from "@/modules/ee/analysis/charts/lib/chart-utils";
+import {
+  getMeasureAxisLabel,
+  sortMeasureIdsForCategoryAxis,
+} from "@/modules/ee/analysis/lib/schema-definition";
+import type { TChartDataRow } from "@/modules/ee/analysis/types/analysis";
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/modules/ui/components/tooltip";
+
+const formatPercent = (percent: number): string => `${Math.round(percent * 100)}%`;
```
kiln-agent: **Whole-percent rounding produces "0%" sections and shares that don't add up.**

Two concrete cases:
- 500 records with 2 in one bucket → the section *is* rendered (`min-w-[3px]`, hoverable, in the legend) but both its tooltip and its legend row read `2 (0%)`.
- three equal buckets → the legend reads `33%` / `33%` / `33%`, i.e. 99%.

The pie rendering of the same query uses `(percent * 100).toFixed(1)` in `renderPieLabel`, so it shows `0.4%` and `33.3%` — the two renderings of one chart type print different numbers for identical data. Matching `renderPieLabel`'s precision fixes both.

Separately, AGENTS.md wants user-facing numbers going through a locale-aware formatter rather than a hardcoded `%` suffix (`Intl.NumberFormat(locale, { style: "percent" })`), which also gets the digit grouping/decimal separator right for the non-English locales this string was just generated for.

---

apps/web/modules/ee/analysis/charts/lib/chart-utils.ts
```diff
@@ -144,6 +144,64 @@ export const prepareMeasureSliceData = (
     tooltipLabel: labelFor(key),
   }));
 
+/** One section of the single-bar distribution chart (the "sentiment" chart type). */
```
kiln-agent: Stale after the type removal — this PR deletes the `sentiment` chart type (enum value, migration, picker entry, prefill), so there is nothing in the codebase left to match this name. Whoever greps for it next will go looking for a `ChartType` that no longer exists.

```suggestion
/** One section of the single-bar distribution chart (a pie chart's "breakdown" display). */
```

The `CHART_SENTIMENT_COLORS` docstring a few lines above has the same problem — it still ends with "Groundwork for the sentiment-only chart (ENG-1558)".

---

apps/web/modules/ee/analysis/dashboards/lib/dashboards.ts
```diff
@@ -20,8 +20,12 @@ import {
 
 const MAX_NAME_ATTEMPTS = 5;
 
+const DEFAULT_WIDGET_LAYOUTS: Partial<Record<TChartType, TWidgetLayout>> = {
+  big_number: { x: 0, y: 0, w: 3, h: 2 },
+};
```
kiln-agent: This map earned its keep while the branch had a second per-type layout (the short-and-wide sentiment widget). With that type removed it holds exactly one entry and encodes the same rule as the ternary it replaced, but now a reader has to jump to a module-level constant to learn that `big_number` is the only special case — and `Partial<Record<TChartType, …>>` will happily keep a key for a chart type that no longer exists.

Suggest reverting to the one-liner and reintroducing the map when a second entry actually lands:

```suggestion
const getDefaultWidgetLayout = (chartType: TChartType): TWidgetLayout =>
  chartType === "big_number" ? { x: 0, y: 0, w: 3, h: 2 } : { x: 0, y: 0, w: 4, h: 4 };
```

(Related, and worth a line in **Risks** rather than a code change: the PR notes a breakdown pie now gets the standard 4×4 widget. Given the renderer just dropped its `min-h-64`, a 4×4 widget holding a single bar plus a legend is mostly empty space — the reverse of the overflow the last commit fixed.)

---

kiln-agent (commented):
Reviewed the diff on top of #8884 (base `dhruwang/eng-1550-add-chart-display-settings`, 26 files). The core mechanics hold up: the pie/breakdown branch in `chart-renderer` picks the same data shape the pie does (`hasCategoryAxis` lines up with `PieChartView`'s `useMeasureSlices` in every combination of dimension / time dimension / measure count), `sanitizeChartDisplay`'s new two-setting form is behaviour-preserving for `barOrientation`, `chartConfig` is seeded from the saved chart on open so `pieDisplay` survives a reopen, and the removal of the `sentiment` chart type is clean — no leftovers in the Prisma enum, `ZChartType`, the picker list, `getChartTypePrefillQuery`, or the locale files. `npx vitest run modules/ee/analysis/charts/{lib,hooks}` → 135 passed (the one failing suite is the documented stale `@formbricks/database` `dist/`, unrelated); eslint and prettier clean on the touched files.

Five findings, none blocking on its own, but the first two are worth fixing before merge:

1. **`breakdown-bars.tsx`** — palette colours and section order diverge from the pie's, so toggling `Display as` moves and recolours every group. This is the one QA step in the description that will fail as written.
2. **`manual-chart-builder.tsx`** — `grid-cols-3` is left over from the six-type list; five tiles now wrap 3 + 2.
3. **`breakdown-bars.tsx`** — whole-percent rounding: real sections read `0%`, three equal groups sum to 99%, and the pie prints one decimal for the same data.
4. **`chart-utils.ts`** — two doc comments still name the removed `sentiment` chart type.
5. **`dashboards.ts`** — single-entry layout map left behind by the type removal.

---

mattinannt (commented):
Looks much better now from the product-side; thanks for fixing :-)

Ready to merge from my side if all AI comments are addressed and merge conflicts are resolved.

apps/web/modules/ee/analysis/charts/components/breakdown-bars.tsx
Dhruwang: Fixed — sorted, in `buildDistributionSegments` rather than at the call site, so the ordering and the palette handout come from one place: sections are now largest share first, the order `preparePieData` sorts slices into. The sort is stable, so equal shares keep the caller's order (which is what keeps the sentiment-scale ordering meaningful for a measure-only query where several counts tie). `chart-utils.test.ts` asserts both. Left the duplicated "semantic colour else next palette hue" loop alone for now — three copies of it exist (`preparePieData`, `buildDistributionSegments`, `pivotMeasuresToCategories`) and factoring them out touches the pie and bar paths this PR otherwise doesn't.

apps/web/modules/ee/analysis/charts/components/manual-chart-builder.tsx
Dhruwang: Fixed — `grid-cols-5`, back to the single row the base branch shows. Called out in the description now, under Risks, since it changes the picker's shape and not just its contents.

apps/web/modules/ee/analysis/charts/components/breakdown-bars.tsx
Dhruwang: Fixed — `(percent * 100).toFixed(1)`, matching `renderPieLabel`, so both displays of one chart print the same share. Left the `Intl.NumberFormat(locale, { style: "percent" })` change out of this PR: the surrounding `distribution_value_share` / `distribution_segment_label` strings are what carry punctuation and word order per locale, and the numeral itself would want the same treatment in `renderPieLabel` and `formatCellValue` to be consistent — worth doing across the chart module in one go rather than on this one call site.

apps/web/modules/ee/analysis/charts/lib/chart-utils.ts
Dhruwang: Fixed — both docstrings. `TDistributionSegment` now says "a pie chart's \"Breakdown bars\" display", and the trailing "Groundwork for the sentiment-only chart (ENG-1558)" is gone from `CHART_SENTIMENT_COLORS` (the palette rationale above it still stands on its own).

apps/web/modules/ee/analysis/dashboards/lib/dashboards.ts
Dhruwang: Fixed — back to the ternary. The Risks note is in the description too: a 4×4 widget holding one bar and a legend is mostly empty space now that the height floor is gone, and a per-display default layout would need the display known at add-to-dashboard time.

PR #8902, opened 2026-08-18
fix(surveys): open overflowing welcome and CTA cards at the top [ENG-2289]

### LINKED ISSUE #8787: Welcome / CTA cards with overflowing text open scrolled to the bottom instead of the top
### Issue Summary

When a **welcome card** or a **Call to Action** question has enough text to overflow the card’s scroll area, the card loads already scrolled so respondents see the **end** of the text (near the continue / Next button), not the beginning.

Other overflowing question types (e.g. rating, open text) still open at the **top**, which is what you’d expect. Welcome/CTA cards also show the scroll-to-bottom chevron when content overflows, which implies the intended starting position is the top.

### Likely cause

Autofocus targets a button that sits **below** the long text, and that focus is allowed to scroll the button into view.

- **Welcome card** (`packages/surveys/src/components/general/welcome-card.tsx`): the start `SubmitButton` is after the headline/subheader and is focused via `focus={autoFocusEnabled}`. `SubmitButton` then calls `.focus()` (and sets `autoFocus`) without `preventScroll`, inside `ScrollableContainer`.
- **CTA as a question** (`block-conditional.tsx`): on mount, `shouldFocusOnMount` → `focusFirstControl`. For CTA, the first focusable control is usually the block **Next** button (non-external CTA has no input of its own), which is rendered **after** the element content, still inside the same `ScrollableContainer`. `focusFirstControl` uses `focus({ preventScroll: preferInvalid })`, so on normal mount `preventScroll` is `false` and the browser scrolls Next into view.

Other question types autofocus a control near the **top**, so they don’t jump to the end.

The modal focus trap already uses `focus({ preventScroll: true })` in `use-focus-trap.ts`; welcome/CTA mount focus does not.

### Suggested fix

Keep autofocus on the continue/Next button for keyboard/a11y, but focus with `preventScroll: true` (and/or reset the scroll container to `scrollTop = 0` after focusing) so long intro text still starts at the top.

### Expected Behavior

When a welcome or CTA card overflows, it should open scrolled to the **top** of the content. The continue/Next button may still receive focus for keyboard users, but focusing it should not scroll the long text out of view. Respondents should be able to read from the start and use the existing scroll-down control if needed.

### Steps to reproduce
1. Enable the survey **welcome card**, or add a **Call to Action** question as the first card.
2. Put a long headline/subheader (enough to exceed the card’s max height / scroll area).
3. Open the survey as a respondent (or use preview) so that card is shown on load.
4. Observe: the visible area starts near the bottom / button, not at the start of the text.
5. Compare with a long rating or open-text question that also overflows — those open at the top.
### Relevant files
- `packages/surveys/src/components/general/welcome-card.tsx`
- `packages/surveys/src/components/buttons/submit-button.tsx`
- `packages/surveys/src/components/general/block-conditional.tsx` (`focusFirstControl`, `shouldFocusOnMount`)
- `packages/surveys/src/components/wrappers/scrollable-container.tsx`

### Your Environment

Formbricks Cloud (app.formbricks.com)

itsjavi:
Fixes formbricks/formbricks#8787<br>Fixes formbricks/formbricks#8787

## What & why

A welcome card and a CTA block both move focus to a button that is rendered **below** the card's text and inside the same scroll area. Neither focus call passed `preventScroll`, so the browser scrolled that button into view: an overflowing card opened at the **end** of its intro text, with everything the respondent had yet to read scrolled out of view. Question types whose first control sits near the top (open text, rating) never showed the symptom.

Two independent focus paths caused it, so both are fixed:

* `packages/surveys/src/components/buttons/submit-button.tsx` — the delayed mount focus now uses `focus({ preventScroll: true })`, and the redundant `autoFocus` attribute is gone. The browser's autofocus step has no `preventScroll` knob, so it fired a *second* scroll on mount; the delayed effect (which waits out the card transition) is the intended focus path and now also clears its timer on unmount. This covers the welcome card and the ending card, which use the same button.
* `packages/surveys/src/components/general/block-conditional.tsx` — `focusFirstControl` now always focuses with `{ preventScroll: true }` instead of `{ preventScroll: preferInvalid }`. A non-external CTA renders no control of its own, so the block's Next button — after the element content — was the first focusable element on mount. Only the mount path changes behaviour: the `preferInvalid = true` callers (failed submit) already `scrollIntoView` the field they focus, so their scroll is unchanged and no longer competes with focus.

Both buttons still end up focused, so keyboard and screen-reader users keep their entry point on the card.

The shipped diff is those two files only. A Playwright spec was written, run green, and then removed on purpose — see Coverage.

## Recordings

> **Rendered from the built** `@formbricks/surveys` **bundle, not from a real** `/s/{surveyId}` **page.** No Docker<br>daemon here, so no database to boot the app against. Both sides are real compiled bundles driven in<br>Chromium — the "before" side built from `main`'s version of the two changed files, the "after" side from<br>this branch — with a long headline plus subheader in the `simple` card arrangement at 720×520. These are<br>captures of the **page loading, with no clicking, typing or scrolling**: two separate recordings of the<br>same fixture, aligned at page open and played side by side. The focus ring on the button is the autofocus<br>this PR keeps.

### Welcome card

Welcome card, before and after, on load

Both sides paint at the headline. Then the *before* side yanks itself to the very bottom — `scrollTop` 436 of 436 — leaving the respondent staring at sentences 26–40 and the focused start button, having read nothing. The *after* side stays where it painted, with the existing scroll-down chevron offering the rest.

### CTA card

CTA card, before and after, on load

Same divergence through the other focus path, and a starker failure: a card whose whole point is to be read before continuing ends up with the Next button already under the cursor (`scrollTop` 433 of 433). After, the headline "Read this before you continue" is what loads.

### Same two moments as stills

Same fixtures at 2× device scale, captured on load with no interaction, for anything the animation is too coarse to show.

**Welcome card**

| Before | After |
| -- | -- |
|  |  |

**CTA card**

| Before | After |
| -- | -- |
|  |  |

## Breaking changes

None

## Migrations & env

* none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| -- | -- | -- |
| Overflowing welcome card opens at the top, start button focused | e2e (run, then removed) | `apps/web/playwright/survey-card-scroll.spec.ts` ran **green in CI** — [run 32131113045](<https://github.com/formbricks/formbricks/actions/runs/32131113045>) on `0e17c55`, all four cases across the card and cardless arrangements. It asserted the fixture really overflows, that no scroll box (and not the page) is scrolled away from the top, that the headline is in the viewport, and that the button is focused. The spec was then deleted in `78630e8`: the E2E suite is reserved for larger features and too expensive to grow per bug fix |
| Overflowing CTA block opens at the top, Next button focused | e2e (run, then removed) | Same spec and same green run: it advanced past the welcome card with Space on the focused start button, then asserted the same invariants with focus on the block's Next button. An earlier revision of the spec failed on `f75b6e4` for a test-only reason — an element's headline is a `Label` from survey-ui's `ElementHeader`, not a heading, so `getByRole("heading")` never matched it; `0e17c55` matched it by text and the whole spec passed |
| Both cards render, and stay, at the top of their text on load | manual | The recordings and stills above, from the two compiled bundles. Welcome card `scrollTop` **436 → 0** (of 436 scrollable); CTA card **433 → 0** (of 433). Focus stayed on the start / Next button in every run — the fix is not "drop the autofocus" |
| Same fix in the cardless arrangement (the default for link surveys, also affected) | manual | Not pictured, measured: welcome card `scrollTop` **155 → 0** (client/scroll height 700/855). Card arrangement at a shorter viewport also checked: **232 → 0** (372/604) |
| Space (not Enter) advances the welcome card | manual | Same harness, counting clicks that reach the focused start button: Space → 1 click, Enter → 0. Enter is swallowed by the `onKeyDown` the welcome card passes to `SubmitButton` |
| Existing surveys unit suites | unit (guard) | `pnpm --filter @formbricks/surveys test` — 27 files, 668 tests, green, and green in CI on every commit here |

**Open gaps**

* **No automated regression guard ships with this fix.** The spec that covered it was deliberately removed (see above), and the repo's guidelines also rule out component unit tests for `.tsx`; the behaviour is real-browser scroll and focus, which jsdom cannot observe. If this regresses, the `scrollTop` numbers above are the repro recipe.
* The media comes from the surveys bundle rendered standalone, so it proves the card's own behaviour but not the full link-survey page around it. A pass on a real link survey is still owed by a human.
* The ending card uses the same `SubmitButton` and is fixed by the same change; it has neither media nor a case of its own.

**Risks**

* `focusFirstControl` is shared with the failed-submit path. That path is unchanged in effect (it already passed `preventScroll: true`), but if focus-after-invalid-submit ever stops scrolling the bad field into view, look at the callers' own `scrollIntoView` rather than at focus. `apps/web/playwright/survey-keyboard.spec.ts` covers that behaviour and stayed green.
* Dropping `autoFocus` means the button is focused \~200ms after mount instead of at mount. Anything that raced that attribute (e.g. an embedding page pulling focus immediately) now sees focus land slightly later.

---

> [!NOTE]
> **AI model used** — Claude Code, reasoning effort high.


jobenjada (approved):
looks good to me, thanks for adding visual proof :)

PR #8907, opened 2026-08-18
fix(auth): tag Better Auth error captures with the request endpoint (ENG-2259)

xernobyl:
Ref ENG-2259

## What & why

`FORMBRICKS-183` (`TypeError: Cannot read properties of null (reading 'id')`) reached ~242 events on every pod and **could not be triaged at all**: the Sentry event carried an exception and nothing else — no transaction, no URL, no route.

That is not a source-map problem. Better Auth's router discards the request context *before* our logger is called — for a non-`APIError` it does `ctx.logger?.error(e.name, e)` (`better-auth@1.6.23`, `dist/api/index.mjs:210`), so the message is the literal string `"TypeError"` and the endpoint is gone. `better-auth-observability.ts` then captured with no scope at all. ENG-2321 will give the throw *site*; this gives the *route*, and the two are independent.

- The `[...all]` route opens an `AsyncLocalStorage` scope carrying `{ path, method }`; the logger reads it and attaches it to `Sentry.captureException` and to the local log context (self-hosters have no Sentry and got the same untriageable line).
- Opening it in the **route** rather than in `hooks.before` also covers the middleware Better Auth runs ahead of the hooks — origin check, CSRF, rate limiter — which a before-hook stash cannot see.
- Tags don't affect grouping, so FORMBRICKS-183 stays one issue with `auth.path` as a breakdown facet. With ~25 endpoints the label alone should name the throw site even while frames stay minified.

### The path is never emitted raw — this is the part to review

Better Auth declares **`createAuthEndpoint("/reset-password/:token")`**: the password-reset token is a **path segment**, and better-call resolves both `request.url` and `ctx.path` to the *concrete* path, not the pattern (`better-call/dist/router.mjs:38` → `dist/context.mjs:20`). So the obvious implementation — and the one ENG-2259's description proposes — would have written a live account-takeover credential into Sentry and the logs:

| Source | Value on a reset-password request |
| --- | --- |
| `new URL(request.url).pathname` | `/api/auth/reset-password/<live token>` |
| `ctx.path` in `hooks.before` | `/reset-password/<live token>` |

Rather than deny-list the sensitive paths, the label vocabulary is derived from Better Auth's own registry: better-call stamps each endpoint function with its **declared** path (`dist/endpoint.mjs:56`), so `Object.values(auth.api).map((e) => e.path)` yields the route *patterns*. Only a parameter-free declared path is emitted verbatim; a concrete parameterized request can never exact-match its own `:param` pattern, so it degrades to the first segment (`reset-password`). A token-bearing endpoint added by a future upgrade — ENG-2343 is in flight — is covered by the same rule with no change here. The query string is never read at all (`?code=`, `?state=`, `?token=`).

`/oauth2/token`, `/oauth2/userinfo` and `/oauth2/authorize` stay distinct, which is what the ticket's MCP-OAuth lead needs; a first-segment-only scheme would have collapsed them.

Not a pre-existing leak: every current `ctx.path` use in the auth surface is a comparison, never an emission — so nothing to rotate. This change would have been the first to emit an auth path.

### Also corrected

`onAPIError.onError` is deliberately still **not** wired: Better Auth calls it and `return`s (`dist/api/index.mjs:194-197`), skipping the logger path and silencing this capture entirely. `route.ts` claimed that port had landed; it never did, and the comment is fixed.

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The label survives the real Better Auth router → `onError` → logger chain | integration (mutation) | `better-auth-error-context.integration.test.ts` drives the real handler with a fault injected into a before-hook. Mutation: dropping `runWithBetterAuthRequestContext` from `route.ts:43` → 2 of 3 cases red. Log output shows `authPath: "/sign-in/email"`, `authMethod: "POST"` |
| A `/reset-password/:token` fault never carries the token, to Sentry or the log | integration + unit (mutation) | Asserted on tags *and* the whole serialized capture context. Mutation: `literalPaths.has(authPath)` → `startsWith` at `better-auth-path-label.ts:63` → 3 unit + 1 integration case red |
| Query-string secrets are dropped | unit (mutation) | Same mutation covers it; `?code=`/`?state=` values asserted absent |
| Parameterized routes degrade for *any* declared pattern, not just the known one | unit (guard) | Loops every `:param` path in the fixture set, so a new one added by a Better Auth upgrade is covered without editing the test |
| Undeclared / unparseable / non-auth URLs bound tag cardinality | unit (guard) | `unknown`, incl. an encoded-traversal path |
| basePath subpath deployments still resolve | unit (guard) | `/formbricks/api/auth/sign-in/email` → `/sign-in/email` |
| Building the label vocabulary can never break the auth route | unit (mutation) | `Object.values(auth.api).map(...)` runs at **module load** in the catch-all, so a nullish entry would throw there and take down every `/api/auth/*` request, not just the tag. `route.test.ts`'s `auth.api` mock now carries a `null` entry, so importing the file at all asserts it. Mutation: `endpoint?.path` → `endpoint.path` fails the file to load with `TypeError: Cannot read properties of null (reading 'path')` |
| The route opens the scope around `auth.handler` | unit (mutation) | `route.test.ts` uses the **real** context modules, not mocks, so it proves the wiring |
| The ENG-2037 capture gate is unchanged | unit + integration (guard) | A handled `APIError` is still not captured (403, `captureException` never called) |
| The rules hold for the **real** registry, not just the fixture | integration (mutation) | Four guards over `auth.api` itself: vocabulary populated, no parameter in the first segment, every parameterized endpoint degrades, every path-less endpoint `SERVER_ONLY`. Verified at 1.6.23: 69 declared paths, 3 parameterized, 0 first-segment, 5 path-less (all server-only). Same prefix-match mutation turns the degrade guard red |
| The Better Auth message never reaches Sentry at all | unit (mutation) | The capture stays `Error` + bounded tags — `redactEmailsInLogMessage` strips emails only, so no message string is a safe payload. Asserts no `extra` key and that a token in a message never appears. Mutation: re-adding `extra` → 5 cases red |
| One label shape across the facet, and no literal/truncated collision | unit (mutation) | Every emitted value is rooted or `unknown`; `/reset-password` and `/reset-password/*` stay distinct. Mutation: returning a bare `firstSegment` → 4 cases red |
| A trailing slash does not split an endpoint across two facet values | unit (mutation) | `/api/auth/get-session/` → `/get-session`. Mutation: dropping the trim → red |
| No regression on the rest of the auth surface | unit (guard) | 395 tests / 39 files green across `modules/auth` + `app/api/auth` |

Also: `tsc --noEmit` clean, eslint clean, prettier clean, `gitleaks protect --staged` → no leaks.

**Open gaps**

- **The tag's real value only lands with the next production occurrence.** This is diagnostics; it does not fix the null dereference. Worth stating that a next event arriving *still untagged* is itself a finding — it would narrow the throw to something ahead of the router.
- `semgrep` is not installed locally and the shared `security-scan.sh` isn't present in this checkout, so the deterministic pass here was gitleaks plus a manual review against the data-exposure rules. The Sentry/logging surface is the only risk surface this diff touches.
- 11 pre-existing failures in the local `modules/auth` **integration** run (`headers()` was called outside a request scope) are unrelated to this change — identical count and set on a clean stash of `origin/main`, and none in the files touched here.
- Root `pnpm typecheck` is red locally for an unrelated reason: turbo does not forward the inline env vars, so `pnpm typegen` fails env validation before `tsc` runs. `tsc` itself passes.

**Risks**

- Tag names: `auth.path` (Better Auth endpoint) and `http.method` (the verb). The log context uses `authPath` / `httpMethod` — deliberately **not** `authMethod`, which already means `password` / `sso` in this same file's sign-in audit.
- The label is computed once at module scope from `auth.api`. If a future refactor makes `auth` lazy, the vocabulary would be built from an empty registry and every request would label `unknown` — degraded, never leaky, but the tags would quietly stop being useful. Now covered by a guard asserting the vocabulary is populated.
- The vocabulary derives from *declared* paths, so a routable endpoint without one would label `unknown`. Today every path-less endpoint is `SERVER_ONLY` and unreachable through `auth.handler`; better-call is moving toward path-less endpoints generally, so that invariant is asserted rather than assumed.
- `IS_PRODUCTION` is overridden in the new integration file to exercise the Sentry gate; it is kept in its own file so it cannot perturb the existing observability integration assertions.

<details>
<summary>Review passes — three self-review findings (<code>759e55e44</code>), four from review (<code>5fee78b89</code>), then four more plus two of my own (<code>55a0a6749</code>)</summary>

**Second review round (`55a0a6749`).** Four findings, each verified against the code before changing it.

An assertion that could not fail: `not.toHaveProperty("tags.auth.path")` reads the dotted string as a property *path*, so it looked for `tags` → `auth` → `path` — a nesting that never exists — and passed whether or not the tag was set. Reproduced against this repo's vitest with the flat key present: the dot form still passed, the array form found it. Now `["tags", "auth.path"]`. The `toHaveBeenCalledWith` above it remains the stronger check, so this line is documentation that is now also true rather than documentation that was false. Plus three doc-drift spots left by `5fee78b89`, where the emitted label became `/reset-password/*` but the prose still said a bare `reset-password` — including the field doc for the value the store carries, in a module that already had it right two other places.

**Two more from re-reviewing the whole diff.** `Object.values(auth.api).map((endpoint) => endpoint.path)` runs at module load in this catch-all, so one nullish entry would throw there and take down every auth request — the opposite of the principle this PR states in `better-auth-request-context.ts`, where the runtime path is explicitly best-effort. Now `endpoint?.path`, covered by the guard above. And two of the four citations into third-party internals go stale under the Better Auth 1.7 upgrade (#8835): better-call moved `internalHandler.path` from `endpoint.mjs:56` to `:55`, and `router.mjs:38` is now a different branch, so both cite the symbol instead. The two `better-auth/dist/api/index.mjs` citations were checked against 1.7.0 and are still exact. Useful by-product: the mechanism this PR rests on — the router logging a non-`APIError` without its endpoint, and `onAPIError.onError` returning before the logger — is byte-identical in 1.7.0, so this fix survives that upgrade and the two PRs need no sequencing.

**From review.** `extra.betterAuthMessage` dropped entirely rather than gated on `string`: the gate closed the object case but the string case was itself a new Sentry sink, and email-only redaction is not a safe basis for forwarding arbitrary strings. That also kept the module header's "an `Error` or nothing, never a string built from `message`" claim true, which a gated `extra` would have quietly falsified. Fallback labels are now `/<segment>/*` — rooted like the literals so one facet stays filterable, with the `*` keeping the truncated form distinct from a same-named literal that a plain `/<segment>` would have merged. Trailing slashes are trimmed before matching. And the HTTP verb moved off `authMethod`, which already means the *authentication* method in this same file.

Reviewed at `c23d943c9` against merge-base `12b8aa1a3`.

- **`extra` was unbounded in type.** `redactEmailsInLogMessage` only redacts strings, so an object message would have reached Sentry unredacted. Not live — Better Auth passes `e.name` / `e.message` / `e.status` — but the guarantee did not cover what was being fed to it, and the non-string case is the Error already captured. Gated on `string`.
- **The labeller's tests were fixture-bound.** They prove the rules, not that the rules hold for the endpoint set that actually ships. Added four guards over the real `auth.api`. This also answered an open question: 5 endpoints carry no declared path, and all 5 are `SERVER_ONLY`, so the vocabulary covers the whole HTTP surface.
- **`AUTH_BASE_PATH` duplicated** the literal in `oauth-urls.ts:5`. Considered importing it and rejected: that module reads `@/lib/env`, which would cost the labeller the no-dependency property that makes it exhaustively testable. Kept local with the reasoning recorded so it is not re-raised.

Considered and dropped: `new URL()` per request on the `/get-session` hot path (~µs against a session lookup, and it is what makes query-stripping correct); case-sensitive `indexOf("/api/auth")` (labels `unknown`, degrades safely); `indexOf` matching an `/api/authx/…` prefix (falls through to `unknown`); a secret reaching the first-segment fallback (impossible unless the secret equals a declared first segment); construction-time logs inheriting an unrelated request's store (auth is built at module load); `auth.method` cardinality (the route exports only GET and POST).

</details>


kiln-agent (commented):
## Review — ENG-2259 auth error-path tagging

Read the full diff and verified the load-bearing claims against the code on `main`. This is a tightly-scoped diagnostics change with unusually rigorous tests. No blocking findings.

### What I verified (rather than took on trust)

- **The security invariant holds across every branch of `createAuthPathLabeller`.** The emitted value is only ever a member of the declared allowlist (`literalPaths` or `knownFirstSegments`) or the `"unknown"` constant — a caller-supplied path segment or query value is never emitted. `/reset-password/:token` degrades to `reset-password`, `?code=`/`?state=` are dropped with the query string, and an undeclared first segment → `unknown`. A secret could only surface if it exactly equalled a declared literal path or first segment, which a random credential cannot. Encoded-traversal and double-`/api/auth` pathnames also fall through safely. Airtight.
- **The prose corrections in `route.ts` are accurate.** `onAPIError` is genuinely not configured anywhere in the auth surface (only referenced in comments), and `auditFailedAuthAfter` really is wired into `hooks.after` via `runAfterAuthHooks`. The old comment claiming the failure path landed via `onAPIError.onError` was indeed stale.
- **The `extra` string-gate is a real protection, not dead code.** `redactEmailsInLogMessage` (`better-auth-observability.ts:125-126`) passes non-strings through untouched, so `typeof safeMessage === "string"` is what stops an object message reaching Sentry `extra` unredacted. The updated deadlock test pins exactly that.
- **The tests can actually fail.** The integration test drives the real Better Auth router with an injected fault and asserts on the captured tags (with a `capturedFault()` gate that throws if nothing was captured); the registry-binding guards run over the live `auth.api`; the route test uses the real context modules rather than mocking the thing under test. No tautological assertions found.

### Non-blocking observation (optional)

- **Mixed `auth.path` tag format.** Declared literals are emitted with a leading slash (`/sign-in/email`), while the degraded first-segment fallback is emitted without one (`reset-password`, `oauth2`, `callback`). Both formats are intentional and tested, but they land as different values in the same Sentry facet — someone building a breakdown or filtering `auth.path:/reset-password` would miss the bare `reset-password` bucket. Normalizing the fallback to a leading slash (or documenting the two-format contract where the facet is consumed) would avoid that gotcha. Purely a dashboard-ergonomics nit; the diagnostic and security behavior is correct either way.

The PR's own "Open gaps" section already states the honest limitation that the tag only pays off on the next production occurrence and doesn't fix the underlying null dereference. Agreed on all of it.

---

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -159,7 +169,21 @@ export const betterAuthLogger: NonNullable<BetterAuthOptions["logger"]> = {
...
+          Sentry.captureException(cause, {
+            tags: {
+              component: "better-auth",
+              ...(request && { "auth.path": request.path, "auth.method": request.method }),
+            },
+            // Only when it is a string. `redactEmailsInLogMessage` passes non-strings through
+            // UNTOUCHED, so an object message would reach Sentry unredacted — and the only non-string
+            // Better Auth passes as `message` is the Error already being captured here, so there is
+            // nothing to gain by forwarding it.
+            ...(typeof safeMessage === "string" && { extra: { betterAuthMessage: safeMessage } }),
```
kiln-agent: The `string` gate closes the *object* leak the self-review describes, but it opens a narrower, adjacent one: this diff changes the capture from a bare `captureException(cause)` to one carrying `extra.betterAuthMessage`, so a **string** `message` now reaches Sentry where it never did before. `redactEmailsInLogMessage` only strips emails, so any non-email secret embedded in an error-level string message (a token, a URL, an identifier) lands in Sentry verbatim.

Better Auth core is safe here — its non-`APIError` path logs `(e.name, e)`, so `message` is just the class name (`"TypeError"`). The exposure is a plugin (or a future BA version) that logs `logger.error("… <token> …", err)`: the string passes the `typeof === "string"` guard and is forwarded. The comment on line 181-184 only justifies dropping the non-string case; it doesn't cover a secret carried in the string.

Note the content is already written to the local application log via `contextLogger.error(safeMessage)` on line 165, so this is incremental exposure to Sentry specifically, not a brand-new sink. Given how carefully the path labeller avoids emitting tokens, it seems worth either (a) dropping `betterAuthMessage` entirely — the `Error` is already the capture and `cause.message` rides along — or (b) noting in the comment why forwarding the redacted string is acceptable despite the email-only redaction.

---

apps/web/modules/auth/lib/better-auth-path-label.ts
```diff
@@ -0,0 +1,74 @@
...
+    // `indexOf` rather than `startsWith` so a Next.js basePath deployment (the app served from a
+    // subpath) still resolves to the Better Auth path instead of labelling every request `unknown`.
+    const baseIndex = pathname.indexOf(AUTH_BASE_PATH);
+    if (baseIndex === -1) return UNKNOWN_AUTH_PATH_LABEL;
+
+    const authPath = pathname.slice(baseIndex + AUTH_BASE_PATH.length);
+    if (literalPaths.has(authPath)) return authPath;
+
+    const firstSegment = getFirstSegment(authPath);
+    if (firstSegment && knownFirstSegments.has(firstSegment)) return firstSegment;
```
kiln-agent: The two emit paths produce inconsistently-shaped labels: a literal match returns `authPath` **with** its leading slash (`/sign-in/email`, `/oauth2/token`), while the first-segment fallback returns `firstSegment` **without** one (`reset-password`, `sign-in`). So in the `auth.path` facet, `/api/auth/sign-in/email` shows as `/sign-in/email` but `/api/auth/sign-in/passkey` degrades to `sign-in` — same endpoint family, two formats, and a reader can't tell a full-path label from a degraded first-segment label by shape alone. Returning `` `/${firstSegment}` `` on line 70 would make the facet uniform.

Related edge case on line 67: because the exact-match set is literal, a trailing slash silently degrades a known endpoint — `/api/auth/get-session/` isn't in `literalPaths` (which holds `/get-session`), so it falls through to the first-segment branch and labels `get-session` instead of `/get-session`. Minor, and may never occur in practice since BA emits canonical paths, but it adds avoidable facet noise if it ever does.

---

kiln-agent (commented):
Reviewed the ENG-2259 diff — Better Auth error captures tagged with the request endpoint. This is unusually careful work: the core security invariant (parameterized routes like `/reset-password/:token` never emit their parameter) holds because only parameter-free declared paths enter `literalPaths` and the fallback only emits a request's first segment when it matches a fixed declared first segment. The `AsyncLocalStorage` scope is wired correctly (opened in the route so it covers BA's pre-hook middleware), the `APIError` capture gate is unchanged, and the test coverage — including the four guards over the real `auth.api` registry and the mutation-based integration test — is thorough.

Two points worth a look before merge, neither a blocker; both are inline. Nothing else in the diff needs changes.

1. **`extra.betterAuthMessage` is a new Sentry sink for string log messages** (observability). Previously a bare `captureException(cause)`; now a redacted-emails-only string can carry a non-email secret to Sentry. Core BA is safe (`message` is `e.name`); the exposure is plugin/future-BA string messages.
2. **Label shape is inconsistent** between the literal branch (`/sign-in/email`, leading slash) and the first-segment fallback (`reset-password`, no slash), which muddies the `auth.path` facet the PR exists to create.

The self-review's dropped items (`new URL()` cost, case-sensitive `indexOf`, the first-segment-collision impossibility) were all reasoned through correctly — I didn't find anything to add there.

---

apps/web/modules/auth/lib/better-auth-observability.ts
kiln-agent: This new `extra: { betterAuthMessage: safeMessage }` makes the pre-existing header comment stale. Lines 113–114 (unchanged by this diff) still read:

> Only the logger needs this: since ENG-2037 (below) Sentry receives a real `Error` object or nothing, never a string built from `message`.

After this change that is no longer true — the Sentry capture now *does* carry a string built from `message`, and the redaction it describes is now load-bearing on the Sentry path too (which your inline comment right here correctly relies on). The risk isn't the code — it's that a future maintainer reading 113–114 could reasonably conclude the Sentry path never sees `message` and drop the `typeof safeMessage === "string"` gate, reintroducing exactly the unredacted-object exposure you just closed.

Worth updating the 113–114 comment to note that as of ENG-2259 the redacted `message` also reaches Sentry as `extra.betterAuthMessage`, so the string gate + email redaction now protect both sinks. Non-blocking.

---

kiln-agent (commented):
## Review — ENG-2259 auth error path tagging

Reviewed the full diff at head `759e55e4`. This is careful, well-reasoned work. I focused on what CI/ESLint/Sonar can't see: the security invariant of the labeller, whether the new tests can actually fail, tenant/cardinality bounds, and prose the diff may have invalidated.

### Security invariant holds
The central claim — that no path secret (the `/reset-password/:token` credential, OAuth `?code=`/`?state=`) is ever emitted — checks out, and is actually *more* robust than the PR description sells it as:
- The labeller only ever returns (a) an exact-match parameter-free declared path, (b) a **known first segment**, or (c) `unknown`. All three draw from a bounded, non-secret vocabulary, so Sentry tag cardinality is bounded too.
- Even the theoretical first-segment-parameter case is safe *by construction*, not just by the guard test: a declared `/:param/...` stores the literal `":param"` in `knownFirstSegments`, so a concrete request's real first segment can never match it and degrades to `unknown`. The "no declared parameter sits in the first segment" test is defense-in-depth, not the load-bearing protection — good.
- Query string is dropped via `new URL().pathname`; `indexOf` basePath handling degrades to `unknown`/first-segment on odd inputs without ever leaking. The `/api/authx/…` case degrades to a (harmless, unreachable) first-segment label rather than `unknown` as the self-review notes — but it's not routable and carries no secret, so immaterial.

### Tests can fail
The observability + route tests deliberately use the **real** context/labeller modules rather than mocking them, and the integration test drives the real Better Auth router with an injected fault — so they exercise the wiring, not their own arrangement. The described mutations (dropping `runWithBetterAuthRequestContext`, `startsWith` vs `has`) would genuinely turn them red. The real-registry guards are a nice touch against silent Better Auth upgrades.

### One non-blocking finding
- **Stale rationale comment** (`better-auth-observability.ts:113–114`, inline). The diff now forwards a redacted `message` string into Sentry `extra`, which contradicts the pre-existing "Sentry … never a string built from `message`" comment and makes the redaction load-bearing on the Sentry path. Worth a one-line update so a future maintainer doesn't remove the `typeof safeMessage === "string"` gate. Details inline.

Nothing here blocks merge.

---

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -149,7 +150,16 @@ export const betterAuthLogger: NonNullable<BetterAuthOptions["logger"]> = {
...
-    const contextLogger = logger.withContext({ source: "better-auth" });
+    // Endpoint label for the request being served, or undefined outside the HTTP handler (instance
+    // construction, a server-side `auth.api.*` call). Already reduced to a safe label — never a raw
+    // path, which on `/reset-password/:token` would be a live credential (better-auth-path-label.ts).
+    const request = getBetterAuthRequestContext();
+    const contextLogger = logger.withContext({
+      source: "better-auth",
+      // Self-hosters have no Sentry, so the label has to reach the application log too — otherwise
+      // their copy of this fault stays as untriageable as FORMBRICKS-183 was.
+      ...(request && { authPath: request.path, authMethod: request.method }),
```
kiln-agent: Minor (naming): `authMethod: request.method` puts the HTTP verb (`POST`/`GET`) under a key that already has an established meaning across the auth surface. `logAuthAttempt`/`logAuthSuccess` in `modules/auth/lib/utils.ts` emit `authMethod` to mean the *authentication method* (`password`, `totp`, `email_verification`, …). A self-hoster building a log query or dashboard on `authMethod` — the exact audience this branch adds the field for — would see HTTP verbs mixed into a field they expect to hold auth methods. Consider `httpMethod` (or reusing `method`) to avoid the overload. The Sentry tag `auth.method` is at least namespaced but carries the same semantic mislabel.

Low severity / arguably intentional given `BetterAuthRequestContext.method` is documented as "HTTP method" — flagging for your call.

---

kiln-agent (commented):
Reviewed the two-commit diff (9 files, +616/-8) across all review angles — correctness, security, simplification, efficiency, and test coverage.

**This is a well-reasoned change.** The security invariant (never emit the `/reset-password/:token` credential) holds: the labeller only emits a declared path verbatim on an exact match, and any concrete parameterized request degrades to its first segment, which can itself only ever be a `:param`-bearing (unmatchable) entry — so a caller-supplied secret can never surface in logs or Sentry. The PR's own "Considered and dropped" section already correctly dismisses every substantive candidate I generated (`indexOf` prefix/case behavior degrading safely to `unknown`; the module-load `auth.api` coupling, now guarded by a populated-vocabulary test; the non-string `extra` gate; first-segment-secret impossibility). Test coverage is thorough (96.5% on new code per Sonar).

No correctness, security, or efficiency defects found. One low-severity naming nit left inline.

---

apps/web/modules/auth/lib/better-auth-observability.ts
xernobyl: Good catch, and a better framing than mine — I was reasoning about the non-string case and missed that the string case is itself a new sink. The capture was previously `captureException(cause)` with no message at all, so any forwarded string is incremental exposure to Sentry, and email-only redaction is not a safe basis for forwarding arbitrary strings.

Took option (a) and dropped `betterAuthMessage` entirely in 5fee78b89, rather than documenting why the string is acceptable. It never earned its place: on the capture path `message` is either `e.name` (which Sentry already has from the Error) or a sentence accompanying the same Error, so the diagnostic value was ~zero against a real if narrow exposure. The tags carry the actual new information and are bounded by construction.

Pinned it so it does not creep back: a test now asserts the capture context has no `extra` at all and that a token embedded in an error-level message string never appears in it. Re-adding `extra` turns 5 cases red.

xernobyl: Right, and dropping `extra` for the finding above resolves this one at the source rather than by editing the comment — lines 113-114 are true again, since Sentry is back to receiving a real `Error` or nothing.

I did leave a note at the capture site recording that this is deliberate and load-bearing, precisely so the next person does not re-add a message payload and quietly falsify the header: "Keeping Sentry to `Error`-or-nothing is what makes the header note above stay true."

apps/web/modules/auth/lib/better-auth-path-label.ts
xernobyl: Both valid, both fixed in 5fee78b89.

On the shape: I went with `/${firstSegment}/*` rather than `/${firstSegment}`. A plain leading slash makes the facet uniform but merges two distinct endpoints — `/reset-password` is the POST that performs the reset, while the degraded value stands for the GET callback carrying a token, and they should not share a bucket. The trailing `*` keeps the facet rooted and sortable while still marking the label as truncated. A test now asserts every emitted value is either rooted or the `unknown` marker, and that the literal and truncated forms stay distinct.

On the trailing slash: agreed it is unlikely from Better Auth itself, but it costs one `replace` and the failure mode is silent facet-splitting, so the path is now trimmed before matching. `/api/auth/get-session/` resolves to `/get-session`; covered by a test, and dropping the trim turns it red.

apps/web/modules/auth/lib/better-auth-observability.ts
xernobyl: Agreed, and it is worse than flagged — `authMethod` carries the authentication method in this same file, not just in `utils.ts`: the sign-in audit at line 78 emits `newObject: { authMethod, sessionStrategy }` from `getSignInAuthMethod`, whose values are `password` / `sso`. So the collision was inside one module, and the local log context is precisely the branch that audience was added for.

Renamed in the latest commit: log context is now `httpMethod`, and the Sentry tag `auth.method` becomes `http.method`. Kept `auth.path` prefixed — the Better Auth endpoint genuinely is an auth-domain concept, whereas the verb is plain HTTP, so the split reads correctly rather than inconsistently.

apps/web/modules/auth/lib/better-auth-observability.test.ts
```diff
@@ -370,3 +377,78 @@ describe("betterAuthLogger (Sentry capture gating, ENG-2037)", () => {
...
+    const cause = fault();
+
+    log("error", "TypeError", cause);
+
+    expect(Sentry.captureException).toHaveBeenCalledWith(cause, {
+      tags: { component: "better-auth" },
+    });
+    // An untagged capture is itself diagnostic: it means the throw did not come through auth.handler.
+    const [, captureContext] = vi.mocked(Sentry.captureException).mock.calls[0];
+    expect(captureContext).not.toHaveProperty("tags.auth.path");
```
BhagyaAmarasinghe: This assertion can never fail, so it is not guarding what it looks like it guards.

Vitest and Jest treat a dotted string as a **property path**, so `"tags.auth.path"` resolves `tags` → `auth` → `path`. That nesting never exists; the actual key is `tags["auth.path"]`. The assertion therefore passes whether or not the tag is set.

Verified directly against this repo's vitest:

```
dot-form PASSED -> vacuous
array-form FAILED (correct): expected { tags: { …(2) } } to not have property "tags.auth\.path"
```

The behaviour *is* covered — `toHaveBeenCalledWith(cause, { tags: { component: "better-auth" } })` four lines up is a deep-equality check that an extra tag would fail. So nothing is currently untested. But given how much of this PR rests on assertions that provably go red under mutation, a line that cannot go red under any mutation is worth either fixing or deleting, so it does not get trusted later.

The array form escapes the dot:

```ts
expect(captureContext).not.toHaveProperty(["tags", "auth.path"]);
```

Deleting it is equally fine — the `toHaveBeenCalledWith` above is the stronger assertion of the two.

apps/web/modules/auth/lib/better-auth-request-context.ts
```diff
@@ -0,0 +1,41 @@
...
+ * handler — including the middleware that runs AHEAD of `hooks.before` (origin check, CSRF, rate
+ * limiter), which a before-hook stash cannot see. The store survives the awaited handler because the
+ * async work starts synchronously inside `run()`, the same guarantee `sso-request-context.ts` relies
+ * on.
+ *
+ * `path` is already reduced to a safe label by `createAuthPathLabeller` — never a raw path. See
+ * `better-auth-path-label.ts` for why that matters (`/reset-password/:token`).
+ */
+export interface BetterAuthRequestContext {
+  /** Safe endpoint label from `createAuthPathLabeller`, e.g. `/sign-in/email` or `reset-password`. */
```
BhagyaAmarasinghe: Stale after the `/<segment>/*` fix in `5fee78b89`: the fallback no longer emits a bare segment, so `reset-password` is not a value this field can hold.

This is the worst of the three drift spots because it is the field doc for the value the store carries — it tells a future reader that a bare segment is a valid `path`, which is exactly the shape the fix removed.

```suggestion
  /** Safe endpoint label from `createAuthPathLabeller`, e.g. `/sign-in/email` or `/reset-password/*`. */
```

apps/web/modules/auth/lib/better-auth-path-label.ts
```diff
@@ -0,0 +1,86 @@
...
+ * `createAuthEndpoint("/reset-password/:token")`, so on that endpoint the password-reset token is a
+ * PATH SEGMENT — and better-call resolves both `request.url` and `ctx.path` to the CONCRETE path, not
+ * the route pattern (`better-call/dist/router.mjs:38` → `dist/context.mjs:20`). Emitting either would
+ * write a live account-takeover credential into Sentry and the logs.
+ *
+ * The vocabulary is therefore derived from Better Auth's own endpoint registry rather than
+ * hand-maintained: better-call stamps each endpoint function with its DECLARED path
+ * (`better-call/dist/endpoint.mjs:56`), so every parameterized route arrives here containing `:` and
+ * can never be matched verbatim by a concrete request. `/reset-password/<token>` degrades to
+ * `reset-password`, and a token-bearing endpoint added by a future Better Auth upgrade is covered the
```
BhagyaAmarasinghe: Same drift from `5fee78b89` — the module header still describes the pre-fix label shape, and the file now contradicts itself: lines 16–17 and 45 say `reset-password`, while lines 28 and 78–79 correctly say `/reset-password/*`.

The header is the most-read doc in the module, so it is the one most likely to mislead.

```suggestion
 * can never be matched verbatim by a concrete request. `/reset-password/<token>` degrades to
 * `/reset-password/*`, and a token-bearing endpoint added by a future Better Auth upgrade is covered the
```

apps/web/modules/auth/lib/better-auth-path-label.ts
```diff
@@ -0,0 +1,86 @@
...
+const AUTH_BASE_PATH = "/api/auth";
+
+/** Emitted when the URL is unparseable or names no endpoint we serve. Bounds tag cardinality. */
+export const UNKNOWN_AUTH_PATH_LABEL = "unknown";
+
+const getFirstSegment = (path: string): string | undefined => path.split("/")[1] || undefined;
+
+export const createAuthPathLabeller = (declaredPaths: Iterable<string>): ((url: string) => string) => {
+  // Only parameter-free declared paths may be emitted verbatim. Patterns still contribute their first
+  // segment, so `/reset-password/<token>` degrades to a recognized `reset-password` rather than to
```
BhagyaAmarasinghe: Third drift spot. The mechanism this comment describes is right — a pattern contributes its first segment so the endpoint is still named rather than collapsing to `unknown` — but the emitted value is `/reset-password/*`, not `reset-password`.

```suggestion
  // segment, so `/reset-password/<token>` degrades to a recognized `/reset-password/*` rather than to
```

BhagyaAmarasinghe (changes requested):
Reviewed at `49cae35eb` against merge-base `12b8aa1a3`. CI is green across the board and the two prior review rounds are addressed.

**The security design holds.** I tried to break the labeller rather than read it, and could not. Every value it can emit is drawn from `{declared literal paths} ∪ {/<knownFirstSegment>/*} ∪ {"unknown"}`, all built at module load from `auth.api`, none of it from request input. Percent-encoded slashes (`/reset-password%2F<token>`) fall to `unknown` rather than leaking, because the fallback requires an *exact* segment match; the query string is never read; cardinality is bounded at ~95 values. Deriving the vocabulary from the registry instead of a deny-list is the right call, and the guards over the real `auth.api` — no first-segment parameter, every parameterized endpoint degrades, path-less endpoints all `SERVER_ONLY` — are what make it survive a dependency bump rather than merely pass today.

Also confirmed: every existing `ctx.path` use in the auth surface is a comparison inside `getSignInAuthMethod`, never an emission, so "nothing to rotate" is accurate. And `httpMethod` over `authMethod` is correct and non-obvious — `authMethod` genuinely carries `password`/`sso` in the sign-in audit at `better-auth-observability.ts:78`.

Requesting changes on two items, neither of which affects behaviour:

1. **A test line that reads as a guard and cannot fail** (`better-auth-observability.test.ts:430`). Verified empirically, not inferred.
2. **Comment drift from the `/<segment>/*` fix** in three places, one of which is the field doc for the value the store carries. `better-auth-path-label.ts` now contradicts itself — lines 16 and 45 say `reset-password`, lines 28 and 78–79 say `/reset-password/*`. A prior round on this PR was itself about a header comment the diff had falsified, so flagging the same class rather than letting it land.

Two notes that need no change:

- **The `/api/authx/…` line in the PR description is wrong.** "Considered and dropped: … `indexOf` matching an `/api/authx/…` prefix (falls through to `unknown`)" — it doesn't. `indexOf` matches at 0, `slice(9)` yields `x/sign-in/email`, and `split("/")[1]` is `"sign-in"`, a known segment, so it labels `/sign-in/*`:

  ```
  "https://x.com/api/authx/sign-in/email"     -> "/sign-in/*"
  "https://x.com/api/authorize/sign-in/email" -> "/sign-in/*"
  ```

  Not a bug — Next routes only `/api/auth/**` to this handler, so it is unreachable, and the output is still from the bounded vocabulary either way. But `createAuthPathLabeller` is exported, and a recorded rationale that is false is worse than no rationale, because the next reviewer trusts it. Worth correcting in the description (the honest version is "unreachable through the route", not "falls through to `unknown`").
- `better-auth` is pinned exact at `1.6.23`, so the `dist/api/index.mjs:210`-style citations are accurate today and a bump is a deliberate act. Cheap hardening: put the version next to the line number so whoever bumps it knows what to re-verify.

**What I could not run:** the suite. This worktree has no root `node_modules`, so `vitest` fails on tsconfig resolution before collecting. Relying on the green CI runs for the assertions themselves; the two findings below were verified independently of the suite.

**What looks good:** opening the scope in the route rather than `hooks.before`, which structurally covers the origin check, CSRF and rate limiter that a before-hook stash cannot see; the integration test driving the real router and asserting on the whole serialized capture context rather than just tags; dropping `extra` entirely instead of gating it on `string`, since email-only redaction is not a basis for forwarding arbitrary strings; keeping `onAPIError.onError` unwired with the reason recorded; and isolating the `IS_PRODUCTION` override to its own file so it cannot perturb the existing observability integration assertions.

apps/web/modules/auth/lib/better-auth-observability.test.ts
xernobyl: Confirmed and fixed in 55a0a6749 — you were exactly right, and I verified it before touching anything rather than take it on trust.

Reproduced against this repo's vitest with the flat key **present**:

```
not.toHaveProperty("tags.auth.path")      -> PASSED   (vacuous: looks for tags -> auth -> path)
toHaveProperty(["tags", "auth.path"])     -> PASSED   (array form finds the flat key)
```

So the dot form passes whether or not the tag is set, precisely as you described.

Went with the array form rather than deleting, because the line is the only place that states the claim in the test's own words — but your analysis of which assertion is load-bearing is right and I have not pretended otherwise: the `toHaveBeenCalledWith` above is a deep-equality check that any extra tag fails, so it is what actually catches a leak. This line is now documentation that is *also true*, instead of documentation that was false. I added a comment saying why the array form is required, since the trap is invisible at a glance.

Mutation-checked the pair: making the tag leak when there is no request context (`"auth.path": request ? request.path : "leaked"`) turns this file red — 4 tests, including this one.

apps/web/modules/auth/lib/better-auth-request-context.ts
xernobyl: Fixed in 55a0a6749, applied as suggested.

Agreed this was the worst of the three: it is the field doc for the value the store actually carries, so it taught a reader that a bare segment is a legal `path` — the exact shape `5fee78b89` removed. Verified against the emitting code first (`return \`/${firstSegment}/*\``), so the doc now matches what the function can return.

apps/web/modules/auth/lib/better-auth-path-label.ts
xernobyl: Fixed in 55a0a6749, applied as suggested.

You were right that the file contradicted itself — lines 28 and 78-79 already had `/reset-password/*` while the header did not, and the header is what gets read first.

I also made two nearby citations version-proof while I was in this header, which turned out to matter: under the Better Auth 1.7 upgrade (ENG-2343, #8835) better-call moves `internalHandler.path` from `endpoint.mjs:56` to `:55`, and `router.mjs:38` becomes a different branch. Both now name the symbol instead of the line. The two `better-auth/dist/api/index.mjs` citations I checked against 1.7.0 and left alone — they are still exact, line numbers included.

apps/web/modules/auth/lib/better-auth-path-label.ts
xernobyl: Fixed in 55a0a6749, applied as suggested.

Your reading of the mechanism was right — the pattern contributes its first segment so the endpoint stays named rather than collapsing to `unknown` — and only the emitted value was wrong. It now says `/reset-password/*`, which is what the line below it actually returns.

xernobyl:
All four review threads addressed in 55a0a6749 and resolved, each verified against the code before changing it rather than applied on trust. Thanks — the vacuous-assertion catch was a good one, and the reasoning about *which* assertion is load-bearing was right.

Then re-reviewed the whole diff and ran a security pass. **Two further findings, both mine, both fixed in the same commit:**

**1. The label build could take down all of `/api/auth/*`.** `Object.values(auth.api).map((endpoint) => endpoint.path)` runs at **module load** in the catch-all route, so a single nullish entry in `auth.api` throws there — killing the route, not just the Sentry tag. That directly contradicts the principle this PR states in `better-auth-request-context.ts` ("observability must never be the thing that breaks an auth request"): the runtime path was defended, the module-load path was not. Now `endpoint?.path`, and the labeller already drops non-strings so a nullish entry degrades that one endpoint to `unknown`.

Pinned by a `null` entry in the route test's `auth.api` mock. Reverting to `endpoint.path` fails the file to load with `TypeError: Cannot read properties of null (reading 'path')` — the same shape as the ticket this PR fixes.

**2. Two citations into third-party internals rot under the 1.7 upgrade.** Checked all four against the versions #8835 installs:

| citation | today (1.6.23 / better-call 1.3.7) | after ENG-2343 (1.7.0 / 1.4.0) |
| --- | --- | --- |
| `api/index.mjs:194-197` | correct | still exact |
| `api/index.mjs:210` | correct | still exact |
| `endpoint.mjs:56` | correct | moved to `:55` |
| `router.mjs:38` | correct | different branch |

The two better-call ones now name the symbol (`internalHandler.path = path`) instead of the line; the two better-auth ones are left as-is because they are still exact.

**The useful half of that check:** the mechanism this PR depends on is unchanged in 1.7.0 — the router still logs a non-`APIError` as `(e.name, e)` with the endpoint discarded, and `onAPIError.onError` still `return`s before the logger. Those blocks are byte-identical between 1.6.23 and 1.7.0, so **this fix survives the Better Auth upgrade** and the two PRs do not need sequencing.

**Security pass — clean, with what I checked:** the emitted label is bounded to `{declared literal paths} ∪ {/<known-segment>/*} ∪ {unknown}`, so tag cardinality is bounded and an attacker cannot inject an arbitrary value. Enumerated every parameterized endpoint better-auth actually declares (`/reset-password/:token`, `/callback/:id`, `/oauth2/callback/:providerId`, `/oauth2/client/:id`) and each truncates correctly; the query string is dropped before matching, so `?code=`/`?token=` never reach a tag; the message is deliberately never forwarded. `/reset-password` (literal) and `/reset-password/*` (truncated) stay distinct rather than merging.

**Considered and dropped:** the new integration test uses the same unguarded `.map((endpoint) => endpoint.path)` at line 122. Left as-is — there a nullish entry fails the test loudly, which is the desired outcome for a canary, rather than breaking production.

Verification: 28 auth-related test files / 289 tests green, typecheck 0 errors, prettier clean. Re-requesting review.

BhagyaAmarasinghe (approved):

PR #8908, opened 2026-08-18
fix(unify): retire stale coming-soon on feedback source options

xernobyl:
Fixes ENG-2431

## What & why

The "Create feedback source" dialog's API ingestion and MCP server options were stuck behind a stale "Coming soon" badge (`disabled: true`), with dead doc links behind their (unreachable, since disabled) click handlers. Per product direction, restore the original interaction: both are selectable cards again, and the dialog's main CTA reads "Learn more" and opens the correct, verified-live docs in a new tab when one is selected — instead of a disabled dead end.

<details>
<summary>Timeline / why these were disabled in the first place</summary>

Both options were flipped to `disabled: true` + "Coming soon" in an earlier PR, citing an authorization concern with the underlying feedback-records gateway. The actual MCP tools and the API ingestion route shipped in a later PR, but the dialog's UI state and doc links were never updated to match — leaving two real, working features hidden behind a stale label pointing at pages that don't exist.

</details>

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `getFeedbackSourceOptions` returns all 4 options enabled, with no `disabled`/`badge` fields | unit (mutation) — `utils.test.ts` `describe("getFeedbackSourceOptions")` | pass |
| Selecting API ingestion / MCP server switches the CTA to "Learn more" and opens the correct docs URL in a new tab | manual — clicked through the real dialog on a real logged-in session; also confirmed programmatically via `window.open` instrumentation (`(url, "_blank", "noopener,noreferrer")` for both) | pass — screenshot below |
| No dead code or unused i18n keys left behind (`ApiIngestionSetupAlert`, the old broken doc-URL constants, `common.coming_soon`) | unit (guard) — full `pnpm test` (114 tests in this module) + `pnpm --filter @formbricks/i18n-utils scan-translations` | pass, zero unused keys |

MCP server option selected, showing the Learn more CTA

**Open gaps**

- No Playwright/e2e coverage exists for this dialog at all (pre-existing gap, not introduced here).

**Risks**

- none — this is UI copy and dead-code removal; no queries, routes, or auth logic touched.

---

> [!NOTE]
> **AI model used** — Claude Opus 5, reasoning effort high.


jobenjada (approved):
looks good :)

PR #8909, opened 2026-08-18
chore: integrate current main into AuthZed epic

BhagyaAmarasinghe:
Fixes ENG-2447

## What & why

- Integrates the current `origin/main` product and security behavior into `epic/authzed` with a real two-parent merge commit.
- Replays the previously reviewed AuthZed reconciliation ledger plus the eight AuthZed commits that landed after that ledger.
- Resolves conflicts deliberately:
  - preserves current-main integration rate limits while retaining AuthZed feedback-source, historical-import, chart, feedback-directory, and feedback-record limits;
  - preserves the central `organization.manage_api_keys` authorization gate while using the current-main API-key component contract;
  - preserves current-main feedback, survey, and API-key product request shapes alongside the AuthZed authorization and projection paths.
- Adds `authzed/INTEGRATION.md` with the exact parent SHAs, reconciliation source, conflict decisions, and validation contract.

> [!IMPORTANT]
> This PR must be merged with a **merge commit**, not squash-merged. The acceptance contract requires `origin/main` to remain an ancestor of the synchronized epic; squash merging would destroy that ancestry.

## Breaking changes

None

## Migrations & env

- Includes the current-main feedback-source element-scope Prisma migration.
- No new AuthZed environment variables or deployment cutover in this PR.
- Local validation of the newly required current-main Hub/Cube environment used non-secret placeholder URLs and key values.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Real repository ancestry | manual | Commit `ba820343e` has epic `3dbebde2d` and current main `12b8aa1a3` as parents; `git merge-base --is-ancestor origin/main HEAD` succeeds. |
| AuthZed schema contract | unit (guard): `pnpm authzed:validate` | 36 relationships, 157 assertions, and 6 expected relations validated. |
| AuthZed and central authorization regression | unit (guard): `pnpm --filter @formbricks/web exec vitest run lib/authzed lib/authorization` | 46 files and 656 tests passed. |
| Complete monorepo regression | unit (guard): `pnpm exec turbo run test --no-cache --env-mode=loose` | 672 files and 8,920 tests passed. |
| Docker AuthZed runtime contract | unit (guard): `bash docker/authzed-compose-contract.sh` and focused installer tests | Compose contract and 8 installer/setup tests passed. |
[...]

PR #8910, opened 2026-08-18
docs: define direct AuthZed cutover contract

BhagyaAmarasinghe:
Fixes ENG-2448

## What & why

- Replaces the project's shadow/cohort release strategy with one canonical direct-cutover contract.
- Defines the immutable legacy-authoritative projection bridge and the no-fallback SpiceDB-authoritative candidate.
- Locks PostgreSQL/SpiceDB data ownership, durable-delivery invariants, 60-second revocation freshness, fail-closed behavior, artifact compatibility, abort triggers, rollback, evidence, approver responsibilities, and the self-hosted v6 upgrade boundary.
- Makes `authzed-sandbox` a mandatory direct-authority and rollback rehearsal before staging, with a 24-hour gate; staging retains the seven-day authoritative soak.
- Defines EU-before-KSA production sequencing and the 24-hour regional gate.
- Marks the old shadow-mode Linear proposal as superseded, updates the project overview, creates a canonical Linear contract document, and links all 27 implementation/rollout tickets to it.
- Updates authorization, projection, performance, runbook, monitoring, environment, and self-hosting documentation without changing runtime behavior.

This PR is stacked on [#8909](https://github.com/formbricks/formbricks/pull/8909). After that merge lands in `epic/authzed`, rebase this branch onto the updated epic and retarget this PR to `epic/authzed`.

## Breaking changes

None

## Migrations & env

- none
- The contract defines a future breaking self-hosted v6 upgrade; ENG-2452 implements its preflight and deployment behavior.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Canonical schema contract remains unchanged | unit (guard): `pnpm authzed:validate` | 36 relationships, 157 assertions, and 6 expected relations passed. |
| All changed Markdown/MDX is repository-formatted | unit (guard): focused `pnpm exec prettier --check` | All eight changed files passed. |
| Markdown patch integrity | manual: `git diff --check` | No whitespace errors. |
| Shadow proposal no longer appears as the approved release strategy | manual document audit | Repository runbook, authorization docs, project overview, and historical Linear proposal now point to direct cutover. |
| Every rollout workstream has one contract | manual Linear audit | 27 foundation, sandbox, staging, production, and hypercare tickets link to the canonical Linear contract. |
| Sandbox precedes staging | manual contract audit | ENG-2469/ENG-2470 and the 24-hour gate are explicit prerequisites to staging. |
| Rollback cannot silently select legacy in the candidate | manual contract audit | Rollback is pinned bridge redeployment; the candidate has no runtime selector. |

**Open gaps**

- Product, security, infrastructure, and operations approval is still required in review.
- Runtime outbox, evaluator, list lookup, telemetry, cleanup, and self-hosted upgrade behavior are intentionally implemented by their linked tickets.
- Live sandbox/staging/production evidence cannot exist until those implementation artifacts are built.

**Risks**

- This is a release-contract change: downstream implementations must be rejected if they weaken an invariant even when their local tests pass.
- Public docs distinguish current v5 best-effort bridge behavior from planned v6 direct authority so operators are not told that unshipped outbox/metrics commands already exist.
- The Linear contract mirrors the checked-in file and must be kept synchronized until this PR merges; after merge the repository file is the reviewable source.

---

> [!NOTE]
> **AI model used** — GPT-5.6, reasoning effort high.


authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Invariants
+
+The following are release-blocking invariants:
+
+1. Every authorization-relevant PostgreSQL mutation and its projection-outbox event commit atomically.
+2. Relationship delivery is idempotent and at least once. A process crash or SpiceDB outage cannot silently lose
+   a committed grant, downgrade, or revocation.
+3. A pending revocation emits a warning at 15 seconds, a critical alert at 45 seconds, and makes protected
+   authorization fail closed at 60 seconds.
```
kiln-agent, line 35: **The revocation fail-closed guard has no stated blast radius.**

This invariant says a pending revocation "makes protected authorization fail closed at 60 seconds", and the failure table (line 196) says a dead-lettered item means "Fail closed for protected authorization until replay and clean audit". Neither scopes the guard to the principal, resource, or tenant the pending revocation touches.

Read literally: one poison-pill outbox row — a revocation SpiceDB keeps rejecting — exhausts the 20 attempts from line 62 overnight, dead-letters, and every user in every organization is denied every protected operation until an operator wakes up, replays it, and gets a clean audit. That is a full-platform outage triggered by one bad row, and it is the kind of thing an implementer will get wrong in either direction if the contract does not say.

Please state the scope explicitly: scope the guard to the affected tenant/principal subgraph, or say it is deployment-wide and that the availability trade-off is accepted.

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+
+The projection bridge writes an outbox row in the same Prisma transaction as the source mutation. PostgreSQL is
+the durable queue; BullMQ is only the recurring trigger that wakes workers.
+
+Workers claim rows with leases and `FOR UPDATE SKIP LOCKED`, reconcile the complete current PostgreSQL state for
+the affected target, and acknowledge the row only after the SpiceDB write succeeds. Delivery is safe under
+duplicate, reordered, delayed, and concurrently claimed events because reconciliation is idempotent and derives
+the desired state from PostgreSQL rather than trusting an event payload as current truth.
+
+An item is attempted at most 20 times before entering dead letter. Operators must be able to inspect sanitized
```
kiln-agent, line 62: **The delivery design cannot meet the 60-second revocation budget as specified.**

Two constraints here are in tension with invariant 3 (15s warning / 45s critical / 60s fail-closed):

- "BullMQ is only the recurring trigger that wakes workers" — a *recurring* trigger, interval unspecified.
- "An item is attempted at most 20 times before entering dead letter" — retry schedule unspecified.

Concretely: a worker crashes right after claiming a revocation lease. The row is not reclaimed until the next tick, so with any interval at or above 15s the first retry alone lands past the warning threshold. And 20 attempts cannot fit inside 60 seconds under any non-degenerate backoff. Routine worker churn therefore trips the 60-second guard — which line 152 then converts into a mandated post-deployment abort and rollback.

Pin the trigger interval and the retry/backoff schedule so worst-case delivery latency provably sits inside the 60-second budget, or decouple the dead-letter attempt count from the freshness SLA (e.g. revocations get a short, bounded schedule; other event types get the 20 attempts).

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+### Staging
+
+Staging receives production-like private SpiceDB infrastructure with two replicas, a PodDisruptionBudget,
+topology spread, dedicated PostgreSQL credentials, backups, and point-in-time recovery. It uses the sandbox-
+validated bridge and candidate digests and the same schema digest.
+
+After establishing a clean graph, staging performs the controlled cutover, complete UI/API/MCP/gateway and
+failure/restore/capacity suites, and a seven-day authoritative soak. The gate is:
+
+- no unexplained authorization decisions or operational errors above 0.1%;
```
kiln-agent, line 132: **This gate can be read as tolerating 0.1% wrong authorization decisions.**

"no unexplained authorization decisions or operational errors above 0.1%" attaches the qualifier ambiguously to both clauses. An approver at the end of the seven-day soak can read it as a single threshold covering both and sign off with a few hundred unexplained allow/deny outcomes at staging volume — and that same build then cuts over to EU and KSA.

`RUNBOOK.md` §8 applies 0.1% only to the operational-error rate, so the intent is clearly two separate criteria. Split them:

```suggestion
- no unexplained authorization decisions;
- operational errors at or below 0.1%;
```

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+- no dead-letter revocations or freshness-guard activation;
+- clean scheduled audits;
+- authorization p95 below 250 ms and p99 below one second;
```
kiln-agent, line 135: **This latency gate is 25–30x looser than our own measured baseline, so it gates nothing.**

`authzed/PERFORMANCE.md` (which this PR annotates as the reference benchmark) measures the full SpiceDB `can()` path at **8–10 ms p95** and **23–30 ms p99**. A gate of 250 ms p95 / 1 s p99 admits a ~20x regression without anyone noticing — a lost dispatch cache, an undersized `pgxpool`, or a pre-check resolver that stops benefiting from `reactCache` would all still measure ~200 ms p95 and pass, then ship to both production regions.

Suggest expressing the gate as a multiple of the ENG-2453 revalidated staging baseline (e.g. "p95 no more than 3x the recorded staging baseline") rather than an absolute number three decades above the measured value.

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Artifact freeze and compatibility
+
+After the durable bridge passes its implementation gates:
+
+1. Build the bridge once and record the application image digest, source commit, schema digest, Prisma migration
+   head, SpiceDB image digest, and operator/chart versions.
+2. Prove that this exact image can read the final outbox migration and final current-model schema.
+3. Keep legacy authorization authoritative, set AuthZed consistency to `fully_consistent`, and remove all shadow
+   and enforcement target configuration.
```
kiln-agent, line 79: **Removing shadow + enforcement config removes the only way the bridge can evaluate a SpiceDB decision — but line 103 still requires the bridge to verify the permission fixture.**

In the current code (`apps/web/lib/env.ts:350-415` plus the coordinator), a SpiceDB permission evaluation on the bridge only happens when a shadow or enforcement target matches. This step removes both, and line 23 additionally forbids configuring comparison telemetry in sandbox/staging/production. What is left can verify *relationships* (audit, `zed` assertions) but never a *decision*.

So the sandbox-bridge bullet "verify the complete owner/manager/member/billing, team, workspace, API-key, resource, and cross-tenant fixture" (line 103) has no mechanism behind it, and the first environment where a SpiceDB permission decision is ever compared against the legacy one is the environment where SpiceDB is already sole authority.

Two smaller consequences of the same step: setting `AUTHZED_CONSISTENCY=fully_consistent` on the bridge is inert, since no check is issued; and `env.ts` only *requires* `fully_consistent` when enforcement targets are set, so this is a convention rather than something validation enforces.

Please name the mechanism that establishes decision parity before the candidate takes authority (offline assertion suite? a one-off harness run outside the frozen config?), or carve out an explicit, time-boxed exception.

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Workstream ownership
+
+| Workstream                                           | Tickets                                                              |
+| ---------------------------------------------------- | -------------------------------------------------------------------- |
+| Contract and repository integration                  | ENG-2448, ENG-2447                                                   |
+| Durable bridge                                       | ENG-2408                                                             |
+| Direct-authority application                         | ENG-2446, ENG-2449, ENG-2451, ENG-2450, ENG-2452                     |
+| Sandbox graph and cutover                            | ENG-2469, ENG-2470                                                   |
+| Staging infrastructure, graph, cutover, QA, and soak | ENG-2454, ENG-2458, ENG-2455, ENG-2364, ENG-2457, ENG-2453, ENG-2456 |
```
kiln-agent, line 238: `ENG-2364` looks like a transposition typo. Every other ticket in this table falls in the contiguous block ENG-2446–ENG-2470, plus ENG-2408 for the durable bridge — 2364 is the only outlier.

If it is wrong, this row points reviewers at an unrelated ticket while the real staging ticket has no contract owner, which matters given the PR claims all 27 tickets were linked to the canonical contract. Worth a quick check against Linear.

---

authzed/RUNBOOK.md
```diff
@@ -261,23 +265,45 @@ shared with another installation?
...
 What matters is afterwards:
 
-1. Every projection attempted during the outage was dropped and will not be retried.
+1. Assume every projection attempted during the outage was dropped and will not be retried.
 2. Once SpiceDB is healthy, `pnpm authzed:health` returns `healthy`.
 3. Run the backfill (§3). Until it reports a clean run, assume the graph is incomplete.
-4. **Keep shadow evaluation and enforcement off until then.** A clean run reports
-   `completedAtSnapshot` — hand that revision to shadow evaluation (ENG-1738) as its freshness floor
-   rather than guessing whether the graph is warm.
+4. Keep every direct-authority deployment blocked until the run is clean.
```
kiln-agent, line 279: **This replaced an actionable guard with one that no shipped build can act on.**

The old step 4 was "Keep shadow evaluation and enforcement off until then." The replacement talks about blocking a "direct-authority deployment" — a state that does not exist in any build an operator can run today. Meanwhile `AUTHZED_AUTHORIZATION_ENABLED` and `AUTHZED_ENFORCEMENT_TARGETS` are still live in `apps/web/lib/env.ts:284-290`, so an internal deployment *can* be enforcing on SpiceDB right now.

Scenario: such a deployment survives a SpiceDB outage. §5 tells the operator to run the backfill and to block a cutover that isn't scheduled; nothing tells them to turn enforcement off. SpiceDB keeps deciding against a graph the same section says to assume is incomplete, and entitled users get denied until the backfill finishes.

Keep both instructions — the enforcement-off guard for the shipped bridge, and the cutover block for the direct-authority path.

---

authzed/RUNBOOK.md
```diff
@@ -432,17 +525,19 @@ histogram_quantile(0.95, sum(rate(formbricks_authzed_projection_duration_seconds
...
 | `unmanaged` relationships reported                       | Something other than Formbricks is writing to this SpiceDB, or the schema moved ahead of its projector. Never pruned; investigate before enforcing.                                                     |
 | Schema check reports `drifted`                           | `pnpm authzed:schema apply --expected-current-digest <remoteDigest>`. Relationship repair against a drifted schema is not meaningful.                                                                   |
-| Sync cannot be restored and enforcement is pending       | `AUTHZED_ENABLED=0` and note that a full backfill is required before re-enabling.                                                                                                                       |
+| Durable delivery or a clean graph cannot be restored     | Block cutover. If already authoritative, roll back to the pinned bridge digest, drain/replay, and require a clean audit before another attempt.                                                         |
```
kiln-agent, line 543: **Same issue as §5: this escalation row now has no executable remedy on the current release.**

Both remedies here require unshipped artifacts — "Block cutover" (none is scheduled) and "roll back to the pinned bridge digest" (does not exist yet). The previous row's remedy, `AUTHZED_ENABLED=0` plus a full backfill before re-enabling, was the one thing an operator on today's build could actually do, and it's gone.

Suggest keeping two rows: one for the shipped bridge (`AUTHZED_ENABLED=0`, full backfill before re-enabling) and one for the direct-authority artifact (block cutover / roll back to the pinned digest).

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+| Redis/BullMQ unavailable                     | PostgreSQL outbox continues accumulating; a later trigger resumes delivery.                                            | Existing graph decisions continue until revocation age reaches 60 seconds, then protected authorization fails closed. |
+| Worker crashes after source commit           | Leased event becomes reclaimable and is reconciled idempotently.                                                       | Same; freshness guard limits stale-revocation exposure.                                                               |
+| Event is duplicated or reordered             | Reconciliation reads current PostgreSQL state and converges.                                                           | Same.                                                                                                                 |
+| Item enters dead letter                      | Alert and block cutover.                                                                                               | Fail closed for protected authorization until replay and clean audit.                                                 |
```
kiln-agent, line 196: This row contradicts invariant 3 and the alert/abort rules, which scope the fail-closed guard to **pending revocations** older than 60 seconds.

Scenario: a non-revocation item — say a new team grant — exhausts its 20 attempts and dead-letters at 03:00. Read literally, this row requires the direct-authority artifact to deny *every* protected operation platform-wide until an operator replays it and a full audit passes: a self-inflicted total authorization outage triggered by a single stuck grant. Invariant 3, the abort trigger ("any revocation is pending, dead-lettered, or older than its SLA"), and RUNBOOK §8 ("any dead-letter revocation is critical") all scope this to revocations, so nothing should fail closed here.

ENG-2451 can implement either reading and claim contract compliance. Please scope this cell explicitly, e.g. "Revocation dead letter: fail closed for protected authorization until replay and clean audit; other dead letters: critical alert, no fail-closed."

---

authzed/RUNBOOK.md
```diff
@@ -432,17 +525,19 @@ histogram_quantile(0.95, sum(rate(formbricks_authzed_projection_duration_seconds
 # for: 15m
 ```
 
-Every one of these resolves to the same first action: **run the backfill and confirm a clean run.**
+Every one of these resolves to the same first action: inspect and drain the durable outbox, then run the full
+audit and confirm a clean result. On a pre-outbox bridge, run the backfill immediately because failed writes were
+not retained.
```
kiln-agent, line 530: **The two clauses are in the wrong order for the alerts they sit under.**

The PromQL rules immediately above are all built on `formbricks_authzed_projection_total` / projection duration — metrics that only exist on the pre-outbox bridge. An operator paged by `AuthzedProjectionFailing` on the shipped release reads "inspect and drain the durable outbox" first, goes looking for an outbox inspect/drain subcommand that `formbricks-authzed` does not have, and only then reaches the sentence that actually applies to them.

Lead with the pre-outbox action, then the outbox procedure:

```suggestion
Every one of these resolves to the same first action. On the current pre-outbox bridge, run the backfill
immediately and confirm a clean run — failed writes were not retained. On the durable bridge and the
direct-authority artifact, inspect and drain the outbox first, then run the full audit and confirm a clean result.
```

---

authzed/RUNBOOK.md
```diff
@@ -1,31 +1,35 @@
...
-> Drift accumulates until something reads SpiceDB and disagrees with PostgreSQL.
+The currently checked-in bridge still performs post-commit best-effort projection. ENG-2408 replaces that
+temporary behavior with a transactional PostgreSQL outbox before any direct-authority deployment is allowed.
+Until that lands, treat every failed projection as potentially lost and prohibit cutover.
+
+> **Direct authority requires durable delivery.** A committed authorization mutation must enqueue its
+> relationship reconciliation atomically, and a revocation that cannot be delivered within 60 seconds must make
+> protected authorization fail closed.
 
 Everything below exists to make that visible and recoverable.
```
kiln-agent, line 14: Minor: "that" lost its antecedent. It used to point at "Sync failures are silent by design. Drift accumulates…"; now it resolves to the new blockquote, which is a forward-looking *requirement* — not something a runbook makes "visible and recoverable".

Suggest naming the problem the sections actually address, e.g. "Everything below exists to make projection loss and graph drift visible and recoverable."

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+SpiceDB becomes the sole authorization decision engine. PostgreSQL remains the business record for users,
+memberships, roles, teams, workspace grants, API-key scopes, feedback directories, and every other fact that is
+projected into SpiceDB.
+
+The release uses two immutable application artifacts:
+
+| Artifact                   | Authority        | Purpose                                                                                                                                                  |
+| -------------------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Projection bridge          | Legacy evaluator | Delivers every committed relationship change durably, establishes and audits the graph, and is the deployment rollback artifact.                         |
+| Direct-authority candidate | SpiceDB          | Evaluates every scalar and list authorization decision, fails closed on operational errors, and contains no runtime legacy evaluator or shadow selector. |
```
kiln-agent, line 21: The candidate's Purpose lists only decision evaluation. Durable delivery is assigned exclusively to the bridge on line 20, and "Durable relationship delivery" opens with "**The projection bridge** writes an outbox row in the same Prisma transaction as the source mutation" (line 54). Nowhere does the contract state that the candidate must also keep the transactional outbox and its workers.

Scenario: ENG-2446/ENG-2450 build the candidate from this table — it evaluates every `can()`/`assertCan()` against SpiceDB and drops the bridge-era projection code as part of "legacy removal". After cutover no membership, role, team, or API-key mutation ever reaches SpiceDB. The graph silently freezes at its pre-cutover state while decisions are authoritative, and the 60-second revocation guard never fires because nothing is ever enqueued to age.

Invariant 1 is unqualified and Rollback step 4 says "durable outbox delivery is **still** active", so the intent is clearly that the candidate delivers too — but the two definitional places don't say it. Suggest adding delivery to the candidate's Purpose cell and rewording line 54 to "Both artifacts write an outbox row…".

---

apps/web/lib/authorization/README.md
```diff
@@ -209,13 +217,11 @@ These read a role but do not decide access, so they stay as they are:
...
-  single shadow-only `LookupResources(workspace, read)` comparison after the response and counts that
-  list observation as one central authorization operation. It never changes the returned list and does
-  not expose generic lookup or enforcement semantics; those remain ENG-1713. Comparison telemetry is
-  directional: an identical set emits one `match` sample, while drift in both directions emits one
-  `legacy_allow_authzed_deny` and one `legacy_deny_authzed_allow` sample for the same observation.
+- **List scoping during the bridge.** Workspace list queries remain PostgreSQL-authoritative and narrow by
+  role while the bridge is deployed. MCP `list_workspaces` has one bounded `LookupResources(workspace, read)`
+  migration observation rather than one check per row. ENG-2449 replaces every current-model organization
+  and workspace authorization list with an authoritative bounded lookup before the direct-authority artifact
+  can ship.
```
kiln-agent, line 224: The rewrite dropped the binding constraint from the old bullet: *"It never changes the returned list and does not expose generic lookup or enforcement semantics."* That was the safety property of the MCP `list_workspaces` observation, and it matches what `workspace-list-observer.ts` actually does — record comparison telemetry, never mutate the result.

What replaced it only says the observation is "bounded" and happens "rather than one check per row", which describes cost, not authority. An implementer picking up ENG-2449 from this paragraph could reasonably wire `LookupResources` into the returned list while the bridge is still deployed and PostgreSQL is still authoritative, changing what users see before any parity evidence exists.

Worth keeping the "never changes the returned list" clause as an explicit property of the bridge behaviour.

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+  database connectivity;
+- deploy the immutable bridge digest and guarded canonical schema digest;
+- drain the outbox, reconcile the complete graph, and require two consecutive clean audits;
+- observe one clean scheduled six-hour audit;
+- create mutations during a SpiceDB outage, restore it, and prove replay converges without dead letters; and
+- verify the complete owner/manager/member/billing, team, workspace, API-key, resource, and cross-tenant fixture.
+
+### Sandbox direct authority
+
+The sandbox cutover uses the same procedure as later environments:
```
kiln-agent, line 107: "the same procedure as later environments" makes all eight steps below binding for staging and both production regions — including step 7's roll-back-to-bridge-and-forward rehearsal and step 8's 24-hour hold. But the Production section requires neither rehearsal, and RUNBOOK §7's "Cut an environment to direct authority" checklist stops at step 6 and then scopes the rehearsal to sandbox only ("Sandbox must pass rollback/forward recovery and 24 continuous healthy hours").

Scenario: an operator cutting EU production follows the six-step RUNBOOK checklist and never redeploys the bridge digest mid-window. A later approval audit reads this line, concludes the mandatory eight-step procedure was not executed, and the EU evidence record cannot certify KSA's 24-hour gate.

Either say the shared procedure is steps 1–6 and that 7–8 are sandbox-specific, or add the rehearsal to the Production section and to RUNBOOK §7.

---

kiln-agent (commented):
Reviewed the diff as a release-contract change. No runtime code is touched, so the findings are contract/runbook defects rather than code bugs — but a few of them would produce real operational failures once implemented.

Verified as correct along the way: the 35-action vocabulary count matches `AUTHORIZATION_PERMISSION_MAP`; `../../../../authzed/CUTOVER.md` and `./RUNBOOK.md#7-direct-authority-cutover` both resolve; RUNBOOK section renumbering (6→9) is internally consistent and nothing else in the repo links the old `#6-shadow-and-enforcement-rollout` anchor; the API-key-principals-are-routed claim matches the call sites; `fully_consistent` is a real `AUTHZED_CONSISTENCY` value and setting it without enforcement targets passes `env.ts` validation; all eight files are Prettier-clean; the 27-ticket count in the workstream table adds up.

Highest-value items: the unscoped revocation fail-closed guard, the retry schedule vs the 60-second budget, and the two staging gates (0.1% ambiguity, latency threshold 25–30x above our own measured baseline). The RUNBOOK ones are smaller but concrete: the rewrite consistently replaced advice that shipped operators can act on today with advice that only applies to artifacts that don't exist yet.

---

apps/web/lib/authorization/README.md
```diff
@@ -95,24 +95,32 @@ evaluator must continue honoring `USER_MANAGEMENT_MINIMUM_ROLE`:
...
 actor/action/resource types.
 
-## Migration inventory
+## Migration inventory and authority contract
 
-The central interface uses the legacy evaluator unless an internal ENG-1738
-rollout rule selects the current request surface and organization. Migration
-means callers use semantic actor/action/resource decisions; deployments with
-authorization rollout disabled have no AuthZed read dependency.
+The bridge implementation still uses the legacy evaluator while the durable
```
kiln-agent, line 100: This drops the qualifier the old text carried ("uses the legacy evaluator **unless** an internal ENG-1738 rollout rule selects the current request surface and organization"), and the ENG-1738 bullet below loses "with fail-closed operational behavior" in the same pass. The result is that this module README no longer documents a code path that is still live: with `AUTHZED_ENFORCEMENT_TARGETS` + `AUTHZED_ENFORCEMENT_ORGANIZATION_IDS` set, the coordinator makes SpiceDB authoritative inline and fails closed on operational errors.

That path has not been removed yet — `apps/web/lib/env.ts` still validates both vars, `rollout-config.ts` still builds an `enforcement` rule, and RUNBOOK §6 still documents the configuration (ENG-2450 is the ticket that removes it).

Scenario: an engineer changing `coordinator.ts` reads this paragraph, concludes SpiceDB can never decide a request on the current code, and skips testing the enforcement branch. Worth keeping one sentence noting the branch still exists in the checked-in code and is removed by ENG-2450.

---

docs/self-hosting/configuration/environment-variables.mdx
```diff
@@ -208,9 +208,10 @@ This check is intentionally a CLI rather than a browser page or HTTP endpoint. A
 general `/health` response, Kubernetes readiness, or application startup, so a transient SpiceDB outage does
 not restart or mark the rest of Formbricks unhealthy.
 
-When authorization enforcement is enabled, failed or unavailable permission checks must deny the protected
-operation. Do not make the general application readiness probe depend on AuthZed availability; this avoids a
-transient authorization outage restarting unrelated Formbricks workloads.
+In the v6 direct-authority release, failed or unavailable permission checks fail the protected operation closed;
```
kiln-agent, line 211: This paragraph promises v6 fail-closed behaviour but never mentions the consistency mode v6 requires, while the same page a few lines up still says "The supported consistency values are `minimize_latency` and `fully_consistent`; the default is `minimize_latency`."

Scenario: a self-hoster upgrades to v6, leaves `AUTHZED_CONSISTENCY` at the documented default, and gets snapshot-stale permission checks — a revoked member keeps passing authorization for as long as the read revision lags. That silently violates `CUTOVER.md` invariant 6 ("Permission checks use `fully_consistent` consistency"), and nothing on the public configuration page warns about it.

Worth stating here that v6 requires `AUTHZED_CONSISTENCY=fully_consistent` (or that v6 changes the default).

---

docs/self-hosting/setup/monitoring.mdx
```diff
@@ -202,10 +202,19 @@ Use these endpoints for monitoring system health in container orchestration and
...
-Formbricks exposes bounded projection and authorization-comparison metrics when AuthZed is enabled. Monitor
-projection failures and disabled outcomes, shadow mismatches, authorization operational errors, and authorization
-latency. Also monitor SpiceDB pod availability, datastore migrations, PostgreSQL pool saturation, dispatch
-latency, and cache behavior.
+Current v5 bridge releases expose bounded projection and authorization-comparison metrics when AuthZed is
+enabled. Monitor projection failures, disabled outcomes, migration mismatches, authorization operational errors,
+and authorization latency.
+
+The v6 direct-authority release removes comparison telemetry and requires authoritative allow/deny/operational-
+error and latency metrics plus outbox backlog, oldest pending revocation, dead-letter, reconciliation drift, and
```
kiln-agent, line 210: Line-ending hyphen: `operational-` ends line 209 and `error` starts line 210, so the Markdown soft break renders on the published docs site as "allow/deny/operational- error" — a stray hyphen plus a space inside the compound word.

```suggestion
The v6 direct-authority release removes comparison telemetry and requires authoritative
allow/deny/operational-error and latency metrics plus outbox backlog, oldest pending revocation, dead-letter,
reconciliation drift, and
```

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+8. Require 24 continuous healthy hours after the final forward deployment.
+
+Staging work is blocked by any unexpected decision, unexplained operational error, dead letter, normal-operation
+60-second freshness guard, non-clean scheduled audit, rollback/recovery failure, or identifier/secret leak.
+
+### Staging
+
+Staging receives production-like private SpiceDB infrastructure with two replicas, a PodDisruptionBudget,
+topology spread, dedicated PostgreSQL credentials, backups, and point-in-time recovery. It uses the sandbox-
+validated bridge and candidate digests and the same schema digest.
```
kiln-agent, line 127: Line-ending hyphen: `sandbox-` ends line 126 and `validated` starts line 127, so the soft break renders as "It uses the sandbox- validated bridge and candidate digests".

```suggestion
topology spread, dedicated PostgreSQL credentials, backups, and point-in-time recovery. It uses the
sandbox-validated bridge and candidate digests and the same schema digest.
```

---

authzed/CUTOVER.md
```diff
@@ -0,0 +1,245 @@
...
+clean audits and revocation-delivery evidence.
+
+## Cutover abort triggers
+
+Abort before deploying the direct-authority image when any of the following is true:
+
+- the mutation freeze would exceed 15 minutes;
+- outbox drain or the final audit does not complete within 10 minutes;
+- the audit is not clean or contains cross-tenant, unmanaged, invalid, mismatched-parent, or stale higher-
+  permission findings;
```
kiln-agent, line 159: Same line-ending-hyphen issue as line 126: `higher-` ends this line and `permission` starts the next, so the abort trigger renders as "…or stale higher- permission findings".

```suggestion
- the audit is not clean or contains cross-tenant, unmanaged, invalid, mismatched-parent, or stale
  higher-permission findings;
```

---

kiln-agent (commented):
Follow-up pass on the same diff, covering the four files the earlier review round did not reach (module README, the two public self-hosting docs) plus two rendering defects in `CUTOVER.md`.

Also re-verified and found correct: the 35-action vocabulary count matches `AUTHORIZATION_PERMISSION_MAP`; `../../../../authzed/CUTOVER.md` and `./RUNBOOK.md#7-direct-authority-cutover` both resolve; the RUNBOOK 6→9 renumbering is internally consistent and nothing in the repo still links the old `#6-shadow-and-enforcement-rollout` anchor; `rollout-runbook.test.ts` keys on the "Valid targets are:" block, which is preserved, so no test breaks; removing cohort/shadow/enforcement config while setting `fully_consistent` passes `env.ts` validation; all eight files are Prettier-clean.

---

authzed/CUTOVER.md
mattinannt, line 1: I don't think we should store these documents in the codebase. They will just rot there and pollute the repository.

authzed/PERFORMANCE.md
mattinannt, line 1: same here. why do we need this in the codebase. I think maybe a document in the Linear project would be a better place for this.

authzed/RUNBOOK.md
mattinannt, line 1: what's the difference between RUNBOOK and README? What should go where?
Still wondering if in general we should have these files in the repo or at least where in the repo. Or if they should go into the `docs/`.

Currently this is not following a good pattern in my opinion.

mattinannt (commented):
Thanks for the PR :-) 

Please also take a look at the automatic review comments and resolve them.

Also I'm a bit sceptical about these new `.md` files in the repo. Can we e.g. make this more concise and only have a `README.md` here and keep temporary information somewhere else?

BhagyaAmarasinghe:
> Thanks for the PR :-)
> 
> Please also take a look at the automatic review comments and resolve them.
> 
> Also I'm a bit sceptical about these new `.md` files in the repo. Can we e.g. make this more concise and only have a `README.md` here and keep temporary information somewhere else?

@mattinannt 
You're correct, this is a misinformed docs PR created by Claude. Closing.

xernobyl (approved):
Reviewed at `158672ba8`. Approving on scope and security posture — this is genuinely docs-only (8 files, all `.md`/`.mdx`, zero runtime files), so the "no runtime behavior change" claim holds.

> [!IMPORTANT]
> **Please merge this with a merge commit, and do not rebase it first.**
>
> #8909 was squash-merged, so the epic tip `16ae08621` has a single parent and `origin/main` is **not** an ancestor of `epic/authzed` — the acceptance criterion ENG-2447 was written around. The trees are identical, so nothing is functionally wrong today, but git's merge-base between `main` and the epic is still the pre-sync commit, which means the next main→epic sync redoes a ~1,500-file three-way merge against a stale base and re-surfaces every deliberate divergence as a conflict.
>
> This branch still contains the original two-parent merge `ba820343e`, so **merging this PR with a merge commit restores the ancestry for free.** The description says to rebase onto the updated epic first — that would drop `ba820343e` and close the window permanently.

### Security review

Nothing to flag, and the direction is a tightening. Every secret- or credential-related line this PR *adds* raises the bar: invariant 9 forbids exposing tokens, database credentials, raw SDK errors, schema contents, relationship strings, and actor/resource/tenant identifiers in logs, metrics, command output and alerts; the sandbox/staging sections require private networking, dedicated PostgreSQL credentials, backups and restore evidence; and the approver split gives Security explicit ownership of fail-closed, tenant-isolation, revocation and secret-handling evidence.

It does not touch the existing `AUTHZED_INSECURE` / plaintext-gRPC guidance. The one removal — the note exploring `at_least_as_fresh` + ZedTokens for the enforcement path — is replaced by an explicit statement that changing consistency is outside this contract and needs a separately reviewed revocation/read-after-write design. That is stricter than what it replaced.

### What I tested

The risk this PR names for itself is telling operators that unshipped v6 outbox/metrics commands already exist, so I checked that mechanically against this branch rather than by reading:

- **Metric names** — all nine `formbricks_authzed_*` names referenced in the changed docs exist in `lib/authzed/metrics.ts` / `lib/authorization/metrics.ts` on this branch. The only two that don't resolve directly are `..._duration_seconds_bucket` pairs, which are the Prometheus bucket series of histograms that do exist, so the PromQL is correct.
- **CLI commands** — every `formbricks-authzed` subcommand the changed docs mention (`health`, `schema apply`, `schema check`, `backfill` with `--apply` / `--organization-id` / `--workspace-id`) exists in the shipped CLI here. **No `outbox` subcommand is documented anywhere in the diff**, which is exactly the claim, verified.
- **Public docs framing** — outbox, fail-closed and consistency language in `docs/self-hosting/**` is consistently scoped to "the planned v6 direct-authority release", and explicitly defers command names with "the detailed v6 commands remain release-specific until the release".
- **Internal links** — every relative `.md`/`.mdx` link in the changed files resolves.

`PERFORMANCE.md` invalidating its own benchmark against the new contract and handing revalidation to ENG-2453 is the right call, and easy to have skipped.

### Not covered by this approval

The 22 open threads are unresolved and none of them are mine to close — in particular Matti's question about whether these documents belong in the repository at all is a direction decision that could moot a lot of the line-level feedback. I'm approving that this carries no runtime or security risk, not that the contract's content is settled.

Two things I did not see raised anywhere, both cross-PR so understandable:

- **Invariant 3 says "a pending *revocation*", but #8912 implements `isRevocation` as `TG_OP <> 'INSERT'`.** I verified against PostgreSQL 16 that invite-acceptance and a `member → owner` upgrade are both recorded as revocations there. The contract is narrower and more correct than its implementation, which is useful — #8912 can be held to this invariant. Worth defining "revocation" explicitly here as a permission-narrowing change so it cannot drift again.
- **Invariant 4 ("including calls outside a request surface") quietly widens SpiceDB to non-request callers.** Today an unscoped check returns `legacyEvaluator.can(...)` unconditionally; once ENG-2446 removes legacy, scripts like `authzed:backfill` and `authzed:perf` plus background jobs start hitting SpiceDB and the freshness guard for the first time. No ticket is named as auditing those callers.

PR #8911, opened 2026-08-18
fix(authorization): scope the page surface to the request, not the callback (ENG-2444)

xernobyl:
Fixes ENG-2444

## What & why

The `page` authorization surface previously lived only for the awaited choke-point callback. Later authorization checks in the same React server render therefore appeared as `unscoped`, splitting checks-per-request telemetry and hiding the true page-level request boundary.

This PR stores the page context in a real React `cache()` request slot while retaining `AsyncLocalStorage` for API, MCP, gateway, and server-action surfaces.

- Layouts and pages in the same render share one page context and one checks-per-request observation.
- An enclosing non-page surface keeps precedence, including when another wrapper is nested inside the page callback.
- The cached slot factory is pinned on `globalThis`, so duplicated Next.js server bundles share the same React cache key.
- Callers outside a React request scope retain the previous async-scoped fallback.
- The dedicated Vitest RSC project resolves React with the `react-server` condition; the redundant top-level resolver condition has been removed.

The PR is stacked on #8936 because it depends on the direct-authority telemetry introduced by that stack. Merge #8936 first, then retarget this PR to `epic/authzed`.

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Page context survives the choke-point callback for the rest of the React request | unit (mutation): `context.rsc.test.ts` | A later check remains attributed to `page` and is included in the same observation |
| Layout and page checks share one request observation | unit (mutation): `context.rsc.test.ts` | Multiple page scopes produce one callback and the combined check count |
| Nested non-page wrappers cannot split an established page observation | unit (mutation): `context.rsc.test.ts` | The page surface keeps precedence and emits exactly one page metric |
| Concurrent and sequential requests do not leak context | unit (guard): `context.rsc.test.ts` | Each request starts with an independent cached slot |
| API/server-action precedence and non-RSC fallback remain unchanged | unit (guard): `context.test.ts` and `context.rsc.test.ts` | All boundary and fallback cases passed |
| Broader authorization helpers remain compatible | unit (guard): `pnpm --filter @formbricks/web exec vitest run lib/authorization modules/organization/lib modules/workspaces/lib` | 20 files and 156 tests passed |
| AuthZed runtime remains operational after the stacked change | integration: `pnpm authzed:smoke` | Schema lifecycle, projections, user/API-key decisions, revocations, outages, migrations, and persistence passed |
| Schema, types, and formatting remain valid | integration: `pnpm authzed:validate`, web typecheck, and Prettier | All passed; schema validation ran 157 assertions and 6 expected relations |

**Open gaps**

- Browser-level authoritative UI coverage remains owned by ENG-2457 because the shared Playwright workflow does not currently provision SpiceDB.
- Live sandbox validation follows after both stack PRs merge and does not belong to this request-context implementation PR.

**Risks**

- The RSC Vitest project intentionally does not inherit the shared setup that mocks React `cache()` to an identity function; future shared test setup changes may need an explicit RSC equivalent.
- The `globalThis` slot stores only the cache factory, not request data. Request data remains scoped by React's cache dispatcher.

---

> [!NOTE]
> **AI model used** — `GPT-5.6`, reasoning effort `high`.

### Restack status

Rebased cleanly onto reviewed #8936 head `75293099f`; the stacked head is now `43b7c8680`. The authorization-focused suite still passes (20 files, 156 tests). Merge #8936 first, then retarget this PR to `epic/authzed`.

### DIFF
--- a/apps/web/lib/authorization/context.ts
+++ b/apps/web/lib/authorization/context.ts
@@ -1,6 +1,7 @@
 import "server-only";
 import { after } from "next/server";
 import { AsyncLocalStorage } from "node:async_hooks";
+import { cache as reactCache } from "react";
 import { recordAuthorizationChecksPerRequest } from "./metrics";
 
 /**
@@ -9,22 +10,30 @@ import { recordAuthorizationChecksPerRequest } from "./metrics";
  * Unlike every other surface, it is not established at a single request boundary: Next.js gives no RSC
  * equivalent of the action-client or API wrapper, and a layout's render and its page's render are
  * separate async contexts. It is therefore opened at the authorization choke points every product
- * route already funnels through. `withAuthorizationSurface` returns early when a surface is already
- * open, so opening it at more than one choke point is idempotent rather than nesting.
+ * route already funnels through, and opening it more than once is idempotent rather than nesting.
  *
- * Two consequences follow, and the second is a real coverage limit rather than a reporting quirk.
+ * **`page` is request-scoped; every other surface is callback-scoped.** That difference is ENG-2444.
+ * `AsyncLocalStorage` closes when the awaited callback returns, so the surface used to end with the
+ * choke-point helper: a page that awaited `getWorkspaceAuth()` and then authorized anything else did
+ * so with no surface, and those decisions were labelled `unscoped` in the authoritative decision
+ * telemetry the direct-authority rollout is monitored on. Nine routes were affected, two of them
+ * issuing one such check *per feedback directory* or *per dashboard widget*.
  *
- * The histogram splits: one navigation that runs both a layout check and a page check records two
- * observations rather than one. That weakens the N+1 signal slightly; it does not make it wrong.
+ * `page` is now held in a React `cache()` slot, which is scoped to the whole render pass, so a layout
+ * and its page share one context and it outlives every helper. Two consequences, both improvements on
+ * what this comment used to record:
  *
- * **The surface does not span the whole page.** `AsyncLocalStorage` scopes to the awaited callback,
- * so it closes when the choke-point helper returns — a page that calls `getWorkspaceAuth()` and then
- * issues further central checks makes those checks without a page label. They still use authoritative
- * SpiceDB, but are tagged `unscoped` in decision telemetry.
+ * A navigation records ONE checks-per-request observation rather than one per choke point, which is
+ * the N+1 signal that histogram exists for.
  *
- * Closing it needs a boundary that survives past the helper — a request-scoped store (React `cache`)
- * rather than an async-scoped one. That telemetry attribution improvement is separate from evaluator
- * authority and requires a render harness or E2E proof.
+ * `getAuthorizationSurface()` reports `page` for the whole render instead of `unscoped` after the first
+ * helper returns, so `formbricks_authzed_authorization_decisions_total` attributes page traffic
+ * correctly.
+ *
+ * Outside a React request scope — scripts, unit tests, any non-RSC caller — `cache()` does not
+ * memoize, so there is no slot to hold. `page` then falls back to the `AsyncLocalStorage` boundary,
+ * which is the pre-ENG-2444 behaviour: narrower than a render scope, but correct for a caller with no
+ * render to scope to.
  */
 export type TAuthorizationSurface =
   | "server_action"
@@ -49,6 +58,58 @@ const authorizationContext =
 
 globalForAuthorization.formbricksAuthorizationContext = authorizationContext;
 
+type TPageSurfaceSlot = { context: TAuthorizationContext | null };
+
+/**
+ * One slot per React request scope. In an RSC render that is the whole render pass, so a layout and
+ * its page resolve the same slot — which is what lets the `page` surface outlive the choke-point
+ * helper that opened it.
+ *
+ * `reactCache` is already this codebase's memoization idiom (resolvers.ts, and both choke-point
+ * modules); here it is used for its scope rather than to cache a value.
+ */
+const getPageSurfaceSlot = reactCache((): TPageSurfaceSlot => ({ context: null }));
+
+/**
+ * Whether React is holding a request scope we can hang the `page` surface on.
+ *
+ * `cache()` only memoizes inside one — outside it (scripts, unit tests, the non-RSC bundle, where the
+ * client build's `cache` is a permanent no-op) every call returns a fresh object, so identity is a
+ * direct, dependency-free probe. Deliberately not named "is rendering": Next also establishes a scope
+ * outside component rendering, and `page` is only ever *selected* by the caller — the ALS store is
+ * consulted first, so an enclosing server-action or API surface always keeps precedence.
+ */
+const hasReactRequestScope = (slot: TPageSurfaceSlot): boolean => slot === getPageSurfaceSlot();
+
+const createSurfaceContext = (surface: TAuthorizationSurface): TAuthorizationContext => ({
+  checksIssued: 0,
+  surface,
+});
+
+/**
+ * Register the one post-response observation for a surface. Shared by both boundaries so they cannot
+ * drift, and fail-safe: an unavailable `after()` costs the histogram observation, never the decision.
+ */
+const scheduleChecksPerRequestObservation = (context: TAuthorizationContext): void => {
+  try {
+    after(() => {
+      try {
+        recordAuthorizationChecksPerRequest(context.checksIssued, context.surface);
+      } catch {
+        // Telemetry must never alter an authoritative response.
+      }
+    });
+  } catch {
+    // A wrapper can be invoked outside a Next.js request in scripts/tests — `after()` is unavailable
+    // there. The authorization decision remains authoritative; only the request histogram observation
+    // is omitted.
+  }
+};
+
+/** The surface answering for the current caller: an explicit ALS boundary first, else the page slot. */
+const getActiveContext = (): TAuthorizationContext | null =>
+  authorizationContext.getStore() ?? getPageSurfaceSlot().context;
+
 export const withAuthorizationSurface = async <T>(
   surface: TAuthorizationSurface,
   callback: () => T | Promise<T>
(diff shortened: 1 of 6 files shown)


apps/web/lib/authorization/context.ts
```diff
@@ -49,6 +58,58 @@ const authorizationContext =
...
+
+/**
+ * One slot per React request scope. In an RSC render that is the whole render pass, so a layout and
+ * its page resolve the same slot — which is what lets the `page` surface outlive the choke-point
+ * helper that opened it.
+ *
+ * `reactCache` is already this codebase's memoization idiom (resolvers.ts, and both choke-point
+ * modules); here it is used for its scope rather than to cache a value.
+ */
+const getPageSurfaceSlot = reactCache((): TPageSurfaceSlot => ({ context: null }));
```
kiln-agent: **Page-surface slot lacks the `globalThis` singleton guard the ALS relies on.**

The `AuthorizationContext` above is deliberately pinned to `globalThis` (lines 52-59) because Next.js can evaluate this server-only module in more than one server bundle, and the writer (`withAuthorizationSurface`) and readers (`recordAuthorizationCheckIssued` / `getAuthorizationSurface`) must share one instance. `getPageSurfaceSlot` has no such guard.

Failure scenario: if the choke-point caller binds to instance A's `withAuthorizationSurface` (setting the slot via cache-wrapper A) while `can()` in `index.ts` resolves `recordAuthorizationCheckIssued` / `getAuthorizationSurface` via instance B (cache-wrapper B — a different React-cache memo key), the post-choke-point checks resolve a fresh, empty slot and are silently attributed to `unscoped` again — the exact regression ENG-2444 claims to fix, invisibly. The ALS surfaces are immune to this because of the `globalThis` guard; the page slot is not.

Consider pinning `getPageSurfaceSlot` (or the underlying `cache()`-wrapped value) to `globalThis` the same way `AuthorizationContext` is, so both surfaces survive module duplication.

---

apps/web/vite.config.mts
```diff
@@ -37,6 +38,27 @@ export default defineConfig({
...
+        // does anything in the react-server build — the default build ships it as a permanent no-op,
+        // and the shared vitestSetup mocks it to identity on top of that. So the `unit` project cannot
+        // exercise this surface at all. Here React resolves the way Next.js resolves it for RSC, so
+        // the real implementation runs. `ssr.resolve.conditions` is the knob that works: Vitest loads
+        // test modules through the SSR environment, so a top-level `resolve.conditions` is ignored.
+        //
+        // Deliberately NOT `extends: true`: that merges the root `setupFiles`, which import
+        // react-dom/client (forbidden under the react-server condition) and mock `cache` away.
+        plugins: [tsconfigPaths()],
+        resolve: { conditions: ["react-server", "node", "import", "default"] },
```
kiln-agent: **Redundant `resolve.conditions` contradicts the comment right above it.**

Lines 46-47 state: `Vitest loads test modules through the SSR environment, so a top-level resolve.conditions is ignored`. Line 52 nonetheless declares `resolve: { conditions: [...] }`. By the diff's own stated rationale this line is dead for the behavior it appears to enable — only line 53's `ssr.resolve.conditions` is load-bearing for the `react-server` resolution during Vitest.

A maintainer reading this will believe the top-level `resolve.conditions` is doing work. Either remove it, or replace the comment with a note explaining why it's kept (e.g. for a non-Vitest consumer of this same config).

---

apps/web/lib/authorization/context.ts
```diff
@@ -57,23 +118,25 @@ export const withAuthorizationSurface = async <T>(
...
+  if (surface === "page") {
+    const slot = getPageSurfaceSlot();
+    if (hasReactRequestScope(slot)) {
+      // Opened once per render, then left open: every later check in the same render — including the
+      // ones a page issues long after this helper returned — resolves through this slot.
+      if (!slot.context) {
+        slot.context = createSurfaceContext(surface);
+        scheduleChecksPerRequestObservation(slot.context);
+      }
+      return callback();
```
kiln-agent, line 143: **Latent: nested non-page surface inside a page callback splits attribution and double-counts.**

Pre-change, the page callback ran inside `authorizationContext.run`, so a nested `withAuthorizationSurface(<non-page>)` hit the same-surface early-return and inherited `page` with a single observation. Post-change, the page branch returns `callback()` without setting the ALS — so a nested non-page surface sees an empty store, opens its own ALS context, and records a **second** per-request observation labelled as that nested surface.

This doesn't manifest with today's choke-point call shapes (page callbacks are tight `can()` calls with no nested `withAuthorizationSurface`), so it's latent. But any future code that opens a non-page surface inside a page-surface callback will silently split attribution and double-count the request, contradicting the "one observation per request" invariant the fix advertises.

If you want to keep the page branch cheap, consider at least a dev-only assert that no nested `withAuthorizationSurface` is called while `getPageSurfaceSlot()` is set, so the invariant fails loudly if a future call site breaks it.

---

kiln-agent (commented):
Reviewed the request-scoped page-surface change. Core logic — page surface held in a React `cache()` slot, ALS precedence, fallback outside a request scope, fail-safe telemetry — is sound and the tests exercise the important precedence and isolation cases. Correctness-critical SpiceDB behavior is unaffected by design.

Three findings, none crash-level:

1. **`context.ts:71`** — `getPageSurfaceSlot` lacks the `globalThis` singleton guard the ALS uses (lines 52-59). If Next.js duplicates this server-only module across bundles, writer and readers could bind to different `cache()` wrappers and silently regress to `unscoped` attribution — the exact case ENG-2444 is meant to close.
2. **`vite.config.mts:52`** — The top-level `resolve.conditions` block contradicts the comment right above it that says Vitest ignores it. Either remove it or explain why it stays.
3. **`context.ts:130`** — Latent: a nested non-page `withAuthorizationSurface` inside a page-surface callback would now open its own ALS context and record a second observation, breaking the "one observation per request" invariant. No current call site hits this, so it's a footgun for future code — consider a dev-only assert.

Details inline.

---

kiln-agent:
## Review — no findings

Read the diff, `context.ts`, both test files, `vite.config.mts`, `vitestSetup.rsc.ts`, the README delta, and the existing callers (`modules/organization/lib/utils.ts`, `modules/organization/settings/api-keys/page.tsx`, the onboarding layouts, and the integration tests under `lib/authorization/`). Everything I hunted for came back clean, and each concern I raised on the way survived a serious attempt to refute it before I discarded it.

**What I checked, and what did not survive refutation:**

- **Does the RSC test actually catch the bug?** Yes. Revert the request-scoped branch and: `getAuthorizationSurface()` after the callback returns `"unscoped"` (test 1 fails), two page choke points schedule two `after()` callbacks (test 2's `toHaveLength(1)` fails), and `recordAuthorizationChecksPerRequest` gets called twice with mismatched counts. Not a test that cannot fail.
- **Is `hasReactRequestScope` a valid probe?** Yes, in both React builds. The react-server build without a dispatcher returns `fn.apply(null, arguments)` — a fresh `{ context: null }` each call → identity comparison fails → probe returns false. The client build ships `cache` as a permanent no-op with the same behaviour. The `unit` project's identity-mock of `cache` also returns fresh objects each call, so the fallback path is uniform across all three environments.
- **Race on `slot.context` between two concurrent choke points in the same render?** No — the check-and-set from `slot = getPageSurfaceSlot()` through the assignment is synchronous, no `await` in between; JS's single-threaded model rules out a race even when the outer awaits interleave.
- **Does the page slot leak precedence to a nested `server_action`?** No, in the direction that matters. The tested case (server_action outer, page inner) early-returns on `authorizationContext.getStore()`. The reverse (page outer, server_action inner) does correctly attribute the inner checks to `server_action` via `getActiveContext`'s ALS-first check — and it doesn't happen in practice, since server actions are client-triggered rather than called from a page render.
- **`??` in `getActiveContext` — extra `getPageSurfaceSlot()` call per authorization decision under ALS surfaces?** No: `??` short-circuits, so `server_action`, `api_*`, `mcp`, and `feedback_gateway` don't pay for a slot lookup at all.
- **A failed `after()` on the first page choke point permanently suppresses scheduling for that request?** True mechanically (`slot.context` is set before `scheduleChecksPerRequestObservation` runs, so subsequent choke points skip the schedule), but `after()`'s availability is a per-request property — a call that throws once throws every time in the same request, so the observation was unrecoverable regardless.
- **Vitest project isolation.** The `rsc` project deliberately does not `extends: true`, `unit` excludes `**/*.rsc.test.ts`, and `components` matches only `*.test.tsx`. No overlap. The setup file mocks only `server-only` — leaving `react` un-mocked is the whole point, and the guard test at the top of the RSC suite (`slot() !== slot()` outside scope, `slot() === slot()` inside) fails vacuously if the `react-server` condition regresses.
- **Integration tests still work.** They call `withAuthorizationSurface("page", ...)` with no React request scope; `getPageSurfaceSlot()` returns a fresh object each call → `hasReactRequestScope` false → falls through to the pre-ENG-2444 ALS boundary. `getIssuedAuthorizationCheckCount()` called inside the callback goes through `getActiveContext()` → ALS store → correct count. Pre-existing contract preserved.
- **Prose vs. code.** `context.ts` docstring and the README delta both reflect the new behaviour accurately, and they carry the "telemetry-accuracy, not security" framing consistently with the PR body's rebased-onto-ENG-2448 note.

Ready to merge from my side.

---

apps/web/lib/authorization/context.ts
BhagyaAmarasinghe: Fixed in ad5d3a81e. The React cache wrapper is now stored on the shared global authorization object, so duplicated Next.js server module instances use the same cache key while request data remains React-cache scoped. Added RSC coverage still verifies request isolation.

apps/web/vite.config.mts
BhagyaAmarasinghe: Fixed in ad5d3a81e. Removed the redundant top-level resolve.conditions entry; the RSC project now relies only on ssr.resolve.conditions, matching the documented Vitest resolution path.

apps/web/lib/authorization/context.ts
BhagyaAmarasinghe: Fixed in ad5d3a81e. An established page request context now takes precedence over nested non-page wrappers, so the nested callback reuses the page observation instead of opening a second ALS observation. Added a regression test that asserts one page metric and one after callback.

BhagyaAmarasinghe (approved):

PR #8912, opened 2026-08-18
feat: make AuthZed relationship delivery durable

BhagyaAmarasinghe:
Fixes ENG-2408

## What & why

- PostgreSQL source mutations currently project relationships synchronously and best-effort, so a SpiceDB outage can leave durable authorization drift with no guaranteed replay.
- Adds a transactional PostgreSQL outbox populated by database triggers for every Phase 1 relationship source, leased delivery with revocation priority, bounded retry/dead-letter handling, and idempotent reconciliation through the existing projectors.
- Adds a 60-second stale-revocation authorization guard, a six-hour full audit/repair job, bounded operational metrics, and sanitized `formbricks-authzed outbox status|drain|replay` commands.
- Keeps PostgreSQL authoritative, legacy authorization authoritative, and SpiceDB startup/health independent while producing the durable bridge required by the direct-cutover stack.

The guard is the load-bearing part, and it is global and unscoped: `spicedb-evaluator.ts` calls it on every permission check and `coordinator.ts` rethrows under enforcement with no legacy fallback, so **anything that arms it — spuriously or permanently — denies every enforced check in the deployment**. Three properties exist to keep that from happening:

- **An undelivered event only arms the guard if it could actually leave stale access.** The trigger classifies through a deny-by-default predicate rather than treating every `UPDATE` as a revocation. Only three transitions are grants, each proven from the projectors' own write shape: reactivating a user, unarchiving a feedback directory that stays in its organization, and any membership update leaving `role` unchanged (the projected snapshot ignores `accepted`, so accepting an invite writes byte-identical relationships). A mass invite acceptance therefore cannot arm the guard. The role ladder is deliberately **not** encoded in SQL — it is rankable, but a rank table in a migration has no compile-time backstop the way `relationship-map.ts` does, so a change to `schema.zed` would silently make it wrong in the fail-*open* direction.
- **A failure is charged only to the events it is attributable to.** A claimed batch delivers as six independent groups; a failing group takes only its own events. A retryable failure releases every remaining group untried. When a code could plausibly belong to one event (`authzed_projection_invalid_source`, `authzed_invalid_request`) the group is halved until the culprit is alone; codes that describe the instance are not split.
- **Dead-lettering requires the failure to *name* an event.** A new `permanentFailures` column gates it, and three things must hold together before it increments: the code is non-retryable, the attempt covered exactly one event, and the code is one an event can actually cause. The third condition is the one that is easy to miss — on a five-second cadence most groups hold a single event, so size alone would charge whichever revocations happened to be travelling alone when SpiceDB rejected a credential. The consequence is the property worth remembering: **no SpiceDB outage, of any duration or kind, can dead-letter an event that was never the problem.** The six-hour audit replays dead letters after a clean run, bounding a fail-closed denial at six hours rather than at operator attention.

## Breaking changes

None. The outbox table, its columns and its indexes are all introduced by this PR, so there is nothing deployed for them to break. No API, SDK, route, env var, config key, default or webhook payload changes.

## Migrations & env

- Adds the `AuthzedProjectionOutbox` table (including the `permanentFailures` column), three indexes — `_claim_idx`, `_undelivered_idx`, `_processed_idx` — and trigger functions in migration `20260818120000_add_authzed_projection_outbox`.
- The migration installs idempotent triggers on organization, membership, user, team, workspace, API-key, feedback-directory, grant, and deletion-cascade sources, plus the `authzed_projection_is_grant` classifier function.
- All three indexes are **partial** (`WHERE "processedAt" IS NULL AND …`), which Prisma's `@@index` cannot express — so `model AuthzedProjectionOutbox` deliberately declares no indexes and points at the migration. The convention and its `db push` caveat are documented in `packages/database/README.md`. A dev who runs `db push` restores them by rerunning the migration, the same contract the triggers already live under.
- No new environment variables. Existing AuthZed and jobs/Redis configuration is reused.
- Deploy the jobs worker with the application migration; legacy authorization remains authoritative on this branch.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Grants are not classified as revocations | integration (real PostgreSQL): `outbox-trigger.integration.test.ts` | Drives the **real triggers**, not the SQL function directly. Invite acceptance and reactivation enqueue `isRevocation = false`; role moves, unmapped enum moves, unarchive-plus-tenant-move and deletes enqueue `true`; a key move enqueues the old pair `true` and the new pair `false`. |
| A grant pending far past the window leaves the guard disarmed | integration: `outbox-trigger.integration.test.ts` | An hour-old non-revocation returns `false` from `hasStaleAuthzedRevocation`; an hour-old revocation and a *fresh* dead letter both return `true`. |
| One poison event does not take healthy events with it | unit (mutation): `outbox-processor.test.ts` — `pnpm --filter=@formbricks/web test lib/authzed` | A failing group is released while the other five are delivered. A 4-event group failing on `authzed_projection_invalid_source` isolates one and delivers three. A group failing on `authzed_internal` is **not** split — asserted at exactly one reconciler call, which is the test that stops a 2×N regression. |
| A SpiceDB outage never dead-letters | integration + unit (mutation): `outbox-trigger.integration.test.ts`, `outbox-repository.test.ts`, `outbox-processor.test.ts` | 15 consecutive retryable failures leave `permanentFailures` at 0 and `deadLetteredAt` null. A group failure at `MAX-1` also does not dead-letter. Crucially, neither does a *single-event* group failing on `authzed_unauthenticated` / `authzed_internal` / `authzed_permission_denied` — those describe the instance, not the event. Only solitary, event-attributable failures dead-letter, at the tenth. `authzed_projection_unstable` is treated as retryable at the outbox boundary without changing the projector contract. |
| The guard ignores revocations that were actually delivered | integration: `outbox-trigger.integration.test.ts` | An hour-old revocation with `processedAt` set leaves the guard disarmed. Delivered rows are retained seven days, so without this predicate a healthy deployment would deny permanently — and every other guard test uses undelivered rows, so nothing else would have noticed. |
| The 5-second status aggregate does not scan retained history | integration: `outbox-trigger.integration.test.ts`; measured with `EXPLAIN (ANALYZE, BUFFERS)` | Against 200,000 delivered + 50 live rows the aggregate is an **Index Only Scan** on `_undelivered_idx` touching 5 buffers / 50 rows (0.04 ms), versus a parallel seq scan over all 200,050 before. The index predicate is the whole undelivered set precisely because a bare `processedAt IS NULL` implies neither complementary partial predicate. |
| Backoff grows per row and is capped | integration: `outbox-trigger.integration.test.ts` | Computed in SQL from each row's own `attempts`; a later attempt backs off further, and `attempts = 40` stays under the five-minute ceiling instead of overflowing. A release by a lease the row no longer owns changes nothing. |
| Freshness guard costs one query, not one per check | unit (mutation): `outbox-freshness.test.ts` | Five concurrent checks collapse to one query; a re-read happens after the TTL; a **failed** read is not memoized and still denies. React `cache()` is gone — the Route Handler runtime carries no cache dispatcher and 9 of 11 rollout targets are Route-Handler-only, so it was previously one uncached query per authorization *check*. |
| Claims are served in index order | integration: `outbox-trigger.integration.test.ts` | With a 2,000-row backlog and `ANALYZE`, `EXPLAIN` shows `AuthzedProjectionOutbox_claim_idx` and **no `Sort` node**. Seeded deliberately — on an empty table the planner picks a seq scan whatever indexes exist, so an unseeded plan assertion measures nothing. |
| Dead letters clear themselves after a clean audit | unit (mutation): `scheduled-reconciliation.test.ts` | Replay runs on `reconciled` and not on `drifted` or `failed`. |
| Source mutation and outbox insertion share one transaction | unit (mutation): `outbox-migration.test.ts`; manual PostgreSQL 16 migration exercise | Rollback produced zero events; committed writes across all 11 trigger sources produced durable events. |
| Leases, key moves, irrelevant updates, replay | unit (mutation): `outbox-repository.test.ts`, `outbox-processor.test.ts`, `outbox-cli*.test.ts` | Revocations are claimed first, key moves enqueue old/new targets, irrelevant API-key use updates are ignored, replay restores both `attempts` and the permanent-failure budget. |
| Six-hour reconciliation repairs only attributable safe drift | unit (mutation): `scheduled-reconciliation.test.ts` | Missing and mismatched-permission edges repair; parent mismatches, invalid, unmanaged, and destructive prune remain manual. |
| Worker registration and recurring cadence | unit (mutation): `instrumentation-jobs.test.ts`, `packages/jobs` tests | Delivery is scheduled every 5 seconds and reconciliation every 6 hours without duplicate registrations. |
| CLI and telemetry remain bounded and secret-safe | unit (mutation): CLI and metrics tests | Output carries aggregate counts and stable codes only. The delivery warning logs a sorted set of enumerable error codes — no target IDs, relationships, tokens, or raw SDK errors. |
| Canonical schema and live SpiceDB behavior | integration: `pnpm authzed:validate`, `pnpm authzed:smoke` | 36 relationships, 157 assertions, and 6 expected relations validated; isolated persistence/outage/recovery smoke passed. |
| Deployment manifests retain their AuthZed contracts | static integration: `docker/authzed-compose-contract.sh`; Helm lint | Production/dev Compose contracts and both Formbricks/operator charts validated. |
| Migration is rerunnable | manual PostgreSQL 16 exercise | Applied twice in a row against a live database; second run is a clean no-op (`CREATE TABLE IF NOT EXISTS`, `ADD COLUMN IF NOT EXISTS`, `CREATE OR REPLACE FUNCTION`, `DROP TRIGGER IF EXISTS`). |
| No build-time AuthZed RPC | production builds with AuthZed disabled and with `127.0.0.1:9` configured | Both builds completed successfully without contacting SpiceDB. |

Every `unit (mutation)` and `integration` row above was checked by reverting the behaviour it names and confirming the named test goes red — including the three fixes in `cc40a095`, which came out of a self-review pass over `3a7c428` and are the reason that commit exists. Whole-suite state: `apps/web` unit suite 679 files / 8,974 tests green; outbox integration 17 tests green; `packages/database` and `packages/jobs` green; typecheck, ESLint and `prettier --check .` clean.

**Open gaps**

- **The freshness guard is still global.** Tenant-scoping it was considered and deliberately not done — it is not fully achievable (user deactivation is genuinely cross-tenant, and 5 of the 11 sources carry no `organizationId` the trigger can reach without a parent lookup on every source mutation). Fixing the classification removes the bulk amplifier that made a global guard a practical outage; scoping remains available later if a real single-tenant backlog appears.
- **Parent moves are unprojectable, not merely unclassified.** `createParentUpdate`/`parentUpdate` only ever *touch* the new parent, so moving a Team/Workspace/ApiKey/FeedbackDirectory between organizations leaves the old organization's owners with access. These are classified as revocations, but the event then delivers "successfully" and disarms the guard — so the guard does not protect against this class. `backfill.ts` already reports it as `mismatchedParents` and deliberately does not repair it. Pre-existing; worth its own ticket.
- The immutable bridge image cannot be frozen or deployed to `authzed-sandbox` until this stacked PR and its prerequisite PRs merge.
- Live crash/replay, alert timing, rollback, and 24-hour sandbox-soak evidence belongs to ENG-2469 and ENG-2470 after the bridge image exists.
- Staging and production rollout are intentionally blocked by the sandbox exit gate.
- Four integration files (`modules/auth/*`, `modules/ee/workflows/*`) fail locally on missing SMTP and mail-URL configuration. Confirmed pre-existing: identical failures on the unmodified parent commit.

**Risks**

- Database triggers are the atomicity boundary; source-table schema changes must update the trigger contract, the classifier's watched-column reasoning, and the migration regression test.
- The classifier is deny-by-default, so the failure mode of getting it wrong is denial, not leakage — but adding a target type to the grant `CASE` without re-reading that projector's write shape would invert that.
- Projection is at-least-once. A worker can replay after a lost lease, so projectors must remain idempotent.
- A stale/dead-letter revocation intentionally fails protected SpiceDB operations closed while `/health`, readiness, and startup stay independent.
- The freshness memo widens the effective window by up to 1s on a 60s budget — bounded and explicit, in exchange for removing an uncached query from every authorization check. It reads the monotonic clock, so a backwards NTP step cannot extend that bound.
- Attribution is deliberately conservative: an event carrying a genuinely poisonous payload under a code *outside* `PER_EVENT_ERROR_CODES` will retry indefinitely at the five-minute ceiling rather than dead-letter. That keeps it visible (the guard stays armed if it is a revocation) instead of silently parked, but it means the dead-letter queue is not a catch-all for every stuck event.
- The scheduled repair never prunes or repairs tenant-parent mismatches automatically; operators must investigate those security-sensitive cases.

---

> [!NOTE]
> **AI model used** — original implementation GPT-5.6, reasoning effort high; review-fix commits `3a7c428` and `cc40a095` Claude Opus 5, reasoning effort high.

**CI follow-up**

- The E2E and Better Auth integration workflows now start a pinned ephemeral SpiceDB fixture and apply the canonical schema before application code runs. This removes projection retry amplification while preserving real AuthZed behavior.
- Local evidence: the fixture health/schema bootstrap passed, and the real Better Auth integration suite completed with 32 files and 141 tests passing. Failure diagnostics include sanitized SpiceDB container logs.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -142,6 +142,10 @@ jobs:
           echo "S3_FORCE_PATH_STYLE=1" >> .env
         shell: bash
 
+      - name: Start AuthZed CI fixture
+        run: bash scripts/start-authzed-ci.sh
+        shell: bash
+
       - name: Start RustFS Server
         run: |
           set -euo pipefail
@@ -382,3 +386,8 @@ jobs:
           else
             echo "app.log not found because the Run App step did not execute or failed before log creation."
           fi
+
+      - name: Output AuthZed logs
+        if: failure()
+        run: docker logs formbricks-authzed-ci 2>&1 || true
+        shell: bash
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -123,6 +123,11 @@ jobs:
           sed -i "s|PASSWORD_RESET_DISABLED=.*|PASSWORD_RESET_DISABLED=0|" .env
         shell: bash
 
+      - name: Start AuthZed CI fixture
+        if: steps.harness.outputs.present == 'true'
+        run: bash scripts/start-authzed-ci.sh
+        shell: bash
+
       # Build the workspace packages the tests import (@formbricks/logger, cache, types, email, …) so
       # vite can resolve their package.json `exports` → dist. Deps only (`^...`), not the Next app.
       - name: Build workspace package dependencies
@@ -165,4 +170,9 @@ jobs:
         run: cd apps/web && pnpm test:integration
         env:
           TEST_DB_PROVISIONED: "1"
+        shell: bash
+
+      - name: Output AuthZed logs
+        if: failure() && steps.harness.outputs.present == 'true'
+        run: docker logs formbricks-authzed-ci 2>&1 || true
         shell: bash
--- a/apps/web/instrumentation-jobs.test.ts
+++ b/apps/web/instrumentation-jobs.test.ts
@@ -7,6 +7,8 @@ const mockRemoveSurveyArchivePurge = vi.fn();
 const mockUpsertSurveyArchivePurge = vi.fn();
 const mockRemoveWorkflowRunReconcile = vi.fn();
 const mockUpsertWorkflowRunReconcile = vi.fn();
+const mockUpsertAuthzedProjectionDelivery = vi.fn();
+const mockUpsertAuthzedReconciliationAudit = vi.fn();
 const mockDebug = vi.fn();
 const mockError = vi.fn();
 const mockWarn = vi.fn();
@@ -17,6 +19,8 @@ const mockProcessSurveySchedulingJob = vi.fn();
 const mockProcessSurveyArchivePurgeJob = vi.fn();
 const mockProcessWorkflowRunJob = vi.fn();
 const mockProcessWorkflowRunReconcileJob = vi.fn();
+const mockProcessAuthzedProjectionDeliveryJob = vi.fn();
+const mockProcessAuthzedScheduledReconciliationJob = vi.fn();
 const TEST_TIMEOUT_MS = 15_000;
 
 const slowTest = (name: string, fn: () => Promise<void>): void => {
@@ -31,6 +35,18 @@ vi.mock("@formbricks/jobs", () => ({
     workflowRun: "workflow-run.process",
   },
   recurringJobs: {
+    authzedProjectionDelivery: {
+      name: "authzed-projection.deliver",
+      scheduleId: "authzed-projection-delivery",
+      scope: "global",
+      upsert: mockUpsertAuthzedProjectionDelivery,
+    },
+    authzedReconciliationAudit: {
+      name: "authzed-reconciliation.audit",
+      scheduleId: "authzed-reconciliation-audit",
+      scope: "global",
+      upsert: mockUpsertAuthzedReconciliationAudit,
+    },
     surveyArchivePurge: {
       name: "survey-archive-purge.process",
       remove: mockRemoveSurveyArchivePurge,
@@ -88,6 +104,14 @@ vi.mock("@/modules/ee/workflows/lib/runner/process-workflow-run-job", () => ({
 
 vi.mock("@/modules/ee/workflows/lib/runner/process-workflow-run-reconcile-job", () => ({
   processWorkflowRunReconcileJob: mockProcessWorkflowRunReconcileJob,
+}));
+
+vi.mock("@/lib/authzed/outbox-processor", () => ({
+  processAuthzedProjectionDeliveryJob: mockProcessAuthzedProjectionDeliveryJob,
+}));
+
+vi.mock("@/lib/authzed/scheduled-reconciliation", () => ({
+  processAuthzedScheduledReconciliationJob: mockProcessAuthzedScheduledReconciliationJob,
 }));
 
 describe("instrumentation-jobs", () => {
@@ -176,6 +200,8 @@ describe("instrumentation-jobs", () => {
     expect(mockStartJobsRuntime).toHaveBeenCalledWith({
       concurrency: 4,
       jobHandlerOverrides: {
+        "authzed-projection.deliver": expect.any(Function),
+        "authzed-reconciliation.audit": expect.any(Function),
         "response-pipeline.process": expect.any(Function),
         "survey-scheduling.reconcile": expect.any(Function),
         "survey-archive-purge.process": expect.any(Function),
@@ -416,6 +442,16 @@ describe("instrumentation-jobs", () => {
       // The schedule identity and payload now belong to the job declaration in @formbricks/jobs (and are
       // asserted there); what this app owns, and what is asserted here, is the timing per job.
       expect(mockStartJobsRuntime).not.toHaveBeenCalled();
+      expect(mockUpsertAuthzedProjectionDelivery).toHaveBeenCalledOnce();
+      expect(mockUpsertAuthzedProjectionDelivery).toHaveBeenCalledWith({
+        everyMs: 5_000,
+        kind: "every",
+      });
+      expect(mockUpsertAuthzedReconciliationAudit).toHaveBeenCalledOnce();
+      expect(mockUpsertAuthzedReconciliationAudit).toHaveBeenCalledWith({
+        everyMs: 6 * 60 * 60 * 1_000,
+        kind: "every",
+      });
       expect(mockUpsertSurveyScheduling).toHaveBeenCalledTimes(1);
       expect(mockUpsertSurveyScheduling).toHaveBeenCalledWith({
         cronPattern: SURVEY_SCHEDULING_DAILY_CRON_PATTERN,
--- a/apps/web/integration/reset-db.ts
+++ b/apps/web/integration/reset-db.ts
@@ -15,6 +15,13 @@ import { prisma } from "@formbricks/database";
  * INVARIANT: this relies on every table a flow writes being FK-cascade-reachable from one of these
  * three roots. A future flow that writes a non-descendant table (or one with `onDelete: SetNull` /
  * `Restrict`) must be added here, or its rows will bleed across tests.
+ *
+ * `AuthzedProjectionOutbox` is the first table to hit that invariant. It has no foreign keys at all —
+ * the durable projection queue records identifiers by value so a row survives the deletion it
+ * describes — so nothing above cascades to it, and database triggers write to it on every mutation
+ * the three roots cascade through.
  */
 export const resetDb = (): Promise<unknown> =>
-  prisma.$executeRawUnsafe('TRUNCATE "User", "Organization", "Team" RESTART IDENTITY CASCADE;');
+  prisma.$executeRawUnsafe(
+    'TRUNCATE "User", "Organization", "Team", "AuthzedProjectionOutbox" RESTART IDENTITY CASCADE;'
+  );
--- a/apps/web/lib/authorization/resource-inventory.ts
+++ b/apps/web/lib/authorization/resource-inventory.ts
@@ -24,6 +24,7 @@ export const PRISMA_AUTHORIZATION_RESOURCE_INVENTORY = {
   ActionClass: "workspace_inherited_resource",
   ApiKey: "relationship_or_grant_source",
   ApiKeyWorkspace: "relationship_or_grant_source",
+  AuthzedProjectionOutbox: "authentication_or_application",
   Chart: "workspace_inherited_resource",
   Contact: "workspace_inherited_resource",
   ContactAttribute: "parent_derived_or_data_integrity",
--- a/apps/web/lib/authorization/spicedb-evaluator.test.ts
+++ b/apps/web/lib/authorization/spicedb-evaluator.test.ts
@@ -12,6 +12,7 @@ vi.mock("@/lib/constants", () => ({
   },
 }));
 vi.mock("@/lib/authzed/client", () => ({ getAuthzedClient: vi.fn() }));
+vi.mock("@/lib/authzed/outbox-freshness", () => ({ assertAuthzedProjectionFreshness: vi.fn() }));
 vi.mock("./source-scope", () => ({ resolveAuthorizationScope: vi.fn() }));
 
 const checkPermission = vi.fn();
--- a/apps/web/lib/authorization/spicedb-evaluator.ts
+++ b/apps/web/lib/authorization/spicedb-evaluator.ts
@@ -1,5 +1,6 @@
 import "server-only";
 import { getAuthzedClient } from "@/lib/authzed/client";
+import { assertAuthzedProjectionFreshness } from "@/lib/authzed/outbox-freshness";
 import { USER_MANAGEMENT_MINIMUM_ROLE } from "@/lib/constants";
 import {
   AUTHORIZATION_PERMISSION_MAP,
@@ -78,6 +79,8 @@ export const checkSpicedbPermissionAtScope = async <TAction extends TAuthorizati
   scope: TResolvedAuthorizationScope
 ): Promise<boolean> => {
   if (!scope.actorValid) return false;
+
+  await assertAuthzedProjectionFreshness();
 
   const permission = getPermission(actor, action, resource.type);
   if (!permission) return false;
--- a/apps/web/lib/authzed/backfill-apply.ts
+++ b/apps/web/lib/authzed/backfill-apply.ts
@@ -1,0 +1,29 @@
+import "server-only";
+import { reconcileApiKeyRelationships } from "./api-key";
+import type { TAuthzedBackfillApply } from "./backfill";
+import {
+  deleteFeedbackDirectoryAssignmentRelationships,
+  reconcileFeedbackDirectoryRelationships,
+} from "./feedback-directory";
+import { reconcileOrganizationMemberships } from "./organization-membership";
+import { reconcileTeamWorkspaceRelationships } from "./team-workspace";
+
+const INERT_RESULT = { passes: 0, status: "projected" } as const;
+
+/** No-op write capability used when the shared orchestrator runs in dry-run mode. */
+export const createAuthzedBackfillNoopApply = (): TAuthzedBackfillApply => ({
+  deleteFeedbackDirectoryAssignmentResources: async () => INERT_RESULT,
+  reconcileApiKeys: async () => INERT_RESULT,
+  reconcileFeedbackDirectories: async () => INERT_RESULT,
+  reconcileMemberships: async () => INERT_RESULT,
+  reconcileTeamWorkspace: async () => INERT_RESULT,
+});
+
+/** Internal write capability shared by the operator CLI and the scheduled attributable repair. */
+export const createAuthzedBackfillApply = (): TAuthzedBackfillApply => ({
+  deleteFeedbackDirectoryAssignmentResources: deleteFeedbackDirectoryAssignmentRelationships,
+  reconcileApiKeys: reconcileApiKeyRelationships,
+  reconcileFeedbackDirectories: reconcileFeedbackDirectoryRelationships,
+  reconcileMemberships: reconcileOrganizationMemberships,
+  reconcileTeamWorkspace: reconcileTeamWorkspaceRelationships,
+});
--- a/apps/web/lib/authzed/backfill-cli.ts
+++ b/apps/web/lib/authzed/backfill-cli.ts
@@ -1,22 +1,16 @@
 import "server-only";
 import { env } from "@/lib/env";
-import { reconcileApiKeyRelationships } from "./api-key";
 import {
   type TAuthzedBackfillApply,
   type TAuthzedBackfillRequest,
   type TAuthzedBackfillResult,
   runAuthzedBackfill,
 } from "./backfill";
+import { createAuthzedBackfillApply, createAuthzedBackfillNoopApply } from "./backfill-apply";
 import type { TAuthzedBackfillCliCommand } from "./backfill-cli-command";
 import { closeAuthzedClient, configureAuthzedClientForBulkWork, getAuthzedClient } from "./client";
 import { isAuthzedEnabled } from "./config";
 import { AUTHZED_ERROR_CODES, AuthzedError, type TAuthzedErrorCode, mapAuthzedError } from "./errors";
-import {
-  deleteFeedbackDirectoryAssignmentRelationships,
-  reconcileFeedbackDirectoryRelationships,
-} from "./feedback-directory";
-import { reconcileOrganizationMemberships } from "./organization-membership";
-import { reconcileTeamWorkspaceRelationships } from "./team-workspace";
 
 export { parseAuthzedBackfillCommand } from "./backfill-cli-command";
 export type { TAuthzedBackfillCliCommand } from "./backfill-cli-command";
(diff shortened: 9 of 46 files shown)


xernobyl (commented):
Reviewed at `48d9d0961` against its real base `158672ba8`. Flagging up front that **no bot has reviewed this PR**: CodeRabbit skipped it because auto-review is scoped to `^epic/` bases and this targets `bhagya/eng-2448-…`, and Sonar's gate passes without new issues. That applies to the whole stacked chain, so the usual second pair of eyes isn't there.

Two design-level findings, both latent today (legacy stays authoritative on this branch) but load-bearing the moment enforcement flips — which is what this stack exists to enable. I verified the first against a real PostgreSQL 16 by applying the migration's trigger function to a scratch database.

---

### 1. Grants are classified as revocations, and they arm a global fail-closed guard

`migration.sql` sets `isRevocation` to `TG_OP <> 'INSERT'`, so *every* UPDATE is recorded as a revocation. Driving a membership lifecycle through the real trigger:

| Event | `isRevocation` |
| --- | --- |
| invite created (INSERT) | `false` |
| **invite accepted** (pure grant) | **`true`** |
| **role `member` → `owner`** (pure grant) | **`true`** |
| role `owner` → `member` | `true` |
| membership deleted | `true` |

That matters because `hasStaleAuthzedRevocation` (`outbox-repository.ts:130-141`) is **global — no tenant or target scoping** — `spicedb-evaluator.ts:83` calls it on every check, and under enforcement `coordinator.ts` rethrows rather than falling back to legacy. Running the guard's exact SQL against an aged *invite-acceptance* row returns `true`.

So any tracked UPDATE that goes undelivered for 60 seconds fails **every enforced authorization check in the deployment** — including updates that only widened access and therefore cannot leave stale access behind. A bulk operation (mass invite acceptance, team import, reactivation sweep) is enough to build that backlog.

Worth deciding between: classify narrowing-vs-widening in the trigger by comparing old/new `role`, `permission`, `isActive` and `accepted`; or scope the guard to the affected organization so one tenant's backlog can't deny the rest.

### 2. Batch failure is all-or-nothing, so one poison event can dead-letter up to 199 healthy ones — permanently

`outbox-processor.ts:168-186`: `reconcileEventGroups` throws on the first failure, and the `catch` then calls `markAuthzedOutboxEventsFailed(leaseOwner, events, …)` for the **entire claimed batch**. Every event in it takes an attempt, so one permanently-failing event walks up to 199 healthy ones to dead-letter alongside it. Claim ordering (`isRevocation DESC, createdAt ASC`) re-selects that event first each round, so it keeps taking fresh batches down with it.

Dead-letter then has **no age bound** in the guard — I confirmed a freshly-created row with `deadLetteredAt` set still returns `true`. Combined with finding 1, that turns into a global authorization outage that persists until someone runs `outbox replay` by hand.

Per-event attribution (deliver what succeeded, fail only what didn't) would contain it.

---

### Lower severity

- **`outbox-repository.ts:96-116`** — `markAuthzedOutboxEventsFailed` awaits one `updateMany` per event, up to 200 sequential round-trips. This runs only when delivery is already failing, i.e. when the database is least likely to have headroom.
- **`outbox-processor.ts`** — `for (const event of grouped.get("user") ?? []) await reconcileUser(…)` is serial: 200 user events is 200 sequential `findUnique` calls plus projections, inside a 5-second cadence. That throughput ceiling feeds directly back into the 60-second window above.
- **`outbox-freshness.ts:6`** — `cache(hasStaleAuthzedRevocation)` only memoizes inside a React request scope; outside one it is a pass-through. Worth confirming route handlers (`api_v1/v2/v3`, `mcp`) actually get a scope, otherwise this is one extra query per authorization *check* rather than per request.
- **`migration.sql`** — none of the four indexes are partial, and processed rows are retained 7 days, so the hot-path indexes carry all of that history. A partial index (`WHERE "processedAt" IS NULL AND "deadLetteredAt" IS NULL`) would keep them small. Separately, neither pending index matches the claim's `ORDER BY "isRevocation" DESC, "createdAt" ASC`, so claims will sort.

### Verified and clean

`FOR UPDATE SKIP LOCKED` is the right claim pattern, and every state transition re-checks `leaseOwner`, so a stolen lease can't double-complete. CLI output really is aggregate-only — no identifiers, tokens or raw SDK errors, matching the PR's claim. Corrupt target types dead-letter instead of leasing forever, without logging identifiers. Pruning exists and is batched. The migration is genuinely rerunnable.

### Housekeeping

ENG-2408 is assigned to me in Linear but implemented here — worth reassigning so the ticket reflects reality, given it blocks ENG-2450/2453/2458.

BhagyaAmarasinghe:
Thanks — both design findings were real, and the second one was worse than the write-up suggested. Addressed in 3a7c428, plus cc40a095, which fixes a bug I introduced in the first one (details under 2b — worth reading, since my first attempt made one failure mode *worse* than the code you reviewed).

Taking them in order.

### 1. Grants classified as revocations

Fixed at the root: `isRevocation` is now a **deny-by-default classifier**, `authzed_projection_is_grant(target_type, previous_source, source)`, instead of `TG_OP <> 'INSERT'`. Only three transitions are grants, each proven from the projectors' *write shape* rather than from the permission closure:

| target | grant when | why |
| --- | --- | --- |
| `user` | `isActive` false→true | `reconcileUsers` deletes every relationship while inactive, so the pre-state is ∅ |
| `feedback_directory` | `isArchived` true→false **and** `organizationId` unchanged | assignment edges are `delete` while archived, `touch` once not; the parent edge is only ever touched, never re-pointed |
| `membership` | `role` unchanged | `readSnapshot` selects only `organizationId, role, userId` and carries an explicit comment that the legacy evaluator does not gate on `accepted` |

Your table now reads: invite created `false`, **invite accepted `false`**, `member → owner` `true`, `owner → member` `true`, deleted `true`.

Worth being explicit that invite acceptance is not a grant that was misfiled — it is a **no-op** that was misfiled. The projected snapshot ignores `accepted` entirely, so an acceptance writes byte-identical relationships. That is the cheapest and most provable case, and it is also the bulk one.

**On `member → owner`, I deliberately kept it a revocation.** The `OrganizationRole` lattice is rankable — `{billing:1, member:1, manager:2, owner:3}` with "grant iff rank(new) > rank(old)" is correct on all twelve ordered pairs, because the one incomparable pair shares a rank and fails the strict inequality. I still didn't encode it, for three reasons:

1. It is a second copy of `authzed/schema.zed` in a place nothing keeps in step. `relationship-map.ts` is `satisfies Record<PrismaEnum, string>`, so a new enum value is a compile error; a rank table in a migration has no such backstop, and it would go wrong in the **fail-open** direction — the one direction this guard must never be wrong in.
2. The rank rule breaks as soon as a target watches more than one non-key column. `feedback_directory` and `api_key` both watch `organizationId` alongside their state column, so each ladder needs an ad-hoc "and these columns did not move" clause anyway.
3. Blast radius doesn't justify it. The damage here is a *bulk* operation arming a global guard — mass acceptance, reactivation sweep, unarchive sweep. Role promotion is a one-at-a-time admin action, and if a single promotion goes undelivered for 60s the projector is already broken.

There's a note in `authzed/README.md` on when to revisit (a bulk re-roling path).

### On tenant scoping — not done, and why

You offered it as the alternative to (1), and I want to be straight that I chose only (1).

Full scoping isn't achievable: `user` revocations (deactivation) are genuinely cross-tenant, and `team_membership`, `workspace_team`, `api_key_workspace` and `feedback_directory_assignment` carry no `organizationId` the trigger can reach without a parent lookup on **every source mutation**. Partial scoping is possible for the six types that do carry one, but it needs a column, an index, and a plumb-through from `TResolvedAuthorizationScope` into `assertAuthzedProjectionFreshness`.

My read is that fixing the classification removes the amplifier that made the global guard a practical outage, and (2) below removes the permanence — which together are the reported harm. Happy to be argued out of this; if you think a genuine single-tenant revocation backlog denying everyone else is still a live risk before enforcement flips, say so and I'll do the partial scoping in this PR rather than leaving it.

### 2. Batch failure is all-or-nothing

Fixed in three layers, because per-event splitting alone would have been the wrong trade — the reconcilers do a read → write → re-read verify loop, so naive per-event delivery turns 2 Postgres queries + 1 gRPC write into ~400 + 200 on the happy path.

- **Per-group attribution.** A claimed batch now delivers as six independent groups (the six reconciler calls that were already there). A failing group takes only its own events. Happy-path cost is unchanged.
- **Retryable vs not.** A retryable failure releases every remaining group *untried* — same reasoning `backfill.ts:461` already spells out about not spending three-attempt budgets against an instance known to be unreachable. A non-retryable one is local, so the other groups still run.
- **Quarantine.** When the code could plausibly belong to one event — `authzed_projection_invalid_source` (your cross-tenant `FeedbackDirectoryWorkspace` case), `authzed_invalid_request` — the group is halved until the culprit is alone, ~2·log₂(n) extra calls. Codes that describe the *instance* (`authzed_unauthenticated`, `authzed_internal`) are **not** split; there's a test asserting exactly one reconciler call for those, specifically to stop a 2×N regression.

I also flipped one classification at the outbox boundary only: `AuthzedProjectionUnstableError` carries `retryable: false` but means "the source row moved under me across 3 passes" — transient, and likeliest on the hot rows the outbox carries. Overridden in the processor rather than on the class, since four projector test files pin that contract.

### 2b. Dead letter → permanent global denial

This was the sharpest part of the finding and it needed its own fix. I did **not** put an age bound on dead letters — an old dead letter is more dangerous than a fresh one, so ageing it out is a silent fail-open on the exact event class the guard exists for. Instead:

- **New `permanentFailures` column gates dead-lettering**, separate from `attempts` (which still drives backoff; decrementing it would have frozen the backoff into a 5-second hot loop).

**This is where I got it wrong first time, and it's the most useful thing in this reply.** In 3a7c428 the gate was "non-retryable AND the attempt covered exactly one event". I called that flag `isolated` and treated it as meaning "attributable to this event" — but those are two different claims, and only the first was checked.

On a five-second cadence most groups hold exactly one event, so "covered one event" is nearly always true and says nothing about fault. Meanwhile `getErrorDescriptor` maps `UNAUTHENTICATED` and `PERMISSION_DENIED` to `retryable: false`, and its `default` branch maps *any* unmapped error — gRPC `INTERNAL`, or a plain throw inside a reconciler — to `authzed_internal`, also non-retryable. None are in `PER_EVENT_ERROR_CODES`, so no split happens and the group was charged as-is.

Net effect: rotate the SpiceDB preshared key wrongly, deactivate a user during the outage, and that revocation dead-letters in **~8.5 min** — *sooner* than the ~58 min the code you reviewed would have taken, and then keeps the guard armed after recovery. I had also written the opposite invariant into the README and the RUNBOOK, so the runbook would have misdirected an operator during exactly that incident.

The module already knew the answer: it refuses to *split* on those codes precisely because they "describe the instance, not an event". That set gated the splitting decision and not the attribution decision. cc40a095 makes it gate both, and renames the flag `attributable` so the two meanings can't drift apart again. Dead-lettering now needs all three: non-retryable, single-event, **and** a code an event can actually cause.

- **Consequence, now actually true: no SpiceDB outage — of any duration or kind — can dead-letter anything.** The ~64-minute path you traced to 20 attempts no longer exists, and neither does the ~8.5-minute one I briefly introduced.
- **Dead letters self-heal.** The six-hour audit now replays them whenever it comes back `reconciled` — a clean full audit means Postgres and SpiceDB already agree, so whatever the dead letter was saying has been said by other means. That bounds a global denial at six hours instead of at operator attention. A still-poisoned event just dead-letters again, so it's a six-hourly retry, not a loop.

### Lower severity

- **`markAuthzedOutboxEventsFailed`** — now one statement instead of up to 200, computing the backoff in SQL from each row's own `attempts` rather than grouping by attempt count in TS. Two things that bit while doing it and are worth a reviewer's eye: raw SQL bypasses Prisma's `@updatedAt` (set explicitly), and the exponent needs `LEAST("attempts", 20)` since `attempts` now grows unbounded during an outage.
- **Serial `reconcileUser`** — replaced with `reconcileUsers`: one `findMany`, bounded-concurrency deletes for inactive users, and one chunked call each to the membership and team-workspace reconcilers. Missing rows still count as inactive (`findMany` silently omits what `findUnique` returned as `null`). This is also why `runChunked` moved out of `backfill.ts` into a shared `projection-chunks.ts` — a bounded batch of user events expands into an unbounded number of membership targets.
- **`cache(hasStaleAuthzedRevocation)`** — you asked me to confirm whether route handlers get a scope. **They don't**, and it's worse than "one extra query per check". `app-route.runtime.prod.js` carries zero `getCacheForType`, and React's `cache` falls straight through to `fn.apply` with no dispatcher — so it was a pass-through in **9 of the 11 rollout targets**, plus in the shadow comparisons that drain inside `after()`. Combined with `canUserNavigateWorkspace` (3 checks), `checkAuthorizationUpdated` (1+N across 118 call sites) and `getAuthorizedWorkspaceFeedbackDirectories` (N concurrent), it was an uncached Postgres round trip per authorization *check*. Replaced with a 1s process-level memo with in-flight sharing — which also has the advantage of being testable without a render harness, unlike the old code whose test `vi.mock`'d React away entirely. The repo had already documented this limitation three times (`context.ts:36`, `resolvers.ts:14`, `PERFORMANCE.md:115`); this is the first place it was load-bearing.
- **Indexes** — all four replaced with three partials, one of which (`_undelivered_idx`) is wider than I first made it. My initial version indexed dead letters only, and I wrote a comment claiming the scoped status aggregate was therefore off the retained history. It wasn't: a bare `processedAt IS NULL` implies neither complementary partial predicate, so the aggregate still seq-scanned everything, twelve times a minute. Widened to the whole undelivered set: measured against 200k delivered + 50 live rows it's now an Index Only Scan touching 5 buffers (0.04 ms) instead of a parallel seq scan over 200,050. `leaseExpiresAt_idx` and `target_idx` were dead (no query leads with either). The claim index is now keyed exactly on its `ORDER BY`, and `EXPLAIN` against a 2,000-row backlog shows an ordered index scan with **no `Sort` node**. The guard's `OR` is split into two `EXISTS` so each branch hits its own partial index. Prisma can't express `WHERE` on `@@index`, so the model deliberately declares none and points at the migration — convention written up in `packages/database/README.md`.
- **One you didn't flag**, found while tracing: `getAuthzedOutboxStatus` ran a **full-table aggregate every 5 seconds** (`drainAuthzedOutbox` calls it unconditionally, delivery runs at 5s), which with 7-day retention is a seq scan of the whole history table twelve times a minute. Now scoped `WHERE "processedAt" IS NULL`.

### How I found the above, and what else it caught

Rather than take my own fix on trust I ran an adversarial pass over 3a7c428 — six independent lenses (SQL classifier, attribution logic, regressions vs the parent, raw SQL, the memo, and test effectiveness), then a refutation pass over each finding. Six candidates, four refuted, two confirmed. Both confirmed ones are fixed in cc40a095, along with three of the refuted ones that were judged low-impact but still made a statement I'd written untrue:

1. The attribution bug above.
2. **A test of mine that couldn't fail.** The "unarchive that also moves organization is still a revocation" case cleared the outbox once, before its *setup* mutation, so it asserted `.at(0)` of a two-row result sorted revocations-first — and passed with the `organizationId` clause deleted from the classifier. That clause had no other coverage anywhere.
3. **Nothing pinned `processedAt IS NULL` in the guard.** Every guard test used undelivered rows, so a guard ignoring the predicate entirely would have passed — while in production a healthy deployment holds seven days of delivered revocations, every one of them older than 60 seconds.
4. The index comment above.
5. The memo TTL read the wall clock, so a backwards NTP step would freeze the guard's answer for the length of the step — the one way its "bounded and explicit" cost could be exceeded without limit. Monotonic clock now.

Every fix is mutation-tested: I reverted each one and confirmed the named test goes red. I'd rather hand you that list than have you find it.

### Testing

The migration test was string-matching SQL, which can tell you the SQL *says* something but not that it *does* it — and a plpgsql classifier is only observable against a real database. Added `outbox-trigger.integration.test.ts` on the existing harness: it drives the **real triggers** (not the SQL function directly) through the transition table above, plus the backoff/dead-letter arithmetic and the index shapes. `AuthzedProjectionOutbox` had to go into `integration/reset-db.ts` — it has no FKs at all, so nothing cascaded to it and rows bled across files, which is exactly the case that file's own `INVARIANT` comment asks to be added.

Migration edited in place rather than stacked, since it's unmerged and exists only on this stack; applied twice against a live PG to confirm the rerun is a clean no-op.

Full state: `apps/web` unit 679 files / 8,974 tests green, outbox integration 17 green, `packages/database` and `packages/jobs` green, typecheck + ESLint + `prettier --check .` clean. Four integration files (`modules/auth/*`, `modules/ee/workflows/*`) fail locally on missing SMTP and mail-URL config — confirmed pre-existing by running them on the unmodified parent commit.

### Housekeeping

Reassigned ENG-2408 to me.

### Still open, deliberately

- The guard remains global (above).
- **Attribution is now conservative in the other direction.** An event with a genuinely poisonous payload under a code outside `PER_EVENT_ERROR_CODES` retries forever at the five-minute ceiling instead of dead-lettering. I think that's the right side to err on — it stays visible, since the guard stays armed if it's a revocation, rather than being silently parked — but it does mean the dead-letter queue isn't a catch-all for every stuck event. If you'd rather it were, the fix is widening that set, not loosening the gate.
- **Parent moves are unprojectable, not merely unclassified** — `createParentUpdate`/`parentUpdate` only ever *touch* the new parent, so moving a Team/Workspace/ApiKey/FeedbackDirectory between organizations leaves the old organization's owners with access. The classifier calls these revocations, but the event then delivers "successfully" and disarms the guard, so the guard doesn't actually protect against this class. `backfill.ts` already reports it as `mismatchedParents` and deliberately doesn't repair it. Pre-existing and out of scope here, but it's in `Open gaps` and I think it wants its own ticket.

pandeymangg (approved):

PR #8919, opened 2026-08-19
fix(a11y): give the public survey a semantic heading structure [ENG-2336]

itsjavi:
Fixes ENG-2336

## What & why

The public survey UI rendered **no heading on a question card**. The VPAT records this as *Partially Supports* against WCAG 2.0 SC 2.4.6 (Level AA) — "All fields have descriptive labels. Headings are not present."

Why the existing code looked like it already handled this but didn't:

- `packages/surveys`' own `Headline` did have an `h2` branch, but 14 of the 15 element types render through `@formbricks/survey-ui`'s `ElementHeader` → `Label`, which emits a `<label>` or a `<span>`. Only `cal-element.tsx` still used `Headline`, so the `h2` branch was effectively dead.
- Welcome and ending cards each carried an `h1`, but they are *separate cards* — never on screen at the same time as a question. A respondent answering question 3 had nothing to navigate by at all.

After this PR the survey exposes one `h1` (the survey name, visually hidden, rendered once) with every card headline — welcome, each element prompt in a block, ending — as an `h2` beneath it. Multi-element blocks yield sibling `h2`s; no level is skipped.

```
h1  Q3 Customer Satisfaction        ← survey name, sr-only, rendered once
h2  Welcome to our feedback survey
h2  What feedback do you have for us?
h2  Which plan are you on?
h2  Thank you!
```

The heading **wraps** the existing `Label` rather than retagging it, because three things hang off that element and all had to survive:

| Depends on | Call sites | How it survives |
| --- | --- | --- |
| `htmlFor` — the prompt *is* the input's `<label>` | `open-text`, `cta`, `ranking`, `consent`, `date`, `file-upload` | `<label>` nested in a heading is valid HTML; the binding is untouched |
| `headlineId` — names grouped questions via `aria-labelledby` | `matrix`, `picture-select`, `date`, `rating`, `single-select`, `multi-select`, `nps` | the id stays on the `Label`, so the computed accessible name is unchanged |
| `label-headline` class — runtime theming targets it with `!important` (`styles.ts` `addCustomThemeToDom`) | all | class stays on the `Label`, not moved to the tag |

The heading inherits the `flex` class the wrapper `<div>` already carried, so layout is byte-identical.

<details>
<summary>Rendered DOM, before → after</summary>

```html
<!-- before: ElementHeader with htmlFor -->
<div class="flex">
  <label for="my-input" class="… label-headline select-text">What feedback do you have?</label>
</div>

<!-- after -->
<h2 data-slot="element-headline" class="flex">
  <label for="my-input" class="… label-headline select-text">What feedback do you have?</label>
</h2>

<!-- before: ElementHeader with headlineId (grouped question) -->
<div class="flex">
  <span id="q1-headline" class="… label-headline select-text">Which plan are you on?</span>
</div>

<!-- after -->
<h2 data-slot="element-headline" class="flex">
  <span id="q1-headline" class="… label-headline select-text">Which plan are you on?</span>
</h2>
```
</details>

On the **modal** path the heading is rendered inside the `role="dialog"` element rather than on the `#fbjs` root: `aria-modal="true"` makes assistive tech ignore everything outside the dialog, so a root-level heading would have been unreachable for app surveys with an overlay.

Also in this diff:

- **`survey-ui/src/styles/globals.css`** gains the `h1…h6` reset it lacked. `packages/surveys/preflight.css` already neutralises heading typography inside `#fbjs`, but that preflight ships with the surveys package and this stylesheet is consumed standalone too (Storybook / any React consumer), where the prompt would otherwise pick up UA heading size and margins.
- The ending card's "sending responses" status text was a **second `h1`**, outside `Headline` and without `label-headline`. It is a transient status message, so it is now a `<p>`.
- `Headline` takes an explicit `headingLevel` instead of inferring the level by comparing `elementId` against the string literals `"welcomeCard"` / `"EndingCard"`. `elementId` had no other use and is gone.
- The editor's 13 e2e question-heading locators are scoped to the editor `<main>`. The editor renders a **live preview of the survey into the same document** (`#survey-preview`, in an `<aside>` visible from the `md` breakpoint up), so an unscoped `getByRole("heading", { name })` now matches the preview's `h2` as well and would fail Playwright's strict mode.
- Both new `headingLevel` props are optional and default to `2`, so no existing call site had to change to keep its current output. `Headline` is internal to `packages/surveys` (not exported from the package entry) and its three call sites are all updated; `ElementHeader`'s is public API on `@formbricks/survey-ui`.

### Screenshots

This is a semantics-only change, so **the correct screenshot result is no difference** — and that is what these are evidence of. Both rendered from the real built survey bundle (`surveys.umd.cjs`), before and after, at 2× device scale.

**Question card** — the card that previously had no heading at all

| Before | After |
| --- | --- |
|  |  |
| headings in `#fbjs`: **none** | `h1` Q3 Customer Satisfaction (sr-only) · `h2` What feedback do you have for us? |

**Welcome card** — headline moves `h1` → `h2` under the survey name

| Before | After |
| --- | --- |
|  |  |
| `h1` Welcome to our feedback survey | `h1` Q3 Customer Satisfaction (sr-only) · `h2` Welcome to our feedback survey |

**Ending card** — headline moves `h1` → `h2`, and the "sending responses" status text stops being a second `h1`

| Before | After |
| --- | --- |
|  |  |
| `h1` Thank you! | `h1` Q3 Customer Satisfaction (sr-only) · `h2` Thank you! |

The `h1` → `<p>` demotion on the ending card's transient "sending responses" state is not screenshotted, because it is a flash between submit and the thank-you card — capturing it reliably is not possible here. What it actually depends on is measured instead: an `h1` and a `p` carrying the same class, injected into `#fbjs` and read back through `getComputedStyle`, resolve identically on **both** sides — `font-size: 16px`, `font-weight: 400`, `margin: 0`, `color: rgb(30, 64, 175)`, `line-height: normal` — because `preflight.css` already neutralises heading typography inside `#fbjs`. So the demotion removes a heading from the a11y tree and changes nothing on screen.

The two images in each row are **byte-identical**, not merely similar:

```
f38705cf6e7929b7a6f8af4f62121b61  before-question-card.jpg
f38705cf6e7929b7a6f8af4f62121b61  after-question-card.jpg
1e005f3fe8bb9d8507057baca2895351  before-welcome-card.jpg
1e005f3fe8bb9d8507057baca2895351  after-welcome-card.jpg
53f2ede5da1fd866c52e2df2ddfe5f9e  before-ending-card.jpg
53f2ede5da1fd866c52e2df2ddfe5f9e  after-ending-card.jpg
```

## Breaking changes

None — both new `headingLevel` props are optional and default to `2`, so every existing consumer renders exactly as before. Nothing is removed, renamed or re-typed, and `packages/surveys/src/index.ts` and the `js-core` exports are untouched.

## Migrations & env

- none

## How this was tested

**Coverage**

Per AGENTS.md, `.tsx` components are covered by Playwright rather than component unit tests, so the heading semantics are asserted in the existing axe suite. Note that axe **cannot** gate this: `page-has-heading-one` and `heading-order` are best-practice rules, which this suite runs as warn-only (`WARN_TAGS`) — only `wcag2a/2aa/21a/21aa/22a/22aa` fail the build. Hence the explicit assertions.

| Behaviour | How | Outcome |
| --- | --- | --- |
| Survey exposes exactly one `h1`, carrying the survey name | e2e | New `heading structure: one h1 names the survey, card headlines are h2` in `survey-accessibility.spec.ts`. Also re-asserted *after* advancing to a question card — the stacked layout renders up to three peeking cards, so a heading placed per-card would show up here as a count > 1 |
| No skipped heading level | e2e | Same test: `widget.locator("h3, h4, h5, h6")` is empty, so `h1 → h2` is the whole tree |
| Welcome and ending headlines are `h2`, not styled divs | e2e | Same test for welcome; `endingCardLocator` in the same spec now resolves the ending card by `level: 2` and is exercised by all nine existing walk variants |
| `htmlFor` label association survives the heading wrapper | e2e | Same test: the `h2` contains exactly one `label[for]`, **and** the input still resolves via `getByLabel(OPEN_TEXT_HEADLINE)` — both halves, so retagging the label in future fails here rather than silently stripping the input's accessible name |
| `headlineId` still names a grouped question | e2e | Same test: `getByRole("radiogroup", { name: SINGLE_SELECT_HEADLINE })` on the single-select card, which is named purely through `aria-labelledby` → the id nested inside the new `h2` |
| No new WCAG AA violations on any card | e2e | The nine pre-existing variants in `survey-accessibility.spec.ts` (desktop / mobile / tablet / forced-colors / reduced-motion / dark / empty-submit / back-nav / RTL Arabic) walk every card type with an empty allowlist |
| Real rendered heading tree, from the built bundle | manual | Rendered the built `surveys.umd.cjs` and read the headings out of `#fbjs` per card: question card went from `[]` to `["H1: Q3 Customer Satisfaction", "H2: What feedback do you have for us?"]`; the ending card from `["H1: Thank you!"]` to `["H1: Q3 Customer Satisfaction", "H2: Thank you!"]`; same `h1` on every card, never duplicated |
| Rendering is unchanged | manual | Byte-identical screenshots above (md5 match) for the question, welcome **and ending** cards. Separately pixel-diffed the PNG originals: 0 of 924 000 pixels differ on each of the welcome, question, ending and switcher-collapsed frames. The ending card's `h1` → `<p>` status text is covered by the computed-style probe described with the screenshots, since that state is too transient to capture |
| Rendered DOM keeps every association | manual | Rendered `ElementHeader` (both `htmlFor` and `headlineId` paths) and `SurveyContainer` (inline + modal) under happy-dom and inspected the HTML: wrapper went from `<div class="flex">` to `<h2 class="flex">` with identical classes and children; `label-headline`, `for` and `id` all present; modal `h1` resolves inside `[role="dialog"]` |
| Monorepo still builds and passes | manual | `pnpm typecheck` (27 tasks), `pnpm lint`, `pnpm test` (615 files / 8203 tests), `pnpm format:check` — all clean |

**Open gaps**

- The `sr-only` `h1` is not verified against a real screen reader. The ticket asks for an NVDA `H`-key and VoiceOver-rotor pass, and for an IBM Equal Access + ANDI re-run; neither is available in CI. The axe suite plus the DOM assertions above are what can be automated — a manual AT pass is still worth doing before the VPAT row is flipped to *Supports*.
- The `h1` carries `survey.name`, which is an internally-authored field. It is already the basis of the link survey's `<title>` (`metadata-utils.ts` falls back to it), so this exposes nothing new, but a workspace whose survey names are cryptic gets a cryptic heading. No product decision was made here.
- The visually-hidden `h1` is a11y-only by design — the card layouts have no room for a visible title. Flagged in the ticket as wanting sign-off from Kristian / pain.design.

**Risks**

- **Playwright strict mode in the editor.** The live preview now emits `h2` prompts into the editor's document. The 13 known locators in `playwright/utils/helper.ts` are scoped, but any *new* unscoped `getByRole("heading")` written against the editor will be ambiguous whenever the previewed card is the one being edited. The new `editorElementHeading` helper carries that warning at its definition.
- **Third-party `dist/` consumers of `@formbricks/survey-ui`.** A consumer that renders `ElementHeader` outside `#fbjs` gets an unreset heading, since both the new rule and the surveys preflight are `#fbjs`-scoped. That was already true of every `label-*` class in this package (they resolve `--fb-*` vars only defined under `#fbjs`), so it is not a new constraint.
- If a card headline ever needs to sit under another heading, pass `headingLevel={3}`. Nothing uses it yet, so that path is typed and implemented but exercised by no test — the e2e assertions only cover the `h1`/`h2` tree that ships.

> [!NOTE]

### DIFF
--- a/apps/web/playwright/survey-accessibility.spec.ts
+++ b/apps/web/playwright/survey-accessibility.spec.ts
@@ -3,8 +3,12 @@ import { type Locator, type Page, expect } from "@playwright/test";
 import { logger } from "@formbricks/logger";
 import { test } from "./lib/fixtures";
 import {
+  A11Y_SURVEY_NAME,
   ENDING_CARD_HEADLINE,
+  OPEN_TEXT_HEADLINE,
+  SINGLE_SELECT_HEADLINE,
   type SeededAccessibilitySurveys,
+  WELCOME_CARD_HEADLINE,
   seedAccessibilitySurveys,
 } from "./utils/accessibility";
 import { mockStorageUploads } from "./utils/helper";
@@ -241,13 +245,13 @@ const getActiveCardId = async (page: Page): Promise<string | null> => {
 
 const advanceButton = (card: Locator): Locator => card.locator(ADVANCE_BUTTON_SELECTOR);
 
-// The ending card renders its (fixture-controlled) headline as the page's <h1>; that
-// is a stable positive signal that the survey completed (unlike "no nav button",
-// which is also briefly true during the 600ms card transition animation). The welcome
-// card also renders an <h1>, so the ending is matched by its exact headline text —
-// endings are not language-patched, so this holds in the RTL variant too.
+// The ending card renders its (fixture-controlled) headline as an <h2>; that is a stable
+// positive signal that the survey completed (unlike "no nav button", which is also briefly
+// true during the 600ms card transition animation). Every card headline is an h2 — the survey
+// name is the page's only h1 (ENG-2336) — so the ending is matched by its exact headline text.
+// Endings are not language-patched, so this holds in the RTL variant too.
 const endingCardLocator = (page: Page): Locator =>
-  page.getByRole("heading", { level: 1, name: ENDING_CARD_HEADLINE });
+  page.getByRole("heading", { level: 2, name: ENDING_CARD_HEADLINE });
 
 const isOnEndingCard = (page: Page): Promise<boolean> =>
   endingCardLocator(page)
@@ -281,6 +285,32 @@ const waitForSettled = async (page: Page, selector: string): Promise<void> => {
 
 const waitForCardSettled = (page: Page, cardId: string): Promise<void> =>
   waitForSettled(page, `[id="${cardId}"]`);
+
+/**
+ * Locator-based twin of `waitForSettled`, for elements that cannot be named by a stable CSS
+ * selector. The ending card is matched by role + heading text because peeking cards carry
+ * identical markup, so `document.querySelector` would settle on the wrong one.
+ */
+const waitForLocatorSettled = async (locator: Locator): Promise<void> => {
+  await expect
+    .poll(
+      () =>
+        locator
+          .evaluate((node: HTMLElement) => {
+            let el: HTMLElement | null = node;
+            while (el) {
+              if (getComputedStyle(el).opacity !== "1") return false;
+              el = el.parentElement;
+            }
+            return true;
+          })
+          .catch(() => false),
+      { timeout: 5000, intervals: [100] }
+    )
+    .toBe(true)
+    // Mirrors waitForSettled: a timeout is not an error here, the caller's assertions decide.
+    .catch(() => undefined);
+};
 
 /**
  * After clicking advance, wait for a STABLE next state: either a different active card
@@ -425,7 +455,7 @@ const walkAndScan = async (
     endingCardLocator(page).first(),
     `ending card should be visible for variant "${variant}"`
   ).toBeVisible({ timeout: CARD_TIMEOUT });
-  await waitForSettled(page, "#fbjs h1");
+  await waitForLocatorSettled(endingCardLocator(page).first());
   await scan(page, variant, "ending-card", failSink);
 };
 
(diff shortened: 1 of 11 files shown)


mattinannt (approved):
@itsjavi Thanks for the PR :-) Is there a good way to also test for a11y of the surveys package in CI? linting, playwright,..
Let's discuss in Slack so we can make sure that these things don't break again in a different PR since they need to be tested manually at the moment from what I understand.

PR #8920, opened 2026-08-19
fix(a11y): identify the active language in the survey language switcher [ENG-2335]

itsjavi:
Fixes ENG-2335

## What & why

The language switcher was an unlabelled globe icon that was **never told which language was active** — `setSelectedLanguageCode` is write-only, so the component had no way to know. Nothing in the dropdown marked the current option either. The VPAT records this as *Partially Supports* against WCAG 2.0 SC 3.1.1 (Level A) — "Current language is not identified in the language switch dropdown."

That is both an a11y failure and a plain UX bug for the audience that needs it most: Electrolux alone runs surveys in ~33 languages, and a respondent could not tell what they were reading, let alone which entry would keep them where they are.

Three changes:

1. **The trigger names the active language** — endonym beside the icon, plus an accessible name that states it. The visible label drops the region ("Deutsch", not "Deutsch (Deutschland)") because the `h-8` chrome row would ellipsise the full name, and the trigger only ever shows *one* language so there is nothing to disambiguate against. The accessible name keeps the full name and still contains the visible text, so WCAG 2.5.3 Label in Name holds.
2. **The active option is marked** — `aria-current="true"` plus a bold weight. `aria-current` rather than `aria-selected`, because these are `<button>`s in a plain popup, not options in a listbox, where `aria-selected` is ignored.
3. **`#fbjs` declares the survey's language.** This is the part that only matters for **embedded and app surveys**. A link survey's host page already syncs `<html lang>` through `onLanguageChange`, but the JS widget is dropped into someone else's document and must never mutate it, so an embedded survey previously inherited the host page's language no matter what language the survey was in.

`RenderSurvey` wraps the `onLanguageChange` callback `Survey` already fires, rather than adding a second channel out of it — the active language keeps one source of truth. Only the code is held in state; both the `lang` tag and `dir` resolve from it, through the same resolver, so they cannot drift apart.

`getSurveyLanguageTag` resolves the renderer's `"default"` sentinel to a real code and validates it against the survey's **enabled** languages, so neither the sentinel nor a language the survey no longer has can reach a `lang` attribute. `isRTLLanguage` had the same resolution inlined and now shares it.

Each option also gets its own `lang`. The labels were already endonyms written in their own script, so without it a screen reader pronounces 日本語 with the page language's rules.

### Merged with `main`, and one refactor

`main` now carries ENG-2336 (#8919), which touched the same two files this branch does — both add a
prop to `SurveyContainerProps` and edit the `#fbjs` roots. The merge is additive and in its own
commit (`141d8ed`): `SurveyContainer` takes **both** `surveyName` (the sr-only `h1`) and `lang`, and
the inline root renders the heading above `children`. This branch's direction effect was kept over
`main`'s, because dropping its `exhaustive-deps` suppression is the point of `47ac5f2`.

A second commit (`03c1f95`) answers SonarQube's `typescript:S3776` on `LanguageSwitch` — cognitive
complexity 16 against a limit of 15. The three things pushing it over are pure decisions about a
language list with no business being inline in a component, so they moved to
`packages/surveys/src/lib/language-options.ts`: `canonicalizeLanguageCode` (which was inlined in
**four** places across two files), `isSameLanguageCode`, and the dedupe-by-canonical loop as
`getVisibleSurveyLanguages`. `resolveSelectedLanguageCode` had its own private copy of the canonical
helper and `getSurveyLanguageTag` inlined the expression twice more; all now share one definition,
and that `find` callback collapses to a single expression.

Sixteen unit tests came with it, which is the more useful half: the switcher's dedupe and
active-marking logic was previously verified only by a manual bundle run.

### Why the visible label is not `sr-only`

The `aria-label` on the trigger already carries the language, so an `sr-only` span would announce it a second time and leave sighted users with the same bare globe they have today. The visible span is the *other* half of the fix — it is `aria-hidden="true"` precisely so it does not duplicate the accessible name. The ticket asks for the switcher to communicate the current language "visibly **and** programmatically", and the finding is framed as a UX bug as much as an a11y one, which is why the label is on screen.

If we would rather ship zero visual change here, the alternative is dropping the visible span and keeping only the `aria-label` plus the bold `aria-current` option — that satisfies the VPAT remark literally ("current language is not identified in the language switch **dropdown**") at the cost of the "visibly" criterion. One line either way; say the word.

### Screenshots

Rendered from the real built survey bundle (`surveys.umd.cjs`) with a three-language survey — English (default), German, Japanese.

**Switcher trigger, collapsed**

| Before | After |
| --- | --- |
|  |  |
| Bare globe. Accessible name: `"Language switch"`. `#fbjs` has no `lang`. | Globe + `English`. Accessible name: `"Language: American English"`. `#fbjs` is `lang="en-US"`. |

**Dropdown open**

| Before | After |
| --- | --- |
|  |  |
| No option marked — nothing says which one you are on. | `American English` is bold and carries `aria-current="true"`. |

**After switching to German** — the trigger, the marker and the root `lang` all follow:

Accessible name becomes `"Sprache: Deutsch (Deutschland)"` (the survey chrome is localized too), visible label `Deutsch`, `#fbjs` becomes `lang="de-DE"`, and the marker moves to `Deutsch (Deutschland)`.

## Breaking changes

**One emitted value changes.** The box is ticked because the template's definition covers it ("changes an API/SDK shape or emitted value"), so it belongs in the release notes rather than buried — even though nothing a consumer *passes in* changes and no action is required of anyone.

| `response.language` after a respondent explicitly picks the default language, on a survey whose **default** language row stores a legacy code (`hi`) *and* which also has the canonical row (`hi-IN`) | `hi-IN` | `hi` | analytics/exports that group responses by language on such a survey | none |

Why it moves: the option list is deduped by canonical code and keeps the canonical row, so the visible default option reads `hi-IN`. `changeLanguage` compared that against the default row's **raw** code (`hi`), missed the match, and stored `hi-IN` instead of the `"default"` sentinel — which `survey.tsx:926` then resolves to the default language's code. So picking the default language explicitly recorded a *different* value than not touching the switcher at all. Canonicalizing both sides makes the two paths agree, and `hi` is the code the survey itself stores for its default language. Surveys whose default row already uses a canonical code are unaffected.

No other breaking surface: `packages/surveys/src/index.ts` and the `js-core` exports are untouched, no route/env/migration changes, and `LanguageSwitch`'s new required `selectedLanguageCode` prop is internal (one call site, `survey.tsx`, not exported from the package entry).

## Migrations & env

- none

One new i18n string, `common.language_switch_current` (`"Language: {language}"`), added to `packages/surveys/locales/en-US.json` and fanned out to all 22 target locales with `pnpm i18n`, which also rebuilt `packages/surveys/i18n.lock`. The non-English files were not hand-edited.

## How this was tested

**Coverage**

Unit tests on the pure helpers plus manual verification of the rendered bundle — no Playwright suite, since the DOM contract here is small enough not to warrant one.

Rerun command — every mutation below was re-run against head `03c1f95`, and **line numbers are against
that head** (they moved when the helpers were extracted):

```
cd packages/surveys && npx vitest run --project surveys-unit \
  src/lib/utils.test.ts src/lib/language-display-name.test.ts src/lib/language-options.test.ts
```

Baseline: **118 passed** across those three files (`@formbricks/surveys` as a whole: 704).

| Behaviour | How | Outcome |
| --- | --- | --- |
| **`response.language` keeps the `"default"` sentinel for a legacy-alias default** — the emitted-value change above | `unit (mutation)`<br>`packages/surveys/src/lib/utils.test.ts` → `describe("resolveSelectedLanguageCode")`<br>`packages/surveys/src/lib/language-options.test.ts` → `describe("isSameLanguageCode")` | `resolveSelectedLanguageCode`, extracted out of the component precisely so this is testable. Mutate `packages/surveys/src/lib/language-options.ts:15` — compare raw instead of canonical (`return code === other;`) → **5 failed / 113 passed**, both alias directions plus the tag lookup that now shares the helper. Mutate `packages/surveys/src/lib/utils.ts:270` — drop the comparison, `return languageCode;` → **3 failed / 115 passed**. Six cases: default pick, non-default pick, canonical pick vs legacy default, legacy pick vs canonical default, two genuinely different languages not collapsed, and no-default-configured pass-through |
| `getSurveyLanguageTag` refuses a language the survey does not have, or has disabled | `unit (mutation)`<br>`packages/surveys/src/lib/utils.test.ts` → `describe("getSurveyLanguageTag")` | Mutate `packages/surveys/src/lib/utils.ts:251` — `if (configured) return configured.language.code;` → `return languageCode;` → **4 failed / 114 passed**: unconfigured code, configured-but-disabled, legacy alias → stored canonical code, no-match-no-default → `null` |
| `getShortLanguageDisplayName` drops the region | `unit (mutation)`<br>`packages/surveys/src/lib/language-display-name.test.ts` → `describe("getShortLanguageDisplayName")` | Mutate `packages/surveys/src/lib/language-display-name.ts:273` — guard → `if (false) return short;` (never shorten) → **1 failed / 117 passed** |
| `getShortLanguageDisplayName` keeps a *known* full name rather than degrading to an echoed code | `unit (mutation)`<br>`packages/surveys/src/lib/language-display-name.test.ts` → `describe("getShortLanguageDisplayName")` | Mutate the same line, `packages/surveys/src/lib/language-display-name.ts:273` → `return short;` (shorten unconditionally) → **1 failed / 117 passed**, the `xx-YY` case |
| `getSurveyLanguageTag` resolves the `"default"` sentinel and empty input to the default language | `unit (guard)`<br>`packages/surveys/src/lib/utils.test.ts` → `describe("getSurveyLanguageTag")` | Asserted, but **honestly a guard, not a mutation row**: mutating the `!== "default"` check at `packages/surveys/src/lib/utils.ts:246` leaves the suite green (`118 passed`). Once the configured-language lookup was added that check became redundant for correctness — `"default"` matches no configured language, so it falls through to the same fallback. Kept for clarity and to skip a pointless lookup on the common path, not because a test detects its removal |
| `getShortLanguageDisplayName` keeps the script, so 简体中文 and 繁體中文 stay distinct | `unit (guard)`<br>`packages/surveys/src/lib/language-display-name.test.ts` → `describe("getShortLanguageDisplayName")` | Asserted for `zh-Hans-CN` / `zh-Hant-TW`, also a guard: dropping the script from that expression is absorbed by the unknown-code guard (bare `zh` has no `NATIVE_NAMES` entry, so it falls through to the full name) and the suite stays green |
| The dropdown offers each language once, keeping the canonical code when a legacy alias is also configured | `unit (mutation)`<br>`packages/surveys/src/lib/language-options.test.ts` → `describe("getVisibleSurveyLanguages")` | Mutate `packages/surveys/src/lib/language-options.ts:34` — drop the canonical preference (`if (!existing)`) → **1 failed / 117 passed**, the case where the alias is listed second. Eight cases, including a disabled canonical row not shadowing its enabled alias |
| A disabled language is never offered | `unit (mutation)`<br>`packages/surveys/src/lib/language-options.test.ts` → `describe("getVisibleSurveyLanguages")` | Mutate `packages/surveys/src/lib/language-options.ts:28` — remove the `enabled` skip → **2 failed / 116 passed** |
| WCAG 2.5.3 Label in Name — the visible short name is contained in the accessible name | `manual` | Independently verified in review across **every** entry in `CANONICAL_LANGUAGE_CODES`: the short form is a case-insensitive substring of the full name for all of them, 0 mismatches |
| `isRTLLanguage` still resolves `"default"` after being refactored onto the shared helper | `unit (guard)`<br>`packages/surveys/src/lib/utils.test.ts` → `describe("isRTLLanguage")` | Pre-existing `isRTLLanguage` cases, unchanged and still green |
| Trigger names the active language, visibly and in its accessible name | `manual` | Built `surveys.umd.cjs`, rendered a three-language survey, read the DOM at the same moment as the screenshot. `aria-label` `"Language switch"` → `"Language: American English"`; visible text `""` → `"English"`; `title` is the language name alone, so it is not the accessible name repeated |
| `#fbjs` carries the active language, and never a language the survey lacks | `manual` | Same run: `lang` `null` → `"en-US"`, → `"de-DE"` for the German variant. `?lang=fr-FR` and `?lang=hi` (neither configured) both resolve to `lang="en-US"` with the marker on the default option — before the fix they declared the unconfigured tag and reverted the trigger to a bare globe. Also confirmed `lang={null}` omits the attribute rather than rendering `lang="null"` |
| `lang` and `dir` never disagree | `manual` | Read both off `#fbjs` in the built bundle across default, a configured language, an **unconfigured RTL** language and a live switch: `en-US/ltr`, `de-DE/ltr`, `?lang=he → en-US/ltr` (not `rtl`), `ja-JP/ltr` after switching |
| Only one switcher is ever in the DOM, so the single-`aria-current` claim holds | `manual` | Verified in review: `getShowLanguageSwitch(offset)` is true for `offset <= 0`, but `StackedCard` only calls `getCardContent` when `delayedOffset === 0` |
| Monorepo still builds and passes, after the merge and after the refactor | `manual` | Both times: `pnpm typecheck` (27 tasks), `pnpm lint` (0 errors, 40 warnings — same count as `main`), `pnpm test` (627 files / 8289 tests; `@formbricks/surveys` 704), `pnpm format:check`, `pnpm scan-translations` — all clean |
| Nothing #8919 added was lost in the merge | `manual` | Rebuilt `surveys.umd.cjs` from scratch and re-read `#fbjs` across six variants: exactly **one** `h1` (the survey name) on every one, the card headline as `h2`, no `h3`–`h6`. Then `aria-current` with the dropdown **open** — exactly one option marked, the correct one, on default / `de-DE` / an unconfigured `?lang=hi`, each option carrying its own `lang` |

**Open gaps**

- **No automated coverage of the `.tsx` changes.** Narrower than it was — `getVisibleSurveyLanguages` and `isSameLanguageCode` are now unit-tested, so the dedupe and which-option-is-active decisions are covered, alongside `resolveSelectedLanguageCode`, `getSurveyLanguageTag` and `getShortLanguageDisplayName`. What is still only verified by the manual bundle run is the **wiring**: that `language-switch.tsx` actually puts `aria-current` on the option `isActive` returns true for, that the trigger renders the label, and that `survey-container.tsx` passes `lang` to the root. A regression that drops one of those attributes will not be caught by CI. Deliberate — an e2e suite was judged disproportionate for this change — but it is the remaining gap here.
- **The offline-restore path is reasoned about, not exercised.** It is what makes the stale-language and `dir`-divergence cases reachable, and the resolver it flows through is unit-tested, but no test drives IndexedDB restore end to end.
- No screen reader was actually driven. The ticket asks for a manual NVDA/VoiceOver pass switching language mid-survey, plus an IBM Equal Access + ANDI re-run; neither is available in this environment.
- **The survey chrome can still disagree with the declared language.** For an unconfigured `?lang=hi`, content falls back to the default (English) and `lang` is now `en-US`, but `getI18nLanguage` still resolves the Hindi chrome bundle, so the button reads "भाषा:". Content and the declared tag agree where they previously did not; the chrome is a third answer. Pre-existing `getI18nLanguage` behaviour and arguably intentional, so left alone.
- **The endonym list is not audited for completeness.** The ticket explicitly asks to "confirm the endonym list is complete for all languages Formbricks supports". `NATIVE_NAMES` is compile-time-keyed to `iso639Languages`, so nothing can be *missing*, but a number of entries are English names rather than endonyms (`aa: "Afar"`, `an: "Aragonese"`, `bh: "Bhojpuri"`, …) and those languages will show an English label. Pre-existing, out of scope here, worth its own pass.
- `en-US` reads as "American English" in the dropdown and "English" on the trigger. Both come from the existing `NATIVE_NAMES` table; no wording decision was made in this PR.
- The visible label widens the trigger from `w-8` to `w-auto px-2`. Sanity-checked at 390 px and up, but no designer has signed it off.
- **`getSurveyLanguageTag` is now a near-duplicate of the copy in `apps/web/modules/survey/link/lib/utils.ts`** — see the open thread. A copy is structurally required today (`apps/web` consumes this package as a prebuilt UMD bundle), but the logic deserves hoisting into a package both already depend on. Filed as a follow-up rather than grown into this PR; the reasoning is in the thread.

**Risks**

- **Keyboard behaviour in the popup is unchanged and still thin**: no Escape handler, no arrow-key roving, no focus return to the trigger, and `aria-haspopup="true"` with no `role` on the popup. That is 2.1.1/2.4.3 territory, not 3.1.1, so it is deliberately left alone here — but a reviewer looking at this file will see it, and it deserves its own ticket.
- Long endonyms truncate rather than wrap on the trigger. The full name is always in the accessible name and `title`, so nothing is lost to AT or on hover.
- `getSurveyLanguageTag` now filters on `enabled`, and `isRTLLanguage` shares it — so a survey rendered with a disabled language's code gets the default language's direction rather than that language's. That is the intended reading (a language not offered should not drive layout), but it is a behaviour change on a shared helper.
- The direction effect now depends on `props.survey` as well as the active language, having previously suppressed that dependency. In the editor preview the survey object is rebuilt on every render, so the effect runs more often there; `setDir` with an unchanged value is a no-op, so the cost is one `isRTLLanguage` call, but it is more work per keystroke than before.
- `getSurveyLanguageTag` and `resolveSelectedLanguageCode` now delegate their canonical comparison to a shared helper. That is one definition instead of four, but it also means a bug in `isSameLanguageCode` would surface in both the `lang` attribute and the emitted `response.language`. The mutation rows above are what stand behind it.

> [!NOTE]


packages/surveys/src/lib/utils.ts
```diff
@@ -218,6 +218,23 @@ export const getDefaultLanguageCode = (survey: TJsWorkspaceStateSurvey): string
...
+/**
+ * Resolves the survey's active language to a real language tag, usable as a `lang` attribute.
+ *
+ * The renderer tracks the active language as either a stored language code or the sentinel
+ * `"default"`. `"default"` is not a language tag and must never reach the DOM, so it is resolved
+ * to the code of the survey's default language. Returns `null` when the survey has no languages
+ * configured at all: such a survey has no language to declare and should inherit the host
+ * document's, rather than assert a guess.
+ */
+export const getSurveyLanguageTag = (
```
kiln-agent, line 242: This is a second, byte-for-byte copy of `getSurveyLanguageTag` — `apps/web/modules/survey/link/lib/utils.ts:64` already has one (same signature, same `"default"` handling, same `null` return), added for the `<html lang>` sync that `survey-client-wrapper.tsx:183` performs.

Two independent copies now decide the *same* thing for the same page: on a link survey `<html lang>` comes from the apps/web copy and `#fbjs lang` from this one. Any future change to one silently desynchronises the two attributes on the same document — and they have already diverged in this very PR: `isRTLLanguage` here was refactored onto the shared helper (so `languageCode === ""` now resolves to the default language instead of returning `true` from `"".startsWith("")`), while the apps/web copy of `isRTLLanguage` still inlines the old branch.

`normalizeLanguageCode` is imported from `@formbricks/i18n-utils/src/canonical` by both `apps/web` and `packages/surveys`, so that package is an existing shared home for this — worth hoisting `getSurveyLanguageTag` (and ideally `isRTLLanguage`) there and deleting both copies.

---

packages/surveys/src/lib/utils.ts
```diff
@@ -218,6 +218,23 @@ export const getDefaultLanguageCode = (survey: TJsWorkspaceStateSurvey): string
...
+  survey: TJsWorkspaceStateSurvey,
+  languageCode: string
+): string | null => {
+  if (languageCode && languageCode !== "default") return languageCode;
```
kiln-agent: `getSurveyLanguageTag` returns any non-`"default"` code verbatim, without checking it is one of the survey's configured languages. That was harmless while the only consumer was `isRTLLanguage` (a direction guess), but the return value now lands in a DOM `lang` attribute and drives the "current language" marker, where an unconfigured code is actively wrong.

Concrete scenario: `Survey` restores offline progress and calls `setSelectedLanguage(progress.selectedLanguage)` (survey.tsx:526). That restore validates `blockId` against the current survey but not the language, so after a language is removed or disabled in the editor, a returning respondent gets `selectedLanguage: "fr-FR"` on a survey that no longer has French. `getLocalizedValue` then falls back to the `default` content (i18n.ts:32-37), so English text is rendered while `#fbjs` declares `lang="fr-FR"` — a screen reader reads English with French pronunciation rules, which is worse than declaring nothing. In the same state `visibleLanguages.find(isActive)` returns `undefined`, so the trigger reverts to the unlabelled globe and no option is marked current, i.e. exactly the bug this PR sets out to fix, silently. The same applies to a direct `window.formbricksSurveys.renderSurvey({ languageCode: "fr" })` embed (js-core guards this path, the public entry point does not).

Resolving against the survey's configured languages — `getI18nLanguage`-style matching, falling back to the default code — would keep the declared tag and the rendered content in agreement.

---

packages/surveys/src/components/general/language-switch.tsx
```diff
@@ -94,10 +123,11 @@ export function LanguageSwitch({
   return (
     <div className="z-1001 flex w-fit items-center">
       <button
-        title={t("common.language_switch")}
+        title={triggerLabel}
```
kiln-agent: `title` and `aria-label` now carry the *same* string. Per HTML-AAM, `aria-label` wins the accessible **name** and `title` then supplies the accessible **description**, so NVDA/JAWS/VoiceOver announce the identical sentence twice: "Language: American English, button — Language: American English".

That duplication existed before too, but it was two words (`"Language switch"`); the new label is a full sentence and it repeats on every focus, which is exactly the verbosity a 3.1.1 fix is supposed to avoid. Since the visible endonym is already `aria-hidden`, the hover tooltip is the only thing `title` still buys — and only for the truncation case flagged under *Risks*.

Suggest keeping `title` as the untruncated **language name only**, so the tooltip stays useful without restating the name:

```suggestion
        title={activeDisplayName ? getLanguageDisplayName(activeLanguage.language.code) : triggerLabel}
```

(or drop `title` altogether and leave the endonym's own `title` on the span.) Flagging as PLAUSIBLE — worth a quick NVDA check during the manual screen-reader pass the description already lists as an open gap.

---

packages/surveys/src/components/general/render-survey.tsx
```diff
@@ -1,23 +1,42 @@
...
+  //
+  // The active language lives in Survey's own state, and it already reports every change through
+  // onLanguageChange. Wrapping that callback keeps a single source of truth rather than adding a
+  // second channel out of Survey. Re-firing with an unchanged code is a no-op in both setters.
+  const handleLanguageChange = useCallback(
+    (languageCode: string) => {
+      setLanguageTag(getSurveyLanguageTag(props.survey, languageCode));
+      onLanguageChange?.(languageCode);
+    },
+    [props.survey, onLanguageChange]
```
kiln-agent: `props.survey` in the dep list turns what used to be the caller's own (usually stable) `onLanguageChange` into a callback whose identity changes whenever the survey object identity changes. `Survey`'s reporting effect is keyed on it — `useEffect(() => { onLanguageChange?.(selectedLanguage); }, [selectedLanguage, onLanguageChange])` (survey.tsx:467-469) — so that effect now re-fires on every such change, not just on an actual language switch.

That is not hypothetical: `preview-survey/index.tsx:278` passes `survey={toJsWorkspaceStateSurvey(survey)}`, an unmemoized fresh object on every render, and `SurveyInline` re-invokes `renderSurvey({...props})` on every parent render (`modules/ui/components/survey/index.tsx:87`). In the editor preview the effect therefore runs on every keystroke. Today nothing breaks — the preview passes no `onLanguageChange`, and `setLanguageTag` bails on an unchanged value — but `onLanguageChange` is a public `SurveyContainerProps` field, so any host that wires up a non-idempotent handler (an analytics event, a URL rewrite, a fetch) will get a burst of spurious calls, and the cause will be invisible from their side.

Keeping the survey out of the dependency (a ref, or depending on the resolved default code rather than the whole object) preserves the "single source of truth" design without making the callback identity churn.

---

kiln-agent (commented):
Code review (correctness pass, effort `high`). Three findings; one of them blocks CI.

The production logic holds up well — I verified the two claims the description leans on:

- `getSurveyLanguageTag` / `isActive` dedupe matching: `normalizeLanguageCode` preserves region variants (`en-GB` → `en-GB`, not `en-US`), so an alias-coded active language resolves to its canonical option and `en-US`/`en-GB` are not collapsed. ✔
- WCAG 2.5.3 (Label in Name): I ran `getShortLanguageDisplayName(code)` against every entry in `CANONICAL_LANGUAGE_CODES` and the short form is a case-insensitive substring of the full name for **all** of them — 0 mismatches. The `short.toLowerCase() !== languageWithScript.toLowerCase()` guard also correctly rescues `zh-CN`/`zh-TW`/`zh-HK` (no bare `zh` entry) back to the full name. ✔
- The `isRTLLanguage` refactor is behaviour-preserving except for `languageCode === ""`, where the old inline branch returned `true` (`"ar".startsWith("")`) and the new one resolves to the default language. That's a fix, not a regression, and no caller passes `""` today (`getLanguageCode` in `survey-renderer.tsx` and `getLanguageCode` in `js-core/lib/common/utils.ts` both yield `"default"` or a stored code).
- Also checked the `toHaveCount(1)` assertions can't be tripped by multiple simultaneously-rendered switchers: `getShowLanguageSwitch(offset)` is true for `offset <= 0`, but `StackedCard` only calls `getCardContent` when `delayedOffset === 0`, so exactly one switcher is in the DOM at a time. ✔

---

### 1. HIGH — `apps/web/playwright/survey-language-switch.spec.ts:188`: unescaped regex makes the German assertion unsatisfiable

GitHub has not yet materialised this file in the PR diff (`changed_files: 33`, but the files API and `.diff` return only the 32 `packages/surveys` entries), so I could not attach this inline. It is in the head commit `808c826`.

```ts
await expect(switcherTrigger(page)).toHaveAccessibleName(new RegExp(FULL_NAME[GERMAN_CODE]));
```

`FULL_NAME["de-DE"]` is `"Deutsch (Deutschland)"`, and `new RegExp` treats the parentheses as a capture group. The pattern is `/Deutsch (Deutschland)/`, which matches the literal text `Deutsch Deutschland` — not `Deutsch (Deutschland)`. The accessible name after switching is `Sprache: Deutsch (Deutschland)`, so the assertion can never pass and the test fails on timeout:

```
$ node -e 'console.log(new RegExp("Deutsch (Deutschland)").test("Sprache: Deutsch (Deutschland)"))'
false
$ node -e 'console.log(new RegExp("Deutsch (Deutschland)").test("Sprache: Deutsch Deutschland"))'
true
```

Line 152 carries the same latent hazard and passes only because `"American English"` happens to contain no regex metacharacters — worth fixing both, e.g.

```ts
const rx = (s: string) => new RegExp(s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"));
await expect(switcherTrigger(page)).toHaveAccessibleName(rx(FULL_NAME[GERMAN_CODE]));
```

or drop the regex and assert the full deterministic `Sprache: …` string.

### 2. LOW — duplicated `getSurveyLanguageTag`
See inline on `packages/surveys/src/lib/utils.ts`.

### 3. LOW (PLAUSIBLE) — `title` + `aria-label` announce the same sentence twice
See inline on `packages/surveys/src/components/general/language-switch.tsx`.

---

packages/surveys/src/lib/utils.ts
itsjavi: Agreed on the substance, but I'm leaving the hoist out of this PR — it's a cross-package refactor and this branch is a Level A a11y fix.

Two things confirmed from your note:

**The `""` divergence is real, and it's a fix rather than a regression.** The old inlined branch computed `baseCode = "".split("-")[0] = ""`, and `RTL_LANGUAGES.some(rtl => rtl.startsWith(""))` is `true` for every entry — so an empty `languageCode` used to report the survey as RTL. Routing through the shared helper resolves `""` to the default language instead. No test covered it either way, so it was silent. The apps/web copy of `isRTLLanguage` still has that branch and still returns `true` for `""`.

**A copy is structurally required today, not just convenient.** `apps/web` consumes `@formbricks/surveys` as a prebuilt UMD bundle served from `public/js/`, so it cannot import from `packages/surveys/src` without pulling Preact into the Next bundle; and `packages/surveys` obviously cannot import from `apps/web`. So the two `lang` writers genuinely need two reachable implementations unless the logic moves to a package both already depend on.

On the proposed home: `@formbricks/i18n-utils` is a good instinct — both sides already deep-import `./src/canonical` from it — but `getSurveyLanguageTag` takes a *survey*, so putting it there either drags survey types into a language-code package or needs a `(languages, code)` signature plus an adapter at each call site, which leaves two functions anyway. My preference would be a small shared module that takes `TSurveyLanguage[]`, hosting both `resolveLanguageTag` and the RTL lookup, with `getSurveyLanguageTag` / `isRTLLanguage` reduced to one-line survey-shaped wrappers on each side — that kills the *logic* duplication, which is the part that can desynchronise, and fixes the apps/web `""` case in the same move.

That's a bigger change than belongs here, so I've filed it as a follow-up rather than growing this PR. What this PR does do is stop the divergence getting worse in the meantime: the local helper now resolves against the survey's enabled languages (b07e646), which is a behaviour the apps/web copy does not have yet and which the follow-up should carry across.

---

packages/surveys/src/lib/utils.ts
itsjavi: Good catch — fixed in b07e646. The helper now resolves the code against the survey's **enabled** languages, canonical-aware, returning the stored code and falling back to the default when nothing matches. That is the same rule the server's `getLanguageCode` applies to `?lang=`, so client and server now agree.

Verified in the built bundle: `?lang=fr-FR` and `?lang=hi` on an en-US/de-DE/ja-JP survey both now give `lang="en-US"` with the marker on the default option, where before they declared the unconfigured tag and reverted the trigger to the bare globe.

Four unit cases cover it (`packages/surveys/src/lib/utils.test.ts`) — unconfigured code, configured-but-disabled, legacy alias resolving to the stored canonical code, and no-match-with-no-default returning `null`. Confirmed they bite: replacing `if (configured) return configured.language.code;` with `return languageCode;` at `packages/surveys/src/lib/utils.ts:238` turns all four red.

One thing your scenario surfaced that this does *not* fix: for an unconfigured `?lang=hi`, the survey chrome still renders in Hindi (`getI18nLanguage` resolves any tag against the bundled locales), so the button reads "भाषा:" while content and `lang` are both `en-US`. Content and the declared tag agree now, which they did not before, but the chrome is a third answer. That is pre-existing `getI18nLanguage` behaviour and arguably intentional — a respondent who asked for a Hindi UI gets one — so I have left it alone and noted it in the PR description rather than widening this change.

---

mattinannt (dismissed):

mattinannt (dismissed):

mattinannt (approved):

PR #8921, opened 2026-08-19
fix(analysis): close the chart gaps found rebuilding a customer dashboard [ENG-2473]

Dhruwang:
## What & why

Rebuilding a customer's reporting deck against the current charting stack (46 elements as 50 live widgets, ~3,000 generated feedback records across five surveys) surfaced six gaps. Five are fixed here; two of the eight items on the ticket are still open, hence `Ref` rather than `Fixes`.

**1. An unasked measure reported itself as zero.** Cube returns NULL for a measure with nothing to compute — `npsScore` and `csatScore` do that deliberately when a survey never asked the question. The client's `tablePivot()` substitutes a fill value, defaulting to `0`, so on a four-survey NPS chart three bars read `0` when only one had been measured, and one of those was a *genuine* 0.00. Note `fillWithValue: null` does **not** fix this: the client resolves each cell as `row[measure] ?? fillWithValue ?? 0`, so a null fill falls straight back through to 0. Fills with a sentinel and maps it back to null. Two downstream coercions would have flattened it again — `pivotMeasuresToCategories` (`Number(null)` is 0, which is finite) and `big_number`'s `|| 0`.

**2. Averages printed three decimals.** `4.705` where a 1–5 scale wants `4.71`, on all 37 journey cards at once. `big_number` now formats through `formatCellValue`, so a card and a bar label agree on precision.

**3. Dashboard widgets can show their data as a table.** Two deck layouts are tables and a widget could only draw a chart. Rather than add a `table` chart type with a second rendering path, this reuses the chart builder's existing `DataViewer`: each widget gets a two-icon toggle beside its menu. The choice is per viewer, not saved on the chart — one person reading exact numbers does not change the dashboard for everyone — stored in `localStorage` keyed by **widget** id, so the same chart on two dashboards is remembered separately.

**5. Widget titles clipped.** A score card's title is the survey question, and at a quarter of the grid a widget is ~270px wide, so titles were cut off after one line — on a dashboard of 37 cards the title is the only thing telling them apart. Wraps to two lines, full text in the native tooltip.

**6. The default-language bucket had no name.** A response given in its survey's default language stores no `language` at all (`transform.ts` only writes it when it is not `"default"`), so grouping by language produced an unlabelled bar beside `ar-SA` and `en-GB` — the entire English audience showing as blank. It is a real group, not missing data, so it gets a name, mirroring how an empty enrichment dimension already renders as "Not enriched".

**4. Choice questions now steer toward option grouping — as a nudge, not a rewrite.** Grouping by "Value (Text)" looks like the natural pick and then splits one option into several buckets the moment the text differs (a translated label, a label edited after responses arrived, or a free-text "other" answer). "Value (Option)" is now listed first and says what it is for, "Value (Text)" spells out how it buckets, and picking text without option raises an inline hint.

This deliberately stops short of a silent `valueText → valueId` swap. ENG-1673 decided the query must preserve the dimension the user picked — see the comment and four tests in `charts/actions.test.ts` — and regrouping saved charts under people would also collapse free-text "other" answers that someone may be charting on purpose. I implemented the swap first, those tests failed by design, and I backed it out.

## Linear ticket

Ref ENG-2473

## How this was tested

**Automated**

| Suite | Result |
| --- | --- |
| `modules/ee/analysis` (whole surface) | **425 passed / 27 files**, two consecutive green runs |
| `cube-client.test.ts` | 11 passed — new: pivot is called with the sentinel, the sentinel maps back to null, a genuine `"0.00"` survives as zero, rows without a sentinel pass through untouched |
| `widget-view.test.ts` (new) | 8 passed — per-widget independence, unknown stored values, default clears the entry, both throwing-storage paths (Safari private mode throws rather than returning null) |
| `chart-utils.test.ts` | 45 passed |

Two pre-existing `pivotMeasuresToCategories` tests asserted the old `0` behaviour; they encoded the bug, so they now assert null, plus a new case pinning that a genuine `0` and a null stay distinguishable.

One earlier run had a single failure in `charts/actions.test.ts` that did not reproduce on either subsequent full run — flaky, not related to this diff, but worth knowing.

**Manual, in a browser against Cube-backed data**

Item 1 — the exact chart from the report, NPS by survey across four surveys where only two asked it:

| Survey | Before | After |
| --- | --- | --- |
| Survey A (never asked) | `0` | *no bar, no label* |
| Survey B (never asked) | `0` | *no bar, no label* |
| Survey C | `72.92` | `72.92` |
| Survey D (**genuine** zero) | `0.00` | `0.00` |

A card whose filter matches nothing shows a muted en dash, verified with a throwaway widget (since removed).

Item 2 — Overall CSAT `4.558` → `4.56`; arrival easiness `4.705` → `4.71`.

Item 3 — toggle renders on every widget; `aria-label` Chart/Data with `aria-pressed` tracking state; tooltip opens and reads "Chart"; Data view renders the `DataViewer` table inside the widget; the view survived a reload **and** a navigation away and back. Isolation proved deterministically: cleared storage → all 50 widgets showed charts → clicked exactly one Data icon → exactly one widget flipped and exactly one key was written → switching back removed the key.

Item 5 — at 1400px (12-column grid, 268px widget) the title renders on two lines instead of one plus ellipsis, with `title` carrying the full string. Below 996px the grid drops to 6 columns and titles already fitted, unchanged.

Item 6 — the by-language chart now reads `ar-SA / en-GB / Default language`; previously the third bucket was blank.

Item 4 — dropdown order confirmed as `Value (Option)` → `Value (Text)` → `Value (Number)` …, and selecting Value (Text) alone raises exactly the intended hint.

**Not verified end to end**

Item 4's *effect* on real fragmentation could not be reproduced locally: it needs a choice question mapped through a Formbricks feedback source, and this environment has exactly one mapping and it is a `rating`. The bilingual fragmentation itself was reproduced earlier with synthetic records (a gender chart returning `Male 151 / Female 54 / ذكر 40`), which is what the hint warns about.

**Visual evidence**

Two kinds of artefact below, and it is worth knowing which is which. The **charts** (items 1 and 6) are each chart's own SVG output, pulled from the running app's DOM and rasterised — geometry, colours, ticks and value labels are exactly what the app drew, with axis labels re-emitted as `<text>` because recharts draws them in a `foreignObject`. They live on the orphan branch `assets/eng-2473`. Everything else is a plain screen capture. Every pair was taken from the same dashboard, the same data and the same viewport, with only the diff differing.

**Item 1 — an unasked measure no longer reports itself as zero**

Four surveys, only two of which ask an NPS question.

| Before (`main`) | After |
| --- | --- |
| NPS by survey on main: two surveys that never asked NPS print 0 | NPS by survey after the fix: the unasked surveys draw nothing |
| Two of the four surveys print **`0`** despite never asking the question — indistinguishable from the fourth, whose `0.00` is a real score | The unasked surveys draw nothing at all; the genuine `0.00` is untouched |

**Item 6 — the default-language bucket has a name**

| Before (`main`) | After |
| --- | --- |
| CSAT by language on main: the third bucket has no label | CSAT by language after the fix: the third bucket reads Default language |
| The third bar — every response given in the survey's default language — carries no label | The same bar reads **Default language** |

**Item 5 — long widget titles wrap instead of clipping**

A score card's title is the survey question it measures, and at a quarter of the grid a widget is ~270px wide.

| Before (`main`) | After |
| --- | --- |
| Widget header on main: the title is clipped after one line | Widget header after the fix: the title wraps onto two lines |
| `KAS Pre-match: Ease of getting access to acco…` — clipped mid-word after one line | `KAS Pre-match: CSAT with arrival to KSA` wraps onto two lines, with the full string in the `title` tooltip |

Only the widget header was reverted for the left-hand capture, so the number in it already shows the item-2 precision fix — read that pair for the title treatment alone.

**Item 3 — chart / data toggle**

Switching one widget between its chart and the underlying data table, including the tooltip on each icon. The other widgets on the dashboard are unaffected, which is the point: the choice is per viewer and per widget, not saved on the chart.

https://github.com/user-attachments/assets/073f5d29-8d59-4500-bdab-6414da368d2c

**Item 4 — grouping hint in the chart builder**

Group By set to Value (Text), with an inline hint recommending Value (Option)

Selecting **Value (Text)** on its own raises the hint; adding **Value (Option)** clears it. Note the dropdown now lists Value (Option) first, and both options explain when to use them.

**Item 2 — precision, no dedicated capture**

`csatAverage` printing three decimals was the defect; it now formats through `formatCellValue` like every bar label. Observed on the same dashboard: Overall CSAT `4.558` → **`4.56`**, arrival easiness `4.705` → **`4.71`**. The cards in the item 5 pair show the fixed formatting (`4.8`, `4.51`).

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Open a dashboard with a chart whose measure has no data for some groups — e.g. a bar chart of `NPS: Score` grouped by survey where only some surveys ask an NPS question → the surveys that never asked it show **no bar and no number**; a survey that genuinely scored zero still shows `0.00`
- [ ] Open a big-number card whose filter matches nothing (any question id that has no answers) → shows a greyed **–**, not `0`
- [ ] Look at any card showing an average on a 1–5 scale → at most two decimals (`4.71`, not `4.705`), and the same number in a bar label matches
- [ ] On any dashboard widget, use the two-icon toggle in the top right → left icon shows the chart, right icon shows a data table of the same numbers; hovering each icon shows a "Chart" / "Data" tooltip
- [ ] Switch one widget to Data, reload the page → that widget is still on Data and **every other widget is still a chart**
- [ ] Switch it back to Chart, reload → still a chart
- [ ] Add the same chart to a second dashboard, set one to Data → the other dashboard's copy is unaffected
- [ ] Enter dashboard edit mode and click the toggle → the view changes and the widget does **not** start dragging
- [ ] Find a widget whose title is longer than its width (a score card at full desktop width) → the title wraps to two lines; anything beyond that is available by hovering the title
- [ ] Build a chart grouped by response language on a survey with more than one language → the bucket for responses in the survey's default language reads "Default language" instead of being blank
- [ ] Create a chart → enable **Group data** → open the dropdown → `Value (Option)` is listed above `Value (Text)` and both explain when to use them; select only `Value (Text)` → an inline hint recommends Value (Option); also select `Value (Option)` → the hint disappears
- [ ] Existing saved charts: open a few that group by Value (Text) → they render exactly as before (no regrouping)

**Preconditions / test data**

- Dashboards entitlement enabled for the organization, and a feedback directory with data behind it
- For the null/zero cases: at least two surveys in the same directory where only some ask an NPS (or CSAT) question — one asking it with a real 0 score is what makes the distinction visible
- For the language case: a survey with two languages and responses in both, including some in the default language
- No feature flag; behaves the same on Cloud and self-host

**Risks & regressions**

- Charts that legitimately show `0` must keep showing `0` — the sentinel only replaces values Cube returned as NULL. The genuine `0.00` case is covered by a unit test and was checked on screen
- The sentinel string (`__formbricks_null__`) would be indistinguishable from a real answer of that exact text. No real answer looks like this, and it is only ever compared against measure cells the pivot filled
- Widget header height changed from a fixed to a minimum height; check widgets with short titles still look right and the grid does not shift
- Two new en-US strings (`language_value_default`, `prefer_option_grouping`) — other locales come from Lingo on commit; until then they fall back to English
- Item 4 changes only the builder's suggestion order and copy. If a reviewer would rather the query were rewritten server-side, that reverses the ENG-1673 decision and its four tests — worth a deliberate call, not a drive-by

- none


kiln-agent:
## Review

Reviewed at `5889293` (the branch moved twice past the review-request SHA `2420d07`; `11feb0b` and `5889293` are included below).

I verified the core mechanism rather than taking the description on trust. It holds up: `@cubejs-client/core` resolves a pivot cell as `row[measure] ?? fillWithValue ?? 0` (`ResultSet.js:416`), so `fillWithValue: null` really does fall through to `0` and only a non-nullish sentinel survives — and upstream has its own test that a genuine `0` is never replaced by the fill. The measure-scoping in `5889293` is the right correction. `modules/ee/analysis` is green locally: **427 passed / 27 files**.

Two findings I'd want resolved before merge, then some smaller things.

---

### 1. Zero-response buckets in a time series now render as gaps, not zeros

`fillWithValue` is not only applied to measures Cube computed as NULL. Cube's pivot also uses it for the rows it *synthesizes* for `fillMissingDates`, which defaults to `true` whenever `x` is a single granular time dimension (`ResultSet.js:294`, `:417-427`). Those synthesized rows are `{ row: {} }`, so the measure is `undefined` → sentinel → `null` after `restoreNullMeasures`.

Reproduced against the installed client — a `count` by day where Jan 2–3 have no rows at all:

```
BEFORE  tablePivot()                          AFTER  tablePivot({fillWithValue: '__formbricks_null__'})
2026-01-01  "5"                               2026-01-01  "5"
2026-01-02  0                                 2026-01-02  "__formbricks_null__"   → null
2026-01-03  0                                 2026-01-03  "__formbricks_null__"   → null
2026-01-04  "7"                               2026-01-04  "7"
```

For a *score* measure that is the fix working as intended. For a **count** it inverts the meaning: "5 responses Monday, none Tuesday, 7 Wednesday" is a true zero, and `connectNulls={false}` (`chart-renderer.tsx:455`, `:482`) now breaks the line into disconnected segments, while the bar and its value label disappear. "No responses that day" and "this question was never asked" become the same picture — the exact conflation the PR set out to remove, pointed the other way.

`restoreNullMeasures` receives `input.query.measures`, so it already knows which measures these are. Scoping the sentinel→null mapping to the measures that can legitimately be NULL (the score/average family) and leaving count-family measures at `0` would keep both readings honest. Passing `fillMissingDates: false` also removes the conflation, but at the cost of dropping empty dates from the axis entirely — probably not what you want.

Same theme, smaller: `big_number` now shows the muted en dash whenever `numericValues` is empty (`chart-renderer.tsx:505-513`), which includes a count card whose result set is simply empty. The QA plan lists that as intended, but for a count measure `0` was the accurate answer and `–` reads as "we couldn't tell you".

Worth noting the axis-scale path is fine here — `computeYAxis` already tolerates `null` (covered by `y-axis-scale.test.ts:57-60`). And no existing test exercises the `fillMissingDates` interaction, which is why the suite stayed green through this.

### 2. The sentinel can reach the UI when a measure column key isn't a bare measure id

`restoreNullMeasures` matches with `measures.has(key)` (`cube-client.ts:83`), which assumes a `tablePivot` measure column is always named exactly as the measure. That is true for the queries the builder emits today, but not in general: when `pivotQuery.dimensions` contains `compareDateRange`, `getNormalizedPivotConfig` unshifts it onto `y` (`ResultSet.js:322-324`) and `tablePivot` joins the y-values into the column name. Probed against the installed client:

```
column keys: [ 'FeedbackRecords.collectedAt.day', ',FeedbackRecords.count' ]
→ { 'FeedbackRecords.collectedAt.day': '2026-01-02T00:00:00.000',
    ',FeedbackRecords.count': '__formbricks_null__' }     // never restored
```

`compareDateRange` is part of `ZChartQuery` (`packages/types/analysis.ts:9`) and nothing in the builder sets it, so this is unlikely rather than routine — but `generateAIChartQuery` output is only validated against `ZChartQuery`, so a model emitting it is enough. The failure mode is bad out of proportion to its odds: the literal string `__formbricks_null__` rendered in a chart label or the data table.

Cheap to make robust either way — match a column key as ``key === m || key.endsWith(`,${m}`)``, or keep the current check and add a final sweep that nulls (and logs) any residual sentinel before returning, so the string can never escape whatever the pivot decides to call its columns.

---

### Smaller things

**3. "Default language" also names buckets that have no default language.** `isDefaultLanguageDimensionValue` (`schema-definition.ts:440`) keys purely off "language dimension + empty value", but `csv-transform.ts` never writes `language` unless the importer happened to map a column to it, and a CSV source has no survey and so no default language at all. A directory mixing survey responses and CSV imports will file every CSV record under "Default language". Also, per the comment at `lib/feedback-source/transform.ts:371-374`, a response in the survey's default language can carry its concrete code (`en-US`) — so the same audience can split across a named bucket *and* "Default language", which then reads as a distinct group it isn't. Gating on `source_type` (or naming it something that doesn't assert a survey, e.g. "Not specified") would avoid asserting more than the data supports.

**4. `chart-renderer.tsx:502` points the reader at the approach the PR disproved.** The comment reads "comes back as NULL (see `fillWithValue: null` in cube-client)" — but cube-client deliberately does *not* use `fillWithValue: null`; the whole point of the sentinel is that a null fill resolves back to `0`. As written it invites the next person to "simplify" it straight back into the bug.

**5. Orphaned JSDoc in `schema-definition.ts:645-652`.** The new `getTranslatedFieldDescription` was inserted between `getTranslatedFieldLabel` and its doc comment, so two comment blocks now stack above the new function and `getTranslatedFieldLabel` is left undocumented — tooling attaches only the nearest block, so "Translate a field/measure ID" now documents the wrong function.

**6. The option-grouping hint has no question-type context.** `suggestsOptionGrouping` (`dimensions-panel.tsx:33`) fires on any `valueText`-without-`valueId` selection, so a chart grouping open-text answers — where Value (Option) means nothing — permanently carries a warning telling the user to switch. The copy hedges with "For single-select and multi-select questions", so it isn't wrong, just always-on. If the panel can see the `fieldType` filter, gating on it would make the nudge land only where it applies.

---

Item 3 (the per-viewer widget view) reads well — `localStorage` keyed by widget id with the default clearing its entry, the effect-based read to keep hydration honest, throwing-storage paths covered, and the drag-suppression moved onto the buttons in `2420d07` rather than a handler-bearing wrapper. No concerns there beyond orphaned keys accumulating for deleted widgets, which isn't worth code.

The write-up is unusually good, and backing out the `valueText → valueId` swap because ENG-1673's tests said so was the right call.

---

apps/web/modules/ee/analysis/api/lib/cube-client.ts
```diff
@@ -91,7 +120,10 @@ export async function executeTenantScopedQuery(input: TScopedCubeQueryInput) {
   try {
     const client = cubejs(token, { apiUrl });
     const resultSet = await client.load(expandPresetDateRanges(input.query) as Query);
```
kiln-agent, line 137: The sentinel can't distinguish "Cube returned NULL for a measure" from "the pivot invented this row", and the second case is a live path here: `fillMissingDates` defaults to `true`, so any query with a time-dimension **granularity** gets synthetic rows for gaps in the range, and those rows go through the same `measureValue()` fill.

Confirmed against the installed `@cubejs-client/core` — same query, only the pivot call differing:

```
BEFORE  2026-01-02 → { "FeedbackRecords.count": 0 }
AFTER   2026-01-02 → { "FeedbackRecords.count": "__formbricks_null__" }   → null
```

In `ResultSet.js` the branch fires whenever `x.length === 1` and x equals the granular time dimensions, which is exactly what `getNormalizedPivotConfig` produces for these queries (`x: [timeDimension]`, `y: dimensions`) — so it holds for grouped time series too, not just the ungrouped case.

Downstream, for a **count over time** chart a zero-activity day is real data, not "not measured":

- line / area: `connectNulls={false}` (chart-renderer.tsx:455,482) now breaks the line across empty days instead of dipping to 0
- bar: the 0 bar disappears
- `big_number` over an all-empty range: en dash instead of `0`

The NULL-vs-zero fix itself is right and worth having — it's the conflation with date-gap fills that needs narrowing. Cheapest fix I can see: only restore nulls when the query has no granular time dimension, or pass `fillMissingDates: false` (the app fills its own date axis elsewhere anyway) so the sentinel can only ever come from a real NULL cell. Either way it'd be worth a test pinning that a count-by-day query with a gap still yields `0`.

---

apps/web/modules/ee/analysis/dashboards/components/dashboard-widget-data.tsx
```diff
@@ -27,6 +34,16 @@ export function DashboardWidgetData({ dataPromise, chartType }: Readonly<Dashboa
     );
   }
 
+  // The data view reuses the chart builder's table, so the numbers behind a widget read the same
+  // way in both places. It scrolls inside the widget rather than resizing it.
+  if (view === "data") {
```
kiln-agent, line 40: Reusing `DataViewer` is the right call, but it was written as a standalone block in the builder and brings chrome that doesn't fit inside a widget: its own `rounded-lg border bg-gray-50 p-4` card, a `DatabaseIcon` + "Chart data" heading (data-viewer.tsx:40-44), and a hard `max-h-64` scroll box (line 45).

Nested in `h-full overflow-auto` that gives two problems at the two ends of the size range:

- a default `h: 3` widget → outer and inner scrollbars side by side
- an `h: 6` widget → table capped at 256px with dead space below it

Plus the heading duplicates the widget title bar right above it. A `compact`/`bare` prop on `DataViewer` (drop the card + heading, let the height come from the parent) would keep the single rendering path you were after without the doubled chrome.

---

apps/web/modules/ee/analysis/charts/components/dimensions-panel.tsx
```diff
@@ -20,10 +26,18 @@ export function DimensionsPanel({
 }: Readonly<DimensionsPanelProps>) {
   const { t } = useTranslation();
 
+  // Grouping a choice question by its answer text is the trap this hint exists for: it looks like
+  // the natural pick, then splits one option into several buckets as soon as a translated label, an
+  // edited label or a free-text "other" answer shows up. Nudge rather than rewrite the query, so a
+  // chart never silently regroups itself under someone.
```
kiln-agent, line 32: The condition is purely "valueText selected, valueId not", with no notion of the question type — but the panel is shared with charts over **open text**, where `valueText` is the only workable dimension and the hint's advice actively breaks the chart: `transform.ts` only sets `value_id` for answers that matched a known choice (`normalizeChoiceValue`, and the `other` fallback in the multi-select expansion), so open-text records carry no `value_id` and grouping by Value (Option) collapses the whole chart into one null bucket.

Scoping it to charts whose filters pin a choice-ish `fieldType` would keep the nudge where it helps. If that's more plumbing than it's worth right now, wording it as a conditional ("if this is a choice question, prefer…") would at least stop it from pointing text charts at a dead end.

Unrelated to whether the hint fires, one nuance for the copy: for records written since ENG-1673, `value_text` is already canonicalised to the choice's **default-language** label, so the translated-label split it warns about only affects older records and labels edited after responses arrived. Free-text "other" answers still fragment as described.

---

apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx
```diff
@@ -159,16 +167,35 @@ const MemoizedWidgetItem = memo(function WidgetItem({
...
+
+  const handleViewChange = useCallback(
+    (nextView: TWidgetView) => {
+      setView(nextView);
+      writeStoredWidgetView(widget.id, nextView);
+    },
+    [widget.id]
+  );
+
+  const hasData = Boolean(widget.chart && dataPromise);
```
kiln-agent, line 186: Minor: the `view` prop is documented as "Omitted for widgets with no data behind them (skeletons, load errors)", but this only checks that a promise *exists*. `executeWidgetQuery` resolves with `{ error }` rather than rejecting, so a widget whose query failed still gets the toggle — and since `DashboardWidgetData` returns the error branch before the `view === "data"` check, both icons render the same error state and clicking does nothing visible.

The parent can't know the outcome without awaiting, so honouring the doc would mean moving the toggle inside the Suspense boundary. Not worth restructuring for; softening the comment to say it covers widgets with no chart or no query would keep the contract honest.

---

apps/web/modules/ee/analysis/charts/components/chart-renderer.tsx
```diff
@@ -495,15 +499,27 @@ export function ChartRenderer({ chartType, data, query, optionLabels }: Readonly
         />
       );
     case "big_number": {
-      const total =
-        data.length === 1
-          ? Number(data[0]?.[dataKey]) || 0
-          : data.reduce((sum, row) => sum + (Number(row[dataKey]) || 0), 0);
-      const formatted = total.toLocaleString();
+      // A measure with nothing to compute comes back as NULL (see fillWithValue: null in
```
kiln-agent: Nit: the comment points at `fillWithValue: null in cube-client`, which is the approach the PR body explains *doesn't* work — the implementation fills with `NULL_FILL_SENTINEL` and maps it back. Worth pointing at `restoreNullMeasures` instead so the next reader doesn't go looking for a null fill that isn't there.

---

kiln-agent (commented):
Reviewed all 28 files at `11feb0b` (10 commits on `4b6824b`). Read every hunk plus the surrounding renderer / util / UI files, checked the installed `@cubejs-client/core` pivot implementation, and ran the affected unit suites — the analysis surface is green locally.

The write-up made this reviewable, and the restraint on item 4 (implementing the `valueText → valueId` swap, watching the ENG-1673 tests fail, backing it out) is the right call. Five findings, one of which I think should be settled before merge.

**Needs a decision before merge**

- **The null sentinel also catches date-gap fills.** `fillMissingDates` defaults to `true`, so any query with a granularity gets synthetic rows for empty buckets, and those go through the same fill. I reproduced it against the installed client: a count-by-day query with a gap returns `count: 0` per empty day today and the sentinel — hence `null` — after this change. For a count-over-time chart a zero-activity day is real data, so line/area charts (`connectNulls={false}`) break the line, bar charts lose the 0 bar, and an all-empty range in `big_number` shows the en dash. Details and two suggested fixes inline on `cube-client.ts`. The NULL-vs-zero fix is right; it just needs narrowing to cells Cube actually returned as NULL.

**Worth addressing**

- `DataViewer` brings its own card, "Chart data" heading and `max-h-64` into the widget, so small widgets get nested scrollbars and tall ones a 256px-capped table.
- The option-grouping hint fires on any `valueText` grouping, including open-text charts, where `value_id` is null and following the advice collapses the chart into one bucket.

**Minor**

- The toggle renders on widgets whose query failed, which the `view` prop's doc comment says it won't.
- A comment in `chart-renderer.tsx` cites `fillWithValue: null`, the approach the PR explains doesn't work.

Nothing else on the diff concerned me: the sentinel's collision risk is genuinely negligible and scoped to filled measure cells, the `localStorage` handling covers throwing storage and keys per widget rather than per chart, the `fieldset` grouping with `aria-pressed` and the `onPointerDown` stop-propagation against react-grid-layout are all handled, and the two pre-existing `pivotMeasuresToCategories` tests did encode the old behaviour, so flipping them is correct.

---

apps/web/modules/ee/analysis/lib/schema-definition.ts
```diff
@@ -430,6 +433,13 @@ const isEmptyDimensionValue = (value: unknown): boolean =>
 export const isNotEnrichedDimensionValue = (dimensionId: string, value: unknown): boolean =>
   isEnrichmentDimensionId(dimensionId) && isEmptyDimensionValue(value);
 
+/** A response given in its survey's default language stores no language code (transform.ts only
+ * writes `language` when it is not "default"), so grouping by language yields an unlabelled bucket
+ * beside the explicit codes. It is a real group — the default-language responses — not missing data,
+ * so it gets a name rather than a blank axis tick. */
+export const isDefaultLanguageDimensionValue = (dimensionId: string, value: unknown): boolean =>
```
kiln-agent: The predicate is "language dimension + empty value", but an empty language only means *default language* on the survey ingestion path. `transform.ts:174` omits `language` when it equals `"default"` — that's the case this label is for. Every other way a feedback record is created leaves the same column empty for a different reason:

- `language` is optional for **all** source types, and `SOURCE_TYPE_PRESET_OPTIONS` is `survey, review, feedback_form, support, social, interview, usability_test, nps_campaign` — a review or support record never captures one.
- the manual record drawer round-trips a blank field as `language: ""` (`unify-feedback/lib/utils.ts:103`).

So on a directory that mixes surveys with imported reviews (the point of a unified feedback directory), a chart grouped by language now labels that bucket **"Default language"**, asserting a language those records never had. Before, the blank tick was at least honestly ambiguous.

The `isNotEnrichedDimensionValue` precedent it mirrors is safe because enrichment is one global pipeline, so empty has exactly one meaning there. Language doesn't. Either gate the label on the record being survey-sourced, or name it for what the data actually says (e.g. "Not specified"), which stays true in both cases.

---

kiln-agent (commented):
One new finding on top of the comments already on this PR (the null-sentinel/`fillMissingDates` conflation, the grouping-hint scope, the `DataViewer` chrome inside a widget, the `view` prop contract and the stale `fillWithValue: null` comment are all still open above).

Also still unaddressed, and already flagged by CodeRabbit outside the diff: the toggle's `onPointerDown` stopPropagation cannot stop a grid drag — react-grid-layout 2.2.2 drags via react-draggable 4.5.0, which binds `mousedown`/`touchstart` and has no pointerdown path, so pressing an icon in edit mode and moving a few pixels still drags the widget and dirties the layout.

---

apps/web/modules/ee/analysis/api/lib/cube-client.ts
Dhruwang: Confirmed and fixed. `fillMissingDates` is on by default and the branch fires exactly when the x axis is the granular time dimensions, so a zero-activity day was becoming "no data" — my own responses-per-day chart would have shown it.

First pass (eaa55e09) kept the client's default fill for any granular query. That was too broad, as CodeRabbit then pointed out, so 6d6d4866 narrows it: a second `tablePivot({ fillMissingDates: false })` lists the buckets Cube actually returned, so a filled cell in one of those is a real NULL and a filled cell anywhere else belongs to a row the pivot invented and stays 0. Tests cover both in one result set.

apps/web/modules/ee/analysis/dashboards/components/dashboard-widget-data.tsx
Dhruwang: Agreed — added a `bare` variant to `DataViewer` that drops the card, the heading and the `max-h-64` cap and fills its parent; the widget passes it. That removes the doubled scrollbars at `h: 3`, the dead space at `h: 6` and the heading duplicating the widget title, while keeping the single rendering path.

apps/web/modules/ee/analysis/charts/components/dimensions-panel.tsx
Dhruwang: Both points taken. The hint now states the condition rather than asserting it ("If this is a single-select or multi-select question…") and says explicitly that Value (Text) is the right choice for open text, so it no longer points text charts at a dimension that would collapse them into one null bucket.

The copy was also overstating the translation case: since ENG-1673 ingestion canonicalises `value_text` to the default-language label, so that only splits older records or a survey whose default language differs. Reworded to name the causes that still hold — an edited label, a differing default language, and free-text "other" answers. Proper scoping by `fieldType` is the better fix and needs filter context plumbed into the panel; left as a follow-up.

apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx
Dhruwang: Right, the comment claimed more than the check delivers. Reworded to say it covers widgets with no chart or no query, and to note that a widget whose query failed does still get the toggle because the parent cannot know that without awaiting the promise. Left the structure alone.

apps/web/modules/ee/analysis/charts/components/chart-renderer.tsx
Dhruwang: Fixed — it now points at `restoreNullMeasures`, which is what actually happens, instead of a null fill that the PR body explains does not work.

apps/web/modules/ee/analysis/lib/schema-definition.ts
Dhruwang: Good catch, and the precedent comparison is the part that convinced me: enrichment has one global pipeline, language does not. On a directory mixing surveys with imported reviews the label was asserting a language those records never had. Renamed to "Not specified", which is true whether the value is missing because the response was in the survey's default language or because the source never captured one.

apps/web/modules/ee/analysis/api/lib/cube-client.test.ts
```diff
@@ -101,6 +101,131 @@ describe("executeTenantScopedQuery", () => {
...
+        measures: ["FeedbackRecords.count"],
+        timeDimensions: [{ dimension: "FeedbackRecords.collectedAt", granularity: "day" }],
+      },
+    });
+
+    // fillMissingDates is on by default, so the pivot invents a row per empty day and fills it the
+    // same way a NULL measure is filled. A day with no responses is a measured zero, so these
+    // queries keep the client's default fill instead of being turned into "no data".
+    expect(mockTablePivot).toHaveBeenCalledWith();
+  });
```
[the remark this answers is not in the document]
Dhruwang: Tests rewritten to assert the returned rows — one result set now carries a real NULL bucket and a synthesized one, and checks they come back as `null` and `0` respectively. On the second point: there is no measure-key normalization in the implementation, the requested measure names are used as-is, so nothing to remove.

apps/web/modules/ee/analysis/api/lib/cube-client.ts
```diff
@@ -91,7 +124,15 @@ export async function executeTenantScopedQuery(input: TScopedCubeQueryInput) {
...
-    const result = resultSet.tablePivot();
+    // See NULL_FILL_SENTINEL: a granular time dimension means the pivot also fabricates rows for
+    // empty date buckets, and those must stay 0 rather than becoming "no data".
+    const fillsEmptyDateBuckets = (input.query.timeDimensions ?? []).some((td) => Boolean(td.granularity));
+    const result = fillsEmptyDateBuckets
+      ? resultSet.tablePivot()
+      : restoreNullMeasures(
+          resultSet.tablePivot({ fillWithValue: NULL_FILL_SENTINEL }),
+          input.query.measures ?? []
+        );
```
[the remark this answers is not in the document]
Dhruwang: Fixed in 6d6d4866. `tablePivot` takes its config per call and is pure over the loaded result, so a second pivot with `fillMissingDates: false` lists the buckets that genuinely came back; a filled cell in one of those becomes `null`, and a filled cell in a bucket the pivot invented stays `0`. No extra round trip. `restoreNullMeasures` takes that decision as a predicate and defaults to treating every row as real, so the non-time-series path is unchanged.

jobenjada (approved):
thanks, great catches!

PR #8922, opened 2026-08-19
fix(security): stop publishing the bundled Valkey session store to the host [ENG-2184]

mattinannt:
Fixes ENG-2184

## What & why

The self-hosting `docker/docker-compose.yml` published the bundled Valkey service on the host with `ports: - "6379:6379"`, which binds `0.0.0.0:6379` with no password. That instance is Better Auth's `secondaryStorage` (`apps/web/modules/auth/lib/auth.ts`), holding live sessions — keyed by the raw session token — plus verification tokens and rate-limit counters.

Because the app resolves identity from the cached session JSON (better-auth's `findSession` returns the cached `{session, user}` without re-reading the DB on a cache hit, and the app DAL `getSession` only re-checks `isUserActive`), anyone able to reach the port could read every live session token and overwrite their own cached session to be served as any other user, including owners — no victim password and no `BETTER_AUTH_SECRET` required. Docker's published-port iptables rule also bypasses host firewalls like `ufw`, so a host-level firewall does not mitigate it.

The fix removes the host port mapping so Valkey is reachable only on the internal compose network. The app already connects via `REDIS_URL=redis://redis:6379` (service DNS), so nothing about app behaviour changes. This aligns the `redis` service with:
- the `postgres` service in the same file, which publishes no host port;
- `docker/migrate-to-v4.sh`'s `add_redis_service`, which already injects redis without a `ports:` mapping;
- the Helm chart, which ships `--requirepass` + `ClusterIP`.

The docs are updated too: the "Port conflicts" note no longer implies `6379` must be free; a warning tells **existing** self-hosters to remove the mapping themselves (updating does not rewrite an already-downloaded `docker-compose.yml`); and, for installs whose port was reachable from an untrusted network, a remediation step to invalidate potentially-forged sessions (Valkey's AOF volume persists them across the fix).

## Breaking changes

| Bundled Valkey (`redis` service) host port | `6379` published on `0.0.0.0` (host-reachable) | not published — internal compose network only | Self-hosters using the bundled compose who reached Valkey from the host, and **all existing installs** (updating does not rewrite their `docker-compose.yml`) | New installs: none. Existing installs: delete `ports: - "6379:6379"` from the `redis:` service and run `docker compose up -d`. If the port was reachable from an untrusted network, also invalidate potentially-forged sessions — set a new `NEXTAUTH_SECRET` in `.env` and `docker compose up -d --force-recreate formbricks` (not `restart`, which keeps the old env). If an external tool genuinely needs access, bind it to loopback (`127.0.0.1:6379:6379`). The Formbricks app itself is unaffected — it uses the internal service address. |

## Migrations & env

- No change required for the fix itself. Existing self-hosters whose bundled Valkey port was reachable from an untrusted network should follow the remediation in the Breaking-changes table (remove the mapping; rotate `NEXTAUTH_SECRET` and recreate the `formbricks` container). Full steps in `docs/self-hosting/setup/docker.mdx`.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `redis` service must not publish a host port | `unit (red on main)` — `docker/__tests__/formbricks-script.test.ts` → "the redis service publishes no host port". Asserts on the js-yaml-resolved compose model (`services.redis.ports` absent/empty), so anchors/`<<` merges can't smuggle a publish past it. | Re-adding `ports: - "6379:6379"` to the `redis` service and running `cd docker && pnpm test` → **1 failed / 5 passed**. With the fix → **all 6 pass**. |
| Existing-install remediation guidance is accurate | `manual` — code inspection of `update_formbricks()` in `docker/formbricks.sh` (rerun: `sed -n '980,990p' docker/formbricks.sh`) | Confirmed it runs only `docker compose pull` / `down` / `up -d` and never re-fetches the template, so the warning's premise (existing installs keep the mapping after updating) holds. |

`docker compose -f docker/docker-compose.yml config -q` still validates after the change.

**Open gaps**

- The "resolve identity from the cached blob" behaviour is upstream better-auth and is unchanged here — this PR closes the network exposure rather than altering session validation.
- Any other container on the compose default bridge can still reach Valkey unauthenticated. Adding `--requirepass` for defense-in-depth is intentionally deferred to a follow-up to keep this fix minimal and backportable.
- Existing installs are remediated by documentation only; the installer/update path is not auto-patched to strip the mapping. Propagating the callout into the dedicated one-click setup doc and a coordinated release advisory are follow-ups for a maintainer (raised in review).

**Risks**

- Only affects reachability of the bundled Valkey from the host. Anything that relied on `localhost:6379` against the compose Valkey must switch to the internal network address (see Breaking changes).

---

> [!NOTE]
> **AI model used** — `claude-opus-4-8`, reasoning effort `high`.

### DIFF
--- a/docker/__tests__/formbricks-script.test.ts
+++ b/docker/__tests__/formbricks-script.test.ts
@@ -1,3 +1,4 @@
+import { load } from "js-yaml";
 import { execFileSync } from "node:child_process";
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
@@ -75,21 +76,20 @@ describe("docker/docker-compose.yml Redis/Valkey exposure (ENG-2184)", () => {
   // The bundled Valkey is Better Auth's session/token store (secondaryStorage). Publishing it to
   // the host binds 0.0.0.0:6379 with no password, exposing every live session token — and Docker's
   // port rule bypasses host firewalls like ufw. The app reaches it over the internal compose
-  // network (REDIS_URL=redis://redis:6379), so no host publish is needed. It must stay internal,
-  // exactly like the postgres service.
-  test("does not publish the session store to the host", () => {
-    const composeContents = readFileSync(dockerComposeTemplatePath, "utf8");
-    const redisBlock = getServiceBlock(composeContents, "redis");
+  // network (REDIS_URL=redis://redis:6379), so no host publish is needed. Assert on the *resolved*
+  // compose model — js-yaml expands anchors and merge keys — so a port smuggled in via an alias or
+  // a `<<` merge cannot slip past a raw-text check.
+  test("the redis service publishes no host port", () => {
+    const doc = load(readFileSync(dockerComposeTemplatePath, "utf8")) as {
+      services?: Record<string, { ports?: unknown }>;
+    };
+    const redis = doc.services?.redis;
 
-    expect(redisBlock).not.toMatch(/^\s*ports:/m);
-    expect(redisBlock).not.toContain("6379:6379");
-  });
-
-  test("keeps postgres internal too, as the reference pattern", () => {
-    const composeContents = readFileSync(dockerComposeTemplatePath, "utf8");
-    const postgresBlock = getServiceBlock(composeContents, "postgres");
-
-    expect(postgresBlock).not.toMatch(/^\s*ports:/m);
+    expect(redis, "the redis service is missing from docker-compose.yml").toBeTypeOf("object");
+    expect(
+      redis?.ports ?? [],
+      "the redis (Valkey) service must not publish any host port — it is reachable only on the internal compose network (ENG-2184)"
+    ).toEqual([]);
   });
 });
 
--- a/docker/__tests__/formbricks-script.test.ts
+++ b/docker/__tests__/formbricks-script.test.ts
@@ -71,6 +71,28 @@ describe("docker/docker-compose.yml Cube configuration", () => {
   });
 });
 
+describe("docker/docker-compose.yml Redis/Valkey exposure (ENG-2184)", () => {
+  // The bundled Valkey is Better Auth's session/token store (secondaryStorage). Publishing it to
+  // the host binds 0.0.0.0:6379 with no password, exposing every live session token — and Docker's
+  // port rule bypasses host firewalls like ufw. The app reaches it over the internal compose
+  // network (REDIS_URL=redis://redis:6379), so no host publish is needed. It must stay internal,
+  // exactly like the postgres service.
+  test("does not publish the session store to the host", () => {
+    const composeContents = readFileSync(dockerComposeTemplatePath, "utf8");
+    const redisBlock = getServiceBlock(composeContents, "redis");
+
+    expect(redisBlock).not.toMatch(/^\s*ports:/m);
+    expect(redisBlock).not.toContain("6379:6379");
+  });
+
+  test("keeps postgres internal too, as the reference pattern", () => {
+    const composeContents = readFileSync(dockerComposeTemplatePath, "utf8");
+    const postgresBlock = getServiceBlock(composeContents, "postgres");
+
+    expect(postgresBlock).not.toMatch(/^\s*ports:/m);
+  });
+});
+
 describe("docker/formbricks.sh Traefik label injection", () => {
   test("adds HTTPS Traefik labels to the formbricks service only", () => {
     const composePath = writeDockerComposeTemplate();
--- a/docs/self-hosting/setup/docker.mdx
+++ b/docs/self-hosting/setup/docker.mdx
@@ -618,11 +618,17 @@ Common checks:
 
   **If the port was reachable from an untrusted network, also invalidate existing sessions.** Valkey persists
   to its `redis` volume, so removing the port does not clear session or token state an attacker may already
-  have written — a forged session can remain valid afterwards. After closing the port, rotate `NEXTAUTH_SECRET`
-  (and `BETTER_AUTH_SECRET`, if you set it) and restart the stack: this invalidates every session cookie, so
-  all users must sign in again, while stored data, the job queue, and the cache stay intact. Leave
-  `ENCRYPTION_KEY` unchanged, and do **not** run a blanket `FLUSHDB` — the same Valkey also holds BullMQ job
-  queues and the application cache.
+  have written — a forged session can remain valid afterward. After closing the port, set a new
+  `NEXTAUTH_SECRET` in your `.env`, then recreate the app container so it picks up the change:
+  ```bash
+  docker compose up -d --force-recreate formbricks
+  ```
+  A plain `docker compose restart` keeps the container's original environment, so the rotated secret would not
+  take effect. Recreating invalidates every session cookie, so all users must sign in again, while stored
+  data, the job queue, and the cache stay intact. (If you added `BETTER_AUTH_SECRET` to the `formbricks`
+  service environment yourself, rotate that instead — Better Auth prefers it over `NEXTAUTH_SECRET`; the
+  bundled compose ships only `NEXTAUTH_SECRET`.) Leave `ENCRYPTION_KEY` unchanged, and do **not** run a blanket
+  `FLUSHDB` — the same Valkey also holds BullMQ job queues and the application cache.
 </Warning>
 
 <Note>
--- a/docs/self-hosting/setup/docker.mdx
+++ b/docs/self-hosting/setup/docker.mdx
@@ -588,10 +588,26 @@ Common checks:
 - **Hub auth errors**: Confirm the same `HUB_API_KEY` from `.env` is used by Formbricks Web and the Hub
   service.
 - **Port conflicts**: Confirm port `3000` and any optional service ports you enabled are not already in use on
-  the host. The bundled Valkey is not published to the host — it is reachable only on the internal compose
-  network — so it does not need port `6379` free on the host.
+  the host. The current template does not publish the bundled Valkey (`redis` service) to the host — it is
+  reachable only on the internal compose network — so a fresh install does not need port `6379` free on the
+  host.
 - **Optional Qwen/vLLM issues**: Check GPU availability, NVIDIA Container Toolkit installation, and
   `docker compose --profile qwen logs --tail=200 vllm`.
+
+<Warning>
+  **Existing installs: remove the bundled Valkey host port.** Older templates published the `redis`
+  (Valkey) service on `0.0.0.0:6379` with no password. That instance stores Better Auth sessions and
+  verification tokens, so an exposed port lets anyone who can reach it read live sessions and take over
+  accounts. Updating Formbricks (`docker compose pull` / `up -d`) does **not** rewrite your existing
+  `docker-compose.yml`, so the mapping stays until you remove it. Open your `docker-compose.yml`, delete the
+  ```yaml
+  ports:
+    - "6379:6379"
+  ```
+  lines from the `redis:` service, then run `docker compose up -d`. The app reaches Valkey over the internal
+  compose network, so nothing else needs to change. Only keep a publish if an external tool genuinely needs
+  it, and then bind it to loopback (`127.0.0.1:6379:6379`) rather than all interfaces.
+</Warning>
 
 <Note>
   To edit any of the available environment variables, check out our
--- a/docs/self-hosting/setup/docker.mdx
+++ b/docs/self-hosting/setup/docker.mdx
@@ -587,8 +587,9 @@ Common checks:
   exist, then inspect `docker compose logs cube`.
 - **Hub auth errors**: Confirm the same `HUB_API_KEY` from `.env` is used by Formbricks Web and the Hub
   service.
-- **Port conflicts**: Confirm ports `3000`, `6379`, and any optional service ports you enabled are not already
-  in use on the host.
+- **Port conflicts**: Confirm port `3000` and any optional service ports you enabled are not already in use on
+  the host. The bundled Valkey is not published to the host — it is reachable only on the internal compose
+  network — so it does not need port `6379` free on the host.
 - **Optional Qwen/vLLM issues**: Check GPU availability, NVIDIA Container Toolkit installation, and
   `docker compose --profile qwen logs --tail=200 vllm`.
 
--- a/docs/self-hosting/setup/docker.mdx
+++ b/docs/self-hosting/setup/docker.mdx
@@ -309,6 +309,14 @@ See our [migration guide](/self-hosting/advanced/migration) for version-specific
    docker compose up -d
    ```
 
+<Warning>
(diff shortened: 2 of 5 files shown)


docs/self-hosting/setup/docker.mdx
```diff
@@ -587,8 +587,9 @@ Common checks:
   exist, then inspect `docker compose logs cube`.
 - **Hub auth errors**: Confirm the same `HUB_API_KEY` from `.env` is used by Formbricks Web and the Hub
   service.
-- **Port conflicts**: Confirm ports `3000`, `6379`, and any optional service ports you enabled are not already
-  in use on the host.
+- **Port conflicts**: Confirm port `3000` and any optional service ports you enabled are not already in use on
+  the host. The bundled Valkey is not published to the host — it is reachable only on the internal compose
+  network — so it does not need port `6379` free on the host.
```
kiln-agent: Existing installs stay exposed, and this line now tells their operators otherwise.

The template fix only reaches installs created after it lands. `update_formbricks` (`docker/formbricks.sh:980`) is `pull` / `down` / `up -d` — it never re-fetches `docker-compose.yml`; the installer curls it once at install time (`formbricks.sh:569`), and `migrate-to-v4.sh`'s `add_redis_service` short-circuits on `has_service redis`. So every already-deployed instance keeps `ports: - "6379:6379"` after updating and keeps binding `0.0.0.0:6379` unauthenticated.

That makes this new sentence read as a false all-clear for exactly the population that is still vulnerable. Worth either a remediation note here ("if your `docker-compose.yml` predates X, delete the `ports:` block under `redis:` and `docker compose up -d`") or a patch step in the update path that strips the mapping.

---

docker/__tests__/formbricks-script.test.ts
```diff
@@ -71,6 +71,28 @@ describe("docker/docker-compose.yml Cube configuration", () => {
...
+    expect(redisBlock).not.toMatch(/^\s*ports:/m);
+    expect(redisBlock).not.toContain("6379:6379");
+  });
+
+  test("keeps postgres internal too, as the reference pattern", () => {
+    const composeContents = readFileSync(dockerComposeTemplatePath, "utf8");
+    const postgresBlock = getServiceBlock(composeContents, "postgres");
+
+    expect(postgresBlock).not.toMatch(/^\s*ports:/m);
+  });
```
kiln-agent: Nit: this postgres guard covers a service the diff does not touch, and it is aimed at the wrong block.

`getServiceBlock` bounds a block by the next `^ {2}[A-Za-z0-9_-]+:` line. The four new comment lines you added sit at 2-space indent above `  redis:` and start with `#`, so they do not terminate the block — they land inside the *postgres* slice. It passes today, but if that comment ever grows the `127.0.0.1:6379:6379` opt-in snippet the PR description suggests, this test fails with a message about postgres while the redis assertions stay green. Confusing failure for a service nobody changed.

Also, in the test above, `not.toContain("6379:6379")` is subsumed by `not.toMatch(/^\s*ports:/m)` — one assertion carries the invariant.

Both are optional; the 7 tests pass as written.

---

kiln-agent (commented):
## Review

The compose change is correct and safe to land. I verified nothing in the stack needs host access to Valkey:

- `REDIS_URL: redis://redis:6379` comes from the compose `environment` anchor, which overrides any `.env` value, so the app always reaches it over the internal network.
- Cube runs `CUBEJS_CACHE_AND_QUEUE_DRIVER: memory` — it never touches Valkey.
- Neither `formbricks.sh` nor `migrate-to-v4.sh` references the `6379:6379` mapping, and `migrate-to-v4.sh`'s `add_redis_service` already injects the service without `ports:`, so the template is now consistent with it.
- CI's Redis is a GitHub Actions service container, not this compose service.
- `docker-compose.dev.yml` keeps its own publish, correctly — that one is host-run dev.

`getServiceBlock` bounds the redis block correctly (the services-level `redis:` at line 261 precedes the top-level volume entry, and the `depends_on` entry is indented 6 spaces), and all 7 tests in `docker/__tests__/formbricks-script.test.ts` pass locally.

**One thing to resolve before this counts as fixed:** the template change does not remediate any existing deployment. Details inline on `docker.mdx` — the update path never re-fetches `docker-compose.yml`, so deployed instances keep the published port while the new troubleshooting line tells their operators Valkey is not published. The second comment is an optional test nit.

---

kiln-agent (commented):
## Review

The core change is correct and well-scoped. I verified the removal of `ports: - "6379:6379"` breaks nothing in the self-hosting stack:

- The only consumer of the bundled Valkey is the `formbricks` service via `REDIS_URL: redis://redis:6379` (`docker/docker-compose.yml:30`) — service DNS on the internal network, unaffected.
- The healthcheck (`valkey-cli ping`) runs inside the container.
- `hub` and `cube` carry no Redis config, so nothing else in the file needs the port.
- `postgres` in the same file publishes no host port, so the two stateful services are now consistent; `migrate-to-v4.sh:828` injects redis without `ports:` already, and the Helm chart uses `ClusterIP` + `--requirepass`.
- `docker-compose.dev.yml:79` and `.github/workflows/docker-build-validation.yml:40` keep their own published Redis — correct, the app runs on the host there.
- No doc instructs reaching the bundled Valkey from the host: `cluster-setup.mdx:124` and `environment-variables.mdx:117` describe an *external* Redis, and `job-runner.mdx` already uses `redis://redis:6379`.

`docker compose -f docker/docker-compose.yml config -q` validates, and all 7 tests in `docker/__tests__/formbricks-script.test.ts` pass (`cd docker && pnpm test`).

The threat model in the description also checks out — `apps/web/modules/auth/lib/auth.ts:96` wires this instance as better-auth's `secondaryStorage`, and the file's own comment (lines 69–72) confirms sessions and verification tokens live there.

### Existing installs stay exposed after upgrading, with no signal

One gap worth deciding on before merge, since this is a security fix rather than a config tidy-up: the change only protects **new** installs.

`update_formbricks()` (`docker/formbricks.sh:980`) runs `docker compose pull` → `down` → `up -d` and never rewrites the local `docker-compose.yml` that `install_formbricks` curled from the old template. `migrate-to-v4.sh:822` also returns early on `has_service redis` without stripping a published port. So every instance installed from the previous template keeps `0.0.0.0:6379` published across the upgrade, and nothing in this diff or the docs tells the operator to remove the block.

Two options, either fine:

1. Minimal: add a note to the Docker self-hosting docs (and the release notes) telling existing self-hosters to delete the `ports: - "6379:6379"` mapping from their `docker-compose.yml` and re-run `docker compose up -d`.
2. Thorough: have `update_formbricks()` (or a small migration step) strip a `6379` publish from the `redis` service block in place — the same awk-surgery pattern the script already uses for volumes and `depends_on`.

Without one of these, the advisory closes for fresh installs while the exposed population stays exposed and unaware.

### Minor

- The `keeps postgres internal too, as the reference pattern` test asserts on config this PR does not change. It is defensible (publishing postgres would be its own exposure), but it does lock unrelated config into this PR's test block — worth a thought against the "no low-signal tests" guidance in `AGENTS.md`. Not blocking.
- `expect(redisBlock).not.toMatch(/^\s*ports:/m)` guards the template only. A publish reintroduced through the `formbricks.sh` injection path would not be caught — fine as a scope note, not a request.

Nothing else blocking from my side; the diff itself is right.

---

mattinannt:
Thanks for the reviews — addressed in `230bced`:

- **Existing installs stay exposed after upgrading** (both reviews): correct and important. `update_formbricks()` (`docker/formbricks.sh:980`) only runs `pull` / `down` / `up -d` and never re-fetches the template, so deployed instances keep the `6379` publish, and my new troubleshooting line read as a false all-clear for exactly that population. Fixed the wording and added a `<Warning>` in `docs/self-hosting/setup/docker.mdx` telling existing self-hosters to delete the `ports: - "6379:6379"` block from their `redis:` service and re-run `docker compose up -d` (with the loopback-bind note if they genuinely need host access). I've left the *thorough* option — auto-stripping the mapping in the update path via the same awk surgery as `cleanup_rustfs_init()` — as an open follow-up rather than expanding this PR; noted under Open gaps.
- **Test robustness against merged/aliased ports** (CodeRabbit): the check now parses the compose file with `js-yaml` and asserts `services.redis.ports` is absent/empty on the *resolved* model, so a port smuggled in via an anchor or `<<` merge can't slip past a raw-text scan.
- **`postgres` guard / redundant assertion** (both): dropped the postgres case (it asserted on config this PR doesn't touch) and collapsed to the single resolved-model assertion.

Breaking-changes section now uses the table (the `breaking-change` label is on), and the PR body's non-template footer is removed.

---

docs/self-hosting/setup/docker.mdx
```diff
@@ -587,11 +587,28 @@ Common checks:
...
+  verification tokens, so an exposed port lets anyone who can reach it read live sessions and take over
+  accounts. Updating Formbricks (`docker compose pull` / `up -d`) does **not** rewrite your existing
+  `docker-compose.yml`, so the mapping stays until you remove it. Open your `docker-compose.yml`, delete the
+  ```yaml
+  ports:
+    - "6379:6379"
+  ```
+  lines from the `redis:` service, then run `docker compose up -d`. The app reaches Valkey over the internal
+  compose network, so nothing else needs to change. Only keep a publish if an external tool genuinely needs
+  it, and then bind it to loopback (`127.0.0.1:6379:6379`) rather than all interfaces.
```
BhagyaAmarasinghe: Removing the port and recreating Valkey does not clear AOF-backed data in the named volume, so session state forged while port 6379 was exposed can remain authoritative. Replace the claim that nothing else needs to change with a maintainer-approved session and cookie invalidation procedure; avoid a blanket FLUSHDB unless its queue and cache impact is explicitly accepted.

docs/self-hosting/setup/docker.mdx
```diff
@@ -587,11 +587,28 @@ Common checks:
...
-- **Port conflicts**: Confirm ports `3000`, `6379`, and any optional service ports you enabled are not already
-  in use on the host.
+- **Port conflicts**: Confirm port `3000` and any optional service ports you enabled are not already in use on
+  the host. The current template does not publish the bundled Valkey (`redis` service) to the host — it is
+  reachable only on the internal compose network — so a fresh install does not need port `6379` free on the
+  host.
 - **Optional Qwen/vLLM issues**: Check GPU availability, NVIDIA Container Toolkit installation, and
   `docker compose --profile qwen logs --tail=200 vllm`.
 
+<Warning>
```
BhagyaAmarasinghe, line 605: This warning lives only in the manual Docker guide, while production one-click users run a local updater that only pulls images and recreates the unchanged Compose file. Add the same Valkey removal and session-invalidation guidance to the One-Click Setup Update section and coordinated release advisory so existing installations do not keep publishing port 6379.

BhagyaAmarasinghe (commented):
Reviewed at exact head 230bceda3c4dfd467c5f3658f666c73bd1015c64. Two security remediation gaps remain.

docs/self-hosting/setup/docker.mdx
mattinannt: Good catch — fixed in `bc39e98`. Removed the "nothing else needs to change" all-clear and added a remediation step: since Valkey persists to its named volume (AOF), closing the port does not clear a session an attacker may have forged, so after removing the mapping the operator should **rotate `NEXTAUTH_SECRET` (and `BETTER_AUTH_SECRET` if set) and restart**. That invalidates every session cookie's signature — the forged cookie no longer verifies and the stale cached blob becomes unreachable — while leaving stored data, the BullMQ queues, and the app cache intact. I explicitly warned **against** a blanket `FLUSHDB` (it would drop the queues + `fb:`-namespaced cache) and against rotating `ENCRYPTION_KEY`.

I went with secret rotation rather than a targeted key delete because the better-auth session keys are the bare session tokens (no shared prefix), so there's no safe scriptable `SCAN`/`DEL` pattern that wouldn't risk the co-located queue/cache keys. Happy to adjust the exact wording if you'd prefer a different maintainer-approved procedure.

---

docs/self-hosting/setup/docker.mdx
mattinannt: Addressed in `bc39e98`. Added a security `<Warning>` to the **Update** section (the `pull` / `down` / `up -d` steps that the one-click updater also runs), stating that neither the manual steps nor the one-click updater rewrite an existing `docker-compose.yml`, so a pre-existing `6379:6379` publish stays exposed after updating — and pointing to the Debug section for removal + session invalidation. So the guidance is now on the update path itself, not only in the manual troubleshooting note.

Two parts I did **not** put in this PR, since they're maintainer/product calls rather than docs edits in this file — flagging for your decision:
- Propagating the same callout into the dedicated **one-click setup doc** (`/self-hosting/setup/one-click`).
- A **coordinated release advisory** (release notes / security advisory) for already-deployed installs.

I can take on either if you'd like — say the word and I'll extend this PR or open a follow-up.

---

BhagyaAmarasinghe (dismissed):
Thank you for the PR, LGTM!

docs/self-hosting/setup/docker.mdx
```diff
@@ -587,11 +595,36 @@ Common checks:
...
+  compose network, so no other configuration change is needed. Only keep a publish if an external tool
+  genuinely needs it, and then bind it to loopback (`127.0.0.1:6379:6379`) rather than all interfaces.
+
+  **If the port was reachable from an untrusted network, also invalidate existing sessions.** Valkey persists
+  to its `redis` volume, so removing the port does not clear session or token state an attacker may already
+  have written — a forged session can remain valid afterwards. After closing the port, rotate `NEXTAUTH_SECRET`
+  (and `BETTER_AUTH_SECRET`, if you set it) and restart the stack: this invalidates every session cookie, so
+  all users must sign in again, while stored data, the job queue, and the cache stay intact. Leave
+  `ENCRYPTION_KEY` unchanged, and do **not** run a blanket `FLUSHDB` — the same Valkey also holds BullMQ job
+  queues and the application cache.
```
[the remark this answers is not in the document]
mattinannt: Fixed in `8ad5f2a` — this was a real gap, thanks. The guidance now says to set the new secret in `.env` and run `docker compose up -d --force-recreate formbricks` (recreating so the new value is applied), and explicitly notes that a plain `docker compose restart` keeps the container's original environment and would not take effect.

On the secret name: the bundled `docker/docker-compose.yml` wires only `NEXTAUTH_SECRET` into the `formbricks` service (`app-environment`), and `auth.ts` uses `BETTER_AUTH_SECRET ?? NEXTAUTH_SECRET` — so for the bundled stack the effective secret is `NEXTAUTH_SECRET`, and that's what the docs now tell operators to rotate. I kept it to a note that `BETTER_AUTH_SECRET` applies only if the operator added it to the service environment themselves, rather than adding `BETTER_AUTH_SECRET` to the bundled compose in this PR — wiring a new secret into the template is a separate change beyond this security fix's scope.

---

BhagyaAmarasinghe (approved):

PR #8923, opened 2026-08-19
fix(a11y): title public survey pages per step and keep instructions reachable [ENG-2334]

itsjavi:
Fixes ENG-2334

## What & why

Two halves of one VPAT finding, *Partially Supports* against WCAG 2.0 SC 2.4.2 at **Level A** — the lowest bar, and the most damaging kind of gap in a procurement review:

> "Input fields have labels, but forms themselves have no titles. Survey instructions are not persistent (not displayed on each page)."

**The form had no accessible name.** Every input had a label, but the form they belonged to was an anonymous `<div>`. `#fbjs` is now `role="form"` named by the survey. This is the only name an **embedded** survey can carry — it is rendered into someone else's document and cannot own its `<title>`. A modal survey's dialog gets the survey name too, instead of the generic `"Survey Dialog"`, which stays as the fallback for a survey rendered without a name.

**The document title never changed as the respondent moved through the survey.** `generateMetadata` cannot fix that — the current card lives in the renderer's Preact state and the server never sees it. `Survey` now reports its position through a new `onPageChange` callback, and the link-survey host appends it:

```
before:  Page title a11y survey | Formbricks          ← identical on every card
after:   Page title a11y survey | Formbricks — Page 1 of 4
         Page title a11y survey | Formbricks — Page 2 of 4
         Page title a11y survey | Formbricks — Page 4 of 4   ← ending card
```

Two deliberate choices there:

- **The base title is the server-rendered one, captured once on mount** — not recomputed. So an author's custom link-metadata title and the root layout's `"| Formbricks"` template both survive, without reimplementing `getBasicSurveyMetadata`'s priority chain (link metadata → welcome headline → survey name) on the client, where `survey.metadata` is not even available.
- **The step label is localized inside the surveys package**, then handed over pre-formatted. The survey knows its own active language; the host's i18n runs in the *viewer's UI locale*. Assembling the title host-side would mix the two — a German survey viewed by an English-locale browser would read "Umfrage — Page 2 of 4". Verified: switching to German gives `"Seite 1 von 4"`, Japanese `"4ページ中1ページ目"`.

**The instructions were unreachable after one click.** The welcome card's subheader is the only instructions text a survey has (`welcomeCard.html` was migrated to `subheader`, and there is no `survey.description` — `survey.metadata.description` is meta-tag-only). The moment the respondent advanced, it was gone with no way back. It is now rendered once as a visually-hidden region that the form's `aria-describedby` points at, so it is announced on entry to every page.

*Once*, not per card: `getCardContent` runs for each of the stacked layout's peeking cards and would have emitted a duplicate id, breaking the very reference the region exists to serve. `aria-describedby` is also omitted entirely when the survey has no instructions, rather than left dangling.

`getSurveyPagePosition` counts cards the way `progress-bar.tsx` already does — welcome when enabled, one per block, one for the ending — so the two can never disagree about how long the survey is. It clamps rather than reporting "page 0 of 4" for a block id that no longer resolves after a mid-session edit.

### Based on #8920, plus the Sonar fixes

**Base branch.** This PR now targets `claude/…-eng-2335` (#8920) rather than `main`, so the props that
overlap between the two meet here instead of after #8920 lands. Two side effects worth knowing: the
diff below is ENG-2334's changes only, and CodeRabbit's auto-review is off on this PR — `.coderabbit.yaml`
only auto-reviews bases matching `^epic/`, so a re-review needs `@coderabbitai review`.

`dc9f1fa` merges #8920 in. Additive in the same two files: `SurveyContainer` carries `surveyName`,
`hasInstructions` and `lang` together; the inline `#fbjs` root carries `dir`, `lang`, `role="form"`,
`aria-label` and `aria-describedby` and renders the sr-only `h1` above `children`; the modal root keeps
`dir` and `lang` while the dialog element keeps its own name and description. Earlier, `a95b26e` merged
`main` (ENG-2336, #8919) — where the interesting part was that **both sides had introduced a
`surveyName` prop for different reasons**: #8919 as the survey's one top-level heading (2.4.6), this
branch as the form's accessible name (2.4.2). Same string, so the two declarations collapsed into one
whose doc comment records both jobs.

Then the SonarQube findings, each in its own commit:

- **`59025e6` — `typescript:S6819`**, which asks for a native `<form>` instead of `role="form"`.
  Suppressed with the reason in code, the same call `progress.tsx` and `auto-close-progress-bar.tsx`
  already make for this rule in this package. Neither thing a real form brings is safe here: an inline
  survey renders into a container the host supplies by id (`packages/surveys/src/index.ts`), which can
  sit inside *their* `<form>` — nested forms are invalid HTML and the browser drops the inner one,
  silently taking this accessible name with it — and a real form makes Enter in any text input submit
  and navigate away from a half-finished survey.
- **`1634d30` — `typescript:S3776`**, cognitive complexity 18 against a limit of 15.

  Two earlier attempts (`62bd3e1`, `d21e0c5`) did **not** fix this, and the reason is worth recording
  because it is easy to get wrong: SonarQube scores each function separately, so a conditional inside a
  nested arrow — a `useEffect` callback, or `getPlacementStyle` — never counted against
  `SurveyContainer` at all. Not one of the 17 flagged locations was in the code those commits moved.
  Every point is in the render, and six of them were two class computations, now module-scope
  `getModalLayerClass` and `getOverlayBackdropClass`. That should land the component at 12.

  Those two commits stay, on their own merits rather than as the fix: hoisting `getPlacementStyle`
  stops it being rebuilt every render, and `useNoOverlayModal` keeps two unrelated concerns out of the
  component. `d21e0c5` exists because `62bd3e1` first put that hook in `lib/use-no-overlay-modal.ts`,
  which **failed the Sonar gate at 47.7% coverage on new code** — `**/*.tsx` is coverage-excluded, `.ts`
  is not, so a 22-line hook with no unit tests became 22 uncovered new lines. Moving it to module scope
  in the `.tsx` restored coverage to 100% and cost nothing, since the rule does not care which file a
  function sits in.

### Screenshots

Nothing about this is visible: the instructions region is `sr-only`, `role`/`aria-*` do not paint, and the title lives in browser chrome that a headless screenshot does not capture. So the screenshots are here as **evidence of no visual regression**, and the observable behaviour is quoted as text above and below.

| Before | After |
| --- | --- |
|  |  |
|  |  |

Pixel-diffed the PNG originals against `main`: **0 of 924 000 pixels differ** on both frames. The uploaded JPEGs are byte-identical too (`f38705cf…` for the question card, `1e005f3f…` for the welcome card).

What actually changed, read out of the real built bundle at runtime:

```
#fbjs  role="form"
       aria-label="Q3 Customer Satisfaction"
       aria-describedby="fb__survey-instructions"     ← absent when there are no instructions

#fb__survey-instructions   1 node, sr-only
  en-US → "It only takes a minute — your answers help us decide what to build next."
  de-DE → "Es dauert nur eine Minute."

onPageChange  en-US → {index: 1, total: 4, label: "Page 1 of 4"} → {index: 2, …, "Page 2 of 4"}
              de-DE → {index: 1, total: 4, label: "Seite 1 von 4"} → {index: 2, …, "Seite 2 von 4"}
              welcome card disabled → total 3, not 4
```

## Breaking changes

** `onPageChange` is additive only — no action required for anyone.**. Consumers calling `window.formbricksSurveys.renderSurvey(...)` won't be affected.

## Migrations & env

- none

One new i18n string, `common.page_x_of_y` (`"Page {current} of {total}"`), added to `packages/surveys/locales/en-US.json` and fanned out to all 22 target locales with `pnpm i18n`, which also rebuilt `packages/surveys/i18n.lock`. The non-English files were not hand-edited.

## How this was tested

**Coverage**

Every mutation below was actually run, and the row records what broke. **Line numbers are against head
`dc9f1fa`**; neither file the mutations target was touched by the merge or the Sonar commits, so they
still point at the code quoted. Reruns:
`cd packages/surveys && npx vitest run --project surveys-unit src/lib/survey-page.test.ts` (baseline `17 passed`)
`cd apps/web && npx vitest run --project unit modules/survey/link/lib/document-title.test.ts` (baseline `7 passed`)

| Behaviour | How | Outcome |
| --- | --- | --- |
| `getSurveyPagePosition` counts cards the way the progress bar does: welcome on/off, multiple endings as one card, no ending | `unit (mutation)`<br>`packages/surveys/src/lib/survey-page.test.ts` → `describe("getSurveyPagePosition")` | 12 cases. Mutated `packages/surveys/src/lib/survey-page.ts:29` (`welcomeOffset` → `0`) → **5 failed / 12 passed** |
| `getSurveyPagePosition` clamps instead of reporting "page 1 of 0" or a negative index | `unit (mutation)`<br>`packages/surveys/src/lib/survey-page.test.ts` → `describe("getSurveyPagePosition")` | Same file. Mutated `packages/surveys/src/lib/survey-page.ts:32` (dropped `Math.max(1, …)`) → **1 failed / 16 passed**. `findIndex` returning `-1` for a block deleted mid-session is covered separately by the "unresolvable block id" case |
| `hasSurveyInstructions` ignores blank text and finds translation-only content | `unit (mutation)`<br>`packages/surveys/src/lib/survey-page.test.ts` → `describe("hasSurveyInstructions")` | 5 cases in the same file. Mutated `packages/surveys/src/lib/survey-page.ts:55` (`value.trim().length > 0` → `value !== undefined`) → **1 failed / 16 passed**, i.e. the whitespace-only subheader case |
| `buildSurveyDocumentTitle` never emits a dangling separator | `unit (mutation)`<br>`apps/web/modules/survey/link/lib/document-title.test.ts` → `describe("buildSurveyDocumentTitle")` | 7 cases. Mutated `apps/web/modules/survey/link/lib/document-title.ts:20` (`!suffix` → `suffix === undefined`) → **1 failed / 6 passed**, i.e. the blank-step case that would have produced `"Survey — "` |
| Title names the survey **and** the current page, and follows navigation | `e2e`<br>`apps/web/playwright/survey-page-title.spec.ts` → "the title names the survey and the current page, and follows navigation" | Walks a 3-block survey and asserts the title at each step, ending exactly on `Page 4 of 4` rather than overshooting. Kept as e2e because this is the one behaviour that only exists across the Preact bundle → `onPageChange` → the React host's `document.title`, and axe cannot judge title descriptiveness at all |
| Form landmark is named, and describes itself with the instructions | `e2e`<br>`apps/web/playwright/survey-page-title.spec.ts` → "the survey exposes a named form landmark with its instructions" | `role="form"`, `aria-label`, `aria-describedby` on `#fbjs` |
| Instructions persist past the welcome card, and stay singular | `e2e`<br>`apps/web/playwright/survey-page-title.spec.ts` → "the survey exposes a named form landmark with its instructions" | Asserted after advancing, with `toHaveCount(1)` — the stacked layout keeps several cards mounted, so a duplicated id is the specific way this breaks |
| **The modal dialog's own accessible name and description** — the surface the inline e2e spec does not reach | `manual` | Read the dialog's a11y attributes in the built bundle on **both** overlay settings: `aria-label` is the survey name (`"Q3 Customer Satisfaction"`, not the generic `"Survey Dialog"`) and `aria-describedby` is `fb__survey-instructions`, with exactly one instructions region and one `h1` |
| Both modal paths still behave the same after the effects moved into a hook | `manual` | The one risky part of the S3776 work, so it was driven in the built bundle rather than reasoned about — and re-run from a fresh bundle after `d21e0c5` moved the hook, with identical results. **No overlay**: `aria-modal` absent, the live region receives "A survey has opened at the end of the page.", and Escape closes the survey. **Dark overlay**: `aria-modal="true"`, *nothing* announced (the focus trap moving focus is the announcement), Escape still closes via the trap |
| The class refactor changes nothing that renders | `manual` | The S3776 fix rewrites two `className` expressions, so it was proven rather than eyeballed: dumped the `class` attribute of **every** node under `#fbjs` from the built bundle across inline, modal/none, modal/dark and modal/light — 44/44/44/40 elements — before and after. Byte-identical in all four |
| All three a11y fixes coexist on this branch | `manual` | With #8920 merged in, read the whole contract from one fresh bundle across inline, `?lang=de-DE`, modal/no-overlay and modal/dark: exactly one `h1` and no `h3`–`h6` (2.4.6), `lang` tracking the active language with `dir` agreeing (3.1.1), `role="form"` + name + description on the inline root and the dialog named on the modal paths with exactly one instructions region (2.4.2), and `onPageChange` reporting `"Seite 1 von 4"` under `de-DE` |
| Localized step label and instructions in the real bundle | `manual` | Built `surveys.umd.cjs`, rendered with `onPageChange` attached, read the callback payload and DOM: `"Page 1 of 4"` → `"Seite 1 von 4"` on `?lang=de-DE`, instructions text switching with it, `aria-describedby` absent when the survey has no subheader |
| No visual regression | `manual` | Pixel-identical to `main` (0/924 000 px) on the welcome and question cards; screenshots above |
| Monorepo still builds and passes, after each merge and each Sonar commit | `manual` | Each time: `pnpm typecheck` (27 tasks), `pnpm lint` (0 errors, 40 warnings — same count as `main`), `pnpm test` (628 files / 8296 tests), `pnpm format:check`, `pnpm scan-translations` — all clean |

**Open gaps**

- **The `document.title` restore on unmount is not exercised.** The e2e spec walks the survey and asserts the title at each step, but never navigates away, so the cleanup that puts the server-rendered title back is reasoned about rather than tested. It matters on a client-side navigation out of a survey; the `lang`/`dir` effect it mirrors in the same component has the same shape and the same gap. Worth an assertion if this file is touched again.
- **The base title does not re-localize on a mid-survey language switch.** It is captured once from the server-rendered title, so on `?lang=de` the base is already German, but switching language *during* the survey moves only the step label. Fixing it properly means exposing `survey.metadata` to the client and duplicating the server's title-priority chain; that felt worse than the partial. The ticket lists this under Risks rather than success criteria. Called out because it is a real, deliberate limitation.
- **Instructions are visually hidden**, so a sighted respondent still cannot get back to them. This is the design call the ticket flags as needing Kristian / pain.design: a visible persistent block conflicts with the minimal single-question card. This ships the accessibility outcome without pre-empting that decision.
- **"Page n of m" is a position, not a path.** Branching logic means a respondent may skip blocks, so the count can imply more pages than they will actually see. That is the same approximation the progress bar already makes, and making it exact would require knowing the future.
- No screen reader was driven. The ticket asks for a manual NVDA/VoiceOver pass on a 3+ page survey and an IBM Equal Access + ANDI re-run; neither is available in this environment.
- The PIN gate and email-verification gate render before `SurveyClientWrapper`, so they keep the plain server title. Correct — there is no survey position yet — but worth knowing.
- **`useNoOverlayModal` has no unit test.** Its behaviour is entirely DOM listeners and a timer, so the honest coverage is the e2e keyboard suite plus the bundle run above rather than a mocked-DOM unit test — which is also why it is colocated in the `.tsx` rather than promoted to `lib/`. Worth knowing that the Escape path it owns is asserted only end-to-end.
- **`Shift+Escape` still closes the survey**, in both the no-overlay handler here and the overlay one in `use-focus-trap.ts:273` — both exclude Alt, Ctrl and Meta but not Shift. Raised in review and deliberately not fixed here: patching one side alone would make the two paths disagree, and the shared focus-trap module is out of scope for this PR. Follow-up, ideally hoisting the modifier check so there is one definition.

**Risks**

- `document.title` is now mutated client-side, the first place in the repo that does so. It is restored on unmount, mirroring the existing `lang`/`dir` effect in the same component, so a client-side navigation away cannot leave a stale title. The embedded widget cannot reach this component (`survey-client-wrapper.tsx` is link-survey only), so no host page is affected.
- `role="form"` adds a landmark to the survey. On a link survey it sits inside the shell's existing `<main>`; nesting a form landmark in main is valid and expected, but it does change what a screen reader's landmark list shows.
- **Escape handling and the open announcement now live in `useNoOverlayModal`.** Behaviour-preserving and verified on both modal paths above, but it moved code that landed in #8919's neighbourhood rather than in this PR's. It is in its own commits precisely so it can be dropped without touching the merge or the feature if a reviewer would rather keep those effects inline.
- The instructions region uses a fixed id (`fb__survey-instructions`). Two surveys rendered into the same document — possible in the editor preview alongside a live widget — would collide. Not reachable today (the preview and the widget are never both mounted on a respondent-facing page), but it is the constraint to remember if that changes.

> [!NOTE]


packages/surveys/src/lib/survey-page.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Whether the survey has instructions worth exposing on every page.
+ *
+ * The welcome card's subheader is the only instructions text a survey has — there is no
+ * `survey.description` — and today it disappears the moment the respondent advances past the
+ * welcome card. Checks every localization rather than just `default`, because a survey may leave
+ * the default empty and fill only its translations.
+ */
+export const hasSurveyInstructions = (survey: TJsWorkspaceStateSurvey): boolean =>
+  Object.values(survey.welcomeCard.subheader ?? {}).some((value) => value.trim().length > 0);
```
kiln-agent: **`hasSurveyInstructions` ignores `welcomeCard.enabled`, so a hidden welcome card's leftover subheader is exposed on every page.**

Toggling the welcome card off in the editor only writes `enabled: false` (`edit-welcome-card.tsx:110` — `updateSurvey({ enabled: newEnabledState })`); `headline`/`subheader` stay in the JSON. So for any survey where an author once wrote a welcome subheader and later hid the welcome card, this returns `true`, `SurveyContainer` sets `aria-describedby="fb__survey-instructions"`, and `survey.tsx` renders that text into the `sr-only` region on every card.

Concrete scenario: author enables the welcome card, writes "Answer honestly, there are no wrong answers.", later disables the welcome card and folds the intro into the first question. Sighted respondents never see the old text; every screen-reader respondent now gets it as the form's description for the whole survey. The reasoning in the test ("The instructions are exposed on every page, so they outlive the card that authored them") only holds for a card the respondent actually saw — with `enabled: false` nobody ever read it, so it is stale content, not instructions.

```suggestion
export const hasSurveyInstructions = (survey: TJsWorkspaceStateSurvey): boolean =>
  survey.welcomeCard.enabled &&
  Object.values(survey.welcomeCard.subheader ?? {}).some((value) => value.trim().length > 0);
```

---

packages/surveys/src/lib/survey-page.ts
```diff
@@ -0,0 +1,55 @@
...
+ * answering, and an ending id at the end.
+ */
+export const getSurveyPagePosition = (
+  survey: TJsWorkspaceStateSurvey,
+  blockId: string
+): TSurveyPagePosition => {
+  const welcomeOffset = survey.welcomeCard.enabled ? 1 : 0;
+  const hasEnding = survey.endings.length > 0;
+  // A survey always has at least one block, but clamp anyway rather than report "page 1 of 0".
+  const total = Math.max(1, welcomeOffset + survey.blocks.length + (hasEnding ? 1 : 0));
```
kiln-agent: **`total` does not match the progress bar, contradicting the doc comment above ("Counts cards the way the progress bar already does ... so the two can never disagree").**

`ProgressBar`/`calculateProgress` (`progress-bar.tsx:20-27`) computes `totalCards = survey.blocks.length` and adds 1 only for the ending — the welcome card is **not** counted, and the welcome card is reported as progress `0`. This function adds `welcomeOffset` to `total` and shifts every block index by it.

Concrete scenario: welcome card enabled, 3 blocks, 1 ending. On the first question the title says "Page 2 of 5" (40%) while the `role="progressbar"` next to it reports `aria-valuenow=25` (1/4). On the last question: "Page 4 of 5" vs. 75%. A screen-reader user hears both numbers and they never agree; the last question is also never "Page 4 of 4"-equivalent to the bar's 3/4.

Either drop `welcomeOffset` from `total` and treat `"start"` as page 0/`total`, or fix `calculateProgress` to include the welcome card — but the two must be derived from one helper if the comment's invariant is to hold.

---

packages/surveys/src/components/wrappers/survey-container.tsx
```diff
@@ -158,7 +181,10 @@ export function SurveyContainer({
             // assistive tech ignore everything outside it, so setting it on a corner survey hides the
             // host page from screen-reader users while they can still see and use it.
             aria-modal={hasOverlay ? "true" : undefined}
-            aria-label={t("common.survey_dialog")}
+            // The survey name is strictly more informative than the generic "Survey Dialog", which
+            // stays as the fallback for a survey rendered without one (previews).
+            aria-label={surveyName ?? t("common.survey_dialog")}
```
kiln-agent: **`??` does not catch an empty survey name, so the dialog can end up with no accessible name at all — a regression from the previous unconditional `"Survey Dialog"`.**

`surveyName` is always a `string` here (`RenderSurvey` passes `props.survey.name`, and `TJsWorkspaceStateSurvey.name` is required), so `??` can only ever fall through when the prop is omitted entirely. `ZSurveyBase.name` / `ZSurveyCreateInput.name` are plain `z.string()` with no `.min(1)`, so a survey created through the management API with `"name": ""` is valid.

Concrete scenario: `POST /api/v1/management/surveys` with `name: ""`, type `app`, overlay set. The modal renders `aria-label=""` on `role="dialog"`, i.e. an unnamed dialog — worse than before this change, and an axe `aria-dialog-name` failure. `||` (or a `.trim()` check) fixes it, and would match the truthiness guard already used on line 153 for `role`.

```suggestion
            aria-label={surveyName || t("common.survey_dialog")}
```

---

packages/surveys/src/components/general/survey.tsx
```diff
@@ -1328,6 +1348,28 @@ export function Survey({
...
+  // Survey instructions, available on every page. They live on the welcome card, which the
+  // respondent leaves after one click and can never get back to — so the text that explains what the
+  // survey is for was gone for the rest of it (the second half of the 2.4.2 VPAT finding).
+  //
+  // Rendered once here rather than inside getCardContent, which runs per card and would emit a
+  // duplicate id for every peeking card in the stacked layout. Visually hidden: the card designs
+  // have no room for a persistent block, and a sighted respondent has already read it on the
+  // welcome card. SurveyContainer points the form's aria-describedby at this id, so a screen reader
+  // announces it on entry to each page.
+  const instructions = hasSurveyInstructions(localSurvey) ? (
```
kiln-agent: **`hasSurveyInstructions` and the rendered text can disagree, leaving `aria-describedby` pointing at an empty element.**

`hasSurveyInstructions` returns `true` if *any* localization is non-blank, but the rendered value comes from `getLocalizedValue(subheader, selectedLanguage)`, which returns `value[selectedLanguage]` only when non-blank and otherwise falls back to `value.default` — with no further fallback (`i18n.ts:32-37`). `Subheader` returns `null` for an empty string.

Concrete scenario: the exact case the new unit test blesses — `subheader = { default: "", "de-DE": "Kurz." }` (test: "is true when only a translation is filled in"). A respondent on the survey's default language gets `hasInstructions === true`, so `#fbjs` carries `aria-describedby="fb__survey-instructions"`, but `<div id="fb__survey-instructions">` renders empty. The form landmark gains a description that says nothing, and the "instructions persist on every page" fix silently does nothing for that respondent.

Deciding `hasInstructions` from the same resolved-and-localized string the region renders (rather than from raw `Object.values`) would keep the two in lockstep.

---

packages/surveys/src/components/wrappers/survey-container.tsx
```diff
@@ -127,9 +134,25 @@ export function SurveyContainer({
...
     return (
-      <div id="fbjs" className="formbricks-form" style={{ height: "100%", width: "100%" }} dir={dir}>
+      <div
+        id="fbjs"
+        className="formbricks-form"
+        style={{ height: "100%", width: "100%" }}
+        dir={dir}
+        role={surveyName ? "form" : undefined}
+        aria-label={surveyName}
+        aria-describedby={instructionsId}>
```
kiln-agent: **`aria-describedby` on a non-focusable landmark is not announced, so the "announced on entry to each page" mechanism does not hold.**

`#fbjs` is a plain `div` with `role="form"` and no `tabindex`. Screen readers announce `aria-describedby` reliably only when the described element receives focus (or is an interactive/labelled widget); NVDA, JAWS and VoiceOver do not read a landmark's description when the reading cursor enters the region, and the survey never moves focus to `#fbjs`.

Concrete scenario: a respondent tabs from the question input to Next, advances to page 2, focus lands inside the new card. Nothing announces the instructions text — the only way to reach it is browse/virtual-cursor mode, where the `sr-only` div is discoverable *because it is in the DOM*, not because of `aria-describedby`. The reference is therefore doing less than the comment claims.

If the goal is a per-page announcement, this needs a different mechanism (e.g. including the instructions text in the per-card region/heading that already receives focus on navigation, or the existing `ensureLiveRegion()` used on modal open). Worth confirming against a real screen reader before this is treated as closing the VPAT item — a Playwright `toHaveAttribute` assertion cannot distinguish "referenced" from "announced".

---

packages/surveys/src/components/general/render-survey.tsx
```diff
@@ -68,7 +69,9 @@ export function RenderSurvey(props: SurveyContainerProps) {
       clickOutside={props.clickOutside}
       onClose={close}
       isOpen={isOpen}
-      dir={dir}>
+      dir={dir}
+      surveyName={props.survey.name}
```
kiln-agent: **This makes the *internal* survey name respondent-facing for in-app/embedded surveys, where it never was before.**

`survey.name` is the name authors pick in the survey list for their own bookkeeping — it is not the respondent-facing title. For link surveys it already leaks into the `<title>` as a last-resort fallback in `getBasicSurveyMetadata`, but for **app surveys** (the embedded widget) nothing ever exposed it: the modal announced the generic `common.survey_dialog`, and inline `#fbjs` had no name.

Concrete scenario: a workspace with a survey named `"Churn survey – Acme pilot – DO NOT LAUNCH"` triggered in-app. Every screen-reader respondent now hears that string as the dialog/form name, and it is readable in the DOM by anyone inspecting the page. Since `TJsWorkspaceStateSurvey.name` is required, the `surveyName ? …` / `?? t("common.survey_dialog")` fallbacks are effectively dead — the internal name is always used.

Given the PR is already marked `breaking-change`, this deserves an explicit decision: either accept it and document it (authors need to know the name is now public), or name the form from something respondent-facing (welcome headline / link-metadata title / first block headline) and keep `survey.name` as the last resort.

---

mattinannt (dismissed):

packages/surveys/src/components/wrappers/survey-container.tsx
```diff
@@ -1,15 +1,107 @@
...
+    if (!enabled) return;
+
+    const container = containerRef.current;
+    if (!container) return;
+
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key !== "Escape" || event.altKey || event.ctrlKey || event.metaKey) return;
+
+      event.preventDefault();
+      onClose?.();
```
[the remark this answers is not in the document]
itsjavi: Not taking this one here, because the one-line fix would create a worse inconsistency than the one it removes.

That guard is pre-existing — it is byte-identical on `main` (`survey-container.tsx:83`); `d21e0c5` only moved it into the hook, which is why it shows as a changed line.

The reason not to patch just this copy: the **overlay** path has its own Escape handler in `use-focus-trap.ts:273`, and it makes exactly the same choice —

```ts
const hasModifierKey = event.altKey || event.ctrlKey || event.metaKey;
if (event.key === "Escape" && !hasModifierKey && onEscapeKeyDownRef.current) {
```

So the two paths agree today. Adding `shiftKey` to only the no-overlay one would mean Shift+Escape closes a modal survey **with** an overlay but not **without** one — a divergence in the same behaviour across two code paths, which is harder to reason about than the current uniform choice.

The finding itself is fair, and there is a real argument for it beyond tidiness: Chrome binds Shift+Esc to its task manager, so `preventDefault()` on that combination swallows a browser shortcut. But the correct fix changes both handlers, and `use-focus-trap.ts` is a shared focus-management module this PR does not otherwise touch — widening a page-title/form-name PR into it is how an unrelated regression gets in. Filing it as a follow-up to fix in one place, ideally by hoisting the modifier check so there is one definition instead of two.

---

mattinannt (approved):

PR #8928, opened 2026-08-20
chore: add consistent UTM tags to all outbound formbricks.com links

harshsbhat:
## What & why

The app links out to `formbricks.com` from ~20 places — the "Powered by Formbricks" badge on surveys and emails, the login page logo, EE upgrade/upsell prompts, the security settings tip, and the default CTA URLs baked into new surveys and survey templates. Some already carried a UTM tag, but the scheme was inconsistent (`utm_source=survey_branding`, `utm_source=email_branding`, `utm_source=ce`, `utm_source=email_header`, one with `utm_medium` but no `utm_campaign`), and several links — the EE "learn more" upsell links, the two "upgrade to self-hosted license" prompts, the security tip, and the default survey/template CTA URLs — had no UTM at all. That makes it impossible to tell, from formbricks.com's analytics, which in-app surface actually drove a given visit.

This PR puts every one of them on one scheme: `utm_source=formbricks-app&utm_medium={survey|email|webapp}&utm_campaign=<unique_placement>`, so each link can be attributed to its exact surface.

<details>
<summary>Full list of links touched (24 files)</summary>

- `packages/surveys/.../formbricks-branding.tsx` — survey "Powered by" badge → `utm_campaign=powered_by_badge`
- `apps/web/.../survey-completed-message.tsx` — link-survey completed page → `utm_campaign=survey_completed`
- `apps/web/.../survey-inactive.tsx` — "create your own" CTA and footer logo → `utm_campaign=create_your_own_cta` / `powered_by_badge`
- `packages/email/.../email-template.tsx` — email header logo link and footer text link → `utm_campaign=email_logo` / `email_footer_text`
- `apps/web/.../preview-email-template.tsx` — embed-survey email "Powered by" → `utm_campaign=powered_by_badge`
- `apps/web/.../form-wrapper.tsx` — login/signup page logo → `utm_campaign=auth_logo`
- `MainNavigation.tsx` / `workspace-breadcrumb.tsx` — "upgrade to self-hosted license" prompts → `utm_campaign=upgrade_prompt_nav` / `upgrade_prompt_breadcrumb`
- `SecurityListTip.tsx` — security settings tip → `utm_campaign=security_tip`
- 9 EE-gated "learn more" upsell links (profile 2FA, personal links, contacts, feedback directory, quotas, unify-feedback sources/topics, email whitelabel, survey targeting) → one unique `utm_campaign=ee_lock_*` each
- `survey-builder.ts` / `end-screen-form.tsx` — default "Ending" screen CTA on new surveys → `utm_campaign=default_ending_cta`
- `templates.ts` — 4 built-in survey templates' default CTA button URLs → `utm_campaign=template_cta_*`

</details>

## Breaking changes

None — only query parameters are added/changed on outbound links; no route, API, or payload shape changes.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Embed-survey preview email "Powered by Formbricks" link carries the new UTM params and still renders/matches the static fixture | unit (guard) — `apps/web/modules/email/embed-survey-preview-email.test.ts` | Pass — updated the fixture (`packages/email/src/lib/fixtures/embed-survey-preview-email-html.ts`) and assertions to the new href; full suite green |
| Same link, rendered in a real preview iframe | e2e (guard) — `apps/web/playwright/survey-email-preview.spec.ts` → "keeps the powered-by link's UTM tag intact" | Pass — updated the expected `href` |
| Every other touched link (survey branding badge, login logo, upgrade prompts, EE upsell links, default template/ending CTAs) | manual — read diff, grepped for remaining bare `formbricks.com` occurrences outside test/mock/fixture/seed files | Confirmed only the intended 24 files changed; no leftover unparameterized links |
| No regressions elsewhere | `pnpm build --filter=@formbricks/web^...` then `pnpm --filter=@formbricks/web test` (627 files / 8289 tests) and `pnpm typecheck` on `apps/web`, `packages/email`, `packages/surveys` | All green |

**Open gaps**

- none

**Risks**

- Low. Purely additive query-string changes to outbound links; no behavior change for anything other than the destination URL's tracking params.

---

> [!NOTE]
> **AI model used** — Claude Sonnet 5, reasoning effort medium.

### DIFF
--- a/apps/web/app/lib/survey-builder.ts
+++ b/apps/web/app/lib/survey-builder.ts
@@ -84,7 +84,8 @@ export const getDefaultEndingCard = (languages: TSurveyLanguage[], t: TFunction)
     headline: createI18nString(t("templates.default_ending_card_headline"), languageCodes),
     subheader: createI18nString(t("templates.default_ending_card_subheader"), languageCodes),
     buttonLabel: createI18nString(t("templates.default_ending_card_button_label"), languageCodes),
-    buttonLink: "https://formbricks.com",
+    buttonLink:
+      "https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=default_ending_cta",
   };
 };
 
--- a/apps/web/modules/survey/editor/components/end-screen-form.tsx
+++ b/apps/web/modules/survey/editor/components/end-screen-form.tsx
@@ -119,7 +119,8 @@ export const EndScreenForm = ({
               } else {
                 updateSurvey({
                   buttonLabel: { default: t("workspace.surveys.edit.create_your_own_survey") },
-                  buttonLink: "https://formbricks.com",
+                  buttonLink:
+                    "https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=default_ending_cta",
                 });
               }
               setshowEndingCardCTA(!showEndingCardCTA);
--- a/apps/web/modules/survey/link/components/survey-completed-message.tsx
+++ b/apps/web/modules/survey/link/components/survey-completed-message.tsx
@@ -30,7 +30,7 @@ export const SurveyCompletedMessage = async ({
       </div>
       {(!workspace || workspace.linkSurveyBranding) && (
         <div>
-          <Link href="https://formbricks.com">
+          <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=survey_completed">
             <Image src={footerLogo} alt="Brand logo" className="mx-auto w-40" />
           </Link>
         </div>
--- a/apps/web/modules/survey/link/components/survey-inactive.tsx
+++ b/apps/web/modules/survey/link/components/survey-inactive.tsx
@@ -65,13 +65,15 @@ export const SurveyInactive = async ({
         <p className="text-lg leading-10 text-slate-500">{description}</p>
         {showCTA && (
           <Button className="mt-2" asChild>
-            <Link href="https://formbricks.com">{t("s.create_your_own")}</Link>
+            <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=create_your_own_cta">
+              {t("s.create_your_own")}
+            </Link>
           </Button>
         )}
       </div>
       {(!workspace || workspace.linkSurveyBranding) && (
         <div>
-          <Link href="https://formbricks.com">
+          <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=powered_by_badge">
             <Image src={footerLogo} alt="Brand logo" className="mx-auto w-40" />
           </Link>
         </div>
--- a/packages/surveys/src/components/general/formbricks-branding.tsx
+++ b/packages/surveys/src/components/general/formbricks-branding.tsx
@@ -4,7 +4,10 @@ export function FormbricksBranding() {
   const { t } = useTranslation();
   return (
     <span className="flex justify-center">
-      <a href="https://formbricks.com?utm_source=survey_branding" target="_blank" rel="noopener">
+      <a
+        href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=powered_by_badge"
+        target="_blank"
+        rel="noopener">
         <p className="text-signature text-xs">
           {t("common.powered_by")}{" "}
           <b>
--- a/apps/web/app/(app)/account/settings/profile/page.tsx
+++ b/apps/web/app/(app)/account/settings/profile/page.tsx
@@ -76,7 +76,7 @@ const Page = async () => {
                     text: t("common.learn_more"),
                     href: IS_FORMBRICKS_CLOUD
                       ? billingUpgradeHref
-                      : "https://formbricks.com/learn-more-self-hosting-license",
+                      : "https://formbricks.com/learn-more-self-hosting-license?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=ee_lock_two_factor",
                   },
                 ]}
               />
--- a/apps/web/app/(app)/workspaces/[workspaceId]/components/MainNavigation.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/components/MainNavigation.tsx
@@ -325,7 +325,7 @@ export const MainNavigation = ({
         text: t("workspace.settings.billing.upgrade"),
         href: isLicenseActive
           ? `/organizations/${organization.id}/settings/enterprise`
-          : "https://formbricks.com/upgrade-self-hosted-license",
+          : "https://formbricks.com/upgrade-self-hosted-license?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=upgrade_prompt_nav",
       },
       {
         text: t("common.cancel"),
--- a/apps/web/app/(app)/workspaces/[workspaceId]/components/workspace-breadcrumb.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/components/workspace-breadcrumb.tsx
@@ -124,7 +124,7 @@ export const WorkspaceBreadcrumb = ({
         text: t("workspace.settings.billing.upgrade"),
         href: isLicenseActive
           ? `/organizations/${currentOrganizationId}/settings/enterprise`
-          : "https://formbricks.com/upgrade-self-hosted-license",
+          : "https://formbricks.com/upgrade-self-hosted-license?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=upgrade_prompt_breadcrumb",
       },
       {
         text: t("common.cancel"),
--- a/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/general/components/SecurityListTip.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/settings/organization/general/components/SecurityListTip.tsx
@@ -13,7 +13,7 @@ export const SecurityListTip = () => {
         <p className="text-sm">
           {t("workspace.settings.general.security_list_tip")}{" "}
           <Link
-            href="https://formbricks.com/security#stay-informed-with-formbricks-security-updates"
+            href="https://formbricks.com/security?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=security_tip#stay-informed-with-formbricks-security-updates"
             target="_blank"
             rel="noopener noreferrer"
             className="underline hover:text-blue-700">
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/shareEmbedModal/personal-links-tab.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/components/shareEmbedModal/personal-links-tab.tsx
@@ -177,7 +177,7 @@ export const PersonalLinksTab = ({
             text: t("common.learn_more"),
             href: isFormbricksCloud
               ? `/organizations/${workspace?.organizationId}/settings/billing`
-              : "https://formbricks.com/learn-more-self-hosting-license",
+              : "https://formbricks.com/learn-more-self-hosting-license?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=ee_lock_personal_links",
           },
         ]}
       />
--- a/apps/web/app/lib/templates.ts
+++ b/apps/web/app/lib/templates.ts
@@ -571,7 +571,8 @@ const churnSurvey = (t: TFunction): TTemplate => {
               subheader: t("templates.churn_survey_question_3_html"),
               headline: t("templates.churn_survey_question_3_headline"),
               required: false,
-              buttonUrl: "https://formbricks.com",
+              buttonUrl:
+                "https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=template_cta_churn_survey",
               buttonExternal: true,
               ctaButtonLabel: t("templates.churn_survey_question_3_button_label"),
             }),
@@ -1000,7 +1001,8 @@ const improveTrialConversion = (t: TFunction): TTemplate => {
               subheader: t("templates.improve_trial_conversion_question_4_html"),
               headline: t("templates.improve_trial_conversion_question_4_headline"),
               required: false,
-              buttonUrl: "https://formbricks.com/github",
+              buttonUrl:
+                "https://formbricks.com/github?utm_source=formbricks-app&utm_medium=survey&utm_campaign=template_cta_trial_conversion",
               buttonExternal: true,
               ctaButtonLabel: t("templates.improve_trial_conversion_question_4_button_label"),
             }),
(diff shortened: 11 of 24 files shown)


jobenjada (approved):
looks good :)

apps/web/modules/survey/link/components/survey-completed-message.tsx
```diff
@@ -30,7 +30,7 @@ export const SurveyCompletedMessage = async ({
       </div>
       {(!workspace || workspace.linkSurveyBranding) && (
         <div>
-          <Link href="https://formbricks.com">
+          <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=survey_completed">
```
kiln-agent, line 33: Same badge, two different campaigns — this will split the "Powered by" numbers.

This block and the one in `survey-inactive.tsx:76` are byte-for-byte identical: same `footerLogo` import, same `alt="Brand logo"`, same `mx-auto w-40`, same `{(!workspace || workspace.linkSurveyBranding) && …}` gate. But this one gets `utm_campaign=survey_completed` while the other gets `utm_campaign=powered_by_badge`.

So the campaign dimension is mixing two conventions: here it names the *page*, there it names the *element*. A report filtered on `utm_campaign=powered_by_badge` will silently under-count link-survey branding clicks by however many land on the already-answered page, and `survey_completed` will read as page-level attribution when it is really the same badge.

Suggest `powered_by_badge` here too (the badge is the thing being clicked, and the page is already implied by `utm_medium=survey`). If the intent really is per-page attribution, then `survey-inactive.tsx:76` should follow suit rather than the other way around — either way the two should agree.

---

packages/surveys/src/components/general/formbricks-branding.tsx
```diff
@@ -4,7 +4,10 @@ export function FormbricksBranding() {
   const { t } = useTranslation();
   return (
     <span className="flex justify-center">
-      <a href="https://formbricks.com?utm_source=survey_branding" target="_blank" rel="noopener">
+      <a
+        href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=powered_by_badge"
```
kiln-agent, line 8: Two distinct surfaces collapsed into one identical tag.

This URL is now character-for-character the same as `survey-inactive.tsx:76` (`utm_source=formbricks-app&utm_medium=survey&utm_campaign=powered_by_badge`), but the two surfaces are very different: this one is the in-survey badge rendered by the `@formbricks/surveys` bundle **on the customer's own website** (and in link surveys), while the other is the footer logo on a Formbricks-hosted inactive-survey page. After this change nothing in the tags tells them apart.

That's the inverse of the problem on `survey-completed-message.tsx` in the same PR: there one element is split across two campaigns, here two elements are merged into one. Worth picking one axis for `utm_campaign` and holding it — e.g. `powered_by_badge_widget` vs `powered_by_badge_page`.

Related: `utm_source=formbricks-app` is a stretch for this component specifically. It renders to *respondents* on third-party sites who have usually never seen the Formbricks app — the old `utm_source=survey_branding` at least carried that. Since `utm_source` conventionally means the referring property, `formbricks-app` here points at the wrong origin. Same applies to the link-survey pages. If the collapse is deliberate (source = "our product surfaces", medium = which surface), that's defensible — but please note it in the description, because it also drops the cloud-vs-self-hosted signal that `utm_source=ce` used to give on the auth logo, and `form-wrapper.tsx` renders on both.

---

apps/web/app/lib/survey-builder.ts
```diff
@@ -84,7 +84,8 @@ export const getDefaultEndingCard = (languages: TSurveyLanguage[], t: TFunction)
     headline: createI18nString(t("templates.default_ending_card_headline"), languageCodes),
     subheader: createI18nString(t("templates.default_ending_card_subheader"), languageCodes),
     buttonLabel: createI18nString(t("templates.default_ending_card_button_label"), languageCodes),
-    buttonLink: "https://formbricks.com",
+    buttonLink:
+      "https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=default_ending_cta",
```
kiln-agent, line 88: This exact 99-character literal is duplicated verbatim in `end-screen-form.tsx:123`.

Both spell out `https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=default_ending_cta` by hand, and they have to stay identical — `checkExternalUrlsPermission` compares `oldEnding.buttonLink !== newEnding.buttonLink` for its grandfathering check, so a one-character drift between the builder default and the editor toggle would make a free-plan org's save fail with `OperationNotAllowedError` on a link they never edited. That was cheap to keep in sync when both sides read `"https://formbricks.com"`; at 99 characters it isn't.

Zooming out, the PR hand-writes 29 of these strings with no shared constant: the `learn-more-self-hosting-license` base appears in 9 files and the `?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=` prefix about ten times. A small `apps/web/lib/formbricks-links.ts` holding the base URLs plus a `withUtm(base, { medium, campaign })` helper would collapse all of that, remove the drift risk here, and make the next taxonomy change one line instead of 24 files. This PR is itself the second rewrite of these params — it replaces four earlier conventions (`ce`, `survey_branding`, `email_branding`, `email_header`) — so a third is likely. A helper would also have caught the two campaign inconsistencies flagged elsewhere in this review, since the surfaces would be enumerated in one place.

(`packages/surveys` and `packages/email` can't import from `apps/web`, so they'd need their own tiny equivalents or a shared package — but the 24 call sites inside `apps/web` are the bulk of it.)

---

kiln-agent (commented):
Mechanically clean — the URL rewrites themselves are all correct, and the old conventions are fully retired (no `utm_source=ce`, `survey_branding`, `email_branding` or `email_header` left anywhere in the tree). My findings are about the **tag taxonomy**, not the code: for a PR whose whole purpose is inbound attribution, two of the campaign values won't aggregate the way the reports will assume. Details in the three inline comments:

1. `survey-completed-message.tsx` and `survey-inactive.tsx` render the *identical* branding badge but get different `utm_campaign` values (`survey_completed` vs `powered_by_badge`) — one element split across two campaigns.
2. `formbricks-branding.tsx` (the widget badge on customers' sites) and `survey-inactive.tsx` now carry *identical* tags — two very different surfaces merged into one.
3. No shared constant: 29 hand-written URLs, including one 99-char literal duplicated verbatim across `survey-builder.ts:88` and `end-screen-form.tsx:123` that `checkExternalUrlsPermission` requires to stay byte-identical.

Two smaller notes:

**Scope vs. the commit message.** "all outbound formbricks.com links" isn't quite what landed — 73 `https://formbricks.com/docs` links across 34 files are untagged (`user-dropdown.tsx`, the integrations pages, `follow-ups-view.tsx`, the webhook and enterprise-feature tables, …). Those hit the same domain, so the inbound report will still mix tagged and untagged formbricks.com traffic. Either bring them in (the helper from #3 makes it cheap) or narrow the claim to marketing/upgrade links. Not a blocker, just worth the description being accurate about what's measurable after this merges.

**I'd leave the breaking-changes box unticked.** CodeRabbit's pre-merge check wants it ticked over the changed `buttonLink`/`buttonUrl` defaults in `survey-builder.ts` and `templates.ts`, but those defaults only apply to *newly created* surveys — existing rows keep their bare URLs, there's no migration, and nothing self-hosters need to do. Ticking it would put a phantom entry in the release notes and the migration guide. Your current answer looks right to me; the checkbox prose is where to say so.

**Verification I ran** (rather than restating CI): `app/lib/survey-builder.test.ts` + `app/lib/templates.test.ts` pass (9 tests), and `modules/email/embed-survey-preview-email.test.ts` passes (20 tests) — that last one is the useful one, since it confirms the hand-edited `embed-survey-preview-email-html.ts` fixture still matches the live render including the `&amp;` escaping, so the two assertions replacing the old `utm_source=email_branding` check do hold against both. Also checked: nothing compares these URLs for equality against a bare `https://formbricks.com`, `ZEndingCardUrl`/`getZSafeUrl` both accept query strings so no validation path rejects the tagged URLs, and the query-before-fragment ordering in `SecurityListTip.tsx` is correct (`?utm…#stay-informed`). I could not run `playwright/survey-email-preview.spec.ts`, but the `toHaveAttribute` expectation with a raw `&` is right — Playwright returns the decoded attribute value.

Nothing here blocks merging; #1 is the one I'd fix before it starts collecting data, since backfilling a mis-tagged campaign after the fact isn't possible.

---

apps/web/modules/survey/editor/components/end-screen-form.tsx
```diff
@@ -119,7 +119,8 @@ export const EndScreenForm = ({
               } else {
                 updateSurvey({
                   buttonLabel: { default: t("workspace.surveys.edit.create_your_own_survey") },
-                  buttonLink: "https://formbricks.com",
+                  buttonLink:
+                    "https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=default_ending_cta",
```
kiln-agent, line 123: This default value is load-bearing for the free-plan grandfathering in `check-external-urls-permission.ts`, and changing it locks those users out of saving.

`checkExternalUrlsPermission` grandfathers an existing ending-card link by **exact string equality** ([`check-external-urls-permission.ts:41`](``https://github.com/formbricks/formbricks/blob/f36a56bd708e9b09710eabe45179401badf9e953/apps/web/modules/survey/editor/lib/check-external-urls-permission.ts#L41)):``

```ts
if (!oldEnding || oldEnding.type !== "endScreen" || oldEnding.buttonLink !== newEnding.buttonLink) {
  throw new OperationNotAllowedError(...)
}
```

Failure scenario, on an org **without** the external-URLs entitlement:

1. Survey created before this PR — stored `buttonLink` is `https://formbricks.com`.
2. In the editor the user toggles "Show button" off (writes `buttonLink: undefined`) and back on — this handler now writes `https://formbricks.com?utm_source=…&utm_campaign=default_ending_cta`.
3. Save → `oldEnding.buttonLink !== newEnding.buttonLink` → `OperationNotAllowedError`. The survey can no longer be saved at all, including unrelated edits made in the same session.
4. They can't type the old value back: the URL input is `disabled={!isExternalUrlsAllowed}` (line 205 of this file). The only escape is turning the CTA off entirely.

Before this PR step 2 wrote a byte-identical string, so the off/on round-trip saved fine. Both call sites of `checkExternalUrlsPermission` ([`editor/actions.ts:231`](https://github.com/formbricks/formbricks/blob/f36a56bd708e9b09710eabe45179401badf9e953/apps/web/modules/survey/editor/actions.ts#L231) and `:289`) are update paths, so creation is unaffected — this is purely about editing pre-existing surveys.

Same default in [`app/lib/survey-builder.ts:88`](https://github.com/formbricks/formbricks/blob/f36a56bd708e9b09710eabe45179401badf9e953/apps/web/app/lib/survey-builder.ts#L88) has the same property.

Cheapest fix: compare origin+path instead of the raw string in the grandfathering check, or normalize away Formbricks-owned UTM params before the comparison, so the entitlement check keys on "did the user point this somewhere new" rather than "is this byte-identical". Tagging the default is fine once the check tolerates it.

---

apps/web/modules/survey/link/components/survey-inactive.tsx
```diff
@@ -65,13 +65,15 @@ export const SurveyInactive = async ({
...
+            <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=create_your_own_cta">
+              {t("s.create_your_own")}
+            </Link>
           </Button>
         )}
       </div>
       {(!workspace || workspace.linkSurveyBranding) && (
         <div>
-          <Link href="https://formbricks.com">
+          <Link href="https://formbricks.com?utm_source=formbricks-app&utm_medium=survey&utm_campaign=powered_by_badge">
```
kiln-agent, line 76: The `utm_campaign` assignment here cuts across surfaces in a way that will make the resulting data harder to read, not easier — which is the one thing this PR is for.

Three links, all `utm_medium=survey`:

| Surface | File | `utm_campaign` |
| --- | --- | --- |
| Footer logo, survey-inactive page | this line | `powered_by_badge` |
| Footer logo, survey-completed page | [`survey-completed-message.tsx:33`](``https://github.com/formbricks/formbricks/blob/f36a56bd708e9b09710eabe45179401badf9e953/apps/web/modules/survey/link/components/survey-completed-message.tsx#L33)`` | `survey_completed` |
| In-survey branding badge | [`formbricks-branding.tsx:8`](``https://github.com/formbricks/formbricks/blob/f36a56bd708e9b09710eabe45179401badf9e953/packages/surveys/src/components/general/formbricks-branding.tsx#L8``) | `powered_by_badge` |

So the same footer-logo element is split across two campaigns depending on which page it renders on, while this one is merged with an unrelated surface (the in-survey badge). Clicks on `powered_by_badge` become un-splittable between "someone hit a closed survey" and "someone was mid-survey", and footer-logo performance can't be compared page to page.

Note the neighbouring CTA on this same page already uses a surface-specific campaign (`create_your_own_cta`), so the pattern is right there. Suggest giving each element+surface its own campaign — e.g. `survey_inactive_badge` here, `survey_completed_badge` for the completed page, `powered_by_badge` reserved for the in-survey badge only.

---

kiln-agent (commented):
Reviewed the full 24-file diff at `f36a56b`. The mechanical part of this is clean — I checked that no references to the five old UTM schemes survive anywhere in the repo, that the URL Zod refinements (`ZEndingCardUrl` / `getZSafeUrl` in `packages/types/common.ts`) accept query strings, that the ending-card and CTA render paths don't append their own query params (so the added `?…` can't be corrupted downstream), that all 11 self-hosting-license links are covered, and that the email fixture, the vitest guard and the Playwright expectation agree on `&amp;` vs `&`. Affected suites pass locally after rebuilding stale package `dist/`s: `modules/email` (58 tests), `app/lib/templates.test.ts`, `app/lib/survey-builder.test.ts`, `app/api/v3/surveys/templates`.

Two things I'd want addressed, both left inline:

1. **`end-screen-form.tsx:122` (and the same default in `app/lib/survey-builder.ts:88`) — free-plan users can be locked out of saving a survey.** The external-URLs entitlement check grandfathers existing links by exact string equality, so changing the default means a pre-existing survey's CTA off/on toggle now writes a *different* URL than the stored one and every save throws `OperationNotAllowedError` — with the URL input disabled, there's no way to type the old value back. This is the one I'd call blocking; details and a suggested fix inline.

2. **`survey-inactive.tsx:76` — `utm_campaign` values don't partition the surfaces cleanly.** The same footer logo gets two different campaigns depending on page, and this one collides with the unrelated in-survey branding badge. Worth settling before this lands, since the whole point is comparable attribution data and these values are what the reports key on.

Two smaller notes, not worth inline comments:

- `app/lib/templates.ts:2645` and `:3732` still point at `https://app.formbricks.com/auth/signup` untagged. `modules/auth/lib/attribution.ts` does capture UTM params on that page, so nothing is lost server-side — but they're the last untagged outbound links in the diff's scope, so flagging in case the omission wasn't deliberate.
- `packages/email/src/components/email-template.tsx` uses `formbricks.com?utm…` for the logo and `formbricks.com/?utm…` for the footer text. Functionally identical, just inconsistent within one file.

One process note: CodeRabbit's `Breaking Changes Match The Diff` warning is worth a real answer rather than a dismissal. The changed default CTA values aren't a breaking change in the API-contract sense, but finding 1 is exactly the kind of user-visible regression that check exists to surface — resolving finding 1 also resolves the warning.

---

PR #8929, opened 2026-08-20
fix(telemetry): send usage updates independently of survey traffic [ENG-2107]

Dhruwang:
Fixes ENG-2107

## What & why

`sendTelemetryEvents` — the function that POSTs `/api/v1/instances/{instanceId}/usage-updates` — had exactly **one** production caller: `processResponsePipelineJob`. A usage update was therefore only ever sent as a side effect of processing a survey response, so an instance that identified itself against the license server but never collected a response reported no usage at all. That is the reported symptom: identified instances with no usage update.

- Registers the usage update as a recurring job on the existing BullMQ recurring-job framework (`usage-telemetry.process`, daily at 02:15 UTC) so reporting no longer depends on survey traffic.
- An instance that is **not up at 02:15 UTC** — the case the ticket calls out ("maybe they don't run the app all the time") — still reports, because a missed tick is not skipped: the boot-time upsert re-adds the overdue iteration with its original timestamp, so its delay clamps to 0 and it runs at that boot.
- The schedule also sets `immediately: true`, which fires **once per scheduler**, not once per boot. BullMQ's repeat strategy does return `now` when it is set, but `addJobScheduler-11.lua` discards that: when the upsert removed a pending job it sets `nextMillis = prevMillis` ("lets use the same millis"), which is every boot after the first. That still lands where it is wanted — this scheduler is new, so every instance upgrading past this change registers it for the first time and reports on that boot instead of waiting for the first 02:15 slot. *(Corrected: this bullet previously claimed a run on each boot. Thanks @BhagyaAmarasinghe — the mechanism is in the review thread.)*
- Idempotency is unchanged and now carries the second trigger: the existing three-layer guard (process memory → shared 24h `telemetry_last_sent_ts` in Redis → distributed lock) makes a boot-time run a single Redis read whenever an update already went out. `TELEMETRY_DISABLED` still gates it for CE, and the dev/E2E hard skip still applies.

Two defects in the same path also delayed the *first* update, and both had to go for the new trigger to help a fresh instance:

- With no organization yet, `sendTelemetry` returned early but the caller still wrote `telemetry_last_sent_ts` — burning the 24h window without sending anything. It now reports whether an update went out, and takes a 1h cooldown instead. (The pre-existing test asserted this behaviour with the comment *"This might be a bug, but we test the actual behavior"*; it now asserts the fix.)
- The `fetch` response status was never checked, so a rejected update was recorded as sent and the instance stayed silent for another 24h. A non-2xx now raises into the existing failure cooldown.

The module moves from `modules/response-pipeline/lib/telemetry.ts` to `lib/telemetry/usage-update.ts`; it is no longer response-pipeline specific now that a job triggers it too. Content is unchanged apart from the two fixes above.

## Breaking changes

None — no env var, endpoint, API/SDK shape, config key or default changes, and no manual action for self-hosters. Self-hosted instances that had telemetry enabled will now report on a schedule rather than only after a response, which is the intended fix; `TELEMETRY_DISABLED=1` opts out exactly as before.

## Migrations & env

- No env vars and no DB migrations.
- Deploy-visible (no action needed): one new recurring schedule, `usage-telemetry.process`, is upserted at boot alongside the existing ones, and appears in the Job Runner's job table — `docs/self-hosting/configuration/job-runner.mdx` is updated. Deployments that run without the job runner / Redis get no recurring jobs, same as today; telemetry already required Redis for its shared 24h gate, so nothing regresses there.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| No org yet → the 24h window is not consumed, so the next trigger can send once an org exists | unit (red on main) | `apps/web/lib/telemetry/usage-update.test.ts` "should skip if no organization exists". Verified red against the old behaviour by neutering the fix — `usage-update.ts:134` `if (!sent) {` → `if (false) {` — then `cd apps/web && npx vitest run lib/telemetry/usage-update.test.ts`: 2 failed / 10 passed. Restored → 12 passed |
| A rejected usage update (non-2xx) is not recorded as sent | unit (red on main) | `usage-update.test.ts` "should not record a send when the usage update endpoint rejects it". Red in the same run above with `usage-update.ts:329` (the `throw` on `!res.ok`) removed; asserts `telemetry_last_sent_ts` is not written, the 1h-cooldown warning carries `status 503`, and the lock is still released |
| The usage telemetry schedule is registered with the daily UTC pattern **and** `immediately` | unit (mutation) | `apps/web/instrumentation-jobs.test.ts` "registers recurring schedules once when queueing is enabled without an in-process worker" asserts the exact upsert payload, so dropping `immediately` or changing the pattern/zone fails. `immediately` is asserted explicitly, not via `objectContaining` |
| The job's handler is bound to `usage-telemetry.process` and receives the global-scope payload | unit (mutation) | `instrumentation-jobs.test.ts` invokes the resolved override and asserts `processUsageTelemetryJob` was called with `{ scope: "global" }` + context; the existing "each recurring registration is paired with its own job handle" test covers the mispairing that type-checks |
| The handler sends the usage update, and does not swallow a dispatch failure | unit (mutation) | `apps/web/lib/telemetry/process-usage-telemetry-job.test.ts`, 2 tests |
| Scheduler id `usage-telemetry.process:global:daily-usage-telemetry` stays stable (a change orphans the live schedule rather than updating it) | unit (guard) | Added a row to the existing pinned-id table in `packages/jobs/src/queue.test.ts` |
| The response-pipeline trigger still works across the module move | unit (guard) | `modules/response-pipeline/lib/process-response-pipeline-job.test.ts` (mock path repointed), incl. the existing "logs telemetry failures without failing the responseCreated job" |

Full sweep, all green: `pnpm lint` (0 errors), `pnpm typecheck` (27/27 tasks), `pnpm test` (628 files / 8292 tests in `apps/web`; 26/26 workspace tasks), `pnpm format:check`. `pnpm install` left `pnpm-lock.yaml` untouched.

**Open gaps**

- Not exercised against a live Redis/BullMQ. The scheduling semantics are read off the bullmq source rather than observed: `defaultRepeatStrategy` in `classes/job-scheduler.js` returns `new Date().getTime()` when `opts.immediately` is set, and `addJobScheduler-11.lua` then overrides it with `nextMillis = prevMillis` on any upsert that removed a pending job. The first reading is what the original description got wrong; both halves have now been read by two people, but neither has been watched against a real Redis.
- No end-to-end check against the real `ee.formbricks.com` endpoint: telemetry hard-skips in dev and E2E, so the delivery path itself is only covered with a mocked `fetch`.
- Whether the license server now records usage for previously-silent Lite instances can only be confirmed on its side after this ships — that half is not in this repo.
- The daily cron time (02:15 UTC) is a fixed literal, not configurable, matching how the survey sweeps pick their offsets. Not verified under a clock/DST edge, which UTC is chosen to avoid.

**Risks**

- ~~A crash-looping or frequently-restarting pod queues one usage-telemetry job per restart.~~ Not a risk: `immediately` fires once per scheduler, so restarts after the first re-add the pending iteration at its existing timestamp rather than queueing extra runs.
- Persistent non-2xx from the endpoint turns a silent failure into a visible, logged one. It does **not** mean hourly retries: the 1h value is a floor on the next attempt, not a schedule, so the retry rate is however often a trigger calls in (a response, the daily tick, or an overdue tick at boot) — at most hourly, and on a traffic-free instance, daily.
- A brand-new install with no organization no longer burns its 24h window on a send that reported nothing — that quiet window was the bug. But the recovery is *not* hourly: with no response traffic the first update after an org is created lands on the next daily tick, so up to a day out rather than up to an hour. Making it faster means enqueueing a delayed follow-up when nothing was sent; deliberately not done here, and worth its own ticket if a day is too slow for the GTM need.
- If the move left a stale importer behind, the response pipeline would silently stop reporting; `grep` for the old path returns nothing and the suite is green, but it is the thing to re-check if usage updates drop after this lands.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F3yZWMqQUS4dLfk1VeqCAF)_


Dhruwang:
Tested this locally against real Postgres + Redis (valkey) by restarting `apps/web` twice with the built dist and watching the BullMQ scheduler state in Redis.

**Confirmed:**
- `usage-telemetry.process` registers with `pattern: "15 2 * * *"`, `tz: UTC` — matches the daily 02:15 UTC claim.
- `immediately` is stripped from the persisted repeat options (checked the raw scheduler hash in Redis) — the load-bearing claim from the "Open gaps" section holds.
- On first-ever registration (no existing schedule), boot correctly queued and ran an immediate job, then scheduled the next iteration for the following day's 02:15 UTC slot.

**One correction to the PR description:** "It applies per upsert, not once per scheduler ... every boot queues one run" isn't quite right. I restarted the app a second time while the next daily iteration was still pending in Redis, and the second `upsert` did *not* queue a new immediate run — the existing pending job for the next cron slot was left untouched. `upsertJobScheduler` doesn't stack an ad-hoc immediate run on top of one already scheduled.

This probably doesn't undermine the fix's actual goal: if the app is down when the 02:15 slot passes, that job sits overdue in Redis and BullMQ should fire it as soon as a worker reconnects — which is the "instance barely runs" case ENG-2107 cares about. I didn't independently verify that overdue-catch-up path (would need a longer-running test), so flagging it as expected-but-unconfirmed rather than proven.

Worth tightening the "every boot" wording in the description, or leaving a note that it only fires on the schedule's initial creation.

apps/web/lib/telemetry/usage-update.ts
```diff
@@ -41,6 +45,51 @@ const isTelemetryDisabledForCE = async (): Promise<boolean> => {
...
+    const sent = await sendTelemetry(lastSent);
+
+    if (!sent) {
+      // Nothing was reported (no organization exists yet), so the 24h window must not be consumed:
+      // recording it here would delay the instance's *first* usage update by a day. Retry in 1h.
+      logger.info(
+        { hashedLicenseKey },
+        "Telemetry skipped - no organization to report on yet, applying 1h cooldown"
+      );
+      nextTelemetryCheck = now + 60 * 60 * 1000;
```
BhagyaAmarasinghe, line 70: **🟠 Major · correctness**
The 1h cooldown is a floor, not a retry — nothing re-invokes telemetry hourly

`nextTelemetryCheck` only permits an earlier attempt; nothing schedules one.

- Callers are exactly two: `process-response-pipeline-job.ts:14` and the daily job (`recurring-registrations.ts:80`). No response traffic — this PR's target population — means no hourly trigger.
- Breaks when: fresh install boots with no org → `sent` is `false` → 1h cooldown; org created ten minutes later; no responses. Next attempt is the 02:15 UTC tick, not 1h. So the body's "retries hourly until an org exists" and `process-usage-telemetry-job.ts:20-21` don't hold, and the first usage update is still a day out.
- Fix: enqueue a `runAt` follow-up when `sent === false`, or say the retry is the next daily tick.

apps/web/lib/jobs/recurring-registrations.ts
```diff
@@ -74,6 +76,26 @@ export const RECURRING_JOB_REGISTRATIONS_BY_KEY: Record<TRecurringJobKey, Recurr
...
+    handler: processUsageTelemetryJob,
+    job: recurringJobs.usageTelemetry,
+    schedule: {
+      cronPattern: USAGE_TELEMETRY_DAILY_CRON_PATTERN,
+      // The daily pattern keeps a long-running instance reporting; `immediately` is what covers an
+      // instance that is *not* up at 02:15 UTC, which is the case the GTM need calls out — an instance
+      // may be identified and then barely run (ENG-2107).
+      //
+      // It applies per upsert, not once per scheduler: BullMQ's repeat strategy returns "now" whenever
+      // `immediately` is set, and `immediately` is stripped from the persisted repeat options, so every
```
BhagyaAmarasinghe: **🟡 Minor · drift**
In-code rationale still claims every boot queues a run, which your own testing disproved

Your PR comment records that a second boot queues nothing; this comment — and the body's *What & why* — still say the opposite at this head.

- Mechanism: `Queue.upsertJobScheduler` passes `override: true` (bullmq `queue.js:267`), and `addJobScheduler-11.lua:123-125` sets `nextMillis = prevMillis` whenever the upsert removed a pending job, discarding the `now` from `defaultRepeatStrategy`. So `immediately` fires only on the first-ever registration.
- That changes who benefits: an upgrading instance has an org, so that one run reports. A fresh install fires before any org exists, reports nothing, and never fires again.
- Also makes the *Risks* bullet on crash-looping pods wrong.

BhagyaAmarasinghe (commented):
**1 major · 1 minor** — reviewed at `ac45e28966`

**Checked**

- `immediately` is plumbed to BullMQ — `schedules.ts:99`, not dropped
- Overdue 02:15 job fires at next boot — lua clamps delay 0
- No importer left on the old `response-pipeline/lib/telemetry` path
- `TELEMETRY_DISABLED` and the dev/E2E skip still gate the new job
- No other docs page states a telemetry cadence this diff invalidates
- No local checks run in this environment — read-only pass

**Findings**

- **🟠 Major · correctness** `apps/web/lib/telemetry/usage-update.ts:64` — The 1h cooldown is a floor, not a retry — nothing re-invokes telemetry hourly
- **🟡 Minor · drift** `apps/web/lib/jobs/recurring-registrations.ts:89` — In-code rationale still claims every boot queues a run, which your own testing disproved

_Reply per finding: the fix, why it's wrong, or a ticket._

_Opus 5 · high_

BhagyaAmarasinghe (changes requested):
**Requesting changes** — 1 major · 1 minor, reviewed at `ac45e28966`

Blocking on the major: the 1h cooldown at `apps/web/lib/telemetry/usage-update.ts:64` is a floor on `nextTelemetryCheck`, not a scheduled retry. `sendTelemetryEvents` has exactly two callers — the response pipeline and the new daily job — so an instance with no response traffic has nothing invoking it hourly. A fresh install boots with no org → `sent === false` → cooldown → next real attempt is the 02:15 UTC tick, so the first usage update is still a day out and the body's "retries hourly until an org exists" does not hold.

Also open: `apps/web/lib/jobs/recurring-registrations.ts:89` still states that every boot queues one run, which your own testing on this PR disproved. The mechanism is in that thread — `addJobScheduler-11.lua` reuses `prevMillis` whenever the upsert removed a pending job, so `immediately` only fires on the first-ever registration.

Evidence, failure scenarios and suggested fixes are in the two inline threads on the review above.

apps/web/lib/telemetry/usage-update.ts
Dhruwang: Correct, and taken in dc94fd8be — the copy was wrong, not the code, so this is a comment/description fix rather than a behaviour one.

You're right that `nextTelemetryCheck` only permits an earlier attempt and that nothing schedules one. I traced the callers again and they are exactly the two you name, so on an instance with no response traffic the "1h" is a floor and nothing more.

I took your second option — say what the retry actually is — rather than enqueueing a `runAt` follow-up, for two reasons. The follow-up would have to be enqueued from `sendTelemetryEvents`, which is also the response-pipeline path, so it couples the telemetry module to the job queue on a path that does not need it. And the exposure is smaller than the thread implies once the triggers are listed honestly: a response, the next 02:15 tick, or that tick running overdue at the next boot — which, as you checked yourself, does fire. So the worst case for your scenario (fresh install, org ten minutes later, no responses) is the first usage update landing up to a day out, not never.

Whether a day is fast enough for the GTM need is a product call rather than mine, so it is written down instead of hidden: the `Risks` bullet now states the up-to-a-day recovery explicitly and names the delayed-follow-up as the fix if that is too slow. Happy to file it if you want the tighter bound.

Changed here:

- `usage-update.ts:64` — the no-organization branch now says the 1h is a floor so the next trigger is not blocked for 24h, and names the three triggers. Its log line changes from "applying 1h cooldown" to "not consuming the 24h window", which is what it actually did.
- The failure branch below carries the same clarification.
- `process-usage-telemetry-job.ts` — the docblock said a rejected update "is retried on that cooldown instead of through BullMQ attempts", which read as a schedule. It now says the retry is whichever trigger calls in next.
- The PR body's *Risks* bullet claiming hourly retries on persistent non-2xx, and the one claiming hourly retries until an org exists, are both corrected.

apps/web/lib/jobs/recurring-registrations.ts
Dhruwang: Confirmed and fixed in dc94fd8be. I read the lua rather than taking it on trust, and it is exactly as you describe:

```lua
if hadPrevJob then
    -- The jobs has been removed and we want to replace it, so lets use the same millis.
    nextMillis = prevMillis
```

`hadPrevJob` is set by `removeJobFromScheduler` whenever the upsert found a pending delayed/prioritized/waiting job for this scheduler — which is every boot after the first, because the daily cron always has its next iteration sitting in `delayed`. So the `now` that `defaultRepeatStrategy` returns for `immediately` is computed and then thrown away. My original reading stopped at the JS and never got to the lua, which is how a wrong claim ended up load-bearing in the description.

Two things follow, and both are now written down.

**Who benefits is not who I said, but it is still the right set.** `usageTelemetry` is a *new* scheduler, so the first boot after this ships has no `prevMillis` for it on any instance — every upgrading instance fires the immediate run once, and those have an organization, so they report. Fresh installs fire it before an org exists and get nothing, which is the case in your other thread.

**What actually covers "not up at 02:15" is the overdue iteration, not `immediately`** — the thing you checked and I had not leaned on. That is now the load-bearing sentence in both the registration comment and the description, with `immediately` demoted to what it really is: a first-registration nicety.

Two consequences you flagged, both corrected:

- The *Risks* bullet about crash-looping pods queueing a job per restart is gone — it cannot happen, since restarts after the first re-add the pending iteration at its existing timestamp.
- The *Open gaps* bullet no longer presents the per-boot semantics as verified-by-reading. It now says which two files the behaviour comes from and that neither half has been watched against a real Redis.

No code change — the schedule keeps `cronPattern` + `immediately`, and `instrumentation-jobs.test.ts` still pins that payload.

BhagyaAmarasinghe:
@Dhruwang please fix the merge conflicts and its good to go

BhagyaAmarasinghe (approved):

PR #8930, opened 2026-08-20
fix(storage): delete uploaded files when a survey's responses are reset

jobenjada:
## What & why

A customer asked whether deleting a response through the management API is
supposed to leave its uploaded files in storage — theirs stayed downloadable
with an API key. Chasing that down turned up a related leak that is still
live on `main`.

`deleteResponsesAndDisplaysForSurvey` — the "delete all responses" / reset
action on the survey summary page — wipes every response row for a survey with
`prisma.response.deleteMany` and never touches storage. Every file-upload
answer in those responses becomes an orphaned object: no longer referenced by
any response, still served by `/storage/...` to anyone with an API key. This is
the bulk counterpart of the single-response cleanup that
`findAndDeleteUploadedFilesInResponse` has always done in
`lib/response/service.ts`, and that #8861 just hardened.

The fix:

- Collect the file URLs **before** the rows are deleted — they only exist
  inside `response.data`, so after `deleteMany` there is nothing left to tell
  storage which objects went unreferenced.
- Hand them to `deleteResponseFileUrls` (added in #8861) **after** the delete
  commits, so a storage failure can never remove files whose responses
  survived, and so workspace scoping is enforced there rather than re-derived
  here.
- Build the file-upload id set from the union of `blocks` and `questions` via
  `getSurveyFileUploadConfigs`, matching the single-response path — keying off
  one shape silently skips surveys stored in the other.
- Page the response scan (500 at a time) so resetting a survey with a large
  response count doesn't hold every `data` blob at once, and skip the scan
  entirely when the survey has no file-upload element. Ordering is
  `(createdAt, id)` with an explicit keyset predicate rather than a cursor on
  `id`, so the scan rides the existing `@@index([surveyId, createdAt])` — there
  is no `(surveyId, id)` index, and ordering by `id` alone would have made the
  paging cost grow with the very input it was added to bound.
- Issue the storage deletes in chunks of 100. `deleteResponseFileUrls` fans out
  with `Promise.all` over everything it is handed and this is its first bulk
  caller, so an unchunked call on a survey with many uploads would open one
  storage request per file. Paging bounds the response blobs; chunking bounds
  the fan-out they feed.
- Read the survey through the `reactCache`d `getSurvey` rather than a second raw
  `findUnique`: the reset action fetched the same survey immediately before, so
  this is a cache hit, and it returns typed `blocks`/`questions` instead of raw
  JSON columns needing a cast.
- Log and continue if cleanup throws. The rows are already committed as deleted
  by then, so a storage problem must not surface as a failed reset the caller
  would retry.

Non-array values under a file-upload key are skipped rather than cast, so
malformed data can't produce a bogus delete target.

## Breaking changes

Not breaking: no API shape, route, env var or default changes. The reset action
returns the same counts as before; only the storage side effect is added.

## Migrations & env

- none

Note this does not retroactively clean objects orphaned by past resets — it
only stops new ones. Existing orphans would need a separate sweep.

## How this was tested

All rows are in
`apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts`.
Rerun one with `cd apps/web && pnpm vitest run "app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts" -t "<test name>"`.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Uploaded files of all deleted responses reach storage cleanup | unit (red on main) | `Deletes the uploaded files held by the deleted responses` — asserts the exact URL list and workspace id |
| Scan precedes the row delete, storage cleanup follows it | unit (red on main) | `Reads the file-upload answers before the responses are deleted` — asserts the exact `["scan", "delete", "storage"]` order |
| Files on a later page are still collected, via an index-friendly keyset | unit (red on main) | `Collects files from every page when responses span the scan page size` — full 500-row first page forces a second query; asserts the `(createdAt, id)` keyset `where`, the `orderBy`, the absence of `skip`/`cursor`, and that all 501 URLs arrive |
| Storage fan-out stays bounded | unit (red on main) | `Issues storage deletes in bounded chunks` — 250 URLs arrive as `[100, 100, 50]`, so no single fan-out exceeds the chunk size |
| A committed reset is not failed by broken storage | unit (red on main) | `Reports the reset as successful when storage cleanup fails` — counts still return, error is logged |
| A failed row delete leaves storage untouched | unit (guard) | `Throws DatabaseError on PrismaClientKnownRequestError occurrence` — now also asserts `deleteResponseFileUrls` was never called, locking the converse of the ordering guarantee |
| No file-upload element → no response scan, no storage call | unit (guard) | `Skips the response scan when the survey has no file-upload element` |
| Missing survey → no scan, no storage call | unit (guard) | `Skips storage cleanup when the survey no longer exists` |
| Non-array answer under a file-upload id is ignored | unit (guard) | `Ignores non-array answers stored under a file-upload element id` — `"not-an-array"` and `[42, null]` produce no storage call |
| Existing reset behaviour (counts, error mapping) | unit (guard) | `Deletes all responses and displays for a survey`, `Handles case with no responses or displays to delete`, `Throws a generic Error for other exceptions` — pre-existing, still green |

Red-on-main verified by restoring `origin/main`'s `survey.ts` under the new test
file: **5 failed / 11 passed**. With the fix: **16 passed**. Full `apps/web`
suite: **627 files, 8297 tests passed**. `tsc --noEmit` clean on the changed
files, eslint and prettier clean.

**Open gaps**

- Not exercised against a live S3/MinIO bucket — `deleteResponseFileUrls` is
  mocked at the module boundary, so this PR proves the URLs reach it with the
  right workspace id, not that the objects disappear. That helper's own S3
  behaviour is covered by `delete-response-files.test.ts` from #8861.
- No e2e coverage of the reset button itself.
- **A file-upload element deleted from a live survey still orphans its files.**
  Its answers stay in `response.data` but its id is gone from
  `getSurveyFileUploadConfigs`, so a later reset skips them (and the
  no-file-upload fast path skips the scan entirely). This is parity with the
  single-response path, not a regression introduced here. Worth stating that
  broadening the key set to "any answer that parses as a storage URL" is *not*
  the fix: `delete-response-files.ts` validates workspace, not survey, so a
  respondent could plant another survey's file URL in a text answer and have an
  admin's reset delete a live file. The narrow keying is deliberate.
- The extraction rule ("build the id set, pull the URLs out of `data`") now
  exists in three places — this file, `lib/response/service.ts` and
  `modules/api/v2/management/responses/[responseId]/lib/utils.ts` — and they have
  already drifted: the two older copies do a bare `elementResponse as string[]`
  with no array guard, so a malformed non-array answer that this file skips still
  reaches storage through them. Extracting one
  `collectResponseFileUrls(data, survey)` next to `deleteResponseFileUrls` would
  fix the drift and give the older paths the guard. Deliberately left out of this
  diff so the leak fix does not carry a refactor of two working paths; happy to
  do it as an immediate follow-up.
- Chunking bounds concurrent storage requests, not total wall-clock. A reset on
  a survey with a very large number of uploads still issues one `DeleteObject`
  per file and could outlive a platform timeout, leaving orphans with no retry
  path. S3's `DeleteObjects` (1000 keys per request) or a durable cleanup queue
  is the real answer at that scale; both are larger changes than this fix and
  neither is needed to close the leak.

**Risks**

- The scan adds a `getSurvey` call (request-cache hit in the action's flow) plus,
  for surveys that actually have a file-upload element, a paged
  `response.findMany` before the delete. Surveys without file uploads add no
  queries beyond the cached survey read. If a reset on a very large survey feels
  slower, `RESPONSE_FILE_SCAN_PAGE_SIZE` and `STORAGE_DELETE_CHUNK_SIZE` are the
  knobs.
- Responses created between the scan and the `deleteMany` have their rows
  deleted but their files kept — the same narrow race the single-response path
  has, and it fails toward keeping files rather than deleting live ones.

## Related, not fixed here

`deleteSurvey` (`modules/survey/lib/surveys.ts`) hard-deletes a survey and
cascades its responses without any storage cleanup, so permanently deleting a
survey — including via the archive purge cron — orphans the same files. Left
out to keep this diff to the response-deletion path; happy to follow up if you
want it in the same cycle.

---

> [!NOTE]
> **AI model used** — `claude-opus-5`, reasoning effort `medium`.

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts
@@ -4,6 +4,7 @@ import {
   SurveyFileUploadFields,
   fileUploadElement,
   responseWithFiles,
+  scanTimestamp,
   storageUrl,
   surveyId,
   surveyWithFileUpload,
@@ -78,6 +79,7 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
       mockResponsePages([
         {
           id: "response-1",
+          createdAt: scanTimestamp(0),
           data: {
             [fileUploadElement.id]: [storageUrl("file1.png"), storageUrl("file2.pdf")],
             "other-element": "not a file",
@@ -124,9 +126,9 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
       // A full first page (500) forces a second cursor-based query; the file on the later page must
       // still reach storage cleanup.
       const firstPage = Array.from({ length: 500 }, (_, index) =>
-        responseWithFiles(`response-${index}`, [`page1-${index}.png`])
+        responseWithFiles(`response-${index}`, [`page1-${index}.png`], index)
       );
-      const secondPage = [responseWithFiles("response-500", ["page2.png"])];
+      const secondPage = [responseWithFiles("response-500", ["page2.png"], 500)];
 
       mockSurvey(surveyWithFileUpload);
       mockResponsePages(firstPage, secondPage);
@@ -138,11 +140,22 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
       // so exactly two queries are expected here.
       expect(prisma.response.findMany).toHaveBeenCalledTimes(2);
 
-      // The second query pages past the last id of the first page.
-      expect(vi.mocked(prisma.response.findMany).mock.calls[1][0]).toMatchObject({
-        skip: 1,
-        cursor: { id: "response-499" },
+      // The second query pages past the last row of the first page with a (createdAt, id) keyset, and
+      // orders by createdAt so it can use the existing (surveyId, createdAt) index.
+      const secondQuery = vi.mocked(prisma.response.findMany).mock.calls[1][0];
+      expect(secondQuery).toMatchObject({
+        orderBy: [{ createdAt: "asc" }, { id: "asc" }],
+        where: {
+          surveyId,
+          OR: [
+            { createdAt: { gt: scanTimestamp(499) } },
+            { createdAt: scanTimestamp(499), id: { gt: "response-499" } },
+          ],
+        },
       });
+      // Keyset paging replaces cursor/skip entirely — a leftover offset would double-read rows.
+      expect(secondQuery).not.toHaveProperty("skip");
+      expect(secondQuery).not.toHaveProperty("cursor");
 
       const deletedUrls = deleteResponseFileUrls.mock.calls.flatMap(([urls]) => urls);
       expect(deletedUrls).toHaveLength(501);
@@ -187,9 +200,13 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
     test("Ignores non-array answers stored under a file-upload element id", async () => {
       mockSurvey(surveyWithFileUpload);
       mockResponsePages([
-        { id: "response-1", data: { [fileUploadElement.id]: "not-an-array" } },
+        { id: "response-1", createdAt: scanTimestamp(0), data: { [fileUploadElement.id]: "not-an-array" } },
         // Numbers and nulls inside the array are dropped rather than cast to a delete target.
-        { id: "response-2", data: { [fileUploadElement.id]: [42, null] as unknown as string[] } },
+        {
+          id: "response-2",
+          createdAt: scanTimestamp(1),
+          data: { [fileUploadElement.id]: [42, null] as unknown as string[] },
+        },
       ]);
       vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 2 }, { count: 0 }]);
 
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts
@@ -1,66 +1,54 @@
+import { deleteResponseFileUrls } from "./__mocks__/delete-response-files.mock";
+import {
+  ScannedResponse,
+  SurveyFileUploadFields,
+  fileUploadElement,
+  responseWithFiles,
+  storageUrl,
+  surveyId,
+  surveyWithFileUpload,
+  surveyWithoutFileUpload,
+  workspaceId,
+} from "./__mocks__/survey-reset.mock";
+import { getSurvey } from "./__mocks__/survey-service.mock";
+import { prisma } from "@/lib/__mocks__/database";
 import { beforeEach, describe, expect, test, vi } from "vitest";
-import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
 import { PrismaErrorType } from "@formbricks/database/types/error";
+import { logger } from "@formbricks/logger";
 import { DatabaseError } from "@formbricks/types/errors";
-import { deleteResponseFileUrls } from "@/modules/storage/lib/delete-response-files";
+import { TSurvey } from "@formbricks/types/surveys/types";
 import { deleteResponsesAndDisplaysForSurvey, getQuotasSummary } from "./survey";
 
-// Mock prisma
-vi.mock("@formbricks/database", () => ({
-  prisma: {
-    response: {
-      deleteMany: vi.fn(),
-      findMany: vi.fn(),
-    },
-    display: {
-      deleteMany: vi.fn(),
-    },
-    survey: {
-      findUnique: vi.fn(),
-    },
-    $transaction: vi.fn(),
-    surveyQuota: {
-      findMany: vi.fn(),
-    },
-  },
-}));
+/**
+ * The fixtures declare only the three fields the service reads, so the hand-off to `getSurvey` (typed
+ * `TSurvey | null`) is cast once here. The fixture fields themselves stay typed against
+ * `@formbricks/types`, so a wrong block or element shape still fails typecheck.
+ */
+const mockSurvey = (survey: SurveyFileUploadFields | null) => {
+  getSurvey.mockResolvedValue(survey as TSurvey | null);
+};
 
-vi.mock("@/modules/storage/lib/delete-response-files", () => ({
-  deleteResponseFileUrls: vi.fn(),
-}));
-
-const surveyId = "clq5n7p1q0000m7z0h5p6g3r2";
-const workspaceId = "u8qa6u0tlxb6160pi2jb8s4p";
-const fileUploadElementId = "y3ydd3td2iq09wa599cxo1me";
-
-const fileUploadSurvey = {
-  workspaceId,
-  questions: [],
-  blocks: [
-    {
-      id: "block-1",
-      elements: [{ id: fileUploadElementId, type: "fileUpload" }],
-    },
-  ],
+const mockResponsePages = (...pages: ScannedResponse[][]) => {
+  const findMany = vi.mocked(prisma.response.findMany);
+  findMany.mockReset();
+  for (const page of pages) {
+    findMany.mockResolvedValueOnce(page as never);
+  }
+  // Anything past the configured pages reads as "no more rows".
+  findMany.mockResolvedValue([] as never);
 };
 
 beforeEach(() => {
-  vi.resetModules();
-  vi.resetAllMocks();
   // Default: a survey with no file-upload element, so the response scan is skipped.
-  vi.mocked(prisma.survey.findUnique).mockResolvedValue({
-    workspaceId,
-    questions: [],
-    blocks: [],
-  } as any);
+  mockSurvey(surveyWithoutFileUpload);
+  deleteResponseFileUrls.mockReset();
+  deleteResponseFileUrls.mockResolvedValue(undefined);
 });
 
 describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
   describe("Happy Path", () => {
     test("Deletes all responses and displays for a survey", async () => {
-      const { prisma } = await import("@formbricks/database");
-
       // Mock $transaction to return the results directly
       vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 5 }, { count: 3 }]);
 
@@ -74,8 +62,6 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
     });
 
     test("Handles case with no responses or displays to delete", async () => {
-      const { prisma } = await import("@formbricks/database");
-
       // Mock $transaction to return zero counts
       vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 0 }, { count: 0 }]);
 
@@ -88,33 +74,24 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
     });
 
     test("Deletes the uploaded files held by the deleted responses", async () => {
-      vi.mocked(prisma.survey.findUnique).mockResolvedValue(fileUploadSurvey as any);
-      vi.mocked(prisma.response.findMany).mockResolvedValue([
+      mockSurvey(surveyWithFileUpload);
+      mockResponsePages([
         {
           id: "response-1",
           data: {
-            [fileUploadElementId]: [
-              `https://example.com/storage/${workspaceId}/private/file1.png`,
-              `https://example.com/storage/${workspaceId}/private/file2.pdf`,
-            ],
+            [fileUploadElement.id]: [storageUrl("file1.png"), storageUrl("file2.pdf")],
             "other-element": "not a file",
           },
         },
-        {
-          id: "response-2",
-          data: { [fileUploadElementId]: [`https://example.com/storage/${workspaceId}/private/file3.png`] },
-        },
-      ] as any);
+        responseWithFiles("response-2", ["file3.png"]),
+      ]);
       vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 2 }, { count: 0 }]);
 
       await deleteResponsesAndDisplaysForSurvey(surveyId);
 
+      expect(deleteResponseFileUrls).toHaveBeenCalledTimes(1);
       expect(deleteResponseFileUrls).toHaveBeenCalledWith(
-        [
-          `https://example.com/storage/${workspaceId}/private/file1.png`,
-          `https://example.com/storage/${workspaceId}/private/file2.pdf`,
-          `https://example.com/storage/${workspaceId}/private/file3.png`,
-        ],
+        [storageUrl("file1.png"), storageUrl("file2.pdf"), storageUrl("file3.png")],
         workspaceId
       );
     });
@@ -122,29 +99,70 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
     test("Reads the file-upload answers before the responses are deleted", async () => {
       const callOrder: string[] = [];
 
-      vi.mocked(prisma.survey.findUnique).mockResolvedValue(fileUploadSurvey as any);
-      vi.mocked(prisma.response.findMany).mockImplementation((async () => {
+      mockSurvey(surveyWithFileUpload);
+      vi.mocked(prisma.response.findMany).mockReset();
+      vi.mocked(prisma.response.findMany).mockImplementation((() => {
         callOrder.push("scan");
-        return [
-          {
-            id: "response-1",
-            data: { [fileUploadElementId]: [`https://example.com/storage/${workspaceId}/private/f.png`] },
-          },
-        ];
-      }) as any);
-      vi.mocked(prisma.$transaction).mockImplementation((async () => {
+        return Promise.resolve([responseWithFiles("response-1", ["f.png"])]) as never;
+      }) as never);
+      vi.mocked(prisma.$transaction).mockImplementation((() => {
         callOrder.push("delete");
-        return [{ count: 1 }, { count: 0 }];
-      }) as any);
-      vi.mocked(deleteResponseFileUrls).mockImplementation((async () => {
+        return Promise.resolve([{ count: 1 }, { count: 0 }]) as never;
+      }) as never);
+      deleteResponseFileUrls.mockImplementation(async () => {
         callOrder.push("storage");
-      }) as any);
+      });
 
       await deleteResponsesAndDisplaysForSurvey(surveyId);
 
       // The scan must precede the row delete (the URLs live in response.data), and storage cleanup must
       // follow it so files are never removed while their responses survive.
       expect(callOrder).toEqual(["scan", "delete", "storage"]);
+    });
+
+    test("Collects files from every page when responses span the scan page size", async () => {
+      // A full first page (500) forces a second cursor-based query; the file on the later page must
+      // still reach storage cleanup.
+      const firstPage = Array.from({ length: 500 }, (_, index) =>
+        responseWithFiles(`response-${index}`, [`page1-${index}.png`])
+      );
+      const secondPage = [responseWithFiles("response-500", ["page2.png"])];
+
+      mockSurvey(surveyWithFileUpload);
+      mockResponsePages(firstPage, secondPage);
+      vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 501 }, { count: 0 }]);
+
+      await deleteResponsesAndDisplaysForSurvey(surveyId);
+
+      // Third call returns [] and ends the loop: 500 == page size, then 1 < page size would stop it,
+      // so exactly two queries are expected here.
+      expect(prisma.response.findMany).toHaveBeenCalledTimes(2);
+
+      // The second query pages past the last id of the first page.
+      expect(vi.mocked(prisma.response.findMany).mock.calls[1][0]).toMatchObject({
+        skip: 1,
+        cursor: { id: "response-499" },
+      });
+
+      const deletedUrls = deleteResponseFileUrls.mock.calls.flatMap(([urls]) => urls);
+      expect(deletedUrls).toHaveLength(501);
+      expect(deletedUrls).toContain(storageUrl("page1-0.png"));
+      expect(deletedUrls).toContain(storageUrl("page2.png"));
+    });
+
+    test("Issues storage deletes in bounded chunks", async () => {
+      const responses = Array.from({ length: 250 }, (_, index) =>
+        responseWithFiles(`response-${index}`, [`file-${index}.png`])
+      );
+
+      mockSurvey(surveyWithFileUpload);
+      mockResponsePages(responses);
+      vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 250 }, { count: 0 }]);
+
+      await deleteResponsesAndDisplaysForSurvey(surveyId);
+
+      // 250 URLs at a chunk size of 100 => 100 + 100 + 50, so no single storage fan-out exceeds 100.
+      expect(deleteResponseFileUrls.mock.calls.map(([urls]) => urls.length)).toEqual([100, 100, 50]);
     });
 
     test("Skips the response scan when the survey has no file-upload element", async () => {
@@ -156,12 +174,23 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
       expect(deleteResponseFileUrls).not.toHaveBeenCalled();
     });
 
+    test("Skips storage cleanup when the survey no longer exists", async () => {
+      mockSurvey(null);
+      vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 0 }, { count: 0 }]);
+
+      await deleteResponsesAndDisplaysForSurvey(surveyId);
+
+      expect(prisma.response.findMany).not.toHaveBeenCalled();
+      expect(deleteResponseFileUrls).not.toHaveBeenCalled();
+    });
+
     test("Ignores non-array answers stored under a file-upload element id", async () => {
-      vi.mocked(prisma.survey.findUnique).mockResolvedValue(fileUploadSurvey as any);
-      vi.mocked(prisma.response.findMany).mockResolvedValue([
-        { id: "response-1", data: { [fileUploadElementId]: "not-an-array" } },
-        { id: "response-2", data: { [fileUploadElementId]: [42, null] } },
-      ] as any);
+      mockSurvey(surveyWithFileUpload);
+      mockResponsePages([
+        { id: "response-1", data: { [fileUploadElement.id]: "not-an-array" } },
+        // Numbers and nulls inside the array are dropped rather than cast to a delete target.
+        { id: "response-2", data: { [fileUploadElement.id]: [42, null] as unknown as string[] } },
+      ]);
       vi.mocked(prisma.$transaction).mockResolvedValue([{ count: 2 }, { count: 0 }]);
 
       await deleteResponsesAndDisplaysForSurvey(surveyId);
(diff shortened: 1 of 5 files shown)


apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts
```diff
@@ -23,6 +111,13 @@ export const deleteResponsesAndDisplaysForSurvey = async (
       }),
     ]);
 
+    // Runs after the rows are gone so a storage failure can never delete files whose responses
+    // survived. deleteResponseFileUrls logs and swallows per-file errors, so a storage outage does not
+    // fail the reset — it leaves objects behind, which is the pre-existing behaviour, not a new one.
+    if (fileUrls.length > 0) {
+      await deleteResponseFileUrls(fileUrls, workspaceId);
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts]: **Unbounded fan-out: the whole survey's URL list goes into one `Promise.all`.**

`deleteResponseFileUrls` maps every URL through `Promise.all` with no concurrency cap (`modules/storage/lib/delete-response-files.ts:47`). That was fine for its two existing callers — both single-response, so a handful of URLs. This caller hands it the union of *every* response in the survey.

A survey with 20k responses and two upload answers each produces 40k concurrent `deleteFile` calls. Practically that means socket exhaustion, S3 throttling, mass per-file `logger.error` — and because the storage call is awaited inside the server action, a reset that has *already committed* the row delete hangs or times out. It fails in the direction of "rows gone, files kept, user sees an error", which is the outcome this PR exists to prevent.

Two things worth doing:

1. **Dedupe.** `fileUrls` can contain the same URL many times across responses; each duplicate is a redundant `deleteFile`. `new Set(fileUrls)` is nearly free here.
2. **Bound the concurrency**, either by chunking at the call site or by capping inside `deleteResponseFileUrls` (better — it fixes both existing callers too and keeps the batching concern next to the S3 code).

Given the row delete is already committed by this point, the stronger option is to hand the URLs to `packages/jobs` and let the reset return immediately. That also removes the timeout risk entirely and gives you retries on a storage outage, instead of the current "log and leave it orphaned". Reasonable to defer if you'd rather keep the diff tight, but the bare `Promise.all` at survey scale shouldn't ship as-is.

---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts
```diff
@@ -2,12 +2,100 @@ import "server-only";
...
+  const fileUrls: string[] = [];
+  let cursor: string | undefined;
+
+  for (;;) {
+    const responses = await prisma.response.findMany({
+      where: { surveyId },
+      select: { id: true, data: true },
+      orderBy: { id: "asc" },
+      take: RESPONSE_FILE_SCAN_PAGE_SIZE,
+      ...(cursor ? { skip: 1, cursor: { id: cursor } } : {}),
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts]: **`orderBy: { id: "asc" }` has no supporting index.**

`Response` carries `@@unique([surveyId, singleUseId])`, `@@index([createdAt])`, `@@index([surveyId, createdAt])`, `@@index([contactId, createdAt])` (`packages/database/schema/main.prisma:187-190`) — nothing on `(surveyId, id)`. Neither available plan is good, and both repeat per page:

- planner uses the `(surveyId, …)` index for the filter → it must materialise and sort the survey's entire response set on every page, so the paging loop is quadratic in response count;
- planner uses the PK for the ordering → it scans the table by `id` filtering `surveyId`, cost scaling with the whole `Response` table rather than this survey.

Either way the paging you added to bound *memory* trades it for a query cost that grows with the thing you're trying to protect against. Two fixes:

- `orderBy: [{ createdAt: "asc" }, { id: "asc" }]` with a composite cursor — rides the existing `(surveyId, createdAt)` index, no migration; or
- add `@@index([surveyId, id])` and keep the current ordering.

The first is cheaper to land. Worth mentioning in `Migrations & env` if you take the second.

---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts
```diff
@@ -2,12 +2,100 @@ import "server-only";
...
+  const fileUploadElementIds = new Set(
+    getSurveyFileUploadConfigs({
+      blocks: survey.blocks as TSurveyBlock[],
+      questions: survey.questions as TSurveyQuestion[],
+    }).map((config) => config.id)
+  );
+
+  // A survey with no file-upload element can have no uploads to clean up, so skip the response scan.
+  if (fileUploadElementIds.size === 0) {
+    return { fileUrls: [], workspaceId: survey.workspaceId };
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts:107]: **The comment claims more than the code delivers, and the gap is the exact leak this PR is fixing.**

> `// A survey with no file-upload element can have no uploads to clean up, so skip the response scan.`

That holds for the survey's *current* definition, not for its responses. `getSurveyFileUploadConfigs` reads today's `blocks`/`questions`, but `response.data` is historical: an upload element that was later deleted (or removed and re-added, getting a fresh id) leaves answers under a key that is no longer in the set. Those URLs are skipped, the rows are deleted anyway, and the objects are orphaned permanently — with no element left at all, the early return skips the scan entirely and *every* response goes with zero cleanup.

Editing questions on a survey with responses is normal usage, so this isn't a corner case. It's the same shape of bug as the blocks/questions split you fixed here.

Note the id filter isn't a security boundary — `deleteResponseFileUrls` re-resolves each URL's storage id and drops anything outside the survey's workspace, precisely because it can't trust URLs that came from response data (see its doc comment). So loosening the filter doesn't widen the blast radius.

That said, don't just drop it: multi-select answers are also `string[]`, so an unfiltered scan would feed choice labels to storage. Filter on *shape* instead — keep strings that `parseStorageFileUrl` accepts, whatever key they sit under. That covers removed elements and stays narrow.

Cleanest fix for this and the two comments above at once: replace the scan + `deleteMany` with a single `DELETE FROM "Response" WHERE "surveyId" = $1 RETURNING data` (`$queryRaw`, inside the existing transaction). One statement, no paging, no cursor, no index question, no scan/delete race — and `data` comes back from the rows actually deleted, so historical keys are included by construction. Worth considering before adding the index.

At minimum, fix the comment so it doesn't assert a guarantee the code doesn't make.

---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts
```diff
@@ -2,12 +2,100 @@ import "server-only";
...
+  const survey = await prisma.survey.findUnique({
+    where: { id: surveyId },
+    select: { blocks: true, questions: true, workspaceId: true },
+  });
+
+  if (!survey) {
+    return { fileUrls: [], workspaceId: undefined };
+  }
+
+  const fileUploadElementIds = new Set(
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.ts]: **Third copy of "build the id set, pull the URLs out of `data`" — and the copies have already drifted.**

The other two:

- `apps/web/modules/api/v2/management/responses/[responseId]/lib/utils.ts:14`
- `apps/web/lib/response/service.ts:632`

All three build the same `Set` from `getSurveyFileUploadConfigs(...).map(c => c.id)` and carry a near-identical explanatory comment about the blocks/questions union. The drift is already visible: this PR guards with `Array.isArray(answer)` and filters to strings, while both older copies do a bare `elementResponse as string[]` and `flatMap` it — so a malformed non-array answer that this file correctly skips still reaches storage through the other two paths. The next fix to the extraction rule has to be made in three places, and the odds of all three staying in sync are what this diff already demonstrates.

Worth extracting one helper next to `deleteResponseFileUrls` — something like `collectResponseFileUrls(data, survey): string[]` in `modules/storage/` — and having all three call it. Cheap now (the two older call sites are one-liners), and it puts the shape filter from the comment above in a single place. Doing it as part of this PR would also let the older paths inherit the array guard.

---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts
```diff
@@ -58,6 +86,88 @@ describe("Tests for deleteResponsesAndDisplaysForSurvey service", () => {
         deletedDisplaysCount: 0,
       });
     });
+
+    test("Deletes the uploaded files held by the deleted responses", async () => {
+      vi.mocked(prisma.survey.findUnique).mockResolvedValue(fileUploadSurvey as any);
+      vi.mocked(prisma.response.findMany).mockResolvedValue([
```
Comment (kiln-agent) [inline apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/summary/lib/survey.test.ts]: **Asking for the coverage listed under `Open gaps`: the cursor loop.**

You flag this yourself ("The paging loop is verified only through a single-page mock; multi-page cursor advancement is not covered by a test"), and per AGENTS.md that's my cue to push back rather than wave it through — the paging loop is the riskiest new code in the diff, and it's the one part no test touches.

Every `findMany` mock here returns fewer than `RESPONSE_FILE_SCAN_PAGE_SIZE` rows, so `responses.length < PAGE_SIZE` breaks on the first iteration and the `cursor` / `skip: 1` branch never executes. An off-by-one in the cursor — or a `skip: 1` that silently re-reads or drops a row — would lose URLs for every survey over 500 responses and all six tests would still be green. That's a bug class that only shows up on exactly the large surveys the paging exists for.

A two-page mock is cheap and doesn't need 500 real rows: export `RESPONSE_FILE_SCAN_PAGE_SIZE` (or thread it in) and drive the mock off it — first call returns `PAGE_SIZE` rows, second returns 2, then assert both that `findMany` was called twice with the second call carrying `cursor: { id: <last id of page 1> }, skip: 1`, and that URLs from *both* pages reach `deleteResponseFileUrls`.

The other two gaps you list (no live S3, no e2e on the reset button) I'm fine with as stated — `deleteResponseFileUrls` has its own S3 coverage from #8861, and mocking at that module boundary is the right seam here.

---

kiln-agent (dismissed):
Reviewed `749f8be`. The bug is real and the diagnosis is right: `deleteResponsesAndDisplaysForSurvey` really does `deleteMany` its way through every response without touching storage, and the core shape of the fix — scan before delete, storage cleanup after the commit, id set from the union of `blocks` and `questions`, `Array.isArray` guard instead of a cast — is the correct design. The PR description is unusually good about its own limits, and the red-on-main verification is the right way to prove a cleanup fix.

What I'd want addressed before merge, all of it about the jump from one response to a whole survey:

1. **Unbounded `Promise.all` at survey scale** (`survey.ts:118`) — `deleteResponseFileUrls` was written for single-response callers. Handing it a survey's entire URL list means tens of thousands of concurrent S3 deletes on a large survey, awaited inside a server action *after* the rows are already committed. Needs deduping and a concurrency bound; a `packages/jobs` handoff would be better still.
2. **`orderBy: { id: "asc" }` is unindexed** (`survey.ts:56-62`) — no `(surveyId, id)` index exists, so each page either sorts the survey's full response set or scans the table by PK. The paging bounds memory but hands back a query cost that grows with response count.
3. **Historical element ids are missed** (`survey.ts:47-50`) — the id set comes from the survey's current definition, so uploads under a since-deleted element are never collected, and a survey with no upload element left skips the scan entirely and orphans everything. The comment asserts the opposite. Note the id filter isn't the security boundary — `deleteResponseFileUrls` re-validates workspace ownership per URL — but the fix is to filter on storage-URL *shape*, not to drop the filter (multi-select answers are `string[]` too).
4. **Third copy of the extraction logic** (`survey.ts:35-41`) — and it's already drifted: this file guards with `Array.isArray`, the two older copies still do `as string[]`. Worth one shared helper.
5. **The paging loop has no test** (`survey.test.ts:92`) — you list it under `Open gaps`; it's also the riskiest new code, and a cursor bug would lose URLs on exactly the large surveys paging exists for while leaving all six tests green. A two-page mock covers it.

Worth considering before you do 1–3 individually: a single `DELETE FROM "Response" WHERE "surveyId" = $1 RETURNING data` collapses the scan and the delete into one statement, which removes the paging loop, the index question, the scan/delete race you note under `Risks`, and the historical-id gap all at once. The concurrency bound in (1) is still needed either way.

Also agreed on scoping `deleteSurvey` out — the same orphaning applies there via the archive purge cron, and it's a fair follow-up rather than something to grow this diff with.

---

kiln-agent:
## Review

The leak is real and the fix is in the right place. I reproduced the claims rather than taking them on trust:

- `origin/main`'s `survey.ts` + this PR's test file → **2 failed / 10 passed** (`Deletes the uploaded files…` and `Reads the file-upload answers before…`). Head → **12 passed**. The red-on-main claim holds exactly as described.
- `tsc --noEmit` over `apps/web`: no error in either changed file (the 159 errors in the project are pre-existing, all in `playwright/*.spec.ts` and `proxy.test.ts`).
- Ordering (scan → `deleteMany` → storage), the `skip: 1` cursor idiom, the `Array.isArray` guard, and sourcing the id set from `getSurveyFileUploadConfigs` (union of `blocks` + `questions`) all check out. `ZSurveyBlock.elements` is flat, so the `flatMap` can't miss nested elements.

Nothing here blocks merge. Four things worth acting on, roughly in order of value.

### 1. The paging bounds the `data` blobs, but not the fan-out it feeds

`RESPONSE_FILE_SCAN_PAGE_SIZE` keeps every `response.data` blob out of memory at once — but `fileUrls` accumulates across all pages, and it is then handed to `deleteResponseFileUrls` in one call, which does:

```ts
await Promise.all(fileUrls.map(async (fileUrl) => { … await deleteFile(…) }))
```

The three existing callers are all single-response paths (`lib/response/service.ts:646`, `modules/api/v2/management/responses/[responseId]/lib/utils.ts:26`), so that unbounded `Promise.all` has only ever seen a handful of files. This PR is its first bulk caller. Reset a survey with 100k file-upload answers and it becomes 100k live promises and 100k queued `DeleteObject` calls. `createS3ClientFromEnv` sets no `requestHandler`, so the SDK default caps sockets at 50 — the socket layer won't blow up, but wall-clock scales linearly and the server action can hit its platform timeout partway through, leaving orphans with no retry path (the rows are already gone).

Chunking the storage call would keep the second axis bounded too:

```ts
for (let i = 0; i < fileUrls.length; i += STORAGE_DELETE_BATCH_SIZE) {
  await deleteResponseFileUrls(fileUrls.slice(i, i + STORAGE_DELETE_BATCH_SIZE), workspaceId);
}
```

(Costs one `findWorkspaceByIdOrLegacyEnvId` per distinct storage id per batch, since that cache is per call — bounded and cheap.) Longer term, S3's `DeleteObjects` takes 1000 keys per request, which is where the real win is. Either way, the `RESPONSE_FILE_SCAN_PAGE_SIZE` doc comment currently reads as if memory is fully bounded; it's bounded on one axis.

### 2. The extra `survey.findUnique` duplicates a fetch the caller already made

`actions.ts:110` calls `getSurvey(parsedInput.surveyId)` (from `@/lib/survey/service`, `reactCache`d) immediately before `deleteResponsesAndDisplaysForSurvey`, and `selectSurvey` already includes `workspaceId`, `questions` and `blocks` (`lib/survey/service.ts:47-52`). So `collectSurveyResponseFileUrls`'s raw `prisma.survey.findUnique` is a second round-trip for data already sitting in the request-scoped cache. `getSurvey(surveyId)` would be a cache hit, it drops the two `as TSurveyBlock[]` / `as TSurveyQuestion[]` casts, and it's what the single-response path this diff mirrors actually does (`findAndDeleteUploadedFilesInResponse` receives the survey from `getSurvey`). Trade-off to weigh: `getSurvey` throws `ResourceNotFoundError` on a missing survey instead of returning `{ fileUrls: [], workspaceId: undefined }` — harmless here, since the action already checks, but it is a behaviour change in the service and would need the test mocks reworked.

### 3. Two cheap test additions

- The sad-path test asserts the `DatabaseError` mapping but not that storage was left alone. The converse of the comment's guarantee — a failed row delete must never delete files — is the whole reason the storage call sits after the transaction, and nothing currently locks it. One line in the existing test: `expect(deleteResponseFileUrls).not.toHaveBeenCalled()`.
- Cursor advancement is the only new branch with no coverage, as `Open gaps` says. A two-page mock is a few lines and would catch a dropped `skip: 1` — which doesn't loop forever, but re-scans one response per page and produces duplicate delete attempts that surface only as `Failed to delete a response file from storage` noise.

### 4. One case missing from `Open gaps`

If a file-upload element is deleted from a live survey (the editor allows this on surveys with responses, behind the caution dialog), its answers stay in `response.data` but its id is gone from `getSurveyFileUploadConfigs` — so a later reset leaves those files behind, and the `size === 0` fast path skips the scan entirely. Same hole as the single-response path, so it's parity, not a regression.

Worth stating explicitly that broadening the key set to "any answer that parses as a storage URL" is **not** the fix: `delete-response-files.ts` validates workspace, not survey, so a respondent could plant another survey's file URL in a text answer and have an admin's reset delete a live file. The narrow keying is the right call — this just belongs in `Open gaps` rather than in code.

### Nit

`Fixes ENG-` at the top has no ticket id.

---

Scope call on `deleteSurvey` — flagging it in `Related, not fixed here` rather than widening this diff — is the right one, and the `Risks` note about the scan/delete race failing toward keeping files is accurate.

---

jobenjada:
Addressed in df3597c. Thanks both — the fan-out point was the real one and I'd missed that this is `deleteResponseFileUrls`'s first bulk caller.

**Taken**

- **Bounded fan-out** (CodeRabbit #4, kiln #1) — storage deletes now go out in chunks of 100, exactly the loop kiln sketched. Also fixed the `RESPONSE_FILE_SCAN_PAGE_SIZE` doc comment, which did read as if memory were fully bounded when it only bounds the `data` blobs.
- **`getSurvey` instead of a second `findUnique`** (kiln #2) — cache hit in the action's flow and it drops both JSON-column casts. Worth noting the trade-off you flagged doesn't apply: `getSurvey` returns `TSurvey | null` (`lib/survey/service.ts:169`), it doesn't throw, so the missing-survey branch is unchanged. Added a test for it.
- **Mocks into `__mocks__`** (CodeRabbit #1) — now uses the shared `@/lib/__mocks__/database` prisma mock, plus local `__mocks__` files for the storage and survey boundaries.
- **Typed fixtures** (CodeRabbit #2) — fixtures are typed against `@formbricks/types` (`TSurveyBlock`, `TSurveyFileUploadElement`, `TResponseData`), so a wrong element or response shape now fails typecheck. One cast remains where the fixture declares only the three fields the service reads and hands them to `getSurvey`; it's documented.
- **Pagination boundary** (CodeRabbit #3, kiln #3) — full 500-row first page forces a second query; asserts `skip: 1` + `cursor.id` and that all 501 URLs arrive.
- **Failed row delete leaves storage untouched** (kiln #3) — added to the `DatabaseError` test, locking the converse of the ordering guarantee.
- **The deleted-file-upload-element hole and the "don't broaden to any storage URL" reasoning** (kiln #4) — both now in `Open gaps`, including why the narrow keying is deliberate rather than an oversight. That security argument is worth having written down somewhere permanent.

**One addition of my own:** cleanup now logs and continues instead of propagating. The rows are already committed as deleted at that point, so an unexpected throw would report a failed reset for work that actually succeeded. Covered by `Reports the reset as successful when storage cleanup fails`.

**Not taken**

- **Durable, transactionally-persisted cleanup jobs** (CodeRabbit #4). That needs a table, a migration and a worker — a much larger change than the leak it would guard, and the leak closes without it. Chunking removes the unbounded-concurrency failure mode; what remains is wall-clock on pathologically large surveys, which is now written up in `Open gaps` along with kiln's `DeleteObjects` (1000 keys/request) suggestion as the cheaper next step if it ever bites.

**Nit:** dropped the empty `Fixes ENG-` line — there's no ticket for this, and the comment above it says why.

Verification: 5 tests red on `origin/main`'s `survey.ts`, 16 green with the fix; full `apps/web` suite 627 files / 8297 tests green.

jobenjada:
Second round addressed in 184c681.

**Taken — typed `getSurvey` mock.** Now `vi.fn<typeof getSurvey>()` off a type-only import, so `mockResolvedValue` is checked against `Promise<TSurvey | null>`.

**Not taken — making collection and `deleteMany` atomic.** I looked at this properly rather than just deferring it, and I think the suggested cure is worse than the disease.

The window is real and already in `Risks`: a response committing between the scan and the delete gets its row removed without its file URL collected, orphaning one object.

The two suggested fixes both cost more than they save:

- **`DELETE … RETURNING data`** would genuinely close it — one statement, exactly the deleted rows, no window. But Prisma's `deleteMany` can't return rows, so it means raw SQL, and it loads *every* response's `data` blob into one result set. That is precisely the unbounded-memory failure mode `RESPONSE_FILE_SCAN_PAGE_SIZE` exists to avoid, and it would happen on every reset rather than only when a race occurs. Trading a rare leak for a systematic memory risk is the wrong direction. Batched `DELETE … RETURNING` with `LIMIT` inside one transaction would bound memory *and* be atomic, but it turns a two-statement reset into N statements holding row locks on `Response` for the duration — a real availability change for a storage leak.
- **Serializing response writes for the reset** would need a lock every response-write path takes, i.e. contention on the hot ingest path to protect a cleanup routine.

The deeper reason I don't think atomicity buys much here: **the reset is not atomic with respect to incoming responses either way.** A response committing 1ms *after* the transaction survives the reset completely — a more visible outcome than an orphaned file, and no locking fixes it because it's outside the transaction by definition. So both sides of the commit boundary are "reset raced with a submission"; this change only decides which of the two outcomes you get in a millisecond-wide window, and the orphaned file is the milder one. It also fails in the safe direction — it leaks storage rather than deleting a live file.

Happy to be overruled if a maintainer wants the batched-`RETURNING` version; it's a contained change, just a bigger one than closing this leak calls for. Left documented in `Risks` with the reasoning.

mattinannt (commented):
@jobenjada please also fix the sonarqube issues.

I'm looping in @pandeymangg for a review since he can evaluate the risks that this change can bring better than me :-)

jobenjada:
Third round: SonarQube's two findings plus the three inline review threads that were resolved but not actually covered yet. `88f51fe` and `ad4946e`.

**SonarQube (both fixed)**

- **S3776 (critical)** — `collectSurveyResponseFileUrls` was at cognitive complexity 18/15. Extracted the per-page extraction into `collectFileUrlsFromPage`, which removes two levels of nesting from the paging loop.
- **S7755** — `responses.at(-1)` instead of `responses[responses.length - 1]`.

Verified locally with `eslint-plugin-sonarjs` (the same rule implementations SonarQube runs): 18/15 reproduced on the previous revision, no violation now, and no other sonarjs finding across the changed files.

**@kiln-agent — `orderBy: { id: "asc" }` has no supporting index.** This was the best catch on the PR and it was a real regression I introduced. Confirmed: `Response` carries `@@index([createdAt])`, `@@index([surveyId, createdAt])`, `@@index([contactId, createdAt])` and `@@unique([surveyId, singleUseId])`, with nothing on `(surveyId, id)` — so paging added to bound memory was trading it for a query cost growing with the same input.

Took the migration-free option: `orderBy: [{ createdAt: "asc" }, { id: "asc" }]` with an explicit keyset `where` (`createdAt > c OR (createdAt = c AND id > lastId)`) instead of Prisma's `cursor`/`skip`, so it rides the existing `(surveyId, createdAt)` index. `id` is only the tiebreaker keeping the order total, so rows sharing a `createdAt` are neither skipped nor double-read. The pagination test now asserts the keyset `where`, the `orderBy`, and the *absence* of `skip`/`cursor`.

**@kiln-agent — the comment claims more than the code delivers.** Correct, and fixed. It now says the check reflects the survey's *current* definition rather than its response history, and that answers left by a since-deleted upload element are not cleaned up here or by the single-response path.

One correction on the suggested fix, though — I don't think shape-based filtering is safe, and I'd rather not do it:

> Note the id filter isn't a security boundary — `deleteResponseFileUrls` re-resolves each URL's storage id and drops anything outside the survey's workspace

That's right about the *workspace* boundary but it's the wrong granularity. Storage keys are `/storage/{workspaceId}/private/{fileName}` — workspace-scoped, not survey-scoped — so a URL belonging to a *different survey in the same workspace* passes that check. Filter on shape and a respondent to survey A can paste survey B's file URL into a text answer, and resetting A deletes B's live file. Same reason the raw `DELETE … RETURNING data` variant doesn't rescue it: the problem is which keys we consider deletable, not where they came from. Keeping the narrow id filter and documenting the gap is the conservative call; closing it properly needs survey provenance in the storage path or a file-ownership table, which is its own change.

**@kiln-agent — third copy of the extraction rule, already drifted.** Verified, including the drift: this file guards with `Array.isArray` and filters to strings, while `lib/response/service.ts:632` and `modules/api/v2/management/responses/[responseId]/lib/utils.ts` both do a bare `elementResponse as string[]` — so a malformed non-array answer that this file skips still reaches storage through the other two. Extracting `collectResponseFileUrls(data, survey)` next to `deleteResponseFileUrls` is the right end state.

Left out of this diff on purpose: it changes behaviour on two currently-working paths (they'd inherit the array guard) and touches their tests, and I'd rather the leak fix not carry that. It's written into `Open gaps` and I'm happy to open it as a follow-up immediately after this merges — say the word if you'd rather have it here.

**Still open — the scan/delete race.** CodeRabbit raised it again on the new revision. My reasoning is unchanged and in the [earlier reply](https://github.com/formbricks/formbricks/pull/8930#issuecomment-5353975130): the reset isn't atomic with respect to incoming responses either way, and an orphaned file is the milder of the two race outcomes. Leaving that thread open for a maintainer rather than resolving it myself.

CI green on `88f51fe`, and I'll confirm the fresh Sonar scan comes back clean.

jobenjada:
Third round: SonarQube's two findings plus the three inline review threads that were resolved but not actually covered yet. `88f51fe` and `ad4946e`.

**SonarQube (both fixed)**

- **S3776 (critical)** — `collectSurveyResponseFileUrls` was at cognitive complexity 18/15. Extracted the per-page extraction into `collectFileUrlsFromPage`, which removes two levels of nesting from the paging loop.
- **S7755** — `responses.at(-1)` instead of `responses[responses.length - 1]`.

Verified locally with `eslint-plugin-sonarjs` (the same rule implementations SonarQube runs): 18/15 reproduced on the previous revision, no violation now, and no other sonarjs finding across the changed files.

**@kiln-agent — `orderBy: { id: "asc" }` has no supporting index.** This was the best catch on the PR and it was a real regression I introduced. Confirmed: `Response` carries `@@index([createdAt])`, `@@index([surveyId, createdAt])`, `@@index([contactId, createdAt])` and `@@unique([surveyId, singleUseId])`, with nothing on `(surveyId, id)` — so paging added to bound memory was trading it for a query cost growing with the same input.

Took the migration-free option: `orderBy: [{ createdAt: "asc" }, { id: "asc" }]` with an explicit keyset `where` (`createdAt > c OR (createdAt = c AND id > lastId)`) instead of Prisma's `cursor`/`skip`, so it rides the existing `(surveyId, createdAt)` index. `id` is only the tiebreaker keeping the order total, so rows sharing a `createdAt` are neither skipped nor double-read. The pagination test now asserts the keyset `where`, the `orderBy`, and the *absence* of `skip`/`cursor`.

**@kiln-agent — the comment claims more than the code delivers.** Correct, and fixed. It now says the check reflects the survey's *current* definition rather than its response history, and that answers left by a since-deleted upload element are not cleaned up here or by the single-response path.

One correction on the suggested fix, though — I don't think shape-based filtering is safe, and I'd rather not do it:

> Note the id filter isn't a security boundary — `deleteResponseFileUrls` re-resolves each URL's storage id and drops anything outside the survey's workspace

That's right about the *workspace* boundary but it's the wrong granularity. Storage keys are `/storage/{workspaceId}/private/{fileName}` — workspace-scoped, not survey-scoped — so a URL belonging to a *different survey in the same workspace* passes that check. Filter on shape and a respondent to survey A can paste survey B's file URL into a text answer, and resetting A deletes B's live file. Same reason the raw `DELETE … RETURNING data` variant doesn't rescue it: the problem is which keys we consider deletable, not where they came from. Keeping the narrow id filter and documenting the gap is the conservative call; closing it properly needs survey provenance in the storage path or a file-ownership table, which is its own change.

**@kiln-agent — third copy of the extraction rule, already drifted.** Verified, including the drift: this file guards with `Array.isArray` and filters to strings, while `lib/response/service.ts:632` and `modules/api/v2/management/responses/[responseId]/lib/utils.ts` both do a bare `elementResponse as string[]` — so a malformed non-array answer that this file skips still reaches storage through the other two. Extracting `collectResponseFileUrls(data, survey)` next to `deleteResponseFileUrls` is the right end state.

Left out of this diff on purpose: it changes behaviour on two currently-working paths (they'd inherit the array guard) and touches their tests, and I'd rather the leak fix not carry that. It's written into `Open gaps` and I'm happy to open it as a follow-up immediately after this merges — say the word if you'd rather have it here.

**Still open — the scan/delete race.** CodeRabbit raised it again on the new revision. My reasoning is unchanged and in the [earlier reply](https://github.com/formbricks/formbricks/pull/8930#issuecomment-5353975130): the reset isn't atomic with respect to incoming responses either way, and an orphaned file is the milder of the two race outcomes. Leaving that thread open for a maintainer rather than resolving it myself.

CI green on `88f51fe`, and I'll confirm the fresh Sonar scan comes back clean.

pandeymangg (approved):
LGTM

PR #8932, opened 2026-08-20
fix(surveys): keep mount autofocus from painting an option as hovered or selected [ENG-2288]

### LINKED ISSUE #8786: Rating scale (numbers): first option shows grey hover background on load until another option is hovered
Image

### Issue Summary

On a rating element with scale format numbers, category 1 appears with a translucent grey background as soon as the survey card loads, even though nothing is selected and the pointer is not over the scale. The grey stays on option 1 until the mouse moves over another rating category (or focus moves away).

This looks like a selected/hovered state and is confusing for respondents.

Likely cause: After the WCAG keyboard refactor in #8566 (useRovingRadioGroup), the first unselected option gets tabIndex={0}. Survey cards autofocus the first interactive control on mount (shouldFocusOnMount → focusFirstControl in block-conditional.tsx). That focuses rating 1.

In packages/survey-ui/src/components/elements/rating.tsx, number cells treat focus the same as hover via onFocus → setHoveredValue(number), which applies bg-input-selected-bg when isHovered && !isSelected. So autofocus on load paints option 1 with the hover grey overlay.

NPS number cells in packages/survey-ui/src/components/elements/nps.tsx use the same focus→hover pattern and may show the same issue on 0.

Suggested fix: Drive the number-scale (and NPS) grey preview from pointer hover only (onMouseEnter / onMouseLeave), not from onFocus / onBlur. Keep keyboard focus indication via the existing data-fb-scale-cell focus-ring styles in globals.css.

### Expected Behavior

On load, with no selection and no pointer over the scale, all number options should look idle (no hover/selected grey). Hover styling should appear only while the pointer is over an option. Keyboard focus should show the focus ring, not the hover fill.

**Steps to reproduce**
1. Create or open a survey with a Rating question, scale = numbers.
2. Open the survey as a respondent (link survey or preview) so the rating card is the focused card on load.
3. Observe option 1 without moving the mouse.
4. Move the mouse over another number — the grey on 1 clears / moves with hover.

**Environment notes**
- Introduced with / after #8566 (refactor(surveys): WCAG-compliant keyboard and focus model for survey elements).
- Relevant code: packages/survey-ui/src/components/elements/rating.tsx (number option onFocus/hoveredValue + bg-input-selected-bg), packages/survey-ui/src/lib/use-roving-radio-group.ts, autofocus in packages/surveys/src/components/general/block-conditional.tsx.

### Your Environment

Formbricks Cloud (app.formbricks.com)

itsjavi:
Fixes formbricks/formbricks#8786

Ref [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load)

## What & why

A survey card moves focus to its first control when it mounts (`focusFirstControl`, there on purpose for keyboard and screen-reader users). Several elements treated that focus as a pointer hover, or styled it exactly like a chosen answer — so a card opened looking as if it had already been answered, before the respondent touched anything:

* **rating (number) and NPS** — `onFocus` set the same `hoveredValue` state as `onMouseEnter`, so option 1 (NPS 0) opened carrying the grey hover fill. This is the reported bug.
* **rating (star, smiley)** — same coupling, louder: star 1 opened *filled yellow*, smiley 1 filled and brand-stroked, reading as a 1-star rating already given.
* **single-select, multi-select, ranking** — each option row carried `focus-within:bg-option-selected-bg focus-within:border-brand`, which is character-for-character its `isSelected` styling, so the first choice opened looking answered (the checkbox screenshot on [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load)).
* **any element with a link in its prose** — `focusFirstControl` accepted `a[href]`, and a link inside the question text sits *before* the element's own control in the DOM. The card opened with a ring around "Privacy Policy" (the second screenshot on [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load)), and a screen reader announced that link instead of the question's control.

The fix separates *the respondent previewing a value* from *focus landing somewhere*:

* **the scales preview from** `hoveredValue ?? keyboardValue`**.** `useRovingRadioGroup` already owns the explicit arrow / Home / End moves in `focusValue`, which a card's mount autofocus never goes through, so the hook now reports that position as `keyboardValue`. Pointer hover and arrow keys preview identically; mount focus, a pointer click and Tab entering the group all leave it null, and a click after arrowing clears the preview instead of stranding it.
* **the option rows** in `single-select.tsx`, `multi-select.tsx` and `ranking.tsx` no longer restyle themselves on `:focus-within` — that rule painted the *selected* colours for any focus, whatever its origin.
* `focusFirstControl` **never targets a link.** Anchors are excluded from the candidate selector, so the block's own control wins; a prose link in a headline or subheader is no longer what a card opens on.

Focus itself is unchanged everywhere: the same kind of element receives it, and survey-ui's uniform ring (`globals.css`: `label:has(input:focus-visible)`, `label[data-fb-scale-cell]`, `button:focus-visible`) still marks it. What is gone is a *preview that nobody asked for* — the same confusion formbricks/formbricks#8566 removed when it decoupled selection from arrow-key focus.

**Deliberately unchanged:** the focus ring still appears on load, because that entry point is what formbricks/formbricks#8566 added for keyboard and screen-reader users. [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load) asks more broadly for nothing to look highlighted on load, so it is referenced rather than closed: after this PR the first option no longer looks *answered*, but it does still carry a focus ring. Dropping that is a separate decision about the autofocus itself — see the first Open gap, which is the same decision from the other side.

## Before / after

> Rendered from the **compiled** `@formbricks/surveys` **bundle**, not from a real `/s/{surveyId}` page: no Docker daemon here, so no database to boot the app against. Both sides are real compiled bundles driven in Chromium — the "before" side built from `main`'s version of the changed files, the "after" side from this branch — same fixture, same 720×560 viewport at 2× scale, one block per case so the element under test is the card's first control. Every capture is **the card on load: no click, no keypress, no pointer movement.** The ring on the first option is the autofocus this PR keeps. (GitHub strips `<details>` from descriptions written through the API, so the remaining nine figures are linked rather than embedded, to keep this readable.)

**Rating, number scale** — the reported bug. Option 1 opens filled grey, as if hovered or picked; after, it is the same colour as 2–5 and only the focus ring marks it.

Rating number scale, before and after, on load

**Multi-select** — the customer's own screenshot on [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load), reproduced and fixed. "Writing new content" opened in the selected colours with a brand border, next to four unanswered options.

Multi-select, before and after, on load

**A card with links in its prose** — the other [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load) screenshot. A non-external CTA contributes no control of its own, so mount focus used to ring "Privacy Policy" mid-sentence; now it goes to the card's own button.

CTA card with prose links, before and after, on load

The same capture for every other affected element, on load with no interaction:  <br>[star scale](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-rating-star.png>) (a filled yellow star 1 is not "focused", it is a 1-star review) ·  <br>[smiley scale](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-rating-smiley.png>) ·  <br>[NPS](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-nps.png>) ·  <br>[single-select](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-single-select.png>) ·  <br>[ranking](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-ranking.png>) (the first item opened in the *ranked* colours while its rank badge was still empty) ·  <br>[consent](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-consent.png>) (focus moves from the prose link to the checkbox)

## Breaking changes

None. No API, SDK, prop, route, env var or emitted value changes. `useRovingRadioGroup` gains a returned field (`keyboardValue`); it is an internal hook of `@formbricks/survey-ui`, and its existing consumers destructure `getRadioProps` only.

## Migrations & env

* none

## How this was tested

Same harness as the figures above (compiled bundle in Chromium, before side built from `main`). Captures are on load with no interaction unless a row says otherwise; computed values come from `getComputedStyle`.

**Coverage**

| Behaviour | How | Outcome |
| -- | -- | -- |
| Rating, number scale: option 1 is idle on load | manual | Option 1's background `bg-input-selected-bg` **→** `bg-input-bg` (`color(srgb 0.882941 0.894118 0.927647)` → `rgb(237, 240, 249)`, identical to options 2–5). Focus still lands on `input[aria-label="Rate 1 out of 5"]`, ringed. First figure above |
| Rating, star scale: star 1 is not pre-filled on load | manual | Star 1's fill **yellow-400 → slate-300** (`oklch(0.852 0.199 91.936)` → `oklch(0.869 0.022 252.894)`); all five idle. Star figure linked above |
| Rating, smiley scale: smiley 1 is not pre-filled on load | manual | Smiley 1 opens outline-only instead of filled yellow with the brand stroke. Figure linked above |
| NPS: 0 is idle on load | manual | Cell 0 `bg-input-selected-bg` **→** `bg-input-bg`. Figure linked above |
| Single-select and multi-select: the first option does not look answered | manual | Option 1 `bg-option-selected-bg` **+** `border-brand` **→** `bg-option-bg` **+** `border-option-border` (`rgb(30, 64, 175)` border gone). Second figure above |
| Ranking: the first item does not look already ranked, and still shows focus | manual | Item 1 `bg-option-selected-bg` **+** `border-brand` **→ idle**, while the brand ring is still painted on the item's button (`#fbjs button:focus-visible` in `globals.css`) — WCAG 2.4.7 intact, visible in the ranking walk linked below |
| Consent: mount focus goes to the checkbox, not to a link in the question text | manual | `document.activeElement` on load `a` **"Privacy Policy" → the consent checkbox** `button`; both prose links render unringed. Figure linked above |
| **A card whose element contributes no control of its own, with links in its prose** | manual | The shape that lost the `a[href]` fallback. Non-external CTA (renders no control — see `cta.tsx`) with two links in its subheader: `document.activeElement` on load `a` **"Privacy Policy" →** `button` **"Finish"**, the block's own nav button. Third figure above. A card with *no* control at all is not constructible through the survey schema — every block renders Submit and/or Back — which is why the fallback was dead code; the hypothetical is in Risks |
| The pointer still previews exactly as before | manual | Pointer over option 4 → 4 takes the hover fill and 1 stays idle (before: 1 *and* 4 were grey, focus and pointer painting the same thing); over star 4 → stars 1–4 fill; pointer away → all idle. Walk figures linked below |
| **The keyboard previews the same value the pointer does** | manual | `→ →` from mount focus: star scale fills **stars 1–3** with the ring on 3, number scale tints **cell 3**, NPS tints **cell 2** — i.e. arrow keys and hover paint the same thing, which is what the first revision of this PR had lost. Walk figures linked below |
| …but only for a move the respondent made | manual | Mount focus: **all five idle** (`fill: oklch(0.869 0.022 252.894)` on every star). Tab entering the group: still idle. Tab out: preview cleared, selection kept. Clicking option 5 after arrowing to 3: cell 3 returns to `rgb(237, 240, 249)` and only 5 is painted — no stranded preview |
| Selection is untouched | manual | `Space` on the focused option still checks it (`input:checked`) and paints the selected styling: number cell → `bg-brand-20` + brand border, stars → 1–3 filled, ranking → item ranked and its button becomes "Remove Price from ranking" |
| A card reached by a **pointer** click on Next paints no focus indicator | manual | Measured on both bundles, two-block fixture, real mouse click on Next: card 2's first option has `:focus` but **not** `:focus-visible`, so no ring on either side; before it still showed the `:focus-within` fill, after it shows nothing. `Space` then checks the option on **both** bundles. This is the first Open gap, not a fixed row |
| Nothing regressed in the two packages | unit (guard) | `pnpm test --filter=@formbricks/surveys --filter=@formbricks/survey-ui` — 35 files, 829 tests, green |

Interaction walks on the **after** bundle — load, pointer over an option, two arrow presses, then Space, four moments per figure:  <br>[star scale](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-rating-star.png>) (hover previews 1–4, arrows preview 1–3, Space commits) ·  <br>[number scale](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-rating-number.png>) (the grey follows the pointer and the arrow keys, but not the card opening) ·  <br>[ranking](<https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-ranking.png>) (no fill on focus, ring still on the item's button; ranking has no roving group, so arrows are a no-op)

**Open gaps**

* **A card reached by a pointer click on Next shows no focus indicator at all** — measured above, and worth a decision that is bigger than this PR. `:focus-within` matched any focus; the ring is `:focus-visible`, which Chromium and WebKit withhold from a script `focus()` when the last input modality was a pointer. So after a mouse-driven Next, focus sits on option 1 invisibly, and a Space press (a plausible "scroll the card" reflex) checks it — on **both** bundles, so the footgun is not introduced here, but this PR removes the misleading paint that used to mark the target. Every remedy cuts against something: a modality-independent mount ring is exactly the "first option looks highlighted on load" that [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load) asks us to stop doing; skipping mount autofocus for pointer navigation would strand VoiceOver/TalkBack users, whose navigation *is* pointer-driven (the regression formbricks/formbricks#8566 fixed); focusing the card container (`tabindex="-1"`, already ring-exempt) on the pointer path looks best to me and is still a change to the autofocus model. It belongs to the [ENG-2415](https://linear.app/formbricks/issue/ENG-2415/rating-scale-numbers-first-option-shows-grey-hover-background-on-load) decision, with this data attached — discussion is on the open `multi-select.tsx` thread.
* **No automated regression guard ships with this fix.** The repo rules out component unit tests for `.tsx`, and the behaviour is real-browser paint and focus, which jsdom/happy-dom cannot observe. A Playwright spec would be the right guard, but there is no database here to run one against, and shipping a spec I could not run would be worse than none. Repro recipe: open any rating/NPS/choice card as the first card and read the first option's `backgroundColor` (or star 1's `fill`) before touching anything.
* The media comes from the surveys bundle rendered standalone, so it proves the elements' own behaviour, not the full link-survey page around them. A pass on a real link survey is still owed by a human — including a touch device, where `onMouseEnter` can fire on tap.
* `picture-select` and `matrix` also consume `useRovingRadioGroup`; they ignore `keyboardValue` and have no focus/selection coupling, so they are unchanged by inspection, not by capture.
* Consent's own `focus-within:border-ring` on the checkbox box is left alone — it is a focus affordance on that element's single control, not a "this is selected" look.

**Risks**

* `useRovingRadioGroup` is shared (rating, NPS, single-select, picture-select). The change is additive — a new returned field plus one extra state write per focus event — and `focusValue` still sets the roving position first, so the tabindex model is untouched. If a scale ever previews the wrong option, the suspect is `isKeyMove`: it is a ref that is true only for the synchronous `.focus()` inside `focusValue`.
* `focusFirstControl` is shared with the failed-submit path (`preferInvalid`), which is unchanged: it still targets `aria-invalid` inputs first, and those callers still scroll the field they focus into view themselves. `apps/web/playwright/survey-keyboard.spec.ts` covers that path; its fixtures have no prose links, so the anchor exclusion cannot touch them.
* Anchors are now excluded from mount focus with no fallback. That is safe because every block renders its Submit and/or Back button inside the focus root — the shape with neither (an auto-progress element on the first block) is always a radio group. A future card type with no control and only a link would receive no focus at all; the guard against that is `FOCUSABLE_CONTROL_SELECTOR`'s docblock, which says so.

---

> [!NOTE]
> **AI model used** — Claude Opus 5, reasoning effort: Max.

### DIFF
--- a/packages/survey-ui/src/components/elements/multi-select.tsx
+++ b/packages/survey-ui/src/components/elements/multi-select.tsx
@@ -92,7 +92,9 @@ const getOptionContainerClassName = (isSelected: boolean, isDisabled: boolean): 
     "relative flex flex-col border transition-colors outline-none",
     "rounded-option px-option-x py-option-y",
     isSelected ? "bg-option-selected-bg border-brand" : "bg-option-bg border-option-border",
-    "focus-within:border-brand focus-within:bg-option-selected-bg",
+    // No focus-within fill: it repainted the option in the *selected* colors, so the card's
+    // mount autofocus made option 1 look answered (ENG-2288). Focus has its own uniform ring
+    // on the option label, from survey-ui's globals.css.
     "hover:bg-option-hover-bg",
     isDisabled ? "cursor-not-allowed opacity-50" : "cursor-pointer"
   );
--- a/packages/survey-ui/src/components/elements/rating.tsx
+++ b/packages/survey-ui/src/components/elements/rating.tsx
@@ -174,6 +174,11 @@ function Rating({
 }: Readonly<RatingProps>): React.JSX.Element {
   const errorAria = getElementErrorAria(inputId, errorMessage);
 
+  // Pointer-only hover preview. The player autofocuses a card's first control on mount
+  // (focusFirstControl in block-conditional.tsx), so driving this from onFocus too painted
+  // option 1 as hovered — grey fill on the number scale, a filled star/smiley — before the
+  // respondent had touched anything (ENG-2288). Keyboard focus is indicated on its own by the
+  // uniform focus ring in survey-ui's globals.css.
   const [hoveredValue, setHoveredValue] = React.useState<number | null>(null);
 
   // Ensure value is within valid range
@@ -247,14 +252,6 @@ function Rating({
         }}
         onMouseLeave={() => {
           setHoveredValue(null);
-        }}
-        onFocus={() => {
-          if (!disabled) {
-            setHoveredValue(number);
-          }
-        }}
-        onBlur={() => {
-          setHoveredValue(null);
         }}>
         {colorCoding ? (
           <div
@@ -301,14 +298,6 @@ function Rating({
             }
           }}
           onMouseLeave={() => {
-            setHoveredValue(null);
-          }}
-          onFocus={() => {
-            if (!disabled) {
-              setHoveredValue(number);
-            }
-          }}
-          onBlur={() => {
             setHoveredValue(null);
           }}>
           <input
@@ -358,14 +347,6 @@ function Rating({
             }
           }}
           onMouseLeave={() => {
-            setHoveredValue(null);
-          }}
-          onFocus={() => {
-            if (!disabled) {
-              setHoveredValue(number);
-            }
-          }}
-          onBlur={() => {
             setHoveredValue(null);
           }}>
           <input
--- a/packages/survey-ui/src/components/elements/rating.tsx
+++ b/packages/survey-ui/src/components/elements/rating.tsx
@@ -174,11 +174,6 @@ function Rating({
 }: Readonly<RatingProps>): React.JSX.Element {
   const errorAria = getElementErrorAria(inputId, errorMessage);
 
-  // Pointer-only hover preview. The player autofocuses a card's first control on mount
-  // (focusFirstControl in block-conditional.tsx), so driving this from onFocus too painted
-  // option 1 as hovered — grey fill on the number scale, a filled star/smiley — before the
-  // respondent had touched anything (ENG-2288). Keyboard focus is indicated on its own by the
-  // uniform focus ring in survey-ui's globals.css.
   const [hoveredValue, setHoveredValue] = React.useState<number | null>(null);
 
   // Ensure value is within valid range
@@ -195,13 +190,22 @@ function Rating({
   // the focusable control), but selection is decoupled from arrow-key focus
   // moves because selecting can trigger auto-progress (see useRovingRadioGroup).
   const ratingValues = Array.from({ length: range }, (_, i) => String(i + 1));
-  const { getRadioProps } = useRovingRadioGroup({
+  const { getRadioProps, keyboardValue } = useRovingRadioGroup({
     values: ratingValues,
     selectedValue: currentValue === undefined ? undefined : String(currentValue),
     onSelect: (v) => {
       handleSelect(Number(v));
     },
   });
+
+  // The value the scale previews before anything is selected: the pointer first, otherwise the
+  // option the respondent arrowed to. Deliberately NOT every focus: cards autofocus their first
+  // control on mount (focusFirstControl in block-conditional.tsx), and previewing that painted
+  // option 1 as hovered — grey on the number scale, a filled star or smiley — on a card nobody had
+  // touched yet (ENG-2288). `keyboardValue` is only set by an explicit arrow/Home/End move, so the
+  // preview a pointer user gets on hover is the one a keyboard user gets on arrow, and mount focus
+  // gets none (it is marked by the focus ring in survey-ui's globals.css, like every other card).
+  const previewValue = hoveredValue ?? (keyboardValue === null ? null : Number(keyboardValue));
 
   // Get number option color for color coding
   const getRatingNumberOptionColor = (ratingRange: number, idx: number): string => {
@@ -222,7 +226,7 @@ function Rating({
   // Render number scale option
   const renderNumberOption = (number: number, totalLength: number): React.JSX.Element => {
     const isSelected = currentValue === number;
-    const isHovered = hoveredValue === number;
+    const isHovered = previewValue === number;
     const isLast = totalLength === number;
     const isFirst = number === 1;
 
@@ -279,8 +283,8 @@ function Rating({
   // Render star scale option
   const renderStarOption = (number: number): React.JSX.Element => {
     const isSelected = currentValue === number;
-    // Fill all stars up to the hovered value (if hovering) or selected value
-    const activeValue = hoveredValue ?? currentValue ?? 0;
+    // Fill all stars up to the previewed value (pointer or arrow key) or the selected value
+    const activeValue = previewValue ?? currentValue ?? 0;
     const isActive = number <= activeValue;
 
     return (
@@ -328,8 +332,7 @@ function Rating({
   // Render smiley scale option
   const renderSmileyOption = (number: number, index: number): React.JSX.Element => {
     const isSelected = currentValue === number;
-    const isHovered = hoveredValue === number;
-    const isActive = isSelected || isHovered;
+    const isActive = isSelected || previewValue === number;
 
     return (
       // The flex-1 wrapper keeps the even column layout; the label is sized to the icon so the
--- a/packages/survey-ui/src/components/elements/single-select.tsx
+++ b/packages/survey-ui/src/components/elements/single-select.tsx
@@ -373,7 +373,9 @@ function getOptionContainerClassName(isSelected: boolean, disabled: boolean): st
     "relative flex cursor-pointer flex-col border transition-colors outline-none",
     "rounded-option px-option-x py-option-y",
     isSelected ? "bg-option-selected-bg border-brand" : "bg-option-bg border-option-border",
-    "focus-within:border-brand focus-within:bg-option-selected-bg",
+    // No focus-within fill: it repainted the option in the *selected* colors, so the card's
+    // mount autofocus made option 1 look answered (ENG-2288). Focus has its own uniform ring
+    // on the option label, from survey-ui's globals.css.
     "hover:bg-option-hover-bg",
     disabled && "cursor-not-allowed opacity-50"
   );
--- a/packages/surveys/src/components/general/block-conditional.tsx
+++ b/packages/surveys/src/components/general/block-conditional.tsx
@@ -27,13 +27,23 @@ const FOCUSABLE_CONTROL_SELECTOR = [
   "textarea:not(:disabled)",
   "select:not(:disabled)",
   "button:not(:disabled)",
-  "a[href]",
-  '[tabindex="0"]',
+  '[tabindex="0"]:not(a)',
 ].join(", ");
 
 /**
+ * Links are focusable but are not what a card asks of the respondent: a headline or subheader may
+ * carry one in its prose (a consent element pointing at a privacy policy), and that link sits
+ * *before* the element's own control in the DOM. Focusing it on mount opened the card with a ring
+ * around a word in the question text, reading as a highlighted suggestion (ENG-2415), and gave a
+ * screen-reader user "Privacy Policy, link" as their orientation instead of the question's control.
+ * Kept as a fallback so a card that really has nothing else still receives focus.
+ */
+const FOCUSABLE_LINK_SELECTOR = "a[href]";
+
+/**
  * Focuses the first interactive control inside `root`. With `preferInvalid`,
- * controls flagged aria-invalid win.
+ * controls flagged aria-invalid win; a prose link is only a fallback (see
+ * FOCUSABLE_LINK_SELECTOR).
  *
  * Scrolling is always left to the caller. The first control can sit *below* the card's content — a
  * CTA block has no input of its own, so its first control is the Next button rendered after the
@@ -46,7 +56,10 @@ const focusFirstControl = (root: HTMLElement, preferInvalid = false): void => {
         ':is(input, textarea, select)[aria-invalid="true"]:not([tabindex="-1"]):not(:disabled)'
       )
     : null;
-  const target = invalidTarget ?? root.querySelector<HTMLElement>(FOCUSABLE_CONTROL_SELECTOR);
+  const target =
+    invalidTarget ??
+    root.querySelector<HTMLElement>(FOCUSABLE_CONTROL_SELECTOR) ??
+    root.querySelector<HTMLElement>(FOCUSABLE_LINK_SELECTOR);
   target?.focus({ preventScroll: true });
 };
 
--- a/packages/surveys/src/components/general/block-conditional.tsx
+++ b/packages/surveys/src/components/general/block-conditional.tsx
@@ -22,6 +22,17 @@ import { getFirstErrorMessage, validateBlockResponses } from "@/lib/validation/e
 
 const AUTO_PROGRESS_SUBMIT_DELAY_MS = 350;
 
+/**
+ * Anchors are deliberately absent: they are focusable, but a link is never what a card asks of the
+ * respondent. A headline or subheader may carry one in its prose (a consent element pointing at a
+ * privacy policy), and that link sits *before* the element's own control in the DOM — so mount
+ * focus landed on a word in the question text, ringed, reading as a highlighted suggestion
+ * (ENG-2415), and gave a screen-reader user "Privacy Policy, link" as their orientation instead of
+ * the question's control. Excluding `a` from `[tabindex="0"]` too keeps that uniform for an anchor
+ * made focusable by hand. Nothing is left unfocused by the exclusion: every block renders its
+ * Submit and/or Back button inside `root`, and the only shape without one (an auto-progress
+ * element on the first block) is a radio group.
+ */
 const FOCUSABLE_CONTROL_SELECTOR = [
   'input:not([type="hidden"]):not([tabindex="-1"]):not(:disabled)',
   "textarea:not(:disabled)",
@@ -31,19 +42,9 @@ const FOCUSABLE_CONTROL_SELECTOR = [
 ].join(", ");
 
 /**
- * Links are focusable but are not what a card asks of the respondent: a headline or subheader may
- * carry one in its prose (a consent element pointing at a privacy policy), and that link sits
- * *before* the element's own control in the DOM. Focusing it on mount opened the card with a ring
- * around a word in the question text, reading as a highlighted suggestion (ENG-2415), and gave a
- * screen-reader user "Privacy Policy, link" as their orientation instead of the question's control.
- * Kept as a fallback so a card that really has nothing else still receives focus.
- */
-const FOCUSABLE_LINK_SELECTOR = "a[href]";
-
-/**
  * Focuses the first interactive control inside `root`. With `preferInvalid`,
- * controls flagged aria-invalid win; a prose link is only a fallback (see
- * FOCUSABLE_LINK_SELECTOR).
+ * controls flagged aria-invalid win. Prose links are not candidates at all (see
+ * FOCUSABLE_CONTROL_SELECTOR).
  *
  * Scrolling is always left to the caller. The first control can sit *below* the card's content — a
  * CTA block has no input of its own, so its first control is the Next button rendered after the
@@ -56,10 +57,7 @@ const focusFirstControl = (root: HTMLElement, preferInvalid = false): void => {
         ':is(input, textarea, select)[aria-invalid="true"]:not([tabindex="-1"]):not(:disabled)'
       )
     : null;
-  const target =
-    invalidTarget ??
-    root.querySelector<HTMLElement>(FOCUSABLE_CONTROL_SELECTOR) ??
-    root.querySelector<HTMLElement>(FOCUSABLE_LINK_SELECTOR);
+  const target = invalidTarget ?? root.querySelector<HTMLElement>(FOCUSABLE_CONTROL_SELECTOR);
   target?.focus({ preventScroll: true });
 };
 
--- a/packages/survey-ui/src/components/elements/nps.tsx
+++ b/packages/survey-ui/src/components/elements/nps.tsx
@@ -60,6 +60,10 @@ function NPS({
 }: Readonly<NPSProps>): React.JSX.Element {
   const errorAria = getElementErrorAria(inputId, errorMessage);
 
+  // Pointer-only hover preview: driving this from onFocus too made the card's mount autofocus
+  // (focusFirstControl in block-conditional.tsx) paint 0 with the grey hover fill before the
+  // respondent had touched anything (ENG-2288). Keyboard focus is indicated on its own by the
+  // uniform focus ring in survey-ui's globals.css.
   const [hoveredValue, setHoveredValue] = React.useState<number | null>(null);
 
   // Ensure value is within valid range (0-10)
@@ -123,14 +127,6 @@ function NPS({
           }
         }}
         onMouseLeave={() => {
-          setHoveredValue(null);
-        }}
-        onFocus={() => {
-          if (!disabled) {
-            setHoveredValue(number);
-          }
-        }}
-        onBlur={() => {
           setHoveredValue(null);
         }}>
         {colorCoding ? (
--- a/packages/survey-ui/src/components/elements/nps.tsx
+++ b/packages/survey-ui/src/components/elements/nps.tsx
@@ -60,10 +60,6 @@ function NPS({
 }: Readonly<NPSProps>): React.JSX.Element {
   const errorAria = getElementErrorAria(inputId, errorMessage);
 
-  // Pointer-only hover preview: driving this from onFocus too made the card's mount autofocus
-  // (focusFirstControl in block-conditional.tsx) paint 0 with the grey hover fill before the
-  // respondent had touched anything (ENG-2288). Keyboard focus is indicated on its own by the
-  // uniform focus ring in survey-ui's globals.css.
   const [hoveredValue, setHoveredValue] = React.useState<number | null>(null);
 
   // Ensure value is within valid range (0-10)
@@ -80,13 +76,18 @@ function NPS({
   // decoupled from arrow-key focus moves because selecting can trigger
   // auto-progress (see useRovingRadioGroup).
   const npsValues = Array.from({ length: 11 }, (_, i) => String(i));
-  const { getRadioProps } = useRovingRadioGroup({
+  const { getRadioProps, keyboardValue } = useRovingRadioGroup({
     values: npsValues,
     selectedValue: currentValue === undefined ? undefined : String(currentValue),
     onSelect: (v) => {
       handleSelect(Number(v));
     },
   });
+
+  // Pointer hover, or the cell the respondent arrowed to — but never the card's mount autofocus,
+  // which painted 0 with the grey hover fill before the respondent had touched anything (ENG-2288).
+  // See `keyboardValue` in useRovingRadioGroup, and the same comment in rating.tsx.
+  const previewValue = hoveredValue ?? (keyboardValue === null ? null : Number(keyboardValue));
 
   // Get NPS option color for color coding
   const getNPSOptionColor = (idx: number): string => {
@@ -98,7 +99,7 @@ function NPS({
   // Render NPS option (0-10)
   const renderNPSOption = (number: number): React.JSX.Element => {
     const isSelected = currentValue === number;
-    const isHovered = hoveredValue === number;
+    const isHovered = previewValue === number;
     const isLast = number === 10; // Last option is 10
     const isFirst = number === 0; // First option is 0
 
--- a/packages/survey-ui/src/components/elements/ranking.tsx
+++ b/packages/survey-ui/src/components/elements/ranking.tsx
@@ -80,7 +80,10 @@ function RankingItem({
       className={cn(
         "rounded-option flex h-12 cursor-pointer items-center border px-3 transition-all",
         "bg-option-bg border-option-border",
-        "hover:bg-option-hover-bg focus-within:border-brand focus-within:bg-option-selected-bg focus-within:shadow-sm",
+        // No focus-within fill: it repainted the item in the *ranked* colors, so the card's mount
+        // autofocus made the first item look already ranked (ENG-2288). Focus has its own uniform
+        // ring, painted on the item's button by survey-ui's globals.css.
+        "hover:bg-option-hover-bg",
         isRanked && "bg-option-selected-bg border-brand",
         disabled && "cursor-not-allowed opacity-50"
       )}>
--- a/packages/survey-ui/src/lib/use-roving-radio-group.ts
+++ b/packages/survey-ui/src/lib/use-roving-radio-group.ts
@@ -20,6 +20,9 @@ interface RovingRadioProps {
  * - Home/End jump to the first/last option.
  * - Space (native) or Enter selects the focused option.
  *
+ * `keyboardValue` reports the roving position for those explicit key moves only, so a consumer can
+ * preview a value on arrow-key navigation without also previewing on the card's mount autofocus.
+ *
  * Spread the returned props onto every radio `<input>` of the group, in the
  * same order as `values`.
  */
@@ -36,17 +39,32 @@ export function useRovingRadioGroup({
   onSelect: (value: string) => void;
 }): {
   getRadioProps: (value: string) => RovingRadioProps;
+  /** The option the respondent moved focus to with a key, or null. See the state's comment. */
+  keyboardValue: string | null;
 } {
   const inputRefs = React.useRef<Map<string, HTMLInputElement>>(new Map());
   // The roving position while focus is inside the group; null = follow selection.
   const [focusedValue, setFocusedValue] = React.useState<string | null>(null);
+  // The same position, but only while the respondent is the one who moved it, with a key. Focus
+  // that arrives any other way — the card's mount autofocus, a pointer click — leaves this null,
+  // so a consumer can drive a value *preview* from it (the star scale's cumulative fill) without
+  // painting an answer onto a card the respondent has not touched yet (ENG-2288).
+  const [keyboardValue, setKeyboardValue] = React.useState<string | null>(null);
+  // True only for the synchronous .focus() inside focusValue, which is reached from the arrow /
+  // Home / End handlers alone. A ref, not state: the focus event fires during that same call.
+  const isKeyMove = React.useRef(false);
 
   const rovingValue =
     focusedValue ?? (selectedValue && values.includes(selectedValue) ? selectedValue : values[0]);
 
   const focusValue = (value: string): void => {
     setFocusedValue(value);
+    isKeyMove.current = true;
     inputRefs.current.get(value)?.focus();
+    isKeyMove.current = false;
+    // Also set here, not only from the focus handler: focusing the already-focused option (Home at
+    // the first option, an arrow in a one-option group) fires no focus event.
+    setKeyboardValue(value);
   };
 
   const handleKeyDown = (value: string) => (e: React.KeyboardEvent<HTMLInputElement>) => {
@@ -94,7 +112,10 @@ export function useRovingRadioGroup({
     // at the selected option, matching native radio-group behavior.
     const next = e.relatedTarget;
     const leavingGroup = !next || ![...inputRefs.current.values()].includes(next as HTMLInputElement);
-    if (leavingGroup) setFocusedValue(null);
+    if (leavingGroup) {
+      setFocusedValue(null);
+      setKeyboardValue(null);
+    }
   };
 
   const getRadioProps = (value: string): RovingRadioProps => ({
@@ -106,9 +127,12 @@ export function useRovingRadioGroup({
     onKeyDown: handleKeyDown(value),
     onFocus: () => {
       setFocusedValue(value);
+      // Mount autofocus, a click, or Tab entering the group: no preview, and no stale one left
+      // behind from an earlier key move (clicking option 5 after arrowing to 3).
+      setKeyboardValue(isKeyMove.current ? value : null);
     },
     onBlur: handleBlur,
   });
 
-  return { getRadioProps };
+  return { getRadioProps, keyboardValue };
 }


itsjavi:
@coderabbitai review

The "Coverage accounts for the whole diff" warning was evaluated against the first revision of the description, which said figures would be attached in the next revision. They are attached now: eight before/after figures (one per affected element, captured on load with no interaction) under **Before / after**, and three four-step interaction walks — load, pointer hover, arrow keys, Space — folded inside **How this was tested**, under the Coverage rows they back up.

---

kiln-agent:
## Review

The fix is the right shape and the diff is minimal: decoupling `hoveredValue` from `onFocus` and dropping the `focus-within:` *selected* repaint is exactly what ENG-2288 needs, and the comments left in place explain why each removal is not dead code. One substantive gap below, plus what I verified so you know what is already cleared.

### Pointer-initiated card changes now have **no** focus affordance at all

Every remaining focus indicator for these five elements is `:focus-visible`-gated — `globals.css` `label:has(input[type="radio"]:focus-visible)`, `label[data-fb-scale-cell]:has(…)`, `button:focus-visible` — and there is no plain `:focus` fallback anywhere in the five files (the only `focus:` left is `ranking.tsx`'s `focus:outline-none`). Chromium does **not** match `:focus-visible` on a programmatic `.focus()` when the last user input was a pointer. Probed in Chromium against the same markup shape (sr-only radio inside its label, plus a plain button):

| last interaction before `el.focus()` | radio → `:focus-visible` | button → `:focus-visible` |
| --- | --- | --- |
| none (fresh load) | ✅ | ✅ |
| a mouse click | ❌ | ❌ |
| a keypress | ✅ | ✅ |

That middle row is a live path, not a hypothetical: `shouldFocusOnMount={autoFocusEnabled || hasUserNavigatedRef.current}` (`survey.tsx:1258`) is true after *any* user navigation, so a respondent who clicks **Next** with a mouse — or clicks an option on an auto-progress card — lands on card 2 with focus on an `sr-only` radio and nothing painted. Before this PR that path did get an indicator: `focus-within` matches regardless of modality, so the option row lit up. It lit up in the *selected* colors, which is the bug you are fixing — but the replacement for that path is currently nothing rather than something quieter.

The concrete consequence is a hidden-focus hazard: after a mouse click on Next, the respondent's next keystroke (Space to scroll the card, Enter) acts on a control they never saw was focused, selecting option 1 or ranking item 1. WCAG 2.4.7 is not failed — a keyboard-operating user sets the keyboard-modality flag and gets the ring — so this is a UX gap, not a compliance one.

Suggestion, and it is already the pattern this PR blesses: keep a `focus-within` affordance but in *focus* colors instead of *selected* colors. `consent.tsx:85` does precisely that (`focus-within:border-ring focus-within:ring-ring/50 focus-within:shadow-sm`), and the description explicitly rules it legitimate — "a focus affordance on that element's single control, not a 'this is selected' look". Applying the same to the three option rows would close ENG-2288 (nothing looks answered) without leaving the pointer path blind.

Worth noting for the description too: the coverage table is measured "on load, with no click, keypress or pointer movement", so it structurally cannot observe this case, and "the focus ring still appears on load" holds for the first card and for keyboard-driven navigation only. If you keep the current behaviour deliberately, this belongs under **Open gaps** as the one path that now has no indicator.

### Verified, no action needed

* `ranking.tsx`'s `focus:outline-none` on the item button does **not** defeat the global ring: utilities are emitted `!important` (`globals.css:15`, `tailwind.config.ts` `important: "#fbjs"`), and for important declarations the earlier `components` layer wins — which is what the header comment at `globals.css:279` is protecting.
* `'[tabindex="0"]:not(a)'` closes the `<a tabindex="0">` hole that plain `[tabindex="0"]` would have left open after `a[href]` moved to the fallback. No element uses an anchor as its primary control (`cta.tsx` renders a `<Button>`), so the reordering cannot change which control a CTA or consent card focuses beyond skipping prose links — the intended effect.
* Roving tabindex means a re-visited or prefilled card focuses the **selected** radio, not option 1 (`use-roving-radio-group.test.ts`: `selectedValue: "b"` → `[-1, 0, -1]`), and that row is legitimately painted selected. No regression from dropping `focus-within` there.
* Completeness of the sweep checks out: after this diff `focus-within:` appears exactly once across `packages/survey-ui/src` and `packages/surveys/src` (consent), and `picture-select`, `matrix`, `date`, `file-upload`, `open-text` have no focus↔hover/selected coupling.
* `apps/web/public/js` is gitignored, so no stale prebuilt bundle needs to ride along.

Nit, optional: with `focus-within:` gone, the `outline-none` still in `getOptionContainerClassName` (both selects) is unambiguously dead — the `<label>` never takes focus itself.


---

kiln-agent:
## Review

I traced every claim in the description against the tree at `a6a9ce9`. The diagnosis is right and the fix is the minimal one: the removed classes were **character-for-character** the `isSelected` styling (`focus-within:bg-option-selected-bg focus-within:border-brand` vs `isSelected ? "bg-option-selected-bg border-brand"`), so there is no reading of the old code under which mount focus did not look like an answer. Approving in substance; one thing I would like checked before merge, and a suggestion that closes part of the stated gap.

### What I verified independently

- **The sweep is complete.** `focus-within` and `peer-focus`/`focus:bg-*`/`group-focus` are now absent from every element in `packages/survey-ui/src/components/elements/` except `consent.tsx:85` — and that one is `focus-within:border-ring focus-within:ring-ring/50 focus-within:shadow-sm`, a ring on the element's own single control, not a selected look. Leaving it is the right call, as the description argues.
- **The focus ring cannot have been suppressed by what was removed.** The rules in `globals.css:316-341` land in the `components` layer with `!important` on both `outline` and `box-shadow`, so neither the surviving `outline-none` on the option containers nor the deleted `focus-within:shadow-sm` on the ranking item can win against them. Star/smiley labels carry no `data-fb-scale-cell`, so they fall to the halo rule at `globals.css:328-329`; the number/NPS cells hit the inset rule at `globals.css:364`. Ranking's inner `focus:outline-none` (`ranking.tsx:103`) is likewise overridden, and the ring there is a box-shadow anyway.
- **`a[href]` really did only ever match prose.** No element component in `survey-ui` renders an anchor at all, so the only `<a>` inside a card comes from headline/description rich text — which is exactly the case the demotion targets, and means no element's own primary action moved. The two `tabIndex={0}` sites in `block-conditional.tsx` (491, 501) are `SubmitButton`/`BackButton`, already matched by `button:not(:disabled)`, and Submit precedes Back in DOM despite the `flex-row-reverse`, so a control-less card still focuses Next rather than Back.
- **Handlers came out cleanly** — the only remaining `onFocus`/`onBlur` occurrences in `rating.tsx` and `nps.tsx` are inside the new comments; all four `onMouseEnter`/`onMouseLeave` pairs are intact.
- **No stale-artifact concern.** `apps/web/public/js` is gitignored (`.gitignore:64`), so the bundle is built downstream and there is nothing to re-commit.

### Please check before merge: the remaining indicator is `:focus-visible`-only now

This is the one thing the diff changes that I cannot settle by reading, and it is worth a few minutes because it is the fix's own accessibility premise.

`:focus-within` matched **any** focus, programmatic included. `:focus-visible` does not — whether it matches an `element.focus()` issued before the user has touched a keyboard is an engine heuristic, not a spec guarantee. So on the three choice elements the card previously had two indicators on mount (the wrong selected-fill, plus the ring) and now has exactly one, whose presence is engine-dependent:

| | before | after, if `:focus-visible` matches | after, if it does not |
| --- | --- | --- | --- |
| single/multi-select, ranking | selected-fill **and** ring | ring | **nothing** |

The harness in the description is Chromium-only, and it does show the ring — good, that is the engine most respondents are on. Chromium is also the permissive end of this behaviour. WebKit is the historically conservative one, which matters here because link surveys get opened on iOS. Firefox I would expect to be fine (`-moz-focusring` has always been generous on programmatic focus), so **WebKit is the one run worth adding** — the fixture and script already exist, it should be a `--browser=webkit` away.

If it turns out WebKit paints nothing on mount, the practical impact is bounded (a respondent who has not touched a keyboard yet is arguably not looking for a focus location) but it does hollow out what #8566 added, and the fix would be to key the mount ring off something deterministic — `focusFirstControl` marking its target with a `data-` attribute the ring rule can read — rather than to restore `focus-within`. Not asking for that speculatively; just naming the exit if the check comes back red.

### Suggestion: the `block-conditional.tsx` half of this is unit-testable today

The `Open gaps` section says no automated guard ships because the behaviour is real-browser paint, which jsdom cannot observe. That is true of the five styling changes — but not of the sixth. `focusFirstControl`'s new ordering is pure DOM-selector logic: *given a subtree with a prose link before a checkbox, which node does `querySelector` return?* That is fully observable in happy-dom, needs no paint, no database, and no Playwright — and it is the change most likely to be quietly undone later, because the `:not(a)` and the `??` fallback both look like noise to someone tidying up.

It is currently unreachable only because `focusFirstControl` is a non-exported `const` inside a `.tsx` file (`block-conditional.tsx:46`), which the repo's no-component-unit-tests rule then covers. Lifting it and the two selector constants into a sibling `.ts` module would make it ordinary testable logic under the same rule — the rule is about React components, and this function is not one. Three cases would lock the whole intent: link-before-control → control; control-less card with a link → link; `preferInvalid` with an `aria-invalid` input → that input. Worth doing in this PR while the reasoning is fresh, though I would not block on it.

### Two smaller things

- **`'[tabindex="0"]:not(a)'` (`block-conditional.tsx:30`) is load-bearing but unexplained.** The docstring on `FOCUSABLE_LINK_SELECTOR` explains beautifully why links are demoted, but nothing says why the tabindex clause needed narrowing — and without the `:not(a)`, an author-supplied `<a href tabindex="0">` in prose would still be caught by the primary selector and defeat the demotion entirely. That is a real reason, and it is invisible: the next reader sees a `:not()` guarding against an element type the file never renders and deletes it. One clause on the existing docstring is enough.
- **Star/smiley keyboard preview — a trade-off worth an explicit design nod, not a defect.** With `onFocus` gone, `isActive` falls back to `number <= value`, so a respondent *changing* an existing 4-star answer now arrows through a ring that floats over four still-filled stars until Space commits. That is consistent with #8566's focus≠selection principle and I think it is the right default, but it is a slightly odd frame that the number scale does not have (no fill to be detached from), and the description files it under `Risks` rather than as a decision. Worth someone on design saying "yes, that" out loud.

### On the description itself

`Breaking changes` is correctly unticked — nothing in the diff touches an API, prop, emitted value or stored response, only which styles paint during focus and which node receives it. The `Open gaps` list is unusually honest and matches what I found, including the touch-device caveat: with `onBlur` gone, nothing clears `hoveredValue` on a device that fires `mouseenter` on tap without a paired `mouseleave`, so a tapped cell can keep its preview until the card unmounts. Low impact — single shared state, last tap wins, card unmount clears it — but the open gap names the right hazard.

CI at the time of writing: unit tests, typecheck, linters, SonarQube, CodeQL, Better Auth all green; E2E and Docker build still running.


---

packages/survey-ui/src/components/elements/single-select.tsx
```diff
@@ -373,7 +373,9 @@ function getOptionContainerClassName(isSelected: boolean, disabled: boolean): st
     "relative flex cursor-pointer flex-col border transition-colors outline-none",
     "rounded-option px-option-x py-option-y",
     isSelected ? "bg-option-selected-bg border-brand" : "bg-option-bg border-option-border",
-    "focus-within:border-brand focus-within:bg-option-selected-bg",
+    // No focus-within fill: it repainted the option in the *selected* colors, so the card's
```
kiln-agent, line 376: **Low severity / plausible** — after this change, mount autofocus following a *pointer*-driven navigation leaves focus with no indicator at all, not just without the "answered" look.

`focus-within` matched plain `:focus`; the replacement indicator in `globals.css` is `:focus-visible` only. Chromium (and WebKit) only match `:focus-visible` on a script `focus()` when the previously focused element matched it — i.e. when the last interaction was keyboard. So:

1. Respondent **clicks** "Next" with the mouse. The Next button takes `:focus` but not `:focus-visible`.
2. The next card mounts, `focusFirstControl` (block-conditional.tsx) calls `.focus({preventScroll:true})` on the first option's radio/checkbox.
3. `label:has(input:focus-visible)` does not match → no ring, and with `focus-within` gone there is now no fill either. Focus sits on option 1 completely invisibly.
4. The respondent presses **Space** (a very common "scroll the card" reflex) → the invisible focused option is selected, and on an auto-progress element the card submits and advances.

Before this PR step 4 was still a footgun, but the option was at least visibly marked; now the state that Space acts on is unobservable. Same applies to `multi-select.tsx`, `ranking.tsx`, and the scale cells in `rating.tsx`/`nps.tsx` (their inset ring is also `:focus-visible`-gated).

The PR body's claim that "keyboard users lose nothing … the uniform focus ring still says where focus is" holds only for keyboard-initiated focus; it does not hold for the mount autofocus after a pointer click, which is the common path.

Worth a conscious decision rather than a silent gap. Two shapes that keep the fix intact: (a) skip the mount autofocus when the navigation that caused it was pointer-driven (the ring is there for keyboard/SR users anyway), or (b) mark mount-focused controls with a data attribute and paint the *ring* (not the selected fill) from it, so the indicator no longer depends on the `:focus-visible` heuristic.

---

kiln-agent (commented):
Careful single pass over the whole diff (6 files, 36/-39). I could not find a correctness bug: the `onFocus`/`onBlur` removals in `rating.tsx`/`nps.tsx` leave `hoveredValue` driven only by pointer events with no stale-state path, the `focus-within` removals are colour-only (no layout shift, no tailwind-merge group interaction with the `isSelected`/`isRanked` classes), and I verified that every element whose focus fill was removed still has a `globals.css` ring path (`label:has(input[type=radio|checkbox]:focus-visible)` for single/multi-select and the star/smiley labels, `label[data-fb-scale-cell]…` for the number scale and NPS, `#fbjs button:focus-visible` for the ranking item's button — the components-layer `!important` does beat its own `focus:outline-none`). The `focusFirstControl` reordering is sound: `useRovingRadioGroup`'s roving `tabindex="-1"` radios are still excluded by the unchanged `input` clause, and I found no element in `packages/surveys` or `packages/survey-ui` whose primary control is an anchor (the only `a[href]` inside a card is `element-media`'s "Open in new tab" affordance, which is `opacity-0` until hover — so focusing it was invisible, and deprioritising it is a strict improvement).

One inline comment, low severity, about a gap the PR body does not cover: the replacement indicator is `:focus-visible`-gated, so the mount autofocus after a *pointer*-driven Next paints nothing at all.

Two non-blocking notes, no change required:
- `FOCUSABLE_LINK_SELECTOR` is effectively unreachable. On the mount path `root` is `containerRef`, which always contains the Submit and/or Back `button`s, so the control selector always matches; on the `preferInvalid` path `root` is one element's `<form>`, and every element type that can fail validation renders an input or button. The doc comment's "a card that really has nothing else" case cannot occur — harmless, but the fallback is dead code today.
- `'[tabindex="0"]:not(a)'` also drops `<a tabindex="0">` *without* `href` from both the control selector and the `a[href]` fallback. Nothing in the repo renders that shape, so it is theoretical.

---

packages/survey-ui/src/components/elements/multi-select.tsx
```diff
@@ -92,7 +92,9 @@ const getOptionContainerClassName = (isSelected: boolean, isDisabled: boolean):
     "relative flex flex-col border transition-colors outline-none",
     "rounded-option px-option-x py-option-y",
     isSelected ? "bg-option-selected-bg border-brand" : "bg-option-bg border-option-border",
-    "focus-within:border-brand focus-within:bg-option-selected-bg",
+    // No focus-within fill: it repainted the option in the *selected* colors, so the card's
```
kiln-agent, line 95: **Mount autofocus reached by pointer now paints no focus cue at all** (same applies to `single-select.tsx:376` and `ranking.tsx:83`).

The replacement indicator is `:focus-visible`; the thing being removed is `:focus-within`. Those differ in exactly the case this PR is about — script-moved focus:

- `:focus-within` matched regardless of modality, so the option was always painted while it held focus.
- `:focus-visible` on a programmatic `.focus()` of a non-text control matches only if the last input modality was keyboard (or nothing has happened yet).

Scenario: respondent clicks **Next** with the mouse. The next card is a multi-select; `focusFirstControl` calls `.focus({preventScroll:true})` on option 1's `sr-only` checkbox. Last modality is pointer → `:focus-visible` does not match → `#fbjs label:has(input[type="checkbox"]:focus-visible)` in `globals.css` never paints. With `focus-within:*` gone there is now **zero** on-screen indication that focus sits on option 1, and pressing Space (a common way to scroll a long card) checks it. On ranking the equivalent Space press mutates the ranking.

This is invisible in the PR's captures because they are all on initial page load, where no pointer interaction has occurred yet and Chrome does grant `:focus-visible`. Worth re-checking with the "click Next, then screenshot the next card" path before merging — if it reproduces, the honest fix is a modality-independent (but non-selected-looking) mount indicator rather than relying on `:focus-visible` alone.

---

packages/surveys/src/components/general/block-conditional.tsx
```diff
@@ -27,13 +27,23 @@ const FOCUSABLE_CONTROL_SELECTOR = [
...
 
+/**
+ * Links are focusable but are not what a card asks of the respondent: a headline or subheader may
+ * carry one in its prose (a consent element pointing at a privacy policy), and that link sits
+ * *before* the element's own control in the DOM. Focusing it on mount opened the card with a ring
+ * around a word in the question text, reading as a highlighted suggestion (ENG-2415), and gave a
+ * screen-reader user "Privacy Policy, link" as their orientation instead of the question's control.
+ * Kept as a fallback so a card that really has nothing else still receives focus.
+ */
+const FOCUSABLE_LINK_SELECTOR = "a[href]";
```
kiln-agent: **This fallback is unreachable, so the doc comment above it (and the "Risks" bullet in the description) promise something the code cannot do.**

`FOCUSABLE_CONTROL_SELECTOR` still contains `button:not(:disabled)`, and:

- **Mount path** — `root` is `containerRef`, which always contains the block's `SubmitButton` / `BackButton` rendered *after* the elements. First `querySelector` always hits, so `?? root.querySelector(FOCUSABLE_LINK_SELECTOR)` never evaluates. The only shape with no nav button (`shouldHideSubmitButton` **and** `isFirstBlock`) is an auto-progress element — single-select / rating / nps / picture-select — all of which render radio `<input>`s.
- **`preferInvalid` path** — `root` is an element `<form>`, and every element type renders its own control: CTA renders a real `<Button>`, consent a Radix `<button role="checkbox">`.

Concretely: a consent-style card whose only in-body focusable thing is a prose "Privacy Policy" link now moves mount focus to the **Next button below the card content**, not to the link — the opposite of "Kept as a fallback so a card that really has nothing else still receives focus", and the opposite of the Risks bullet "A card whose only focusable thing is a link still focuses that link — the fallback is kept."

That behaviour may well be the one you want, but please either drop the dead selector or correct the comment, so the next reader does not rely on a safety net that cannot fire.

---

packages/survey-ui/src/components/elements/rating.tsx
```diff
@@ -302,14 +299,6 @@ function Rating({
...
             setHoveredValue(null);
-          }}
-          onFocus={() => {
-            if (!disabled) {
-              setHoveredValue(number);
-            }
-          }}
-          onBlur={() => {
-            setHoveredValue(null);
           }}>
```
kiln-agent, line 306: **Keyboard users lose the star/smiley range preview that pointer users keep.**

Only *mount* focus had to stop painting, but `onFocus`/`onBlur` were removed wholesale, so all keyboard-initiated focus loses the preview too:

- Keyboard respondent tabs into a 5-star rating and presses `ArrowRight` ×3. Focus is on star 4, but all five stars stay `fill-slate-300` — the only cue for "this would be 4 stars" is the ring's position, which they have to count.
- A mouse user hovering star 4 still sees stars 1–4 filled yellow, i.e. the real value preview. Same asymmetry on the smiley scale, where hover still fills and brand-strokes the hovered face.

The Risks section says telling mount focus from keyboard focus is the hard part because `:focus-visible` matches both — but you don't need `:focus-visible` here. `useRovingRadioGroup` already owns the explicit user-initiated moves in `focusValue` (arrows / Home / End). Surfacing the roving position from the hook (or accepting an `onFocusValueChange`) and driving the preview from that would keep parity for keyboard users without reintroducing the mount-focus paint, since mount focus never goes through `focusValue`.

---

kiln-agent (commented):
Code review (correctness pass, effort: high) — 3 findings, no blocking correctness bug.

The core fix is sound: decoupling `hoveredValue` from `onFocus` in `rating.tsx` / `nps.tsx` and dropping `focus-within:bg-option-selected-bg` from the three option rows removes the "looks answered on load" paint without touching selection, and `focusFirstControl`'s reordering correctly prefers the block's own control over a prose link (`querySelector` with a selector list already returns document order, so the two-step `??` is the right shape). `tsc --noEmit` on `@formbricks/survey-ui` is clean, and `@formbricks/surveys` unit tests pass (696) — the 2 failing files are the stale `@formbricks/survey-ui` `dist/` trap documented in AGENTS.md, not this diff.

Verified while reviewing (so these are *not* findings): the replacement rings do exist and win the cascade — `label:has(input[type=radio|checkbox]:focus-visible)` covers the select option labels, `label[data-fb-scale-cell]` covers the number/NPS cells, and `#fbjs button:focus-visible` beats ranking's own `focus:outline-none` because `globals.css` puts it in the earlier `components` layer with `!important`. `useRovingRadioGroup`'s own `onFocus`/`onBlur` are spread onto the `<input>`, so removing the label-level handlers does not disturb the roving tabindex. Radix's `BubbleInput` carries `tabIndex={-1}`, so the consent checkbox `<button>` is what mount focus lands on. `picture-select` and `matrix` genuinely have no focus/selection coupling.

The three comments below are one plausible regression (`:focus-visible` vs `:focus-within` for pointer-initiated navigation), one dead-code/doc-accuracy issue in `focusFirstControl`, and one keyboard-parity regression that I think is cheaper to avoid than the description suggests.

---

packages/survey-ui/src/components/elements/single-select.tsx
kiln-agent: Superseded — my review ran twice and this thread is the earlier, partial pass. The same finding is written up more fully on `multi-select.tsx:95` (https://github.com/formbricks/formbricks/pull/8932#discussion_r3821162094); resolving this one so the discussion stays in one place. Nothing here is withdrawn.

---

packages/survey-ui/src/components/elements/rating.tsx
itsjavi: Fixed in 062b91d — you were right on both halves, including that `focusValue` is the seam. Mount focus reaches the input's own `onFocus` and never `focusValue`, so the hook can tell the two apart without `:focus-visible`; my Risks bullet claiming otherwise was wrong and is gone.

`useRovingRadioGroup` now returns `keyboardValue`, set only on the arrow/Home/End path. The guard is an `isKeyMove` ref that is true exclusively for the synchronous `.focus()` inside `focusValue`, so every other arrival (mount autofocus, a click, Tab entering the group) sets it to `null` — which also clears a stale preview rather than stranding it. `focusValue` re-sets it afterwards to cover focusing the already-focused option, where no focus event fires at all (Home at the first option). Both scales preview from `hoveredValue ?? keyboardValue`.

Measured on the rebuilt bundle, star scale:

| Step | Stars |
| --- | --- |
| mount autofocus | all five `oklch(0.869 0.022 252.894)` — idle |
| `→ →` | 1–3 filled, ring on 3 |
| click option 5, pointer away | 1–5 filled, `checked=[5]` — no preview stuck at 3 |
| Tab out | preview cleared, selection kept |

The number and NPS cells got the same treatment rather than a special case, so hover and arrow keys now paint identically on all four scales. Walk figures in the description are regenerated from this bundle — the old captions said "ring moves to 3, no fill", which is exactly the behaviour you flagged.

---

packages/surveys/src/components/general/block-conditional.tsx
itsjavi: Fixed in 062b91d. Your path analysis holds: `containerRef` always contains the Submit and/or Back button, so the control selector matched first every time and the `??` never evaluated.

Rather than keep a comment describing a safety net that cannot fire, the selector and the fallback are gone. `FOCUSABLE_CONTROL_SELECTOR` is a single list again, and the docblock now says what the code actually does — anchors are excluded on purpose (that exclusion is the ENG-2415 fix), `a` is excluded from `[tabindex="0"]` too so a hand-focusable anchor behaves the same, and nothing is left unfocused because every block renders a nav button inside the root, the one shape without one being a radio group.

The Risks bullet is replaced with the real consequence: a hypothetical future card with no control and only a prose link would receive no focus at all.

---

packages/survey-ui/src/components/elements/multi-select.tsx
itsjavi: Reproduced exactly as you described, and I ran the check you asked for — two-block fixture, real mouse click on Next, then probe card 2's first option on **both** bundles:

| | `:focus` | `:focus-visible` | ring | fill | after `Space` |
| --- | --- | --- | --- | --- | --- |
| card 1 on load | ✓ | ✓ | painted | — | — |
| card 2 after a **mouse** Next, before | ✓ | ✗ | none | `:focus-within` fill | `checked` |
| card 2 after a **mouse** Next, after | ✓ | ✗ | none | none | `checked` |

So: mechanism confirmed, and `Space` checks the option on both bundles — the footgun itself is not introduced here, but this PR does remove the paint that used to mark its target. It's now the first Open gap in the description with these numbers.

I'm not taking either remedy in this PR, because both cut against something load-bearing:

- **(b) a modality-independent mount ring** is precisely what ENG-2415 asks us to stop doing. That ticket's second screenshot is a customer objecting to a ring on the first focusable element on load; painting one deliberately on every card change trades their complaint straight back in.
- **(a) skipping mount autofocus for pointer navigation** would strand VoiceOver and TalkBack users, whose navigation *is* pointer-driven — a tap is a pointer event. That is the regression #8566 fixed ("VoiceOver re-announce the whole survey dialog on every card change").

There's a third shape neither of us listed that I think is the real answer, and it's also out of scope here: on the pointer path, focus the card container (`tabindex="-1"`, already excluded from the ring in `globals.css`) instead of the first control. Nothing paints, nothing is selectable by Space, and the SR announcement is preserved. That is a change to the autofocus model, so it belongs with the ENG-2415 decision rather than bolted onto a styling fix — which is why this PR says `Ref ENG-2415`, not `Fixes`.

Leaving this thread open on purpose: it's a decision for @itsjavi and the ticket, not something I've silently dismissed.

---

itsjavi:
The nine linked figures in the description lost their URLs when the body was re-synced (they render as `[star scale]()`), so here they are as working links. All are the card **on load, with no click, keypress or pointer movement**, before | after, from the two compiled bundles.

Per element:

- [star scale](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-rating-star.png) — a filled yellow star 1 is not "focused", it is a 1-star review
- [smiley scale](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-rating-smiley.png)
- [NPS](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-nps.png)
- [single-select](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-single-select.png)
- [ranking](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-ranking.png) — the first item opened in the *ranked* colours while its rank badge was still empty
- [consent](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/pair-consent.png) — focus moves from the prose link to the checkbox

Interaction walks on the **after** bundle (load → pointer hover → two arrow presses → Space):

- [star scale](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-rating-star.png) — hover previews 1–4, arrows preview 1–3, Space commits
- [number scale](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-rating-number.png) — the grey follows the pointer and the arrow keys, but not the card opening
- [ranking](https://raw.githubusercontent.com/formbricks/formbricks/assets-itsjavi-prs/pr-8932/walk-ranking.png) — no fill on focus, ring still on the item's button

The three figures embedded in the description (number scale, multi-select, CTA-with-links) are unaffected.

---

jobenjada (approved):
thanks for visual proof, looks good :)

PR #8933, opened 2026-08-20
fix(auth): stop paging on client-caused OAuth state rejections (ENG-2471)

xernobyl:
Fixes ENG-2471

## What & why

`BetterAuthError: State mismatch: verification not found` cleared the ENG-2037 gate and pages — 52 Sentry events in 14 days, 25 of them on `5.3.4` (`FORMBRICKS-16G`). ENG-2037 enumerated two shapes of expected-but-logged-at-error rejection (bare string codes, `APIError`s); this is a third it never covered — a `StateError extends BetterAuthError`, which `isAPIError()` is false for, so it clears `if (cause && !isAPIError(cause))` and reaches Sentry.

The gate now also skips a `StateError` whose `code` is client- or timing-caused, matching on **`code`** rather than the message — the message is not a contract. **Exactly one of the five qualifies:**

| code | thrown when (our strategy) | verdict |
| --- | --- | --- |
| `state_mismatch` | no verification record for this `state` — purged after its 10-minute TTL, already consumed, or never issued — or the parsed state is past `expiresAt` | **suppressed** |
| `state_not_found` | the callback carried no `state` — **never reaches this gate on 1.7** | kept (inert) |
| `state_security_mismatch` | the state does not match the stored one, or the signed `state` cookie fails verification | kept |
| `state_generation_error` | the adapter could not write the verification row | kept — a real fault |
| `state_invalid` | undecryptable state — cookie branch only, **unreachable on this config** | kept |

Everything still logs at `error`. The gate fails **closed**: anything not positively identified as an unactionable code is still captured.

### `state_not_found` — kept out of the set, and on 1.7 that is inert

Two earlier drafts of this section were wrong in opposite directions, so here is what the running code actually does. Smoke-tested against 1.7.0 on the rebased branch: **an OAuth callback with no `state` never produces a `StateError` at all.** The callback route short-circuits before `parseGenericState` (`better-auth/dist/api/routes/callback.mjs:72-76`):

```js
if (!state) { c.context.logger.error("State not found", error); throw c.redirect(`…error=state_not_found`); }
```

That second argument is the OAuth `error` **query parameter**, not an `Error`, so our `cause` lookup finds nothing, the Sentry gate's `cause &&` is already false, and the event is logged with no `stateErrorCode`. It was never captured, so it was never part of the over-capture problem, and the `StateError` carrying that code (`state.mjs:90`) is unreachable from this route.

Listing it or omitting it therefore changes nothing today. It stays omitted so that a future upstream change routing it through as a real `StateError` pages, rather than being dropped by a stale allow-list — and the test asserting that is labelled as the defensive contract it is.

**One gap this does not close, found the same way:** because that path carries no `Error`, an IdP that stops echoing `state` — a provider-wide sign-in outage — produces **no Sentry event at all**, only a log line. Independent of this change and not fixable from this gate.

### Replay visibility, stated rather than implied

A *replayed* callback — the same `state` presented again after the real flow consumed it — also lands on `state_mismatch`, so replay attempts stop appearing in Sentry. Accepted: the replay fails closed (the verification record and the IdP's authorization code are both single-use), so this is visibility into a *failed* attempt, not a control. The events stay in the log with their code, which is where a campaign shows up as volume.

## Read the right branch — this is where the first draft was wrong

`better-auth/dist/state.mjs` has two branches that throw *different codes for the same underlying cause*, and `auth.ts` sets `secondaryStorage`, so Better Auth resolves `storeStateStrategy` to `"database"` — the cookie branch never executes here. Two conclusions changed once that was checked against the throw sites rather than assumed:

- **`state_invalid` is unreachable**, so justifying its capture by a cross-replica `BETTER_AUTH_SECRET` divergence was wrong. That divergence actually surfaces as `state_security_mismatch`, because the signed state cookie cannot be verified with a different secret. It is a genuine outage, so that code keeps paging.
- **The dominant benign case is not suppressed.** The state cookie's `maxAge` is 300s against the record's 10 minutes, so a user who spends 5–10 minutes at the identity provider loses the cookie first and also lands on `state_security_mismatch`. So **this likely leaves residual noise** — it closes the reported shape, not the whole class. Code cannot separate that from a forged callback, which is why the decision is deferred to the `auth.path` tag from ENG-2259 rather than guessed. Suppressing a signal before measuring it is how ENG-2037 left this shape behind in the first place.

The state code now reaches the log context as `stateErrorCode`, extracted once and used for both the log and the gate. Without it the "still visible in logs" justification was hollow — only upstream's message text distinguished a Redis-eviction burst from anything else. It is never widened to the error's `details`, which holds the raw single-use `state`.

## Breaking changes

None — a Sentry capture filter. No API, schema, env or config surface.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The one client-caused code is not captured | unit | One code path, asserted with the three real message literals upstream throws for it (verification not found, cookie-branch label, request expired). The gate reads `code`, so those messages are documentation rather than three behaviours — the behavioural assertion is that each is still logged at `error`, i.e. suppression is Sentry-only |
| The other four still are | unit | `state_invalid`, `state_security_mismatch`, `state_generation_error` and `state_not_found` each assert a `captureException` with the `component: "better-auth"` tag. The last two are defensive contracts, not live paths — `state_invalid` is cookie-branch-only and `state_not_found` never arrives as a `StateError` (above) |
| The gate fails closed | unit | A `BetterAuthError` with no `code`, and a look-alike carrying `name: "BetterAuthError"` + `code: "state_mismatch"` that is not one — both still captured |
| **The REAL `StateError` is suppressed** | unit (real instance) | `StateError` is internal to `better-auth/dist/state.mjs` and unimportable, so the stand-in tests would keep passing if the real shape drifted. A real `betterAuth` (memoryAdapter, `logger: betterAuthLogger`) is driven at `/api/auth/callback/google?state=no-such-state` — the reported failure — and asserted not captured, plus an anti-vacuity assertion that it was logged at all |
| The real error's shape is pinned | unit (real instance) | Same throw captured through a recording logger: asserts `name === "BetterAuthError"` and `code === "state_mismatch"`, so an upstream rename fails loudly instead of quietly restoring the noise |
| The guards bind | manual | Mutation-tested: removing the gate fails **4** tests including the contract test; suppressing all five codes fails **4**; and putting `state_not_found` back into the suppressed set fails **1**, so the narrowing above is itself pinned rather than a comment |
| Runs on 1.7.0, which is now `main` | manual, live | #8835 has merged, so this branch is rebased onto 1.7.0 and re-verified there rather than on the 1.6.23 it was written against. Re-read the throw sites on 1.7.0: `state_invalid` is still inside `if (storeStateStrategy === "cookie")` (`state.mjs:108` in `:93`), the strategy default is unchanged, and the cited `api/index.mjs:210` context-discarding `logger?.error` is still exactly there. Suite 59/59 on 1.7.0, including the contract test that provokes a real `StateError` |
| The gate through #8835's pinned legacy callback | manual, live | A combination neither PR could test alone. A bogus `state` sent to the **pinned** `/api/auth/oauth2/callback/openid` — which `legacy-sso-callback.ts` rewrites — still logs `authPath: "/callback/*"`, `stateErrorCode: "state_mismatch"`. Confirms the external/internal path split holds end to end: observability keys off the internal `/callback/`, not the external `/oauth2/`, exactly as #8835 documents |
| Both codes end to end through the real route | manual, live | The contract test drives a real `betterAuth`, but nothing covered the two request-scoped contexts *composing* through our own route handler. Ran the app: `?state=no-such-state&code=…` → `302 …error=state_mismatch`, logged `ERROR … authPath: "/callback/*" httpMethod: GET stateErrorCode: "state_mismatch"`; `?code=…` with no `state` → same shape with `state_not_found`. That is the justification for suppressing anything — a suppressed event stays queryable by the value that suppressed it — holding live, with ENG-2259's `authPath` and this PR's `stateErrorCode` on one line. Also confirms `authPath` is the *parameterized* label, never a raw path |

`pnpm test` (this suite 59/59, full suite 8311 uncached), `pnpm typecheck` (27/27), `pnpm lint` (0 errors), `pnpm format:check` all green.

**Open gaps**

- **No measurement of the genuine-failure fraction.** ENG-2037's step 1 was never run for this shape and still has not been — I cannot read Sentry. Everything above is reasoning from the throw sites plus tests, not from event data. This is why only the two codes I can justify from code are suppressed, and why `state_security_mismatch` is explicitly left paging.
- **Residual noise is expected**, per the section above. The follow-up is to size `state_security_mismatch` once ENG-2259's `auth.path` tag has produced a few days of tagged events, then decide. Not this PR.
- **That follow-up is gated on a deploy, not a merge.** ENG-2537 shows with `git merge-base` that #8907 is not in production — prod has been on `5.3.4` since Aug 11, #8907 merged Aug 19 — so the `auth.path` tag is producing no data yet and the deferred decision cannot be taken until a release carries it. Flagged so it does not quietly never happen.
- No log-based alert rule exists for `stateErrorCode`, so a suppressed burst is discoverable rather than announced. **Now tracked as ENG-2551**, filed off this review — which also surfaced a sharper case than the eviction one this PR had: replicas that do not share one Redis fail **100% of SSO sign-ins** onto the suppressed code, because the record lookup precedes the signed-cookie check and the shared `BETTER_AUTH_SECRET` keeps the cookie valid. That is a deploy-time total outage with no Sentry event. The docblock now names it; the alert is out of scope here.
- Not exercised against a deployed image — this is a logger gate with no request surface, so a staging pass adds little beyond what the real-instance test covers.

**Risks**

- A *hard* Redis outage does not hide behind this: `secondary-storage.ts` rethrows on connect failure, which is not a `StateError` and still pages. Only a reachable-but-empty Redis (eviction, flush, failover to an empty replica) yields `state_mismatch`, and those are real login failures now visible only in the log.
- The three kept codes reach Sentry with no stack — `BetterAuthError` sets `this.stack = ""`. Pre-existing, not introduced here, but it caps how actionable those captures are.

### DIFF
--- a/apps/web/modules/auth/lib/better-auth-observability.test.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.test.ts
@@ -413,6 +413,8 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
   // expired, or no `state` arrived at all. Nothing to act on, so these must not page.
   test.each([
     ["state_mismatch", "State mismatch: verification not found"],
+    // Cookie-branch message, kept as a label only: the code is what the gate reads, and this variant
+    // is unreachable on our database strategy.
     ["state_mismatch", "State mismatch: auth state cookie not found"],
     ["state_mismatch", "Invalid state: request expired"],
     ["state_not_found", "State not found in OAuth callback"],
@@ -430,10 +432,12 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
   /**
    * The other three codes stay captured, and each for its own reason:
    * - `state_generation_error` — the adapter could not write the verification row: a real fault.
-   * - `state_invalid` — undecryptable state, which is what replicas running different
-   *   `BETTER_AUTH_SECRET` values look like. Suppressing it would bury an outage.
-   * - `state_security_mismatch` — the state does not match the stored one: the forged/mixed-up callback
-   *   shape. ENG-2471 defers judging it until the ENG-2259 `auth.path` tag has sized it.
+   * - `state_security_mismatch` — the state does not match the stored one, OR the signed state cookie
+   *   fails verification. That second half is where a cross-replica `BETTER_AUTH_SECRET` divergence
+   *   surfaces (a real outage) and also where the benign 5–10 minute cookie-expiry case lands, so it is
+   *   mixed and ENG-2471 defers judging it until the ENG-2259 `auth.path` tag has sized the split.
+   * - `state_invalid` — cookie-branch only, so unreachable on this configuration. Kept because
+   *   suppressing a code that never fires buys nothing.
    */
   test.each([
     ["state_generation_error", "Unable to create verification"],
@@ -447,6 +451,38 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
     expect(Sentry.captureException).toHaveBeenCalledWith(stateError, {
       tags: { component: "better-auth" },
     });
+  });
+
+  /**
+   * The log field is the whole justification for suppressing anything: a suppressed event survives only
+   * in the application log, so it has to be queryable by the same value that suppressed it. Without this
+   * test the field can be deleted and the suite stays green — verified, it did.
+   *
+   * Recorded for captured codes too, so one query covers the class rather than only its quiet half.
+   */
+  test.each([
+    ["state_mismatch", "suppressed"],
+    ["state_security_mismatch", "captured"],
+  ])("records %s on the log context (%s)", (code) => {
+    log("error", "State mismatch", new StateErrorLike("State mismatch", code));
+
+    expect(logger.withContext).toHaveBeenCalledWith({
+      source: "better-auth",
+      stateErrorCode: code,
+    });
+  });
+
+  // The raw `state` is a single-use credential for the in-flight flow and lives on the error's
+  // `details`. Only the code is ever read, so it cannot reach the log through this path.
+  test("never puts the raw state value on the log context", () => {
+    const stateError = new StateErrorLike("State mismatch: verification not found", "state_mismatch");
+    Object.assign(stateError, { details: { state: "super-secret-state-value" } });
+
+    log("error", "State mismatch: verification not found", stateError);
+
+    expect(JSON.stringify(vi.mocked(logger.withContext).mock.calls)).not.toContain(
+      "super-secret-state-value"
+    );
   });
 
   // The gate fails CLOSED: anything it cannot positively identify as one of the two suppressed codes is
--- a/apps/web/modules/auth/lib/better-auth-observability.test.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.test.ts
@@ -437,11 +437,12 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
    *   mixed and ENG-2471 defers judging it until the ENG-2259 `auth.path` tag has sized the split.
    * - `state_invalid` — cookie-branch only, so unreachable on this configuration. Kept because
    *   suppressing a code that never fires buys nothing.
-   * - `state_not_found` — no `state` on the callback at all. A real flow cannot produce this (the IdP
-   *   echoes the state it was given), so it is either a bare scanner or something upstream dropping the
-   *   parameter — an IdP regression, a proxy, or a callback rewrite mishandling the query string. The
-   *   latter is a provider-wide sign-in outage and this is its canary, so it keeps paging. None of the
-   *   reported FORMBRICKS-16G events are this code.
+   * - `state_not_found` — a defensive entry rather than a live one. Verified against 1.7.0 that the
+   *   callback route short-circuits a missing `state` before any `StateError` is thrown, and logs it
+   *   with the OAuth `error` query param rather than an `Error` — so it reaches neither this gate nor
+   *   Sentry, and the assertion below is a contract for a shape the callback route does not currently
+   *   produce. Kept captured so that if upstream ever does route it through as a real `StateError`, it
+   *   pages instead of being dropped by a stale allow-list.
    */
   test.each([
     ["state_generation_error", "Unable to create verification"],
--- a/apps/web/modules/auth/lib/better-auth-observability.test.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.test.ts
@@ -409,15 +409,14 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
     vi.clearAllMocks();
   });
 
-  // Client- or timing-caused: the state cookie expired, the verification record was purged, the request
-  // expired, or no `state` arrived at all. Nothing to act on, so these must not page.
+  // Client- or timing-caused: the verification record was purged or already consumed, or the parsed
+  // state is past its `expiresAt`. Nothing to act on, so these must not page.
   test.each([
     ["state_mismatch", "State mismatch: verification not found"],
     // Cookie-branch message, kept as a label only: the code is what the gate reads, and this variant
     // is unreachable on our database strategy.
     ["state_mismatch", "State mismatch: auth state cookie not found"],
     ["state_mismatch", "Invalid state: request expired"],
-    ["state_not_found", "State not found in OAuth callback"],
   ])("does not capture the client-caused %s (%s)", (code, message) => {
     const stateError = new StateErrorLike(message, code);
 
@@ -438,11 +437,17 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
    *   mixed and ENG-2471 defers judging it until the ENG-2259 `auth.path` tag has sized the split.
    * - `state_invalid` — cookie-branch only, so unreachable on this configuration. Kept because
    *   suppressing a code that never fires buys nothing.
+   * - `state_not_found` — no `state` on the callback at all. A real flow cannot produce this (the IdP
+   *   echoes the state it was given), so it is either a bare scanner or something upstream dropping the
+   *   parameter — an IdP regression, a proxy, or a callback rewrite mishandling the query string. The
+   *   latter is a provider-wide sign-in outage and this is its canary, so it keeps paging. None of the
+   *   reported FORMBRICKS-16G events are this code.
    */
   test.each([
     ["state_generation_error", "Unable to create verification"],
     ["state_invalid", "State invalid: Failed to decrypt or parse auth state"],
     ["state_security_mismatch", "State mismatch: OAuth state parameter does not match stored state"],
+    ["state_not_found", "State not found in OAuth callback"],
   ])("still captures %s", (code, message) => {
     const stateError = new StateErrorLike(message, code);
 
--- a/apps/web/modules/auth/lib/better-auth-observability.test.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.test.ts
@@ -429,7 +429,7 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
   });
 
   /**
-   * The other three codes stay captured, and each for its own reason:
+   * The other four codes stay captured, and each for its own reason:
    * - `state_generation_error` — the adapter could not write the verification row: a real fault.
    * - `state_security_mismatch` — the state does not match the stored one, OR the signed state cookie
    *   fails verification. That second half is where a cross-replica `BETTER_AUTH_SECRET` divergence
@@ -491,8 +491,8 @@ describe("betterAuthLogger — OAuth state errors (ENG-2471)", () => {
     );
   });
 
-  // The gate fails CLOSED: anything it cannot positively identify as one of the two suppressed codes is
-  // still captured. A wrong answer should add noise, never silence a fault.
+  // The gate fails CLOSED: anything it cannot positively identify as THE one suppressed code
+  // (`state_mismatch`) is still captured. A wrong answer should add noise, never silence a fault.
   test("captures a BetterAuthError carrying no code", () => {
     const bare = new BetterAuthError("something upstream broke");
 
--- a/apps/web/modules/auth/lib/better-auth-observability.test.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.test.ts
@@ -1,4 +1,7 @@
+import { BetterAuthError } from "@better-auth/core/error";
 import * as Sentry from "@sentry/nextjs";
+import { betterAuth } from "better-auth";
+import { memoryAdapter } from "better-auth/adapters/memory";
 import { APIError } from "better-auth/api";
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { prisma } from "@formbricks/database";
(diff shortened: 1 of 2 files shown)


xernobyl:
Re-review, security review and a live smoke test. One substantive change, plus one thing on the ticket that has moved since this was written.

## Narrowed the suppression set — `state_not_found` keeps paging

This reverses part of my own earlier recommendation on the ticket, which had listed three codes to suppress.

`state_not_found` means the callback arrived with **no `state` parameter at all**. A real user-initiated flow cannot produce that — the IdP echoes back the state it was given. So it is either a bare scanner hitting the callback path, or something upstream dropping the parameter: an IdP regression, a proxy stripping the query string, or a callback-URL rewrite mishandling it. That last one stopped being hypothetical this week — #8835 routes *every* SSO callback through `legacy-sso-callback.ts`. A provider that stops returning `state` is a total sign-in outage for that provider, and this code is its canary.

And none of the reported FORMBRICKS-16G events are this code — they are all `verification not found`, i.e. `state_mismatch`. So suppressing it would have traded an unmeasured amount of scanner noise for the loss of that canary, on no evidence — the opposite of what the ticket asks ("investigate the split before suppressing"). If it turns out to be scanner-dominated once the `auth.path` tag is actually deployed, adding it back is one line and will be data-backed.

Net effect: this PR now suppresses exactly one code, `state_mismatch`, which is the one shape the ticket reports and the only one with a deterministic argument behind it.

## The prerequisite this PR defers to is blocked on a release, not a merge

Worth flagging because it changes what "later" means here. This PR's docblock defers the `state_security_mismatch` judgement to ENG-2259's `auth.path` tag. ENG-2537 (filed this morning) shows with `git merge-base` that **#8907 is not in production** — prod has been on `5.3.4` since Aug 11, #8907 merged Aug 19 — so that tag is not producing data and won't until a release carries it. Part 2 of this work is therefore gated on a deploy, not on a merge. Recorded on the ticket too, so it does not silently never happen.

## Security review

- **The security-relevant codes all stay captured.** `state_security_mismatch` (forgery, mix-up, and cross-replica `BETTER_AUTH_SECRET` divergence), `state_generation_error` (adapter fault), and now `state_not_found`. The gate fails closed: anything not positively identified is captured.
- **Not spoofable.** The gate requires `instanceof BetterAuthError` *and* an exact upstream code; neither is attacker-settable. There is a test that a look-alike object with `name: "BetterAuthError"` and a suppressed `code` is still captured.
- **No credential in the log.** `getStateErrorCode` reads only `code`, never `details`, which holds the raw `state` — a single-use credential for the in-flight flow. Verified live below that `authPath` is the *parameterized* label (`/callback/*`), not a raw path, so the `reset-password/:token` class of leak cannot occur here either.
- **Replay visibility is the one real loss**, and it was implicit rather than stated. A replayed callback lands on `state_mismatch`, so replay attempts stop appearing in Sentry. Judged acceptable — the replay fails closed, since the verification record and the IdP's authorization code are both single-use, so this is visibility into a *failed* attempt, not a control. Now written down in the docblock rather than left to inference.
- **Residual:** a Redis eviction or failover drops in-flight verification records and produces `state_mismatch` for genuine login failures, which this suppresses. A hard Redis outage still pages (`secondary-storage.ts` rethrows, not a `StateError`). There is no log-based alert on `stateErrorCode` volume today, so a burst is discoverable rather than announced — worth its own follow-up, not this PR.

## Smoke test — live, through the real route

The contract test already drives a real `betterAuth` instance, but nothing covered the two request-scoped contexts composing through our actual route handler. Ran the app and hit the callback:

| request | result |
| --- | --- |
| `?state=no-such-state&code=…` | `302 → /api/auth/error?error=state_mismatch`, logged `ERROR … authPath: "/callback/*" httpMethod: GET stateErrorCode: "state_mismatch"` |
| `?code=…` (no `state`) | `302 → …error=state_not_found`, same shape with `stateErrorCode: "state_not_found"` |

That is the whole justification for suppressing anything — a suppressed event has to stay queryable by the value that suppressed it — and it now holds end to end, with ENG-2259's `authPath` and this PR's `stateErrorCode` on the same line.

Gates: typecheck 27/27, 8311 tests (uncached), lint 0 errors, format clean.

xernobyl:
Rebased onto `main` now that #8835 has merged, so this runs on Better Auth **1.7.0** rather than the 1.6.23 it was written against. Clean rebase, no conflicts. Re-reviewed and smoke-tested on the new base — and the smoke test disproved something I had written into this PR, so that is corrected rather than left standing.

## The correction: `state_not_found` never reaches this gate on 1.7

I had claimed keeping it captured preserved a canary for "IdP stops echoing `state`". Verified against a running 1.7.0 instance: **an OAuth callback with no `state` never produces a `StateError` at all.** The callback route short-circuits before `parseGenericState` (`api/routes/callback.mjs:72-76`):

```js
if (!state) { c.context.logger.error("State not found", error); throw c.redirect(`…error=state_not_found`); }
```

That second argument is the OAuth `error` **query parameter**, not an `Error` — so our `cause` lookup finds nothing, the Sentry gate's `cause &&` is already false, and it logs with no `stateErrorCode`. It was never captured, so it was never part of the over-capture problem, and the `StateError` carrying that code (`state.mjs:90`) is unreachable from this route.

So listing it or omitting it changes nothing today. It stays omitted as a defensive entry — if upstream ever routes it through as a real `StateError` it should page, not be dropped by a stale allow-list — and the test is now labelled as that contract instead of implying live behaviour. Both earlier drafts of this reasoning were wrong, in opposite directions; the docblock now records what the code does.

**A gap this does not close, found the same way:** because that path carries no `Error`, an IdP that stops echoing `state` — a provider-wide sign-in outage — produces **no Sentry event whatsoever**, only a log line. Independent of this change and not fixable from this gate; worth its own ticket.

## Re-review on 1.7.0

Re-read the throw sites rather than trusting the 1.6.23 analysis: `state_invalid` is still inside `if (storeStateStrategy === "cookie")` (`state.mjs:108` within `:93`), the `storeStateStrategy` default is unchanged, both TTLs are unchanged, and the cited `api/index.mjs:210` context-discarding `logger?.error` is still exactly there. So the suppression set is still correct, and `state_mismatch` remains the only live suppression.

## Smoke test — including one combination neither PR could test alone

| check | result |
| --- | --- |
| `state_mismatch` via the current path | `302 …error=state_mismatch`, logged `authPath: "/callback/*"` `stateErrorCode: "state_mismatch"` |
| **`state_mismatch` via #8835's pinned legacy callback** | Same, on `/api/auth/oauth2/callback/openid` — which `legacy-sso-callback.ts` rewrites. Confirms observability keys off the **internal** `/callback/`, not the external `/oauth2/`, exactly as #8835 documents |
| `state_not_found` | logs `"State not found"` with `authPath`, **no** `stateErrorCode`, no capture — the behaviour described above |

Gates on the rebased branch: suite 59/59, typecheck 27/27, lint 0 errors, format clean.

apps/web/modules/auth/lib/better-auth-observability.test.ts
```diff
@@ -380,6 +383,213 @@ describe("betterAuthLogger (Sentry capture gating, ENG-2037)", () => {
...
+    log("error", message, stateError);
+
+    expect(Sentry.captureException).not.toHaveBeenCalled();
+    // Still visible in the application log — suppression is Sentry-only. A burst of these is how a
+    // Redis eviction would present, since verification records live in Redis only.
+    expect(contextLoggerMock.error).toHaveBeenCalledWith(message);
+  });
+
+  /**
+   * The other three codes stay captured, and each for its own reason:
```
kiln-agent: Stale count from the pre-narrowing draft: "The other three codes stay captured" now precedes **four** bullets (`state_generation_error`, `state_security_mismatch`, `state_invalid`, `state_not_found`) and a `test.each` with four rows. `d75af25` moved `state_not_found` from the suppressed set into this block but left the count at three.

Concretely misleading: a reader counting the kept codes against `UNACTIONABLE_STATE_ERROR_CODES` (one entry) and `state.mjs`'s five throw sites gets 1 + 3 = 4 and concludes a code is unaccounted for — the same class of comment/code drift this PR's third commit was written to fix.

```suggestion
   * The other four codes stay captured, and each for its own reason:
```

---

apps/web/modules/auth/lib/better-auth-observability.test.ts
```diff
@@ -380,6 +383,213 @@ describe("betterAuthLogger (Sentry capture gating, ENG-2037)", () => {
...
+    Object.assign(stateError, { details: { state: "super-secret-state-value" } });
+
+    log("error", "State mismatch: verification not found", stateError);
+
+    expect(JSON.stringify(vi.mocked(logger.withContext).mock.calls)).not.toContain(
+      "super-secret-state-value"
+    );
+  });
+
+  // The gate fails CLOSED: anything it cannot positively identify as one of the two suppressed codes is
```
kiln-agent: Same stale count in the other direction: "one of the **two** suppressed codes" describes `f5f1769`'s set (`{state_mismatch, state_not_found}`). After `d75af25` narrowed it, `UNACTIONABLE_STATE_ERROR_CODES` holds exactly one code.

This one is worse than a typo because the comment sits directly above the fail-closed tests: someone reading "two suppressed codes" here and then looking for the second one will either add `state_not_found` back to the set — silently undoing the narrowing that `d75af25` argued for at length — or assume a code is missing from it. The PR body already says "exactly one of the five qualifies"; the tests should agree.

```suggestion
  // The gate fails CLOSED: anything it cannot positively identify as the suppressed code is
```

---

kiln-agent:

I re-derived every load-bearing claim in the diff from `node_modules/better-auth@1.7.0` rather than from the description, and independently re-ran the mutation checks. Everything holds. Details below, because the value of this PR is entirely in whether its reasoning about upstream is right, and it is.

### What I verified against the 1.7.0 source

| Claim in the docblock | Verified |
| --- | --- |
| `storeStateStrategy` resolves to `"database"` here | `create-context.mjs:136` → `isStateful ? "database" : "cookie"`, and `store-capabilities.mjs` is literally `!!options.database \|\| !!options.secondaryStorage`. `auth.ts` sets both. ✔ |
| The cookie branch never executes, so `state_invalid` is unreachable | `state.mjs:107-112` sits inside `if (storeStateStrategy === "cookie")` at `:93`. ✔ |
| `state_not_found` never arrives as a `StateError` | `callback.mjs:72-76` short-circuits with `logger.error("State not found", error)` where `error` is the OAuth query param, not an `Error`. The `StateError` at `state.mjs:90` is unreachable — I also checked every other `parseGenericState`/`parseState` caller (`oauth2/state.mjs`, `plugins/oauth-proxy`, and the plugin dirs); the only one reachable from this app is the callback route, and oauth-proxy logs at `warn`. ✔ |
| The 300s cookie vs 600s record split | `state.mjs:72` sets the `state` cookie at `maxAge: 300`; `oauth2/state.mjs` sets `expiresAt: Date.now() + 600 * 1e3` and the record at +10 min. So <5 min → success, 5–10 min → `state_security_mismatch`, >10 min → `state_mismatch`. The description's residual-noise conclusion is exactly right, and `skipStateCookieCheck` is not set anywhere in `apps/web`, so that third check is live. ✔ |
| A hard Redis outage still pages | `internal-adapter.mjs:721` awaits `secondaryStorage.get` with **no** try/catch, and `secondary-storage.ts:27-29` rethrows on connect failure. So it surfaces as a non-`BetterAuthError` → `getStateErrorCode` returns undefined → captured. This was the claim most likely to be wrong (a swallowed error would have made a hard outage invisible); it isn't. ✔ |
| Verification records are Redis-only | `verification` in `auth.ts` sets only `storeIdentifier`, so `findVerificationValue` hits `if (!options.verification?.storeInDatabase) return null` — no DB fallback. Confirms eviction → `state_mismatch`. ✔ |

Two things I checked that the docblock doesn't state and that support the approach:

- **`StateError` is the only `BetterAuthError` subclass in 1.7.0** (`grep "extends BetterAuthError"` → one hit), and `APIError` extends `better-call/error`, not `BetterAuthError`. So `getStateErrorCode` reading `code` off any `BetterAuthError` cannot pick up a non-state error today, and the `stateErrorCode` log field can't be misnamed by accident.
- **There is exactly one `Sentry.captureException` in the auth surface**, and `onAPIError.onError` is deliberately unset (`route.ts:52`). So the fix is complete — nothing re-captures the same throw. The redirect that follows is an `APIError`, already excluded by the existing gate.

### Behaviour on the real path

Drove a real `betterAuth` through the production logger. One log line per failure, and:

```
message: "Failed to parse state"
context: { source: "better-auth", stateErrorCode: "state_mismatch" }
captureException calls: 0
```

Worth noting because it's the sharp end of the "still greppable" argument: upstream logs *every* state error under the same message, so `stateErrorCode` is not a convenience — it's the only thing that distinguishes these in the log at all. Good call adding it in the same change rather than after.

### Mutation checks, re-run independently

| Mutation | Result |
| --- | --- |
| Remove `!isUnactionableStateError(...)` from the gate | 4 fail, including `is NOT captured to Sentry` (the real-`StateError` contract test) |
| Suppress all five codes | 4 fail (`still captures …`) |
| Drop `...(stateErrorCode && { stateErrorCode })` | 2 fail |

59/59 green unmutated. The `IS_PRODUCTION: true` / `SENTRY_DSN` mock at the top of the file means the `not.toHaveBeenCalled()` assertions are load-bearing rather than vacuous — worth saying explicitly, since that is the usual way a suppression test passes for the wrong reason.

### Two non-blocking notes

**1. Please file the log-alert ticket now rather than leaving it as "worth a rule".** The suppressed bucket is the *only* automated signal for a class of total-outage misconfiguration, not just for eviction: a deployment whose replicas don't share one Redis writes the verification record on pod A and looks it up on pod B. The lookup is the first check in the database branch, before the signed-cookie check, so that fails 100% of SSO sign-ins and lands on `state_mismatch` — suppressed, with the signed cookie verifying fine because the secret *is* shared. The description's risk list names eviction/flush/failover; the split-Redis case is the same blind spot but arrives at deploy time rather than at runtime, which is when someone is most likely to be watching Sentry and not the logs. Nothing to change in this diff — just don't let the rule land in the same "later" bucket the `auth.path` follow-up is already sitting in, especially given ENG-2537 shows that one is gated on a release.

If you'd rather not depend on an external alert rule at all, a sampled capture (keep ~1 in 50 `state_mismatch` events) would cut the noise ~50× while leaving a burst visible in Sentry as volume. I'd take the log rule over this — it's deterministic and the field is already there — but it's a one-line alternative if the rule turns out to be awkward.

**2. Description nits.** The `Both codes end to end through the real route` row is separated from the table above it by a blank line, so it renders as its own headerless one-row table. And the three `state_mismatch` rows in the suppressed `test.each` exercise one code path with three inert message literals (the gate reads `code`) — real messages, so they're useful as documentation of what upstream throws, but they aren't three behaviours; the coverage table reads as if they are.

Everything else — narrowing to the one code the ticket actually reports, keeping `state_security_mismatch` paging until the `auth.path` tag can size it, the fail-closed gate, never reading `details` — I agree with as written, including the two self-corrections. CI is green across all 19 checks.

---

kiln-agent (commented):
Single careful pass over the diff (`ab8fc21...655d37a`), hunting correctness bugs. I verified every substantive claim in the new docblock against the installed `better-auth@1.7.0` / `@better-auth/core@1.7.0` sources rather than taking them on trust, and they hold:

- `StateError extends BetterAuthError` imported from the **same** `@better-auth/core/error` module `better-auth/dist/state.mjs` imports, and there is exactly one `@better-auth/core` in `node_modules`, so the `instanceof` gate binds. `APIError` extends `better-call`'s `APIError`, **not** `BetterAuthError`, so `getStateErrorCode` cannot mislabel a handled `APIError` as a state error.
- `storeStateStrategy` → `"database"`: `context/store-capabilities.mjs` is `!!options.database || !!options.secondaryStorage`, and `auth.ts` sets both. Cookie branch (and therefore `state_invalid`) is unreachable, as documented.
- "verification records live in Redis only" is right: `verification.storeInDatabase` is unset, and `findVerificationValue` returns `null` without touching the DB when `secondaryStorage` misses.
- The `BETTER_AUTH_SECRET`-divergence claim is right for the right reason: `verification.storeIdentifier: "hashed"` is a plain SHA-256 (`db/verification-token-storage.mjs`), keyless — so a secret split cannot cause a *lookup* miss (`state_mismatch`); it surfaces on `getSignedCookie` at `state.mjs:133` → `state_security_mismatch`, which stays captured.
- `state_not_found` really is inert: `callback.mjs:72-76` logs `("State not found", error)` where `error` is the OAuth **query param**, so `cause` is undefined and the gate is already false.
- `state_generation_error` really does still page, though not via the route the docblock's ENG-2259 note describes — `oauth2/state.mjs:36-43` logs it as `("Failed to create verification", error)` *before* wrapping it in an `APIError`, so the `StateError` reaches the gate with its `code` intact. (Worth knowing it does **not** survive the router's own path: `api/index.mjs:201-206` swallows any error whose message contains `"table"`, which that message does, logging the string with no `Error`.)
- The `details.state` credential does not reach Sentry either: `sentry.server.config.ts` overrides no `integrations`, and `extraErrorDataIntegration` is not a default in `@sentry/node` 10, so own error properties are not serialised.
- `secondary-storage.ts` propagates from `get`, so a hard Redis outage is not a `StateError` and still pages, as the risk note says.

Behaviour-wise the diff is a two-predicate addition with `cause` hoisted out of the `SENTRY_DSN && IS_PRODUCTION` guard; the error path is unchanged and the only new effect on non-error levels is the extra log field. `pnpm typecheck` (apps/web) clean, `prettier --check` clean, suite 59/59 locally.

**No correctness findings.** The two comments below are stale counts left behind by `d75af25`'s narrowing — low severity, but they misdescribe the size of the suppressed set in a file whose whole purpose is to pin it.

One more of the same kind, outside the diff so not inline-able: the PR body's **Open gaps** section still says "the **two** codes I can justify from code are suppressed", while the table above it correctly says exactly one does.

---

apps/web/modules/auth/lib/better-auth-observability.test.ts
xernobyl: Fixed in `b62aa6ebe` — now "The other four codes stay captured". You're right that it's the same drift class the third commit was written to fix; `d75af25` moved `state_not_found` into this block and left the count behind.

I swept both changed files for the shape rather than just this instance — `grep -nE '\b(two|three|four|five) (suppressed|codes|of the)'` — and the only other hit was the one you and CodeRabbit flagged below. No third.

apps/web/modules/auth/lib/better-auth-observability.test.ts
xernobyl: Fixed in `b62aa6ebe`: now "anything it cannot positively identify as THE one suppressed code (`state_mismatch`)".

Agreed this was the worse of the two, and for the reason you give — someone reading "two" directly above the fail-closed tests goes looking for the second code, and the two ways that ends are adding `state_not_found` back (silently undoing the narrowing) or assuming the set is incomplete. Named the code inline rather than just correcting the number, so there's nothing left to go looking for.

apps/web/modules/auth/lib/better-auth-observability.test.ts
```diff
@@ -380,6 +383,213 @@ describe("betterAuthLogger (Sentry capture gating, ENG-2037)", () => {
...
+  // still captured. A wrong answer should add noise, never silence a fault.
```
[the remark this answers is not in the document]
xernobyl: Fixed in `b62aa6ebe` — the comment now reads "THE one suppressed code (`state_mismatch`)" rather than "one of the two". Correct catch: the set was narrowed to a single entry in this PR and this comment still described the earlier policy, in the test that pins the fail-closed contract.

Same fix as kiln-agent's note on the adjacent line; I also swept both changed files for other stale counts and found only the one at `:432` ("three" → "four").

xernobyl:
Addressed both reviews. All three inline threads fixed, replied to, and resolved.

## What was wrong — comment/code drift from my own narrowing

Both reviewers caught the same class, in two places, and they were right: the narrowing commit (`d75af25`) left counts describing the previous set.

- "The other **three** codes stay captured" → now **four**, matching the four bullets and four test rows below it.
- "one of the **two** suppressed codes" → now "**THE one** suppressed code (`state_mismatch`)". This was the worse of the pair for the reason kiln-agent gave: it sits directly above the fail-closed tests, so someone looking for the missing second code either adds `state_not_found` back — silently undoing the narrowing — or concludes the set is incomplete. I named the code inline rather than just correcting the number, so there's nothing left to hunt for.

I swept both changed files for the same shape instead of only fixing what was reported. No third instance.

## The split-Redis case — added to the docblock

This is the substantive part of the review and it's a case my analysis missed. My cost note named eviction, flush and failover: all *runtime* events. Replicas that don't share one Redis are worse and different. The verification record is written by the pod that starts the flow and looked up by whichever pod serves the callback, and that lookup is the **first** check in the database branch, before the signed-cookie check — so a split-Redis deployment fails **100% of SSO sign-ins** onto the suppressed code, while the signed state cookie verifies perfectly because `BETTER_AUTH_SECRET` is shared even when the store is not. It keeps it off `state_security_mismatch`, and it lands at deploy time, when someone is watching Sentry rather than tailing logs.

The docblock now names both shapes explicitly rather than implying eviction is the only one.

**Filed the alert as ENG-2551** rather than leaving it as "worth a rule" — agreed with the reasoning that it shouldn't join the "later" bucket the `auth.path` follow-up is already stuck in, especially with ENG-2537 showing that one is gated on a release. Medium, backlog, no due date. It records the log-rule approach as preferred and the 1-in-50 sampled capture as the fallback, with your reasoning for preferring the rule.

## Full re-review after the changes

Since these edits touched the file the whole PR is about, I re-derived rather than assumed:

- **Every upstream citation in the docblock re-checked line by line against the installed 1.7.0**: `state.mjs:90` (the unreachable `state_not_found` throw), `callback.mjs:72-73` (the short-circuit and its `logger.error("State not found", error)`), `api/index.mjs:210` (the context-discarding logger call), `state.mjs:93`/`:108` (cookie branch and `state_invalid` inside it). All exact.
- **Invariants intact**: the set holds exactly `["state_mismatch"]`; the gate is still `cause && !isAPIError(cause) && !isUnactionableStateError(...)`, i.e. fails closed; and there is still no `.details` access anywhere.
- **Mutation re-checked** after the edits: removing the gate fails 4, unmutated 59/59.
- Gates: typecheck 27/27, lint 0 errors, format clean, 8497 tests passing.

Also fixed the two description nits — the blank line that split the coverage table into a headerless one-row table, and the row that read as if three `state_mismatch` message literals were three behaviours when the gate reads `code` (they're documentation of what upstream throws; the behavioural assertion is that each still logs at `error`).

Still a draft, pending @Tiago opening it for review.

xernobyl:
@kiln-agent could you re-run over the head commit when you get a chance? Your review landed on `655d37a8a` and there is one commit on top of it, `b62aa6ebe`, which is entirely your two notes:

- **The stale counts** you and CodeRabbit both flagged — "the other three codes" → four, and "one of the two suppressed codes" → "THE one suppressed code (`state_mismatch`)". I named the code inline rather than only fixing the number, so there is nothing left to go looking for, and I swept both changed files for the same shape rather than just the two reported lines (no third instance). All three threads replied to and resolved; CodeRabbit has confirmed its one.
- **The split-Redis case**, which is now in the docblock. Your framing was better than what I had: my note listed eviction/flush/failover, all runtime, and missed that replicas not sharing one Redis fail 100% of SSO sign-ins onto the suppressed code — the record lookup precedes the signed-cookie check, and the shared `BETTER_AUTH_SECRET` keeps the cookie valid, so it never reaches `state_security_mismatch`. The docblock now names both shapes and says plainly that the second produces no Sentry event at all.

Also filed **ENG-2551** for the log alert rather than leaving it as "worth a rule", per your first note — it prefers the rule on `stateErrorCode` volume and records the 1-in-50 sampled capture as the fallback, with your reason for preferring the rule.

No behavioural change in that commit: the suppression set is still exactly `["state_mismatch"]` and the gate is unchanged. What I re-verified after it, since the edits touched the file the PR is about — every upstream citation re-checked line by line against the installed 1.7.0 (`state.mjs:90`, `callback.mjs:72-73`, `api/index.mjs:210`, `state.mjs:93`/`:108`, all exact), the fail-closed gate intact, no `.details` access, and the mutation check still biting (removing the gate fails 4; 59/59 unmutated). Sonar re-ran on the head: gate passed, 0 new issues, 100% coverage on new code. CI 19/19 green.

mattinannt (approved):

PR #8934, opened 2026-08-20
docs(pr-template): say how to look up the model and effort for the agent note

itsjavi:
No ticket — process fix for the PR template, prompted by getting this exact field wrong on #8932.

## What & why

The template ends with a note the author fills in:

```
**AI model used** — `<model>`, reasoning effort `<level>`.
```

The shape is fine. What was missing is any statement of *what those two values are* and *where to read them*, so both get filled from memory:

* **`<model>` reads as "what are you".** Two comments higher the template tells agents not to advertise their tooling, so an agent resolves "model" to the product it runs in and writes the harness name. On #8932 that produced `Claude Code, reasoning effort high` while the session was serving `claude-opus-5` at effort `max` — @itsjavi corrected it by hand.
* **"reasoning effort" is Claude-shaped.** Codex exposes a different knob, some tools expose none, and an agent with nothing to put there invents a plausible level rather than saying so.

**The rendered note is unchanged.** This PR only fills in the HTML comment above it, which never reaches the description:

* model is the vendor's exact model id (`claude-opus-5`, `gpt-5.1-codex`), not the harness — Claude Code, Codex CLI and Cursor are harnesses; the harness may follow in parentheses when it adds something;
* effort is that vendor's own knob in its own units — an effort level, a thinking budget, or `n/a` where there is no such setting;
* both are read out of the tool: `/status` or the session's `model` / `effort_level` in Claude Code, `/model` or the startup line in Codex CLI, whatever it reports about itself elsewhere;
* a value that cannot be looked up is `unknown`, never a plausible guess and never the harness standing in for the model.

Vendor-neutral on purpose: the answer differs per tool, not just per model, so the guidance has to hold for whatever opens the next PR.

The header comment's cross-reference is also renamed from "The model note at the bottom" to "The agent note at the bottom", so it names the block rather than one of its fields.

## Breaking changes

None — one HTML comment and one word in another. No API, SDK, route, env var, config key or emitted value, and nothing an integrator or self-hoster can observe.

## Migrations & env

- none

## How this was tested

Nothing executable changed. The risks are that something *parses* the note, that repo prose goes stale, or that the template stops rendering as intended — checked rather than assumed.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The rendered template is byte-identical to `main` | manual | `git diff origin/main -- .github/pull_request_template.md` touches only the HTML comment block and the word "model" → "agent" inside another comment. The `> [!NOTE] **AI model used** — …` line is unchanged, so every future PR description renders exactly as it does today. An earlier revision of this PR restructured that visible line into four bullets; that is reverted |
| Nothing parses the note | manual | `grep -rn "AI model\|reasoning effort\|ai_model" .github/ .coderabbit.yaml` returns one hit — the unchanged template line. `pr-label-sync.yml` reads only the Breaking-changes checkbox; `.coderabbit.yaml`'s `pre_merge_checks.description` is `mode: warning` with no requirements, and its three `custom_checks` (breaking-changes, magic-word, coverage) don't reference the note |
| No repo prose describes the old wording | manual | `grep -rn "AI model\|model note\|agent note\|reasoning effort"` over `AGENTS.md`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/` and `.claude/` finds only the unrelated `AI_MODEL` env-var row in `docs/self-hosting/…/environment-variables.mdx`. AGENTS.md refers to the template generically, which stays true; the one in-file cross-reference is updated here |
| The guidance actually produces the right answer | manual | Followed it on this PR: `get_session` returns `session_context.model` `claude-opus-5` and `effort_level` `max`, which is what the note below says. That is the same lookup the comment now points at, and the value I failed to check on #8932 |
| Prettier cannot reflow this file, so the comment's wrapping is safe | manual | `.prettierignore:24` exempts `.github/pull_request_template.md` (for the intentional empty list items). Stated because it cuts both ways: it means Prettier can't strip a blockquote marker or re-wrap the comment — and it means **`pnpm format:check` never inspects this file**, so citing that check as evidence here would be vacuous. `npx prettier --check` on the path prints "All matched files use Prettier code style!" precisely because zero files matched. Verified by reading `.prettierignore`, not by running the check — thanks to @kiln-agent for catching that an earlier revision of this description made exactly that vacuous claim |

**Open gaps**

* Honour-system either way — nothing can verify that a filled-in model id is the one that actually served the turn. This only removes the excuses for getting it wrong.
* Lookup instructions are concrete for Claude Code and Codex CLI, generic for everything else. Whoever first fills this in from Cursor or Copilot Workspace should add that tool's incantation.
* The comment's `claude-opus-5 (Claude Code 2.1.237)` example pins a patch version that will age. Kept because the version *is* the thing being illustrated, and it sits inside the same comment that says to read the value from the tool — but it is an example an inattentive agent could copy.

**Risks**

* The comment is longer than the one it replaces, but it is a comment: it costs template-reading agents a few lines and costs the rendered description nothing.
* If a workflow ever wants to parse authorship, it should key on the visible `**AI model used**` label, which this PR deliberately leaves alone.

---

> [!NOTE]
> **AI model used** — `claude-opus-5 (Claude Code 2.1.237)`, reasoning effort `max`.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@ opens on demand (background, file audits, setup steps). Never fold the Coverage 
 Breaking changes — long media inside Coverage may still be folded. -->
 
 <!-- AI agents: no promotional footers, don't advertise yourself or any tool/service (e.g. "Generated
-by …"). The model note at the bottom is the one exception. -->
+by …"). The agent note at the bottom is the one exception. -->
 
 <!-- Complete the line below: `Fixes ENG-<id>`. A bare URL does NOT link the PR — the magic word has
 to come first. Use `Ref ENG-<id>` instead if this PR only partly addresses the ticket, so merging
@@ -90,7 +90,20 @@ to challenge this list, so an empty one is a claim rather than a formality. -->
 
 ---
 
-<!-- Fill in if an AI agent wrote code or this description; delete the note if none did. -->
+<!-- Fill in if an AI agent wrote code or this description; delete the note if none did.
+
+`<model>` is the exact model id the vendor serves — `claude-opus-5`, `gpt-5.1-codex` — not the
+product it runs in: Claude Code, Codex CLI and Cursor are harnesses, not models. Name the harness
+and its version in parentheses when it adds something: `claude-opus-5 (Claude Code 2.1.237)`.
+
+`<level>` is whatever reasoning knob that vendor exposes, in that vendor's own units: an effort
+level (`max`, `high`), a thinking budget (`32k tokens`), or `n/a` where there is no such setting.
+
+Read both values out of the tool rather than from memory. Claude Code reports them in `/status`,
+or as the session's `model` and `effort_level`; Codex CLI in `/model` or the line it prints at
+startup; other tools, whatever they report about themselves. A value you cannot look up is
+`unknown` — never a plausible-looking guess, and never the harness name standing in for the
+model. -->
 
 > [!NOTE]
 > **AI model used** — `<model>`, reasoning effort `<level>`.


kiln-agent (commented):
Reviewed the full diff (`.github/pull_request_template.md`, +22/-3). No correctness problems — the change is mechanically safe and the reasoning in the description holds up. Two low-severity notes below.

**What I verified rather than assumed**

- Nothing parses the old wording. `pr-label-sync.yml` strips HTML comments and matches only `/^this pr contains breaking changes/i` checkboxes; `.coderabbit.yaml`'s three `custom_checks` and its `pre_merge_checks.description` (`mode: warning`, no requirements) never reference the note. `pr-size-check.yml` is the only other workflow touching the body, and it writes only.
- The in-file cross-reference rename at line 8 is the complete set — grepping `.github/`, `.coderabbit.yaml`, `AGENTS.md`, `CLAUDE.md`, `.claude/` and `.agents/` for `AI model` / `reasoning effort` / `model note` / `agent note` finds hits only inside the template itself.
- Markdown mechanics of the new block are sound: the multi-paragraph HTML comment is a CommonMark type-2 HTML block, so the blank lines inside it don't terminate it (the Breaking-changes comment already relies on this), and the `Scope` code span wrapping across lines 114–115 with a two-space continuation inside the blockquote renders as a single list item.

**Notes**

1. **`.prettierignore` covers this file**, so the `pnpm format:check` line under Coverage is true but for a different reason than stated — the file is out of scope entirely, not merely clean. Worth correcting so the next person doesn't read it as evidence the formatter validated the block.
2. **Line 111 — the `Agent` placeholder ships a concrete plausible version** (`Claude Code 2.1.237`). That is exactly the copy-from-memory failure mode this PR exists to remove: a copied version number is near-right and a reviewer cannot falsify it, unlike the `Model` examples where the cross-vendor pairing makes a wrong id obvious. `Claude Code <version>` would force the lookup the comment above asks for.
3. **Minor, and partly outside your control:** the hunk at line 8 forbids promotional footers ("Generated by …") while this description — offered as the worked example of the new note — ends with one. Since the note is now the documented single exception, it's worth deciding whether the harness-appended footer is a second exception or should be stripped when the note is present.

Both notes are non-blocking; the substance of the change is an improvement.

---

.github/pull_request_template.md
```diff
@@ -90,7 +90,26 @@ to challenge this list, so an empty one is a claim rather than a formality. -->
...
 
 > [!NOTE]
-> **AI model used** — `<model>`, reasoning effort `<level>`.
+> **Written with an AI agent**
+>
+> - **Agent** — `<product and version, e.g. Claude Code 2.1.237, Codex CLI 0.55, Cursor>`
+> - **Model** — `<exact model id, e.g. claude-opus-5, gpt-5.1-codex>`
+> - **Reasoning** — `<as the vendor names it, e.g. effort max, 32k thinking tokens, n/a>`
+> - **Scope** — `<what the agent produced — code, tests, this description — and what a human
+>   changed afterwards>`
```
kiln-agent: The `Scope` placeholder is the only bullet that wraps, and the wrap puts an unterminated code span on a continuation line that Prettier will strip the blockquote marker from. Reproduced on a copy of the file (Prettier markdown parser, `printWidth: 110`):

```
> - **Scope** — `<what the agent produced — code, tests, this description — and what a human
changed afterwards>`
```

The `> ` is gone from line 2, so the closing backtick and the rest of the placeholder fall outside the blockquote — the `[!NOTE]` alert ends early and the text renders as a stray paragraph with an unterminated code span.

This is currently masked: `.prettierignore:24` exempts this file, so `pnpm format:check` never inspected it (which also makes the Formatting row in the description vacuous — Prettier was a no-op because the file is ignored, not because the file is clean). The exemption's stated reason is unrelated — "intentional empty list items ... Prettier folds them into the label above" — so whoever removes it after the placeholders change shape will silently mangle this line.

Cheapest fix is to keep it on one line; it's a template placeholder, so the long line costs nothing:

```suggestion
> - **Scope** — `<what the agent produced — code, tests, this description — and what a human changed afterwards>`
```

---

.github/pull_request_template.md
```diff
@@ -90,7 +90,26 @@ to challenge this list, so an empty one is a claim rather than a formality. -->
...
+else: whatever it reports about itself. A field you cannot look up is `unknown` — never a
+plausible-looking value, and never the harness name standing in for the model. -->
 
 > [!NOTE]
-> **AI model used** — `<model>`, reasoning effort `<level>`.
+> **Written with an AI agent**
+>
+> - **Agent** — `<product and version, e.g. Claude Code 2.1.237, Codex CLI 0.55, Cursor>`
+> - **Model** — `<exact model id, e.g. claude-opus-5, gpt-5.1-codex>`
+> - **Reasoning** — `<as the vendor names it, e.g. effort max, 32k thinking tokens, n/a>`
```
kiln-agent: These three visible placeholders carry concrete, copyable values in exactly the fields whose instructions say "never a plausible-looking value":

- `Claude Code 2.1.237` — a real-looking patch version that is stale the week after it lands
- `claude-opus-5`, `gpt-5.1-codex` — valid model ids
- `effort max` — a valid level

An agent that skips `/status` now has a well-formed answer sitting in the template. The failure mode this PR is fixing (#8932's `Claude Code, reasoning effort high`) got caught precisely *because* it looked wrong; `claude-opus-5` at `effort max` copied verbatim into a Codex-authored PR reads as authoritative and nobody re-checks it. Precise-and-wrong is harder to spot than vague-and-wrong.

The HTML comment above already carries every one of these examples, where it can't be lifted into the rendered description. Suggest the visible bullets stay shape-only:

```suggestion
> - **Agent** — `<product and version>`
> - **Model** — `<exact model id>`
> - **Reasoning** — `<as the vendor names it>`
```

Not blocking — your call whether the discoverability of inline examples is worth the copy risk.

---

kiln-agent (commented):
Reviewed the single-file template change (+22/-3). The substance is sound — splitting agent/model/reasoning/scope into separate facts, naming the lookup for each, and telling the agent to write `unknown` all address the real failure on #8932. The header cross-reference rename is consistent.

I re-ran the description's verification rather than taking it:

- **No consumer parses the note** — confirmed. Repo-wide grep for `AI model` / `reasoning effort` / `model note` / `agent note` across `.github/`, `.coderabbit.yaml`, `AGENTS.md` and `CLAUDE.md` returns nothing outside the template itself. `pr-label-sync.yml` reads only the breaking-changes checkbox; the three `.coderabbit.yaml` `custom_checks` don't reference it. Safe to reshape.
- **Rendering** — the `> [!NOTE]` alert with a nested list renders as intended *as committed*.
- **Formatting** — this is the one claim that doesn't hold. `.prettierignore:24` exempts this file, so `pnpm format:check` passing says nothing about it. See the inline comment on the `Scope` bullet: Prettier does reflow that line, and the reflow breaks the alert.

Two inline comments — one real latent break (the wrapped `Scope` placeholder), one judgement call (copyable example values in the visible placeholders). Neither is a blocker for a docs-only change; the `Scope` one is worth fixing now since it's a one-line edit and the `.prettierignore` entry that hides it is documented as existing for an unrelated reason.

Breaking-changes checkbox is correctly unticked — a comment and a template block change nothing consumable.

---

.github/pull_request_template.md
itsjavi: Both halves land, and the file no longer has the shape you reviewed — the four-bullet block is reverted, so the visible note is back to the single-line `**AI model used** — <model>, reasoning effort <level>` and nothing wraps.

The `.prettierignore:24` half is the real catch and it outlives the revert: my Formatting row cited `pnpm format:check` as evidence about a file that check never inspects, which is vacuous exactly as you say — `prettier --check` on that path reports success because zero files matched. The description now states the exemption, notes that it also means Prettier cannot reflow the comment, and says it was verified by reading `.prettierignore` rather than by running the check. Resolving as stale.

---

.github/pull_request_template.md
itsjavi: Stale in the specific — those three visible placeholders are gone, the rendered note is back to `<model>` / `<level>` — but you were right about the general point, and it is why the revert is the better shape: concrete values belong in the HTML comment, where an agent reads them as instruction and cannot lift them into a description.

Kept your `2.1.237` objection in Open gaps rather than silently dropping the version: the version is the thing that example illustrates, and it sits inside the comment that says to read the value from the tool, so I'd rather keep the illustration and flag the copy risk. Resolving.

---

.github/pull_request_template.md
```diff
@@ -90,7 +90,26 @@ to challenge this list, so an empty one is a claim rather than a formality. -->
...
+Read the values out of the tool instead of guessing. Claude Code: `/status`, or the session's
+`model` and `effort_level`. Codex CLI: `/model`, or the model line printed at startup. Anything
+else: whatever it reports about itself. A field you cannot look up is `unknown` — never a
+plausible-looking value, and never the harness name standing in for the model. -->
 
 > [!NOTE]
-> **AI model used** — `<model>`, reasoning effort `<level>`.
+> **Written with an AI agent**
+>
+> - **Agent** — `<product and version, e.g. Claude Code 2.1.237, Codex CLI 0.55, Cursor>`
```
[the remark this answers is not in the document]
itsjavi: Moot now — the `Agent` placeholder no longer exists. The visible note is back to the original `**AI model used** — <model>, reasoning effort <level>`, and the harness example (with its version) lives in the HTML comment above it. Resolving as stale.

---

xernobyl (approved):
LGTM

Issue #123, opened 2022-10-26
token-reward after finishing form


jobenjada:
### Is your proposal related to a problem?

Get more people in web3 communities to fill out a survey by incentivizing

(Write your answer here.)

### Describe the solution you'd like

so in a Web3 context
i would do it like this
1. user connects wallet
2. user finishes the form
3. at thank you page, user clicks on some "gift" 
4. gift is opened and user gets a random POAP NFT / amount of Token

### Describe alternatives you've considered

(Write your answer here.)

### Additional context

(Write your answer here.)

PR #8935, opened 2026-08-20
feat(embedded-data): ingest rules — type coercion, size limits, collisions, precedence [ENG-1845]

itsjavi:
## What & why

Embedded Data has two ingestion paths coming (URL params, `setEmbeddedData`) and nothing that says what either is allowed to write. This adds that one contract and enforces it everywhere `response.data` is written, so neither path has to restate a rule.

`packages/types/embedded-data-ingest.ts` owns the rules: only a survey's `ingested` fields are settable, `locked` ones ignore outside writes, values are coerced to their declared `dataType` rather than rejected, oversize values are truncated, and a question answer is never rewritten. Nothing is fatal — a bad key or a wrong-typed value can never be the reason a response fails to save.

The invariant that keeps it honest: every value the normalizer accepts without a flag is checked against `coerceToEmbeddedDataType` (the read seam) before it is returned. That seam is narrower than a permissive reading of the rules in three places — booleans are only `"true"`/`"false"`, dates are `Z`-only, and a non-numeric number resolves to its default — so validating through it makes write/read drift unrepresentable instead of review-dependent. A value that would ingest "successfully" and then read back as unset comes back flagged.

**Server — three boundaries, all public, none trusting the client.** The v1 and v2 client POSTs and the client PUT each run the contract before validation and quota evaluation (so both see what will be stored) and before the verified-email gate (which stamps `verifiedEmail` from the token and has to stay the last writer).

The PUT is the boundary that matters most and the easiest to miss: `validateResponseData` only validates keys matching element ids, so everything else passed straight through — a crafted PUT could write a locked key, an unknown key or a 2 MB value into `response.data`, which exports, filters and the summary read directly.

**Flags are persisted now, because there is nothing to surface later otherwise.** A wrong-typed value is stored raw and flagged; without the flag it is indistinguishable from an absent one. New nullable `Response.ingestFlags` column, additive, no backfill — null means no ingest boundary has written it, which is a different claim from an empty list. Server-computed and threaded as a separate parameter rather than a key on the parsed input, because a client-sent flag list could claim "no flags" — the same trust problem as the filtering. A PUT unions by key: keys it rewrote take their new verdict, the rest keep theirs. Deliberately *not* a `_flags` key inside `response.data`: `isLegacyIdCharset` permits leading underscores, so a survey could declare that `storageKey`.

**Client — one call site.** The renderer applies the same contract to the raw bag at display time, so the SDKs stay dumb pipes and the four mobile ones inherit the rules without shipping a copy (ENG-2472). It logs what it refused, for developer feedback only; the server recomputes and stores. js-core's own hidden-field filter is deleted rather than ported, which also removes its `hiddenFields.enabled` gate — see **Risks** below.

Coercion decisions worth naming:

| dataType | accepted | stored as | on failure |
| --- | --- | --- | --- |
| `string` | any scalar | `String(raw)`, `""` kept | non-scalar dropped |
| `number` | number, numeric string | JS `number` | raw text + `coercion_failed` |
| `boolean` | `true/false`, `"true"/"false"`, `1/0`, `yes/no`, `on/off` | the **string** `"true"`/`"false"` — `ZResponseDataValue` has no boolean member | raw text + `coercion_failed` |
| `date` | ISO date, ISO datetime, explicit offset, `Date` | ISO date or `…Z` datetime | raw text + `coercion_failed` |

- a date the read seam already accepts is returned untouched, so a date-only value is never promoted to midnight UTC and normalization stays idempotent across the renderer and the server;
- a zone-less datetime is read as UTC, never as the local zone of whoever ran the contract, so the same param cannot ingest as two different instants; a colon-less offset (`+0200`) is canonicalized before parsing, for the same reason;
- a blank string is never `0` for a number field;
- oversize values are cut on a code-point boundary against a UTF-8 byte budget, so the `Json` column never receives a lone surrogate — and only the *writing* boundary cuts, because truncation is the one rule whose enforcement destroys the evidence it happened: hand the server a value that has already been cut and it fits the budget, so no `truncated` flag is ever recorded. The renderer flags without cutting;
- what the contract *reports* is keyed by the declared `storageKey`, the same as the flags, so the two lists join by key — the one exception is `unknown_key`, where nothing declares it and the spelling that arrived is the whole finding.

Two rules from the ticket landed differently, on purpose: ingest never writes `defaultValue` (`resolveEmbeddedValue` already applies it on read, and two writers would mean a later default change applies on one path and not the other), and the `name ≤ 255` cap has no boundary to enforce it at — `name` is not in the response payload and the library `key` is already capped.

Out of scope, as the ticket says: the two ingestion paths themselves (ENG-1843 / ENG-1844), surfacing the flag in the UI, arrays, and rule 6's *read*-side precedence, which belongs to ENG-1840/1853.

### Why this isn't split further (re: the PR size bot)

Nearly all of the counted lines are the contract module and its enforcement points (the bot's count excludes tests). Splitting the module out on its own would ship a file nothing calls, and splitting the enforcement points apart would leave one public boundary hardened and the others open — which is exactly the drift the module exists to prevent. The commits are already the reviewable units: `the contract`, then `the wiring`, then the review fixes; the wiring commit is mostly the four small server signature changes plus the tests the AC asks for.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1845/ingest-rules-type-coercion-size-limits-collisions-precedence

## How this was tested

- `pnpm test` on every touched package ✅ — `@formbricks/types` 356, `@formbricks/js-core` 284, `@formbricks/surveys` 697 (counts as of `b0d8813`; they move when the base branch lands tests of its own)
- `apps/web` unit suite ✅ — 598 files, 7779 tests
- `turbo run typecheck` for `@formbricks/types`, `@formbricks/js-core`, `@formbricks/surveys` ✅; `@formbricks/web` diffed against the base branch — **zero new errors** (the pre-existing set is unchanged)
- `turbo run lint` on the touched packages ✅, `eslint` on every changed `apps/web` file ✅, `prettier --check` ✅, `prisma format` ✅

Tests added:

- `packages/types/embedded-data-ingest.test.ts` — the read-seam round-trip invariant and an idempotency invariant, both table-driven over all four `dataType`s and ~35 inputs; per-type normalization; locked dropped; unknown key dropped; computed key dropped; element-id collisions; truncation on a code-point boundary; exact-beats-inexact case matching independent of arrival order; drops keyed by the declared name rather than the one that arrived; the renderer-then-server round trip that keeps a `truncated` verdict; a `__proto__` key surviving as a stored value.
- The **crafted-body** tests the AC asks for, on all three server boundaries: `app/api/v1/client/[workspaceId]/responses/route.test.ts`, the v2 equivalent, and the PUT handler. These post raw JSON that never went through the renderer, which is the only way to prove the re-validation. Each was checked to *fail* with the wiring removed (5/7 on PUT, 7/9 on POST — the rest assert the flags rather than the data).
- One test per survey-payload path pinning the Embedded Data join (`getSurvey`, the SDK workspace-state payload, the link-survey payload, the response pipeline), since the allow-list fails closed and a select that dropped the join would ingest nothing at all.
- `apps/web/lib/response/ingest.test.ts` — the shared server helper, including the log cap, the drift alarm, and the `hiddenFields.enabled` decision below.
- `packages/js-core/src/lib/survey/tests/widget.test.ts` — the hidden-field bag reaches the renderer verbatim, entered at `triggerSurvey` because that is the hop the deleted filter occupied.
- Flag persistence: `updateResponse`'s union, and `buildPrismaResponseData`'s three states (omitted / `[]` / flags) — `null` stays reserved for "no ingest boundary has written this".

Re-run in full after the automated review round (`900d3a3`), which fixed five real findings: `ingestFlags` collapsing an empty list to `null` (which contradicted this PR's own null-state contract), `isResponseDataValue` accepting `Date`/`Map`/class instances through an `Object.values(...).every(...)` hole, colon-less zone offsets (`+0200`) reaching `Date` in an implementation-defined form, the renderer retaining element-id keys that `ResponseQueue` then merged over real answers, and a question answer of an unstorable shape being dropped without a notice. One finding was left for follow-up with a reason on its thread: pre-existing lost updates from concurrent PUTs. (The other, the `as any` Prisma-mock fixtures, was initially deferred and then done — see `b4db31b` below.)

Re-run again after the human review round (`91f9ffc`), which took both of @pandeymangg's diagnostics findings: `dropped[].key` reporting the incoming spelling for `locked_field` and `unsupported_value` while `flags[].key` and `element_id_collision` reported the declared one (so the two lists could not be joined, and a field declared `Plan` was not greppable when sent as `plan`), and total ingestion failure on a survey whose rows are missing being logged identically to a single typo'd param. Each new test was checked to fail without its fix.

Re-run once more after a second review round (`c617c67`), which fixed two real bugs and two smaller things: the `truncated` flag never reaching storage on the path the feature ships on (the renderer cut the value, so the server's re-run saw one that fit and recorded nothing — `coercion_failed` survives the same trip because the raw text is kept and the verdict re-derived); a `__proto__` storage key or element id evaporating into `Object.prototype`'s setter with no stored value, no drop and no flag, which is the one outcome the module documents as impossible; a console warning fired at every respondent on every email-verified link survey, because the wrapper puts `verifiedEmail` in the bag and `FORBIDDEN_IDS` guarantees no survey can declare it; and nothing pinning js-core's post-deletion pass-through.

`b0d8813` merges `epic/embedded-data-v1` back in after ENG-2538 landed there, and everything above was re-run on the merge. The one conflict was textual — both sides inserted into the same alphabetical slot of the `embedded-data-resolver` import list in `survey.tsx` — and the two changes are independent despite sharing that file: ENG-2538 adds a *read* path (dropping reserved catalog entries shadowed by a declared field or element id, for recall and logic), while this PR owns the *write* path. They consult different resolvers on purpose, since shadowing counts every declared field while the ingest allow-list is `ingested` fields only. Worth stating explicitly for review: the incoming commits do not touch `coerceToEmbeddedDataType`, so the read seam this PR validates every accepted value against did not move.

`b4db31b` closes the last open review thread by typing the Prisma fixture in `apps/web/lib/response/service.test.ts` and removing all 33 `as any` casts. Typing it surfaced something the thread had not: `vi.mocked(prisma.response.findUnique)` types `mockResolvedValue` against the method's default instantiation, so the `select` at the call site is invisible to it and it demands the whole `Response` model — while the fixture supplied only the fourteen `responseSelection` names, missing `contactId` outright and `ingestFlags` once this PR added it. The casts were hiding a real fixture/schema gap rather than style. Since `apps/web/tsconfig.typecheck.json` excludes `**/*.test.ts`, this was verified with `tsc --noEmit` against a config that includes the file: 33 errors before, 0 after.

Not tested manually against a running stack: nothing here has a UI, and the two paths that a tester could click through are ENG-1843/1844, which land next. The QA plan below is API-level for that reason.

## Breaking changes

None

## QA / Test Plan

**How to test**

- [ ] Add an ingested Embedded Data field `plan` (type *string*) to a link survey, publish it, open `?plan=gold`, submit → the response shows `plan = gold`.
- [ ] Add `seats` (type *number*), open `?seats=12`, submit → the response stores `12` as a number, not `"12"`.
- [ ] Same field, open `?seats=many`, submit → **the response still saves** and shows `seats = many`. This is the differentiator: a wrong-typed value is kept, not discarded.
- [ ] Add `trial` (type *boolean*), open `?trial=yes` → stored as `true`. Repeat with `on`, `1`, `no`, `off`, `0`.
- [ ] Add `signedUpAt` (type *date*), open `?signedUpAt=2026-08-06` → stored as `2026-08-06` (not converted to a midnight timestamp). Then `?signedUpAt=2026-08-06T10:30:00%2B02:00` → stored as `2026-08-06T08:30:00.000Z`, and `?signedUpAt=2026-08-06T10:30:00%2B0200` (no colon in the offset) → the same value.
- [ ] Open the survey with a param the survey does not declare (`?rogue=x`) → it never appears on the response, and the survey works normally.
- [ ] `POST /api/v1/client/{workspaceId}/responses` with `data` containing a key the survey does not declare → HTTP 200, and the stored response does not contain that key.
- [ ] Same POST against `/api/v2/client/...` → same result.
- [ ] `PUT /api/v1/client/{workspaceId}/responses/{responseId}` on an unfinished response with an undeclared key in `data` → HTTP 200, key not stored. Before this change it was stored.
- [ ] POST with a declared string field carrying ~20 KB of text → HTTP 200, the stored value is cut to 16 KB.
- [ ] Same ~20 KB value, but sent through the **JS SDK** rather than a direct POST (`track("evt", { hiddenFields: { note: "…" } })`) → the stored value is cut to 16 KB *and* the response carries a `truncated` flag. The flag is the point: it must not appear only for a direct POST.
- [ ] A link survey with email verification on: complete the verification and load it → the respondent's console shows **no** Formbricks warning about `verifiedEmail`.
- [ ] A survey with email verification on: complete it → the response still shows the verified address (the gate must remain the last writer of `response.data`).
- [ ] An app survey via the JS SDK: `formbricks.track("evt", { hiddenFields: { plan: "gold", rogue: "x" } })` → `plan` lands, `rogue` does not, and the browser console explains why.
- [ ] Same survey, but pass a hidden field named after one of its question ids (`{ hiddenFields: { <questionId>: "injected" } }`), then answer that question → the stored response holds the respondent's own answer, not `injected`, and the console says the key belongs to a question. (Before this change the hidden field won, on every submit.)
- [ ] Declare a field as `Plan` (capital P) and lock it, then open `?plan=gold` → the browser console names `"Plan"`, the field as declared, rather than echoing the `plan` that was sent.

**Preconditions / test data**

- A workspace with a published link survey and a published app survey, each with at least one ingested Embedded Data field. Locked fields need to be set through the API — the editor has no lock toggle yet.
- An API client (curl/Postman) for the POST/PUT steps; the client endpoints need no auth beyond the workspace id.
- Cloud and self-host behave identically; no feature flag.
- The 16 KB limit and the accepted boolean/date spellings are **inferred from the implementation** — worth a human confirming they are the product intent.

**Risks & regressions**

- **A survey whose rows are missing ingests nothing.** The allow-list is the joined Embedded Data rows and nothing else (ENG-2412 removed the legacy fallback), so it fails closed. Every survey-payload path carries the join today and each now has a test pinning it, but a survey with populated legacy `hiddenFields.fieldIds` and no rows would silently collect nothing — the same state in which its recall, logic and exports are already broken. Worth spot-checking an older survey on staging. That state now warns on its own log line rather than blending into the generic drop line, with `embeddedFieldCount` in the payload to separate a missing join from rows that loaded with none of them `ingested`.
- **`hiddenFields.enabled` is no longer an ingest gate.** Removing js-core's filter removes the only reader that honoured it. For an app survey in the (near-unreachable) state of `enabled: false` with populated `fieldIds`, values passed via legacy `track({ hiddenFields })` now land where they were previously dropped. Purely additive — no stored data changes and nothing that worked stops working — but it is a silent change for anyone who read the flag as a kill switch. `locked` is the per-field control now, and acceptance into such a survey is logged so a customer report is greppable.
- **Values that used to reach `response.data` unvalidated no longer do.** A client-sent `verifiedEmail` on a survey with verification *off*, and any undeclared key on a PUT, were previously written as-is. Both are now dropped. A hidden field named after a question id no longer overwrites that question's answer on submit.
- **Coercion changes stored value types.** A number field that used to store `"12"` now stores `12`. Filters, exports and quotas read through the same seam and were exercised by the suite, but response filters on a number field are worth a look.
- **Not fixed here, and pre-existing:** `updateResponse` merges `data`, `ttc`, `variables` and now `ingestFlags` from a pre-update snapshot, so concurrent PUTs can lose each other's writes. Raised on the review; it needs its own ticket because the fix changes the concurrency semantics of every response update.
- Re-check: the response table's verified-email column, exports of a survey with hidden fields, and quota rules whose operand is an ingested field.

- One additive migration: `20260820000000_add_response_ingest_flags` adds a nullable `Response.ingest_flags` JSONB column. No default, no backfill, no cutover step; safe to deploy before or after the app.


apps/web/lib/response/service.test.ts
```diff
@@ -94,6 +94,75 @@ describe("updateResponse", () => {
...
+        mockResponseId,
+        createMockResponseInput({ data: { plan: "gold" } }),
+        undefined,
+        []
+      );
+
+      expect(updateArgs().data).toMatchObject({
+        ingestFlags: [{ key: "seats", reason: "coercion_failed" }],
+      });
+    });
```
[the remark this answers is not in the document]
itsjavi: Not doing this one here, and it's a fair catch that I added to it.

I checked the counts: this file had 25 `as any` casts on the base branch and my new tests added 8 more, following the surrounding style. The problem with removing only mine is that they all go through the shared `createMockCurrentResponse` helper, which returns a partial literal missing most of the model's columns — so passing it to `mockResolvedValue` without a cast needs the helper itself typed as the selected Prisma payload, and that changes it for all 33 call sites. That's a file-wide test refactor, which is wider than this ticket and mostly touches tests this PR doesn't otherwise go near.

Worth doing as its own change — typing the fixture once would delete all 33 casts rather than 8 of them. Noting that `apps/web/tsconfig.typecheck.json` excludes `**/*.test.ts`, so there's no typecheck signal being suppressed today either way.

---

itsjavi:
Worked through the review — `900d3a3`. Five of the seven findings were real; I verified each against the code before touching it.

**Fixed**

- **`ingestFlags` collapsed `[]` to `null`.** The sharpest one, because it contradicted this PR's own documented contract: `null` is supposed to mean "no ingest boundary has written this", and on a create an omitted nullable field is already `NULL` — so a response the contract checked and found clean was indistinguishable from a legacy one. Both the create and update paths now write whatever list they were given, empty included.
- **`isResponseDataValue` accepted any class instance.** `Object.values(new Date())` is `[]` and `[].every(…)` is `true`, so a `Date`/`Map`/`Set` passed the exact guard whose docstring claims it keeps non-JSON shapes out of a `Json` column. Prototype-checked now, with a test that would have caught it.
- **Colon-less zone offsets.** `±HHMM` reaches the parser from query strings and V8 accepts it, but it isn't in ECMAScript's Date Time String Format — so the same param would resolve to an instant on the server and flag as uncoercible in an engine that refuses it. That's precisely the cross-runtime disagreement the date rules exist to prevent, so it's canonicalized before parsing.
- **Element-id keys kept from the host's bag in the renderer.** Correct and worth spelling out: the contract passes question answers through because at a server boundary `incoming` *is* `response.data`, but on the client it's only the host's bag — and `ResponseQueue` merges that record **over** `data` on every submit, so a hidden field named after a question id re-applied its display-time value on top of whatever the respondent answered. Pre-existing (the raw prop did the same), now closed, and the filtered keys are logged rather than vanishing silently.
- **A question answer of an unstorable shape was silently omitted** with no `dropped` entry. Reported now, like everything else.

**Not doing here**

- **`as any` in the Prisma mocks** — reasoning on [that thread](https://github.com/formbricks/formbricks/pull/8935#discussion_r3825503292). Short version: 25 pre-existing in the same file, 8 added by me, and removing only mine requires retyping the shared fixture all 33 call sites use.
- **Lost updates from concurrent PUTs** (`apps/web/lib/response/service.ts`) — real, and I'm leaving it. `updateResponse` has merged `data`, `ttc` and `variables` from a pre-update snapshot since long before this PR; all this change did was add `ingestFlags` to that existing pattern, so the exposure is unchanged rather than introduced. The fix — row locking, optimistic concurrency with retry, or a database-side merge — changes the concurrency semantics of *every* response update and needs its own ticket and its own QA. Flagging it rather than smuggling it in behind an ingest change; happy to file it if you'd like.

Verification on the new head: `apps/web` unit suite 596 files / 7758 tests green (real vitest exit code, not a piped one), `@formbricks/types` 342, `@formbricks/js-core` 283, `@formbricks/surveys` 696, typechecks clean and diffed against the base branch for zero new errors, lint and prettier clean.

---

apps/web/lib/response/ingest.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  // The allow-list is the stored rows, and a row carries no `enabled` concept — so `locked` is the
+  // per-field control for "stop accepting writes" and the legacy flag is not an ingest gate
+  // (ENG-1845, decision 5). Three of the four readers already ignored it, and the state below is
+  // near-unreachable: the editor writes `enabled: true` whenever a field is added or removed, and a
+  // freshly created survey has the flag off with no field ids and therefore no rows. Only an
+  // API-authored or pre-editor survey can hold both, and for those the flag was never a working kill
+  // switch, because the link-survey URL path has always ignored it. Logged so that if a customer
+  // ever reports it, this is greppable.
+  if (survey.hiddenFields?.enabled === false && Object.keys(result.data).length > 0) {
```
pandeymangg, line 99: There's a louder version of this worth adding right here: `survey.hiddenFields?.fieldIds?.length > 0 && ingestedFields.length === 0`. That means the survey declares fields but has no rows — either a select dropped the join, or the column and the rows have drifted. Right now that case logs the same "dropped or flagged incoming keys" line as a single typo, so a regression that kills ingestion for every response looks like routine noise.

packages/types/embedded-data-ingest.ts
```diff
@@ -0,0 +1,444 @@
...
+    // that differ only by case, which `@@unique([surveyId, storageKey])` allows and the shipped URL
+    // reader already does. Case tolerance is not optional: surveys relying on that drift are
+    // collecting data today.
+    const matchedKey = matchDeclaredFieldName(ingestableKeys, storageKey);
+    // Rule 8: nothing arrived, so the key stays omitted rather than being written as a default.
+    if (matchedKey === undefined) continue;
+    consumedKeys.add(matchedKey);
+
+    if (field.locked) {
+      dropped.push({ key: matchedKey, reason: "locked_field" });
```
pandeymangg: `dropped[].key` means two different things depending on the reason: the incoming key here and for `unknown_key`, but the declared `storageKey` for `element_id_collision` above. With case-insensitive matching, a locked field declared as `Plan` and sent as `plan` reports `plan`, so grepping for the field name finds nothing. Worth picking one and noting which it is, since the v2 UI will group these by key.

pandeymangg (commented):
Read the contract, the three server boundaries, the migration and the js-core deletion. Two small things below, both about diagnostics rather than behaviour — the logic itself checks out, and validating the accepted value through `coerceToEmbeddedDataType` is a genuinely good call.

apps/web/lib/response/ingest.ts
itsjavi: Added in `91f9ffc`, as its own `logger.warn` right after the generic line.

You're right that this is the case that shouldn't blend in — it's the top risk in the PR description and today it produces the *same* line as one typo, only repeated once per response, since in that state every key falls through to `unknown_key`.

Two things I did on top of your condition:

- **Gated it on a value actually having been dropped** (`result.dropped.some(({ reason }) => reason === "unknown_key")`). The state can hold on a survey nobody sends params to, and there's no data loss there — so this way every occurrence of the line means ingestion actually cost us something on that response.
- **Added `embeddedFieldCount` to the payload**, because `ingestedFields.length === 0` has two causes and the line couldn't tell them apart: `0` means the join is missing (a select dropped `embeddedDataLinks`), non-zero means rows loaded but none of them are `ingested`. Different fix, same symptom.

Tests in `apps/web/lib/response/ingest.test.ts` cover the alarm firing, and staying quiet both when nothing arrived and when rows resolved normally — checked that the first one fails without the change.

---

packages/types/embedded-data-ingest.ts
itsjavi: Fixed in `91f9ffc` — picked the declared `storageKey`, and documented it on `TIngestDrop.key`.

The deciding argument was one you half-made: `flags[].key` has always been declared-side (its doc says "keyed by `storageKey`, never by the incoming name"), so as it stood the two lists couldn't be joined by key at all. A v2 UI grouping "what happened to this field" would have shown `Plan` on one side and the `plan` that arrived on the other. Declared side also keeps the identity stable, since the incoming spelling varies per request under case-tolerant matching while the declared one doesn't.

So `locked_field` and the declared-field `unsupported_value` now report `storageKey`, joining `element_id_collision` which already did. `unknown_key` deliberately keeps the incoming key — nothing declares it, so the spelling that arrived *is* the whole finding. That asymmetry is the part worth having written down, so it's on the field's doc rather than only in a commit.

Your exact example is now a test (`Plan` locked, sent as `plan`, reports `Plan`), plus one asserting a flag and a drop key the same field the same way. Both fail without the change.

Nice side effect: the renderer's console warning now reads `"Plan" is locked and ignores values set from outside` — naming the field the developer configured rather than echoing their own param back at them.

---

kiln-agent (commented):
Read the contract module, the three server boundaries, the flag persistence, the renderer call site and the js-core deletion; ran the `@formbricks/types` suite plus a few probes of my own against `applyIngestContract`. The design holds up — validating every accepted value through `coerceToEmbeddedDataType` is the right seam, and the three boundary tests that POST raw JSON are what make the re-validation claim checkable rather than assertable. One real gap and three smaller things.

## 1. The `truncated` flag never survives the renderer path

`packages/types/embedded-data-ingest.ts:425-429` truncates and flags in the same step, and `packages/surveys/src/components/general/survey.tsx:358` runs the whole contract client-side and keeps its **output**. So the server re-runs the contract over a value that has already been cut, `truncateToMaxUtf8Bytes` returns it unchanged, and no flag is produced. Verified against the module directly:

```
client: { note: "a".repeat(MAX+500) }        → data.note = 16384 bytes, flags = [{ note, truncated }]
server: { note: <that 16384-byte value> }    → data.note = 16384 bytes, flags = []   ← what gets persisted
```

`coercion_failed` survives the same round trip, because the client stores the raw text and the server re-derives the verdict from it. Truncation is the one rule whose enforcement is lossy, which is exactly why re-running it cannot recover the event — and the module's idempotency guarantee is what makes that certain rather than incidental.

The consequence is that the flag only ever lands for a crafted direct POST, and never for the path the feature ships on: `track({ hiddenFields })` / `setEmbeddedData` (ENG-1844) has no URL-length ceiling keeping a >16 KB value out. A truncated value then reads back as a perfectly valid 16 KB string — which is the failure mode the PR argues the column exists to prevent. The write-up's own justification ("without the flag it is indistinguishable from an absent one") applies here in a sharper form: indistinguishable from a *complete* one.

Cheapest fix that keeps the trust model intact: don't truncate in the renderer. Let it warn and keep the full value in its own state, and let the server — which sees the real length — truncate and flag. The stored value is byte-identical either way, and the only client-side divergence is a >16 KB recall/logic value, which is already meaningless at that size.

## 2. Every respondent on an email-verified link survey gets a console warning

`apps/web/modules/survey/link/components/survey-client-wrapper.tsx:240-242` spreads `getVerifiedEmail` into `hiddenFieldsRecord`, and `verifiedEmail` is in `FORBIDDEN_IDS`, so the contract drops it as `unknown_key` and `logIngestResult` emits:

> Formbricks: "verifiedEmail" is not an ingested Embedded Data field on this survey, so the value was ignored.

on every load of every such survey, in every respondent's console. The `INGEST_DROP_MESSAGES` docstring anticipates this key arriving and justifies the *wording* of the message, but nothing suppresses the line — and it is first-party noise no integrator can act on, which is the fastest way to teach developers to ignore the channel. Either give the verified address its own prop (it is not Embedded Data and the server writes it from the token anyway) or exclude it from the log.

## 3. A `__proto__` storage key or element id loses its value silently

`ZSurveyEmbeddedData.storageKey` (`packages/types/embedded-data.ts:221-224`) only requires `isLegacyIdCharset` — deliberately, per its own docstring, so migrated legacy names keep loading — and that charset admits `__proto__`. Element ids are governed by the same rule. `data` is a plain `{}` at `embedded-data-ingest.ts:355`, so `data["__proto__"] = "gold"` reaches `Object.prototype`'s setter, is a no-op for a primitive, and the key never appears:

```
incoming { __proto__: "gold", plan: "gold" }, both declared
→ data = { plan: "gold" }, dropped = [], flags = []
```

No stored value, no drop, no flag — the one outcome the module states is impossible ("Every key either becomes a stored value, is reported in `dropped`, or — when it is a question answer — passes through untouched"). It also differs from the old behaviour, where a `__proto__` key parsed from the body was an own property and persisted. Obscure, and it takes a legacy survey to reach, but `Object.create(null)` for `data` closes both the ingest and the element-id passthrough loop in one line.

## 4. Nothing pins js-core's new dumb-pipe behaviour

`handleHiddenFields` and its 76 lines of tests are gone, and `triggerSurvey` now forwards `properties?.hiddenFields` verbatim (`packages/js-core/src/lib/survey/widget.ts:76`), but nothing asserts that. `widget.test.ts` already asserts `renderSurvey`'s arguments in several places, so a single `hiddenFieldsRecord` expectation would pin the ENG-2472 decision against someone re-adding a filter in the SDK — which is the drift the deletion exists to prevent.

---

Checked and satisfied, for the record: `elementIds` comes from `blocks` alone while `validateResponseData` still falls back to `questions`, but `20251118032116_migrate_questions_to_blocks` moves `questions` into `blocks` and empties it, and both v1 management write paths derive `blocks` on write, so the empty-`blocks` state that would drop every answer is not reachable. The four join-pinning tests assert against `selectSurveyEmbeddedDataLinks` rather than a literal, so they fail if the constant is dropped from a select. The byte budget is honoured on multi-byte input (8192 `é` → exactly 16384 bytes, cut on a code-point boundary). `normalizeIngestedValue` compares `!== undefined` rather than testing truthiness, so `0` and `"false"` are accepted rather than flagged. And `ZResponseInput` parses before the contract runs on both POST boundaries, so the `isResponseDataValue` guard on element-id keys is belt-and-braces rather than load-bearing — worth knowing if it ever looks like dead weight.

---

itsjavi:
Thanks @kiln-agent — all four verified and fixed in `c617c67`. Finding 1 is the one that mattered; I reproduced each claim against the module before touching anything.

**1. The `truncated` flag never survived to storage.** Confirmed exactly as described:

```
client (renderer):  { note: "a"×(MAX+500) } → 16384 bytes, flags [{ note, truncated }]
server (re-run):    { note: <that 16384-byte value> } → 16384 bytes, flags []   ← persisted
```

And the asymmetry you drew is the real diagnosis: `coercion_failed` round-trips because the raw text is kept and the verdict re-derived from it, so truncation is the only rule whose *enforcement destroys the evidence it happened*. The module's idempotency guarantee is what makes that certain rather than luck. Your sharper framing of the PR's own justification is right too — not indistinguishable from absent, indistinguishable from **complete**.

`applyIngestContract` now takes `enforceSizeLimit`, and the renderer passes `false`. Two deliberate choices on top of your proposal:

- **It defaults to `true`**, so a caller that forgets gets the bound rather than an unbounded write. The opt-out is the exception and has to be spelled.
- **The flag is still raised when the value *would* be cut**, so the renderer keeps warning about a value it isn't itself cutting — it's describing what the server will store. Dropping the flag along with the cut would have traded a lost DB flag for a lost console warning.

On payload size, which was my one hesitation about not cutting client-side: it isn't a regression. The deleted `handleHiddenFields` never truncated either, so an oversize hidden field already travelled in the request body before this PR. Covered by a test that runs the contract renderer-then-server and asserts on the *server's* flags, since those are the persisted ones.

**2. `verifiedEmail` console warning on every load.** Confirmed — the wrapper spreads it in, it's in `FORBIDDEN_IDS` so no survey can declare it, and nothing in `packages/surveys/src` ever reads it. I went with your second option (exclude it from the log) rather than the first (give it its own prop), because the prop is used *only* by that memo, so removing the spread would cascade into the parent's interface for what is a logging problem.

The suppression is on reserved names generally, not `verifiedEmail` specifically — a name `FORBIDDEN_IDS` covers can never be declared, so a drop against it is plumbing rather than feedback. `RESERVED_DECLARED_FIELD_NAMES` costs no bundle size here: `surveys/validation` is already in the renderer's graph via `embedded-data-resolver`, which imports `getTextContent` from it. Your "teaches developers to ignore the channel" point is the reason this was worth fixing rather than documenting.

**3. `__proto__` loses its value silently.** Confirmed, including that it produces `dropped: []` and `flags: []` — the outcome the module's own docstring rules out. Fixed with `Object.create(null)` as you suggested, plus one addition: the result is spread back into a literal on return, so the key survives as an own property *and* callers still get an ordinary-prototype object. That matters because this object becomes React state and is spread by `ResponseQueue`, and I'd rather not have a null-prototype object loose in either. Verified the key survives a `JSON.stringify`/`parse` round trip (relevant since it lands in a JSONB column) and that nothing reaches the real prototype. Both the declared-key and element-id paths are covered.

**4. Nothing pinned the dumb-pipe behaviour.** Fair, and the first version of my test was wrong in an instructive way: an assertion on `renderWidget` passed even with `triggerSurvey`'s forwarding replaced by `{}`. The filter used to sit on *that* hop, so the test now enters at `triggerSurvey`, and I confirmed it fails when that hop is broken.

Also noted the four things you checked and cleared — particularly the `blocks`/`questions` reachability argument and that `ZResponseInput` parses before the contract runs, which makes `isResponseDataValue` belt-and-braces on the element-id path. That last one is genuinely useful to have on the record; it's the kind of guard that looks like dead weight to a future reader.

Re-ran everything: `@formbricks/types` 348, `@formbricks/js-core` 284, `@formbricks/surveys` 696, `apps/web` 597 files / 7768 tests, typecheck and lint clean on all three packages. Each new test was checked to fail without its fix.

---

itsjavi:
Not acting on CodeRabbit's "Description check" warning, because its three specifics don't match `.github/pull_request_template.md`. Noting it so nobody spends time on it — following it would move this description *away* from the template.

| It asks for | What the template actually says |
| --- | --- |
| Add `Fixes ENG-1845` | `Closing a GitHub issue? Add "Fixes #123".` — a GitHub issue reference, which this PR doesn't have. What it does require is the full Linear URL, which the description carries. |
| Add "the required unchecked breaking-change checkbox" | There is no checkbox. `If there are no breaking changes, leave "None" below.` The label and the table are both conditional on there *being* one. |
| Add "the required Coverage table, Open gaps, and Risks sections" | No Coverage table or Open gaps section exists in the template. `**Risks & regressions**` does, and it's filled in. |

The description has all five template headings — What & why, Linear ticket, How this was tested, Breaking changes, QA / Test Plan — plus all four QA subsections (How to test, Preconditions / test data, Risks & regressions, Migrations / env / cutover). This check passed on the three previous renders with the same structure, so the flip looks like non-determinism rather than anything the last edit changed; that edit only rewrote one sentence about the size bot's line count.

Same disposition as the Docstring Coverage warning next to it, which reports 0.00% after analyzing **1 function across 34 files** on a diff that is disproportionately JSDoc — a scoping failure in the checker rather than undocumented code. Both are non-blocking, and neither is worth a CI cycle. If a human reviewer disagrees on either, say so and I'll change it.

---

apps/web/lib/response/service.test.ts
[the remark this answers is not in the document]
itsjavi: Reversing my earlier call — done in `b4db31b`, and all 33 casts are gone, not just the 8 this PR added.

Worth recording what typing it actually surfaced, because it wasn't the reason either of us gave. I'd assumed the fixture just needed the *selected* payload. It doesn't: `vi.mocked(prisma.response.findUnique)` types `mockResolvedValue` against the method's default instantiation, so the `select` at the call site is invisible to it and it demands the **whole `Response` model**. The fixture only ever supplied the fourteen names in `responseSelection` — it was missing `contactId` outright, plus `ingestFlags` once this PR added the column. So the casts weren't suppressing style, they were suppressing a genuine gap between the fixture and the schema.

The type is therefore the selection plus exactly those two scalars, which together are all sixteen columns, with `responseSelection` carrying the `contact` and `tags` relations the code under test reads:

```ts
type MockCurrentResponse = Prisma.ResponseGetPayload<{
  select: typeof responseSelection & { contactId: true; ingestFlags: true };
}>;
```

`overrides` is now `Partial<MockCurrentResponse>` instead of `Record<string, unknown>`.

On your point that `apps/web/tsconfig.typecheck.json` excludes `**/*.test.ts` — that cut both ways here, so I didn't take it on trust. Nothing in CI would have told me whether the casts were load-bearing, so I ran `tsc --noEmit` against a throwaway config that includes this file: **33 errors** with the casts removed and the fixture untyped, **0** once it matches the model. Suite unchanged at 598 files / 7779 tests.

That exclusion does mean this fixture can drift again without CI noticing — it now fails to *compile* rather than silently passing `undefined`, but only for someone who typechecks tests locally. Putting test files into a typecheck project that runs in CI is a repo-wide decision well beyond this PR, so I've left it alone rather than smuggling it in here.

---

pandeymangg (approved):
LGTM 🚀

PR #8937, opened 2026-08-20
chore(deps): adopt pnpm catalogs for shared dependency versions

mattinannt:
Fixes ENG-1689

## What & why

Nothing kept a shared dependency at one version across the 21 workspaces — alignment was manual discipline, and `nodeLinker: hoisted` hid a split until it broke. One split had already become a latent type bug.

Every dependency declared by **two or more** workspaces is now pinned once in a `catalog:` block in `pnpm-workspace.yaml` (64 entries, derived by script rather than by hand) and referenced as `"catalog:"` from each `package.json`. Bumping a shared dep now means editing the catalog, which moves every consumer at once.

Drift resolved:

- **`redis`** — `apps/web` was on 4.7.1, `@formbricks/cache` on 5.11.0. `apps/web` had no nested copy, so `apps/web/lib/cache/index.ts` resolved `RedisClientType` from the **hoisted v4** and used it to type the **v5** client that `getCacheService()` actually returns. It typechecked only because that facade type is hand-written and structural. Unified on 5.11.0. Note this is a real runtime major bump for Better Auth session/token/rate-limit storage, not only a type alignment — the only value-level use is `modules/auth/lib/secondary-storage.ts`, whose API surface is unchanged across the major.
- **`vite`** — `packages/ai` was on 8.0.16, the other fourteen on 7.3.5. Unified **down** to 7.3.5, because up is blocked: `@vitejs/plugin-react` has no release accepting both majors (5.1.4 caps at vite 7; 6.1.0 requires `^8.0.0` exclusively, plus three new peers), `vite-prerender-plugin` (transitive of `@preact/preset-vite`, so `packages/surveys`) caps at 7.x, and Vite 8 swaps Rollup for rolldown while the `database`, `survey-ui` and `js-core` configs use Rollup-shaped API. `packages/ai` uses no Vite plugins, which is why its major went unnoticed.
- **Tailwind** — `tailwindcss`, `@tailwindcss/postcss` and `@tailwindcss/vite` unified on 4.3.1 (`apps/web` was already there; `survey-ui` and `surveys` were on 4.2.4).
- **`eslint-plugin-react-hooks`** — `config-eslint` was on 5.2.0 while `apps/storybook` was on 7.1.1. Unified on 7.1.1, which *requires* moving to `configs.flat.recommended` (v7's `recommended-latest` is the legacy eslintrc format and crashes ESLint 9). The compiler-era rules with pre-existing violations are parked under the existing **ENG-2366** opt-out, exported once from `base.mjs` rather than duplicated — version alignment here, rule adoption there. This also unblocks ENG-2366's third scope item, which was explicitly waiting on the shared tiers leaving v5; `packages/config-eslint/README.md` is corrected accordingly.
- **`packages/email`** declared `vitest` without `vite`, so `autoInstallPeers` floated the peer to the newest in range and pulled vite 8 + rolldown back in. Now pinned.

A catalog only aligns names that are *in* it, so two guards back it:

- **`scripts/check-catalog.mjs`** (runs in `pnpm lint`) fails if a workspace declares a literal version for a catalogued name, shares a dependency with another workspace without catalogueing it, or points `catalog:` at a name the catalog does not define. Peer *ranges* are exempt — a peer range is a compatibility declaration for consumers, not an install pin (`survey-ui` declares react `^19.0.0` while pinning 19.2.6 to build against); "is this a single pinned version" is tested positively, so no range spelling can slip through. Workspace globs come from `pnpm-workspace.yaml` itself, and any pattern the resolver cannot expand is a hard error rather than silently uncovered.
- **`packages/survey-ui/scripts/assert-pnpm-publish.mjs`** (`prepublishOnly` + `prepack`). `survey-ui` is the one workspace package that is actually published, and 13 of its runtime dependencies are now `catalog:`. `pnpm publish` resolves those; npm does not, and would ship `"clsx": "catalog:"` verbatim. Both hooks are needed: `prepublishOnly` catches `npm publish .`, `prepack` catches `npm pack` (whose tarball is publishable later via `npm publish <tarball>`, which runs no scripts). `--ignore-scripts` stays uncloseable from a manifest and is documented as such. Relevant because every release so far was cut by hand — 1.0.0 through 1.0.4 all went out in one 66-minute session, with no release workflow.

**The guard has already earned its keep.** Merging `main` in brought #8922, which added `js-yaml` and `@types/js-yaml` to `docker/package.json`. That moved `@types/js-yaml` from one consumer to two (`docker` + `packages/workflows`), and `check-catalog` correctly demanded it be catalogued — so it now is, at 4.0.9. `js-yaml` was already catalogued at exactly the version #8922 wanted. That is the mechanism catching the first shared dependency to arrive from outside this branch, which is precisely the drift class ENG-1689 exists to close.

**Why this isn't split** (the size bot asks): a catalog migration is atomic. A `catalog:` reference with no catalog entry fails `pnpm install`, and a catalogued name left at a literal version fails the new guard — so the catalog block and all 21 manifests have to land together or the intermediate commits don't install. The diff is mostly mechanical: `+222/-217` across the 21 `package.json` files and `+540/-603` in the lockfile (excluded from the count), leaving the 101-line catalog block and the two guard scripts as the only code to read.

## Breaking changes

None. Every change is an internal dependency version or lint/build config: no API or SDK shape, endpoint, route, env var, config key, default or webhook payload changes, and nothing requires manual action. `@formbricks/survey-ui`'s published dependency versions will differ at its *next* release, but this PR does not publish, and its `peerDependencies` (`react`/`react-dom` `^19.0.0`) are untouched.

## Migrations & env

- none

## How this was tested

The Tailwind 4.2.4 → 4.3.1 bump is the only change that can alter what a user sees, so it is covered twice over, on **both** Tailwind pipelines this repo uses:

- **`@tailwindcss/vite`** (how `apps/storybook` builds survey-ui) — covered at the pixel level by the 270-story comparison below.
- **`@tailwindcss/postcss`** (how the published `survey-ui` bundle and the `@formbricks/surveys` widget are built) — covered behaviourally by CI's E2E job, which builds that exact bundle and drives it in a real browser, plus the declaration-level CSS diff.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| **survey-ui renders identically on Tailwind 4.3.1 and 4.2.4** | manual (visual regression, 270 stories) | Built `apps/storybook` statically on each Tailwind version and screenshotted **every** story with Playwright/Chromium — 270 stories across all 23 survey-ui story files, fixed 900×700 viewport, `reducedMotion`, `animations: "disabled"`, `document.fonts.ready` await. **All 270 PNG pairs are byte-identical.** Not "no visible difference" — the files compare equal. Method and re-run steps in the fold below |
| **A live survey still renders and can be completed, on the built bundle** | e2e | CI **Run E2E Tests** green on `526dffe`. Eight specs drive a real rendered survey — `survey-keyboard`, `survey-progress-a11y`, `survey-page-title`, `survey-accessibility`, `survey-email-preview`, `survey`, `survey-overview`, `cross-tenant-survey-access`. `survey-keyboard.spec.ts` alone walks single-select → NPS → dropdown with arrows/Space/Enter and clicks the visible label for `Rate 9 out of 10`. This is the **PostCSS** path: the job runs `pnpm build --filter=@formbricks/web...`, which has `@formbricks/survey-ui` and `@formbricks/surveys` in scope, and surveys' `copyCompiledAssetsPlugin` copies the built UMD bundle to `apps/web/public/js/` — so E2E exercises the same artifact self-hosters ship |
| The guard actually runs as part of `pnpm lint` | manual | `pnpm lint` → 21/21 turbo tasks pass, then `api:v3:lint`, then `$ node scripts/check-catalog.mjs` / `✓ catalog check passed (64 catalogued dependencies)`. This is the integration that matters, since the guard is only useful if CI invokes it |
| Every Vite-consuming workspace resolves one Vite | manual | `for d in apps/web apps/storybook docker packages/{ai,cache,database,email,survey-ui,surveys,vite-plugins}; do (cd $d && node -e 'console.log(require("vite/package.json").version)'); done` → all 10 print 7.3.5. One `vite` in `node_modules`; zero `vite@8` in the lockfile's `importers:` block |
| `apps/web` types its redis client against v5 | manual | `node -e 'console.log(require("./apps/web/node_modules/redis/package.json").version)'` → 5.11.0 (previously resolved the hoisted 4.7.1). Root still holds 4.7.1, but only as `typeorm`'s optional peer via `@boxyhq/saml-jackson` — no workspace declares it |
| Guard rejects a literal version for a catalogued name | unit (mutation) | Set `devDependencies.typescript` to `"5.9.4"` in `packages/logger/package.json`, then `node scripts/check-catalog.mjs` → exit 1 naming file, block, literal and catalog value |
| Guard rejects a shared dep absent from the catalog | unit (mutation) | Delete `redis: 5.11.0` from `pnpm-workspace.yaml`'s `catalog:`, set `"redis": "5.11.0"` in `apps/web/package.json` + `packages/cache/package.json`, then `node scripts/check-catalog.mjs` → exit 1 naming both consumers |
| Guard covers a package that did not exist when it was written | unit (mutation) | `mkdir packages/zz-probe`, write `{"name":"@formbricks/zz-probe","devDependencies":{"typescript":"5.9.1"}}`, then `node scripts/check-catalog.mjs` → exit 1 (glob resolution, not a hardcoded list) |
| Guard exempts partial/wildcard peer ranges (review finding) | unit (mutation) | Set `peerDependencies.react` in `packages/survey-ui/package.json` to each of `19`, `19.0`, `19.x`, `1.2.3 - 2.0.0` → exit 0 for all four. Before `b5bd9e7` all four exited 1 |
| Guard enforces exact peers, including `=` form (review finding) | unit (mutation) | Same field set to `19.2.6`, `=19.2.6`, `=19.2.6-beta.1` → exit 1 for all three. `=` forms exited 0 before `8a7e627` |
| Guard rejects an unresolvable `catalog:` reference (review finding) | unit (mutation) | Add `"definitely-not-catalogued": "catalog:"` to `packages/database/package.json`, and separately `"typescript": "catalog:react19"` → exit 1 on both. Both exited 0 before `b5bd9e7` |
| Guard fails loudly on a glob it cannot expand (review finding) | unit (mutation) | Change `pnpm-workspace.yaml`'s `packages` to `packages/**`, then `!docker`, then `nope/*` → exit 1 each, naming the pattern. Before `8a7e627`, `packages/**` scanned only the root manifest and still exited 0 |
| Tailwind 4.2.4 → 4.3.1 changes no generated CSS that matters | manual | Declaration-level diff of the built `survey-ui.css` on both versions: 19 of 23 changed declarations are constant-folding (`calc(var(--spacing) * 0)` → `0`, `* 1` → `var(--spacing)`), computed values identical; the only removed rules are `#fbjs .start` / `#fbjs .end`, dead CSS no markup uses. Fold below |
| Survey stylesheet stays `#fbjs`-scoped after the bump | unit (guard) | `assert-scoped-css.mjs` passes on a forced rebuild with `packages/survey-ui/dist`, `packages/surveys/dist`, `apps/web/public/js/surveys.*` and `node_modules/.cache/turbo` cleared first |
| `packages/ai` export surface survives the Rollup rebundle | manual | `node -e 'import("./packages/ai/dist/index.js").then(m=>console.log(Object.keys(m).sort()))'` → all 11 runtime exports match `src/index.ts`, `.d.ts` emitted. Bundle bytes differ (rolldown → rollup), exports do not |
| `survey-ui` publish/pack blocked on npm, allowed on pnpm | unit (mutation) | In `packages/survey-ui`: `npm run prepublishOnly` → exit 1 listing all 13 `catalog:` specifiers; `npm pack` → exit 1, **no tarball produced**; `pnpm run prepublishOnly` and `pnpm pack` → exit 0; `env -u npm_config_user_agent node scripts/assert-pnpm-publish.mjs` → exit 1 (fails closed) |
| The guards do not disturb pack/build/install | manual | `pnpm pack`, `pnpm install` and `pnpm build --filter=@formbricks/survey-ui` all succeed (`prepack` does not run on install) |
| A published tarball carries no unresolved specifiers | manual | `pnpm pack` then `tar -xzOf *.tgz package/package.json` → concrete versions for all 14 deps, zero `catalog:`/`workspace:` entries |
| The merge of `main` (#8922 conflict) is sound | manual | Conflict was in `docker/package.json`; resolved keeping #8922's two new deps with all five on `catalog:`. Lockfile auto-merged but was regenerated with `pnpm install` rather than trusted. On the merged tree: `pnpm typecheck` 27/27, `pnpm lint` 21/21 + guard at 64, `pnpm test` 26/26, `pnpm format:check` clean. CI on `526dffe`: all 20 checks green, including E2E and Validate Docker Build |

<details>
<summary>Visual regression: 270 survey-ui stories, Tailwind 4.2.4 vs 4.3.1, byte-identical</summary>

The app itself needs Docker (Postgres + Redis), which was unavailable in the authoring environment — but Storybook does not. `apps/storybook` is a Vite app that consumes `packages/survey-ui/src/**` as raw source, so it renders every component with no database at all.

Method, per Tailwind version:

1. Pin the Tailwind trio in `pnpm-workspace.yaml`'s `catalog:` (`tailwindcss`, `@tailwindcss/postcss`, `@tailwindcss/vite`), `pnpm install`
2. `pnpm --filter storybook build-storybook`
3. Serve `apps/storybook/storybook-static` over HTTP and drive Playwright/Chromium over every entry in its `index.json` with `type === "story"`, navigating to `iframe.html?id=<id>&viewMode=story`
4. Wait for `#storybook-root` to have children, then `document.fonts.ready`, then screenshot with `animations: "disabled"`

Determinism controls: fixed 900×700 viewport, `deviceScaleFactor: 1`, `reducedMotion: "reduce"`, `colorScheme: "light"`, and `Math.random` stubbed to a constant via an init script.

Result: **270/270 stories captured on each version, and all 270 pairs are byte-identical** (`cmp -s` equal for every file). This covers all 23 survey-ui story files — consent, cta, date, file-upload, form-field, matrix, multi-select, nps, open-text, picture-select, ranking, rating, single-select, plus the general components (alert, button, checkbox, dropdown-menu, element-header, input, label, progress, radio-group, textarea) — including their disabled, required, error, RTL, colour-coding and styling-playground variants.

This is the **`@tailwindcss/vite`** pipeline. The `@tailwindcss/postcss` pipeline — the one that produces the published bundle — is covered behaviourally by the E2E row above, which drives that built bundle in a real browser. The two are complementary: pixels on one path, behaviour on the other.

</details>

<details>
<summary>Tailwind 4.2.4 → 4.3.1: the full generated-CSS diff</summary>

Reproduce by pinning the Tailwind trio in `pnpm-workspace.yaml`'s `catalog:` back to `4.2.4`, `pnpm install`, `pnpm build --filter=@formbricks/survey-ui --force`, and diffing `packages/survey-ui/dist/survey-ui.css` against the 4.3.1 build. Stylesheet went 54,818 → 54,375 bytes; 23 declarations removed, 19 added.

**Constant folding — identical computed values, simpler output:**

```
- bottom:calc(var(--spacing) * 0)     + bottom:0
- inset:calc(var(--spacing) * 0)      + inset:0
- margin:calc(var(--spacing) * 1)     + margin:var(--spacing)
- padding:calc(var(--spacing) * 1)    + padding:var(--spacing)
- gap:calc(var(--spacing) * 1)        + gap:var(--spacing)
  …and 14 more of the same two shapes (`* 0` → `0`, `* 1` → `var(--spacing)`)
```

**The only genuinely removed rules:**

```
- #fbjs .start{inset-inline-start:var(--spacing)!important}
- #fbjs .end{inset-inline-end:var(--spacing)!important}
```

These were dead CSS. 4.2.4 resolved a bare `start`/`end` candidate — which Tailwind's scanner extracts as a token from compound classes like `text-start` and `justify-end` — into a real utility; 4.3.1 no longer does. No element uses `class="start"` or `class="end"` (`grep -rnE 'class(Name)?="(start|end)"' packages/{survey-ui,surveys}/src` → no matches), and every compound utility the source actually uses is still emitted in both builds (`justify-start`, `text-start`, `flex-col` all present, byte-for-byte). That accounts for the 443-byte reduction.

</details>

**Open gaps**

- **No pixel-level assertion on the PostCSS-built bundle specifically.** E2E drives it in a real browser and would catch a functional break, but E2E asserts behaviour and accessibility, not appearance — a purely cosmetic regression could pass it. The pixel evidence is on the Vite path (270 stories) and the CSS diff shows the version delta is a null-effect codegen change, so the residual is narrow, but it is not zero.
- **Chromatic does not run on pull requests.** `.github/workflows/chromatic.yml` triggers only on `push` to `main` and `workflow_dispatch`, so the repo's visual-regression tooling gave this PR no signal and would only have caught a survey-ui regression *after* merge. The 270-story comparison above is a local stand-in. Wiring Chromatic to `pull_request` would make this a permanent per-PR gate — worth its own change rather than an add-on here.
- **The guard's mutation cases have no spec file.** Every row marked `unit (mutation)` was verified by hand and re-verified after each fix, but nothing re-runs them in CI. The root workspace has no test runner (`pnpm test` is `turbo run test`), so a spec for a root `.mjs` script has no natural home, and I would rather say so than bolt one into an unrelated workspace. Tracked in ENG-2534 — until then a future refactor of `check-catalog.mjs` could regress these without CI noticing.
- `apps/web` reports 40 `react-hooks/exhaustive-deps` **warnings** (0 errors; the rule is explicitly `warn` with no `--max-warnings 0`, and that predates this PR). I did not establish the count is identical to v5's, so some may be new. Non-blocking either way.
- The lockfile still resolves a `vite@8.0.16`, and this predates the PR: `@vitest/eslint-plugin` declares `vitest` as an **optional** peer with range `*`, so `autoInstallPeers` materialises the newest vitest, whose own vite peer picks 8. No importer references it and nothing builds with it — but it is why "the Vite split is resolved" and "no vite 8 in the lockfile" are not the same claim.
- **Single-consumer siblings of catalogued packages are not covered** ([thread](https://github.com/formbricks/formbricks/pull/8937#discussion_r3823827956)). `prisma`, `@prisma/adapter-pg` and `@prisma/instrumentation` must match catalogued `@prisma/client`, and `playwright` must match `@playwright/test`; all are aligned today, but bumping the catalogued half would split them silently. Closing it needs declared lockstep groups plus a change to the "single consumer stays in its own package.json" policy — deferred to ENG-2534.
- `--ignore-scripts` bypasses the publish guard entirely, and `survey-ui` has no release workflow to move the check into. Also its npm release stays at the 8-month-stale 1.0.4 — this PR only makes the next publish safe, it does not cut one.
- `packages/survey-ui`'s own `lint` script is `eslint src`, so its `scripts/` directory is linted only by lint-staged, never by `pnpm lint`. Pre-existing — `assert-scoped-css.mjs` has the same hole — and noted in ENG-2534.

**Risks**

- A catalog entry now moves every consumer at once. That is the point, but it widens the blast radius of a bad bump compared with a single-package edit.
- react-hooks v7's other twelve rules are now live in the shared tiers, so new code in `survey-ui`, `surveys` and `email` can trip `static-components`, `use-memo` or `set-state-in-render` where it previously could not.
- redis v5 in `apps/web`: if session, token or rate-limit storage misbehaves, start at `modules/auth/lib/secondary-storage.ts`.
- Tailwind: between 270 byte-identical stories, the green E2E run against the built bundle, and the CSS diff, this is about as closed as evidence gets. If survey rendering still looks wrong, start at `packages/survey-ui/dist/survey-ui.css` and the `#fbjs` scoping.

---

> [!NOTE]
> **AI model used** — `claude-opus-5 (Claude Code 2.1.237)`, reasoning effort `high`.


scripts/check-catalog.mjs
```diff
@@ -0,0 +1,130 @@
...
+import yaml from "js-yaml";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const DEP_BLOCKS = ["dependencies", "devDependencies", "peerDependencies", "optionalDependencies"];
+
+// Specifiers that name no single version to align on.
+const isRange = (spec) => /^[\^~><*]/.test(spec) || spec.includes("||");
```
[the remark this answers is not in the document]
mattinannt: Confirmed and fixed in `b5bd9e7`, though not the suggested way — flagging the deviation.

The bug is real and slightly wider than the three cases named: `1.2.3 - 2.0.0` (hyphen range), a dist-tag, and the empty string were also misread as pins, so a peer declared with any of them would have been reported as a catalog violation.

I did not reach for a SemVer range parser. Enumerating range syntaxes is what broke here, and a parser is still an enumeration — just someone else's, plus a new root dependency for one predicate. Inverting the test removes the failure mode instead of widening it:

```js
const isExactVersion = (spec) => /^\d+\.\d+\.\d+(?:[-+][0-9A-Za-z.-]+)?$/.test(spec);
```

The guard only ever needs to answer "is this a single pinned version?", because that is the only specifier a catalog entry can replace. Everything that is not one is exempt by construction, so no future range spelling can slip through — which `semver.validRange` would not give me, since it accepts an exact version as a valid range and so cannot distinguish the two cases on its own.

Verified by mutation on `packages/survey-ui` `peerDependencies.react`:

| spec | before | after |
| --- | --- | --- |
| `19`, `19.0`, `19.x`, `1.2.3 - 2.0.0` | exit 1 (wrong) | exit 0 |
| `19.2.6` | exit 1 | exit 1 (still enforced) |

The sibling finding about unresolvable `catalog:` references is in the same commit: both `catalog:` and `catalog:<name>` now resolve against `pnpm-workspace.yaml`, and an undefined name or group fails with its own message. Agreed that pnpm would fail the install anyway, but the guard owns this contract, so it should be the thing that says so.

---

packages/survey-ui/scripts/assert-pnpm-publish.mjs
```diff
@@ -0,0 +1,49 @@
...
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+
+/* eslint-disable-next-line turbo/no-undeclared-env-vars -- injected by the package manager at
+   publish time, not build config: it is never a turbo task input, so declaring it in turbo.json
+   would only make every task hash against the client's UA string. */
+const userAgent = process.env.npm_config_user_agent ?? "";
+const isPnpm = userAgent.startsWith("pnpm/");
+
+if (isPnpm) process.exit(0);
```
kiln-agent, line 33: **The guard doesn't cover `npm pack` + `npm publish <tarball>`, which is the actual break path.**

`prepublishOnly` runs on `npm publish <dir>` / `pnpm publish`, but it does **not** run on `npm pack`. I verified: `npm pack` here produces a tarball whose `package/package.json` still contains 33 `"catalog:"` specifiers (pnpm rewrites them, npm does not). Anyone who then runs `npm publish survey-ui-x.y.z.tgz` — or `npm publish --ignore-scripts` — ships exactly the unresolvable specifiers this guard exists to prevent, and no package script fires to stop them.

`prepublishOnly` is the right hook for the common `npm publish .` mistake, so this is a real (if partial) safety net. But since the comment claims the guard prevents shipping broken `catalog:` specifiers, it's worth either tightening (e.g. also assert in `prepack`, which *does* run on `npm pack`) or narrowing the comment to state the tarball-publish hole remains open.

---

scripts/check-catalog.mjs
```diff
@@ -0,0 +1,163 @@
...
+  spec.startsWith("npm:") ||
+  spec.startsWith("git");
+
+// `catalog:` uses the default catalog; `catalog:<name>` a named one.
+const CATALOG_REF = /^catalog:(.*)$/;
+
+// Deliberate exceptions to rule 2: a dependency shared by 2+ workspaces that must NOT be
+// catalogued. Add an entry only with a comment saying why the two consumers are meant to
+// hold different versions — if they are meant to match, the fix is a catalog entry, not
+// an exemption. Empty today, and that is the healthy state.
```
kiln-agent, line 55: **`resolveWorkspaceDirs` silently ignores any glob that isn't `dir/*` or a literal directory — the exact silent escape the file's comment forbids.**

The `packages` field in `pnpm-workspace.yaml` accepts full glob syntax (`apps/**`, `packages/*/`, negations, etc.). This handles only `pattern.endsWith("/*")` and the literal-directory fallback; everything else is dropped with no warning.

Verified: changing the workspace globs to `apps/**` / `packages/**` makes the check scan only the root manifest yet still print `✓ catalog check passed (63 catalogued dependencies)` and exit 0 — coverage collapses to zero packages while CI stays green. Today's globs happen to be `apps/*` / `packages/*` so it works, but a future glob edit would quietly disable the whole check. Consider using pnpm's own workspace resolution (`@pnpm/find-workspace-packages`) or at least failing loudly on an unrecognized glob shape.

---

scripts/check-catalog.mjs
```diff
@@ -0,0 +1,163 @@
...
+import yaml from "js-yaml";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const DEP_BLOCKS = ["dependencies", "devDependencies", "peerDependencies", "optionalDependencies"];
+
+// A single pinned version — the only kind of specifier the catalog can replace. Tested
+// positively rather than by listing range syntaxes: a range has too many spellings to
```
kiln-agent, line 33: **`isRange` misses several valid semver range forms, so their peer-dependency exemption is lost.**

`/^[\^~><*]/ || includes("||")` doesn't recognize `x`-ranges (`19.x`, `1.2.x`), hyphen ranges (`1.2.3 - 2.0.0`), or `=`-prefixed specs. Since the `peerDependencies` exemption at line 87 keys off `isRange`, those forms are treated as pinned versions and flagged.

Verified: putting `"react": "19.x"` in survey-ui's `peerDependencies` fails the check with a "use `catalog:`" message — which would be wrong advice, since narrowing a published package's peer range to a single version is a regression. A peer range is a compatibility declaration and should stay exempt regardless of which valid range syntax it uses. Worth broadening the detector (or reusing `semver.validRange` vs `semver.valid`).

---

scripts/check-catalog.mjs
```diff
@@ -0,0 +1,163 @@
...
+    for (const [name, spec] of Object.entries(manifest[block] ?? {})) {
+      if (isNotAVersion(spec)) continue;
+
+      // A `catalog:` reference that names nothing is a typo pnpm only catches at install
+      // time; report it here, where the rest of the contract is checked.
+      const catalogRef = CATALOG_REF.exec(spec);
+      if (catalogRef) {
+        const group = catalogRef[1];
+        if (!group && !(name in catalog)) {
+          brokenRefs.push(`  ${label}: ${block}.${name} is "catalog:" but the catalog has no ${name} entry`);
```
kiln-agent, line 131: **The `uses.size >= 2` rule doesn't catch single-consumer siblings that must version-match a catalogued package.**

The rule enforces "a name used by 2+ workspaces must be catalogued." But several packages have a single consumer yet are meant to move in lockstep with a *different*, already-catalogued name — and those pairs slip through:

- `prisma` / `@prisma/adapter-pg` vs catalogued `@prisma/client`
- `playwright` vs `@playwright/test`
- `eslint` vs `@eslint/js`
- `@storybook/addon-*` vs `storybook`

Bumping the catalogued half silently splits the pair (e.g. `@prisma/client` 6.x in the catalog, `prisma` CLI 5.x still literal), which is precisely the version-skew class this migration is meant to close. This is a design gap rather than a bug in the current tree — consider a small "keep in sync with" grouping so related-but-single-consumer packages are caught too.

---

kiln-agent (commented):
## Review: pnpm-catalog migration (ENG-1689)

Reviewed both commits (`c9fa440` catalog adoption + `5eb29c1` publish guard), all 30 files, the two new scripts, the eslint tier changes, and the lockfile. The migration itself is solid — I verified the version deltas hidden behind `catalog:` and none is an accidental regression:

- **redis 4.7.1 → 5.11.0 (`apps/web`)** is a real runtime major bump for Better Auth session/token/rate-limit storage, not just the type alignment the catalog comment describes. All five call sites (`createClient`, `get`, `del`, `getDel`, `eval`, and the deprecated `{ EX }` on `SET`) are compatible with v5. **Worth calling out in the PR description / breaking-change checkbox**, but no defect.
- vite 8→7 downgrade in `packages/ai` builds clean, all 90 tests pass.
- react-hooks v7 flat configs: eslint clean across all consumers (`--max-warnings 0`); `reactCompilerRulesOptOut` placement is safe.
- tailwind 4.2.4 → 4.3.1: surveys build succeeds, CSS still fully `#fbjs`-scoped.
- Lockfile is byte-stable under `--frozen-lockfile`; catalog is complete for the current tree.

### Findings (4, none blocking)

All four are in the two guard/lint **scripts**, not the manifests — they're gaps in the safety net rather than breakage in the current tree. Each is inline with a reproduction:

1. `assert-pnpm-publish.mjs` — `prepublishOnly` doesn't fire on `npm pack` + `npm publish <tarball>` (or `--ignore-scripts`); the packed tarball still carries 33 `catalog:` specifiers.
2. `check-catalog.mjs` `resolveWorkspaceDirs` — only handles `dir/*`/literal globs; any other glob silently collapses coverage to zero while the check still passes.
3. `check-catalog.mjs` `isRange` — misses `x`-ranges/hyphen/`=` specs, so those peer ranges lose their exemption.
4. `check-catalog.mjs` `uses.size >= 2` — single-consumer siblings that must match a catalogued package (`prisma`↔`@prisma/client`, `playwright`↔`@playwright/test`, etc.) aren't caught.

Nice work — clean, well-documented change with genuinely useful guard rails. The findings are about hardening those guards, not the migration.

---

scripts/check-catalog.mjs
mattinannt: Agreed, and this was the worst of the four — fixed in `8a7e627`. Reproduced exactly as described: with `packages/**` the check scanned only the root manifest and still printed `✓ catalog check passed (63 catalogued dependencies)` and exited 0.

I deliberately did **not** widen the parser, because "understands more globs" just moves the cliff. Anything unexpandable now throws:

- an unsupported shape (`**`, `!`, braces, `?`) → error naming the pattern
- a `dir/*` whose parent does not exist → error
- a glob set matching zero workspaces → error, rather than reporting success

Verified: `packages/**`, `!docker` and `nope/*` each exit 1 with the offending pattern in the message.

I skipped `@pnpm/workspace.find-packages` for now — it would be the right answer if we ever need real glob support, and the error message says so explicitly so the next person doesn't quietly re-implement globbing. Adding a pnpm internal as a root dependency to resolve two static patterns felt like the wrong trade today; failing loudly gets the safety without the dependency.

---

packages/survey-ui/scripts/assert-pnpm-publish.mjs
mattinannt: Confirmed and tightened in `8a7e627`. I reproduced it before fixing: `npm pack` produced a tarball whose manifest carried **13 unresolved runtime `catalog:` specifiers** (33 including devDependencies, though only the runtime ones break a consumer's install).

`prepack` now runs the same assertion, so `npm pack` fails before a tarball exists — verified: it exits 1 and produces no file. `pnpm pack` still succeeds and still emits a manifest with zero `catalog:`/`workspace:` left, and `pnpm publish`, `pnpm build` and `pnpm install` are all unaffected (`prepack` doesn't run on install).

You were right that the comment overclaimed, so I did both things you suggested rather than picking one: the hook is wider *and* the doc now states the remaining hole plainly —

> Neither hook fires under `--ignore-scripts`; nothing in package.json can close that, so it stays a known hole.

That one is genuinely uncloseable from a manifest. If we want it closed, it needs to move outside the package — a release workflow that runs `pnpm publish`, or a CI check that packs and asserts the tarball is clean. Worth a ticket alongside the fact that this package has no release workflow at all (1.0.0–1.0.4 all went out by hand in one session), but I didn't want to build a release pipeline inside a dependency-refactor PR.

---

scripts/check-catalog.mjs
mattinannt: Both halves of this are fixed, in two commits — and they turned out to be bugs in opposite directions, which is worth spelling out.

**False positive** (`19.x`, `19`, `19.0`, hyphen ranges, dist-tags, empty string), fixed in `b5bd9e7`: these were treated as pins and flagged, exactly as you describe. Rather than extend the range list I inverted the predicate, since enumerating range syntaxes is the fragile part:

```js
const isExactVersion = (spec) => /^=?\d+\.\d+\.\d+(?:[-+][0-9A-Za-z.-]+)?$/.test(spec);
```

**False negative** (`=1.2.3`), fixed in `8a7e627`: you and CodeRabbit both flagged `=`, but note it fails the *other* way. `=1.2.3` is an exact pin, so exempting it meant a pinned peer silently escaped enforcement. That's why it isn't simply grouped with the ranges above — it's now accepted as a pin (hence the `=?`), not exempted.

On `semver.validRange` vs `semver.valid`: that pairing would work, but the guard only needs "is this a single pin?", and `validRange` accepts an exact version too, so it can't separate the two cases alone. The positive test answers it directly with no new root dependency.

Verified by mutation on `packages/survey-ui` `peerDependencies.react`: `^19.0.0`, `19.x`, `19`, `1.2.3 - 2.0.0` all exit 0; `19.2.6`, `=19.2.6`, `=19.2.6-beta.1` all exit 1.

One honest gap: these regression cases are verified by hand, not by a test file. The root workspace has no test runner (`pnpm test` is `turbo run test`), so a spec for a root `.mjs` script has nowhere natural to live, and I'd rather say that than bolt one into an unrelated workspace. Worth a ticket to give the repo-level scripts a home.

---

scripts/check-catalog.mjs
mattinannt: This one I'm deliberately **not** fixing in this PR, and I want to be explicit about why rather than quietly leaving it.

The gap is real and I agree with the diagnosis. I checked the current tree and every pair is aligned today — `prisma`, `@prisma/adapter-pg`, `@prisma/instrumentation` and catalogued `@prisma/client` are all `7.8.0`; `playwright` and catalogued `@playwright/test` are both `1.58.2`; `eslint` and `@eslint/js` are both `9.39.5`. So there is no live skew, but bumping the catalogued half would split them silently, exactly as you say.

The reason it isn't a quick fix is that it contradicts a documented decision this PR makes. `pnpm-workspace.yaml` and `AGENTS.md` both state that single-consumer deps deliberately stay in their own `package.json`, because a catalog entry with one referent is indirection for nothing. Closing this gap means changing that rule to "single consumer **and** not part of a catalogued family", plus a new mechanism, since cataloguing the siblings alone wouldn't *enforce* anything — someone could still write `prisma: 7.9.0` next to `@prisma/client: 7.8.0` in the catalog. Enforcement needs declared lockstep groups:

```js
// scripts/check-catalog.mjs
const LOCKSTEP = [
  ["prisma", "@prisma/client", "@prisma/adapter-pg", "@prisma/instrumentation"],
  ["playwright", "@playwright/test"],
  ["storybook", "@storybook/react", "@storybook/react-vite", "@storybook/addon-a11y", /* … */],
];
// fail if the catalog resolves two members of a group to different versions
```

That's ~15 lines plus roughly a dozen new catalog entries and a policy change to AGENTS.md — a membership-policy decision worth its own review, not a late addition to a PR that CI already flags at 828 lines. It's also strictly additive: nothing here blocks it later.

Storybook is the strongest case (its packages hard-require matching versions and there are seven of them), Prisma next, `eslint`/`@eslint/js` the weakest — those only need matching majors, so a lockstep rule would be stricter than the actual constraint.

I'll raise this as a follow-up ticket referencing this thread. Leaving it unresolved so it's the author's call whether to pull it forward — happy to do it here instead if you'd rather not split it.

---

mattinannt: Filed as **ENG-2534** — https://linear.app/formbricks/issue/ENG-2534/enforce-lockstep-versions-for-catalogued-package-families-prisma

It carries the pair table, the `LOCKSTEP` sketch, the reason it wasn't folded into this PR, and the `eslint` ↔ `@eslint/js` caveat (majors only, so a strict rule would over-constrain it). I also recorded the two adjacent gaps from this review round there rather than losing them: repo-level scripts having no test home, and `--ignore-scripts` bypassing the publish guard on a package with no release workflow.

Leaving this thread open until someone decides whether to pull it into this PR or take it in ENG-2534.

---

mattinannt: Filed as [ENG-2535](https://linear.app/formbricks/issue/ENG-2535/catch-version-skew-between-catalogued-packages-and-their-single) with the family table, the policy change it implies, and the note that `eslint`/`@eslint/js` only need matching majors so exact lockstep would be stricter than the real constraint. This thread is linked from the ticket.

Also recorded under `Open gaps` in the PR description, so it doesn't merge as an unstated hole.

---

mattinannt: Correction to my previous reply: the right ticket is **[ENG-2534](https://linear.app/formbricks/issue/ENG-2534/enforce-lockstep-versions-for-catalogued-package-families-prisma)**, not ENG-2535.

ENG-2534 was already filed for this gap fifteen minutes before mine, and it is the better writeup — it covers the same family table plus the two adjacent gaps from this review (repo-level scripts having no test home, and `--ignore-scripts` bypassing the publish guard). It is also the ticket the PR description references. I have marked ENG-2535 as a duplicate of it and cancelled it.

No change to the substance of the reply above: still deferred, still the author's call whether to pull it into this PR.

---

mattinannt: Resolving: auto-merge is now enabled on this PR, so the author's call is to take this in **ENG-2534** rather than pull it forward. Nothing here blocks it — the change is strictly additive.

Recap of what carries over: `prisma` / `@prisma/adapter-pg` / `@prisma/instrumentation` ↔ catalogued `@prisma/client` (all `7.8.0` today), `playwright` ↔ `@playwright/test` (both `1.58.2`), `eslint` ↔ `@eslint/js` (both `9.39.5`, majors-only constraint so strict lockstep would over-constrain), and the `@storybook/*` family. All aligned in the current tree; the gap is that bumping the catalogued half would split them silently. It stays recorded under `Open gaps` in the description too, so it doesn't merge as an unstated hole.

---

mattinannt:
@coderabbitai review

The `Coverage Accounts For The Whole Diff` warning cites an `Open gaps` line that the description no longer contains — it was rewritten before that check reported, so please re-evaluate against the current body.

For the record, the Tailwind 4.2.4 → 4.3.1 coverage the check asks about is now stated on both pipelines this repo uses:

- **`@tailwindcss/vite`** — 270 survey-ui stories built and screenshotted on each Tailwind version with Playwright/Chromium (fixed viewport, `reducedMotion`, `animations: "disabled"`, `document.fonts.ready`, `Math.random` stubbed). All 270 PNG pairs are **byte-identical** — `cmp -s` equal, not "no visible difference".
- **`@tailwindcss/postcss`** — the pipeline that produces the published bundle — covered behaviourally by the green **Run E2E Tests** job on `526dffe`, which builds that exact UMD bundle (`pnpm build --filter=@formbricks/web...` → surveys' `copyCompiledAssetsPlugin` → `apps/web/public/js/`) and drives it in a real browser across eight survey specs.

I can't attach an image to the description from this environment, and 270 byte-identical pairs is strictly stronger evidence than one screenshot would be. The one thing that genuinely remains open is stated as such: there is no *pixel-level* assertion on the PostCSS-built bundle specifically, since E2E asserts behaviour and accessibility rather than appearance.

---

pnpm-workspace.yaml
```diff
@@ -17,6 +17,107 @@ allowBuilds:
...
+  globals: 16.5.0
+  typescript-eslint: 8.63.0
+  eslint-plugin-react-hooks: 7.1.1
+  # --- types ---
+  "@types/js-yaml": 4.0.9
+  "@types/node": 25.4.0
+  "@types/react": 19.2.14
+  "@types/react-dom": 19.2.3
+  # --- storybook ---
+  storybook: 10.3.6
```
itsjavi, line 80: do we need storybook deps in the catalog?
same question for other libs that are only used in one package or app.

I think we should only keep here what is common, and most importantly framework versions such as: next, react, react-dom (and their types), preact. to avoid issues like having different react versions.

itsjavi (changes requested):

pnpm-workspace.yaml
mattinannt: Good question, and it's the one real judgement call in this PR — so I'd rather give you the numbers than argue. **Not changing anything until you and @mattinannt decide**, since narrowing the catalog also means changing the guard, and that's a policy call rather than a cleanup.

**On "libs only used in one package or app": there are none.** The rule this PR implements is exactly yours — a dependency is catalogued *only* if two or more workspaces declare it. I re-ran the census on the current head to be sure rather than trusting the intent:

```
catalog entries: 64
catalogued with fewer than 2 consumers: (none)
```

**Storybook is the strongest case for cataloguing, not against.** Only 3 of the 9 storybook-named deps are in the catalog, and each is genuinely shared:

| entry | consumers |
| --- | --- |
| `storybook` | `apps/storybook`, `apps/web` |
| `@storybook/react` | `apps/web`, `packages/survey-ui` |
| `@storybook/react-vite` | `apps/storybook`, `apps/web`, `packages/survey-ui` |

The five single-consumer addons (`@storybook/addon-a11y`, `addon-docs`, `addon-links`, `addon-onboarding`, `@chromatic-com/storybook`) and `eslint-plugin-storybook` are all `apps/storybook`-only and deliberately left literal. Storybook's packages also hard-require matching versions across the family, so three workspaces declaring `@storybook/react-vite` independently is the same skew risk you're describing for react — just less famous.

**Your core concern is already fully covered:** `react` (3 consumers), `react-dom` (2), `@types/react` (**5**), `@types/react-dom` (3), `preact` (2) — all catalogued. A react split can't happen silently any more.

**One inconsistency worth flagging:** `next` has exactly **one** consumer (`apps/web`), so the 2+ rule excludes it. If you want framework versions pinned centrally regardless of consumer count, that's a coherent position — but it's a *different* rule from "only keep what is common", and I'd want it stated explicitly rather than have `next` sit in the catalog as the one entry nothing else references.

**The cost of narrowing.** A frameworks-only catalog would be 5 entries and would leave 66 shared deps unpinned — including all four drifts this PR exists to fix, none of which is a framework:

| drift | consumers | in a frameworks-only catalog? |
| --- | --- | --- |
| `redis` 4.7.1 vs 5.11.0 — *the latent type bug* | 2 | ✗ |
| `vite` 7 vs 8 (rollup vs rolldown) — the AC | **16** | ✗ |
| `tailwindcss` / `@tailwindcss/postcss` 4.2.4 vs 4.3.1 | 3 | ✗ |
| `eslint-plugin-react-hooks` 5 vs 7 | 2 | ✗ |

The structural half matters more than the entry count: `check-catalog.mjs`'s second rule is *"a dep declared by 2+ workspaces must be catalogued"*, and that rule is what makes this durable rather than a one-time cleanup — it already caught `@types/js-yaml` arriving from #8922 mid-review. A frameworks-only catalog means deleting that rule and replacing it with a hand-maintained allowlist, at which point the next redis-shaped split is back to review discipline.

**Three options, your call — I'll implement whichever:**

- **A — keep as-is (my recommendation).** 64 entries, mechanically derived, guard-enforced. The indirection cost is that a `package.json` no longer shows the version; the payoff is that no shared dep can drift.
- **B — middle: frameworks + toolchain only.** Keep react/preact/tailwind/vite/vitest/typescript/eslint/storybook, drop incidental shared runtime libs (`clsx`, `date-fns`, `lucide-react`, `uuid`, …). Rule 2 becomes an allowlist and stops being automatic. Catches 3 of the 4 drifts — misses `redis`, the one that was already a bug.
- **C — frameworks only,** as described above. Rule 2 goes away.

Happy to be wrong about A — if the 64 entries read as too much indirection for the codebase you maintain, that's a legitimate reason and I'll cut it back. I just don't want to trade the enforcement away by accident.

---

itsjavi: ok I was wrong then, all good.

itsjavi (approved):
after resolving some doubts, I think the PR is good

PR #8938, opened 2026-08-20
docs(agents): tell agents when an e2e test is warranted

itsjavi:
Ref ENG-2517

## What & why

Agents have been adding granular e2e specs — a breadcrumb's visibility, a progress bar's `aria-valuenow`, a Tab keypress inside one editor — and the shared context is what asked for them. Six places pointed at Playwright, and one made it the only sanctioned home for UI behaviour:

- **`AGENTS.md`** banned `.tsx` unit tests and added *"React components are covered by Playwright E2E tests **instead**"*. The second clause reads as an instruction: component changed → unit test forbidden → write a spec.
- **`.coderabbit.yaml`** asked, on any newly added `.test.tsx`, for *"the coverage to move to a Playwright `.spec.ts`"*, and licensed the same on any `.tsx` diff (*"Suggesting an E2E spec is fine"*). The review bot was requesting specs.
- Its `playwright/**` rules policed *how* a spec is written (flake, mega-specs, `@slow`) and never *whether* it should exist, so a component-level spec passed review clean.
- **`testing-methodology.mdx`** made an e2e test a per-feature `Must`, read as per-ticket, and asserted an *"85% code coverage requirement"* that nothing enforces.
- **The PR template** asked for one row per behaviour with `e2e` among the options and no cost attached, so `e2e` looked like the strongest row available.

This states the rule instead, in the four places agents actually read:

- **Level.** Prove a behaviour at the cheapest level that can fail on it. Feature happy path or a journey across surfaces → e2e; logic, invariants, anything pure → unit; a route's authorization/response shape/scoping → a test on that route.
- **Unit of coverage.** One spec per **feature area**, not per ticket and not per component — and the spec filenames in `apps/web/playwright/` are named as the inventory, so "is this area covered?" is a lookup rather than a judgement call.
- **"Neither" is an option.** A granular UI detail inside a covered feature gets manual QA and a screenshot, not a spec — with named examples so it needs no judgement.
- **Cost, where the decision is made.** An e2e test is paid on every PR, by everyone, forever; the job is the gate's critical path and its wall clock can never drop below its slowest single test.
- **The floor moves with the ceiling.** Every feature area ships a happy-path e2e, and an area with none is a gap rather than a saving — Dashboards and Workflows named, pointing at ENG-2314. Without this the change would just read as "write fewer tests".

`.coderabbit.yaml` now mirrors the same rule from the reviewer's side: a new spec file is a finding unless it opens an uncovered feature area, component-level assertions and variant matrices are findings, and UI-clicked state that Prisma or `/api/v3` could seed is a finding.

Four factual corrections fall out of it: the only enforced coverage threshold is `modules/auth/lib/**` at 80% (`apps/web/vite.config.mts`), `@slow` is metadata no project or script reads, and the mdx's Configuration section named three things that do not exist — `vite.config.ts`, `packages/lib/vitestSetup.ts`, and a stale `exclude` list — now synced to the real config.

Docs and review-bot configuration only. No spec was trimmed here and no runtime code changed.

## Breaking changes

None. No API/SDK shape, endpoint, route, env var, config key, default or webhook payload is touched — the diff is four documentation and review-configuration files.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The cost figures in the new guidance (~13 min job of which ~6 min is the Playwright step, ~30 browser-minutes, ~110 tests, wall clock floored by the slowest test) | manual | From the Playwright HTML report of run [31204161814](https://github.com/formbricks/formbricks/actions/runs/31204161814) and its job steps: e2e job 13m02s vs unit 5m04s (so it is the critical path), Playwright step 6m58s, 78 tests / 1,474 s measured, single longest test 303 s, 35% worker utilisation. Independently re-derived by a reviewer on [32417433654](https://github.com/formbricks/formbricks/actions/runs/32417433654) at this PR's base: job 12m59s, Playwright step 6m03s. `apps/web/playwright` holds 41 spec files / 109 tests at this commit |
| `@slow` described as metadata nothing reads | manual | `grep -rn "slow" playwright.config.ts playwright.service.config.ts package.json .github/workflows/e2e.yml` returns no consumer — no project, `grep`/`grepInvert`, or CI step, while 7 spec files carry the tag |
| The one enforced coverage threshold | manual | The only `thresholds` block in the repo is `apps/web/vite.config.mts:140-144` (`modules/auth/lib/**` at 80%); no other vitest config sets one, and `sonarqube.yml` scans without a quality-gate step (the gate lives in SonarQube, which is what the doc now says). Sonar's own check on this PR passes with 0.0% coverage on new code, which is the same claim from the other direction |
| Every path named anywhere in the diff resolves | manual | `playwright/utils/accessibility.ts`, `settings-tags.spec.ts`, `survey-overview.spec.ts`, `survey-accessibility.spec.ts`, `apps/web/vite.config.mts` all exist. The three that did **not** (`vite.config.ts`, `packages/lib/vitestSetup.ts`, the stale `exclude` list) were the review finding fixed in 6efefc4 |
| The level rule reads the same in all three places it appears | manual | Re-read after 6efefc4: `AGENTS.md` table + floor paragraph, the mdx "Unit of coverage" item, and the mdx New Features bullet now all say journey → extend the existing spec, logic → `.ts` unit test, UI detail → manual QA. Before that commit the three gave three different fallbacks |
| The a11y route is reachable for non-survey surfaces | manual | `survey-accessibility.spec.ts` seeds a link survey and only visits `/s/<id>`, so the unconditional "extend the axe gate" wording was unfollowable for a billing or settings page. Now scoped to the rendered survey, with the feature-area spec named for everything else |
| `.coderabbit.yaml` stays valid after both rounds of edits | manual | YAML parses, 19 `path_instructions` entries; the `path_filters`-are-exclusions guard from `coderabbit-config-validation.yml` passes locally (4 entries, all exclusions) and the check is green on the PR. Only string values changed — no schema keys added or renamed |

**Open gaps**

- **The same instructions exist outside this repo and are untouched.** The installed agent skills carry them: the E2E Tester brief's *"cover the themes, viewports, keyboard paths, roles… that the acceptance criteria **or the adjacent suite** establish"* (the ratchet that produces variant matrices), `formbricks-rules.md`'s repeat of the `.tsx` substitution plus *"treat [the axe gate] as the bar for all product UI"*, and `pr-review`'s stale *"the 25 specs"* (41 now). Those need a separate PR in the robots repo; this PR cannot reach them.
- **Nothing enforces the rule mechanically.** No CI check notices the suite growing — it went from 78 to 109 tests in about two weeks unremarked. Wiring `@slow` into `pr` / `nightly` Playwright projects and adding a duration budget check belong with ENG-2006, not here.
- **The ticket's "done when" cannot be verified yet**: it asks for a week of agent-authored PRs with no e2e specs Javier would have removed.
- **The existing suite is untouched.** The per-file inventory (which specs to trim, merge, or move to nightly — starting with `survey.spec.ts` at 32% of all browser time and the 7-variant axe matrix) is analysis, not part of this diff. Two current specs — `survey-progress-a11y.spec.ts` and `survey-keyboard.spec.ts` — are close to literal instances of the new "do not" examples; the new rules only bite on new files, so they are the trim pass's first entries rather than this PR's problem.

**Risks**

- Lowering the e2e ceiling while `.tsx` unit tests stay banned could open a real gap if "neither" is read too liberally. The counterweight is that the table's "Neither" row is scoped to *a detail inside a feature that already has a happy-path spec* — the escape hatch is unavailable exactly where coverage is absent — plus the floor sentence naming Dashboards and Workflows.
- CodeRabbit `path_instructions` are prose fed to a model, so *"a new spec file is a finding unless it opens a feature area the suite does not cover"* will occasionally fire on a legitimate new-area spec (exactly the ENG-2314 work). 6efefc4 mitigates both halves: the `.tsx` bullet now has the same new-area escape as the `.spec.ts` one, and both point at the spec-filename inventory instead of leaving the bot to guess. If it still proves noisy, narrow the bullet rather than dropping the cost framing.
- The a11y rule routes new survey-a11y work into `survey-accessibility.spec.ts`, which is itself the suite's clearest variant matrix and a trim target. Extending still beats a new file on cost, so this is a deliberate choice, not an oversight — worth revisiting when ENG-2006 picks that file up.
- The cost figures are a dated snapshot ("as of Aug 2026") and will drift as the suite changes.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -86,15 +86,20 @@ reviews:
   path_instructions:
     - path: "**/*.test.tsx"
       instructions: |
-        AGENTS.md forbids React component unit tests in this repo — components are covered by
-        Playwright specs in `apps/web/playwright/`. If this file is newly added, say so and ask for
-        the coverage to move to a Playwright `.spec.ts` instead. Do not raise this on edits to the
-        handful of `.test.tsx` files that already exist.
+        AGENTS.md forbids React component unit tests in this repo. If this file is newly added, say
+        so and ask for the logic under test to move into a `.ts` module with a unit test. Do not ask
+        for a Playwright spec instead: deleting a `.test.tsx` creates no E2E obligation, and a
+        component is not an E2E subject. Do not raise this on edits to the handful of `.test.tsx`
+        files that already exist.
 
     - path: "**/*.tsx"
       instructions: |
-        - Never suggest adding unit tests for `.tsx` files. AGENTS.md forbids them; UI behaviour is
-          covered by Playwright specs in `apps/web/playwright/`. Suggesting an E2E spec is fine.
+        - Never suggest adding unit tests for `.tsx` files — AGENTS.md forbids them, and their
+          absence is not a coverage gap to fill somewhere else. Ask for a Playwright spec only when
+          the diff changes a feature's happy path or a journey across several surfaces, and then ask
+          for it in that feature's existing spec. Never ask for one to cover component detail (a
+          label, a breadcrumb, an ARIA attribute, a keystroke inside one widget, a field's
+          validation message) — that belongs in a `.ts` unit test or in manual QA.
         - New client data flows go through an `/api/v3` route with TanStack Query, not a new Server
           Action. Server data lives in the query cache only — flag it being mirrored into
           `useState` or Jotai, and flag `router.refresh()` used as a data-refresh mechanism.
@@ -122,9 +127,24 @@ reviews:
 
     - path: "apps/web/playwright/**/*.spec.ts"
       instructions: |
-        - Flag timing hacks (`waitForTimeout`, arbitrary sleeps) and dependencies on data another
-          spec created — these are the repo's main source of flake.
-        - Specs must stay small and single-purpose; flag mega-specs. Slow suites need a `@slow` tag.
+        This suite is the critical path of every PR and every test is paid on every PR forever, so
+        cost is a review dimension here rather than a nit (AGENTS.md "Testing Guidelines"):
+        - A new spec file is a finding unless it opens a feature area the suite does not cover yet —
+          the default is assertions or a `test.step` added to that feature's existing spec. A spec
+          that asserts component-level detail (a label, a breadcrumb, a sidebar's link list, an ARIA
+          attribute, a keystroke inside one widget, a field's validation message) is a finding
+          whatever file it lands in: ask for a `.ts` unit test or manual QA instead.
+        - Flag variant matrices. A second viewport, theme, locale, role or layout needs its own
+          stated reason; "the adjacent spec does it" is not one. New a11y coverage extends
+          `survey-accessibility.spec.ts` rather than adding a per-ticket a11y spec.
+        - Flag state clicked into existence through the UI where Prisma or `/api/v3` could seed it
+          (`playwright/utils/accessibility.ts` is the pattern), and a per-test `users.create()` plus
+          login where a worker-scoped fixture would do.
+        - Flag timing hacks (`waitForTimeout`, arbitrary sleeps, `slowMo`) and dependencies on data
+          another spec created — these are the repo's main source of flake.
+        - Specs stay single-purpose per feature area: flag a spec that walks several unrelated
+          features, and a spec that exists for one detail of a covered one. `@slow` is metadata that
+          nothing reads, so a tag is not an answer to a cost finding.
 
     - path: "**/actions.ts"
       instructions: |
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -97,7 +97,9 @@ reviews:
         - Never suggest adding unit tests for `.tsx` files — AGENTS.md forbids them, and their
           absence is not a coverage gap to fill somewhere else. Ask for a Playwright spec only when
           the diff changes a feature's happy path or a journey across several surfaces, and then ask
-          for it in that feature's existing spec. Never ask for one to cover component detail (a
+          for it in that feature's existing spec — or, if the diff opens a feature area the suite does
+          not cover yet, for one happy-path spec named for that area. Never ask for one to cover
+          component detail (a
           label, a breadcrumb, an ARIA attribute, a keystroke inside one widget, a field's
           validation message) — that belongs in a `.ts` unit test or in manual QA.
         - New client data flows go through an `/api/v3` route with TanStack Query, not a new Server
@@ -129,14 +131,17 @@ reviews:
       instructions: |
         This suite is the critical path of every PR and every test is paid on every PR forever, so
         cost is a review dimension here rather than a nit (AGENTS.md "Testing Guidelines"):
-        - A new spec file is a finding unless it opens a feature area the suite does not cover yet —
-          the default is assertions or a `test.step` added to that feature's existing spec. A spec
+        - A new spec file is a finding unless it opens a feature area the suite does not cover yet
+          (the spec filenames in `apps/web/playwright/` are the inventory — check before claiming an
+          area is uncovered); the default is assertions or a `test.step` added to that feature's
+          existing spec. A spec
           that asserts component-level detail (a label, a breadcrumb, a sidebar's link list, an ARIA
           attribute, a keystroke inside one widget, a field's validation message) is a finding
           whatever file it lands in: ask for a `.ts` unit test or manual QA instead.
         - Flag variant matrices. A second viewport, theme, locale, role or layout needs its own
-          stated reason; "the adjacent spec does it" is not one. New a11y coverage extends
-          `survey-accessibility.spec.ts` rather than adding a per-ticket a11y spec.
+          stated reason; "the adjacent spec does it" is not one. A11y coverage for the rendered
+          survey extends `survey-accessibility.spec.ts`; elsewhere it belongs in that feature area's
+          own spec. Neither case is a per-ticket a11y spec.
         - Flag state clicked into existence through the UI where Prisma or `/api/v3` could seed it
           (`playwright/utils/accessibility.ts` is the pattern), and a per-test `users.create()` plus
           login where a worker-scoped fixture would do.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,8 +21,8 @@ Every `packages/*` workspace therefore exposes the standard `lint` / `typecheck`
 `test:coverage` scripts (plus `build` where there is a compile step). Deliberate exceptions:
 `config-*` packages hold only config files (no scripts beyond `clean`); `types` has no runtime logic
 to test; `email`, `types`, and `vite-plugins` are consumed from source, so they have no `build`;
-`apps/storybook` has no unit tests by policy (UI is covered by Playwright). Keep new packages on this
-matrix or document the exception here.
+`apps/storybook` has no unit tests by policy (its components are exercised by the feature journeys in
+`apps/web/playwright`). Keep new packages on this matrix or document the exception here.
 
 ### Survey Packages Build & Cache
 
@@ -171,8 +171,9 @@ Principles:
 - Prove a behavior at the cheapest level that can fail on it. An E2E test is not a stronger unit test; it
   has a different subject — the journey, not the logic.
 - **An E2E test is paid on every PR, by everyone, forever.** The Playwright job is the critical path of the
-  PR gate (as of Aug 2026: ~13 min, ~30 browser-minutes across ~110 tests), and its wall clock can never
-  drop below its slowest single test. Weigh that before adding one — sometimes the right answer is no test
+  PR gate (as of Aug 2026: a ~13 min job, of which ~6 min is the Playwright step itself — the rest is
+  install, build and boot — over ~110 tests and ~30 browser-minutes), and its wall clock can never drop
+  below its slowest single test. Weigh that before adding one — sometimes the right answer is no test
   at this level.
 
 Which level, concretely:
@@ -187,10 +188,13 @@ Which level, concretely:
 A journey across several surfaces means something like survey list → editor → public survey → response,
 where the behavior only exists once browser, survey bundle, and server are wired together.
 
+The spec filenames in `apps/web/playwright/` are the inventory of covered areas — check there before
+concluding an area has no spec.
+
 This raises a floor as well as lowering a ceiling. Every feature area ships a happy-path E2E, and an area
 with none is a gap rather than a saving (Dashboards and Workflows are the current examples — ENG-2314). A
-bug fix inside a feature that already has one almost never needs a second spec: add an assertion to the
-existing spec, or prove it in a unit test.
+bug fix inside a feature that already has one almost never needs a second spec — the level still follows
+the table above: journey behavior extends that spec, logic goes to a unit test, UI detail to manual QA.
 
 Do:
 
@@ -223,8 +227,9 @@ Do not:
   none of these justify a browser, a login, and a seeded tenant.
 - Do not build a variant matrix. Cover the one case that carries the risk; a second viewport, theme,
   locale, role, or layout needs its own stated reason, and "the adjacent spec does it" is not one.
-  Accessibility work extends the existing axe gate (`survey-accessibility.spec.ts`) instead of adding a
-  per-ticket a11y spec.
+  Accessibility work on the rendered survey extends the existing axe gate
+  (`survey-accessibility.spec.ts`); elsewhere it becomes an assertion in that feature area's own spec.
+  Either way, not a per-ticket a11y spec.
 - Do not add coverage-driven or low-signal tests.
 - Do not write tests that lock implementation details, markup, snapshots, or create churn — an assertion on
   an exact list of nav labels is churn, not coverage.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -168,33 +168,69 @@ Always mark React component props as `Readonly<>` (e.g., `({ children }: Readonl
 Principles:
 
 - Confidence over coverage. Test behavior and outcomes; avoid brittle implementation-detail tests.
+- Prove a behavior at the cheapest level that can fail on it. An E2E test is not a stronger unit test; it
+  has a different subject — the journey, not the logic.
+- **An E2E test is paid on every PR, by everyone, forever.** The Playwright job is the critical path of the
+  PR gate (as of Aug 2026: ~13 min, ~30 browser-minutes across ~110 tests), and its wall clock can never
+  drop below its slowest single test. Weigh that before adding one — sometimes the right answer is no test
+  at this level.
+
+Which level, concretely:
+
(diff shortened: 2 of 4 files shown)


AGENTS.md
```diff
@@ -168,33 +168,69 @@ Always mark React component props as `Readonly<>` (e.g., `({ children }: Readonl
...
+  test instead**: the absence of a component unit test creates no coverage obligation. If a component holds
+  logic worth proving, lift that logic into a `.ts` module and unit-test it there; the rendering is
+  exercised incidentally by the feature journeys that already cross it.
+- Do not E2E a component. A language selector, a breadcrumb, a sidebar's link list, an ARIA attribute on
+  one widget, a keystroke inside one editor, a field's validation message, a search box filtering a list —
+  none of these justify a browser, a login, and a seeded tenant.
+- Do not build a variant matrix. Cover the one case that carries the risk; a second viewport, theme,
+  locale, role, or layout needs its own stated reason, and "the adjacent spec does it" is not one.
+  Accessibility work extends the existing axe gate (`survey-accessibility.spec.ts`) instead of adding a
+  per-ticket a11y spec.
```
kiln-agent: The a11y instruction is stated unconditionally, but the spec it names only covers one surface, so for a11y work anywhere else it points agents at a spec that physically cannot reach their page — and it contradicts the "one spec per feature area" rule two paragraphs up.

`survey-accessibility.spec.ts:17` describes itself as the *"Accessibility gate for the rendered link survey (axe-core)"*: it seeds a published link survey through Prisma and `goto`s `/s/<id>` (`playwright/utils/accessibility.ts`). An agent adding an a11y assertion for a billing page, a settings surface, or an admin screen has nowhere to put it under this sentence — the axe gate never navigates to those pages, and "instead of adding a per-ticket a11y spec" reads as closing the other door too. Meanwhile the feature-area rule would say the assertion belongs in that area's own spec.

Suggest scoping it to the surface the gate actually covers, and naming the fallback:

```suggestion
  locale, role, or layout needs its own stated reason, and "the adjacent spec does it" is not one.
  Accessibility work on the rendered survey extends the existing axe gate
  (`survey-accessibility.spec.ts`); elsewhere it becomes an assertion in that feature area's own
  spec. Either way, not a per-ticket a11y spec.
```

The same narrowing applies to the mirrored bullet in `.coderabbit.yaml:138-139` ("New a11y coverage extends `survey-accessibility.spec.ts` rather than adding a per-ticket a11y spec") — otherwise the bot will push non-survey a11y work into the survey gate.

Worth noting for context, not as a change request here: the spec this sentence blesses is also the suite's clearest variant matrix (desktop / mobile / tablet full walks, plus the RTL variant), which the bullet directly above now forbids. The PR is explicit that the existing suite is out of scope, so that's a note for the ENG-2314 / trim pass rather than this diff.

---

.coderabbit.yaml
```diff
@@ -86,15 +86,20 @@ reviews:
...
+        files that already exist.
 
     - path: "**/*.tsx"
       instructions: |
-        - Never suggest adding unit tests for `.tsx` files. AGENTS.md forbids them; UI behaviour is
-          covered by Playwright specs in `apps/web/playwright/`. Suggesting an E2E spec is fine.
+        - Never suggest adding unit tests for `.tsx` files — AGENTS.md forbids them, and their
+          absence is not a coverage gap to fill somewhere else. Ask for a Playwright spec only when
+          the diff changes a feature's happy path or a journey across several surfaces, and then ask
+          for it in that feature's existing spec. Never ask for one to cover component detail (a
```
kiln-agent: This bullet is missing the new-feature-area escape hatch its sibling has, so on the one case the PR body flags as a known risk the bot gives advice that cannot be followed.

The `.spec.ts` instruction below gets this right — *"A new spec file is a finding **unless it opens a feature area the suite does not cover yet**"* (line 132). Here the ask is unconditional: *"then ask for it in that feature's existing spec."* A `.tsx` diff that introduces a new feature area — Dashboards or Workflows, exactly the ENG-2314 work the PR names — has no existing spec, so the reviewer bot will point at a file that does not exist rather than asking for the area's first happy-path spec, which AGENTS.md's table explicitly wants.

```suggestion
        - Never suggest adding unit tests for `.tsx` files — AGENTS.md forbids them, and their
          absence is not a coverage gap to fill somewhere else. Ask for a Playwright spec only when
          the diff changes a feature's happy path or a journey across several surfaces, and then ask
          for it in that feature's existing spec — or, if the diff opens a feature area the suite
          does not cover yet, for one happy-path spec named for that area. Never ask for one to
          cover component detail (a
```

---

kiln-agent (commented):
Reviewed the diff and re-derived every number in it against this commit. The change does what it says: it replaces a substitution instruction ("component changed → no unit test → write a spec") with a level-selection rule, and it puts the rule in the four places that actually drive behaviour. The floor sentence (every feature area ships a happy-path E2E, Dashboards and Workflows named) is doing the load-bearing work — without it this would read as "write fewer tests," and I think it holds. Two small findings inline, plus one consistency point below.

**Claims I verified**

| Claim | Result |
| --- | --- |
| Exactly one enforced Vitest threshold, `modules/auth/lib/**` at 80% | ✅ `apps/web/vite.config.mts:140` is the only `thresholds` block in the repo |
| Coverage is otherwise reported, gate lives in SonarQube | ✅ reporters are `["text","html","lcov"]`; `sonarqube.yml` scans with no quality-gate step |
| `@slow` is metadata nothing reads | ✅ `slow` appears in neither `playwright.config.ts`, `playwright.service.config.ts`, nor any workflow — while 7 specs carry the tag |
| "~110 tests" | ✅ 109 top-level `test(` across 41 spec files |
| Every exemplar path resolves | ✅ `playwright/utils/accessibility.ts`, `settings-tags.spec.ts`, `survey-overview.spec.ts`, `survey-accessibility.spec.ts` |
| `playwright/utils/accessibility.ts` is the Prisma-seeding pattern | ✅ seeds through Prisma and reuses `transformQuestionsToBlocks`, no login or API key |
| `.coderabbit.yaml` still valid, `path_filters` all exclusions | ✅ parses; 4 filters, all `!`-prefixed; the config-validation check is green |
| Formatting | ✅ `prettier --check` clean on all four files |

No contradicting copy left elsewhere in the repo either: no other file claims an 85% threshold, and `technical-handbook/overview.mdx` only lists Playwright as the E2E tool. The new PR-template note also lines up with the existing `Coverage accounts for the whole diff` pre-merge check, which already passes a `manual` row with no test named — so the template change doesn't strand anyone against the bot.

**One consistency point, outside the diff**

`AGENTS.md:24` still reads *"`apps/storybook` has no unit tests by policy (UI is covered by Playwright)."* That parenthetical is the exact substitution phrasing this PR purges from the Testing Guidelines 140 lines further down, and it's the first thing an agent reading the file top-to-bottom meets. The sentence's real job is explaining why the storybook workspace has no `test` script, which doesn't need the claim — `(UI is exercised by the feature journeys in apps/web/playwright)`, or just dropping the parenthetical, would keep the intent without re-arming the inference.

**On the risk you flagged**

> `path_instructions` are prose fed to a model, so "a new spec file is a finding unless it opens a feature area the suite does not cover yet" will occasionally fire on a legitimate new-area spec.

Agreed that narrowing the bullet is the right response if it gets noisy, rather than dropping the cost framing. The `.tsx` bullet has a sharper version of this problem today — see the inline comment; it's a wording fix, not a design change.

Nothing blocking. Docs and review-config only, CI green apart from jobs still running.

---

kiln-agent (commented):
## Review

Docs and review-bot configuration only, so the review is about whether the added prose is *true* and whether it contradicts anything left standing. I re-derived every factual claim from the repo and from CI rather than taking the Coverage table on trust. **All of them hold** — including the cost figures, which I initially thought were stale and were not. Three soft findings below, none blocking.

### Claims I verified independently

| Claim | Verdict |
| --- | --- |
| `modules/auth/lib/**` at 80% is the only enforced Vitest threshold | ✅ `apps/web/vite.config.mts:140-144` is the sole `thresholds` block in the repo |
| `@slow` is metadata nothing reads | ✅ No hit in `playwright.config.ts`, `playwright.service.config.ts`, `package.json`, or `e2e.yml`. The config has one project, `testMatch: "**/*.spec.ts"`, no `grep`/`grepInvert`. 7 spec files carry the tag; nothing consumes it |
| v8 provider, lcov emitted, gate lives in SonarQube | ✅ `vite.config.mts:42-43`; `sonar-project.properties` sets `sonar.javascript.lcov.reportPaths`; `sonarqube.yml` has no quality-gate step |
| All four exemplar paths resolve | ✅ `playwright/utils/accessibility.ts`, `settings-tags.spec.ts`, `survey-overview.spec.ts`, `survey-accessibility.spec.ts` |
| 41 spec files / ~110 tests | ✅ 41 files, 109 `test(` blocks at `a09ea54` |
| `.coderabbit.yaml` still valid | ✅ Parses; 19 `path_instructions` entries, all `{path, instructions}`; `path_filters` = 4 entries, all exclusions, so the `coderabbit-config-validation.yml` guard passes |
| No orphaned references to the removed "Heuristic" block | ✅ Nothing in the repo references the old wording; `CONTRIBUTING.md`, `.cursor/`, `.codex/`, `ACCESSIBILITY.md` carry no competing e2e guidance |
| `prettier --check` on all four changed files | ✅ Clean (`.mdx` and the PR template are `.prettierignore`d anyway) |

**On the cost figures specifically.** The measured run cited in the description ([31204161814](https://github.com/formbricks/formbricks/actions/runs/31204161814)) is from Aug 7 at 78 tests, and the suite is at 109 today — so I expected "~13 min" to understate the current job by ~40% and was ready to flag it. It does not. On the most recent merge-queue run at this PR's base commit ([32417433654](https://github.com/formbricks/formbricks/actions/runs/32417433654), `9f186e2`), the E2E job ran **12m59s** against Unit at 5m12s, Linters 7m27s→job end, Typecheck 7m06s. It is unambiguously the critical path and "~13 min" is accurate for the current suite. Withdrawn.

One refinement if you want the number to carry more weight: of that 12m59s, only **6m03s** is the Playwright step itself (21:09:30→21:15:33); the rest is container init, install, a 3m35s build, migrations and app boot. The sentence as written is correct — it says the *job* is ~13 min — but a reader optimising against it might assume trimming specs recovers 13 minutes when the recoverable envelope is about half that. Worth a parenthetical, entirely optional.

### 1. `AGENTS.md:24` still carries the substitution the PR removes — nit

```
`apps/storybook` has no unit tests by policy (UI is covered by Playwright).
```

This is the exact clause the description identifies as the root cause, and it now sits ~165 lines *above* the corrected guidance. In context it is only justifying why `apps/storybook` has no `test` script, so it is much weaker than the line you removed at 190 — but an agent scanning `AGENTS.md` for "what covers UI" hits this one first. A one-clause fix keeps the packaging point intact: `(UI is exercised by the Playwright journeys that cross it — see "Testing Guidelines")`.

### 2. "feature area" is load-bearing but never defined — suggestion

It is the unit the whole rule turns on, in both directions:

- `AGENTS.md:197` — "one spec per **feature area**"; `:190` — "an area with none is a gap rather than a saving"
- `.coderabbit.yaml` — "A new spec file is a finding **unless it opens a feature area the suite does not cover yet**"

Neither says how to determine whether an area is already covered, so both the agent and the bot have to guess — and the description already flags the bot false-positive as a live risk. The answer in practice is discoverable but implicit: the spec filenames in `apps/web/playwright/` *are* the inventory. One sentence saying so turns a judgement call into a lookup, and it is the cheapest available mitigation for the risk you named.

### 3. The a11y rule points new work at the spec the description calls the problem — note, not a defect

Within the same `.coderabbit.yaml` block:

> Flag variant matrices. A second viewport, theme, locale, role or layout needs its own stated reason […] New a11y coverage extends `survey-accessibility.spec.ts` rather than adding a per-ticket a11y spec.

`survey-accessibility.spec.ts` is the 7-variant axe matrix the description names under Open gaps as a trim target. So the rule routes new a11y work into the file it also wants smaller. That is still the right default — extending beats a new file on cost — and the inventory is explicitly out of scope here, so this is not a change request. Flagging it so the tension is a conscious choice rather than a surprise when ENG-2006 picks the file up.

### On the stated risk

The "read too liberally" risk is real, and the floor sentence at `:190` is doing a lot of work alone. I think it holds, for a reason worth making explicit: the table's fourth row scopes "Neither" to *"a UI detail inside a feature that already has a happy-path spec"* — the escape hatch is unavailable precisely where coverage is absent, which is the case that would actually hurt. The `Do not E2E a component` list then makes the boundary concrete enough not to need judgement. I would not weaken it further, and I would not strengthen it either.

Worth noting for whoever picks up the inventory: `survey-progress-a11y.spec.ts` and `survey-keyboard.spec.ts` are close to literal instances of the new "do not" examples ("an ARIA attribute on one widget", "a keystroke inside one editor"). Correctly out of scope here — the new rules only bite on new files — but they are the first entries on that list.

Nothing here blocks. Leaving the approve to @mattinannt.

---

itsjavi:
All four findings landed in 6efefc4; description updated to match. Both inline threads resolved.

**Taken as suggested** — the a11y route is now scoped to the surface the gate can actually reach (rendered survey → the axe gate; anywhere else → that feature area's own spec), and the `.tsx` bullet got the same new-feature-area escape its `.spec.ts` sibling already had. Both were unfollowable as written: one pointed at a spec that never navigates to the page, the other at a spec that doesn't exist yet.

**"Feature area" is now a lookup, not a judgement call** — added to both `AGENTS.md` and the `.spec.ts` instruction: the spec filenames in `apps/web/playwright/` are the inventory, check there before concluding an area is uncovered. That was the cheapest available mitigation for the bot false-positive risk, and it's a better answer than the one I'd written into Risks.

**Cost figure sharpened** — you're right that "~13 min job" invites the wrong optimisation. It now reads: a ~13 min job, of which ~6 min is the Playwright step itself, the rest install/build/boot. So trimming specs recovers about half the job, not all of it.

**`AGENTS.md:24`** — fixed. I'd deliberately left it to keep the diff tight; "first thing an agent meets reading top-to-bottom" is the better argument.

**The a11y tension is a conscious choice, not an oversight.** The rule does route new survey-a11y work into the file that is itself the suite's clearest variant matrix and a named trim target. Extending still beats a new file on cost today, so I'd rather carry the tension than invert the default — and it's now recorded under Risks so ENG-2006 meets it deliberately. Same for `survey-progress-a11y.spec.ts` and `survey-keyboard.spec.ts` being near-literal instances of the new "do not" examples: the rules only bite on new files, so they're the trim pass's first entries rather than something this diff should touch.

One note on CI: the `PR Check Summary` failure showing against `a09ea54` is my own push superseding that run — every job on it passed except E2E, which was `cancelled` mid-step at 21:42:32 when `6efefc4` landed (`pr.yml` runs with `cancel-in-progress: true`, and the summary job treats `cancelled` like `failure`). The run against the current head is what counts.

---

Dhruwang (approved):

PR #8939, opened 2026-08-21
feat(auth): ask how users found Formbricks on sign-up

harshsbhat:
No ticket — ad-hoc request to capture sign-up attribution beyond UTM params, no Linear ticket exists for this.

## What & why

We had no self-reported signal for how a new user found Formbricks, only URL/cookie-based UTM attribution (`apps/web/modules/auth/lib/attribution.ts`). This adds an optional "How did you find out about Formbricks?" dropdown to the sign-up form (Blog post, AI tool, "Powered by Formbricks" badge, Saw a survey somewhere, Search engine, Social media, Friend or colleague, Conference or event, Other), with a conditional follow-up text field for Blog ("Which article or page?"), AI tool ("What did you ask it?"), and Other ("Please specify").

Both fields are entirely optional — nothing blocks submission, and leaving them blank is a first-class case. The answer is not persisted anywhere in the database; it's forwarded only as `discovery_source` / `discovery_source_detail` properties on the existing `user_signed_up` PostHog event, alongside the current UTM attribution, so no migration is needed.

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Dropdown renders on sign-up form, both fields optional, submits fine with none selected | manual | Verified in browser — signup form renders correctly, submission succeeds with the field left blank |
| Follow-up text field only appears for Blog / AI tool / Other, with the right label + placeholder per option | manual | Verified by selecting each option in the browser |
| `discoverySource` / `discoverySourceDetail` forwarded onto the `user_signed_up` PostHog event when provided | `unit (mutation)` — `apps/web/modules/auth/signup/actions.test.ts` "forwards discoverySource/discoverySourceDetail into the user_signed_up PostHog event" | Pass |
| Both properties report `null` (not omitted/undefined) when the user leaves the question blank | `unit (mutation)` — same file, "reports discoverySource/discoverySourceDetail as null when the user leaves them blank" | Pass |
| `discoverySourceDetail` is trimmed and dropped when `discoverySource` is absent or doesn't take a follow-up (e.g. a direct POST to the unauthenticated action bypassing the form's own clearing logic) | `unit (mutation)` — `apps/web/modules/auth/lib/discovery-source.test.ts`, 5 cases: trims + keeps for a follow-up source, drops whitespace-only, drops when no source, drops when the source doesn't take a follow-up, returns undefined when no detail given | Pass |
| `discoverySourceDetail`'s max-length check runs on the trimmed value, not the raw one — whitespace padding can't push valid content over the limit and get rejected before it's ever trimmed | `unit (red on main)` — same file, `ZDiscoverySourceDetail` schema tests: content padded to just over the limit is accepted after trimming; content still too long after trimming is rejected | Pass |
| Existing sign-up behaviour (invite flow, domain blocking, duplicate handling, etc.) unaffected | `unit (guard)` — full existing `actions.test.ts` suite (24 tests) plus all 40 auth test files (408 tests) | All pass, no changes to existing assertions |
| Translations complete across all locales | `pnpm i18n` (Lingo.dev regeneration + validation) | 0 missing keys, 0 unused keys, all 23 locales complete |
| Typecheck / lint / production build | `pnpm turbo run typecheck/lint --filter=@formbricks/web`, `pnpm build --filter=@formbricks/web` | All pass, 0 errors (pre-existing warnings only, unrelated to this change) |

**Open gaps**

- No E2E/Playwright coverage added for the new dropdown — this is a low-risk, purely additive UI field with no gating logic, so it didn't seem to warrant a new spec per the "confidence over coverage" testing principle. Happy to add one if reviewers want it.

**Risks**

- None expected — the new fields are optional and additive to both the form schema and the server action's Zod schema; nothing existing changes shape.

---

> [!NOTE]
> **AI model used** — Claude Sonnet 5, reasoning effort medium.

Screenshot 2026-08-21 at 10 03 43 AM


apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
+                              htmlFor="discovery-source"
+                              className="mb-1.5 block text-xs font-medium text-slate-500">
+                              {t("auth.signup.discovery_source_label")}
+                            </label>
+                            <Select
+                              value={field.value}
+                              onValueChange={(value: TDiscoverySource) => {
+                                field.onChange(value);
+                                if (!DISCOVERY_SOURCES_WITH_FOLLOWUP.has(value)) {
+                                  form.setValue("discoverySourceDetail", "");
```
kiln-agent, line 370: `setValue` without `{ shouldValidate: true }` leaves `formState.isValid` stale, and the submit button is gated on it — the user can end up permanently unable to submit.

Repro: select "AI tool", paste >256 chars into "What did you ask it?" (`ZSignupInput` caps it at `MAX_DISCOVERY_SOURCE_DETAIL_LENGTH`), then switch the dropdown to e.g. "Conference or event". Sequence in RHF 7.71.2:

1. `field.onChange(value)` runs the registered `onChange`; with the default `mode: "onSubmit"` it takes the `shouldSkipValidation` path and calls `_setValid()`, which invokes the zod resolver **synchronously** on `_formValues` (verified: `zodResolver` calls `parseAsync(schema, values)` synchronously and zod 4.3.6 reads the object synchronously) — so it snapshots the still-too-long value → `isValid: false`.
2. `form.setValue("discoverySourceDetail", "")` then clears the value but never calls `_setValid` (RHF's `setValue` only re-validates when `shouldValidate` is passed).

Result: the detail field is now unmounted, its value is empty, yet `formState.isValid` stays `false`, so `disabled={showLogin && !form.formState.isValid}` keeps "Continue with email" greyed out with no error anywhere on screen. The only way out is to re-pick a follow-up source and type into the field again.

```suggestion
                                  form.setValue("discoverySourceDetail", "", { shouldValidate: true });
```


---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
+                              <label
+                                htmlFor="discovery-source-detail"
+                                className="mb-1.5 block text-xs font-medium text-slate-500">
+                                {discoverySource === "blog"
+                                  ? t("auth.signup.discovery_source_blog_followup_label")
+                                  : discoverySource === "llm"
+                                    ? t("auth.signup.discovery_source_llm_followup_label")
+                                    : t("auth.signup.discovery_source_other_followup_label")}
+                              </label>
+                              <Input
```
kiln-agent, line 423: This input has no `maxLength` and its `FormField` renders no `FormError`, but the value is length-capped in `ZSignupInput` and the submit button is gated on `form.formState.isValid` — so an over-long entry silently blocks the whole sign-up.

Concrete case: a user picks "AI tool" and pastes the 300-character prompt they actually used into "What did you ask it?". zod fails `max(256)`, `isValid` flips to `false`, and "Continue with email" greys out for the *entire* form (name/email/password all valid) with no message rendered under the field and no visible reason. Contrast the name/email/password fields above, which all render `{error?.message && <FormError …>}`.

Either render the field error like the other three fields, or cap the input so the invalid state is unreachable:

```suggestion
                              <Input
                                id="discovery-source-detail"
                                data-testid="signup-discovery-source-detail"
                                value={field.value}
                                name="discoverySourceDetail"
                                maxLength={MAX_DISCOVERY_SOURCE_DETAIL_LENGTH}
                                onChange={(e) => field.onChange(e.target.value)}
```


---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
                   />
                 </div>
                 <PasswordChecks password={form.watch("password")} />
+                <div className="mt-4 space-y-2 border-t border-slate-100 pt-4">
```
kiln-agent, line 350: This block renders on every deployment, but the answer only goes anywhere when PostHog is configured — `capturePostHogEvent` starts with `if (!posthogServerClient) return;` (`apps/web/lib/posthog/capture.ts`), which is the normal state for self-hosted instances. So a self-hoster's users are asked "How did you find out about Formbricks?" on their own sign-up screen (in 15 locales) and the answer is dropped on the floor; nothing is persisted, since the PR deliberately writes no DB column.

`isFormbricksCloud` is already a prop on this component and is used a few lines below to switch between the product-updates and security-updates checkbox, so gating this `<div>` on it would match the existing pattern for cloud-only marketing UI.

---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
+                                  form.setValue("discoverySourceDetail", "");
+                                }
+                              }}>
+                              <SelectTrigger
+                                id="discovery-source"
+                                data-testid="signup-discovery-source"
+                                className="bg-white font-normal text-slate-700">
+                                <SelectValue placeholder={t("auth.signup.discovery_source_placeholder")} />
+                              </SelectTrigger>
+                              <SelectContent>
```
kiln-agent, line 379: The nine option values are hand-written here, duplicating `DISCOVERY_SOURCES`, and the two lists can only be kept in sync by eye. They agree today, but the failure mode of drift is worse than lost analytics: `discoverySource` is part of the same `ZCreateUserAction` schema as name/email/password, so a value that isn't in the enum makes the *entire* action fail input validation. The user then gets `getFormattedErrorMessage` output in a toast — a raw, untranslated `discoverySource: Invalid option …` — and no account is created.

Adding a source to `DISCOVERY_SOURCES` also silently does nothing here, since the list of `SelectItem`s is what the UI renders.

Deriving the options from the constant removes both failure modes, e.g. `DISCOVERY_SOURCES.map((source) => <SelectItem key={source} value={source}>{t(`auth.signup.discovery_source_${source}`)}</SelectItem>)` — the label keys already follow exactly that pattern, and `scan-translations` picks up the template-literal form (`/\bt\s*\(\s*\`(?<temp1>[^\`]+)\`/g` in `packages/i18n-utils/src/scan-translations.ts`), though you'd want to confirm the interpolated key doesn't get reported as unused.

---

apps/web/modules/auth/signup/actions.test.ts
```diff
@@ -318,6 +318,39 @@ describe("createUserAction — signup verification email callbackURL", () => {
 
   // The catch branch is still live: flipping EMAIL_VERIFICATION_DISABLED / autoSignIn makes Better Auth
   // throw USER_ALREADY_EXISTS instead of answering synthetically, so both signals must classify.
+  // Guard: the sign-up form's optional "how did you find out about Formbricks" answer must reach
+  // the analytics event so it can be used for attribution — but never anything else (no DB write).
+  test("forwards discoverySource/discoverySourceDetail into the user_signed_up PostHog event", async () => {
```
kiln-agent, line 323: These two tests were inserted between the pre-existing comment on line 319 ("The catch branch is still live: flipping EMAIL_VERIFICATION_DISABLED / autoSignIn makes Better Auth throw USER_ALREADY_EXISTS…") and the test it documents, `treats a thrown duplicate as already-existed too` (now line 354). As it reads now, that comment looks like documentation for the discovery-source test, and the duplicate test has lost its rationale. Moving the new block below `treats a thrown duplicate as already-existed too` (or above line 319) keeps both comments attached to the right test.

Also worth noting for the record: `actionClient.inputSchema` is mocked to `mockReturnThis()` in this file, so `parsedInput` bypasses `ZCreateUserAction` — these tests prove the forwarding into the PostHog payload, but not that the new zod fields accept the values the form sends.

---

kiln-agent (commented):
Reviewed the diff for correctness. The server-side wiring (optional zod fields → `user_signed_up` properties, `?? null` for blanks, nothing persisted) looks right, and the i18n side checks out: all 15 locales carry the 17 new keys with no drift, and every new `i18n.lock` checksum matches `objectHash.MD5` of the en-US value (the two removed keys, `common.coming_soon` and `…api_ingestion_setup_description`, are already absent from en-US, so that's stale-key cleanup, not a regression).

Five findings, all in the new form UI plus one test-placement nit. The first two are the substantive ones: because the follow-up field participates in `formState.isValid` and the submit button is `disabled` on it, an over-long entry silently blocks the whole sign-up, and one of the two paths leaves the form permanently unsubmittable.

---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
+                            <label
+                              htmlFor="discovery-source"
+                              className="mb-1.5 block text-xs font-medium text-slate-500">
+                              {t("auth.signup.discovery_source_label")}
+                            </label>
+                            <Select
+                              value={field.value}
+                              onValueChange={(value: TDiscoverySource) => {
+                                field.onChange(value);
+                                if (!DISCOVERY_SOURCES_WITH_FOLLOWUP.has(value)) {
```
kiln-agent: The reset only fires when the *new* source has no follow-up, so an answer typed for one follow-up question survives into a different one.

Repro: pick "Blog post", type `formbricks.com/blog/open-source-surveys`, then change your mind and pick "AI tool". `DISCOVERY_SOURCES_WITH_FOLLOWUP.has("llm")` is `true`, so the reset is skipped and the blog URL stays in the field — now labelled "What did you ask it?". Submitting reports `discovery_source: "llm"` with `discovery_source_detail` = a blog URL, i.e. an answer to a question the user never answered, in the very dataset this PR exists to produce. Same in the other direction (llm → other, blog → other, …).

Clearing on every source change is both simpler and correct — the field is empty exactly when it appears:

```suggestion
                                if (value !== field.value) {
```
(and drop the `DISCOVERY_SOURCES_WITH_FOLLOWUP` check here, keeping `form.setValue("discoverySourceDetail", "", { shouldValidate: true })` unconditional).

---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
                 <PasswordChecks password={form.watch("password")} />
+                <div className="mt-4 space-y-2 border-t border-slate-100 pt-4">
+                  <FormField
+                    control={form.control}
+                    name="discoverySource"
+                    render={({ field }) => (
+                      <FormItem className="w-full">
+                        <FormControl>
+                          <div className="text-left">
+                            <label
```
kiln-agent, line 358: `<label htmlFor>` does not name a Radix `SelectTrigger`. The trigger renders a `<button role="combobox">`, and per HTML-AAM a `button`'s accessible name is computed from its *subtree contents* — an associated `label` element is not part of that computation (unlike `input`/`select`).

Effect: a screen-reader user tabbing here hears only "Select an option, combobox" — the placeholder — and never hears "How did you find out about Formbricks?", so the question is unanswerable non-visually. The visible text is fine; it just isn't wired to the control.

Point the trigger at the label instead (keeps the visual markup as-is):

```suggestion
                            <label
                              id="discovery-source-label"
                              htmlFor="discovery-source"
                              className="mb-1.5 block text-xs font-medium text-slate-500">
```
…and add `aria-labelledby="discovery-source-label"` to the `SelectTrigger` below (or, if you'd rather not add the id, `aria-label={t("auth.signup.discovery_source_label")}` on the trigger).

---

apps/web/modules/auth/signup/components/signup-form.tsx
```diff
@@ -294,6 +320,103 @@ export const SignupForm = ({
...
+                              htmlFor="discovery-source"
+                              className="mb-1.5 block text-xs font-medium text-slate-500">
+                              {t("auth.signup.discovery_source_label")}
+                            </label>
+                            <Select
+                              value={field.value}
```
kiln-agent, line 364: `field.value` is `undefined` until the user picks something (`defaultValues.discoverySource: undefined`), so this `Select` starts **uncontrolled** and flips to **controlled** on the first selection.

`@radix-ui/react-use-controllable-state` recomputes `isControlled = prop !== undefined` on every render and warns on exactly this transition (`dist/index.mjs:16-29`): the first pick logs `Select is changing from uncontrolled to controlled` in the console, and until then Radix's internal state — not RHF — is the source of truth for the value.

`?? ""` keeps it controlled from the first render, and Radix still shows the placeholder for an empty string:

```suggestion
                              value={field.value ?? ""}
```

---

kiln-agent (commented):
Three additional findings on the new dropdown, on top of the ones already posted (submit-blocking `maxLength`/`FormError` gap, stale `isValid` after `setValue`, hand-duplicated option list, test placement): the follow-up answer carries over between two follow-up sources, the question text is not the select's accessible name, and the `Select` flips from uncontrolled to controlled on the first pick.

For the record, two things I checked that are **not** problems: the server wiring (optional zod fields → `discovery_source`/`discovery_source_detail` on `user_signed_up`, `?? null` for blanks, nothing persisted) is correct and `pnpm scan-translations` is clean across all locales; and the earlier bot comment about `onValueChange: (value: TDiscoverySource)` failing under `strict` is a false positive — Radix declares it as a method shorthand (`onValueChange?(value: string): void`, `dist/index.d.ts:26`), so parameter checking there is bivariant and it type-checks.

---

PR #8940, opened 2026-08-21
perf(ci): halve the e2e test step and drop the dead docker build cache

Dhruwang:
Ref ENG-2006

## What & why

E2E and Docker Build Validation are the two jobs that decide how long a PR takes; every other check finishes in under 6 min. This PR takes a measured chunk out of each: **E2E 12m43s → 10m30s** (its test step 598s → 288s) and **Docker Build Validation 12–17m → 8.7m**.

**1. `survey.spec.ts` paced every action with `launchOptions.slowMo: 150`**, with a comment saying it was load-bearing and that removing it needed hardened waits first (the leftover of ENG-2006).

- E2E median was **12m43s**, worst **16m29s** over 14 runs; next-longest job 5m.
- Of its 598s test step in run [32447746140](https://github.com/formbricks/formbricks/actions/runs/32447746140), `survey.spec.ts` was **876s of the suite's 1799s CPU-seconds in 3 tests** — and with 10 Azure workers the wall clock is bounded by the longest single test, not the total.
- The trace of the retried test shows 726 actions / 171s, ~110s of it slowMo sleep, plus 25 `pressSequentially` calls averaging 2756ms because `delay: 50` and slowMo both apply per keystroke.

Removing the sleep meant fixing the two editor behaviours it was hiding:

- **Text has to be entered in one operation.** `pressSequentially` at full speed races Lexical's re-render and truncates (`"Picture Select Question"` arriving as `"Picture S"`). `fill()` replaces the value with a single `insertText`; verified it commits by saving the draft and reloading.
- **Structural clicks drop in-flight edits.** `handleAddLabel` (matrix), `updateChoice` (multiple-choice), the element add, and the "Add description" / "Add “Other”" buttons each rebuild their array from the `element`/`survey` prop of the render they were created in. A click landing before the previous edit committed writes a stale array back, silently reverting it — surfacing as `Choice 1 in question 1 of block 2 is missing` blocking Publish. Adds are now separated from fills, the fills converge against the **live preview** (the only surface reflecting committed state — the label inputs keep their own), and the ~40 structural seams pace themselves instead of all 726 actions.

Two cleanups fell out. The add-block trigger was a Radix `asChild` `div`, so specs matched it positionally via `locator("div").filter({hasText}).nth(1)` — a match that re-resolves to a different node on every editor re-render. Review surfaced that the `div` was also an **accessibility bug**: `asChild` forwards the click handler and aria state but adds no role and no `tabIndex`, so the control was unreachable by keyboard and a keyboard-only user could not add a block at all. Dropping `asChild` lets Radix render its own `<button>`, which fixes that and gives the specs a real button; the `data-testid` stays because the live preview renders a second copy of the block text. And `publishSurvey` reports the editor's auto-dismissing `react-hot-toast` validation message instead of failing as an opaque `waitForURL` timeout — which is how the stale-write symptom was identified at all.

**2. `docker-build-validation.yml` exported a BuildKit layer cache that costs far more than it restores.** Decomposing one 10.4 min run ([32452739610](https://github.com/formbricks/formbricks/actions/runs/32452739610)):

| BuildKit step | Time |
| --- | --- |
| `pnpm install` | 34s |
| build `@formbricks/database` | 33s |
| build `@formbricks/web...` | 239s |
| exporting to docker image format | 45s |
| importing to docker (`load: true`) | 28s |
| **exporting to GitHub Actions Cache** | **222s** |
| base image, apk, ~106 runner layers | ~30s |

That export was 222s here, 286s in [32449834984](https://github.com/formbricks/formbricks/actions/runs/32449834984) and 407s in [32453148066](https://github.com/formbricks/formbricks/actions/runs/32453148066). What the cache can restore is **12s** of work: three of four runs reported `CACHED` on nothing, and the one that hit (8 steps) hit only the apk/corepack prelude, measured at 12s cold. Two independent reasons it cannot do better as written, both recorded in a comment at the step:

- `COPY . .` precedes `pnpm install` in `apps/web/Dockerfile`, so any diff invalidates every layer after it — only the prelude is ever reusable.
- GHA caches are branch-scoped. A PR may read its own scope and `main`'s, but no workflow runs on pushes to `main`, and `merge_group` runs write to throwaway `gh-readonly-queue/...` scopes, so the readable scope is never populated.

Making it pay needs both halves — reorder the Dockerfile so the install layer survives a source change, *and* give the cache a producer PRs can read (a push-to-main build, or a registry cache, which isn't branch-scoped). Until then, paying 222–407s to save 12s is worse than not caching. No other workflow uses `type=gha`, so nothing else is affected.

**The two jobs on a PR's critical path, before and after:**

| Job | Before | After |
| --- | --- | --- |
| `Run E2E Tests` | Run E2E Tests — Successful in 12m | Run E2E Tests — Successful in 10m |
| `Validate Docker Build` | Validate Docker Build — Successful in 18m | Validate Docker Build — Successful in 8m |

The two run in parallel, so a PR waits on the slower of them: 18m before, 10m after.

## Breaking changes

None — test-side changes, two `data-testid` attributes, a keyboard-reachability fix on the add-block trigger (a `div` becomes the `<button>` Radix renders by default, same classes and layout), and one workflow step's caching options. No API, SDK or config surface changes; the image plus every validation run against it are unchanged.

## Migrations & env

- none

## How this was tested

E2E was verified locally against the full stack (Postgres, Valkey, RustFS, `pnpm build` + `pnpm start` in `NODE_ENV=test`, `E2E_TESTING=1`) — the same config CI uses. Baselines come from `git stash`ing the spec changes and re-running the identical command, so both sides ran against the same build and machine. The docker change is evidenced by measurement of existing runs; its check is the job on this PR, which must stay green and get faster.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `survey.spec.ts` passes without `slowMo`, on CI | e2e | Two green runs on this branch: test step **303s** then **288s**, against 598s on the run analysed above; E2E job 10m54s / 10m30s against a 12m43s median. Local, 2 workers: 3 passed in 151s vs 286s baseline — longest test 198s → 114s, which is what bounds CI wall clock. |
| Rich-text entry commits, not just renders | e2e | Probe spec typed via `fill()`, saved the draft, reloaded: value persisted. Same probe with `pressSequentially` (no delay) failed on truncation — this is why `fill()` and not merely faster typing. |
| Matrix rows/columns and choice options survive to Publish | e2e | Covered by the two `survey.spec.ts` tests that build the 13-element survey and publish it; before the converge-loop fix these failed with `Choice 1 in question 1 of block 2 is missing` and a missing `#column-3`. |
| Other specs using the changed helpers | e2e | `storage-smoke.spec.ts` + `survey-editor-rich-text-tabs.spec.ts` → 3 passed (12.0s). The only other specs importing `fillRichTextEditor` / `createSurvey` / the add-block pattern. |
| Add-block trigger is keyboard reachable | manual | `asChild` dropped, so Radix renders a `<button>`: the control now takes focus and opens the picker from the keyboard, where the `div` took no focus at all. Layout is unchanged (same classes, same inner markup). |
| Review round re-verified end to end | e2e | After the review fixes: `survey.spec.ts` 3 passed in **144s** (37.4s / 28.3s / 108s), `storage-smoke` + `survey-editor-rich-text-tabs` 3 passed in 12.8s, against the rebuilt app. The first attempt at the per-item add loop failed here with `Expected 4, Received 3` — a single `toPass` stops at the first successful add — which is why the loop is now driven from outside the retry. |
| `data-testid` reaches the rendered editor | manual | Rebuilt and grepped the emitted chunk for `add-element-trigger`; the spec's `getByTestId` resolves against the running app. A first run against a stale build failed here, which is how the check earned its place. |
| Docker image still builds and passes every validation step, faster | e2e | This PR's `Validate Docker Build` run: **pass, job 8.7m / build step 7.1m**, against 12–17m / 10.4–15.3m across the previous six runs. Same steps throughout (npm-exclusion check, env-rejection cases, Compose migration validation, startup-migration skip, health check) — only the cache options are gone. |
| The removed docker cache was a net loss | manual | Four runs via `gh run view <id> --log`: `CACHED` counts 0/0/0/8, cache export 222s / 286s / 407s, and the cold cost of the only ever-cached steps (apk, corepack ×3, chmod, WORKDIR, runner apk) summing to 12s. `grep -rn "type=gha\|cache-from\|cache-to" .github/workflows/` matches only this workflow. |

**Open gaps**

- Both savings are measured over one or two runs, not a distribution: docker export time varied 222–407s before this change, so the 8.7m job is one sample of a range rather than a new ceiling.
- The rest of the e2e suite (the other 25 specs) was not run locally — they don't import the changed helpers, so this PR's own E2E run is the check.
- Local e2e numbers are single runs on one machine, not repeated samples. The previously flaky `Multi Language Survey Create` needs several runs before its flake rate can be called improved.
- The underlying editor bug is untouched: fast successive edits can still drop a label for a real user. The specs work around it rather than fixing it — worth its own ticket.
- The address rows in `survey.spec.ts` still use positional `getByRole("cell").nth(2)` / `getByRole("switch").nth(1)` locators (raised in review). Those lines are unchanged by this diff; fixing them means giving the address switches accessible names in `address-element-form.tsx`, which belongs with the editor-bug ticket rather than here.
- Three larger docker costs are left alone: the 239s `apps/web` build (built again in the E2E job, with no Turbo remote cache anywhere in the repo to share it), the 45s image export, and the ~106-layer runner stage built from `COPY` + `RUN chown -R` + `RUN chmod -R` triples.

**Risks**

- The converge loops assert against `#survey-preview`. If a future change stops rendering an element type's labels there, those helpers spin until their 60s cap rather than failing fast.
- Structural adds retry while the expected input count is short. The `count < total` guard cannot overshoot, but a genuinely stuck add now surfaces after the `toPass` timeout instead of immediately.
- `RENDER_SETTLE_MS = 150` in `helper.ts` still paces the ~40 structural seams — the residue of the old global pacing, and now deliberately sized against the editor's `debounce(handleUpdate, 100)`. If CI proves slower than local, that constant is the first knob.
- `ensureChoiceCount` picks its control by which one exists ("Add option" in the ranking form, else the per-row "Add choice below"). A form that renders both would take the ranking path.
- If someone later adds a push-to-main build that populates the docker cache scope, its removal becomes the wrong default; the comment at the step says so, with what has to change first.

---

> [!NOTE]

### DIFF
--- a/apps/web/modules/survey/editor/components/add-element-button.tsx
+++ b/apps/web/modules/survey/editor/components/add-element-button.tsx
@@ -93,7 +93,11 @@ export const AddElementButton = ({ addElement, workspace, isCxMode }: AddElement
         "group w-full overflow-hidden rounded-lg border border-slate-300 bg-white duration-200 hover:cursor-pointer hover:bg-slate-50"
       )}>
       <Collapsible.CollapsibleTrigger asChild className="group h-full w-full">
-        <div className="inline-flex">
+        {/* data-testid: the trigger is a `div` (Radix `asChild`), so it exposes no button role for
+            tests to target. Without a stable hook, E2E had to match it by its rendered text through
+            a positional `locator("div").nth(1)`, which re-resolved to a different node on every
+            editor re-render and produced "element detached from the DOM" flakes. */}
+        <div className="inline-flex" data-testid="add-element-trigger">
           <div className="flex w-10 items-center justify-center rounded-l-[7px] bg-brand-dark group-aria-expanded:rounded-br group-aria-expanded:rounded-bl-none">
             <PlusIcon className="size-5 text-white" />
           </div>
--- a/apps/web/modules/survey/editor/components/add-element-button.tsx
+++ b/apps/web/modules/survey/editor/components/add-element-button.tsx
@@ -92,12 +92,13 @@ export const AddElementButton = ({ addElement, workspace, isCxMode }: AddElement
         open ? "shadow-lg" : "shadow-md",
         "group w-full overflow-hidden rounded-lg border border-slate-300 bg-white duration-200 hover:cursor-pointer hover:bg-slate-50"
       )}>
-      <Collapsible.CollapsibleTrigger asChild className="group h-full w-full">
-        {/* data-testid: the trigger is a `div` (Radix `asChild`), so it exposes no button role for
-            tests to target. Without a stable hook, E2E had to match it by its rendered text through
-            a positional `locator("div").nth(1)`, which re-resolved to a different node on every
-            editor re-render and produced "element detached from the DOM" flakes. */}
-        <div className="inline-flex" data-testid="add-element-trigger">
+      {/* Not `asChild` with a `div`: Radix forwards the click handler and aria state to the child but
+          adds no role and no tabIndex, so the control was unreachable by keyboard — a user could not
+          open the element picker at all. Letting Radix render its own `<button>` fixes that and
+          gives tests a real button role; the data-testid keeps them off the rendered copy, which
+          also appears in the live preview. */}
+      <Collapsible.CollapsibleTrigger className="group h-full w-full" data-testid="add-element-trigger">
+        <div className="inline-flex">
           <div className="flex w-10 items-center justify-center rounded-l-[7px] bg-brand-dark group-aria-expanded:rounded-br group-aria-expanded:rounded-bl-none">
             <PlusIcon className="size-5 text-white" />
           </div>
@@ -109,7 +110,9 @@ export const AddElementButton = ({ addElement, workspace, isCxMode }: AddElement
           </div>
         </div>
       </Collapsible.CollapsibleTrigger>
-      <Collapsible.CollapsibleContent className="overflow-hidden data-[state=closed]:animate-collapsible-up data-[state=open]:animate-collapsible-down">
+      <Collapsible.CollapsibleContent
+        data-testid="add-element-picker"
+        className="overflow-hidden data-[state=closed]:animate-collapsible-up data-[state=open]:animate-collapsible-down">
         <div className="grid grid-cols-1 gap-x-6 px-3 pb-3 sm:grid-cols-2">
           <div>{visibleCategories.filter((category) => category.column === 1).map(renderCategory)}</div>
           <div>{visibleCategories.filter((category) => category.column === 2).map(renderCategory)}</div>
--- a/apps/web/playwright/survey.spec.ts
+++ b/apps/web/playwright/survey.spec.ts
@@ -10,17 +10,6 @@ import {
   uploadImageChoicesForPictureSelection,
 } from "./utils/helper";
 
-// NOTE: slowMo paces every action in this spec's heavy survey-editor flows.
-// It is load-bearing: without it, the question-builder interactions (matrix
-// columns, ranking options) hit "element detached from the DOM" races and the
-// tests fail. Removing it requires first hardening those waits in
-// createSurvey/createSurveyWithLogic (utils/helper.ts). Tracked as a follow-up.
-test.use({
-  launchOptions: {
-    slowMo: 150,
-  },
-});
-
 test.beforeEach(async ({ page }) => {
   await helper.mockStorageUploads(page);
 });
@@ -77,10 +66,7 @@ test.describe("Survey Create & Submit Response without logic", async () => {
       await page.getByRole("button", { name: "Save as draft", exact: true }).click();
       await expect(page.getByText("Changes saved.")).toBeVisible();
 
-      await Promise.all([
-        page.waitForURL(/\/workspaces\/[^/]+\/surveys\/[^/]+\/summary(\?.*)?$/, { timeout: 120000 }),
-        page.getByRole("button", { name: "Publish", exact: true }).click(),
-      ]);
+      await helper.publishSurvey(page);
       await page.getByLabel("Copy survey link to clipboard").click();
       url = await page.evaluate("navigator.clipboard.readText()");
     });
@@ -325,29 +311,15 @@ test.describe("Multi Language Survey Create", async () => {
     await page.locator("#welcome-toggle").click();
 
     // Add questions in default language
-    await page.getByText("Add Block").click();
-    await page.getByRole("button", { name: "Single-Select" }).click();
+    await helper.addElement(page, "Single-Select");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.singleSelectQuestion.question);
-    await page.getByPlaceholder("Option 1").fill(surveys.createAndSubmit.singleSelectQuestion.options[0]);
-    await page.getByPlaceholder("Option 2").fill(surveys.createAndSubmit.singleSelectQuestion.options[1]);
+    await helper.fillChoiceOptions(page, surveys.createAndSubmit.singleSelectQuestion.options);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Multi-Select", exact: true }).click();
+    await helper.addElement(page, "Multi-Select", { exact: true });
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.multiSelectQuestion.question);
-    await page.getByPlaceholder("Option 1").fill(surveys.createAndSubmit.multiSelectQuestion.options[0]);
-    await page.getByPlaceholder("Option 2").fill(surveys.createAndSubmit.multiSelectQuestion.options[1]);
-    await page.getByPlaceholder("Option 3").fill(surveys.createAndSubmit.multiSelectQuestion.options[2]);
+    await helper.fillChoiceOptions(page, surveys.createAndSubmit.multiSelectQuestion.options);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Picture Selection" }).click();
+    await helper.addElement(page, "Picture Selection");
     await helper.fillRichTextEditor(
       page,
       "Question*",
@@ -356,75 +328,28 @@ test.describe("Multi Language Survey Create", async () => {
 
     await uploadImageChoicesForPictureSelection(page);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Rating" }).click();
+    await helper.addElement(page, "Rating");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.ratingQuestion.question);
     await page.getByPlaceholder("Not good").fill(surveys.createAndSubmit.ratingQuestion.lowLabel);
     await page.getByPlaceholder("Very satisfied").fill(surveys.createAndSubmit.ratingQuestion.highLabel);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Net Promoter Score (NPS)" }).click();
+    await helper.addElement(page, "Net Promoter Score (NPS)");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.npsQuestion.question);
     await page.getByLabel("Lower label").fill(surveys.createAndSubmit.npsQuestion.lowLabel);
     await page.getByLabel("Upper label").fill(surveys.createAndSubmit.npsQuestion.highLabel);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Date" }).click();
+    await helper.addElement(page, "Date");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.dateQuestion.question);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "File Upload" }).click();
+    await helper.addElement(page, "File Upload");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.fileUploadQuestion.question);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-
-    await page.getByRole("button", { name: "Matrix" }).scrollIntoViewIfNeeded();
-    await page.getByRole("button", { name: "Matrix" }).click();
+    await helper.addElement(page, "Matrix");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.matrix.question);
-    await page.locator("#row-0").click();
-    await page.locator("#row-0").fill(surveys.createAndSubmit.matrix.rows[0]);
-    await page.locator("#row-1").click();
-    await page.locator("#row-1").fill(surveys.createAndSubmit.matrix.rows[1]);
-    await page.getByRole("button", { name: "Add row" }).click();
-    await expect(page.locator("#row-2")).toBeEditable();
-    await page.locator("#row-2").fill(surveys.createAndSubmit.matrix.rows[2]);
-    await page.locator("#column-0").click();
-    await page.locator("#column-0").fill(surveys.createAndSubmit.matrix.columns[0]);
-    await page.locator("#column-1").click();
-    await page.locator("#column-1").fill(surveys.createAndSubmit.matrix.columns[1]);
-    await page.getByRole("button", { name: "Add column" }).click();
-    await expect(page.locator("#column-2")).toBeEditable();
-    await page.locator("#column-2").fill(surveys.createAndSubmit.matrix.columns[2]);
-    await page.getByRole("button", { name: "Add column" }).click();
-    await expect(page.locator("#column-3")).toBeEditable();
-    await page.locator("#column-3").fill(surveys.createAndSubmit.matrix.columns[3]);
+    await helper.fillMatrixLabels(page, "row", surveys.createAndSubmit.matrix.rows);
+    await helper.fillMatrixLabels(page, "column", surveys.createAndSubmit.matrix.columns);
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Address" }).click();
+    await helper.addElement(page, "Address");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.address.question);
     await page.getByRole("row", { name: "Address Line 2" }).getByRole("switch").nth(1).click();
     await page.getByRole("row", { name: "City" }).getByRole("cell").nth(2).click();
@@ -432,26 +357,9 @@ test.describe("Multi Language Survey Create", async () => {
     await page.getByRole("row", { name: "Zip" }).getByRole("cell").nth(2).click();
     await page.getByRole("row", { name: "Country" }).getByRole("switch").nth(1).click();
 
-    await page
-      .locator("div")
-      .filter({ hasText: /^Add BlockChoose the first question on your Block$/ })
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Ranking" }).click();
+    await helper.addElement(page, "Ranking");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.ranking.question);
-    await page.getByPlaceholder("Option 1").click();
-    await page.getByPlaceholder("Option 1").fill(surveys.createAndSubmit.ranking.choices[0]);
-    await page.getByPlaceholder("Option 2").click();
-    await page.getByPlaceholder("Option 2").fill(surveys.createAndSubmit.ranking.choices[1]);
-    await page.getByRole("button", { name: "Add option" }).click();
-    await page.getByPlaceholder("Option 3").click();
-    await page.getByPlaceholder("Option 3").fill(surveys.createAndSubmit.ranking.choices[2]);
-    await page.getByRole("button", { name: "Add option" }).click();
-    await page.getByPlaceholder("Option 4").click();
-    await page.getByPlaceholder("Option 4").fill(surveys.createAndSubmit.ranking.choices[3]);
-    await page.getByRole("button", { name: "Add option" }).click();
-    await page.getByPlaceholder("Option 5").click();
-    await page.getByPlaceholder("Option 5").fill(surveys.createAndSubmit.ranking.choices[4]);
+    await helper.fillChoiceOptions(page, surveys.createAndSubmit.ranking.choices);
 
     // Navigate to Language tab to enable translations and add German
     await page.getByText("Language").click();
@@ -747,10 +655,7 @@ test.describe("Multi Language Survey Create", async () => {
     await page.getByRole("button", { name: "Save as draft", exact: true }).click();
     await expect(page.getByText("Changes saved.")).toBeVisible();
 
-    await Promise.all([
-      page.waitForURL(/\/workspaces\/[^/]+\/surveys\/[^/]+\/summary(\?.*)?$/, { timeout: 120000 }),
-      page.getByRole("button", { name: "Publish", exact: true }).click(),
-    ]);
+    await helper.publishSurvey(page);
     await page.getByLabel("Select Language").click();
     await page.getByText("German").click();
     await page.getByLabel("Copy survey link to clipboard").click();
@@ -791,10 +696,7 @@ test.describe("Testing Survey with advanced logic", async () => {
       await page.getByRole("button", { name: "Save as draft", exact: true }).click();
       await expect(page.getByText("Changes saved.")).toBeVisible();
 
-      await Promise.all([
-        page.waitForURL(/\/workspaces\/[^/]+\/surveys\/[^/]+\/summary(\?.*)?$/, { timeout: 120000 }),
-        page.getByRole("button", { name: "Publish", exact: true }).click(),
-      ]);
+      await helper.publishSurvey(page);
 
       // Get URL
       await page.getByLabel("Copy survey link to clipboard").click();
--- a/apps/web/playwright/utils/helper.ts
+++ b/apps/web/playwright/utils/helper.ts
@@ -424,14 +424,19 @@ export const signupUsingInviteToken = async (page: Page, name: string, email: st
 };
 
 /**
- * Wait for React to commit and paint whatever the previous interaction changed.
+ * Wait for the previous interaction to reach the survey state, then for a frame after it.
  *
- * The survey editor's "add" handlers (`handleAddLabel`, `addChoice`, `addElement`, the "Add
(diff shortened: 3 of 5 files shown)


apps/web/playwright/utils/helper.ts
```diff
@@ -423,6 +423,32 @@ export const signupUsingInviteToken = async (page: Page, name: string, email: st
...
+    (settleMs) =>
+      new Promise<void>((resolve) => {
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            setTimeout(resolve, settleMs);
+          });
+        });
+      }),
+    RENDER_SETTLE_MS
+  );
```
kiln-agent, line 452: **The two frames run before the sleep, so nothing waits for a frame *after* the commit.**

`ElementFormInput` pushes the edit upstream through `debounce(handleUpdate, 100)` (`apps/web/modules/survey/components/element-form-input/index.tsx:305`). This wait resolves its two rAFs at ~16 ms and ~32 ms — before the debounce has even fired — and then just sleeps 150 ms. So the rAF pair contributes nothing, the only margin over the debounce is ~50 ms, and there is no frame/paint guarantee after the `updateElement` the debounce triggers.

On a contended CI runner (4 workers plus the app server in the same box) that 50 ms margin disappears and the next structural click (`addElement`, `addOtherChoice`, `ensureChoiceCount`'s "Add option") again reads a stale `element` prop and silently drops the previous edit — the exact `Choice 1 in question 1 of block 2 is missing` failure this helper exists to prevent. Inverting the order gives what the JSDoc promises:

```suggestion
const flushRender = async (page: Page): Promise<void> => {
  await page.evaluate(
    (settleMs) =>
      new Promise<void>((resolve) => {
        // Sleep first so the input's 100ms debounce has fired and dispatched its
        // `updateElement`, then wait two frames for React to commit and paint it.
        setTimeout(() => {
          requestAnimationFrame(() => {
            requestAnimationFrame(() => resolve());
          });
        }, settleMs);
      }),
    RENDER_SETTLE_MS
  );
};
```

Separately, the JSDoc above is inaccurate about the cost: it says "Two animation frames are enough to guarantee the commit has painted, at ~30ms a seam instead of 150ms on every single action", but the code sleeps `RENDER_SETTLE_MS = 150` on every seam — the same pacing as the `slowMo: 150` it replaced, just at ~40 seams instead of 726 actions. Worth rewording so the next person tuning the constant isn't misled.

---

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+  await expect(field).toHaveValue(value);
+  await flushRender(page);
+};
+
+/** Click "Add option" until the choice list holds `total` inputs (same rationale as the matrix). */
+export const ensureChoiceCount = async (page: Page, total: number) => {
+  // Counting by placeholder, not by `id^="choice-"`: the "Other" and "None of the above" choices
+  // share that id prefix, so an id-based count would never match the number of real options.
+  const choices = page.locator('input[placeholder^="Option "]');
+  const addButton = page.getByRole("button", { name: "Add option" });
```
kiln-agent: **"Add option" only exists in the ranking form — for select elements this button never resolves.**

`workspace.surveys.edit.add_option` is rendered in exactly one place: `ranking-element-form.tsx:248`. The multiple-choice form (`multiple-choice-element-form.tsx`) has no "Add option" button at all — its only add control is the per-row icon button `aria-label="Add choice below"` in `element-option-choice.tsx`, alongside "Add “Other”" / "Add None of the above" / "Bulk edit".

`fillChoiceOptions` is now the shared path for single-select, multi-select *and* ranking (5 call sites across `createSurvey`, `createSurveyWithLogic` and the multi-language spec). It happens to work today only because the fixtures match the presets exactly — single-select has 2 options and a 2-choice preset, multi-select 3 and 3 (`modules/survey/lib/elements.tsx:77`/`:98`), so the `count < total` branch is never entered for a select element.

Concrete break: add a third entry to `surveys.createAndSubmit.singleSelectQuestion.options` in `apps/web/playwright/utils/mock.ts` (or reuse this exported helper for any select element with more values than the preset) and `expect(addButton).toBeEnabled()` spins for the whole 30 s `toPass` budget, then fails on a locator that matches nothing — with no hint that the button belongs to a different element type.

Either make the add control element-aware (fall back to the last row's "Add choice below" when "Add option" is absent), or note in the JSDoc that the helper only grows ranking lists and that select callers must pass exactly the preset count.

---

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+
+  await expect(async () => {
+    if ((await choices.count()) < total) {
+      await flushRender(page);
+      await expect(addButton).toBeEnabled();
+      await addButton.click();
+    }
+    await expect(choices).toHaveCount(total, { timeout: 2000 });
+  }).toPass({ timeout: 30000 });
```
kiln-agent: **One click per attempt + waiting for the *final* count burns ~2 s per missing item, and the overshoot guard doesn't hold.**

Applies equally to `ensureMatrixLabelCount` above.

Cost: ranking goes 2 → 5. Attempt 1 clicks, count becomes 3, then `toHaveCount(5, {timeout: 2000})` burns the full 2 s and fails; same for attempts 2 and 3 (plus `toPass` backoff). That's ~4–5 s of pure timeout per ranking element, and there are four ranking/matrix-heavy call sites in the suite — meaningful in a PR whose whole point is wall clock.

Correctness: the comment says "`count < total` means a retry cannot overshoot", but that only holds if each click's re-render is observed *within* the 2000 ms cap. On the 13-element editor under CI contention a commit can take longer than that, so the next attempt still reads the pre-click count, clicks again, and the list grows past `total`. There is no delete path, so `toHaveCount(total)` can then never be satisfied and the helper hard-fails after the 30 s budget with a count that is too *high* — a confusing failure mode.

Asserting growth after each click fixes both at once:

```suggestion
  await expect(async () => {
    const before = await choices.count();
    if (before < total) {
      await flushRender(page);
      await expect(addButton).toBeEnabled();
      await addButton.click();
      // Wait for *this* click to land, not for the final total: the assertion then
      // succeeds as soon as the add commits, and the next attempt can never
      // double-click a list whose previous add is still in flight.
      await expect(choices).toHaveCount(before + 1, { timeout: 5000 });
    }
    await expect(choices).toHaveCount(total, { timeout: 2000 });
  }).toPass({ timeout: 30000 });
```

---

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+  const publishButton = page.getByRole("button", { name: "Publish", exact: true });
+  const summaryUrl = /\/workspaces\/[^/]+\/surveys\/[^/]+\/summary(\?.*)?$/;
+  // react-hot-toast renders every toast into a role="status" container.
+  const toasts = page.locator('div[role="status"]');
+
+  await expect(publishButton).toBeEnabled();
+  await publishButton.click();
+
+  const seen = new Set<string>();
+  const deadline = Date.now() + 60000;
```
kiln-agent: **This halves the publish timeout: all three replaced call sites waited 120 s.**

Every site this helper replaces was `page.waitForURL(/…\/summary(\?.*)?$/, { timeout: 120000 })` — i.e. effectively "until the 120 s per-test timeout" (`playwright.config.ts:19`). The new deadline is 60 s.

Publishing the 13-element survey is the slowest single step in the suite, and it's the one running on a 4-worker CI runner that shares CPU with the app server. A publish that takes 61 s now throws `Publish did not reach the summary page` while the request is still in flight — a slow-but-passing run becomes red. And because no validation toast was shown in that case, the new diagnostic reports `[]`, which reads as a mystery rather than a timeout.

```suggestion
  const deadline = Date.now() + 120000;
```

If the 60 s was deliberate (fail fast rather than eat the test budget), it's worth saying so in the JSDoc and including "no toast seen — likely a timeout, not a validation error" in the thrown message so the empty array isn't misread.

---

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+
+/**
+ * Fill one matrix row/column label and wait for the value to be committed.
+ *
+ * `handleAddLabel` and `updateMatrixLabel` in matrix-element-form.tsx both rebuild the whole label
+ * array from the `element` prop of the render they were created in, so an add that lands while an
+ * edit is still in flight overwrites it — the edited label reverts to "" and the new label is lost.
+ * Every label is therefore added first (see `ensureMatrixLabelCount`) and only then filled.
+ */
+export const fillMatrixLabel = async (page: Page, type: "row" | "column", index: number, value: string) => {
```
kiln-agent: **Dead code: `fillMatrixLabel` (here) and `fillChoiceOption` (line 633) have no callers.**

Grepping `apps/web/playwright` for either name returns only their own definitions plus one doc reference — the array-level `fillMatrixLabels` / `fillChoiceOptions` fully supersede them.

Beyond the AGENTS.md "remove dead code" rule, keeping them exported is actively risky here: they look like the blessed single-field API, but they only assert `toHaveValue` on the input and skip the converge-against-`#survey-preview` loop that this PR argues is the *only* reliable proof an edit committed (see the `surveyPreview` doc comment). A future spec reaching for `fillChoiceOption` would silently reintroduce the dropped-label failure the PR just fixed.

---

kiln-agent (commented):
Reviewed the diff for correctness. The direction is right and the product-code change (`data-testid` on the Radix `asChild` trigger) is clean — `AddElementButton` is rendered once (`elements-view.tsx:905`), so `getByTestId` is unambiguous, and Radix does put `data-state` on the `asChild` child, so both state assertions in `addElement` are sound.

Five findings on the helper, in severity order:

1. `flushRender` runs its two animation frames *before* the 150 ms sleep, so nothing waits for a frame after the commit it's meant to flush — and the input's own 100 ms debounce means the rAF pair fires before the update is even dispatched.
2. `ensureChoiceCount` clicks "Add option", which only the ranking form renders — `fillChoiceOptions` can't grow a select element's choice list at all.
3. The add-loops burn ~2 s per missing item and their overshoot guard doesn't hold under a slow commit.
4. `publishSurvey` drops the publish timeout from 120 s to 60 s.
5. `fillMatrixLabel` / `fillChoiceOption` are exported but never called.

Things I checked that are fine: `input[placeholder^="Option "]` and `input[id^="row-|column-"]` do match real `<input>`s (`ElementFormInput` only renders a rich-text editor for `headline`/`subheader`/`html`); collapsed element forms unmount (no `forceMount`), so the page-wide counts are safe; `#survey-preview` is in the same document, not an iframe, and is visible at the 1280×720 Desktop Chrome viewport; matrix rows render as `<th scope="row">` independently of column labels, so filling rows before columns converges; publish uses `router.push`, a soft navigation, so `allInnerTexts()` in the poll loop won't hit a destroyed execution context; and the `fillLabelsUntilCommitted` re-fill guard does work, because `ElementFormInput` re-syncs its local `text` from props (`useEffect(() => setText(elementText), [elementText])`) — though that also means the `surveyPreview` doc comment's claim that "an input can still show text the survey no longer holds" isn't quite what happens.

---

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+ *
+ * The editor validates client-side and reports problems only through a toast that auto-dismisses,
+ * so a failed publish otherwise shows up as an opaque `waitForURL` timeout. This polls for the
+ * toast alongside the navigation and fails with the message the editor actually gave.
+ */
+export const publishSurvey = async (page: Page): Promise<void> => {
+  const publishButton = page.getByRole("button", { name: "Publish", exact: true });
+  const summaryUrl = /\/workspaces\/[^/]+\/surveys\/[^/]+\/summary(\?.*)?$/;
+  // react-hot-toast renders every toast into a role="status" container.
+  const toasts = page.locator('div[role="status"]');
```
kiln-agent: **`div[role="status"]` is not toast-specific — the shared `Alert` uses the same role.**

`grep -rn 'role="status"' apps/web/modules` finds it on `Alert` all over the editor and settings surfaces the spec is on when it clicks Publish: `survey-menu-bar.tsx:636` (storage not configured), `block-card.tsx:399` (caution when `responseCount > 0`), `how-to-send-card.tsx:118` (`status === "inProgress"`), `response-options-card.tsx:477` (spam protection). Whenever one of those conditions holds, its text lands in `seen` and the failure reads `Editor reported: ["<unrelated warning>"]` — which is worse than the opaque `waitForURL` timeout it replaces, because it now points confidently at the wrong cause.

The `message !== "Formbricks"` filter is itself evidence that non-toast nodes already match. Scoping to the react-hot-toast container would keep the diagnostic honest:

```suggestion
  // react-hot-toast renders its toasts inside the <Toaster/> container; scope to it so the
  // editor's own `Alert role="status"` regions don't get reported as publish validation errors.
  const toasts = page.locator('[data-testid="toaster"] div[role="status"]');
```

(That testid does not exist yet — either add it to the `<Toaster/>` wrapper, or use whatever stable hook the Toaster renders. The point is the selector should only match toasts.)

---

apps/web/modules/survey/editor/components/add-element-button.tsx
```diff
@@ -93,7 +93,11 @@ export const AddElementButton = ({ addElement, workspace, isCxMode }: AddElement
         "group w-full overflow-hidden rounded-lg border border-slate-300 bg-white duration-200 hover:cursor-pointer hover:bg-slate-50"
       )}>
       <Collapsible.CollapsibleTrigger asChild className="group h-full w-full">
-        <div className="inline-flex">
+        {/* data-testid: the trigger is a `div` (Radix `asChild`), so it exposes no button role for
+            tests to target. Without a stable hook, E2E had to match it by its rendered text through
+            a positional `locator("div").nth(1)`, which re-resolved to a different node on every
+            editor re-render and produced "element detached from the DOM" flakes. */}
+        <div className="inline-flex" data-testid="add-element-trigger">
```
kiln-agent: **The comment diagnoses a real a11y bug and then works around it.**

`Collapsible.CollapsibleTrigger asChild` forwards `onClick`, `aria-expanded`, `aria-controls` and `data-state` onto this child, but it adds no `role` and no `tabIndex` — so the add-block control is not focusable. A keyboard-only or screen-reader user cannot open the element picker at all, i.e. cannot add a block. Radix renders a real `<button>` when `asChild` is dropped, which is presumably why this was `asChild` + `div` by accident rather than by design (the child adds nothing but `className="inline-flex"`).

Dropping `asChild` (or making the child a `<button type="button">`) fixes the a11y gap *and* gives the specs the stable `getByRole("button", { name: /Add Block/ })` hook that `data-testid` is standing in for — one change instead of a test hook layered on top of the defect. Pre-existing, so not a blocker, but this diff is the moment it's cheapest to fix.

---

kiln-agent (commented):
Second correctness pass on the same commit. Four of my findings duplicate the review already on this PR (`flushRender`'s frame/debounce ordering, `ensureChoiceCount`'s "Add option" only existing in the ranking form, the 120s→60s publish timeout, and the never-called `fillMatrixLabel`/`fillChoiceOption` exports) — I confirmed all four independently and have nothing to add beyond what is already written there. The two comments below are new.

One extra data point on `flushRender` that the earlier thread does not mention: because the promise resolves only from inside the second `requestAnimationFrame` callback, a page whose frames stop firing (throttled or non-painting renderer) leaves `page.evaluate` hanging until the enclosing 5–8 minute test timeout, reported at an arbitrary helper call. If the frame barrier is kept, it is worth knowing that a plain `page.waitForTimeout(RENDER_SETTLE_MS)` has no such failure mode.

Also verified as fine, for the record: `AddElementButton` is rendered exactly once (`elements-view.tsx:905`), so `getByTestId` is unambiguous; Radix does put `data-state` on the `asChild` child, so both `addElement` state assertions are sound; collapsed element forms unmount (no `forceMount`), so the page-wide `input[placeholder^="Option "]` / `input[id^="row-"]` counts cannot pick up a neighbouring card; and every current `fillChoiceOptions` / `fillMatrixLabels` call site matches the element presets, so today's specs never hit the add-path limitation.

---

apps/web/playwright/utils/helper.ts
Dhruwang: Confirmed and fixed in 05757f36 — `debounce(handleUpdate, 100)` at element-form-input:305 is real, so the frames were resolving at ~16/32ms, before the debounce had fired. `flushRender` now sleeps first and takes the two frames after, and the JSDoc names the debounce and the ~180ms/seam cost instead of the old ~30ms claim (which is also what CodeRabbit flagged on the same lines).

apps/web/playwright/utils/helper.ts
Dhruwang: Correct — `add_option` is only in ranking-element-form.tsx:248; the multiple-choice forms only have the per-row `aria-label="Add choice below"`. `ensureChoiceCount` now uses whichever control exists, so the select path no longer depends on the fixtures happening to match the 2- and 3-choice presets.

apps/web/playwright/utils/helper.ts
Dhruwang: Both points were right, and the first version I pushed was worse than you described: a single `toPass` that "added one" reported success and stopped, so it never reached the target at all — the matrix run failed with `Expected 4, Received 3` locally. Now one item per loop iteration, each add retried against its own target count (`count + 1`), so intermediate passes cost one fast assertion instead of the full 2s cap and an add can never overshoot.

apps/web/playwright/utils/helper.ts
Dhruwang: Restored to 120000ms to match the replaced call sites and the per-test timeout. The thrown message now distinguishes the two cases explicitly: with no toast it says the publish was slow rather than rejected, instead of reporting an empty array.

apps/web/playwright/utils/helper.ts
Dhruwang: Both removed. They were left over from an earlier iteration of the converge loop and, as you say, would have been the tempting single-field API while skipping the `#survey-preview` check that makes the fix work.

apps/web/playwright/utils/helper.ts
Dhruwang: Right, and `role="status"` on `Alert` is exactly why the `!== "Formbricks"` filter existed. Scoped to `.formbricks__toast__error` instead — react-hot-toast already gets that class from modules/ui/components/toaster-client, so no product change was needed and only genuine error toasts are collected.

apps/web/modules/survey/editor/components/add-element-button.tsx
Dhruwang: Agreed, and fixed rather than worked around: `asChild` is gone, so Radix renders its own `<button>` and the add-block control is keyboard reachable. The `data-testid` stays, because the editor's live preview renders a second copy of the block text and the specs should not be able to match that one.

apps/web/playwright/utils/helper.ts
```diff
@@ -423,6 +423,32 @@ export const signupUsingInviteToken = async (page: Page, name: string, email: st
...
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            setTimeout(resolve, settleMs);
+          });
+        });
+      }),
+    RENDER_SETTLE_MS
+  );
+};
+
```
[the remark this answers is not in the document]
Dhruwang: Fixed — but by keeping the sleep and correcting the comment, because the sleep turns out to be load-bearing: `ElementFormInput` pushes edits upstream through `debounce(handleUpdate, 100)` (element-form-input:305), so two animation frames resolve at ~16/32ms, before the debounce has fired, and guarantee nothing on their own. The order was also wrong: `flushRender` now sleeps past the debounce first and takes the frames after, so the commit it waits for is the one the debounce triggers. The JSDoc names the debounce and the honest ~180ms/seam cost.

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+  // Radix flips `data-state` on the trigger, so this waits for the panel to be open rather than
+  // for a timeout to expire.
+  await expect(trigger).toHaveAttribute("data-state", "open");
+
+  const option = page.getByRole("button", { name: elementType, exact: options.exact ?? false });
+  await expect(option).toBeVisible();
+  // The picker is a two-column grid that overflows on shorter viewports (e.g. "Matrix").
+  await option.scrollIntoViewIfNeeded();
+  await option.click();
```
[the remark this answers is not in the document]
Dhruwang: Good catch, fixed. Scoped via a `data-testid="add-element-picker"` on `Collapsible.CollapsibleContent` rather than a `:has()` on the trigger, since the trigger and content are siblings — and the trigger is now a real `<button>` (dropping `asChild` also fixed a keyboard-reachability bug on that control, per another review thread).

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+  while (Date.now() < deadline) {
+    if (summaryUrl.test(page.url())) return;
+
+    for (const text of await toasts.allInnerTexts()) {
+      const message = text.trim();
+      if (message && message !== "Formbricks") seen.add(message);
+    }
+
+    await page.waitForTimeout(250);
+  }
```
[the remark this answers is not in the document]
Dhruwang: Both right. The toast reads are now raced against the navigation promise, so once the summary page starts loading no further locator query is issued, and the read itself is guarded for the case where navigation begins between the race and the query. `waitForTimeout(250)` is still the poll interval — it is bounded by the navigation rather than pacing an interaction, but say the word if you would rather see it driven off toast visibility.

apps/web/playwright/survey.spec.ts
```diff
@@ -356,102 +328,38 @@ test.describe("Multi Language Survey Create", async () => {
...
-      .nth(1)
-      .click();
-    await page.getByRole("button", { name: "Address" }).click();
+    await helper.addElement(page, "Address");
     await helper.fillRichTextEditor(page, "Question*", surveys.createAndSubmit.address.question);
     await page.getByRole("row", { name: "Address Line 2" }).getByRole("switch").nth(1).click();
     await page.getByRole("row", { name: "City" }).getByRole("cell").nth(2).click();
     await page.getByRole("row", { name: "State" }).getByRole("switch").nth(1).click();
     await page.getByRole("row", { name: "Zip" }).getByRole("cell").nth(2).click();
     await page.getByRole("row", { name: "Country" }).getByRole("switch").nth(1).click();
```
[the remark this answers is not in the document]
Dhruwang: Skipping this one, with a reason: those five lines are untouched by this diff — the block reads as changed only because the `addElement` call above them replaced the positional add-block locator. Fixing them properly means giving the address switches accessible names in address-element-form.tsx, which is a product change with no bearing on the wall-clock goal here, and it would be silently reverted by the same stale-write bug this PR works around unless done carefully. Worth its own ticket alongside the underlying editor bug noted in Open gaps.

apps/web/playwright/utils/helper.ts
```diff
@@ -463,15 +495,208 @@ export const fillModalRichTranslation = async (page: Page, path: string, text: s
...
+ * edit is still in flight overwrites it — the edited label reverts to "" and the new label is lost.
+ * Every label is therefore added first (see `ensureMatrixLabelCount`) and only then filled.
+ */
+export const fillMatrixLabel = async (page: Page, type: "row" | "column", index: number, value: string) => {
+  const field = page.locator(`#${type}-${index}`);
+  await expect(field).toBeEditable();
+  await field.fill(value);
+  await expect(field).toHaveValue(value);
+  await flushRender(page);
+};
```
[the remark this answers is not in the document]
Dhruwang: Addressed by removing both single-field helpers — `fillMatrixLabel` and `fillChoiceOption` had no callers left after the array helpers landed, so the stale doc reference is gone with them.

mattinannt (approved):
Nice, thank you for the fix. Great to see that we got the e2e test finishing faster again. Nearly half the time it took a few weeks ago 💪

PR #8941, opened 2026-08-21
perf(ci): drop the docker build cache that costs more than it restores

Dhruwang:
No Linear ticket — spun out of a CI wall-clock investigation while working on ENG-2006's leftover; small and self-contained enough that a ticket would only restate this description.

## What & why

`Validate Docker Build` is the longest job on a PR (12–17 min over the last six runs; its `Build Docker Image` step alone is 10.4–15.3 min). A measurable chunk of that is the layer cache it exports to the GitHub Actions cache — which is a net loss today, not an optimization.

Decomposing one 10.4 min run ([32452739610](https://github.com/formbricks/formbricks/actions/runs/32452739610)):

| BuildKit step | Time |
| --- | --- |
| `pnpm install` | 34s |
| build `@formbricks/database` | 33s |
| build `@formbricks/web...` | 239s |
| exporting to docker image format | 45s |
| importing to docker (`load: true`) | 28s |
| **exporting to GitHub Actions Cache** | **222s** |
| base image, apk, ~106 runner layers | ~30s |

The export was 222s here, 286s in [32449834984](https://github.com/formbricks/formbricks/actions/runs/32449834984) and 407s in [32453148066](https://github.com/formbricks/formbricks/actions/runs/32453148066). Against that, what the cache can restore is **12s** of work: three of the four runs reported `CACHED` on nothing at all, and the one that did hit (8 steps) hit only the apk/corepack prelude — measured at 12s total when cold.

Two independent reasons it cannot do better as written, both left in a comment at the step:

- `COPY . .` sits ahead of `pnpm install` in `apps/web/Dockerfile`, so any diff in the repo invalidates every layer after it. Only the prelude is ever reusable.
- GHA caches are branch-scoped: a PR may read its own scope and `main`'s, but no workflow runs on pushes to `main`, and `merge_group` runs write to throwaway `gh-readonly-queue/...` scopes. So the scope PRs are allowed to read is never populated.

Making the cache pay needs both halves — reorder the Dockerfile so the install layer survives a source change, *and* give the cache a producer PRs can read (a push-to-main build, or a registry cache, which isn't branch-scoped). Until someone does both, paying 222–407s a run to save 12s is strictly worse than not caching. No other workflow uses `type=gha`, so nothing else is affected.

## Breaking changes

None — CI-only change to one workflow step's caching options. The image built, and every validation run against it, are unchanged.

## Migrations & env

- none

## How this was tested

The evidence is measurement of existing runs rather than a new test; the check on this PR is the job itself, which must stay green and get faster.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Docker image still builds and passes every validation step | e2e | This PR's own `Validate Docker Build` run — same steps (npm-exclusion check, env-rejection cases, Compose migration validation, startup-migration skip, health check), only the cache options removed. |
| The removed cache was a net loss | manual | Four runs inspected via `gh run view <id> --log`: `CACHED` counts 0/0/0/8, cache export 222s / 286s / 407s, and the cold cost of the only ever-cached steps (apk, corepack ×3, chmod, WORKDIR, runner apk) summing to 12s. |
| Nothing else consumes this cache | manual | `grep -rn "type=gha\|cache-from\|cache-to" .github/workflows/` matches only this workflow. |

**Open gaps**

- The expected saving is inferred from the export timings above, not yet observed on this branch — the first run on this PR is what confirms it. Worth a second look after it lands, since export time varied 222–407s.
- The three larger costs are untouched: the 239s `apps/web` build (built a second time in the E2E job, with no Turbo remote cache anywhere in the repo to share it), the 45s image export, and the ~106-layer runner stage built from `COPY` + `RUN chown -R` + `RUN chmod -R` triples. Each is its own change.

**Risks**

- If someone later adds a push-to-main build that populates the cache scope, this removal becomes the wrong default — the comment at the step says so, with what has to change first.
- Build times become slightly more variable run-to-run without any cache, though at 12s of restorable work the effect is inside the noise of the 239s build.

---

> [!NOTE]

### DIFF
--- a/.github/workflows/docker-build-validation.yml
+++ b/.github/workflows/docker-build-validation.yml
@@ -53,6 +53,15 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
 
+      # NOTE: deliberately no `cache-from`/`cache-to: type=gha`. Measured over four runs, exporting
+      # the cache cost 222-407s per run while the layers it could restore add up to 12s of work: the
+      # only steps that ever reported CACHED were the apk/corepack prelude, because `COPY . .` sits
+      # ahead of `pnpm install` in apps/web/Dockerfile and invalidates everything after it on any
+      # diff. Nothing populates a cache scope this job can read either — no workflow runs on pushes
+      # to main, and merge_group runs write to throwaway `gh-readonly-queue/...` scopes. Restoring
+      # the cache is worth revisiting only together with both halves of the fix: reorder the
+      # Dockerfile so the install layer survives a source change, and give the cache a producer PRs
+      # can read (a push-to-main build, or a registry cache, which is not branch-scoped).
       - name: Build Docker Image
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
         env:
@@ -63,8 +72,6 @@ jobs:
           push: false
           load: true
           tags: formbricks-test:${{ env.GITHUB_SHA }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
           secrets: |
             database_url=${{ secrets.DUMMY_DATABASE_URL }}
             encryption_key=${{ secrets.DUMMY_ENCRYPTION_KEY }}


.github/workflows/docker-build-validation.yml
```diff
@@ -53,6 +53,15 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
 
+      # NOTE: deliberately no `cache-from`/`cache-to: type=gha`. Measured over four runs, exporting
+      # the cache cost 222-407s per run while the layers it could restore add up to 12s of work: the
+      # only steps that ever reported CACHED were the apk/corepack prelude, because `COPY . .` sits
+      # ahead of `pnpm install` in apps/web/Dockerfile and invalidates everything after it on any
+      # diff. Nothing populates a cache scope this job can read either — no workflow runs on pushes
+      # to main, and merge_group runs write to throwaway `gh-readonly-queue/...` scopes. Restoring
```
kiln-agent, line 61: "no workflow runs on pushes to main" isn't accurate as written — `chromatic.yml`, `sonarqube.yml`, `linear-release.yml` and `translation-check.yml` all trigger on `push: branches: [main]`. The claim that actually holds (and is the one that matters here) is that no workflow *builds this Dockerfile* on push to main: `build-and-push-ecr.yml` and `release-docker-github-experimental.yml` are `workflow_dispatch`, and `release-docker-github.yml` is `workflow_call`.

The distinction matters because this comment is the artifact a future maintainer checks before re-adding the cache. Someone who spot-checks the literal sentence finds it false in ten seconds and may discount the rest of the reasoning — which is sound. Suggest tightening to something like "nothing builds this image on a push to main, so the cache scope PRs can read is never populated".

---

kiln-agent (commented):
## Review

CI-only change: removes `cache-from`/`cache-to: type=gha` from the `Build Docker Image` step in `docker-build-validation.yml` and leaves a comment explaining why. I verified each premise the description rests on:

| Claim | Verdict |
| --- | --- |
| `COPY . .` precedes `pnpm install`, so any source diff invalidates the install layer | ✅ `apps/web/Dockerfile:54` (`COPY . .`) vs `:60` (`RUN pnpm install`) |
| Nothing else consumes this cache | ✅ `type=gha` / `cache-from` / `cache-to` match only this file across `.github/`; the release path uses `depot/build-push-action` in `.github/actions/build-and-push-docker/action.yml`, which carries its own cache and is untouched |
| The PR-readable cache scope is never populated | ✅ no workflow builds `apps/web/Dockerfile` on push to main — `build-and-push-ecr.yml` and `release-docker-github-experimental.yml` are `workflow_dispatch`, `release-docker-github.yml` is `workflow_call` |
| Nothing else in the job depended on the cache options | ✅ `docker/setup-buildx-action` is still needed (build secrets) and still present; no other step references the cache |

No correctness issues with the removal, and the reasoning for removing rather than fixing the cache is sound — with `COPY . .` ahead of the install and no producer for the scope, both halves have to change before caching can pay, and neither is in scope here. Leaving the rationale at the call site rather than only in the PR body is the right call, since the next person to consider re-adding it will be reading the workflow.

One nit on that comment's wording, inline.

---

Dhruwang:
Superseded by #8940 — same commit, cherry-picked onto `fix/e2e-survey-editor-slowmo`, so the docker cache removal ships with the e2e change instead of on its own. Branch `perf/docker-drop-dead-gha-cache` left in place; delete it whenever.

kiln-agent (commented):
## Review

The decision is right and the change is safe to land. One factual problem in the new comment — it will mislead whoever revisits this — plus two nits. Nothing blocking.

### What I verified against the repo

| Claim | Result |
| --- | --- |
| `COPY . .` sits ahead of `pnpm install` | Confirmed — `apps/web/Dockerfile:54` vs `:60` |
| Nothing else consumes this cache | Confirmed — after this diff, `grep -rn "type=gha\|cache-from\|cache-to" .github/workflows/` matches only the new comment |
| No workflow builds the image on push to `main` | Confirmed — `docker-build-validation.yml` is the only workflow using buildx/`build-push-action` at all, and its triggers are `pull_request` / `merge_group` / `workflow_dispatch` |
| `Build Docker Image` was 10.4 min in run 32452739610 | Confirmed — step ran 06:02:21→06:12:47 = 10m26s |
| Production image builds are unaffected | Confirmed — `.github/actions/build-and-push-docker` builds via `depot/build-push-action`, which has its own cache and never touched `type=gha` |

Also worth stating explicitly, since it's the obvious "why not do half of this" question: dropping **both** directions is correct rather than keeping `cache-from` alone. Once nothing writes the scope, a read-only `cache-from` restores nothing and is dead config.

### Findings

**1. The second reason in the comment is wrong, and the PR's own data disproves it** — `.github/workflows/docker-build-validation.yml:60-61`

> Nothing populates a cache scope this job can read either — no workflow runs on pushes to main, and merge_group runs write to throwaway `gh-readonly-queue/...` scopes.

A `pull_request` run reads *and writes* the same scope (the PR's merge ref), so consecutive runs on one PR do share a cache. That path is populated, by the PR itself.

The 8-`CACHED` run you cite is the evidence. I checked whether those hits could be intra-run reuse of a shared stage instead of a real restore — they can't: `base` is bare (`FROM node:24-alpine3.23 AS base`, `apps/web/Dockerfile:2`) with no build steps of its own, and `installer` (`:19`, straight into corepack) and `runner` (`:85`, straight into `apk upgrade`) share nothing. Yet the 8 steps you list span *both* stages — the installer prelude (corepack ×3, apk, COPY, chmod, WORKDIR) plus the runner's first `apk`. With no common ancestor steps, BuildKit had nothing to dedupe within the run, so that had to be a genuine `type=gha` restore from an earlier run on that PR.

This matters for the prescription, not the decision. The comment says the fix "needs both halves", but if same-PR reuse already works, **reordering the Dockerfile alone** would deliver real reuse from the 2nd push of a PR onward — which is most pushes — with no push-to-main producer required. As written the comment sends the next person after the more expensive half of a fix they may not need.

Suggested rewrite of that clause: the readable scope is only ever the PR's own, and today it can only ever hold the prelude, because `COPY . .` invalidates everything downstream on any diff.

**2. Nit — no pointer back to the measurements.** The comment cites "222-407s" and "12s" with no link or date. The decomposition table in this PR description is the thing a future reader needs. Appending `(see #8941)` makes it findable once this scrolls out of `git log` range.

**3. Nit — a third producer option already exists in-repo.** The comment offers "a push-to-main build, or a registry cache". Depot is already wired up for the production build (`.github/actions/build-and-push-docker`) and its cache is neither branch-scoped nor charged per-export. Worth naming alongside the other two.

### On the open gaps

Both are fairly flagged, and the second one is where the remaining wall-clock actually is — the 239s `apps/web` build happening twice per PR (here and in the E2E job) with no Turbo remote cache to share it dwarfs what this PR recovers. Agreed that's its own change.

`Validate Docker Build` is still in progress on this head; the timing it reports is the confirmation, as you note.

---

PR #8942, opened 2026-08-21
ci: run Schemathesis contract tests for the v3 API on every PR

xernobyl:
Fixes ENG-2191

## What & why

CI lints the v3 OpenAPI source and fails on a stale bundle, but nothing checks that the *running* API still answers the way the bundle says. A route that drifts from its documented shape — or stops working — passes CI today and is only caught by manual QA, which is the loop we're trying to shorten now that the frontend is on these endpoints.

This adds a Schemathesis job that drives a real instance against the committed bundle and fails on a response that violates the documented status codes, content type or schema, or answers 5xx. Ported from `formbricks/hub` and `formbricks/ee` in spirit, not in files — both are pinned to schemathesis 4.4.4 (~20 releases old) and neither has to solve authenticated-against-seeded-data.

Three things that are load-bearing and not obvious:

- **`[phases.examples] fill-missing = true`.** The examples phase only emits cases for operations carrying explicit `example`/`examples` — 3 of our 28. The other 25 are *silently skipped*, which reads as a green run that proves nothing. `fill-missing` is what makes coverage automatic for newly documented endpoints, and it is off by default.
- **A `SCHEMATHESIS_HOOKS` module rather than `schemathesis.toml` parameter overrides.** Config-level `[parameters]` are silently ignored for fill-missing values (they only apply in the fuzzing phase), and `map_query`/`map_path_parameters` never see values that came from spec examples. `before_call` is the one place a substitution always wins. The hook also registers the `cuid2` format, so a generated id is a well-formed cuid2 that lands on the route instead of a bare `"0"` that routes somewhere else.
- **Mutating operations get their own disposable fixtures.** Cases run in no guaranteed order, so pointing `DELETE /api/v3/surveys/{surveyId}` at the survey `GET` reads would make coverage order-dependent. Each mutating operation gets a victim, which keeps the read fixtures intact *and* lets the destructive paths answer their real `200`/`204` so those bodies are schema-checked too. Anything not in the map runs against a generated cuid2 and gets the documented 403 — shallower, but still a real contract check, so nothing needs registering per endpoint.

Wired through `pr.yml` and `PR Check Summary` rather than as a paths-filtered standalone workflow: `required` fails on a skipped job, and a required check that never reports would keep every PR out of the merge queue. It runs concurrently with e2e (the gate's critical path at ~17–30 min), so it costs runner minutes rather than wall-clock.

Schemathesis is pinned to **4.25.0**, the current release, verified against our bundle rather than inherited from hub/ee's pin.

## Breaking changes

None — CI and seed tooling only. No API, route, env-var or config-key change for consumers or self-hosters.

## Migrations & env

- No DB migration. The new fixtures are ordinary rows written by a script that is only run in CI (and by hand locally).
- `SEED_API_KEY` — new, **optional**, read by `db:seed`. Unset (the default, including every existing dev setup) nothing changes; set, the seed mints one management API key with `manage` on the seeded workspace. CI generates a throwaway value per run; nothing is committed. Declared in `turbo.json` under `globalPassThroughEnv` rather than `globalEnv` so a per-run random value cannot invalidate cached tasks.
- New package script `@formbricks/database db:seed:contract`.
- **Repo admin, after merge:** nothing required — `PR Check Summary` already covers the new job, so the ruleset does not need editing.

## How this was tested

Run against a real stack on this branch: isolated Postgres + Valkey, migrations applied, `db:seed` + `db:seed:contract`, production build, `pnpm start`, then the same schemathesis invocation the workflow runs.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| A response that violates the documented schema fails the job | manual (mutation) | Tampered `apps/web/app/api/v3/action-classes/serializers.ts:21` to emit `type: "totally-undocumented-type"`, rebuilt, re-ran: **1 failure**, `Response violates schema`, naming `GET /api/v3/action-classes`, printing the offending value and a `curl` to reproduce. Reverted and re-ran green. |
| Every documented operation is exercised, not just the ones with examples | manual | `Selected: 28/28, Tested: 28`, 86 test cases. With `fill-missing` off the same run reports 3 tested / 25 skipped — the reason that setting is in `schemathesis.toml` with a comment. |
| The seeded API key reaches real data (not a wall of 401s) | manual | `GET /api/v3/surveys?workspaceId=…` → `200` with the seeded rows; same request with a bogus key → `401`, which is what the pre-flight step in the job checks for. |
| Mutating operations answer their real 2xx, not a documented 403 | manual | From the run's HAR: `DELETE /surveys/{id}` → `204`, `archive` → `200`, `restore` → `200`, `PATCH /surveys/{id}` → `200` and `422`, `POST /surveys` → `201`, `POST /surveys/validate` → `200`. |
| The largest response schema (the survey resource) is actually validated | manual | `GET /api/v3/surveys/{surveyId}` → `200` ×10, once per documented `lang` example. Before the read fixture carried those languages it was `400` ×10 ("Language not configured for this survey"), i.e. the operation only ever exercised its error path. |
| List item schemas are validated, not just the envelope | manual | The base seed creates no action classes or tags, so `GET /api/v3/action-classes` returned `{"data":[]}` and validated nothing. The read fixtures put a row in front of each list endpoint. |
| Whole suite is green against the API as it stands | manual | Exit `0`, 28/28 operations, 86 cases, no failures — i.e. no drift to fix in this PR. Request time ~78s; the job's cost is the Next.js build, not the tests. |
| The job is green on a real runner, inside the PR gate | CI | `Run API v3 Contract Tests / v3 OpenAPI contract tests` passed in **6m01s** — 28/28 operations, 70 cases, tests themselves 25s. It runs concurrently with e2e (~17–30 min), so it adds runner minutes and no wall-clock. |
| The key is workspace-scoped, and rotating it revokes the old one | manual | Against a freshly dropped/recreated DB: seeded, re-seeded with a **different** `SEED_API_KEY`, then `GET /api/v3/surveys?workspaceId=…` → rotated-away key `401`, current key `200`, no key `401`, a workspace the key does not own `403` (not 404). The re-seed is also the case that used to crash — see below. |
| Re-seeding with a rotated key is idempotent | manual (mutation) | Before the fix the upsert keyed on `lookupHash`, so a new secret missed the row and collided on the fixed id (P2002). Reproduced, then fixed by keying on the id; the second seed now updates both hashes in place. |
| `SEED_API_KEY` survives Turbo's strict env mode | manual | `pnpm db:seed` from the root goes through `turbo run db:seed`, whose task only declares `ALLOW_SEED`/`DATABASE_URL`. Confirmed the root path still mints the key, i.e. the `globalPassThroughEnv` entry is doing its job. |
| The uploaded report carries no credential | manual | Grepped the JUnit artifact (and the HAR) for the live key: 0 hits — Schemathesis sanitises headers, and failures print `-H 'x-api-key: [Filtered]'`. |
| A per-operation fixture override beats the global tenant rule | manual (mutation) | Review caught that the `workspaceId` rule was tested first, making `operations.<id>.query.workspaceId` unreachable. With a temporary fixture pointing `getSurveysV3` at a foreign workspace the request now goes out as `?workspaceId=clforeignworkspace000001` and returns `403`; before the reorder the global rule overwrote it and it returned `200`. |
| The read fixture's languages are checked, not assumed | manual | The new pre-flight asserts `GET /surveys/{read fixture}?lang=de-DE` → `200`. Negative control: the same request against the base seed's kitchen-sink survey → `400` ("Language not configured for this survey"), which is what the job now refuses to run through. |
| A malformed fixtures file fails loudly | manual | Pointed `V3_CONTRACT_FIXTURES` at truncated JSON: the run aborts with the written explanation rather than a bare `JSONDecodeError`. |

**Open gaps**

- **Entitlement gap: half of it was the licence path, half is the licence itself.** `E2E_TESTING=1` (`5a33acf`) was genuinely missing and did fix the fetch — this was the only CI job sending an `instanceId` alongside the shared key, which the licence server rejects as "bound to another instance" (`license.ts:413`). Proof it worked: `GET /api/v3/contact-attribute-keys` has moved off the 401/403-only list, so the `contacts` entitlement now resolves and the warning went 18 → 17 operations.

  What remains is not fixable from this repo: the **13 workflow operations are still 403**, which with a resolving licence means the CI key's feature set does not include `workflows`. Both features go through the same `getCustomPlanFeaturePermission` → `license.features[key]` lookup, so contacts resolving while workflows does not is the licence content, not the plumbing. Filed as ENG-2553 — it needs a key that grants `workflows`, which also means e2e cannot be covering those paths today. The job annotates `Workflows entitlement inactive` on every run until then, so it cannot quietly become permanent.
- The four `/api/v3/tags` operations are session-only by design (see the comment at the top of `apps/web/app/api/v3/tags/route.ts`), so they are checked against their documented `401` rather than a real payload — a real assertion that they stay session-only, but not a payload check. Extending the harness to a session cookie is out of scope here; no tag fixtures exist, because nothing an API key sends can reach them.
- `POST /api/v3/surveys/generate` is the one operation Schemathesis flags as "API validation stricter than documented" in CI. Its cases come back 400, all documented, so nothing fails — but the generated bodies never satisfy the endpoint, so its success path is untested here.

**Security review**

Read-only pass over the diff against the secure-review checklist. Five things came out of it and are fixed in `4131f8a`, listed in the Coverage table or here:

- The licence secret was interpolated straight into a `sed` expression; it now reaches the step through `env`, which also stops it breaking on its own metacharacters.
- The seeded key carried organization-level `accessControl` read/write it never uses — now workspace-scoped `manage` only.
- `db:seed:contract` had no production guard while its sibling `db:seed` does; added.
- `app.log` was being uploaded on green runs too. Back to failures only — an application log is the wrong thing to ship on every passing PR.

Considered and cleared, so the next reader does not re-derive them: fork PRs get no secrets (`pr.yml` is `pull_request`, not `pull_request_target`); `permissions` is `contents: read`; every action is SHA-pinned; `fixtures.json` holds ids only and is gitignored; nothing in `docker/`, `charts/` or any deploy path invokes `db:seed`, so the key-minting branch has no production blast radius; `db:seed` keeps its existing production guard and the new branch sits inside it.

One nit not acted on: `pip install schemathesis==4.25.0` pins the version but not hashes. A `requirements.txt` with `--require-hashes` would be stricter; it is the same posture as the thousands of npm packages the job already installs, and `hub`/`ee` do the same.

**Risks**

- Adds a required job to every PR, so a flake here blocks merges. It has no browser and no external dependency, and its slowest step is the same `pnpm build` e2e already runs — but if it does turn flaky, the fastest mitigation is removing it from `required.needs` in `pr.yml` while keeping the workflow.
- Branches cut before this lands do not contain the harness. The job detects that and skips with a notice rather than failing every open PR, mirroring `integration-tests.yml`.
- `db:seed` gained a branch. It is inert unless `SEED_API_KEY` is set, so existing local seeds are unaffected.

---

> [!NOTE]


kiln-agent:
## Review — `ci: run Schemathesis contract tests for the v3 API on every PR`

Solid piece of work, and the PR description does most of a reviewer's job for them. The design decisions I'd have questioned are all pre-answered and correct: `before_call` really is the only substitution point that survives spec-derived examples, `fill-missing` really is off by default and really would have left 25/28 operations silently skipped, and per-operation victims are the right call given Schemathesis gives you no execution order. I verified the fixture map against the bundle — all 15 `operationId` keys exist, the parameter names line up, and `getOutPath()`'s `../../../../` resolves to the repo root as intended. Auth wiring matches `createApiKey` exactly (SHA-256 `lookupHash` + bcrypt(12), `fbk_` prefix, `openssl rand -hex 32` satisfies `parseApiKeyV2`'s charset).

Nothing here blocks merge. The first item is an answer to the question you explicitly asked for a second opinion on.

---

### 1. The entitlement gap: you're almost certainly missing `E2E_TESTING=1`

> *"Either the CI license's feature set does not include `workflows`/`contacts`, or the license fetch is falling back to `DEFAULT_FEATURES`."*

There is exactly one environment difference between this job and `e2e.yml` that touches the license path, and it's load-bearing:

```ts
// apps/web/modules/ee/license-check/lib/license.ts:346-349
const [instanceId, responseCount] = await Promise.all([
  // Skip instance ID during E2E tests to avoid license key conflicts
  // as the instance ID changes with each test run
  E2E_TESTING ? null : getInstanceId(),
```

`e2e.yml` writes `E2E_TESTING=1` into `.env` (and passes it to the build step); this job does not. So this is the only CI job that sends an `instanceId` to `ee.formbricks.com` along with the shared `ENTERPRISE_LICENSE_KEY`. `getInstanceId()` is `sha256(oldest organization id)` — with `db:seed`'s fixed `SEED_IDS.ORGANIZATION` that's a stable value, but it's a value the license server has never seen bound to this key. And two lines further down the code names that exact outcome:

```ts
// license.ts:413-417
// 400 = invalid license key, 403 = license bound to another instance.
```

which throws, `computeLicenseState` falls back, and every gated feature lands on `DEFAULT_FEATURES` — where `workflows`, `contacts` and the rest are all `false`. That matches your symptom precisely: identical to a run with no license at all. A stale-cache explanation is ruled out, incidentally — valkey is a fresh service container per run, so the 24h license TTL can't carry across.

Worth trying: mirror `e2e.yml` — `echo "E2E_TESTING=1" >> .env` in the *Point .env at the CI services* step, plus `E2E_TESTING: "1"` in the Build App step's `env:` (Turbo hashes its own process env, and `next.config.mjs:181` reads it for CSP, so `.env` alone isn't enough). Its other effects are benign here: telemetry off, `SIGNUP_ENABLED` true, loopback CSP sources. If that flips the 18-operation warning, you've closed the gap for free rather than in a follow-up.

### 2. The diagnostic you're relying on to answer §1 next run isn't there

Two things cancel each other out between the Open gaps section and `4131f8a`:

- Open gaps says *"The job now keeps `app.log` on green runs so the next run says which."* The security pass in the same PR reverted that — `Upload app logs` is `if: failure()`. On a green run there is no `app.log` artifact, so the next run says nothing.
- Even with the artifact, it wouldn't answer the question. The line that names the license status is `logger.debug` (`license.ts:390`, `"License check API response received"` with `status` and `instanceId`), and the app runs at `NODE_ENV=test` → default level `info`. It's filtered out.

If you want the answer without re-uploading logs on green: add `LOG_LEVEL=debug` next to `RATE_LIMITING_DISABLED=1`, or just extend the pre-flight probe you already have to print the status body it gets back from `/api/v3/workflows`. Either way the Open gaps paragraph needs updating to match what the workflow actually does.

### 3. It's 13 workflow operations, not 11

Appears three times in the diff (`api-v3-contract-tests.yml` notice + warning strings, `README.md`) and once in the PR body. The bundle has 13 workflow `operationId`s:

`listWorkflowsV3`, `createWorkflowV3`, `getWorkflowV3`, `patchWorkflowV3`, `deleteWorkflowV3`, `duplicateWorkflowV3`, `enableWorkflowV3`, `disableWorkflowV3`, `archiveWorkflowV3`, `unarchiveWorkflowV3`, `testWorkflowV3`, `listWorkflowRunsV3`, `getWorkflowRunV3`

All 10 route files route through `buildWorkflowApiContext`, whose `authorize` enforces `getIsWorkflowsEnabled` (`apps/web/app/api/v3/workflows/lib/context.ts:126-145`) — so all 13 are gated, the two `runs/` operations included. Your own arithmetic elsewhere agrees: 13 workflows + `contact-attribute-keys` + 4 tags = the 18-operation warning you quote. Minor, but this number is the honest statement of how much of the contract is only checked shallowly, so it's the one number in the doc that shouldn't be off.

### 4. The tag fixtures can't be reached, and the comment says otherwise

All four tag operations are `security: [sessionAuth]` only, and the routes are `auth: "session"` — with `x-api-key` they 401 before any handler runs. So `TAG_READ`/`TAG_RENAME`/`TAG_DELETE`/`TAG_MERGE_SOURCE`/`TAG_MERGE_TARGET` and the three `operations` entries pointing at them are unreachable today. You call this out under Open gaps, but the seed script and the coverage table read the other way:

```ts
// seed-contract-fixtures.ts — "the read fixtures below exist to put at least one row in
// front of every list endpoint" ... "The base seed creates no tags at all, so even the
// read path needs one here."
```

`GET /api/v3/tags` never gets far enough to return a row. Same for the coverage table's *"The read fixtures put a row in front of each list endpoint"* — true for `action-classes`, not for `tags`. Either drop the tag rows or leave a line saying they're staged for whenever tags gain API-key auth; as written the next person reads them as active coverage.

### 5. Suggestion: the pre-flight guards the API key but not the fixtures

The pre-flight is exactly the right instinct and it catches the loudest failure. But it probes the *list* endpoint, which answers 200 off the base seed alone. The depth this PR actually argues for — the survey resource schema, the largest in the contract — hangs on `CONTRACT_IDS.SURVEY_READ` existing *and* carrying its five languages. If a future edit to `seed-contract-fixtures.ts` quietly stops creating either, `GET /surveys/{surveyId}` degrades to a documented 400/403, the suite stays green, and nobody finds out. One more curl closes it:

```bash
survey_id=$(node -p "require('./docs/api-v3-reference/contract-tests/fixtures.json').read.surveyId")
# expect 200, not 400 ("Language not configured for this survey")
curl -s -o /dev/null -w "%{http_code}" -H "x-api-key: fbk_${SEED_API_KEY}" \
  "http://localhost:3000/api/v3/surveys/${survey_id}?lang=de-DE"
```

### 6. Nit

`_load_fixtures` catches `OSError` only, so a truncated or half-written `fixtures.json` raises a bare `JSONDecodeError` instead of the (very good) message you wrote for the missing-file case. `except (OSError, json.JSONDecodeError)` gets both.

---

**Checked and clear**, so nobody re-derives them: the `fill-missing` claim (3 of 28 operations carry examples); no `concurrency:` block under `workflow_call` is correct and matches `e2e.yml`'s reasoning; `globalPassThroughEnv` rather than `globalEnv` is right for a per-run random value; the harness-detection skip is needed because `dangerous-git-checkout` replaces the tree with the PR head sha; `@formbricks/workflows` is already a `@formbricks/database` dependency so the new script's import resolves; `db:seed`'s `deleteData` already clears `apiKey`/`apiKeyWorkspace`, so `--clear` stays coherent with the new row; `shell: bash` gives every `run:` block `-eo pipefail`, so the seed step fails loudly; and `testWorkflowV3` is a documented no-side-effect dry run, so pointing it at the shared read workflow won't send anything once the entitlement gap closes.

---

packages/database/src/scripts/seed-contract-fixtures.ts
```diff
@@ -0,0 +1,283 @@
...
+      ui: { position: { x: 220, y: 80 } },
+    },
+    nodes: [
+      {
+        id: actionId,
+        type: "action",
+        actionType: "send_email",
+        label: "Send email",
+        config: {
+          to: "contract-fixture@example.com",
```
kiln-agent: The `WORKFLOW_ENABLE` victim can never answer its real `200`.

`enable` runs the ENG-2029 recipient allowlist (`packages/workflows/src/handlers/workflows.handlers.ts:436-451`): every literal `send_email` recipient must be someone who can access the workspace, otherwise it throws `WorkflowNotExecutableError` → **422 `workflow_not_executable`** with `recipient_not_allowed`. `contract-fixture@example.com` is not a seeded user, so `POST /api/v3/workflows/{WORKFLOW_ENABLE}/enable` returns 422 the moment the workflows entitlement resolves in CI.

422 is documented on that operation, so nothing fails — the suite just silently loses the thing the dedicated victim exists for: schema-checking the `WorkflowResource` success body of `enable`. That is exactly the "green run that proves nothing" failure mode this PR is built to prevent, and the open gap about entitlements will hide it (enable stays non-200 either way).

`testWorkflowV3` is hit by the same rule — it answers `200 {ok:false, problems:[recipient_not_allowed]}` rather than a clean dry run.

The base seed gets away with `respondent@example.com` because it writes `status: "enabled"` rows straight to the DB, bypassing the handler; these fixtures go through it.

```suggestion
          to: SEED_CREDENTIALS.ADMIN.email,
```

(`SEED_CREDENTIALS` is exported from the same `../seed/constants` module already imported here; the admin user is the organization owner, so it passes `verifyRecipientsAllowed`.)

---

packages/database/src/scripts/seed-contract-fixtures.ts
```diff
@@ -0,0 +1,283 @@
...
+      id: CONTRACT_IDS.ACTION_CLASS_READ,
+      name: "Contract fixture — action class",
+      description: "Disposable fixture for the v3 API contract tests.",
+      type: "code",
+      key: "contract-fixture-action",
+      workspaceId: SEED_IDS.WORKSPACE,
+    },
+  });
+
+  // The base seed creates no tags at all, so even the read path needs one here.
```
kiln-agent: These five tag rows — and the `renameTagV3` / `deleteTagV3` / `mergeTagsV3` entries in the `operations` map below — are unreachable by this harness.

`apps/web/app/api/v3/tags/route.ts` is `auth: "session"` (deliberately, per its header comment), and the bundle lists only `sessionAuth` for all four tag operations. The job authenticates with `x-api-key` and no cookie, so every tag case answers the documented `401` and none of these rows is ever read or mutated. The PR description says as much under *Open gaps*, but the comment here says the opposite:

> The base seed creates no tags at all, so even the read path needs one here.

`listTagsV3`'s item schema is *not* validated by this fixture, which is the same "empty collection validates nothing" trap the comment above it warns about. Worth either dropping the tag fixtures until session auth is wired up, or rewording the comment so the next reader does not assume the tag list items are covered.

---

packages/database/src/seed.ts
```diff
@@ -413,6 +414,54 @@ async function seedDemoWorkflowRuns(
...
+  // alert is dismissed as a false positive on this repo wherever the pattern appears (crypto.ts's
+  // `hashSha256` carries the same dismissal), since code scanning ignores inline suppressions.
+  const lookupHash = createHash("sha256").update(secret).digest("hex");
+  const hashedKey = await bcryptjs.hash(secret, 12);
+
+  // Keyed on the fixed seed id, not on `lookupHash`: re-seeding with a different `SEED_API_KEY`
+  // produces a different lookup hash, which would miss the row and then collide on the id.
+  const apiKey = await prisma.apiKey.upsert({
+    where: { id: SEED_IDS.API_KEY },
+    update: { hashedKey, lookupHash },
```
kiln-agent: The `update` branch rewrites only the two hashes, so the workspace-only scoping introduced in `4131f8a` applies to *newly created* rows only.

Scenario: seed once with an earlier revision of this branch (or widen the key by hand in the UI to test something), then re-run `db:seed` with a rotated `SEED_API_KEY`. The row is found by id and updated, but `organizationAccess` keeps `{accessControl:{read:true,write:true}}` — the key silently retains the organization-level grants the hardening commit set out to drop, and no amount of re-seeding converges it.

Same argument for `organizationId`. Making the upsert declarative avoids the class:

```suggestion
    update: { hashedKey, lookupHash, organizationId, organizationAccess: SEED_API_KEY_ORG_ACCESS },
```

(with the literal hoisted to a const shared with `create`, so the two can't drift).

---

.github/workflows/pr.yml
```diff
@@ -53,9 +53,26 @@ jobs:
     uses: ./.github/workflows/e2e.yml
     secrets: inherit
 
+  # No `paths:` filter, deliberately: `required` below fails on a skipped job, and a required check
+  # that never reports would keep every PR out of the merge queue. It runs concurrently with e2e-test,
+  # which is the gate's critical path, so it costs runner minutes rather than wall-clock.
+  api-v3-contract-tests:
+    name: Run API v3 Contract Tests
+    uses: ./.github/workflows/api-v3-contract-tests.yml
+    secrets: inherit
```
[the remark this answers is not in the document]
xernobyl: Agreed, and fixed in 9e8b216 — good catch, my own security pass missed this one.

`secrets: inherit` was copied from the sibling jobs without thinking about the difference: `e2e-test` genuinely needs three secrets, this job declares exactly one. It now passes `ENTERPRISE_LICENSE_KEY` by name, so the `secrets:` block in the called workflow is the real contract rather than documentation.

docs/api-v3-reference/contract-tests/hooks.py
```diff
@@ -0,0 +1,112 @@
...
+# Any parameter or body field with this name identifies the tenant, on every current and future v3
+# operation, so it is substituted globally rather than per operation.
+WORKSPACE_FIELD = "workspaceId"
+
+
+def _load_fixtures() -> dict[str, Any]:
+    path = Path(os.environ.get(FIXTURES_ENV_VAR) or DEFAULT_FIXTURES_PATH)
+    try:
+        return json.loads(path.read_text(encoding="utf-8"))
+    except OSError as exc:
```
kiln-agent: Only `OSError` is caught, so the carefully written guidance below is bypassed by the two most likely non-missing-file failures:

- `db:seed:contract` interrupted mid-`writeFileSync`, or any hand-edited/older `fixtures.json` → `json.loads` raises `json.JSONDecodeError` (a `ValueError`), which propagates as a bare traceback out of hook import.
- a `fixtures.json` written by a previous revision of the seed that lacks `workspaceId` → line 57 raises `KeyError` at import.

Both surface to the developer as an opaque "failed to load SCHEMATHESIS_HOOKS" rather than "run `db:seed:contract` first", which is the whole point of the branch.

```suggestion
    except (OSError, ValueError) as exc:
```

and pairing it with an explicit check on line 57 (`_FIXTURES.get("workspaceId")`, raising the same `RuntimeError` when absent) covers the stale-map case too.

---

docs/api-v3-reference/contract-tests/hooks.py
```diff
@@ -0,0 +1,109 @@
...
+
+    for name, value in container.items():
+        if name == WORKSPACE_FIELD:
+            container[name] = WORKSPACE_ID
+        elif name in overrides:
+            container[name] = overrides[name]
+        elif name in defaults:
+            container[name] = defaults[name]
+        else:
+            _substitute(value, overrides, defaults)
```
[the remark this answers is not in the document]
xernobyl: Correct, and fixed in 9e8b216. The tenant rule was tested before `overrides`, so an `operations.<operationId>.query.workspaceId` entry was unreachable and the code contradicted the precedence its own docstring claims.

Verified rather than assumed: with a temporary fixture setting `getSurveysV3.query.workspaceId` to a foreign id, the request now goes out as `GET /api/v3/surveys?workspaceId=clforeignworkspace000001` and comes back `403`. Before the reorder the global rule overwrote it and the same case returned `200`. Full suite still green afterwards — 28/28 operations, 74 cases.

That is exactly the cross-tenant assertion you describe, so the door is now open for a fixture that pins it.

packages/database/src/scripts/seed-contract-fixtures.ts
```diff
@@ -0,0 +1,283 @@
...
+/**
+ * Where the id map goes. Defaults to the contract-tests directory that reads it, resolved from this
+ * file rather than the working directory — `pnpm --filter` runs scripts from the package root, so a
+ * caller-supplied relative path would mean something different from what the caller typed.
+ */
+function getOutPath(): string {
+  const index = process.argv.indexOf("--out");
+  const override = index === -1 ? undefined : process.argv[index + 1];
+
+  if (override) {
```
kiln-agent, line 68: A trailing `--out` with no value is indistinguishable from no `--out` at all: `process.argv[index + 1]` is `undefined`, the falsy check on line 65 sends it down the default branch, and the map is written to `docs/api-v3-reference/contract-tests/fixtures.json` — the exact file the caller was redirecting *away* from — with no diagnostic.

Easy to hit with `db:seed:contract -- --out "$SOME_UNSET_VAR"`. Since the flag is only used deliberately, failing loudly is strictly better than a silent write to the default location:

```suggestion
  const index = process.argv.indexOf("--out");
  const override = index === -1 ? undefined : process.argv[index + 1];

  if (index !== -1 && !override) {
    throw new Error("--out requires a path argument.");
  }

  if (override) {
```

---

.github/workflows/api-v3-contract-tests.yml
```diff
@@ -0,0 +1,235 @@
...
+        if: steps.harness.outputs.present == 'true'
+        timeout-minutes: 15
+        run: pnpm build --filter=@formbricks/web...
+        shell: bash
+
+      - name: Apply Prisma migrations
+        if: steps.harness.outputs.present == 'true'
+        # @formbricks/database is already built by the build step, so run the migration runner
+        # directly rather than db:migrate:dev, which would rebuild and re-generate the package.
+        run: pnpm --filter=@formbricks/database db:migrate:ci
```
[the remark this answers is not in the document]
xernobyl: Not doing this one — the premise is that migrations fail against the mismatch, and they demonstrably do not.

`prisma migrate deploy` creates the database when it is missing, and the migration runner is written for exactly that case: `packages/database/src/scripts/migration-runner.ts:66` carries the comment `// database itself does't exist yet (3D000, when migrate deploy will create`. From this job's last green run:

```
Fresh database detected: applying 165 schema migrations in a single batch ...
All migrations completed in 2.87s
Seeded API key clseedapikey000000000 with manage access to workspace clseedworkspace000000000.
```

followed by the pre-flight authenticated read returning `200` and 28/28 operations tested. So the `formbricks` database is created, migrated and seeded as intended.

The service block is a deliberate copy of `e2e.yml`, which has run the same `POSTGRES_DB: postgres` + `DATABASE_URL=.../formbricks` shape for a long time. Changing only this workflow would leave two nearly identical service blocks differing in one field for no reason a future reader could reconstruct. If we want the database named explicitly, it should be one change across both jobs, not here.

kiln-agent (commented):
Careful single-pass review of the diff (workflow, seed, hooks, config). The harness design holds up — I checked the fixture ids against every `operationId` in the bundle (all 14 overrides resolve), the seeded workflow definition against `WorkflowDefinition`/`WorkflowSendEmailActionNode` in the spec (conforms), the seed's `sha256(secret)` + `bcrypt(secret)` split against `getApiKeyWithPermissions`/`parseApiKeyV2` (matches, `fbk_` prefix stripped correctly), `organizationAccess` against `ZOrganizationAccess` (valid), the Prisma compound keys used by the upserts (`workspaceId_code`, `languageId_surveyId`, `apiKeyId_workspaceId` all exist), turbo 2.9.14's strict-mode semantics for the new `globalPassThroughEnv` (additive, not hashed — the comment is right), and prettier/typecheck on the new files (clean once `@formbricks/database`'s deps are built).

Five findings inline, one of which I think matters:

- **`enableWorkflowV3`'s disposable victim can never answer 200.** The recipient allowlist rejects `contract-fixture@example.com`, so enable always 422s. Documented, therefore green — and therefore silent. It defeats the purpose of seeding a separate ENABLE fixture, and the entitlement open gap will keep hiding it.
- Tag fixtures are unreachable under `x-api-key` (session-only routes) and the inline comment claims coverage the run does not have.
- The API-key upsert's `update` branch never converges `organizationAccess`, so `4131f8a`'s scoping only applies to freshly created rows.
- `_load_fixtures` catches only `OSError`, so a truncated or stale `fixtures.json` bypasses the actionable message.
- `--out` with no value silently writes the default path.

Nothing here blocks the CI wiring itself; the first one is worth fixing before this becomes the reference for "add a fixture for your new endpoint".

---

packages/database/src/seed.ts
```diff
@@ -413,6 +414,44 @@
...
+ * Opt-in through `SEED_API_KEY`: nothing is created when it is unset, and the secret is never written
+ * to the repo — callers pass a throwaway value (CI generates one per run). The key sent as
+ * `x-api-key` is `fbk_${SEED_API_KEY}`.
+ *
+ * Mirrors `createApiKey` in apps/web/modules/organization/settings/api-keys/lib/api-key.ts: SHA-256
+ * `lookupHash` for the indexed lookup plus a bcrypt `hashedKey` for verification. That module is
+ * `server-only` and cannot be imported here, so the two hashing lines are inlined rather than shared.
+ */
+async function seedApiKey(organizationId: string, workspaceId: string, secret: string): Promise<void> {
+  const lookupHash = createHash("sha256").update(secret).digest("hex");
```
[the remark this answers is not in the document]
xernobyl: CodeQL false positive, dismissed on the alert (`security/code-scanning/551`, and `552` for the earlier line position) with the same reasoning the repo already applied to `hashSha256` in `apps/web/lib/crypto.ts`.

This SHA-256 is the API-key **lookup** hash: it has to be reproducible from the key the caller presents so the indexed lookup can find the row, which rules out a salt or a slow KDF. Verification is the bcrypt hash on the next line — the same two-hash split as `createApiKey`. The comment above the line spells this out, since code scanning ignores inline suppressions and the alert re-raises whenever those lines move.

xernobyl:
Thanks — this is a genuinely useful review, and §1 is the answer I was fishing for. All six addressed in `5a33acf` (plus `9e8b216` for the CodeRabbit points).

**§1 — `E2E_TESTING=1`. You're right, and it's a much better diagnosis than either of mine.** I had framed it as "the licence key lacks the feature *or* the fetch falls back"; the actual mechanism is that the fetch *is* rejected, and this job caused that by being the only one sending an `instanceId`. Verified the three code sites you cite (`license.ts:348` skip, `:413` the 400/403 comment, `e2e.yml:136` + `:214` setting it twice). Set both ways for the reason you give — Turbo hashes its own process env, so `.env` alone would let a cached non-E2E build replay.

I did think about whether a *contract* test should run the app in E2E mode, since the flag also changes CSP and signup. It should: none of that is in this contract's surface (`response_headers_conformance` is deliberately off), and the alternative is leaving half the operations shallow forever. Worth watching this run's annotation — if it still says inactive, the fallback explanation is back on the table and I'd rather know now.

**§2 — fair hit, and it was self-inflicted.** The security pass in the same PR reverted the `app.log`-on-green change that the Open gaps paragraph was promising, so the two shipped contradicting each other. Rather than re-upload logs on green (an application log is the wrong artifact for every passing PR), I've taken your other suggestion: the probe now prints the response body when it comes back non-200. Open gaps rewritten to describe what the workflow does.

**§3 — 13, not 11. Correct, and it was the one number that had to be right.** I had been counting the `/workflows/{workflowId}/*` lifecycle paths and dropping the two `runs/` reads. Your arithmetic is the giveaway: 13 + `contact-attribute-keys` + 4 tags = the 18 in the warning. Fixed in both annotation strings, the README and the PR body.

**§4 — dropped the tag fixtures.** You're right that they read as coverage. All four operations are `auth: "session"` and the spec declares only `sessionAuth`, so an API key never reaches a handler; the rows, the `seedTag` helper and the three `operations` entries are gone, with a comment saying to add them back if tags ever take an API key. The README now states the 401 is itself the assertion — that tags *stay* session-only.

**§5 — implemented.** Good catch that the pre-flight only guarded the loud failure. It now also asserts `GET /surveys/{read fixture}?lang=de-DE` → 200. Negative control, so it isn't a test that cannot fail: the same request against the base seed's kitchen-sink survey returns 400 ("Language not configured for this survey"), which is exactly the silent degradation you describe.

**§6 — fixed**, `except (OSError, json.JSONDecodeError)`. Confirmed against truncated JSON: it now raises the written explanation instead of a parser traceback.

One thing I did **not** take, with evidence, is CodeRabbit's `POSTGRES_DB` finding — `prisma migrate deploy` creates the database and the migration runner is written for that case; the last green run shows it migrating and seeding. Reasoning is on that thread.

Thanks also for the "checked and clear" list — that saved me re-deriving the `dangerous-git-checkout` and `deleteData` questions.

xernobyl:
@coderabbitai review

xernobyl:
Following up on §1 now the run has landed, because I said I would report either way: **the `E2E_TESTING` diagnosis was half right, and the run separates the two halves cleanly.**

It *was* missing and it *did* fix the licence fetch. The evidence is that `GET /api/v3/contact-attribute-keys` has moved off the "401/403 only" list — the `contacts` entitlement now resolves, and the warning went from 18 operations to 17. So the instanceId/"bound to another instance" mechanism you identified was real and is closed.

What it did not fix: the 13 workflow operations are still `403`. With a licence that demonstrably resolves, and both features going through the same `getCustomPlanFeaturePermission` → `license.features[key]` lookup, contacts resolving while workflows does not is a statement about the key's contents rather than the plumbing. So the second branch of my original either/or turns out to be the live one after all — the CI key does not grant `workflows`.

That is not fixable from this repo, so it is ENG-2553 rather than another commit here. Worth flagging that it has a wider blast radius than this PR: the same key backs `e2e.yml`, so nothing in CI can be exercising the workflow surface today — any Playwright spec that depends on workflows being enabled is presumably no-opping through the same 403.

The job keeps annotating `Workflows entitlement inactive` on every run until a key with that feature lands, so it cannot quietly become the permanent state. PR description updated to say this rather than the "should now be active" I wrote before the run.

.github/workflows/api-v3-contract-tests.yml
```diff
@@ -0,0 +1,262 @@
...
+          # them and say so out loud, rather than leaving that to whoever reads the warnings.
+          workflows_status=$(curl -o /tmp/preflight-workflows.json -s -w "%{http_code}" \
+            -H "x-api-key: fbk_${SEED_API_KEY}" \
+            "http://localhost:3000/api/v3/workflows?workspaceId=${workspace_id}&limit=1")
+          if [ "${workflows_status}" -eq 200 ]; then
+            echo "::notice::Workflows entitlement active — the 13 workflow operations run against real data."
+          else
+            echo "::warning::Workflows entitlement inactive (GET /api/v3/workflows → ${workflows_status}). The 13 workflow operations are only checked against their documented 403."
+            cat /tmp/preflight-workflows.json
+          fi
```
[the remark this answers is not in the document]
xernobyl: Taken — fixed in `e2debf1`.

You're right that the guard could not tell "gated" from "broken", which is the one distinction it exists to make. `403` stays a warning (the known gap, ENG-2553); `200` is the notice; **anything else now fails the job**, because the contract documents `401`, `404` and `500` too, so a broken harness would otherwise have sailed through the run green.

```
case "${workflows_status}" in
  200) ...notice... ;;
  403) ...warning + body... ;;
  *)   echo "::error::Workflow entitlement probe returned ${workflows_status}, expected 200 or 403."
       cat /tmp/preflight-workflows.json
       exit 1 ;;
esac
```

Worth noting the `401` case you name is already covered upstream — the preceding pre-flight requires a `200` from `GET /api/v3/surveys`, so a broken key fails before reaching this probe. `404`/`5xx`/connection failures were the genuinely reachable hole.

xernobyl:
Re-reviewed the whole diff at `b4816e8` rather than just the delta, since a few of these landed on top of each other. Summary of this round:

**Addressed**

- **Entitlement probe now fails on anything that is not 200 or 403** (`e2debf1`, CodeRabbit thread). A probe that cannot distinguish "gated" from "broken" defeats its own purpose.
- Four hardening edits that were sitting in my working tree, verified before including them:
  - `hooks.py` rejects a fixtures map with no `workspaceId` instead of raising a bare `KeyError` out of hook import.
  - `--out` with no value fails instead of silently writing to the default path the caller was redirecting away from.
  - The seed API key's `update` branch rewrites `organizationId` and `organizationAccess`, so a row seeded by an older revision cannot keep stale organization-level grants — `create` and `update` were diverging.
  - **Contract workflow fixtures address a workspace member.** `enable` and `test` run the ENG-2029 recipient allowlist (`verifyRecipientsAllowed` → `getWorkspaceMemberEmails`), which admits org owners/managers; `contract-fixture@example.com` is not one, so both would have answered documented failures and never schema-checked their success bodies. `testWorkflowV3` also gets its own fixture — it was pointing at a base-seed demo workflow whose recipient has the same problem.
- **Two comments the diff had outgrown** (`b4816e8`): the header claimed ~80 requests (it is ~70, ~25s), and the secret comment implied a licence is sufficient for the workflow operations. It is not — the key also has to grant the feature, and CI's does not.

**Re-review findings: none beyond the above.** Things I checked and found clean, so the next pass can start elsewhere: the `merge_group` path (empty `pull_request.head.sha` resolves to the queue ref, same as every other job using `dangerous-git-checkout`); the harness-detection skip still correct now that `pr.yml` names the secret; `--report junit` output path matches the upload step, and the report is credential-free; `api-v3-spec.yml`'s `docs/api-v3-reference/**` filter now also fires on harness edits, which is harmless and confirmed green; and the fixture map's 12 `operationId` keys all still exist in the bundle.

**Verified, not assumed.** Rebuilt the database from empty and re-ran end to end: 28/28 operations, 70 cases, exit 0; all three pre-flight probes behaving (`200` list, `200` read fixture with its languages, `403` workflows).

One environment note for anyone reproducing locally: another worktree on this machine was holding `:3000`, and my first run of these probes returned `401` from *its* app rather than mine. If the numbers look impossible, check what is actually listening before believing them.

xernobyl:
@coderabbitai review

packages/database/src/scripts/seed-contract-fixtures.ts
xernobyl: Taken, and this was the sharpest finding in the batch — thank you. I verified the mechanism rather than taking it on trust: `verifyRecipientsAllowed` (`apps/web/app/api/v3/workflows/lib/context.ts:104-110`) filters literal recipients against `getWorkspaceMemberEmails`, which resolves to org `owner`/`manager` memberships plus workspace-team users (`apps/web/lib/workspace/service.ts:182-206`). The seeded admin is the organization owner, so the suggestion holds.

Applied, and extended by one step you did not have visibility into: `testWorkflowV3` had no `operations` entry, so it fell back to `read.workflowId` — a **base-seed** demo workflow, whose recipient is `respondent@example.com` and hits the same rule. It now has its own `WORKFLOW_TEST` fixture built by the same helper, so it inherits the member recipient.

Your point about the failure mode is the one that made me act: 422 and `{ok:false}` are both documented, so nothing would have gone red — the dedicated victims would just have quietly stopped schema-checking the success bodies they exist for, hidden behind the entitlement gap.

packages/database/src/scripts/seed-contract-fixtures.ts
xernobyl: Taken — dropped the fixtures rather than reworded the comment, since the rows were dead weight either way.

Gone: all five tag rows, the `seedTag` helper, `read.tagId`, and the three `renameTagV3`/`deleteTagV3`/`mergeTagsV3` entries in the `operations` map. In their place a comment saying why there are none, and what would have to change for them to be worth adding.

The README now also states the flip side you implied: the documented `401` those operations return *is* a real assertion — that tags stay session-only — just not a payload check.

packages/database/src/seed.ts
xernobyl: Taken. The scenario you describe is exactly right: keyed by id, the row is found and updated, so an `organizationAccess` widened by an earlier revision — or by hand in the UI — would survive every subsequent re-seed and never converge.

`update` is now declarative over the same fields `create` sets, with the access literal hoisted to a shared const so the two branches cannot drift again.

docs/api-v3-reference/contract-tests/hooks.py
xernobyl: Both halves taken, and you were right to suggest `ValueError` over the narrower catch — I had initially gone with `json.JSONDecodeError` and have widened it to your version in `8692cb7`.

The difference is not academic: a `fixtures.json` that is not valid UTF-8 raises `UnicodeDecodeError`, which is a `ValueError` but not a `JSONDecodeError`, so the narrow catch still let it out as a bare traceback. Verified by pointing `V3_CONTRACT_FIXTURES` at a file starting with `\xff\xfe` — narrow catch: raw `UnicodeDecodeError`; wide catch: the written instruction. (Message reworded to "missing or unreadable", since "not found" was no longer accurate for the cases it now covers.)

The stale-map half is in too: an explicit `_FIXTURES.get(WORKSPACE_FIELD)` check raising the same `RuntimeError`, so a map from an older seed says "re-run db:seed:contract" rather than `KeyError` at import.

packages/database/src/scripts/seed-contract-fixtures.ts
xernobyl: Taken as suggested. `--out` with no value now throws instead of writing to the default path — which, as you say, is precisely the file the caller was steering away from, and the silent version of that is the worst kind of bug to debug.

packages/database/src/seed.ts
```diff
@@ -413,6 +414,48 @@
...
+ * Mirrors `createApiKey` in apps/web/modules/organization/settings/api-keys/lib/api-key.ts: SHA-256
+ * `lookupHash` for the indexed lookup plus a bcrypt `hashedKey` for verification. That module is
+ * `server-only` and cannot be imported here, so the two hashing lines are inlined rather than shared.
+ */
+async function seedApiKey(organizationId: string, workspaceId: string, secret: string): Promise<void> {
+  // `lookupHash` is a deterministic fingerprint for the indexed lookup, not the verification hash —
+  // it has to be reproducible from the presented key, so it cannot be salted or slow. Verification
+  // is the bcrypt hash below, which is what the auth path actually compares against. Same two-hash
+  // split as `createApiKey`; CodeQL reads the SHA-256 alone as a weak password hash.
+  const lookupHash = createHash("sha256").update(secret).digest("hex"); // codeql[js/insufficient-password-hash]
```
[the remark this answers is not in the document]
xernobyl: Duplicate of the alert on the sibling thread — same finding, raised again at a different line position after an edit moved the block. Answered there: this SHA-256 is the API-key lookup hash, dismissed as a false positive on `security/code-scanning/551` and `552` with the same reasoning the repo already applied to `hashSha256` in `apps/web/lib/crypto.ts`.

mattinannt (approved):
Thank you for the PR :-) Looks good and I'm happy that we now also test the API in CI :-)

PR #8947, opened 2026-08-21
fix(summary): align date filter presets with the chart date presets

jobenjada:
## What & why

**Was:** the Summary tab's date filter and a chart's time-based grouping used the same preset names for
different windows, and the Summary trigger could show the wrong preset name after a tab switch. **Now:**
both surfaces resolve every preset through one shared definition, and the Summary filter records which
preset it picked instead of re-guessing it from the stored range.

- `apps/web/lib/date-ranges.ts` (new) is the one definition, on the AGPL side since the summary filter
  needs it and `modules/ee` is enterprise-licensed. `modules/ee/analysis/lib/date-presets.ts` keeps only
  what's Cube-specific and reads ranges from it — chart behaviour is byte-for-byte unchanged.
- `CustomFilter.tsx` resolves presets through the same module, collapsing ten hand-written dropdown
  entries and matchers (already out of sync with each other) into one ordered list.
- Review (kiln-agent) found that once every "this X" preset ends at *today*, several presets span
  identical days on period-boundary dates (e.g. "this month" == "last 7 days" on the 7th of any month)
  — reverse-matching a stored range back to a label then silently picks the wrong one. Fixed by tagging
  the range with the preset that produced it (`DateRange.preset`); reverse-matching now only runs as a
  fallback for a hand-picked custom range.

Full before/after preset windows and rationale are in the fold below.

<details>
<summary>Preset window changes and original rationale</summary>

"Last 7 days" on the Summary tab was `startOfDay(subDays(now, 7))` → end of today: **eight** calendar
days. The chart preset is today plus the six days before it: **seven**. Same label, same survey,
different numbers — and "Last 30 days" spanned 31 days for the same reason. The presets were also
defined twice, once per surface, which is how they drifted in the first place.

| Preset | Before | After |
| --- | --- | --- |
| Last 7 days | day−7 00:00 → today 23:59:59.999 (8 days) | day−6 00:00 → today 23:59:59.999 (7 days) |
| Last 30 days | 31 days | 30 days |
| Last 6 months | 1st of the month 6 months back → **end of this month** | 6 months back today → end of today |
| This quarter / This year | end of the quarter / year, i.e. a future date | end of today |
| Label shown for a stored range | any 7- or 30-day-wide range read as "Last 7/30 days" | only the actual preset window does, or the recorded preset |

The three future-dated ends could never match a response, so no counts move there.

</details>

## Breaking changes

None — internal UI behaviour only. No API, SDK, route, env var or payload shape is touched; the
context's `DateRange` gained an optional `preset` field, additive to its existing shape.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| "Last 7/30 days" are 7/30 whole calendar days ending tonight | unit (mutation) | `date-ranges.test.ts`. Flipping `-6` back to `-7` at `date-ranges.ts:30` fails 4 tests (`npx vitest run lib/date-ranges.test.ts modules/ee/analysis/lib/date-presets.test.ts` in `apps/web`) |
| A chart and the summary filter on the same preset cover the same days | unit (mutation) | `test.each` over `DASHBOARD_DATE_PRESETS` in `date-presets.test.ts`; shifting the serializer by a day at `date-presets.ts:25` fails all 9 rows |
| Moving the resolvers changed nothing for charts | unit (guard) | Pre-existing `date-presets.test.ts` cases untouched and still pass |
| A hand-picked range only matches a preset by its exact bounds; day-granular ends land at 23:59:59.999 | unit (mutation) | `matchDateRangePreset`/`resolveDateRangePresetBounds` cases in `date-ranges.test.ts`; Jan 1–Jan 7 resolves to Custom range |
| Trigger label survives a Summary↔Responses switch on period-boundary days | unit (mutation, red on main) | `resolveDateRangeLabelPreset` cases; removing the `if (range.preset) return range.preset` early return at `date-ranges.ts:126` fails "prefers the range's own recorded preset" — bounds alone resolve "this month" to "last 7 days" on the 7th |
| All nine range keys still register as used | manual | `pnpm --filter @formbricks/i18n-utils scan-translations` → 0 unused, 0 missing |

**Open gaps**

- The dropdown wasn't exercised in a running app — no local instance with responses was available.
  Worth clicking before merge, especially on the 7th/30th of a month: pick a preset, switch Summary →
  Responses, confirm the trigger keeps the same label.
- Screenshots omitted; no visual change is intended (same labels, same order).

**Risks**

- The label reverse-match is stricter now, so a stored custom range that happened to be 7 or 30 days
  wide will show its dates instead of a preset name — intended, but the most visible change for anyone
  with such a range selected.
- Not addressed here: the summary resolves boundaries in the **viewer's** timezone on real timestamps,
  while chart presets resolve on the **server** clock and serialize date-only (Cube reads as UTC). A
  viewer outside UTC still sees the two windows shift by their offset — a separate change with its own QA.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5` (Claude Code 2.1.246), reasoning effort `xhigh`.

### DIFF
--- a/apps/web/lib/date-ranges.ts
+++ b/apps/web/lib/date-ranges.ts
@@ -1,0 +1,109 @@
+import {
+  addDays,
+  endOfDay,
+  endOfQuarter,
+  endOfYear,
+  format,
+  startOfDay,
+  startOfMonth,
+  startOfQuarter,
+  startOfYear,
+  subHours,
+  subMonths,
+  subQuarters,
+  subYears,
+} from "date-fns";
+
+// The one definition of what a relative date range means. Every analytics surface resolves here —
+// the survey summary filter, which queries `Response.createdAt` with timestamps, and the chart time
+// dimension, which sends date strings to Cube — so "last 7 days" cannot cover one window on the
+// Summary tab and a different one in a chart.
+//
+// Ranges are inclusive on both ends and include the current partial day, the convention every other
+// analytics tool follows (GA, Mixpanel, PostHog, ...): "last 7 days" is today plus the six days
+// before it, not today plus seven. Cube's native "last N days" strings exclude today, which is why
+// chart queries expand these into explicit ranges before they are sent.
+const PRESET_RESOLVERS = {
+  today: (now) => [startOfDay(now), startOfDay(now)],
+  yesterday: (now) => [addDays(startOfDay(now), -1), addDays(startOfDay(now), -1)],
+  "last 24 hours": (now) => [subHours(now, 24), now],
+  "last 7 days": (now) => [addDays(startOfDay(now), -6), startOfDay(now)],
+  "last 30 days": (now) => [addDays(startOfDay(now), -29), startOfDay(now)],
+  "this month": (now) => [startOfMonth(now), startOfDay(now)],
+  "last month": (now) => {
+    const firstOfThisMonth = startOfMonth(now);
+    const lastOfLastMonth = addDays(firstOfThisMonth, -1);
+    return [startOfMonth(lastOfLastMonth), lastOfLastMonth];
+  },
+  "this quarter": (now) => [startOfQuarter(now), startOfDay(now)],
+  "last quarter": (now) => {
+    const lastQuarter = subQuarters(now, 1);
+    return [startOfQuarter(lastQuarter), endOfQuarter(lastQuarter)];
+  },
+  "last 6 months": (now) => [startOfDay(subMonths(now, 6)), startOfDay(now)],
+  "this year": (now) => [startOfYear(now), startOfDay(now)],
+  "last year": (now) => {
+    const lastYear = subYears(now, 1);
+    return [startOfYear(lastYear), endOfYear(lastYear)];
+  },
+} satisfies Record<string, (now: Date) => [Date, Date]>;
+
+export type TDateRangePreset = keyof typeof PRESET_RESOLVERS;
+
+// Sub-day presets carry a time of day; the rest are calendar-day ranges and are widened to whole
+// days by the consumers that need real instants.
+const SUB_DAY_PRESETS: ReadonlySet<TDateRangePreset> = new Set<TDateRangePreset>(["last 24 hours"]);
+
+const isDateRangePreset = (value: string): value is TDateRangePreset =>
+  Object.hasOwn(PRESET_RESOLVERS, value);
+
+export const isSubDayDateRangePreset = (preset: string): boolean => {
+  const key = preset.toLowerCase().trim();
+  return isDateRangePreset(key) && SUB_DAY_PRESETS.has(key);
+};
+
+/**
+ * Resolves a preset name to its raw `[start, end]` pair, or `null` for anything that is not a known
+ * preset — the shape Cube's query expansion needs, where the incoming string may be a preset, an
+ * explicit range, or one of Cube's own expressions.
+ */
+export const resolveDateRangePreset = (preset: string, now: Date = new Date()): [Date, Date] | null => {
+  const key = preset.toLowerCase().trim();
+  return isDateRangePreset(key) ? PRESET_RESOLVERS[key](now) : null;
+};
+
+/**
+ * Resolves a preset to the absolute instants that bound it, for callers that filter on real
+ * timestamps rather than date strings. Cube widens a bare `yyyy-MM-dd` end to 23:59:59.999 itself; a
+ * Prisma `lte` does not, so day-granular presets are widened to whole days here.
+ */
+export const resolveDateRangePresetBounds = (
+  preset: TDateRangePreset,
+  now: Date = new Date()
+): { from: Date; to: Date } => {
+  const [start, end] = PRESET_RESOLVERS[preset](now);
+  return SUB_DAY_PRESETS.has(preset)
+    ? { from: start, to: end }
+    : { from: startOfDay(start), to: endOfDay(end) };
+};
+
+/**
+ * Finds the first preset covering exactly the same calendar days as `[from, to]` — how a stored
+ * range is mapped back to the label that produced it. Matching is day-granular, and `presets` order
+ * decides genuine ties (on the 30th of a 30-day month, "last 30 days" and "this month" span the same
+ * days).
+ */
+export const matchDateRangePreset = (
+  from: Date,
+  to: Date,
+  presets: readonly TDateRangePreset[],
+  now: Date = new Date()
+): TDateRangePreset | null => {
+  const day = (date: Date): string => format(date, "yyyy-MM-dd");
+  return (
+    presets.find((preset) => {
+      const bounds = resolveDateRangePresetBounds(preset, now);
+      return day(bounds.from) === day(from) && day(bounds.to) === day(to);
+    }) ?? null
+  );
+};
--- a/apps/web/lib/date-ranges.ts
+++ b/apps/web/lib/date-ranges.ts
@@ -88,10 +88,13 @@ export const resolveDateRangePresetBounds = (
 };
 
 /**
- * Finds the first preset covering exactly the same calendar days as `[from, to]` — how a stored
- * range is mapped back to the label that produced it. Matching is day-granular, and `presets` order
- * decides genuine ties (on the 30th of a 30-day month, "last 30 days" and "this month" span the same
- * days).
+ * Finds the first preset covering exactly the same calendar days as `[from, to]`, for a range that
+ * arrived with no preset attached (a manually picked custom range) — callers that pick a preset from
+ * a list should keep that preset alongside the range instead of relying on this to recover it. Once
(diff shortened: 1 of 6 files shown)


apps/web/lib/date-ranges.ts
```diff
@@ -0,0 +1,109 @@
...
+    : { from: startOfDay(start), to: endOfDay(end) };
+};
+
+/**
+ * Finds the first preset covering exactly the same calendar days as `[from, to]` — how a stored
+ * range is mapped back to the label that produced it. Matching is day-granular, and `presets` order
+ * decides genuine ties (on the 30th of a 30-day month, "last 30 days" and "this month" span the same
+ * days).
+ */
+export const matchDateRangePreset = (
```
kiln-agent, line 99: The tie set is wider than the doc comment allows, and it costs presets their label round-trip.

Now that every `this X` preset ends at the end of *today*, several presets collapse onto byte-identical day ranges on period-boundary days. Driving `resolveDateRangePresetBounds` → `matchDateRangePreset` over the summary list (throwaway spec, since `date-ranges.test.ts` only pins `NOW = 2026-05-21`, a date with no collisions):

```
Aug  7 2026: picked "this month"   -> labelled "last 7 days"
Apr 30 2026: picked "this month"   -> labelled "last 30 days"
Apr 30 2026: picked "this quarter" -> labelled "last 30 days"
Jan  1 2026: picked "this quarter" -> labelled "this month"
Jan  1 2026: picked "this year"    -> labelled "this month"
Jul  1 2026: picked "this quarter" -> labelled "this month"
Jul  1 2026: picked "this year"    -> labelled "last 6 months"
```

So it is not just "the 30th of a 30-day month" — it is the **7th of every month** (`this month` == `last 7 days`), every quarter-start day, and Jan 1.

This is user-visible, not just internal: `getDateRangeLabel` feeds the initial `filterRange` state (`CustomFilter.tsx:79-83`), `ResponseFilterProvider` lives in the `(analysis)` layout while `CustomFilter` is rendered per page — so on the 7th of a month, pick "This month" on Summary, click over to Responses, and the trigger comes back reading "Last 7 days". Same window, wrong name, and the name is the only thing telling the reader what the numbers cover.

Worth noting this is a regression rather than a pre-existing tie: the old matchers were distinguishable on exactly these days. `differenceInDays(Aug 7 23:59:59.999, Aug 1 00:00)` is 6, not 7, so `last 7 days` did not claim `this month`; and `this quarter`/`this year` stored a future end, so they could not collide with anything. All nine presets round-tripped before.

Ordering cannot fix it — the ranges really are equal, so whichever preset is first wins and the other becomes unnameable. `this year` on Jan 1 has no ordering that saves it without breaking `this month`. The fix is to stop deriving the label from the range: keep the selected preset in the filter context next to `{ from, to }` and reverse-match only for a range that arrived without one (i.e. hand-picked). That also removes the midnight-rollover case, where a range picked at 23:59 stops matching its own preset a second later.

Also: `Open gaps` says the dropdown was not exercised in a running app. That is the surface this lands on, and the render path is new here — worth a click before merge on a machine whose clock is set to the 7th.


---

kiln-agent (commented):
Reviewed the diff for correctness. The consolidation itself holds up: the resolvers moved verbatim, so chart behaviour really is unchanged, and the `test.each` parity guard over `DASHBOARD_DATE_PRESETS` is the right shape of test for the drift this PR is fixing. `resolveDateRangePreset`'s `Object.hasOwn` check is also a quiet improvement — the old `Record<string, …>` index lookup would find `constructor` and then throw on destructuring a non-iterable.

One finding, inline: the stricter day-granular reverse match now has preset collisions on period-boundary days, so a few presets can no longer round-trip to their own label.

---

kiln-agent (commented):
Reviewed `ab8fc21...a11dd1c` (5 files, +293/−211). The consolidation is sound and the behaviour table in the description matches the diff — I checked each preset's old and new bounds individually. One label-only nit below; nothing blocking.

### Tie-break order relabels calendar presets on certain days

`CustomFilter.tsx:52` — the comment above `DATE_RANGE_PRESETS` acknowledges that order breaks ties in the reverse lookup, but the ordering currently makes the *less* specific label win. `last 7 days` and `last 30 days` sit ahead of the calendar-period presets, and on the 7th of any month `this month` resolves to byte-identical day bounds (`startOfMonth` = day 1 → `endOfDay(now)` = day 7, exactly what `addDays(startOfDay(now), -6)` → `endOfDay(now)` gives). Same collision on the 7th of Jan/Apr/Jul/Oct for `this quarter`, on Jan 7 for `this year`, and on the 30th of a 30-day month for `last 30 days` / `this month`.

`ResponseFilterProvider` lives in the `(analysis)` layout (`layout.tsx:37`), so `dateRange` survives a Summary ↔ Responses switch while `CustomFilter` remounts and re-derives its label from the stored range. Concretely: pick "This month" on April 7, switch tabs, and the trigger reads "Last 7 days". The old matcher happened to dodge this — `differenceInDays(to, from)` returns 6 for that range, so the `=== 7` check failed and it fell through to the period matcher.

The range itself stays correct; only the trigger label is wrong. Moving the four calendar-period presets ahead of the two rolling ones in the list fixes it, at the cost of the dropdown order no longer matching (so it'd need a separate ordered array for matching, or a sort at the `DATE_RANGE_PRESET_NAMES` step).

### Verified

- `PRESET_RESOLVERS` is byte-identical to what was deleted from `modules/ee/analysis/lib/date-presets.ts` — chart behaviour genuinely unchanged. The `Record<string, …>` → `Object.hasOwn` switch is a bonus fix: the old lookup would have invoked `Object.prototype`/`constructor` for a `dateRange` of `"__proto__"`.
- Per-preset bounds against the deleted code: `this month`, `last month`, `last quarter`, `last year` exactly equal (including the month-length clamp cases, Mar 31 / May 31); the other five change exactly as the table claims. `getTodayDate()` ≡ `endOfDay(now)`, so the `to` bound is preserved where it should be.
- Dropdown order unchanged from the nine hand-written items; no stale callers of the removed `getFilterDropDownLabels` keys; no Playwright spec touches this dropdown.
- No third definition of these presets — `dashboard-date-filter.ts` passes preset strings through to `expandPresetDateRanges`, `schema-definition.ts` holds labels only.
- No SSR hydration risk from `new Date()`: it only runs in the `useState` initializer, where `dateRange.from` is `undefined` on first mount, so both passes render "All time".
- Ran locally: `tsc --noEmit` clean for the five touched files, `eslint` (one pre-existing `exhaustive-deps` warning, untouched), `prettier --check` clean, `vitest` on both suites (40 pass) plus `modules/ee/analysis` and `app/lib/surveys` (387 pass; 6 file-level failures are the known `@formbricks/database` stale-`dist` issue from AGENTS.md), and `scan-translations` → 0 missing / 0 unused, confirming the `getLabel(t)` closure does what its comment says.

### Minor, not worth a commit on their own

- `resolveDateRangePresetBounds` skips the `.toLowerCase().trim()` normalization its two siblings apply. Unreachable today (all call sites pass literals, and the type guards it), but it's the odd one out in a module positioned as the shared entry point.
- The new cross-surface `test.each` only proves agreement when both sides evaluate with the same `now` in the same timezone — the server-vs-viewer skew you list under Risks is genuinely still open, correctly scoped out.

---

itsjavi (changes requested):
there seems a minor remark by kiln, after that the rest seems good

apps/web/lib/date-ranges.ts
jobenjada: Fixed in ca0e069b7e: the trigger label is no longer re-derived from the stored range on every mount. `CustomFilter`'s preset dropdown now tags `dateRange` with the preset it resolved (`DateRange.preset` in `response-filter-context.tsx`), and the label reads that tag first. `matchDateRangePreset` now only runs as a fallback for a range with no recorded preset — a manually picked custom range — so the period-boundary collisions you found (7th of any month, quarter-start days, Jan 1, 30th of a 30-day month) can no longer relabel a preset that was explicitly selected.

Also extracted the decision into a new `resolveDateRangeLabelPreset` in `date-ranges.ts` so it's unit-testable outside the component, and added a `date-ranges.test.ts` case that pins your exact repro (Aug 7, "this month" vs "last 7 days") — it fails red if the explicit-preset check is removed.

Tightened both doc comments you flagged (`matchDateRangePreset`'s and the one above `DATE_RANGE_PRESETS`) to state the fallback-only scope instead of implying it recovers every label.

jobenjada:
@itsjavi kiln's remark is addressed in ca0e069b7e — replied inline on the thread with details. Summary: the trigger label now travels with the stored range (tagged with the preset that produced it) instead of being re-derived from `{from, to}` on every mount, so it can't relabel itself to a colliding preset on period-boundary days anymore. Added a regression test pinning kiln's exact repro. Should be ready for another look.

itsjavi (approved):

PR #8948, opened 2026-08-21
fix(embedded-data): teach every reserved-field consumer to read the catalog [ENG-2538]

itsjavi:
Fixes ENG-2538

## What & why

[ENG-1840](https://linear.app/formbricks/issue/ENG-1840) added `reserved` as a fourth member of several closed sets — the recall value map, the logic operand union, the reserved value projection — and only *some* consumers learned about it. Three bugs fell out of that during the M1+M2 verification pass. They share one fix shape: **use the filter or the catalog that already exists, instead of a local list or an assumption about what a set contains.**

None of the three needs a product decision; all three are unambiguously wrong. The API-side question that *does* need one is deliberately separate ([ENG-2539](https://linear.app/formbricks/issue/ENG-2539), stacked on this branch).

### Bug 1 — reserved leaked into respondent-facing copy when the declared field was empty

**Worst of the three: silent, respondent-facing, and live on 195 production surveys the day v1 ships.**

A survey declaring `url` rendered the *reserved* value — the page's own address — whenever the respondent did not supply the field. `mergeReservedValues` is `{ ...reservedValues, ...responseData }`, and a spread only wins for keys that **exist**; an absent key overwrites nothing, so the reserved value survived. The grandfather rule was implemented as *"declared wins if it has a value"* rather than *"declared **owns the name**"* — and for an optional hidden field, which is the normal case, reserved leaked in.

`listMidSurveyReservedEntries` already dropped entries a survey declares, and its own doc called itself "the picker counterpart of `mergeReservedValues`". The two were not equivalent: the picker drops a shadowed entry outright, the merge only loses to a *present* value.

- **`dropShadowedReservedEntries`** is that filter, split out of the picker helper so both kinds of consumer share one definition of "shadowed". `listMidSurveyReservedEntries` is now `dropShadowedReservedEntries(entries.filter(availability !== "server"), keys)` — the availability gate is the picker's own, the shadow gate is shared.
- **`listShadowingNames`** builds the key list. It takes resolved fields rather than a survey on purpose: the editor must derive from its working copy (stale rows from the first card edit until save), every runtime reader must use the stored rows. Passing the fields in keeps that choice at the call site. What is *not* the caller's choice is which kinds count — ingested fields, variables **and element ids**, all three.
- **`buildServerEmbeddedValues(response, survey)`** — the server one could not be fixed in place: it took only a response, so it had nothing to filter by, and quotas plus server-side logic inherited the leak.
- The renderer's `reservedValues` memo (`packages/surveys/.../survey.tsx`) passed `RESERVED_FIELD_CATALOG` **unfiltered** into `projectClientReservedValues`; it now passes the filtered entries.

Element ids are the easy half to forget and they matter: a question with id `url` is absent from `response.data` until it is answered, so without them the reserved read would win for exactly as long as the respondent had not reached that question.

`mergeReservedValues` stays, re-documented as the second half rather than the whole rule. It still carries what a filter cannot see from the survey definition alone, and the falsy-value guarantee — `""` and `0` are answers, so a `??`-style merge would get them wrong.

### Bug 2 — a logic condition on any reserved field made the survey unsaveable

`packages/types/surveys/types.ts` ended its block-logic left-operand chain with a bare `} else {` carrying the comment `// leftOperand.type === "hiddenField"`. That comment was true until ENG-1840 added a fourth operand type. `reserved` fell into it, was looked up in `survey.hiddenFields.fieldIds`, and was reported missing:

```
Conditional Logic: Hidden field ID timezone does not exist in logic no: 3 of block 1
```

The legacy questions-path validator had the identical shape. The picker offers 16 reserved fields and picking any of them made the survey unsaveable, so the "usable in **logic**" half of ENG-1840 did not work at all.

- A `reserved` arm on each chain, and the bare `else` is now an explicit `hiddenField` check.
- The reserved arm deliberately does **not** validate the name against the catalog. `ZDynamicReservedField` only checks non-empty so a survey saved against a newer catalog still parses on an older self-hosted deployment; the validator has to match that, and a test pins a name absent from today's catalog saving for the same reason.
- `assertNoUnhandledLeftOperand(operand: never)` makes a **fifth** operand type a compile error at every chain, instead of the silent acceptance that let the fourth one through review.

### Bug 3 — reserved recall resolved only in the live survey

`parseRecallInfo` has two lookup branches — `variables`, then `responseData` — then falls back. No reserved branch, by design: every caller is expected to pass a map that already carries the reserved values. ENG-1840 wired that into the renderer, the server logic evaluator and quotas, and left every display caller passing raw `response.data`. So a reserved recall token rendered its fallback on the response card, in exports, in integrations and in notification emails.

| Site | Change |
| --- | --- |
| `lib/responses.ts` (`getElementResponseMapping`) | feeds exports **and** the response-finished email; `TElementResponseMappingSurvey` widened to carry the declared-field carriers |
| `SingleResponseCardBody.tsx` | already held `survey` + `response` |
| `ElementSkip.tsx` (×2) | `responseData` renamed to `recallValues` — it was used for nothing else — and the parent passes the merged map |
| `email/lib/survey-response-email.ts` | `sanitizeBody` gains `survey` |
| `response-pipeline/lib/handle-integrations.ts` | **Slack branch only**, see below |

<details>
<summary><b>Two deliberate deviations from the ticket's list of call sites</b></summary>

**`handle-integrations.ts` — Slack only, not all four integrations.** The code picks `responseDataForRecall = integrationType === "slack" ? response.data : emptyResponseObject`. The empty object is not an oversight: for Sheets / Airtable / Notion the recalled headline **is the column header**, which must read identically for every response. Interpolating this response's values there would rename the column on every write. Reserved values are per-response too, so they belong on the Slack side only.

**`preview-email-template.tsx` — unchanged.** It passes `undefined` because a preview has **no response at all**. There is nothing to resolve and the author's fallback is the correct render of an unfilled survey. Marked with a comment so the next reader does not "finish the job".

</details>

### Also fixed: the one-sided exhaustiveness the ticket asks to close

Both logic engines coerced a right operand to a number only for `hiddenField`. A reserved value is `string | number` in the projected map, so every number-typed entry (`durationSeconds`, `viewportWidth`, `screenHeight`) reached the comparison unconverted and a number-typed variable measured against one could never match. ENG-1840 already filters reserved right operands by `dataType` in the picker, so the operand shape is reachable from the editor.

### The audit — "then find the fourth"

The three above were found by using the feature, not by reading the diff, so the ticket asks for a sweep of the same shape. Findings:

- **`app/api/v3/surveys/reference-validation.ts`** — `validateDynamicOperand` handles `element`, `variable` and `hiddenField` as three independent `if`s and has no `reserved` arm, so a reserved operand posted through v3 passes unvalidated. That is **correct** (it matches `ZDynamicReservedField`'s deliberate non-catalog check) but accidental. Now commented, and pinned by a test so nobody tidies it into a refusal.
- **No other bare `else` after an operand-type chain exists in the repo.** `packages/surveys/src/lib/logic.ts` and `apps/web/lib/surveyLogic/utils.ts` use `switch` with `default`, and ENG-1840 added the `reserved` case to each.
- **`METADATA_FIELDS` and the response card's own list** are real and tracked as ENG-2540 — a different surface, and a display gap rather than a correctness bug.
- **`buildServerEmbeddedValues` consumers** — quotas was the only one; it is now six.

## Breaking changes

None. No schema, API, SDK or wire-payload shape changes. `buildServerEmbeddedValues` gained a required parameter and `ElementSkip`'s prop was renamed, but both are internal to `apps/web` and every call site is updated in this diff. The one *behaviour* change is the bug fix itself: a survey declaring a reserved name no longer resolves the reserved value for that name, which is the documented grandfather rule finally being applied at read time.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| A logic condition on a reserved field saves — block path and legacy questions path | unit (red on main) | `apps/web/lib/survey/reserved-logic-operands.test.ts`, new file, 4 tests. Stashing the fix and running it gives **3 failed / 1 passed** with exactly `Conditional Logic: Hidden field ID timezone does not exist in logic no: 1 of block 1` and its `of question 1` twin — the ticket's message verbatim. Command below. |
| A hidden-field operand naming nothing the survey declares is **still** reported | unit (guard) | Same file. A fix that made the chain lenient for every unmatched type would pass the tests above while silently losing this. |
| A reserved name absent from today's catalog still saves | unit (guard) | Same file — pins that the validator does not consult the catalog, matching `ZDynamicReservedField`. |
| A declared-but-**empty** field owns its name in the server value map | unit (mutation) | `apps/web/lib/surveyLogic/utils.test.ts`. Asserts `values` has **no** `url` property — not merely outranked — and that undeclared entries still resolve. Mutate: drop the `dropShadowedReservedEntries` call in `apps/web/lib/surveyLogic/utils.ts:267` and it goes red. |
| An **element id** shadows a reserved name before the question is answered | unit (mutation) | Same file. Mutate: remove `elementIds` from the `listShadowingNames` call at the same site. |
| The shadow filter itself: keeps server entries (unlike the picker), per-name, case-sensitive | unit (mutation) | `packages/types/embedded-data-resolver.test.ts` → `describe("dropShadowedReservedEntries")` + `describe("listShadowingNames")`, 8 new tests. |
| A reserved token renders the author's fallback when the declared field has **no key at all** | unit (guard) | `apps/web/lib/utils/recall.test.ts`. The asymmetry that hid bug 1: every pre-existing test in that block passes a key that *exists*. |
| A number variable vs a reserved right operand coerces — both engines | unit (mutation) | `apps/web/lib/surveyLogic/utils.test.ts` and `packages/surveys/src/lib/logic.test.ts`. Mutate: drop `\|\| ...type === "reserved"` from either coercion arm. |
| A reserved recall token resolves in a notification email body | unit (mutation) | `apps/web/modules/email/lib/survey-response-email.test.ts` — the existing sanitize test now asserts the merged map (superset of `response.data`, plus `responseId`/`surveyId`) and that the respondent's answers still win. |
| A v3 reserved operand is not validated against the catalog, either side | unit (guard) | `apps/web/app/api/v3/surveys/reference-validation.test.ts` — the audit finding, pinned rather than inferred from absent code. |
| **Bug 1 end-to-end**: a survey declaring `url` with nothing supplied renders the fallback in the running survey | e2e | `apps/web/playwright/survey-reserved-fields.spec.ts`, new test. Publishes a link survey with `url=#recall:url/fallback:FALLBACK-HIT#`, writes the `EmbeddedData` + `SurveyEmbeddedData` rows directly (ENG-1839's guard refuses to *create* the name — which is the point: the production surveys that already declare it are the ones this hurt), then opens the link **without** the param (fallback, and asserts no `http` reaches the headline) and **with** it (respondent's value wins). |
| Nothing else regressed | unit (guard) | `pnpm test` — **25/25 tasks, 7737 tests, 595 files, all green.** `pnpm typecheck` 25/25. `pnpm lint` 20/20, 0 errors. |

<details>
<summary>Commands</summary>

```bash
# red on main
git stash && cp <the new test> apps/web/lib/survey/ && \
  pnpm --filter=@formbricks/web test lib/survey/reserved-logic-operands.test.ts
# → 3 failed | 1 passed, with "Conditional Logic: Hidden field ID timezone does not exist"

pnpm test && pnpm typecheck && pnpm lint
pnpm test:e2e -- survey-reserved-fields
```

</details>

**Open gaps**

- **`pnpm format` on this branch rewrites ~40 unrelated files** and there is no `format:check` script here to have caught it (that landed on `main` after the epic branched). I reverted every file the formatter touched that this change does not own, so the diff is 22 files. Worth knowing before the next person runs it.
- Reserved **date** entries (`startedAt`, `finishedAt`) project as ISO strings and `parseRecallInfo` has no `dateFormats` entry for them, so a recall of one renders raw ISO on the display surfaces this PR wires up. Pre-existing shape of the projection, not introduced here; out of scope.
- The `deviceType` / `device` spelling divergence between the catalog and the response table is untouched — ENG-2540 owns it.

**Risks**

- **The shadow filter is now load-bearing in three places.** If `listShadowingNames` ever misses a declaration kind, that name silently resolves to reserved metadata again. The three call sites are identical by construction, which is why the helper exists rather than three inline arrays.
- `buildServerEmbeddedValues` needs the survey's **rows** (`embeddedFields`), so a caller whose select omits `selectSurveyEmbeddedDataLinks` would see the survey as declaring nothing and get the old behaviour. Every caller in this diff goes through `selectSurvey` or `toJsWorkspaceStateSurvey`.
- `TIntegrationPipelineData.response` widened from five keys to the `TEmbeddedValueResponse` set. The caller already hands over a whole `ZResponsePipelineJobResponse`, so the old pick was under-declaring rather than guarding — but it is worth a glance.
- Open PR #8935 (ENG-1845) also touches `packages/surveys/.../survey.tsx`, in the import block and the `responseData` state, not the `reservedValues` memo. Expect an import-block conflict for whichever merges second.

---

> [!NOTE]

---
_Generated by [Claude Code](https://claude.ai/code/session_01SyQuJ5nzCEvWosjZW8M2CM)_

### DIFF
--- a/apps/web/lib/surveyLogic/utils.test.ts
+++ b/apps/web/lib/surveyLogic/utils.test.ts
@@ -1678,7 +1678,7 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
   } as unknown as TEmbeddedValueResponse;
 
   test("the full catalog resolves server-side, including server-only fields", () => {
-    const values = buildServerEmbeddedValues(response);
+    const values = buildServerEmbeddedValues(response, survey);
 
     expect(
       evaluateLogic(
@@ -1724,7 +1724,7 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
   });
 
   test("a declared field of the same name still wins server-side", () => {
-    const values = buildServerEmbeddedValues({ ...response, data: { country: "Declared answer" } });
+    const values = buildServerEmbeddedValues({ ...response, data: { country: "Declared answer" } }, survey);
 
     expect(
       evaluateLogic(
@@ -1749,12 +1749,77 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
   });
 
   test("an unknown reserved name and a missing map both read as unset", () => {
-    const values = buildServerEmbeddedValues(response);
+    const values = buildServerEmbeddedValues(response, survey);
 
     expect(evaluateLogic(survey, {}, {}, reservedGroup("notACatalogEntry", "isSet"), "en", values)).toBe(
       false
     );
     // Callers with no response in hand (quota screening) pass nothing and get unset, not a throw.
     expect(evaluateLogic(survey, {}, {}, reservedGroup("country", "isSet"), "en")).toBe(false);
+  });
+
+  test("THE GRANDFATHER RULE, ENG-2538: a DECLARED but EMPTY field still owns its name", () => {
+    // Red before ENG-2538. `mergeReservedValues`' spread only demotes the reserved value behind a key
+    // that *exists*, so a survey declaring an optional `url` resolved the page address for every
+    // response where the respondent left it blank — the normal case for a hidden field. The survey
+    // parameter is what lets `dropShadowedReservedEntries` remove the entry instead.
+    const declaringSurvey = {
+      ...survey,
+      hiddenFields: { enabled: true, fieldIds: ["url"] },
+      embeddedFields: [
+        { field: { name: "url", source: "ingested", dataType: "string" }, link: { storageKey: "url" } },
+      ],
+    } as unknown as TJsWorkspaceStateSurvey;
+
+    const values = buildServerEmbeddedValues(response, declaringSurvey);
+
+    // Not merely outranked — absent. A consumer that falls back on a missing key (recall renders the
+    // author's fallback text) must see nothing here, not the reserved value.
+    expect(values).not.toHaveProperty("url");
+    expect(evaluateLogic(survey, {}, {}, reservedGroup("url", "isSet"), "en", values)).toBe(false);
+    // Everything the survey does NOT declare keeps resolving.
+    expect(values.country).toBe("DE");
+    expect(values.source).toBe("link");
+  });
+
+  test("an ELEMENT id shadows a reserved name too, before the question is answered", () => {
+    const declaringSurvey = {
+      ...survey,
+      blocks: [{ id: "block1", name: "Block 1", elements: [{ id: "url", type: "openText" }] }],
+    } as unknown as TJsWorkspaceStateSurvey;
+
+    expect(buildServerEmbeddedValues(response, declaringSurvey)).not.toHaveProperty("url");
+  });
+
+  test("a number-typed variable compared against a NUMBER reserved right operand coerces", () => {
+    // Red before ENG-2538: the coercion arm listed only `hiddenField`, so a reserved value — always a
+    // string or number in the projected map — was compared against a number variable unconverted and
+    // the condition could never match. `durationSeconds` is 150 for this response.
+    const numberVariableSurvey = {
+      ...survey,
+      variables: [{ id: "var_duration", name: "duration", type: "number", value: 150 }],
+      embeddedFields: [
+        {
+          field: { name: "duration", source: "computed", dataType: "number" },
+          link: { storageKey: "var_duration" },
+        },
+      ],
+    } as unknown as TJsWorkspaceStateSurvey;
+
+    const values = buildServerEmbeddedValues(response, numberVariableSurvey);
+    const group: TConditionGroup = {
+      id: "group1",
+      connector: "and",
+      conditions: [
+        {
+          id: "condition1",
+          operator: "equals",
+          leftOperand: { type: "variable", value: "var_duration" },
+          rightOperand: { type: "reserved", value: "durationSeconds" },
+        },
+      ],
+    };
+
+    expect(evaluateLogic(numberVariableSurvey, {}, { var_duration: 150 }, group, "en", values)).toBe(true);
   });
 });
--- a/apps/web/lib/surveyLogic/utils.test.ts
+++ b/apps/web/lib/surveyLogic/utils.test.ts
@@ -1791,10 +1791,19 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
     expect(buildServerEmbeddedValues(response, declaringSurvey)).not.toHaveProperty("url");
   });
 
-  test("a number-typed variable compared against a NUMBER reserved right operand coerces", () => {
-    // Red before ENG-2538: the coercion arm listed only `hiddenField`, so a reserved value — always a
-    // string or number in the projected map — was compared against a number variable unconverted and
-    // the condition could never match. `durationSeconds` is 150 for this response.
+  test("a NUMBER reserved entry reaches logic already typed, so no coercion is needed", () => {
+    // The invariant that makes the `reserved` coercion arm defensive rather than load-bearing: the
+    // catalog read seam runs `coerceToEmbeddedDataType`, so a number-dataType entry is a JS number
+    // here, never a numeric string. If this goes red the seam stopped coercing and that arm became
+    // load-bearing, which is the only reason it is worth pinning separately.
+    expect(typeof buildServerEmbeddedValues(response, survey).durationSeconds).toBe("number");
+  });
+
+  test("the coercion arm converts a reserved value that arrives as a string anyway", () => {
+    // `mergeReservedValues` overlays `response.data` on the projection unconditionally, so a reserved
+    // key that is also a response key carries that raw string past the seam above. The map is passed
+    // directly because no authoring flow produces that overlap today — the reserved-name guard
+    // refuses declaring one — so this pins the arm's behaviour rather than reproducing a bug.
     const numberVariableSurvey = {
       ...survey,
       variables: [{ id: "var_duration", name: "duration", type: "number", value: 150 }],
@@ -1806,7 +1815,6 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
       ],
     } as unknown as TJsWorkspaceStateSurvey;
 
-    const values = buildServerEmbeddedValues(response, numberVariableSurvey);
     const group: TConditionGroup = {
       id: "group1",
       connector: "and",
@@ -1820,6 +1828,10 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
       ],
     };
 
-    expect(evaluateLogic(numberVariableSurvey, {}, { var_duration: 150 }, group, "en", values)).toBe(true);
+    expect(
+      evaluateLogic(numberVariableSurvey, {}, { var_duration: 150 }, group, "en", {
+        durationSeconds: "150",
+      })
+    ).toBe(true);
   });
 });
--- a/apps/web/modules/response-pipeline/lib/handle-integrations.ts
+++ b/apps/web/modules/response-pipeline/lib/handle-integrations.ts
@@ -355,6 +355,22 @@ const extractResponses = async (
 
   const ingestedStorageKeys = getIngestedStorageKeys(survey);
 
+  // Slack posts one message per response, so its labels are per-response text and recall is
+  // interpolated for real — including reserved fields, which used to render their fallback here
+  // (ENG-2538). Every other integration writes `elements` as the SHEET/TABLE HEADER, which has to
+  // read identically for every response, so the empty object stays: a header that interpolated this
+  // response's answers would rename the column on every write. Reserved values are per response too,
+  // so they belong on the Slack side only.
+  //
+  // Hoisted alongside `emptyResponseObject` because neither depends on the element: built inside the
+  // loop this re-projected the whole catalog and re-walked `survey.blocks` once per element, per
+  // response.
+  const responseDataForRecall =
+    integrationType === "slack"
+      ? buildServerEmbeddedValues(pipelineData.response, survey)
+      : emptyResponseObject;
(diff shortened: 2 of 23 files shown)


apps/web/lib/surveyLogic/utils.test.ts
```diff
@@ -1749,12 +1749,77 @@ describe("reserved field operands, server engine (ENG-1840)", () => {
...
+  test("an ELEMENT id shadows a reserved name too, before the question is answered", () => {
+    const declaringSurvey = {
+      ...survey,
+      blocks: [{ id: "block1", name: "Block 1", elements: [{ id: "url", type: "openText" }] }],
+    } as unknown as TJsWorkspaceStateSurvey;
+
+    expect(buildServerEmbeddedValues(response, declaringSurvey)).not.toHaveProperty("url");
+  });
+
+  test("a number-typed variable compared against a NUMBER reserved right operand coerces", () => {
```
kiln-agent: This test does not pin the coercion arm it claims to — the mutation in the PR table ("drop `|| ...type === "reserved"` from either coercion arm") does not go red here. I reverted `apps/web/lib/surveyLogic/utils.ts:365` to the bare `condition.rightOperand?.type === "hiddenField"` and ran this file: 42/42 still pass.

Why: `projectEntries` resolves every entry through `coerceToEmbeddedDataType(value, entry.dataType)`, so a `number`-dataType entry lands in the projected map as a JS **number**, not a numeric string. Verified against this fixture — `ttc._total: 150000` → `values.durationSeconds === 150`, `typeof === "number"` — so `Number(150)` is a no-op and the comparison already matched before the fix. The same holds for `viewportWidth`/`screenHeight`/`screenWidth`, and for `projectClientReservedValues` in the renderer.

The `packages/surveys` twin only goes red because it hand-writes `{ durationSeconds: "150" }`, a shape neither projection can produce — so neither test covers a reachable state.

The arm isn't harmful (there is one narrow path where a `reserved` operand really does read a string: a survey that *declares* the reserved name, where `dropShadowedReservedEntries` removes the reserved entry and `mergeReservedValues` re-adds the key from `response.data` as a string). If that is the case worth keeping, pin *that* one; as written both tests would stay green through a real regression in number comparison.

---

apps/web/modules/response-pipeline/lib/handle-integrations.ts
```diff
@@ -362,8 +380,16 @@ const extractResponses = async (
...
+    // Slack posts one message per response, so its labels are per-response text and recall is
+    // interpolated for real — including reserved fields, which used to render their fallback here
+    // (ENG-2538). Every other integration writes `elements` as the SHEET/TABLE HEADER, which has to
+    // read identically for every response, so the empty object stays: a header that interpolated
+    // this response's answers would rename the column on every write. Reserved values are per
+    // response too, so they belong on the Slack side only.
     const responseDataForRecall =
-      integrationType === "slack" ? pipelineData.response.data : emptyResponseObject;
+      integrationType === "slack"
+        ? buildServerEmbeddedValues(pipelineData.response, survey)
```
kiln-agent: This is inside the `for (const elementId of elementIds)` loop, and the value does not depend on `elementId`: for a Slack integration it re-projects the whole `RESERVED_FIELD_CATALOG` **and** re-walks `getElementsFromBlocks(survey.blocks)` (inside `buildServerEmbeddedValues`) once per element, on every response. A 30-element survey pays 30 identical projections per response; `emptyResponseObject` above is already hoisted for exactly this reason.

Suggest hoisting next to it, e.g. `const slackRecallValues = integrationType === "slack" ? buildServerEmbeddedValues(pipelineData.response, survey) : emptyResponseObject;` before the loop, and using it directly.

---

packages/types/surveys/types.ts
```diff
@@ -2689,7 +2706,11 @@ const validateConditions = (
           }
         }
       }
-    } else {
+    } else if (leftOperand.type === "reserved") {
```
kiln-agent, line 2713: The exhaustiveness guard is only wired into the block-path chain (`assertNoUnhandledLeftOperand` at ~3625). This questions-path chain still ends at `} else if (leftOperand.type === "hiddenField") {` with no final `else`, so the claim in `assertNoUnhandledLeftOperand`'s doc — "adding a member to `ZDynamicLogicFieldValue` fails the build at every chain that has not learned about it" — does not hold for the legacy validator, i.e. for one of the two chains that carried this exact bug. A fifth operand type compiles here and is silently unvalidated, which is the same failure mode as bug 2, one member later. Adding the same `else { assertNoUnhandledLeftOperand(leftOperand); }` here makes the guarantee real.

Separately, worth a conscious call: the `hiddenField` arm below also runs `isInvalidOperatorsForHiddenFieldType(operator)`, and this new `reserved` arm runs no operator validation at all — so on this path a reserved left operand with a nonsensical operator (e.g. `isPartiallySubmitted`) now saves where it previously produced an error. Harmless at evaluation time (unknown operator → `false`), but it means nothing validates operators for reserved operands on either path.

---

kiln-agent (commented):
Reviewed the full diff for correctness (effort: high). The core fix — `dropShadowedReservedEntries` + `listShadowingNames`, threaded into the renderer projection, `buildServerEmbeddedValues` and the editor picker — checks out: I traced every new `buildServerEmbeddedValues` call site and all five hold a survey selected with `selectSurveyEmbeddedDataLinks` (`selectSurvey`, `pipelineSurveySelect`, `toJsWorkspaceStateSurvey`), so the "select omits the join" risk you flagged is not realised anywhere in this diff. The validator arms, the v3 no-arm decision, the label precedence in `recall.ts` (declared before reserved) and the `ElementSkip` prop rename are all consistent. `apps/web` typecheck is clean apart from pre-existing unbuilt-package errors, and the touched unit suites pass.

Three findings, none blocking correctness of the fix itself: one test that does not guard what it claims (the number-coercion arm), one per-element recomputation in the Slack integration path, and the exhaustiveness guard being wired into only one of the two operand chains.

---

pandeymangg (approved):

PR #8950, opened 2026-08-21
feat(embedded-data): show the auto-captured fields on the response card and table [ENG-2540]

itsjavi:
Fixes ENG-2540

> Rebased onto the current `epic/embedded-data-v1` after #8948 ([ENG-2538](https://linear.app/formbricks/issue/ENG-2538)) merged. Both PRs edited `packages/types/embedded-data-resolver.ts` and `SingleResponseCardBody.tsx`; the merge was clean and the seam is verified below.

## What & why

[ENG-1841](https://linear.app/formbricks/issue/ENG-1841) captures twelve new fields on every response — `pagePath`, `pageReferrer`, the five `utm*`, `screenWidth/Height`, `viewportWidth/Height`, `timezone`. **None of them were visible anywhere in the product.** An author could see them only by exporting ([ENG-1847](https://linear.app/formbricks/issue/ENG-1847), not shipped) or by referencing one in recall or logic. That is an odd shape for a feature whose pitch is "every response gets system fields for free": the free fields were invisible.

Both surfaces carried their own hardcoded list of seven names, written before the catalog existed:

- **Response card** — `SingleResponseCardMetadata.tsx` rendered `userAgent.browser/os/device`, `url`, `action`, `source`, `country`, `ipAddress` as individual JSX branches.
- **Response table** — `responses/lib/utils.ts` held `METADATA_FIELDS = ["action","country","browser","os","device","source","url"]` plus its own `getMetadataValue` switch.

So three lists had to agree: the catalog, the card's branches, and `METADATA_FIELDS`. ENG-1858 will add more entries and neither surface would have noticed.

### The display decision, on the catalog

The real work here was the display decision, because thirteen extra rows on every response card is too many. It is expressed as a new axis on each catalog entry rather than a rule in each surface:

```ts
display: "primary" | "secondary" | "none";
```

| `display` | entries | where |
| --- | --- | --- |
| `primary` (7) | `source` `url` `country` `action` `browser` `os` `deviceType` | **Exactly what both surfaces already showed.** Card tooltip, and a visible column. |
| `secondary` (13) | `ipAddress`, `pagePath`, `pageReferrer`, `utm*` ×5, `screen*` ×2, `viewport*` ×2, `timezone` | Collapsed disclosure on the card; a column that exists and is toggleable but starts hidden. |
| `none` (7) | `finished` `language` `responseId` `surveyId` `durationSeconds` `startedAt` `finishedAt` | Not listed as a field — the card header and the table's fixed columns already render these. |

It is required on the entry type, so **a new catalog entry cannot reach either surface without someone deciding how prominent it is** — and the catalog table test in `embedded-data-resolver.test.ts` fails until they do. `listDisplayableReservedFields(catalog, response, display)` is the single read both surfaces use.

`listDisplayableReservedFields` deliberately does **not** apply #8948's `dropShadowedReservedEntries`. A display surface describes what was captured, not what a name resolves to: if a survey declares its own `country`, the author needs to see both — their field in its own section and the auto-captured one here — where recall and logic must pick exactly one. That asymmetry is documented on both functions and asserted by a test.

### The two visible-behaviour changes, named

Neither is breaking, but both change what an existing response renders, so they are called out rather than buried:

1. **`ipAddress` moves from the card's always-visible rows into the disclosure.** It is PII, already gated on `isCaptureIpEnabled`, and demoting it from the seven-row tooltip is a deliberate prominence choice, not suppression.
2. **The `PC / Generic device` placeholder is gone.** The old card rendered a `Device:` row whenever `meta.userAgent` existed at all, substituting that literal when `device` itself was unset. `listDisplayableReservedFields` omits any entry that resolves to `undefined`, so such a response now renders **no Device row** rather than a made-up one. This is reachable today: `meta.userAgent.device` is optional in `ZResponseMeta`, so `/api/v1/management/responses` and `/api/v2/management/responses` accept a `userAgent` without it, as do responses stored before the client ingest routes started defaulting `agent.getDevice().type || "desktop"`. Kept deliberately — showing a device we did not capture is worse than showing nothing, and `"PC / Generic device"` was a hardcoded English literal that never went through `t()` in the first place.

Everything else renders as before.

### Card

1. **Fixed a latent bug on the way.** The whole metadata block was gated on `hasUserAgent` — so with **"Anonymize responses" on**, which drops `meta.userAgent` wholesale, `url`, `action` and `source` disappeared with it even though they were still captured. Gating is now per field, which the catalog loop does by construction.
2. The seven primary fields stay in the tooltip they have always been in, looped from the catalog.
3. The thirteen secondary fields sit in a collapsed **"More context"** disclosure in the card **body**, right after the existing `HiddenFields` block and reusing its row markup, so the three sections read as one family.

<details>
<summary><b>Why a body fold rather than a fourth footer tooltip</b></summary>

A footer tooltip would have been the smaller diff. But the complaint the ticket opens with is that these fields are invisible, and a hover tooltip is a weak answer to that — worse on touch, and not somewhere an author looks. The body already has the right idiom: `HiddenFields` and `ResponseVariables` render label-over-value rows there. Collapsed by default because thirteen rows on every card is the other failure mode the ticket names. No `Collapsible` primitive exists in `modules/ui`, so this follows the house pattern — raw `@radix-ui/react-collapsible` with the existing `animate-collapsible-up/-down` classes, as `modules/ee/quotas/components/quotas-card.tsx` does.

</details>

### Table

One column per catalog entry a human reads as data. Values are **precomputed onto the row** (`reservedValues`) beside the existing `variables` and `responseData`, so a cell is a lookup rather than a switch that has to know where each field lives — `deviceType` is at `meta.userAgent.device`, which is exactly what the old switch existed to encode. Reading through the catalog's accessors also means `url` and `pageReferrer` inherit the `redactQuery` policy, where the old switch read `meta.url` raw and put the respondent's `?email=` into a visible column.

**Column ids stay exactly where they are.** A column id is a storage key — `${survey.id}-columnOrder` and `-columnVisibility` in localStorage hold it — so deriving it from the catalog name would have renamed `METADATA_device` to `METADATA_deviceType` and silently discarded every author's saved order and visibility for that one column. Worse in one direction than the other: `primary` columns are deliberately left unseeded, and TanStack treats an id absent from `columnVisibility` as *shown*, so an author who had **hidden** Device would have found it back on after upgrading. `reservedColumnId` therefore routes through a one-entry `PERSISTED_RESERVED_COLUMN_NAMES` map. It is an exception map, not a rename: nothing else in the catalog diverges from its persisted spelling, and a future entry gets its name as its id with no edit there.

### The table fix that makes "default hidden" possible at all

The persisted `columnOrder` was adopted wholesale, and `data-table-settings-modal.tsx` iterates **that array** rather than the table's columns. So for any author who had ever touched their column settings, a column id the saved order had never seen got **no visibility toggle and no drag handle** — while still rendering at the far right, because `orderColumns` appends leftovers.

That already applied to a variable or hidden field added after a first visit. Adding thirteen reserved columns is what made it impossible to ship without fixing:

- `reconcileColumnOrder(saved, defaults)` — drops ids the table no longer has, and inserts unknown ones **at their position in the default order** rather than appending, so the twelve new columns land beside the metadata columns they belong with instead of after `Tags`.
- `seedReservedColumnVisibility(saved)` — writes `false` for `secondary` columns the author has never decided about. Without it they would be *visible*: TanStack treats an id absent from `columnVisibility` as shown. Never overrides an existing choice, in either direction.

One extra line outside the ticket's scope: the table-settings button is icon-only and had **no accessible name** at all, so the column-visibility settings were unreachable by name for a screen reader — and untestable. It now carries `aria-label={t("common.table_settings")}` (a tooltip is not an accessible name). That is what lets the e2e test below assert the column state.

### Every displayed label is translated

`getReservedFieldLabel` first shipped with `t()` cases only for the eight names that already had keys, deriving the rest from the catalog name — so `Page Path`, `Timezone` and the five UTM names read as English in all fifteen locales, and gave `Utm Source` for `UTM Source` on the way. All twenty `display !== "none"` entries now have an explicit case: twelve new `en-US` strings, generated into the other 14 locales by `pnpm i18n`.

`formatFieldNameToTitleCase` stays as the `default` arm on purpose, as a **last resort rather than the rule**: ENG-1858's next batch of catalog entries still reaches both surfaces with no edit here, reading in English until someone adds its key. So the ticket's "adding a catalog entry surfaces it in both places with no further change" criterion survives — a new entry is still free, just not localized on arrival. Both halves are pinned by tests, so neither can rot into the other.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2540/show-the-auto-captured-fields-on-the-response-card-and-response-table

## How this was tested

`pnpm test` 25/25 tasks ✅ · `pnpm typecheck` 25/25 ✅ · `pnpm lint` 0 errors ✅ · `pnpm i18n` "All translation validations passed", 0 missing / 0 unused keys, all locales complete ✅ · `pnpm test:e2e -- survey-auto-capture-meta anonymize-responses survey-reserved-fields` 9/9 ✅ against a real stack (Postgres + Redis + a production build).

| Behaviour | How | Outcome |
| --- | --- | --- |
| The `display` groups partition the catalog, and `primary` is exactly the seven both surfaces already showed | unit (guard) | `packages/types/embedded-data-resolver.test.ts` → `describe("listDisplayableReservedFields")`, 9 tests. The `primary` set is pinned as a literal list on purpose: an entry silently joining it is a visible regression, not a new feature. |
| A new catalog entry cannot be added without a display decision | unit (guard) | Same file — the catalog table test now asserts `display` per entry, so a new entry fails until classified. |
| A pre-ENG-1841 response renders exactly as today — no empty rows, no placeholders | unit (mutation) | Same block: a response with `meta: {}` yields zero fields for both groups. Mutate: drop the `values[entry.name] !== undefined` filter in `listDisplayableReservedFields`. |
| With Anonymize on, the suppressed fields are absent rather than blank | unit (guard) | Same block. Driven by what `anonymize.ts` already removes at ingest — no read-time policy of its own, because duplicating that decision is how the two drift. |
| A display surface shows a declared field's reserved twin, unlike the value maps | unit (guard) | Same block — pins the deliberate asymmetry with #8948's `dropShadowedReservedEntries`. |
| Values inherit the projection's `redactQuery` | unit (mutation) | Same block, and `responses/lib/utils.test.ts` → *"inherits the projection's redaction rather than reading meta again"*. Red if a cell went back to `meta.url`. |
| Neither surface holds a hardcoded list of reserved field names any more | unit (guard) | `responses/lib/utils.test.ts` → `describe("RESERVED_COLUMN_ENTRIES")` derives the expected set from the catalog, so reintroducing a local `METADATA_FIELDS` fails it. `METADATA_FIELDS`, `userAgentFields`, `COLUMNS_ICON_MAP` and `getMetadataValue` are deleted; the card's seven JSX branches are one loop. |
| **A saved column layout survives the upgrade, id for id** | unit (mutation) | `responses/lib/utils.test.ts` → *"the Device column keeps its persisted id, not the catalog's spelling"* and *"every other reserved column takes its catalog name as its id"*; `column-state.test.ts` → *"a saved Device column choice survives, because its id never changed"*, which asserts both halves — reconciliation keeps the id, and seeding does not flip a stored `false` back on. Mutate: drop `PERSISTED_RESERVED_COLUMN_NAMES` and all three go red. |
| **Every label either surface shows is translated**, and the derived fallback still works for an entry nobody has keyed yet | unit (mutation) | `responses/lib/utils.test.ts` → `describe("getReservedColumnLabel")`. *"routes every field either surface displays through `t()`, so none ships as English"* asserts the key **and** the rendered label for all twelve new names — so one of them silently going back to derived English is red. *"still derives a label for a catalog entry nobody has written a key for yet"* pins the other half. *"the catalog's `deviceType` keeps the column's existing `Device` header"* is the one case where deriving would change shipped copy for every survey. |
| Today's columns stay visible by default; the thirteen new ones do not | unit (mutation) | `column-state.test.ts` → *"hides every secondary column an author has never decided about"* and *"leaves the primary columns alone"*. Mutate: return `true` from `isReservedColumnVisibleByDefault` and the table grows thirteen columns. |
| **A newly added column gets a toggle and a drag handle** | unit (red on main) | `column-state.test.ts` → `describe("reconcileColumnOrder")`, 8 tests. `reconcileColumnOrder` did not exist; the effect used the saved array verbatim, so this is red by absence. Includes the runtime invariant — the result is a permutation of the real columns, never a duplicate, never a gap. |
| An author's own column arrangement survives, and an already-made visibility choice is never overridden | unit (guard) | Both files above. The highest-risk part of this change is silently rewriting someone's saved layout. |
| **End to end: the captured page, campaign and timezone are visible in the product without exporting** | e2e | `apps/web/playwright/survey-auto-capture-meta.spec.ts`, new test in the spec that already proves those values reach Postgres. Submits a response with UTM params, signs in as the owner, then asserts both halves: **table** — the settings modal lists `Page Path` and `UTM Source`, with `METADATA_pagePath`'s switch `unchecked` and `METADATA_url`'s `checked`, located by the column id the modal puts on each Switch; **card** — the classic tooltip still carries `Browser:`, `OS:`, `Device:` and `URL:` (that last one is what proves the `hasUserAgent` gating fix), the disclosure's values are hidden until clicked, and after clicking, `Page Path`, `UTM Source`, `Timezone` and `Viewport Width` are all visible. The values are asserted **against the row this test wrote to Postgres** — `pagePath`, `timezone` and `viewportWidth` are read back out of the stored response and matched to what the disclosure renders, so a wrong value projection cannot pass on the strength of the labels alone. |
| Anonymize's existing card assertions still hold | e2e (unchanged) | `anonymize-responses.spec.ts` asserts `/^Country:/`, `/^Device:/`, `/^Browser:/` have count 0. The catalog loop keeps the same `Label: value` shape and the `deviceType` override keeps saying `Device`, so those regexes still match what they are meant to match. |
| The merge with #8948 did not break either change at the seam | unit + e2e | Full suite on the merged tree: **596 files, 7750 tests green**, plus the 9/9 e2e above. |

<details>
<summary><b>What the first push got wrong about the e2e test, and how</b></summary>

The first version of this PR shipped the Playwright test typechecked but **unrun**, and said so under Open gaps. CI ran it and it failed. All three causes were in the test; the production code needed no change:

1. It reused the user from the shared `beforeEach`. `createUserFixture` binds `login()` to the page of the test that created it, so by the fourth test that page is closed and the sign-in POST fails with `Test ended`. It now creates its own user and seeds its own survey.
2. It looked for the response card on the responses page. The table is the default view and the card lives in the response modal, opened by clicking any cell but `select` — the dedicated "Expand response" button is `group-hover:flex`.
3. It asserted `getByText(/^Browser:/)`. Each row renders `{label}: {value}` across separate text nodes, so that never matched. Asserting on the tooltip container with `toContainText` is robust and lets the same step check `URL:`.

The table half, which the first version had cut as "hard to drive", is now covered — that is what the `aria-label` addition unlocked.

</details>

**Open gaps**

- No component tests for `AutoCapturedFields.tsx` or the card metadata component, per `AGENTS.md` ("do not write component or UI unit tests for `.tsx`"). Their logic lives in `listDisplayableReservedFields` and `getReservedFieldLabel`, which are unit-tested; the rendering is the Playwright test's job.
- **`getReservedFieldLabel` is a `switch` of literal `t("…")` calls, and it costs a Sonar CPD exclusion.** It cannot be a name → key lookup map — the change that would actually make it DRY — because `packages/i18n-utils/src/scan-translations.ts` only counts a key as used when it appears as a literal first argument to `t(`; a map makes all twenty read as dead and `pnpm i18n` offers to delete them from fifteen locale files. CPD normalises string literals away, so past ~11 arms such a switch is a duplicate of itself shifted by two lines *and* of every other switch of the same shape — Sonar's report named lines 40-79 vs 42-81 of this file plus the segments `convertOperatorToText`/`convertOperatorToTitle` pair. That failed the gate at 5.1% duplication on new code (43 lines, all in this one file). The file is now in `sonar.cpd.exclusions` with the reasoning recorded there, beside the existing `schema-definition.ts` entry excluded for the same class of reason. **This is a silenced smell, not a fixed one** — flagged explicitly because it is a reviewable choice. The alternative I rejected was splitting the switch to keep each uniform run under the token threshold: that shapes code to a counter and re-trips silently the next time a catalog entry is added. The real fix is teaching the scanner to recognise a key map, which is shared-tooling work and does not belong in this PR.
- **`pnpm format` on this branch rewrites ~40 unrelated files** and there is no `format:check` script here to have caught it (that landed on `main` after the epic branched). I formatted only this change's files.
- The disclosure's own keyboard and screen-reader behaviour is Radix's default. I did not audit it beyond confirming the trigger is a real `button` with a text label.

## Breaking changes

None.

No API or SDK contract moves: no schema, endpoint, env var, config or payload shape changes, and no migration. `TReservedFieldCatalogEntry` gained a required `display` field and `ZResponseTableData` gained `reservedValues`, but both are internal types with every construction site updated in this diff.

No persisted state changes either: every column id is byte-identical to what authors already have in localStorage — see **Column ids stay exactly where they are** above. The two changes to what an existing response *renders* are named in their own section above; neither is a contract an integration can depend on.

## QA / Test Plan

**How to test**

- [ ] Publish a link survey with one open-text question, open it with campaign params — `/s/<surveyId>?utm_source=newsletter&utm_medium=email&utm_campaign=spring` — and submit an answer.
- [ ] Go to the survey's **Responses** tab and click anywhere on the new row → the response card opens. Hover the **Device info** button in the card footer → the tooltip still lists `Browser:`, `OS:`, `Device:`, `URL:`, `Country:`, `Source:`, `Action:` exactly as before. `IP address` is **no longer** in this tooltip.
- [ ] In the card body, below the answers, a collapsed **More context (n)** row is present. Its values are hidden. Click it → `Page Path` (`/s/<surveyId>`), `UTM Source` (`newsletter`), `UTM Medium`, `UTM Campaign`, `Timezone`, `Screen Width/Height`, `Viewport Width/Height` and `Page Referrer` appear. `IP address` appears here too, if IP capture is on for the instance.
- [ ] Switch the app language to German (or any non-English locale) and reopen the card → those labels are translated (`Seitenpfad`, `UTM-Quelle`, `Zeitzone`), not English.
- [ ] Close the card. Click the **Table settings** button in the toolbar (the sliders icon, right of the search field) → the list now contains `Page Path`, `Page Referrer`, `UTM Source/Medium/Campaign/Term/Content`, `Screen Width/Height`, `Viewport Width/Height`, `Timezone` and `IP address`, each with a toggle. All of them start **off**; `URL`, `Country`, `Source`, `Action`, `Browser`, `OS`, `Device` start **on**.
- [ ] Toggle `Page Path` on and close the modal → a `Page Path` column appears in the table showing `/s/<surveyId>`, positioned with the other metadata columns rather than at the far right. Reload the page → it is still on, still in that position.
- [ ] Toggle `Device` **off**, reload → it stays off. This is the upgrade-safety case: an author who had hidden Device before this release must not find it back on.
- [ ] Edge — **existing saved layout**: on a survey whose column order you had previously customised (drag a column, hide one), reload after the release → your order and your hidden columns are unchanged, and the new columns are all present in the settings list with working toggles.
- [ ] Edge — **Anonymize responses on** (survey Settings → Response Options): submit a new response, open its card → no `Country:`, `Browser:`, `OS:`, `Device:` rows, but `URL:`, `Action:` and `Source:` **are** shown. Before this PR they were hidden too, which was the bug.
- [ ] Edge — **an old response** (created before the ENG-1841 capture shipped): its card shows the classic tooltip and **no** *More context* row at all — no empty rows, no blank values.
- [ ] Edge — **a response created through the management API** with `meta.userAgent` set but `meta.userAgent.device` omitted: the tooltip shows `Browser:` and `OS:` but **no** `Device:` row. Previously it read `Device: PC / Generic device`, a value we never captured.
- [ ] Edge — **a survey that declares its own hidden field named `country`** (or `url`): the response card shows both — the author's field under *Hidden fields* with the respondent's value, and the auto-captured one in the tooltip. A display surface deliberately shows both, unlike recall and logic, which resolve to the declared one.
- [ ] Edge — a **respondent-facing URL carrying query params** (`/s/<id>?email=a@b.com`): the `URL` column and the card's `URL:` row show the path without the query string. Reading through the catalog means the existing `redactQuery` policy now applies to this column, where before it printed the raw URL.

**Preconditions / test data**

- Any workspace with a published **link** survey and at least one response. Nothing plan- or entitlement-gated; no feature flag.
- For the "existing saved layout" case you need column state saved *before* installing this build — customise the columns on the previous release first.
- For the IP-address rows, the instance needs IP capture enabled (unchanged gating).
- For the management-API case, a v1 or v2 API key and a `POST /api/v1/management/responses` body whose `meta.userAgent` has `browser`/`os` but no `device`.

**Risks & regressions**

- **A saved column layout being silently rewritten** is the worst outcome here — re-check order and hidden columns on a survey you had customised, on both a big and a small survey.
- The response card is a heavily used surface: re-check that the footer tooltips (Device info, Language, Tags) all still open, and that surveys with hidden fields and variables still render those sections.
- Anonymize responses — the gating changed shape (per field rather than all-or-nothing), so re-check both on and off.
- Nothing outside the responses page and the response card should be touched; the survey editor, summary tab and exports are unchanged.

- none. Fourteen new `en-US` strings — `more_context`, `auto_captured_field`, and one per newly displayed field; `pnpm i18n` generated all of them into the other 14 locales (`0 missing keys, 0 unused keys, all locales complete`). No DB migration, no env var, no cutover step.


apps/web/playwright/survey-auto-capture-meta.spec.ts
```diff
@@ -186,4 +190,64 @@ test.describe("Auto-captured browser context on responses @slow", () => {
...
+      // The seven the card has always shown, now read from the catalog. `Browser` is server-derived
+      // from the request header, so it is present for every response including historical ones.
+      const deviceInfo = page.getByLabel("Device info").first();
+      await expect(deviceInfo).toBeVisible({ timeout: 60000 });
+      await deviceInfo.hover();
+      await expect(page.getByText(/^Browser:/)).toBeVisible();
+    });
+
+    await test.step("the auto-captured fields sit behind a collapsed disclosure", async () => {
+      const disclosure = page.getByTestId("auto-captured-fields").first();
```
kiln-agent: **The card UI these two steps assert is not on `/responses`.** That page renders only `ResponseTable` (`ResponseDataView` returns `<ResponseTable/>` and nothing else). `SingleResponseCard` — and therefore `SingleResponseCardMetadata`'s `aria-label="Device info"` button and the new `AutoCapturedFields` disclosure — lives exclusively inside `ResponseCardModal`, which `ResponseTable` renders closed and only opens once `setSelectedResponseId` fires from a cell click.

So after the `page.goto(...)` on line 205, `page.getByLabel("Device info")` (210) and `page.getByTestId("auto-captured-fields")` (217) both match zero elements, and `expect(deviceInfo).toBeVisible({ timeout: 60000 })` times out — the test fails before it ever reaches the disclosure assertions.

The steps need to open the card first, e.g. click the newest row's answer cell (`page.getByText("Seen in the product").first().click()`) and scope the subsequent locators to the modal dialog.


---

apps/web/playwright/survey-auto-capture-meta.spec.ts
```diff
@@ -186,4 +190,64 @@ test.describe("Auto-captured browser context on responses @slow", () => {
...
+      await expect(disclosure.getByText("Timezone", { exact: true })).toBeVisible();
+      await expect(disclosure.getByText("Viewport Width", { exact: true })).toBeVisible();
+      // The values, not just the labels.
+      await expect(disclosure.getByText("newsletter", { exact: true })).toBeVisible();
+      await expect(disclosure.getByText("1280", { exact: true })).toBeVisible();
+    });
+
+    await test.step("the new table columns exist, listed but unchecked", async () => {
+      await page.getByRole("button", { name: "Table" }).click();
+      await page.getByRole("button", { name: "Settings" }).click();
```
kiln-agent: **Neither of these two locators can resolve.**

- `{ name: "Table" }` — there is no such control anywhere in the app; `grep` finds no `"table"` label in `apps/web/locales/en-US.json` and no view-mode toggle on the responses page (it is table-only, see the comment above).
- `{ name: "Settings" }` — the table-settings button in `modules/ui/components/data-table/components/data-table-toolbar.tsx` is

  ```tsx
  <TooltipRenderer tooltipContent={t("common.table_settings")} shouldRender={true}>
    <button type="button" onClick={() => setIsTableSettingsModalOpen(true)} className="…">
      <SettingsIcon strokeWidth={1.5} className="m-1 size-6 p-0.5" />
    </button>
  </TooltipRenderer>
  ```

  no text child, no `aria-label`. `TooltipRenderer` wraps the trigger in a `<span>` and its content becomes tooltip text, not `aria-label`/`aria-labelledby`, so the button's accessible name is the empty string. (The `{ name: "Settings", exact: true }` matches in `survey.spec.ts` / `anonymize-responses.spec.ts` are the survey **editor** tab, a different control.)

Line 237 throws first. Suggest locating the toolbar button structurally (or adding an `aria-label` to it, which is a genuine a11y gap worth fixing anyway) and dropping the "Table" click.


---

apps/web/playwright/survey-auto-capture-meta.spec.ts
```diff
@@ -186,4 +190,64 @@ test.describe("Auto-captured browser context on responses @slow", () => {
...
+  }) => {
+    // ENG-2540. ENG-1841 captures twelve fields on every response and, until this, none of them were
+    // visible anywhere: an author could only reach them by exporting. The tests above prove the
+    // values reach Postgres; this one proves an author can see them.
+    await page.setViewportSize({ width: 1280, height: 800 });
+    await page.goto(`${surveyUrl}?${UTM_QUERY}`);
+    await submitAnswer(page, "Seen in the product");
+    await readStoredMeta(surveyId ?? "");
+
+    await owner?.login();
```
kiln-agent: **`owner.login()` signs in the wrong browser context.**

`beforeEach` short-circuits on `if (surveyUrl) return`, so `owner` is only ever assigned during the **first** test in the describe. The fixture binds the page at construction time:

```ts
// playwright/fixtures/users.ts
export const createUserFixture = (user, page, ids?) => ({
  login: async () => { await login(user, page); },   // ← captures THIS page
  …
});
// login() posts to /api/auth/sign-in/email via page.context().request
```

Playwright hands every test a fresh `page` / `BrowserContext` by default, so this call writes the Better Auth session cookie into test #1's context jar. The `page.goto(...)` on the next line runs on *this* test's page, which has no session — it redirects to the login screen and every assertion below fails.

Fix by creating (or re-logging) the owner against the current test's fixture, e.g. drop the `if (surveyUrl) return` short-circuit for the login half, or take `users` in this test and call `users.create({ email: ownerEmail })`-equivalent / re-bind `login` to the live page.


---

packages/types/embedded-data-resolver.ts
```diff
@@ -822,6 +914,44 @@ export const projectClientReservedValues = (
  * answers respondents actually gave. It lives here, called from every consumer, rather than being
  * open-coded per surface, so there is exactly one line to audit and one line to test.
  */
+/** One reserved field as a human-facing surface renders it: the entry, plus the value it resolved to. */
```
kiln-agent, line 931: **The new declarations were inserted between `mergeReservedValues`' JSDoc and `mergeReservedValues`.**

The block ending `…so there is exactly one line to audit and one line to test.` (lines 903–916) is the pre-existing **grandfather rule** doc for `mergeReservedValues`. It now sits immediately above `export interface TDisplayableReservedField`, so:

- hovering `TDisplayableReservedField` shows a paragraph about spread ordering, `FORBIDDEN_IDS` and `#recall:country#` that has nothing to do with it;
- `mergeReservedValues` (line 955) is left with **no** doc — which is exactly the audit trail that comment says must stay attached to that one line ("Flipping the two spreads is the whole regression").

Move the block back down to directly above `export const mergeReservedValues`.


---

apps/web/modules/analysis/components/SingleResponseCard/components/SingleResponseCardMetadata.tsx
```diff
@@ -82,54 +97,26 @@ export const SingleResponseCardMetadata = ({ response, locale }: SingleResponseC
...
+        {primaryFields.map(({ entry, value }) => (
+          <p
+            key={entry.name}
+            // `url` is the one value long enough to need wrapping rather than truncation.
+            className={entry.name === "url" ? "break-all" : "truncate"}
+            title={`${getReservedFieldLabel(entry.name, t)}: ${value}`}>
+            {getReservedFieldLabel(entry.name, t)}: {value}
+          </p>
+        ))}
+      </div>
```
kiln-agent, line 118: **A third visible-behaviour change, beyond the two the description names: the `PC / Generic device` fallback is gone.**

The deleted branch rendered a `Device:` row whenever `meta.userAgent` existed at all, substituting a literal when `device` was unset:

```tsx
{response.meta.userAgent && (
  <p … >
    {t("…device")}: {response.meta.userAgent.device ? response.meta.userAgent.device : "PC / Generic device"}
  </p>
)}
```

`listDisplayableReservedFields` omits any entry that resolves to `undefined`, so a response with `userAgent.browser`/`os` but no `device` now renders **no Device row** instead of `Device: PC / Generic device`. Reachable today via `/api/v1/management/responses` and `/api/v2/management/responses`, which accept caller-supplied `meta` (`meta.userAgent.device` is optional in `ZResponseMeta`), and for any response stored before the client ingest routes started defaulting `agent.getDevice().type || "desktop"`.

Not necessarily wrong — arguably better than a fake value — but it contradicts "Everything else renders exactly as before", so it should either be called out in **Breaking changes** or preserved (e.g. keep the fallback as the `deviceType` arm of the catalog accessor, or leave `PC / Generic device` in the label override layer).


---

kiln-agent (commented):
Reviewed the full diff (32 files) for correctness. The production logic holds up well — I traced `reconcileColumnOrder` against saved-order permutations, stale ids and the `METADATA_device` → `METADATA_deviceType` rename and could not break the "permutation of the real columns" invariant; I also confirmed `defaultColumnOrder`'s `useMemo([table])` is computed on the first render while `columnOrder` state is still `[]`, so TanStack's `getAllLeafColumns()` (which *does* apply `_getOrderColumnsFn`) genuinely yields definition order there, and that `quotas`/`isReadOnly`/`isVerifyEmailEnabled` are all server-provided at first render, so the drop-stale-ids rule cannot eat a legitimate saved column. `reservedValues` has exactly one construction site, and the skeleton rows that lack it are only ever rendered with `SkeletonCell`. `pnpm vitest` on the two touched spec files: 36/36 green.

Five findings, four of them in the new e2e test — which the description flags as unrun, and which cannot pass as written:

1. **`survey-auto-capture-meta.spec.ts:210,217`** — the first two steps assert response-**card** UI on `/responses`, which renders only the table; the card lives in `ResponseCardModal` and needs a row click first.
2. **`survey-auto-capture-meta.spec.ts:237-238`** — `getByRole("button", { name: "Table" })` matches nothing (no such control exists), and the toolbar's settings button has no accessible name.
3. **`survey-auto-capture-meta.spec.ts:204`** — `owner.login()` is bound to the first test's `page`/context, so the current test stays unauthenticated.
4. **`embedded-data-resolver.ts:917`** — `mergeReservedValues`' "grandfather rule" JSDoc got reattached to the new `TDisplayableReservedField` interface.
5. **`SingleResponseCardMetadata.tsx:105`** — the `Device: PC / Generic device` fallback is silently dropped; a third visible-behaviour change beyond the two documented.

Nothing here blocks the design; (1)–(3) do need fixing before the e2e claim in the Coverage table means anything.


---

apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/lib/utils.ts
```diff
@@ -46,58 +44,53 @@ export const getContactInfoFieldLabel = (field: string, t: TFunction) => {
...
-  source: ArrowUpFromDotIcon,
-  url: GlobeIcon,
-};
+/** Prefix for a reserved column's TanStack id. Unchanged, so persisted column state still matches. */
+export const METADATA_COLUMN_PREFIX = "METADATA_";
 
-const userAgentFields = ["browser", "os", "device"];
-export const METADATA_FIELDS = ["action", "country", ...userAgentFields, "source", "url"];
+/** The column id for a reserved field. */
+export const reservedColumnId = (name: string): string => `${METADATA_COLUMN_PREFIX}${name}`;
```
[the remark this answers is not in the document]
Comment (itsjavi) [reply apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/lib/utils.ts]: Fixed exactly as proposed — a one-entry `PERSISTED_RESERVED_COLUMN_NAMES` map, so `reservedColumnId("deviceType")` keeps returning `METADATA_device`.

Worth recording why this was worse than "an author loses their ordering". `primary` columns are deliberately left unseeded in `columnVisibility`, and TanStack treats an id absent from that map as *visible* — so an author who had **hidden** Device would not merely have found it moved, they would have found it switched back on.

Three regression tests: `responses/lib/utils.test.ts` pins the id and asserts every other reserved column still takes its catalog name (so the map stays an exception rather than becoming a rename table), and `column-state.test.ts` asserts both halves of the upgrade path — reconciliation keeps the saved id, and seeding does not overwrite a stored `false`.

---

packages/types/responses.ts
```diff
@@ -497,6 +497,14 @@ export const ZResponseTableData = z.object({
...
   contactAttributes: ZResponseContactAttributes,
   meta: ZResponseMeta,
+  /**
+   * Reserved-field values already resolved and rendered, keyed by column id (ENG-2540). Precomputed
+   * on the row like `variables` and `responseData`, rather than read out of `meta` by a per-column
+   * switch: resolution belongs to the catalog's own accessors, which apply the `redactQuery` policy
+   * and the dataType coercion, and `meta` alone cannot express a field that is not stored under its
+   * own name (`deviceType` lives at `meta.userAgent.device`).
+   */
+  reservedValues: z.record(z.string(), z.string()),
```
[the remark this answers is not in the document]
itsjavi: Skipping this one, with a reason rather than a shrug.

The skeleton rows are `{ responseId } as TResponseTableData`, and that cast already omits eight other required fields — `responseData`, `variables`, `tags`, `status`, `verifiedEmail`, `language`, `notes`, `createdAt`. Adding `reservedValues: {}` on its own would not make the cast honest, only less obviously dishonest.

It is also never read: `ResponseTable` swaps every cell for `SkeletonCell` while `isFetchingFirstPage`, so the reserved cell renderer that would do `row.original.reservedValues[…]` never runs against a skeleton row.

A dedicated skeleton type is the right fix and I agree with it — but the shape predates this PR, and doing it properly means changing `ResponseTable`'s row type for every column, not just the ones added here. That belongs in its own change rather than smuggled into a display PR.

---

apps/web/modules/analysis/lib/reserved-field-display.ts
```diff
@@ -0,0 +1,82 @@
...
+    case "ipAddress":
+      return t("workspace.surveys.responses.ip_address");
+    case "os":
+      return t("workspace.surveys.responses.os");
+    case "source":
+      return t("workspace.surveys.responses.source");
+    case "url":
+      return t("common.url");
+    default:
+      return formatFieldNameToTitleCase(name);
```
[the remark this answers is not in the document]
itsjavi: Fixed, and it was the right call to raise — this was a real trade-off rather than an oversight, so I put it to the PR author before changing it.

All twenty `display !== "none"` catalog entries now have an explicit `t()` case. Twelve new `en-US` strings, generated into the other 14 locales by `pnpm i18n` (`0 missing keys, 0 unused keys, all locales complete`). One bonus from spelling them out: `utm*` now reads `UTM Source` rather than the derived `Utm Source`.

The one part of the suggestion I did **not** take is "avoid using `formatFieldNameToTitleCase` as the displayable-field fallback". It stays as the `default` arm, deliberately, because it is what keeps ENG-2540's acceptance criterion — *adding a catalog entry surfaces it in both places with no further change* — true for ENG-1858's next batch. So it is a last resort rather than the rule: a new entry still reaches both surfaces for free, reading in English until someone adds its key.

Both halves are pinned so neither can rot into the other, in `responses/lib/utils.test.ts`:

- *"routes every field either surface displays through `t()`, so none ships as English"* — asserts the key **and** the rendered label for all twelve, so one silently going back to derived English is red.
- *"still derives a label for a catalog entry nobody has written a key for yet"* — asserts the fallback still works, so deleting it is red too.

---

apps/web/modules/analysis/components/SingleResponseCard/components/SingleResponseCardMetadata.tsx
itsjavi: Verified — the finding is correct. That branch was the only place the literal appeared, and it is gone.

Keeping the new behaviour rather than restoring it, for two reasons: rendering a device we never captured is worse than rendering no row, and `"PC / Generic device"` was a hardcoded English string that never went through `t()`, so bringing it back would also mean re-adding an untranslated literal to a surface this PR just finished localizing.

What was wrong was the description, not the code — "Everything else renders exactly as before" was an overclaim. It now names this alongside the `ipAddress` move under **The two visible-behaviour changes, named**, with the management-API path that makes it reachable spelled out, and there is a QA step for it.

---

pandeymangg (approved):

PR #8951, opened 2026-08-21
fix(sso): write the canonical Account.issuer so Google sign-in stops looping

xernobyl:
Fixes ENG-2555

## What & why

Google SSO has been broken since the Better Auth 1.7 upgrade (#8835): the callback 500s, a refresh lands on `?error=OAuthAccountNotLinked`, and every retry repeats it. Caught on staging; #8835 is in no release tag, so no production data is affected and this lands before 5.4.

1.7 keys accounts on `(issuer, accountId)` and filters every lookup on it. The canonical issuer is **provider-dependent**, and google is the odd one out — it is a built-in social provider that declares its own `accountIssuer` upstream:

| provider | canonical issuer | source |
| --- | --- | --- |
| `credential` | `local:credential` | `createLocalAccountIssuer` |
| `google` | `https://accounts.google.com` | declared by the provider itself |
| github / azuread / openid / saml | `local:oauth:<provider>` | upstream fallback, and what we pin |

The ENG-2343 SQL backfill encoded that correctly — its comment even warns that *"getting google wrong would leave every existing Google user unmatched at sign-in"*. But `ssoAccountIssuer`, which SSO recovery used for every write, returns the synthetic form unconditionally. So linking Google for an existing user wrote `local:oauth:google`, a key Better Auth never looks under: the link was invisible, recovery fired again on the next sign-in, and it never converged.

Two changes:

- **`canonicalAccountIssuer`** mirrors the SQL `CASE`. `ssoAccountIssuer` keeps its one real job — pinning `accountIssuer` on the generic providers, where the synthetic form is correct and deliberate. Two helpers because they answer different questions: *what we pin* vs *what an existing row's issuer must be*. The [1.7 upgrade guide](https://www.better-auth.com/docs/guides/1-7-upgrade-guide) prescribes exactly this ("build an explicit `providerId`-to-issuer map for your deployment").
- **The branch that made it unbreakable.** With a canonical row present and no legacy row, `syncSsoIdentityForUser` updated tokens only and never touched `issuer` — so a row written with a wrong value could not heal: sign-in could not see it, recovery ran again, and the update left it alone. It now writes the canonical issuer, so the next sign-in repairs the row.

**Why the suite stayed green.** Every assertion in `account-linking.test.ts` used google — the one provider the helper got wrong — and asserted the value *our own code passed*, not the one Better Auth looks up. A mocked test cannot notice that upstream reads a different key.

## Breaking changes

None. It repairs data written by an unreleased version and restores the intended lookup; no API, schema, env or config surface changes.

## Migrations & env

- `20260821165535_repair_account_issuer` (data): rewrites `Account.issuer` rows that disagree with the canonical value for their provider. Idempotent, no-op on an already-correct database, and baselined without running on a fresh one. `IS DISTINCT FROM` means it also repairs any `NULL` issuer left behind.
- Cannot merge identities: that would need two rows sharing a subject *and* mapping to one issuer, which `@@unique([provider, providerAccountId])` plus the injective provider→issuer map forbids. If that reasoning were ever wrong, `@@unique([issuer, providerAccountId])` aborts the migration rather than merging silently.
- No env changes.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Recovery links google under the key Better Auth actually reads | integration, real Postgres (mutation) | `better-auth-recovery-signin.integration.test.ts` links via `syncSsoIdentityForUser`, then looks the row up by `(issuer, providerAccountId)` built the way upstream builds it. Mutation: deleting the google arm in `constants.ts` fails this **and** the repair case, leaving github passing — the exact signature the bug had |
| github still uses the synthetic form | integration, real Postgres (guard) | Same test, second case. Pins that the fallback stays right for providers that declare no issuer |
| A row already carrying a wrong issuer heals | integration, real Postgres (mutation) | Third case: pre-seed `local:oauth:google`, run recovery, assert it is findable under `https://accounts.google.com` and that no duplicate row appeared |
| No SQL copy of the mapping can drift — in either migration | unit (mutation) | `constants.test.ts` **parses** every `CASE` block in BOTH migrations — the backfill's one and the repair's two (`SET` + the self-excluding `WHERE`) — and asserts each identical to the canonical parse, with a copy-count guard so dropping the repair's `WHERE` copy also fails. Mutations: a trailing slash on google in the repair's `SET` fails 1; dropping an arm from only the `WHERE` copy (the pair diverging, which would break idempotency) fails 1; a trailing slash in the backfill fails 2 |
| …or from upstream | unit (guard) | Same file asserts against Better Auth's own exports — `google({…}).accountIssuer`, `createOAuthAccountIssuer`, `createLocalAccountIssuer` — and that github has **no** `accountIssuer` property at all. A release that changes google's issuer, or gives github one, fails `pnpm test` instead of production sign-in |
| The fourth branch writes the issuer | unit (mutation) | `account-linking.test.ts`; dropping the write fails 2. A pre-existing test also caught it, which is how I knew the change was real |
| Both arms exercised through account-linking | unit (guard) | Added a non-google (github) case — previously every assertion used google, which is why the wrong branch looked right |
| The repair migration, end to end | manual + **integration (mutation)** | Manual: seeded the four-row matrix and ran the real migration runner — google fixed, github/credential untouched, NULL repaired, second run a no-op, fresh DB baselined without running. Now also automated: `repair-account-issuer.integration.test.ts` reproduces that matrix against real Postgres — repair, idempotency, empty-table no-op, and the repaired row resolving by Better Auth's own `(issuer, providerAccountId)` key. DB-level mutations: neutralising the google arms fails 2 tests; `IS DISTINCT FROM` → `<>` fails the NULL-row case |
| The deliberate TS/SQL encoding asymmetry is documented, not latent | unit (guard) | The SQL `ELSE` concatenates the raw provider id; the helper percent-encodes. Identity for every provider id in use. A test pins `team/github` as the divergence case, and both docblocks state the mirror includes the missing encoding on purpose |

| Live SSO smoke: registration → login → recovery loop → logout → deletion | manual, live | Full stack (real Postgres + Redis + Mailhog + stub OIDC IdP). **Registration**: JIT-provisioned user lands on `/`, account written `openid / local:oauth:openid`, session created. **Login**: second sign-in from a fresh browser goes straight through, still exactly 1 account row. **Verify-before-link recovery**: seeded an email-only user, SSO sign-in → `verification-requested`, no account minted before the email; clicked the Mailhog link → session + account `local:oauth:openid`, `identityProvider` flipped; **the loop test**: next sign-in lands on `/` directly, 1 account row, 0 recovery emails sent. **Logout**: `sign-out` 200, DB session row gone, `get-session` → null. **Deletion**: credential user deletes with password confirmation — user, account and session rows all cascade to 0 |
| Deletion & registration guards hold | manual, live | Bare `delete-user` for an SSO (password-less) user is refused with "Password confirmation is required" — the ENG-1054 confirmation-bypass guard, since SSO deletion goes through the email-token action. Native `sign-up/email` on an instance with existing users is refused `signup_disabled` — the ENG-2293 closed-instance gate |
| Credential sign-in resolves through `issuer = local:credential` | manual, live | Seeded a credential account the post-cutover way (bcrypt on `Account.password`) and signed in over HTTP → 200. Exercises `findCredentialAccount`'s issuer filter against a real row |
| **Real Google, end to end — the case this PR exists for** | manual, live | Against a real Google Workspace identity on a throwaway DB, using the dev OAuth client (registered for `localhost:3000`). Existing user, no Google link — the exact staging state. Sign-in → `account_not_linked` → recovery email, and **no account row or session created before the inbox was proven**. Completing the emailed link wrote `provider=google, issuer=https://accounts.google.com` — asserted equal to `google({…}).accountIssuer` read from upstream in the same run. **Pre-fix this write produced `local:oauth:google`.** Then the loop test: a further Google sign-in landed past login (not the "confirm your email" page), still exactly 1 account row, and **0** new recovery emails — the reported symptom, gone |

**Open gaps**

- ~~Not yet verified against real Google.~~ **Done** — see the real-Google row above. Verified locally against a real Google identity rather than a stub, including the loop test. What remains is confirming it on **staging** after deploy, which is a deployment check rather than a gap in the change: staging uses a different Google OAuth client, so its own registration is the only thing left unproven by this run.
- The live smoke needed a local, uncommitted `sso` entitlement override (the dev license falls back to `default`, and `ssoLicenseGateBeforeHandler` 403s the callback without it). Reverted before committing; verified the tracked tree is clean.
- The migration's `logger.info` produces no visible output through `db:migrate:deploy` locally — the runner emitted nothing at all for a run that demonstrably did work. Pre-existing (every data migration logs the same way) and not worth diverging from house convention over, but it means the repaired count is not currently observable at deploy time.
- 8 unrelated `widget-view.test.ts` failures appear locally on my machine only; they are green in CI and come from my Node being 26.4.0 against the repo's `.nvmrc` 24.14.0.

**Risks**

- Nearby: `linkSocial` also resolves accounts by `(issuer, accountId)`, so it was affected by the same wrong value and is fixed by the same change — worth a click if you are exercising connected accounts.
- The connected-accounts UI lists rows by `userId`, so it showed "Google connected" throughout even while sign-in could not use the row. Don't read that list as proof the link works.
- Rolling deploys: a pod still on the old code could re-stomp a row during the window. The remedy is the branch-3 healing write this PR adds — a re-stomped row is repaired by the next sign-in, with no operator action. **Do not reach for re-running the migration**: `runDataMigration` returns early on `status = 'applied'` (`migration-runner.ts:304`) and there is no force flag, so a second `db:migrate:deploy` logs `already completed. Skipping...` and touches nothing. Re-running would need the `DataMigration` row deleted by hand.

---

> [!NOTE]

### DIFF
--- a/apps/web/modules/ee/sso/lib/constants.test.ts
+++ b/apps/web/modules/ee/sso/lib/constants.test.ts
@@ -1,0 +1,115 @@
+import { createLocalAccountIssuer, createOAuthAccountIssuer } from "@better-auth/core/db";
+import { github, google } from "@better-auth/core/social-providers";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, test } from "vitest";
+import { canonicalAccountIssuer, ssoAccountIssuer } from "./constants";
+
+/**
+ * `Account.issuer` is spelled in four places that cannot import each other, and ENG-2555 happened
+ * because two of them disagreed about google: the SQL backfill had a `CASE` arm for it, the TypeScript
+ * helper did not, and every Google sign-in broke while the whole suite stayed green.
+ *
+ * So this pins `canonicalAccountIssuer` against the two sources it has to agree with:
+ *
+ * - the SQL literal, parsed out of the migration rather than restated here — restating it would just
+ *   move the drift into this file;
+ * - Better Auth's own exports, which are what actually key the row at sign-in.
+ *
+ * The upstream leg is the one with a future in it: if a Better Auth release changes google's issuer, or
+ * gives github one it does not have today, this fails at `pnpm test` instead of at somebody's login.
+ */
+const ISSUER_BACKFILL_MIGRATION = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "../../../../../../packages/database/migration/20260812110000_eng_2343_better_auth_17_resource_model/migration.sql"
+);
+
+/**
+ * The `CASE` arms of the `UPDATE "Account" SET "issuer" = …` backfill, as `{ provider: issuer }`, plus
+ * the `ELSE` template. Parsed, not transcribed.
+ */
+const parseIssuerCase = (): { arms: Record<string, string>; elseTemplate: string } => {
+  const sql = readFileSync(ISSUER_BACKFILL_MIGRATION, "utf8");
+  const block = /UPDATE "Account"\s+SET "issuer" = CASE([\s\S]*?)END/.exec(sql);
+  if (!block) throw new Error("issuer backfill CASE not found — did the migration move or get renamed?");
+
+  const arms: Record<string, string> = {};
+  const armPattern = /WHEN "provider" = '([^']+)' THEN '([^']+)'/g;
+  let match = armPattern.exec(block[1]);
+  while (match) {
+    arms[match[1]] = match[2];
+    match = armPattern.exec(block[1]);
+  }
+
+  const elseArm = /ELSE '([^']+)' \|\| "provider"/.exec(block[1]);
+  if (!elseArm) throw new Error("issuer backfill ELSE arm not found");
+
+  return { arms, elseTemplate: elseArm[1] };
+};
+
+describe("canonicalAccountIssuer ↔ the ENG-2343 SQL backfill", () => {
+  const { arms, elseTemplate } = parseIssuerCase();
+
+  // Guard the guard: if the regex silently matched nothing, every assertion below would pass against an
+  // empty object and prove precisely nothing.
+  test("the migration's CASE parsed, and still special-cases exactly credential and google", () => {
+    expect(Object.keys(arms).sort()).toEqual(["credential", "google"]);
+    expect(elseTemplate).toBe("local:oauth:");
+  });
+
+  test.each(Object.entries(parseIssuerCase().arms))("agrees with the SQL for %s", (provider, expected) => {
+    expect(canonicalAccountIssuer(provider)).toBe(expected);
+  });
+
+  test.each(["github", "azuread", "openid", "saml"])("agrees with the SQL ELSE arm for %s", (provider) => {
+    expect(canonicalAccountIssuer(provider)).toBe(`${elseTemplate}${provider}`);
+  });
+});
+
+describe("canonicalAccountIssuer ↔ Better Auth's own account key", () => {
+  const providerOptions = { clientId: "pin-test", clientSecret: "pin-test" };
+
+  /**
+   * How upstream resolves the issuer (`better-auth/dist/oauth2/account-key.mjs`): a provider's declared
+   * `accountIssuer` wins, and only its absence falls back to the synthetic form.
+   */
+  test("google declares its own issuer, and we use that rather than the synthetic form", () => {
+    const declared = google(providerOptions).accountIssuer;
+
+    expect(declared).toBe("https://accounts.google.com");
+    expect(canonicalAccountIssuer("google")).toBe(declared);
+    // The regression itself: this is the value that was written, and it is not the one BA looks under.
+    expect(canonicalAccountIssuer("google")).not.toBe(ssoAccountIssuer("google"));
+  });
+
+  /**
+   * Pins the *assumption* behind the fallback, not just the fallback. If a future release gives github a
+   * declared issuer, `local:oauth:github` silently becomes wrong for it in exactly the way it was wrong
+   * for google — and this is what notices.
+   */
+  test("github declares none, so the synthetic fallback is the right answer for it", () => {
+    // `in` rather than reading the property: upstream's type for github has no `accountIssuer` at all,
+    // so this is checked at compile time as well as here. If a release adds one, the property appears
+    // and this fails — which is the notification we want.
+    expect("accountIssuer" in github(providerOptions)).toBe(false);
+    expect(canonicalAccountIssuer("github")).toBe(createOAuthAccountIssuer("github"));
+  });
+
+  test("credential matches Better Auth's local account issuer", () => {
+    expect(canonicalAccountIssuer("credential")).toBe(createLocalAccountIssuer("credential"));
+  });
+});
+
+describe("ssoAccountIssuer stays the pinning helper", () => {
+  // It is still correct for the generic providers we pin `accountIssuer` on, and this records that the
+  // two helpers are deliberately different rather than one being a leftover.
+  test.each(["azuread", "openid", "saml"])("%s keeps the synthetic form", (provider) => {
+    expect(ssoAccountIssuer(provider)).toBe(`local:oauth:${provider}`);
+    expect(canonicalAccountIssuer(provider)).toBe(ssoAccountIssuer(provider));
+  });
+
+  test("percent-encodes a provider id that needs it", () => {
+    expect(ssoAccountIssuer("team/github")).toBe("local:oauth:team%2Fgithub");
+  });
+});
--- a/apps/web/modules/ee/sso/lib/constants.test.ts
+++ b/apps/web/modules/ee/sso/lib/constants.test.ts
@@ -20,50 +20,100 @@ import { canonicalAccountIssuer, ssoAccountIssuer } from "./constants";
  * The upstream leg is the one with a future in it: if a Better Auth release changes google's issuer, or
  * gives github one it does not have today, this fails at `pnpm test` instead of at somebody's login.
  */
-const ISSUER_BACKFILL_MIGRATION = join(
+const MIGRATION_DIR = join(
   dirname(fileURLToPath(import.meta.url)),
-  "../../../../../../packages/database/migration/20260812110000_eng_2343_better_auth_17_resource_model/migration.sql"
+  "../../../../../../packages/database/migration"
 );
 
 /**
- * The `CASE` arms of the `UPDATE "Account" SET "issuer" = …` backfill, as `{ provider: issuer }`, plus
- * the `ELSE` template. Parsed, not transcribed.
+ * The two SQL spellings of the canonical mapping. The ENG-2343 backfill has one `CASE`; the ENG-2555
+ * repair restates it twice (`SET` and the self-excluding `WHERE`). All copies in all files must agree
+ * with each other and with `canonicalAccountIssuer` — a repair that drifts would *un-fix* rows through
+ * the migration meant to cure them.
  */
-const parseIssuerCase = (): { arms: Record<string, string>; elseTemplate: string } => {
-  const sql = readFileSync(ISSUER_BACKFILL_MIGRATION, "utf8");
-  const block = /UPDATE "Account"\s+SET "issuer" = CASE([\s\S]*?)END/.exec(sql);
-  if (!block) throw new Error("issuer backfill CASE not found — did the migration move or get renamed?");
+const ISSUER_MIGRATION_SOURCES = {
+  "eng-2343 backfill": "20260812110000_eng_2343_better_auth_17_resource_model/migration.sql",
+  "eng-2555 repair": "20260821165535_repair_account_issuer/migration.ts",
+} as const;
 
-  const arms: Record<string, string> = {};
-  const armPattern = /WHEN "provider" = '([^']+)' THEN '([^']+)'/g;
-  let match = armPattern.exec(block[1]);
-  while (match) {
-    arms[match[1]] = match[2];
-    match = armPattern.exec(block[1]);
-  }
+interface TParsedIssuerCase {
+  arms: Record<string, string>;
+  elseTemplate: string;
+}
 
-  const elseArm = /ELSE '([^']+)' \|\| "provider"/.exec(block[1]);
-  if (!elseArm) throw new Error("issuer backfill ELSE arm not found");
+/**
+ * Every `"issuer" = CASE … END` block in a migration file, as `{ provider: issuer }` arms plus the
+ * `ELSE` template. Parsed, not transcribed — restating the values here would just move the drift into
+ * this file.
+ */
+const parseIssuerCases = (relativePath: string): TParsedIssuerCase[] => {
+  const source = readFileSync(join(MIGRATION_DIR, relativePath), "utf8");
+  const blocks = [...source.matchAll(/"issuer" (?:= CASE|IS DISTINCT FROM \(\s*CASE)([\s\S]*?)END/g)];
+  if (blocks.length === 0) throw new Error(`no issuer CASE found in ${relativePath} — moved or renamed?`);
 
-  return { arms, elseTemplate: elseArm[1] };
+  return blocks.map(([, body]) => {
+    const arms: Record<string, string> = {};
+    for (const [, provider, issuer] of body.matchAll(/WHEN "provider" = '([^']+)' THEN '([^']+)'/g)) {
+      arms[provider] = issuer;
+    }
+    const elseArm = /ELSE '([^']+)' \|\| "provider"/.exec(body);
+    if (!elseArm) throw new Error(`issuer CASE in ${relativePath} has no ELSE arm`);
+    return { arms, elseTemplate: elseArm[1] };
+  });
 };
 
-describe("canonicalAccountIssuer ↔ the ENG-2343 SQL backfill", () => {
-  const { arms, elseTemplate } = parseIssuerCase();
+const parsedSources = Object.entries(ISSUER_MIGRATION_SOURCES).map(([label, path]) => ({
+  label,
+  cases: parseIssuerCases(path),
+}));
 
+// One canonical parse for the per-arm assertions below; the cross-copy equality tests prove every
+// other copy is identical to it, so asserting against one is asserting against all.
+const { arms, elseTemplate } = parsedSources[0].cases[0];
+
+describe("canonicalAccountIssuer ↔ every SQL spelling of the mapping", () => {
   // Guard the guard: if the regex silently matched nothing, every assertion below would pass against an
-  // empty object and prove precisely nothing.
-  test("the migration's CASE parsed, and still special-cases exactly credential and google", () => {
+  // empty object and prove precisely nothing. The repair file must contain exactly two copies (SET and
+  // the self-excluding WHERE) — a refactor that drops one would weaken its idempotency, and this is
+  // what notices.
+  test("both migrations parsed, with the expected number of CASE copies", () => {
+    expect(parsedSources.map(({ label, cases }) => [label, cases.length])).toEqual([
+      ["eng-2343 backfill", 1],
+      ["eng-2555 repair", 2],
+    ]);
     expect(Object.keys(arms).sort()).toEqual(["credential", "google"]);
     expect(elseTemplate).toBe("local:oauth:");
   });
 
-  test.each(Object.entries(parseIssuerCase().arms))("agrees with the SQL for %s", (provider, expected) => {
+  // The drift that shipped ENG-2555 was two spellings of this mapping disagreeing. Every copy in every
+  // migration must therefore be byte-equal to every other — including the repair's SET and WHERE pair,
+  // which could otherwise drift apart and make the repair non-idempotent.
+  test.each(
+    parsedSources.flatMap(({ label, cases }) => cases.map((c, i) => [`${label} copy ${i + 1}`, c] as const))
+  )("%s is identical to the canonical parse", (_label, parsed) => {
+    expect(parsed.arms).toEqual(arms);
+    expect(parsed.elseTemplate).toBe(elseTemplate);
+  });
+
+  test.each(Object.entries(arms))("agrees with the SQL for %s", (provider, expected) => {
     expect(canonicalAccountIssuer(provider)).toBe(expected);
   });
 
   test.each(["github", "azuread", "openid", "saml"])("agrees with the SQL ELSE arm for %s", (provider) => {
     expect(canonicalAccountIssuer(provider)).toBe(`${elseTemplate}${provider}`);
+  });
+
+  /**
+   * Documents the one input class where the TS and SQL sides deliberately disagree: the SQL ELSE arm
+   * concatenates the raw provider id, while the helper percent-encodes. Identity for every provider id
+   * in use (all encoding-neutral) — but a future provider id carrying a reserved character would make
+   * the migration write a value the app never looks up. This pins the divergence as known and
+   * deliberate rather than letting it look like an oversight; enabling such a provider means adding an
+   * explicit arm to the SQL, per the ENG-2343 migration's own comment.
+   */
+  test("the SQL ELSE arm cannot express an encoded provider id", () => {
+    expect(canonicalAccountIssuer("team/github")).toBe("local:oauth:team%2Fgithub");
+    expect(canonicalAccountIssuer("team/github")).not.toBe(`${elseTemplate}team/github`);
   });
 });
 
--- a/apps/web/modules/ee/sso/lib/constants.ts
+++ b/apps/web/modules/ee/sso/lib/constants.ts
@@ -2,20 +2,53 @@ export const OAUTH_ACCOUNT_NOT_LINKED_ERROR = "OAuthAccountNotLinked";
 export const SSO_RECOVERY_COMPLETION_PATH = "/api/auth/sso/recovery/complete";
 
 /**
- * The synthetic `Account.issuer` for our generic-OAuth providers (ENG-2343).
+ * The synthetic `Account.issuer` for the providers we configure ourselves (ENG-2343).
  *
- * Lives here, in a dependency-free module, because THREE places must produce byte-identical values and
- * a literal repeated three times is a silent-divergence bug waiting to happen:
+ * This is the value we PIN via `accountIssuer` on the generic-OAuth providers in
+ * `better-auth-providers.ts` (azuread / openid / saml). Because it is pinned, Better Auth stores and
+ * looks up exactly what we hand it, so upstream's own format never enters the picture — which is why
+ * this is deliberately NOT `createOAuthAccountIssuer` from `@better-auth/core/db` even though the two
+ * are currently identical. Tracking upstream here would drift us away from rows already written, and
+ * the SQL backfill could not follow.
  *
- * 1. `better-auth-providers.ts` — `accountIssuer`, what Better Auth writes and looks up.
- * 2. `account-linking.ts` — the rows SSO recovery writes itself.
- * 3. `migration/20260812110000_…` — the backfill, as a SQL literal (`'local:oauth:' || "provider"`),
- *    which cannot call TypeScript and is therefore the copy this one has to match.
- *
- * Deliberately NOT `createOAuthAccountIssuer` from `@better-auth/core/db`, even though it is public and
- * currently identical: because `accountIssuer` is set explicitly, Better Auth stores and looks up
- * whatever we hand it, so upstream's format never enters the picture — while the SQL literal in (3)
- * cannot follow an upstream change. Tracking upstream would drift us away from rows already written.
+ * It is NOT the answer to "what issuer does an existing row for provider X have" — a built-in social
+ * provider can declare its own. Use `canonicalAccountIssuer` for that (ENG-2555).
  */
 export const ssoAccountIssuer = (providerId: string): string =>
   `local:oauth:${encodeURIComponent(providerId)}`;
+
+/**
+ * The canonical `Account.issuer` for a given `Account.provider` — the value Better Auth 1.7 actually
+ * keys the row on, and therefore the only value a write may use (ENG-2555).
+ *
+ * 1.7 keys accounts on `(issuer, accountId)` and filters every lookup on it, so a row written with the
+ * wrong issuer is invisible to sign-in. Google is the trap: it is a BUILT-IN social provider that
+ * declares its own `accountIssuer` upstream, so the synthetic `local:oauth:` form is wrong for it.
+ * Writing `local:oauth:google` is what broke Google sign-in on 5.4-rc — the link was created, the user
+ * got a session, and every subsequent sign-in bounced back through verify-before-link forever.
+ *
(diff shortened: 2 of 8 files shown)


xernobyl:
Self-review pass (fresh read of the diff), security review, and a full live SSO smoke. Two comment-level fixes pushed as `4878f3c`; no code changes needed.

## Review

- **Fixed two comments the fix itself made stale.** The create-branch note still claimed the written value is "the same as the provider config" — false for google since this change, and comment/code disagreement on exactly this point is how the bug shipped. Both now state the canonical-vs-pinned distinction and point at the test pin.
- **Swept for other spellings of an issuer value**: exactly one definition site (`constants.ts`), two helpers, three consumers (`account-linking.ts`, the provider pins, `seed.ts`'s correct `local:credential` literal). Nothing else in `apps/web`, `packages/database/src`, or the auth module constructs one.
- **Migration ordering verified**: the repair (`…165535`) sorts after every migration that writes `Account`, and nothing between the ENG-2343 backfill and it touches the table.

## Security review

- **The new issuer write on the update branch cannot rebind a row.** The ownership guard (`existingCanonicalAccount.userId !== userId → throw`) runs before every branch, the row is found *by* `(provider, providerAccountId)`, and the issuer derives from that same `provider` — so the written issuer always matches the row's own provider, for a row already proven to belong to the caller.
- **The repair migration is a static tagged template** — zero interpolation, nothing attacker-influenced. It touches only `issuer`; `userId`/`provider`/`providerAccountId` are untouched, so no row can move between users or orgs, and an (unreachable, argued in the docblock) key collision aborts on the unique index rather than merging.
- No secrets or PII in fixtures or logs — the migration logs a count only; the test's provider credentials are `pin-test` literals.

## Live smoke — SSO registration / login / recovery / logout / deletion

Real Postgres + Redis + Mailhog + a stub OIDC IdP, driving the actual HTTP endpoints:

| scenario | result |
| --- | --- |
| **Registration** (JIT, fresh user) | lands on `/`; account `openid / local:oauth:openid`; session created |
| **Login** (existing linked, fresh browser) | straight to `/`; still exactly 1 account row |
| **Verify-before-link recovery** (existing email-only user) | sign-in → `verification-requested`, **no account minted before the email**; Mailhog link → session + correctly-issued account, `identityProvider` flipped |
| **The loop test** | next sign-in lands on `/` directly — 1 account row, **0** recovery emails. The loop is dead |
| **Logout** | `sign-out` 200; DB session gone; `get-session` → `null` |
| **Deletion** (credential user, password confirm) | user, account, session rows all cascade to 0 |
| **Guards** | bare `delete-user` for a password-less SSO user refused ("password confirmation required" — the ENG-1054 bypass guard); `sign-up/email` on a populated instance refused `signup_disabled` (ENG-2293) |
| **Credential sign-in** | 200 against a hand-seeded post-cutover row — exercises `findCredentialAccount`'s `issuer = local:credential` filter live |

**Honest limits.** The stub cannot reproduce the *google* loop — google's endpoints are hardcoded upstream, and for `openid` the canonical and synthetic issuers coincide. So the live smoke proves the recovery machinery end to end (and that this change doesn't regress it), while the google class rests on the real-Postgres integration tests, which mutation-fail when the google arm is removed. Real Google on staging remains the closing check. The smoke also needed a local, uncommitted `sso` entitlement override (dev license falls back to `default` and the license gate 403s the callback) — reverted before committing, tracked tree verified clean.

Also answered in review: **no breaking changes for self-hosters** — no released version ever contained the bug, so the repair migration no-ops on their data, and there are no surface changes of any kind.

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
```diff
@@ -0,0 +1,60 @@
...
+ * - Only the `issuer` column is touched; `userId`, `provider` and `providerAccountId` are untouched, so
+ *   no row can move between users or organizations.
+ */
+export const repairAccountIssuer: MigrationScript = {
+  type: "data",
+  id: "ghqt118bfqdjs8tpzv1dvbqi",
+  name: "20260821165535_repair_account_issuer",
+  run: async ({ tx }) => {
+    const repaired = await tx.$executeRaw`
+      UPDATE "Account"
```
kiln-agent, line 40: The fourth spelling of the `CASE` is pinned by nothing.

`constants.ts` enumerates four sites that must agree and adds "`constants.test.ts` pins this function against BOTH the SQL literal and upstream's own exports" — but `parseIssuerCase()` reads only `20260812110000_eng_2343_better_auth_17_resource_model/migration.sql`. This file (site 4) restates the same `CASE` twice, in `SET` here and in `WHERE` below, and no test reads either copy.

Concrete drift: a trailing slash on `'https://accounts.google.com/'`, an arm added here for a newly enabled built-in social provider, or the two copies diverging from each other — all leave `pnpm test` green while the repair writes an issuer Better Auth never looks under, reproducing ENG-2555 through the migration meant to cure it. That is precisely the failure class the PR closes for site 3 by parsing rather than transcribing.

Cheap fix: parameterise `ISSUER_BACKFILL_MIGRATION` over both migration files and run the existing arm assertions against each, which also catches the `SET`/`WHERE` pair drifting apart.

---

kiln-agent (commented):
Reviewed the fix itself against the installed `better-auth@1.7.0`: `dist/oauth2/account-key.mjs` resolves `provider.accountIssuer ?? createOAuthAccountIssuer(provider.id)`, `@better-auth/core/dist/social-providers/google.mjs:57` declares `accountIssuer: "https://accounts.google.com"`, and github declares none — so `canonicalAccountIssuer` is right for every provider this deployment registers (github/google unpinned social; azuread/openid/saml pinned to the synthetic form in `better-auth-providers.ts`). The repair migration's collision argument also holds: the provider→issuer map is injective and `@@unique([provider, providerAccountId])` forbids two rows per provider+subject, so no pair can converge on one `(issuer, providerAccountId)` — not even transiently inside the single `UPDATE`, which is the case a non-deferrable unique index would otherwise reject. No correctness objection to the change.

Two things worth acting on.

**1. Site 4 of the four-site invariant is unpinned** — inline on the repair migration.

**2. Outside this diff, but the same verify-before-link flow, and it looks serious.**

`reclaimUnverifiedLocalAuthIfNeeded` (`apps/web/modules/ee/sso/lib/sso-recovery.ts:89-100`) sets `password: null` on the **`User`** row. Post-ENG-1054 the credential hash does not live there — `apps/web/lib/user/password.ts:10` says so outright ("Post-ENG-1054 the password lives here, NOT on `User.password`"), and `emailAndPassword` in `auth.ts` verifies against `Account.password`. Nothing in `modules/ee/sso` touches the credential `Account` row; the only `account.delete` in the flow is `account-linking.ts:91`, for the legacy row.

So the comment above it — "strip any untrusted local auth factors before the SSO account becomes the canonical way back in" — does not hold, and the same write sets `emailVerified: true`, which is what turns the leftover hash from inert into usable:

1. On an instance with signup enabled, an attacker signs up natively as `victim@corp.com` and sets a password. `identityProvider=email`, `emailVerified=false`, and `requireEmailVerification` blocks sign-in. `Account(provider="credential", password=<attacker hash>)` now exists.
2. The real owner signs in via SSO and completes verify-before-link. This function runs: `emailVerified: true`, `password: null` on `User`. The credential `Account` row is untouched.
3. `POST /api/auth/sign-in/email` with the attacker's password now succeeds — the only gate that was holding it back was email verification, and the owner's recovery just satisfied it.

Same transaction already has `tx`, so deleting (or nulling `password` on) the `provider="credential"` row alongside the `User` update is a small change. Worth its own ticket with an integration case asserting `signInEmail` fails after recovery, since a mocked test cannot see which table the hash is in — the same blind spot that let ENG-2555 ship.

---

> [!NOTE]

---

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
```diff
@@ -0,0 +1,60 @@
...
+ *   outcome we want — no silent merge.
+ * - Only the `issuer` column is touched; `userId`, `provider` and `providerAccountId` are untouched, so
+ *   no row can move between users or organizations.
+ */
+export const repairAccountIssuer: MigrationScript = {
+  type: "data",
+  id: "ghqt118bfqdjs8tpzv1dvbqi",
+  name: "20260821165535_repair_account_issuer",
+  run: async ({ tx }) => {
+    const repaired = await tx.$executeRaw`
```
kiln-agent, line 39: The repair migration itself has no automated coverage — it's the one piece of this PR verified only by hand.

That matters more here than it usually would, because this migration is a second, independent transcription of the same `CASE` that the PR argues must never drift. `constants.test.ts` pins the helper against the *ENG-2343* SQL and against upstream, but nothing pins this file: swap `IS DISTINCT FROM` for `<>`, or drop the google arm, and it silently stops repairing rows while `pnpm test` stays green. That is the exact failure mode — green suite, broken sign-in — this PR exists to fix.

`apps/web/integration/credential-backfill.integration.test.ts` already shows the pattern for driving a migration module against real Postgres, and you've already done this run manually, so the test is mostly the seed data you used: a stomped `local:oauth:google` row, a correct github row, a correct credential row, a NULL-issuer openid row → assert the google and NULL rows are canonical afterwards and the others are byte-identical, then run it twice and assert the second run changes nothing.

---

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
```diff
@@ -0,0 +1,60 @@
...
+      UPDATE "Account"
+      SET "issuer" = CASE
+        WHEN "provider" = 'credential' THEN 'local:credential'
+        WHEN "provider" = 'google' THEN 'https://accounts.google.com'
+        ELSE 'local:oauth:' || "provider"
```
kiln-agent, line 44: Nit, and pre-existing rather than introduced here: this `ELSE` arm is `'local:oauth:' || "provider"`, while both `ssoAccountIssuer` and upstream's `createOAuthAccountIssuer` percent-encode the provider id. Unreachable for every provider id in use today, and inherited verbatim from the ENG-2343 backfill (which is the point — it's a faithful mirror).

Worth noting only because the new drift guard doesn't cover it: `constants.test.ts` checks the `ELSE` arm against four hardcoded ids (`github`, `azuread`, `openid`, `saml`), all encoding-neutral, so a future provider id with a reserved character would diverge silently — and this migration would then *un-fix* rows that `ssoAccountIssuer` had written correctly. The `percent-encodes a provider id that needs it` test pins the helper's behaviour but has no SQL-side counterpart.

Not asking for a change to the SQL — mirroring ENG-2343 exactly is the right call. A sentence in the docblock saying the mirror deliberately includes the missing encoding would keep the next reader from "fixing" one side.

---

apps/web/modules/ee/sso/lib/constants.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  const { arms, elseTemplate } = parseIssuerCase();
+
+  // Guard the guard: if the regex silently matched nothing, every assertion below would pass against an
+  // empty object and prove precisely nothing.
+  test("the migration's CASE parsed, and still special-cases exactly credential and google", () => {
+    expect(Object.keys(arms).sort()).toEqual(["credential", "google"]);
+    expect(elseTemplate).toBe("local:oauth:");
+  });
+
+  test.each(Object.entries(parseIssuerCase().arms))("agrees with the SQL for %s", (provider, expected) => {
```
kiln-agent: `parseIssuerCase()` runs a second time here (re-reading and re-parsing the migration file) when `arms` from line 52 is already in scope.

Minor, but it slightly weakens the "guard the guard" test above it: that test validates the *line 52* parse, while these cases assert against a separate one. Same file, same result today — just `Object.entries(arms)` and the two are the same object.

---

kiln-agent (commented):
## The fix is correct

I checked the premise against the installed Better Auth 1.7.0 rather than taking the description on trust, and it holds on every leg:

- `better-auth/dist/oauth2/account-key.mjs` resolves the issuer as `provider.accountIssuer ?? createOAuthAccountIssuer(provider.id)`, and `findAccountByKey` filters on `(issuer, accountId)` **only** — no `providerId` anywhere in the predicate. A wrong `issuer` really is invisible at sign-in, exactly as described.
- `@better-auth/core/dist/social-providers/google.mjs` declares `accountIssuer: "https://accounts.google.com"`; `github.mjs` declares none. The only other declaring built-ins (apple, facebook, line, cognito, entra-id, paybin) aren't configured in this repo, and `azuread`/`openid`/`saml` pin `accountIssuer` themselves via `ssoAccountIssuer` in `better-auth-providers.ts`. So the three-arm map in `canonicalAccountIssuer` is **complete**, not just correct for the reported case.
- The two-helper split is the right shape — *what we pin* and *what an existing row must carry* genuinely are different questions, and collapsing them is what produced the bug.
- The branch-3 healing write is the substantive part, and it converges: `syncSsoIdentityForUser`'s only account lookup is by `provider_providerAccountId`, which is issuer-independent, so the repair path can always find the row it needs to fix.
- Unique-index safety holds. The provider→issuer map is injective and `@@unique([provider, providerAccountId])` already forbids two rows per provider+subject, so a collision on `@@unique([issuer, providerAccountId])` is unreachable — and the 500 the bug produced means no correct-issuer duplicate can exist to collide with either.

`constants.test.ts` is the right test to have written: parsing the SQL instead of restating it, and asserting against upstream's own exports, is what turns "these four sites agree today" into something that stays true.

**Checks I ran locally:** all 161 unit tests under `modules/ee/sso` pass, `tsc --noEmit` clean for every changed file in both `apps/web` and `packages/database`, ESLint and Prettier clean, and the new migration builds to `dist/migration/…/migration.js` so the runner auto-discovers it — its `DataMigration` id and name are both unique across all migrations. Integration tests I could not run (no Docker in this environment), so the real-Postgres legs rest on your run.

## One correction to the PR description

Under **Risks**, the rolling-deploy remedy — *"the migration is idempotent, so re-running it after the rollout settles is the cheap answer"* — is a silent no-op as written. `runDataMigration` short-circuits on `status = 'applied'` (`migration-runner.ts:304`) and there's no force flag, so a second `db:migrate:deploy` logs `already completed. Skipping...` and touches nothing. Re-running it would need the `DataMigration` row deleted by hand.

Which is fine, because the actual answer is better than the one you wrote down: a row re-stomped during the rollout window is healed by the **new branch-3 write** on the next sign-in — that's precisely the property this PR adds. Worth fixing the note so nobody reaches for a re-run that won't do anything.

## Findings

Three inline, none blocking. The first is the one I'd actually like addressed: the repair migration is the only piece of this change verified by hand alone, and it's a second independent transcription of the `CASE` that the rest of the PR goes to real lengths to pin. The other two are nits (an un-pinned encoding asymmetry in the SQL mirror, a redundant re-parse in the new test).

Not approving — I'll leave that to @mattinannt — and the staging check against real Google is still the one that closes this out, as your **Open gaps** already says.

---

xernobyl:
**Verified against a real Google identity.** This was the one gap left on this PR, and it's now closed locally rather than deferred to staging.

Setup: throwaway database, the dev Google OAuth client (registered for `localhost:3000`), a real Google Workspace account, and an existing user carrying that email with **no** Google link — the exact staging state.

| step | result |
| --- | --- |
| Google sign-in for the existing user | `account_not_linked` → verify-before-link recovery email. **No account row, no session** before the inbox was proven — ENG-554 holding |
| Completing the emailed link | `provider=google`, **`issuer=https://accounts.google.com`**, linked to the right user, `identityProvider` flipped. Asserted equal to `google({…}).accountIssuer` read from upstream in the same run. **Pre-fix this exact write produced `local:oauth:google`** |
| **Loop test** — sign in with Google again | Landed past login, not on "confirm your email". Still exactly **1** account row, issuer unchanged, and **0** new recovery emails |

That last row is the reported symptom directly refuted: on staging every attempt bounced back through recovery and re-sent an email. Until now the google case rested on reading upstream's source plus the mutation-checked integration tests; it's measured.

**Two incidental findings, neither related to this diff** — recording them because they cost real diagnosis time:

1. **`unable_to_create_user` masks the SSO provisioning gate.** My first attempt failed there: with `SKIP_INVITE_FOR_SSO` unset, a single-org non-fresh instance requires a matching invite for SSO sign-up, so an uninvited Google identity is refused. Correct behaviour, and it **rolled back cleanly** — 0 users, 0 accounts, 0 sessions, confirming the hook's orphan-safe `return false` against real Google. But the user-facing error is the generic `unable_to_create_user`, with the actual reason only in the server log. For a self-hoster hitting the invite gate that is close to undiagnosable — same theme as ENG-2558, worth folding in there.
2. Two failures during the run were **my test data, not the code**: hand-written user ids (`usr_tiago`) fail the app's cuid2 validation via `getLocale` in the root layout, and a bare seeded org with 0 Workspaces gives "Error loading resources" after login. Both post-authentication; neither touches this change.

Teardown: scratch DB dropped (the real Google identity with it), Mailhog cleared, `:3000` released, and the uncommitted `sso` entitlement override reverted — tracked tree verified clean before this comment.

apps/web/modules/ee/sso/lib/constants.ts
```diff
@@ -2,20 +2,53 @@ export const OAUTH_ACCOUNT_NOT_LINKED_ERROR = "OAuthAccountNotLinked";
...
+ */
+export const canonicalAccountIssuer = (provider: string): string => {
+  // Better Auth's own `createLocalAccountIssuer("credential")`. Unreachable from the SSO recovery
+  // caller (its provider is an `IdentityProvider`, which has no `credential` member) but kept so this
+  // stays a faithful mirror of the SQL, and so a future non-SSO caller cannot get it wrong.
+  if (provider === "credential") return "local:credential";
+  // Declared upstream in `@better-auth/core/dist/social-providers/google.mjs`.
+  if (provider === "google") return "https://accounts.google.com";
+  return ssoAccountIssuer(provider);
+};
```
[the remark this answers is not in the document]
xernobyl: Fixed in `0fd431c`, both halves. The docblock no longer claims "byte for byte" — it now names the asymmetry (SQL ELSE concatenates raw, the helper percent-encodes, identity for every id in use) and says a provider id ever needing escaping must get an explicit CASE arm in SQL. And your suggested test is in, slightly strengthened: it asserts the helper's encoded value AND that it differs from the raw ELSE-arm concatenation, so the divergence reads as documented-deliberate rather than an oversight.

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
xernobyl: Fixed in `0fd431c` — this was the best catch of the round. `parseIssuerCases()` now parses every `"issuer" = CASE` / `IS DISTINCT FROM (CASE` block across BOTH migration files and asserts each copy identical to the canonical parse, with a guard that the repair file contains exactly two copies (so dropping the WHERE copy — which would silently weaken idempotency — also fails). Mutation-checked with your named drifts: a trailing slash on google in the repair's SET copy fails 1 test, and dropping the google arm from only the WHERE copy (SET/WHERE diverging) fails 1. The constants.ts docblock's pin claim now says "every SQL copy in both migrations" instead of overstating what the old test covered.

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
xernobyl: Fixed in `0fd431c` — `apps/web/integration/repair-account-issuer.integration.test.ts`, following the credential-backfill pattern as you suggested. Four cases reproducing the manual matrix: stomped-google + NULL-issuer rows repaired while canonical github/credential rows stay byte-identical; idempotency (second run changes nothing); no-op on an empty table; and the repaired row resolving by Better Auth's own `(issuer, providerAccountId)` unique — the property the ticket is about. Mutation-checked against real Postgres with your two named mutations: neutralising the google arms fails 2 tests, and `IS DISTINCT FROM` → `<>` fails the NULL-row case specifically. Exactly the green-suite-broken-signin shape, now bound.

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
xernobyl: Done in `0fd431c` — kept the SQL untouched as you asked, and both docblocks now state the mirror deliberately includes the missing encoding ("'fixing' one side is how ENG-2555 happened", which is the sentence I want the next reader to trip over). The new divergence test in constants.test.ts additionally pins `team/github` → the helper encodes, the ELSE arm can't express it — so the encoding-neutral assumption is now asserted rather than implicit.

apps/web/modules/ee/sso/lib/constants.test.ts
xernobyl: Fixed in `0fd431c` — fell out of the restructure for the cross-file pin: the per-arm cases now assert against the same in-scope `arms` object the guard-the-guard test validates, and the canonical parse happens exactly once.

packages/database/migration/20260821165535_repair_account_issuer/migration.ts
```diff
@@ -0,0 +1,60 @@
...
+      END
+      WHERE "issuer" IS DISTINCT FROM (
+        CASE
+          WHEN "provider" = 'credential' THEN 'local:credential'
+          WHEN "provider" = 'google' THEN 'https://accounts.google.com'
+          ELSE 'local:oauth:' || "provider"
+        END
+      )
+    `;
```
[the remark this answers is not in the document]
xernobyl: Declining the batching, with the reasoning on the record — the suggestion as written doesn't achieve its stated goals inside this runner, and the underlying risk is bounded.

**Why batching inside `tx` buys nothing here.** The runner wraps every data migration in ONE interactive transaction with a 30-minute timeout (`migration-runner.ts:41,334`), and the `DataMigration` bookkeeping commits in that same transaction. So a loop of 1000-row UPDATEs on `tx`: (a) is **not** resumable — a failure rolls back every batch plus the applied-marker, identical to the single statement; (b) does **not** release locks earlier — row locks are held until the one commit either way; (c) does not escape the 30-minute ceiling, which covers the whole `run`. "Progress is monotonic" only holds across transactions, and there is exactly one. The house pattern for genuinely high-volume work is the other lever entirely: the autocommit `prisma` handle with chunked commits, as `20260625104904_canonicalize_language_codes` does — trading atomicity for resumability.

**Why that trade isn't warranted here.** The matched set is structurally tiny: the only writer of a non-canonical issuer is the SSO verify-before-link recovery path on 1.7 code (google links made through it, plus any stray NULL), and no released version ever ran that code. The ENG-2343 backfill — a single unbounded UPDATE over the **entire** Account table, strictly more work than this — ran in this same pipeline one migration earlier. On "never load a whole table into memory": nothing is loaded — it is one set-based statement, zero rows in app memory.

And the migration now has real-Postgres integration coverage (`repair-account-issuer.integration.test.ts`, added this round), so its behaviour — including idempotency and the NULL case — is asserted rather than argued. If a reviewer still prefers chunked-autocommit for defence in depth I'll switch it to the `prisma`-handle pattern, but I'd rather keep single-transaction atomicity for a repair whose expected row count is single-digits.

xernobyl:
All six review comments addressed in `0fd431c` — five fixed, one declined with reasoning on its thread. All threads replied to and resolved.

**Fixed:**

- **The unpinned fourth spelling** (kiln's strongest catch): `constants.test.ts` now parses every `CASE` copy in **both** migrations — including the repair's `SET`/`WHERE` pair, with a copy-count guard — and asserts each identical to the canonical parse. Mutation-checked with the exact drifts named in review: a trailing slash in the repair's `SET`, and the `SET`/`WHERE` pair diverging. Both fail now; both were green before.
- **Integration coverage for the repair migration** — previously the one piece verified only by hand. Four real-Postgres tests reproduce the manual matrix, ending with the repaired row resolving by Better Auth's own `(issuer, providerAccountId)` key. DB-level mutations: neutralised google arms fail 2, `IS DISTINCT FROM` → `<>` fails the NULL-row case — the two silent-breakage shapes named in review.
- **The encoding asymmetry** (both reviewers): docblocks in the helper and the migration now state the SQL mirror deliberately includes the missing percent-encoding, and a test pins `team/github` as the input class where the two sides part ways — documented-deliberate rather than latent.
- The double `parseIssuerCase()` nit fell out of the restructure.

**Declined, on its thread:** the batched-UPDATE suggestion. Short version: the runner wraps the whole `run` in one 30-minute interactive transaction, so batches on `tx` are neither resumable (rollback undoes all of them, marker included) nor lock-shortening (held to the single commit) — the stated benefits only exist across transactions, and there is one. The house lever for genuinely high-volume work is the autocommit `prisma` handle (see `canonicalize_language_codes`), and the matched set here is structurally single-digits: the only writer of a non-canonical issuer is the 1.7 recovery path, which no released version ever ran. Happy to switch to chunked-autocommit if a human reviewer prefers it; full reasoning on the thread.

**Full re-review + fresh smoke on `0fd431c`:** every upstream citation and invariant re-verified (suppression set, gate shape, no `.details` reads — unchanged; this round touched tests, docblocks and nothing behavioural). Gates: typecheck 27/27, lint 0 errors, format clean, unit 8503 passing, integration 96/98 (the 2 are the known local-env password-reset failures in other files). Live smoke re-run on the final commit: SSO registration → account `local:oauth:openid` → repeat login (1 account row, straight through) → logout (200, `get-session` null). License-gate override used for the smoke was reverted before committing; tracked tree verified clean.

@kiln-agent could you re-run over `0fd431c`? The delta is exactly your findings plus CodeRabbit's two. @coderabbitai review

xernobyl:
Closing out the two items that lived in **review bodies** rather than inline threads, so they never showed as unresolved and I had missed them. All six inline threads were already fixed/declined and resolved in `0fd431c`.

**1. The Risks note was wrong — corrected in the description.** @kiln-agent flagged that *"the migration is idempotent, so re-running it after the rollout settles is the cheap answer"* is a silent no-op, and that checks out: `runDataMigration` returns early on `status = 'applied'` (`packages/database/src/scripts/migration-runner.ts:304-307`, logging `already completed. Skipping...`) and there is no force flag anywhere in that file. A second `db:migrate:deploy` would touch nothing, which is a bad thing to have written down for someone reaching for it mid-incident.

The replacement says what actually covers it, which is better than what I'd written: a row re-stomped during the rollout window is healed by the **branch-3 write this PR adds**, on the next sign-in, with no operator action. Re-running would require deleting the `DataMigration` row by hand.

**2. The `reclaimUnverifiedLocalAuthIfNeeded` finding is fixed, in its own PR.** Also from @kiln-agent, outside this diff but the same verify-before-link flow: the strip nulls `User.password` while the live credential sits on `Account.password` post-ENG-1054, so an attacker-set password survives recovery — and the same write flips `emailVerified: true`, which is what makes it usable.

Confirmed real and tracked as ENG-2557, fixed in #8962. Both of the remedies suggested here were taken: the credential is cleared in the same transaction (nulled rather than deleted, so the surviving row still marks the user as a password user and `forgotPasswordAction` can offer them a reset — deleting it traded the takeover for a lockout), and there is an integration case asserting `signInEmail` returns 401 after recovery. That test was verified red against `origin/main`, plus a live before/after through the real HTTP routes: **200 signed-in before, 401 after**. The point about a mocked test being unable to see which table the hash is in was the right one — it is exactly why that assertion goes through `auth.api.signInEmail` against real Postgres instead of reading a column.

Thanks for both — the second one was the more valuable find of the two, and it was outside the diff.

Dhruwang (approved):

PR #8952, opened 2026-08-21
fix(sso): write the canonical Account.issuer so Google sign-in stops looping (backport 5.4)

xernobyl:
Ref ENG-2555

## What & why

Backport of #8951 to `release/5.4` — Google SSO is broken on any build carrying #8835 (Better Auth 1.7), which `release/5.4` does: linking Google for an existing user writes `Account.issuer = local:oauth:google`, a key Better Auth never looks under, so every sign-in loops back through verify-before-link recovery forever. Root cause, blast radius and evidence are on #8951 and the linked ticket.

No cherry-pick drift is possible on this one: `release/5.4`'s tip is `ab8fc21ff`, the exact commit the fix branch was cut from, so this branch is **byte-identical** to #8951 — the same four commits on the same base. Fix, canonical-issuer helper, the self-healing update branch, the drift-pinning tests, the data-repair migration, and the real-Postgres integration coverage all come along unchanged.

Per the backport policy: one fix, minimal diff (identical to the reviewed PR), critical — it breaks Google sign-in for every existing user on the release.

## Breaking changes

None — repairs data written by unreleased code and restores the intended lookup; no API, schema shape, env or config changes.

## Migrations & env

- `20260821165535_repair_account_issuer` (data): rewrites `Account.issuer` rows that disagree with the canonical value. Idempotent, no-op on correct/empty databases, baselined without running on fresh ones. Same migration as #8951, unchanged.
- No env changes.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Everything | see #8951 | This branch is byte-identical to #8951 (same commits, same base — verified `merge-base = ab8fc21ff = release/5.4 tip`), so its full coverage table applies verbatim: real-Google end-to-end incl. the loop test, real-Postgres integration + mutation checks on the repair migration, the three-way drift pins, and the live SSO smoke matrix |

**Open gaps**

- Same as #8951: the staging pass after deploy (different Google OAuth client there).
- CI on this PR is the only thing not shared with #8951; wait for it before merging.

**Risks**

- Same as #8951. Merge order note: whichever of this and #8951 lands second is a no-op content-wise, but both must land — main and the release each need the migration.

---

> [!NOTE]


kiln-agent (commented):

The backport is sound. Because the description asks the reviewer to trust #8951's coverage, I re-derived the parts that carry the risk rather than taking the identity claim on trust.

### The identity claim holds, stronger than stated

Not just "byte-identical" — it is **the same commit object**:

```
git rev-parse HEAD pr8951      → 0fd431c… / 0fd431c…   (identical SHA, not just identical tree)
git merge-base HEAD release/5.4 → ab8fc21ff = release/5.4 tip
git log release/5.4..HEAD       → the same 4 commits
```

So the diff needs no separate content review, and `release/5.4` does carry #8835 (`better-auth 1.7.0` in `apps/web/package.json` on this base, untouched by the 4 commits) — the precondition for the bug being live on the release.

### Root cause and fix, verified against the installed library

- `node_modules/@better-auth/core/dist/social-providers/google.mjs:57` declares `accountIssuer: "https://accounts.google.com"`.
- `better-auth/dist/oauth2/account-key.mjs` resolves `issuer = provider.accountIssuer ?? createOAuthAccountIssuer(provider.id)`, then keys on `(issuer, accountId)`.
- `ssoSocialProviders` registers google/github as **built-in** providers with no `accountIssuer` override, so upstream's declared value applies — the synthetic form was genuinely unreachable by BA's lookup. Fix is right.

**Provider coverage is complete, not just correct for google.** `normalizeSsoProvider` (`provider-normalization.ts`) constrains everything reaching `canonicalAccountIssuer` to a closed set of five, and `account-linking.ts` is the only writer of `Account.issuer` in the tree:

| provider | canonical | why |
| --- | --- | --- |
| `google` | `https://accounts.google.com` | declared upstream |
| `github` | `local:oauth:github` | built-in, declares none → upstream fallback |
| `azuread` / `openid` / `saml` | `local:oauth:<id>` | genericOAuth, `accountIssuer` pinned at `better-auth-providers.ts:219/249/281` |

`email` cannot reach it (not in `SSO_PROVIDER_MAP`), and `credential` is handled.

**The repair's `ELSE` arm cannot corrupt anything.** The only `local:` issuers Better Auth 1.7 writes are `credential` and `siwe` (grep of `createLocalAccountIssuer` call sites across the installed package); `siwe`/`phoneNumber`/`emailOTP`/`admin` are not in `auth.ts`'s plugin list, so no row exists that the raw `'local:oauth:' || "provider"` concat would rewrite wrongly.

**Migration mechanics.** Auto-discovery confirmed (`loadMigrations` scans dirs for `migration.ts`/`migration.js`, no registry to update); `dist/migration/20260821165535_repair_account_issuer/migration.js` is emitted by the build, so the Docker path resolves it. `id` is unique across all 26 data migrations. Timestamp sorts after `20260812110000`, so the `issuer` column exists when it runs. The injectivity argument for `@@unique([issuer, providerAccountId])` checks out: any row already holding the target pair would have to share `(provider, providerAccountId)`, which the other unique index forbids.

### The tests can actually fail

Four mutations, all caught:

| Mutation | Result |
| --- | --- |
| `constants.ts:57` google arm → `ssoAccountIssuer(provider)` | 8 failed / 2 files |
| `account-linking.ts:133` drop `issuer` from the self-heal branch | 2 failed |
| `migration.ts:50` `IS DISTINCT FROM` → `<>` | 1 failed |
| `migration.ts:44` drop the google arm from the `SET` `CASE` | 1 failed |

`constants.test.ts` + `account-linking.test.ts` + `better-auth-providers.test.ts`: 46 passing on this SHA locally, and all 35 checks (incl. **Better Auth integration tests**, which glob `**/*.integration.test.ts` and so pick up both new integration files) are green on the head commit.

---

## Finding (non-blocking, suggestion)

**The drift guard is keyed to a hardcoded provider list, so it will not catch the *next* instance of this exact bug.**

`constants.test.ts` pins google and github by name, and the `ELSE`-arm test iterates a literal `["github", "azuread", "openid", "saml"]`. Enabling a third built-in social provider — `apple` (`https://appleid.apple.com`), `facebook`, `cognito`, or `microsoft` (whose `accountIssuer` is a *function*, `({ profile }) => profile.iss`, which neither the helper nor the SQL can express) — reintroduces ENG-2555 with a fully green suite.

That is the same failure mode this PR is fixing: the ENG-2343 SQL comment already said *"Any future built-in social provider must be checked the same way before it is enabled"*, and a comment did not stop the drift. Worth making it executable. `@better-auth/core/social-providers` exports a `socialProviders` registry, so it can be derived rather than listed — I verified this passes today and does bite on `apple`:

```ts
import { socialProviders } from "@better-auth/core/social-providers";

// mirrors SSO_PROVIDER_MAP's values — every id that can reach canonicalAccountIssuer
test.each(["google", "github", "azuread", "openid", "saml"])(
  "%s does not disagree with an upstream built-in of the same id",
  (provider) => {
    const builtIn = socialProviders[provider];
    if (!builtIn) return; // generic + pinned (azuread/openid/saml)
    const declared = builtIn({ clientId: "x", clientSecret: "x" }).accountIssuer;
    if (declared === undefined) return; // github
    expect(canonicalAccountIssuer(provider)).toBe(declared);
  }
);
```

Severity is low — it needs a deliberate change across `IdentityProvider`, `SSO_PROVIDER_MAP`, `ssoSocialProviders` and the SQL, not an accidental edit — and it is arguably follow-up work rather than backport scope. Since this branch must stay identical to #8951, if you take it, take it there and re-cut, or file it. I would not hold the release for it.

## Notes

- **This PR is still a draft** while carrying a review request. CI is green on the head commit, so nothing is left of the "wait for CI before merging" gap in the description — it can be marked ready.
- **Coverage table.** For a same-commit backport, deferring to #8951 is defensible, and I verified the identity that makes it so. Strictly, the template's `How` column takes one of `unit (red on main) / unit (mutation) / unit (guard) / e2e / manual`, and `see #8951` is not one of them. A one-line note that the SHA is identical (`git rev-parse` output above) would satisfy the same reviewer need without restating #8951's table.
- **Breaking-changes box correctly unticked.** No API/SDK shape, route, env var, config key, default or webhook payload changes; the data migration needs no manual action and is a no-op on both a fresh DB (baselined) and an already-correct one. The v5.4 rollback note in `docs/self-hosting/advanced/migration.mdx:150` stays accurate — repaired values are still additive and unread by v5.3.
- **Merge order.** Both PRs adding the same directory with the same migration `id` is what makes the second landing free: `runDataMigration` keys on that id and returns early on `status = 'applied'`. `release/5.4` is currently an ancestor of `main` (1 commit behind), so the eventual converge is clean either way.

---

Dhruwang (approved):
Thank you for the PR. Looks good 🚀. I just stripped out the additional test.

PR #8955, opened 2026-08-24
fix(contacts): gate contact write paths and attribute-key APIs on the contacts entitlement [ENG-2295]

pandeymangg:
Fixes ENG-2295

## What & why

[ENG-1768](https://linear.app/formbricks/issue/ENG-1768) gated the **read** side of contacts (EE) on the `contacts` entitlement. The **write** side of the same module never got the gate: an authenticated member of an org *without* the entitlement could still mint personal survey links, bulk-import contacts from CSV, edit contact attributes, manage attribute keys, enumerate distinct attribute values, and drive the v1 by-id / v2 attribute-key management APIs. Tenancy was enforced everywhere — this is a licence bypass only, no cross-tenant angle.

The by-id v1 route even carried a **dead catch** matching the "Contacts are only enabled for Enterprise Edition…" message that nothing throws anymore — the check was evidently lost in a refactor. That is why this PR adds a **shared guard** at the module boundary instead of seven more inline checks ([ENG-2507](https://linear.app/formbricks/issue/ENG-2507)'s "close the class, not the instance"):

- **New:** two guard files, split by who may import them. `modules/ee/contacts/lib/contacts-entitlement.ts` — `ensureContactsEnabled(organizationId)` (throws `OperationNotAllowedError`, for the module's server actions) plus the v1 message as a shared constant; only EE code imports it. `modules/ee/license-check/lib/contacts-api-guard.ts` — `checkContactsEnabledApiV2` (returns the `forbidden` error object, for v2 handlers); it lives in license-check because the v2 attribute-key routes under the OSS `modules/api/v2` path consume it, and license-check is the sanctioned EE import boundary for OSS code (per review — the wider `modules/api/v2` → `modules/ee/contacts` imports that already exist on `main` are left for a separate cleanup).
- **Gated server actions:** `generatePersonalSurveyLinkAction`, `createContactsFromCSVAction`, `updateContactAttributesAction`, `deleteContactAction`, attribute-key create/update/delete, `getDistinctAttributeValuesAction` (this one via the segments file's existing `checkAdvancedTargetingPermission`, like every sibling there). `getContactsAction`'s existing inline check is migrated to the shared guard — same message, one way to do it.
- **Gated API routes:** v1 `contact-attribute-keys/[id]` GET/PUT/DELETE (one check inside `fetchAndAuthorizeContactAttributeKey` covers all three; dead catch removed), v2 `contact-attribute-keys` GET/POST and `[id]` GET/PUT/DELETE.
- `deleteContactAction` is not in the ticket's list, but v1 management contact DELETE already gates — leaving the UI action ungated would have kept it the module's one open write path, so it is gated for consistency.

Error strings are kept verbatim (the v1 EE-upgrade message, the action message `getContactsAction` already used), so no consumer-visible message changes for entitled orgs.

## Breaking changes

Breaking only for organizations **without** the contacts entitlement — this enforces the licence the gated sibling routes already enforce, and entitled orgs see no change in shape, status or message. It is ticked because the emitted value of public management endpoints changes for unlicensed consumers, which is exactly what the label, release notes and self-hoster migration guide exist to announce.

| `GET/PUT/DELETE /api/v1/management/contact-attribute-keys/{id}` | 2xx with data for orgs without the contacts entitlement | 403 "Contacts are only enabled for Enterprise Edition, please upgrade." | Self-hosters / API consumers without the contacts entitlement | Obtain the contacts entitlement, or stop calling these endpoints |
| `GET/POST /api/v2/management/contact-attribute-keys` and `GET/PUT/DELETE …/{id}` | 2xx with data for orgs without the contacts entitlement | 403 `forbidden` error (`Contacts feature is not enabled for this organization`) | Same | Same |

The gated web-app actions (CSV import, attribute editing, personal links, contact delete) are not listed: server actions are not a public API surface, and the contacts UI is already entitlement-gated since ENG-1768.

## Migrations & env

- none

## How this was tested

**Coverage**

Up front: **this bug class has no red-on-main seam under the repo's test conventions.** The missing checks lived in server actions and route handlers, and neither `actions.ts` nor `route.ts` files are unit-tested here (both are in the Sonar coverage exclusions; the convention is to test what they call). The guard itself is new code, so its rows are mutation-proven instead; the wiring — "each action/route calls the guard first" — is the review's job, plus the diff is mechanically uniform (same one-liner after the authz block in every action, same two patterns in the routes as their gated siblings).

| Behaviour | How | Outcome |
| --- | --- | --- |
| `ensureContactsEnabled` throws `OperationNotAllowedError` with the exact message when the entitlement is missing, resolves when present | unit (mutation) | `modules/ee/contacts/lib/contacts-entitlement.test.ts`. Inverting the check at `contacts-entitlement.ts:24` → 2/2 tests fail. |
| `checkContactsEnabledApiV2` returns the `forbidden` error object when missing, `null` when present | unit (mutation) | `modules/ee/license-check/lib/contacts-api-guard.test.ts`. Inverting the check at `contacts-api-guard.ts:19` → 2/2 tests fail. |
| `getContactsAction`'s existing read gate (ENG-1768) is behaviour-identical after migrating to the shared guard | unit (guard) | The guard test asserts the exact error type and message the action previously threw inline (`OperationNotAllowedError`, "Contacts are not enabled for this organization"), so callers see byte-identical failures. The untouched ENG-1920 suite (`segments/actions.test.ts`) still passes against the modified `segments/actions.ts`. |
| Nothing else regressed | unit (guard) | `pnpm test --filter=@formbricks/web`: 641/641 files green. `pnpm typecheck --filter=@formbricks/web` 12/12. ESLint on all touched files: clean. |

**Open gaps**

- **The wiring of the guard into the 8 actions and 8 route handlers has no automated test** — that is the repo convention for actions/routes, not an oversight, but it is exactly how the original check got lost (the dead catch is the fossil). The shared guard reduces the surface to "is the one-liner present", which a reviewer can check per call site from the diff alone.
- **No e2e row.** The Playwright environment runs with `ENTERPRISE_LICENSE_KEY` set (`e2e.yml`), so the entitlement-off behaviour cannot be exercised there at all.
- The dead catch's removal rests on a repo grep: the EE-upgrade message is still *returned* by other v1 routes as an explicit `forbiddenResponse`, but no code path **throws** it as an `Error`, which is all the catch matched.

**Risks**

- An org that loses the entitlement can no longer delete contacts or attribute keys from the app (the v1 management API already refused contact DELETE this way; the module is now uniform). If a downgraded org needs PII removed, that goes through the existing support path — unchanged for the API, new for the UI action.
- v2 attribute-key endpoints now return 403 for unlicensed orgs where they previously returned data; any such consumer was relying on the licence gap this PR closes.

---

> [!NOTE]

### DIFF
--- a/apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
+++ b/apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
@@ -13,6 +13,7 @@ import {
   ZContactAttributeKeyUpdateSchema,
 } from "@/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/types/contact-attribute-keys";
 import { ApiErrorResponseV2 } from "@/modules/api/v2/types/api-error";
+import { checkContactsEnabledApiV2 } from "@/modules/ee/contacts/lib/contacts-entitlement";
 import { hasPermission } from "@/modules/organization/settings/api-keys/lib/utils";
 
 export const GET = async (
@@ -27,6 +28,11 @@ export const GET = async (
     externalParams: props.params,
     handler: async ({ authentication, parsedInput }) => {
       const { params } = parsedInput;
+
+      const contactsNotEnabledError = await checkContactsEnabledApiV2(authentication.organizationId);
+      if (contactsNotEnabledError) {
+        return handleApiError(request, contactsNotEnabledError);
+      }
 
       const res = await getContactAttributeKey(params.contactAttributeKeyId);
 
@@ -61,6 +67,11 @@ export const PUT = async (
 
       if (auditLog) {
         auditLog.targetId = params.contactAttributeKeyId;
+      }
+
+      const contactsNotEnabledError = await checkContactsEnabledApiV2(authentication.organizationId);
+      if (contactsNotEnabledError) {
+        return handleApiError(request, contactsNotEnabledError, auditLog);
       }
 
       const res = await getContactAttributeKey(params.contactAttributeKeyId);
@@ -133,6 +144,11 @@ export const DELETE = async (
 
       if (auditLog) {
         auditLog.targetId = params.contactAttributeKeyId;
+      }
+
+      const contactsNotEnabledError = await checkContactsEnabledApiV2(authentication.organizationId);
+      if (contactsNotEnabledError) {
+        return handleApiError(request, contactsNotEnabledError, auditLog);
       }
 
       const res = await getContactAttributeKey(params.contactAttributeKeyId);
--- a/apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
+++ b/apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
@@ -13,7 +13,7 @@ import {
   ZContactAttributeKeyUpdateSchema,
 } from "@/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/types/contact-attribute-keys";
 import { ApiErrorResponseV2 } from "@/modules/api/v2/types/api-error";
-import { checkContactsEnabledApiV2 } from "@/modules/ee/contacts/lib/contacts-entitlement";
+import { checkContactsEnabledApiV2 } from "@/modules/ee/license-check/lib/contacts-api-guard";
 import { hasPermission } from "@/modules/organization/settings/api-keys/lib/utils";
 
 export const GET = async (
--- a/apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
+++ b/apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
@@ -4,6 +4,8 @@ import { RequestBodyTooLargeError, parseJsonBodyWithLimit } from "@/app/lib/api/
 import { responses } from "@/app/lib/api/response";
 import { transformErrorToDetails } from "@/app/lib/api/validator";
 import { TApiKeyAuthentication, THandlerParams, withV1ApiWrapper } from "@/app/lib/api/with-api-logging";
+import { CONTACTS_API_V1_NOT_ENABLED_MESSAGE } from "@/modules/ee/contacts/lib/contacts-entitlement";
+import { getIsContactsEnabled } from "@/modules/ee/license-check/lib/utils";
 import { hasPermission } from "@/modules/organization/settings/api-keys/lib/utils";
 import {
   deleteContactAttributeKey,
@@ -14,15 +16,22 @@ import { ZContactAttributeKeyUpdateInput } from "./types/contact-attribute-keys"
 
 async function fetchAndAuthorizeContactAttributeKey(
   attributeKeyId: string,
-  workspacePermissions: NonNullable<TApiKeyAuthentication>["workspacePermissions"],
+  authentication: NonNullable<TApiKeyAuthentication>,
   requiredPermission: "GET" | "PUT" | "DELETE"
 ) {
+  // Entitlement first, matching the plural route: without the contacts feature the caller may
+  // not interact with attribute keys at all, regardless of workspace permissions.
+  const isContactsEnabled = await getIsContactsEnabled(authentication.organizationId);
+  if (!isContactsEnabled) {
+    return { error: responses.forbiddenResponse(CONTACTS_API_V1_NOT_ENABLED_MESSAGE) };
+  }
+
   const attributeKey = await getContactAttributeKey(attributeKeyId);
   if (!attributeKey) {
     return { error: responses.notFoundResponse("Attribute Key", attributeKeyId) };
   }
 
-  if (!hasPermission(workspacePermissions, attributeKey.workspaceId, requiredPermission)) {
+  if (!hasPermission(authentication.workspacePermissions, attributeKey.workspaceId, requiredPermission)) {
     return { error: responses.unauthorizedResponse() };
   }
 
@@ -42,7 +51,7 @@ export const GET = withV1ApiWrapper({
 
       const result = await fetchAndAuthorizeContactAttributeKey(
         params.contactAttributeKeyId,
-        authentication.workspacePermissions,
+        authentication,
         "GET"
       );
       if (result.error) {
@@ -55,14 +64,6 @@ export const GET = withV1ApiWrapper({
         response: responses.successResponse(result.attributeKey),
       };
     } catch (error) {
-      if (
-        error instanceof Error &&
-        error.message === "Contacts are only enabled for Enterprise Edition, please upgrade."
-      ) {
-        return {
-          response: responses.forbiddenResponse(error.message),
-        };
-      }
       return handleErrorResponse(error);
     }
   },
@@ -85,7 +86,7 @@ export const DELETE = withV1ApiWrapper({
     try {
       const result = await fetchAndAuthorizeContactAttributeKey(
         params.contactAttributeKeyId,
-        authentication.workspacePermissions,
+        authentication,
         "DELETE"
       );
 
@@ -132,7 +133,7 @@ export const PUT = withV1ApiWrapper({
     try {
       const result = await fetchAndAuthorizeContactAttributeKey(
         params.contactAttributeKeyId,
-        authentication.workspacePermissions,
+        authentication,
         "PUT"
       );
       if (result.error) {
--- a/apps/web/modules/api/v2/management/contact-attribute-keys/route.ts
+++ b/apps/web/modules/api/v2/management/contact-attribute-keys/route.ts
@@ -12,7 +12,7 @@ import {
 } from "@/modules/api/v2/management/contact-attribute-keys/types/contact-attribute-keys";
 import { resolveBodyIdsV2 } from "@/modules/api/v2/management/lib/workspace-resolver";
 import { ApiErrorResponseV2 } from "@/modules/api/v2/types/api-error";
-import { checkContactsEnabledApiV2 } from "@/modules/ee/contacts/lib/contacts-entitlement";
+import { checkContactsEnabledApiV2 } from "@/modules/ee/license-check/lib/contacts-api-guard";
 
 export const GET = async (request: NextRequest) =>
   authenticatedApiClient({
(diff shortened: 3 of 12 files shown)


kiln-agent (commented):
## Review

The gate itself is right and the coverage is complete — I re-derived the call-site list independently rather than trusting the description, and every write path into the six contact mutation functions is now behind an entitlement check. Three things below: one is a decision I think is answered wrong, two are cheap consistency fixes.

### 1. The breaking-changes box should probably be ticked

`- [ ] This PR contains breaking changes` is unticked, but the diff takes five v2 endpoints from 200/201 to 403 for a whole class of deployments:

- `GET|POST /management/contact-attribute-keys`
- `GET|PUT|DELETE /management/contact-attribute-keys/{id}`

`getIsContactsEnabled` → `getCustomPlanFeaturePermission` returns `false` on self-hosted whenever the licence is inactive or lacks the `contacts` feature (`modules/ee/license-check/lib/utils.ts:60-64`). So **every Community-Edition self-hoster currently using these v2 endpoints loses them on upgrade** — and unlike the v1 plural route, these were never gated, so that usage is real rather than hypothetical.

The template's test is consumer impact ("breaks something for API/SDK consumers or self-hosters"), not intent, and AGENTS.md notes the tick is the sole input to `breaking-change`, which feeds the release notes and the self-hoster migration guide. A CE operator who wakes up to 403s on `/management/contact-attribute-keys` is exactly the reader that guide exists for. The Risks section already spells the impact out — which is the argument for ticking, not against it.

I can see the counter-case (this is closing a licence bypass, so arguably nothing "changed" that was ever intended to work), and it's your call as author. But note CodeRabbit's "Breaking changes match the diff" check compares the tick against the diff, so leaving it unticked will likely need defending there too.

### 2. Denied writes now produce audit records with no org and no target

In the six `withAuditLogging`-wrapped actions the guard sits *above* the audit-context assignments:

```ts
await ensureContactsEnabled(organizationId);

ctx.auditLoggingCtx.organizationId = organizationId;   // never runs on denial
ctx.auditLoggingCtx.contactId = parsedInput.contactId;
```
(`modules/ee/contacts/actions.ts:84-87`, and the same shape at `:125`, `:190`, `attributes/actions.ts:64`, `:128`, `:177`)

`withAuditLogging` does log the failure — `status = "failure"` and the event is emitted before the rethrow (`modules/ee/audit-logs/lib/handler.ts:242-355`) — but with the context unset it resolves `organizationId` to `UNKNOWN_DATA` and `targetId` to `UNKNOWN_DATA` (`:245-262`, `:335`). The `workspaceId` fallback at `:250` cannot rescue `createContactsFromCSVAction` either: `organizationId` was already defaulted to the truthy `UNKNOWN_DATA` string one line up, so `if (!organizationId)` never fires. That's a pre-existing bug in the handler, not yours — but it means the fallback you might expect to cover this doesn't.

Net: on a licence-enforcement PR, the audit trail of every newly-denied privileged write says "unknown org deleted unknown contact". Moving the guard two lines down, below the `ctx.auditLoggingCtx.*` assignments, fixes all six at zero cost — and it's what this PR already does in the v2 routes, where the check is deliberately placed *after* `auditLog.targetId = params.contactAttributeKeyId`.

Fair caveat: the current placement matches `segments/actions.ts:327-328` exactly, so it's the existing house style and authz denials already behave this way. I'd still align the six, since the v2 half of this same diff makes the better choice.

### 3. The shared constants are only half-adopted

The stated goal is one way to say this, but after the diff v2 has three wordings for the same condition:

| Site | `issue` string |
| --- | --- |
| new `checkContactsEnabledApiV2` | `Contacts feature is not enabled for this organization` |
| `ee/contacts/api/v2/management/contacts/route.ts:31` (+ `bulk`) | `Contacts feature is not enabled for this environment` |
| `api/v2/management/surveys/[surveyId]/contact-links/**` | `Contacts are only enabled for Enterprise Edition, please upgrade.` |

Yours is the most accurate of the three (the check *is* org-scoped, so the "environment" one is wrong). Same story on v1: `CONTACTS_API_V1_NOT_ENABLED_MESSAGE` is used at 2 sites while 4 others still hardcode the literal (`api/v1/management/contacts/route.ts:18`, `contacts/[contactId]/route.ts:41,86`, `contact-attributes/route.ts:18`). Pointing those six at the new module is a mechanical follow-up and would make the "close the class" framing actually hold. Fine as a separate PR if you'd rather keep this diff tight.

### Verified

Things I checked rather than took on trust:

- **Coverage is complete.** All 8 actions in `ee/contacts/{actions,attributes,[contactId]}` call `ensureContactsEnabled`; all 8 in `segments/actions.ts` call `checkAdvancedTargetingPermission`. Grepping every caller of `deleteContact` / `createContactsFromCSV` / `updateContactAttributes` / `{create,update,delete}ContactAttributeKey` outside `lib/` returns 13 sites, and all 13 now sit behind a gate. Every other contact route (v1 client identify, v1 contacts, v1 contact-attributes, v2 ee contacts + bulk, v2 contact-links) already gated before this PR.
- **The dead-catch removal is safe.** Repo-wide, `"Contacts are only enabled for Enterprise Edition, please upgrade."` appears only as a `forbiddenResponse` / `details.issue` payload — nothing constructs it as an `Error`, so the catch could never fire. Grep for `throw new Error("Contacts…` is empty.
- **`authentication.organizationId` is sound** on `NonNullable<TApiKeyAuthentication>` (`packages/types/auth.ts:20-26`), and the three v2 `[id]` handlers already destructured `authentication`.
- **v1 `[id]` ordering** puts the entitlement check ahead of both the not-found and the `hasPermission` check, so an unentitled caller gets 403 rather than a 404 that would confirm key existence. Right way round.
- **`getDistinctAttributeValuesAction`'s** only caller is `segments/components/attribute-value-input.tsx:40`, so the "Advanced targeting is not allowed for this organization" wording it inherits is accurate there — no misleading message leaks into the contacts UI.
- **Tests pass**: `npx vitest run modules/ee/contacts/lib/contacts-entitlement.test.ts` → 4/4.
- **The e2e gap is real, not an excuse**: `e2e.yml:131-133` writes `ENTERPRISE_LICENSE_KEY` into `.env` and `:315-317` hard-fails if it's empty, so entitlement-off simply cannot be exercised there.
- **No doc drift**: the v2 OpenAPI files document only success responses, and the already-gated sibling (`ee/contacts/api/v2/management/contacts/lib/openapi.ts`) documents neither a 403 nor its EE requirement — so the newly-gated endpoints are no less accurate than their gated peers.

I'll also flag that the review harness's `pr-review` skill wasn't available in this container, so this is a hand-rolled pass over the same ground rather than the packaged workflow.

---

apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
```diff
@@ -14,15 +16,22 @@ import { ZContactAttributeKeyUpdateInput } from "./types/contact-attribute-keys"
 
 async function fetchAndAuthorizeContactAttributeKey(
   attributeKeyId: string,
-  workspacePermissions: NonNullable<TApiKeyAuthentication>["workspacePermissions"],
+  authentication: NonNullable<TApiKeyAuthentication>,
   requiredPermission: "GET" | "PUT" | "DELETE"
 ) {
+  // Entitlement first, matching the plural route: without the contacts feature the caller may
+  // not interact with attribute keys at all, regardless of workspace permissions.
+  const isContactsEnabled = await getIsContactsEnabled(authentication.organizationId);
```
Comment (kiln-agent) [inline apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts:24]: The `## Breaking changes` box is unticked ("None"), but the diff changes the emitted value of endpoints that public API consumers already call: `GET/PUT/DELETE /api/v1/management/contact-attribute-keys/{id}` and `GET/POST /api/v2/management/contact-attribute-keys` (+ `/{id}`) go from 2xx-with-data to 403 for any organization without the `contacts` entitlement. The PR's own **Risks** section states this ("v2 attribute-key endpoints now return 403 for unlicensed orgs where they previously returned data"), which is exactly the template's definition of breaking ("changes an API/SDK shape or emitted value … needs manual migration action").

Concrete scenario: a self-hoster on Community Edition (no enterprise license, so `getIsContactsEnabled` → false) has a script that syncs attribute keys through the v1 by-id endpoints. They upgrade to the release containing this commit and every call starts returning 403. Because `pr-label-sync.yml` reads only the tick, no `breaking-change` label is applied, so the release notes and the self-hoster migration guide carry no entry and the operator gets no warning — the one signal that would have told them is suppressed.

Suggest ticking the box and replacing "None" with a row per affected endpoint group (Before: 200 + payload; After: 403; Who: self-hosters / API consumers without the contacts entitlement; Action: obtain the entitlement or stop calling these endpoints). The functional gating itself looks correct and complete — this is only about the label/migration-guide signal.

---

kiln-agent (commented):
Reviewed the full diff for correctness. The entitlement gating is complete and consistent: all 9 server actions in `modules/ee/contacts` (4 in `actions.ts`, 3 in `attributes/actions.ts`, 1 in `[contactId]/actions.ts`, 8/8 in `segments/actions.ts` via `checkAdvancedTargetingPermission`) and all 8 route handlers are gated, `organizationId` is in scope at every new call site, and `authentication.organizationId` exists on `TAuthenticationApiKey` so the v2 handlers type-check. Confirmed the removed `catch` in the v1 by-id GET was genuinely dead — nothing in the repo `throw`s that message, it is only ever `return`ed as a `forbiddenResponse`. New unit test file passes locally (4/4).

One non-code finding on the breaking-change checkbox, inline.

---

apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
```diff
@@ -13,6 +13,7 @@ import {
   ZContactAttributeKeyUpdateSchema,
 } from "@/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/types/contact-attribute-keys";
 import { ApiErrorResponseV2 } from "@/modules/api/v2/types/api-error";
+import { checkContactsEnabledApiV2 } from "@/modules/ee/contacts/lib/contacts-entitlement";
```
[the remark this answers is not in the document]
Comment (pandeymangg) [reply apps/web/modules/api/v2/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts]: Moved in 1cfd50cdc — `checkContactsEnabledApiV2` now lives in `modules/ee/license-check/lib/contacts-api-guard.ts` and both v2 routes import it through the license-check boundary. Behavior and tests unchanged (tests moved alongside). Note: `modules/api/v2` already imports `modules/ee/contacts` libs in several places on main (contact-survey-link, attribute-key-policy, segments filter) — leaving those for a separate cleanup.

apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts
Comment (pandeymangg) [reply apps/web/modules/ee/contacts/api/v1/management/contact-attribute-keys/[contactAttributeKeyId]/route.ts:24]: Agreed — ticked, with a Before/After table for the v1 by-id and v2 attribute-key endpoints. Server actions left off the table since they aren't a public API surface.

itsjavi (approved):
LGTM

PR #8956, opened 2026-08-24
chore: enable the eslint/typescript-eslint recommended baseline as warnings in apps/web (ENG-2264)

pandeymangg:
Ref ENG-2264

## What & why

- `packages/config-eslint/next.mjs` deliberately omitted the eslint/typescript-eslint recommended baselines, so rules like `@typescript-eslint/no-explicit-any` never ran in `apps/web` — our largest codebase — while being errors in `packages/*`.
- This PR adds `js.configs.recommended` + `tseslint.configs.recommended` to the `next` tier, with every rule downgraded to `warn` (options preserved). Warnings don't fail CI (`eslint .` has no `--max-warnings`), so no PR is blocked; rules get promoted to `error` one by one as their backlog clears.
- Also adds the repo's `_`-prefix `no-unused-vars` convention (shared `unusedVarsConvention`, at warn) so intentionally-unused bindings aren't flagged, and updates the `AGENTS.md` "avoid `any`" line to state how it's now enforced.
- Deliberately **not** `recommendedTypeChecked` (used by the library/react tiers): apps/web measured 9,204 violations and OOMs at Node's default heap under it — that stays a separate decision on the ticket.

**Measured backlog** (recorded on the ticket with the ratchet plan): 1,892 warnings across 386 of 2,656 files. 1,552 are in test/mock files, only 300 in product code. `no-explicit-any` is 1,583 of the total (1,476 test/mock, 107 source).

## Breaking changes

None — lint-severity change only (new warnings, zero new errors); no runtime code touched.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Baseline rules now report in `apps/web` | manual | `pnpm exec eslint . --format json` in `apps/web`: 1,892 warnings from the expected rule set (`no-explicit-any`, `prefer-const`, `no-unused-vars`, …) |
| Nothing newly fails — merge behaviour unchanged | manual | Same run exits 0: zero error-level messages, warnings only |
| Other packages unaffected | manual | `next.mjs` has exactly one consumer (`apps/web/eslint.config.mjs`, verified by grep); library/react tiers untouched |

**Open gaps**

- Root `pnpm lint` (turbo, all workspaces) not run locally — only `apps/web` was linted directly; CI covers the rest. The change is scoped to the `next` tier, which no other workspace consumes.
- `_`-prefix handling relies on reusing the shared `unusedVarsConvention` options; I didn't construct a dedicated fixture to confirm an `_arg` goes unflagged in apps/web.

**Risks**

- Editor/PR-annotation noise from the new warnings until the ratchet clears them; if a rule proves too noisy, its severity can be adjusted in one place in `next.mjs`.
- `eslint-disable` comments in apps/web referencing typescript-eslint rules were previously inert; the plugin now being loaded can change whether they count as used or unused directives (unused ones surface as warnings via `reportUnusedDisableDirectives`, already `warn`).

---

> [!NOTE]
> **AI model used** — `claude-fable-5` (Claude Code, version unknown), reasoning effort `unknown`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -239,7 +239,10 @@ Heuristic:
 
 - Keep code DRY and small; remove dead code and unused imports.
 - Follow React hooks rules, keep effects focused, and avoid unnecessary `useMemo`/`useCallback`.
-- Prefer type inference, avoid `any`, and use shared types from `@formbricks/types`.
+- Prefer type inference, avoid `any`, and use shared types from `@formbricks/types`. This is enforced:
+  `@typescript-eslint/no-explicit-any` is an error in `packages/*` and a warning in `apps/web`, where the
+  typescript-eslint baseline is being ratcheted to error rule by rule (ENG-2264). Never add new `any`s —
+  a warning today becomes an error once its rule's backlog is cleared.
 - Keep components focused, avoid deep nesting, and ensure basic accessibility.
 
 ## Commit & Pull Request Guidelines
--- a/packages/config-eslint/next.mjs
+++ b/packages/config-eslint/next.mjs
@@ -1,16 +1,50 @@
+import js from "@eslint/js";
 import nextConfig from "eslint-config-next";
-import { base, commonIgnores, typescriptParsing } from "./base.mjs";
+import tseslint from "typescript-eslint";
+import { base, commonIgnores, typescriptParsing, unusedVarsConvention } from "./base.mjs";
 
 /*
  * Flat config for the Next.js app — the successor of `legacy-next.js`:
  * eslint-config-next + turbo + prettier + the vitest convention, with the same rule
- * overrides the legacy config carried. Deliberately does not add the
- * eslint/typescript-eslint recommended baselines yet to keep rule parity with the
- * previous setup — tightening is a separate step (ENG-2264).
+ * overrides the legacy config carried.
+ *
+ * The eslint/typescript-eslint recommended baselines are active but downgraded to
+ * warnings (ENG-2264): apps/web was never linted against them, so the backlog is
+ * large and enabling them as errors in one step would block every PR. Rules are
+ * promoted to "error" individually as their violation count reaches zero — the
+ * per-rule counts and ratchet plan live on the ticket. `recommendedTypeChecked`
+ * (used by the library/react tiers) is a separate decision, also tracked there.
  */
+
+// The recommended baselines ship rules at "error"; keep their options but lower the
+// severity so the backlog surfaces without failing CI. "off" entries (e.g. the core
+// rules that eslint-recommended disables for TS files) pass through untouched.
+const isErrorSeverity = (severity) => severity === "error" || severity === 2;
+
+const downgradeToWarn = ({ rules, ...config }) => ({
+  ...config,
+  ...(rules && {
+    rules: Object.fromEntries(
+      Object.entries(rules).map(([id, entry]) => {
+        if (Array.isArray(entry)) {
+          const [severity, ...options] = entry;
+          return [id, isErrorSeverity(severity) ? ["warn", ...options] : entry];
+        }
+        return [id, isErrorSeverity(entry) ? "warn" : entry];
+      })
+    ),
+  }),
+});
+
 export const next = [
   commonIgnores,
   typescriptParsing,
+  downgradeToWarn(js.configs.recommended),
+  ...tseslint.configs.recommended.map(downgradeToWarn),
+  // Same underscore convention as the other tiers, at warning severity like the rest
+  // of the baseline; without the options, default no-unused-vars would flag the
+  // repo-sanctioned `_`-prefixed intentionally-unused bindings.
+  downgradeToWarn(unusedVarsConvention),
   ...nextConfig,
   {
     rules: {


kiln-agent (commented):
## Review — enable eslint/tseslint baseline as warnings in `apps/web`

Read the whole diff plus `base.mjs`, the sole consumer (`apps/web/eslint.config.mjs`), and `eslint-config-next@16.2.10`'s resolved config. This is a clean, well-scoped, and — unusually for a lint-config change — thoroughly self-tested PR. I could not find a correctness defect. **No blocking findings.**

### Verified against the resolved config (not just the source)

I ran `eslint --print-config` on a real `apps/web` file and on a `packages/*` file to confirm the composition actually resolves as intended, rather than trusting flat-config ordering by eye:

| Rule | `apps/web` | Expected | ✓ |
| --- | --- | --- | --- |
| `@typescript-eslint/no-explicit-any` | `warn` | warn | ✓ |
| `@typescript-eslint/no-unused-vars` | `warn` **with** `^_` options | underscore convention survives | ✓ |
| core `no-unused-vars` / `no-undef` | `off` | tseslint-recommended disables for TS | ✓ |
| `no-restricted-syntax` (env guard) | `error` | must **not** be downgraded | ✓ |
| `@typescript-eslint/no-explicit-any` (`packages/ai`) | `error` | AGENTS.md prose accurate | ✓ |

So the two claims the diff makes in prose — "warning in `apps/web`, error in `packages/*`" — both hold, and the underscore convention is not clobbered by the later `...nextConfig` / `...base` spreads.

### Concerns I considered and refuted

- **Double-registration of the `@typescript-eslint` plugin** (once via `tseslint.configs.recommended`, once via `eslint-config-next`'s `next/typescript` block). Flat config throws on redefining a plugin key with a *different* instance — but under the hoisted linker + pinned catalog both resolve to the same module object, so the config composes without error (`--print-config` succeeded, and the author's full `eslint .` exits 0).
- **`downgradeToWarn` mutating shared module state** — it doesn't: it rebuilds `rules` via `Object.fromEntries` into a fresh object and only shares `plugins`/`languageOptions` by reference (never written), so the library/react tiers that also import `tseslint.configs.recommended` are unaffected.
- **`no-restricted-syntax` env guard getting downgraded** — safe, because it's set by the consumer *after* `...next`, outside the reach of `downgradeToWarn`; confirmed `error` above.
- **`downgradeToWarn` severity handling** — covers strings (`"error"`/`"warn"`/`"off"`), numerics (`2`/`1`/`0`), and array forms with options; `off`/`warn` entries pass through untouched. Complete.
- **Warnings blocking CI** — `apps/web` lint is `eslint .` with no `--max-warnings`; the `--max-warnings 0` in the repo is `apps/storybook`, which consumes a different tier. Claim holds.

### Minor, non-blocking

- `js.configs.recommended` is applied with no `files` filter, so core rules run on plain `.js`/`.mjs` in `apps/web` too (e.g. `no-undef` stays `warn` there since the TS turn-off is `.ts`/`.tsx`-scoped). Only annotation noise, never a failure — and already covered by your "Risks" note.

Nice touches: keeping rule *options* while lowering severity, the pass-through of `off` entries, and the "How tested" table measuring the actual backlog. The `AGENTS.md` update lands the enforcement detail exactly where a contributor will look.

---

kiln-agent:
## Code review — no blocking issues

I reviewed this PR (`34b1c0e`). It touches two files: `AGENTS.md` (doc note) and `packages/config-eslint/next.mjs` (the actual logic).

I ran all finder angles and then **empirically verified** the candidates by loading `apps/web/eslint.config.mjs` through the real `ESLint` API and linting probe files across `.ts`, `.js`, and `.mjs`. Everything checked out:

- **`downgradeToWarn`** correctly maps error→warn (`no-explicit-any`, `no-cond-assign`, `ban-ts-comment`, `no-empty`, …) while `"off"` entries and non-error severities pass through untouched.
- **Underscore convention preserved** — `@typescript-eslint/no-unused-vars` fires at warn *with* the `^_` options because `downgradeToWarn(unusedVarsConvention)` is placed after the recommended map, so it wins over the baseline's option-less version.
- **No duplicate `no-unused-vars`** — tseslint recommended turns core `no-unused-vars` off, so JS files don't double-warn.
- **`no-undef` scoping** — off for TS, warn for `.mjs`; node globals resolve via eslint-config-next's globals, so no spurious `no-undef`.
- **No parity regression** — baselines sit before `...nextConfig`/`...base`, so no pre-existing error rule is weakened; the change is strictly additive at warn severity.
- **AGENTS.md** note matches actual behavior (error in `packages/*`, warn in `apps/web`).

**One theoretical, non-blocking note:** `js.configs.recommended` applies to all files, but eslint-config-next's globals glob excludes `.cjs`, so a future `.cjs` file in `apps/web` using `require`/`module` would draw `no-undef` *warnings*. There are currently zero `.cjs` files in `apps/web`, and these would be non-blocking warnings — not a defect today.

No correctness, reuse, simplification, efficiency, or convention findings survived verification. LGTM. ✅

---

mattinannt (approved):

PR #8957, opened 2026-08-24
fix(segments): bound the survey-interaction filter IN queries (ENG-2305)

pandeymangg:
Fixes ENG-2305

## What & why

Follow-up to #8772 (ENG-2004): that PR bounded `getSurveyWorkspaceIdMap`, but two sibling helpers reachable from the same actions kept the unbounded Prisma `IN (...)` shape. `ZSegmentSurveyInteractionFilterValue.surveyIds` caps a single filter at 100 ids, but `ZSegmentFilters` is a recursive array with no length or depth cap — so N filters × 100 ids made the total unbounded: a cheap authenticated DoS lever through `createSegmentAction`, `updateSegmentAction`, and the v3 targeting route.

Same two layers as #8772, hardened further after review:

- **Pre-parse bounds gate** (`packages/types/segment.ts`): an iterative (explicit-stack, non-recursive) walker measures the **raw** tree — total nodes, nesting depth, total survey-interaction ids — and rejects before the recursive Zod parse runs, wired as `z.unknown().superRefine(bounds).pipe(recursive schema)`. Pre-parse is load-bearing, not style: a deep enough tree overflows the call stack *inside* the recursive parse with a `RangeError` that even `safeParse` doesn't catch (escapes as a 500), so a post-parse `.refine` can never guard it. Three exported caps: `MAX_SEGMENT_FILTERS_PER_TREE = 1000` (total nodes; nesting can't dodge it since every level costs nodes), `MAX_SEGMENT_FILTER_DEPTH = 50` (~20x below the measured stack limit; the editor produces single-digit depth), `MAX_SEGMENT_SURVEY_INTERACTION_IDS_PER_TREE = 1000` (bounds write-time lookups to ≤5 batches per request instead of ~500).
- **Draft-save gate** (`apps/web/lib/survey/service.ts`): the survey-editor draft save (`updateSurveyDraftAction` → `updateSurveyInternal` with `skipValidation`) used to persist the segment without ever parsing its schemas — an open bypass around every cap. Two unconditional checks now run before the segment persist: the filter-tree bounds check, and `ZSegmentSurveyIds` (the ENG-2004 `ZId` + 500-link cap, now exported instead of duplicated) on `segment.surveys` before the ids drive the batched workspace lookup. Full semantic validation stays skippable, so drafts with half-built filter values save exactly as before.
- **Defensive batching in the helpers**: `assertSurveyInteractionSurveyIds` (`helper.ts`) and `getExistingWorkspaceSurveyIds` (`segments.ts`) look ids up in sequential 200-size batches via the now-shared `SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE` (moved to `segments/lib/utils.ts`, single definition). Semantics unchanged: the assert helper throws the same `InvalidInputError` on the first missing/foreign id (and stops querying), `getExistingWorkspaceSurveyIds` still returns only the found subset. With the id cap this layer is defense in depth for within-cap totals and any future unparsed caller.

On the cap values: hand-built segment trees are tens of nodes at single-digit depth, so 1000/50/1000 are orders of magnitude above real usage. `ZSegmentFilters` also parses trees read back from the DB (clone, publish validation, editor), so the caps must never brick a legacy row — hence generous, and with the draft bypass closed there is no writer left that can persist an over-bounds tree.

Also removed `ZBaseFilter`/`ZBaseFilters`: a dead, unrefined duplicate of the same tree schema (no consumers anywhere, verified by grep and typecheck). Keeping an uncapped copy around would let a future consumer silently bypass this fix; the widely-used `TBaseFilter`/`TBaseFilters` TS types are defined independently and stay.

## Breaking changes

None — the new validation only rejects segment filter trees beyond 1000 total nodes, 50 nesting levels, or 1000 survey-interaction ids, all far beyond anything the product can produce; no API shape, endpoint, env var, default or payload change.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Filter tree over the node cap is rejected at the Zod boundary, including the nesting-bypass shape (small arrays at every level) | unit (red on main) | `segment-schema.test.ts`: accepts exactly 1000 nodes, rejects 1001 with the cap message, rejects the nested-bypass tree and over-cap input through `ZSegmentUpdateInput`. Fails on main, where no cap existed — rerun: `pnpm test --filter=@formbricks/web` (file: `apps/web/modules/ee/contacts/segments/lib/segment-schema.test.ts`) |
| A deep left-spine tree **within** the node cap fails with a clean depth `ZodError` — not a `RangeError` escaping `safeParse` as a 500 | unit (red on main) | `segment-schema.test.ts`: 1000-deep spine → depth message; depth exactly 50 accepted, 51 rejected. On main the same tree throws `RangeError: Maximum call stack size exceeded` out of `safeParse` |
| Survey-interaction ids capped tree-wide, not just per filter | unit (red on main) | `segment-schema.test.ts`: 10×100 ids accepted, 11×91 = 1001 rejected with the id-cap message |
| Draft save can no longer persist an over-bounds tree (`skipValidation` bypass closed), while a within-bounds draft with semantically invalid filters still saves | unit (red on main) + unit (guard) | `apps/web/lib/survey/service.test.ts`: over-bounds tree with `skipValidation: true` → `InvalidInputError`, no `segment.update`; junk-but-small filters still persist (locks draft UX) |
| Draft save rejects an over-cap or non-id `segment.surveys` list **before any survey lookup runs** (500-cap + `ZId` now enforced on the draft path too), while a within-cap valid list still reaches the ownership check | unit (red on main) + unit (guard) | `apps/web/lib/survey/service.test.ts`: 501 valid ids and a junk string each → `InvalidInputError` with zero `survey.findMany` / `segment.update` calls (both red on the previous head); within-cap list proceeds unchanged |
| `assertSurveyInteractionSurveyIds` batches: 450 ids → queries of [200, 200, 50] with `workspaceId` on every query; first missing id throws the same error and later batches are never queried; a parsed-tree-max payload (1000 ids) resolves in exactly 5 batches | unit (mutation) | `helper.test.ts` — asserts exact batch contents and call counts |
| `getExistingWorkspaceSurveyIds` dedupes + batches, still returns only the found subset | unit (mutation) | `segments.test.ts` |
| Tenancy write-time guard unchanged — foreign/unknown survey ids still rejected (the ENG-1749 / ENG-1920 behavior) | unit (guard) | zero pre-existing tests modified; foreign-id rejection re-asserted on the batched path |

Full root gate after the review-fix round: typecheck 27/27, lint 21/21 (0 errors), test 26/26 tasks (`@formbricks/web`: 640 files / 8508 tests).

**Open gaps**

- The caps weren't checked against real production segment sizes (no data access from here). They're orders of magnitude above what the editor can build, and the draft-save path — until this PR the only writer that could persist an unvalidated tree — is now gated. But a segment created programmatically before this PR beyond 1000 nodes / depth 50 would become unsavable on paths that parse `ZSegmentFilters` (clone, publish validation, editor). Worth a one-off prod query if we want certainty.
- No manual UI pass — unit-only change through the same schemas the editor already exercises.

**Risks**

- Read-time cost within the cap is unchanged: one saved segment with ~1000 attribute filters still drives ~1000 sequential evaluation queries per request on the SDK-facing user endpoint. Pre-existing (there was no cap at all before), out of this PR's scope — candidate follow-up ticket.
- Draft saves now reject over-bounds filter trees and over-cap/malformed `segment.surveys` lists with `InvalidInputError` where they previously persisted (or looked up) anything; no known client builds such drafts.

---

> [!NOTE]

### DIFF
--- a/apps/web/lib/survey/service.ts
+++ b/apps/web/lib/survey/service.ts
@@ -10,7 +10,11 @@ import {
   OperationNotAllowedError,
   ResourceNotFoundError,
 } from "@formbricks/types/errors";
-import { TBaseFilters, ZSegmentFilters } from "@formbricks/types/segment";
+import {
+  TBaseFilters,
+  ZSegmentFilters,
+  getSegmentFilterTreeBoundsViolation,
+} from "@formbricks/types/segment";
 import { TSurveyBlock } from "@formbricks/types/surveys/blocks";
 import { TSurvey, TSurveyCreateInput, ZSurvey, ZSurveyCreateInput } from "@formbricks/types/surveys/types";
 import { scheduleFeedbackSourceReconciliation } from "@/lib/feedback-source/mapping-reconciliation";
@@ -429,6 +433,15 @@ export const updateSurveyInternal = async (
     // if the survey body has type other than "app" but has a private segment, we delete that segment, and if it has a public segment, we disconnect from to the survey
     if (segment) {
       if (type === "app") {
+        // ENG-2305: tree bounds are enforced UNCONDITIONALLY — the draft save (skipValidation)
+        // deliberately skips full semantic validation so half-built filters can be saved, but an
+        // over-bounds tree persisted through it would break every consumer that parses the row
+        // back (publish validation, clone, evaluation).
+        const boundsViolation = getSegmentFilterTreeBoundsViolation(segment.filters);
+        if (boundsViolation) {
+          throw new InvalidInputError(boundsViolation);
+        }
+
         // parse the segment filters:
         const parsedFilters = ZSegmentFilters.safeParse(segment.filters);
         if (!skipValidation && !parsedFilters.success) {
--- a/apps/web/lib/survey/service.ts
+++ b/apps/web/lib/survey/service.ts
@@ -11,8 +11,10 @@ import {
   ResourceNotFoundError,
 } from "@formbricks/types/errors";
 import {
+  MAX_SEGMENT_SURVEYS,
   TBaseFilters,
   ZSegmentFilters,
+  ZSegmentSurveyIds,
   getSegmentFilterTreeBoundsViolation,
 } from "@formbricks/types/segment";
 import { TSurveyBlock } from "@formbricks/types/surveys/blocks";
@@ -446,6 +448,17 @@ export const updateSurveyInternal = async (
         const parsedFilters = ZSegmentFilters.safeParse(segment.filters);
         if (!skipValidation && !parsedFilters.success) {
           throw new InvalidInputError("Invalid user segment filters");
+        }
+
+        // ENG-2305 sibling of the filter-tree gate above: on the draft path (skipValidation)
+        // segment.surveys reaches this point unvalidated — ZSurveyDraft.segment is an untyped
+        // record, so neither the ZId format rule nor the MAX_SEGMENT_SURVEYS cap has applied. Both
+        // must hold unconditionally BEFORE the ids drive the batched workspace lookup below; the
+        // validated (non-draft) path re-checks the same schema it already passed, a no-op.
+        if (segment.surveys && !ZSegmentSurveyIds.safeParse(segment.surveys).success) {
+          throw new InvalidInputError(
+            `Invalid segment surveys: at most ${MAX_SEGMENT_SURVEYS} valid survey ids are allowed`
+          );
         }
 
         // ENG-1749/ENG-1920: the connected survey ids are client-supplied; ensure each belongs to
--- a/apps/web/modules/ee/contacts/segments/lib/helper.test.ts
+++ b/apps/web/modules/ee/contacts/segments/lib/helper.test.ts
@@ -1,6 +1,10 @@
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { InvalidInputError } from "@formbricks/types/errors";
-import { TBaseFilters, TSegmentWithSurveyRefs } from "@formbricks/types/segment";
+import {
+  MAX_SEGMENT_FILTERS_PER_TREE,
+  TBaseFilters,
+  TSegmentWithSurveyRefs,
+} from "@formbricks/types/segment";
 import {
   assertSurveyInteractionSurveyIds,
   checkForRecursiveSegmentFilter,
@@ -335,5 +339,82 @@ describe("assertSurveyInteractionSurveyIds", () => {
     await expect(
       assertSurveyInteractionSurveyIds(buildSpecificFilter(["s1", "s2"]), "workspace-1")
     ).rejects.toThrow(new InvalidInputError("Survey not found in workspace: s2"));
+  });
+
+  test("deduplicates ids collected across filters before querying", async () => {
+    const filters = [
+      ...buildSpecificFilter(["s1", "s2"]),
+      ...buildSpecificFilter(["s2", "s1"]),
+    ] as unknown as TBaseFilters;
+    mockSurveyFindMany.mockResolvedValue([{ id: "s1" }, { id: "s2" }]);
+
+    await expect(assertSurveyInteractionSurveyIds(filters, "workspace-1")).resolves.toBeUndefined();
+
+    expect(mockSurveyFindMany).toHaveBeenCalledTimes(1);
+    expect(mockSurveyFindMany).toHaveBeenCalledWith({
+      where: { id: { in: ["s1", "s2"] }, workspaceId: "workspace-1" },
+      select: { id: true },
+    });
+  });
+
+  test("splits an oversized id list into bounded sequential batches, all workspace-scoped", async () => {
+    const surveyIds = Array.from({ length: 450 }, (_, index) => `s_${index}`);
+    mockSurveyFindMany.mockImplementation(async ({ where }: any) =>
+      where.id.in.map((id: string) => ({ id }))
+    );
+
+    await expect(
+      assertSurveyInteractionSurveyIds(buildSpecificFilter(surveyIds), "workspace-1")
+    ).resolves.toBeUndefined();
+
+    // 450 ids at a batch size of 200 -> 200 / 200 / 50, each query scoped to the workspace.
+    const batchSizes = mockSurveyFindMany.mock.calls.map(([args]: any) => args.where.id.in.length);
+    expect(batchSizes).toEqual([200, 200, 50]);
+    for (const [args] of mockSurveyFindMany.mock.calls) {
+      expect((args as any).where.workspaceId).toBe("workspace-1");
+    }
+  });
+
+  test("throws on the first missing id in collection order and stops querying further batches", async () => {
+    const surveyIds = Array.from({ length: 450 }, (_, index) => `s_${index}`);
+    // Two ids in the second batch are foreign/unknown; the earlier one must win, and the third
+    // batch must never be queried.
+    mockSurveyFindMany.mockImplementation(async ({ where }: any) =>
+      where.id.in.filter((id: string) => id !== "s_205" && id !== "s_210").map((id: string) => ({ id }))
+    );
+
+    await expect(
+      assertSurveyInteractionSurveyIds(buildSpecificFilter(surveyIds), "workspace-1")
+    ).rejects.toThrow(new InvalidInputError("Survey not found in workspace: s_205"));
+
+    expect(mockSurveyFindMany).toHaveBeenCalledTimes(2);
+  });
+
+  test("a pre-existing tree over the Zod tree cap still flows through without throwing on size", async () => {
+    // Trees persisted before MAX_SEGMENT_FILTERS_PER_TREE existed can exceed it; the write-time
+    // guard must still process their ids (batched), not reject them for size.
+    const filterCount = MAX_SEGMENT_FILTERS_PER_TREE + 100;
+    const filters = Array.from({ length: filterCount }, (_, index) => ({
+      id: `f_${index}`,
+      connector: index === 0 ? null : "and",
+      resource: {
+        id: `si_${index}`,
+        root: { type: "surveyInteraction" },
+        qualifier: { operator: "haveSeen" },
+        value: {
+          surveyScope: "specific",
+          surveyIds: [`s_${index}`],
+          within: { amount: 1, unit: "months" },
+        },
+      },
+    })) as unknown as TBaseFilters;
+    mockSurveyFindMany.mockImplementation(async ({ where }: any) =>
+      where.id.in.map((id: string) => ({ id }))
+    );
+
+    await expect(assertSurveyInteractionSurveyIds(filters, "workspace-1")).resolves.toBeUndefined();
+
+    const batchSizes = mockSurveyFindMany.mock.calls.map(([args]: any) => args.where.id.in.length);
+    expect(batchSizes).toEqual([200, 200, 200, 200, 200, 100]);
   });
 });
--- a/apps/web/modules/ee/contacts/segments/lib/helper.test.ts
+++ b/apps/web/modules/ee/contacts/segments/lib/helper.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import { InvalidInputError } from "@formbricks/types/errors";
 import {
-  MAX_SEGMENT_FILTERS_PER_TREE,
+  MAX_SEGMENT_SURVEY_INTERACTION_IDS_PER_TREE,
   TBaseFilters,
   TSegmentWithSurveyRefs,
 } from "@formbricks/types/segment";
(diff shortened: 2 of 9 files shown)


packages/types/segment.ts
```diff
@@ -396,6 +386,28 @@ const refineFilters = (filters: TBaseFilters): boolean => {
...
+/**
+ * Maximum number of filter nodes — leaf filters plus nested groups — across the WHOLE recursive
+ * filter tree. The tree schema has no per-level length cap, so a per-level `.max()` alone would be
+ * bypassable by nesting; only a total bound keeps N filters × the per-filter surveyIds cap (100)
+ * from becoming an unbounded id payload (ENG-2305, sibling of ENG-2004). The same schema also
+ * parses trees read back from the database (clone, publish validation, segment editor), so the cap
+ * is deliberately generous — orders of magnitude above anything the segment editor produces — to
+ * never brick a pre-existing segment on read.
+ */
+export const MAX_SEGMENT_FILTERS_PER_TREE = 1000;
```
kiln-agent, line 398: **The 1000-node cap sits above the stack-safe parse depth, so the depth lever stays open.**

The docstring says a total-node cap "implicitly bounds depth too". That only holds if the node cap is *below* the depth at which the recursive zod parse blows the JS stack — at 1000 it is not.

Reproduced on this branch: a left-spine tree (one nested group per level) of ~991 nodes — **inside** the new cap — makes `ZSegmentFilters.safeParse` throw:

```
RangeError: Maximum call stack size exceeded
 ❯ handleUnionResults node_modules/zod/v4/core/schemas.js:946:25
```

990 levels still parses; 992+ throws. `safeParse` does not turn a thrown `RangeError` into a `ZodError`, so it propagates out of `createSegmentAction`/`updateSegmentAction` (`actions.ts:87` / `:161`) and the v3 targeting schemas (`app/api/v3/surveys/schemas.ts:1118`) as an unhandled 500 rather than a 400. The payload is ~60KB, nowhere near the 2mb serverActions limit. And the threshold is *lower* in a real request handler than in the vitest worker I measured in, because the handler stack is already deeper before zod is entered.

So the exact shape the cap is meant to reject (deep nesting) still crashes the process path instead of being rejected cleanly. Suggest either an explicit depth cap well under the stack-safe depth (tens, not hundreds — the editor produces single-digit depth), or an iterative depth/size pre-check on the raw JSON *before* handing it to zod, since a post-parse refine can never run early enough to help here.

---

apps/web/modules/ee/contacts/segments/lib/helper.ts
```diff
@@ -62,29 +62,36 @@ export const collectSurveyIdsFromSegmentFilters = (filters: TBaseFilters): strin
...
 
-  if (surveyIds.length === 0) {
-    return;
-  }
-
-  const foundSurveys = await prisma.survey.findMany({
-    where: { id: { in: surveyIds }, workspaceId },
-    select: { id: true },
-  });
+  for (let i = 0; i < surveyIds.length; i += SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE) {
```
kiln-agent, line 83: **Within the cap, this turns one query into up to 500 *sequential* round-trips — the connection is held for all of them.**

The description flags "up to 100k ids → ~500 sequential indexed PK lookups" as accepted, but it's worth being explicit about what changed: before this PR `assertSurveyInteractionSurveyIds` made exactly **one** query for that payload. After it, the same request awaits 500 `prisma.survey.findMany` calls back to back, each holding a pooled connection, so per-request connection occupancy went up by roughly 500x for precisely the abuse case the PR targets. A handful of concurrent `createSegmentAction` calls with a max-legal tree can starve the pool — which is a cheaper DoS than the one big `IN (...)` it replaced.

I confirmed the payload is accepted: 1000 `surveyInteraction` filters × 100 distinct ids parses through `ZSegmentFilters` in ~150ms, `success: true`.

`getExistingWorkspaceSurveyIds` (`segments.ts:198`) has the same shape and is reached from the v3 targeting route, which is a route handler — the 2mb `serverActions.bodySizeLimit` doesn't apply there, so the full 100k is reachable.

The batching itself is right; what's missing is a bound on the *total* id count. A cheap fix is a second refine summing `surveyIds` across the tree (the description already calls this out as a follow-up) — capping total ids at, say, a few thousand keeps the batch count in single digits and makes the batching layer purely defensive rather than load-bearing.

---

packages/types/segment.ts
```diff
@@ -408,6 +420,9 @@ export const ZSegmentFilters: z.ZodType<TBaseFilters> = z
   )
   .refine(refineFilters, {
     error: "Invalid filters applied",
+  })
+  .refine((filters) => countSegmentFilterNodes(filters) <= MAX_SEGMENT_FILTERS_PER_TREE, {
```
kiln-agent: **The cap lands on read/round-trip paths too, and the diff itself assumes over-cap rows can exist.**

`helper.test.ts` adds *"a pre-existing tree over the Zod tree cap still flows through"* with a 1100-node tree — i.e. this change explicitly models over-cap rows as a real possibility. But `ZSegmentFilters` is also the round-trip schema via `ZSegment` → `ZSurvey`, with no read-path exemption, so if such a row exists it becomes **unsavable**:

- `GET`-then-`PUT` on `/api/v1/management/surveys/{surveyId}` fails `ZSurveyUpdateInput` (`segment: ZSegment` → `filters: ZSegmentFilters`) with the new "Too many filters" message.
- Saving from the segment editor fails `ZSegmentUpdateInput`, and `updateSurveyInternal` throws `InvalidInputError("Invalid user segment filters")` (`apps/web/lib/survey/service.ts:435`).

There's no UI affordance to shrink a tree that large either, so the row is stuck. The Open gaps section acknowledges this, so the ask is just to close it one of two ways before merge: enforce the cap only on client-submitted input (leave the read/round-trip parse uncapped), or verify against production data that no segment exceeds 1000 nodes. Right now the helper test asserts one behaviour for legacy trees and the schema asserts the opposite.

---

kiln-agent (commented):
Reviewed the diff for correctness. The batching change itself is faithful — dedup order is preserved, first-missing-id semantics are unchanged, and the empty-input short-circuit still holds. Three findings, all about the cap layer rather than the batching layer. The first one I reproduced against this branch and I think it blocks: the chosen cap value doesn't actually cover the nesting case it's meant to.

---

packages/types/segment.ts
kiln-agent: **The cap is bypassable — "all entry points parse through this schema" doesn't hold.**

The survey-editor draft save path never parses `ZSegmentFilters`:

- `updateSurveyDraftAction` (`apps/web/modules/survey/editor/actions.ts:192`) validates with `ZSurveyDraft`, where the segment is `segment: z.record(z.string(), z.unknown())` (`apps/web/modules/survey/editor/types/survey.ts:21`) — entirely unvalidated.
- It calls `updateSurveyDraft` → `updateSurveyInternal(survey, /* skipValidation */ true)`, whose filter gate is `if (!skipValidation && !parsedFilters.success) throw` (`apps/web/lib/survey/service.ts:432-436`). With `skipValidation` true the `safeParse` result is discarded and `filters: segment.filters` goes straight into `prisma.segment.update` (`service.ts:456-473`).

Concrete scenario: a user with `readWrite` on the workspace calls `updateSurveyDraftAction({ id, status: "draft", type: "app", name: "x", segment: { id: <their survey's segment id>, filters: <200k-node tree> } })`. The only bound is the 2mb server-action body limit — the tree is persisted. So the DoS lever this PR sets out to close stays open on that path, and the persisted row then breaks every path that *does* parse:

- publish validation rejects the survey (`modules/survey/editor/lib/validation.ts:280`)
- `cloneSegment` throws `ValidationError` (`segments.ts:312`)
- `evaluateSegmentFilter` returns `false` (`segments.ts:608`) — a nested reference to it silently evaluates to "not in segment", changing targeting with no error

Either gate the draft path (parse `segment.filters` even when `skipValidation` is set — the segment filters aren't the "incomplete draft" data the flag exists for), or drop the "all entry points" claim from the PR body so the remaining hole is visible.

---

packages/types/segment.ts
kiln-agent, line 398: **1000 bounds the write payload, but not the read path the tree actually drives.**

The cap value is argued purely against the id payload ("worst case within the cap is 1000 x 100 ids, which layer 2 chunks into <=200-id queries"). But the same accepted tree is compiled into the runtime evaluation query:

- `segmentFilterToPrismaQuery` -> `processFilters` walks the tree with a sequential `await` **per node** (`apps/web/modules/ee/contacts/segments/lib/filter/prisma-query.ts:498-506`).
- Each numeric attribute filter fires its own `$queryRawUnsafe` — an unindexed join over `ContactAttribute`/`ContactAttributeKey` with a regex predicate and a `::double precision` cast (`prisma-query.ts:180-207`).
- That builder is called from the SDK-facing client endpoint (`apps/web/modules/ee/contacts/api/v1/client/[workspaceId]/user/lib/segments.ts`).

Scenario: a tenant admin saves **one** segment with 1000 numeric-attribute filters — explicitly allowed by this cap. From then on every end-user `/api/v1/client/{workspaceId}/user` request runs ~1000 sequential unindexed queries plus a where clause holding ~1000 relation subqueries. A one-time authenticated write becomes permanent per-request amplification on a public endpoint.

Same budget on the v3 path: `assertV3SurveyTargetingFilterReferences` pushes one `invalidParams` entry per unknown id (`apps/web/app/api/v3/surveys/targeting.ts:300-307`), so 1000 x 100 unknown ids builds a multi-MB 422 body from one modest request.

None of this is a regression (there was no cap before), but it does mean the number chosen doesn't bound the expensive path. A cap in the low hundreds, or the total-ids bound you mention as a follow-up added to this same refine, would.

---

apps/web/modules/ee/contacts/segments/lib/helper.ts
```diff
@@ -62,29 +62,36 @@ export const collectSurveyIdsFromSegmentFilters = (filters: TBaseFilters): strin
   return surveyIds;
 };
 
+// Upper bound on ids per `IN (...)` lookup below, mirroring SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE in
+// segments.ts (ENG-2004/ENG-2305). MAX_SEGMENT_FILTERS_PER_TREE caps what a client can submit, but
+// this guard also runs over trees persisted before that cap existed, so the query itself is bounded
+// instead of trusting the collected array's length.
+const SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE = 200;
```
kiln-agent: `SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE = 200` now exists twice under the same name — here and at `segments.ts:180` — and each comment claims to mirror the other, but neither constant is exported so nothing enforces that.

Failure mode: someone tunes the value in `segments.ts` (say to 50 after a slow-query report) and this copy silently stays at 200, so `assertSurveyInteractionSurveyIds` keeps issuing exactly the query shape the tune was meant to remove — while both comments still assert the two are in sync.

Export it from `segments.ts` (helper.ts already imports `getSegment` from there) and import it here, so the two batch paths and their tests move together.

---

apps/web/modules/ee/contacts/segments/lib/helper.test.ts
```diff
@@ -336,4 +340,81 @@ describe("assertSurveyInteractionSurveyIds", () => {
...
+    );
+
+    await expect(
+      assertSurveyInteractionSurveyIds(buildSpecificFilter(surveyIds), "workspace-1")
+    ).rejects.toThrow(new InvalidInputError("Survey not found in workspace: s_205"));
+
+    expect(mockSurveyFindMany).toHaveBeenCalledTimes(2);
+  });
+
+  test("a pre-existing tree over the Zod tree cap still flows through without throwing on size", async () => {
```
kiln-agent: This test — and the matching rationale in `helper.ts:65-68` ("this guard also runs over trees persisted before that cap existed") — describes a state this function cannot be in.

Both callers parse through `ZSegmentFilters` immediately before calling it: `actions.ts:87-95` (create) and `actions.ts:161-172` (update). `assertSurveyInteractionSurveyIds` is never handed a tree read back from the database, so an over-cap tree can't reach it.

The batching here is still right, just for a different reason: a **within**-cap tree already yields up to 1000 x 100 = 100k ids. Worth fixing the comment to say that, and replacing this test with one that pins the real bound. The path that genuinely does receive unvalidated trees is the draft-save one (see my comment on `packages/types/segment.ts`), and that one has no coverage.

---

kiln-agent (commented):
Reviewed the diff for correctness. The batching changes themselves are sound: both loops are correct (empty input naturally skips the DB, batches run in order so the first missing id in collection order still wins and later batches are skipped), the recursive node count is right, `ZBaseFilter`/`ZBaseFilters` really are dead (no consumers in the repo, and `@formbricks/types` is `private`, so no external ones), and I verified `helper.test.ts` + `segments.test.ts` + `segment-schema.test.ts` + the v3 `schemas`/`targeting` suites green locally, plus `tsc --noEmit` on `@formbricks/types`.

Four comments, two of them about the Zod-boundary layer rather than the batching:

1. **The cap is bypassable** — the draft-survey save path writes `segment.filters` without ever parsing `ZSegmentFilters`, so "all entry points parse through this schema" doesn't hold.
2. **Cap value vs. read-time cost** — 1000 bounds the write payload but permits ~1000 sequential (partly unindexed raw) queries per request on the public SDK `/user` endpoint.
3. Duplicated `SURVEY_WORKSPACE_LOOKUP_BATCH_SIZE` constant across the two modules.
4. A new test and a comment assert a scenario that can't occur for `assertSurveyInteractionSurveyIds`.

---

apps/web/lib/survey/service.ts
```diff
@@ -429,6 +433,15 @@ export const updateSurveyInternal = async (
     // if the survey body has type other than "app" but has a private segment, we delete that segment, and if it has a public segment, we disconnect from to the survey
     if (segment) {
       if (type === "app") {
+        // ENG-2305: tree bounds are enforced UNCONDITIONALLY — the draft save (skipValidation)
+        // deliberately skips full semantic validation so half-built filters can be saved, but an
+        // over-bounds tree persisted through it would break every consumer that parses the row
+        // back (publish validation, clone, evaluation).
+        const boundsViolation = getSegmentFilterTreeBoundsViolation(segment.filters);
```
BhagyaAmarasinghe, line 442: [P2] Cap linked survey IDs on draft saves

This guard only bounds `segment.filters`, but the lenient `ZSurveyDraft` leaves `segment.surveys` unvalidated. An authenticated editor can therefore submit an arbitrarily large unique array; `getSurveyWorkspaceIdMap` performs one sequential Prisma query per 200 values before the ownership check rejects unknown IDs. A targeted probe confirmed that 1,001 arbitrary survey strings pass the draft schema despite the normal 500-link cap. Apply the same `ZId` and `MAX_SEGMENT_SURVEYS` controls before this lookup, and add a test proving an over-limit draft performs no database queries.

BhagyaAmarasinghe (changes requested):
Requesting changes for the P2 resource-boundary issue below. Reviewed against exact head 04c91b042af2663231a3ec7814da83d4f4f30e0d.

apps/web/lib/survey/service.ts
pandeymangg: done, added the cap in `updateSurveyInternal`

BhagyaAmarasinghe (approved):

PR #8958, opened 2026-08-24
fix(logger): make survey-API 500s diagnosable [ENG-2578]

Dhruwang:
Ref ENG-2578

## What & why

ENG-2578 reports "An unexpected error occurred" when creating a survey from a template. The template is a red herring — the reporter later confirmed it hits **every** template, "start from scratch" too, and that retrying sometimes succeeds. Production logs show the same 500 on `GET /api/v3/surveys`:

```
{"requestId":"…","method":"POST","path":"/api/v3/surveys/templates",
 "error":{"clientVersion":"7.8.0"},"statusCode":500,"msg":"V3 API unexpected error"}
```

`clientVersion: "7.8.0"` is a Prisma error — and it is the *only* field that survived. That is the actual bug this PR fixes: **the logs cannot tell you what went wrong.**

- **`packages/logger`** — Pino applies a serializer only to keys that have one registered. The logger registered `err`; essentially every call site in the app logs a caught exception as `{ error }`. `name`, `message` and `stack` live on `Error.prototype` and are non-enumerable, so those call sites emitted only the subclass's own enumerable fields — for a Prisma failure, `clientVersion` alone. No type, no message, no `code`, no stack, so a connection-pool timeout is indistinguishable from a bad query. Registering the same serializer for `error` fixes it for every existing call site at once; it returns non-error values unchanged, so keys carrying something else are unaffected.

- **`apps/web/app/api/v3/surveys`** — the create and patch handlers mapped only errors they knew by type. A bare `ValidationError` — what `validateInputs` throws when `createSurvey`/`updateSurvey` re-check their input against `ZSurveyCreateInput` — fell through to the catch-all and came back as a 500 with no field and no reason, so a survey that broke a document rule looked identical to a server fault. Mapped to 422 with the reason, the way `V3SurveyReferenceValidationError` already is.

That second gap is reachable today from the template catalog: **`employee-satisfaction` cannot be created in `nl-NL`, `pt-BR`, `ro-RO` or `zh-Hant-TW`** (and `career-development-survey` in `tr-TR`), because "Extremely likely" and "Very likely" translate to the same string and trip the "no two choices may share a label" rule. That needs a translation change, not a code change — see Open gaps. The new catalog sweep is what surfaces this class of breakage as a failing test instead of a user's failing request.

<details>
<summary>What the 500 in ENG-2578 is <em>not</em></summary>

Ruled out by reproducing the real route handler against a real Postgres (`createTrustedTemplateSurveyResponse`, auth stubbed, everything else live):

- **Template data** — all 52 catalog templates create successfully in `en-US`, including `churn-survey` and `site-abandonment`, across `link` and `app`. Across all 15 locales, only the 5 duplicate-label combinations above fail.
- **The post-commit path** — `subscribeOrganizationMembersToSurveyResponses` read-modify-writes `user.notificationSettings` after the survey row is committed and looked like a plausible flake. 780 consecutive creates for one user ran it 780 times without a failure.
- **Audit logging / rate limiting** — `queueV3AuditLog` swallows its own errors; a rate-limit failure returns 429, not 500.

Combined with the intermittency, both pods affected, and the same error on the read path, the remaining cause is database connectivity in that environment — pool exhaustion or a flapping connection. Which of those it is should be readable straight off the log line once this PR is deployed.
</details>

## Breaking changes

None. The logger change adds a serializer for a key that previously had none, so `error` fields in log output gain `type`/`message`/`stack` — a shape change in logs, not in any API, env var or config key. The 422 replaces a 500 on a path that was already failing.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| An exception logged as `{ error }` keeps its type, message and stack | unit (red on main) | `packages/logger/src/logger.test.ts` → "serializes the `error` key the same way as `err`". Deleting the `error: stdSerializers.err` line at `packages/logger/src/logger.ts:41` fails it; `pnpm --filter @formbricks/logger test` |
| A service-layer `ValidationError` returns 422 with the reason, not an opaque 500 | unit (red on main) | `apps/web/app/api/v3/surveys/lib/operations.test.ts` → "reports the reason for a service-layer ValidationError instead of an opaque 500". Asserts status 422 and that `invalid_params[0].reason` carries the rule that was broken |
| Every catalog template survives all three create gates | unit (mutation) | `apps/web/app/api/v3/surveys/templates/lib/template-catalog-create.test.ts`. Passes on `main`; pointing `churn_survey_question_1_choice_2` at `..._choice_1` in `apps/web/app/lib/templates.ts:520` fails it with `churn-survey: survey document — blocks.0.elements.0.choices: Element 1 in block 1 has duplicate choice labels` |
| Templates still create end to end after the 422 mapping | manual | Real Postgres 16 + pgvector, migrations applied, `createTrustedTemplateSurveyResponse` with auth stubbed: `churn-survey` and `site-abandonment` → 201; `employee-satisfaction`/`nl-NL` → 422 containing "duplicate choice labels" (was 500 "An unexpected error occurred.") |

**Open gaps**

- **ENG-2578's own intermittent 500 is not fixed here** — the evidence points at database connectivity in that environment, not at code on this path. This PR makes the next occurrence say which Prisma error it is. Hence `Ref`, not `Fixes`.
- **`employee-satisfaction` in `nl-NL`/`pt-BR`/`ro-RO`/`zh-Hant-TW` and `career-development-survey` in `tr-TR` are still uncreatable.** The collision is in the machine translations, and `AGENTS.md` allows edits only to `en-US.json` with `pnpm i18n` to regenerate — which needs a `LINGO_API_KEY` I do not have. Needs either re-translation of those keys or en-US labels that stay distinct once translated. The catalog sweep therefore asserts `en-US` only; widening it to all 15 locales is the follow-up once the translations are fixed.
- Not exercised against a live Redis (the harness ran with caching disabled), so the Redis-backed cache and rate-limit paths are unchanged but unverified here.

**Risks**

- The logger change alters the shape of every `{ error }` field in log output — anything parsing those lines (SigNoz dashboards, alert rules) now sees `error.type`/`error.message` instead of a bare object. Worth a glance at any saved query that reads `error` as a scalar.
- The 422 is a new status on `POST /api/v3/surveys` and `PATCH /api/v3/surveys/:id` for inputs that previously 500'd. Clients treating only 5xx as retryable will now stop retrying — which is the intent, since these are not transient.

---

> [!NOTE]

---
_Generated by [Claude Code](https://claude.ai/code/session_01SkkcSQy3Z8hcpnAnh9CuT4)_

### DIFF
--- a/apps/web/app/api/v3/surveys/lib/operations.ts
+++ b/apps/web/app/api/v3/surveys/lib/operations.ts
@@ -1,7 +1,12 @@
 import "server-only";
 import { z } from "zod";
 import { logger } from "@formbricks/logger";
-import { DatabaseError, InvalidInputError, ResourceNotFoundError } from "@formbricks/types/errors";
+import {
+  DatabaseError,
+  InvalidInputError,
+  ResourceNotFoundError,
+  ValidationError,
+} from "@formbricks/types/errors";
 import { requireV3WorkspaceAccess } from "@/app/api/v3/lib/auth";
 import {
   createdResponse,
@@ -234,6 +239,18 @@ function mapV3SurveyCreateError(
     log.warn({ statusCode: 422, invalidParams: err.invalidParams }, "Survey document validation failed");
     return problemUnprocessableContent(requestId, "Survey document failed validation", {
       invalid_params: err.invalidParams,
+      instance,
+    });
+  }
+  if (err instanceof ValidationError) {
+    // `createSurvey` re-validates its input against ZSurveyCreateInput via `validateInputs`, which
+    // throws a bare ValidationError. That is the same class of failure as above — a well-formed
+    // document that breaks a semantic rule (duplicate choice labels, dangling logic targets) — so it
+    // must report the reason as a 422. Left unmapped it fell through to the catch-all 500 "An
+    // unexpected error occurred.", which told the caller nothing about what to fix (ENG-2578).
+    log.warn({ statusCode: 422, errorCode: err.name }, "Survey document validation failed");
+    return problemUnprocessableContent(requestId, "Survey document failed validation", {
+      invalid_params: [{ name: "body", reason: err.message }],
       instance,
     });
   }
@@ -588,6 +605,16 @@ function mapV3SurveyPatchError(
     );
     return problemUnprocessableContent(requestId, "Survey document failed validation", {
       invalid_params: err.invalidParams,
+      instance,
+    });
+  }
+
+  if (err instanceof ValidationError) {
+    // Same as the create path: `updateSurvey`'s own `validateInputs` throws a bare ValidationError for
+    // a semantic rule break, which would otherwise be reported as an opaque 500.
+    log.warn({ statusCode: 422, workspaceId, errorCode: err.name }, "Survey document validation failed");
+    return problemUnprocessableContent(requestId, "Survey document failed validation", {
+      invalid_params: [{ name: "body", reason: err.message }],
       instance,
     });
   }
--- a/packages/logger/src/logger.ts
+++ b/packages/logger/src/logger.ts
@@ -38,6 +38,14 @@ const baseLoggerConfig: LoggerOptions = {
   level: getLogLevel(),
   serializers: {
     err: stdSerializers.err,
+    // Almost every call site in the app logs a caught exception as `{ error }`, not `{ err }`, and
+    // without a serializer for that key Pino writes the object as-is. An Error's `name`, `message`
+    // and `stack` live on the prototype and are non-enumerable, so all that survives is whatever own
+    // enumerable fields the subclass happens to add — a Prisma failure logged as
+    // `{"error":{"clientVersion":"7.8.0"}}`, with no type, message or code to act on (ENG-2578).
+    // `stdSerializers.err` returns non-error values untouched, so keys that carry something other
+    // than an Error are unaffected.
+    error: stdSerializers.err,
     req: stdSerializers.req,
     res: stdSerializers.res,
   },
--- a/apps/web/app/api/v3/surveys/lib/operations.test.ts
+++ b/apps/web/app/api/v3/surveys/lib/operations.test.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, test, vi } from "vitest";
-import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
+import { DatabaseError, ResourceNotFoundError, ValidationError } from "@formbricks/types/errors";
 import { requireV3WorkspaceAccess } from "@/app/api/v3/lib/auth";
 import { problemForbidden } from "@/app/api/v3/lib/response";
 import { capturePostHogEvent } from "@/lib/posthog";
@@ -482,6 +482,32 @@ describe("createV3SurveyResponse", () => {
         })
       ).status
     ).toBe(500);
+  });
+
+  test("reports the reason for a service-layer ValidationError instead of an opaque 500", async () => {
+    vi.mocked(createV3Survey).mockRejectedValueOnce(
+      new ValidationError(
+        "Validation failed: blocks.5.elements.0.choicesElement 1 in block 6 has duplicate choice labels"
+      )
+    );
+
+    const response = await createV3SurveyResponse({
+      body: parsedCreateBody,
+      authentication,
+      requestId,
+      instance,
+    });
+
+    expect(response.status).toBe(422);
+    await expect(response.json()).resolves.toMatchObject({
+      detail: "Survey document failed validation",
+      invalid_params: [
+        {
+          name: "body",
+          reason: expect.stringContaining("duplicate choice labels"),
+        },
+      ],
+    });
   });
 });
 
--- a/apps/web/app/api/v3/surveys/templates/lib/template-catalog-create.test.ts
+++ b/apps/web/app/api/v3/surveys/templates/lib/template-catalog-create.test.ts
@@ -1,0 +1,107 @@
+import { createInstance } from "i18next";
+import type { TFunction } from "i18next";
+import ICU from "i18next-icu";
+import { beforeAll, describe, expect, test, vi } from "vitest";
+import { ZSurveyCreateInput } from "@formbricks/types/surveys/types";
+import { prepareV3SurveyCreate } from "@/app/api/v3/surveys/prepare";
+import { ZV3CreateSurveyBody } from "@/app/api/v3/surveys/schemas";
+import { templates } from "@/app/lib/templates";
+import { replacePresetPlaceholders } from "@/lib/utils/templates";
+import enUS from "@/locales/en-US.json";
+import { buildV3SurveyCreatePayloadFromTemplate } from "./template-to-v3";
+
+vi.mock("server-only", () => ({}));
+
+/**
+ * The trusted template route rebuilds a template server-side and pushes it through the same three
+ * gates a hand-written create body goes through. Only the first one reports a usable error: the
+ * request schema fails with a 400 naming the field, reference validation with a 422, but
+ * `createSurvey`'s own `validateInputs(ZSurveyCreateInput)` throws a bare ValidationError from deep
+ * inside the service layer. So a template that trips the last gate only ever surfaced as "An
+ * unexpected error occurred." with no indication of which template or which rule (ENG-2578).
+ *
+ * Walking the whole catalog through all three gates is what catches that before a release: the
+ * template bodies are generated from translations, so a rule like "no two choices may share a label"
+ * can break for a single template without any code change.
+ */
+const t = (): TFunction => {
+  const instance = createInstance();
+  // ICU matches the server's `getTranslate`, so a template string that ICU cannot parse fails here
+  // rather than at request time.
+  instance.use(ICU).init({
+    lng: "en-US",
+    fallbackLng: "en-US",
+    resources: { "en-US": { translation: enUS } },
+    interpolation: { escapeValue: false },
+  });
+  return instance.getFixedT("en-US");
+};
+
+const workspace = { id: "cmnh38nzx00003b6r3svd9pv2", name: "Acme" };
+
+let catalogIds: string[];
+let translate: TFunction;
+
+beforeAll(() => {
+  translate = t();
+  catalogIds = templates(translate).map((template) => template.id);
+});
+
+describe("catalog templates are creatable", () => {
+  test("the catalog is non-empty", () => {
+    expect(catalogIds.length).toBeGreaterThan(0);
+  });
+
+  test.each(["link", "app"] as const)("every template passes every create gate as %s", (surveyType) => {
+    const rejected: string[] = [];
+
+    for (const template of templates(translate).map((raw) => replacePresetPlaceholders(raw, workspace))) {
+      const body = ZV3CreateSurveyBody.safeParse(
+        buildV3SurveyCreatePayloadFromTemplate({
+          template,
+          workspaceId: workspace.id,
+          surveyType,
+          defaultLanguage: "en-US",
+        })
+      );
+
+      if (!body.success) {
+        rejected.push(`${template.id}: request schema — ${body.error.issues[0]?.message}`);
+        continue;
+      }
+
+      const prepared = prepareV3SurveyCreate(body.data);
+      if (!prepared.ok) {
+        rejected.push(`${template.id}: references — ${prepared.validation.invalidParams[0]?.reason}`);
+        continue;
+      }
+
+      // The shape `executeV3SurveyCreate` hands to `createSurvey`: a template declares no languages,
+      // triggers or targeting, so the document fields are the only ones that vary per template.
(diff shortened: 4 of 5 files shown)


apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -237,6 +242,18 @@ function mapV3SurveyCreateError(
       instance,
     });
   }
+  if (err instanceof ValidationError) {
+    // `createSurvey` re-validates its input against ZSurveyCreateInput via `validateInputs`, which
+    // throws a bare ValidationError. That is the same class of failure as above — a well-formed
+    // document that breaks a semantic rule (duplicate choice labels, dangling logic targets) — so it
+    // must report the reason as a 422. Left unmapped it fell through to the catch-all 500 "An
+    // unexpected error occurred.", which told the caller nothing about what to fix (ENG-2578).
+    log.warn({ statusCode: 422, errorCode: err.name }, "Survey document validation failed");
```
kiln-agent, line 251: This log line drops the one thing the PR is about. `errorCode` is always the constant `"ValidationError"`, and neither `err.message` nor a stack is recorded — so after this change the reason exists **only** in the 422 the caller receives, which is not retained server-side.

Concretely, for `POST /api/v3/surveys/templates` with `employee-satisfaction` + `nl-NL`: previously it fell into the catch-all, which logs `{ error: err }` (and with this PR's logger fix that line would have carried type/message/stack). Now the log says `warn {statusCode:422, errorCode:"ValidationError"} "Survey document validation failed"` and an operator staring at the ENG-2578 line still cannot tell which rule broke or which template. The sibling `V3SurveyReferenceValidationError` branch 10 lines above logs `invalidParams` for exactly this reason.

```suggestion
    log.warn({ statusCode: 422, errorCode: err.name, reason: err.message }, "Survey document validation failed");
```

---

apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -592,6 +609,16 @@ function mapV3SurveyPatchError(
     });
   }
 
+  if (err instanceof ValidationError) {
+    // Same as the create path: `updateSurvey`'s own `validateInputs` throws a bare ValidationError for
+    // a semantic rule break, which would otherwise be reported as an opaque 500.
+    log.warn({ statusCode: 422, workspaceId, errorCode: err.name }, "Survey document validation failed");
```
kiln-agent, line 615: Same omission as the create branch: no `err.message`, no stack, and `errorCode` is the constant `"ValidationError"`.

`PATCH /api/v3/surveys/:id` with a body that clears `ZV3PatchSurveyBody` and reference validation but trips a rule inside `updateSurvey`'s `validateInputs` (e.g. two choices that collapse to the same label after the patch merge) now leaves a log line that is identical for every such failure — and unlike the previous catch-all it carries no stack, so there is no way to tell which of the several `validateInputs` calls on the path threw.

```suggestion
    log.warn(
      { statusCode: 422, workspaceId, errorCode: err.name, reason: err.message },
      "Survey document validation failed"
    );
```

---

packages/logger/src/logger.ts
```diff
@@ -38,6 +38,14 @@ const baseLoggerConfig: LoggerOptions = {
...
   serializers: {
     err: stdSerializers.err,
+    // Almost every call site in the app logs a caught exception as `{ error }`, not `{ err }`, and
+    // without a serializer for that key Pino writes the object as-is. An Error's `name`, `message`
+    // and `stack` live on the prototype and are non-enumerable, so all that survives is whatever own
+    // enumerable fields the subclass happens to add — a Prisma failure logged as
+    // `{"error":{"clientVersion":"7.8.0"}}`, with no type, message or code to act on (ENG-2578).
+    // `stdSerializers.err` returns non-error values untouched, so keys that carry something other
+    // than an Error are unaffected.
+    error: stdSerializers.err,
```
kiln-agent, line 48: Agreed this is a net win, but it is not purely additive — two details in the comment don't match what `pino-std-serializers@7` actually does, and one of them removes data from existing log lines:

1. **An error's own `raw` field is silently dropped.** `err-proto.js` defines `raw` as a *non-enumerable accessor* pre-initialised to `{}`, so in `errSerializer`'s copy loop `_err['raw'] === undefined` is false and the original value is skipped; then `_err.raw = err` overwrites it, and being non-enumerable it is never serialized. `stripe-webhook.ts:240` logs `{ error, eventId, eventType }` with a caught `StripeError`, and stripe-node sets `this.raw = raw` as an own property (`node_modules/stripe/cjs/Error.js:53`) — so that line loses the whole Stripe API error payload (`code`, `param`, `doc_url`, `request_log_url`). Same for the Stripe call sites in `billing/lib/metering.ts` and `billing/lib/organization-billing.ts`. Cheap fix at those call sites: log `{ error, stripeRaw: error.raw }`.

2. **"returns non-error values untouched" is not the rule.** `isErrorLike` is `err && typeof err.message === 'string'` — duck-typed, not `instanceof Error`. Any plain object logged under `error` that carries a string `message` is reshaped (gains `type:"Object"` and `stack:""`), and an own `type`/`stack`/`message` field is clobbered by the serializer's own. Nothing in the current 18 `{ error: … }` call sites looks harmed (cache/storage errors are `{ code }` only), but the comment shouldn't promise a guarantee the library doesn't give.

Worth noting the new unit test only asserts the reference identity against a mocked `stdSerializers.err`, so neither behaviour is covered.

---

apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -237,6 +242,18 @@ function mapV3SurveyCreateError(
...
   }
+  if (err instanceof ValidationError) {
+    // `createSurvey` re-validates its input against ZSurveyCreateInput via `validateInputs`, which
+    // throws a bare ValidationError. That is the same class of failure as above — a well-formed
+    // document that breaks a semantic rule (duplicate choice labels, dangling logic targets) — so it
+    // must report the reason as a 422. Left unmapped it fell through to the catch-all 500 "An
+    // unexpected error occurred.", which told the caller nothing about what to fix (ENG-2578).
+    log.warn({ statusCode: 422, errorCode: err.name }, "Survey document validation failed");
+    return problemUnprocessableContent(requestId, "Survey document failed validation", {
+      invalid_params: [{ name: "body", reason: err.message }],
```
kiln-agent, line 253: `name: "body"` is right for `POST /api/v3/surveys`, but wrong for the path that actually motivated this change. On `POST /api/v3/surveys/templates` the document is built server-side from the catalog + translations; the caller sent only `{workspaceId, templateId, source, surveyType, defaultLanguage}`. So `employee-satisfaction` + `nl-NL` now answers 422 with `invalid_params: [{name:"body", reason:"Validation failed: blocks.5.elements.0.choices…"}]` — pointing at a field the caller never supplied and cannot fix.

The second-order effect matters more: a built-in template that is uncreatable for a whole locale is a product defect, and reclassifying it from 5xx to 4xx means it will never trip an error alert or show up in a 5xx dashboard. Per the Open gaps section it is broken for 5 template/locale combinations right now.

Options: keep the 422 only where the document came from the request, and let `createTrustedTemplateSurveyResponse` treat a gate-3 failure as a server-side fault (still logging the reason); or at minimum name a param the template caller actually supplied (`templateId`/`defaultLanguage`) so the response is actionable.

---

kiln-agent (commented):
Reviewed the diff for correctness. The direction is right and the catalog sweep is a good addition — no blocking correctness bug found, but four things worth addressing, all inline.

Verified locally (after `pnpm build --filter=@formbricks/web^...`, which the new test needs): `template-catalog-create.test.ts` (3 passed), `operations.test.ts` (34 passed), the whole `app/api/v3/surveys` suite (317 passed) and `packages/logger` (17 passed) are green. I also checked that the third gate the new test reconstructs really matches `executeV3SurveyCreate` → `createSurvey`'s `validateInputs([workspaceId, ZId], [surveyBody, ZSurveyCreateInput])`, and that the template route passes `surveyCreateInputOverrides: undefined`, so the hand-built shape is faithful. `ValidationError` does not shadow any sibling branch either — none of `V3SurveyReferenceValidationError`, `V3SurveyUnsupportedShapeError`, `InvalidInputError`, `DatabaseError` extends it.

Two non-blocking observations beyond the inline comments:

- **The dry-run endpoint still only knows two of the three gates.** `validateV3Survey` runs `prepareV3SurveyCreateInput`, i.e. schema + reference validation, but not `ZSurveyCreateInput`. So `POST /api/v3/surveys/validate` reports a document with duplicate choice labels as valid, and the subsequent create 422s on it — the same "the gates disagree" problem, one layer up. Pre-existing, but this PR is the natural place to close it and it would make the MCP `validate_survey` tool trustworthy.
- **Every `ValidationError` on the create path is now attributed to the caller**, not just the ones from `createSurvey`'s own `validateInputs`. The other `validateInputs` calls reachable from that path (`getActionClasses`, segment creation, …) receive internally-derived ids, so a throw there means a server bug — which now answers 422 at `warn` level instead of 500 at `error`. Narrowing the branch (or keeping `log.error` for it) would preserve that signal.


---

apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -237,6 +242,18 @@ function mapV3SurveyCreateError(
       instance,
     });
   }
+  if (err instanceof ValidationError) {
```
kiln-agent, line 245: This catches every `ValidationError` that reaches the mapper, not just the one `createSurvey`'s own `validateInputs(ZSurveyCreateInput)` throws. `ValidationError` carries no provenance, and `validateInputs` is a generic service-layer assertion helper — including on code that runs **after** the survey row is committed.

Concretely: `createSurvey` (`apps/web/lib/survey/service.ts:757`) commits inside `prisma.$transaction`, and only then calls `subscribeOrganizationMembersToSurveyResponses` → `updateUser(id, { notificationSettings })` (`apps/web/lib/user/service.ts:63`), which opens with `validateInputs([personId, ZId], [data, ZUserUpdateInput.partial()])`. For a creator whose stored `notificationSettings` doesn't satisfy `ZUserNotificationSettings` (an `alert` entry that isn't a boolean, a non-string-array `unsubscribedOrganizationIds`), that throws a bare `ValidationError` *after* the survey exists. `POST /api/v3/surveys` (and `/templates`) then answers:

```
422 Survey document failed validation
invalid_params: [{ name: "body", reason: "Validation failed: notificationSettings…" }]
```

so the caller is told its document is invalid and — per this PR's own rationale — that the request is not retryable, while the survey was in fact created. It also leaks an internal field path into a public API response. The old 500 was uninformative but at least didn't misattribute the fault.

Worth narrowing the mapping to the ValidationError raised by the create input validation itself rather than the whole `try` block — e.g. wrap just the `createV3Survey` input-validation call, or have `createSurvey`/`updateSurvey` rethrow it as a dedicated document-validation error type — so post-commit failures keep falling through to the catch-all. The same applies to the patch branch below.

---

packages/logger/src/logger.ts
kiln-agent, line 48: The registration itself is right, but the last sentence of the comment isn't: `stdSerializers.err` does **not** return non-error values untouched. In pino-std-serializers 7 the guard is duck-typed:

```js
const isErrorLike = (err) => err && typeof err.message === 'string'
```

So any plain object with a string `message` under the `error` key goes through the error serializer. Two consequences worth knowing about:

- The serializer sets `_err.type = err.constructor.name` **first**, and the `for…in` copy only writes keys where `_err[key] === undefined`. So `logger.error({ error: { type: "quota_exceeded", message: "limit reached" } }, …)` logs `error.type: "Object"` — the original `type` is silently dropped, and `error.stack` comes out as `""`.
- It mutates the logged object (`err[seen] = undefined`, then `delete err[seen]`), so a frozen error-like object under `error` turns the log call itself into a `TypeError` (the module is `'use strict'`).

I checked every `{ error: … }` call site that carries a non-Error today — `CacheError` `{ code }`, the `Record<string,string>` maps from `transformErrorToDetails`, `{ error: string }` from `parseAndValidateJsonBody` — and none has a `message`, so nothing regresses now. But the comment as written will lead the next author to assume an immunity that isn't there. Suggest wording it as "values without a string `message` are returned unchanged".

---

kiln-agent (commented):
Reviewed the diff for correctness. Both source changes are sound in intent and the tests pass locally (`pnpm --filter @formbricks/logger test`; `vitest run app/api/v3/surveys/templates/lib/template-catalog-create.test.ts app/api/v3/surveys/lib/operations.test.ts` after `pnpm build --filter=@formbricks/web^...`). I verified the new catalog sweep is not vacuous — `ZSurveyCreateInput`'s `surveyRefinement` does carry the duplicate-choice-label rule (`packages/types/surveys/types.ts:1507`) — and that the new `ValidationError` branch cannot shadow the other mapped errors, since `V3SurveyReferenceValidationError`, `V3SurveyUnsupportedShapeError`, the permission errors and `InvalidInputError` all extend `Error` directly.

Two findings, one medium and one low, inline.

---

PR #8959, opened 2026-08-24
chore: say "window" instead of "modal" in the AI translation toast

jobenjada:
No Linear ticket — small copy fix.

## What & why

- The AI translation toast said "Please keep this modal open." "Modal" is our component vocabulary,
  not the user's, so it now reads "Please keep this window open."
- The translations modal no longer closes when you click the backdrop. It holds unsaved translation
  drafts, and a stray click outside discarded them. The close button and Escape still work.

Only `en-US.json` is edited; `i18n.lock` carries the new hash for that key so `scan-translations`
stays in sync, and Lingo re-translates the other locales from en-US on commit.

Closing the modal while an AI translation is still running reports success over drafts that are then
discarded. That is a separate bug, tracked in ENG-2600, and deliberately not fixed here.

## Breaking changes

User-facing copy and one modal's dismiss behaviour. No API shape, emitted value, route, env var or
default changes.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Backdrop click no longer closes the modal | manual | Mounted the real `DialogContent` with this PR's props and clicked outside the panel with Playwright: modal stays open. A plain `DialogContent` in the same harness closes on that click, so the check can detect a close |
| Close button and Escape still close it | manual | Close button is enabled and closes; `keyboard.press("Escape")` closes |
| Toast copy | manual | `ai_translating` was the only "modal" in en-US, and is used in exactly one place (`manage-translations-modal.tsx:163`) |
| Lockfile in sync with en-US | manual | `pnpm scan-translations` passes — it hashes every en-US value against `i18n.lock` and fails on a stale entry |

**Open gaps**

- Not clicked through in the running app: this modal needs an AI-enabled workspace and a live AI
  provider, so the dismiss behaviour was verified against the real component rather than the real
  screen. Worth a look on staging.
- The 14 non-English locales keep their old translation of the string until Lingo regenerates. Only
  `es-ES` and `pt-BR` actually say "modal" — the rest already said window/dialog/panel.

**Risks**

- Anyone used to dismissing this modal with a backdrop click now has to use the close button, Cancel
  or Escape.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `unknown` (not exposed in this non-interactive session).

### DIFF
--- a/apps/web/i18n.lock
+++ b/apps/web/i18n.lock
@@ -2939,7 +2939,7 @@ checksums:
     workspace/surveys/edit/ai_instance_not_configured: 939ad7c3240fa8de98a325239f1b36bc
     workspace/surveys/edit/ai_smart_tools_disabled: 13df84ae47d35dfa6e86ffa62f29c75d
     workspace/surveys/edit/ai_translate: f25943cdeffe155ee524428f4daa5da2
-    workspace/surveys/edit/ai_translating: 098a2293b39f9f258d67f926cf03df37
+    workspace/surveys/edit/ai_translating: 5f60cc47ce106b98251c46197e4981c0
     workspace/surveys/edit/ai_translation_all_fields_populated: d78f6a663ea19ce77045970179bd200f
     workspace/surveys/edit/ai_translation_complete: f443d0801404f728e68000b46ca67598
     workspace/surveys/edit/ai_translation_failed: fd356a173d0abde7a0fc660394954cc7
--- a/apps/web/locales/en-US.json
+++ b/apps/web/locales/en-US.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "AI is not configured. Contact your administrator.",
         "ai_smart_tools_disabled": "AI smart tools are disabled for this organization.",
         "ai_translate": "Translate with AI",
-        "ai_translating": "Translating with AI... Please keep this modal open.",
+        "ai_translating": "Translating with AI... Please keep this window open.",
         "ai_translation_all_fields_populated": "All fields are already translated",
         "ai_translation_complete": "AI translation complete",
         "ai_translation_failed": "Translation failed",
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -219,7 +219,15 @@ export const ManageTranslationsModal = ({
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogContent width="wide" className="max-h-[85dvh]">
+      {/* While AI translation is running, closing the modal would discard the incoming translations,
+          so the close button and backdrop clicks are blocked. Escape and Cancel stay live as the
+          deliberate way out. */}
+      <DialogContent
+        width="wide"
+        className="max-h-[85dvh]"
+        disableCloseButton={isTranslating}
+        disableCloseOnOutsideClick={isTranslating}
+        closeOnEscape>
         <DialogHeader>
           <DialogTitle>{t("workspace.surveys.edit.manage_translations")}</DialogTitle>
           <div className="mt-2 flex items-center justify-between">
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -219,15 +219,17 @@ export const ManageTranslationsModal = ({
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      {/* While AI translation is running, closing the modal would discard the incoming translations,
-          so the close button and backdrop clicks are blocked. Escape and Cancel stay live as the
-          deliberate way out. */}
+      {/* Closing mid-translation loses the result outright: the modal stays mounted, so the
+          translations still land in draftTranslations and the success toast still fires, but nothing
+          writes them to localSurvey and the init effect above resets the drafts on reopen. So every
+          way out — close button, backdrop, Escape, Cancel and Save — is blocked until the request
+          settles.
+          handleTranslateWithAI clears isTranslating in a finally, so this cannot latch. */}
       <DialogContent
         width="wide"
         className="max-h-[85dvh]"
         disableCloseButton={isTranslating}
-        disableCloseOnOutsideClick={isTranslating}
-        closeOnEscape>
+        disableCloseOnOutsideClick={isTranslating}>
         <DialogHeader>
           <DialogTitle>{t("workspace.surveys.edit.manage_translations")}</DialogTitle>
           <div className="mt-2 flex items-center justify-between">
@@ -328,10 +330,10 @@ export const ManageTranslationsModal = ({
         </DialogBody>
 
         <DialogFooter>
-          <Button variant="secondary" size="sm" onClick={handleCancel}>
+          <Button variant="secondary" size="sm" onClick={handleCancel} disabled={isTranslating}>
             {t("common.cancel")}
           </Button>
-          <Button size="sm" onClick={handleSave}>
+          <Button size="sm" onClick={handleSave} disabled={isTranslating}>
             {t("common.save")}
           </Button>
         </DialogFooter>
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -219,17 +219,10 @@ export const ManageTranslationsModal = ({
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      {/* Closing mid-translation loses the result outright: the modal stays mounted, so the
-          translations still land in draftTranslations and the success toast still fires, but nothing
-          writes them to localSurvey and the init effect above resets the drafts on reopen. So every
-          way out — close button, backdrop, Escape, Cancel and Save — is blocked until the request
-          settles.
-          handleTranslateWithAI clears isTranslating in a finally, so this cannot latch. */}
-      <DialogContent
-        width="wide"
-        className="max-h-[85dvh]"
-        disableCloseButton={isTranslating}
-        disableCloseOnOutsideClick={isTranslating}>
+      {/* This modal holds unsaved translation drafts, so a stray backdrop click must not discard
+          them. closeOnEscape keeps Escape working, which disableCloseOnOutsideClick would otherwise
+          suppress too. */}
+      <DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick closeOnEscape>
         <DialogHeader>
           <DialogTitle>{t("workspace.surveys.edit.manage_translations")}</DialogTitle>
           <div className="mt-2 flex items-center justify-between">
@@ -330,10 +323,10 @@ export const ManageTranslationsModal = ({
         </DialogBody>
 
         <DialogFooter>
-          <Button variant="secondary" size="sm" onClick={handleCancel} disabled={isTranslating}>
+          <Button variant="secondary" size="sm" onClick={handleCancel}>
             {t("common.cancel")}
           </Button>
-          <Button size="sm" onClick={handleSave} disabled={isTranslating}>
+          <Button size="sm" onClick={handleSave}>
             {t("common.save")}
           </Button>
         </DialogFooter>
--- a/apps/web/modules/ui/components/dialog/index.tsx
+++ b/apps/web/modules/ui/components/dialog/index.tsx
@@ -33,8 +33,6 @@ DialogOverlay.displayName = DialogPrimitive.Overlay.displayName;
 
 interface DialogContentProps {
   hideCloseButton?: boolean;
-  /** Render the close button greyed out and unclickable, e.g. while an operation must not be interrupted. */
-  disableCloseButton?: boolean;
   disableCloseOnOutsideClick?: boolean;
   /** Keep Escape closing the dialog even when disableCloseOnOutsideClick is set. */
   closeOnEscape?: boolean;
@@ -57,14 +55,13 @@ const getDialogWidthClass = (width: "default" | "wide" | "full" | "narrow"): str
 
 const DialogContent = React.forwardRef<
   React.ComponentRef<typeof DialogPrimitive.Content>,
-  Readonly<React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & DialogContentProps>
+  React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & DialogContentProps
 >(
   (
     {
       className,
       children,
       hideCloseButton,
-      disableCloseButton,
       disableCloseOnOutsideClick,
       closeOnEscape,
       width = "default",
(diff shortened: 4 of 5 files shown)


kiln-agent (commented):
## Review

The copy fix is right and the `disableCloseButton` prop is a clean, additive addition — greying the X out rather than removing it does keep the header from reflowing, and every existing `DialogContent` call site is unaffected. One substantive gap in the close-guard, then some nits.

---

### 1. `Cancel` stays live during translation, and it discards the in-flight result — with a success toast

`manage-translations-modal.tsx:331`

The guard blocks the X and the backdrop, but the footer `Cancel` button is untouched and does exactly what the X was just blocked from doing:

1. `handleTranslateWithAI` sets `isTranslating = true` and awaits `translateSurveyFieldsAction`.
2. User clicks `Cancel` (still enabled) → `setOpen(false)`.
3. The action resolves → `setDraftTranslations(prev => ({ ...prev, ...translations }))` and `toast.success(t("...ai_translation_complete"))`.
4. `ManageTranslationsModal` is rendered unconditionally (`language-view.tsx:552`), so the state survives — but nothing is ever written to `localSurvey`; only `handleSave` does that.
5. Reopening runs the init effect at `manage-translations-modal.tsx:72-80`, which resets `draftTranslations` from `localSurvey`.

Net: **the user sees "AI translation complete" and the translations do not exist.** That is precisely the loss the PR sets out to prevent, reachable by one click on a large button sitting immediately left of `Save`.

The PR body's rationale — "they are deliberate actions rather than accidental ones" — holds for Escape, but clicking `Cancel` is a stray click in exactly the same way clicking the X is. Escape on its own already satisfies the stated goal of not trapping the user on a hung request, so disabling `Cancel` costs nothing there.

Worth noting: the repo already has prior art for this exact situation. `create-with-ai-dialog.tsx:83,101-105` handles an in-flight AI operation in a dialog by blocking outside clicks **and disabling Cancel** (`disabled={isFormBusy || isBusy}`), leaving the close button alone. This PR inverts that. Aligning is a one-liner:

```tsx
<Button variant="secondary" size="sm" onClick={handleCancel} disabled={isTranslating}>
  {t("common.cancel")}
</Button>
```

I tried to refute this by looking for a path that preserves the drafts — a conditional mount that would keep the effect from re-running, or a write into `localSurvey` on the AI success path. Neither exists: the parent mounts the modal unconditionally, and `setTranslationAtPathMutable` is only reached from `handleSave`.

### 2. Follow-on: the success toast fires on a discarded result whichever exit is used

Same trace as above, and it still applies via Escape after fixing #1. If the modal has been closed by the time the action resolves, `toast.success("AI translation complete")` is an outright false statement. Not introduced here, but this PR is what makes the closed-mid-translation case a documented, supported flow, so it's the natural place to handle it — either skip the success toast when the dialog is no longer open, or don't reset `draftTranslations` on reopen when drafts are already populated. Your call whether that's in scope.

---

### Nits

- **`closeOnEscape` at `manage-translations-modal.tsx:230` has no runtime effect.** `onEscapeKeyDown` is only overridden when `disableCloseOnOutsideClick && !closeOnEscape`, which is `false` in both states here (`isTranslating` true → `true && !true`; false → short-circuits). Escape closes either way. Harmless as self-documentation — flagging only so nobody later reads it as load-bearing and removes the `disableCloseOnOutsideClick` guard expecting Escape to change behaviour.

- **PR body claim is slightly off.** The coverage table says `ai_translating` "is the only occurrence of 'modal' in en-US". `common.centered_modal` = `"Centered Modal"` is also a value containing it. Different context (survey placement setting, arguably a legitimate product term) so leaving it is defensible — but the claim as written isn't accurate.

- **a11y on the disabled X.** `disabled` drops the button from the tab order and the `sr-only` label still reads just "Close", so a screen-reader user gets no signal that closing is temporarily unavailable — the control simply vanishes from navigation. No trap, since Escape works, so this is minor; `aria-disabled` with a no-op click handler would keep it announced and focusable if you want to be thorough.

---

**Verified clean:** the new prop types without a `StoryProps` change (it derives from `React.ComponentProps<typeof DialogContent>`); `disabled:opacity-50` cannot affect any existing dialog, since nothing else passes the prop and `disabled:pointer-events-none` keeps `hover:opacity-100` from competing; the `i18n.lock` hash is in sync (`Validate Translation Keys` passes); `isTranslating` is cleared in `finally` on both the success and throw paths, so the X cannot get stuck dead.

---

kiln-agent (commented):
## Review

Copy change is right and the lockfile work checks out. The close-guard is where I'd push back: it blocks two of the four ways out, and the two it leaves open don't just discard the drafts — they report success.

---

### 1. Escape/Cancel while translating end in a green "AI translation complete" toast for work that is gone

`language-view.tsx:552` renders `ManageTranslationsModal` unconditionally — closing only flips `open`, the component never unmounts. So when Escape (`manage-translations-modal.tsx:230`, `closeOnEscape`) or Cancel (`:331`, not disabled while `isTranslating`) closes the dialog mid-flight:

- `handleTranslateWithAI` keeps running; its `finally` and `toast.success(t("...ai_translation_complete"))` still fire, and react-hot-toast renders outside the dialog — so the user sees **"AI translation complete"** on a closed modal.
- `setDraftTranslations` writes into the state of a closed modal. Next open, the effect at `:72-80` re-seeds `draftTranslations` from `localSurvey`, and since only `handleSave` ever writes drafts back, the AI output is discarded.

Net: the outcome of Escape/Cancel is identical to the X-click this PR exists to prevent — the toast just claims otherwise. That's the part worth fixing **regardless of which exit policy you pick**: a success toast for discarded work is wrong under either. Options, in increasing strictness:

- Keep Escape/Cancel live, but don't claim success when the dialog is no longer open — swap in a "translation discarded" message (an `openRef` updated in an effect reads correctly inside the async closure; the `open` prop captured at handler-creation time will be stale).
- Or seal it: `disabled={isTranslating}` on the Cancel button and drop `closeOnEscape`. That's the reading the new copy ("Please keep this window open") actually promises, and it matches the X.

I don't think leaving it exactly as-is is defensible, because the toast lies. Your call on which direction.

### 2. `DisabledCloseButton` story demonstrates stricter behaviour than the app ships

`renderModal` (`stories.tsx:255-260`) forwards `hideCloseButton` / `disableCloseButton` / `disableCloseOnOutsideClick` but never `closeOnEscape`. The new story sets `disableCloseOnOutsideClick: true`, so `onEscapeKeyDown` is prevented — and the footer buttons are plain `<Button>`s with no handler or `DialogClose`. Result: X disabled, outside click blocked, Escape blocked, footer inert — the story dialog can't be closed at all, and what it shows isn't the shipped modal, where Escape stays live by design.

`RestrictClose` already has the same trap, so this isn't a regression — but the story added specifically to make the new state inspectable ends up misrepresenting it. Forwarding `closeOnEscape` through `renderModal` and setting it on the new story fixes both.

### 3. `disabled` gives assistive tech nothing to go on (non-blocking)

`disabled` plus the pre-existing `disabled:pointer-events-none` takes the X out of the tab order, announces no reason for the change, and rules out a tooltip. Since the point of the change is telling the user *why* they can't close, `aria-disabled` with a guarded `onClick` would keep the button focusable and let you attach "Translation in progress". Reasonable to defer.

### 4. Nit: the "only occurrence" claim

The coverage table says `ai_translating` is the only occurrence of "modal" in en-US. After this change `common.centered_modal` → `"Centered Modal"` is still a user-facing value containing it. Presumably deliberate (placement is a product term, not component jargon) — just flagging that the claim isn't literally true.

---

### Verified, no action needed

- **Lockfile hash is genuinely correct, not just plausible.** `validateLockfile` in `packages/i18n-utils/src/scan-translations.ts:566-567` hashes each en-US value with `objectHash.MD5(val)` (not raw md5 — I checked, raw md5 of the string doesn't match any entry). `pnpm scan-translations` passes locally against the updated `i18n.lock`.
- **`disableCloseButton` is safely additive.** It's the only new prop, optional, and `manage-translations-modal.tsx:228` is the sole call site; `disabled:opacity-50` can only apply when `disabled` is set, and Radix's `Close` never sets it on its own — so no existing dialog can reach the faded state.
- **Storybook types.** `StoryProps` derives from `React.ComponentProps<typeof DialogContent>`, so the new prop flows through with no story-side type change.
- All 18 checks green, Sonar quality gate passed.

---

jobenjada:
Both findings were valid and are fixed in f3dafb9.

**The `Cancel` gap (@kiln-agent):** confirmed by reading the call site — `ManageTranslationsModal` is rendered unconditionally at `language-view.tsx:552`, so closing it does not unmount, and the init effect resets `draftTranslations` on the next open. The success toast over discarded translations was reachable through `Cancel`, and through `Escape` for the same reason. `Save` had the same hole: it closes the modal, so translations arriving after it would have been lost too.

Rather than keep a "deliberate way out", every exit is now blocked while `isTranslating` — close button, backdrop, Escape, Cancel and Save. `isTranslating` is cleared in a `finally` on both the success and throw paths, so the seal cannot latch.

**`Readonly<>` (@coderabbitai):** applied to the forwarded props intersection as suggested; typecheck is clean.

Re-verified by mounting the real `DialogContent`/`Button` against the app's own `globals.css` and driving them with Playwright. The idle case is the control that proves the probe can see a close at all:

| | idle | translating |
| --- | --- | --- |
| Escape | closes | **stays open** |
| Backdrop click | — | **stays open** |
| Close X | enabled | **disabled**, `opacity: 0.5`, `pointer-events: none` |
| Cancel / Save | enabled | **disabled** |

One thing this PR still does not do, now listed under Open gaps: the toast/discard bug is prevented, not fixed. Writing the translations through to `localSurvey` as they arrive would fix it at the root, and is worth a follow-up.

jobenjada:
**Scope reduced — earlier review threads on this PR are now moot.**

The in-flight close guard is reverted in 81ac10b: no `disableCloseButton` prop, no disabled Cancel/Save, no Escape blocking. `dialog/index.tsx` and its stories are back to their state on `main`, so the `Readonly<>` suggestion no longer applies to this diff either.

The data-loss bug those threads were about is real and now tracked as **ENG-2600** rather than worked around here.

What is left is the copy change plus one line: the translations modal sets `disableCloseOnOutsideClick` unconditionally (with `closeOnEscape`, since this `DialogContent` otherwise suppresses Escape along with backdrop clicks), so a stray click outside cannot discard unsaved drafts. The close button stays clickable.

Verified against the real component with Playwright, with a plain `DialogContent` as the control:

| | this PR | plain dialog |
| --- | --- | --- |
| Backdrop click | **stays open** | closes |
| Escape | closes | closes |
| Close button | enabled, closes | enabled, closes |

BhagyaAmarasinghe (changes requested):
## Review

Reviewed against `f3dafb9d6`, including how `i18n.lock` is actually consumed. All 19 checks are green — the one finding below is structurally invisible to CI.

The diagnosis in the description is correct and I verified it independently: `ManageTranslationsModal` is rendered unconditionally in `language-view.tsx:552`, so `open={false}` only unmounts the Radix portal — the component keeps its state, `draftTranslations` still receives the AI result, the success toast still fires, and only `handleSave` writes to `localSurvey`. Sealing the modal is a legitimate fix. The `finally` genuinely covers every path, and the seal is bounded: `AI_TRANSLATION_TIMEOUT_MS = 45_000` in `apps/web/modules/ee/ai-translation/lib/translate-fields.ts:15`, so worst case is ~45s locked, not indefinite.

One blocker, in the copy half rather than the logic half.

---

### 🔴 Hand-bumping the lockfile checksum stops the 14 target locales from ever being re-translated

`apps/web/i18n.lock:2942`

`i18n.lock` is Lingo's change-detection ledger, not a CI checksum artifact. `pnpm i18n` runs `lingo.dev run` (translate whatever the lock says changed) **then** `lingo.dev lockfile --force` (rewrite the hashes) — in that order, deliberately. Writing the new hash first destroys the delta signal, so the next run sees `ai_translating` as already handled and skips it.

The committed hash is exactly the new en-US value's:

```
node -e "console.log(require('object-hash').MD5('Translating with AI... Please keep this window open.'))"
# 5f60cc47ce106b98251c46197e4981c0  ← the value in the diff
```

Which is precisely why **Validate Translation Keys passes** — and nothing downstream can catch it either: `validateLockfile` (`packages/i18n-utils/src/scan-translations.ts:546`) compares en-US against the lock and the lock against en-US, never the target locales. Current state on the branch:

```
es-ES: Traduciendo con IA... Por favor, mantén este modal abierto.
pt-BR: Traduzindo com IA... Por favor, mantenha este modal aberto.
```

Two supporting facts. In the last 60 commits touching `apps/web/i18n.lock`, no commit has ever changed an existing key's hash in place — only additions and removals. And every merged en-US change lands with all 14 target locales in the same commit (16 files: en-US + 14 targets + lock). Nothing in the repo re-translates on commit, either: no lingo step in `.husky/pre-commit`, `.lintstagedrc.mjs`, or any workflow — so "Lingo re-translates the other locales from en-US on commit" is doing load-bearing work in the description that the repo does not back up.

As it stands this merges with 14 stale locale files and a lock asserting they are in sync, and no future check will ever flag it.

**Fix:** revert the lock line to `098a2293b39f9f258d67f926cf03df37` and run `pnpm i18n` (needs `LINGODOTDEV_API_KEY`), so `run` sees the stale hash, regenerates the targets, and `lockfile --force` writes the new hash itself.

---

### 🟡 Sealing a dialog takes three coordinated props and no single one does it

`apps/web/modules/ui/components/dialog/index.tsx:36`

`DialogContent` now carries four close-related booleans, and the correct seal is "set `disableCloseButton`, set `disableCloseOnOutsideClick`, and do *not* set `closeOnEscape`". `disableCloseButton` alone leaves the backdrop and Escape open — the exact half-guarded state the description says the first pass shipped. The escape hatch that made the bug reachable was the Cancel button, but the API shape is what allowed it.

A single `preventClose` driving all three vectors, with the granular props left for existing callers, would make the correct usage unmissable. Not blocking.

### 🟡 The comment names Escape as blocked but not which prop blocks it

`apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx:222`

Escape is blocked only as a side effect of `disableCloseOnOutsideClick && !closeOnEscape` (`dialog/index.tsx:88`). Someone later removing `disableCloseOnOutsideClick` — a prop whose name says nothing about keyboards — silently reopens the data-loss path. One clause noting that `disableCloseOnOutsideClick` is also what blocks Escape here would carry the coupling.

### 🟢 The new story cannot be dismissed once opened

`apps/web/modules/ui/components/dialog/stories.tsx:449`

`DisabledCloseButton` sets `disableCloseButton: true` + `disableCloseOnOutsideClick: true`, and `renderModal`'s footer buttons are plain `Button`s with no `onClick` or `DialogClose`. X disabled, backdrop prevented, Escape prevented, footer inert — the only way out is reloading the iframe. `RestrictClose` already has this shape so it is consistent with precedent, but `closeOnEscape: true` would make the story actually inspectable, which is the stated reason it exists.

---

### What looks good

- `finally` clears both `isTranslating` and `translatingPaths` on the success path, the `!result?.data?.translations` early return, and the throw — the seal cannot latch.
- `disabled:opacity-50` cannot be overridden by the pre-existing `hover:opacity-100`: `disabled:pointer-events-none` stops `:hover` matching at all, and Tailwind orders `disabled` after `hover` regardless.
- `disableCloseButton` over `hideCloseButton` is the right call — no header reflow mid-operation.
- `StoryProps` picks the new prop up through `React.ComponentProps<typeof DialogContent>`, so no story type drift.
- The copy claim checks out: walking en-US *values* rather than key names finds only `ai_translating` and `common.centered_modal`, and the latter is the survey placement option — real product terminology.

### Two things worth adding to the description

- **Risks** only considers the latch case. The realistic exposure is the bounded one: the modal is genuinely unclosable for up to the 45s server timeout plus network, and `disabled` on Save means manual edits made *before* clicking Translate cannot be saved during that window. Deliberate, but it is the risk a user will actually meet.
- **Open gaps** misses one residual path: the init effect at `manage-translations-modal.tsx:72` depends on `strings`, memoized on `localSurvey`. Any new `localSurvey` identity while the modal is open wipes `draftTranslations` mid-flight, seal or no seal. Pre-existing and probably unreachable behind the overlay, but "every way out is blocked" reads stronger than what is actually guaranteed.

---

**Verdict:** the lockfile edit needs to be undone and replaced with a real `pnpm i18n` run — without it the PR ships the "modal" wording permanently in es-ES and pt-BR while asserting the opposite. The dialog logic itself is sound and I would approve it on its own.

PR #8960, opened 2026-08-24
feat: add Formbricks logo to MCP OAuth consent page

jobenjada:
No Linear ticket — small UI polish requested directly by the team.

## What & why

The `/account/authorize` page (shown when an MCP client like Claude requests access to a user's Formbricks account) rendered as a plain, unbranded card — no logo, nothing to indicate it's actually Formbricks. That makes it look untrustworthy/phishy right at the moment a user is asked to grant account access.

Add the same `Logo` component used on login/signup (`@/modules/ui/components/logo`), linked to formbricks.com, above both the consent card and the invalid-request error card, matching the branding pattern already established by `FormWrapper` on the auth pages.

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Logo renders above the consent card (client name, scopes, allow/deny) | manual (static reproduction of the JSX/Tailwind markup rendered in-browser) | Logo displays centered above the card, linked out, matches auth-page branding |
| Logo renders above the invalid-request error card | manual (code inspection — same `OAuthConsentLogo` component used in both branches) | Present in both render paths |

**Open gaps**

- Not exercised against the actual running `/account/authorize` route end-to-end (no Postgres/Docker available in this environment to boot the full app + a real MCP OAuth handshake). Verified the exact markup/classes visually via a standalone reproduction instead. Worth a quick manual check on a preview deploy before merge.

**Risks**

- Low risk: purely additive JSX (a linked logo) in a Server Component; no logic, props, or existing markup changed.

---

Written by claude-sonnet-5 (Claude Code) at effort level: medium.

### DIFF
--- a/apps/web/app/(app)/account/authorize/page.tsx
+++ b/apps/web/app/(app)/account/authorize/page.tsx
@@ -59,13 +59,14 @@ const getPublicOAuthClient = async (clientId: string): Promise<TOAuthPublicClien
   }
 };
 
-const OAuthConsentLogo = () => (
+const OAuthConsentLogo = ({ label }: Readonly<{ label: string }>) => (
   <div className="mb-6 flex justify-center">
     <Link
       target="_blank"
       href="https://formbricks.com?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=oauth_consent_logo"
-      rel="noopener noreferrer">
-      <Logo className="h-8 w-auto" />
+      rel="noopener noreferrer"
+      aria-label={label}>
+      <Logo aria-hidden="true" className="h-8 w-auto" />
     </Link>
   </div>
 );
@@ -89,7 +90,7 @@ const Page = async ({ searchParams }: Readonly<{ searchParams: Promise<TSearchPa
     return (
       <div className="flex min-h-screen items-center justify-center bg-slate-50 px-4 py-10">
         <div className="w-full max-w-xl rounded-lg border border-slate-200 bg-white p-6 shadow-sm">
-          <OAuthConsentLogo />
+          <OAuthConsentLogo label={t("auth.oauth.formbricks_website")} />
           <Alert variant="error" role="status">
             <AlertTitle>{t("auth.oauth.invalid_oauth_request")}</AlertTitle>
             <AlertDescription>{t("auth.oauth.invalid_oauth_request_description")}</AlertDescription>
@@ -105,7 +106,7 @@ const Page = async ({ searchParams }: Readonly<{ searchParams: Promise<TSearchPa
   return (
     <div className="flex min-h-screen items-center justify-center bg-slate-50 px-4 py-10">
       <div className="w-full max-w-xl rounded-lg border border-slate-200 bg-white p-6 shadow-sm">
-        <OAuthConsentLogo />
+        <OAuthConsentLogo label={t("auth.oauth.formbricks_website")} />
         <div className="space-y-1">
           <p className="text-sm font-medium text-slate-500">{t("auth.oauth.authorization_request")}</p>
           <h1 className="text-2xl font-semibold text-slate-900">
--- a/apps/web/app/(app)/account/authorize/page.tsx
+++ b/apps/web/app/(app)/account/authorize/page.tsx
@@ -1,4 +1,5 @@
 import { headers } from "next/headers";
+import Link from "next/link";
 import { redirect } from "next/navigation";
 import { getTranslate } from "@/lingodotdev/server";
 import { auth } from "@/modules/auth/lib/auth";
@@ -10,6 +11,7 @@ import {
 } from "@/modules/auth/lib/oauth-client-metadata";
 import { getSession } from "@/modules/auth/lib/session";
 import { Alert, AlertDescription, AlertTitle } from "@/modules/ui/components/alert";
+import { Logo } from "@/modules/ui/components/logo";
 import { OAuthConsentActions } from "./components/OAuthConsentActions";
 
 type TSearchParams = Record<string, string | string[] | undefined>;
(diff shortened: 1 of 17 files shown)


apps/web/app/(app)/account/authorize/page.tsx
```diff
@@ -57,6 +59,17 @@ const getPublicOAuthClient = async (clientId: string): Promise<TOAuthPublicClien
   }
 };
 
+const OAuthConsentLogo = () => (
+  <div className="mb-6 flex justify-center">
+    <Link
```
kiln-agent, line 64: The new link has no accessible name. `Logo` renders a bare `<svg>` with no `<title>` or `role="img"`, and neither the anchor nor the SVG carries an `aria-label` — so a screen reader announces an unnamed link and falls back to reading the raw utm-laden href, and voice-control users have nothing to speak.

Worth fixing here specifically because of what this page is: the branding cue this PR adds is the trust signal shown while a user grants an MCP client account access, and it is the one part of that signal a non-sighted user currently cannot perceive. `apps/web` doesn't load `eslint-plugin-jsx-a11y`, so CI won't flag it.

```suggestion
    <Link
      target="_blank"
      href="https://formbricks.com?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=oauth_consent_logo"
      rel="noopener noreferrer"
      aria-label="Formbricks">
```

Same gap exists at `apps/web/modules/auth/components/form-wrapper.tsx:13`, so fixing it inside `Logo` (a `<title>` plus `role="img"`) would cover both — your call whether that belongs in this PR.

---

apps/web/app/(app)/account/authorize/page.tsx
```diff
@@ -57,6 +59,17 @@ const getPublicOAuthClient = async (clientId: string): Promise<TOAuthPublicClien
...
+      target="_blank"
+      href="https://formbricks.com?utm_source=formbricks-app&utm_medium=webapp&utm_campaign=oauth_consent_logo"
+      rel="noopener noreferrer">
+      <Logo className="h-8 w-auto" />
```
kiln-agent: **Duplicates the brand-link block in `FormWrapper`** — low severity, non-blocking.

`apps/web/modules/auth/components/form-wrapper.tsx:13-19` is the same `Link` + `Logo` + utm markup; the only differences here are `utm_campaign` and the sizing (`h-8 w-auto` vs `mx-auto w-3/4`). The two surfaces already render the logo at different sizes, and anything that changes the brand link (domain, `rel`, an added accessible name — see CodeRabbit's comment above) now has to be changed in two places.

Worth extracting a small shared `LogoLink` that takes `utmCampaign` + `className`, so the a11y fix lands once rather than twice. Fine to do as a follow-up if you'd rather keep this PR to the one-line addition.

---

kiln-agent (commented):
Reviewed the diff (single file, 15 added lines, purely additive JSX). One finding, inline — a missing accessible name on the new logo link. Nothing blocking.

**What I checked and cleared**

- **Props/rendering** — `Logo` spreads props onto the root `<svg>`, so `className` lands. `h-8 w-auto` differs from `FormWrapper`'s `w-3/4`, but the SVG has a `viewBox` and no `width`/`height` attributes, so browsers derive the intrinsic 697:150 ratio and resolve `width: auto` from the definite 32px height (~149px wide). Renders about 0.6× the auth-page size — intentional-looking, not a defect.
- **Duplicate DOM ids** — the SVG hardcodes `mask0_2810_6010`, `filter0_d_…`, `paint0_linear_…`, which would collide if `Logo` rendered twice on one page. It doesn't: `OAuthConsentLogo` sits in two mutually exclusive return branches, and `app/(app)/layout.tsx` renders no logo.
- **Link hygiene** — `target="_blank"` + `rel="noopener noreferrer"` is right, and the new tab means clicking the logo doesn't abandon the in-flight consent flow. The literal `&` separators are a JS string, not HTML-entity-parsed, so the utm params survive.
- **Callers/tests** — no unit or Playwright spec asserts on this page's markup (`consent-redirect.test.ts` and `better-auth-schema-contract.test.ts` don't go through the DOM). No logic, props, or existing markup changed; the `redirect()` auth guard is untouched.
- **Conventions** — no new user-facing strings, so no `en-US.json` / `pnpm i18n` obligation; import order matches the enforced grouping; `OAuthConsentLogo` takes no props, so `Readonly<>` is moot; no `process.env` read.

**One non-code note:** the description says the logo sits "above both the consent card and the invalid-request error card," but the code places it *inside* each card as its first child. The result looks fine — just worth correcting so the description matches the diff.

---

jobenjada:
ill add the screenshots

itsjavi:
**Visual evidence**

Captured on the real authenticated `/account/authorize` route at 1280 × 820 CSS px, 2× device scale, in light mode. Both pairs use the same dynamically registered “Claude Desktop” OAuth client and identical interaction state.

<details>
<summary>Valid OAuth consent request</summary>

| Before | After |
| --- | --- |
| OAuth consent card without Formbricks branding | Formbricks logo centered above the Claude Desktop OAuth consent request |

</details>

<details>
<summary>Invalid OAuth request</summary>

| Before | After |
| --- | --- |
| Invalid OAuth request card without Formbricks branding | Formbricks logo centered above the invalid OAuth request alert |

</details>

itsjavi (approved):

itsjavi (changes requested):
LGTM but kiln has some valid concerns, specially about accessible name

apps/web/app/(app)/account/authorize/page.tsx
jobenjada: Fixed in 3c9a2886cd: `OAuthConsentLogo` now takes a `label` prop, rendered as `aria-label` on the `Link` and passed `t("auth.oauth.formbricks_website")` from both call sites (with a new `en-US.json` key generated across all locales via `pnpm i18n`). The `Logo` SVG is marked `aria-hidden="true"` since it's now purely decorative.

apps/web/app/(app)/account/authorize/page.tsx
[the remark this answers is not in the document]
jobenjada: Applied — `OAuthConsentLogo` now takes a `Readonly<{ label: string }>` prop, both call sites pass `t("auth.oauth.formbricks_website")`, the `Link` gets `aria-label`, and `Logo` is `aria-hidden`. New key added to `en-US.json` and propagated to every locale via `pnpm i18n`. See 3c9a2886cd.

jobenjada: Fair point, and agreed it's real duplication now that both spots need the same a11y label. Keeping this PR to the branding fix per your note that it's fine as a follow-up — flagging a `LogoLink` extraction (shared `utmCampaign` + `className` + label) separately rather than growing this diff.

itsjavi (approved):

PR #8961, opened 2026-08-24
fix: default question headlines to normal weight instead of bold

jobenjada:
No ticket — small default-value change requested directly.

## What & why

Survey question headlines render bold by default (`font-semibold` / CSS weight `500`), but recall/fallback substitution inside a headline can't currently be rendered in bold, so a headline mixing literal text and a recall/fallback value looks visibly inconsistent (part bold, part not).

Changes the default headline font-weight to normal (`400`) for every question type, matching the existing description (`400`) and upper-label (`400`) defaults:

- `apps/web/lib/styling/constants.ts` — `STYLE_DEFAULTS.elementHeadlineFontWeight`: `500` → `400`. This is the single source of truth used by the editor's styling form, the "Restore defaults" action, the survey renderer, and the email preview template.
- `packages/survey-ui/src/styles/globals.css` — `--fb-element-headline-font-weight` static CSS fallback: `500` → `400`. Governs every question type except the three below (openText, single/multi-select, rating, NPS, matrix, consent, contactInfo, address, fileUpload, ranking, pictureSelection, CTA, date).
- `packages/surveys/src/components/general/headline.tsx` — removed a hardcoded `font-semibold` class that always forced bold regardless of the styling settings above. This is the `Headline` component used by the welcome card, ending/thank-you card, and the Cal.com booking question — it was also internally inconsistent, since the HTML-headline branch right next to it never had this class.

Workspaces/surveys that already set an explicit `elementHeadlineFontWeight` (via Advanced Styling → Headlines & Descriptions → Headline Font Weight) are unaffected — this only changes what renders when no explicit value is set.

## Breaking changes

None — this only changes a visual default (headline font-weight). No API/SDK shape, env var, route, or config key changes, and nothing requires self-hoster migration action. Surveys/workspaces without an explicit headline-weight override will simply render headlines slightly lighter after deploy.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Question headlines render at normal (400) weight by default | manual | Storybook `Elements/Rating/Default` story: `getComputedStyle` on `.label-headline` returns `font-weight: 400` (was `500`) |
| Hardcoded bold removed from welcome/ending-card/Cal.com headline | manual (built-bundle inspection) | Rebuilt `@formbricks/surveys` and grepped the compiled `apps/web/public/js/surveys.umd.cjs`: the plain-text headline branch's className is now `label-headline text-base` (no `font-semibold`), and `--fb-element-headline-font-weight:400` is the compiled default |
| Existing behaviour (custom headline weight, recall/fallback substitution, HTML headlines) unaffected | unit (guard) | `pnpm test --filter=@formbricks/surveys --filter=@formbricks/survey-ui` — 721 + existing survey-ui tests pass unchanged, including `styles.test.ts` cases that set an explicit `elementHeadlineFontWeight` |

**Open gaps**

- Didn't visually screenshot the welcome card / ending card / Cal.com booking question specifically (no Storybook stories exist for them) — verified via compiled-bundle inspection instead, since they share the exact same `Headline` component and CSS variable as the Storybook-verified path.
- No full end-to-end run against a live DB-backed app in this environment (worktree lacks `DATABASE_URL`/`ENCRYPTION_KEY`/etc.), so the change wasn't clicked through in the actual survey editor/link-survey UI.

**Risks**

- This changes a global rendering default: any live survey/workspace that hasn't explicitly set a headline font weight will visibly render lighter headlines immediately after deploy. This is the intended, requested behavior change, not a regression, but worth calling out since it's a visible change with no opt-in.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5`, reasoning effort `medium`.

### DIFF
--- a/apps/web/lib/styling/constants.ts
+++ b/apps/web/lib/styling/constants.ts
@@ -115,7 +115,7 @@ export const STYLE_DEFAULTS: TWorkspaceStyling = {
   // Headlines & Descriptions
   elementHeadlineColor: { light: _colors["elementHeadlineColor.light"] },
   elementHeadlineFontSize: 16,
-  elementHeadlineFontWeight: 500,
+  elementHeadlineFontWeight: 400,
   elementDescriptionColor: { light: _colors["elementDescriptionColor.light"] },
   elementDescriptionFontSize: 14,
   elementDescriptionFontWeight: 400,
--- a/packages/surveys/src/components/general/headline.tsx
+++ b/packages/surveys/src/components/general/headline.tsx
@@ -47,9 +47,7 @@ export function Headline({
             dangerouslySetInnerHTML={{ __html: safeHtml }}
           />
         ) : (
-          <HeadingTag
-            data-testid="fb__surveys__headline-text-test"
-            className="label-headline text-base font-semibold">
+          <HeadingTag data-testid="fb__surveys__headline-text-test" className="label-headline text-base">
             {headline}
           </HeadingTag>
         )}
--- a/packages/survey-ui/src/styles/globals.css
+++ b/packages/survey-ui/src/styles/globals.css
@@ -81,7 +81,7 @@
 
   /* ── Headlines ─────────────────────────────────────────────────────── */
   --fb-element-headline-font-family: inherit;
-  --fb-element-headline-font-weight: 500;
+  --fb-element-headline-font-weight: 400;
   --fb-element-headline-font-size: 16px;
   --fb-element-headline-color: #142a72;
   --fb-element-headline-opacity: 1;


apps/web/lib/styling/constants.ts
```diff
@@ -115,7 +115,7 @@ export const STYLE_DEFAULTS: TWorkspaceStyling = {
   // Headlines & Descriptions
   elementHeadlineColor: { light: _colors["elementHeadlineColor.light"] },
   elementHeadlineFontSize: 16,
-  elementHeadlineFontWeight: 500,
+  elementHeadlineFontWeight: 400,
```
kiln-agent, line 118: **Incomplete fix: already-saved styling keeps `500`, and there is no migration.**

This default only reaches styling objects that were never persisted. Both styling forms seed themselves from `STYLE_DEFAULTS` and then persist the *entire* form object, not just dirty fields:

- `apps/web/modules/workspaces/settings/look/components/theme-styling.tsx:66` — `defaultValues: { ...STYLE_DEFAULTS, ...cleanSaved }`, and `onSubmit` saves `styling: data`.
- `apps/web/modules/survey/editor/components/styling-view.tsx:73` / `:132` — same seed, and the `form.watch(...)` subscription writes the whole `data` into `localSurvey.styling`.

So any workspace that ever pressed Save in Look & Feel — and any survey whose Styling tab was mounted and saved — already stores `elementHeadlineFontWeight: 500`, even though the user never touched the Headline Font Weight field. For those, `styles.ts` still emits `--fb-element-headline-font-weight: 500` plus the unlayered `!important` rule, so headlines stay bold after deploy while a brand-new workspace renders 400.

Concretely: the recall/fallback weight mismatch this PR is meant to remove persists for existing installs, and the description's claim that only surveys which *explicitly* set the weight are unaffected doesn't hold — the form persists defaults, so "explicit" and "never configured" are indistinguishable in the stored data. Either migrate the stored value or state in the description that the change applies only to new/never-saved styling.

---

packages/surveys/src/components/general/headline.tsx
```diff
@@ -47,9 +47,7 @@ export function Headline({
             dangerouslySetInnerHTML={{ __html: safeHtml }}
           />
         ) : (
-          <HeadingTag
-            data-testid="fb__surveys__headline-text-test"
-            className="label-headline text-base font-semibold">
+          <HeadingTag data-testid="fb__surveys__headline-text-test" className="label-headline text-base">
```
kiln-agent, line 50: `text-base` on this same className shadows `elementHeadlineFontSize` for exactly the reason `font-semibold` shadowed the weight — so this fixes half of the same defect.

Tailwind utilities are imported as `@import "tailwindcss/utilities.css" layer(utilities) important` (`packages/surveys/src/styles/global.css:3`, `packages/survey-ui/src/styles/globals.css:14`), so `text-base` emits `font-size: 1rem !important` **inside a layer**, which outranks both the unlayered `#fbjs .label-headline { font-size: var(--fb-element-headline-font-size) }` rule and the unlayered `!important` rule `addCustomThemeToDom` injects. That's the cascade already documented in `apps/web/playwright/survey-styling.spec.ts:371-374`.

Repro: set Advanced Styling → Headlines & Descriptions → Headline Font Size to 30px. Element prompts rendered through survey-ui's `ElementHeader`/`Label` grow to 30px, but the welcome card, ending card, and Cal.com booking headline rendered by this component stay at 16px.

Dropping `text-base` from both branches here would make the size setting apply too (the base rule already supplies 16px via `--fb-element-headline-font-size`):

```suggestion
          <HeadingTag data-testid="fb__surveys__headline-text-test" className="label-headline">
```

---

kiln-agent (commented):
Reviewed the 3-line default change (`high` effort). The cascade reasoning behind removing `font-semibold` checks out: Tailwind's important-layered utilities outranked the unlayered `!important` from `addCustomThemeToDom`, so that class was overriding even explicitly configured headline weights on the welcome/ending/Cal.com cards. `STYLE_DEFAULTS` and the `--fb-element-headline-font-weight` fallback are the only two sources for this value, and both were updated consistently (no other CSS copy, no test or doc pins the old `500`).

Two findings, both about the change being incomplete rather than wrong — details inline.

---

kiln-agent (commented):
## Review

The `font-semibold` removal in `headline.tsx` is the load-bearing part of this change and it's correct — that utility was actually beating the injected `!important` var rule (Tailwind utilities are imported with `layer(utilities) important` in both `packages/surveys/src/styles/global.css` and `packages/survey-ui/src/styles/globals.css`, and important declarations reverse layer order), so welcome/ending/Cal.com headlines were ignoring the user's configured weight entirely. Same cascade quirk is documented at `apps/web/playwright/survey-styling.spec.ts:372`. Removing it also fixes the internal inconsistency with the HTML-headline branch right beside it, which never had the class.

One issue with the other two hunks:

### The new `400` default is inert for existing workspaces and surveys

Both styling forms seed `defaultValues` from `STYLE_DEFAULTS` and then persist the **whole** merged form object, not just the fields the user touched:

- `apps/web/modules/workspaces/settings/look/components/theme-styling.tsx:66` — `defaultValues: { ...STYLE_DEFAULTS, ...cleanSaved }`, and `onSubmit` saves `styling: data` verbatim.
- `apps/web/modules/survey/editor/components/styling-view.tsx:73` — `defaultValues: { ...STYLE_DEFAULTS, ...cleanWorkspace, ...cleanSurvey }`, and the `form.watch` subscription merges all of `data` into `localSurvey.styling`.

So any workspace whose Look settings have ever been saved, and any survey whose Styling tab has ever been opened and saved, already has `elementHeadlineFontWeight: 500` stored in the DB — the user never explicitly chose it, it just came along for the ride. `packages/surveys/src/lib/styles.ts:259-260` then emits the `!important` var rule from that stored `500`, and the new CSS fallback at `packages/survey-ui/src/styles/globals.css:84` is never consulted.

Net effect: headlines stay at `500` for most existing installs, so the mixed-weight recall/fallback problem this PR targets isn't actually fixed there. Only brand-new workspaces (Prisma default `{"allowStyleOverwrite":true}`), workspaces that never saved Look settings, and users who hit "Reset to default" get `400`. Welcome/ending/Cal.com headlines do get lighter for everyone — from `600` down to the stored `500` — which is a change but not the intended `400`.

Worth deciding explicitly: if the goal is that existing surveys render at `400`, this needs a data migration normalizing the persisted `500` (or dropping the key where it equals the old default). If shipping only for new/reset workspaces is acceptable, the PR description's "Workspaces/surveys that already set an explicit `elementHeadlineFontWeight` are unaffected" should be corrected — in practice that's most of them, not the ones that deliberately configured it.

### Checked and cleared

- **CSS variable coverage** — `--fb-element-headline-font-weight` is defined in exactly one place, and `packages/surveys` pulls that stylesheet in via `import surveyUiCss from "@formbricks/survey-ui/styles?inline"`. No second stale fallback missed. The file's "Non-color values MUST match STYLE_DEFAULTS" invariant is satisfied by the paired edit.
- **No stale committed build artifacts** — `apps/web/public/js` is gitignored and neither package's `dist/` is tracked, so nothing ships at `500`.
- **No broken assertions** — the tests touching this var set it explicitly (`packages/surveys/src/lib/styles.test.ts:381,423`), the Playwright styling spec only asserts user-set values, and `preview-email-template-styles.test.ts` makes no weight assertions. `getPreviewFontWeight` handles `400`, and the editor's `NumberField` allows `min={100} max={900} step={100}`.
- **No other hardcoded headline bold** in `packages/surveys`, `packages/survey-ui`, or the email templates (the `fontWeight: 500` at `preview-email-template.tsx:375` is a consent label).

### Pre-existing, not from this diff

The HTML-headline branch carries `htmlbody`, whose rule (`packages/surveys/src/styles/global.css:52`) forces `color: var(--fb-subheading-color) !important` and `text-sm` on the headline element — so HTML headlines take the *description* color and 14px for direct text nodes, while `<p>` children get the headline color/size from `.label-headline *`. Not introduced or worsened here, but worth a look if headline styling gets revisited.

---

PR #8963, opened 2026-08-24
fix: correct vertical misalignment of type select in survey variables row

jobenjada:
## What & why

The "Add variable" row in the survey editor's Variables card (`survey-variables-card-item.tsx`) rendered the "Number"/"Text" type dropdown visibly offset upward relative to the surrounding `Field name` and `Initial value` inputs, even though all three controls share the same `h-10` height.

Root cause: Radix's `<Select>` renders a visually-hidden native `<select>` as a second child of its wrapper `<div>` (used for native form/accessibility bubbling). That wrapper is a `FormItem`, which applies Tailwind's `space-y-2` (`margin-bottom` on every child except the last). Because the hidden native select becomes the *last* child, the margin lands on the visible `SelectTrigger` button instead of being absorbed — shifting it up by half the margin relative to its siblings, which have no such hidden sibling.

Fix: set `space-y-0` on that specific `FormItem` (it only ever has one visible child, so the spacing utility was never doing anything useful there anyway).

**Before / after** (isolated repro using the real `Input`/`Select`/`Button`/`FormItem` components and the app's actual Tailwind build — see "How this was tested"):

| Before | After |
| --- | --- |
|  |  |

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Type `Select` trigger vertically aligned with the `Field name`/`Initial value` inputs in the "Add variable" row | manual | Built an isolated Vite+React repro importing the actual production `Input`, `Select`, `Button`, `FormItem`/`FormField`/`FormControl` components (`apps/web/modules/ui/components/*`) plus `react-hook-form`, compiled through the app's real `globals.css` (Tailwind v4 + `@tailwindcss/forms`). Confirmed via `getBoundingClientRect()` that before the fix the trigger's `top`/`bottom` were offset by 4px from the inputs (traced to the `space-y-2` margin landing on the trigger, per the root cause above); after adding `space-y-0` all four controls report identical `top`/`bottom`. Screenshots above captured from this repro with Playwright. |

**Open gaps**

- Not verified against a live survey editor page (no local DB/auth environment available in this session) — only the isolated component repro above. The change is a single Tailwind class on an existing `FormItem`, so risk is low, but worth a quick visual sanity check on staging.

**Risks**

- None expected — `space-y-0` only removes now-unnecessary spacing on a `FormItem` with a single visible child; no effect on the `name`/`value` fields or other `FormItem` usages.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5` (Claude Code), reasoning effort `unknown`.

### DIFF
--- a/apps/web/modules/survey/editor/components/survey-variables-card-item.tsx
+++ b/apps/web/modules/survey/editor/components/survey-variables-card-item.tsx
@@ -202,6 +202,7 @@ export const SurveyVariablesCardItem = ({
               name="type"
               render={({ field }) => (
                 <FormItem
+                  className="space-y-0"
                   onBlur={mode === "edit" ? () => form.handleSubmit(editSurveyVariable)() : undefined}>
                   <Select
                     {...field}


kiln-agent (commented):

I re-derived the root cause from the installed dependencies rather than taking the description on trust, and it holds up.

**Verified**

- `FormItem` is `cn("space-y-2", className)` (`apps/web/modules/ui/components/form/index.tsx:74`), and `cn` is `twMerge(clsx(...))` (`apps/web/lib/cn.ts:4`) — so `space-y-0` genuinely replaces `space-y-2` rather than fighting it on specificity.
- Compiled `space-y-2` with this repo's pinned Tailwind (4.3.1) to check which child actually gets the margin:

  ```css
  .space-y-2 { :where(& > :not(:last-child)) { margin-block-end: calc(var(--spacing) * 2) } }
  ```

  v4 puts `margin-block-end` on every child *except the last* — which is the detail that makes this bug possible. (In v3 it was `margin-top` on later siblings, and the trigger would have been untouched.)
- Radix renders `SelectBubbleInput` (a visually-hidden native `<select>`) as the last child whenever the trigger is inside a `<form>` — `isFormControl = trigger ? form || !!trigger.closest("form") : true` (`node_modules/@radix-ui/react-select/dist/index.js:142`), and it is inside the `<form>` at line 145. So the trigger is not `:last-child` and collects `margin-bottom: 8px`; the bubble input's own margin is inert (`position: absolute`, inline `margin: -1`).
- Result: `FormItem` measures 48px while `Input` (`h-10`, `input/index.tsx:19`), the overridden `SelectTrigger` (`h-10`) and the submit `Button` (`h-10`) are all 40px. Under `items-center` on the row, the 40px trigger sits in the top of a 48px box → the 4px upward offset reported. `space-y-0` removes it; all four boxes become 40px.
- The `name`/`value` `FormItem`s are unaffected, matching the symptom: their single child *is* the last child, so `space-y-*` never applied. Nothing regresses.
- No Playwright spec asserts screenshots or positions on the variables card, so the row getting 8px shorter breaks no test.

**Nits — neither blocks merge**

1. **Same latent margin elsewhere, currently masked.** `page-url-selector.tsx:239` has the identical shape (a `FormItem` whose only visible child is a `Select`), so its trigger also carries the phantom 8px. It just isn't visible because that row uses `items-start` instead of `items-center`. Not this PR's job to fix, but the class of bug will resurface on the next `items-center` row — the durable fix is `FormItem` using `flex flex-col gap-2` instead of `space-y-2`, since `gap` ignores absolutely-positioned and hidden children by construction. That is a repo-wide change and deliberately out of scope here; worth a follow-up ticket rather than growing this diff.

2. **The before/after screenshots are hosted from an unreachable commit.** `3ad20fe` ("chore: PR screenshots…", adding `before.png`/`after.png` at repo root) is on no branch — `before.png` 404s on `main`. `raw.githubusercontent.com` serves unreachable objects only until GitHub prunes them, so the visual evidence can silently vanish from the PR record. Re-uploading them as comment attachments (`user-attachments` URLs) would make them permanent. No harm to the repo either way — nothing from that commit lands on `main`.

**On testing:** I'm not asking for a test here. `AGENTS.md` rules out unit tests for `.tsx`, and a Playwright assertion on bounding-box equality is exactly the brittle, implementation-locking test the guidelines warn against. The stated open gap (not exercised in a live editor) is honest and proportionate for a one-class change; a visual glance on staging is enough.

---

kiln-agent (commented):
## Review — looks correct, no blocking findings

I re-derived the root cause independently rather than taking the description on trust, and it holds up at every link:

- **Tailwind is v4.3.1** (`pnpm-workspace.yaml`), where `space-y-2` compiles to `> :not(:last-child) { margin-block-end }`. The description's "margin-bottom on every child except the last" is therefore right — worth noting because this would have been *wrong* on v3, which put `margin-top` on `~` siblings and would have landed the margin on the hidden native select instead of the trigger. The explanation is version-correct, not accidentally correct.
- **Radix really does render the bubble input as a sibling.** In `@radix-ui/react-select`, `SelectBubbleInput` is emitted after `children` in the root's fragment, gated on `isFormControl` (`trigger.closest("form")`). This row *is* inside a `<form>`, so it renders — and it carries `VISUALLY_HIDDEN_STYLES` (`position: absolute`), so it contributes no height itself but does make the visible `SelectTrigger` a non-last child. Trigger gets 8px bottom margin → `FormItem` measures 48px → in an `items-center` row it sits 4px high. Matches the reported 4px exactly.
- **The fix depends on tailwind-merge, and that dependency is satisfied.** Worth spelling out since it's the one non-obvious part: `.space-y-0 > :not(:last-child)` and `.space-y-2 > :not(:last-child)` have equal specificity, and Tailwind v4 emits `space-y-0` *before* `space-y-2`, so on source order alone `space-y-2` would still win. It works because `FormItem` uses `cn()` → `twMerge`, which drops the loser. Verified against the repo's pinned tailwind-merge 3.5.0: `twMerge("space-y-2", "space-y-0")` → `"space-y-0"`.
- The `FormItem` at line 204 has no `FormLabel`/`FormError` child, so zeroing the spacing removes nothing that was in use. Both `create` and `edit` modes share this row, so both are fixed.

Also checked and clean: breaking-change checkbox correctly unticked (CSS-only, no API/schema/env surface), no new i18n strings, and no unit test is the right call here — per `AGENTS.md` this is a `.tsx` UI concern, and a 4px-alignment assertion is exactly the kind of implementation-detail lock the testing guidelines tell us not to add.

One side note on the screenshots: they resolve against `refs/heads/pr-assets/second-field-height-ui`, a live ref rather than a dangling commit, so the images won't rot and no binaries land on `main`. Just worth keeping that branch around until this merges.

### Non-blocking: the same construct exists one place else

I swept `apps/web` for other `FormItem`s wrapping a `Select` as their only visible child inside a `<form>`. Of the 12 `FormItem`+`Select` blocks repo-wide, exactly one other matches:

`apps/web/modules/ui/components/no-code-action-form/components/page-url-selector.tsx:239` — bare `<FormItem>` (so `space-y-2` is live) wrapping only `FormControl > Select`, inside a `<form>` via both call sites (`create-new-action-tab.tsx` and `ActionSettingsTab.tsx`). The 8px margin *does* land on its trigger. It's invisible today only because the row at line 223 is `flex items-start` rather than `items-center`, so nothing shifts — the row is just 8px taller than intended, adding a stray gap before the next URL-filter row. One alignment keyword away from being the identical visible bug.

I'd leave it out of this PR — it's a different surface and a separate visual check — but it's a reasonable follow-up.

Two things I looked at and am explicitly *not* recommending, so they don't get re-litigated later:

- **Fixing this in `FormItem` itself** (e.g. `space-y-2` → a gap-based flex column) is tempting, but `FormItem` is used across essentially every form in the app and the blast radius dwarfs the bug. The targeted one-liner is the right call for this PR.
- **"Only `FormItem`s lacking a `FormError` are affected"** is a tempting generalization but false, so don't rely on it when auditing: `FormError` returns `null` when there's no error (`form/index.tsx:132`), so it isn't a real last child. It happens not to matter for `Select` cases — the bubble input is last regardless — but it would mislead anyone using "has a `FormError`" as an all-clear.

Nothing here blocks merging. The one open item is the author's own: a visual sanity check on staging, since the row's height drops 8px (the select's `FormItem` was the tallest item in it), which slightly changes the vertical rhythm against the surrounding card.

---

BhagyaAmarasinghe (approved):
LGTM!

PR #8967, opened 2026-08-24
fix(organization): explain why owners cannot leave an organization

jobenjada:
No Linear ticket — drive-by UX fix spotted while looking at the members table.

## What & why

An organization owner opening Settings → Teams → Manage members sees **no** "Leave organization" button: it is gated on `role !== "owner"`, so it simply isn't rendered. Nothing tells them ownership is the reason — the row looks the same as on a plan where leaving isn't supported at all. Members see the button, owners see a blank space.

The button is now rendered for every role when multi-org is enabled, and is disabled with a tooltip for owners. On plans with role management, it reads: _"You cannot leave an organization you own. Ask another owner to change your role first."_ On RBAC-less plans — where every member is an owner and nobody can change anyone's role — it drops the unactionable second sentence: _"You cannot leave an organization you own."_ Both mirror what `leaveOrganizationAction` already throws (`OperationNotAllowedError: You cannot leave an organization you own`).

- Disabled buttons swallow hover events, so the owner variant also carries `pointer-events-none` on the button, and `cursor-not-allowed` on `TooltipRenderer`'s trigger span (the button's own `disabled:cursor-not-allowed` never paints once pointer events are redirected to the wrapper).
- Non-owners are untouched: same enabled button, same dialog, same only-organization guard.
- The button still isn't rendered at all when multi-org is disabled, and the server-side owner check is unchanged — no permission boundary moves.

| Before (owner) | After (owner) |
| --- | --- |
| before | after |

## Breaking changes

None — UI only; no API, SDK, route, env var or config key touched.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Owner: button rendered but inert, with an actionable (or, on RBAC-less plans, honest) reason | manual | Real `OrganizationActions` mounted in a throwaway Vite harness (real `Button`/`TooltipRenderer`, real `en-US` strings and Tailwind tokens), driven by Playwright: `disabled: true`, `opacity: 0.5`, `pointer-events: none`; hover opens a tooltip reading "You cannot leave an organization you own. Assign the owner role to another member first."; a forced click leaves `[role=dialog]` absent. Screenshot above. |
| Owner tooltip copy varies by `isAccessControlAllowed`; `cursor-not-allowed` renders on the trigger span | code review + `pnpm typecheck` + `pnpm lint` | `TooltipRenderer`'s `triggerClass` prop (`apps/web/modules/ui/components/tooltip/index.tsx:41`) is applied to the `<span>` that actually receives the pointer once `pointer-events-none` redirects hit-testing there; the ternary passes each `t(...)` call as a full literal so the i18n key scanner still finds both keys. Not re-captured with the Playwright harness in this pass — see Open gaps. |
| Control — the same harness against the pre-change file | manual | `role=owner` renders **no** leave button at all, so the row above is this diff's doing, not the harness's. |
| Member: nothing changes | manual | Same harness, `role=member`: `disabled: false`, `opacity: 1`, `pointer-events: auto`, `cursor: pointer`. <details><summary>Screenshot</summary></details> |
| `cannot_leave_organization_as_owner` + new `cannot_leave_organization_as_owner_no_access_control` reach all locales | manual | `pnpm i18n` (Lingo) generated the 14 target locales — one changed/inserted line each, no unrelated churn — and `scan-translations` reports "All translation validations passed". |

**Open gaps**

- The updated tooltip copy (both variants) and the `cursor-not-allowed` fix weren't re-captured in the Vite/Playwright harness this pass — verified by code inspection, `pnpm typecheck`, and `pnpm lint` instead. Worth a hover in a real browser during review, in both an RBAC-allowed and an RBAC-less org.
- Not exercised in a running app: the worktree has no database, so the states above are component renders rather than captures of the real settings page. The wiring around them (`membershipRole` → `role`) is unchanged by this diff.
- No E2E spec covers the leave button today, and none was added — this is a `.tsx`-only change, which this repo covers through Playwright rather than component tests, and adding a members-page spec is more than this fix warrants.

**Risks**

- The button now exists in the DOM for owners, so anything reading "button present" as "user can leave" would have to read the disabled state instead. No spec does that today.
- `pointer-events-none` is what makes the tooltip work; if it were ever dropped, the button would go silently back to explaining nothing. Worth a hover in a real browser during review.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5 (Claude Code 2.1.237)`, reasoning effort `medium`.


kiln-agent (commented):
## Review

The gap this fixes is real — an owner currently gets blank space where members get a button, and nothing explains it. The mechanism is right too: `TooltipRenderer` wraps children in a `<span>` trigger, so `pointer-events-none` on the disabled button is genuinely what lets the hover reach it. CI is green across all 19 checks and the `.tsx`-only, no-unit-test shape matches `AGENTS.md`.

One finding I'd want resolved before merge, because it lands in the one string this PR exists to add.

### 1. The tooltip prescribes a step that does not unblock leaving

> "You cannot leave an organization you own. **Assign the owner role to another member first.**"

Assigning owner to another member does not let you leave. The server gate is *being an owner at all*, not being the only one — `leaveOrganizationAction` (`apps/web/modules/organization/settings/teams/actions.ts:498-504`):

```ts
const { isOwner } = getAccessFlags(membership.role);
if (isOwner) {
  throw new OperationNotAllowedError("You cannot leave an organization you own");
}
```

So after following the instruction you are still an owner, `role === "owner"` is still true, the button is still disabled, and the tooltip still shows the same sentence. Dead end with no new guidance.

The cleanest case: an organization that **already** has two owners. The advice is already satisfied for both of them, and both still see a disabled button telling them to do the thing that is already done.

Nor can an owner act on it themselves — `EditMembershipRole` disables your own role select unconditionally (`apps/web/modules/ee/role-management/components/edit-membership-role.tsx:54-58`):

```ts
const disableRole =
  isUserManagementDisabledFromUi ||
  memberId === userId ||        // <- your own row, always
  (memberRole === "owner" && !doesOrgHaveMoreThanOneOwner) ||
  (currentUserRole === "manager" && memberRole === "owner");
```

`members-info.tsx:147-150` passes `memberId={member.userId}` and `userId={currentUserId}`, so that middle clause is always the current user's own row. The actual path out is: **another owner** demotes you (a manager cannot — see the last clause), then you can leave. A sole owner therefore has to promote someone to owner *and then get demoted by them*, which is two steps the string does not hint at.

I suspect the phrasing was borrowed from `ACCOUNT_DELETION_SOLE_OWNER_BLOCK_MESSAGE` (`apps/web/modules/account/constants.ts:8`), which reads almost identically. There it is correct — account deletion is gated on `getOrganizationsWhereUserIsSingleOwner` (`better-auth-account-deletion.ts:43-45`), so adding a second owner really does clear it. Leaving is gated on `isOwner`, so the same sentence stops being true.

Suggested:

```
"cannot_leave_organization_as_owner": "You cannot leave an organization you own. Another owner has to change your role to manager or member first."
```

Note this needs a re-run of `pnpm i18n`.

### 2. The reason is mouse-only

`disabled` takes the button out of the tab order and the `<span>` trigger isn't focusable, so a keyboard or screen-reader user gets "Leave organization, unavailable" and no reason — which is the one thing the PR set out to provide. Not a regression (they previously got nothing at all), but the fix doesn't reach them.

`IconBar` already has the pattern for this — `aria-label={action.tooltip}` on a possibly-disabled button (`apps/web/modules/ui/components/iconbar/index.tsx:39`). Either that, or `aria-disabled` + a no-op `onClick` guard instead of `disabled`, which keeps the button focusable so the tooltip opens on focus.

Worth weighing against the alternative the PR description itself cites as the model: the sibling "only organization" case keeps the button **enabled** and explains itself inside the dialog (`organization-actions.tsx:229-233`). Doing the same here — enabled button, owner message rendered next to `cannot_leave_only_organization` — would match that sibling exactly, reach keyboard and screen-reader users for free, and drop the need for both `pointer-events-none` and the tooltip. The server already rejects the call, so nothing depends on the button being inert. Your call; the current shape is defensible, it just isn't the shape the description claims to mirror.

### 3. `cursor: not-allowed` in the coverage table isn't what the owner sees

The table lists `cursor: not-allowed` as observed evidence of the disabled affordance. `pointer-events: none` removes the button from hit-testing, so the cursor resolves from the wrapping `<span>` instead — the owner sees a plain arrow. `getComputedStyle` still reports `not-allowed` (the `disabled:cursor-not-allowed` rule in `buttonVariants` matches regardless of pointer-events), so a computed-style assertion passes while the rendered cursor never changes. The assertion holds for the wrong reason; the affordance it stands in for isn't there.

Minor, and the tooltip arguably carries the signal on its own. Flagging it because the coverage table is load-bearing under the PR policy.

### Nits

- `isOrganizationOwner = role === "owner"` (line 78) duplicates `isOwner` from `getAccessFlags(membershipRole)` two lines up, and `members-view.tsx:69-70` passes the same value to both `role` and `membershipRole`. Reusing `isOwner` keeps one source of truth.
- Register drift in the generated locales: the new line is informal where its immediate neighbour is formal — `fr-FR` "Tu ne peux pas" next to "Vous ne pouvez pas", `nl-NL` "Je kunt" next to "U kunt", `pt-PT` "Não podes" next to "Não pode"; `hu-HU` also adds a "Kérem" the English doesn't have. Adjacent lines in the same block, so it's visible. Not hand-fixable per `AGENTS.md` — if it's worth addressing it's a Lingo formality/glossary setting, not an edit to those files.
- Beyond this diff: if `pointer-events-none` is what makes a tooltip work over a disabled button, then `editor-card-menu.tsx`, `block-menu.tsx` and `IconBar` have the same silent failure for their disabled states — none of them carries it. This PR has the right fix; the others may be worth a follow-up.

### Checked and fine

- Server-side owner check untouched, and the "no permission boundary moves" claim holds — the button is presentation only.
- Non-owner path unchanged: `shouldRender=false` returns children bare, so same button, same dialog, same only-organization guard.
- Still gated behind `isMultiOrgEnabled`, so the free plan (where every member is an owner) never renders it.
- `i18n.lock` has exactly the one expected checksum line; `Validate Translation Keys` green.
- `billing` role can't reach this page (`redirectBillingRoleFromRestrictedOrgSettings`), so the `role` union's third value isn't a case here.
- The `<span>` trigger blockifies as a flex item at the button's own size — no layout shift versus the non-owner path.

---

kiln-agent (commented):
## Review — `fix(organization): explain why owners cannot leave an organization`

The gap this fixes is real: an owner currently gets a blank space where the button should be, with nothing saying why. All 19 checks are green, the i18n key is placed correctly and the lock updated, only `en-US.json` was hand-edited, and the before/after screenshots resolve (`assets/leave-organization-owner-tooltip`). Two things I'd want changed before merge, plus two nits.

---

### 1. The tooltip tells the owner to do something that doesn't unblock them

The message ends with _"Assign the owner role to another member first."_ Following it changes nothing:

- The server rejects **any** owner, not the sole owner — `teams/actions.ts:502`, `if (isOwner) throw new OperationNotAllowedError("You cannot leave an organization you own")`. Adding a second owner leaves your own `isOwner` true.
- Promoting someone else doesn't demote you, and you cannot demote yourself: `edit-membership-role.tsx:54-58` has `disableRole = … || memberId === userId || …`, unconditional for your own row. `EditMembershipRole` is the only role-change surface in the app (sole consumer is `members-info.tsx:146`).

So the owner promotes a colleague, returns to this page, and finds the same disabled button with the same message. The only thing that actually unblocks them is **another owner changing their role** — something the string doesn't mention and they can't do themselves.

I think the phrasing was borrowed from `warning_cannot_delete_account` ("You are the only owner of this organization. Please transfer ownership to another member first."). That one sits behind a genuine *sole-owner* rule (`better-auth-account-deletion.ts:43`, `getOrganizationsWhereUserIsSingleOwner`), where adding a second owner really does unblock. The leave rule isn't the sole-owner rule, so the advice doesn't carry over.

Suggestion: name the action that works — e.g. _"You cannot leave an organization you own. Another owner has to change your role first."_ (Whether the server rule *should* be sole-owner-only, matching account deletion, looks like a fair question, but a much bigger one and not this PR's.)

Secondary: the role column only renders when `isAccessControlAllowed` (`members-info.tsx:139`). That's a per-org plan permission (`getCustomPlanFeaturePermission(orgId, "accessControl")`), while multi-org is a license feature flag (`getSpecificFeatureFlag("isMultiOrgEnabled")`) — independent mechanisms, so multi-org-on/access-control-off is reachable, and there the current wording points at a UI that isn't on the page at all.

### 2. `pointer-events-none`: the stated reason doesn't apply here, and it silently drops the `not-allowed` cursor

The comment says a disabled button swallows hover events. That's true when the listener is on the button — but `TooltipRenderer` puts `asChild` on a wrapper `<span>` (`ui/components/tooltip/index.tsx:51-53`), so the trigger is the span, and the button is its child.

I checked this in Chromium (standalone page, disabled button inside a span with listeners on the span). The wrapper receives `pointerenter` **and** `pointermove` — the events Radix's Trigger opens on — in both the plain-`disabled` case and the `pointer-events: none` case. The class isn't what makes the tooltip work.

The repo agrees: three existing TooltipRenderer-over-disabled-Button call sites have no such class — `delete-team.tsx:52-64`, `archive-feedback-directory.tsx:63-75`, `purge-feedback-directory-data.tsx:99-111`. If those tooltips work, this one works without it; if they don't, the fix belongs in `TooltipRenderer` (or via its existing `triggerClass`) for all four sites rather than one-off at a call site.

It isn't free, either. `pointer-events: none` makes the span the hit-test target instead of the button, so the Button variant's `disabled:cursor-not-allowed` (`ui/components/button/index.tsx:8`) never applies — the owner hovers and gets a plain arrow. `document.elementFromPoint` at the button's centre returns the *wrapper*, and the effective cursor is the wrapper's (`auto`), where the plain-disabled control returns the button itself.

Worth noting because the PR's evidence can't detect this: `cursor: not-allowed` read from the button's computed style stays `not-allowed` whether or not the class is there — computed style doesn't know about hit-testing. Likewise _"a forced click leaves `[role=dialog]` absent"_ passes on the `pointer-events-none` rule alone, so it would still pass if `disabled` were accidentally dropped. Neither assertion can fail in the way that matters.

Ask: drop the class and the comment, and hover once in a real browser to confirm. If a non-Chromium engine genuinely needs it, put it in `TooltipRenderer` so the three siblings get it too.

### 3. `isOrganizationOwner` restates `isOwner` two lines up — nit

`isOwner` already exists at line 76 from `getAccessFlags(membershipRole)`; line 78 adds `isOrganizationOwner = role === "owner"`. The only caller passes the same value to both props (`members-view.tsx:69-70`, `membershipRole={membershipRole} role={membershipRole}`). One fact, two names, derived from two different props — if they ever diverge the component renders contradictory state. Either use `isOwner`, or retire the duplicate prop.

### 4. Keyboard and screen-reader users still get no explanation — nit, pre-existing

`disabled` takes the button out of the tab order, and the wrapper span has no `tabIndex`, so the tooltip is pointer-only. For keyboard and AT users the owner row is still the blank, unexplained state this PR sets out to fix. Shared with the three sibling call sites, so a `TooltipRenderer`-level follow-up (focusable trigger + `aria-disabled` instead of `disabled`) rather than something to solve here.

---

**FYI, not actionable:** the generated locales drift in register from their neighbours — fr-FR uses "Tu" where the adjacent lines use "Vous", nl-NL "Je" against "U", pt-PT "tu" against "pode". Per AGENTS.md those files can't be hand-edited, so this is only a note in case it's worth raising with Lingo config.

---

BhagyaAmarasinghe (changes requested):
Small, well-scoped diff, and the mechanism is right: `TooltipRenderer` wraps children in a `<span>` trigger (`apps/web/modules/ui/components/tooltip/index.tsx:57`), so `pointer-events-none` on the disabled button correctly lets the hover land on the span — the comment on `organization-actions.tsx:186` is accurate and worth keeping. `Button` carries `disabled:opacity-50 disabled:cursor-not-allowed` but no `disabled:pointer-events-none`, so the class is genuinely load-bearing. i18n is clean: key alphabetically placed, all 14 locales + `i18n.lock` updated with exactly one line each. No spec asserts on this button, so nothing breaks. Breaking-changes box correctly unticked.

Requesting changes on one of these — the first.

### 1. The tooltip's remedy doesn't actually unblock leaving

> Assign the owner role to another member first.

This implies that doing so lets you leave. It doesn't. `leaveOrganizationAction` checks *your own* membership role (`apps/web/modules/organization/settings/teams/actions.ts:500`), which promoting somebody else does not change. And you can't fix that yourself: `apps/web/modules/ee/role-management/components/edit-membership-role.tsx:54` disables the role dropdown when `memberId === userId`, so **another** owner has to demote you. An owner who follows the tooltip literally lands back on the same disabled button with the same tooltip.

It gets worse where access control isn't licensed. The Role column isn't rendered at all (`apps/web/modules/organization/settings/teams/components/edit-memberships/members-info.tsx:139`) and `updateMembershipAction` throws `Role management is not allowed for this organization` — and those are exactly the orgs where *every* member is an owner by construction (`individual-invite-tab.tsx:67`, `bulk-invite-tab.tsx:136`). So on a multi-org-enabled, RBAC-less plan the new button appears for **every member**, each advising a step none of them can take. Before this diff they saw nothing; after it they see instructions that dead-end.

Either drop the second sentence, or make it true — something along the lines of "Ask another owner to change your role first." `isAccessControlAllowed` is already a prop on this component if you'd rather vary the copy by plan.

### 2. `cursor-not-allowed` never actually renders

`pointer-events-none` takes the button out of hit-testing, so the cursor the owner sees comes from the parent `<span>` — a plain arrow. The Coverage table cites `cursor: not-allowed`, which is true of the computed style but not of what's on screen. `triggerClass="cursor-not-allowed"` on the `TooltipRenderer` fixes it, and is worth doing since the disabled affordance is the whole point here.

### 3. Keyboard and screen-reader users get no explanation

`disabled` removes the button from the tab order, and Radix's `asChild` trigger is a bare `<span>` with no `tabIndex`, so the tooltip only ever opens on hover — the reason stays invisible to precisely the users who can't hover. `aria-disabled` plus an early return in `onClick` (keeping the button focusable) would close it. `apps/web/modules/ee/teams/team-list/components/team-settings/workspace-row.tsx:88` has the same limitation today, so this is fair as a follow-up rather than something to fix here.

### 4. Nit — two sources of truth for one fact

`isOrganizationOwner = role === "owner"` (`organization-actions.tsx:78`) duplicates `isOwner` from `getAccessFlags(membershipRole)` two lines above, and the only caller passes the same value to both props (`members-view.tsx:69-70`). Reusing `isOwner` drops the new local.

---

Everything else checks out: prettier passes, import order is correct, and the `role !== "owner"` → `isMultiOrgEnabled` gate change is safe (`role` has no other use in the component). No permission boundary moves.

jobenjada:
Thanks — fixed all three in f738fa5f0a:

1. **Tooltip remedy** — now conditional on `isAccessControlAllowed`: "Ask another owner to change your role first." on RBAC-enabled plans, and just "You cannot leave an organization you own." (no false remedy) on RBAC-less ones, where role management doesn't exist at all.
2. **`cursor-not-allowed`** — moved to `TooltipRenderer`'s `triggerClass` prop, so it actually lands on the span doing the hit-testing.
3. **Duplicate source of truth** — dropped `isOrganizationOwner` and the `role` prop it depended on; the component now reuses `isOwner` from `getAccessFlags(membershipRole)` directly, and `members-view.tsx` no longer passes the redundant `role={membershipRole}`.

Left #3 (keyboard/SR discoverability) as the follow-up you suggested — `pnpm typecheck`, `pnpm lint`, and `pnpm i18n` all pass; PR description updated to match.

PR #8968, opened 2026-08-24
fix(survey-editor): say "window" instead of "modal" and stop backdrop clicks closing the translations modal

jobenjada:
Ref ENG-2600

## What & why

Two small things in the survey editor's **Manage translations** modal:

- **Copy.** The AI-translation toast said _"Translating with AI… Please keep this modal open."_
  "Modal" is our component vocabulary, not the user's — it now says **window**. `en-US.json` is the
  only hand-edited locale; the 14 others were regenerated from it with `pnpm i18n`, and `i18n.lock`
  carries the new hash for the key so `scan-translations` stays in sync.
- **Dismiss behaviour.** The modal holds unsaved drafts — `draftTranslations` lives in component
  state and only `handleSave` writes it into `localSurvey`
  ([manage-translations-modal.tsx:199](https://github.com/formbricks/formbricks/blob/f96ccf5e4c4035b38a2985ac38042ab7b17e7277/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx#L199))
  — so a stray click on the backdrop threw away everything you had typed or translated.
  `disableCloseOnOutsideClick` stops that.

Only 4 of the 14 non-English locales actually said "modal"-ish (`es-ES`, `pt-BR`, `hu-HU`
"párbeszédablak", `sv-SE` "dialogruta"); the other 10 already named a window and were restyled by the
regeneration anyway. See **Risks** for the drift that introduced.

This does **not** fix ENG-2600 (closing the modal mid-run reports success over drafts that are then
discarded) — it only removes one way of triggering it. The root fix is still open.

## Breaking changes

None — user-facing copy plus one modal's dismiss behaviour. No API or SDK shape, emitted value,
endpoint, route, env var, config key, default or webhook payload changes, and nothing to migrate.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| en-US toast no longer says "modal" | manual | `ai_translating` was the only en-US **value** containing the word; after the change the sole remaining hit is the key `common.centered_modal`, whose value "Centered Modal" is a survey-placement label. The key has exactly one call site, `manage-translations-modal.tsx:163` |
| Every locale carries the new wording | manual | All 15 locale files changed by exactly one line, the `ai_translating` value, and each now names a window (`ventana` / `janela` / `ablak` / `fönster` / `ウィンドウ` / `窗口` / …). No other key moved in any file |
| Backdrop click no longer dismisses the modal | manual | `disableCloseOnOutsideClick` maps to `onPointerDownOutside → preventDefault` in the shared `DialogContent` ([dialog/index.tsx:86](https://github.com/formbricks/formbricks/blob/f96ccf5e4c4035b38a2985ac38042ab7b17e7277/apps/web/modules/ui/components/dialog/index.tsx#L86)), the same wiring ~15 other modals already rely on. Verified by reading that prop path, not by clicking it in a running app — see **Open gaps** |
| Escape no longer dismisses it either | manual | Same flag also gates `onEscapeKeyDown` unless `closeOnEscape` is passed ([dialog/index.tsx:87-89](https://github.com/formbricks/formbricks/blob/f96ccf5e4c4035b38a2985ac38042ab7b17e7277/apps/web/modules/ui/components/dialog/index.tsx#L87-L89)), and this call site does not pass it. The remaining exits are the X button, Cancel and Save. This is a side effect of the flag, not a decision — see **Open gaps** |

No new test was added: the copy change is data, and the dismiss behaviour is one prop on a shared
`DialogContent` that already has a Storybook story for it (`dialog/stories.tsx:425`). Per AGENTS.md
this is UI detail inside a feature, so it does not earn an e2e spec of its own.

**Open gaps**

- **The Escape suppression is unintended.** `disableCloseOnOutsideClick` is overloaded in this repo
  and kills Escape too. CodeRabbit and the automated review both flagged it. Needs a call before
  merge: either pass `closeOnEscape`, or gate the flag on `isTranslating` the way
  `create-with-ai-dialog.tsx:83` does. Not resolved in this diff.
- **Not clicked through in the running app.** This modal needs an AI-enabled workspace and a live AI
  provider, so both dismiss rows above are code-level verification against the shared Dialog wiring.
  Worth a look on staging.
- **The non-English wording is machine output.** No native-speaker review of the 10 locales that were
  restyled without needing the fix.
- ENG-2600 stays open; this PR narrows how often it is hit, it does not fix it.

**Risks**

- Anyone used to dismissing this modal with a backdrop click or Escape now has to use the X button or
  Cancel. If that is reported, add `closeOnEscape` at the call site.
- `hu-HU` now says "AI fordítás" where its siblings on the same screen say "MI" (`ai_translate` =
  "Fordítás MI-vel", `ai_translation_complete` = "Az MI-fordítás befejeződött"). Terminology drift
  inside one dialog.
- Politeness register churned in both directions: `ru-RU` went formal ("не закрывайте") while its
  siblings stay informal ("Свяжись с администратором"), `tr-TR` went the other way ("tutun" → "tut").
  A by-product of the regeneration, not a decision.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code 2.1.201), reasoning effort `unknown` (not exposed
> in this non-interactive session).

### DIFF
--- a/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
+++ b/apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
@@ -219,7 +219,7 @@ export const ManageTranslationsModal = ({
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogContent width="wide" className="max-h-[85dvh]">
+      <DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick>
         <DialogHeader>
           <DialogTitle>{t("workspace.surveys.edit.manage_translations")}</DialogTitle>
           <div className="mt-2 flex items-center justify-between">
--- a/apps/web/i18n.lock
+++ b/apps/web/i18n.lock
@@ -2939,7 +2939,7 @@ checksums:
     workspace/surveys/edit/ai_instance_not_configured: 939ad7c3240fa8de98a325239f1b36bc
     workspace/surveys/edit/ai_smart_tools_disabled: 13df84ae47d35dfa6e86ffa62f29c75d
     workspace/surveys/edit/ai_translate: f25943cdeffe155ee524428f4daa5da2
-    workspace/surveys/edit/ai_translating: 098a2293b39f9f258d67f926cf03df37
+    workspace/surveys/edit/ai_translating: 5f60cc47ce106b98251c46197e4981c0
     workspace/surveys/edit/ai_translation_all_fields_populated: d78f6a663ea19ce77045970179bd200f
     workspace/surveys/edit/ai_translation_complete: f443d0801404f728e68000b46ca67598
     workspace/surveys/edit/ai_translation_failed: fd356a173d0abde7a0fc660394954cc7
--- a/apps/web/locales/de-DE.json
+++ b/apps/web/locales/de-DE.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "KI ist nicht konfiguriert. Kontaktiere deinen Administrator.",
         "ai_smart_tools_disabled": "KI-Smart-Tools sind für diese Organisation deaktiviert.",
         "ai_translate": "Mit KI übersetzen",
-        "ai_translating": "Übersetze mit KI... Bitte lasse dieses Fenster geöffnet.",
+        "ai_translating": "Übersetze mit KI... Bitte lass dieses Fenster geöffnet.",
         "ai_translation_all_fields_populated": "Alle Felder sind bereits übersetzt",
         "ai_translation_complete": "KI-Übersetzung abgeschlossen",
         "ai_translation_failed": "Übersetzung fehlgeschlagen",
--- a/apps/web/locales/en-US.json
+++ b/apps/web/locales/en-US.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "AI is not configured. Contact your administrator.",
         "ai_smart_tools_disabled": "AI smart tools are disabled for this organization.",
         "ai_translate": "Translate with AI",
-        "ai_translating": "Translating with AI... Please keep this modal open.",
+        "ai_translating": "Translating with AI... Please keep this window open.",
         "ai_translation_all_fields_populated": "All fields are already translated",
         "ai_translation_complete": "AI translation complete",
         "ai_translation_failed": "Translation failed",
--- a/apps/web/locales/es-ES.json
+++ b/apps/web/locales/es-ES.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "La IA no está configurada. Contacta con tu administrador.",
         "ai_smart_tools_disabled": "Las herramientas inteligentes de IA están deshabilitadas para esta organización.",
         "ai_translate": "Traducir con IA",
-        "ai_translating": "Traduciendo con IA... Por favor, mantén este modal abierto.",
+        "ai_translating": "Traduciendo con IA... Mantén esta ventana abierta.",
         "ai_translation_all_fields_populated": "Todos los campos ya están traducidos",
         "ai_translation_complete": "Traducción con IA completada",
         "ai_translation_failed": "La traducción ha fallado",
--- a/apps/web/locales/fr-FR.json
+++ b/apps/web/locales/fr-FR.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "L'IA n'est pas configurée. Contacte ton administrateur.",
         "ai_smart_tools_disabled": "Les outils intelligents IA sont désactivés pour cette organisation.",
         "ai_translate": "Traduire avec l'IA",
-        "ai_translating": "Traduction en cours avec l'IA... Garde cette fenêtre ouverte.",
+        "ai_translating": "Traduction avec l'IA en cours... Merci de garder cette fenêtre ouverte.",
         "ai_translation_all_fields_populated": "Tous les champs sont déjà traduits",
         "ai_translation_complete": "Traduction IA terminée",
         "ai_translation_failed": "La traduction a échoué",
--- a/apps/web/locales/hu-HU.json
+++ b/apps/web/locales/hu-HU.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "Az MI nincs beállítva. Vegye fel a kapcsolatot az adminisztrátorral.",
         "ai_smart_tools_disabled": "Az MI intelligens eszközei le vannak tiltva ennél a szervezetnél.",
         "ai_translate": "Fordítás MI-vel",
-        "ai_translating": "Fordítás MI-vel… Tartsa nyitva ezt a párbeszédablakot.",
+        "ai_translating": "AI fordítás folyamatban... Kérem, tartsa nyitva ezt az ablakot.",
         "ai_translation_all_fields_populated": "Az összes mező le van már fordítva",
         "ai_translation_complete": "Az MI-fordítás befejeződött",
         "ai_translation_failed": "A fordítás nem sikerült",
--- a/apps/web/locales/ja-JP.json
+++ b/apps/web/locales/ja-JP.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "AIが設定されていません。管理者にお問い合わせください。",
         "ai_smart_tools_disabled": "この組織ではAIスマートツールが無効になっています。",
         "ai_translate": "AIで翻訳",
-        "ai_translating": "AIで翻訳中... このモーダルを開いたままにしてください。",
+        "ai_translating": "AIで翻訳中...このウィンドウを開いたままにしてください。",
         "ai_translation_all_fields_populated": "すべてのフィールドは既に翻訳されています",
         "ai_translation_complete": "AI翻訳が完了しました",
         "ai_translation_failed": "翻訳に失敗しました",
--- a/apps/web/locales/nl-NL.json
+++ b/apps/web/locales/nl-NL.json
@@ -3056,7 +3056,7 @@
         "ai_instance_not_configured": "AI is niet geconfigureerd. Neem contact op met je beheerder.",
         "ai_smart_tools_disabled": "AI slimme tools zijn uitgeschakeld voor deze organisatie.",
         "ai_translate": "Vertalen met AI",
-        "ai_translating": "Vertalen met AI... Laat dit venster open.",
+        "ai_translating": "Vertalen met AI... Houd dit venster open.",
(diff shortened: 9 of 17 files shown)


apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
```diff
@@ -219,7 +219,7 @@ export const ManageTranslationsModal = ({
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogContent width="wide" className="max-h-[85dvh]">
+      <DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick>
```
kiln-agent, line 222: `disableCloseOnOutsideClick` also suppresses Escape in `DialogContent` — the prop wires up both `onPointerDownOutside` **and** `onEscapeKeyDown` (`apps/web/modules/ui/components/dialog/index.tsx:86-89`), and Escape only survives when `closeOnEscape` is passed alongside it. So this change goes beyond the stated intent ("prevents accidental modal closing through outside click"): a user who opens Manage translations, changes nothing, and presses Escape (no translation running) now stays stuck in the dialog and has to find the X or Cancel — a regression against the standard dialog dismissal affordance (WAI-ARIA dialog pattern requires Escape). `create-chart-view.tsx:137` is the precedent for keeping it.

```suggestion
      <DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick closeOnEscape>
```

---

kiln-agent, line 222: The guard only covers outside clicks, so the data-loss path the toast warns about is still open: while `isTranslating` is true the header X button (`DialogPrimitive.Close`) and the footer Cancel button (line 323) both still close the dialog. Concrete scenario — start "Translate with AI" on a 40-field survey, click the X while the request is in flight; the component stays mounted (`language-view.tsx:552` renders it unconditionally), the action resolves and writes into `draftTranslations`, but nothing is ever saved, and reopening the modal runs the `open` effect (lines 72-80) which re-initializes drafts from `localSurvey` and discards every translation the user just paid for.

Consider `disabled={isTranslating}` on the Cancel button plus `hideCloseButton={isTranslating}` (or gating `setOpen` in `onOpenChange` on `isTranslating`) so the "keep this window open" instruction is actually enforceable.

---

kiln-agent (commented):
Reviewed the wording change plus the new `disableCloseOnOutsideClick`. The locale updates (15 locales + `en-US` + `i18n.lock` checksum) are consistent and no key is missing. Two notes on the dialog change, both inline.

---

apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
kiln-agent: Correction: the `suggestion` block above rendered empty — do **not** commit it, it would delete the opening tag. The intended patch is:

```suggestion
      <DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick closeOnEscape>
```

---

kiln-agent, line 222: This also disables Escape-to-close, which is broader than the stated intent ("prevents accidental modal closing through outside click").

In this repo's `DialogContent`, the flag is overloaded — `apps/web/modules/ui/components/dialog/index.tsx:86-89`:

```tsx
onPointerDownOutside={disableCloseOnOutsideClick ? (e) => e.preventDefault() : undefined}
onEscapeKeyDown={
  disableCloseOnOutsideClick && !closeOnEscape ? (e) => e.preventDefault() : undefined
}
```

So as written, the only way out of this dialog is the X button or Cancel. Two options depending on what you actually want:

1. **Outside click only** (matches the PR description) — add `closeOnEscape`. That prop exists for exactly this case, and `create-chart-view.tsx:136` pairs them.
2. **Block both, but only while it matters** — gate on the in-flight state, as `create-with-ai-dialog.tsx:83` does with `disableCloseOnOutsideClick={isBusy}`:

```tsx
<DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick={isTranslating}>
```

I'd lean toward (1) or (2) rather than the unconditional form. Escape is the expected keyboard dismissal for a modal dialog, and the hazard the `ai_translating` toast warns about only exists during an in-flight translation — outside of that, the dialog holds unsaved drafts like any other editor dialog here (`edit-segment-modal`, `webhook-detail-modal`), where blocking outside click alone is the established pattern.

Worth noting either way: X and Cancel stay live during translation, so the accidental-close path isn't fully closed. If that's the real goal, disabling those while `isTranslating` would be the complete fix.


---

kiln-agent (commented):
## Review

Small, sensible change — the `en-US` wording fix is correct and the locale regeneration looks like a clean `pnpm i18n` run. One substantive point on the dialog prop, plus two smaller notes.

### 1. `disableCloseOnOutsideClick` also kills Escape (inline comment)

Left inline on `manage-translations-modal.tsx:222`. Short version: in this repo the flag is overloaded and suppresses `onEscapeKeyDown` too unless `closeOnEscape` is passed, so the change is broader than the PR description claims. Either add `closeOnEscape`, or gate the flag on `isTranslating` the way `create-with-ai-dialog.tsx` does. Not a blocker, but I think it's the wrong default as written.

### 2. hu-HU terminology drift: `MI` → `AI`

The regenerated Hungarian string switched the term for "AI", and it's now the odd one out among its own siblings:

```
ai_translate      => Fordítás MI-vel
ai_translating    => AI fordítás folyamatban... Kérem, tartsa nyitva ezt az ablakot.   ← new
ai_translation_complete => Az MI-fordítás befejeződött
ai_features_not_enabled => Az MI-funkciók nincsenek engedélyezve ennél a szervezetnél.
```

Every other `ai_*` key in `workspace.surveys.edit` uses `MI` (27 `MI` vs 17 `AI` across the whole file), and this key itself used `Fordítás MI-vel…` before. So the same modal will now say both. Per AGENTS.md the non-English files are machine-generated and shouldn't be hand-edited — the fix would be a Lingo.dev glossary entry pinning `AI → MI` for hu-HU. Low severity and not this PR's fault, but it regressed here so flagging it.

The rest of the locale churn is fine — the reworded-but-already-correct ones (de, fr, nl, pt-PT, ro, tr, zh) are expected re-translation noise now that the source string changed, and tr-TR actually moved *toward* the file's informal register. ru-RU went the other way (`не закрывай` → `не закрывайте`, now formal next to the informal `Свяжись с администратором`), same glossary/tone-config territory.

### 3. PR description doesn't use the template

AGENTS.md treats `.github/pull_request_template.md` as mandatory and the source of truth for PR structure. The description is currently a single sentence, missing the ticket line, the `## Breaking changes` checkbox (the only input to the `breaking-change` label), `## Migrations & env`, and the `How this was tested` coverage table. Given the diff, all of those are quick to fill: no breaking change, no migrations, and the behaviour to cover is the dialog dismissal — which, worth noting, has no automated coverage today. The Playwright helpers (`fillModalTranslation` / `fillModalRichTranslation`, `apps/web/playwright/utils/helper.ts:449-465`) drive this modal by testid and never press Escape, so nothing existing breaks, but nothing verifies the new behaviour either. A `manual` row is fine.

Minor: the description quotes the strings as `"Please don't close modal"` → `"Please don't close window"`; the actual change is `Please keep this modal open.` → `Please keep this window open.`

### Verified clean

- `disableCloseOnOutsideClick` is a real `DialogContent` prop, correctly spelled, widely used elsewhere — no type/API mismatch.
- All 15 locale files updated; none left referring to a modal/dialog. `ai_translating` checksum in `i18n.lock` bumped consistently with the `en-US.json` edit.
- `Validate Translation Keys`, `PR title`, `check-pr-size` and `Validate Helm Chart` are green; the remaining checks were still running at review time.
- New line is ~82 chars, within the 110-char Prettier width.


---

apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx
kiln-agent, line 222: Two things worth considering here.

**1. This also disables Escape, not just outside-click.** In `modules/ui/components/dialog/index.tsx:87-89`, `disableCloseOnOutsideClick` suppresses `onEscapeKeyDown` too unless `closeOnEscape` is passed (the prop is documented at line 37 precisely for this). So the diff silently removes keyboard dismissal on top of the intended change. `create-chart-view.tsx:136-137` is the precedent for keeping it:

```tsx
<DialogContent width="wide" className="max-h-[85dvh]" disableCloseOnOutsideClick closeOnEscape>
```

If losing Escape is deliberate (protecting unsaved drafts), fine — but it's worth stating in the PR body, since it's an a11y-visible behaviour change that the title doesn't cover.

**2. The guard doesn't actually close the data-loss path the copy warns about.** During `isTranslating`, the X button and the Cancel button both stay active. Closing mid-flight leaves the in-flight promise running against a still-mounted component, so on reopen the init effect (lines 72-80) rebuilds `draftTranslations` from `localSurvey` and the AI results are wiped — while an "AI translation complete" toast still fires. Outside-click and Escape were the two least likely ways for a user to lose that work; the two buttons that are most likely are untouched.

If the intent is to protect the translation run rather than the modal generally, gating on `isTranslating` covers it — `create-with-ai-dialog.tsx:83` does exactly that with `disableCloseOnOutsideClick={isBusy}`, and pairing that with `hideCloseButton={isTranslating}` plus a disabled Cancel would make the "keep this window open" copy true.

---

kiln-agent (commented):
Reviewed the full diff: one line of real code, the `ai_translating` copy change in `en-US.json`, its Lingo-regenerated values across 14 locales, and the matching `i18n.lock` checksum. All locale files parse, no key-set changes, and nothing in the unit or E2E suites depends on the old copy or on Escape-to-close. The copy change itself reads well — "window" is the right call over "modal".

Two notes on the `disableCloseOnOutsideClick` line, left inline: it also turns off Escape as a side effect (`closeOnEscape` exists for that), and the close/cancel buttons stay live during translation, so the accidental-loss path the toast warns about is still open.

**Nit (non-blocking, not a hand-edit):** the regenerated `hu-HU.json:3059` switched the product term from "MI" to "AI", so the toast now reads "AI fordítás folyamatban..." while the button right above it is labelled "Fordítás MI-vel" and the sibling keys keep "MI" (`ai_translate`, `ai_translation_complete`). Per `AGENTS.md` the non-English locales aren't edited by hand, so if this is worth correcting it belongs in the Lingo glossary rather than in this PR.

Nothing here is blocking — the change does what the title says.

---

BhagyaAmarasinghe (approved):
LGTM!

PR #8969, opened 2026-08-24
fix(sso): strip the live local auth factors on SSO recovery (backport 5.4)

xernobyl:
Ref ENG-2557

## What & why

Backport of #8962 to `release/5.4`. SSO verify-before-link recovery is supposed to strip a recovered account's local auth factors before the SSO identity becomes the way back in, but the strip was written before ENG-1054 moved the credential to `Account.password` and 2FA into the `TwoFactor` table and was never updated — so post-cutover it nulled three already-null legacy `User` columns and stripped nothing, while reading in code as though it did. An attacker who set a password on an unverified account keeps it across the victim's recovery. Root cause, blame, blast radius and the full evidence are on #8962 and the ticket.

`release/5.4` needs it on its own merits rather than as a routine follow-the-main backport: **5.4 is not released yet** (`5.4.0-rc.1` is the latest tag on this line), so landing here is what keeps the vulnerability out of 5.4.0 GA instead of shipping it and fixing it afterwards. The affected range is 5.2.0–5.3.4.

**No cherry-pick drift.** `release/5.4`'s tip is `ab8fc21ff` (#8835, Better Auth 1.7) and all five source files this fix touches are byte-identical between `release/5.4` and `main`, so the eight commits applied with zero conflicts and the result is **byte-identical to the reviewed head of #8962** (`1473fceb8`) across all twelve files — verified with `git diff 1473fceb8 HEAD -- <the twelve paths>`, empty. The only tree difference against that head is `b66c1dd97` (Johannes's accessibility docs), which is on `main`, not on this line, and is unrelated.

This also lands natively rather than adapted: the fix targets the Better Auth 1.7 account model, and `ab8fc21ff` is exactly the commit that brought 1.7 to this branch. `createLocalAccountIssuer` — which the last commit uses to match Better Auth's own credential-account predicate — is an upstream `@better-auth/core/db` export already in use on `release/5.4`.

**Independent of the other SSO work in flight.** #8952 (the ENG-2555 5.4 backport) touches a disjoint file set — `account-linking.ts`, `constants.ts` and a data migration — so the two can land in either order without conflicting.

Per the backport policy: one fix, minimal diff identical to the reviewed PR, and urgent — it is an account-takeover chain.

## Breaking changes

None — no API/SDK shape, endpoint, route, env var, config key or default changes, and no deployer action. The behaviour change is the security fix itself.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Everything | see #8962 | This branch is byte-identical to #8962's reviewed head across all twelve touched files (`git diff 1473fceb8 HEAD -- <paths>` → empty; base files identical between `release/5.4` and `main`), so its full coverage table applies verbatim: the live before/after through real HTTP (**200 → 401**, `Account.password` present → null), the real-Postgres/real-Redis integration suite asserting through `auth.api.*`, and fourteen mutations each proven to turn the suite red |

The equivalence is the argument here rather than a re-run: re-testing identical bytes on an identical base re-proves nothing that #8962 has not already proven, and #8962's own suites passed 18/18 on `1473fceb8`. What is genuinely unshared is this branch's CI, below.

**Open gaps**

- **CI on this PR is the only thing not shared with #8962** — it runs against the 5.4 base rather than main. Wait for it before merging; that is the one check the equivalence argument cannot stand in for.
- Everything else is as #8962: no live IdP handshake (the recovery tokens are minted directly — ENG-2445 covers the handshake), the forgot-password gate not driven through the UI, `startSsoRecovery` untested at the integration layer (ENG-2592), revocation not instantaneous under `cookieCache` (ENG-2591), a squatter's JWT access token valid for up to 15 minutes, and API keys surviving recovery.

**Risks**

- Same as #8962, in particular that a "dead column" cleanup could silently re-open the 2FA half (the legacy `User` nulls are load-bearing while the backfill shim exists — ordering on ENG-1826) and that simplifying the null into a row delete re-introduces the password-reset lockout.
- **Merge order:** keep this draft until #8962 merges, so the release branch never carries a fix that `main` does not. Both must land — the release line does not inherit from `main`.
- **5.3 is not covered by this PR.** Production runs 5.3.4 and `release/5.3` is on Better Auth 1.6.23 with no `Account.issuer` column and no `@better-auth/core/db` import anywhere, so the last commit cannot be cherry-picked there and that backport is a genuinely different patch, not a mechanical one. Tracked separately rather than bundled here.

---

> [!NOTE]

### DIFF
--- a/apps/web/app/api/auth/sso/recovery/complete/route.ts
+++ b/apps/web/app/api/auth/sso/recovery/complete/route.ts
@@ -1,12 +1,12 @@
 import { NextResponse } from "next/server";
 import { logger } from "@formbricks/logger";
 import { verifySsoRelinkIntent } from "@/lib/jwt";
-import { deleteSessionBySessionToken } from "@/modules/auth/lib/auth-session-repository";
 import { getSession } from "@/modules/auth/lib/session";
 import {
   BETTER_AUTH_SESSION_COOKIE_NAMES,
   getSessionTokenFromCookieHeader,
 } from "@/modules/auth/lib/session-cookie";
+import { revokeSessionByToken } from "@/modules/auth/lib/session-revocation";
 import { completeSsoRecovery, getSsoRecoveryFailureRedirectUrl } from "@/modules/ee/sso/lib/sso-recovery";
 
 const clearSessionCookies = (response: NextResponse) => {
@@ -31,7 +31,9 @@ const buildFailedRecoveryResponse = async (request: Request, callbackUrl?: strin
   }
 
   try {
-    await deleteSessionBySessionToken(sessionToken);
+    // Through the two-store revocation, not a raw Prisma delete: sessions live in Redis too, and a
+    // DB-only delete would leave this one resolvable by `getSession` until its TTL (ENG-2557).
+    await revokeSessionByToken(sessionToken);
   } catch (error) {
     logger.error(error, "Failed to delete SSO recovery session after recovery completion error");
   }
--- a/apps/web/app/api/auth/sso/recovery/complete/route.ts
+++ b/apps/web/app/api/auth/sso/recovery/complete/route.ts
@@ -52,6 +52,8 @@ export const GET = async (request: Request) => {
     const callbackUrl = await completeSsoRecovery({
       intentToken,
       sessionUserId: session?.user.id,
+      // Spared by the post-commit session sweep, so the redirect below still lands signed in.
+      sessionToken: getSessionTokenFromCookieHeader(request.headers.get("cookie")) ?? undefined,
     });
 
     return NextResponse.redirect(callbackUrl);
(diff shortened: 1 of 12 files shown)


xernobyl:
Closing this for now — reopening as a fresh backport once #8962 merges.

The reason is in this PR's own body: its argument was that it is byte-identical to #8962's reviewed head. That stopped being true when the review fixes landed there, and #8962 is likely to move again while Matti's review is open. Keeping two backports in sync with a PR still under review means re-cherry-picking on every round, and it leaves a green draft sitting here that looks ready but is missing those fixes.

Cleaner to cherry-pick the merge commit once, which is what our backport flow assumes anyway. The branch stays; nothing is lost.

PR #8970, opened 2026-08-24
fix(sso): strip the live local auth factors on SSO recovery (backport 5.3)

xernobyl:
## What & why

Backport of #8962 to `release/5.3` — **the line production runs** (5.3.4). The affected range is 5.2.0–5.3.4, so this is the branch where the vulnerability is actually live; #8969 covers `release/5.4`, which is not yet GA.

SSO verify-before-link recovery is supposed to strip a recovered account's local auth factors before the SSO identity becomes the way back in. The strip was written before ENG-1054 moved the credential to `Account.password` and 2FA into the `TwoFactor` table, and was never updated — so post-cutover it nulled three already-null legacy `User` columns and stripped nothing, while reading in code as though it did. An attacker who set a password on an unverified account keeps that password across the victim's recovery. Root cause, blame and the full evidence are on #8962 and the ticket.

**Seven of the original eight commits, byte-identical.** The five source files the fix touches are identical between `release/5.3` and `main`, so seven commits cherry-picked with zero conflicts and the tree matches #8962's reviewed commit 7-of-8 (`fbb8c2b6a`) exactly across all eleven files.

**Two deliberate deviations, both forced by this line being on Better Auth 1.6.23:**

1. **The eighth commit is dropped** (`match Better Auth's own credential-account predicate`). It filters `hasCredentialAccount` on `issuer: createLocalAccountIssuer("credential")`, and 1.6 has no `Account.issuer` column — the column and that `@better-auth/core/db` export both arrive with 1.7. This is not a compromise: 1.6's own `findCredentialAccount` doesn't filter on `issuer` either, so the pre-`63ef4bd21` predicate is the *correct* one here.
2. **The access-token revocation is removed** (one commit, `bc44adc3c`). `oauthAccessToken.revoked` also arrives with 5.4, so the write does not compile here — the typecheck caught this, not review. Nothing is lost: every access token is a self-contained JWT that `createJwtAccessToken` signs without persisting, and `/api/mcp` verifies bearers against JWKS without reading that table, so the column would only ever have mattered for an opaque-token configuration this build does not run. **The refresh token — the 30-day persistence that actually outlives the session — is still revoked, as is consent.** The residual is identical to `main`'s: a squatter's JWT stays valid for up to `accessTokenExpiresIn` (15 min).

`oauthRefreshToken.revoked`, `Account.password`, `TwoFactor` and all four `oauth*` models are present on this line, and `internalAdapter.deleteSessions` exists in 1.6.23 (`dist/db/internal-adapter.mjs:408`) with the same array-of-tokens signature and the same `secondaryStorage`-then-rows semantics the session sweep depends on — so nothing else needed adapting.

## Linear ticket

https://linear.app/formbricks/issue/ENG-2557

## How this was tested

Run in a worktree cut from `origin/release/5.3` with the dependency graph built (12/12), against Better Auth 1.6.23:

- **Typecheck** ✅ `pnpm -C apps/web run typecheck` — clean. This is what found deviation 2 above; the first run failed with two `TS2353` errors on `oauthAccessTokenWhereInput` having no `revoked` property.
- **Unit** ✅ `npx vitest run modules/auth modules/ee/sso lib/user` — **548 passed / 51 files**, no failures. The five directly touched test files are 50/50.
- **Mutation, 2 applied and reverted** — because deviation 2 required editing a test, a green suite alone was not adequate evidence that it still binds the fix:
  - neutering the credential strip (`data: { password: null }` → `data: {}` in `sso-recovery.ts`) → **2 tests red**
  - dropping the refresh-token revocation → **1 test red**
  - both reverted → 13/13 green again
- **Not run locally: the integration suite.** It needs a Postgres provisioned from *this* branch's schema, and pointing it at another worktree's container clones that branch's schema instead — which produces ~20 schema-shape failures that look like regressions and aren't. CI provisions its own DB from this PR's schema and is the authoritative signal; the `Better Auth integration tests` job here is what to read.
- Everything else derives from #8962, whose own suites passed 18/18 on `1473fceb8`: the live before/after through real HTTP (**200 → 401**, `Account.password` present → null) and fourteen mutations.

## Breaking changes

None — no API/SDK shape, endpoint, route, env var, config key or default changes, and no deployer action. The behaviour change is the security fix itself.

## QA / Test Plan

**How to test**

- [ ] **The core fix.** On an instance with SSO configured and email verification **required**: register a local account with an address you control but do *not* verify it, and note the password. Then complete SSO sign-in with that same address (the verify-before-link recovery flow) and let it link. Now return to `/auth/login` and try the original email + password → **expected: sign-in is refused** (invalid-credentials, no session). Before this fix the same attempt signs in successfully.
- [ ] **The recovered user can still get a password back.** As that user, go to `/auth/forgot-password`, request a reset for the address → **expected: the reset email arrives**, the link sets a new password, and the new password signs in while the old one is still refused. (This is the lockout the fix deliberately avoids by nulling the password rather than deleting the credential row.)
- [ ] **2FA does not survive.** Repeat the first step with 2FA enabled on the unverified account before recovery → **expected: after recovery, sign-in via SSO does not prompt for a 2FA code, and no 2FA remains configured on the account.**
- [ ] **The caller stays signed in.** At the end of the recovery redirect → **expected: you land signed in**, not bounced to the login page. Other sessions that account held elsewhere should be signed out (check from a second browser signed in as that account before recovery → it should be logged out).
- [ ] **An already-verified account is untouched.** Take an account whose email *is* verified and with a working password, and complete SSO sign-in for it → **expected: nothing is stripped** — the password still signs in and 2FA, if configured, still applies.
- [ ] **Edge case — recovery failure.** If the recovery completion fails (e.g. an expired/invalid `intent` token on `/api/auth/sso/recovery/complete`) → **expected: redirect to `/auth/login?error=oauth_account_not_linked`, and the session that request carried is no longer valid.**

**Preconditions / test data**

- An enterprise licence, since SSO is licence-gated — without one no SSO buttons render and the provider callback 403s.
- An SSO provider configured (Google is easiest) and `EMAIL_VERIFICATION_DISABLED` unset/`0`, i.e. verification **required** — that is the posture where this matters.
- A mail catcher (or real inbox) for the reset-email step.
- Two accounts: one unverified-with-password (the attacker-shaped case) and one verified-with-password (the control).
- Cloud and self-host behave identically here; no feature flag gates the fix.

**Risks & regressions**

- **Sign-in and password reset for existing password users** — the strip now actually writes, so re-check that an ordinary, never-recovered password user still signs in and can still reset.
- **A "dead column" cleanup could silently re-open the 2FA half.** The legacy `User` nulls are load-bearing while the `better-auth-two-factor-backfill` shim exists — it re-materialises a `TwoFactor` row from `twoFactorEnabled && twoFactorSecret` on every successful credential sign-in, so dropping them would let the next sign-in re-arm a factor recovery had just removed. Ordering recorded on ENG-1826.
- **Simplifying the null into a row delete re-introduces a lockout** — a recovered user would have no password *and* no way to ask for one. Bound by the second QA step above.
- **A legitimate recovered user loses their MCP tokens** and must re-authorize. Correct for a handover, but a real consequence.
- **Revocation is not instantaneous**: `cookieCache` (`maxAge: 5 * 60`) answers `getSession` from a signed cookie without consulting either store, so a revoked session's holder can keep a cached session for up to 5 minutes. Upstream by design, true of every revocation path; filed as ENG-2591.
- **Merge order:** keep this draft until #8962 merges, so a release branch never carries a fix `main` does not. All three need to land — the release lines do not inherit from `main`.

- none

### DIFF
--- a/apps/web/app/api/auth/sso/recovery/complete/route.ts
+++ b/apps/web/app/api/auth/sso/recovery/complete/route.ts
@@ -1,12 +1,12 @@
 import { NextResponse } from "next/server";
 import { logger } from "@formbricks/logger";
 import { verifySsoRelinkIntent } from "@/lib/jwt";
-import { deleteSessionBySessionToken } from "@/modules/auth/lib/auth-session-repository";
 import { getSession } from "@/modules/auth/lib/session";
 import {
   BETTER_AUTH_SESSION_COOKIE_NAMES,
   getSessionTokenFromCookieHeader,
 } from "@/modules/auth/lib/session-cookie";
+import { revokeSessionByToken } from "@/modules/auth/lib/session-revocation";
 import { completeSsoRecovery, getSsoRecoveryFailureRedirectUrl } from "@/modules/ee/sso/lib/sso-recovery";
 
 const clearSessionCookies = (response: NextResponse) => {
@@ -31,7 +31,9 @@ const buildFailedRecoveryResponse = async (request: Request, callbackUrl?: strin
   }
 
   try {
-    await deleteSessionBySessionToken(sessionToken);
+    // Through the two-store revocation, not a raw Prisma delete: sessions live in Redis too, and a
+    // DB-only delete would leave this one resolvable by `getSession` until its TTL (ENG-2557).
+    await revokeSessionByToken(sessionToken);
   } catch (error) {
     logger.error(error, "Failed to delete SSO recovery session after recovery completion error");
   }
--- a/apps/web/app/api/auth/sso/recovery/complete/route.ts
+++ b/apps/web/app/api/auth/sso/recovery/complete/route.ts
@@ -52,6 +52,8 @@ export const GET = async (request: Request) => {
     const callbackUrl = await completeSsoRecovery({
       intentToken,
       sessionUserId: session?.user.id,
+      // Spared by the post-commit session sweep, so the redirect below still lands signed in.
+      sessionToken: getSessionTokenFromCookieHeader(request.headers.get("cookie")) ?? undefined,
     });
 
     return NextResponse.redirect(callbackUrl);
--- a/apps/web/lib/user/password.test.ts
+++ b/apps/web/lib/user/password.test.ts
@@ -2,12 +2,13 @@ import { beforeEach, describe, expect, test, vi } from "vitest";
 import { prisma } from "@formbricks/database";
 import { InvalidInputError } from "@formbricks/types/errors";
 import { verifyPassword } from "@/modules/auth/lib/utils";
-import { getCredentialPasswordHash, verifyUserPassword } from "./password";
+import { getCredentialPasswordHash, hasCredentialAccount, verifyUserPassword } from "./password";
 
 vi.mock("@formbricks/database", () => ({
   prisma: {
     account: {
       findUnique: vi.fn(),
+      count: vi.fn(),
     },
   },
 }));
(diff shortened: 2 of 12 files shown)


xernobyl:
Closing this for now — reopening as a fresh backport once #8962 merges. Same reason as #8969: keeping backports in sync with a PR still under review means re-cherry-picking every round, and a green draft that is missing the review fixes is a merge hazard rather than a head start.

Keeping the 5.3-specific research here so it is not rediscovered, since `release/5.3` is on Better Auth 1.6.23:

- `Account.issuer` does not exist, so the "match Better Auth's own credential-account predicate" commit cannot apply — and should not. 1.6's own `findCredentialAccount` does not filter on issuer either, so the older predicate is the correct one on this line.
- `oauthAccessToken.revoked` does not exist either, so the access-token revocation drops entirely. That costs nothing: access tokens are self-contained JWTs that are never persisted. With the audit counts now split, 5.3 gets `oauthRefreshTokensRevoked` only, not the pair.
- Everything else is present: `oauthRefreshToken.revoked`, `TwoFactor`, `Account.password`, all four `oauth*` models, and `internalAdapter.deleteSessions` with the same signature and the same dual-store semantics.

The branch stays.

PR #8971, opened 2026-08-24
fix(auth): hand the post-verification session only to the browser that signed up

xernobyl:
Fixes ENG-2562

Reported by **[frandle331-yh](https://github.com/frandle331-yh)**, who asked to be named — please carry the credit into the release notes for whichever release ships this.

## What & why

`emailVerification.autoSignInAfterVerification` was `true`, justified in code as *"Safe — clicking the signed link proves email ownership."* It does prove that. It does **not** prove the clicker chose the account's password, and treating those as one fact is textbook account pre-hijacking (Sudhodanan & Paverd, USENIX '22):

1. Attacker registers an address that has no account yet, with a password they pick.
2. Formbricks mails the verification link to the real owner.
3. The owner clicks it and is signed **into the attacker's account** — in an organization the attacker provisioned, since `handlePostUserCreation` runs at sign-up time.
4. The attacker logs back in with the password they set and reads whatever the victim created.

Verified live end to end, before and after. On `main` step 3 issues `formbricks.session_token`, writes a `Session` row, and `get-session` returns the victim's own address — the victim authenticated inside someone else's account.

The flag is now `false`, so Better Auth mints nothing, and a new `hooks.after` handler mints the session **only** when the request carries a sign-up intent cookie naming the user who was just verified — i.e. only when the browser finishing the verification is the browser that started the sign-up. Everything else gets a verified email and no session, and is sent to `/auth/login?verified=1` with an explanation. The ENG-1746 land-in-the-app UX survives for exactly the case it was meant to serve.

**Deliberately non-destructive.** The stored password is never touched. That is what makes a mail-security scanner prefetching the link cost a session instead of somebody's password — a real operational hazard for the alternative remediation the report offers.

Two design notes worth a reviewer's attention:

- **Success is taken from Better Auth's own signal, not inferred.** `afterEmailVerification` fires once per user, immediately after the `emailVerified` write, and never on the already-verified early return. It is carried to the after-hook through a request-scoped store. Reading the response instead would be wrong: a successful verification carrying a `callbackURL` finishes as a *thrown redirect*, so it is an `APIError` — structurally identical to a failure.
- **The intent cookie does not go through `verifyToken` in `lib/jwt.ts`.** That helper's purpose resolution fails open (an unknown `purpose` is silently rewritten to `email_verification`), and its signature-failure fallback does a DB lookup — which would put one or two queries behind every malformed cookie on an unauthenticated GET, including every scanner prefetch. `signup_intent` is kept out of `VERIFICATION_TOKEN_PURPOSES` for the same reason: nothing lenient can resolve one of these. `lib/jwt.ts` is untouched by this PR.

## Breaking changes

None by the criteria above — no API/SDK shape, endpoint, route, env var, config key or default changes, and no deployer action. Judged rather than assumed, because there *is* an observable behaviour change: `/api/auth/verify-email` no longer sets a session cookie unless the intent cookie is present, and redirects to `/auth/login?verified=1` when it withholds. That is the security fix itself, needs no migration, and affects no documented contract. One consequence to know about: sign-ups driven through Better Auth's native `/sign-up/email` never receive an intent cookie, so their verification withholds the session — fail-closed, and deliberate.

## Migrations & env

- none

## How this was tested

**Coverage**

Asserted through sessions rather than columns throughout: `emailVerified` is `true` on every path here — that write is correct, the mailbox really was proven — so a column-level assertion would call the vulnerable behaviour a pass.

Integration commands (real Postgres + real Redis):
`cd apps/web && pnpm test:integration modules/auth`

| Behaviour | How | Outcome |
| --- | --- | --- |
| The victim is no longer signed into the attacker's account | manual (live, before/after) | Same stack, same script, only the code swapped — `origin/main`'s auth code checked out for the "before". **BEFORE: 302 → `/`, `formbricks.session_token` issued, 1 `Session` row, `get-session` → signed in as `victim@example.com`. AFTER: 302 → `/auth/login?verified=1`, zero `Set-Cookie`, 0 sessions, `email_verified=t`.** The ticket's PoC steps 1–4, run verbatim |
| A verification with no sign-up proof grants nothing | integration (red on main) — `modules/auth/lib/better-auth-verification-autosignin.integration.test.ts` | `emailVerified` true, `getSession` null, zero `Session` rows, and at the HTTP layer no session cookie plus a 3xx to `/auth/login?verified=1` |
| The same browser that signed up still lands in the app | integration (red on main) + manual (live) | Session minted for that user, `session_token` on the response, spent intent cookie cleared on the same response. Live: landed on `/setup/organization/create`, `get-session` → the signer-upper |
| An intent cookie for a different account buys nothing | integration (red on main) | Zero sessions. Proof of registering account A is not proof of registering account B |
| A token from another flow is refused | integration (red on main) | A correctly-signed `sso_recovery` token, minted by the app's own `createToken`, is rejected — the shared-keyspace confusion the dedicated verifier exists to prevent |
| A garbage cookie withholds rather than 500s | integration (guard) | Verified, not signed in. The user is never stranded by a bad cookie |
| **Verification-disabled (the shipped self-host default) keeps an existing session** | manual (live) + unit | Signed in *before* verifying, then clicked the link: 302 → `/`, `email_verified` flipped, **the same `Session` row preserved**, original cookie still authenticated. This is the common path on default self-hosted configs, not an edge case |
| The `signedIn` audit trail survives the move | integration (guard) | The session is now minted by this hook rather than by `autoSignInAfterVerification`, and still emits `signedIn` / `authMethod: "password"`. Losing it would have been a silent hole in the sign-in trail, not a visible failure |
| The withheld path is recorded | unit (mutation) | Audited *and* logged — audit logging is enterprise-gated, so the log line is what a self-hoster gets |
| Rate limits actually bind | manual (live) | `/send-verification-email`: attempts 1–3 → 200, 4th → **429**. `/verify-email`: 8 consecutive requests, **no 429** (401s only). The asymmetry is deliberate — see Risks |
| The withheld-path banner renders | manual (live) | Verified in the real DOM on a signed-out browser — a success-variant alert above the login form reading *"Email verified — Your email address is confirmed. Sign in to continue. If you did not create this account, do not sign in — reset the password instead to take control of it."* Screenshot to be attached (it cannot be uploaded from the CLI); the prop wiring is also pinned by the RSC payload carrying `emailJustVerified: true` only with `?verified=1` |
| Ordinary auth is unaffected | manual (live) | Correct password → 200, wrong password → 401, unknown email → 401, reset request → 200 with the mail delivered, replayed verify link idempotent with no extra session |

**Nine mutations on the handler, each applied and reverted — all nine turn the suite red**: path scoping, the just-verified marker, the intent/user binding, the same-user early return, the already-granted guard, the snapshot ordering, the withheld redirect, the withheld audit, and the single-use cookie clear. Two vacuous tests were found and fixed by that process: an `alg: none` row that passed with the algorithm pin removed (relabelled — jsonwebtoken already refuses `none` with a string secret), and a marker-guard row that passed for the wrong reason, which surfaced a real latent bug where an absent session and an absent marker both read `undefined` and matched each other.

Red-on-main is structural, not a claim: flip `autoSignInAfterVerification` back to `true` and 5 of the 6 new integration cases fail, including the core withhold. Verified in both env postures — verification required *and* `EMAIL_VERIFICATION_DISABLED=1`, which is what CI runs.

**Open gaps**

- **The non-English locale strings are hand-written, not Lingo output.** No Lingo API key was available, and `Validate Translation Keys` blocks without them, so the 14 target locales and the `i18n.lock` entries were written by hand — normally forbidden, since those files are generated. Register was matched per locale from the neighbouring `auth.login` strings (informal `du`/`tu` where those files already use it, `vous` in fr-FR, formal `u` in nl-NL, polite ですます in ja-JP, pt-BR/pt-PT kept distinct). The lock checksums use the same `object-hash` MD5 the validator does, verified against an existing entry rather than guessed. **Correcting what an earlier revision of this section claimed: Lingo will NOT regenerate these.** A key is re-translated only when its `i18n.lock` checksum no longer matches the en-US source, and these entries were written to match — which is what makes the validator pass, and equally what makes Lingo skip the keys. So this wording ships as-is and stays until someone with a Lingo key deliberately busts the lock for these two keys and re-runs `pnpm i18n:web:generate`. Worth a reviewer who reads a target language correcting the wording here rather than assuming a later run will. Removing the two lock entries instead does not work: `scan-translations` also runs `validateLockfile`, and I tried it — it fails with `LOCKFILE OUT OF SYNC: Keys missing from lockfile (2)`. The lock entries and the translations cannot be decoupled. Only the English render was confirmed visually; for the other locales I verified key presence, valid JSON, and that `Accept-Language` resolves the locale correctly.
- **No security notification to the verified address on the withheld path.** That branch is the only moment the system knows an address was registered by someone who did not finish the verification, and telling the owner would make the residual below visible to the one person who can act on it. It needs a new email template plus more locale keys, so it is filed rather than grown into a fix that also has to backport — worth doing.
- **A resend after the intent cookie has expired withholds the session even for the browser that signed up.** A resent link gets a fresh hour while the cookie's clock started at sign-up. The resend action now re-pairs the cookie with the new link when the requesting browser still holds a valid one for the same user — strictly a refresh, so an unauthenticated caller cannot arm sign-up proof for an arbitrary account by asking for a resend. What remains is the post-expiry case: once the cookie is gone, that browser cannot be told apart from one that never signed up, so a resend more than an hour after sign-up stays fail-closed — the user verifies, lands on `/auth/login?verified=1`, and signs in with the password they just chose. Lengthening the cookie instead would widen the window in which a stale cookie is spendable; left as the accepted residual.
- **A scanner prefetch consumes the `?verified=1` explanation.** If a mail scanner opens the link first, its request is the one that verifies (and gets the withheld redirect); the human's later click hits Better Auth's already-verified early return at `email-verification.mjs:285`, which redirects to the `callbackURL` (`/`) *before* `afterEmailVerification` fires — so this hook returns without acting and middleware bounces them to `/auth/login` with no banner. They can still sign in; they just are not told why they were not signed in automatically. Pre-existing shape of the already-verified branch rather than something this change introduces, and not fixable from the after-hook, which is why it is listed rather than fixed.
- **The intent cookie's expiry is not covered at the integration layer.** It is one hour and unit-tested with fake timers; no integration row waits it out.
- **No E2E.** Checked deliberately: the suite never follows a verification link — `signup.spec.ts` asserts `/auth/signup-without-verification-success`, and the user fixture writes `emailVerified: true` straight to the DB — so there is no existing spec to extend and nothing here regresses it.

**Risks**

- **The attacker's password survives, so they can sign in until the victim resets.** This removes the data-comingling harm — the victim never operates inside the attacker's account — but it does not evict the squatter, and the victim's click does leave that account verified. The report's other remediation (invalidate the credential) would close it, at the cost of a mail scanner's prefetch nulling a legitimate password. Chosen deliberately; pinned as a test so making it destructive later is an explicit edit rather than a surprise.
- **This does not close the ENG-2557 interaction.** A pre-hijacked account ends up `emailVerified: true`, which is the guard the SSO-recovery strip keys on, so recovery will not strip the attacker's credential either. Needs its own ticket — `sso-recovery.ts` has open PRs on it and is deliberately untouched here.
- **Cross-device verification now costs one extra step**: the email is verified, then the user signs in with the password they already chose. Confirmed live to recover in a single sign-in, no reset needed. A mail-scanner prefetch lands in the same branch and costs the same.
- **`/send-verification-email` is capped at 3/60s**, the same budget as `/request-password-reset`. Shared-NAT users could reach it on repeated resends; that exposure already exists for the sibling endpoint and is accepted. `/verify-email` is left generous (30/60s) on purpose — it is a top-level GET out of a mail client, often behind corporate NAT and often prefetched, and it already requires a signed expiring token, so a tight per-IP cap buys little and risks locking people out of verifying at all.
- **Backports needed**: the flag is `true` on both `release/5.3` and `release/5.4`. 5.3 runs Better Auth 1.6.23 — confirm the `afterEmailVerification` and dispatch hook semantics there before assuming a clean cherry-pick, the way the ENG-2557 5.3 backport turned out to need adapting.

---

> [!NOTE]


apps/web/i18n.lock
```diff
@@ -44,6 +44,8 @@ checksums:
     auth/last_used: 9049bf77c57c92b7a5a760ba2a40128c
     auth/login/backup_code: 870711a6df11ae55366d018255b466cf
     auth/login/create_an_account: bc53a57281d2701563f94f3169518028
+    auth/login/email_verified_sign_in_description: 9937ca8606fed9bfd7d83902fa7346b5
+    auth/login/email_verified_sign_in_title: f960899104c4e22c27ab699272b2bbc1
```
kiln-agent, line 48: These two lock entries are the reason the hand-written translations will **not** be regenerated — which inverts the Open-gaps claim that "Lingo will regenerate and overwrite these on its next run, which is the intended end state".

I recomputed both checksums and they are exactly right:

```
$ node -e "const oh=require('object-hash'); …"
"Email verified"                 f960899104c4e22c27ab699272b2bbc1
"Your email address is confirme" 9937ca8606fed9bfd7d83902fa7346b5
```

Which matches the two added lines verbatim. That is the problem: the lockfile checksum is a hash of the **en-US source**, and a matching checksum is exactly how Lingo decides a key is already up to date. Both keys are also present in all 14 target locales, so `lingo.dev run` sees "source unchanged, target present" and skips them. Whatever wording is in this PR ships permanently, until someone edits the English string.

Two ways out, both cheap:

1. **Drop just these two lines from `i18n.lock`** and keep the hand-written target strings. The keys then look new to Lingo and get regenerated on its next run, while `scan-translations` still passes today — I confirmed that locally, it only checks key presence/completeness per locale, not the lock. This gets you the disclosed end state for real.
2. Have a maintainer with the Lingo key run `pnpm i18n:generate` before merge.

Option 1 seems strictly better than the current state, since it costs nothing and makes the disclosure accurate. Either way, worth updating that Open-gaps bullet — as written it tells a reviewer they can skip reading the target languages, and right now they can't.

---

apps/web/modules/auth/lib/signup-intent.ts
```diff
@@ -0,0 +1,114 @@
...
+/** Cookie attributes, matching Better Auth's `advanced.defaultCookieAttributes`, with one exception. */
+export const SIGNUP_INTENT_COOKIE_OPTIONS = {
+  httpOnly: true,
+  secure: USE_SECURE_COOKIES,
+  path: "/",
+  // `lax`, not `strict`: the verification link is a top-level GET navigation arriving from a mail
+  // client, i.e. cross-site. `strict` would withhold the cookie on exactly the request that needs it,
+  // turning every legitimate same-browser sign-up into a withheld session.
+  sameSite: "lax",
+  maxAge: EMAIL_VERIFICATION_TTL_SECONDS,
```
kiln-agent, line 72: The "cookie outlives the link it is paired with" invariant in `auth-cookies.ts` holds for the *first* link only, and the resend paths break it.

The cookie's `maxAge` starts at sign-up. A resent verification token is minted fresh, with its own full `EMAIL_VERIFICATION_TTL_SECONDS`. So:

1. Sign up at T+0 → cookie and link both die at T+60m.
2. Email lands in spam; user hits resend at T+70m → new link valid until T+130m, cookie already gone.
3. User clicks it **in the same browser that signed up** → `readSignupIntentUserId` returns `null` → withheld → `/auth/login?verified=1`.

Fail-closed, so not a security issue — but it is a permanent loss of the ENG-1746 UX for a legitimate same-browser sign-up, and it isn't in Open gaps (the listed gap is only that the expiry lacks integration coverage). Given how often a verification mail is missed on the first pass, I'd guess this is a more common path than the one the integration suite covers.

The fix that seems most worth it is on the `sendOnSignIn` branch specifically, because the proof available there is *stronger* than the sign-up cookie's. Better Auth verifies the password before it looks at `emailVerified`:

```js
// node_modules/better-auth/dist/api/routes/sign-in.mjs
:332  if (!await ctx.context.password.verify({ … password })) { … throw INVALID_EMAIL_OR_PASSWORD }
:339  if (…requireEmailVerification && !user.emailVerified) {
:341    if (…emailVerification?.sendOnSignIn) {
```

So a browser that reaches the resend has just demonstrated it knows the credential — which is precisely the fact clicking the link fails to establish, and precisely what the attacker cannot obtain in the victim's browser (the victim doesn't know the attacker's password). Re-issuing the intent cookie there is safe against the pre-hijack and restores land-in-the-app for the "signed up, link expired, signed in again" path.

`/send-verification-email` is the weaker case — unauthenticated, so a cookie issued there would be attacker-obtainable — I'd leave that one withholding. Happy for this to be a follow-up ticket rather than growing this PR; mostly it should be in Open gaps either way.

---

apps/web/modules/auth/lib/better-auth-verification-autosignin.ts
```diff
@@ -0,0 +1,163 @@
...
+    // Whether the ENDPOINT already attached a session cookie, meaning it made its own sign-in decision
+    // and this hook must not second-guess it. Unreachable in today's config — the only `/verify-email`
+    // branch that does this is Better Auth's `updateTo` email-change flow and `user.changeEmail` is not
+    // enabled — but that branch also fires `afterEmailVerification`, so the guard keeps the hook correct
+    // if it is ever enabled. Server-controlled header, not spoofable.
+    //
+    // Read before `getSessionFromCtx` as belt-and-braces, NOT because a live bug needs it. That helper
+    // appends any Set-Cookie the session read produced — `getSession` re-issues the cookie once
+    // `updateAge` elapses — but it appends to better-call's `ctx.responseHeaders`, which is a DIFFERENT
+    // Headers instance from Better Auth's `ctx.context.responseHeaders` read here (verified at runtime
```
kiln-agent, line 116: `hadIntentCookie` overstates what this boolean knows. `readSignupIntentUserId` collapses absent, malformed, expired, wrong-signature and wrong-purpose all to `null`, so `intentUserId !== null` is really "a **valid** intent cookie naming a **different** user".

The consequence is on the `false` side, which is where the ambiguity lands: no cookie at all (the ordinary cross-device click, and the mail-scanner prefetch) reads identically to an expired one and to a correctly-signed `sso_recovery` token being probed here. `true` is unambiguous and genuinely suspicious; `false` covers both the boring case and the interesting ones.

Since this log line is the only view a self-hoster gets — the audit half is enterprise-gated, as the comment right below says — that's the one distinction you'd actually want during triage. Either rename to something the value supports (`intentCookieNamedAnotherUser`), or have the reader hand back a small reason so you can log `reason: "absent" | "invalid" | "other_user"`. The latter costs a return-type change in `signup-intent.ts` but makes the withheld path diagnosable.

---

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -238,6 +242,35 @@ export const auditFailedAuthAfter = async (ctx: AuthHookContext): Promise<void>
...
+ * ENG-2562: a verification completed, but the browser presenting it was not the one that signed up, so
+ * no session was granted. Worth a record because it is the observable footprint of an attempted account
+ * pre-hijack — someone registered this address and someone else finished the verification.
+ *
+ * Uses the `updated` + marker idiom of `auditPasswordReset` below rather than a new `ZAuditAction`
+ * value: it is the established shape for auth-internal events, and it keeps a shared enum out of a fix
+ * that has to land on two release branches as well as main. Audit logging is enterprise-gated, so the
+ * caller also logs — a self-hoster must still see this.
+ */
+export const auditVerificationSessionWithheld = async (userId: string): Promise<void> => {
```
kiln-agent: Nit: this function landed between `auditPasswordReset`'s doc block and its declaration, so the file now reads as two stacked JSDoc comments on `auditVerificationSessionWithheld` (lines 238–243 and 245–253), and `auditPasswordReset` at line 274 has lost its documentation entirely. TypeScript takes the nearest block, so the password-reset rationale — the `onPasswordReset` wiring and why the snapshots were dropped — no longer surfaces anywhere.

Moving the new function (with its own comment) above line 238 restores both.

---

kiln-agent (commented):
## Review — ENG-2562 pre-hijack via post-verification auto-sign-in

The security reasoning here is correct and the fix is well-placed. Clicking a signed link proves mailbox control and nothing about who chose the password; `autoSignInAfterVerification: true` conflated the two, and moving the grant behind a sign-up intent cookie is the right discriminator. I traced the attack paths and could not find a way through:

- **Can the attacker get an intent cookie for the squatted account into the victim's browser?** No — `createUserAction` issues it only on `outcome.status === "created"`, and a second sign-up for an existing address returns `already_existed`. That exclusion is genuinely load-bearing and the test pinning it is the right test.
- **Can the attacker get the token into their own browser, where they do hold a cookie?** No — the token is a signed JWT that only reaches the victim's inbox, and `sendOnSignIn` resends go to the same place.
- **Can a token from another flow be spent as an intent cookie?** No, and keeping `signup_intent` out of `VERIFICATION_TOKEN_PURPOSES` is the right call given that `getVerificationTokenPurpose` rewrites an unknown purpose to `email_verification`. Avoiding `verifyToken`'s DB fallback on an unauthenticated GET is a real win, not just tidiness.

Two things I want to call out as unusually good, because they're the kind of thing that normally goes wrong here. Taking success from `afterEmailVerification` rather than inferring it from the response is correct for exactly the reason stated — a successful verification carrying a `callbackURL` finishes as a thrown redirect and is structurally an `APIError`. And keeping `throw ctx.redirect(...)` outside the `try` is the difference between a login page and a 500 on a link that has already flipped `emailVerified`, which would strand the user permanently since the retry hits the already-verified early return.

Verified locally rather than taken on trust:

- `pnpm build --filter=@formbricks/web^...` then the auth unit suite — **510 passing across 46 files**, including all 46 new/changed cases in the four touched test files.
- `pnpm run scan-translations` — passes.
- Recomputed both `i18n.lock` checksums (see the inline comment — this is the one finding I'd act on before merge).
- Read `better-call`'s `getCookie`/`setCookie` and `getCookieKey` to confirm passing the already-`__Secure-`-prefixed name with no `prefix` argument is right, and read Better Auth's `sign-in.mjs` to confirm the password-before-`emailVerified` ordering the second comment relies on.

The test suite is the strongest part of the PR. Asserting through sessions rather than `emailVerified` is the correct choice — a column assertion would have graded the vulnerable behaviour a pass — and the mutation pass that caught the two vacuous rows (including the `undefined === undefined` marker-guard bug) is real work. Pinning the surviving-password residual as a test so a later destructive change is deliberate is the right instinct.

### Findings

Four inline comments, none blocking on correctness:

1. **`apps/web/i18n.lock`** — the hand-written lock checksums are *correct*, which is why Lingo will skip these keys rather than regenerate them. That inverts the Open-gaps claim, and means the hand-written wording ships permanently. Cheap fix; worth doing before merge.
2. **`signup-intent.ts`** — the cookie/link co-expiry invariant holds only for the first link; resends outlive the cookie, so a legitimate same-browser sign-up loses land-in-the-app after an hour. Fail-closed, so not a security issue, but not in Open gaps either. Includes a safe suggestion for the `sendOnSignIn` branch.
3. **`better-auth-verification-autosignin.ts`** — `hadIntentCookie` claims more than the value knows, and the ambiguity sits on the side that matters for triage.
4. **`better-auth-observability.ts`** — the new function split `auditPasswordReset` from its doc comment. Pure nit.

### On the PR description

The breaking-change judgement looks right to me: no API/route/env/config surface moves, and the observable behaviour change is the fix itself. The fail-closed consequence for Better Auth's native `/sign-up/email` is correctly identified as deliberate rather than hidden.

One note on the rate limits, not a finding. `/verify-email` at 30/60s is described as "generous", but measured against Better Auth's global default it is a substantial tightening, not a loosening — the manual result ("8 requests, no 429") doesn't probe near the boundary. 30/min per IP is very likely fine, and the reasoning for keeping it loose is sound; I'd just not lean on "generous" as the justification if a large shared-egress tenant ever shows up in the 429 logs.

The ENG-2557 interaction and the surviving-password residual are correctly scoped out with reasons, and the 5.3 backport caveat about Better Auth 1.6.23's hook semantics is the right thing to have flagged.

---

apps/web/modules/auth/lib/better-auth-verification-autosignin.ts
```diff
@@ -0,0 +1,163 @@
...
+    // in this hook: `ctx.responseHeaders !== ctx.context.responseHeaders`; the two are merged only after
+    // the hook returns, by `mergeResponseHeaders` in dispatch.mjs). So a session refresh cannot reach
+    // this value either way. The ordering is kept so the guard's meaning stays unambiguous — it asks
```
kiln-agent, line 119: This event fires on ordinary cross-device verification, so it can't carry the meaning `better-auth-observability.ts` gives it ("the observable footprint of an attempted account pre-hijack").

The branch is reached whenever a first verification arrives without a matching intent cookie. That set includes the pre-hijack, but it also includes every case the PR itself calls common and expected:

- sign up on the laptop, click the link on the phone (no cookie on that device);
- a mail-security scanner prefetching the link — the hazard the non-destructive design is explicitly chosen for, and the reason `/verify-email` is left at 30/60s ("often prefetched by a mail-security scanner before the human clicks");
- any click more than an hour after sign-up, since the cookie's `maxAge` starts at sign-up.

So a self-hoster reading the log line, or an enterprise reading the audit row, gets `verificationSessionWithheldMarker` with no way to separate "someone registered this address and someone else finished the verification" from "user clicked on their phone". That is the one distinction the record exists to make, and the payload carries only `userId`, so it can't be recovered after the fact. Worth distinguishing at the source — the sibling comment thread on `hadIntentCookie` suggests a `reason` from `readSignupIntentUserId`, and `reason: "absent" | "invalid" | "other_user"` would do it here too: `other_user` is the genuinely suspicious one, and only `invalid`/`other_user` are worth a security-flavoured record at all.

The scanner case has a second edge worth knowing about, because it defeats the `?verified=1` mitigation for the browser that *did* sign up:

1. Scanner prefetches → `emailVerified := true`, `afterEmailVerification` fires, no cookie on that request → this audit + a 302 to `/auth/login?verified=1` that nothing reads.
2. The human then clicks the same link in the browser that signed up, valid intent cookie and all. `user.emailVerified` is already `true`, so Better Auth takes its already-verified early return, `afterEmailVerification` never fires, `getJustVerifiedUserId()` is `undefined`, and this handler returns at line 64 before looking at the cookie.
3. The endpoint redirects to `callbackURL` (`/`), middleware bounces the sessionless request to a bare `/auth/login?callbackUrl=%2F` — no `verified=1`, so none of the new banner copy renders.

Step 3 matches `main`'s behaviour (there the scanner ate the session instead), so it is a gap rather than a regression — but it is the most likely non-happy path, and the PR text says everything withheld "is sent to `/auth/login?verified=1` with an explanation", which does not hold for it. Redirecting the already-verified-and-sessionless case to `?verified=1` instead of `callbackURL` would close it without minting anything.

---

apps/web/modules/auth/lib/signup-intent.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * ## Why this does not use `verifyToken` from `lib/jwt.ts`
+ *
+ * Two reasons, both found in the security review of the ENG-2562 plan:
+ *
+ * 1. `getVerificationTokenPurpose` there **fails open** — an absent or unrecognised `purpose` claim is
+ *    silently rewritten to `"email_verification"` and returned as if it had been asserted. That makes
+ *    `VERIFICATION_TOKEN_PURPOSES` a rewriting allowlist rather than a rejecting one, so `signup_intent`
+ *    is deliberately NOT added to it: kept out of that shared keyspace, no lenient consumer can ever
+ *    resolve one of these tokens into a purpose it was not issued for.
```
kiln-agent: This conclusion is inverted, and it matters because it's the stated justification for keeping `signup_intent` out of `VERIFICATION_TOKEN_PURPOSES`.

Because `getVerificationTokenPurpose` fails open, staying *out* of the list is exactly what makes a `signup_intent` token resolve as something it was never issued for. `verifyToken` doesn't reject the unknown purpose, it rewrites it:

```ts
// lib/jwt.ts:59
const getVerificationTokenPurpose = (purpose: unknown): TVerificationTokenPurpose => {
  if (purpose && VERIFICATION_TOKEN_PURPOSES.includes(purpose as TVerificationTokenPurpose)) return …;
  return DEFAULT_VERIFICATION_TOKEN_PURPOSE; // "email_verification"
};
```

And the token this file mints is shape-identical to `createToken`'s output — `{ id: symmetricEncrypt(userId, ENCRYPTION_KEY), purpose }`, HS256 — signed with `BETTER_AUTH_SECRET ?? NEXTAUTH_SECRET`, which on any deployment that has only ever set `NEXTAUTH_SECRET` (both are `optional()` in `lib/env.ts`) is the same key `verifyToken` verifies with. So `verifyToken(<intent cookie value>)` returns `{ id, email, purpose: "email_verification" }`: `payload.id` decrypts, `getUserEmailForLegacyVerification` resolves the row, and the purpose claim is silently discarded. The intent cookie *is* a valid email-verification token by that helper's reckoning — the opposite of "no lenient consumer can ever resolve one of these tokens into a purpose it was not issued for".

Not exploitable today: `verifyToken`'s only caller is `better-auth-recovery-signin.ts`, which demands `purpose === "sso_recovery"` and refuses. So this is about the comment, not the behaviour — but it's a comment that tells the next person the shared keyspace is safe by construction, and any future consumer that trusts a returned `email_verification` would accept this cookie. Reason 2 (no DB lookups on a malformed cookie) already justifies the dedicated verifier on its own; I'd either drop the claim or make it true — a rejecting check (`purpose` must be in a known set, unknown → throw) in `lib/jwt.ts`, or a distinct claim name/secret here so the two keyspaces genuinely can't overlap.

---

kiln-agent (commented):
Correctness pass over the ENG-2562 diff (high effort). The security core holds up: I traced the granted/withheld paths through Better Auth 1.7's `dispatch.mjs` + `better-call` and the mechanics are right — the after-hook's `ctx.setCookie`/`setSessionCookie` writes reach the final response through `mergeResponseHeaders`, the withheld `throw ctx.redirect` replaces the endpoint's `location` (and keeps 302 even when the endpoint returned JSON, since `responseStatus` is undefined and `toResponse` falls back to `data.statusCode`), `ctx.getCookie` reads real request cookies in the router path (`router.mjs:64` passes `headers: request.headers`), and the inactive-user gate still binds because it throws inside the try. The `i18n.lock` checksums also verify against `object-hash` MD5.

Two new findings inline; both are about the *meaning* of things rather than a broken path, and neither blocks. The resend/cookie-expiry gap and the misplaced `auditPasswordReset` JSDoc are already covered in existing threads, so I didn't duplicate them. One smaller thing I'd also fix while in there: the `} else { sessionWithheld = true; }` at line 136 (user or session missing) withholds with no log and no audit, which is the one withheld path that leaves no trace — CodeRabbit's thread flags the same gap but against a `grantSessionToSignupBrowser` helper that doesn't exist in this diff.

---

apps/web/i18n.lock
xernobyl: You're right, and the Open gaps bullet was inverted — fixed the PR description to say these ship permanently unless someone busts the lock. One correction on the suggested fix though: dropping the two lock lines doesn't work, `scan-translations` also runs `validateLockfile` and fails with `LOCKFILE OUT OF SYNC: Keys missing from lockfile (2)` (tried it). The lock entries and the translations can't be decoupled, so the wording here is what ships — happy to take corrections from anyone who reads one of the target languages.

apps/web/modules/auth/lib/signup-intent.ts
xernobyl: Good catch — fixed the within-TTL half in 8ea53a9e: the resend action now re-issues the cookie alongside the fresh link, but only when the requesting browser already holds a valid cookie naming the same user. Strictly a refresh, so a resend can't be used to arm sign-up proof for an arbitrary account. The post-expiry case (your step 2, resend at T+70m) stays fail-closed on purpose — once the cookie is gone that browser is indistinguishable from one that never signed up, and lengthening the maxAge would widen the stale-cookie window. Documented as the residual in Open gaps.

apps/web/modules/auth/lib/better-auth-verification-autosignin.ts
xernobyl: Agreed — the boolean threw away the only distinction worth having. Reworked in 1786841: `readSignupIntent` now returns why it refused, and the handler logs/audits `absent` / `invalid` / `other_user` (plus `grant_failed` for a refused mint). All three flow through the integration suite end-to-end.

apps/web/modules/auth/lib/better-auth-observability.ts
xernobyl: Fixed in 1786841 — moved the new helper (with its own docblock) above `auditPasswordReset`, so the password-reset rationale is attached to its function again.

apps/web/modules/auth/lib/better-auth-verification-autosignin.ts
xernobyl: Agreed. The event now carries the reason (`absent` / `invalid` / `other_user` / `grant_failed`) and its docblock says explicitly that it means "nobody was signed in", not "an attack happened" — `absent` is the ordinary cross-device click or a scanner prefetch, `other_user` is the one worth looking at. Same for the log line, since that's all a self-hoster gets. Also dropped a leftover test comment that still called this path "the footprint of an attempted pre-hijack" (8ea53a9e).

apps/web/modules/auth/lib/signup-intent.ts
xernobyl: You're right — staying out of the list was exactly what would make `verifyToken` rewrite the purpose, not what protected against it. Fixed in 1786841 by changing the claim names: the cookie now carries `uid`/`kind` instead of `id`/`purpose`, so `verifyToken` bails on `if (!payload?.id)` before `getVerificationTokenPurpose` ever sees it. The header comment now states that correctly, and there's a test calling the real `verifyToken` against a minted cookie and asserting the rejection.

apps/web/modules/auth/lib/signup-intent.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  // in lib/jwt.ts FAILS OPEN, rewriting an unrecognised purpose to "email_verification". Staying out of
+  // VERIFICATION_TOKEN_PURPOSES is therefore not protection by itself — carrying no `id` claim is, since
+  // `verifyToken` bails on `if (!payload?.id)` before any purpose is considered.
+  test("is not parseable as a lib/jwt.ts verification token", () => {
+    const decoded = jwt.decode(createSignupIntentToken("user_1")) as Record<string, unknown>;
+
+    expect(decoded.id).toBeUndefined();
+    expect(decoded.purpose).toBeUndefined();
+    expect(decoded.kind).toBe("signup_intent");
+  });
```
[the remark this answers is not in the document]
xernobyl: Done in 8ea53a9e — the test now also calls the real `verifyToken` and asserts it rejects the token. The signature is genuinely valid to it (same secret on a NEXTAUTH_SECRET-only deployment), so the call binds the `if (!payload?.id)` bail-out rather than just the claim layout.

itsjavi (changes requested):
LGTM the approach is the right one IMHO, similar to oauth state cookies.

I'd still need a screencast to confirm how/when the new alert is shown

xernobyl:
New instance;
Register user:
image

Open verification email in incognito window:
image


And you end up with this message:
image

apps/web/locales/en-US.json
```diff
@@ -56,6 +56,8 @@
     "login": {
       "backup_code": "Backup code",
       "create_an_account": "Create an account",
+      "email_verified_sign_in_description": "Your email address is confirmed. Sign in to continue. If you did not create this account, do not sign in — reset the password instead to take control of it.",
```
itsjavi:
```suggestion
      "email_verified_sign_in_description": "Your email address is confirmed. Sign in to continue or reset your password if you didn't create this account.",
```

This sounds more friendly. "To take control" sounds more serious, hacker-level.

xernobyl:
Good call — went with a slightly shorter variant of your suggestion:

> Your email address is confirmed. Sign in to continue. If you didn't create this account, reset the password instead.

Kept "instead" so resetting still reads as *replacing* the sign-in rather than a second option on a menu, but dropped "do not sign in" and "take control of it". You're right that it was incident-response register for what is usually just someone clicking the link on their phone instead of their laptop. And "do not sign in" wasn't really doing any work anyway: whoever didn't create the account doesn't have its password, so they can't sign in even if they try — resetting is the only actionable advice, which is what your version already said.

Heads-up that the screenshots I posted earlier show the *old* wording, since they predate this change. The flow itself is unchanged — to answer the "how/when" directly, there are exactly two outcomes:

- link opened in the **same browser that signed up** → session granted, lands straight in the app, no alert
- link opened **anywhere else** (different browser, incognito, or a mail-security scanner prefetching it) → email still gets verified, but no session is issued, and you land on `/auth/login?verified=1` with the alert

The incognito screenshots above are that second path; the only thing different now is the sentence.

Pushed in 23bb67c1d, together with the 14 other locales and the `i18n.lock` checksum.

One caveat since you're the one who touched the copy: the non-English strings are hand-written rather than Lingo output, and the lock checksum now matches the new English, which means Lingo will *skip* this key rather than retranslate it. So this wording ships as-is in all 15 locales until someone with a working Lingo key deliberately busts the lock — the `LINGO_API_KEY` in the repo env is currently rejected as invalid, so I couldn't regenerate. If you read any of the target languages and the phrasing is off, much better to say so now than after it ships.

itsjavi (approved):

PR #8973, opened 2026-08-25
fix(dashboards): clean up chart-builder confusion from lead gen survey review

Dhruwang:
Ref ENG-2541

## What & why

Johannes and Harsh hit several points of confusion building charts for their lead-gen survey. This addresses the four that are same-PR-sized fixes; the other two need a schema migration and a product decision first, so they're split into their own tickets (see below).

- **Rename "Responses" measure to "Feedback Records"** (`schema-definition.ts`, `en-US.json`) — the count measure was labeled "Responses" while every other part of this feature calls the underlying entity a Feedback Record.
- **Clarify Unique respondents vs Unique responses** — both already had a description shown under the option in the measures dropdown, but neither said what it deduplicated by. Reworded to state it explicitly: respondents dedupe by person, responses dedupe by submission.
- **Hide time-based *grouping* for Big Number and Pie** (`chart-display.ts`, `advanced-chart-builder.tsx`, `time-dimension-panel.tsx`) — those chart types show a single point-in-time value/composition, so bucketing by granularity produced a chart that no longer answered the question its type implies. The toggle itself stays available on every chart type (it doubles as the date-range filter); only the granularity control is hidden for Big Number/Pie, with the toggle's copy switching to "Add a date range filter" for those types.
- **Move Line next to Area** in the chart type selector (`chart-types.ts`) — was `area, bar, line, pie, big_number`, now `area, line, bar, pie, big_number`.

**Split into their own tickets** (each needs a decision this PR shouldn't make silently):
- [ENG-2610](https://linear.app/formbricks/issue/ENG-2610) — importing Hidden Field values into Feedback Records. Not a tweak: hidden field values live in `response.hiddenFields`, a separate object from where the mapping pipeline reads (`response.data`), and hidden fields have no label/type anywhere in the schema to derive a Feedback Record's `field_label`/`hubFieldType` from.
- [ENG-2611](https://linear.app/formbricks/issue/ENG-2611) — disabling Sources instead of deleting them on workspace-access removal. `FeedbackSource` has a composite FK to `FeedbackDirectoryWorkspace` (`onDelete: NoAction`), so pausing a source's status doesn't free up the join row for deletion — needs a migration.
- ~~[ENG-2612](https://linear.app/formbricks/issue/ENG-2612)~~ — **no longer split out.** #8985 merged Line into Area as a `config.areaDisplay` style, was reviewed on its own, and merged into this branch, so its diff (including the enum migration) is part of this PR now. `Migrations & env` and `Breaking changes` below cover it.

**Review fixes (later commits):**
1. The first pass hid the *entire* time dimension for Big Number/Pie, but `TimeDimensionConfig` is `dimension` + optional `granularity` + `dateRange` — with no granularity it's a date-range **filter**, not grouping (exactly what `ai-schema-context.ts` tells the AI to emit for date-only requests, and what `dashboard-date-filter.ts` appends). Nulling the whole thing would have silently widened those charts to all-time on open. Fixed to strip only `granularity` and keep the toggle rendered with a range-only copy variant.
2. Moved the `uniqueRespondents`/`uniqueResponses`/`count` measure descriptions into the i18n path (`getTranslatedFieldDescription`) — they were hardcoded English, unreachable by non-English locales, inconsistent with the label already going through i18n.
3. `uniqueRespondents` is `countDistinct(user_id)`, and `COUNT(DISTINCT x)` skips NULLs — anonymous feedback (no identified respondent) contributed 0 despite counting as a Feedback Record, which reads as a bug for e.g. an anonymous link survey. Description now says it only counts identified respondents.
4. Renaming the count measure's label removed the only measure labeled "Responses" from the AI system prompt, risking a "responses per week" prompt landing on `uniqueResponses` (dedup by submission) instead of `count`. Added a Metric alias line, same pattern as the existing NPS/CSAT/CES aliases.

**Review fixes (this round, from @xernobyl's review and CodeRabbit):**
5. **`getTranslatedFieldDescription` had no test.** The coverage table below claimed one; there wasn't one anywhere in the repo. That mattered because review fix 2 exists precisely so these descriptions reach non-English locales — dropping a key from the `descriptions` map left every non-English user silently on the hardcoded English while the suite, `pnpm i18n` and Validate Translation Keys all stayed green. Now covered, including the fallback path.
6. **The ENG-2612 migration was not idempotent.** Guarded on the enum catalog. Verified against a real Postgres — see the coverage rows; the failure mode was worse than "errors on re-run", because it failed at the *backfill*, before the swap, with `invalid input value for enum ChartType: "line"`, which reads as data corruption rather than as "already applied".
7. **The "Responses" rename was half-done.** `CSAT: Responses`, `CES: Responses` and `Rating: Responses` still sat in the same dropdown as `Unique responses` — the exact ambiguity the rename existed to remove, and they are wrong twice over: they count feedback records (one per answered question), not submissions, so a survey with two rating questions reports two "Rating: Responses" for one response. Renamed to `CSAT/CES/Rating: Records`, matching the `Feedback Records` vocabulary this PR established rather than introducing a third noun, and their descriptions now say "feedback records" too. Say the word if you'd rather have a different label — it is three en-US strings and a `pnpm i18n`.
8. **The deployed Cube schemas had drifted.** `docker/cube/schema/FeedbackRecords.js` *and* `charts/formbricks/cube/schema/FeedbackRecords.js` (both, not just the one flagged) still carried the pre-rewrite copy for `count`, `uniqueRespondents`, `uniqueResponses`, `valueText` and `valueId`. Realigned with the app's copy so the second source of truth does not rot.

9. **The CSAT/CES/Rating descriptions dropped "answered".** All three Cube measures filter on `value_number IS NOT NULL`, so dismissed records are excluded — the Cube-side copy says so and my rewrite kept it only for Rating. All three now read "Number of answered feedback records from *X* questions (dismissed excluded)".
10. **The description tests asserted a key prefix, so a wrong id→key mapping still passed.** Each id now asserts its exact key, with a `t` that returns something distinguishable from the key it was handed. Writing the prototype-chain case turned it red: the lookup was an object literal, so `getTranslatedFieldDescription("constructor", …)` returned `[Function Object]` where a string is declared. Converted to a `Map` — the same fix #8985 made to the same shape. (`getTranslatedFieldLabel` below has the identical hole; left alone as ~55 entries this PR does not otherwise touch, and equally unreachable.)
11. **The backfill could corrupt a non-object `config`.** `jsonb ||` against a scalar, array or JSON `null` does not fail — it concatenates as an array — so such a row would have migrated to `[null, {"areaDisplay": "line"}]` and stopped parsing as `ZChartConfig`, with the migration reporting success. Guarded on `jsonb_typeof`.

**Not changed, with reasons:**
- **Bar keeps its time grouping.** ENG-2541 item 4 names Big Number, Pie *and* Bar, and only the first two are gated. A bar chart over a time dimension is a perfectly good bar chart — a trend over time is what bar and area/line are for — whereas Big Number and Pie show a single point-in-time value or composition, which is what makes a granularity meaningless there rather than merely unusual. The reasoning was only in a code comment at `chart-display.ts:28`; it is here now so Johannes gets an answer.
- **The backfill is not batched.** The `ALTER TABLE … ALTER COLUMN TYPE` three statements later rewrites the whole table under `ACCESS EXCLUSIVE`, in the same transaction the backfill must share (or a failure between them leaves rows written against an enum that no longer holds their value). Batching would release row locks moments before a table-level exclusive lock is taken — no saving. What bounds this is `Chart` being small; that assumption is stated under `Migrations & env`.
- **`getTranslatedFieldDescription` does not fall back when i18next cannot resolve a mapped key.** `Validate Translation Keys` fails CI on a key used in code but missing from `en-US.json`, so that state is a red job rather than a user-visible string; a self-healing branch would paper over the check. It would also have to be added to `getTranslatedFieldLabel` to keep the pair consistent, which is its own change.
- **`MultiSelect` chip labels freeze at mount** (`useState(() => …)`), so a mid-session language switch leaves English chips beside translated options. Pre-existing, `multi-select` is not in this diff, and it is only reachable by changing language in-session. Flagged so it is not misread as this PR's.

## Breaking changes

None. The ENG-2541 half is UI labels/copy, an AI-prompt alias, and a builder-only option gate. The ENG-2612 half drops `line` from the `ChartType` enum, which is internal: no API v1/v2/v3 route, SDK surface, OpenAPI schema or docs page references chart types, charts are reached only through server actions and module code, and every saved `line` chart is backfilled by the migration rather than left invalid. A self-hoster runs migrations as usual and needs no other action.

## Migrations & env

- `packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area` (from #8985, now guarded) — backfills `Chart` rows with `type = 'line'` to `type = 'area'` + `config.areaDisplay = "line"`, then recreates the `ChartType` enum without `line`. Backfill and swap share one transaction, so a failure in either leaves the column and the enum consistent. The whole block is now wrapped in a `pg_enum` catalog check, so a second run and a `db:push` database are both no-ops instead of errors. Runs automatically; no manual step. The enum recreate takes a brief `ACCESS EXCLUSIVE` lock on `Chart` (a small table).
- No env changes.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `FeedbackRecords.count` measure label is "Feedback Records" | unit (guard) — `schema-definition.test.ts` | pass |
| Unique respondents / Unique responses / count descriptions render translated, state their dedup key, and (respondents) the anonymous-feedback caveat | unit (mutation) + manual | **The earlier claim here was wrong — there was no test for `getTranslatedFieldDescription` anywhere in the repo.** New `describe("getTranslatedFieldDescription")` in `schema-definition.test.ts` now covers all five described members resolving to a `field_description_*` key, plus the fallback and undefined-fallback paths. Deleting `field_description_unique_responses` from the map turns it red; before, nothing did. Measures dropdown still checked manually. |
| The chart-type migration is idempotent and convergent, and cannot corrupt a config | manual (real Postgres) | Ran against `postgres:15-alpine` with six seeded charts (two `line`, one `area`, one `bar`, plus one `line` holding `config = 'null'` and one holding `'[1,2]'`). Fresh run: the two line charts become `area` + `{"areaDisplay": "line"}`, the area chart's config is untouched, the enum ends as `area, bar, pie, big_number`. Second run: `BEGIN / DO / COMMIT`, zero rows changed. Against a database built the `db:push` way (enum already without `line`): same clean no-op. The pre-fix version, for contrast, fails **both** cases with `ERROR: invalid input value for enum "ChartType": "line"` at the backfill on line 21 — before the enum swap it was supposed to be guarding. The two non-object rows end as `{"areaDisplay": "line"}` (unguarded, they become `[null, {…}]` and `[1, 2, {…}]`), and `count(*) FILTER (WHERE jsonb_typeof(config) <> 'object')` is 0 across all six afterwards. |
| Each described member resolves to its own i18n key, and an inherited property does not | unit (mutation) | `describe("getTranslatedFieldDescription")` asserts the exact key per id, the schema fallback, an absent fallback, and that `"constructor"`/`"toString"` resolve to the fallback. Deleting a key from the map, or swapping two, turns it red; before the `Map` conversion the prototype case failed outright. |
| No "Responses" measure remains beside "Unique responses" | manual + grep | `CSAT/CES/Rating: Responses` → `: Records`, in `schema-definition.ts` and `en-US.json`; `grep` for the old labels across `apps/web`, `packages`, `docker` and `charts` returns only the locale files. `pnpm i18n` regenerated all 15 locales (ja-JP spot-checked: `CSAT: レコード数`). |
| Both deployed Cube schemas carry the current copy | inspection | `docker/` and `charts/formbricks/` schemas updated together; `schema-definition.test.ts` already asserts the two stay mirrored on membership. |
| `supportsTimeGrouping` returns false only for `big_number`/`pie` | unit (guard) — `chart-display.test.ts` | pass |
| Time-grouping toggle shows "Add time-based grouping" (Field + Granularity + Date Range) for Bar/Line/Area, and "Add a date range filter" (Field + Date Range only, no Granularity) for Big Number/Pie | manual — Playwright capture against seeded data, before/after pair below | matches; granularity control confirmed absent for Big Number/Pie |
| Chart type selector order is area, line, bar, pie, big_number | unit (guard) — `chart-types.test.ts`; screenshot below | pass |
| AI prompt has a "responses" → `FeedbackRecords.count` alias, distinct from `uniqueResponses` | unit (guard) — `ai-schema-context.test.ts` | pass |
| i18n: en-US string changes propagate to other locales | `pnpm i18n` | ran clean, 0 missing/unused keys; spot-checked de-DE/ja-JP |

**Screenshots**

Updated wordings 

Screenshot 2026-08-26 at 12 00 15 PM

Granularity hidden for pie charts and big number chart

Screenshot 2026-08-26 at 12 00 23 PM

**Open gaps**

- The migration was verified on a throwaway Postgres with a hand-built `Chart` table matching the schema, not against a restored production-shaped database. Row counts there are small (charts are per-workspace), so the `ACCESS EXCLUSIVE` lock window was not timed.
- The `Records` rename is my call, not a product decision — see review fix 7. It is three en-US strings if you want different wording.

- No new e2e spec added for the time-dimension gating — verified manually against seeded data (screenshots above) rather than via `apps/web/playwright`, since this is an option-visibility change on an existing dialog rather than a new user flow.
- Existing saved charts (if any exist in production) that already have a granularity configured on a Big Number or Pie chart keep rendering as-is; the new gating only strips a stale granularity once someone reopens that chart in the editor and saves. No data migration was run to strip it proactively.

**Risks**

- The `chart-types.ts` reorder changes only display order, not the underlying `ChartType` enum/values, so no risk to saved charts.
- The time-dimension `useEffect` cleanup in `advanced-chart-builder.tsx` strips only `granularity`, never `dateRange`/`dimension` — verified against `query-builder.ts`, which already treats a time dimension with no granularity as a valid filter-only shape, so nothing downstream assumes granularity is always present.


apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
```diff
@@ -116,6 +117,16 @@ export function AdvancedChartBuilder({
...
+  const timeDimensionSupported = supportsTimeDimension(chartType);
+
+  // Switching to a chart type that doesn't support time grouping (Big Number, Pie) drops any
+  // grouping left over from a previous type, matching how sanitizeChartDisplay drops other
+  // per-type display settings rather than saving them as dead values.
+  useEffect(() => {
+    if (!timeDimensionSupported && state.timeDimension) {
+      dispatch({ type: ACTION.SET_TIME_DIMENSION, payload: null });
+    }
+  }, [timeDimensionSupported, state.timeDimension]);
```
kiln-agent: **A time dimension without a granularity is a date-range *filter*, not grouping — this effect drops both.**

`TimeDimensionConfig` is `dimension` + optional `granularity` + `dateRange`. With no granularity it is filter-only, which is exactly what the AI is told to emit (`ai-schema-context.ts`: *"For time-based filtering (date range only, no time grouping): add a timeDimension … Do NOT include granularity"*) and what `dashboard-date-filter.ts` appends for the dashboard-level filter.

Concrete failure: user asks the AI *"how many feedback records in the last 30 days?"* → `chartType: "big_number"` with `timeDimensions: [{ dimension: "FeedbackRecords.collectedAt", dateRange: "last 30 days" }]` and no granularity. The builder mounts, `INIT_FROM_QUERY` loads it, this effect nulls it → `currentQueryJson` drifts from `lastRunQueryJsonRef` → the auto-run effect re-executes the query 500 ms later without the date range → `handleChartGenerated` replaces `chartData` → the preview silently flips from the 30-day count to the all-time count with no user action, and Save persists the widened query (`handleSaveChart` saves `chartData.query`).

Same on opening any existing Big Number / Pie chart that has a saved date range: the number changes on open. A Pie whose only grouping was the granularity collapses to a single slice.

Suggest gating the *grouping* only and keeping the range:

```suggestion
  // Switching to a chart type that doesn't support time grouping (Big Number, Pie) drops the
  // granularity, but keeps the time dimension itself: with no granularity it is the chart's
  // date-range filter (see ai-schema-context / dashboard-date-filter), not a grouping.
  useEffect(() => {
    if (!timeDimensionSupported && state.timeDimension?.granularity) {
      const { granularity: _granularity, ...rest } = state.timeDimension;
      dispatch({ type: ACTION.SET_TIME_DIMENSION, payload: rest });
    }
  }, [timeDimensionSupported, state.timeDimension]);
```


---

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
```diff
@@ -239,32 +250,34 @@ export function AdvancedChartBuilder({
...
-        customContainerClass="mt-2 px-0"
-        childrenContainerClass="flex-col gap-3 p-4"
-        childBorder>
-        <TimeDimensionPanel
-          hideTitle
-          timeDimension={state.timeDimension}
-          onTimeDimensionChange={(config) => dispatch({ type: ACTION.SET_TIME_DIMENSION, payload: config })}
-        />
-      </AdvancedOptionToggle>
+      {timeDimensionSupported && (
```
kiln-agent: Hiding the whole toggle removes **date scoping** for Big Number and Pie, not just bucketing. `TimeDimensionPanel` renders three controls — time field, granularity (with an explicit *"No grouping"* entry), and date range presets — so after this change *"Feedback records in the last 7 days"* as a Big Number is no longer buildable at all.

The filters panel is not a substitute: `FILTER_OPERATORS.time` is `equals/notEquals/gt/gte/lt/lte/set/notSet` against absolute dates, so a rolling window (`last 7 days`, `this month`) cannot be expressed there.

Consider keeping the toggle rendered for every chart type and hiding only the granularity select inside `TimeDimensionPanel` when `!supportsTimeDimension(chartType)` — the title/description copy ("Add time-based grouping" / "Monitor trends over time") would then need a range-only variant.

---

apps/web/modules/ee/analysis/lib/schema-definition.ts
```diff
@@ -244,24 +244,26 @@ export const FEEDBACK_FIELDS = {
...
+        "Unique people who gave feedback, deduplicated by person — one respondent answering 3 questions counts once",
     },
     {
       id: "FeedbackRecords.uniqueResponses",
       label: "Unique responses",
       type: "number",
       group: "count",
-      description: "Number of unique survey submissions",
+      description:
+        "Unique survey submissions, deduplicated by submission — one respondent submitting twice counts twice",
```
kiln-agent, line 266: These two descriptions are user-facing copy rendered verbatim: `measures-panel.tsx` maps `description: m.description` — it never goes through `getTranslatedFieldDescription` (which only covers `valueId`/`valueText`). So a de-DE user opening the measures dropdown gets the translated label next to an English sentence, and `pnpm i18n` can't pick the string up because it never reaches `en-US.json`.

The label change in this same hunk did add an en-US key (`field_label_count`), so the two are now inconsistent. Worth adding `field_description_unique_respondents` / `field_description_unique_responses` keys and extending the map in `getTranslatedFieldDescription` (plus using it from `measures-panel`) while you're touching this copy.

---

kiln-agent (commented):
Code review (high effort) — 3 findings, one of which I think is a real data-scope regression.

The time-dimension gate treats the whole `timeDimension` as "grouping", but in this codebase a `timeDimension` with no `granularity` is the chart's **date-range filter** (that is what `ai-schema-context.ts` instructs the model to emit, and what `dashboard-date-filter.ts` appends). Dropping it for `big_number`/`pie` therefore silently widens those charts to all-time on open — and hiding the panel removes the only way to scope a Big Number to a period at all, since time filters in the filters panel only take absolute dates. Details inline.

---

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
kiln-agent: This drops the whole `timeDimension`, but a time dimension **without granularity is the chart's date-range filter**, not a grouping — `applyDashboardDateFilter` calls exactly that shape "a filter-only time dimension (no granularity)" and appends it to any chart, big_number included.

Concrete failure: a saved Big Number with `{measures:["FeedbackRecords.count"], timeDimensions:[{dimension:"FeedbackRecords.collectedAt", dateRange:"last 30 days"}]}` (buildable today) is opened in the edit dialog. `chartData.query` seeds `state.timeDimension` via `INIT_FROM_QUERY`, this effect immediately dispatches `SET_TIME_DIMENSION: null`, `currentQueryJson` then drifts from `lastRunQueryJsonRef`, and the 500 ms auto-run re-executes without the `dateRange`. The preview silently switches from the 30-day count to the all-time count with no UI to explain it (the toggle is hidden), and Save — even for someone who only came to rename the chart — persists the stripped query through `updateChartAction`.

Gating on `granularity` instead of on the whole time dimension keeps the date scoping while still removing the time bucketing that ENG-2541 flagged.

---

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
kiln-agent: The AI path still produces the queries this gate destroys. `generateSchemaContext()` (`ai-schema-context.ts`) tells the model: *"For time-based filtering (date range only, no time grouping): add a timeDimension with dimension `FeedbackRecords.collectedAt` and dateRange. Do NOT include granularity"* and *"big_number (for single-number queries)"*.

So: user types "How many responses in the last 30 days?" in `AIQuerySection` → `generateAIChartAction` returns `chartType: "big_number"` + `timeDimensions:[{dimension: collectedAt, dateRange: "last 30 days"}]` → `handleChartGenerated` stores it → the builder receives it as `initialQuery` → the new effect strips it → the auto-run replaces the answer with an all-time count. The chart the user saves answers a different question than the one they asked, with no warning.

Whatever the resolution on the comment above, the prompt and the gate need to agree — either stop the model from emitting time dimensions for `big_number`/`pie`, or keep the date-range control for them.

---

apps/web/modules/ee/analysis/charts/lib/chart-display.ts
```diff
@@ -13,6 +13,15 @@ export const DEFAULT_PIE_DISPLAY: TPieDisplay = "pie";
...
 export const supportsPieDisplay = (chartType: TChartType | undefined): boolean => chartType === "pie";
 
+/**
+ * Big Number shows a single snapshot value and Pie shows composition at a point in time — bucketing
+ * either by date range/granularity produces a chart that no longer answers the question its type
+ * implies, which is what ENG-2541 flagged as confusing. Bar and Line/Area keep it: a trend over time
+ * is exactly what those types are for.
+ */
+export const supportsTimeDimension = (chartType: TChartType | undefined): boolean =>
+  chartType !== "big_number" && chartType !== "pie";
```
kiln-agent, line 32: `TimeDimensionPanel` owns two different things: the **granularity** select (the actual time bucketing, defaults to "None") and the **date range** select (a filter). Gating the whole panel on chart type removes date scoping from Big Number and Pie altogether, not just the bucketing.

After this change there is no way to build "Feedback Records collected in the last 30 days" as a Big Number: the date-range select is the only control that emits `dateRange`, and `FiltersPanel` can't stand in — `FILTER_OPERATORS.time` is `equals/notEquals/gt/gte/lt/lte/set/notSet`, none of which expresses a rolling preset. Meanwhile `applyDashboardDateFilter` injects a `collectedAt` `dateRange` into that very same big_number chart when it sits on a dashboard, so builder and dashboard now disagree about whether a Big Number may be date-scoped.

Suggest keeping the panel for all types and gating only the granularity select (`supportsTimeGrouping`), which is the thing that actually makes a Big Number or Pie stop answering its own question.

---

apps/web/modules/ee/analysis/lib/schema-definition.ts
```diff
@@ -244,24 +244,26 @@ export const FEEDBACK_FIELDS = {
...
+      description: "Total number of feedback records",
     },
     {
       id: "FeedbackRecords.uniqueRespondents",
       label: "Unique respondents",
       type: "number",
       group: "count",
-      description: "Number of unique users who provided feedback",
+      description:
+        "Unique people who gave feedback, deduplicated by person — one respondent answering 3 questions counts once",
```
kiln-agent: The new copy nails the dedup key but keeps the part that is arguably the bigger trap: this measure is `type: countDistinct, sql: user_id` (`docker/cube/schema/FeedbackRecords.js:12-16`), and SQL `COUNT(DISTINCT x)` skips NULLs — records with no user id contribute nothing.

For a link survey with 500 anonymous records the dropdown promises "Unique people who gave feedback" and the chart renders `Feedback Records = 500, Unique respondents = 0`, which reads as a bug. Since this PR exists to remove exactly this class of confusion, worth saying it counts *identified* respondents (feedback without a user id isn't counted).

---

apps/web/modules/ee/analysis/lib/schema-definition.ts
```diff
@@ -244,24 +244,26 @@ export const FEEDBACK_FIELDS = {
   measures: [
     {
       id: "FeedbackRecords.count",
-      label: "Responses",
+      label: "Feedback Records",
```
kiln-agent, line 247: Worth checking the AI side of this rename: `ai-schema-context.ts` renders every measure into the system prompt as `- <id>: <label> (<description>)`, so after this change no measure is labelled "Responses" any more — the only label still containing the word is `FeedbackRecords.uniqueResponses` ("Unique responses"), a different measure, and the "Metric aliases" section covers NPS/CSAT/CES/rating/sentiment but says nothing about responses.

"Show me responses per week" used to hit `FeedbackRecords.count` on an exact label match; now the model can plausibly land on `uniqueResponses` (`countDistinct(submission_id)`) and the chart under-reports — a submission answering 5 questions counts once instead of five. One alias line in the prompt (`"responses" / "response count" means FeedbackRecords.count`) would pin it.

---

apps/web/modules/ee/analysis/lib/schema-definition.ts
kiln-agent, line 266: These two reworded descriptions are user-facing but never go through `t()`. `MeasuresPanel` maps the option as `label: getTranslatedFieldLabel(m.id, t)` + `description: m.description` — there's no measure-side equivalent of `getTranslatedFieldDescription`, so the raw English string is what renders.

Result for a de-DE user: a German label with an English clarification directly beneath it, i.e. the fix this PR is making doesn't land in any non-English locale. AGENTS.md requires user-facing text to come from `t()` with the string added to `en-US.json` (the i18n table in the PR body only covers `field_label_count`). Either add `field_description_unique_respondents` / `field_description_unique_responses` keys and extend the description lookup to measures, or note the gap explicitly.

---

kiln-agent (commented):
Code review of the diff (6 findings, inline).

The headline issue is that the new `supportsTimeDimension` gate treats the time-dimension panel as pure "time grouping", but that panel also owns the chart's **date range** — and a `timeDimension` without `granularity` is precisely what the codebase elsewhere calls a filter-only time dimension (`applyDashboardDateFilter`). So the gate plus the cleanup effect don't just hide bucketing for Big Number/Pie, they strip date scoping from existing charts on open (and persist that on save), remove any way to build a date-scoped Big Number/Pie, and quietly mutilate AI-generated queries that the prompt still instructs the model to produce. Gating on `granularity` rather than the whole time dimension would keep the intended UX fix without any of that.

The remaining three are smaller: copy accuracy on Unique respondents, the AI prompt losing its "Responses" anchor after the label rename, and the new descriptions being English-only.

---

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
Dhruwang: Good catch — fixed in 2c7df7dd7e. Now stripping only `granularity` (kept the toggle rendered, copy switches to "Add a date range filter" for Big Number/Pie) instead of nulling the whole timeDimension.

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
Dhruwang: Fixed in 2c7df7dd7e — the toggle now stays rendered for every chart type; only the granularity control inside TimeDimensionPanel is hidden for Big Number/Pie (via the new `hideGranularity` prop), so the date-range filter is still buildable there.

apps/web/modules/ee/analysis/lib/schema-definition.ts
Dhruwang: Fixed in 2c7df7dd7e — moved these into en-US.json (field_description_unique_respondents/unique_responses/count) and wired them through getTranslatedFieldDescription, same path measures-panel already used for labels.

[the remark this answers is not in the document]
Dhruwang: Fixed in 2c7df7dd7e — moved into en-US.json (field_description_count/unique_respondents/unique_responses) and threaded through getTranslatedFieldDescription; measures-panel.tsx now calls that instead of reading m.description raw.

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
Dhruwang: Duplicate of the same finding above — fixed in 2c7df7dd7e (granularity-only strip, toggle stays rendered).

apps/web/modules/ee/analysis/charts/components/advanced-chart-builder.tsx
Dhruwang: Same fix covers this — 2c7df7dd7e strips only granularity, so an AI-generated range-only timeDimension (no granularity) now survives untouched for big_number/pie.

apps/web/modules/ee/analysis/charts/lib/chart-display.ts
Dhruwang: This is what 2c7df7dd7e does — TimeDimensionPanel now always renders for every chart type; only the granularity select is hidden via the new hideGranularity prop (driven by supportsTimeGrouping), so the date-range control stays available for Big Number/Pie.

apps/web/modules/ee/analysis/lib/schema-definition.ts
Dhruwang: Good catch, real gap — fixed in ad4e4c65a5. Updated the description to say it counts identified respondents only, since COUNT(DISTINCT user_id) skips anonymous feedback.

apps/web/modules/ee/analysis/lib/schema-definition.ts
Dhruwang: Fixed in ad4e4c65a5 — added a Metric aliases line pointing "responses"/"response count"/"feedback records" at FeedbackRecords.count and explicitly away from uniqueResponses, same pattern as the existing NPS/CSAT/CES aliases. Added test coverage in ai-schema-context.test.ts.

apps/web/modules/ee/analysis/lib/schema-definition.ts
Dhruwang: Fixed in 2c7df7dd7e — moved into en-US.json and wired through getTranslatedFieldDescription; measures-panel.tsx now calls that instead of reading m.description raw.

xernobyl:
One thing on the coverage table: the row for the reworded measure descriptions cites "existing `schema-definition.test.ts` coverage of `getTranslatedFieldDescription`", but I can't find a test for that function anywhere — grepping every `*.test.ts`/`*.test.tsx` in the repo turns up zero references to it, and the two changed test files only touch the `label` assertions.

So the thing the earlier review rounds flagged (a measure description bypassing i18n) is genuinely fixed, it's just unguarded. If someone later drops or typos one of the three keys in the descriptions map, every non-English locale silently falls back to the hardcoded English string, `pnpm i18n` still passes because the en-US key is still there, Validate Translation Keys passes, and the suite stays green.

Should only be a few lines in `schema-definition.test.ts` — the three new ids resolving through `t()`, plus an unmapped id still hitting the fallback. Everything else I checked holds up, so happy to approve once that's in 🙏

(Separately, and not blocking: the docblock above `getTranslatedFieldDescription` still says it's for "the dimensions whose copy guides a chart-building decision" — it covers three measures now.)

xernobyl (commented):
Reviewed against `ad4e4c65a`. The code itself is in good shape — the four review fixes landed and I confirmed each one works rather than taking the description's word for it. One thing does need fixing before this merges, and two scope questions are worth a sentence each.

## Needs fixing

**The coverage table claims a test that doesn't exist.** The row for the description changes says *"unit (guard) — existing `schema-definition.test.ts` coverage of `getTranslatedFieldDescription`"*. There's no test anywhere for `getTranslatedFieldDescription`, nothing references the new `field_description_*` keys, and `charts/components/` has no test files at all — so `measures-panel.tsx` is uncovered too.

That matters more than a bookkeeping slip, because review fix 2 exists precisely so these descriptions reach non-English locales. Right now you can delete a line from the `descriptions` map in `schema-definition.ts:661` and every suite still passes, while every non-English user silently falls back to the English string — the exact bug the fix was for. It works today (I checked, see below), but nothing holds it there.

A three-line test asserting the three ids resolve to `workspace.analysis.charts.field_description_*` would close it, in the same shape as the existing `getTranslatedFieldLabel` assertion at `schema-definition.test.ts:250`.

## Scope questions

**Bar.** ENG-2541 item 4 names Big Number, Pie *and Bar*; only the first two are gated. The reasoning is in the code comment at `chart-display.ts:28` ("a trend over time is exactly what those types are for") and I agree with it — but the description doesn't mention Bar at all, so Johannes, who asked the question, gets no answer. Worth a line saying it was considered and kept, rather than leaving it to be inferred from a code comment.

**The rename is partial.** `FeedbackRecords.count` is now "Feedback Records", but `CSAT: Responses`, `CES: Responses` and `Rating: Responses` still sit in the same dropdown right next to it and next to "Unique responses". Possibly deliberate — "CSAT: Feedback Records" reads badly — but if so that's worth stating, because item 2 exists to remove exactly this ambiguity and a user opening the measures list still meets three "Responses".

**Nit:** `docker/cube/schema/FeedbackRecords.js:15,21` still carries the old descriptions, now drifted from the copy this PR rewrote. Invisible while the app reads `FEEDBACK_FIELDS`, but it's a second source of truth that just went stale.

## What I checked

- **The tests that exist can fail.** Four mutations, four reds: inverting `supportsTimeGrouping`, reverting the count label, dropping the AI alias line, un-swapping the line/bar order.
- **The UI behaves as the screenshots show.** Mounted the real `TimeDimensionPanel` and `MeasuresPanel` at this head: selector order is `area → line → bar → pie → big_number`, and the granularity control is absent for Pie/Big Number while Field and Date Range stay.
- **Review fix 2 lands.** Switched the same harness to `ja-JP` and the new respondent/response descriptions render in Japanese.
- **The anonymous-respondent copy is accurate.** `docker/cube/schema/FeedbackRecords.js:12` is `countDistinct(user_id)`, so `COUNT(DISTINCT)` skipping NULLs does mean anonymous feedback contributes 0. Good catch to put that in the description.
- **No auto-run race.** I expected the type switch to fire one query with the stale granularity before the strip effect ran, but the auto-run is behind `AUTO_RUN_DEBOUNCE_MS`, so granularity is gone well before any request goes out. Non-issue.
- Typecheck is clean in the analysis module; the whole `modules/ee/analysis` suite passes.

## Security

Nothing here touches auth, tenant scoping, queries or dependencies. Two things I checked rather than assumed: `getTranslatedFieldDescription` does `descriptions[id] ?? fallback`, the same object-literal lookup #8985 had to convert to a `Map` so `"constructor"` couldn't resolve through the prototype chain — not reachable here, since all four call sites pass ids from the hardcoded `FEEDBACK_FIELDS` arrays and `formatCubeColumnHeader` gates on `getFieldById` first. And the new AI alias interpolates only `CUBE_NAME`, so no user input reaches the system prompt.

Worth stating once so it isn't mistaken later: `supportsTimeGrouping` is a UI gate, not validation — the AI path or a hand-built query can still send a granularity for a Big Number.

## Unrelated, flagging so it isn't misread as yours

`MultiSelect` seeds its selected chips with `useState(() => …)`, so chip labels freeze at mount and a mid-session language switch leaves English chips beside translated options. Pre-existing, `multi-select` isn't in this diff, and it's only reachable by changing language in-session.

---

Everything CodeRabbit and kiln-agent raised is fixed at head — I checked each against the current code rather than the description, and nothing bot-flagged is still open.

Dhruwang (commented):
All four taken in 1507a3ea70. Thanks for grepping rather than trusting the table — the missing test was the one that mattered, and you were right about why.

**The coverage table claimed a test that doesn't exist.** It did not, anywhere: no `*.test.ts` referenced `getTranslatedFieldDescription`, and I had written that row from the adjacent `getTranslatedFieldLabel` coverage without checking. Fixed, and the table now says so rather than quietly correcting itself.

New `describe("getTranslatedFieldDescription")` in `schema-definition.test.ts`, in the same shape as the `getTranslatedFieldLabel` assertion at line 250. It covers all five described members — not just the three measures, since `valueId`/`valueText` went through the same map and were equally unguarded — plus the fallback path and the undefined-fallback path. Verified it can fail: deleting `field_description_unique_responses` from the map turns it red, where before every suite stayed green.

**Bar.** Kept, and the reasoning is in the description now instead of only in the code comment. A bar chart over a time dimension is a perfectly good bar chart — a trend over time is what bar and area are for — whereas Big Number and Pie show a single point-in-time value or composition, which is what makes a granularity meaningless there rather than merely unusual. So it was considered and deliberately excluded, and Johannes has an answer rather than an inference.

**The rename is partial.** Not deliberate — you found a real gap, and it is worse than cosmetic. `CSAT: Responses`, `CES: Responses` and `Rating: Responses` count *feedback records* (one per answered question), so besides sitting in the same dropdown as "Unique responses", they are wrong on their own terms: a survey with two rating questions reports two "Rating: Responses" for one response.

I went with `CSAT: Records` / `CES: Records` / `Rating: Records` rather than `CSAT: Feedback Records`, which you're right reads badly at that length beside `CSAT: Average` and `CSAT: Neutral`. "Records" is the short form of the vocabulary this PR just established, so it does not introduce a third noun the way "Answers" would. Their descriptions now say "feedback records" too. `pnpm i18n` regenerated all 15 locales. This one is my call rather than a product decision, so say the word if you'd rather have different wording — it is three en-US strings and a regeneration.

**The Cube schema nit.** Taken, and in both copies: `charts/formbricks/cube/schema/FeedbackRecords.js` had drifted identically, and `schema-definition.test.ts` already asserts the two stay mirrored, so fixing one would have left the other as the stale source of truth. Realigned `count`, `uniqueRespondents`, `uniqueResponses`, `valueText` and `valueId` with the app's copy.

**The docblock.** Fixed — it now says "two dimensions and the three count measures" rather than "the dimensions".

**On the `MultiSelect` chip freeze** — agreed it is not this PR's, and thank you for flagging it rather than filing it against this diff. I have left it alone; it wants its own ticket.

Two notes on what you saw at `ad4e4c65a`: #8985 has since merged into this branch, so the enum migration is part of this diff now, and the description's `Migrations & env` and `Breaking changes` sections cover it. CodeRabbit's idempotency finding on that migration was real and is fixed — details in that thread.

packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+-- Backfill: every line chart becomes an area chart displayed as a line.
+UPDATE "public"."Chart"
+SET "type"   = 'area',
+    "config" = "config" || '{"areaDisplay": "line"}'::jsonb
+WHERE "type" = 'line';
+CREATE TYPE "public"."ChartType_new" AS ENUM ('area', 'bar', 'pie', 'big_number');
+ALTER TABLE "public"."Chart" ALTER COLUMN "type" TYPE "public"."ChartType_new" USING ("type"::text::"public"."ChartType_new");
+ALTER TYPE "public"."ChartType" RENAME TO "ChartType_old";
+ALTER TYPE "public"."ChartType_new" RENAME TO "ChartType";
+DROP TYPE "public"."ChartType_old";
```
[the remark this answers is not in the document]
Dhruwang: Valid, and fixed in 1507a3ea70 — with one correction to the diagnosis that changed how the guard is written.

You said it fails on the enum swap ("`ChartType_new` cannot be created again"). It fails earlier than that: on line 21, at the *backfill*. Once `line` is not a member, `WHERE "type" = 'line'` cannot even be parsed —

```
ERROR:  invalid input value for enum "ChartType": "line"
LINE 4: WHERE "type" = 'line';
```

That matters because the operator-visible symptom is not "this migration already ran", it is an enum coercion error on a table of user data, which reads as corruption. So the guard goes around the whole block rather than around each DDL statement: a single `pg_enum` lookup for the `line` label, `RETURN` if it is gone. Same convergence, one catalog query, and the intent — "there is nothing here to do" — is stated once instead of being spread over four `IF EXISTS` clauses.

It follows the house pattern already in `20260812120000_add_feedback_source_element_scope`, which guards its `CREATE TYPE` for exactly this reason.

Verified against a real Postgres rather than by reading, in all four states, on `postgres:15-alpine` with four seeded charts (two `line`, one `area`, one `bar`):

| | old migration | guarded migration |
| --- | --- | --- |
| fresh database with `line` charts | applies | applies — line charts become `area` + `{"areaDisplay": "line"}`, the existing area chart's config untouched, enum ends as `area, bar, pie, big_number` |
| same database, run again | `ERROR: invalid input value for enum "ChartType": "line"` | `BEGIN / DO / COMMIT`, zero rows changed |
| database created with `db:push` | same error | same clean no-op |

On the Squawk warning about the `ACCESS EXCLUSIVE` lock: not addressed, deliberately. It is inherent to a Postgres enum-value removal, `Chart` is a small per-workspace table, and the guard now means the rewrite happens once rather than on every attempted re-run.

apps/web/modules/ee/analysis/lib/schema-definition.ts
```diff
@@ -337,10 +339,10 @@ export const FEEDBACK_FIELDS = {
     },
     {
       id: "FeedbackRecords.csatCount",
-      label: "CSAT: Responses",
+      label: "CSAT: Records",
       type: "count",
       group: "count",
-      description: "Number of CSAT responses",
+      description: "Number of feedback records from CSAT questions",
```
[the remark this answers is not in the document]
Dhruwang: Correct, fixed in 5f5ece4922 — and it applies to `ratingCount` too, which I had also reworded.

Verified the filters you cite: all three are `field_type = '<x>' AND value_number IS NOT NULL` (`docker/cube/schema/FeedbackRecords.js:69, 117, 130`, mirrored in `charts/formbricks/`), and the Cube-side descriptions say "answered … (dismissed responses excluded)". My rewrite kept the exclusion only for `ratingCount` and dropped it for the other two, which was an inconsistency rather than a decision.

All three app-side descriptions now read "Number of answered feedback records from <X> questions (dismissed excluded)", which keeps both facts the rename was for: they count *records*, not submissions, and unanswered ones are not among them.

These three are not in the `getTranslatedFieldDescription` map, so they are the inline English in `FEEDBACK_FIELDS` and need no locale regeneration.

apps/web/modules/ee/analysis/lib/schema-definition.test.ts
```diff
@@ -281,6 +282,40 @@ describe("schema-definition", () => {
...
+        "FeedbackRecords.valueText",
+        "FeedbackRecords.count",
+        "FeedbackRecords.uniqueRespondents",
+        "FeedbackRecords.uniqueResponses",
+      ];
+
+      for (const id of describedIds) {
+        expect(getTranslatedFieldDescription(id, "english fallback", t)).toMatch(
+          /^workspace\.analysis\.charts\.field_description_/
+        );
```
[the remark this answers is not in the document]
Dhruwang: Half taken, half declined — and writing the stricter version found a real bug, so thank you for pushing on it.

**Taken: exact assertions.** You are right that a key-echoing `t` let a wrong id→key mapping pass. Each id now asserts its exact key through a `test.each` table, with `t` returning `translated:${key}` so an assertion cannot be satisfied by the key alone. Fixed in 5f5ece4922.

**Found while doing it:** I added a case for an id that collides with a prototype member, expecting it to be a no-op. It failed — `getTranslatedFieldDescription("constructor", …)` returned `[Function Object]`, because the lookup was an object literal and `descriptions[id]` resolves inherited members. That is the same shape #8985 had to convert, so this one is now a `Map` too. Not reachable from today's call sites (every one passes an id from the hardcoded `FEEDBACK_FIELDS`), but it returned a function where a string was declared, and closing it cost three lines.

**Declined: making the helper return `fallback` when i18next cannot resolve a mapped key.** Two reasons.

That failure cannot reach production silently. `Validate Translation Keys` runs on every PR and fails on a key used in code but missing from `en-US.json` — a mapped key that i18next cannot resolve is a red CI job, not a user-visible string. Adding a self-healing branch would make the helper paper over exactly the condition that check exists to surface.

And it would only paper over half of it: `getTranslatedFieldLabel` sits directly below with the identical shape and no such guard, so the helper pair would disagree about what an unresolvable key means. If we want that behaviour it belongs to both, as a deliberate change, not as a side effect of a test.

On that note — `getTranslatedFieldLabel` has the same object-literal lookup and the same prototype-chain hole. I left it alone here: ~55 entries this PR does not otherwise touch, and equally unreachable. Happy to file it as a follow-up if you'd rather not leave the pair inconsistent.

packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area/migration.sql
```diff
@@ -0,0 +1,50 @@
...
+      AND e.enumlabel = 'line'
+  ) THEN
+    RETURN;
+  END IF;
+
+  -- Backfill: every line chart becomes an area chart displayed as a line. `config` is
+  -- `Json @default("{}")` and never null, so `||` cannot blank it out.
+  UPDATE "public"."Chart"
+  SET "type"   = 'area',
+      "config" = "config" || '{"areaDisplay": "line"}'::jsonb
```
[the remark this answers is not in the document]
Dhruwang: Confirmed and fixed in 5f5ece4922 — and it is worse than the comment says, which is what makes it worth the guard.

I expected `||` against a non-object to raise, so the failure mode would at least be loud. It does not:

```
=> SELECT 'null'::jsonb || '{"a":1}'::jsonb, '"x"'::jsonb || '{"a":1}'::jsonb, '[1,2]'::jsonb || '{"a":1}'::jsonb;
 [null, {"a": 1}] | ["x", {"a": 1}] | [1, 2, {"a": 1}]
```

Postgres concatenates as an array. So a `line` chart holding a scalar, an array or JSON `null` would migrate to a `config` that is an array — no error, migration reports success, and the chart stops parsing as `ZChartConfig` at read time, long after the migration is the obvious suspect.

The backfill now branches on `jsonb_typeof("config") = 'object'`. A non-object gets `'{"areaDisplay": "line"}'` outright rather than a merge: the value being replaced was already unusable as a config, so the choice is between a valid config carrying the display style and a corrupt one, not between old and new data.

Re-verified on `postgres:15-alpine` with the fixture extended by two rows — one `config = 'null'`, one `config = '[1,2]'`:

| | guarded | unguarded |
| --- | --- | --- |
| `'null'` config | `{"areaDisplay": "line"}` (object) | `[null, {"areaDisplay": "line"}]` (array) |
| `'[1,2]'` config | `{"areaDisplay": "line"}` (object) | `[1, 2, {"areaDisplay": "line"}]` (array) |
| object configs | merged, unchanged otherwise | merged |

After the run, `count(*) FILTER (WHERE jsonb_typeof(config) <> 'object')` is 0 across all six rows, and a second run is still a clean no-op.

Worth saying plainly: every row the app writes goes through `ZChartConfig`, so this should match nothing in practice. It is a guard against a shape the column does not constrain, not a known bad state.

packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area/migration.sql
```diff
@@ -0,0 +1,50 @@
...
+  WHERE "type" = 'line';
```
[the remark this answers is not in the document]
Dhruwang: Not taking this one, and the reason is specific to what this migration does immediately after the `UPDATE`.

Batching would buy nothing, because the locks it is meant to release are taken anyway, in the same transaction, three statements later:

```sql
ALTER TABLE "public"."Chart" ALTER COLUMN "type" TYPE "public"."ChartType_new" USING (...)
```

A column type change rewrites the whole table under `ACCESS EXCLUSIVE` — Squawk flags exactly this on line 23 of the original. So the transaction holds a table-level exclusive lock regardless of how the backfill was written, and row locks from an unbatched `UPDATE` are strictly contained by it. Splitting the `UPDATE` into primary-key batches would add loop machinery whose only effect is to release row locks a moment before a lock that blocks the entire table is taken.

Batching is the right call when a long `UPDATE` is the *whole* migration and can be committed incrementally. It cannot be here: the backfill and the enum swap must share one transaction, or a failure between them leaves rows already rewritten against an enum that no longer has the value they hold — which is the reasoning in the header comment, and the thing that makes this safe to re-run at all.

What actually bounds the exposure is the table's size, and `Chart` is small: one row per saved chart per workspace, on a feature that is EE and new. If it were large, the fix would not be batching the backfill — it would be avoiding the whole-table rewrite, e.g. `ALTER TYPE ... ADD VALUE` semantics or a two-deploy column swap, which is a much bigger change than this ticket.

Recorded rather than dismissed: `Migrations & env` in the description already names the `ACCESS EXCLUSIVE` lock and the small-table assumption, so a self-hoster is not finding out at run time.

xernobyl (approved):
Approving — this holds up.

What I checked beyond reading the diff: ran the migration against a real Postgres seeded with the awkward cases (configs holding `null`, `[1,2]`, a scalar, and a `line` chart that already had an `areaDisplay`) — fresh run, second run, and against a `db:push`-shaped database where the enum never had `line`; second run is byte-identical state, both no-op cases are clean. Also applied the full migration history to an empty database and diffed the result against `main.prisma`: no `ChartType` drift, and no ordering conflict with what's on main.

For the line/area merge I rendered the base commit's renderer alongside this one and diffed the SVG. Identical in both styles once you normalise attribute order and Recharts' id counter — every painting attribute (`d`, `fill`, `fill-opacity`, `stroke`) matches. The one thing that isn't in the static SVG is `activeDot={undefined}` on the filled style; that still falls through to Recharts' `true` default (`Area` is a class component, and React 19 only dropped `defaultProps` for function components), so no hover-dot regression. Gate behaves across all four types, and the granularity-stripping effect converges rather than looping.

I also mutation-checked the new tests — dropping a description key, swapping two of them, reverting the `Map` to an object literal, forcing `supportsTimeGrouping` true, and removing the AI alias each turn the suite red. They earn their place.

Three small things, none blocking, all yours to take or leave:

**1. The Cube descriptions drifted again for the three measures fix 9 reworded.** `docker/cube/schema/FeedbackRecords.js:70,118,131` (and the same lines in `charts/formbricks/`) still read *"Number of answered CSAT responses (dismissed responses excluded)"*, while `schema-definition.ts` now says *"Number of answered feedback records from CSAT questions (dismissed excluded)"*. Fix 8 realigned five members for exactly this reason; these three then moved app-side only. Worth noting the mirroring test only asserts membership and byte-equality between the two Cube files, so nothing would catch a description divergence — which is probably why it slipped.

**2. The Responses → Records rename stops just short in the same dropdown.** `csatSatisfiedCount`, `csatDissatisfiedCount` and `csatNeutralCount` (`schema-definition.ts:324,331,338`) still read "Number of satisfied/dissatisfied/neutral CSAT responses", and `csatScore:309` / `npsScore:273` describe their denominator as "responses". All three counts are `type: count` filtered on `field_type = 'csat'` — record counts, same shape as the `csatCount` you renamed — so the "wrong twice over" reasoning in fix 7 looks like it lands on these too. Entirely possible you drew the line at labels deliberately and left descriptions for later; if so ignore me.

**3. Two spots in the description went stale when #8985 merged in.** The "Move Line next to Area" bullet and the coverage row both still say the order is `area, line, bar, pie, big_number`. There's no `line` entry any more and `chart-types.test.ts` asserts `["area", "bar", "pie", "big_number"]`. Only worth fixing because the coverage table is the thing a reviewer reads as the claim.

One observation rather than a request: `chart-renderer.tsx` and `chart-display-settings.tsx` have no test file anywhere, and they're the two files this PR changes most. The parity check I ran was throwaway, so nothing in the repo would catch a regression in that merge later. Feels like a follow-up ticket rather than something to grow this PR with.

PR #8975, opened 2026-08-25
fix: remove non-functional single-response-per-email feature (ENG-2294)

jobenjada:
Fixes ENG-2294

## What & why

`isSingleResponsePerEmailEnabled` ("Prevent double submission" in the editor) was a UI-only deterrent, never enforced at the API — it caught casual double submissions but never closed the actual gap:

- The only runtime read of the field was in `verify-email.tsx`'s `submitEmail`, which called `isSurveyResponsePresentAction` to block **sending a second verification email** to an address that already had a response. That's it — a replayed verify token, or skipping the verify-email step entirely, was never blocked.
- The actual submission endpoints (`POST /api/v1/client/.../responses`, and the v2 equivalent) never checked this field. Nothing on the `Response` model enforces uniqueness by email. The verification JWT is stateless and reusable, so even with both "Verify email" and "Prevent double submission" turned on, resubmitting the same verified link (or skipping straight past the email step) always worked.
- The toggle was also structurally nested inside "Verify Email Before Submission" in the editor UI, so it couldn't even be turned on unless email verification was already enabled — making it easy to miss even as a no-op.

This surfaced as [ENG-2294](https://linear.app/formbricks/issue/ENG-2294/security-issingleresponseperemailenabled-never-enforced-on-response), and was independently confirmed while reviewing the feature: the "Only allow 1 response per email address" copy promises protection the code never delivered. Rather than wire up real enforcement for a control that customers may already believe protects them (a false sense of security is worse than an obviously-missing feature), we're sunsetting it: removing the toggle, the dead pre-send check, the DB column, and all related plumbing.

Note: `singleUse` survey links are a separate, genuinely-enforced mechanism (unique constraint on `Response`, 409 on reuse) and are unaffected by this change.

## Breaking changes

| `isSingleResponsePerEmailEnabled` field removed from Survey resource | Management API v2 survey create/read/update accepted and returned this boolean field | Field no longer exists on the resource; requests that set it will simply have it ignored (schema drops the key) | API v2 consumers reading or setting this field | Stop reading/writing `isSingleResponsePerEmailEnabled` in integrations. It never enforced anything server-side, so no behavioral migration is needed — only remove references to the field itself. |
| `isSingleResponsePerEmailEnabled` field removed from Survey resource | Management API v1 (`POST /management/surveys`, `GET`/`PUT /management/surveys/{id}`) accepted and returned this boolean field via `ZSurveyCreateInput`/`ZSurveyUpdateInput` | Same as above — field no longer exists on the resource | API v1 consumers reading or setting this field | Same as above |

## Migrations & env

- New Prisma migration `20260825000000_remove_is_single_response_per_email_enabled_from_survey` drops the `isSingleResponsePerEmailEnabled` column from `Survey`. Applied automatically via the standard migration deploy step; no manual action needed. Any existing value on that column (which never had a functional effect) is discarded.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Field fully removed from schema/zod/types/API surface, no dangling references | `pnpm build` (root), `tsc --noEmit` for `@formbricks/web`, `@formbricks/database`, `@formbricks/types` | unit (guard) — 0 type errors across the workspace |
| Rate-limit config no longer lists the removed `isSurveyResponsePresent` action | `apps/web/modules/core/rate-limit/rate-limit-configs.test.ts` | unit (mutation) — updated assertion, fails if the entry reappears |
| Existing survey create/read/update, editor settings, link-survey rendering, email verification flow all still work with the field gone | Full existing suite: `pnpm test` (root) | unit (guard) — 8486/8486 tests, 640/640 files pass |
| i18n keys (`prevent_double_submission*`, `s.response_already_received`) removed cleanly from all 15 locales | `pnpm i18n` | unit (guard) — 0 missing/unused keys, all locales complete |
| Formatting/lint stay clean after removal | `pnpm format:check`, `pnpm lint` (root) | unit (guard) — clean, only pre-existing unrelated warnings |
| `docs/api-v2-reference/openapi.yml` no longer documents the removed field or lists it as `required` | manual diff review; `js-yaml` parse check that the file is still valid YAML | manual — field and its `required` entry removed (was missed by CI; nothing validates the spec against the live schema) |
| "Verify email before submission" no longer has a nested "Prevent double submission" toggle in the survey editor's Response Options card (link surveys, Settings tab) | manual — live capture via a throwaway Playwright spec against the real dev server (fixture-seeded link survey, `#verifyEmailBeforeSubmission` toggled on to expose the nested-child region) | Nested toggle and its `border`/`bg-slate-50` child container are gone; "Verify email before submission" now renders as a single self-contained row. See screenshots below. |

<details>
<summary>Screenshots</summary>

| Before | After |
| --- | --- |
| Verify email before submission toggle with the nested Prevent double submission toggle underneath | Verify email before submission toggle alone, no nested toggle |

</details>

**Open gaps**

- No e2e test specifically exercises "toggle no longer appears in the editor" — the feature is being removed, not added, and no existing Playwright spec asserted its presence, so there's nothing to update there.
- `openapi.yml` was hand-edited to remove the two lines the field owned, rather than regenerated via `pnpm run generate-api-specs`: that script's vite build (`apps/web/scripts/openapi/vite.config.ts`) fails in this environment trying to bundle `@prisma/client` for a browser target — reproduces identically on `main` with no changes from this branch, so it's a pre-existing local/CI-environment issue, not something this PR introduced. The resulting YAML was validated to still parse (`js-yaml`) and the diff is a straight subset of what the generator would have produced (same two spots the field owned, nothing else touched).
- Did not attempt to count how many production/cloud surveys currently have `isSingleResponsePerEmailEnabled: true` set — that requires cloud DB access this environment doesn't have; flagging per @xernobyl's review so it can be checked before merge if it matters for comms.

**Risks**

- Any self-hosted or cloud survey with `isSingleResponsePerEmailEnabled: true` in the DB silently loses that (non-functional) flag on migration. Since the flag never blocked anything, there's no behavioral regression — only the (already-misleading) toggle state disappears.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5` (Claude Code), reasoning effort `unknown` (not exposed to the harness in this non-interactive session).


xernobyl (commented):
Looked into this — the ticket, the code on main, and the diff. Removing it makes sense to me 👍

The gap is real: the toggle never gated the response create path (a replayed verify token mints unlimited "verified" responses), and the only thing it ever did was the UI-side check in verify-email that refuses to send a second verification link. Single-use links (and personal links) are the enforced way to do this, so sunsetting instead of wiring up enforcement is fine by me — a toggle that promises protection it doesn't deliver is worse than no toggle. Nice side effect: this also removes `isSurveyResponsePresentAction`, which was effectively a public "has this email answered this survey" oracle.

Good to go, but a few small things first, mainly around the OpenAPI spec:

- the v2 API reference is stale: `docs/api-v2-reference/openapi.yml` still documents `isSingleResponsePerEmailEnabled` (~line 5744) and lists it in the survey schema's `required` array (~5832). Once this merges, anyone validating against the published spec sees every survey response fail. Needs a spec regen (`generate-api-specs`) — nothing in CI catches the drift, so it would ship silently.
- the breaking changes table only mentions management API v2, but v1 create/update/read loses the field the same way (via `ZSurveyCreateInput`/`ZSurveyUpdateInput`). Worth a row, same remediation.
- small wording nit: "never actually enforced anything" is a bit strong for something operators had switched on — the UI did block sending a second link to an address that already responded, so casual double submissions were caught. I'd say "UI-only deterrent, never enforced at the API" so the changelog reads accurately for people who relied on it.

Non-blocking thoughts: before merging it'd be good to count how many cloud surveys actually have this enabled (the column drop is unrecoverable, and it tells us whether this needs comms), and maybe give Matti a heads-up — ENG-2507 sketched closing ENG-2294 with the ENG-2094 enforcement pattern, and this closes it by deletion instead. Defensible, just worth saying out loud.

Happy to approve once the spec is regenerated.

jobenjada:
@xernobyl thanks for the thorough review — all addressed:

- `docs/api-v2-reference/openapi.yml` no longer documents `isSingleResponsePerEmailEnabled` or lists it as `required` (had to hand-edit rather than run `generate-api-specs`: that script's vite build fails bundling `@prisma/client` in this environment, reproduces identically on `main`, unrelated to this change — flagged as an open gap in the PR description).
- Breaking changes table now has a row for Management API v1 (`ZSurveyCreateInput`/`ZSurveyUpdateInput`) alongside v2.
- Reworded "never actually enforced anything" → "UI-only deterrent, never enforced at the API" so it doesn't undersell what the toggle did catch.

On the non-blocking points: I don't have access to cloud DB from here to count how many surveys have this enabled, so left that as an explicit open gap for whoever merges to check if it matters for comms. Same for looping in Matti on ENG-2507/ENG-2094 — that's a heads-up call for a human, not something I can act on.

xernobyl (dismissed):
Verified the spec fix at 73e281674 — field gone from the schema and the required list, YAML parses, Mintlify validation green. Breaking table and wording updates look good too. The two open points (cloud usage count, Matti heads-up re ENG-2507) are on me before merge. Approving 👍

xernobyl (approved):
Re-approving at f6634bf05. Reviewed the delta since my last approval — it's exactly the one-line `DROP COLUMN IF EXISTS` change (pushed by me, addressing the CodeRabbit thread with Johannes's OK). Verified locally against real Postgres: the full 168-migration history applies cleanly in order on a fresh DB, end-state schema is correct, and re-running the drop migration skips with a NOTICE instead of wedging the deploy — the exact replay case the thread was about.

Closing the loop on my earlier note: the cloud usage count is consciously dropped — the flag never enforced anything, so the lost toggle state is cosmetic and the breaking-change table already covers API consumers. The ENG-2507 heads-up to Matti is happening on Slack.

PR #8976, opened 2026-08-25
refactor(workflows): drop the dead duplicate workflows list modules

jobenjada:
## What & why

Three files under `apps/web/modules/ee/workflows/list/` were unreachable — they imported only each
other, and nothing outside that directory imported any of them. Each was a superseded copy of a
module the workflows list page actually uses:

| Deleted (dead) | Live counterpart in use |
| --- | --- |
| `list/hooks/use-workflows.ts` | `hooks/use-workflows.ts` |
| `list/lib/v3-workflows-client.ts` | `lib/api-client.ts` |
| `list/lib/query.ts` | `lib/query.ts` |

They are an older shape, not a variant worth keeping: the dead `useWorkflows` accepted no filter
arguments at all — no `nameContains`, `statusIn` or `sortBy` — and its `workflowKeys.list()` keyed on
`{ workspaceId, limit }` alone, so it predates the list filtering the live hook carries.
`pages/workflows-list-page.tsx` imports `../hooks/use-workflows`, never the deleted one.

`use-trigger-survey-picker.ts` was the one live module in that directory (four importers across the
workflow builder), so it moves up to `hooks/` alongside its test and `list/` goes away entirely. The
hook imports nothing relative — only `@/` paths — so the move needed no edits inside it; just the
four import sites. Both moves are recorded as renames, so the diff shows no content change for them.

## Breaking changes

None — the deleted modules were unreachable from any entry point, and the moved hook keeps its
exported names and behaviour. No API, SDK, route, env var or config key is touched.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The three deleted files really were dead | manual | Repo-wide `grep -rn "ee/workflows/list"` before deleting returned five hits: four for `use-trigger-survey-picker` (kept) and one file importing the other two — `list/hooks/use-workflows.ts` itself, which nothing imported. Re-ran on the exported symbol names too (`listWorkflows`, `workflowKeys`, `flattenWorkflowPages`) to catch relative-path importers the alias grep would miss: every live hit resolves to `../lib/query` or `../lib/api-client`. Post-change sweep for `workflows/list` across all file types, configs and docs included, is empty |
| The moved hook still resolves for all four importers | unit (mutation) | `pnpm typecheck` 27/27 green — the code under test is the new import path, so a wrong path fails module resolution rather than passing quietly. Importers: `components/inspector/workflow-trigger-form.tsx`, `components/canvas/workflow-validation-status.tsx`, `hooks/use-reconcile-trigger-ending-cards.ts`, `pages/workflow-builder-page.tsx` |
| The moved hook's own behaviour is unchanged | unit (guard) | `use-trigger-survey-picker.test.ts` moved with it and still passes — it imports via `./use-trigger-survey-picker`, so it followed the file. Whole module green: `npx vitest run modules/ee/workflows` -> 37 files, 409 tests |
| Nothing else in the app regressed | unit (guard) | Full `pnpm test`: 8525 passed. One unrelated failure, `lib/env-access.test.ts`, timed out at 5000 ms under full-suite contention and passes solo in 4.3 s — it lints string sources for `process.env` and shares nothing with this diff |

No visual evidence: the three files this diff modifies change one import line each and nothing else.
No rendered output, styling or copy is touched.

**Open gaps**

- **Not exercised in a running app.** Everything here is resolution and reachability, which
  `typecheck` plus the module's own suite settle statically; there is no runtime behaviour to click.
  The workflow builder screens that import the moved hook were not opened in a browser.
- `lib/env-access.test.ts` is left as-is. It is flaky on this machine at its 5 s timeout
  (4.3 s solo) and unrelated to this change, so raising that timeout does not belong in this PR.

**Risks**

- The deleted `workflowKeys` used the same `["workflows", "list", …]` key prefix as the live one, so
  the two query-key factories overlapped in the cache. Nothing mounted the dead hook, so no live
  entry is being removed — but if a future branch expected that older `{ workspaceId, limit }`
  keying, it is gone and the live `lib/query.ts` shape is the only one.
- If any unmerged branch imports the deleted paths, it will fail to resolve on rebase rather than
  silently pick up the older no-filter hook. That is the intended outcome.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `xhigh`.


kiln-agent (commented):
## Review: no findings

Reviewed `aff2e4e` (6 insertions / 107 deletions, 9 files). This is a clean dead-code deletion plus a byte-identical file move, and every claim in the description holds up.

**What I verified**

- **The three deleted modules were genuinely unreachable.** `list/hooks/use-workflows.ts`, `list/lib/query.ts` and `list/lib/v3-workflows-client.ts` imported only each other. Grepping the exported symbols (`useWorkflows`, `workflowKeys`, `flattenWorkflowPages`, `listWorkflows`, `buildWorkflowListSearchParams`) rather than just the alias path — so relative-path importers can't hide — every live consumer resolves to `modules/ee/workflows/lib/query.ts` or `modules/ee/workflows/lib/api-client.ts`.
- **The live counterparts are strict supersets**, so nothing is lost by preferring them: `lib/api-client.ts` carries the `filters` (`nameContains` / `statusIn` / `sortBy`) the dead copy lacked, and `lib/query.ts`'s `workflowKeys` adds `details` / `detail` plus the filter-aware `list` key.
- **No stale references anywhere.** An unrestricted grep for `workflows/list` and `v3-workflows-client` across the whole tree — all file types, configs, docs, CI workflows, tsconfig/vitest/eslint — returns nothing. The `list/` directory is gone with no orphans.
- **No coverage lost.** `git ls-tree HEAD~1` shows that directory held exactly five files: the three deleted (which had no tests) and the two moved. `use-trigger-survey-picker.test.ts` moved with its subject and imports it as `./use-trigger-survey-picker`, so it followed correctly — 6/6 pass.
- **The move needed no internal edits.** `hooks/use-trigger-survey-picker.ts` has zero relative imports (only `@formbricks/*` and `@/` aliases), so the new depth cannot affect its resolution, and there is no filename collision in the destination.
- **All four import sites updated**, and no fifth exists: `components/inspector/workflow-trigger-form.tsx`, `components/canvas/workflow-validation-status.tsx`, `hooks/use-reconcile-trigger-ending-cards.ts`, `pages/workflow-builder-page.tsx`.
- **Lint and formatting clean.** `prettier --check` and `eslint` are silent on all five touched files; the import reorder in `workflow-trigger-form.tsx` and `workflow-builder-page.tsx` is what the `@trivago` sort plugin requires.

**On the stated gaps** — both are the right call. There is no runtime behaviour here to click: this diff is entirely resolution and reachability, which `typecheck` plus the module's own suite settle statically. And the `lib/env-access.test.ts` timeout is unrelated to this change; raising it here would be scope creep.

Nothing blocking from my side.

---

itsjavi (approved):

PR #8977, opened 2026-08-25
feat(surveys): mirror widget placement for right-to-left languages

jobenjada:
Ref ADM-16

## What & why

Survey window placement is one workspace-level setting (`Workspace.placement`), authored once and in LTR terms. RTL text rendering already works — the survey resolves `dir="rtl"` from its active language and mirrors its own type, close button and buttons — but the popup itself stayed pinned to the same physical corner. An Arabic respondent got it on the side their eye reaches last, on top of the page corner an RTL host page is most likely to already be using.

This mirrors the corner horizontally whenever the active language resolves to `dir="rtl"`, reusing the direction state that already drives the survey's text rather than adding any new setting:

- `bottomRight` ⇄ `bottomLeft`, `topRight` ⇄ `topLeft`; vertical placement never flips (RTL reverses the inline axis only) and `center` has no side to flip.
- The input is the **active** language, not the survey default — switching language mid-survey moves the popup, because `dir` is already recomputed there (`packages/surveys/src/components/general/render-survey.tsx`).
- A survey with no configured languages still mirrors: `isRTLLanguage` sniffs the content, so an Arabic-only survey behaves the same as a configured `ar` one.

The whole widget change is one call site — `getPlacementStyle(mirrorPlacementForDir(placement, dir))` in `SurveyContainer` — because `dir` was already a prop there and placement is applied in exactly one place.

The editor preview positions the popup with its own chrome (`modules/ui/components/preview-survey`) rather than the widget's `SurveyContainer`, so it gets the same mirroring. Without it, the one surface where an author checks placement would contradict what ships.

It keys on the language the **rendered survey** is actually on, not on the preview's `languageCode` prop, and hands that same language to the survey it renders — the two modal previews sit behind `previewMode` gates, so toggling between the phone and browser frames remounts the survey, and seeding it from the prop would reset content and corner out of step. The previewed survey carries its own language switch — `getShowLanguageSwitch` gates on `showLanguageSwitch` and the enabled-language count and nothing else, so it is live in preview mode — and an author using it changes direction without touching that prop. The preview subscribes to `onLanguageChange`, which `Survey` already fires for every active language, and holds the result in local state rather than pushing it back through `setLanguageCode`: that state is the editor's *editing* language and also drives the question pane, so a switch inside the preview must not retarget what the author is typing into.

## Breaking changes

None — no API/SDK shape, endpoint, route, env var, config key, default value or webhook payload changes. Nothing to migrate: `Workspace.placement` keeps its stored value and its meaning for LTR surveys is unchanged. It is a rendering change for RTL surveys only, called out under Risks.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Widget mirrors each corner for an RTL active language, leaves `center` alone | unit (mutation) | `packages/surveys/src/lib/utils.test.ts` → `mirrorPlacementForDir`; new code, so it only fails if the mapping is broken. `pnpm --filter @formbricks/surveys test` — 725 passed |
| LTR and `auto` keep the authored corner, and mirroring is its own inverse | unit (guard) | Same file — a survey switched back to English returns to its corner rather than drifting |
| The shipped widget actually moves: Arabic app survey with workspace placement `bottomRight` | manual | Real UMD bundle (`packages/surveys/dist/index.umd.cjs`) built from this branch and from its merge base, rendered on the same host page at 1280×720. Before: bottom-right. After: bottom-left. English is byte-identical in both. See the pairs below |
| The popup follows a language switch made **inside** the survey, not just the host's initial language | manual | Same harness, `showLanguageSwitch` on, Arabic default + English. Clicking the survey's own globe and picking English moved the card from x=28 to x=868 in a 1280px viewport — left corner to right corner. This is the signal the preview now subscribes to |
| Editor preview mirrors on the survey's active language | unit (mutation) | `apps/web/modules/ui/components/preview-survey/lib/utils.test.ts` covers the mapping and its round trip; `isRTLLanguage` (including its `"default"` branch) is already covered by `modules/survey/link/lib/utils.test.ts`. The state wiring itself is not exercised in a running editor — see Open gaps |

<details>
<summary>Screenshots — Arabic and English app survey, workspace placement <code>bottomRight</code></summary>

Same host page, same seeded survey, same 1280×720 viewport. The only variable is the widget bundle: the merge base on the left, this branch on the right.

| Before | After |
| --- | --- |
| Arabic survey pinned to the bottom-**right** corner, opposite the reading direction<br>Arabic survey popup in the bottom-right corner | Arabic survey pinned to the bottom-**left** corner, on the reading-end side<br>Arabic survey popup in the bottom-left corner |
| English survey in the bottom-right corner<br>English survey popup in the bottom-right corner | English survey unchanged, still bottom-right<br>English survey popup in the bottom-right corner |

| In-survey language switch, after this change | |
| --- | --- |
| The survey's own globe set to Arabic — card in the bottom-left corner<br>Arabic survey with its language switch, popup bottom-left | Same session after picking English in that switch — card has moved to the bottom-right<br>Same survey switched to English, popup bottom-right |

</details>

**Open gaps**

- The editor-preview wiring (`preview-survey/index.tsx`) was verified by unit test, by reading the code path, and — for the widget half of the contract — by the in-survey switch pair above. What nobody has watched is the **preview** popup move: `PreviewSurvey`'s own state wiring (`onLanguageChange` → `activeLanguageCode` → both `placement` and the survey's seed language, plus the effect that lets the chrome dropdown still win) has not been exercised in a running editor, and neither has the phone/browser frame toggle that motivated the seeding. The only dev server reachable from this worktree serves a different branch.
- Not checked below the `sm` breakpoint. The widget's placement classes are all `sm:`-prefixed, so a phone-width modal is full-width and bottom-anchored either way — but that is reasoning, not an observation.

**Risks**

- A survey already authored with an explicit `bottomLeft` (or `topLeft`) **and** an RTL language now renders on the right. That is inherent to mirroring rather than pinning: placement is one setting shared across every language of the survey, so the only way to leave such a survey alone would be to add the per-language override this deliberately avoids. Re-check by flipping the workspace setting to the opposite corner.
- The preview deliberately does **not** report an in-survey language switch back to the editor, so the chrome's globe dropdown and the question pane keep showing the editing language while the survey renders another. That is pre-existing and unchanged; wiring them together would move the author's editing target from inside a preview, which belongs in its own change.
- A language picked inside the modal preview now survives a phone/browser frame toggle, where it used to reset to the editing language. That is the point — content and corner have to agree across the remount — but it is a behaviour change beyond placement. Re-check by using the chrome's globe dropdown, which still overrides it through the prop. Link/fullwidth previews are untouched.
- Nothing else reads placement's horizontal side: `StackedCard` branches only on `top*` vs `bottom*`, so card stacking is untouched. Link surveys never receive a placement at all.

---

> [!NOTE]


packages/surveys/src/lib/utils.test.ts
```diff
@@ -670,3 +671,29 @@ describe("cn", () => {
...
+  });
+
+  test("leaves center untouched in RTL — it has no side to flip", () => {
+    expect(mirrorPlacementForDir("center", "rtl")).toBe("center");
+  });
+
+  test("returns the authored placement unchanged for ltr and auto", () => {
+    expect(mirrorPlacementForDir("bottomRight", "ltr")).toBe("bottomRight");
+    expect(mirrorPlacementForDir("topLeft", "ltr")).toBe("topLeft");
+    expect(mirrorPlacementForDir("bottomRight", "auto")).toBe("bottomRight");
```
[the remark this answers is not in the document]
jobenjada: Half taken.

**Self-inverse in the preview test** — fair, added in 4042815. The two `mirrorPlacementForDir` copies are deliberate duplicates, so their tests drifting apart is exactly how one stops matching the other.

**`mirrorPlacementForDir("auto", "rtl")`** — skipping: `"auto"` is a `dir` value, not a `TPlacement`. `ZPlacement` is `["bottomLeft", "bottomRight", "topLeft", "topRight", "center"]` (`packages/types/common.ts:70`), so that call needs a `@ts-expect-error` to compile and would assert nothing about the contract. The real `auto` case is the second argument, already covered: `mirrorPlacementForDir("bottomRight", "auto")`.

packages/surveys/src/lib/utils.ts
```diff
@@ -302,6 +303,36 @@ const mimeTypes: Record<string, string> = {
...
+    case "bottomLeft":
+      return "bottomRight";
+    case "topRight":
+      return "topLeft";
+    case "topLeft":
+      return "topRight";
+    default:
+      return placement;
+  }
+};
```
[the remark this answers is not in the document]
jobenjada: Skipping — nothing to fix in the diff. `packages/surveys/dist` is gitignored (`packages/surveys/.gitignore:11`); it is a build artifact that Turborepo produces in CI and in `pnpm build`, never a checked-in one, so its absence from the tree is the expected state rather than a missing step. The bundle used for the before/after screenshots in the description was built from this branch with `pnpm build --filter=@formbricks/surveys --force`.

apps/web/modules/ui/components/preview-survey/index.tsx
```diff
@@ -69,7 +71,14 @@ export const PreviewSurvey = ({
   const { overlay: surveyOverlay } = workspaceOverwrites || {};
   const { clickOutsideClose: surveyClickOutsideClose } = workspaceOverwrites || {};
 
-  const placement = surveyPlacement || workspace.placement;
+  // Placement mirrors with the previewed language, exactly as the shipped widget does — see
+  // `mirrorPlacementForDir`. The check runs on the preview's own language selection rather than the
+  // survey's default, so switching the preview to Arabic moves the popup the same way a respondent
+  // reading it in Arabic will see it.
+  const placement = mirrorPlacementForDir(
```
kiln-agent, line 116: The widget keys the mirror on the **active** language (`render-survey.tsx` recomputes `dir` from `activeLanguageCode`, which the in-survey `LanguageSwitch` updates). The preview keys it on the `languageCode` **prop** only — and the previewed survey has its own language switcher, which never reports back here: `SurveyInline` is rendered without `onLanguageChange`, so `setLanguageCode` is only ever called by the preview chrome's globe dropdown.

Concrete divergence — app survey, workspace placement `bottomRight`, default language Arabic + English enabled, `showLanguageSwitch` on (`survey.tsx:324` is the only gate, and it holds in preview mode):

1. Editor preview opens with `languageCode = "default"` → Arabic → popup mirrored to bottom-left. Correct.
2. The author clicks the globe **inside the rendered survey** and picks English. `LanguageSwitch.changeLanguage` calls `setDir("ltr")`, so the survey re-renders LTR — but `PreviewSurvey`'s `languageCode` prop is unchanged, so `placement` stays `bottomLeft`.
3. The preview now shows an LTR survey in the bottom-**left** corner, while the shipped widget would move it back to bottom-right. Same in reverse when the default is English and the author switches to Arabic in-survey: the text flips RTL, the popup does not.

So on that path the preview still contradicts what ships — the failure this PR sets out to remove. Passing `onLanguageChange={setLanguageCode}` on the modal-preview `SurveyInline` (or otherwise deriving the mirror from the survey's active language rather than the prop) closes it.

---

apps/web/modules/ui/components/preview-survey/lib/utils.ts
```diff
@@ -16,3 +16,30 @@ export const getPlacementStyle = (placement: TPlacement) => {
...
+ * Editor-preview twin of `mirrorPlacementForDir` in `packages/surveys/src/lib/utils.ts`. The preview
+ * positions the survey with its own fake-browser chrome rather than the widget's SurveyContainer, so
+ * without this the preview would keep the popup on the authored side while the shipped widget mirrors
+ * it — the one surface where an author checks placement would be the one lying about it.
+ *
+ * Duplicated rather than shared for the same reason `isRTLLanguage` already is: the surveys package
+ * is a Preact bundle the web app loads dynamically, so importing a helper out of it would pull that
+ * bundle into the editor's own chunk.
+ */
+export const mirrorPlacementForDir = (placement: TPlacement, dir: "ltr" | "rtl" | "auto"): TPlacement => {
```
kiln-agent, line 30: The stated reason for duplicating doesn't apply to the one place both consumers already share: `@formbricks/types/common`, where `ZPlacement`/`TPlacement` live. Both sides import it today — this file at line 1, and `packages/surveys/src/components/wrappers/survey-container.tsx:4` — and per AGENTS.md `types` is consumed from source with no build step, so a pure function next to `ZPlacement` pulls in no Preact bundle and no editor-chunk cost. The `isRTLLanguage` precedent doesn't carry: that one closes over survey content types, this one is a five-case map over the enum defined in that very file.

The concrete cost of keeping two copies is already visible one function up: `getPlacementStyle` is the other duplicated placement helper, and the copies have drifted — `sm:top-3 sm:right-3` in the widget (`survey-container.tsx:113`) vs `sm:top-6 sm:right-6` here. A future change to the mapping (a new `TPlacement` member, or a decision about `center`) landing in one copy only reproduces exactly that, and the symptom is the preview showing a different corner from the shipped widget — silently, since each copy has its own passing test file.

---

apps/web/modules/ui/components/preview-survey/index.tsx
```diff
@@ -69,7 +71,14 @@ export const PreviewSurvey = ({
...
+    surveyPlacement || workspace.placement,
+    isRTLLanguage(toJsWorkspaceStateSurvey(survey), languageCode) ? "rtl" : "ltr"
+  );
```
kiln-agent, line 119: Minor, cosmetic, and in untouched code — but this line makes it the default for every RTL app survey: the desktop preview's hide animation is hard-coded to slide right (`components/modal.tsx:118`, `show ? "translate-x-0" : "translate-x-32"`).

Scenario: Arabic app survey, placement `bottomRight`, `clickOutsideClose` on. The card now sits bottom-left; clicking the preview background sets `show = false`, and instead of sliding out past the near (left) edge the card slides 8rem to the right, across the middle of the fake browser, for the second before it reappears. Previously this only happened to authors who explicitly picked a left corner; now it's what every RTL author sees. A direction-aware sign (`translate-x-32` vs `-translate-x-32`, keyed on whether the resolved placement is a left corner) would fix it.

---

kiln-agent (commented):
Reviewed for correctness. The widget change itself checks out: `SurveyContainer` is the only place placement becomes CSS, `dir` there is only ever `"ltr"`/`"rtl"` in practice (`render-survey.tsx` never sets `"auto"`), the classes are physical (`left-*`/`right-*`, no logical-property or RTL plugin in `packages/surveys/tailwind.config.cjs`) so the manual mirror is the right mechanism, and nothing double-mirrors — the editor preview renders through `renderSurvey({ mode: "inline" })`, which skips the modal branch entirely, and `StackedCard` branches only on `top*` vs `bottom*`, so the mirrored value reaches it harmlessly. All five `ZPlacement` members are handled.

Three findings, none blocking the widget behaviour; the first is the one worth acting on.

---

kiln-agent (commented):
## Review — `feat(surveys): mirror widget placement for right-to-left languages`

Reviewed at `442ed73`. The helper is the easy part and it's right: total over the 5-member `TPlacement`, vertical axis and `center` untouched, self-inverse, and both test suites are mutation-detecting rather than coverage padding (I ran `mirrorPlacementForDir` in `packages/surveys` directly — 4 passed). The shipping path is a genuine one-liner because `dir` was already a `SurveyContainer` prop. Everything I found is about **where the mirror is and isn't applied**, not about the mirror itself — nothing here blocks shipping the widget change.

### Findings

**1. The other app-survey preview wasn't mirrored** — `apps/web/modules/ui/components/theme-styling-preview-survey/index.tsx:106`

`renderAppPreview` there goes through the same `Modal` + `getPlacementStyle(placement)` path the editor preview does, but `placement` is still `surveyPlacement || workspace.placement` raw. It's the only other surface that renders an app-survey popup at a placement, so the PR's own justification for touching the editor preview — "the one surface where an author checks placement would contradict what ships" — reads as applying here too.

Narrower than it first looks, and worth saying so: this preview renders the synthetic `previewSurvey(workspace.name, t)` template in the admin's **UI** locale, and `previewSurveyType` defaults to `"link"`, so it only diverges for an RTL-locale admin who switches the toggle to "app". Still, an Arabic-locale admin would then see bottom-right on Look & Feel and bottom-left in the survey editor for the same workspace setting.

**2. The editor preview doesn't re-mirror when the language changes inside the survey** — `apps/web/modules/ui/components/preview-survey/index.tsx:78`

Mirroring keys off the `languageCode` prop, which only the chrome's globe `LanguageSelector` drives. The `SurveyInline` at line 284 isn't passed `onLanguageChange`, so a language change made through the survey's **own** `LanguageSwitch` (rendered whenever `showLanguageSwitch && enabledLanguages.length > 1`, preview included) calls `setDir` inside `RenderSurvey` and flips the content while the chrome keeps the popup on the authored side.

That inverts the intent: in the shipped widget the same switch *does* move the popup, because `SurveyContainer` reads the live `dir` state — so the preview now under-reports the behaviour this PR adds, in the one interaction the description calls out as the reason to key on the active language ("switching language mid-survey moves the popup"). This is the mismatch listed under Open gaps, but it's reachable from the chrome the preview already renders rather than only from a running editor.

Not a clean one-liner, which is why I'm flagging rather than prescribing: `handleLanguageChange` emits the real language code, while the chrome's `languageCode` uses the sentinel `"default"` for the default language (`surveyLang.default ? "default" : surveyLang.language.code`). `isRTLLanguage` resolves either, so mirroring would work — but wiring `onLanguageChange={setLanguageCode}` straight through would stop the dropdown's `languageCode === code` check from highlighting the default entry.

**3. Nit — the `"auto"` case can't be reached** — `packages/surveys/src/lib/utils.ts:317`

`setDir` has exactly one call site (`render-survey.tsx:30`, `setDir(isRTL ? "rtl" : "ltr")`) and the state initialiser is the same ternary, so widget `dir` is never `"auto"`; the preview twin passes a ternary that can't produce it either. Harmless — it falls through `dir !== "rtl"` — and the parameter type reasonably mirrors the existing `dir` type. But the doc paragraph explaining what `auto` means *here* documents a state this function never sees, so it reads as live behaviour when it's inherited type surface.

### Checked and not findings

Recording these so the shape of the change is on the record:

- **No double-mirroring in the editor preview**, even though the already-mirrored `placement` is also handed down to `SurveyInline` (line 298). `SurveyInline` forces `mode: "inline"`, and `SurveyContainer` returns from its `!isModal` branch (line 233) before reaching `getPlacementStyle`. This was my main worry on first read.
- **The placement *pickers* correctly keep the raw value** — `survey/editor/components/placement.tsx:68` and `workspaces/settings/look/components/edit-placement-form.tsx:130`. Both swatches sit next to a radio labelled "Bottom Right"; mirroring them would make the swatch contradict its own label. The setting is authored in LTR terms and should be shown that way, even when the live preview beside it now renders mirrored.
- **`StackedCard` is unaffected** — `stacked-card.tsx:45,49` branch on `bottomLeft||bottomRight` vs `topLeft||topRight`, so the horizontal side is invisible to card stacking. It also receives the unmirrored `props.placement`, which is consistent either way.
- **`TPlacement` is a type-only import**, so nothing is added to the Preact bundle.
- No pre-existing `[dir=rtl]` rule that would double-flip the corner in CSS.

### CI

Green at this SHA except three jobs still running when I looked (E2E, Unit Tests, Validate Docker Build, SonarQube). Locally, `pnpm --filter @formbricks/surveys test` reports 700 tests passing with 2 test *files* failing on `packageEntryFailure` — that's the stale-`dist/` trap documented in `AGENTS.md`, from bypassing the turbo graph in this container, not this diff.

### Verdict

The widget change is sound and I'd ship it. Findings 1 and 2 are both "a preview disagrees with the widget", which is the specific failure mode this PR set out to remove — worth closing before merge, or logging deliberately with the Open gaps entry updated to name the in-preview language switch as reachable.

---

apps/web/modules/ui/components/preview-survey/index.tsx
jobenjada: addressing, seems reasonable

apps/web/modules/ui/components/preview-survey/index.tsx
```diff
@@ -286,6 +311,7 @@ export const PreviewSurvey = ({
                         setBlockId = f;
                       }}
                       onFinished={onFinished}
+                      onLanguageChange={setActiveLanguageCode}
                       placement={placement}
```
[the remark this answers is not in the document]
jobenjada: Confirmed and fixed in 325ac5c — the premise holds: both modal previews sit behind `previewMode === "mobile"` / `=== "desktop"` gates (`index.tsx:269`, `:370`), so toggling the frame really does unmount one and mount the other, and the fresh `RenderSurvey` seeds `activeLanguageCode` from `useState(props.languageCode)`.

Worth noting the failure is transient rather than sticky — the remounted `Survey` fires its report effect on mount (`survey.tsx:474`), which feeds the language back and re-corrects `placement`. But the card carries a 500 ms transition, so the visible result was an Arabic survey sliding across the viewport to the LTR corner and back, and the author's in-survey language choice was discarded on every frame toggle either way. Seeding from `activeLanguageCode` removes both.

Scoped to the two modal instances as suggested. The link/fullwidth `SurveyInline`s keep the prop: they get no `onLanguageChange` and no placement, and `previewType` is fixed per survey type, so no survey ever sees both paths.

pandeymangg (approved):

PR #8978, opened 2026-08-25
docs: rewrite multi-language surveys page for the current translation flow

jobenjada:
Fixes ADM-15

## What & why

The multi-language docs page described a UI that no longer exists: a "Multiple Languages" section at the
bottom of the survey editor, a language dropdown on each question, and a vague "Translate with AI" note at
the end of the page. The current product has a dedicated **Language** tab with per-language visibility and
translation progress, and translating happens in a **Manage translations** editor whose **Translate with AI**
button fills every empty field of one language in a single click. A customer call surfaced the gap (see the
ticket): the reader could not find the one-click flow, and could not tell what happens when a user's language
has no translation.

- Rewrote the page around the real flow: workspace languages → Language tab → Manage translations → publish.
- Documented the one-click auto-translate explicitly: what it fills (empty fields only), that existing
  translations survive a re-run, that nothing is stored until **Save**, and the two org-level conditions that
  gate the button.
- Made the missing-translation fallback an explicit, documented decision rather than a footnote: link surveys
  fall back to the default language; app and website surveys are **skipped**. Added a resolution table for
  the `language` attribute and the reason behind the asymmetry.
- Added the parts of the feature that were undocumented: per-language visibility and its save-blocking
  validation, translation progress and the amber warning on the tab, **Missing first** sorting, the
  respondent-facing language switch, the share modal's per-language link picker, `lang` matching precedence
  (identifier → alias → canonical, so an old `?lang=pt` still resolves to `pt-BR`), per-language link
  previews, and how translated answers aggregate in the summary.
- Replaced all screenshots with captures from the current app and deleted four images whose UI is gone
  (`workspace-configuration`, `surveys-home`, `enable-multi-lang`, `add-language-in-survey`).
- Documented the exact-tag requirement on `setLanguage()`: js-core matches a language's identifier or alias
  and, unlike the link path, never canonicalizes — so `setLanguage("ar")` against a stored `ar-EG` hides the
  survey. See Risks for the underlying asymmetry.

Docs-only change: no product code is touched.

## Breaking changes

None — documentation only, no API, SDK, env var or route change.

## Migrations & env

- none

## How this was tested

Every screenshot was captured from this branch running locally against a seeded workspace with four
languages (English default, German and Arabic complete, French partial), which is also how each behavioural
claim below was checked in the running app. Claims that are not visible on screen were verified against the
code that implements them.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Language tab: activation toggle, per-language `Code`/`Translated`/`Visible` columns, progress colours, `Manage languages`, `Show language switch` | manual | Matches the page's description; captured as `language-tab.webp` (Arabic 17/17, German 17/17, French 7/17) |
| Workspace languages settings: `Identifier (BCP-47)` is read-only, language fixed after creation, alias editable, in-use language cannot be removed | manual + `language-row.tsx`, `edit-language.tsx` | Captured as `survey-languages-settings.webp` / `add-languages.webp`; removal guard confirmed in code |
| Manage translations editor: side-by-side columns, `Translate with AI`, `Missing first` / `Show in order`, `ID` column (`W`, `1.1`), save-on-confirm | manual + `manage-translations-modal.tsx` | Captured as `manage-translations.webp`; AI fills `emptyFields` only and drafts are discarded on cancel |
| `Translate with AI` gating (plan entitlement, org smart-tools setting, configured AI instance) | code — `ai-translation/lib/actions.ts`, `lib/ai/service.ts` | Three reasons map to the tooltip strings quoted in the docs note |
| Visible language with missing required labels blocks save/publish; translations survive toggling visibility off | code — `packages/types/surveys/validation.ts`, `language-view.tsx` | `validateLabelForAllLanguages` filters on `enabled`; disabling a language with translations keeps its keys |
| Language switch renders on every card, needs ≥2 visible languages, lists endonyms | manual + `survey.tsx`, `language-switch.tsx` | Captured as `language-switch.webp` (American English / Deutsch (Deutschland) / العربية (مصر)) |
| Link survey `?lang=` precedence and fallback to default | manual (`?lang=ar`, `?lang=de`) + `survey-renderer.tsx` | Code → alias → canonical; unknown, disabled or default resolves to `default` |
| Share modal language picker appends `?lang=` | manual | Captured as `share-link-language.webp` |
| App survey skipped when the user's `language` has no visible translation; single-language surveys ignore the attribute | code — `js-core/src/lib/survey/widget.ts`, `lib/common/utils.ts` | `getLanguageCode` returns `undefined` → `setIsSurveyRunning(false)`; gate only applies when `languages.length > 1` |
| RTL rendering mirrors the survey | manual (`?lang=ar`) | Captured as `rtl-support.webp` — progress bar, smiley scale, button and footer all mirrored |
| Built-in interface translations: 23 languages | `ls packages/surveys/locales` | 23 files, matching the table (kept from the previous revision) |
| Translated choice answers aggregate onto the default-language choice | code — `summary/lib/surveySummary.ts` | Answers are read in the response's language and counted against the default-language choice |
| Review fix: `#translate-with-ai` anchor kept alive for the two inbound links in `ai-features.mdx` | manual | Subsection renamed back to `### Translate with AI`; both links (lines 20, 44) resolve again — `Mintlify link-rot` only checks paths, not fragments |
| Review fix: link preview image is not per-language | code — `ZSurveyMetadata` (`types.ts:281-287`), `link/lib/metadata-utils.ts:66-83` | `title`/`description` are `ZI18nString` read through `getLocalizedValue`; `ogImage` is a plain `ZStorageUrl` read unlocalized — page now says the image is shared across languages |
| Review fix: guidance to add a language as invisible removed | code — `js-core/src/lib/common/utils.ts` | `!selectedLanguage.enabled` returns `undefined`, same as an unmatched language, so an invisible language is skipped rather than falling back to default — the advice was wrong; the resolution table already states the outcome |
| Review fix: a bare two-letter code cannot be an alias | code — `edit-language.tsx:65`, `iso639Languages` | The guard rejects an alias equal to any catalog language's code (the catalog holds both `de` and `de-DE`), so the `de`-for-`de-DE` example could never save; example is now `en_us` and the rule is stated |
| Review fix: translation completeness gates publish, not a draft save | code — `survey-menu-bar.tsx:372` vs `:402` | `handleSurveySaveDraft` runs neither `validateSurveyWithZod` nor `isSurveyValid`; only publish and saving a live survey gate on it |
| Review fix: the modal's **Save** does not persist | code — `manage-translations-modal.tsx:209` | It calls `setLocalSurvey` only — no action — so the page now says the survey itself must be saved; the old wording could have cost a reader a whole translation pass |
| Review fix: `?lang=de` against `de-DE` uses the canonical pass, not the alias pass | code — `survey-renderer.tsx:228` | `normalizeLanguageCode("de")` → `de-DE`; relabelled so the three passes stay distinct |
| Review fix: `setLanguage()` snippet passed `"de"` | code — `js-core/src/lib/common/utils.ts` | That is exactly the value that silently hides a survey whose language is `de-DE`; snippet now passes the identifier |
| Review fix: heading capitalization | AGENTS.md path instruction | Headings are sentence case; `### Translate with AI` deliberately keeps its case because it is the `#translate-with-ai` anchor, and no other heading on this page is linked from elsewhere (`grep` for `multi-language-surveys#` finds only those two links) |

**Open gaps**

- `Translate with AI` was not exercised end-to-end: it needs an Enterprise license plus a configured AI
  instance, neither of which exists in the local dev environment. The button, its state and its tooltip
  copy are verified; the quality of the returned translations is not.
- Response filtering by **Language** is stated from the filter's own option list in code
  (`ResponseFilter.tsx`); a seeded multi-language response set was not built to click it through.
- The alias rule is verified from `validateLanguages` and the ISO catalog rather than by trying to save a
  rejected alias in the UI. Note the guard is client-side only — `ZLanguageInput` accepts any string — so an
  alias the editor refuses is still settable through the action; the page documents the editor's rule, which
  is what a reader will hit.

**Risks**

- Docs-only, so the risk is drift rather than breakage: the page now quotes UI labels (`Activate
  translations`, `Visible`, `Translate with AI`, `Missing first`) and the app-survey skip rule. If any of
  those change, this page is the thing to update.
- **Product asymmetry this PR documents rather than fixes.** The two language resolvers disagree:
  `survey-renderer.tsx` matches `?lang=` by code → alias → `normalizeLanguageCode` canonical, added so
  legacy links survived the June 2026 canonicalization migration; `js-core`'s `getLanguageCode` matches code
  or alias only (js-core ships with no runtime deps, so it cannot reach `@formbricks/i18n-utils`). After that
  migration relabelled `ar` → `ar-EG` without preserving the old code as an alias, `setLanguage("ar")`
  matches nothing and the app survey is skipped with only a `logger.debug` line — while `?lang=ar` still
  works. The migration's own log message assumes the SDK "falls back to the survey default", which is not
  what the code does. The page now warns readers to set matching aliases; making the SDK canonicalize (or
  fall back) is a code change for its own ticket.

---

> [!NOTE]


kiln-agent (commented):
## Review

Docs-only rewrite. I verified each behavioural claim on the page against the code that implements it rather than re-reading the prose, since that is the only thing CI cannot check here.

**Claims I confirmed against code** (no action needed, listing so the next reader doesn't redo it):

- `?lang=` precedence — identifier → alias → canonical, case-insensitive, with unknown/disabled/default all collapsing to `default` — matches `survey-renderer.tsx:219-246` exactly, including the `?lang=pt` → `pt-BR` example.
- The app-survey resolution table matches `getLanguageCode` in `js-core/src/lib/common/utils.ts:176-195` row for row, and the "single-language surveys ignore the attribute" note matches `isMultiLanguageSurvey = survey.languages.length > 1` in `widget.ts:122`.
- Bundle matching: `de-AT`/`de` → `de-DE`, `pt-PT` → `pt-BR`, and script preservation for `zh-Hant`/`zh-TW` all match `resolveFallbackBundles` in `packages/surveys/src/lib/i18n.config.ts`. 23 locale files, and the table lists exactly those 23.
- Visibility gating: `validateLabelForAllLanguages` filters on `enabled` (`packages/types/surveys/validation.ts:92`), and `handleToggleLanguage` keeps translations when disabling a language that has any (`language-view.tsx:217-223`) — both halves of that paragraph are right.
- `Translate with AI` fills `emptyFields` only and is disabled at zero (`manage-translations-modal.tsx:131-159, 246`); drafts live in local state and are only committed by `handleSave`.
- `W` / `1.1` ID scheme (`multi-language-surveys/lib/utils.ts:73, 112`), `Missing first` sorting, the amber tab icon (`survey-editor-tabs.tsx:56, 98`), the ≥2-visible-languages switch with endonyms (`survey.tsx:324`), RTL = ar/he/fa/ur (`surveys/src/lib/utils.ts:317`), the in-use-language removal guard (`edit-language.tsx:107-125`), and default-language choice aggregation in the summary (`surveySummary.ts:255, 267`).
- The seven referenced images map 1:1 onto the seven in the folder — no orphans, nothing missing.

Two things to fix, then this is good to go.

---

### 1. Renaming the AI heading breaks two inbound links

`docs/platform/features/ai-features.mdx` links to this page's AI section twice:

- line 20: `[AI Survey Translation](/surveys/general-features/multi-language-surveys#translate-with-ai)`
- line 44: `[AI Survey Translation](/surveys/general-features/multi-language-surveys#translate-with-ai)`

That anchor came from `## Translate with AI`, which this PR renames to `### Translate Everything in One Click` (line 97). No heading on the new page generates `#translate-with-ai`, so both links now drop the reader at the top of a fairly long page instead of at the one-click flow — which is the section the AI features page is specifically pointing at.

Worth noting that the `Mintlify Validation - link-rot` check is green: the page *path* still resolves, so a path-level checker has nothing to flag. The fragment is what went stale.

Simplest fix is renaming the heading to `### Translate with AI` — it also matches the button label the section is about, and the surrounding prose already says "in one click". Otherwise update both links in `ai-features.mdx`.

### 2. The link preview image is not per-language

Line 173:

> In **Share survey → Link settings** you can also set the link preview title, description and image per language

Title and description are per-language; the image is not. In `ZSurveyMetadata` (`packages/types/surveys/types.ts:281-287`) `title` and `description` are `ZI18nString`, but `ogImage` is a plain `ZStorageUrl`. `getBasicSurveyMetadata` resolves the first two through `getLocalizedValue(..., langCode)` and then reads `metadata?.ogImage` unlocalized (`link/lib/metadata-utils.ts:66-83`), and the Link settings tab types it as `ogImage?: string` with its language picker only swapping title and description.

The sentence's conclusion still holds — a German-shared link does preview with German text — so dropping "and image" is enough, or say the image is shared across all languages if that's worth calling out.

---

### Nits (take or leave)

- **Line 68** — "To change it later you have to switch **Activate translations** off". Accurate in effect, but there's a **Change default** item in the default language row's ⋮ menu that does exactly this in one step (`language-view.tsx:273-294` strips every non-default language and flips activation off). A reader who spots that menu item may not connect it to the warning; naming it would close the loop.
- **Line 115** — "Nothing is stored until you save." The modal's **Save** writes to the editor's in-memory survey (`setLocalSurvey`), not the database — the survey still has to be saved or published. The editor's unsaved-changes guard covers the accident, but after a one-click fill of every field it's worth an explicit closing step, which the flow currently doesn't have.

Nothing else blocking — the depth of verification in the PR description held up everywhere I checked it.

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
+  <Step title="Add a language">
+    Click **Edit languages**, then **Add language**, and pick the language from the dropdown. The
+    **Identifier** is filled in for you — it is the language's BCP-47 tag, such as `de-DE` or `ar-EG`.
 
-    ![Add Multiple Languages to your Workspace](/images/surveys/general-features/multi-language-surveys/add-languages.webp)
+    ![Adding a language to the workspace](/images/surveys/general-features/multi-language-surveys/add-languages.webp)
 
-    You can come back to this page anytime to add more languages or remove existing ones.
+    Optionally set an **Alias**: a short alternate name you can use instead of the identifier in link survey
+    URLs and in the SDK, for example `de` instead of `de-DE`. Click **Save changes** when you are done.
```
kiln-agent: The alias example can't be entered. `validateLanguages` rejects an alias that equals another language's ISO 639 code:

```ts
// apps/web/modules/survey/multi-language-surveys/components/edit-language.tsx:65
for (const alias of languageAliases) {
  if (iso639Languages.some((language) => language.code === alias && !languageCodes.includes(alias))) {
    toast.error(t("workspace.languages.conflict_between_selected_alias_and_another_language"));
    return false;
  }
}
```

With identifier `de-DE`, `languageCodes` is `["de-de"]`, so alias `de` matches `iso639Languages` and isn't in `languageCodes` → the save is blocked with the "conflict between the selected alias and another language" toast. Any bare ISO 639 code (`de`, `fr`, `pt`, `nl`) is unusable as an alias, which rules out the most natural thing a reader will try first.

Worth picking an example that actually saves, and saying that a bare two-letter language code is rejected — the guard is deliberate (see the comment above it citing ambiguous entries breaking the ISO-based dropdowns), so readers should know the rule rather than discover it via a toast.

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
   </Step>
 
-  <Step title="Add Supported Languages">
-    Add the languages from the dropdown that you want to support in your survey:
+  <Step title="Make languages visible">
+    Every workspace language shows up as a row with its **Code**, a **Translated** progress bar, and a
+    **Visible** toggle. Turn **Visible** on for each language respondents should be able to answer in.
 
-    ![Add Supported Languages](/images/surveys/general-features/multi-language-surveys/add-language-in-survey.webp)
+    A visible language must carry a translation for every required field before the survey can be saved or
```
kiln-agent: "before the survey can be saved or published" holds for publish, but not for a draft save. `handleSurveySaveDraft` runs no validation at all — no `validateSurveyWithZod`, no `isSurveyValid`:

```ts
// apps/web/modules/survey/editor/components/survey-menu-bar.tsx:372
const handleSurveySaveDraft = async (): Promise<boolean> => {
  setIsSurveySaving(true);
  const segment = await handleSegmentUpdate();
  clearSurveyLocalStorage();
  const updatedSurveyResponse = await updateSurveyDraftAction(getDraftSurveyToPersist(localSurvey, segment));
  ...
```

whereas `handleSurveySave` (line 401) gates on both. A draft with a half-translated visible language saves fine, so the workaround in the next sentence ("switch that language back to invisible") is only needed to publish. Suggest scoping it to publishing.

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
+
+  <Step title="Click Translate with AI">
+    The translations appear in the right-hand column, formatting and rich text included.
   </Step>
 
-  <Step title="Publish Your Survey">
-    Once you are done, click on the **Publish** button to save the survey.
+  <Step title="Review, then save">
+    Read the result before you keep it — tone and product terms are the two things worth a second look — edit
+    anything you want to change, and click **Save**. Nothing is stored until you save.
```
kiln-agent: "Nothing is stored until you save" reads as if the modal's **Save** persists the translations, but it only writes to editor state:

```ts
// apps/web/modules/survey/multi-language-surveys/components/manage-translations-modal.tsx:209
setLocalSurvey(updatedSurvey);
setOpen(false);
```

There is no action call — persistence only happens on the survey's own save in the menu bar. So a reader who translates a whole survey, clicks **Save** in the modal, and then leaves the editor loses all of it, having done exactly what this line told them to.

This is the one finding here with a real cost attached. Suggest something like "clicking **Save** returns the translations to the editor — save the survey itself to persist them."

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
-      the language of the user, the survey will not be displayed.
-    </Note>
-  </Step>
+![Choosing a language for the survey link in the share modal](/images/surveys/general-features/multi-language-surveys/share-link-language.webp)
 
-  <Step title="Start Collecting Responses">
-    That's it! Now, users with the language attribute set will see the survey in their preferred language. You can start collecting responses in multiple languages and filter them by language on the summary page.
-  </Step>
-</Steps>
+`lang` accepts either the language's identifier (`de-DE`) or its alias (`de`), and matching is
```
kiln-agent: The precedence explanation on the next two lines is exactly right, but this line mislabels which pass `de` uses. Given identifier `de-DE`, `?lang=de` cannot be an alias (see my comment on line 42 — validation rejects that alias), so it resolves through the *canonical* pass, the same one as the `?lang=pt` → `pt-BR` example:

```ts
// apps/web/modules/survey/link/components/survey-renderer.tsx:228
const langParamLower = langParam.toLowerCase();
const langParamCanonical = normalizeLanguageCode(langParam);
survey.languages.find((l) => l.language.code.toLowerCase() === langParamLower) ??
  survey.languages.find((l) => l.language.alias?.toLowerCase() === langParamLower) ??
  (langParamCanonical && survey.languages.find((l) => normalizeLanguageCode(l.language.code) === langParamCanonical))
```

`normalizeLanguageCode("de")` → `de-DE` via `LANGUAGE_CANONICAL_MAP` (`packages/i18n-utils/src/canonical.ts`), so the link works — just not for the reason stated. Using a genuine alias for the alias example keeps the three passes distinct.

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
+case-insensitive. Formbricks resolves it in a fixed order: an exact identifier first, then an alias, then a
+canonical match — which is why an older link like `?lang=pt` still works after the language was stored as
+`pt-BR`.
+
+Without a `lang` parameter — or with one that names a language the survey does not have, or one that is not
+visible — the survey opens in its default language.
+
+<Note>
+  In **Share survey → Link settings** you can also set the link preview title, description and image per
```
kiln-agent: Title and description are per-language; the image is not. In `ZSurveyMetadata` only the first two are i18n strings:

```ts
// packages/types/surveys/types.ts:281
export const ZSurveyMetadata = z
  .object({
    title: ZI18nString.optional(),
    description: ZI18nString.optional(),
    ogImage: ZStorageUrl.optional(),
  })
```

`ogImage` is a single `ZStorageUrl` shared across every language. Dropping "and image" from this Note keeps the promise accurate — otherwise a reader plans a per-locale preview image and finds one FileInput.

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
-    For example, if you have set the alias for French as `fr`, you can share the survey link as
+Formbricks.setLanguage("de"); // identifier or alias of the language
+```
 
-    [`https://your-survey-url.com?lang=fr`](https://your-survey-url.com?lang=fr)
+Call `setLanguage()` whenever your app's language changes — for example right after the user switches it in
+your own settings — and Formbricks will use the new value for every survey it shows from then on.
 
-    Here are two examples:
+`setLanguage()` accepts the identifier or the alias, case-insensitively, exactly like the `lang` parameter.
```
kiln-agent: "exactly like the `lang` parameter" isn't true — the SDK path has no canonical pass. js-core matches code or alias only:

```ts
// packages/js-core/src/lib/common/utils.ts:180
const selectedLanguage = survey.languages.find((surveyLanguage) => {
  return (
    surveyLanguage.language.code.toLowerCase() === language.toLowerCase() ||
    surveyLanguage.language.alias?.toLowerCase() === language.toLowerCase()
  );
});
```

No `normalizeLanguageCode`, unlike the link renderer quoted in my line-164 comment. So the two surfaces diverge exactly where readers will assume they don't:

| value | `?lang=` | `setLanguage()` |
| --- | --- | --- |
| `de-DE` (stored code) | resolves | resolves |
| `DE-de` | resolves | resolves |
| `de` (survey stored as `de-DE`) | resolves (canonical) | **survey skipped** |
| `en_US` (survey stored as `en-US`) | resolves (canonical) | **survey skipped** |

And per the table just below, a miss on the SDK path isn't a fallback to the default language — the survey is skipped entirely. That makes this the sharpest edge on the page: a reader who ships `setLanguage("de")` after reading this loses the survey silently. Suggest stating that `setLanguage()` needs the exact identifier or alias, with legacy/short codes not normalized.

(`toLegacyLanguageCodes` in `apps/web/lib/i18n/utils.ts:276` de-canonicalizes on the user-state side for older clients, so the practical blast radius may be narrower than the table suggests — but the doc's claim of equivalence still doesn't hold as written.)

---

docs/surveys/general-features/multi-language-surveys.mdx
```diff
@@ -1,216 +1,286 @@
...
+| Matches nothing on the survey, or a language that is not visible | **Nothing — the survey is skipped** |
 
-  <Step title="Start Collecting Responses">
-    You can now start collecting responses in multiple languages!
-  </Step>
-</Steps>
+The last row is an intentional product decision, not a bug: a targeted survey is skipped rather than shown
+in a language the user did not ask for. It also means adding a language to your app before adding it to a
+survey silently narrows that survey's audience, so add the language to the survey — even untranslated but
+invisible — before you start sending users its way.
```
kiln-agent: This mitigation contradicts the table three lines above it, and following it makes the problem worse rather than better.

The table's last row already says an attribute matching "a language that is not visible" means the survey is skipped, and `getLanguageCode` confirms it — `!selectedLanguage.enabled` returns `undefined`:

```ts
// packages/js-core/src/lib/common/utils.ts:189
if (!selectedLanguage || !selectedLanguage.enabled || !availableLanguageCodes.includes(selectedLanguage.language.code)) {
  return undefined;
}
```

So "even untranslated but invisible" doesn't rescue those users. Worse, per the Note directly below, a single-language survey *ignores* the `language` attribute and shows to everyone. Adding a second (invisible) language flips it into the gated path, so the advice takes a survey that was reaching all users and hides it from every user whose language doesn't match — the exact outcome the paragraph is warning about.

The correct advice is the opposite: add the language to the survey **and make it visible** (with translations) before sending those users its way, or leave the survey single-language until you're ready.

---

kiln-agent (dismissed):
## Review: multi-language surveys docs rewrite

This is a real improvement on the page it replaces — the step structure, the missing-translation table, and the explicit `?lang=` precedence are all much clearer than prose. I checked every behavioural claim against the code that implements it, and most hold up: the UI labels (`Activate translations`, `Visible`, `Translated`, `Missing first`, `Show in order`), the `?lang=` resolution order, the app-survey skip rule, the ID column scheme (`W`, `1.1`, `E1`), the RTL set, the language-switch endonym rule, the summary's default-language aggregation, the 23 locale bundles, all 11 image paths, and the two cross-page links all match.

Seven claims don't. Six are inline; one is on a file outside this diff.

### Broken cross-page anchor (not in this diff)

The rewrite drops the `## Translate with AI` heading, but `docs/platform/features/ai-features.mdx` links to that slug twice:

```
docs/platform/features/ai-features.mdx:20:  [AI Survey Translation](/surveys/general-features/multi-language-surveys#translate-with-ai)
docs/platform/features/ai-features.mdx:44:- **[AI Survey Translation](/surveys/general-features/multi-language-surveys#translate-with-ai)**:
```

No heading on the new page produces `#translate-with-ai` (closest is `### Translate Everything in One Click` → `#translate-everything-in-one-click`), so both links now land at the page top. Either rename that heading back or update the two links in `ai-features.mdx`.

### Inline findings, by cost to the reader

| Line | Issue |
| --- | --- |
| 210–212 | Mitigation contradicts the table above it and *narrows* a survey's audience if followed |
| 196 | `setLanguage()` is not equivalent to `?lang=` — no canonical pass, and a miss skips the survey |
| 115 | "Nothing is stored until you save" — the modal's Save doesn't persist; leaving the editor loses the work |
| 42, 164 | The `de` / `de-DE` alias example is rejected by validation; `?lang=de` resolves canonically, not by alias |
| 77 | Translation completeness gates publish, not a draft save |
| 173 | `ogImage` is a single value, not per-language |

### Checked and fine

One thing worth recording so it isn't re-flagged: the `⋮` menu on the default-language row (`language-view.tsx:411`) is labelled **Change default**, which looks like it contradicts the "one-way decision" Note at line 66. It doesn't — `handleChangeDefault` opens a destructive confirmation and then strips every non-default language and deactivates multi-language, i.e. it's a shortcut to exactly the flow the Note describes. The Note is accurate as written; mentioning the shortcut would be a nice-to-have, not a fix.

### Suggestion

The claims that broke are the ones where two code paths differ (link vs. SDK resolution, draft vs. publish validation, i18n vs. plain field, local vs. persisted state). Where the page states an equivalence — "exactly like", "either/or", "saved or published" — it's worth confirming both halves share an implementation, since this page's readers act on those sentences in code.

Nothing here needs new tests; per AGENTS.md a docs change carries no coverage obligation.

---

jobenjada:
All nine findings accepted and fixed in 7efd20c (plus ddca026 for the first three). Each one was checked against the code before changing the prose — summary of what was wrong and why:

| Finding | Verdict | Fix |
| --- | --- | --- |
| `#translate-with-ai` anchor lost | Valid — two links in `ai-features.mdx` (lines 20, 44) target it, and `Mintlify link-rot` only validates paths, not fragments | Subsection renamed back to `### Translate with AI`, and deliberately left in title case when the other headings went to sentence case |
| Link preview image is not per-language | Valid — `ogImage` is a plain `ZStorageUrl` read unlocalized in `metadata-utils.ts:83` | Note now says title and description are per-language and the image is shared |
| "add the language invisible" mitigation | Valid, and worse than reported: `!selectedLanguage.enabled` returns `undefined` exactly like an unmatched language, so it produces the very skip it claimed to avoid | Sentence removed; the resolution table already states the outcome |
| Bare code cannot be an alias | Valid — the catalog holds both `de` and `de-DE`, so `validateLanguages` rejects alias `de` for `de-DE`. This also invalidated the alias-based workaround I had written for the SDK | Example is now `en_us`, the rule is stated, and the SDK note tells readers to pass the identifier instead |
| Completeness gates publish, not draft save | Valid — `handleSurveySaveDraft` (`:372`) runs neither `validateSurveyWithZod` nor `isSurveyValid`; only `handleSurveySave` (`:402`) and publish do | Scoped to publishing and to saving an already-live survey |
| Modal **Save** does not persist | Valid, and the costliest of the set — `handleSave` calls `setLocalSurvey` only | Now: "That hands the translations back to the editor; they are only stored once you save or publish the survey itself" |
| `?lang=de` uses the canonical pass | Valid — `normalizeLanguageCode("de")` → `de-DE`, so it could not have been the alias pass given the rule above | Relabelled so the three passes stay distinct |
| Heading capitalization | Valid per the path instruction | Sentence case throughout, with the one anchor exception above |

One finding not raised, found while checking the alias rule: the `setLanguage()` snippet passed `"de"` — precisely the value that silently hides a survey whose language is stored as `de-DE`. It now passes the identifier.

Worth flagging beyond this PR: the two language resolvers disagree. `survey-renderer.tsx` matches `?lang=` by code → alias → canonical (added so links survived the `canonicalize_language_codes` migration), while `js-core`'s `getLanguageCode` matches code or alias only — it ships with no runtime deps, so it cannot reach `@formbricks/i18n-utils`. After that migration relabelled `ar` → `ar-EG` without keeping the old code as an alias, `setLanguage("ar")` matches nothing and the app survey is skipped behind a single `logger.debug`, while `?lang=ar` still resolves. The migration's own operator log assumes the SDK "falls back to the survey default", which is not what the code does, and the js-core unit tests still use pre-migration bare codes (`en`, `fr`), so the case is never exercised. This PR documents the behaviour; fixing it is a code change for its own ticket.

mattinannt (approved):

PR #8979, opened 2026-08-25
feat: group question type by category in the block dropdowns (ENG-1777)

jobenjada:
Fixes ENG-1777, Fixes ENG-2141

## What & why

The "Add Block" picker groups question types into labelled sections (Input, Choice, Rating & Scoring,
Contact, Content) in a fixed order. Two other pickers offering the exact same choice — the "Add question
to block" button, and the "Change question type" / "Add question below" submenus in a question's "..."
menu — rendered a flat, differently-ordered list instead, which read as inconsistent.

`elements.tsx` already exported the category config (`getElementCategories`) and the grouping logic used
by "Add Block"; it just wasn't reused anywhere else. This PR extracts that grouping into a shared
`getGroupedElementTypes(t, isCxMode)` helper and switches all three flat pickers to render the same
section headers, in the same order — no color coding, just the section name, per the ask.

While capturing before/after screenshots for that change, the "Add question to block" menu (17 items)
reproduced [ENG-2141](https://linear.app/formbricks/issue/ENG-2141/long-dropdown-menus-render-items-above-the-viewport-with-no-way-to)
exactly: `DropdownMenuContent`/`DropdownMenuSubContent` set `overflow-hidden` with no `max-height`, so when
Radix flips a long menu upward against a short viewport, the items that don't fit render **above** the
top of the screen with no scrollbar to reach them. Fixed by capping both at
`max-h-[min(20rem,var(--radix-dropdown-menu-content-available-height))]` with `overflow-y-auto` — the
same pattern the `Select` component already uses for this exact Radix limitation.

## Breaking changes

None

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| "Add question to block" groups items into Input / Choice / Rating & Scoring / Contact / Content, same order as "Add Block" | manual | Matches; see screenshots below |
| "Change question type" submenu groups items the same way (current type excluded from the list) | manual | Matches; see screenshots below |
| "Add question below" submenu groups items the same way | manual | Same rendering path as "Change question type" — verified visually, no separate screenshot |
| Long menus (17+ items) no longer render items above the viewport with no way to reach them (ENG-2141) | manual | Reproduced the ticket's exact repro steps (1280×720 window, "Add question to block"); see screenshots below |

| Before | After |
| --- | --- |
| Add question to block: flat, uncategorized list | Add question to block: grouped into Input/Choice/Rating & Scoring/Contact/Content sections |
| Change question type: flat, uncategorized submenu | Change question type: grouped submenu matching Add Block |
| ENG-2141: "Free text" rendered above the viewport at y=-25, unreachable | ENG-2141: menu capped to the viewport height with a scrollbar, every item reachable |

**Open gaps**

- No automated regression test guards the `max-height`/`overflow-y-auto` fix for ENG-2141 — a future edit
  to `dropdown-menu/index.tsx` that drops those classes wouldn't be caught by CI, only by manual QA.

**Risks**

- `DropdownMenuContent`/`DropdownMenuSubContent` are shared by every dropdown menu in the app (survey list
  actions, workflow list/canvas menus, the API key modal, chart/dashboard dropdowns, conditional logic,
  etc.). None currently render more than `20rem` of content in normal use, but any that do will now scroll
  instead of growing unbounded — worth a re-check if a reviewer knows of one that relies on that.
- `add-api-key-modal.tsx` already sets its own `max-h-[300px] overflow-y-auto` override on
  `DropdownMenuContent`; untouched by this change since the explicit `className` wins over the new default.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5` (Claude Code), reasoning effort `unknown`.

### DIFF
--- a/apps/web/modules/survey/editor/components/add-element-to-block-button.tsx
+++ b/apps/web/modules/survey/editor/components/add-element-to-block-button.tsx
@@ -7,16 +7,13 @@ import toast from "react-hot-toast";
 import { useTranslation } from "react-i18next";
 import { type Workspace } from "@formbricks/database/prisma-browser";
 import { TSurveyBlock } from "@formbricks/types/surveys/blocks";
-import { TSurveyElementTypeEnum } from "@formbricks/types/surveys/elements";
 import { TSurvey } from "@formbricks/types/surveys/types";
 import { addMultiLanguageLabels, extractLanguageCodes } from "@/lib/i18n/utils";
 import { addElementToBlock } from "@/modules/survey/editor/lib/blocks";
 import { scrollElementCardIntoView } from "@/modules/survey/editor/lib/utils";
 import {
-  getCXElementNameMap,
   getElementDefaults,
-  getElementIconMap,
-  getElementNameMap,
+  getGroupedElementTypes,
   universalElementPresets,
 } from "@/modules/survey/lib/elements";
 import { Button } from "@/modules/ui/components/button";
@@ -24,6 +21,8 @@ import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from "@/modules/ui/components/dropdown-menu";
 
@@ -46,8 +45,7 @@ export const AddElementToBlockButton = ({
 }: AddElementToBlockButtonProps) => {
   const { t } = useTranslation();
   const [open, setOpen] = useState(false);
-  const availableElementTypes = isCxMode ? getCXElementNameMap(t) : getElementNameMap(t);
-  const ELEMENTS_ICON_MAP = getElementIconMap(t);
+  const groupedElementTypes = getGroupedElementTypes(t, isCxMode);
 
   const handleAddElement = (elementType: string) => {
     // Get language symbols and add multi-language support
@@ -91,11 +89,22 @@ export const AddElementToBlockButton = ({
         </Button>
       </DropdownMenuTrigger>
       <DropdownMenuContent align="start">
-        {Object.entries(availableElementTypes).map(([type, name]) => (
-          <DropdownMenuItem key={type} className="min-h-8" onClick={() => handleAddElement(type)}>
-            {ELEMENTS_ICON_MAP[type as TSurveyElementTypeEnum]}
-            <span className="ml-2">{name}</span>
-          </DropdownMenuItem>
+        {groupedElementTypes.map((group, index) => (
+          <div key={group.category.id}>
+            {index > 0 && <DropdownMenuSeparator />}
+            <DropdownMenuLabel className="pt-2 text-xs font-semibold tracking-wide text-slate-500 uppercase">
+              {group.category.label}
+            </DropdownMenuLabel>
+            {group.elements.map((elementType) => (
+              <DropdownMenuItem
+                key={elementType.id}
+                className="min-h-8"
+                onClick={() => handleAddElement(elementType.id)}>
+                <elementType.icon className="size-4" />
+                <span className="ml-2">{elementType.label}</span>
+              </DropdownMenuItem>
+            ))}
+          </div>
         ))}
       </DropdownMenuContent>
     </DropdownMenu>
--- a/apps/web/modules/survey/editor/components/editor-card-menu.tsx
+++ b/apps/web/modules/survey/editor/components/editor-card-menu.tsx
@@ -10,18 +10,15 @@ import { TSurveyBlockLogic } from "@formbricks/types/surveys/blocks";
 import { TSurveyElement, TSurveyElementTypeEnum } from "@formbricks/types/surveys/elements";
 import { TSurvey, TSurveyEndScreenCard, TSurveyRedirectUrlCard } from "@formbricks/types/surveys/types";
 import { getElementsFromBlocks } from "@/modules/survey/lib/client-utils";
-import {
-  getCXElementNameMap,
-  getElementDefaults,
-  getElementIconMap,
-  getElementNameMap,
-} from "@/modules/survey/lib/elements";
+import { getElementDefaults, getGroupedElementTypes } from "@/modules/survey/lib/elements";
 import { Button } from "@/modules/ui/components/button";
 import { ConfirmationModal } from "@/modules/ui/components/confirmation-modal";
 import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
+  DropdownMenuLabel,
+  DropdownMenuSeparator,
   DropdownMenuSub,
   DropdownMenuSubContent,
   DropdownMenuSubTrigger,
@@ -73,7 +70,6 @@ export const EditorCardMenu = ({
   isCxMode = false,
 }: EditorCardMenuProps) => {
   const { t } = useTranslation();
-  const ELEMENTS_ICON_MAP = getElementIconMap(t);
   const [logicWarningModal, setLogicWarningModal] = useState(false);
   const [changeToType, setChangeToType] = useState(() => {
     if (card.type !== "endScreen" && card.type !== "redirectToUrl") {
@@ -87,7 +83,7 @@ export const EditorCardMenu = ({
   const isDeleteDisabled =
     cardType === "element" ? elements.length === 1 : survey.type === "link" && survey.endings.length === 1;
 
-  const availableElementTypes = isCxMode ? getCXElementNameMap(t) : getElementNameMap(t);
+  const groupedElementTypes = getGroupedElementTypes(t, isCxMode);
 
   const changeElementType = (type?: TSurveyElementTypeEnum) => {
     if (!type) return;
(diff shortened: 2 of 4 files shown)


kiln-agent (commented):
## Review

Good change, and the ENG-2141 fix is the right one — I verified the mechanism rather than trusting the pattern:

- `@radix-ui/react-dropdown-menu@2.1.16` re-namespaces `--radix-popper-available-height` to `--radix-dropdown-menu-content-available-height` on **both** `Content` and `SubContent` (`node_modules/@radix-ui/react-dropdown-menu/dist/index.mjs`), so the `SubContent` half of the fix is real and not a no-op.
- In `@radix-ui/react-popper`, the `size()` middleware that sets that variable is unconditional — unlike `shift()`/`flip()` it is not gated on `avoidCollisions` — so no consumer can end up with an undefined var (which would have invalidated the whole `min()` and silently dropped the cap).
- `cn` is `twMerge(clsx(...))`, and `max-h-[300px]` / `max-h-[min(...)]` are in the same tailwind-merge group, so the `add-api-key-modal.tsx` override does win as the description claims.
- Incidental win worth noting: `SettingsSidebarContent.tsx:251` sets `max-h-[300px]` with **no** overflow class, so under the old `overflow-hidden` its overflowing items were clipped and unreachable. That menu becomes scrollable for free here.

Four things below; none is a correctness bug.

### 1. The grouping logic was duplicated, not extracted (`elements.tsx` / `add-element-button.tsx`)

The description says `elements.tsx` "already exported … the grouping logic used by 'Add Block'" and that this PR "extracts that grouping into a shared helper". It doesn't — `getGroupedElementTypes` is new, and `add-element-button.tsx` (the "Add Block" picker, the reference implementation) still carries its own byte-for-byte-equivalent copy at lines 41–52:

```ts
const elementsByCategory = new Map<string, TElement[]>();
for (const elementType of availableElementTypes) { ... }
const visibleCategories = categories.filter((c) => (elementsByCategory.get(c.id)?.length ?? 0) > 0);
```

So the source of truth for "which categories, in what order, with what contents" now exists twice, in a PR whose entire purpose is that these pickers can't drift apart. The two-column layout isn't an obstacle — the helper returns exactly what that component needs:

```ts
const groups = getGroupedElementTypes(t, isCxMode);
// renderCategory takes the group instead of looking the elements up again
<div>{groups.filter((g) => g.category.column === 1).map(renderGroup)}</div>
<div>{groups.filter((g) => g.category.column === 2).map(renderGroup)}</div>
```

Either make that switch, or reword the description so it doesn't claim a de-duplication that didn't happen.

### 2. Dead code: `getElementNameMap` and `getCXElementNameMap`

These were the only two consumers. After this diff both helpers have **zero** callers anywhere in `apps/` or `packages/` (`getElementIconMap` is still used by four other files and stays). `elements.tsx` is app-internal, so there's no external API to preserve — please delete them (`elements.tsx:407` and `:425`) per the "remove dead code" line in AGENTS.md.

### 3. The 20rem cap now bites the menus this PR just made taller

`min(20rem, …)` caps at 320px on *every* viewport, not just short ones. Grouping added five labels and four separators to the "Add question to block" menu — roughly 17 × 32px of items plus ~175px of new chrome, ~730px total. Before this PR that menu rendered in full on a 1080p screen; after it, it's a 320px scroll box showing under half its content, on every screen. The screenshots only cover 1280×720, where the cap is doing its job.

That may well be the intended trade (it matches `Select`), but it's the most user-visible consequence of the diff and it isn't in the Coverage table. Two alternatives if scanability matters more than the 320px convention here:

- Drop the fixed cap on `Content` and bound by viewport only: `max-h-[var(--radix-dropdown-menu-content-available-height)]` — fixes ENG-2141 without shrinking anything that used to fit.
- Keep 20rem as the shared default and let the three element pickers opt into a taller cap via `className="max-h-[min(32rem,var(--radix-dropdown-menu-content-available-height))]"`, which twMerge handles cleanly (see above).

Worth a screenshot at a tall viewport either way.

### 4. Nits

- The ~15-line group-render block is now pasted three times across two files, including the header class string `pt-2 text-xs font-semibold tracking-wide text-slate-500 uppercase`. Same drift risk as (1), one level down — a small shared `ElementTypeGroups` component (or at minimum a shared class constant) would collapse all three.
- `DropdownMenuGroup` is exported from `dropdown-menu/index.tsx` and unused; the new code wraps items in a bare `<div>` inside `role="menu"`. Radix's collection-based roving focus and typeahead are unaffected, so nothing breaks, and `editor-card-menu.tsx` already had a bare `<div className="flex flex-col">` wrapper — but `DropdownMenuGroup` is the idiomatic wrapper and gives the sections a `role="group"`.
- The `index > 0 && <DropdownMenuSeparator />` sits *inside* the group wrapper, so each separator is semantically part of the group it precedes. A fragment (`<Fragment key={…}>`) would place it between groups instead — more relevant if you adopt `DropdownMenuGroup`.

### Checked and clear

- Item sets are unchanged in both modes: `getGroupedElementTypes` draws from the same `getElementTypes` / `getCXElementTypes` that the old name maps reduced over, so only ordering and section headers changed.
- The "Change question type" filter still excludes `card.type`, and categories that empty out as a result are dropped — no empty header can render.
- No Playwright spec targets these menus by item position, so the reordering can't break e2e.
- No new i18n keys; the category labels already exist and go through `t()`.
- `pt-2` vs the label's default `py-1.5`, and `text-slate-500` vs `text-slate-900`, both resolve the way you want under twMerge + Tailwind v4 utility ordering.

CI is fully green (typecheck, lint, unit, e2e, Sonar quality gate).

---

kiln-agent (commented):
## Review — grouping question types in the block dropdowns

**Verdict: good to merge with nits.** The regrouping is set-preserving (I checked the element lists really do match), the ENG-2141 cap targets the right Radix variable on both content types, and the two failure modes I most suspected did not reproduce. Findings below: one latent-robustness issue, one half-finished extraction, two nits, and an answer to the risk you flagged.

### What I tried to break and couldn't

1. **No element type is dropped or gained by the refactor.** `getElementNameMap` / `getCXElementNameMap` reduce over `getElementTypes` / `getCXElementTypes` — the exact lists `getGroupedElementTypes` now consumes — and all five `TElementCategory` members are registered in `getElementCategories`. Same set in all three pickers, only re-ordered.
2. **The Radix variable exists on `SubContent`, not just `Content`.** `--radix-dropdown-menu-content-available-height` is written by both (`@radix-ui/react-dropdown-menu` dist lines 130 and 245), and `--radix-popper-available-height` comes from the `size` middleware, which is not gated on `avoidCollisions` — so the cap works on the submenus too, and degrades to "no cap" rather than "zero height" if the var ever goes missing.
3. **Submenus are *not* clipped by the new `overflow` on the parent content.** This was my main worry, and it's non-obvious: `MenuSubContent` renders no portal, so a `DropdownMenuSubContent` is a DOM descendant of the `DropdownMenuContent` that now carries `overflow-x-hidden overflow-y-auto` — and `editor-card-menu.tsx` is exactly that shape. I replicated Radix's box structure in Chromium (fixed + transformed popper wrapper → height-capped overflow content → fixed sub-wrapper) and hit-tested both corners of a submenu deliberately overflowing the parent on both axes: fully painted, not clipped. The fixed sub-wrapper's containing block is the transformed *ancestor*, so the intervening scroll box has no claim on it. No action needed — recording it so nobody has to re-derive it.

### 1. `getGroupedElementTypes` silently drops elements whose category isn't registered

`apps/web/modules/survey/lib/elements.tsx:382`

The helper builds `elementsByCategory` from the element list, then iterates `getElementCategories(t)` to read it back — so anything whose `category` has no entry there is never read out at all. Today all five enum members are registered, so there is no live bug. But it changes the failure mode of three pickers from "always lists every element type" (the old `Record`-based maps had no category dependency whatsoever) to "lists every element type whose category happens to be registered". Add a sixth `TElementCategory` and a question type using it, forget the `getElementCategories` entry, and that question type vanishes from *Add question to block*, *Change question type* and *Add question below* simultaneously — with nothing failing: not typecheck, not lint, not Sonar, and not the E2E suite.

Cheapest guard is a unit test on the pure function — this is the "pure logic and edge cases" case in AGENTS.md; the file is `.tsx` only incidentally:

```ts
it("groups every available element type, dropping none", () => {
  const all = getElementTypes(t).map((e) => e.id).sort();
  const grouped = getGroupedElementTypes(t).flatMap((g) => g.elements).map((e) => e.id).sort();
  expect(grouped).toEqual(all);
});
```

plus the `isCxMode` variant against `getCXElementTypes`. That also retires part of your `Open gaps` note. (A catch-all group for unregistered categories would work too, but the test is the better trade — it fails at the moment the mistake is made rather than rendering something plausible.)

### 2. The extraction is half-done — `add-element-button.tsx` still carries its own copy

The description says the grouping logic "used by 'Add Block' … just wasn't reused anywhere else. This PR extracts that grouping into a shared helper." The Add Block picker itself was never switched over: `apps/web/modules/survey/editor/components/add-element-button.tsx:43-53` still holds the same `elementsByCategory` loop and its own `visibleCategories` filter, line for line. So the new single source of truth ships alongside a second copy, and any drift between them is invisible precisely because the goal is that all four pickers agree on order.

The two-column layout isn't an obstacle — the column lives on the category:

```ts
const groups = getGroupedElementTypes(t, isCxMode);
// ...
<div>{groups.filter((g) => g.category.column === 1).map(renderGroup)}</div>
<div>{groups.filter((g) => g.category.column === 2).map(renderGroup)}</div>
```

Either make that switch, or reword the description so it doesn't claim an extraction that didn't land.

### 3. Nit — the group-rendering JSX is pasted three times

The ~12-line `<div key={group.category.id}>` + separator + `DropdownMenuLabel` + item loop appears at `add-element-to-block-button.tsx:93` and `editor-card-menu.tsx:243` and `:288`, including the same 74-character label className each time. One small local component taking the groups and an item renderer collapses all three — and gives a single place to type the icon: `TElement.icon` is `any`, so `<elementType.icon className="size-4" />` is unchecked at all three new call sites, where the old `getElementIconMap` path returned a typed `Record<TSurveyElementTypeEnum, JSX.Element>`.

### 4. Nit — `DropdownMenuGroup` rather than a bare `<div>`

All three new blocks wrap their items in `<div key={group.category.id}>`. `role="menu"` owns only `menuitem`, `group` and `separator`, so a bare `div` isn't a valid child. `DropdownMenuGroup` is already exported from the same module (and used in `SettingsSidebarContent.tsx:220`), and Radix renders it as `role="group"` (`@radix-ui/react-menu` dist line 352) — a drop-in:

```diff
-<div key={group.category.id}>
+<DropdownMenuGroup key={group.category.id}>
```

Nothing changes functionally: Radix's collection orders items by document position, so roving focus and typeahead already work through the wrapper.

### 5. Answering the risk you raised — one dropdown does rely on unbounded growth

You asked whether any dropdown renders more than `20rem`. `recall-item-select.tsx:178` does: heading + `Input` (`h-10`) + a list capped at `max-h-72`, inside `p-3` with `gap-2` — about 383px of natural content. It does **not** break, because the content is `flex flex-col` and the inner list is a scroll container, so its `min-height: auto` resolves to 0 and it shrinks instead of overflowing. Measured both variants in Chromium:

| | content height | inner list viewport | nested scrollbars |
| --- | --- | --- | --- |
| before | 383px | 288px | no |
| after | 318px | 223px | no |

The search box stays put and there's still exactly one scrollbar — the recall list just shows about two fewer rows. Fine by me, but it is a real visual change to a surface the PR doesn't mention, so worth a line in Coverage or a glance in QA.

---

Nothing here blocks merge; #1 and #2 are the two I'd want addressed before it lands.

---

mattinannt (approved):

PR #8982, opened 2026-08-25
fix(surveys): stop announcing the public-API name placeholder on app surveys [ENG-2617]

Dhruwang:
Fixes ENG-2617

## What & why

On an app-type survey delivered by the js-core widget, the modal's accessible name and its `sr-only` `h1` both read `[deprecated] survey name omitted from public API - will be removed soon` instead of the survey's name — so that sentence is what a screen reader announces as the dialog's name and as its only heading.

The widget reads its surveys from the public client API, which deliberately substitutes that fixed placeholder for every survey name so names aren't exposed over an unauthenticated endpoint (ENG-808). The dialog name and the heading added for ENG-2334 are both taken from `survey.name`, and neither accounted for the substitution. Link surveys are unaffected — their name comes from the server component, never from this endpoint.

- `getSurveyDisplayName` (new, in `packages/surveys/src/lib/survey-page.ts`) returns `undefined` for the placeholder — and for a missing or whitespace-only name — so the renderer treats it as "this survey has no name" and falls back to the generic `Survey Dialog` label it already uses for a survey rendered without one (previews).
- `RenderSurvey` passes `survey.name` through that helper. It's the single choke point: `surveyName` reaches the dialog's `aria-label`, the `sr-only` `h1`, the `role="form"` landmark and `aria-describedby` from this one prop.
- The placeholder string becomes a shared constant so the API side and the widget cannot drift apart. It lives in a new **dependency-free** `packages/types/js-constants.ts` rather than in `types/js.ts`: a *value* import of `types/js` drags that module's zod schema graph into the widget bundle — measured at **+94 kB** on the UMD build every app-survey respondent downloads. The leaf module costs 140 bytes.

## Breaking changes

None. The public client API's response is byte-identical — the placeholder is now referenced through a constant instead of an inline literal. Nothing changes for API/SDK consumers or self-hosters; the only behaviour change is what assistive tech reads out in the rendered widget.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The placeholder is not treated as a survey name (and a real name, including one that merely contains the placeholder, still is) | unit (mutation) | 4 cases in `packages/surveys/src/lib/survey-page.test.ts`. The code under test is new, so mutation is the honest strength: reverting the guard at `packages/surveys/src/lib/survey-page.ts:65` to `if (!name) return undefined;` turns *"drops the public client API's placeholder"* red — `pnpm --filter @formbricks/surveys test` → **1 failed / 724 passed**; restored → **725 passed (29 files)**. |
| End-to-end, against the real public client API: the widget's dialog is not named with the placeholder | e2e | Two assertions added to the **existing** widget journey in `apps/web/playwright/js.spec.ts` (no new spec): the dialog's `aria-label` is `Survey Dialog`, and no node in the widget carries the placeholder text. Red on main by construction — that's exactly the string main renders. Not executed locally, see Open gaps. |
| The public client API still substitutes the placeholder — the security behaviour ENG-808 added is untouched | unit (guard) | `apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.test.ts` now asserts against the shared constant rather than a copied literal, so the two can't drift. 18/18 pass. |
| The widget bundle does not grow | manual | Built `@formbricks/surveys` on each revision: baseline UMD **971,447 B** → this PR **971,587 B** (**+140 B**). An earlier revision importing the constant from `@formbricks/types/js` measured **1,065,940 B** (**+94 kB, +9.7%**) — that measurement is why the constant lives in a leaf module. |

Full gates on the final tree: `pnpm typecheck` 27/27, `pnpm lint` 0 errors, `pnpm format:check` clean, `pnpm test` 26/26 tasks (8,529 web + 725 surveys).

**Open gaps**

- **Playwright was not run.** This environment has no Docker daemon, so Postgres/Redis can't boot and the suite can't start. The two new `js.spec.ts` assertions are therefore unverified locally and rest on the PR's Playwright job — worth a look at that job's result before merging, since it's the only check that exercises the real API → widget → dialog path this bug lives on.
- **The `sr-only` `h1` is now absent on widget-delivered surveys**, because there is no name to put in it. That is the fallback the ticket asks for and matches the documented preview behaviour, but it means WCAG 2.4.6's top-level heading is not present on that delivery path — the placeholder was never a heading worth having, so this is strictly better than before, not complete. Restoring a real heading there requires an API-side decision about exposing survey names to the public client endpoint (the deliberate ENG-808 trade-off), which is out of scope for a QA fix.
- Screen-reader verification was done by reading the accessibility attributes, not by listening with an actual screen reader.

**Risks**

- Inline (non-modal) app surveys delivered by the widget lose their `role="form"` landmark, since the container only declares the role once there is a name to give it. That's the pre-existing, deliberate behaviour for an unnamed survey — an unnamed form landmark is noise in a landmark list — but it is a change on that path, and it goes away on its own if survey names ever become available to the widget.
- `getSurveyDisplayName` matches the placeholder **exactly**; if the API-side string is ever edited without updating the shared constant, the placeholder starts leaking again. That's the failure mode the shared constant exists to prevent, and both call sites now import it.

### DIFF
--- a/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.test.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.test.ts
@@ -3,6 +3,7 @@ import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
 import { logger } from "@formbricks/logger";
 import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
 import { getWorkspaceStateData } from "./data";
 
 vi.mock("server-only", () => ({}));
@@ -120,7 +121,7 @@ describe("getWorkspaceStateData", () => {
       surveys: [
         {
           ...mockWorkspaceData.surveys[0],
-          name: "[deprecated] survey name omitted from public API - will be removed soon",
+          name: PUBLIC_API_SURVEY_NAME_PLACEHOLDER,
         },
       ],
       actionClasses: mockWorkspaceData.actionClasses,
--- a/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts
@@ -9,6 +9,7 @@ import {
   TJsWorkspaceStateSurvey,
   TJsWorkspaceStateWorkspaceSetting,
 } from "@formbricks/types/js";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
 import { type TBaseFilters, buildSurveyInteractionRefreshMap } from "@formbricks/types/segment";
 import { toLegacyLanguageCodes } from "@/lib/i18n/utils";
 import { validateInputs } from "@/lib/utils/validate";
@@ -283,7 +284,7 @@ export const getWorkspaceStateData = async (workspaceId: string): Promise<Worksp
 
       return {
         ...transformed,
-        name: "[deprecated] survey name omitted from public API - will be removed soon",
+        name: PUBLIC_API_SURVEY_NAME_PLACEHOLDER,
         segment: sanitizedSegment,
         ...(interactionRefresh ? { interactionRefresh } : {}),
       };
--- a/apps/web/playwright/js.spec.ts
+++ b/apps/web/playwright/js.spec.ts
@@ -103,6 +103,16 @@ test.describe("JS Package Test", async () => {
 
     await page.goto("http://localhost:3004");
     await expect(page.locator("#formbricks-modal-container")).toHaveCount(1, { timeout: 120000 });
+
+    // The widget reads the survey from the public client API, which substitutes a placeholder for
+    // every survey name so names are not exposed over an unauthenticated endpoint. This is the only
+    // place the real API, the widget and the dialog are wired together, so it is the only place that
+    // can catch the placeholder leaking into what a screen reader announces: the dialog falls back
+    // to its generic name instead, and no heading carries the placeholder either.
+    const widget = page.locator("#formbricks-modal-container");
+    await expect(widget.getByRole("dialog")).toHaveAttribute("aria-label", "Survey Dialog");
+    await expect(widget.getByText(/\[deprecated] survey name omitted/)).toHaveCount(0);
+
     await expect(
       page.locator("#questionCard-0").getByRole("link", { name: "Powered by Formbricks" })
     ).toBeVisible();
--- a/packages/surveys/src/components/general/render-survey.tsx
+++ b/packages/surveys/src/components/general/render-survey.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 import { SurveyContainerProps } from "@formbricks/types/formbricks-surveys";
-import { hasSurveyInstructions } from "@/lib/survey-page";
+import { getSurveyDisplayName, hasSurveyInstructions } from "@/lib/survey-page";
 import { getSurveyLanguageTag, isRTLLanguage } from "@/lib/utils";
 import { SurveyContainer } from "../wrappers/survey-container";
 import { Survey } from "./survey";
@@ -106,7 +106,7 @@ export function RenderSurvey(props: Readonly<SurveyContainerProps>) {
       onClose={close}
       isOpen={isOpen}
       dir={dir}
-      surveyName={props.survey.name}
+      surveyName={getSurveyDisplayName(props.survey.name)}
       hasInstructions={hasSurveyInstructions(props.survey)}
       lang={languageTag}>
       <Survey
--- a/packages/surveys/src/lib/survey-page.test.ts
+++ b/packages/surveys/src/lib/survey-page.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, test } from "vitest";
 import type { TJsWorkspaceStateSurvey } from "@formbricks/types/js";
-import { getSurveyPagePosition, hasSurveyInstructions } from "./survey-page";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
+import { getSurveyDisplayName, getSurveyPagePosition, hasSurveyInstructions } from "./survey-page";
 
 const block = (id: string) => ({ id, name: id, elements: [] });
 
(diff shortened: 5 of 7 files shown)


kiln-agent (commented):
## Review

The bug is real, the diagnosis is right, and the choke point is the right one. Nothing here is a blocker. One thing wants a human decision before merge, and it is the framing of that decision in the description rather than the code.

### 1. The fallback drops three a11y affordances where a generic name would keep two of them — and the stated reason for not doing that isn't accurate (medium)

`getSurveyDisplayName` returning `undefined` is correct. What follows from it in `SurveyContainer` is a wider revert than the description accounts for. With no name, on the widget path:

| | modal (app survey, the common case) | inline |
| --- | --- | --- |
| dialog `aria-label` | falls back to `Survey Dialog` ✅ | n/a |
| `sr-only` `h1` | gone (`survey-container.tsx:214`) | gone |
| `role="form"` landmark | n/a | gone (`survey-container.tsx:241`) |
| `aria-describedby` → instructions | unaffected — gated on `hasInstructions` alone (`:271`) | gone (`:231`) |

Two notes on the description against that table:

- **`aria-describedby` is not a single choke point off `surveyName`.** On the modal path it never was; on the inline path it now goes away, which `Risks` doesn't list (it lists only the landmark). The inline loss is nominal — `aria-describedby` on a role-less `div` is largely ignored anyway — but it is a fourth thing moving, not three.
- **`Open gaps` says restoring a heading "requires an API-side decision about exposing survey names to the public client endpoint."** That isn't so for a *generic* heading. `SurveyContainer` already has `t("common.survey_dialog")` wired up 57 lines below the heading, and `useTranslation()` is already in scope, so:

  ```tsx
  const displayName = surveyName ?? t("common.survey_dialog");
  const surveyHeading = <h1 className="sr-only">{displayName}</h1>;
  // inline: role="form" unconditionally, aria-label={displayName}
  ```

  keeps WCAG 2.4.6's top-level heading and the named form landmark on the widget path with no API change at all. ENG-2334's own rationale — *"a screen reader user pressing H would land inside the survey with nothing naming what they are answering"* — is exactly the state app surveys return to under this diff, and app delivery is the majority path for that heading.

Whether a generic "Survey" heading is better than none is a product a11y call, and "an unnamed landmark is noise" is a defensible answer to it (it's the argument already encoded at `:241`). So this is not a defect — it's a decision. Please either take the generic fallback or record in `Open gaps` that a generic name **was** available and was rejected on purpose, because the current wording reads as "no option existed."

### 2. `data.test.ts` no longer pins the wire value (low)

Swapping the literal for the constant makes the string comparison a tautology: both sides now move together, so nothing in the repo fails if the placeholder's text changes. The *security* behaviour ENG-808 added is still covered and can still fail — remove the substitution and the assertion sees the real survey name — so this is narrow. But the placeholder is a value in an unauthenticated public API response, and it used to be pinned. One literal assertion somewhere (`js-constants` or that test) would make a change to it a deliberate, visible diff instead of a silent one. Given the string is documented as a marker nobody should read, your call whether it's worth the line.

### 3. Nit: `getSurveyDisplayName` returns the name untrimmed

`survey-page.ts:65-67` checks `name.trim().length > 0` but returns `name`, so `"  Product feedback  "` lands in `aria-label` and the `h1` with its padding. `return name.trim() || undefined` after the placeholder check is the whole function and normalises what ships.

### 4. Note: nothing enforces the reason `js-constants.ts` exists

There's no bundle-size gate in `.github/workflows/`, so the +94 kB trap is guarded only by the doc comments in `js-constants.ts` and `survey-page.ts`. They're clear and well-placed, and the next person to "tidy" the constant into `js.ts` will most likely read them — but a size budget on `surveys.umd.cjs` is what would make it durable. Follow-up at most; not this PR's job.

### Checked and clean

- **Mutation claim reproduced.** Reverting the guard at `survey-page.ts:65` to `if (!name) return undefined;` → `1 failed | 20 passed`; restored → `21 passed`. The test can fail for the stated reason.
- **No drift surface left.** `grep` for the placeholder text across the repo finds only `js-constants.ts` and the new spec assertion — the substitution site is singular and both consumers import the constant.
- **`@formbricks/types/js-constants` resolves** — `exports` maps `"./*": "./*.ts"`, no wiring needed.
- **The existing a11y gate isn't broken.** `survey-accessibility.spec.ts:628` asserts exactly one `h1` carrying the survey name, but it navigates a **link** survey (`seeded.surveyUrl`), whose name comes from the server component. Unaffected, as claimed.
- **`common.survey_dialog` is `"Survey Dialog"` in `packages/surveys/locales/en-US.json`**, so the hardcoded string in the new E2E assertion matches.
- **Single consumer.** `getSurveyDisplayName(props.survey.name)` at `render-survey.tsx:109` is the only `survey.name` read in `packages/surveys`; nothing in `packages/js-core` reads it either.
- **E2E ordering considered and refuted.** `toHaveCount(0)` would pass vacuously on an empty widget, but the `aria-label` assertion ahead of it waits for the dialog, which renders in the same pass as the heading. Not a test that cannot fail.
- Typecheck is green on CI. Unit tests, E2E, lint and Sonar were still running at review time — the Playwright job in particular is the one you flagged as the only real check on this path, so it's worth a look before merge.

---

kiln-agent:
## Review: no correctness findings

Reviewed all 7 files hunk by hunk and traced the affected paths. Nothing blocking; no inline comments to leave. What I verified rather than took on trust:

**The choke point claim holds.** `packages/surveys/src/index.ts` funnels every render mode (modal, inline app, inline link) through `RenderSurvey`, and `props.survey.name` appears nowhere else in `packages/surveys` or `packages/js-core` (the only other `.name` reads there are action classes). So the single call site in `render-survey.tsx` really does cover the dialog `aria-label`, the `sr-only` `h1`, the `role="form"` landmark and `aria-describedby`.

**The placeholder has exactly one producer.** Grepping the literal repo-wide returns only the new constant and the Playwright regex. `js-core` fetches surveys solely from `/api/v1/client/{id}/environment`; `/api/v1/client/{id}/user` returns user state only. Git history shows the string never had another spelling, so cached widget state can't hold a variant the exact-match guard misses.

**The `role="form"` regression listed under Risks is unreachable in production.** `packages/js-core/src/lib/survey/widget.ts:164` passes no `mode`, so app surveys always take the modal branch. The inline branch is reached only from `apps/web/modules/ui/components/survey/index.tsx` and the link-survey wrapper, both of which carry a real DB name. Accurate to list it, but it costs nothing today.

**No existing spec regresses.** `survey-accessibility.spec.ts:628` asserts an `h1` carrying the survey name, but against a link survey (`seeded.surveyUrl`, name from the server component); `survey-keyboard.spec.ts:783` uses a seeded fixture with a real name. The documented `h1` disappearance on widget-delivered surveys breaks no assertion.

**The new E2E assertions are sound.** `role="dialog"` is declared in exactly one place, so strict mode is safe; the `sr-only` `h1` uses clip-based hiding, so `getByText(...).toHaveCount(0)` would still catch a leak rather than passing vacuously.

**Module resolution and the bundle argument check out.** `packages/types/package.json` maps `"./*": "./*.ts"`, and `packages/surveys` already value-imports `@formbricks/types/errors` and `/styling`, so the new leaf module resolves in both the Vite build and vitest. Keeping the constant out of `types/js.ts` to avoid pulling the zod graph into the respondent bundle is the right call and the comment explaining it is worth having.

### One non-blocking observation

`apps/web/playwright/js.spec.ts:113` asserts the *translated* string `"Survey Dialog"`. Editing `common.survey_dialog` in `packages/surveys/locales/en-US.json` would break that E2E with a failure that reads as a widget bug rather than a copy change. Not worth a change now — just something to remember if that key is ever reworded.

### On the Open gaps

The unverified Playwright run is the one thing left: the two new assertions are red on `main` by construction, so the PR's Playwright job is the only evidence that the fix works end to end. Worth confirming that job is green before merge, as the description already says.

---

xernobyl (approved):
Thank you for the PR. Looks good.

PR #7218, opened 2026-02-06
fix: fixing issue with saving follow ups

TheodorTomas:
## What does this PR do?

Fixes the Survey Follow-Up save functionality where creating a new follow-up would silently fail unless the "Attach response data" toggle and its sub-options ("Include variables", "Include hidden fields") were explicitly enabled.

The root cause was that `includeVariables` and `includeHiddenFields` were missing from the `useForm` `defaultValues` in the `FollowUpModal` component. The form's Zod schema (`ZCreateSurveyFollowUpFormSchema`) requires both as `z.boolean()`, but they were never initialized in create mode — leaving them as `undefined`, which silently failed validation. In edit mode, a `useEffect` with `form.reset()` happened to set them correctly, so editing worked fine.

The fix adds both fields to the initial `defaultValues` with a fallback of `false`, matching the unchecked UI state and the existing pattern for `attachResponseData`.

Also adds an E2E test covering follow-up creation (with and without optional toggles) and editing.

Fixes [#1325](https://github.com/formbricks/internal/issues/1325)

Loom recording: https://www.loom.com/share/81605af2c2e143069fbfec36b4789188

## How should this be tested?

- Open any survey in the editor and navigate to the **Follow-ups** tab.
- Click **"New follow-up"** to open the create modal.
- Fill in only the name field. **Do NOT enable** "Attach response data" or any sub-toggles.
- Click **Save** — the follow-up should be created successfully (toast appears, follow-up shows in list).
- Click the newly created follow-up to edit it, change the name, click **Save** — it should update successfully.
- Create another follow-up, this time **enabling** "Attach response data" and checking both "Include variables" and "Include hidden fields" — it should also save successfully.
- Run the new E2E test: `pnpm test:e2e -- --grep "Follow-Up"`

### Required

- [x] Filled out the "How to test" section in this PR
- [x] Read [How we Code at Formbricks](https://formbricks.com/docs/contributing/how-we-code)
- [x] Self-reviewed my own code
- [x] Commented on my code in hard-to-understand bits
- [x] Ran `pnpm build`
- [x] Checked for warnings, there are none
- [x] Removed all `console.logs`
- [x] Merged the latest changes from main onto my branch with `git pull origin main`
- [x] My changes don't cause any responsiveness issues

### Appreciated

- [ ] If a UI change was made: Added a screen recording or screenshots to this PR
- [ ] Updated the Formbricks Docs if changes were necessary

PR #8983, opened 2026-08-25
feat(surveys): stop new follow-ups where Workflows can replace them

jobenjada:
Ref PRO-98

## What & why

Survey Follow-ups are being deprecated in favour of Workflows, which already cover the same
`response completed → send email` automation — `packages/workflows/src/types/actions/send-email.ts`
documents its config as 1:1 field parity with `ZSurveyFollowUpAction`, and the `response.completed`
trigger's `endingCardIds` maps onto the follow-up `response`/`endings` trigger.

**This PR removes nothing.** It stops the migration debt growing, so the converter and the packaging
decision that follow it aren't chasing a moving target.

- **The editor tab now survives only where it has to.** `shouldShowFollowUpsTab` keeps it for a
  survey that already has follow-ups (live automation its owner must be able to read, fix and later
  migrate — including when the entitlement has since lapsed), and for a deployment Workflows cannot
  reach. That last case is the one that matters: `getSurveyFollowUpsPermission` returns `true`
  unconditionally when `IS_FORMBRICKS_CLOUD` is false, so self-hosted community gets follow-ups
  free, while Workflows need an active enterprise license. Hiding the tab there would leave that
  deployment with no path forward at all.
- **The follow-ups upsell is gone.** An org with no follow-ups and no entitlement used to get an
  `UpgradePrompt` selling a deprecated feature. It now sees no tab.
- **The entry point becomes a Workflow** wherever Workflows are available: `+ New follow-up` is
  replaced by `New workflow` (new tab — the survey editor has no autosave, so navigating away would
  drop unsaved edits). Self-hosted community keeps `+ New follow-up`.
- **The soft nudge becomes a real deprecation notice** carrying the **1 December 2026** removal date,
  and an org whose entitlement has lapsed is now told its follow-ups aren't being sent rather than
  being shown an upsell — the runtime already refuses to send them. Each row it still lists is
  muted and carries a **Not sending** badge, so the list stops looking live. Muted rather than red
  or amber: sending is disabled, not broken, and the warning colour on that same row already means
  "this follow-up is misconfigured".

The visibility rule reads the *persisted* `survey.followUps`, not the editor's working copy: the
working copy soft-deletes with a `deleted` flag, so counting it would pull the tab out from under
someone who just deleted their last follow-up and hasn't saved.

<details>
<summary>Not in this PR — what the deprecation still needs</summary>

- **Packaging for self-hosted community.** They have follow-ups free today and no Workflows. You
  cannot deprecate a free feature into a licensed one without either giving the equivalent away or
  taking it away. This PR keeps their tab and buys time; it does not decide.
- **A converter.** Parity is 1:1, so it is a field mapping — but it must delete the follow-up in the
  same transaction, because `process-response-pipeline-job.ts` calls `sendFollowUpsForResponse` and
  `enqueueResponseCompletedWorkflowRuns` independently: a survey with both emails the respondent
  twice.
- **Two conversions that can fail.** Workflows require a literal `to` to belong to a workspace member
  (ENG-2029) and cap subject/body length; follow-ups enforce neither, and v1 management API writes
  can violate both.
- **An editor guard gap.** `edit-ending-card.tsx:135` blocks deleting an ending card referenced by a
  follow-up; there is no equivalent guard for workflows. That should land before anyone is pushed to
  migrate, or migrating moves those references out of the editor's sight.
- **Metering.** Follow-up sends are unmetered; each workflow run emits a `workflow_run_created`
  Stripe meter event.
</details>

## Breaking changes

None — no API, SDK, route, env or config change. Existing follow-ups keep running and stay editable;
only where a *new* one can be started narrows, and only where Workflows already cover it.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Tab survives for a survey that already has follow-ups, incl. lapsed entitlement | unit (mutation) — `apps/web/modules/survey/follow-ups/lib/deprecation.test.ts` | `shouldShowFollowUpsTab` returns true for `followUpCount > 0` under both entitlement states |
| Tab survives with 0 follow-ups only where Workflows can't replace it | unit (mutation) — same file | true for `allowed && !workflows` (self-hosted community); false once workflows are available |
| Empty tab hidden for orgs not entitled to follow-ups (retires the upsell) | unit (mutation) — same file | false for both `!allowed` combinations |
| Announced sunset date | unit (guard) — same file | pinned to `2026-12-01T00:00:00.000Z`, so the copy and the docs can't drift apart |
| Tab strip drops Follow-ups where Workflows are available | manual — real app, see below | before/after pair below |
| Deprecation notice, date, and Workflow CTA render | manual — real app, see below | before/after pair below; the CTA needed a fix to render at all (see Risks) |
| Existing follow-up still opens, edits and saves from the tab | e2e — `apps/web/playwright/survey-follow-up.spec.ts` | rewritten to seed a follow-up via Prisma, then assert the notice and the edit journey; green on CI (11m20s run) |
| Sunset date renders as the announced calendar day in every zone | unit (red on broken) — `deprecation.test.ts` | `Dec 1, 2026` under `TZ=America/Los_Angeles`; dropping `timeZone: "UTC"` turns 3 tests red. `cd apps/web && TZ=America/Los_Angeles npx vitest run modules/survey/follow-ups/lib/deprecation.test.ts` |
| Same date asserted end-to-end, not just the sentence | e2e — `survey-follow-up.spec.ts:62` | asserts the literal `Follow-ups stop working on Dec 1, 2026` |
| Form defaults always satisfy the form schema (#7218 guard) | unit (red on broken) — `form-default-values.test.ts` | re-dropping `includeVariables` from the defaults turns 3 tests red; replaces the deleted Playwright `attachResponseData: false` step |
| Duplicate hidden wherever a new follow-up can't be created | manual — real app, see below | copy icon present on `main`, absent on this branch with Workflows available |
| Workflows CTA actually renders on the real page | manual — real app, see below | the button was gated on `useWorkspace()`, which is null in the `(survey-editor)` route group; now passed the workspace id as a prop |
| Form defaults fall back to translated copy, never a hardcoded string | unit (mutation) — `form-default-values.test.ts` | the modal's reset copy had drifted to a hardcoded `"Thanks for your answers!"`; both paths share one builder now. Mutate `subject:` in `form-default-values.ts:36` to turn it red — re-inlining a literal in the `form.reset` effect (`follow-up-modal.tsx:289`) leaves it green, so the row claims the builder, not both call sites |
| Rows are muted and badged where the entitlement has lapsed | manual — real app | screenshot pending; the badge string and the `isSending` branch are the whole change |

Run the unit tests: `cd apps/web && npx vitest run modules/survey/follow-ups/lib/deprecation.test.ts`

<details>
<summary>Screenshots — real app, before/after</summary>

**The editor tab strip** — survey with no follow-ups, organization with Workflows available

| Before | After |
| --- | --- |
| Editor tab strip ending in a Follow-ups tab | Editor tab strip ending at Settings, no Follow-ups tab |

**The Follow-ups tab** — survey that already has one follow-up, same organization

| Before | After |
| --- | --- |
| Follow-up list with a New follow-up button, a duplicate icon on the row, and a soft Workflows nudge below | Deprecation notice naming Dec 1 2026, a New workflow button, and the row without a duplicate icon |

Captured against the running app on this branch and on `main` with the same Playwright fixture user,
seeded survey, route, viewport (1280x820 @2x) and light theme — the only variable is the code.

</details>

**Open gaps**

- The `activeView` reset and the empty-state gating are both verified by reading the render path, not
  by an automated test — they need a delete-then-save round trip against a live editor, which is an
  E2E-shaped cost this deprecation doesn't warrant.
- The `New workflow` button only lands the user on the workflows list. It does not pre-create a
  workflow or prefill the survey trigger — that belongs with the converter.
- The one-notice-for-everyone choice means a self-hosted community deployment now sees a hard
  sunset date for a feature it has free and cannot yet replace. That is deliberate — softening it
  for them would hide the very problem — but it puts a clock on the packaging decision, and it is a
  one-string change if you'd rather it didn't.

**Risks**

- `canDuplicate` is now a required prop rather than defaulting to `true`. It is the gate this
  branch exists to close, so a future caller omitting it must be a typecheck error, not a silently
  reopened duplicate button.

- The `New workflow` CTA was dead on the real page until 74a9cb8: the editor's route group does not
  mount `WorkspaceContextWrapper`, so `useWorkspace()` returns a null workspace there. Anything else
  in the editor reading that context has the same problem — the follow-ups upgrade prompt this branch
  deletes was building its billing href from `workspace?.organizationId`, so that link was broken too.
- The tab now depends on `getIsWorkflowsEnabled`, so a license-server blip that reports Workflows as
  unavailable makes the tab reappear for a survey with none. That fails open, in the safe direction.
- The old spec created a follow-up from the empty state; on a licensed deployment that path no
  longer exists, so UI creation of follow-ups is only covered where Workflows are unavailable — the
  specific regression it guarded is re-pinned as a unit test on the form defaults.
- `FollowUpModal`'s form defaults moved to `lib/form-default-values.ts`, replacing two
  hand-maintained copies (the `useForm` initializer and the reset effect). Behaviour is unchanged for
  English; other locales gain the translated subject fallback the reset path had lost. A missing
  recipient option now yields `""` rather than `undefined`, which surfaces the
  schema's "To is required" error instead of an invalid form value. In practice the options list
  always contains the current user, so that fallback is unreachable.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `unknown` (not reported in this non-interactive session).


apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
```diff
@@ -34,75 +37,85 @@ export const FollowUpsView = ({
...
-          ]}
-        />
-      </div>
-    );
-  }
+  // Follow-ups are deprecated, so the only place a new one can still be started is a deployment
+  // Workflows cannot reach yet. Everywhere else the entry point is a Workflow — that is what stops
+  // the migration debt growing while nothing is removed.
+  const canCreateFollowUps = isSurveyFollowUpsAllowed && !isWorkflowsAllowed;
+  const sunsetDate = formatDateForDisplay(SURVEY_FOLLOW_UPS_SUNSET_DATE, locale);
```
kiln-agent: The announced sunset date shifts a day for every viewer west of UTC.

`SURVEY_FOLLOW_UPS_SUNSET_DATE` is the instant `2026-12-01T00:00:00.000Z`, and `formatDateForDisplay` builds an `Intl.DateTimeFormat` with no `timeZone`, so it renders in the *browser's* zone:

```
TZ=America/Los_Angeles → "Nov 30, 2026"
TZ=Europe/Berlin       → "1. Dez. 2026"
```

So a US user reads "Follow-ups stop working on Nov 30, 2026" while `email-followups.mdx` and this PR both say **1 December 2026**. AGENTS.md calls this out explicitly: locale controls formatting, time zone controls the represented calendar day, and one must not be inferred from the other. The unit test pins the ISO instant, which cannot catch this, and the e2e assertion is `/Follow-ups stop working on/` with no date.

This is a fixed calendar date, not a moment in time, so pin the zone at render:

```suggestion
  const sunsetDate = formatDateForDisplay(SURVEY_FOLLOW_UPS_SUNSET_DATE, locale, {
    year: "numeric",
    month: "short",
    day: "numeric",
    timeZone: "UTC",
  });
```

---

apps/web/modules/survey/editor/components/survey-editor.tsx
```diff
@@ -277,18 +287,17 @@ export const SurveyEditor = ({
             />
           )}
 
-          {activeView === "followUps" && (
+          {activeView === "followUps" && isFollowUpsTabVisible && (
```
kiln-agent, line 296: `activeView` can be left pointing at a tab that has just disappeared, and the editor body then renders nothing.

`isFollowUpsTabVisible` is derived from the **server** prop `survey.followUps.length`, which changes under the user mid-session:

1. Workflows-enabled org (Cloud or licensed self-host), survey with one follow-up → tab visible, user opens it.
2. User deletes the follow-up (`follow-up-item.tsx:226` marks `deleted: true`) and clicks Save.
3. `updateSurvey` really deletes it (`apps/web/lib/survey/service.ts:582` `deleteMany`), then `survey-menu-bar.tsx:387/460` calls `router.refresh()`.
4. The refreshed server render gives `survey.followUps.length === 0` → `isFollowUpsTabVisible` flips to `false`. `activeView` is client state and survives the refresh, so it is still `"followUps"`.

Result: the Follow-ups tab vanishes from the strip *and* this branch is false, so `<main>` renders an empty pane. The user is stuck looking at a blank editor until they happen to click another tab.

Worth resetting the view when the active tab stops existing, e.g.

```ts
useEffect(() => {
  if (!isFollowUpsTabVisible && activeView === "followUps") setActiveView("elements");
}, [isFollowUpsTabVisible, activeView]);
```

---

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
```diff
@@ -34,75 +37,85 @@ export const FollowUpsView = ({
...
-            </div>
-            <div>
-              <p className="text-lg font-semibold text-slate-800">
-                {t("workspace.surveys.edit.follow_ups_empty_heading")}
-              </p>
-              <p className="text-sm text-slate-500">
-                {t("workspace.surveys.edit.follow_ups_empty_description")}
-              </p>
-            </div>
+      {!surveyFollowUps.length && (
```
kiln-agent: The empty state is now reachable in a configuration where it is a dead end selling a feature the viewer cannot create.

The button inside it is gated on `canCreateFollowUps`, but the card itself is not. Concrete path: workflows-enabled org, survey with one persisted follow-up (so the tab is visible), the user deletes it in the editor — `surveyFollowUps` is now empty while `canCreateFollowUps` is `false`. They get the "Send automatic follow-ups / Send messages to respondents, yourself or team mates." promo card for a deprecated feature, with no action on it, directly under a deprecation banner and a `New workflow` button.

Rendering the card only when `canCreateFollowUps` (and otherwise nothing, or a short "no follow-ups on this survey" line) would avoid advertising the thing the PR is trying to stop people starting.

---

docs/surveys/general-features/email-followups.mdx
```diff
@@ -5,9 +5,13 @@ icon: "envelope"
...
 <Warning>
-  Email Follow-ups are being deprecated in favor of [Workflows](/workflows/overview). Workflows
-  offer the same response- and ending-based email automation plus additional triggers and actions. We recommend
-  building new email automations as Workflows.
+  Email Follow-ups are deprecated in favor of [Workflows](/workflows/overview) and stop working on **1
+  December 2026**. Workflows offer the same response- and ending-based email automation plus additional
+  triggers and actions. Build new email automations as Workflows, and rebuild your existing follow-ups
+  as Workflows before that date.
+
+  The Follow-ups tab is now hidden in the survey editor unless the survey already has follow-ups, or
```
kiln-agent, line 13: This states the visibility rule as an OR over two conditions, but `shouldShowFollowUpsTab` requires the follow-ups entitlement in the second branch:

```ts
if (followUpCount > 0) return true;
return isSurveyFollowUpsAllowed && !isWorkflowsAllowed;
```

A Cloud org on a plan with neither entitlement (`isSurveyFollowUpsAllowed === false`, `isWorkflowsAllowed === false`) matches "Workflows are unavailable on your plan or license" as written here, so the docs promise them the tab — and they get nothing, which is exactly the case the PR intends (retiring the upsell). Suggest: "…unless the survey already has follow-ups, or your plan includes follow-ups but not Workflows."

---

kiln-agent (commented):
Careful pass over the diff. The visibility rule itself is sound and the unit tests cover its four corners; four findings, all in how the rule's consequences show up at runtime — one date-formatting bug, one stuck-tab state, plus a dead-end empty state and a docs mismatch.

---

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
```diff
@@ -34,75 +37,85 @@ export const FollowUpsView = ({
...
-            },
-          ]}
-        />
-      </div>
-    );
-  }
+  // Follow-ups are deprecated, so the only place a new one can still be started is a deployment
+  // Workflows cannot reach yet. Everywhere else the entry point is a Workflow — that is what stops
+  // the migration debt growing while nothing is removed.
+  const canCreateFollowUps = isSurveyFollowUpsAllowed && !isWorkflowsAllowed;
```
kiln-agent, line 58: `canCreateFollowUps` gates the two "new follow-up" buttons (L91, L112) but not the **Duplicate** action on each row. `follow-up-item.tsx:109` `duplicateFollowUp` pushes a fresh `createId()` follow-up into `localSurvey.followUps` with no entitlement or Workflows check, and `FollowUpItem` isn't passed anything to gate it with.

Two consequences:

1. **The stated goal doesn't fully hold.** On a Workflows-enabled deployment, any survey that already has one follow-up — exactly the surveys that keep the tab — can still mint new ones: duplicate, rename, edit. That path keeps the migration debt growing.
2. **A lapsed org's saves start failing.** This PR newly shows the *list* to an org whose entitlement lapsed (it used to get an `UpgradePrompt` instead), so Duplicate is now reachable there. `editor/actions.ts:182-189` throws `OperationNotAllowedError("Survey follow ups are not enabled for this organization")` for any follow-up id not already persisted — so one accidental duplicate makes **every** subsequent save of that survey fail, including edits that have nothing to do with follow-ups, until the row is deleted again. That state was unreachable before this PR.

Passing `canCreateFollowUps` down to `FollowUpItem` and hiding the duplicate button when false covers both.

---

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
kiln-agent: The sunset date renders one day early for every viewer west of UTC.

`SURVEY_FOLLOW_UPS_SUNSET_DATE` is the instant `2026-12-01T00:00:00.000Z`, and `formatDateForDisplay` (`lib/utils/datetime.ts:23`) passes it to `Intl.DateTimeFormat(locale, options)` with **no `timeZone`**, so it formats in the *runtime's* zone. In any negative-offset zone that instant is still 30 November, so a user in New York reads "Nov 30, 2026" while the docs (`email-followups.mdx:8`) and the Linear ticket announce **1 December 2026**. The unit test pins the constant but can't catch this — it asserts the ISO instant, not the rendered string.

This is also what AGENTS.md is getting at under *Date and Time Rendering*: locale controls formatting, time zone controls the represented calendar day, and the two shouldn't be conflated. An announced calendar date is a wall-clock date, not an instant — so either pass `timeZone: "UTC"` explicitly here, or make the constant a plain date-parts value that can't shift.

Cheap to prove either way: the existing `deprecation.test.ts` can assert the *formatted* string under `TZ=America/New_York`.

---

apps/web/modules/survey/editor/components/survey-editor.tsx
kiln-agent, line 296: `activeView` is never reset when `isFollowUpsTabVisible` flips from true to false, so the editor can land on a blank main pane.

`isFollowUpsTabVisible` is derived from the **`survey` prop** (L92-96), which the server refreshes: `survey-menu-bar.tsx:387` calls `router.refresh()` after a successful save. So on a Workflows-enabled deployment, a survey with exactly one follow-up → delete it → Save → the refreshed prop has `followUps.length === 0` → `shouldShowFollowUpsTab` returns false. The tab disappears from the strip, but `activeView` is still `"followUps"` — and with the `&& isFollowUpsTabVisible` guard here, nothing renders. Every `activeView` branch is false, so the user is left staring at an empty pane with no tab selected, and has to click another tab to recover.

Pre-PR this couldn't happen: the tab's visibility didn't depend on the follow-up count.

An effect that resets `activeView` to `"elements"` when `activeView === "followUps" && !isFollowUpsTabVisible` fixes it. Not blocking — it's recoverable and needs a specific sequence — but it's a visible dead end in the deletion flow this PR deliberately keeps open.

---

apps/web/playwright/survey-follow-up.spec.ts
```diff
@@ -1,106 +1,85 @@
...
+          replyTo: [to],
+          subject: "Thanks for responding",
+          body: "<p>Thanks!</p>",
+          attachResponseData: false,
+        },
+      },
+    },
+  });
+
+test.describe("Survey Follow-Up deprecation", async () => {
```
kiln-agent, line 36: Seeding through Prisma instead of clicking is exactly right, and the rationale in the docblock is sound — on CI the empty-state creation path genuinely no longer exists.

The gap is what went with it. The deleted spec wasn't only covering "creation works"; its `attachResponseData` toggle left **off** was a deliberate regression guard for a specific bug ("previously save failed silently"). Creation stays live on every self-hosted community instance, and there is now no test at any level that exercises creating a follow-up with `attachResponseData: false`:

```
$ grep -rn "attachResponseData" apps/web --include=*.test.ts
# only follow-ups.test.ts / workflows runner — all `true`, none on the create path
```

`packages/types/surveys/follow-up.ts` has no schema test either. So the guard is gone with nothing behind it.

Per AGENTS.md this shouldn't come back as an E2E — it's validation logic, so the cheapest level that can fail on it is a unit test on the follow-up schema or the create path. The PR's *Risks* section does disclose the reduced creation coverage, which I'd rather see than not; I'd just like the specific regression it was guarding to be re-pinned somewhere before this lands.

---

docs/surveys/general-features/email-followups.mdx
```diff
@@ -5,9 +5,13 @@ icon: "envelope"
...
+  Workflows are unavailable on your plan or license.
```
kiln-agent: The second branch is missing its other condition. The rule is:

```ts
followUpCount > 0 || (isSurveyFollowUpsAllowed && !isWorkflowsAllowed)
```

The docs say the tab survives if "Workflows are unavailable on your plan or license" — but it also requires the **follow-ups** entitlement. So a Cloud org with neither entitlement is told here it will still see the tab, when in fact it won't (that's the retired upsell, which is the point of the change).

Suggested: "…or Workflows are unavailable on your plan or license *and your plan includes follow-ups*." Self-hosted community, the case the sentence is really about, reads the same either way since `getSurveyFollowUpsPermission` returns `true` there unconditionally.

---

kiln-agent (commented):
## Review — `feat(surveys): stop new follow-ups where Workflows can replace them`

Reviewed the single commit `b327611` at high effort: every hunk plus the surrounding editor, permission, save and alert paths, the i18n plumbing, the Prisma schema behind the new fixture, and the new unit test (6 passed locally).

The shape of this is right, and the reasoning in the PR body holds up against the code. Reading the *persisted* `survey.followUps` rather than the editor's working copy is the correct call and the comment explaining why earns its place. Keeping the tab for self-hosted community — where `getSurveyFollowUpsPermission` returns `true` unconditionally while Workflows need a license — is the one decision that would have been easy to get wrong, and the "Not in this PR" section is an unusually honest account of what's still owed. The failure direction on a license-server blip (tab reappears rather than vanishes) is the safe one.

**What I checked that's fine:** the `Promise.all` destructuring in `page.tsx` stays aligned; `getIsWorkflowsEnabled` exists with the expected signature; `selectSurvey` really does include `followUps`, so the persisted count is trustworthy; ICU single-brace `{date}` is correct for this i18next setup and survived translation in all 15 locales; the removed key has no remaining references; `enterpriseLicenseRequestFormUrl` is still consumed elsewhere in `survey-editor.tsx`; `SurveyEditorTabs` has exactly one caller, so the new optional prop can't silently hide the tab somewhere unnoticed; the seeded follow-up's `to` matches a real `buildEmailSendToOptions` id; and the "not being sent" alert matches the runtime guard at `follow-ups.ts:191`.

### Findings

| # | Where | What |
| --- | --- | --- |
| 1 | `follow-ups-view.tsx:54` | **Duplicate isn't gated** — new follow-ups can still be created on Workflows-enabled deployments, and one duplicate makes every later save fail for a lapsed org |
| 2 | `follow-ups-view.tsx:55` | **Sunset date renders a day early** west of UTC — "Nov 30" vs the announced 1 December |
| 3 | `survey-editor.tsx:290` | `activeView` never reset when the tab disappears → blank main pane after deleting the last follow-up and saving |
| 4 | `survey-follow-up.spec.ts:36` | The `attachResponseData: false` regression guard is gone with no replacement at any level |
| 5 | `email-followups.mdx:14` | Visibility sentence omits that the second branch also needs the follow-ups entitlement |

(1) and (2) are the two I'd want addressed before merge — the first because it leaves the PR's own goal partly unmet *and* introduces a save-blocking state that was previously unreachable, the second because the whole point of pinning the date in a constant was that the copy and the docs can't drift, and right now they do drift by a day for a large share of users. (3) is a recoverable dead end, (4) and (5) are small. Details inline.

### On the open gaps

Both listed gaps read as correctly scoped to me. The one-notice-for-everyone choice is the right default — a self-hosted community deployment seeing a hard date for a feature it can't yet replace *is* the problem, and softening the string would hide it. Worth noting that finding (1) interacts with the packaging question: as long as Duplicate is ungated, "new follow-ups have stopped" isn't quite true yet on Cloud either, which weakens the "buys time" argument slightly.

Not approving — leaving that to @mattinannt as the requested reviewer.

---

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
jobenjada: Fixed in a12d67d. Pinned to \`timeZone: "UTC"\`, but moved the formatting next to the constant as \`formatSurveyFollowUpsSunsetDate\` rather than inlining it in the view, so a second consumer of the date cannot reintroduce the shift by forgetting the zone.

Also took the suggestion to prove it at the right level: \`deprecation.test.ts\` now asserts the rendered string (\`Dec 1, 2026\`) and runs green under \`TZ=America/Los_Angeles\`; dropping the \`timeZone\` option turns three tests red. The e2e now asserts the literal date too instead of \`/Follow-ups stop working on/\`.

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
jobenjada: Fixed in a12d67d. \`canDuplicate\` is threaded to \`FollowUpItem\` and the button is hidden when a new follow-up cannot be created — same gate as the two create buttons.

Both consequences check out as described. The second is the one that decided it: this branch is what newly exposes the list to a lapsed org, so the branch owns the fact that one accidental duplicate would wedge every subsequent save of that survey via \`checkSurveyFollowUpsPermission\`. That state genuinely was unreachable before.

apps/web/modules/survey/editor/components/survey-editor.tsx
jobenjada: Fixed in a12d67d with an effect essentially as written. Confirmed the sequence — `router.refresh()` is called at both save sites (`survey-menu-bar.tsx:387` and `:460`), so the refreshed prop drops `followUps` to zero while `activeView` survives as client state.

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
jobenjada: Fixed in a12d67d — the card is now gated on `canCreateFollowUps` as a whole, not just its button. Kept the "render nothing" option rather than adding a "no follow-ups" line: with the deprecation banner and the `New workflow` button already above it, another line there would say nothing the reader cannot see.

apps/web/playwright/survey-follow-up.spec.ts
jobenjada: Agreed, and thanks for digging out what the step was actually guarding — I had read it as creation coverage only. The bug behind it was #7218: `includeVariables` and `includeHiddenFields` were missing from the form `defaultValues`, so they arrived `undefined` against `z.boolean()` and, with `mode: "onChange"`, Save silently did nothing.

A schema-only test would not have caught that, since the defect was on the modal side. So in a12d67d the defaults move to `lib/form-default-values.ts` and `form-default-values.test.ts` asserts they satisfy `ZCreateSurveyFollowUpFormSchema` — mutation-checked: re-dropping `includeVariables` turns three tests red. One correction to my own test while writing it: a fresh form legitimately fails the schema on `followUpName`, so the guard asserts the name is the *only* thing missing.

docs/surveys/general-features/email-followups.mdx
jobenjada: Fixed in a12d67d, using your wording. The docs now read "…unless the survey already has follow-ups, or your plan includes follow-ups but not Workflows."

apps/web/modules/survey/follow-ups/components/follow-ups-view.tsx
jobenjada: Same issue as the thread above — fixed in a12d67d by pinning `timeZone: "UTC"` inside a `formatSurveyFollowUpsSunsetDate` helper next to the constant. Took your suggestion for the proof: `deprecation.test.ts` asserts the formatted string and passes under `TZ=America/New_York`/`America/Los_Angeles`, and fails without the pin.

apps/web/modules/survey/editor/components/survey-editor.tsx
jobenjada: Duplicate of the thread above — fixed in a12d67d with the effect you describe.

docs/surveys/general-features/email-followups.mdx
jobenjada: Duplicate of the other docs thread — fixed in a12d67d with your suggested phrasing.

apps/web/modules/survey/follow-ups/components/follow-up-modal.tsx
```diff
@@ -111,18 +112,12 @@ export const FollowUpModal = ({
...
-      attachResponseData: defaultValues?.attachResponseData ?? false,
-      includeVariables: defaultValues?.includeVariables ?? false,
-      includeHiddenFields: defaultValues?.includeHiddenFields ?? false,
-    },
+    defaultValues: buildFollowUpFormDefaultValues({
+      defaultValues,
+      firstEmailSendToOptionId: emailSendToOptions[0]?.id,
+      userEmail,
+      t,
+    }),
```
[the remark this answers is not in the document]
jobenjada: Good catch — fixed in 195d7bc.

Verified and it was worse than "inconsistent fallback": the reset copy had already drifted to a hardcoded `"Thanks for your answers!"` instead of `t("workspace.surveys.edit.follow_ups_modal_action_subject")`. That literal happens to be the en-US value, so English saw nothing, but every other locale got untranslated copy when editing a follow-up with no stored subject.

This was also half-done work on my side: extracting one of two duplicate defaults implementations left the unit guard covering only the initialization path. Both go through `buildFollowUpFormDefaultValues` now, so the guard covers the reset too, and I added an assertion that the fallbacks are translated rather than literal.

jobenjada:
I wanna first check it myself

jobenjada:
@itsjavi  can you have a look at the UI / UX of this PR? What I intend to do is written in the PR description. Pls deploy and see for yourself what this feels like

itsjavi:
## Visual evidence: every UI surface and behaviour this PR touches

Captured against `pnpm dev` with Playwright fixture users, `main` (`f44bf98ab`) vs this branch (`74a9cb8aa`), same survey, route, viewport (1280x820), light scheme and seeded follow-up on both sides. Stills are 2x; the clips are trimmed and sped up 2x with the phase burnt into the frame. Both entitlement flags were driven through one local cookie override applied identically to both branches, so all five combinations come from the same dev server.

The tab-strip shots are `nav[aria-label="Tabs"]` itself, so the crop is the element rather than a guessed rectangle.

### 1. The editor tab strip: where the tab goes, and where it stays

Two combinations lose the tab. Three keep it, and keep it pixel for pixel.

**Workflows available, survey has no follow-ups** — the case the deprecation is for:

| Before (main) | After (this PR) |
| --- | --- |
| Editor tab strip: Questions, Styling, Language, Settings, Follow-ups | Same strip ending at Settings, no Follow-ups tab |

**No follow-ups entitlement, no follow-ups** — this is the retired upsell:

| Before (main) | After (this PR) |
| --- | --- |
| Tab strip still ending in Follow-ups for an org that never had the feature | Strip ending at Settings; the tab that led to the upsell is gone |

...and what clicking it used to reach, which now has no entry point at all:

Follow-ups tab on main showing an UpgradePrompt titled "Send automatic follow-ups" with Request trial license and Learn more buttons

### 2. The Follow-ups tab, survey with one follow-up, Workflows available

The soft nudge becomes the deprecation notice with the date, `+ New follow-up` becomes `New workflow`, and the row loses its duplicate icon (only the trash icon remains).

| Before (main) | After (this PR) |
| --- | --- |
| Follow-up list with a + New follow-up button, trash and duplicate icons on the row, and a blue "Need more flexibility?" info alert below | Amber notice "Follow-ups are being replaced by Workflows ... Follow-ups stop working on Dec 1, 2026", a New workflow button, and the row with only a trash icon |

### 3. Lapsed entitlement with a follow-up still on the survey

The sharpest improvement in the diff, and worth a second look: on `main` the early `!isSurveyFollowUpsAllowed` return means the upsell **replaces the follow-up list**, so an org whose plan lapsed cannot see, read or fix the automation its survey is still carrying. This branch shows the follow-up and says plainly that it is not being sent.

| Before (main) | After (this PR) |
| --- | --- |
| Follow-ups tab showing only the upsell - the survey's existing follow-up is nowhere on the page | Deprecation notice, a red "These follow-ups are not being sent" alert, and the Thank-you email row listed below it |

### 4. Self-hosted community keeps its creation path

Follow-ups free, Workflows licensed away: the tab, the promo card and `New follow-up` all survive, and the notice is added on top. This is the one-notice-for-everyone choice from Open gaps, rendered.

| Before (main) | After (this PR) |
| --- | --- |
| Follow-ups tab with the dashed promo card and a New follow-up button, plus the blue Workflows nudge | Same promo card and New follow-up button, with the amber deprecation notice above it and no blue nudge |

### 5. Deleting the last follow-up, then saving

This is the `activeView` reset and empty-state gating that Open gaps lists as verified by reading the render path. Here it is running: the working copy soft-deletes so the tab holds through the delete, then the save refreshes the server prop and the tab goes with it, dropping the view back to Questions rather than leaving an empty pane. On `main` the tab stays and offers to create another follow-up.

| Before (main) | After (this PR) |
| --- | --- |
| Clip: delete the only follow-up, save, and the Follow-ups tab stays with a New follow-up card | Clip: delete the only follow-up, save, and the Follow-ups tab disappears while the editor falls back to Questions |

End state, side by side:

| Before (main) | After (this PR) |
| --- | --- |
| Follow-ups tab still selected, showing the create-a-follow-up promo card | Questions view with "Progress saved" and a tab strip that ends at Settings |

Full-rate MP4s: [before](https://github.com/formbricks/formbricks/blob/assets-pr-screenshots/PRO-98/delete-last-followup-before.mp4) - [after](https://github.com/formbricks/formbricks/blob/assets-pr-screenshots/PRO-98/delete-last-followup-after.mp4)

### 6. The subject fallback on the reset path

The drift the shared builder fixes, in the UI: a German-locale author editing a follow-up with no stored subject. The unit test pins it; this is what the author sees.

| Before (main) | After (this PR) |
| --- | --- |
| German modal, Betreff field pre-filled with the English "Thanks for your answers!" | Same field pre-filled with "Danke für deine Antworten!" |

<details>
<summary>The three cases where the tab survives - pixel-identical strips, so no regression</summary>

Each pair below is byte-for-byte identical (ffmpeg difference blend: `YMAX=0`, `YAVG=0`), which is the point: `shouldShowFollowUpsTab` changes who gets the tab without touching anyone who should keep it.

| Case | Before (main) | After (this PR) |
| --- | --- | --- |
| Survey already has a follow-up, Workflows available | Tab strip ending in Follow-ups | Identical strip |
| Entitlement lapsed, follow-up still on the survey | Tab strip ending in Follow-ups | Identical strip |
| Self-hosted community, no follow-ups yet | Tab strip ending in Follow-ups | Identical strip |

</details>

**Not covered here:** the read-only member's view of the same tab, and the v1 management API paths that can write an over-long subject or a non-member `to` (both listed as converter work, not this PR). The `New workflow` CTA was verified to render and point at `/workspaces/<id>/workflows`, which is the fix in `74a9cb8`; I did not follow it through to a created workflow, since it deliberately only lands on the list.

---

itsjavi (changes requested):
@jobenjada  

Shouldn't we add a button to upgrade the plan in the second alert? or since they will be deprecated, that will be misleading?

I would also change the listed follow-ups with a red/orange border or something that indicates that they are not active.

Image

jobenjada:
@itsjavi I wouldn't upsell them on a feature we are sunsetting and red / orange always looks like an error, while it is `disabled`, right? Maybe dark gray bg?

jobenjada:
@itsjavi pls remove the change request and add to merge queue after your review :)

itsjavi (approved):
lgtm

PR #8985, opened 2026-08-26
feat(dashboards): merge the Line chart type into Area as a display style

Dhruwang:
Fixes ENG-2612

> Stacked on #8973 (ENG-2541) — base is that branch, not `main`, because the chart-type selector and `chart-display.ts` changes there are the ground this builds on. Retarget to `main` once #8973 merges.

## What & why

Line and Area were two `ChartType` enum values for one chart. Both rendered the same Recharts `<AreaChart>` + `<Area>` series over the same axes, with the same `connectNulls` and tick handling — the only difference was how the band under the stroke got painted: `line` used a gradient fading to nothing, `area` a flat 0.4 tint. Neither stacks, so the usual semantic split (independent series vs. part-of-whole composition) did not exist here either. Two types, one chart.

They now collapse into a single **Area / Line** type carrying a `config.areaDisplay: "filled" | "line"` style, the same shape as the existing `barOrientation` and `pieDisplay` settings — and the same reasoning already applied to pie/breakdown, which the renderer calls "two renderings of one chart type rather than two chart types".

- `line` leaves the `ChartType` enum; `area` survives, so no existing `area` row is touched.
- The migration backfills `line` charts to `area` + `areaDisplay: "line"` **before** dropping the value, using `config || '{...}'::jsonb` so their other config keys survive.
- `area` charts keep an absent `areaDisplay` and resolve to the `"filled"` default, matching the "empty config falls back to the previous behavior" contract the other display settings already follow.
- Both styles are pixel-identical to the types they replace. `fillOpacity: 0.6` is spelled out for the line style because that was Recharts' own default it had been relying on; `dot: false` / `activeDot` likewise resolve to what each branch rendered before.
- The selector shows one entry labelled "Area / Line Chart" so the type is still findable by "line", with the style exposed through `ChartDisplaySettings`.
- The AI system prompt reads its chart-type list off `CHART_TYPE_IDS` instead of spelling it out, so it cannot offer the model a type `ZChartType` would reject, plus a line telling the model `area` covers a line rendering.
- `resolveChartType` maps a stray `"line"` string forward to `area` instead of letting it fall through to the `"bar"` default — reachable by anything holding a pre-migration value, such as a cached payload. The alias table is a `Map` rather than an object literal so an inherited key like `"constructor"` cannot resolve to a bogus type.

https://github.com/user-attachments/assets/fd56dd48-9f8c-49f6-b007-5964b76289c3

## Breaking changes

None — the `ChartType` enum is internal. No API v1/v2/v3 route, SDK surface, OpenAPI schema or docs page references chart types (`grep` over `apps/web/modules/api`, `apps/web/app/api` and `apps/docs` returns nothing), charts are reached only through server actions and module code, and the enum change needs no action from a self-hoster beyond running migrations as usual.

## Migrations & env

- `packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area` — backfills `Chart` rows with `type = 'line'` to `type = 'area'` + `config.areaDisplay = "line"`, then recreates the `ChartType` enum without `line`. Backfill and swap share one transaction, so a failure in either leaves the column and the enum consistent. Runs automatically; no manual step. The enum recreate takes a brief `ACCESS EXCLUSIVE` lock on `Chart` (a small table).
- No env changes.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Existing `line` charts survive the enum drop with their style and other config intact | manual | Seeded two `line` rows on a local DB at the pre-migration schema — one with `config = {}`, one with `{"xAxisLabel","showLegend","barOrientation"}` — then ran `pnpm db:migrate:dev`. Both came out `type = area` with `areaDisplay: "line"`; the second kept all three original keys. The pre-existing `area` row kept `config = {}`. `pg_enum` ended at `area, bar, pie, big_number`, and `select type, count(*) from "Chart" group by type` showed no `line` rows left. Test rows deleted afterwards. |
| A migrated chart reads back through the app as the merged type with Line selected | manual | Opened the migrated `line` chart in the builder: **Area / Line Chart** selected, and Chart display settings showed `Display as → Filled area \| Line` with **Line** active. Its now-inapplicable `barOrientation` correctly rendered no Bar direction control. |
| Both display styles render as they did as separate types | manual | Switched a 6-category sentiment chart (165/157/89/71/18/14) to the merged type and toggled live, no reload. `filled` = solid flat tint to the baseline; `line` = 2px stroke over a gradient fading to nothing. |
| The style is offered only on the merged type | unit (mutation) | New `describe("area display")` suite in `apps/web/modules/ee/analysis/charts/lib/chart-display.test.ts` covers `supportsAreaDisplay` across all four types plus `undefined`, the `"filled"` default, and `sanitizeChartDisplay` keeping the key on `area` while dropping it elsewhere. Flip `chartType === "area"` to `!== "area"` at `chart-display.ts:18` and the suite goes red. Confirmed in the UI too: Bar shows only Bar direction, Big Number drops the whole settings panel. `pnpm --filter @formbricks/web test -- chart-display` |
| The selector no longer offers Line as its own type | unit (mutation) | `chart-types.test.ts` asserts 4 entries in order `area, bar, pie, big_number` and that `CHART_TYPE_ICONS` has exactly those keys. Re-adding a `line` entry to `getChartTypes` turns it red. `pnpm --filter @formbricks/web test -- chart-types` |
| A stray `"line"` chart type resolves to the merged type, not the `bar` fallback | unit (mutation) | New case in `chart-utils.test.ts`; deleting the `LEGACY_CHART_TYPE_ALIASES` lookup at `chart-utils.ts:85` makes it return `"bar"` and the test fails. `pnpm --filter @formbricks/web test -- chart-utils` |
| An inherited `Object` key does not resolve through the alias lookup | unit (mutation) | `resolveChartType("constructor")` must return `"bar"`. The alias table is a `Map`, so it only sees its own entries; swapping it back to an object literal makes the test fail, because `LEGACY_CHART_TYPE_ALIASES["constructor"]` would find `Object.prototype.constructor`, treat it as a hit and return it as a `TChartType`. Caught in self-review after the first push — fixed in c2a3888. |
| The AI prompt offers only chart types the output schema accepts | unit (red on main) | The prompt listed `bar, line, area, pie, or big_number` while `ZChartType` had already dropped `line`, so a "line chart" request steered the model at a value that fails validation inside `generateObject` — the user got the generic prompt error, and `resolveChartType`'s alias never sees the value. New test in `ai-schema-context.test.ts` pins the guideline to `CHART_TYPE_IDS`; it fails against the pre-fix prompt. `cd apps/web && npx vitest run modules/ee/analysis/lib/ai-schema-context.test.ts` |
| Backfill and enum swap are atomic | manual | Applied `migration.sql` verbatim to a throwaway database seeded with 2 `line` rows, 1 `area` and 1 `bar`: one `BEGIN … COMMIT`, `UPDATE 2`, all four rows correct, enum down to 4. Then re-ran it against a database with a second table on the enum so `DROP TYPE` fails mid-swap — `ROLLBACK`, and the `line` row was still `line` with `config = {}` rather than left rewritten against an unchanged enum. Both test databases dropped afterwards. |
| Stored chart types reach the renderer through `resolveChartType` on the dashboard path too | none | `dashboard-detail-client.tsx` passed `widget.chart.type` raw — the only read that skipped the resolver — so on a database where this migration has not run yet a `line` widget rendered "not yet supported". Now wrapped. No test: it is a `.tsx` component, which this repo covers by e2e rather than unit tests, and the window it guards (app serving traffic ahead of migrations) is not reachable from the e2e stack. Reviewed by inspection against the other call sites in `use-chart-dialog.ts`. |
| No regressions in the analysis module | unit (guard) | 470/470 tests pass across all 28 suites in `apps/web/modules/ee/analysis` (156 of them in `charts/lib`). Zero browser console errors and no new dev-server errors while exercising all four chart types. |

**Open gaps**

- No screenshots attached yet for the two display styles — the outcomes above are described from the manual pass; images to follow on this description.
- No e2e spec. There is no existing Playwright coverage of the chart builder to extend, and a new spec for one display toggle did not seem worth a standalone flow — the display-settings path is covered by unit tests plus the manual pass above.
- The migrated chart was verified in the chart list and builder preview, not inside a dashboard widget. Same `ChartRenderer` and same `config`, threaded through `dashboard-widget-data.tsx` unchanged by this diff, so the risk is low but it is untested.
- AI chart generation was not run against a live model. `ZChartType` no longer admits `"line"`, so the model can only return the merged type; only the mocked `ai-chart-query.test.ts` fixture was updated.

**Risks**

- Anything outside `modules/ee/analysis` holding a hardcoded `"line"` chart type would now hit the legacy alias rather than a real enum value. A repo-wide grep found no such call site, and both read paths (builder and dashboard widget) now go through the alias, but that is the place to look if a chart renders as the wrong type.
- Migrated charts can carry a display key their new type ignores (the seeded row kept `barOrientation`). `sanitizeChartDisplay` drops it on the next save; until then it is inert.

---

> [!NOTE]
> **AI model used** — `claude-opus-5 (Claude Code)`, reasoning effort `high`.


apps/web/modules/ee/analysis/charts/lib/chart-utils.ts
```diff
@@ -74,10 +74,19 @@ export const getSentimentMeasureColor = (measureId: string): string | undefined
...
+ * display style is not recovered here; the migration is what carries it for stored charts.
+ */
+const LEGACY_CHART_TYPE_ALIASES = new Map<string, TChartType>([["line", "area"]]);
+
+/** Validate a chart type string, mapping retired types forward and defaulting to "bar". */
 export const resolveChartType = (raw: string): TChartType => {
   const parsed = ZChartType.safeParse(raw);
-  return parsed.success ? parsed.data : "bar";
+  if (parsed.success) return parsed.data;
+  return LEGACY_CHART_TYPE_ALIASES.get(raw) ?? "bar";
```
kiln-agent, line 89: `LEGACY_CHART_TYPE_ALIASES` is an object literal, so the index lookup also resolves `Object.prototype` members: `resolveChartType("constructor")` returns the `Object` constructor and `resolveChartType("toString")` returns a function — both truthy, so `?? "bar"` never fires and the function returns a non-`TChartType` value despite its signature. Today `raw` only ever comes from the DB enum, so it is not reachable, but the helper is exported and typed `(raw: string)`, and it is the documented landing place for "anything still holding the old value".

```suggestion
  return Object.hasOwn(LEGACY_CHART_TYPE_ALIASES, raw) ? LEGACY_CHART_TYPE_ALIASES[raw] : "bar";
```

---

kiln-agent (commented):
Code review — correctness pass over the diff. Three findings; two sit in files the diff does not touch, so they could not be attached inline.

**1. `apps/web/modules/ee/analysis/lib/ai-schema-context.ts:67` — the AI prompt still offers the retired `line` type (highest impact)**

```
- Choose the most appropriate chart type: bar, line, area, pie, or big_number (for single-number queries).
```

`ZChartType` is the structured-output schema passed to `generateOrganizationAIObject` in `ai-chart-query.server.ts:127`, and it no longer admits `"line"`. A prompt like "line chart of responses per day" now steers the model to `chartType: "line"`; with any provider that does not hard-constrain the enum, the object fails zod validation, `generateAIChartQuery` throws, and AI chart generation surfaces an error instead of the merged Area/Line chart. `resolveChartType`'s legacy alias cannot rescue this path — the value never gets past the schema. `ai-chart-query.test.ts` only swapped the mocked fixture from `line` to `area`, so nothing catches the stale instruction. Suggested fix: drop `line` from that guideline (ideally render the list from `CHART_TYPE_IDS` so prompt and enum cannot drift again).

**2. `apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx:144` — the dashboard widget path bypasses `resolveChartType`**

```tsx
chartType={widget.chart.type}
```

Every other read of a stored chart type goes through `resolveChartType` (`use-chart-dialog.ts:125,158`), which is what the new `LEGACY_CHART_TYPE_ALIASES` entry exists for. This one passes the raw column value straight to `ChartRenderer`, whose `switch` no longer has a `line` arm — so an un-migrated `line` row renders the `default` branch, i.e. the `Chart type "line" not yet supported` message, where it previously rendered fine. Reachable whenever this code serves traffic against a database where `20260826120000_...` has not been applied yet (rolling deploy, deferred migration, a restored snapshot). The PR's risk note says the alias "makes the failure benign"; that holds for the builder but not for dashboards. Wrapping in `resolveChartType(widget.chart.type)` closes it.

**3. `apps/web/modules/ee/analysis/charts/lib/chart-utils.ts:89`** — see the inline comment: prototype-chain lookup in `LEGACY_CHART_TYPE_ALIASES[raw]`.

Everything else in the diff checks out: the two `Area` branches are behaviourally identical to the types they replace (`fillOpacity` 0.6 was indeed Recharts' `Area` default for the old `line` case, and `activeDot={undefined}` still resolves to the default active dot for `filled`), the `<defs>`-or-`null` child is fine for Recharts, `sanitizeChartDisplay`/`resolveChartDisplay` follow the existing per-type pattern, `ChartType` has exactly one column so the enum recreate is complete, `Chart.config` is `JSONB NOT NULL DEFAULT '{}'` and always written through `ZChartConfig`, so the `config || '{"areaDisplay": "line"}'::jsonb` backfill cannot hit a NULL or non-object left operand, the migration timestamp is the newest in `packages/database/migration`, and no reference to the retired type survives outside `modules/ee/analysis` (no API, SDK, or docs surface). 120 tests in `modules/ee/analysis/charts/lib` pass locally (the `charts.test.ts` failure is the unrelated stale `@formbricks/database` `dist/` trap from AGENTS.md).

---

kiln-agent (commented):
## Review — merge Line into Area as a display style

Reviewed the full diff and verified the key claims locally against the head commit (`4a24ac8`). This is a clean, well-scoped refactor. No blocking issues found.

### What I verified

- **Renderer merge is behavior-preserving.** The merged `case "area"` reproduces both old cases exactly:
  - `line` style → gradient fill + `fillOpacity={0.6}` (0.6 was Recharts' own default the old code relied on) + `activeDot={{ r: 5, … }}`.
  - `filled` style → flat `color` fill + `fillOpacity={0.4}` + `activeDot={undefined}`, which resolves to Recharts' default `activeDot` — the same default the old standalone `area` case relied on (it never set the prop). `dot={false}` matches the Area default the old area case used implicitly. So both styles are pixel-identical to the types they replace.
- **Migration is correct and safely ordered.** `Chart.config` is `Json @default("{}")` (non-nullable), so `config || '{"areaDisplay": "line"}'::jsonb` never hits the `NULL || jsonb → NULL` pitfall — the `areaDisplay` key is reliably applied and other config keys survive. The backfill runs *before* the enum recreate, so the `USING ("type"::text::ChartType_new)` cast has no `line` rows left to fail on. This is the standard Prisma enum-value-removal shape.
- **No stray `"line"` chart-type references remain.** Repo-wide grep confirms every remaining `"line"` is either the new `areaDisplay` style value, the explanatory comment, or the alias test. `chart_type_line` is gone from every locale and `i18n.lock`.
- **Legacy alias + AI path.** `resolveChartType` maps a stray `"line"` forward to `"area"` instead of the `"bar"` fallback (covered by a new mutation-style test). AI output is schema-constrained by `ZChartType`, which no longer admits `"line"`, so the model can only return the merged type.
- **Type safety.** `CHART_TYPE_ICONS: Record<TChartType, …>` is exhaustive with the narrowed 4-key `TChartType`; dropping `line` keeps it consistent.
- **i18n handled per repo rules** — only `en-US.json` edited by hand; other locales + lock are the machine-generated output.

### Tests

Ran the affected suites on the head commit after rebuilding the dependency graph (`pnpm build --filter=@formbricks/web^...`):

- `chart-display`, `chart-types`, `chart-utils`, `ai-chart-query` → **89 passed**
- Full `charts/lib` directory → **156/156 passed** across 9 suites (matches the PR's claim)

### Minor observations (non-blocking)

- The backfill `UPDATE` sits outside the explicit `BEGIN; … COMMIT;` block that wraps the enum swap (Prisma's generated shape). It's harmless here because the backfill is idempotent, but worth knowing: if the enum swap ever failed, the backfill would already be committed and a re-run would simply no-op.
- `breaking-change` checkbox correctly left unticked — the enum is internal with no API/SDK/OpenAPI surface, confirmed by the diff and the grep in the description.

### On the open gaps you listed
The remaining gaps (no e2e for the display toggle, migrated chart not exercised inside a live dashboard widget, AI not run against a live model) are reasonable to defer given the unchanged `ChartRenderer`/`config` threading and the schema-constrained AI output. No objection.

LGTM pending the retarget to `main` once #8973 lands.

---

packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+  what lets the enum value be dropped below without data loss.
+
+  `area` charts keep an absent `areaDisplay`: the app defaults it to "filled", so their rendering
+  is unchanged and no row without a `line` type needs touching.
+*/
+-- Backfill: every line chart becomes an area chart displayed as a line.
+UPDATE "public"."Chart"
+SET "type"   = 'area',
+    "config" = "config" || '{"areaDisplay": "line"}'::jsonb
+WHERE "type" = 'line';
```
kiln-agent, line 21: The backfill is not in the same transaction as the enum swap. Prisma Migrate does not wrap a migration script in a transaction — that is why the generated `AlterEnum` carries its own explicit `BEGIN; … COMMIT;` — so this `UPDATE` commits on its own and there is a window between it and the enum recreate.

Concrete failure: during a rolling deploy old instances are still serving and can still write `type = 'line'`. The `UPDATE` commits (no line rows left), an old instance inserts one, then

```
ALTER TABLE "public"."Chart" ALTER COLUMN "type" TYPE "public"."ChartType_new" USING ("type"::text::"public"."ChartType_new");
```

fails with `invalid input value for enum ChartType_new: "line"`. The migration is then recorded as failed with the backfill already committed, and the deploy is stuck until someone runs `prisma migrate resolve` by hand.

Moving the `UPDATE` inside the `BEGIN … COMMIT` makes backfill and enum swap atomic, and the `ACCESS EXCLUSIVE` lock the `ALTER TABLE` takes then also covers the backfill:

```suggestion
-- AlterEnum + backfill, atomically: the ACCESS EXCLUSIVE lock the ALTER TABLE below takes must
-- also cover the backfill, or a 'line' row written in between fails the enum cast.
BEGIN;

-- Backfill: every line chart becomes an area chart displayed as a line.
UPDATE "public"."Chart"
SET "type"   = 'area',
    "config" = "config" || '{"areaDisplay": "line"}'::jsonb
WHERE "type" = 'line';
```


---

kiln-agent (commented):
Code review — 3 findings. Two of them are in files this PR does not touch, so they could not be posted inline; they are below.

### 1. The AI system prompt still offers `line` — `apps/web/modules/ee/analysis/lib/ai-schema-context.ts:67`

```
- Choose the most appropriate chart type: bar, line, area, pie, or big_number (for single-number queries).
```

`ZChartType` no longer admits `"line"`, and it is the structured-output schema for `chartType` (`ai-chart-query.server.ts:91`), so the prompt now steers the model at a value the schema rejects.

Ask for "responses per day over the last 30 days as a line chart":

- Provider without strict enum enforcement → the model returns `chartType: "line"`, `Output.object` validation fails, `NoObjectGeneratedError` → `getAIChartPromptError` → `InvalidInputError(AI_CHART_PROMPT_ERROR_CODE)`, i.e. the user gets the generic "couldn't understand this prompt" error instead of a chart.
- Provider with strict enforcement → the enum is honoured, but the prompt gives the model no hint that Line is now the `area` type, so a "line chart" request lands on some other type (bar) rather than `area`.

The new `LEGACY_CHART_TYPE_ALIASES` does not cover this: validation happens inside `generateObject`, before `normalizeChartQuery` sees the value. Nothing asserts on this string either (`ai-schema-context.test.ts` says nothing about chart types), which is why it rotted silently. Suggest replacing `bar, line, area, pie, or big_number` with the four real ids and, if a line rendering is still wanted from AI, a note that `area` covers it.

### 2. Dashboard widgets bypass the legacy alias — `apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx:144`

`chartType={widget.chart.type}` goes straight into `ChartRenderer`; only `use-chart-dialog.ts` calls `resolveChartType`. So on a database where this migration has not run yet (self-hoster or dev starting the app before `prisma migrate deploy` / `pnpm db:migrate:dev`, or an instance booting ahead of migrations), a dashboard widget for a `line` chart hits the renderer's `default:` branch and shows `Chart type "line" not yet supported` — the case the description says the alias covers.

Second-order effect in that same window: opening such a chart in the builder *does* map it to `area`, but with no `areaDisplay`, so it renders filled. If the user saves, the row is persisted as `type = area` with no `areaDisplay`, and the migration — which only touches `type = 'line'` — can no longer recover the line style. The style choice is then lost for good.

### 3. Backfill is not atomic with the enum swap

See the inline comment on the migration.

---

Everything else in the diff checked out. In particular the claim that both styles are pixel-identical to the types they replace holds against recharts 2.15.3: `Area.defaultProps` is `fillOpacity: 0.6, dot: false, activeDot: true`, so the explicit `0.6` for the line style and `activeDot={undefined}` for the filled style both resolve to what each branch rendered before. `config` is `JSONB NOT NULL DEFAULT '{}'` and every write path goes through `ZChartConfig`, so the `config || '…'::jsonb` concat cannot hit a NULL or a non-object. `sanitizeChartDisplay` is the only allowlist of display keys (nothing else in the repo reads `barOrientation`/`pieDisplay`), `areaDisplay` was added to `ZChartConfig` so chart duplication carries it, `chart-row.tsx` already falls back to a bar icon for an unknown type, and the 120 tests in `modules/ee/analysis/charts/lib` pass (the `charts.test.ts` failure is the known stale `@formbricks/database` `dist/`, unrelated).

<sub>Reviewed with Claude Code (claude-opus-5), effort `high`.</sub>

---

kiln-agent:
Housekeeping: my review tooling posted three reviews above instead of one, and one of them ("LGTM, no blocking issues") contradicts the other two. Sorry for the noise — GitHub won't let me delete submitted reviews, so treat this comment as the single authoritative summary and disregard the LGTM.

Two findings are real; I re-verified both against `4a24ac8`:

**1. The AI prompt still offers the retired `line` type** — `apps/web/modules/ee/analysis/lib/ai-schema-context.ts:67`

```
- Choose the most appropriate chart type: bar, line, area, pie, or big_number (for single-number queries).
```

`ZChartType` is the structured-output schema for `chartType` and no longer admits `"line"`, so the prompt now steers the model at a value its own schema rejects. Ask for "responses per day as a line chart" and either validation fails (`NoObjectGeneratedError` → the generic "couldn't understand this prompt" error) or, with a strict-enum provider, the model lands on some other type because nothing tells it Line is now `area`. `LEGACY_CHART_TYPE_ALIASES` can't help — validation happens inside `generateObject`, before `normalizeChartQuery` sees the value. Rendering that list from `CHART_TYPE_IDS` would stop prompt and enum drifting apart again.

**2. Dashboard widgets bypass the legacy alias** — `apps/web/modules/ee/analysis/dashboards/components/dashboard-detail-client.tsx:144`

`chartType={widget.chart.type}` goes straight into `ChartRenderer`. `resolveChartType` is called only from `use-chart-dialog.ts:125,158`, so on a database where this migration hasn't run yet a `line` widget hits the renderer's `default:` arm and shows `Chart type "line" not yet supported`. The description's risk note says the alias makes that failure benign — true for the builder, not for dashboards. Second-order effect in the same window: the builder maps such a chart to `area` with no `areaDisplay`, and saving persists `type = area` with no style, after which the migration (`WHERE type = 'line'`) can no longer recover the line rendering.

Two minor notes, non-blocking: `LEGACY_CHART_TYPE_ALIASES[raw] ?? "bar"` is a prototype-chain lookup, so `raw = "constructor"` returns a function rather than falling through to `"bar"` (unreachable in practice — `raw` is an enum column or schema-constrained AI output — but `Object.hasOwn` or a `Map` would close it); and the backfill `UPDATE` sits outside the `BEGIN … COMMIT` enum swap, which is Prisma's generated shape and idempotent, so it only matters if a `line` row were written between the two statements during a rolling deploy.

The rest of the diff checks out, and I verified the load-bearing claims rather than taking them from the description: both `Area` branches are behaviourally identical to the types they replace against recharts 2.15.3 (`Area.defaultProps` is `fillOpacity: 0.6, dot: false, activeDot: true`, so the explicit `0.6` and `activeDot={undefined}` each reproduce what the old branch rendered); `Chart.config` is `JSONB NOT NULL DEFAULT '{}'` with every write going through `ZChartConfig`, so the `config || '…'::jsonb` backfill cannot hit a NULL or non-object left operand; `ChartType` has exactly one column, so the enum recreate is complete; `sanitizeChartDisplay` is the only allowlist of display keys, and `areaDisplay` reached `ZChartConfig`, so duplication carries it; no reference to the retired type survives outside `modules/ee/analysis`, and only `en-US.json` was hand-authored.

---

Dhruwang:
Thanks — all three findings were real. Fixed in 5150fc8, plus the prototype-chain one in c2a3888. Taking them in impact order:

**1. AI prompt still offered `line` — confirmed, and the worst of the three.** `ai-schema-context.ts:67` listed `bar, line, area, pie, or big_number` while `ZChartType` had already dropped `line`, and since that schema is what constrains `generateObject`, a "line chart" request steered the model straight at a value validation rejects. `resolveChartType`'s alias is downstream of that, so it could never help. I had grepped `charts/lib` for prompt text and missed that the prompt lives in `analysis/lib`.

Fixed by reading the list off `CHART_TYPE_IDS` rather than spelling it out, so prompt and enum cannot drift again, plus a line telling the model `area` covers a line rendering (your second-order concern — strict-enum providers would otherwise land "line chart" on bar). New test pins the guideline to those ids and fails against the old string.

**2. Dashboard widget bypassed the resolver — confirmed.** `chartType={widget.chart.type}` was the one read of a stored type that skipped `resolveChartType`, so a `line` row on a database ahead of its migrations hit the renderer's `default:` branch. Now wrapped. Your second-order point is the sharper one and I'd missed it: in that same window, opening such a chart in the builder resolves it to `area` with no `areaDisplay`, and saving persists `type = area` — after which the migration's `WHERE type = 'line'` can never recover the line style. Closing the read path closes that too. No unit test, since it is a `.tsx` component and this repo covers those by e2e; noted as such in the description.

**3. Backfill atomicity — taken, and it turned out worth more than a nit.** Moved the `UPDATE` inside the `BEGIN … COMMIT`. I verified both directions on throwaway databases rather than reasoning about it:

- Happy path: seeded 2 `line` rows, 1 `area`, 1 `bar`; one transaction, `UPDATE 2`, all four rows correct, enum down to 4.
- Failure path: added a second table on the enum so `DROP TYPE` fails mid-swap. `ROLLBACK`, and the `line` row was still `line` with `config = {}` — where before the reorder it would have been left rewritten as `area` + `areaDisplay` against an enum that still had `line`. So "idempotent, harmless" was true of a re-run, but the intermediate state was inconsistent, and a re-run is not guaranteed to happen.

That second test also incidentally confirms your note that `ChartType` having exactly one column is what makes the `DROP TYPE` safe — it fails immediately if anything else depends on the type.

On the housekeeping: no problem, and thanks for flagging which review to disregard. Worth noting the LGTM was not merely redundant — it asserted "AI output is schema-constrained by `ZChartType`, so the model can only return the merged type", which is exactly the reasoning that misses finding 1. Right about the constraint, wrong about the consequence: the schema does reject `line`, and that rejection is the bug.

Everything else you verified matches what I had: `fillOpacity` 0.6 as Recharts' `Area` default, `activeDot={undefined}` resolving to the default, `config` being `JSONB NOT NULL` so the `||` concat is safe, and no API/SDK/docs surface on the enum. 470/470 tests in `modules/ee/analysis` pass on 5150fc8 after `pnpm build --filter=@formbricks/web^...` — and yes, the `charts.test.ts` failure you hit was the stale `dist/` trap; I hit it too in a fresh worktree.

apps/web/modules/ee/analysis/charts/lib/chart-utils.ts
Dhruwang: Fixed in c2a3888 — took the `Map` route rather than the `Object.hasOwn` guard, since a `Map` cannot see prototype members at all instead of relying on every future reader remembering the guard. Same effect for `"constructor"` / `"toString"`.

You are right that it is unreachable today via the DB enum, but the signature is `(raw: string)` and the doc comment invites exactly that kind of caller, so the type lie was worth closing. `resolveChartType("constructor")` returning `"bar"` is now pinned by a test.

packages/database/migration/20260826120000_eng_2612_merge_line_chart_type_into_area/migration.sql
Dhruwang: Fixed in 5150fc8 — the `UPDATE` now sits inside the `BEGIN … COMMIT` alongside the swap.

Your premise checks out for this repo specifically: `packages/database/src/scripts/migration-runner.ts` applies schema migrations by spawning `prisma migrate deploy` (batched, but still `migrate deploy`), so there is no outer transaction and the `UPDATE` really was committing on its own.

The rolling-deploy scenario is the one that convinced me — an old instance inserting `type = 'line'` between the committed backfill and the swap makes the `USING ("type"::text::"ChartType_new")` cast fail with `invalid input value for enum ChartType_new: "line"`, leaving the migration recorded failed with the backfill already applied and needing a manual `prisma migrate resolve`. That is materially worse than the "idempotent, harmless" read I had.

Verified both directions on throwaway databases rather than reasoning about it:

- Happy path: 2 `line` rows + 1 `area` + 1 `bar` → single transaction, `UPDATE 2`, all rows correct, enum down to 4.
- Failure path: second table on the enum so `DROP TYPE` fails mid-swap → `ROLLBACK`, and the `line` row was still `line` with `config = {}`. Pre-fix it would have been left rewritten as `area` + `areaDisplay` against an enum that still had `line`.

That second test also confirms your other note: the `DROP TYPE` is only safe because `ChartType` has exactly one column — it fails immediately if anything else depends on the type.

xernobyl (approved):
Approving. I checked the claims rather than re-deriving them, and the ones that mattered held up.

**What I verified against `bfc2cef4c`**

- **Migration.** Applied all 165 base SQL migrations to a throwaway PG 18, seeded `line` rows at the pre-migration schema, then ran `migration.sql` verbatim. The backfill preserved the other config keys, the pre-existing `area` row was untouched, the enum ended at `area, bar, pie, big_number`, and there was no orphan `ChartType_old`. The `BEGIN`/`COMMIT` wraps the backfill now, so that earlier atomicity note is closed.
- **The "pixel-identical" claim.** Rendered `ChartRenderer` from this head and from the base commit and diffed the emitted SVG. Once attribute order is normalised (prop order changed, values didn't), head's `area` + `filled` is identical to the old standalone `area`, and head's `area` + `line` is identical to the old `line`. The explicit `fillOpacity: 0.6` does reproduce what `Area.defaultProps` had been supplying.
- **The inert `barOrientation`.** A migrated row that kept `barOrientation` renders byte-identical SVG to one that didn't, so that second risk note is a non-issue rather than a "probably fine".
- **Tests can actually fail.** Flipped `supportsAreaDisplay` to `!== "area"` (4 red), swapped the alias `Map` back to an object literal (1 red), reverted the prompt guideline (1 red).

**Three non-blocking things** — none worth holding this up for:

1. `use-chart-dialog.ts:125,158` — on a database where the migration hasn't run yet, opening a `line` chart maps the *type* forward but not the *style*, so it comes back Filled. If the user then saves, the row is persisted as `type = area` with no `areaDisplay`, and the migration can no longer recover it. It's the same window you deliberately hardened `dashboard-detail-client.tsx` for, so the defence is a bit lopsided — though I can see an argument that the window is too narrow to be worth more code.
2. `chart-row.tsx:22` — a third raw read of the stored type; an un-migrated `line` row falls back to the bar icon. Purely cosmetic, but the description says "both read paths (builder and dashboard widget) now go through the alias" and there are three.
3. The backfill assumes `config` is a JSON object — a `'null'::jsonb` config would concatenate into an array. Not reachable through the app, since every write goes through `ZChartConfig`, so this one really is just a nit.

Nice call pulling the chart-type list off `CHART_TYPE_IDS` rather than patching the string — that's the drift that caused this in the first place.

Retarget to `main` once #8973 lands.

PR #8986, opened 2026-08-26
fix: use multi-arch Valkey image

BhagyaAmarasinghe:
Ref ENG-2575

Fixes: ENG-2629
## What & why

- The bundled Valkey digest selected only `linux/amd64`, so native ARM64 self-hosted installations failed
  without emulation even though upstream publishes multi-architecture images.
- Pin Valkey 8.1.9 Alpine's immutable OCI index across Docker, Helm and CI so Docker selects the native amd64
  or arm64 child.
- Migrate only the known old one-click pin, including valid non-default Compose indentation, retaining a
  backup and restoring it if Compose validation fails; custom image references and the named AOF volume
  remain untouched.
- Add pin-synchronization tests, a live manifest-platform guard, and self-hoster upgrade guidance.

## Breaking changes

| Bundled Valkey image | Debian Valkey 8.1.1 amd64 child digest | Alpine Valkey 8.1.9 multi-architecture index | Existing self-hosters whose saved Compose file or Helm override retains the old digest | Refresh the one-click updater or replace the documented digest; keep the existing Redis volume/PVC |

## Migrations & env

- No database migrations or environment-variable changes.
- Fresh Docker and Helm installations use the new index automatically.
- Existing one-click installations must download the current updater from the directory containing their
  `formbricks/` folder before running `update`; manual Compose and Helm deployments with an explicit/reused
  digest must follow the v5.4 migration guide.
- The updater never removes the named Redis volume or flushes Valkey data.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Every bundled deployment and CI surface uses one canonical index | unit (mutation): `pnpm exec vitest run docker/__tests__/valkey-image.test.ts` | Docker, Helm, updater and all applicable workflow pins matched the 8.1.9 index; the updater's recognized predecessor was asserted explicitly; the release branch variant also passed without the main-only API workflow |
| One-click migration changes only the known old pin and fails closed | unit (mutation): `pnpm exec vitest run docker/__tests__/formbricks-script.test.ts` | Exact old pin was backed up and migrated under standard and four-space indentation; restricted Compose file permissions were preserved; current/custom refs were no-ops; the exact `docker compose -f <path> config` validation command was asserted; failed validation restored the original file |
| Published digest is a native multi-architecture index | manual: `docker buildx imagetools inspect --raw` | OCI index contained `linux/amd64` child `07d7…dc60` and `linux/arm64` child `7e01…38af` |
| Existing AOF data survives the Debian-to-Alpine image change | manual container upgrade on an ARM64 Docker host | Session-like TTL data, queued data and cached data written by the old amd64 child remained readable under native ARM64 Valkey 8.1.9 |
| Fresh bundled stack runs with native ARM64 Valkey | manual production Compose boot | Migrations completed, `/health` returned `200`, `/setup/intro` rendered, the app connected to Valkey internally, and Valkey exposed no host port |
| Compose and Helm render the immutable index | manual: `docker compose config`, `helm lint`, `helm template` | Production/development Compose validated and the rendered StatefulSet used the exact index digest |
| Both target children meet the vulnerability gate | manual: Trivy scan by exact child digest | Both Alpine 3.24.1 children returned zero known Critical, High, Medium or Low findings with the current database |

**Open gaps**

- A native amd64 full-stack boot was not available locally; the amd64 GitHub service workflows provide that
  platform gate on this PR.
- A live Helm PVC upgrade was not run because no disposable Kubernetes cluster was in scope; the rendered
  StatefulSet and the equivalent named-volume AOF upgrade were verified.
- The isolated fresh stack stopped at the setup page, so authenticated login and survey-response journeys
  remain for release QA.
- The `release/5.4` backport is prepared locally but will be published only after this PR merges.

**Risks**

- The runtime base changes from Debian to Alpine. A real cross-image AOF upgrade and a full native ARM64 stack
  boot passed, but platform-specific behaviour should still be watched in CI and release QA.
- The updater intentionally recognizes only the exact old bundled digest. Operators with custom image
  references must update those references themselves.

---

> [!NOTE]
> **AI model used** — `unknown` (Codex), reasoning effort `unknown`.


kiln-agent:
## Review

The core change is correct and I verified its central claims independently rather than taking the description's word for them. Four findings, none blocking; the doc one is worth fixing before self-hosters follow it.

### Verified independently

| Claim | Result |
| --- | --- |
| New digest is a native multi-arch index | ✅ `e0eb7c48…` is `application/vnd.oci.image.index.v1+json` with `linux/amd64`, `linux/arm64`, `linux/arm/v7`, `linux/ppc64le` children |
| Old digest was amd64-only | ✅ `12ba4f45…` is a bare `manifest.v2+json`, config `architecture: amd64` — so the premise holds |
| Digest matches the documented version | ✅ `e0eb7c48…` is exactly the `8.1.9-alpine` tag; old image self-reports `v=8.1.1`, so "8.1.1 → 8.1.9 Alpine" in the migration guide is accurate |
| Cross-image AOF upgrade preserves data | ✅ Reproduced on the same named volume: wrote a string, a `SETEX` key, a list and a hash under Debian 8.1.1, restarted on Alpine 8.1.9 — `dbsize` 4, TTL preserved (3590), list order preserved, hash intact |
| Container contract unchanged | ✅ Booted with the bundled `valkey-server --appendonly yes --maxmemory-policy noeviction`; healthcheck `valkey-cli ping` → `PONG`, both configs read back correctly, `appendonlydir` created |
| New CI verify step works | ✅ `docker compose -f docker/docker-compose.yml config --format json \| jq -r '.services.redis.image'` resolves to the pinned index locally, and `Validate Docker Build` is green on this head |
| No stragglers on the old pin | ✅ The only remaining occurrences of `12ba4f45…` are the intentional `legacy_valkey_image` in `formbricks.sh` and its test constant |
| Both new tests can actually fail | ✅ Mutation-checked: bogus digest in `.github/workflows/e2e.yml` → `valkey-image.test.ts` fails; bogus `multi_arch_valkey_image` in `formbricks.sh` → `formbricks-script.test.ts` fails |

The `sudo`-stub design in `migrateLegacyValkeyImage` is a good call — driving the "leaves untouched" cases with `validationResult: "failure"` means those tests would fail if the function *did* migrate, so they prove a no-op rather than just asserting one.

---

### 1. The documented backup is taken *after* the edit — `docs/self-hosting/advanced/migration.mdx:47-58`

The Docker Compose tab says "**Back up** your Compose file and replace the `redis` service's old image", then shows only the YAML. The actual `cp` sits in the *next* block:

```bash
cp docker-compose.yml docker-compose.yml.before-valkey-8.1.9
docker compose config >/dev/null
```

An operator following the steps in order hand-edits the image line first, so the "backup" is a copy of the already-modified file — it captures the new digest and there is nothing to roll back to. Worth hoisting the `cp` into its own step ahead of the edit, so the manual path gets the same rollback the One-Click path advertises at line 36.

### 2. The not-found path fails open and silent — `docker/formbricks.sh:993-999`

Every other path in `migrate_legacy_valkey_image` fails closed and says so, but the detection `awk` returning no match is an unconditional silent `return 0`. That correctly covers "already migrated" and "custom image", but it also swallows "the legacy pin is right there, just not in the shape the anchor expects" — the match requires `^  redis:` at exactly two spaces.

Reproduced against the real function:

```
[redis last + top-level volumes:]  rc=0  ✅ migrated
[commented-out legacy + live pin]  rc=0  silent, no migration   ← correct
[legacy pin in a non-redis service] rc=0  silent, no migration  ← correct
[duplicate legacy pins in redis]   rc=1  ❌ fails closed        ← correct
[4-space indented compose]         rc=0  silent, STILL amd64-only
```

So an operator on ARM64 whose Compose file is indented differently runs the documented `./formbricks.sh update`, sees no message about Valkey at all, and stays on the amd64-only image believing the migration applied. Blast radius is genuinely small — the installer only ever `curl`s and `sed`s the 2-space bundled file, so a pure one-click install can't hit this — but `one-click.mdx` does tell people to hand-merge Compose changes for major upgrades, which is how a file stops being byte-identical.

Cheap fix: if `$legacy_valkey_image` appears anywhere in the file but the redis-scoped match didn't fire, warn and point at the migration guide's Docker Compose tab. That distinguishes "nothing to do" from "couldn't do it" without changing any behaviour.

### 3. The validation log is written and never asserted — `docker/__tests__/formbricks-script.test.ts:54,62,74`

The `sudo` stub appends its argv to `VALIDATION_LOG_PATH`, but no test reads that file — only the stub's exit status matters. So nothing verifies the function actually ran `docker compose -f "$compose_file" config`; it would pass just as happily if the validation call were dropped or pointed at the wrong file. Either assert the log contains `docker compose -f <path> config`, or drop the log and keep the stub to its exit code.

### 4. Nit: the fail-closed path leaves its backup behind — `docker/formbricks.sh:1012-1015`

When `replacements != 1`, the temp file is cleaned up but the `.before-valkey-8.1.9` copy created at line 1001 is not, while the message says "The original Compose file is unchanged." True of the Compose file, but a new file did appear. An `rm -f "$backup_file"` alongside the `rm -f "$temp_file"` would make the filesystem match the message. (Harmless either way — the backup is byte-identical to the original.)

### Optional

`valkey-image.test.ts`'s `sharedDirectPinFiles` omits `docker/formbricks.sh`, the one remaining file holding a forward pin. Drift there *is* caught today — I confirmed a mutated `multi_arch_valkey_image` fails `formbricks-script.test.ts` — but only indirectly, via an equality assertion in a migration test. Adding it to the list makes the guard explicit and keeps the failure message on-topic for the next digest bump.

### Considered and dismissed

- **Docker Hub dependency in the new CI step.** `docker buildx imagetools inspect` adds an unauthenticated registry fetch to a required job, but `harden-runner` here is `egress-policy: audit` (not `block`), and the job already pulls the Valkey service container plus Dockerfile base images from the same host — so this introduces no new failure class.
- **`js-yaml` in the new test.** Already a catalogued devDependency of the `docker` workspace; no catalog-check violation.
- **The breaking-change tick.** Defensible: `redis.image.digest` is a config-key default that changes, which the template lists explicitly, and the table plus migration section satisfy the CodeRabbit check. No objection.

---

**Verdict:** approve-with-comments in substance. Finding 1 is a real defect in operator-facing guidance and cheap to fix; 2-4 are hardening. Nothing here needs to block the merge on my account.

<sub>Note: this review was produced without the repo's usual `pr-review` skill — the account skill directory did not sync into this container, so the checks above were assembled from `AGENTS.md` conventions plus direct verification rather than the skill's checklist.</sub>

---

docs/self-hosting/advanced/migration.mdx
```diff
@@ -6,15 +6,81 @@ icon: "arrow-right"
...
+    index:
+
+    ```yaml
+    image: valkey/valkey@sha256:e0eb7c480958d32bdc4357a74bdd70653ae15f2f9b4c93c4a5a9fad1dc471c84
+    ```
+
+    Validate the file, then pull and recreate only Valkey:
+
+    ```bash
+    cp docker-compose.yml docker-compose.yml.before-valkey-8.1.9
```
kiln-agent: The backup is taken *after* the edit, so it isn't a rollback point.

The tab's prose above tells the operator to "Back up your Compose file and replace the `redis` service's old image" and then shows the new `image:` line to paste in — so by the time they reach this `cp`, `docker-compose.yml` already carries the new digest. `docker-compose.yml.before-valkey-8.1.9` therefore holds the *post*-edit file, and an operator who needs to roll back copies the new digest right back over.

That also makes the two documented paths disagree on what that filename means: `migrate_legacy_valkey_image` in `docker/formbricks.sh` does `cp -p "$compose_file" "$backup_file"` *before* rewriting, so on the one-click path the same name genuinely is the pre-upgrade file.

Moving the `cp` ahead of the edit fixes both — e.g. lead the tab with:

```bash
cp docker-compose.yml docker-compose.yml.before-valkey-8.1.9
```

then the `image:` replacement, then `docker compose config >/dev/null` and the rest of the block as written.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,57 @@ stop_formbricks() {
...
+    }
+    { print }
+    END { exit(replacements == 1 ? 0 : 1) }
+  ' "$compose_file" >"$temp_file"; then
+    rm -f "$temp_file"
+    echo "❌ Could not update the bundled Valkey image. The original Compose file is unchanged."
+    return 1
+  fi
+
+  mv "$temp_file" "$compose_file"
```
kiln-agent, line 1050: `mv` from a fresh temp file resets the Compose file's mode to the umask default, silently undoing operator hardening.

`$temp_file` is created by the `>` redirection, so it gets `0666 & ~umask` — typically `644`. Moving it over `$compose_file` replaces the original inode and its permissions along with it. Reproduced on a `600` Compose file:

```
before: 600 docker-compose.yml
after:  644 docker-compose.yml
        600 docker-compose.yml.before-valkey-8.1.9
```

The Compose file next to `formbricks.sh` holds `POSTGRES_PASSWORD`, `ENCRYPTION_KEY` and the other generated secrets, so an operator who tightened it to `600` gets it quietly widened to world-readable by `update`. The backup keeps the old mode, which makes the drift easy to miss.

It's also inconsistent with the surrounding code — both `cp` calls in this function already use `-p` specifically to preserve mode.

Simplest fix is to have the temp file inherit the original's attributes before it's written, so the `mv` carries them:

```bash
cp -p "$compose_file" "$temp_file"
```

right before the rewriting `awk` (whose `>"$temp_file"` then truncates the copy in place, keeping the mode). Alternatively `cat >"$compose_file" <"$temp_file" && rm -f "$temp_file"` in place of the `mv`, which never replaces the inode.

---

docker/__tests__/valkey-image.test.ts
```diff
@@ -0,0 +1,42 @@
...
+const valkeyIndexDigest = "sha256:e0eb7c480958d32bdc4357a74bdd70653ae15f2f9b4c93c4a5a9fad1dc471c84";
+const valkeyImage = `valkey/valkey@${valkeyIndexDigest}`;
+
+const sharedDirectPinFiles = [
+  ".github/workflows/docker-build-validation.yml",
+  ".github/workflows/e2e.yml",
+  ".github/workflows/integration-tests.yml",
+  "docker-compose.dev.yml",
+  "docker/docker-compose.yml",
+  "docker/migrate-to-v4.sh",
```
kiln-agent, line 20: `docker/formbricks.sh` is missing from this list, and it's the one pin that reaches operators' own files.

The list covers the six files that pin the image for *our* stacks and CI, but `formbricks.sh` now carries two copies of the digest — `multi_arch_valkey_image` is what `migrate_legacy_valkey_image` writes into an existing installation's `docker-compose.yml`. That makes it the highest-consequence pin in the PR and the only one not asserted here.

The gap is the one this suite exists to close: on the next Valkey bump, someone updates `docker/docker-compose.yml` and the workflows, the whole suite stays green, and the updater keeps rewriting operators' Compose files to a stale digest — arguably worse than not migrating them at all.

`readFileSync(...).toContain(valkeyImage)` works unchanged on a shell file, so adding it to the array is enough:

```ts
const sharedDirectPinFiles = [
  ".github/workflows/docker-build-validation.yml",
  ".github/workflows/e2e.yml",
  ".github/workflows/integration-tests.yml",
  "docker-compose.dev.yml",
  "docker/docker-compose.yml",
  "docker/formbricks.sh",
  "docker/migrate-to-v4.sh",
];
```

Worth considering a companion assertion that `legacy_valkey_image` in `formbricks.sh` still matches the digest the migration is meant to recognize — that pair is what makes the migration fire, and a bump that rewrites only `multi_arch_valkey_image` would leave it pointed at the wrong predecessor.

---

.github/workflows/docker-build-validation.yml
```diff
@@ -53,6 +53,20 @@ jobs:
...
+      - name: Verify Valkey Image Platforms
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          valkey_image="$(docker compose -f docker/docker-compose.yml config --format json | jq -r '.services.redis.image')"
+          manifest="$(docker buildx imagetools inspect --raw "$valkey_image")"
+
+          jq -e '
+            .mediaType == "application/vnd.oci.image.index.v1+json" and
```
kiln-agent, line 65: The `mediaType` equality makes this guard stricter than the invariant it's protecting.

The thing being guarded is "the pin resolves to a native amd64 *and* arm64 child", and the two `any(.manifests[]; …)` clauses below already state exactly that. This line adds a third, unrelated condition: that the multi-arch container be an **OCI** index specifically.

A multi-arch pin published as `application/vnd.docker.distribution.manifest.list.v2+json` — still the default for a plain `docker manifest create`/`push`, and what plenty of upstream images still ship — satisfies the real requirement but fails here. Because this is a required job, that failure looks identical to the regression it's meant to catch (a single-arch pin sneaking in), so whoever bumps the digest next gets sent hunting for a platform problem that doesn't exist.

Note the current pin is a genuine OCI index, so this isn't failing today — it's a trap set for the next bump.

Dropping the clause loses nothing: a single-platform `manifest.v2+json` has no `.manifests` key at all, so `any(.manifests[]; …)` errors out and `jq -e` still fails the step. If you'd rather keep it explicit, accepting either type keeps the guard honest:

```jq
(.mediaType == "application/vnd.oci.image.index.v1+json" or
 .mediaType == "application/vnd.docker.distribution.manifest.list.v2+json") and
```

---

kiln-agent (commented):
Reviewed the multi-arch Valkey pin. The core change holds up — I verified both digests directly against the registry rather than taking the description's word for it:

- The old pin `sha256:12ba4f45…` really is a single-platform `application/vnd.docker.distribution.manifest.v2+json` (Debian bookworm), so the ARM64 failure this fixes is real.
- The new pin `sha256:e0eb7c48…` is a genuine OCI index with `linux/amd64`, `linux/arm64`, `linux/arm/v7` and `linux/ppc64le` children, all `VALKEY_VERSION=8.1.9` on Alpine 3.24.1.
- Every occurrence of the old digest is replaced, and no other Valkey/Redis image reference was missed.
- The `docker` workspace tests pass (18/18). I also ran `migrate_legacy_valkey_image` against a realistic Compose file: the awk state machine correctly scopes the rewrite to the `redis` service and does **not** misfire on the `redis` entry in the top-level `volumes:` block — that was my main worry with a line-oriented matcher, and it's handled.
- On the Debian→Alpine AOF concern in your Risks section: the `valkey` uid stays 999 across both images (the gid moves 999→1000), so data in the named volume and the Helm PVC stays owner-writable. That matches your manual upgrade result.

Four findings, none blocking the premise of the change. The `formbricks.sh` ones are the two I'd want addressed before merge, since both affect operators' own files rather than our stacks:

| # | Where | What |
| --- | --- | --- |
| 1 | `docker/formbricks.sh:1018` | `mv` from a fresh temp file resets the Compose file's mode to the umask default (reproduced: `600` → `644`), silently widening a file that holds the generated secrets |
| 2 | `docker/__tests__/valkey-image.test.ts:11` | `formbricks.sh` isn't in `sharedDirectPinFiles`, yet it holds the digest the updater writes into operators' Compose files — a future bump can leave it stale with the suite green |
| 3 | `docs/self-hosting/advanced/migration.mdx:57` | The Docker Compose tab backs up *after* the edit, so the rollback file holds the new digest; the one-click path backs up first, so the two paths disagree on what that filename means |
| 4 | `.github/workflows/docker-build-validation.yml:65` | Hardcoded OCI-index `mediaType` would fail a valid multi-arch pin published as a Docker manifest list, indistinguishably from a real single-arch regression |

#3 and #4 are cheap and worth taking now — #4 in particular is a trap for whoever bumps this digest next, since a required job would go red for a reason that isn't the one it names.

On the PR description: the Coverage table and Open gaps are unusually thorough, and the breaking-change tick reads correct to me — existing self-hosters with a retained digest genuinely need action, which is what the template's list calls breaking. No objection there.

---

docker/formbricks.sh
kiln-agent, line 1050: **The `mv` silently widens the Compose file's permissions.**

`>"$temp_file"` creates the temp file with the caller's umask (typically `0644`), and `mv` carries that mode onto `docker-compose.yml`. Because `install_formbricks` seds `NEXTAUTH_SECRET`, `ENCRYPTION_KEY`, `CRON_SECRET`, `SMTP_PASSWORD` and the S3 keys straight into this file, an operator who hardened it with `chmod 600` gets it reset to world-readable by an update that was only supposed to swap an image digest — and nothing in the output says so.

Reproduced against the bundled template:

```
before: 600
after:  644
backup: 600     # cp -p, so only the live file is widened
```

Preserve the mode instead of inheriting the umask, e.g.:

```suggestion
  cat "$temp_file" >"$compose_file"
  rm -f "$temp_file"
```

(`cp -p "$compose_file" "$temp_file"` before the awk redirect, or `chmod --reference="$backup_file" "$temp_file"` before the `mv`, work equally well.)

---

docs/self-hosting/advanced/migration.mdx
kiln-agent: **The backup is taken after the edit, so it cannot roll anything back.**

The tab tells the operator to replace the `redis` image first (line 47-51), and only then run this block. By the time `cp docker-compose.yml docker-compose.yml.before-valkey-8.1.9` executes, the file already contains the new digest — so `docker-compose.yml.before-valkey-8.1.9` is a copy of the *new* state. If `docker compose up -d --force-recreate redis` then fails on their host, the file they were told to keep as a fallback has no trace of the old pin, and they have nothing to restore from.

The one-click tab is correct here (the script backs up before rewriting); this manual path just needs the `cp` moved ahead of the edit — e.g. show `cp docker-compose.yml docker-compose.yml.before-valkey-8.1.9` in its own block above the YAML snippet, and start this block at `docker compose config >/dev/null`.

---

.github/workflows/docker-build-validation.yml
```diff
@@ -53,6 +53,20 @@ jobs:
...
+            any(.manifests[]; .platform.os == "linux" and .platform.architecture == "amd64") and
+            any(.manifests[]; .platform.os == "linux" and .platform.architecture == "arm64")
```
kiln-agent, line 67: **This puts a live Docker Hub query on the critical path of a required PR check, to assert something immutable.**

The image is pinned *by digest*, so the manifest these two `jq` calls inspect can never change: `valkey/valkey@sha256:e0eb7c…` will return the same OCI index with the same amd64/arm64 children forever. The check therefore has exactly two outcomes — pass, or a registry hiccup. Concretely: an anonymous `docker buildx imagetools inspect` that gets a 429 from Docker Hub's unauthenticated pull limit (shared runner egress IPs) or a transient 5xx makes `set -euo pipefail` fail the step, and since it sits *before* `Build Docker Image`, the whole Docker Build Validation job goes red on a PR that never touched Docker.

What it does buy is a guard on a *future* pin change — but that guard can be static: `docker/__tests__/valkey-image.test.ts` already asserts the index digest, so asserting the two child digests (`07d7f634…` amd64, `7e01e837…` arm64) there gives the same protection offline, with no network and no PR-gate cost. If you want to keep the live check, please move it after the build (so a registry blip doesn't mask the build signal) or make it non-blocking.

---

docs/self-hosting/setup/one-click.mdx
```diff
@@ -319,21 +319,30 @@ Y
...
 
-To update Formbricks for a minor or patch release, run:
+To update Formbricks for a minor or patch release, first download the current stable updater and then run it:
 
-```
+```bash
+curl -fsSL https://raw.githubusercontent.com/formbricks/formbricks/stable/docker/formbricks.sh \
+  -o formbricks.sh
+chmod +x formbricks.sh
 ./formbricks.sh update
```
kiln-agent, line 330: **The new download step has no directory context, and `update` only works from the parent of `formbricks/`.**

`update_formbricks` starts with `cd formbricks`. Before this change the operator ran the script that was already sitting in the right place, so the working directory was self-evident. Now the docs hand them a `curl … -o formbricks.sh` that drops a fresh copy into whatever directory they happen to be in — and this page elsewhere routinely tells them to `cd formbricks` (`cd formbricks && sudo docker compose ps` / `logs`). Run verbatim from inside `~/formbricks`, they get `~/formbricks/formbricks.sh`, `cd formbricks` fails, `set -e` aborts, the stack is not updated, and a stray script now sits next to their Compose file.

Worth one line: "from the directory that contains your `formbricks/` folder". Also note the `-o formbricks.sh` overwrites an existing updater without a backup.

---

kiln-agent (commented):
Reviewed the digest swap end to end. The core premise checks out — I verified against Docker Hub directly that `sha256:e0eb7c48…` is the `8.1.9-alpine` OCI index with `linux/amd64` (`07d7f634…`) and `linux/arm64` (`7e01e837…`) children, that the old `12ba4f45…` was a bare single-platform `manifest.v2` (so the ARM64 complaint is real), and I booted the new image with the bundled `valkey-server --appendonly yes --maxmemory-policy noeviction` command (PONG, Alpine 3.24.1, AOF files created, `/data` chowned to `valkey` by the entrypoint). `redis-cli` is still symlinked in the Alpine variant, so `apps/web/integration/global-setup.ts`'s flush keeps working. All pins in the repo are consistent — no stale `12ba4f45…` anywhere except the intentional `legacy_valkey_image` constant and its test fixture.

I also ran `migrate_legacy_valkey_image` against a downgraded copy of the real `docker/docker-compose.yml` with a real `docker compose config` (no mock): it migrates exactly one line, backs up, and validates. The awk block-scoping holds up on the shapes `install_formbricks` actually produces.

Four findings, none blocking the approach:

1. `docker/formbricks.sh:1018` — the `mv` resets the Compose file's mode to the umask default; a `chmod 600` secrets-bearing file comes back `644` (reproduced).
2. `docs/self-hosting/advanced/migration.mdx:57` — the manual backup is taken *after* the edit, so it holds the new digest and can't roll back.
3. `.github/workflows/docker-build-validation.yml:56` — live registry inspect ahead of the build, gating every PR on an assertion that is immutable by construction.
4. `docs/self-hosting/setup/one-click.mdx:324` — the new `curl` step gives no working-directory context, and `update` requires the parent of `formbricks/`.

---

docs/self-hosting/setup/one-click.mdx
BhagyaAmarasinghe: Addressed in 09adb1b0b: the update instructions now require running from the directory that contains `formbricks/` and state that the downloaded stable updater replaces an older `formbricks.sh`.

docker/formbricks.sh
BhagyaAmarasinghe: Addressed in 09adb1b0b: the temporary file now inherits the original attributes through `cp -p` before the rewrite. The migration test starts from mode `0600` and asserts that both the updated Compose file and its backup remain `0600`.

.github/workflows/docker-build-validation.yml
BhagyaAmarasinghe: Kept intentionally: the accepted implementation plan explicitly requires Docker validation to inspect the remote manifest and verify the OCI index platforms. This is the future pin-change gate requested for the deployment workflow; the job already pulls Docker Hub artifacts.

.github/workflows/docker-build-validation.yml
BhagyaAmarasinghe: Kept intentionally: the accepted implementation plan explicitly requires Docker validation to reject anything other than an OCI index with native amd64 and arm64 children. The exact media type is part of the requested invariant, not only an implementation detail.

docker/__tests__/valkey-image.test.ts
BhagyaAmarasinghe: Addressed in 09adb1b0b: the synchronization list now includes `docker/formbricks.sh`, and a companion assertion locks both the current and expected legacy migration pins.

docs/self-hosting/advanced/migration.mdx
BhagyaAmarasinghe: Addressed in ea9f9bdc8: the manual procedure now creates `docker-compose.yml.before-valkey-8.1.9` in a separate step before the operator changes the image reference.

pandeymangg (approved):

PR #8987, opened 2026-08-26
feat(unify): add "Re-import historic data" to Formbricks survey sources

Dhruwang:
Fixes ENG-1889

## What & why

**Was:** fixing a source's question mapping meant deleting the source and rebuilding it to replay the same responses.

**Now:** an active Formbricks survey source with a mapped survey offers **Re-import historic data**, behind a confirmation dialog.

- Reuses `importHistoricalResponsesAction`: it reconciles rather than inserts, and honours the saved `importMode`.
- Replays every distinct mapped survey, sequentially — not `formbricksMappings[0]`.
- Every create conflicts on a re-import, so the SDK's 409 retries are narrowed.

## Where to look

- [`hub-client.ts`](https://github.com/formbricks/formbricks/blob/5463639/apps/web/modules/hub/hub-client.ts#L74-L109) — which of the create's two 409s is terminal
- [`sources/utils.ts`](https://github.com/formbricks/formbricks/blob/5463639/apps/web/modules/ee/unify-feedback/sources/utils.ts#L53-L73) — when the item is offered
- `sources/components/feedback-sources-page-client.tsx:246-294` — the handler and its toasts

## Breaking changes

None — a new menu entry reusing an existing server action with unchanged inputs and authorization.

## Migrations & env

- none

## How this was tested

Rerun: `cd apps/web && npx vitest run modules/hub modules/ee/unify-feedback lib/feedback-source`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Re-import replays a source without duplicating records | manual | 8 responses → `6/0/2`; hub rows 274 → 275; re-run identical |
| Offered only where it can work | unit (mutation), `sources/utils.test.ts` | Mutate `utils.ts:71-73`: paused, error, unmapped, CSV go red |
| Each mapped survey replayed once, totals summed | unit (mutation), `sources/utils.test.ts` | Mutate `utils.ts:54` (`Set` → `[0]`) or `:87` |
| Failures announced as failures at all three import sites | unit (mutation), `sources/utils.test.ts` | Invert `utils.ts:116`: four cases go red |
| `skipped` counts only this survey's mappings | unit (mutation), `import.test.ts` | Mutate `import.ts:32` to `responses.length * mappings.length` |
| Duplicate create not retried, nothing else stops being | unit (mutation), `hub-client-retry.test.ts` | POST attempts 1 / 3 across 7 cases; mutate `hub-client.ts:108` |

<details>
<summary>Walkthrough, and the detail behind the rows</summary>

https://github.com/user-attachments/assets/5cf85975-850e-4086-8d5f-891d0da1669f

**Not duplicating.** Counted `feedback_records` for the source's `tenant_id` + `source_id` in the hub database before and after: **274 → 275**, and `updated_at` moved to now. One record was genuinely missing and got added; the other five were reconciled in place. Six duplicates would have taken it to 280. A second run gave the identical toast (`6 succeeded, 0 failed, 2 skipped`) and left the count at 275. The 2 skipped are the response×mapping pairs with no answer for that question, which is what `skipped` counts.

**The create's two 409s.** Hub answers 409 both for the duplicate `(tenant_id, submission_id, field_id)` (`code: "conflict"`, terminal) and when the insert's purge gate refuses the row (`code: "tenant_write_conflict"`, retryable — `NewTenantWriteConflictError` in its feedback-records repository, documented on this operation in its `openapi.yaml`). So the narrowing reads the problem code off a clone of the response and treats only the duplicate as terminal; an unreadable body leaves the SDK's own policy alone, making the degradation extra retries rather than a suppressed one. `hub-client-retry.test.ts` leaves `@formbricks/hub` real and stubs only the transport, asserting POST attempts: **1** for a duplicate create, **3** for a purge-conflict create, a code-less 409, a rate-limited create, a 409 on `tenants.settings.update`, one on `feedbackRecords.update`, and one whose response carries no URL.

**Why the menu item reads the source's state.** `canReimportHistoricalData` requires `formbricks_survey`, `status: "active"` and a mapping. Pausing is the user's switch for stopping a source writing into its directory — the live pipeline honours it, but neither the action nor the import re-checks, so without that clause the menu would offer to replay a whole history into a directory the user had just paused the source to keep out of.

**What a re-import does not repair.** Hub's PATCH carries only `value_*`, `metadata`, `language` and `user_id` (`UPDATE_FIELD_KEYS` in `lib/feedback-source/reconcile.ts`); `field_label`, `field_type` and `source_name` are create-only. Retyping a mapped question or renaming the survey and re-importing reports full success while those records keep their old label and type. The dialog copy says so; fixing it belongs in `reconcile.ts`, where it affects the create-time import equally.

**Dialog copy.** `ConfirmationModal` otherwise falls back to "This action cannot be undone.", the opposite of what a reconciling import does; it now reads "This can be run again at any time."

**The row dropdown** gained a prop threaded through `feedback-sources-table.tsx`; that plumbing is where to look if a source row menu misbehaves.

</details>

**Open gaps**

- A re-import does not repair `field_label` or `field_type` on already-imported records — see the fold.
- No e2e spec; the multi-survey path is unit-only, since the create dialog cannot produce such a source.
- The narrowing's wall clock comes from a stub Hub (150 POSTs/420ms vs 450/1911ms reverted).

**Risks**

- The import runs synchronously in the server action, so a very large survey can time out.
- `historicalResponseImport` allows 10/hour per user, so a source with more mapped surveys stops mid-run.
- The saving rests on Hub keeping `code: "conflict"` for duplicates; a change there silently restores the retry storm, unguarded.

---

> [!NOTE]
> **AI model used** — `claude-opus-5 (Claude Code)`, reasoning effort `high`.


kiln-agent (commented):
Reviewed `c04a1da` (21 files, +320/-1). Ran `apps/web` → `vitest run modules/ee/unify-feedback/sources/utils.test.ts`: 102/102 pass. Typecheck, Linters, Translation Keys and CodeQL green on this head; Unit Tests, E2E, Sonar, Docker and contract tests still running.

The shape is right, and the reasoning in the description holds up where I checked it. Authorization is genuinely reused rather than re-invented — `importHistoricalResponsesAction` pins the source to `workspaceId` via `getFeedbackSourceWithMappingsById` and separately re-checks `survey.workspaceId === workspaceId`, so the new entry point adds no tenant surface. Read-only is gated one level up: `feedback-sources-table.tsx:175` renders no dropdown at all when `isReadOnly`, so the item cannot appear for a read member. `buttonLoading` disables the confirm button (`button/index.tsx`: `disabled={loading || disabled}`), so no double-submit. And the test level matches AGENTS.md — logic lifted into `.ts` and unit-tested, no component test, no new Playwright spec. Deriving the item from the mappings rather than the type alone, and covering every distinct mapped survey, are both the right calls.

Three things, one of which I'd want addressed before merge.

---

### 1. The dialog copy promises a correction that a re-import cannot make

`reimport_historic_data_confirmation` says "Records whose answers changed are updated". That is true of answers and only of answers. Hub's PATCH accepts exactly `value_*`, `metadata`, `language`, `user_id` — `UPDATE_FIELD_KEYS` in `apps/web/lib/feedback-source/reconcile.ts:96`, enforced by the `satisfies` there. `field_label`, `field_type` and `source_name` are create-only.

The reachable case needs no mapping edit at all, because `field_label` is not stored on the mapping — `transform.ts:413` derives it from the survey's *current* headline on every run:

1. A question's headline has a typo. The source imported 200 responses under `field_label: "Whats your role"`.
2. Someone fixes the headline in the survey editor.
3. They run **Re-import historic data**, which is now the advertised remedy for exactly this.
4. Every record 409s, reconciles, and PATCHes only the value fields. All 200 keep the old label.
5. The toast reads `Import complete: 200 succeeded, 0 failed, 0 skipped` — reconciled records count as successes (`import.ts:54`).

So the operation reports complete success and the thing the user came to fix is unchanged. The same applies to `source_name` after a survey rename.

The `field_type` variant is worse than a no-op. If a question's type changed and the mapping was re-saved (`resolveFormbricksMappingsInput` re-derives `hubFieldType` from the element type on every save), the new type routes the answer to `value_number`, while `value_text` goes out as `record.value_text ?? undefined` — which `reconcile.ts:119` documents as "leave unchanged". The record ends up holding the stale `value_text` *and* the new `value_number` under the old `field_type`.

None of this is introduced here — it is reconcile's contract, and the live pipeline has the same ceiling. But this PR is what invites users to rely on it, and the confirmation dialog is where the expectation gets set. I'd narrow the copy to what actually happens (answers corrected, missing records added) and say plainly that question labels and types on already-imported records are not rewritten. A one-line follow-up ticket for the label case would be worth more than the copy change, but the copy is the part that ships today.

### 2. `skipped` is wrong for the multi-survey path this PR deliberately builds

The description makes a considered case for covering every mapped survey rather than `formbricksMappings[0]`, and `sumImportTotals` has tests proving the totals add up. The number being added up is the problem.

`importHistoricalResponses` hands `feedbackSource.formbricksMappings` — *all* of them — to `processBatch` (`import.ts:97`), which computes `expectedRecords = responses.length * mappings.length` (`import.ts:27`). But `transformResponseToFeedbackRecords` filters to `m.surveyId === survey.id` (`transform.ts:367`). So on a source mapping 3 questions each across survey A and survey B, the call for A over 10 responses reports `expected 60 − produced 30 = 30 skipped`, all of it phantom — and the loop then sums A's and B's phantoms into one figure.

Display-only, and unreachable today since the create dialog binds one survey — but the loop exists precisely for the day it isn't, and that is the day the number goes wrong. Passing only that survey's mappings (filter in the action, or in `processBatch`) closes it; a comment recording the known gap would do if this should stay out of scope.

### 3. Nit: `reimport_no_survey_mapped` cannot be reached

The item renders only under `canReimport`, which *is* `getMappedSurveyIds(...).length > 0`, and the handler then re-checks the same predicate. The toast is dead, and it carries a string into all 16 locale files. Fine as a guard for a future caller of `onReimport` that doesn't gate — worth a comment saying so, or drop the string.

### One question

Re-import is offered on `paused` and `error` sources too. Someone who disabled a source has switched live sync off; writing historic data into it may be exactly what they want after fixing something, or may surprise. Deliberate? A sentence either way is enough.

---

Nothing here blocks on correctness of the happy path — the feature does what the manual testing shows. My ask is (1), and it is about the promise rather than the code.

---

apps/web/modules/ee/unify-feedback/sources/utils.ts
```diff
@@ -38,6 +39,50 @@ export const getSelectableQuestionIds = (survey: TUnifySurvey): string[] =>
...
+  ...new Set(feedbackSource.formbricksMappings.map((mapping) => mapping.surveyId)),
+];
+
+/**
+ * Whether "Re-import historic data" applies to a source. Only Formbricks sources replay responses
+ * (`importHistoricalResponses` rejects every other type), and only one that names a survey has
+ * anything to replay from.
+ */
+export const canReimportHistoricalData = (feedbackSource: TFeedbackSourceWithMappings): boolean =>
+  feedbackSource.type === "formbricks_survey" && getMappedSurveyIds(feedbackSource).length > 0;
```
kiln-agent: **Re-import is still offered on a disabled source.**

The gate is type + mappings, but not `status`. "Disable" sits three items below "Re-import historic data" in the same menu, and it is the user's switch for stopping this source writing into its feedback directory — the live pipeline honours it (`getFeedbackSourcesBySurveyId` filters `status: "active"`, so nothing new is published for a paused source).

Neither `importHistoricalResponsesAction` nor `importHistoricalResponses` re-checks status, so on a source someone paused precisely to stop it feeding the directory, confirming this dialog replays the survey's entire response history into that directory. The create-time import could never reach this case — a source is always active the moment it is created — so it arrives with this entry point.

```suggestion
export const canReimportHistoricalData = (feedbackSource: TFeedbackSourceWithMappings): boolean =>
  feedbackSource.type === "formbricks_survey" &&
  feedbackSource.status === "active" &&
  getMappedSurveyIds(feedbackSource).length > 0;
```

(If replaying into a paused source is intentional, it is worth saying so in the dialog copy rather than leaving the two switches silently disagreeing.)

---

apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
```diff
@@ -233,6 +239,54 @@ export function FeedbackSourcesSection({
...
   };
 
+  /**
+   * Replays the linked survey's historic responses into this source again, so a mapping fixed after
+   * the first import no longer needs the source deleted and rebuilt (ENG-1889).
+   *
+   * Safe to run repeatedly: `importHistoricalResponses` reconciles rather than inserts, and carries
+   * a `snapshotAt` so a record the live pipeline has since corrected is not reverted to this older
+   * copy. It also honours the source's saved `importMode`, so a re-import never widens a
+   * completed-only source to partials.
```
kiln-agent: **"a mapping fixed after the first import" only half works, and the half that does not is silent.**

For a response that was never imported, the corrected mapping is applied on create. For one that was already imported, the record 409s and `reconcileConflict` PATCHes it with `toUpdateParams(record)` — and `toUpdateParams`, checked against Hub's `FeedbackRecordUpdateParams` by `UPDATE_FIELD_KEYS`, carries only `value_*`, `metadata`, `language`, `user_id`. `field_type` and `field_label` are identity fields on create and are not in the update request at all.

Concretely: a question mapped while it was `openText` is later retyped to `rating`. `mapping-reconciliation.ts` does its job and rewrites the stored mapping to `hubFieldType: "rating"` (its `retyped` branch). The user re-imports to fix the history and gets `Import complete: N succeeded, 0 failed`. But every pre-existing record keeps `field_type: "text"` and its stale `value_text`, and merely gains a `value_number` — so the charts still group it as text, and the source has to be deleted and rebuilt after all.

Not asking for it to be fixed here (it lives in `reconcile.ts`), but the comment and the dialog copy currently read as though a mapping fix is fully replayed. Worth either narrowing the claim to "questions added to the mapping" or noting the retype case under `Open gaps`.

---

apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
```diff
@@ -233,6 +239,54 @@ export function FeedbackSourcesSection({
...
+          return;
+        }
+
+        totals.push(importResult.data);
+      }
+
+      const { successes, failures, skipped } = sumImportTotals(totals);
+      // Spelled out rather than spread: t()'s options parameter is typed against i18next's
+      // $Dictionary, which an interface does not satisfy.
+      toast.success(t("workspace.unify.historical_import_complete", { successes, failures, skipped }));
```
kiln-agent: **A wholly failed re-import is reported as a success.**

`failures` is destructured and then only interpolated — the toast is `toast.success` no matter what it holds. A Hub outage does not throw: `createFeedbackRecordsBatch` returns per-record errors, `reconcileFeedbackRecords` folds them into `failures`, and the action resolves normally with `{ successes: 0, failures: 412, skipped: 0 }`.

So the user gets a green *"Import complete: 0 succeeded, 412 failed, 0 skipped"* and a `router.refresh()`, visually identical to the happy path, for an import that wrote nothing. Something like:

```ts
const message = t("workspace.unify.historical_import_complete", { successes, failures, skipped });
if (failures > 0) {
  toast.error(message);
} else {
  toast.success(message);
}
```

(The create-time path above has the same shape, so if you'd rather fix both together that's fair — but this entry point is the one users will reach repeatedly.)

---

apps/web/modules/ee/unify-feedback/sources/components/feedback-source-row-dropdown.tsx
```diff
@@ -155,6 +192,22 @@ export function FeedbackSourceRowDropdown({
...
+        // ConfirmationModal falls back to "This action cannot be undone." when given no
+        // description, which is the opposite of what a re-import does: it reconciles, so running
+        // it again is the remedy rather than the risk.
+        description={t("workspace.unify.reimport_historic_data_description")}
+        body={t("workspace.unify.reimport_historic_data_confirmation")}
+        buttonText={t("workspace.unify.reimport_historic_data_cta")}
+        buttonVariant="default"
+        buttonLoading={isReimporting}
+        onConfirm={handleReimport}
+        Icon={RefreshCwIcon}
```
kiln-agent, line 222: **The dialog can be dismissed mid-import, and then the loading state and the dialog disagree.**

`ConfirmationModal` defaults `closeOnOutsideClick` to `true` and renders the close button, and neither is tied to `isReimporting`. The confirm button is correctly disabled while loading, but Escape / an outside click / the X are not.

On the long imports this feature is for (the copy itself says "Large surveys can take a while"), a stray outside click closes the dialog while the request is still in flight and `isReimporting` stays `true`. Re-opening "Re-import historic data" then shows a spinning, disabled **Re-import** button with nothing explaining why it can't be clicked — and when the original request finally resolves, the `finally` block's `setIsReimportDialogOpen(false)` closes the dialog the user had just re-opened.

```suggestion
      <ConfirmationModal
        open={isReimportDialogOpen}
        setOpen={setIsReimportDialogOpen}
        title={t("workspace.unify.reimport_historic_data")}
        // ConfirmationModal falls back to "This action cannot be undone." when given no
        // description, which is the opposite of what a re-import does: it reconciles, so running
        // it again is the remedy rather than the risk.
        description={t("workspace.unify.reimport_historic_data_description")}
        body={t("workspace.unify.reimport_historic_data_confirmation")}
        buttonText={t("workspace.unify.reimport_historic_data_cta")}
        buttonVariant="default"
        buttonLoading={isReimporting}
        closeOnOutsideClick={!isReimporting}
        hideCloseButton={isReimporting}
        onConfirm={handleReimport}
        Icon={RefreshCwIcon}
      />
```

---

kiln-agent (commented):
Code review (high effort) — 4 findings, all inline. The plumbing itself is clean: mappings-derived gating, dedup across surveys, sequential imports, and the unit tests pass locally (102/102 in `sources/utils.test.ts`).

What I'd want addressed before merge:

1. **`canReimportHistoricalData` ignores `status`** — the entry point is offered on a source the user disabled to stop it writing into its feedback directory, and nothing downstream re-checks. New with this button; the create-time import can't reach it.
2. **The ENG-1889 use case is only half served, silently** — a mapping whose `hubFieldType` changed is not corrected on records already in Hub, because the reconcile PATCH carries no `field_type`/`field_label`. Reported as success.
3. **`toast.success` regardless of `failures`** — a Hub outage yields a green "0 succeeded, 412 failed".
4. **The dialog is dismissible mid-import**, leaving `isReimporting` stranded true.

(2) is inherited from `reconcile.ts` and (3) mirrors the create-time path, so both are fair to scope out — but they should be stated in `Open gaps` rather than left implied by the current comment and copy.

---

apps/web/modules/ee/unify-feedback/sources/components/feedback-source-row-dropdown.tsx
```diff
@@ -35,19 +39,33 @@ export function FeedbackSourceRowDropdown({
...
 
+  const handleReimport = async () => {
+    setIsReimporting(true);
+    try {
+      await onReimport();
+    } finally {
+      setIsReimporting(false);
+      setIsReimportDialogOpen(false);
+    }
+  };
```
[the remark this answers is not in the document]
Dhruwang: Guard added in c1b470d, but the double-import path as described is not reachable — worth recording why, because the reasoning matters for whoever reads this next.

`ConfirmationModal` renders its confirm button as `<Button loading={buttonLoading} disabled={isButtonDisabled} …>`, and `Button` applies `disabled={loading || disabled}` ([modules/ui/components/button/index.tsx:54](https://github.com/formbricks/formbricks/blob/main/apps/web/modules/ui/components/button/index.tsx#L54)). So `buttonLoading={isReimporting}` already disables it. Walking the sequence: confirm → `isReimporting` true → cancel or click outside → modal closes, import still running → reopen from the menu → same component instance, `isReimporting` still true → confirm button disabled. There is no second run to start.

I added the `if (isReimporting) return;` guard anyway, for a different reason than the one given: without it the invariant depends on the shared `Button` keeping `disabled={loading || disabled}`, which is a coupling this component should not rely on. That is a fair quick win, so thanks.

I did not block dismissal, deliberately. `ConfirmationModal.renderCancelButton` takes no `disabled` prop, so I could block the outside click and the X but not Cancel — a dialog you can close with one control and not another is worse than one you can always close. Adding `disabled` support to a shared modal used across the app to cover this single call site is not something this ticket should carry. The real gap left is cosmetic: dismiss mid-import and there is no indication it is still running until the toast lands. Worth fixing properly by making the import a background job (see the Risks section — the synchronous server action is the underlying problem), not by patching the dialog.

apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
```diff
@@ -233,6 +239,54 @@ export function FeedbackSourcesSection({
...
+    } catch {
+      toast.error(t("common.something_went_wrong"));
+      return;
+    }
+
+    router.refresh();
```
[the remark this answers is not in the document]
Dhruwang: Not taking this one in this PR, and I want to be explicit about why rather than silently skipping it.

The rule in AGENTS.md is "Server actions are legacy — do not add new ones. New backend work belongs in an `/api/v3` route consumed from the client with TanStack Query." This diff adds no server action and no backend work: `importHistoricalResponsesAction` already exists, is already called twice from this same page (`handleImportResponses` here and `create-feedback-source-modal.tsx`), and is invoked with unchanged inputs and authorization. What is new is one menu entry that calls it.

Doing what is suggested would mean building a new `/api/v3` route for historical import — i.e. new backend work, which is more than the ticket asks — and it would land in a page that has no TanStack Query wiring at all: `feedbackSources` arrives as a server prop (`initialFeedbackSources`), and create, delete and toggle-status all use server actions plus `router.refresh()`. Converting only the re-import path would leave one of four mutations on a different data model from its three neighbours, with a query to invalidate that nothing else populates. That is a worse state than either end point.

The right shape is to migrate the whole feedback-sources page in one go — route, query, and all four mutations — which is a standalone piece of work rather than a rider on a menu item. Happy to file that as a follow-up ticket if you want it tracked; say the word and I will.

Leaving this thread open rather than resolving it, since I have not acted on it — it is a call for a human reviewer, not for me to close.

apps/web/modules/ee/unify-feedback/sources/utils.ts
Dhruwang: Taken in 4a79c7f — `canReimportHistoricalData` now requires `status === "active"`. Verified your premise: `getFeedbackSourcesBySurveyId` filters `status: "active"` at [service.ts:149](https://github.com/formbricks/formbricks/blob/main/apps/web/lib/feedback-source/service.ts#L149), and neither the action nor `importHistoricalResponses` re-checks it. Two tests added, for `paused` and `error`.

On the `error` case I deliberately went with your `=== "active"` rather than the narrower `!== "paused"` I first reached for. The argument for `!== "paused"` is that `error` is not a user choice, so blocking re-import there could block a repair. But `updateFeedbackSourceWithMappings` already clears `error` back to `active` when a source is re-mapped ([service.ts:427](https://github.com/formbricks/formbricks/blob/main/apps/web/lib/feedback-source/service.ts#L427)), so the repair path restores `active` on its own and nothing is blocked. `=== "active"` also matches the pipeline exactly, which is the property that makes the two switches agree.

That also answers your "One question" at the end of the review: not deliberate, now closed.

apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
Dhruwang: Fixed in 4a79c7f, and thanks — this was a real bug and I had it backwards in my own head, assuming a Hub failure would throw.

Both call sites in this file now go through one helper that picks the toast by `failures`, so the create-time path is fixed alongside the re-import one rather than left divergent from a line I was already touching. `create-feedback-source-modal.tsx` still has the same shape and is listed under `Open gaps` — happy to fix it here too if you would rather it not linger, it is the same three lines.

apps/web/modules/ee/unify-feedback/sources/components/feedback-sources-page-client.tsx
Dhruwang: Confirmed and partly acted on in 4a79c7f. `reconcile.test.ts:94` states it outright — "The identifying fields (source_*, field_*, submission_id, collected_at) are not part of that request and must not be sent" — so `field_type` and `field_label` genuinely survive a reconcile.

Your headline example in the review body is the sharper one and it changed what I did here. `field_label` is not stored on the mapping at all: `transform.ts` derives it from the survey's current headline on every run, so the scenario needs no mapping edit — fix a typo in the question, re-import to backfill it, get `200 succeeded, 0 failed`, and all 200 records keep the old label. Same for `source_name` after a survey rename. A user following the dialog's advice would conclude it worked.

So rather than only narrowing the code comment I also changed the copy. `reimport_historic_data_confirmation` now reads "Answers that changed are updated and missing records are added … Question labels and types are not rewritten on records that were already imported." The handler comment names `UPDATE_FIELD_KEYS` as the reason and says the fix belongs in `reconcile.ts`, and `Open gaps` carries the retype, headline-fix and rename cases.

Not fixing the underlying behaviour here, agreed — it is `reconcile.ts`, it affects the create path equally, and making `field_*` updatable is a Hub contract question rather than a menu-item one.

apps/web/modules/ee/unify-feedback/sources/components/feedback-source-row-dropdown.tsx
Dhruwang: Taken, with one deviation. 4a79c7f adds `closeOnOutsideClick={!isReimporting}` as you suggested, but not `hideCloseButton={isReimporting}`.

The reason is that `ConfirmationModal.renderCancelButton` takes no `disabled`, so hiding the X while Cancel stays live gives a dialog you can close with one control and not the other. Blocking only the *accidental* dismissal and leaving both deliberate exits works is the consistent version of the same idea.

Your last point is the one I had missed and it needed a separate fix: the in-flight run's `finally` closing a dialog the user had just re-opened. The menu item is now `disabled` while `isReimporting`, so that state cannot be entered at all — no re-opening into an unexplained spinner, nothing for the `finally` to yank shut.

CodeRabbit raised the dismissal half of this too and claimed it allowed a second concurrent import; that part is not reachable, since `Button` applies `disabled={loading || disabled}` to `buttonLoading`. I added the in-flight guard in `handleReimport` anyway (c1b470d) so the invariant does not rest on the shared Button.

xernobyl (commented):
This looks good to me — the gating, the dialog copy and the failure toast all behave as described, and I checked the reconcile path end to end against a stub Hub: re-running it is genuinely idempotent, a changed answer gets rewritten, and adding a question to the mapping backfills just that field. Nice work on the `status === "active"` clause, that one would have bitten.

One thing I'd like to fix before this merges, and I don't think it's visible from the manual test.

The Hub SDK retries 409s. `shouldRetry` returns `true` for 409 ("Retry on lock timeouts") and `maxRetries` defaults to 2, and `getHubClient()` overrides neither. On a first import that costs nothing, because nothing conflicts — but a re-import is the case where *every* create conflicts, which `reconcile.ts:17` already says out loud, and `reconcile.ts:11` documents the 409 as terminal. So each record pays three POSTs and ~1.1–1.5s of backoff before the 409 it was always going to get.

60 responses against a localhost stub:

| | POSTs | wall clock |
| --- | --- | --- |
| first import | 60 | 29 ms |
| re-import | 180 | 2963 ms |

100× the wall clock, all of it retry sleep. At 10k records that's 30k POSTs and roughly 5 minutes of pure backoff inside a synchronous server action — which I think makes the timeout risk more than "pre-existing, unchanged": on a create-time import nothing conflicts, on a re-import everything does.

Two one-line options, both of which I measured back to 60 POSTs / 59 ms with identical results: `create(input, { maxRetries: 0 })` in `createFeedbackRecordsBatch`, or a `shouldRetry` override on the `FormbricksHubWithRepeatedArrayParams` subclass returning `false` for 409. I'd lean to the override since it keeps the 429/5xx retries.

xernobyl:
Two smaller things I left out of the review above to keep it on one point. Neither is a blocker.

**`create-feedback-source-modal.tsx:346` still reports a failed import as a success.** You list it under `Open gaps`, so this is only a nudge to take it here rather than let it linger — it's the same shape you already fixed twice in `feedback-sources-page-client.tsx`, and on this path the green toast also carries a link inviting the user to go look at records that were never written. Lifting `notifyImportResult` to take `t` would let all three sites share it instead of the two.

**The `import.ts` change ships untested, and I think a test is cheaper than the reasoning for skipping one.** The PR says it isn't covered because there's no behavioural change while a source binds one survey — but `import.test.ts` already asserts on `skipped` in two tests, so the fixtures are right there. This passes on your branch and fails with `expectedRecords` back to `responses.length * mappings.length`:

```ts
test("counts only this survey's mappings towards skipped", async () => {
  const twoSurveySource: TFeedbackSourceWithMappings = {
    ...mockFeedbackSource,
    formbricksMappings: [
      mockFeedbackSource.formbricksMappings[0],
      {
        ...mockFeedbackSource.formbricksMappings[0],
        id: "mapping-2",
        surveyId: "clxxxxxxxxxxxxxxxx009",
        elementId: "el-2",
      },
    ],
  };

  getResponses.mockResolvedValueOnce([{ id: "r1" }] as never);
  getResponses.mockResolvedValueOnce([]);
  transformResponseToFeedbackRecords.mockReturnValueOnce([{ field: "record1" }] as never);
  reconcileFeedbackRecords.mockResolvedValue({ created: 1, reconciled: 0, superseded: 0, failures: [] });

  const result = await importHistoricalResponses(twoSurveySource, mockSurvey);

  expect(result).toEqual({ successes: 1, failures: 0, skipped: 0 });
});
```

Worth having, I think — it's the one clause standing between an honest `skipped` and one that counts another survey's mappings.

Dhruwang (commented):
Taken in 85041b7f91, and thanks for measuring it — I would not have caught this from the manual run, because the create-time import is the one case where nothing conflicts.

Your diagnosis holds exactly. `shouldRetry` returns `true` for 409 at `client.ts:569` ("Retry on lock timeouts"), `maxRetries` defaults to 2 at `client.ts:205`, and `getHubClient()` overrides neither — so on a re-import, where every create conflicts by design, each record pays three POSTs and the backoff before the 409 it was always going to get.

Two corrections to the fix itself, both of which changed what I wrote.

**The subclass override doesn't compile.** `stringifyQuery` is `protected` in the SDK's types, which is why the existing override is a plain `protected override` method. `shouldRetry` is `private` (`client.d.ts:148`), and TypeScript forbids redeclaring a private base member in a subclass — so `override shouldRetry` is a build error, not a one-liner. It is still an ordinary prototype method at runtime, so the fix is installed on `FormbricksHubWithRepeatedArrayParams.prototype` with the base delegated to, and only when the base method is actually there: if a future SDK renames it, the SDK stops calling ours too, and the behaviour degrades to today's extra retries rather than to a broken client.

**A blanket 409 override would have cost something.** Hub uses 409 for two unrelated things. Besides the uniqueness conflict, `tenant_write_conflict` is returned by taxonomy node writes (`resources/taxonomy/nodes.d.ts:30,50`), tenant settings (`resources/tenants/settings.d.ts:31`) and the tenant delete, which "waits up to a configured lock timeout for them to drain and then returns a retryable 409" (`resources/tenants/tenants.d.ts:23`) — the retry there is the documented behaviour. So the override is scoped to the create rather than to the status.

That scoping has one sharp edge worth recording, because it is the thing to check if this ever misbehaves. `shouldRetry` receives only the `Response`, which does not carry the request method, so the discriminator is the path — and three operations share `/v1/feedback-records` exactly: `create` (POST), `list` (GET) and the delete-by-user bulk delete (DELETE), whose 409 *is* the retryable one. That is safe today for two reasons: a GET cannot conflict, and the bulk delete never goes through this client — its only caller, `feedback-records-proxy.ts`, forwards with raw `fetch`, so the SDK's retry policy never applies to it. If a caller ever issues that bulk delete through `getHubClient()`, this needs a narrower discriminator than the path. The comment says so at the call site.

`maxRetries: 0` on the create was the other option and I passed on it: `createFeedbackRecordsBatch` is shared with the live pipeline, which has no user to click "re-import" when a transient 429 turns into a per-record failure, and a bulk import is exactly where rate limiting bites.

Covered by a new `describe("conflict retries")` in `modules/hub/hub-client.test.ts` — 409 on the collection is terminal, 409 on `/v1/tenants/{id}` and on `/v1/feedback-records/{id}` still retries, 429 still retries, and a `baseURL` carrying a path prefix still resolves. Replacing the guard with any other predicate turns two of them red.

One thing I did **not** do: re-measure the request count and wall clock end to end on this branch. The override is proven at unit level, but the numbers in the PR are yours, not a fresh run here. It is listed under `Open gaps` rather than implied to be verified.

Dhruwang:
Both taken in 85041b7f91.

**The create modal's success toast.** Fixed, and at both of its import sites rather than only the one you named. `notifyImportResult` is lifted to `sources/utils.ts` and takes the already-translated message plus an optional success renderer, which is how the modal keeps its "Feedback records" link on the green toast while a failure gets a plain red one — your point about the link is the sharper half of the bug, since it does not just misreport, it invites the user to go and look for records that were never written. `handleCsvImport` had the identical shape three functions below `handleHistoricalImport`, and fixing one while leaving its neighbour green is exactly the divergence I gave as the reason for fixing the create path alongside the re-import path last round. Both now return `"error"` rather than `"success"` when `failures > 0`; the only thing `TImportState` gates is whether the "source created" toast fires on a *skipped* import, so nothing else in the wizard changes.

Passing the formatted message rather than `t` keeps `utils.ts` free of the two message keys and lets the CSV and historical wordings stay distinct — that was the only reason I did not lift `t` itself as you suggested.

New `describe("notifyImportResult")` in `sources/utils.test.ts`: total failure, partial failure, a clean run, and that the caller's success toast is reached only on success. Inverting the `failures > 0` branch turns all four red.

**The `import.ts` test.** You are right that the reasoning cost more than the test. Added essentially as you wrote it — I kept your fixtures and assertion and only added the comment explaining why the number is `0`. Verified it fails the way you said: reverting `expectedRecords` to `responses.length * mappings.length` gives `skipped: 1` and the test goes red.

```
cd apps/web && npx vitest run lib/feedback-source/import.test.ts
```

The PR description is updated: both rows are now `unit (mutation)` with the mutation named, and the `create-feedback-source-modal.tsx` entry is gone from `Open gaps` since it is no longer a gap.

apps/web/modules/hub/hub-client.test.ts
```diff
@@ -156,3 +163,59 @@ describe("assertRepeatedArrayParams", () => {
...
+  const response = (status: number, url: string): Response => ({ status, url }) as Response;
+
+  const buildClient = async (): Promise<{ shouldRetry: (r: Response) => Promise<boolean> }> => {
+    mutableEnv.HUB_API_KEY = "test-key";
+    globalForHub.formbricksHubClientRepeatArrays = undefined;
+    vi.mocked(FormbricksHub).mockImplementation(function () {} as never);
+
+    const { getHubClient } = await import("./hub-client");
+
+    return getHubClient() as unknown as { shouldRetry: (r: Response) => Promise<boolean> };
```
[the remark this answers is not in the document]
Dhruwang: Right, and acting on it found a real hole in my own test — thank you.

The point stands exactly as put: `hub-client.test.ts` mocks `@formbricks/hub`, so it asserted my predicate against a stand-in base, not that the SDK still routes retries through `shouldRetry`. That is the one thing the type system cannot say here, since `shouldRetry` is `private` and the override is installed on the prototype rather than declared.

Done in 7607172549 as a new `modules/hub/hub-client-retry.test.ts` that leaves the SDK real and stubs only the transport, asserting POST attempts:

| case | attempts |
| --- | --- |
| `feedbackRecords.create` → 409 | 1 |
| `feedbackRecords.create` → 429 | 3 |
| `tenants.settings.update` → 409 | 3 |
| `feedbackRecords.update` → 409 | 3 |

`retry-after-ms: 0` on the stubbed responses is honoured ahead of the SDK's exponential backoff, so the retrying cases cost no wall clock — without it the 429 case sleeps ~1.5s to prove the same thing. Reverting the guard in `hub-client.ts` turns the first row red.

**Writing it caught a bug in the test, and a fragility worth recording.** The first run had the conflicting create at 3 attempts, i.e. the narrowing not firing at all. The cause was the stub: a hand-built `Response` has an empty `url`, and the narrowing reads `response.url` to tell the create apart from the other operations sharing `/v1/feedback-records` (a `Response` does not carry the request method, so the path is the only discriminator available inside `shouldRetry`). A real fetch always populates it; my stub did not. Fixed by setting `url` from the request, which is also the more faithful stub.

The empty-`url` case is now a test of its own rather than left as an accident, so the degradation is a recorded decision: no usable URL means the SDK's own policy stands, which costs the extra retries but can never suppress one that was wanted.

I removed the `describe("conflict retries")` block from `hub-client.test.ts` and the `shouldRetry` I had added to its mock prototype. The new file covers everything it did and more, and keeping the weaker version would have left the misleading impression that the wiring was under test in two places.

xernobyl:
Rebased `7607172` onto `main` locally and smoke-tested the merged state against a stub Hub. Three of your `Open gaps` come out clean: the 3x saving measured end to end (**150 POSTs / 420ms** with the narrowing vs **450 / 1911ms** reverted), CSV/paused/unmapped visibility rendered in the UI, and the failure toast reproduced against simulated 503s (red, `0 succeeded, 150 failed`, no unhandled errors). All five mutation rows go red when reverted.

Two things I'd want addressed before merge:

1. **The feedback-record create *does* return a retryable 409.** `NewTenantWriteConflictError` at `internal/repository/feedback_records_repository.go:143` when the purge gate refuses the insert. The SDK's docstrings only document `tenant_write_conflict` under `tenants/*` and `taxonomy/*`, which is why the enumeration missed it. The behaviour is defensible; the doc comment, the description and the new test all currently assert the opposite.

2. **78 commits behind `main`, and it matters here.** #9089 deleted `checkAuthorizationUpdated` and rewrote `lib/feedback-source/actions.ts` — on `main` the action already has audit logging, a 10/hour rate limit and `assertFeedbackSourceDirectoryAccess`. So no green check on this branch has seen the authorization layer this will actually merge into.

Happy to write up either in detail — say which.

Dhruwang:
Both confirmed, both taken in 5463639, on top of a merge of `main` (d5689bc).

**1. The create's 409 is two errors, and you named the right one.** Read at the source rather than off the SDK:

- `internal/repository/feedback_records_repository.go`, `Create` — a 23505 unique violation becomes `NewConflictError` (`code: "conflict"`); `pgx.ErrNoRows` from the purge gate becomes `NewTenantWriteConflictError` (`code: "tenant_write_conflict"`).
- Hub's `openapi.yaml` documents both under this operation's 409, the second as "retryable – retry after the purge completes".
- `huberrors` draws the split precisely so "retryable lock conflicts are never confused with terminal resource conflicts". Keying on the path re-fused them.

So the path was the wrong discriminator rather than an incomplete one. `isTerminalConflict` now reads the problem `code` off a **clone** of the response — the SDK reads the original immediately after, `text()` for the APIError or `CancelReadableStream` before a retry — and treats only `conflict` as terminal. The path stays as the outer scope, since `conflict` is also raised by taxonomy runs. A code-less or unreadable body falls through to the SDK's own policy, so the degradation is extra retries, never a suppressed one.

Two new tests, both red without the code check: a `tenant_write_conflict` create still gets its 3 POSTs, and a code-less 409 falls back. Seven cases in the file now, all through the SDK's own request loop.

One thing I did **not** change: if a `tenant_write_conflict` outlives its retries, `reconcile.ts` still routes it into the "already exists" branch — a lookup, a 150 ms sleep, a second lookup, then `conflict not visible in tenant`. Wrong label, right outcome (a failure, not a silent skip), and the create-time import has carried it just as long. Say the word if you want it in this PR rather than the `reconcile.ts` follow-up.

**2. Merged rather than rebased** — `main` at cfd27e6, 80 commits, no textual conflicts (this PR never touched `lib/feedback-source/actions.ts`). On the merged tree: 720 unit tests pass across `modules/hub`, `modules/ee/unify-feedback` and `lib/feedback-source`; typecheck and lint clean. `assertFeedbackSourceDirectoryAccess` and the audit logging apply to the re-import unchanged, since it is the same action.

The merge did surface one thing worth recording: [#9089](https://github.com/formbricks/formbricks/pull/9089)'s `historicalResponseImport` limit is 10/hour per user, and the re-import calls the action once per mapped survey — so a source with more than ten mapped surveys would stop mid-run on the eleventh. Today's create dialog cannot produce one, so it is under `Risks` rather than fixed here.

No write-up needed on either, thanks.


---

xernobyl (approved):
Approving. Both points are addressed and I checked the discriminator at the source rather than off the description: `newProblem` defaults every 409 to `code: "conflict"` via `codeForStatus`, and only the tenant-write branch overrides it to `tenant_write_conflict` — and `code` is a non-`omitempty` field, so it is always on the wire. Reading it off a clone is the right call given the SDK reads the original straight after. The three-way split in the tests — duplicate, purge refusal, code-less body — is what makes the narrowing safe to leave in place.

The `main` merge closes the other one: `assertCan`, the audit logging, `assertFeedbackSourceDirectoryAccess` and the rate limit are now what this runs on, and the whole suite is green on the merged tree.

On the `tenant_write_conflict` that outlives its retries — agreed it stays out of this PR. It is pre-existing, it fails rather than silently skips, and this branch is already carrying a cross-cutting SDK change. Filed as [ENG-2800](https://linear.app/formbricks/issue/ENG-2800/feedback-import-mislabels-a-retryable-tenant-write-conflict-as-already) with the routing detail and a note that the `code` check you added here is the discriminator it should reuse rather than re-derive.

One thing for `Risks` rather than for code, whenever you next touch the description: the 3x saving now rests on Hub keeping `code: "conflict"` on the duplicate. If that ever moves, the narrowing quietly reverts to the retry storm — safe, but silent, and nothing would notice. This file already guards the same class of silent revert for query serialization with `assertRepeatedArrayParams`; there is no equivalent for this one, and I do not think there needs to be in this PR.

PR #8989, opened 2026-08-26
feat(js-core): setEmbeddedData() + clearEmbeddedData() for app surveys [ENG-1844]

pandeymangg:
> **Stacked on #8988** (ENG-1843). Diff below is this PR's own changes only.

## What & why

Today the only way a host page attaches context to an app-survey response is `track(action, { hiddenFields })`, which binds the data to the triggering action — and app surveys are multi-action, so an integrator would have to repeat the same values on every possible `track()` call. This is the Electrolux path (GTM pushes data-layer values) and the SJ/JourneyID ask.

New SDK methods, decoupled from the trigger:

```js
formbricks.setEmbeddedData({ pageType: "product", plan: "pro" }); // merge, last write wins per key
formbricks.setEmbeddedData({ pageType: null });                   // null removes the key
formbricks.clearEmbeddedData("pageType");                         // same, explicit
formbricks.clearEmbeddedData();                                   // logout / hard context switch
```

**Four semantics, not three — `undefined` is a no-op** (does not set, does not drop). That is a promise the integration doc already makes: the GTM tag reads keys off the data layer and passes every field unconditionally, so a key absent on the current page type arrives as `undefined` and must not clear the value a previous page set. `undefined` and `null` are one keystroke apart and behave oppositely; both are pinned by tests.

**The bag is in-memory, page-load scoped, never persisted.** Deliberately not `Config` — that class writes to localStorage, and persisting the bag would blur the Embedded Data ↔ contact-attribute boundary and create a PII-at-rest surface. Backed by a `Map` so a `__proto__` key is data, not a prototype write.

**Synchronous, not queued — the `setNonce` precedent, and a deliberate departure from every other method.** The command queue would do two wrong things here: `checkSetup` silently *drops* commands issued before `setup()` completes (the exact GTM failure mode the `formbricks_setup_successful` readiness event exists to prevent — ENG-1846, next in this stack), and `CommandType.GeneralAction` drains the user update queue (network) on every call, while the ticket promises that calling this on every SPA route change is free. A memory write needs neither ordering nor setup.

**js-core does not filter — the SDK stays a dumb pipe** (ENG-1845/2472). `renderWidget` merges the bag *under* the per-trigger `track({ hiddenFields })` values (explicit beats ambient on a shared key) and hands the result to the renderer, whose ingest contract drops unknown/non-ingested/locked keys with a logged reason and coerces the rest; the server re-runs all of it on ingest. Snapshot-at-display: the merge happens inside the delay timeout at the moment the survey shows, from a detached copy — a later `setEmbeddedData` affects the next response, never the one on screen (the renderer's `useState` initializer freezes its side too).

`track({ hiddenFields })` keeps working unchanged (ENG-1838).

## Linear ticket

https://linear.app/formbricks/issue/ENG-1844/sdk-setembeddeddata-clearembeddeddata-app-surveys

## How this was tested

- `@formbricks/js-core` suite ✅ — 21 files, 293 tests
- `turbo run typecheck lint --filter=@formbricks/js-core` ✅ 4/4 · prettier ✅
- Tests added, each checked to **fail with its fix removed**:
  - `embedded-data.test.ts` — merge keeps other keys; `null` drops; **`undefined` no-op asserted separately from `null`** (red when `undefined` is treated like `null`); single clear; full clear; last-write-wins; snapshot detachment; `__proto__` stored as data with `Object.prototype` unpolluted; localStorage untouched and zero `fetch` calls across every method.
  - `widget.test.ts` — bag merges under per-trigger hidden fields (`plan` from `track` beats `plan` from the bag), and a `setEmbeddedData` after render does not reach the already-passed record (red with the merge removed).
- The existing ENG-2472 pass-through test (raw `track` bag reaches the renderer verbatim) still passes, pinning that this PR added no SDK-side filtering.

## Breaking changes

None — two new methods, nothing existing changes shape or behavior.

## QA / Test Plan

**How to test**

- [ ] App survey with an ingested Embedded Data field `pageType`, triggered by action `evt`. On the host page: `formbricks.setEmbeddedData({ pageType: "product" })`, then `formbricks.track("evt")` → complete the survey → the response shows `pageType = product`.
- [ ] Same, but call `formbricks.setEmbeddedData({ pageType: "changed" })` **while the survey is on screen**, then finish it → the response still shows `product`. Trigger the survey again → the new response shows `changed`.
- [ ] `setEmbeddedData({ plan: "pro" })` then `setEmbeddedData({ pageType: "x" })` → both keys land on the next response (merge, not replace).
- [ ] `setEmbeddedData({ plan: null })` then trigger → `plan` absent from the response.
- [ ] `setEmbeddedData({ plan: undefined, pageType: "y" })` after `plan` was set → `plan` **still lands**; only setting `null` removes it.
- [ ] `clearEmbeddedData()` then trigger → no bag keys on the response.
- [ ] `formbricks.track("evt", { hiddenFields: { pageType: "from-track" } })` with `pageType` also in the bag → the response shows `from-track` (explicit per-trigger value wins).
- [ ] Send an undeclared key (`setEmbeddedData({ rogue: "x" })`) → never stored; the browser console explains the drop.
- [ ] Reload the page without re-pushing → the next response has no bag values (nothing persisted; check devtools → Application → Local Storage: no new keys).
- [ ] Network tab: `setEmbeddedData`/`clearEmbeddedData` calls themselves fire zero requests.

**Preconditions / test data**

- An app survey with at least one ingested Embedded Data field and a code action trigger; the JS SDK embedded on a test page.
- Works before `setup()` completes (value is kept, not dropped) — but the reliable pattern for consent-gated setups is triggering on `formbricks_setup_successful`, which lands with ENG-1846 in this stack.

**Risks & regressions**

- The bag rides the same `hiddenFieldsRecord` slot as `track({ hiddenFields })` — re-check that a survey using only legacy `track` hidden fields behaves identically (pinned by the untouched pass-through test).
- Values set via the bag are subject to the same ingest contract as everything else: locked fields ignore them, wrong-typed values store raw + flagged. Nothing here writes to the contact record.

- No DB/env changes. **Release sequencing (from review):** the published `@formbricks/js` npm wrapper (5.0.0, used by `apps/web` and the framework guides) forwards a hard-coded method list — `setEmbeddedData`/`clearEmbeddedData` are `undefined` on that path until a companion wrapper release ships. That release must ride with this stack's release.

### DIFF
--- a/packages/js-core/src/index.ts
+++ b/packages/js-core/src/index.ts
@@ -98,11 +98,12 @@ const setEmbeddedData = (data: TEmbeddedDataInput): void => {
 };
 
 /**
- * Remove one Embedded Data key, or clear the whole bag when called without one — logout, or a hard
- * context switch. Synchronous, no network.
+ * Remove one Embedded Data key, or clear the whole bag when called with no argument — logout, or a
+ * hard context switch. Synchronous, no network. A key that evaluated to `undefined` is a no-op, not
+ * a full clear: the arity is forwarded, so only a literal zero-argument call wipes everything.
  */
-const clearEmbeddedData = (key?: string): void => {
-  EmbeddedDataStore.getInstance().clearEmbeddedData(key);
+const clearEmbeddedData = (...args: [] | [key: string]): void => {
+  EmbeddedDataStore.getInstance().clearEmbeddedData(...args);
 };
 
 /**
--- a/packages/js-core/src/index.ts
+++ b/packages/js-core/src/index.ts
@@ -2,6 +2,7 @@ import { CommandQueue, CommandType } from "@/lib/common/command-queue";
 import * as Setup from "@/lib/common/setup";
 import { getIsDebug } from "@/lib/common/utils";
 import * as Action from "@/lib/survey/action";
+import { EmbeddedDataStore, type TEmbeddedDataInput } from "@/lib/survey/embedded-data";
 import { checkPageUrl } from "@/lib/survey/no-code-action";
 import * as Attribute from "@/lib/user/attribute";
 import * as User from "@/lib/user/user";
@@ -82,6 +83,29 @@ const registerRouteChange = async (): Promise<void> => {
 };
 
 /**
+ * Attach Embedded Data to future responses without tying it to a trigger (ENG-1844). Merges into the
+ * in-memory bag, last write wins per key; `{ key: null }` removes a key and `undefined` values are
+ * skipped. Values land only on the survey's declared *ingested* fields — anything else is dropped
+ * and logged by the renderer, never fatal.
+ *
+ * Synchronous and network-free on purpose (like `setNonce`, unlike the queued methods): calling it
+ * on every SPA route change is free, and routing it through the command queue would silently drop
+ * calls made before `setup()` completes — the exact failure the `formbricks_setup_successful`
+ * readiness event exists to prevent (ENG-1846).
+ */
+const setEmbeddedData = (data: TEmbeddedDataInput): void => {
+  EmbeddedDataStore.getInstance().setEmbeddedData(data);
+};
+
+/**
+ * Remove one Embedded Data key, or clear the whole bag when called without one — logout, or a hard
+ * context switch. Synchronous, no network.
+ */
+const clearEmbeddedData = (key?: string): void => {
+  EmbeddedDataStore.getInstance().clearEmbeddedData(key);
+};
+
+/**
  * Set the CSP nonce for inline styles
  * @param nonce - The CSP nonce value (without 'nonce-' prefix), or undefined to clear
  */
@@ -107,6 +131,8 @@ const formbricks = {
   logout,
   registerRouteChange,
   setNonce,
+  setEmbeddedData,
+  clearEmbeddedData,
 };
 
 // Explicitly assign to globalThis so the wrapper SDK (@formbricks/js) can
--- a/packages/js-core/src/lib/survey/embedded-data.ts
+++ b/packages/js-core/src/lib/survey/embedded-data.ts
@@ -1,3 +1,6 @@
+import { Logger } from "@/lib/common/logger";
+import { type TTrackProperties } from "@/types/survey";
+
 /** What a host page may hand to `setEmbeddedData`. `null` removes the key; `undefined` is a no-op. */
 export type TEmbeddedDataInput = Record<string, string | number | boolean | Date | null | undefined>;
 
@@ -43,6 +46,18 @@ export class EmbeddedDataStore {
    * type arrives as `undefined` and must not clear the value a previous page set.
    */
   public setEmbeddedData(data: TEmbeddedDataInput): void {
+    // Guarded rather than thrown: this is the one SDK entry point outside the command queue's
+    // `wrapThrowsAsync` shield (it is deliberately synchronous), and the GTM pattern this feature
+    // targets can hand over an absent object — `setEmbeddedData(dataLayerObj)` on a page type where
+    // that object is undefined. A broken tag is a worse failure than a skipped write. A primitive is
+    // refused too: `Object.entries("plan")` would spread into junk keys ({0: "p", 1: "l", …}).
+    if (typeof data !== "object" || data === null) {
+      Logger.getInstance().error(
+        `setEmbeddedData: expected an object, got ${data === null ? "null" : typeof data} — nothing was set`
+      );
+      return;
+    }
+
     for (const [key, value] of Object.entries(data)) {
       if (value === undefined) continue;
       if (value === null) {
@@ -53,12 +68,29 @@ export class EmbeddedDataStore {
     }
   }
 
-  /** Remove one key, or everything when called without one (logout / hard context switch). */
-  public clearEmbeddedData(key?: string): void {
-    if (key === undefined) {
+  /**
+   * Remove one key, or everything when called with no argument (logout / hard context switch).
+   *
+   * "No argument" and "an argument that evaluated to `undefined`" are deliberately different:
+   * `setEmbeddedData` treats `undefined` as a no-op because GTM reads values off the data layer
+   * unconditionally, and this method reads from the same dynamic source — so
+   * `clearEmbeddedData(dataLayer.fieldToClear)` with the key absent must skip, not wipe the whole
+   * bag. Only a literal zero-argument call clears everything.
+   */
+  public clearEmbeddedData(...args: [] | [key: string]): void {
+    if (args.length === 0) {
       this.data.clear();
       return;
     }
+
+    const [key] = args;
+    if (typeof key !== "string") {
+      Logger.getInstance().error(
+        "clearEmbeddedData: expected a field name — nothing was cleared (call with no argument to clear everything)"
+      );
+      return;
+    }
+
     this.data.delete(key);
   }
 
@@ -71,3 +103,27 @@ export class EmbeddedDataStore {
     return Object.fromEntries(this.data);
   }
 }
+
+/**
+ * The display-time merge: the ambient bag under the explicit per-trigger `track({ hiddenFields })`
+ * values, so an explicit write wins a shared key.
+ *
+ * Bag keys are folded case-INsensitively against the explicit ones, because the ingest contract
+ * matches declared names case-insensitively with exact-match-first: with a field declared `Plan`, a
+ * bag entry `Plan` and a track entry `plan` would otherwise both survive a plain spread, and the
+ * contract would then pick the bag's exact match — the ambient value silently beating the explicit
+ * one. Folding here keeps "explicit beats ambient" true under every casing.
+ *
+ * The cast at the end: the renderer's contract accepts boolean/Date scalars the narrower legacy
+ * `hiddenFields` type cannot spell, and normalizes them before anything is stored.
+ */
+export const buildDisplayHiddenFields = (
+  explicit?: TTrackProperties["hiddenFields"]
+): TTrackProperties["hiddenFields"] => {
+  const explicitKeysFolded = new Set(Object.keys(explicit ?? {}).map((key) => key.toLowerCase()));
+  const ambient = Object.entries(EmbeddedDataStore.getInstance().getSnapshot()).filter(
+    ([key]) => !explicitKeysFolded.has(key.toLowerCase())
+  );
+
+  return { ...Object.fromEntries(ambient), ...explicit } as TTrackProperties["hiddenFields"];
+};
--- a/packages/js-core/src/lib/survey/embedded-data.ts
+++ b/packages/js-core/src/lib/survey/embedded-data.ts
@@ -50,8 +50,9 @@ export class EmbeddedDataStore {
     // `wrapThrowsAsync` shield (it is deliberately synchronous), and the GTM pattern this feature
     // targets can hand over an absent object — `setEmbeddedData(dataLayerObj)` on a page type where
     // that object is undefined. A broken tag is a worse failure than a skipped write. A primitive is
-    // refused too: `Object.entries("plan")` would spread into junk keys ({0: "p", 1: "l", …}).
-    if (typeof data !== "object" || data === null) {
+    // refused too, and so is an array (`typeof [] === "object"`): either would spread into junk
+    // numeric keys ({0: "p", 1: "l", …} / {0: "a", 1: "b"}) — `ecommerce.items` is the common array case.
+    if (typeof data !== "object" || data === null || Array.isArray(data)) {
       Logger.getInstance().error(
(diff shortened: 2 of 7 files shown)


pandeymangg:
Self-review, raised while thinking through the lifetime implications: `formbricks.logout()` does **not** clear the bag. On a shared device, if the host forgets `clearEmbeddedData()` on a user switch, the previous user's context — including `hashed_email`-class values — rides onto the next user's response within the same page load. The ticket makes clearing host-driven, but auto-clearing on logout is strictly safer and costs a legitimate host nothing (values can be re-pushed after logout). Adding that to this branch.

pandeymangg:
Self-review #2, DX gap flagged for the record: the bag is invisible — it lives in memory (nothing in DevTools → Local Storage) and the API has set/clear but no read. An integrator wiring GTM can't verify what's currently in the bag; the first observable point is the rendered survey's console output. Deliberately **not** adding a public `getEmbeddedData()` here (new public API surface is a product call, and a getter invites using the bag as a state store). Flagging instead for ENG-1849's docs, plus a possible debug-mode log (`formbricksDebug=true`) as a cheap follow-up.

packages/js-core/src/lib/survey/embedded-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Merge — never replace — so refreshing a volatile field (`pageType`) cannot wipe the stable ones
+   * (`plan`) set at page load. Per key: last write wins; `null` removes; `undefined` does nothing.
+   *
+   * The `undefined` no-op is a documented promise, not an accident: the GTM integration reads keys
+   * off the data layer and passes every field unconditionally, so a key absent on the current page
+   * type arrives as `undefined` and must not clear the value a previous page set.
+   */
+  public setEmbeddedData(data: TEmbeddedDataInput): void {
+    for (const [key, value] of Object.entries(data)) {
```
kiln-agent, line 62: `Object.entries(data)` throws a `TypeError` on `null` or `undefined`. Because this method is deliberately synchronous, the throw escapes straight into host page code — every other SDK method is shielded behind the command queue's `wrapThrowsAsync`, so this is the one entry point that can break a customer's page.

That matters for the exact integration this feature targets: a GTM tag reading off the data layer can hand over an absent object (`formbricks.setEmbeddedData(dataLayerObj)` where `dataLayerObj` is undefined on that page type), and a broken tag is a worse failure than a skipped write. A primitive argument is also mis-ingested rather than rejected — `setEmbeddedData("plan")` spreads to `{0: "p", 1: "l", ...}` and stores four junk keys.

Suggest guarding at the top and logging rather than throwing, in keeping with the "never fatal" promise in the public docstring:

```ts
if (typeof data !== "object" || data === null) {
  Logger.getInstance().error("setEmbeddedData: expected an object");
  return;
}
```

---

packages/js-core/src/lib/survey/embedded-data.ts
```diff
@@ -0,0 +1,73 @@
...
+      if (value === null) {
+        this.data.delete(key);
+        continue;
+      }
+      this.data.set(key, value);
+    }
+  }
+
+  /** Remove one key, or everything when called without one (logout / hard context switch). */
+  public clearEmbeddedData(key?: string): void {
```
kiln-agent: Worth a second look at the `undefined` branch here, given how hard `setEmbeddedData` argues the opposite case.

`setEmbeddedData` treats `undefined` as a no-op precisely because the GTM tag passes data-layer values unconditionally and a missing key must not clear anything. `clearEmbeddedData` reads its values from that same dynamic source, but here `undefined` means "wipe everything" — so `formbricks.clearEmbeddedData(dataLayer.fieldToClear)` with the key absent nukes all ambient context instead of doing nothing. Same source, same value, opposite blast radius.

The no-arg overload itself is fine and clearly useful; it's the collapsing of "no argument" and "an argument that evaluated to undefined" that's the footgun. `clearEmbeddedData(...args: [] | [string])` and branching on `args.length` would separate them, and would be worth a test alongside the existing `undefined`-vs-`null` pair. Not blocking if you'd rather document it, but the asymmetry will bite someone.

---

packages/js-core/src/index.ts
```diff
@@ -107,6 +131,8 @@ const formbricks = {
   logout,
   registerRouteChange,
   setNonce,
+  setEmbeddedData,
```
kiln-agent, line 135: The store's own docstring names logout as the reason `clearEmbeddedData()` exists ("logout / hard context switch"), but nothing wires it up — so today the host page has to remember to call it.

`logout()` and `setUserId()` with a changed id both go through `User.logout` / `User.setUserId`, which call `tearDown()` (`packages/js-core/src/lib/common/setup.ts:342`). That resets user state and closes the open survey, but the new singleton is untouched — so on a shared machine, or any app where one page load spans a logout and a re-login, user A's ambient values ride onto user B's responses. That's the same PII boundary the "never persisted" decision was protecting, just leaking through a different door.

One line in `tearDown()` closes it:

```ts
EmbeddedDataStore.getInstance().clearEmbeddedData();
```

Worth a test that a `logout()` drops the bag — it's cheap to pin and easy to regress.

---

packages/js-core/src/lib/survey/widget.ts
```diff
@@ -171,7 +172,15 @@ export const renderWidget = async (
...
       styling: getStyling(settings, survey),
-      hiddenFieldsRecord: hiddenFieldsObject,
+      // The ambient Embedded Data bag (ENG-1844) under the per-trigger `track({ hiddenFields })`
+      // values, so an explicit per-trigger write wins on a shared key. Snapshotted here — inside the
+      // delay timeout, at the moment the survey actually shows — and `getSnapshot` returns a copy, so
+      // a later `setEmbeddedData` affects the next response, never this one. The renderer's ingest
+      // contract accepts boolean/Date scalars the narrower legacy type cannot spell, hence the cast.
+      hiddenFieldsRecord: {
+        ...EmbeddedDataStore.getInstance().getSnapshot(),
+        ...hiddenFieldsObject,
```
kiln-agent: The "explicit per-trigger value beats the ambient bag" rule holds only when the two keys match exactly, because the spread dedupes case-sensitively while the ingest contract does not.

`matchDeclaredFieldName` (`packages/types/safe-identifier.ts:130`) resolves declared-side: exact match first, then case-insensitive. So with a field declared as `Plan`, a bag entry `{ Plan: "ambient" }` and `track("evt", { hiddenFields: { plan: "explicit" } })`, the spread produces `{ Plan: "ambient", plan: "explicit" }` — two distinct keys, neither overwriting the other — and the contract then picks `Plan` as the exact match. The ambient value wins, which is the reverse of what the PR description and the comment right above promise.

Case tolerance is load-bearing on the ingest side (the comment at `embedded-data-ingest.ts:421` explains why surveys depend on it), so the fix belongs here: fold the bag keys against the incoming `hiddenFieldsObject` case-insensitively before spreading, dropping any bag key that a `track` key already covers under case-folding.

Narrow trigger, but silent when it hits — the wrong value lands on the response with nothing logged.

---

kiln-agent (commented):
Reviewed the SDK-side changes against the base branch — the new store, the `renderWidget` merge, the ingest contract it feeds (`packages/types/embedded-data-ingest.ts`, `safe-identifier.ts`), the command-queue error handling, and `tearDown`/`logout`.

The core mechanics hold up, and the design rationale in the docstrings is unusually good. Specifically verified:

- The `Map` really does carry a `__proto__` key as data all the way through `Object.fromEntries` and the object spread in `widget.ts` — `Object.prototype` stays clean.
- The snapshot is genuinely detached, and taking it inside the delay timeout gives the snapshot-at-display semantics the description claims.
- `undefined` vs `null` in `setEmbeddedData` is correct and separately pinned by tests, which is the right call given they're one keystroke apart.
- Boolean and `Date` values are normalized by the renderer before anything is stored, so the cast at `widget.ts:182` isn't lying at runtime.

Four things worth addressing, inline. Two look like real bugs:

1. **`setEmbeddedData(null)` throws into host page code** — the one SDK entry point not shielded by `wrapThrowsAsync`, and the GTM data-layer pattern this feature targets can produce exactly that argument.
2. **The bag survives `logout()` / a `setUserId` switch** — `tearDown()` doesn't touch the singleton, so user A's ambient values can land on user B's responses within one page load. The store's docstring names logout as the reason `clearEmbeddedData()` exists, but nothing calls it.

Plus a narrow correctness gap where the case-insensitive declared-field matching makes the ambient bag beat an explicit `track({ hiddenFields })` value, and a design question about `clearEmbeddedData(undefined)` wiping the whole bag.

One thing outside this diff, for release sequencing rather than this PR: the published `@formbricks/js` wrapper (pinned to `5.0.0`, used by `apps/web` and by the framework guides) forwards a hard-coded method list — I checked the shipped `dist`, and it has `setNonce` and `registerRouteChange` but no `setEmbeddedData`/`clearEmbeddedData`. Both methods will be `undefined` for anyone on that path until a companion wrapper release goes out, and its `TFormbricks` type won't have them either. Worth confirming that release is part of the plan for this stack.

---

packages/js-core/src/lib/survey/embedded-data.ts
pandeymangg: Taken in 947c038bb — type guard at the top, logs and returns instead of throwing; primitives refused too (no junk keys). Pinned: null/undefined no-throw + the "plan"-spread case.

packages/js-core/src/lib/survey/embedded-data.ts
pandeymangg: Taken — the signature is now `(...args: [] | [key: string])` end to end (the index.ts wrapper forwards arity), so only a literal zero-argument call clears everything; an explicit `undefined` logs and no-ops. Pinned by a test right next to the null/undefined pair.

packages/js-core/src/index.ts
pandeymangg: Taken, with one deliberate deviation from the suggested placement: the clear sits in `logout()` and `setUserId`'s different-user branch rather than in `tearDown()` — tearDown's third caller is the setup-*error* path, which isn't an identity switch, and page context should survive a setup retry. First-time identification keeps the bag (hosts legitimately push context before identifying). Tests for all three states.

packages/js-core/src/lib/survey/widget.ts
pandeymangg: Taken — the merge now goes through `buildDisplayHiddenFields`, which folds bag keys case-insensitively against the track keys before spreading. Your `Plan`/`plan` scenario is the pinned test, and removing the fold flips it red.

packages/js-core/src/lib/survey/embedded-data.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export const buildDisplayHiddenFields = (
+  explicit?: TTrackProperties["hiddenFields"]
+): TTrackProperties["hiddenFields"] => {
+  const explicitKeysFolded = new Set(Object.keys(explicit ?? {}).map((key) => key.toLowerCase()));
+  const ambient = Object.entries(EmbeddedDataStore.getInstance().getSnapshot()).filter(
+    ([key]) => !explicitKeysFolded.has(key.toLowerCase())
+  );
+
+  return { ...Object.fromEntries(ambient), ...explicit } as TTrackProperties["hiddenFields"];
```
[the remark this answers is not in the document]
pandeymangg: Taken — `buildDisplayHiddenFields` now strips undefined-valued entries from `explicit` before both the fold set and the spread, so `track("evt", { hiddenFields: { plan: dataLayer.plan } })` with the key absent leaves the ambient value untouched — same promise as `setEmbeddedData`'s undefined no-op. Pinned by a test; reverting the strip flips it red.

packages/js-core/src/lib/survey/embedded-data.ts
```diff
@@ -0,0 +1,129 @@
...
+   * off the data layer and passes every field unconditionally, so a key absent on the current page
+   * type arrives as `undefined` and must not clear the value a previous page set.
+   */
+  public setEmbeddedData(data: TEmbeddedDataInput): void {
+    // Guarded rather than thrown: this is the one SDK entry point outside the command queue's
+    // `wrapThrowsAsync` shield (it is deliberately synchronous), and the GTM pattern this feature
+    // targets can hand over an absent object — `setEmbeddedData(dataLayerObj)` on a page type where
+    // that object is undefined. A broken tag is a worse failure than a skipped write. A primitive is
+    // refused too: `Object.entries("plan")` would spread into junk keys ({0: "p", 1: "l", …}).
+    if (typeof data !== "object" || data === null) {
```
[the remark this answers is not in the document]
pandeymangg: Taken — `|| Array.isArray(data)` added to the guard, with an `ecommerce.items`-style test right next to the primitive one. Reverting the guard flips it red.

itsjavi (approved):
verified and LGTM

PR #8990, opened 2026-08-26
feat(js-core): SDK event bus — emit lifecycle events to the host page [ENG-1846]

pandeymangg:
> **Stacked on #8989** (ENG-1844), which sits on #8988 (ENG-1843). Diff below is this PR's own changes only.

## What & why

The SDK is a black box to the host page: nothing announces that setup finished, that a survey showed, or that a response landed. That forces the two integration hacks this PR retires — polling `if (window.formbricks)` in a 100 ms interval (our own GTM doc ships that pattern today), and the `postMessage` listener customers run on confirmation pages. It also makes `setEmbeddedData` (#8989) unreliable under consent-gated setup: a GTM tag firing at page load hits an undefined global and the value silently vanishes.

Four lifecycle events, a general bus rather than a single hook:

| Event | Fires | Payload |
| --- | --- | --- |
| `formbricks_setup_successful` | fresh `setup()` completes | `{ workspaceId }` |
| `formbricks_action_tracked` | any code/no-code action, hit or miss | `{ action }` |
| `formbricks_survey_shown` | display created | `{ surveyId }` |
| `formbricks_response_submitted` | response acked (`finished: false`) and finished (`finished: true`) | `{ surveyId, responseId, finished }` |

The names are a **committed external contract**: the data-mapping doc sent to Electrolux publishes all four, and their UAT already has a GTM trigger on `formbricks_setup_successful`. Constants hold the full literals so a customer's trigger string greps to one place.

**Two transports per event.** A `CustomEvent` on `window` — always; the in-house pattern (`formbricks:onFilePick`) and vendor-neutral. Plus a dataLayer push via the standard idiom `window.dataLayer = window.dataLayer || []` — the idiom survives load order (GTM drains the pre-created array on init), where pushing only-if-present would silently drop every event emitted before GTM loads. Cost, accepted: a GTM-less site accumulates small event objects nobody drains.

**The dataLayer payload is nested** — `{ event, formbricks: { … } }` — not spread flat. GTM's Data Layer Variables read a *merged* model, so a flat `action` or `finished` persists across pushes and collides with the host's own keys (`action` is everywhere in ecommerce dataLayers). This is a deliberate deviation from the payload *sketch* in the customer doc, which marks payloads as still open; Sean needs a heads-up before he builds the `response_submitted` tag (his existing `setup_successful` trigger matches on the event name only and is unaffected).

**Exactly-once for the readiness event is structural, not flag-based:** the emit sits at the single tail every *fresh* setup converges on in `setup.ts`; the already-set-up and missing-config early returns never reach it, so a repeated `setup()` cannot re-fire the host's tags.

**The one contract change: `responseId` now reaches the host.** `onResponseCreated` fired at enqueue time, before the server ever minted an id. It now fires (still exactly once, same guard) from the ResponseQueue's creation ack, widened to `(responseId?: string)` — same post-ack semantics `onDisplayCreated` already has, and the only way `response_submitted` can carry the persisted id that links a ContentSquare replay. `onFinished` is widened the same way (it is gated on the send finishing, so the id exists by then). Both widenings are additive; preview mode keeps firing at submit time with no id, since nothing is stored. This touches `packages/types` + `packages/surveys`, so the surveys bundle needs the rebuild dance when testing manually.

**App surveys only.** Link surveys never load js-core — they render through the web app. No events fire there; a separate ticket if ever wanted.

**No PII:** ids and the `finished` boolean, never response content or scores — per the ticket's consent note and ENG-1842's Anonymize direction.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1846/sdk-event-bus-emit-lifecycle-events-to-the-host-page

## How this was tested

- `@formbricks/js-core` ✅ 22 files / 302 tests · `@formbricks/surveys` ✅ 697 · full `pnpm test` ✅ 25/25 tasks
- **E2E (added in review): `js.spec.ts` now pins that `formbricks_response_submitted` carries the
  PERSISTED `responseId`**, asserted against the stored `Response` row. This closes the one gap a
  reviewer found: reverting `onFinished?.(surveyState?.responseId ?? undefined)` to `onFinished?.()`
  previously left all 701 `packages/surveys` tests green, because the producer lives in a `.tsx` (not
  unit-tested here) and js-core's `widget.test.ts` feeds `"resp_123"` to a callback it mocks itself —
  it exercises the consumer, never the renderer that supplies the id. Added as a `test.step` to the
  existing SDK journey rather than a new spec: the survey is already created, displayed and completed
  there, so it costs one `page.evaluate` and one query, not a browser. Checked to go red with the fix
  reverted.
- `turbo run typecheck lint` ✅ for js-core, surveys, types **and** `@formbricks/web` (the widened callbacks are consumed by `SurveyInline`/preview — zero-arg handlers remain assignable) · prettier ✅
- Tests added, the two structural ones checked to **fail with their fix removed**:
  - `events.test.ts` — CustomEvent type + `detail`; dataLayer created when absent; **a pre-existing dataLayer is appended to, never replaced** (a seeded host entry survives — the failure that would wipe a customer's queued events); the nested envelope; every name starts with `formbricks_`; both transports carry the same payload.
  - `setup.test.ts` — fresh setup emits exactly once with the resolved `workspaceId` (red with the emit removed); the already-set-up early return does **not** re-emit; a failed setup emits nothing.
  - `action.test.ts` — emits for a tracked action even when no survey matches (the emit sits above the matching loop).
  - `widget.test.ts` — `survey_shown` on display; `response_submitted` with the acked `responseId`, `finished: false` then `true` (red with the emit removed).
- Existing suites pin the regressions that matter: all 697 surveys tests pass with the `onResponseCreated` fire point moved, and the js-core segment-refresh tests still drive the same callbacks.

## Breaking changes

None — new events only; `onResponseCreated`/`onFinished` widen with an optional parameter, so every existing implementation keeps compiling and behaving. The one observable shift: `onResponseCreated` now fires on the server ack rather than at enqueue (milliseconds later on a healthy network), matching `onDisplayCreated`'s existing post-ack semantics.

## QA / Test Plan

**How to test**

- [ ] Page with the JS SDK and this snippet before `setup()`: `window.addEventListener("formbricks_setup_successful", (e) => console.log(e.detail))` → after `setup()` completes, the console logs `{ workspaceId }`, and `window.dataLayer` contains `{ event: "formbricks_setup_successful", formbricks: { workspaceId } }`.
- [ ] Call `formbricks.setup()` a second time → **no** second event (check `window.dataLayer` length).
- [ ] `formbricks.track("some-action")` for an action that triggers no survey → `formbricks_action_tracked` still fires with `{ action }`.
- [ ] Trigger an app survey → `formbricks_survey_shown` with `{ surveyId }` when it appears.
- [ ] Answer the first question → `formbricks_response_submitted` with `{ surveyId, responseId, finished: false }`, where `responseId` matches the response later visible in the app.
- [ ] Finish the survey → the same event again with `finished: true` and the same `responseId`.
- [ ] GTM readiness handshake (the Electrolux pattern): delay `setup()` behind a consent prompt, put a GTM Custom Event trigger on `formbricks_setup_successful` whose tag calls `formbricks.setEmbeddedData({...})` → the values land on the next response. This is the integration the event exists for.
- [ ] Pre-seed `window.dataLayer = [{ event: "host" }]` before the SDK loads → after events fire, the host entry is still index 0.
- [ ] No GTM on the page at all → everything above still works via `addEventListener`; `window.dataLayer` exists and grows only with our events.

**Preconditions / test data**

- An app survey with a code-action trigger; the JS SDK on a test page; GTM (or Tag Assistant preview) for the handshake step.
- Link surveys intentionally emit nothing — verify no `formbricks_*` events on a link survey page.

**Also in this PR (review follow-up)**

- `warnOnMissingIngestRows` moved out of the `useMemo` body in `survey-client-wrapper.tsx` into a
  `useEffect` keyed on the rows and the legacy column. A memo body is not a defined place for a side
  effect, and its deps included `searchParams`, which has nothing to do with what the diagnostic
  reports. Scope stated honestly: this does **not** reduce the duplicate line in dev — measured before
  and after, it prints twice per mount either way, because StrictMode double-invokes effects just as
  it does memo bodies. Hygiene and correct dependencies, not a fix for the duplication.

**Risks & regressions**

- `onResponseCreated` timing moved from enqueue to server ack. js-core's own bookkeeping (responses array, segment refresh) now runs on ack — on a healthy network milliseconds later; if the create request ultimately fails, the bookkeeping (correctly) never runs for a response that was never stored. Re-check "have responded"-style segment targeting right after a first answer.
- Mobile SDKs (ENG-2472) read the same renderer contract — the widened callbacks are the shape they should adopt.
- The `postMessage("formbricksSurveyCompleted")` hack is deliberately left in place; removing it is a separate deprecation once integrators migrate.

- none. Docs for the events land with ENG-1849 (the existing GTM doc's polling example should be replaced by the readiness event).

### DIFF
--- a/packages/js-core/src/lib/common/events.ts
+++ b/packages/js-core/src/lib/common/events.ts
@@ -72,7 +72,14 @@ export const emitFormbricksEvent = (event: TFormbricksEventName, payload: TFormb
     // `.push`. GTM's own snippet only ever creates an array, and its loaded state keeps the array
     // and swaps the `push` method, so `Array.isArray` stays true on every real GTM page.
     if (!Array.isArray(window.dataLayer)) window.dataLayer = [];
-    window.dataLayer.push({ event, formbricks: { ...EMPTY_DATALAYER_PAYLOAD, ...payload } });
+    // Undefined-valued keys are stripped before the merge: `TFormbricksEventPayload` admits
+    // `undefined` (the widened callbacks type `responseId` optional), and a key PRESENT with value
+    // `undefined` would replace the `null` sentinel in the spread — re-opening the recursive-merge
+    // bleed the sentinel exists to stop.
+    const definedPayload = Object.fromEntries(
+      Object.entries(payload).filter(([, value]) => value !== undefined)
+    );
+    window.dataLayer.push({ event, formbricks: { ...EMPTY_DATALAYER_PAYLOAD, ...definedPayload } });
   } catch (error) {
     console.error(`Formbricks: failed to push "${event}" to the dataLayer`, error);
   }
--- a/packages/js-core/src/lib/common/events.ts
+++ b/packages/js-core/src/lib/common/events.ts
@@ -35,12 +35,45 @@ export type TFormbricksEventName = (typeof FORMBRICKS_EVENTS)[keyof typeof FORMB
 
 export type TFormbricksEventPayload = Record<string, string | number | boolean | undefined>;
 
+/**
+ * Every key the event contract can carry, `null` where an event does not set it. The dataLayer push
+ * always carries the FULL set because GTM's data model merges pushes recursively — a partial push
+ * would let a previous event's `formbricks.responseId` or `formbricks.finished` bleed into a later
+ * event's reads (survey A's `responseId` resolving under survey B's `formbricks_survey_shown`).
+ * Pushing `null` overwrites the merged value; omitting the key would not. `CustomEvent.detail` has
+ * no merge semantics, so it carries only the event's own keys.
+ */
+const EMPTY_DATALAYER_PAYLOAD: Record<string, null> = {
+  workspaceId: null,
+  surveyId: null,
+  responseId: null,
+  finished: null,
+  action: null,
+};
+
 export const emitFormbricksEvent = (event: TFormbricksEventName, payload: TFormbricksEventPayload): void => {
   // js-core is imported by SSR bundles; emitting is meaningless (and crashes) off the browser.
   if (globalThis.window === undefined) return;
 
-  window.dispatchEvent(new CustomEvent<TFormbricksEventPayload>(event, { detail: payload }));
+  // Both transports run host-owned code — the page's event listeners, and `dataLayer.push`, which
+  // GTM replaces with its own function once it loads. The emitter's call sites sit at the head of
+  // SDK-critical paths (display bookkeeping, the response queue's ack), so a host-page throw here
+  // must never escape: it would cost us display state or mark a persisted response as failed. Each
+  // transport is isolated separately so one failing cannot silence the other.
+  try {
+    window.dispatchEvent(new CustomEvent<TFormbricksEventPayload>(event, { detail: payload }));
+  } catch (error) {
+    console.error(`Formbricks: a "${event}" event listener threw`, error);
+  }
 
-  window.dataLayer = window.dataLayer ?? [];
-  window.dataLayer.push({ event, formbricks: payload });
+  try {
+    // `Array.isArray`, not `?? []`: a host page that set `window.dataLayer = {}` (a hand-rolled
+    // shim, or another vendor reusing the name) survives nullish coalescing and then throws on
+    // `.push`. GTM's own snippet only ever creates an array, and its loaded state keeps the array
+    // and swaps the `push` method, so `Array.isArray` stays true on every real GTM page.
+    if (!Array.isArray(window.dataLayer)) window.dataLayer = [];
+    window.dataLayer.push({ event, formbricks: { ...EMPTY_DATALAYER_PAYLOAD, ...payload } });
+  } catch (error) {
+    console.error(`Formbricks: failed to push "${event}" to the dataLayer`, error);
+  }
 };
--- a/packages/js-core/src/lib/common/events.ts
+++ b/packages/js-core/src/lib/common/events.ts
@@ -1,0 +1,46 @@
+/**
+ * The SDK's outbound event bus (ENG-1846): lifecycle signals the host page can react to — fire
+ * GA/GTM tags, link session replays, and (most importantly) know when the SDK is ready, which is
+ * what makes a consent-delayed `setup()` integrable without polling for `window.formbricks`.
+ *
+ * Every event goes out on two transports:
+ *
+ * 1. **A `CustomEvent` on `window` — always.** The in-house pattern for host-facing signals
+ *    (`formbricks:onFilePick`, the route-change events) and vendor-neutral: nothing accumulates,
+ *    and no global we do not own is touched.
+ * 2. **A `window.dataLayer.push` — via the standard GTM idiom** (`window.dataLayer = window.dataLayer
+ *    || []`). The idiom survives load order: if GTM initialises after us it drains what is already
+ *    queued, whereas pushing only when the array exists would silently drop every event emitted
+ *    before GTM loads. Cost, accepted knowingly: a site with no GTM gets a `dataLayer` array that
+ *    accumulates small event objects with nothing draining it.
+ *
+ * One underscored name per event, identical on both transports, published to integrators — the
+ * constants hold the full literal so the string a customer's GTM trigger matches is greppable here.
+ *
+ * The payload is nested under a `formbricks` key on the dataLayer (instead of spread flat) because
+ * GTM's Data Layer Variables read a *merged* model: a flat `action` or `finished` would persist
+ * across pushes and collide with the host's own keys — `action` in particular is one of the most
+ * common keys in any ecommerce dataLayer. `CustomEvent.detail` gives the same isolation for free.
+ *
+ * No PII: payloads carry ids and the `finished` boolean, never response content or scores.
+ */
+export const FORMBRICKS_EVENTS = {
+  setupSuccessful: "formbricks_setup_successful",
+  actionTracked: "formbricks_action_tracked",
+  surveyShown: "formbricks_survey_shown",
+  responseSubmitted: "formbricks_response_submitted",
+} as const;
+
+export type TFormbricksEventName = (typeof FORMBRICKS_EVENTS)[keyof typeof FORMBRICKS_EVENTS];
+
+export type TFormbricksEventPayload = Record<string, string | number | boolean | undefined>;
+
+export const emitFormbricksEvent = (event: TFormbricksEventName, payload: TFormbricksEventPayload): void => {
+  // js-core is imported by SSR bundles; emitting is meaningless (and crashes) off the browser.
+  if (globalThis.window === undefined) return;
+
+  window.dispatchEvent(new CustomEvent<TFormbricksEventPayload>(event, { detail: payload }));
+
+  window.dataLayer = window.dataLayer ?? [];
+  window.dataLayer.push({ event, formbricks: payload });
+};
--- a/packages/js-core/src/lib/survey/widget.ts
+++ b/packages/js-core/src/lib/survey/widget.ts
@@ -1,5 +1,6 @@
 import { Config } from "@/lib/common/config";
 import { CONTAINER_ID } from "@/lib/common/constants";
+import { FORMBRICKS_EVENTS, emitFormbricksEvent } from "@/lib/common/events";
 import { Logger } from "@/lib/common/logger";
 import { executeRecaptcha, loadRecaptchaScript } from "@/lib/common/recaptcha";
 import { TimeoutStack } from "@/lib/common/timeout-stack";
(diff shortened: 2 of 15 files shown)


packages/surveys/src/components/general/survey.tsx
```diff
@@ -270,6 +288,7 @@ export function Survey({
           },
           onResponseCreated: (responseId) => {
             void persistSurveyStateSnapshot({ responseId });
+            triggerResponseCreatedOnce(responseId);
```
kiln-agent: Moving the trigger to the queue's ack also moves host-supplied callback code **inside `ResponseQueue.sendResponse`'s try block**, and that changes what a throw costs.

`sendResponse` (response-queue.ts:645-698) calls `this.config.onResponseCreated?.(response.data.id)` at line 676 inside the big `try`. That now reaches js-core's `onResponseCreated` (widget.ts:220), which synchronously runs `emitFormbricksEvent` → `window.dataLayer.push`, then `config.update(...)`, `filterSurveys(...)` and `refreshSegmentsAfterInteraction(...)`.

If any of those throws — the realistic one being a host page where `window.dataLayer` is not an array, so `.push is not a function` (see the comment on `events.ts`) — the catch at line 692 swallows it and returns `err({ code: "internal_server_error" })`. `processQueue` then treats the send as failed → `onResponseSendingFailed` → the respondent sees the "could not send response" error UI and a retry, **for a response the server already created and returned an id for**.

Before this PR `triggerResponseCreatedOnce()` ran from `onResponseCreateOrUpdate` on the caller's stack, outside that try, so this blast radius is new. Cheapest fix is to isolate the callback:

```suggestion
          onResponseCreated: (responseId) => {
            void persistSurveyStateSnapshot({ responseId });
            try {
              triggerResponseCreatedOnce(responseId);
            } catch (error) {
              // Host-supplied callback: never let it turn a persisted response into a send failure.
              console.error("Formbricks: onResponseCreated handler threw", error);
            }
          },
```


---

packages/js-core/src/lib/common/events.ts
```diff
@@ -0,0 +1,46 @@
...
+export type TFormbricksEventPayload = Record<string, string | number | boolean | undefined>;
+
+export const emitFormbricksEvent = (event: TFormbricksEventName, payload: TFormbricksEventPayload): void => {
+  // js-core is imported by SSR bundles; emitting is meaningless (and crashes) off the browser.
+  if (globalThis.window === undefined) return;
+
+  window.dispatchEvent(new CustomEvent<TFormbricksEventPayload>(event, { detail: payload }));
+
+  window.dataLayer = window.dataLayer ?? [];
+  window.dataLayer.push({ event, formbricks: payload });
```
kiln-agent: Two things about this pair of lines.

**1. `?? []` only guards null/undefined, and the emit has no try/catch.** A site that defines `window.dataLayer = {}` (a hand-rolled shim, or another tag vendor that reuses the name) keeps that object through `??`, and `push` throws `TypeError: window.dataLayer.push is not a function`. `dataLayer.push` is host-owned code once GTM replaces it, so we can't assume it's safe either. The throw then escapes `emitFormbricksEvent` into `setup()`, `trackAction()`, `onDisplayCreated`, `onResponseCreated` and `onFinished` — see the separate comments on `widget.ts` and `survey.tsx` for what each one costs (lost display bookkeeping; a persisted response reported to the respondent as failed). A host page's broken `dataLayer` should never be able to break survey delivery:

```suggestion
  if (!Array.isArray(window.dataLayer)) window.dataLayer = [];
  try {
    window.dataLayer.push({ event, formbricks: payload });
  } catch (error) {
    // `dataLayer.push` is host-owned once GTM replaces it — never let it escape into SDK flow.
    console.error("Formbricks: failed to push to dataLayer", error);
  }
```

(the `dispatchEvent` above deserves the same wrapper.)

**2. The nesting doesn't buy what the file comment claims.** It does stop collisions with the host's top-level keys, but GTM merges *nested* objects recursively too, so `formbricks.*` persists across our own pushes exactly like a flat `action`/`finished` would. Concretely: survey A completes → last push is `formbricks: { surveyId: "A", responseId: "r1", finished: true }`. Later survey B shows → we push `formbricks: { surveyId: "B" }` only, and a tag triggered on `formbricks_survey_shown` reading `formbricks.responseId` still resolves `"r1"` and `formbricks.finished` still resolves `true`. Same for `formbricks.action` leaking out of `formbricks_action_tracked` into every later push. Since these names are a committed external contract, worth pushing the full key set on every event (explicit `undefined`/`null` where it doesn't apply) rather than only the keys that event carries.


---

packages/js-core/src/lib/survey/widget.ts
```diff
@@ -185,6 +186,8 @@ export const renderWidget = async (
       isSpamProtectionEnabled,
       getRecaptchaToken,
       onDisplayCreated: () => {
+        emitFormbricksEvent(FORMBRICKS_EVENTS.surveyShown, { surveyId: survey.id });
```
kiln-agent: Ordering: the emit is the *first* statement here, ahead of the display bookkeeping, and the caller swallows throws.

`createDisplay` in survey.tsx:502-525 invokes `onDisplayCreated()` inside `try { ... } catch (err) { console.error("error creating display: ", err) }`. So if `emitFormbricksEvent` throws (a host `window.dataLayer` that isn't an array — see the `events.ts` comment), everything below this line is skipped: `config.update` never appends the new display or `lastDisplayAt`, and `refreshSegmentsAfterInteraction` never runs — while the display *was* created server-side. Display caps and recontact then evaluate against stale local state and the same survey re-shows on every subsequent trigger, with one console.error as the only trace.

Cheapest fix is to emit after the SDK has finished its own state update (the same applies to the `onResponseCreated` emit below), so host-page failures can't cost us our bookkeeping. Guarding inside `emitFormbricksEvent` fixes both at once.


---

kiln-agent (commented):
Code review of the event-bus diff (3 inline findings; the `events.ts` comment covers two related points).

All three trace back to one root cause: `emitFormbricksEvent` runs host-controlled code (`window.dataLayer.push` — replaced by GTM once it loads) with no guard, and its call sites sit at the head of, or inside, SDK-critical paths. Guarding the emitter itself (`Array.isArray` + try/catch on both transports) resolves the widget.ts and survey.tsx findings too.

Verified locally: `pnpm --filter @formbricks/js-core test` (32 tests in the touched suites) and the full `@formbricks/surveys` suite (697) pass, and `tsc --noEmit` for js-core is clean — these are runtime robustness / contract issues, not test failures.


---

packages/surveys/src/components/general/survey.tsx
pandeymangg: Taken verbatim in aa34bce9f — the queue-side callback now isolates `triggerResponseCreatedOnce` in a try/catch with a console.error, so a host-page throw can no longer turn a persisted response into a send failure.

packages/js-core/src/lib/common/events.ts
pandeymangg: Both taken in aa34bce9f. (1) `Array.isArray` check + separate try/catch per transport, so neither a `{}`-shaped dataLayer nor a throwing host `push`/listener escapes — 3 new tests, 6 go red with the hardening reverted. (2) You're right that the nesting didn't stop recursive-merge staleness inside our own namespace — every push now carries the full key set (`workspaceId/surveyId/responseId/finished/action`) with explicit `null`s, so a later event's read can't resolve an earlier event's value. `CustomEvent.detail` stays minimal (no merge semantics there).

packages/js-core/src/lib/survey/widget.ts
pandeymangg: Taken in aa34bce9f — all three emits now run LAST in their callbacks, after `config.update`/segment refresh, so even a failure the emitter's own guards somehow missed could no longer cost the display bookkeeping. Ordering + the guarded emitter are the two walls.

packages/js-core/src/lib/common/events.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.error(`Formbricks: a "${event}" event listener threw`, error);
+  }
+
+  try {
+    // `Array.isArray`, not `?? []`: a host page that set `window.dataLayer = {}` (a hand-rolled
+    // shim, or another vendor reusing the name) survives nullish coalescing and then throws on
+    // `.push`. GTM's own snippet only ever creates an array, and its loaded state keeps the array
+    // and swaps the `push` method, so `Array.isArray` stays true on every real GTM page.
+    if (!Array.isArray(window.dataLayer)) window.dataLayer = [];
+    window.dataLayer.push({ event, formbricks: { ...EMPTY_DATALAYER_PAYLOAD, ...payload } });
```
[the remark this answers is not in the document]
pandeymangg: Taken in eefeb4e4b — undefined-valued payload keys are stripped before the sentinel merge, so a present-but-undefined `responseId` resolves to `null` on the wire instead of re-opening the recursive-merge bleed. Pinned by a test (emit with `responseId: undefined` → `formbricks.responseId === null`); reverting the strip flips it red. Went with the strip rather than relying on GTM assigning undefined — safer under both merge behaviors.

itsjavi (approved):
verified and LGTM. I hardened a test and a fixed double warning.

PR #8991, opened 2026-08-26
fix(api-v3): return 422 instead of 500 for post-schema survey validation failures [ENG-2587]

Dhruwang:
Fixes ENG-2587

## What & why

**Was:** `POST /api/v3/surveys` answered `500` for a document its own request schema accepts but the survey service rejects on write — a CTA `buttonUrl` of `tel:+123456789` was the case QA hit.

**Now:** `422` naming the offending field in `invalid_params`, e.g. `blocks.0.elements.0.buttonUrl`, and no survey row is written. See Risks for the one write that still precedes the 422.

- Two more shapes were the same 500 and are fixed too, both found in review: a plain `http://` CTA link (`isSafeLinkUrl` allows `http:`, `safeUrlRefinement` allows only `https://` and `http://localhost`), and a blank element headline once every declared locale key is present.
- The allowlist divergence itself is ENG-2586.

## Where to look

- [`create.ts:217`](https://github.com/formbricks/formbricks/blob/0308054/apps/web/app/api/v3/surveys/create.ts#L217) — parse before any write
- [`operations.ts:267`](https://github.com/formbricks/formbricks/blob/0308054/apps/web/app/api/v3/surveys/lib/operations.ts#L267) — 422 on the typed error only
- [`schemas.ts:1328`](https://github.com/formbricks/formbricks/blob/0308054/apps/web/app/api/v3/surveys/schemas.ts#L1328) — strips the editor-only `-fLang-` marker, linearly

## Breaking changes

| `POST /api/v3/surveys` status when the body passes the request schema but fails the write schema | `500` | `422` with `invalid_params` | v3 consumers branching on 5xx vs 4xx | Stop retrying; read `invalid_params` |
| `invalid_params[].reason` for language-validation failures (`schemas.ts:1328`) | not emitted — this case was the `500` above, with no `invalid_params` at all | reason with the internal `-fLang-` delimiter stripped, e.g. `…missing for the following languages: de-DE` | v3 consumers reading `reason` on the new 422 | None |

The second row is a consequence of the first rather than a separate break: `-fLang-` has never appeared in any v3 response. The label validators that embed it are reachable only through `surveyRefinement`, and no v3 request schema applies it — `prepare.ts` and `operations.ts` feed `formatV3ZodInvalidParams` from `ZV3CreateSurveyBody`/patch schemas only, so the create parse added here is the first v3 code path that reaches those messages. It changes no previously-shipped value; it shapes the reason of a response that did not exist before. Recorded as its own row because the diff does change an emitted API value.

No self-hoster step: 422 is already documented for this operation, and no 2xx changes.

Before:

Screenshot 2026-08-31 at 12 18 04 PM

After:

Screenshot 2026-08-31 at 12 18 13 PM

## Migrations & env

- none

## How this was tested

Rerun: `npx vitest run --root apps/web app/api/v3/surveys` — 326 passed

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `tel:` CTA rejected pre-write, no survey row | unit (red on main) — `create.test.ts:312`; mutate `create.ts:217-225` by deleting the parse | Path `blocks.0.elements.0.buttonUrl`; `createSurvey` not called. Mutated → the old 500 path returns |
| Plain `http://` CTA rejected the same way | unit (red on main) — `create.test.ts:344` | Same path and typed error; verified it lands on the new parse rather than assumed |
| Blank headline reaches the new parse — no languages declared, and all declared locales present | unit (red on main) — `create.test.ts:413`, two shapes | `V3SurveyInputValidationError` on `blocks.0.elements.0.headline`; both were 500s on `main` |
| Blank headline with a *missing* declared locale goes to the earlier guard instead | unit (guard) — `create.test.ts:388` | `V3SurveyReferenceValidationError`, already 422 before this change. Pins which side of the parse each shape lands on |
| `https` CTA still creates | unit (guard) — `create.test.ts:445` | `createSurvey` called, survey returned |
| Typed error maps to 422 with the path | unit (red on main) — `operations.test.ts:490` | `invalid_params[0].name` is the path; the branch does not exist on main |
| Bare `ValidationError` still maps to 500 | unit (mutation) — test `operations.test.ts:514`; mutate `operations.ts:267` by broadening the branch to `ValidationError` | Mutated → that test fails, proving post-commit faults stay 5xx |
| `-fLang-` stripped, both message shapes | unit (mutation) — tests `schemas.test.ts:888` and `:915`; mutate `schemas.ts:1346` to pass `issue.message` straight through | Whole-string assertions; messages built by the real validator. Mutated → the marker reaches `reason` |
| A reason without the marker is unchanged | unit (guard) — `schemas.test.ts:901` | `invalid_locale` reason byte-identical |
| The strip is linear, not quadratic | unit (mutation) — test `schemas.test.ts:934`, two 100k whitespace-run cases with a 2s timeout; mutate `schemas.ts:1328-1337` by restoring the `\s*`-wrapped regex | 69ms linear; mutated → both cases time out. Guard, not benchmark: ~350x margin |
| Other five mapper branches unchanged | unit (guard) — `operations.test.ts:423` | Pre-existing 422/400/403/403/500 assertions still pass |
| OpenAPI 422 text updated, bundle in sync | manual — `pnpm api:v3:check` | exit 0 |
| Full HTTP round-trip on a live instance | manual — by reviewer, real Postgres + Valkey, AuthZed enabled, over HTTP | `tel:`, `http://` and blank headline → 422 with correct paths, 0 `Survey` rows; `https://` → 201; parse removed → 500; rejected create declaring `fr-FR` → 422 with one new `Language` row left behind |

**Open gaps**

- **Which side of the parse a blank headline lands on depends on the language set**, not on the blankness: with every declared locale key present (or none declared) it reaches the new parse and gets the 422 this PR adds; with a declared locale key missing anywhere in the document — including `metadata.title` — `prepareV3SurveyCreate` rejects that first, as it already did. Both routes are pinned above. Two earlier versions of this description got this wrong in opposite directions.
- **The preflight disagrees with create.** `POST /api/v3/surveys/validate` answers `valid: true` for documents create now rejects — confirmed live for both `tel:` and plain `http://`. It runs only `prepareV3SurveyCreateInput`, which never reaches `surveyRefinement`. Pre-existing; raised on ENG-2586.
- **No length cap on a v3 language code**, so a large `languages[].code` echoes back in full inside `invalid_params[].reason` (`schemas.ts:34` has no `.max()`). Pre-existing, caller's own data; raised on ENG-2586.
- Service-schema paths `languages`/`questions` have no request-body counterpart; PATCH (ENG-2588) unchanged.

**Risks**

- **One write still precedes the 422, accepted deliberately.** `ensureV3WorkspaceLanguages` upserts workspace `Language` rows (`languages.ts:56`) before the parse, so a rejected request can leave a new language row behind — unchanged from `main`, where the same upsert ran and then returned 500. Confirmed live in review. The parse cannot move ahead of it for two separate reasons: an *empty* language set skips `surveyRefinement`'s multi-language label checks, gated on `languages.length > 1` (`types.ts:1109`, `types.ts:1454`), handing multi-language documents back their old 500; and a *fabricated* set is not accepted at all, because `ZLanguage` requires full DB rows (`packages/types/workspace.ts:27`).
- Body failures that returned 500 now return 422; `operations.test.ts:514` pins the other side.
- `ZSurveyCreateInput` parses twice here — schema only, no I/O.

---

> [!NOTE]


kiln-agent:
## Review

The fix is correct and well-scoped. `ValidationError` → 422 with `invalid_params` from the ZodError cause is the right mapping, the branch ordering in `mapV3SurveyCreateError` is safe, and the causeless fallback matches the neighbouring `InvalidInputError` shape. One thing worth recording before merge.

### The preflight now disagrees with create

`POST /api/v3/surveys/validate` is documented as the repair loop for exactly this class of failure — *"Invalid survey documents return `200` with `data.valid=false` so agents and MCP clients can repair payloads from structured `invalid_params`."*

For `create`, that endpoint runs only `prepareV3SurveyCreateInput` (`apps/web/app/api/v3/surveys/lib/operations.ts:765`), which parses `ZV3CreateSurveyBody` → `isSafeLinkUrl` and then does reference validation. It never reaches `ZSurveyCreateInput` / `surveyRefinement` → `getZSafeUrl`.

The two allowlists differ:

| | schemes accepted |
| --- | --- |
| `isSafeLinkUrl` (`packages/types/common.ts:51`) | `https:`, `http:`, `mailto:`, `tel:` |
| `safeUrlRefinement` (`packages/types/common.ts:104`) | `https://`, `http://localhost`, `mailto:` |

So a CTA `buttonUrl` of `tel:+123456789` or `http://example.com` gets `valid: true` from the preflight, and — after this PR — a `422` naming `blocks.0.elements.0.buttonUrl` from create. An agent following the documented validate → repair → create flow gets a false green and no way to repair from the preflight's own output.

This divergence is pre-existing, and the PR does not make the behaviour worse — a 500 was strictly less useful than a 422. What changes is its standing: the 422 is now a documented outcome of the create contract, which makes the preflight's `valid: true` an actively misleading answer rather than an obvious bug. Worth a line in ENG-2586 (or a sibling ticket) so the allowlist decision covers the preflight and the write path together, rather than being closed on `getZSafeUrl` alone.

Not a blocker for this PR.

### Checked and cleared

- **Branch ordering** — no error class in the repo extends `ValidationError`; the new branch sits below `V3SurveyReferenceValidationError` / `InvalidInputError` and above `DatabaseError`. No shadowing, and `DatabaseError`'s 500 still maps.
- **`instanceof z.ZodError`** — zod is catalogued at a single version and both `validate.ts` and `operations.ts` resolve the same entry, so the cross-realm failure mode doesn't apply.
- **Path alignment** — `surveyRefinement` emits `["blocks", i, "elements", j, "buttonUrl"]` (`packages/types/surveys/types.ts:1554`), and `executeV3SurveyCreate` passes `input.blocks` through unreordered, so `invalid_params[].name` indices match the request body. The `languages` / `questions` gap noted in the description isn't reachable in practice: `questions` is always `[]` on this path, and `languages` are built from DB rows that already satisfy their schema.
- **`cause` as an own enumerable property** — no client-facing leak. `handleServerError` returns `e.message` only; all three other `instanceof ValidationError` handlers use `err.message`. pino's `err` serializer folds cause message/stack into the log line rather than dumping the ZodError, and `validateInputs` already logged the full ZodError immediately before throwing.
- **`cause?: unknown` class field** — matches `lib.es2022.error.d.ts`'s optional `cause?: unknown`; `packages/types` compiles under lib ES2021 where `Error` has no `cause`, so there's no conflict either way, and field-init order can't clobber the constructor assignment.
- **OpenAPI** — `InvalidParam.code` is optional and `additionalProperties: false` is satisfied by `{name, reason}`, so the new 422 body matches the committed schema. The `src/` and bundle edits fold to the same string.
- **PATCH asymmetry** — the stated rationale checks out: `patch.ts` writes blocks directly and never reaches `updateSurvey`'s `validateInputs([updatedSurvey, ZSurvey])`, so `surveyRefinement` genuinely does not run there. ENG-2588 is a separate root cause.
- **Post-write misclassification** — a `ValidationError` thrown after the row commits would return a client-blaming 422, but isn't reachable: the only post-write `validateInputs` on this path is `getSurvey([surveyId, ZId])` with a DB-generated cuid2, and `normalizeSurveyScheduling` throws before the transaction.

---

apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -259,6 +277,19 @@ function mapV3SurveyCreateError(
       instance,
     });
   }
+  if (err instanceof ValidationError) {
```
kiln-agent: This branch lives in the response mapper, so it catches *every* `ValidationError` that escapes `createV3Survey` — including ones raised after the survey row is already committed. The coverage row "No survey row written on the rejected request — the throw happens inside `createSurvey`'s `validateInputs`, before any write" only holds for the pre-write pair; `createSurvey` keeps going after the transaction commits.

Concretely: `apps/web/lib/survey/service.ts:885-892` — the `prisma.$transaction` has committed, then `subscribeOrganizationMembersToSurveyResponses(reconciledSurvey.id, createdBy, organization.id)` runs, which calls `updateUser` (`apps/web/lib/user/service.ts:63`) → `validateInputs([personId, ZId], [data, ZUserUpdateInput.partial()])`. `data.notificationSettings` is rebuilt by spreading the user's stored `notificationSettings` JSON column, which is never validated on read, and `ZUserNotificationSettings` requires `alert: z.record(z.string(), z.boolean())`. So a session-authenticated `POST /api/v3/surveys` by a user whose stored `alert` map holds a non-boolean value creates the survey and *then* throws `ValidationError`.

Two consequences of routing that through this branch:

1. The client gets `422 "Survey document failed validation"` for a survey that exists. A caller that reads 4xx as "nothing was written" (the exact behaviour the PR description argues for) retries and ends up with duplicates.
2. `log.warn` instead of the old `log.error` + 500 takes a genuine server-side fault out of 5xx alerting entirely.

Both go away if the mapping is anchored to the pre-write boundary rather than the whole call tree — e.g. catch the `ValidationError` around the `createSurvey(...)` call in `executeV3SurveyCreate` and rethrow it as a `V3SurveyReferenceValidationError`-style typed error, or gate this branch on the cause being a ZodError from the survey-input schema and let anything else keep falling through to the 500.

---

apps/web/app/api/v3/surveys/lib/operations.ts
```diff
@@ -215,6 +221,18 @@ export async function listV3Surveys({
   }
 }
 
+/**
+ * `validateInputs` attaches the underlying ZodError to the `ValidationError` it throws, so the
+ * response can itemize the offending field paths. Falls back to the flattened message when the
+ * `ValidationError` was raised by hand without a cause.
+ */
+function getValidationErrorInvalidParams(err: ValidationError): InvalidParam[] {
```
kiln-agent: Pushing on the `Open gaps` entry about service-schema paths, because there is a case sharper than `languages`/`questions`: `createSurvey` validates **two** pairs — `validateInputs([workspaceId, ZId], [surveyBody, ZSurveyCreateInput])` — and `validateInputs` throws on whichever pair fails first, with no marker of which one it was.

If the `workspaceId` pair fails, the ZodError carries a single issue with an **empty** `path`, so `formatV3ZodInvalidParams(err.cause, "body")` falls back to `{ name: "body", reason: "Invalid cuid2" }`. That reports a purely server-derived value (`authResult.workspaceId`, resolved by `requireV3WorkspaceAccess`, never sent by the caller) as a problem with the caller's body — and, per the other comment, at `warn` with a 4xx instead of a 500.

Not reachable today (workspace ids are cuid2 by construction), but it is the same structural issue as the acknowledged `languages`/`questions` mismatch and it is cheap to close: filter to issues whose path head is an actual `ZV3CreateSurveyBody` key, and treat anything else (empty path included) as unmapped — either `{ name: "body" }` with the flattened message, or let it fall through to the 500 it really is. That also makes the branch robust if a future `validateInputs` call in the create path adds a third pair.

---

apps/web/app/api/v3/surveys/lib/operations.test.ts
```diff
@@ -483,6 +488,60 @@ describe("createV3SurveyResponse", () => {
       ).status
     ).toBe(500);
   });
+
+  // ENG-2587: `createSurvey` runs `validateInputs([surveyBody, ZSurveyCreateInput])`, whose stricter
+  // refinement rejects payloads the route's own `ZV3CreateSurveyBody` accepts (a CTA `buttonUrl` of
+  // `tel:…` was the case QA hit). That ValidationError had no branch here and surfaced as a 500.
+  test("maps a ValidationError raised below the schema layer to 422 naming the offending path", async () => {
```
kiln-agent: The comment on the next line says "built through the real `validateInputs` so the test breaks if the cause stops being attached" — true, but the *schema* is synthetic, so the assertion on `blocks.0.elements.0.buttonUrl` is guaranteed by the shape the test itself declares, not by anything in the product.

I checked the real path and it does line up today (`packages/types/surveys/types.ts:1563` adds the issue at `["blocks", blockIndex, "elements", elementIndex, "buttonUrl"]`), so the PR's claim is correct. But nothing in the suite runs `ZSurveyCreateInput`/`surveyRefinement`, so this test stays green if that path changes, or if the CTA check stops firing — note it is gated on `element.buttonExternal`, so the ticket's repro only reproduces with `buttonExternal: true`. Either regression would silently turn the ENG-2587 case back into a `{ name: "body" }` response or a 500.

Driving the error through the real schema with a CTA element (`buttonExternal: true`, `buttonUrl: "tel:+123456789"`) rather than a locally declared `z.object({ blocks: ... })` makes the test fail on those regressions, and it costs nothing here — `createV3Survey` is already mocked, so no DB is involved.

---

kiln-agent (commented):
Reviewed the diff for correctness. The core fix is right: I confirmed `isSafeLinkUrl` (packages/types/common.ts:51) allows `tel:`/`http:` while `safeUrlRefinement` (line 104) requires `https://`/localhost/`mailto:`, so the route schema really does admit payloads `ZSurveyCreateInput` rejects; the refinement's issue path is `["blocks", i, "elements", j, "buttonUrl"]` so the reported `invalid_params` name is accurate; zod and `@formbricks/types` resolve to one hoisted zod 4.3.6, so the `err.cause instanceof z.ZodError` check holds; the `cause` class field is safe under both `useDefineForClassFields` settings (assignment follows the field definition); the OpenAPI bundle is in sync with the source for POST /api/v3/surveys; and I checked the leak risk from the enumerable `cause` — zod 4 issues carry no received values (only key names), so nothing new reaches a client or Sentry. `pnpm vitest run app/api/v3/surveys/lib/operations.test.ts lib/utils/validate.test.ts` passes (42/42) locally.

Three comments below: one substantive (the 422 branch's blast radius extends past the pre-write boundary), one pushing on an acknowledged `Open gaps` entry, one on the regression test's fixture.

---

apps/web/app/api/v3/surveys/lib/operations.ts
Dhruwang: You're right, and this was the real defect. Fixed in 3ef5045 by changing the approach rather than patching the branch.

I traced your chain end to end before touching anything, and every link holds:

- `lib/survey/service.ts` — `await prisma.$transaction(...)` returns, *then* `reconcilePersistedSurveySchedulingIfDue`, *then* `subscribeOrganizationMembersToSurveyResponses(reconciledSurvey.id, createdBy, organization.id)`. Post-commit, as you said.
- `lib/organization/service.ts:429-442` — it spreads `surveyCreator.notificationSettings` (a Prisma JSON column, never validated on read) into `updatedNotificationSettings`, sets `alert[surveyId] = true`, and calls `updateUser`.
- `lib/user/service.ts:62` — `validateInputs([personId, ZId], [data, ZUserUpdateInput.partial()])`, and `ZUserNotificationSettings.alert` is `z.record(z.string(), z.boolean())`.
- `createSurvey`'s catch only converts `Prisma.PrismaClientKnownRequestError`, so the `ValidationError` propagates untouched into the mapper.

So a session-authenticated create by a user with a corrupt stored `alert` map got the survey written and *then* a 422 telling them the document failed validation. Both consequences you named are real, and the second is the worse one: my branch quietly moved a data-integrity fault from `log.error` + 500 to `log.warn` + 4xx.

**What changed.** The mapping is anchored to provenance now instead of inferred from the error class. `executeV3SurveyCreate` runs `ZSurveyCreateInput.safeParse` on the create input itself, before any DB work, and throws a typed `V3SurveyInputValidationError` carrying the formatted invalid params. The mapper has a branch for *that* error only, so every other `ValidationError` — post-commit ones included — keeps its 500 and its `log.error`. There's now a test asserting exactly that: `"keeps a bare ValidationError on the 500 path, so post-commit faults are not reported as 4xx"`.

Two knock-on effects worth flagging:

- It closes your `workspaceId`-pair comment structurally rather than by filtering. Since I only convert my own explicit parse, that pair can no longer surface as `{ name: "body" }` — no path-head allowlist needed, and a future third `validateInputs` pair can't leak either.
- It **shrank** the PR. The `cause` plumbing existed only to recover paths from a caught `ValidationError`, so `packages/types/errors.ts` and `lib/utils/validate.ts` are reverted to their original state. This no longer touches a shared package at all, which also retires the enumerable-`cause` risk you took the trouble to clear.

Thanks for pushing on this — the first version would have introduced a worse bug than the one it fixed.

---

apps/web/app/api/v3/surveys/lib/operations.ts
Dhruwang: Verified — `createSurvey` does open with `validateInputs([workspaceId, ZId], [surveyBody, ZSurveyCreateInput])`, and `validateInputs` throws on the first failing pair with nothing to say which it was. Your read of the consequence is right: an empty issue path falls back to `{ name: "body" }`, which would have blamed the caller's body for `authResult.workspaceId` — a value `requireV3WorkspaceAccess` derives server-side and the caller never sends.

Closed structurally in 3ef5045 rather than by filtering. `executeV3SurveyCreate` now parses `ZSurveyCreateInput` itself and throws a typed `V3SurveyInputValidationError`; the 422 branch keys on that error alone, so nothing else that `createSurvey` validates can reach it. The `workspaceId` pair, and any pair a future change adds, keep the 500 they should have had.

I went that way over the path-head allowlist you suggested because it removes the guess instead of narrowing it — with the allowlist I'd still be inferring provenance from issue paths, and your other comment showed why that inference is the actual problem. Agreed it was cheap either way; this happened to also delete code.

---

apps/web/app/api/v3/surveys/lib/operations.test.ts
Dhruwang: Fair, and the `buttonExternal` gate is the part that makes it matter — a synthetic `z.object({ blocks: ... })` can't notice if the CTA branch stops firing, so the test would have stayed green while the ticket's own repro silently regressed.

Replaced in 3ef5045. The regression test now lives in `create.test.ts` and drives the real thing: `ZV3CreateSurveyBody.parse` on a CTA element with `buttonExternal: true` and `buttonUrl: "tel:+123456789"`, then `createV3Survey(...)`, which runs the real `ZSurveyCreateInput` / `surveyRefinement` through the new pre-write parse. It asserts `V3SurveyInputValidationError` with `invalidParams` containing `blocks.0.elements.0.buttonUrl` — so the path comes from the product now, not from the fixture. It also asserts `createSurvey` was never called, which is the "no survey row written" claim my description previously only argued for.

Added the positive control alongside it (`"accepts an https CTA buttonUrl through the same path"`), so the test fails if the refinement starts rejecting everything rather than just the disallowed schemes — a mutation the negative case alone can't catch. As you noted, `createSurvey` is mocked there, so no DB.

---

Dhruwang:
**CI note: `Run E2E Tests` failed on `3ef5045`, and I don't believe it's this PR's. Re-running once to confirm.**

What failed ([job](https://github.com/formbricks/formbricks/actions/runs/32953391654/job/98129657336)):

```
1 failed
  [chromium] › apps/web/playwright/workspace-delete.spec.ts:187:5 ›
    keeps the organization in account settings when the delete lands on the onboarding flow
1 flaky
  [chromium] › apps/web/playwright/cross-tenant-survey-access.spec.ts:65:7 ›
    Cross-tenant survey access (ENG-2101) › keeps access to the workspace's own surveys
111 passed (5.7m)
```

`Error: expect(locator).toBeVisible() failed / element(s) not found` at `workspace-delete.spec.ts:260`, on all three attempts.

**Why I don't think it's mine.** This PR changes four things: the v3 create route (`create.ts`, `lib/operations.ts`), their two test files, and the OpenAPI 422 description. `workspace-delete.spec.ts` seeds its fixtures with `prisma.survey.create` / `prisma.survey.createMany` directly — it never calls `createV3Survey` or `POST /api/v3/surveys`, so the pre-write `ZSurveyCreateInput.safeParse` this PR adds is never executed on that path. The only new runtime import is a schema into a `server-only` API module; had that broken anything shared, far more than 1 of 113 specs would be red.

The failing assertion is also inherently timing-dependent by its own description — it targets the window where "the cookie still named the workspace we just deleted until the delete action started repointing it", and it depends on `createdAt` ordering between two organizations created milliseconds apart. Playwright's retries reuse the same app instance and database, so once that state resolves the wrong way it fails all three attempts deterministically rather than looking like a flake.

**Being straight about the counter-evidence:** I can't claim this is red on the base. Of the last 30 `PR Update` runs, this is the only failure — the `gh-readonly-queue/main/*` runs are all green, and this PR's own earlier head `9403ada` passed E2E. So "green on base" is true, which is why I'm spending the one re-run rather than asserting flake and moving on.

**No fix ported, because none exists** — I found no open PR addressing `workspace-delete.spec.ts` ordering. I'm deliberately not touching that spec from here either: making its org-ordering and cookie-repointing assertions robust is a real change to an unrelated feature area's spec, and smuggling it into an API-validation PR is worse than leaving it visible.

If the re-run comes back red, that's a second failure and I'll treat it as real and root-cause it against this diff rather than re-running again.

---

Dhruwang:
@coderabbitai review

Requesting a fresh pass because the two failing pre-merge checks are pinned to `9403ada` and no longer describe this PR. The head is `3ef5045`, two commits later, and the approach changed substantially in between.

The clearest sign they are stale is that the same comment contradicts itself: **Breaking Changes Match The Diff** says "the checkbox is unticked", while **Description check** in that same run says "the breaking-change checkbox correctly reflects the documented HTTP status change". The box is ticked, with the table filled in.

Both checks also cite code that is no longer in the diff:

- `operations.ts:280-291` returning 422 for `ValidationError` — that branch is gone. The 422 is now keyed on a typed `V3SurveyInputValidationError` thrown by an explicit pre-write parse, so every other `ValidationError` keeps its 500. That was the fix for a review finding on the first revision.
- `apps/web/lib/utils/validate.ts` and `packages/types/errors.ts` — reverted to their original state; this PR no longer touches a shared package.
- The Coverage asks (name each test, add a rerunnable command, account for the no-write claim) were all addressed in the rewritten description, which now also carries the `Rerun:` line and `Where to look` that the current template wants.

Since the PR is a draft, auto-review is off, so this needed a manual trigger rather than fixing itself.

---

Dhruwang:
The fresh CodeRabbit pass (`78bec80c`, reviewing `9403ada..3ef5045`) cleared the two checks that were pinned to the old approach, and raised one new point that is correct. Accepting it explicitly rather than fixing it, with the reasoning below — and I've narrowed the description's claim, which was overstated.

## The language upsert really does precede the 422

`ensureV3WorkspaceLanguages` runs at `create.ts:193`, before the `ZSurveyCreateInput.safeParse` at `create.ts:217`, and it is not read-only — `languages.ts:56` is a `prisma.language.upsert` with a populated `create` branch. So a rejected `tel:` request can leave a new workspace `Language` row behind. My "with nothing written" was wrong; it is now "no survey row is written", with the exception spelled out under Risks.

Worth keeping in proportion: the row is workspace-scoped, not survey-scoped, and the upsert no-ops whenever the code already exists — which is the common case. It is also **unchanged from `main`**, where the identical upsert ran and the request then returned 500. This PR does not add the write; it only makes the subsequent failure a 422.

## Why the parse cannot simply move earlier

The obvious fix — parse before resolving languages — is unsound, and I want that on the record rather than left as an obvious-looking improvement for a reviewer to suggest.

`surveyRefinement`'s multi-language label checks are gated on `languages.length > 1` (`types.ts:1109` for questions, `types.ts:1454` for block elements). Parsing with `languages: []` therefore *skips* every one of those checks, and a genuinely broken multi-language document would sail past the pre-write parse and fail inside `createSurvey` instead — back to a 500, for exactly the class of document this PR exists to fix. Fabricating placeholder language rows to satisfy the schema would be worse: it makes the parse validate something the request did not ask for.

Getting the write-free 422 properly means making `ensureV3WorkspaceLanguages` resolve-only and moving the row creation inside `createSurvey`'s transaction. That is a real change to the language path shared with PATCH, well outside a status-code fix, and I'd rather it be its own ticket than a widening of this one. Happy to file it if you agree it's worth having.

## Also addressed

The `Coverage Accounts For The Whole Diff` warning was accurate on the current description too, not just the old one — the rows said `unit (red on main)` without naming anything. Each row now names its test file, line and test name, and the two red-on-main rows carry the mutation that reproduces them.

---

apps/web/app/api/v3/surveys/create.ts
```diff
@@ -184,6 +209,16 @@ export async function executeV3SurveyCreate(params: {
...
   };
 
+  // Run the survey service's own write schema here, before any DB work, so a document this route's
+  // request schema admits but `surveyRefinement` rejects (a CTA `buttonUrl` of `tel:…` being the
+  // known case) fails with a typed error the route can answer 422. `createSurvey` validates this
+  // same input again; catching its `ValidationError` instead would be wrong, because it also throws
+  // one from post-commit work where the survey row already exists. See V3SurveyInputValidationError.
+  const parsedCreateInput = ZSurveyCreateInput.safeParse(surveyCreateInput);
+  if (!parsedCreateInput.success) {
+    throw new V3SurveyInputValidationError(formatV3ZodInvalidParams(parsedCreateInput.error, "body"));
```
xernobyl, line 219: The refinement messages reach the client verbatim from here, and one of them carries an internal delimiter. A declared language with a blank translation gives:

```json
"invalid_params": [{
  "name": "blocks.0.elements.0.headline",
  "reason": "The question in question 1 of block 1 is missing for the following languages:  -fLang- de-DE"
}]
```

`-fLang-` is editor-only — `modules/survey/editor/components/survey-menu-bar.tsx:296` splits on it to build a toast, and it's produced in three places under `packages/types/surveys/`. It has never been client-visible, because on main this request returned a generic 500 and the message never left the server; the 422 is what exposes it.

Confirmed live on this branch with an ordinary multi-language payload, so it isn't a corner case. Cheapest fix is probably to normalise it where the reasons are built rather than touching the shared validators — replacing `" -fLang- "` with `": "` in `formatV3ZodInvalidParams` would do it. Happy to be told it's not worth holding the PR for.

xernobyl (dismissed):
Approving — the fix is right and the evidence holds up. Thanks for the rewrite in `3ef5045`; keying the 422 on a typed error rather than on `ValidationError` is a much better shape than the first revision.

## What I checked

The property that matters is that the pre-write parse can't reject anything `createSurvey` wouldn't have. It runs the same `ZSurveyCreateInput` over the same object that `createSurvey` re-validates at `lib/survey/service.ts:762-765`, so the accepted set is unchanged and only the status code moves. I also confirmed the classes in `packages/types/errors.ts` are all flat `extends Error` — so `ValidationError` genuinely did fall through to the 500 on main, and the new branch can't be shadowed by the `InvalidInputError` one sitting above it.

Your three "red on main" claims hold. I mutated each rather than taking them on faith:

- delete the parse at `create.ts:217-220` -> the `tel:` test fails on the `instanceof` assertion
- broaden the mapper branch to `ValidationError` -> `operations.test.ts:514` fails with `expected 422 to be 500`
- 317/317 green across `app/api/v3/surveys/`

## The round-trip listed under Open gaps

Since the description called it out, I filled it in rather than bouncing the PR back for it — real stack, isolated Postgres + Valkey, migrations off this branch, seeded API key, over HTTP:

| Request | Result |
| --- | --- |
| `tel:+123456789` CTA | 422, `blocks.0.elements.0.buttonUrl`, 0 `Survey` rows |
| `https://` CTA (control) | 201, survey created |
| the same `tel:` request with the parse removed | 500, generic detail |
| rejected create declaring a new `fr-FR` | 422, no survey — one new `Language` row left behind |

So the 422, the path, the no-write claim and the language side-effect under Risks are all confirmed exactly as written.

It also fixes two cases that aren't in the ticket, both previously 500s: a whitespace-only headline (`blocks.0.elements.0.headline`), and a plain `http://` CTA link. The `http://` one is likely the most common of the three in practice — worth adding to the description, since it makes the breaking-change note better justified rather than worse.

## Three non-blocking things

1. Inline comment below on the `-fLang-` marker escaping into `invalid_params[].reason`. That one is new to this PR, and I'd like it fixed here if it's as small as it looks.

2. kiln-agent's preflight point is still open and I'd rather it didn't get lost. `POST /api/v3/surveys/validate` answers `valid: true` for the exact documents create now rejects — I confirmed it live for both `tel:` and plain `http://`. Agreed it isn't this PR's to fix, since the divergence is pre-existing, but it wants a line in ENG-2586 or a sibling ticket, and probably a line in Open gaps here too. An agent following the documented validate -> repair -> create loop gets a false green with nothing in the preflight's own output to repair from.

3. Small one on the Risks reasoning: the multi-language checks don't depend on the language *ids*, only on `default` / `enabled` / `language.code`, so a fabricated set would still trip them. The real blocker is that `ZSurveyLanguage` wants full DB rows — `createdAt`, `updatedAt`, `alias` and `workspaceId` are all required — which I ran into trying exactly that. The conclusion stands, it's just the wrong reason on the record =)

Also worth a rebase before merge — the branch is 78 commits behind and the AuthZed cutover has landed since. No conflicts, and nothing on main touches `create.ts` or `operations.ts`, so it should be uneventful.

apps/web/app/api/v3/surveys/lib/operations.ts
Dhruwang: placeholder

---

apps/web/app/api/v3/surveys/create.ts
Dhruwang: Fixed in `69d931c`. You're right that the 422 is what exposes it — the marker was always in the message, it just never left the server before.

Reproduced it first: `validateElementLabels` embeds the marker in the `invalidLanguageCodes.length` branch (`packages/types/surveys/elements-validation.ts:58`), which is the declared-language-with-a-blank-translation case you hit — the *missing key* branch above it builds a different message without the marker. Three producers, one editor-only consumer that splits on it, exactly as you described.

**One deviation from your suggested patch.** Replacing `" -fLang- "` with `": "` would double the punctuation: `messageSuffix` is already `" is missing for the following languages: "` with a trailing space, and the marker is appended *after* it, so the raw message has two spaces before `-fLang-` and your replacement yields `…languages: : de-DE`. Collapsing the marker and its surrounding whitespace instead:

```ts
const FIELD_LANGUAGE_MARKER_PATTERN = /\s*-fLang-\s*/g;
// reason: message.replace(FIELD_LANGUAGE_MARKER_PATTERN, " ")
```

gives `The question in question 1 of block 1 is missing for the following languages: de-DE`. Same place you proposed — `formatV3ZodInvalidParams`, not the shared validators.

The test drives the real validator rather than a literal message, since a hand-written fixture would assert nothing about whether the marker is still being embedded. Mutated the fix away to confirm it fails for the right reason:

```
AssertionError: expected 'The question in question 1 of block 1…' not to contain '-fLang-'
Received: "…is missing for the following languages:  -fLang- de-DE"
```

319/319 green in `app/api/v3/surveys/`, typecheck clean.

**On your other two points:**

Point 3 is a fair correction and I've fixed the record. My Risks note gave one reason for two different cases. The *empty* set is blocked by the gating I described. A *fabricated* set is not — you're right that the checks only read `default`/`enabled`/`language.code` — it's blocked by `ZLanguage` requiring full DB rows (`packages/types/workspace.ts:27`: `id` and `workspaceId` as cuid2, `createdAt`, `updatedAt`, `alias`). Description now says that instead.

Point 2 — added to Open gaps with the two schemes you confirmed live, and I'll raise the preflight divergence for ENG-2586 rather than let it die in a thread. Also added your `http://` and whitespace-only-headline findings to the description; agreed the `http://` one strengthens the breaking-change note.

Rebase: brought `main` in via merge (`c485f62`) rather than rebasing, to avoid rewriting pushed history. 80 commits, no conflicts as you predicted, and I reinstalled and rebuilt the dependency graph before re-running the checks so the AuthZed cutover was actually exercised.

Thanks for filling in the round-trip instead of bouncing it back — the language side-effect being confirmed against a real Postgres is worth more than the argument I'd made for it.

---

apps/web/app/api/v3/surveys/lib/operations.ts
Dhruwang: Disregard the "placeholder" comment above — I posted it to this thread by mistake (wrong comment id; it was meant for xernobyl's `-fLang-` review comment, which is answered [here](https://github.com/formbricks/formbricks/pull/8991#discussion_r3902451339)). This thread remains resolved and needs nothing further. Sorry for the noise — I can't delete a review reply, so this note is the correction.

---

apps/web/app/api/v3/surveys/schemas.ts
```diff
@@ -1309,14 +1309,27 @@ export const ZV3SurveyValidationRequestBody = z.discriminatedUnion("operation",
...
 export const ZV3EmptyQuery = z.object({}).strict();
 
+/**
+ * `-fLang-` is an editor-only delimiter: the shared label validators embed it in the issue message so
+ * the editor can split the language list off for its toast (`survey-menu-bar.tsx`). v3 clients receive
+ * these messages verbatim as `invalid_params[].reason`, so the marker is stripped here rather than in
+ * the shared validators the editor still depends on. The suffix it follows already ends in ": ", so
+ * the marker and the whitespace around it collapse to a single space.
+ */
+const FIELD_LANGUAGE_MARKER_PATTERN = /\s*-fLang-\s*/g;
```
xernobyl: SonarQube's S8786 on this line is a true positive, and I'd take it rather than mark it — `\s*` followed by a required literal backtracks at every start position, so a long whitespace run with no marker in it is O(n²).

It's reachable from a single request. `createZV3SurveyLanguageTag` puts the caller's raw value into `Language '${value}' is not a valid locale code`, there's no `.max()` on that string, and `.trim()` only strips the ends — so interior whitespace survives into the message and then comes through here.

Live against `POST /api/v3/surveys/validate` with `languages[0].code = "a" + n spaces + "b"`:

| n | server time |
| --- | --- |
| 1 000 | 0.31 s |
| 20 000 | 0.46 s |
| 100 000 | 5.31 s |

I attributed it by toggling only the `toV3InvalidParamReason` call on a running server: 0.29 s without it, 5.33 s with it. So it is this replace rather than the payload size, and it's quadratic.

It needs a valid API key and production rate limiting caps how often you can ask for it, so it isn't dramatic — but a 5 s event-loop stall from one request is a worse trade than the cosmetic marker it replaced, so I'd rather not merge as-is.

Bounding the quantifiers is the smallest fix and keeps the output identical. I checked both message shapes (the single-code one, and the `", "`-joined ending-card one):

```ts
const FIELD_LANGUAGE_MARKER_PATTERN = /\s{0,8}-fLang-\s{0,8}/g;
```

100 K worst case: 4921 ms -> 2.2 ms. If you'd rather drop the regex altogether, `message.includes("-fLang-") ? message.split("-fLang-").map((s) => s.trim()).join(" ") : message` measures 0.003 ms and is linear by construction — both give the same output on the two shapes.

Separately, and pre-existing: that response echoed the whole 100 KB code back in `invalid_params[].reason`, same missing length cap. It's the caller's own data coming back to them so I wouldn't hold anything for it, but it may be worth a small ticket next to the ENG-2586 work rather than folding it in here.

xernobyl (commented):
Re-reviewed at `69d931c`. The `-fLang-` fix is right, and your correction to my suggested patch was correct — I'd have shipped `languages: : de-DE`. Confirmed the clean reason live, and the test pins the exact string so it catches that variant too: I mutated the fix to my version and it failed on the doubled colon.

I also checked the merge properly rather than taking "no conflicts" on trust — `git merge-tree --write-tree 3ef504530 cfd27e6e2` gives tree `8bf176ff9`, byte-identical to `c485f628f^{tree}`, so it is purely automatic with no hand edits. And I re-ran the live set at the new head with AuthZed actually enabled (SpiceDB plus the six `AUTHZED_*` vars): `tel:`, `http://` and the whitespace-only headline all 422 with the right paths, `https://` still 201, one survey row. The cutover didn't disturb this path.

Not re-approving quite yet, though — one thing the fix brought in with it, in the comment below. It looks small to fix and I'll approve straight after.

Thanks for the description corrections, all three landed. I couldn't check the ENG-2586 raise from here, so that one is on your word.

apps/web/app/api/v3/surveys/create.ts
xernobyl: Confirmed fixed live at `69d931c` — the reason now comes back as `…for the following languages: de-DE`, marker gone and a single colon. You were right about the doubled punctuation in what I proposed, too.

Leaving this thread open only because the replacement itself wants one more pass: Sonar's S8786 on the new pattern is a true positive, written up at [`schemas.ts:1319`](https://github.com/formbricks/formbricks/pull/8991#discussion_r3902589187). Happy to resolve both once that's in.

Dhruwang: True positive, and mine. Fixed in `2048aa7`.

Embarrassing detail worth stating plainly: earlier in this PR I wrote that this regex had "no nested or adjacent-ambiguous quantifier, so no S8786 repeat" — and #8993 in the same session exists to fix an S8786 quadratic regex. I reasoned about the wrong property. A `\s*` before a required literal is ambiguous against *itself* across start positions, which is the whole finding. I also guessed the Sonar issue was the global-flag regex and argued from `EMAIL_REACT_SERVER_MARKER_PATTERN` as precedent; that was the wrong suspect entirely.

**I took your second option, not the bounded one.** `\s{0,8}` is the smaller diff and I don't dispute the measurement, but it leaves a magic bound that silently degrades if a future message carries more adjoining whitespace, and it keeps a regex in a spot where I've now misjudged regex complexity twice. Splitting on the literal has no bound and nothing to backtrack:

```ts
const FIELD_LANGUAGE_MARKER = "-fLang-";

function toV3InvalidParamReason(message: string): string {
  if (!message.includes(FIELD_LANGUAGE_MARKER)) return message;
  return message.split(FIELD_LANGUAGE_MARKER).map((segment) => segment.trim()).join(" ");
}
```

The `includes` guard keeps a marker-free reason byte-identical rather than round-tripping it through split/join, which the existing guard test asserts.

**Reproduced your finding locally before fixing it.** Mutating back to the regex turns the two new cases from 69 ms into a timeout — 26.2 s of test time across the pair, ~13 s each at 100 k. So the quadratic behaviour is real and it is this call, independently of your server measurement.

Coverage added: the comma-joined ending-card shape (`", "`-joined codes), plus two 100 k whitespace-run cases — marker present and absent — asserting the caller's own run is preserved and only the marker and its adjoining whitespace are touched. Those two carry a 2 s timeout. That is a complexity guard, not a benchmark: linear needs microseconds, the regex needs ~13 s, so the margin is ~350x and it is not going to flake. Full suite 322 passed, typecheck clean, 0 new lint errors.

**On your separate point** — the missing length cap, so a 100 KB code echoes back in `invalid_params[].reason`. Agreed it shouldn't be folded in here, and I haven't. It is now in this PR's Open gaps, and I've added it to the ENG-2586 thread alongside the preflight divergence so both live next to the allowlist work. Filing it as its own ticket is the author's call, not mine, so I've left the creation to them rather than opening one unasked.

On the ENG-2586 raise you couldn't verify from there — fair to withhold trust on my word for it. It's comment `c6468b17` on ENG-2586, covering the `isSafeLinkUrl` vs `safeUrlRefinement` scheme table, the preflight's `valid: true`, and that both schemes were confirmed live in your review. The length-cap note is a second comment on the same issue.

---

Dhruwang: Taken in `a09bf68`, and acting on it turned up an error in my own description.

**The fragment assertion.** Valid point, and worth having: those three checks would pass while the formatter mangled some other part of the message. Now one assertion on the whole expected string for both `withMarker` values. Note the suggested snippet had the wrong literal — `is not a valid:` where the fixture says `is not valid:` — so applying it verbatim would have gone red. Took the substance with the string corrected.

**What that shook out.** Writing a unit test for the whitespace-only headline case, which the Coverage check had flagged as unaccounted for, showed my description was wrong about it. My `What & why` said this PR gives that case the same 500 → 422 treatment as `tel:`. It does not:

```
AssertionError: expected V3SurveyReferenceValidationError … to be an instance of V3SurveyInputValidationError
```

`prepareV3SurveyCreate` rejects it first (`create.ts:202-205`), throwing `V3SurveyReferenceValidationError`, which already mapped to 422 before this change. So it was never a 500 and this PR does not touch it. I had generalised from the reviewer's live round-trip — which correctly reported a 422 — to "this PR fixed it", without checking which branch produced the 422. Description corrected, and the test now pins the ordering rather than claiming the fix, so a future reorder that lets it fall through to the new parse gets caught.

I did check the other half rather than assume symmetry: a plain `http://example.com` CTA **does** land on the new pre-write parse and produce `V3SurveyInputValidationError` with `blocks.0.elements.0.buttonUrl`. That case is real, and it now has its own test instead of riding on the `tel:` one.

324 passed, typecheck clean, 0 new lint errors, format clean.

---

apps/web/app/api/v3/surveys/create.test.ts
```diff
@@ -306,6 +306,133 @@ describe("createV3Survey", () => {
...
+      expect.arrayContaining([expect.objectContaining({ name: "blocks.0.elements.0.buttonUrl" })])
+    );
+    expect(createSurvey).not.toHaveBeenCalled();
+  });
+
+  // Ordering guard, not a fix by this PR. A whitespace-only headline never reaches the new pre-write
+  // parse: `prepareV3SurveyCreate` rejects it first with V3SurveyReferenceValidationError, which
+  // already mapped to 422 before this change. Pinned so the new parse is not mistaken for the thing
+  // that catches it, and so a future reorder that lets it fall through shows up here.
+  test("leaves a whitespace-only headline to the earlier preparation guard", async () => {
```
xernobyl: This guard is correct and worth keeping, but it pins the one input shape where the earlier branch intercepts — and the conclusion drawn from it in Open gaps is too broad. "The reviewer's live 422 for it was real, but produced by the pre-existing branch" isn't right.

The deciding factor is the language set, not the blank headline. This fixture spreads `rawCreateBody`, which carries `languages: [{ code: "de-DE" }]`, while the headline only has `en-US` — so what `prepareV3SurveyCreate` rejects is the *absent* `de-DE` key, not the blankness. My live payload had no `languages` array at all.

I ran all three shapes through `createV3Survey` at this head:

| document | outcome |
| --- | --- |
| single language, blank headline (my live payload) | `V3SurveyInputValidationError` — the new parse |
| `de-DE` declared, `de-DE` key absent (this test) | `V3SurveyReferenceValidationError` — earlier branch |
| `de-DE` declared, both keys present and blank | `V3SurveyInputValidationError` — the new parse |

So the whitespace-only headline does reach the new parse in two of the three shapes, and for those it was a 500 on `main`. My original claim was too broad, but so is the retraction — and the current wording attributes my live result to a branch that didn't produce it.

Suggestion: keep this guard as it is, add a sibling that drops `languages` and asserts `V3SurveyInputValidationError` on the same blank headline, and reword the Open gaps line so it turns on whether every declared locale key is present rather than on the case not being this PR's at all.

xernobyl (dismissed):
Approving at `a09bf68`. The linear strip is right, and I verified it independently rather than off the diff:

- Correct on all four shapes I tried — single code, comma-joined ending card, marker-free (early return), and two markers in one message.
- Linear: a 1M-character whitespace run in 0.2 ms, both with and without a marker present.
- Live, the vector I reported is closed: 100 K went 5.31 s -> 0.26 s, and 400 K is 0.24 s — flat rather than quadratic.
- Both of your new linearity cases are genuine guards. I restored the `\s*`-wrapped regex and both timed out, ~4.8 s each. Worth saying your marker-present shape is a better test than the one I measured: putting text between the run and the marker defeats V8's literal prescan, which my own microbenchmark had accidentally dodged.
- The functional set still holds live at this head with AuthZed on — `tel:` and `http://` → 422 with the right paths, `https://` → 201, one survey row.
- SonarQube green with 0 open issues, so S8786 is cleared. 324/324 in `app/api/v3/surveys/`.

Splitting on the literal was the better of the two options I offered — no bound to tune and nothing to reason about. Good call.

One thing left, in the comment below: the whitespace-headline correction went a step too far, and I'd like that description line fixed before this merges. It's the only thing outstanding on my side. E2E and the v3 contract checks are still running.

apps/web/app/api/v3/surveys/schemas.ts
xernobyl: Verified fixed at `a09bf68`, closing the loop on this one. Live, the 100 K case went 5.31 s -> 0.26 s and 400 K is 0.24 s — flat rather than quadratic. Independently, a 1M-character run takes 0.2 ms with and without a marker present, and the output is unchanged on all four message shapes I tried (single code, comma-joined ending card, marker-free, two markers). Both of your linearity cases fail as they should when I restore the `\s*`-wrapped regex, ~4.8 s each. SonarQube green with 0 open issues.

apps/web/app/api/v3/surveys/create.ts
Dhruwang: You're right, and this is the more interesting of my two errors on this case. Fixed in `0308054`.

I ran all three shapes through `createV3Survey` at this head rather than take the table on trust, and it reproduces exactly:

```
no languages declared, blank headline      -> V3SurveyInputValidationError
de-DE declared, de-DE key absent           -> V3SurveyReferenceValidationError
de-DE declared, both keys present, blank    -> V3SurveyInputValidationError
```

So the deciding factor is whether every declared locale key is present, not the blankness — and my fixture spread `rawCreateBody`'s `languages: [{ code: "de-DE" }]` against an `en-US`-only headline, which pinned the one shape where the earlier branch intercepts. Two of three shapes do reach the new parse and were 500s on `main`, so the case is partly this PR's after all. Both the over-broad claim and the over-broad retraction were mine; the retraction was worse, because it attributed your live result to a branch that did not produce it.

**What changed.** Kept your guard, renamed to `routes a blank headline with a missing declared locale to the earlier preparation guard` so it states what it actually pins, and added a `test.each` for the two shapes that reach the new parse, asserting `V3SurveyInputValidationError` with `blocks.0.elements.0.headline`.

Worth recording one thing: my first attempt at that sibling failed, and for the rule under test. I had overridden `metadata.title` to `en-US` only while declaring `de-DE`, so preparation rejected the missing *metadata* key before the element was reached — the same missing-declared-locale route, one field over. The test now carries a matching title per shape, with a comment saying why, because it is an easy trap for whoever edits this next. 326 passed, typecheck clean, 0 new lint errors.

Description updated: `What & why` no longer claims or disclaims the whole case, the Coverage rows name all three shapes, and the Open gaps line now turns on the declared-locale rule rather than on the case "not being this PR's".

Thanks for pushing twice on this one. On the microbenchmark point — good to know the marker-present shape defeats V8's literal prescan; that was luck on my part rather than design, and I would not have known to claim it.

---

Dhruwang:
**Note on the red `PR Check Summary` on `2048aa7` and `a09bf68` — both are my own pushes cancelling in-flight jobs, not test failures.** Flagging it because @xernobyl said they were watching E2E and the contract checks, and a red X there is easy to read as a real failure.

`PR Check Summary` is an aggregator whose only step is `fail if conditional jobs failed`, and it counts `cancelled` as failure. Per job:

| head | what was cancelled | everything else |
| --- | --- | --- |
| `2048aa7` | v3 contract tests — all 51 steps had already succeeded, including `Run v3 contract tests` | unit, typecheck, lint, E2E, Helm, AuthZed all green |
| `a09bf68` | E2E, mid-`Run E2E Tests (Playwright Service)` after 8m18s | unit, typecheck, lint, **contract tests**, Helm, AuthZed all green |

Each cancellation lands within seconds of my next push, which is the concurrency group doing its job. Nothing in either run actually failed.

Worth stating the consequence, because it is a real constraint rather than a curiosity: **E2E takes ~8+ minutes, so any push inside that window cancels it, and E2E last ran to completion on `69d931c`** (09:22:58 → 09:28:02, success). Two pushes since then each killed it in flight. So the fix is simply for me to stop pushing: `0308054` is the head I intend to be final, and its run needs to be left alone to produce a genuine E2E result.

I am not re-running anything or pushing to clear the stale reds — they are attached to superseded commits and will be irrelevant once `0308054` completes. If E2E fails on `0308054`, that is a real signal and I will treat it as one.

---

xernobyl (approved):
Re-approving at `0308054` — the push dismissed the previous one.

This closes the last thing on my side, and I checked it rather than assuming:

- Both new shapes are genuine guards. Removing the pre-write parse fails all four cases that assert it, the two blank-headline ones included.
- The renamed earlier-branch guard still passes under that mutation, which is correct — it pins a branch that exists either way.
- The Open gaps line reads accurately now, and both routes are separate rows in the table. Good catch that `metadata.title` has to carry the same declared locales; that is the same rule biting one level up, and it is worth having in the comment.
- 326/326 in `app/api/v3/surveys/`.

No need to re-run the live stack for this one: every change since `2048aa7` — the head I exercised against a real Postgres with AuthZed enabled — is test-only, so the earlier live evidence still covers the shipping code.

CI is re-running on the new commit and the required checks gate the merge independently of this. Nothing further from me — thanks for turning all four of these around so quickly.

xernobyl (approved):
LGTM! Ship it! ;)

PR #8992, opened 2026-08-26
feat(helm): support Hub workload scheduling

BhagyaAmarasinghe:
No Linear ticket: this is a deployment prerequisite for the Artemis staging clone.

## What & why

The Hub API and Hub migration Job could not be constrained to a dedicated tainted node pool. This adds the same optional node selector, toleration, affinity, and topology-spread interfaces already used by the other chart workloads. Defaults remain empty, so existing releases render unchanged.

## Breaking changes

None. The chart interface is additive and every new value defaults to the prior scheduling behaviour.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Hub API and migration inherit dedicated-pool scheduling | unit (guard): `helm template` with Artemis selector/toleration values | Both rendered pod specs contain the expected selector and `NoSchedule` toleration |
| Existing staging output remains stable | manual: byte-for-byte render comparison using refreshed `origin/main` chart and current staging values | No rendered staging differences |
| Chart remains valid with defaults | unit (guard): `helm lint charts/formbricks` | Passed |

**Open gaps**

- Full Artemis deployment remains gated in the linked infra/GitOps rollout.

**Risks**

- A malformed environment overlay could still target a nonexistent pool; GitOps validates every Artemis workload separately.

---

> [!NOTE]

### DIFF
--- a/charts/formbricks/templates/hub-deployment.yaml
+++ b/charts/formbricks/templates/hub-deployment.yaml
@@ -28,6 +28,22 @@ spec:
         app.kubernetes.io/instance: {{ .Release.Name }}
         app.kubernetes.io/component: hub
     spec:
+      {{- with .Values.hub.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.deployment.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml .Values.deployment.imagePullSecrets | nindent 8 }}
--- a/charts/formbricks/templates/hub-migration-job.yaml
+++ b/charts/formbricks/templates/hub-migration-job.yaml
@@ -33,6 +33,22 @@ spec:
         app.kubernetes.io/component: hub-migration
     spec:
       restartPolicy: Never
+      {{- with .Values.hub.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.hub.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
--- a/charts/formbricks/values.yaml
+++ b/charts/formbricks/values.yaml
@@ -1005,6 +1005,12 @@ hub:
   #   TRANSLATION_DEFAULT_LANGUAGE: en-US
   env: {}
 
+  # Scheduling constraints shared by the Hub API and its migration Job.
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+  topologySpreadConstraints: []
+
   # Optional autoscaling for the Hub API deployment.
   autoscaling:
     enabled: false


kiln-agent (commented):
## Review — `feat(helm): support Hub workload scheduling`

Verified locally with `helm v3.16.3` against `a806fcd`: default render is byte-identical to `main` (only the randomly generated secret values differ between runs), `helm lint` passes, indentation and the `with` guards are correct, and the block matches the shape already used by `cube`, `redis.master`, `hub.worker` and `hub.embeddings`. The additive claim in the description holds.

One thing I'd like your call on before this lands.

### The migration Job inherits `affinity` and `topologySpreadConstraints` that were written for a multi-replica Deployment

`hub-migration-job.yaml` is a `helm.sh/hook: pre-upgrade` (and Argo `Sync`, wave `-1`) Job with `activeDeadlineSeconds: 900` and `backoffLimit: 3`. Rendering it with a realistic Artemis-style value set:

```yaml
hub:
  nodeSelector: { formbricks.com/pool: artemis }
  tolerations: [{ key: formbricks.com/pool, operator: Equal, value: artemis, effect: NoSchedule }]
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector: { matchLabels: { app.kubernetes.io/component: hub } }
          topologyKey: kubernetes.io/hostname
  topologySpreadConstraints:
    - { maxSkew: 1, topologyKey: topology.kubernetes.io/zone, whenUnsatisfiable: DoNotSchedule,
        labelSelector: { matchLabels: { app.kubernetes.io/component: hub } } }
```

produces a Job pod labelled `app.kubernetes.io/component: hub-migration` that carries both constraints verbatim. Two consequences:

1. **The anti-affinity can deadlock an upgrade.** Per-node `requiredDuringScheduling` anti-affinity against `component: hub` is the normal way to spread the API across the pool — and it guarantees that when replicas == pool nodes, every node hosts a `hub` pod. The migration pod is then unschedulable on all of them, sits Pending, gets killed by `activeDeadlineSeconds: 900`, and the `pre-upgrade` hook fails the release before the new Hub API rolls. The failure surfaces only during an upgrade on a fully-packed pool, which is the worst place to discover it.
2. **The spread constraint constrains without being useful.** The Job's pod doesn't carry the label the constraint selects on, so its own placement never changes any domain's count (`selfMatchNum = 0`) — it can only ever filter nodes out, never be satisfied by where the Job lands.

Worth noting the chart's own convention points the other way: neither `templates/migration-job.yaml` nor `hub-embedding-backfill-job.yaml` takes any scheduling values, so this would be the first Job in the chart to inherit a Deployment's affinity.

The stated goal — pinning the Hub workloads to a dedicated tainted pool — is fully served by `nodeSelector` + `tolerations`. Two options:

- Render only `nodeSelector` and `tolerations` on the Job, leaving `affinity` / `topologySpreadConstraints` to the Deployment; or
- Give the Job its own keys under the `hub.migration.*` namespace that already exists (`ttlSecondsAfterFinished`, `backoffLimit`, `activeDeadlineSeconds`), defaulting to the API values so the common case stays one-line.

Either way I'd narrow the `values.yaml` comment, since "shared by the Hub API and its migration Job" is what invites the misconfiguration.

### Notes (non-blocking)

- `hub.nodeSelector` sits a level above `hub.worker.*` and `hub.embeddings.*`, so it reads like "all Hub workloads" while only reaching the API and the migration Job. The inline comment says so explicitly and `hub.autoscaling` already means API-only, so this is consistent — just be aware the Artemis overlay needs `hub.worker.*` and `hub.embeddings.*` set separately for the whole Hub to land on the pool.
- `charts/formbricks/README.md` doesn't gain rows for the four new values, but it's equally missing them for `cube`, `taxonomy`, `hub.worker` and `hub.embeddings`, and no CI job runs `helm-docs` — pre-existing drift, not this PR's to fix.

---

kiln-agent (commented):
## Review — `feat(helm): support Hub workload scheduling`

Verified locally with helm v3.15.4 (the version CI pins) against `a806fcd`:

- `helm lint charts/formbricks` passes.
- Rendering with `hub.nodeSelector` / `hub.tolerations` / `hub.affinity` / `hub.topologySpreadConstraints` set puts all four into both the `hub` Deployment and the `hub-migration` Job pod specs, at the correct indentation.
- Default render is byte-identical to `origin/main` apart from the chart's randomly generated secrets and an unrelated valkey digest bump on main — so the additive claim holds.
- Template style (`with` + `toYaml | nindent 8`) matches `cube-deployment.yaml`, `taxonomy-deployment.yaml`, `hub-worker-deployment.yaml`, `hub-embeddings-deployment.yaml` and `redis-statefulset.yaml` exactly.

Nothing blocking. One thing worth a decision before merge, and three notes.

---

### 1. Sharing `affinity` / `topologySpreadConstraints` with the pre-upgrade migration Job can wedge an upgrade

`hub-migration-job.yaml` is a `helm.sh/hook: pre-upgrade` hook (and an Argo `Sync` hook at wave `-1`), so it is scheduled **while the existing Hub API pods are still running**. Its pod template carries `app.kubernetes.io/name: formbricks-hub` — the same label the Hub API pods carry — and only differs in `app.kubernetes.io/component`.

That means a perfectly ordinary HA value for `hub.affinity` gets applied to the hook pod too:

```yaml
hub:
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app.kubernetes.io/name: formbricks-hub
          topologyKey: kubernetes.io/hostname
```

renders into the Job as-is (confirmed by `helm template --show-only templates/hub-migration-job.yaml`). On a small dedicated pool where every node already runs a Hub replica, the migration pod has nowhere to go, sits `Pending`, and is killed by `activeDeadlineSeconds: 900` — a failed upgrade. A `whenUnsatisfiable: DoNotSchedule` spread constraint selecting the same label fails the same way.

The chart already avoids this on the web tier: `migration-job.yaml` takes **only** `deployment.nodeSelector` and `deployment.tolerations`, deliberately not `deployment.affinity` or `deployment.topologySpreadConstraints`. Suggest matching that precedent — keep `nodeSelector` + `tolerations` on the Hub migration Job (which is all the Artemis dedicated-pool use case needs) and drop the other two, or give the Job its own opt-in override.

Not urgent for the Artemis rollout itself, since that only sets a selector and a toleration — but the footgun ships with the interface.

### 2. The Coverage table calls the render checks `unit (guard)`, but no guard was added

`.github/workflows/helm-chart-validation.yml` is where this chart's optional interfaces get pinned, and it carries guards for the comparable additions (`hub.extraVolumes`, `hub.worker.waitForApi`, `deployment.lifecycle`, cube env, embeddings, backfill). This PR adds a new interface with nothing in CI that can fail if the four blocks are deleted again, while the table labels the rows `unit (guard)`.

In fairness: **no** existing scheduling interface is guarded either — `cube.*`, `taxonomy.*`, `redis.master.*`, `hub.worker.*`, `hub.embeddings.*` all render unasserted, so this is a standing convention gap rather than a regression this PR introduces. Either add the ~6-line assertion or relabel those rows as manual, so the table reports what actually exists.

### 3. `hub-embedding-backfill-job.yaml` is now the only Hub pod spec with no scheduling interface

If the Artemis pool is tainted and someone enables the backfill Job, it has no way to tolerate the taint. It is opt-in and off by default, so this is a note rather than a request — but it is the same gap this PR is closing everywhere else.

### 4. Confirm the Artemis overlay also pins `hub.worker.*`

Scoping the new keys to the API + its migration Job is consistent with how `deployment.*` relates to `migration-job.yaml`. The wrinkle is that other **top-level** `hub.*` keys in this chart explicitly reach the worker as well — `hub.env` ("Applied to both the Hub API and hub-worker"), `hub.extraVolumes` / `hub.extraVolumeMounts` ("Hub API and worker"). So `hub.nodeSelector` not covering hub-worker is a reasonable reading of one convention and a surprise under the other. Your inline comment does disambiguate it; just make sure the environment overlay sets `hub.worker.nodeSelector` / `hub.worker.tolerations` too, or the worker silently stays off the dedicated pool.

---

Also checked and fine: values land at the right nesting under `hub:` (between `env` and `autoscaling`), no duplicate keys, no `values.schema.json` to update, and the omission from the README values table matches every sibling scheduling key (`hub.worker.*`, `hub.embeddings.*`, `cube.*`, `taxonomy.*` are all absent there too), so no README drift is introduced. All 18 checks are green; `mergeable_state: blocked` is just the pending human approval.

---

xernobyl (approved):
LGTM

mattinannt (approved):

PR #8993, opened 2026-08-26
fix(survey-editor): hold the no-rules-means-text invariant and reject exponent input [ENG-2419]

Dhruwang:
Fixes ENG-2419

## What &amp; why

**Was:** Deleting the last validation rule on an Open Text question left `inputType` at `number`, so **Long answer** stayed disabled and off while the section read as off. Separately, `1e5` in a number rule stored `100000`.

**Now:** Deleting the last rule resets `inputType` to `text` **for `number` only**, re-enabling Long answer. Exponents are blocked on keypress and rejected on paste, and a numeric value that is not finite never reaches the stored rule.

- `applyRuleDeletion` returns the remaining rules *and* whether to reset, so the two cannot diverge again.
- Scope note: commit `874c5d8` carried an unrelated Add Block alignment fix. #9085 landed the same fix on `main`, so after the `main` merge that file&#39;s diff is empty — 4 files, all ENG-2419, nothing to split.

## Where to look

The deletion transition, narrowed to `number` — `validation-rules-helpers.ts` line 65:
https://github.com/formbricks/formbricks/blob/785b39f/apps/web/modules/survey/editor/lib/validation-rules-helpers.ts#L65

The paste guard — linear regex (S8786) at line 179, finite check at line 187:
https://github.com/formbricks/formbricks/blob/785b39f/apps/web/modules/survey/editor/lib/validation-rules-helpers.ts#L179

## Breaking changes

None. All changes are internal to `apps/web`&#39;s survey editor. Surveys already saved with a non-`text` `inputType` and zero rules keep working — see the first Open gaps entry, which is why the reset is limited to `number`.

## Migrations &amp; env

- none

## How this was tested

Rerun: `npx vitest run --root apps/web modules/survey/editor/lib/validation-rules-helpers.test.ts` — 75 passed (465 across `modules/survey/editor`)

Both files below are under `apps/web/modules/survey/editor/lib/`; every row names its test title and the exact line to mutate. Each mutation was run, not inferred.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Deleting the last rule resets `inputType: &#34;number&#34;`, and leaves `email`/`url`/`phone` **unchanged** | unit (mutation) — `validation-rules-helpers.test.ts:341` &#34;deleting the last rule asks the caller to reset inputType&#34; and `:353` &#34;deleting the last rule leaves inputType %s alone&#34;; mutate `validation-rules-helpers.ts:65` by removing `inputType === &#34;number&#34; &amp;&amp;` | `:341` gives `resetInputTypeToText: true` with rules empty; `:353` gives `false` for all three. Reverting the narrowing reddens exactly `:353` |
| Disabling the section still clears any non-`text` type, matching `main` | unit (guard) — `validation-rules-helpers.test.ts:364` &#34;disabling the section still clears inputType %s, matching main&#34; | `shouldResetInputTypeToText` still `true` for `email`/`url`/`phone`, so a change to `handleDisable` shows as a deliberate diff |
| Deleting one of several rules leaves `inputType` alone | unit (mutation) — `validation-rules-helpers.test.ts:370` &#34;deleting one of several rules leaves inputType alone&#34;; mutate `validation-rules-helpers.ts:31` (`remainingRuleCount === 0`) | `false` with 1, 2 or 5 remaining |
| The number parse accepts plain decimals and nothing else | unit (red on main) — `validation-rules-helpers.test.ts:255` &#34;parses %j as %i&#34; (describe &#34;number value type rejects scientific notation&#34;) and `:289` &#34;still parses %j as %d&#34;; delete the guard at `validation-rules-helpers.ts:179` | Deleting it reddens **16** tests: 5 of the 7 parse cases (`1e5`, `1E5`, `-1e5`, `1e-5`, `0x10`) plus 11 composed-chain tests. `e` and `Infinity` stay green — `Number()` yields `NaN`/`Infinity`, caught by the finite check at `:187` instead, so the two guards each carry part of the load. `10`, `2.5`, `.5`, `-3`, `-2.5`, `&#34; 7 &#34;` still parse exact |
| A non-finite parse never reaches the stored rule | unit (mutation) — `validation-rules-helpers.test.ts:269` &#34;parses %s as 0 rather than a non-finite number&#34; and `:275` &#34;the overflow never reaches the stored rule params&#34;; mutate `validation-rules-helpers.ts:187` back to `Number(value) || 0` | A 400-digit run parsed `Infinity` and stored `{max: null}` (JSON writes Infinity as null). Now `0` → `{max: 100}`. Reverting reddens only the two Infinity cases; `-`/`.`/`-.` stay green because the old truthy guard already caught `NaN` |
| What a rejected paste actually stores, per rule type | unit (guard) — `validation-rules-helpers.test.ts:421` &#34;the pasted magnitude never reaches the stored rule&#34; and `:429` &#34;a literal 0 is replaced by the rule default too&#34; | Rule default, not `0`: `{max:100}`, `{max:3}`, `{min:1}`. `:429` pins the pre-existing literal-`0` path so a future `Number.isFinite` fix to those helpers shows as a deliberate diff |

Long answer re-enabling after the last number rule is deleted was verified **manually** — before/after screencasts:

Before:

https://github.com/user-attachments/assets/aecfd66c-ee6c-4f06-8fb3-c3c25801d1b3

After:

https://github.com/user-attachments/assets/aca86235-b80e-4584-b081-a2f7c17ac4c4

**Open gaps**

- **Why the reset is `number`-only, added in review.** An earlier revision reset every non-`text` type. `app/lib/templates.ts` builds OpenText elements with `inputType: &#34;email&#34;`, `longAnswer: false` and no `validation` key (139, 281, 2431 — the file has zero `validation:` keys), so zero rules with a non-`text` type is shipped data. Since `open-element-form.tsx` sets `longAnswer: newInputType === &#34;text&#34;`, resetting one turned a template email question into a textarea still placeholdered `example@email.com`. Long answer is disabled for every non-`text` type (`open-element-form.tsx:130`), so that disabled state is correct there, not the ENG-2419 symptom. **Not changed:** `handleDisable` still clears any non-`text` type, as on `main` — whether it should is its own ticket.
- **The regex&#39;s linear-time property is not directly asserted.** Line 179 was rewritten for Sonar S8786 (quadratic backtracking on a long non-numeric paste; 203ms at 8k chars before, 0.044ms after). Accepted-language equivalence was verified exhaustively over 66,430 strings, and that equivalence is what the parse tests cover — but no test measures time, so a future regression to a quadratic form would pass. Sonar is the gate for it.
- The keypress guard (`e`/`E`/`+`, with modifier chords let through) is not covered by the screencasts. Reviewer: type `1e5` into a number rule value and confirm `15` lands.
- Negatives reach the field by paste only: a lone `-` sanitizes to empty in a controlled number input. `min={0}` is this ticket&#39;s own out-of-scope item.

**Risks**

- Resetting `inputType` can invalidate saved logic using `isGreaterThan`; `handleDisable` and `handleInputTypeChange` already do this on `main`. Own ticket. Narrowing to `number` shrinks the exposure to the one type the ticket names.
- `createMinParams`/`createMaxParams` are `Number(value) || default`, so a typed `0` already becomes the default on `main`. Untouched; `:429` pins it.
- Two sequential `updateElement` calls are safe — `elements-view.tsx` re-finds the element in the updater.

---

&gt; [!NOTE]
&gt; **AI model used** — `claude-opus-5`, reasoning effort `unknown`.


apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
```diff
@@ -110,6 +158,10 @@ export const parseRuleValue = (
   }
 
   if (config.valueType === "number") {
+    // `onKeyDown` blocks the exponent keys but not paste, and `Number()` happily reads `1e5` as
+    // 100000. Accept only a plain decimal — optionally signed, since the numeric rule params are
+    // bare `z.number()` and a negative threshold is legitimate — and treat anything else as 0.
+    if (!/^-?\d*\.?\d*$/.test(value.trim())) return 0;
```
kiln-agent: The `0` this returns does not reach the stored rule for every numeric rule type, so the documented contract ("treats anything else as `0`") only holds for `minLength` / `minValue` / `isGreaterThan`.

`handleRuleValueChange` feeds the parsed value straight into `createRuleParams`, and `createMinParams` / `createMaxParams` (validation-rules-utils.ts:180-186) are `Number(value) || defaultValue` — `0` is falsy, so it is replaced by the rule's own default.

Ran the two together on this branch:

```
["isLessThan",   "1e3", 0, {"max":100}]
["maxLength",    "1e5", 0, {"max":100}]
["maxSelections","1e5", 0, {"max":3}]
["minSelections","1e5", 0, {"min":1}]
["minLength",    "1e5", 0, {"min":0}]
```

Concretely: a `isLessThan 500` rule, user pastes `1e3` → the rule silently becomes "less than **100**", a threshold they never typed. Better than `1000`? Arguably, but it is still a made-up limit, and the new tests can't see it because they assert `parseRuleValue` in isolation. Worth either asserting `parseRuleValue` + `createRuleParams` together, or making the guard keep the previous value instead of returning `0`.


---

apps/web/modules/survey/editor/components/validation-rule-value-input.tsx
```diff
@@ -129,6 +129,12 @@ export const ValidationRuleValueInput = ({
       type={htmlInputType}
       value={currentValue ?? ""}
       onChange={(e) => onChange(e.target.value)}
+      // Browsers accept scientific notation in a number field, so `1e5` would silently store
+      // 100000 and a bare `e` would store 0 while the field still shows the typed text. `.` and
+      // `-` stay allowed: decimal and negative thresholds are valid per the rule schemas.
+      onKeyDown={(e) => {
+        if (config.valueType === "number" && ["e", "E", "+"].includes(e.key)) e.preventDefault();
```
kiln-agent, line 139: Minor: the guard looks only at `e.key`, so it also cancels modifier chords. On macOS, Ctrl+E (caret to end of line, honoured by Chrome/Safari in text fields) and Cmd+E fire `keydown` with `key === "e"` and get `preventDefault()`d; Ctrl/Cmd+`+` (zoom in) is swallowed too while this field has focus. The `response-options-card.tsx` precedent has the same gap, so this is inherited rather than introduced — but it is one condition to avoid.

```suggestion
      onKeyDown={(e) => {
        if (e.ctrlKey || e.metaKey || e.altKey) return;
        if (config.valueType === "number" && ["e", "E", "+"].includes(e.key)) e.preventDefault();
      }}
```


---

kiln-agent (commented):
Reviewed the diff for correctness. The core fix is sound: `applyRuleDeletion` / `shouldResetInputTypeToText` are correct, `handleDisable`'s rewrite is behaviour-preserving, the two sequential `updateElement` calls are safe (`elements-view.tsx` re-finds the element inside the `setLocalSurvey` updater), the delete button is always rendered so the fixed path is genuinely reachable, and the claim that `handleInputTypeChange` cannot empty the rule list holds (every `RULES_BY_INPUT_TYPE` entry is non-empty). Helper suite passes locally (55/55).

Two findings, both on the number-input half of the PR — see inline comments.

---

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
kiln-agent: **The `0` sentinel does not survive `createRuleParams`.**

`0` is falsy, and `createMinParams`/`createMaxParams` (`validation-rules-utils.ts:180-186`) are `Number(value) || defaultValue`. So the rejected value is not stored as `0` — it is stored as the rule's *default*:

I ran the real chain (`parseRuleValue` → `createRuleParams`) with `"1e5"`:

| rule | parsed | stored params |
| --- | --- | --- |
| `maxValue` | `0` | `{ max: 100 }` |
| `maxLength` | `0` | `{ max: 100 }` |
| `isLessThan` | `0` | `{ max: 100 }` |
| `minSelections` | `0` | `{ min: 1 }` |
| `minValue` | `0` | `{ min: 0 }` ✓ |

Concrete scenario: a user has `maxValue: 5000` and pastes `1e5`. The rule silently becomes `max: 100` — tighter than both what they had and what they pasted, and with no feedback. Only the `min`-family rules with `defaultValue === 0` actually land on `0`.

The new tests can't catch this because they assert `parseRuleValue` in isolation, which is exactly where the invariant holds. The PR description's claim ("treats anything else as `0`") is false for every max-family rule and for `minSelections`/`minRanked`/`minRowsAnswered`.

(Same falsy-default collision is why clearing a `maxValue` field to retype it snaps back to `100` — pre-existing, but it's the same root cause and worth fixing together, e.g. by having these helpers use `Number.isFinite(n) ? n : defaultValue`.)

---

apps/web/modules/survey/editor/components/validation-rules-editor.tsx
```diff
@@ -146,8 +140,18 @@ export const ValidationRulesEditor = ({
...
+    // Deleting the last rule leaves the section reading as "validation off" (isEnabled is derived
+    // from the rule count), so it has to reset inputType exactly like handleDisable does.
+    const { rules, resetInputTypeToText } = applyRuleDeletion(
+      validationRules,
+      ruleId,
+      elementType,
+      inputType
+    );
+    onUpdateValidation({ rules, logic: validationLogic });
+    if (resetInputTypeToText) {
```
kiln-agent, line 152: **This makes deleting a rule able to break saved conditional logic, with a save error that points somewhere else.**

The invariant itself is right, but the `number → text` flip has a consequence outside this component that `handleDisable` (a deliberate, explicit action) could get away with and a trash-icon click can't.

Scenario: an OpenText element with `inputType: "number"`, one `minValue` rule, and a block whose logic uses that element with `isGreaterThan` (or a `calculate` action assigning it to a number variable). The user deletes the single rule. `onUpdateInputType("text")` fires, and on save `ZSurvey`'s superRefine rejects the survey:

- `isInvalidOperator` (`packages/types/surveys/types.ts:2985`) allows only `equals`/`doesNotEqual`/`contains`/… for non-number OpenText — `isGreaterThan` is now invalid.
- the calculate-action check (`types.ts:3645`) rejects an OpenText with `inputType !== "number"` as a number-variable source.

Both messages name the *logic index of another block*, so the user sees "Conditional Logic: Invalid operator … in logic no: 2 of block 3" after deleting a validation rule in block 1, with nothing connecting the two.

Worth either warning before the reset when `inputType` actually changes, or cleaning up dependent logic alongside it. Note `handleInputTypeChange` has the same gap on `main`, so a shared cleanup would cover all three paths.

---

apps/web/modules/survey/editor/components/validation-rule-value-input.tsx
kiln-agent, line 139: **`-` is allowed through here, but a negative value still can't be typed into this field.**

The regex change deliberately keeps `-` and the new tests assert `-3` / `-2.5`, but this input still has `min={0}` two lines down and is a *controlled* `type="number"` field, so the intermediate state can't survive:

1. User focuses the field (say it holds `0`) and presses `-`.
2. Per the HTML value-sanitization algorithm `-` is not a valid floating-point number, so `e.target.value` is `""`.
3. `parseRuleValue("")` → `0`; the stored param doesn't change.
4. React's controlled-input write-back (it explicitly handles `value === 0 && node.value === ''`) sets the DOM value back to `"0"`.

The minus sign disappears; typing `10` then yields `10`, not `-10`. Worse on a `maxValue` field: the write-back restores `100`, so the field visibly jumps.

So the `-3`/`-2.5` cases the tests assert are reachable only by paste. I'd either drop `min={0}` for the signed rules and let the field hold a transient `-` (parse it as "unchanged" rather than `0`), or drop `-` from the regex and stop advertising negative support — right now the code claims one thing and the UI does another.

---

kiln-agent (commented):
Reviewed the diff for correctness. The `applyRuleDeletion` refactor is sound — I traced the two sequential `updateElement` calls through `elements-view.tsx` (functional updater re-finds the element by id, so no clobbering) and confirmed the third path, `handleInputTypeChange`, really can't leave an empty rule list. Types and the new tests check out locally (`tsc -p apps/web` clean on the touched files, 55/55 passing).

Three findings, one of which I think blocks: the `0` sentinel in `parseRuleValue` doesn't survive `createRuleParams`, so pasted scientific notation lands on the rule's *default* (100 / 1) rather than 0. Details inline.

---

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
kiln-agent: Superseded — this finding is stated more completely in [the later thread on the same line](https://github.com/formbricks/formbricks/pull/8993#discussion_r3861196008) (my review tooling posted twice). Resolving this one; please follow the other. Apologies for the duplicate.

---

apps/web/modules/survey/editor/components/validation-rule-value-input.tsx
Dhruwang: Good catch — fixed in 55825de, taking your suggestion as written.

```tsx
onKeyDown={(e) => {
  if (e.ctrlKey || e.metaKey || e.altKey) return;
  if (config.valueType === "number" && ["e", "E", "+"].includes(e.key)) e.preventDefault();
}}
```

Agreed it's inherited rather than introduced, but it's one condition and the chords are real, so no reason to carry it forward. Left `response-options-card.tsx` alone — same gap, but out of this PR's scope and not worth widening the diff for.

---

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
Dhruwang: Confirmed, and thanks — I reproduced the full chain rather than take the table on trust, and it matches yours exactly. One extra data point worth having:

```
maxLength "1e5" -> parsed=0 -> {"max":100}    ← the symptom you reported
maxLength "0"   -> parsed=0 -> {"max":100}    ← same result from a literal typed 0, on main today
```

So the falsy-`0` collision is **pre-existing and already reachable without this PR** — anyone typing `0` into a max-family field gets the default. My change doesn't introduce it; it routes more inputs into it.

**Done in 55825de:** added `describe("parseRuleValue composed with createRuleParams")`, asserting the pair together across all ten numeric rule types with their real stored params (`{max:100}`, `{max:3}`, `{min:1}`, `{min:0}`), plus a row pinning the literal-`0` behaviour so a future fix to the fallback shows up as a deliberate diff. Your point that the isolated tests couldn't see this was exactly right. Also corrected the PR description, which overstated the contract.

**Not done here, deliberately:** the root fix you propose (`Number.isFinite(n) ? n : defaultValue`) is the right shape, but it changes behaviour for a *legitimately typed* `0` across ten rule types — `maxLength: 0`, `maxSelections: 0`, `minSelections: 0` all become storable where today they silently become the default. Whether "max 0" is a value the product should accept isn't mine to decide inside a bug-fix PR, and it isn't what ENG-2419 asked for. Proposed patch for whoever picks it up:

```ts
const toNumberOr = (value: number | string | undefined, defaultValue: number) => {
  const parsed = Number(value);
  return value === undefined || value === null || value === "" || !Number.isFinite(parsed)
    ? defaultValue
    : parsed;
};
const createMinParams = (value?: number | string, defaultValue = 0) => ({ min: toNumberOr(value, defaultValue) });
const createMaxParams = (value?: number | string, defaultValue = 100) => ({ max: toNumberOr(value, defaultValue) });
```

Raised with the author to file separately. The guarantee ENG-2419 actually asked for — that the pasted magnitude never survives — holds either way, and is now asserted.

---

apps/web/modules/survey/editor/components/validation-rules-editor.tsx
Dhruwang: Verified both halves against `packages/types/surveys/types.ts` rather than taking them on trust, and you're right on both:

- `isInvalidOperator` — the OpenText switch restricts `email`/`phone`/`text`/`url` to the string operator set (`equals`, `contains`, `startsWith`, …), so `isGreaterThan` does become invalid the moment `inputType` flips to `text`.
- the calculate-action check — an OpenText with `inputType !== "number"` is rejected as a number-variable source, and the message is `"… in logic no: N of block M"`, naming a block other than the one the user was editing. The disconnect you describe is real.

Where I land: **the hazard is real but not introduced by this PR, and the invariant is still the right fix.** `handleDisable` already performs the identical `number → text` flip on `main` (toggling the Validation rules section off), and `handleInputTypeChange` does too when the dropdown moves off `number`. Both reach exactly the same broken-logic state with exactly the same misdirected save error. This PR adds a third path to a hazard that already has two — and the reason it adds it is that the state it replaces (zero rules with `inputType: "number"`) is itself a bug the ticket exists to remove.

So I'm not folding a fix in here. What you're describing — warn before the reset, or clean up dependent logic alongside it — is a product decision about what should happen to someone's existing logic, and it needs to cover all three paths to be worth anything. Doing it in a two-bug fix PR would be a unilateral design change on the author's behalf.

Recorded under Risks in the description, and raised with the author to file as its own ticket covering `handleDisable` / `handleDeleteRule` / `handleInputTypeChange` together. If they'd rather block this PR on it, that's their call and I'll pick it up.

---

apps/web/modules/survey/editor/components/validation-rule-value-input.tsx
Dhruwang: You're right that the code and the description disagree, and the description is the wrong half. Conceded.

Tracing it: for `type="number"` the HTML value-sanitization algorithm returns `""` for anything that isn't a valid floating-point number, and a lone `-` isn't one — so `e.target.value` is `""` before `min={0}` even enters the picture, `parseRuleValue("")` gives `0`, and React's controlled write-back restores the DOM value. The `-` can't persist as an intermediate state. I can't demonstrate the keystroke sequence end-to-end from here (no browser in this environment — the gap already recorded in the description), so I'm treating your account as correct on the mechanism rather than arguing it.

What that means for the diff: `-?` in the regex is still doing real work — a **pasted** `-3` is honoured instead of being zeroed, and that's the path the tests exercise. What was wrong is my description, which implied typing a negative works. Corrected: the `-3` / `-2.5` cases are now labelled paste-only, and the "negative thresholds are valid" line now says valid *per the rule schemas*, not reachable by typing.

I'm not taking either of your two options in this PR:

- **Dropping `min={0}` for signed rules** is explicitly out of scope on ENG-2419 — the ticket lists it under "Out of scope (worth separate tickets)", verbatim: "`min={0}` is applied to every number rule value input, including `minValue`, where negative thresholds should be allowed". It already has a home; it shouldn't be smuggled in here.
- **Dropping `-` from the regex** would make a pasted `-3` store `0`, which is a real regression for the one path that does work, and it contradicts the ticket's own explicit "do **not** block `-`" instruction.

So the honest end state is: negatives work by paste, not by typing; the schemas allow them; the field doesn't; and closing that gap is the `min={0}` ticket's job. That's now what the description says.

---

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
Dhruwang: One more change landed on this exact line in 5f82d40231, unrelated to the `0`-sentinel point above but worth noting here since it rewrites the guard: Sonar flagged `^-?\d*\.?\d*$` under S8786 (MAJOR) — a digit run can be split between the two `\d*` around the optional `.`, so a long non-numeric paste backtracks quadratically (measured 6ms at 2k chars, 24ms at 4k, 203ms at 8k). It is now `^-?\d*(\.\d*)?$`, where the group can only start at a literal `.`, so shortening `\d*` opens no new path: 0.011 / 0.021 / 0.044ms for the same inputs. Accepted language is unchanged — verified exhaustively over the 66430 strings of length <= 5 from `[- + 0 1 9 . e E space]`, zero divergence — so the parse behaviour your table describes is untouched, `0` sentinel included.

Resolving this thread: the tests you asked for (`parseRuleValue` composed with `createRuleParams`) are in, and the root `Number.isFinite` fix stays out of this PR for the reason given above.

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
```diff
@@ -1,11 +1,59 @@
...
+ */
+export const shouldResetInputTypeToText = (
+  elementType: TSurveyElementTypeEnum,
+  remainingRuleCount: number,
+  inputType?: TSurveyOpenTextElementInputType
+): boolean =>
+  elementType === TSurveyElementTypeEnum.OpenText &&
+  remainingRuleCount === 0 &&
+  inputType !== undefined &&
+  inputType !== "text";
```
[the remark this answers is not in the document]
Dhruwang: Confirmed and fixed in `ac6ba09`. I checked every link before changing anything, and all of them hold:

- `app/lib/templates.ts:139` — `buildOpenTextElement({ inputType: "email", longAnswer: false, placeholder: "example@email.com" })`, same at `281` and `2431`, and `grep -c "validation:"` on that file is **0**. So zero rules with a non-`text` inputType is shipped data.
- `open-element-form.tsx:144-148` — `onUpdateInputType` writes `longAnswer: newInputType === "text"`, so the reset does flip `longAnswer` to true.
- `validation-rules-editor.tsx:152-154` — the delete path calls it.
- `main`'s `handleDeleteRule` is a bare `filter` with no reset, while `main`'s `handleDisable` already clears any non-text type. Your attribution is exactly right: this PR added the downgrade to a path that never had it.

The point that settled it for me was `open-element-form.tsx:130` — `disabled={element.inputType !== "text"}`. Long answer is disabled for **every** non-text type, so for a template email question that disabled toggle is the correct state, not the ENG-2419 symptom. That makes "zero rules ⇒ text" false as a general invariant, which is what I had encoded. `number` is the one case where the leftover type is vestigial: it was set by adding a number rule, and it keeps Long answer disabled with nothing left enforcing it.

**One deviation from your suggested fix.** You proposed narrowing `shouldResetInputTypeToText` itself. I narrowed only the deletion path instead, because that helper is shared with `handleDisable` — this PR refactored `handleDisable` onto it — and `main` clears any non-text type there deliberately, with a comment about the input otherwise keeping browser-native format validation after the user switched validation off. Narrowing the shared helper would have silently reversed that shipped behaviour as a side effect of a refactor. So:

```ts
resetInputTypeToText:
  inputType === "number" && shouldResetInputTypeToText(elementType, remaining.length, inputType),
```

Whether the *disable* path should also stop downgrading email/url is a real question — it does the same damage to the same template, just via the toggle — but it is main's behaviour and not this PR's to decide. Happy to file it.

Tests pin both sides: three cases asserting `email`/`url`/`phone` survive a deletion, and three asserting `shouldResetInputTypeToText` still clears them so a future change to `handleDisable` shows up as a deliberate diff. Reverting the narrowing reddens exactly the first three and leaves the second three green. 450 passed in `modules/survey/editor`, typecheck clean, 0 new lint errors.

---

apps/web/modules/survey/editor/lib/validation-rules-helpers.ts
```diff
@@ -1,11 +1,59 @@
...
+export const applyRuleDeletion = (
+  rules: TValidationRule[],
+  ruleId: string,
+  elementType: TSurveyElementTypeEnum,
+  inputType?: TSurveyOpenTextElementInputType
+): { rules: TValidationRule[]; resetInputTypeToText: boolean } => {
+  const remaining = rules.filter((rule) => rule.id !== ruleId);
+  return {
+    rules: remaining,
+    resetInputTypeToText: shouldResetInputTypeToText(elementType, remaining.length, inputType),
```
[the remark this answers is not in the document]
Dhruwang: Correct on both rows, and the diagnosis of *why* it read as checked is the useful part.

Verified: `:278` opens `describe("shouldResetInputTypeToText")` and `:280` is its `test.each`, which calls the helper directly — so a mutation inside `applyRuleDeletion` cannot reach it. The four that actually go red are in `describe("applyRuleDeletion")` at `:306`, and the count coinciding at four is exactly what let me believe I had checked it. That is a genuine hole in my own process, not just a wrong line number: I ran the mutation, saw four red, and matched the count instead of the names.

Row 4 likewise named `:386` (the `test("the pasted magnitude never reaches the stored rule")`) while its Outcome described the `test.each` above it.

Both rows are now rewritten against `ac6ba09`, which renumbers the file anyway, and each `unit (mutation)` row names the test *and* the production `file:line` to edit. The reset rows also had to change substance, not just line numbers — the behaviour narrowed to `number`, so the old "all four inputTypes reset" claim is gone.

---

Dhruwang:
**CI note: `Better Auth integration tests` and `Run E2E Tests` failed on `ac6ba09` because this branch was 81 commits behind `main`, not because of the code. Fixed by merging `main` in — `79df135`.**

Diagnosed to step level rather than assumed. Both jobs died on the same step, before any test ran:

```
13  Start AuthZed CI fixture   failure   08:50:03 → 08:50:03
14  Build workspace package dependencies   skipped
…
18  Run Better Auth integration tests      skipped
```

The cause is a mismatch between the workflow and the tree:

| | AuthZed fixture |
| --- | --- |
| `main` | `scripts/start-authzed-ci.sh` present; `integration-tests.yml:127` runs it |
| this branch at `ac6ba09` | script absent; its own workflows never mention AuthZed |

So the AuthZed cutover landed on `main` after this branch's last green run (Aug 27), and CI was invoking a step whose script this tree didn't contain. Nothing to do with the `number`-narrowing commit, which touches two editor helper files.

`main` is now merged in (clean, no conflicts, 81 commits). I reinstalled and force-rebuilt the dependency graph before re-checking, because the same drift bites locally in the opposite direction — my first typecheck after the fix failed in `modules/survey/lib/survey.ts` and `modules/survey/link/lib/data.ts` over `isSingleResponsePerEmailEnabled`, a field `main` removed in #8975, purely because my build graph was stale. Rebuilt from this tree: clean.

On the merged tree: 450 passed in `modules/survey/editor`, typecheck exit 0, 0 lint errors, format clean.

I'd also said earlier on this PR that the merge queue made a `main` merge unnecessary here. That was wrong — the queue tests the merged result at merge time, but it can't make the branch's *own* CI green when the workflow needs files only `main` has.

---

itsjavi (dismissed):

Dhruwang:
@coderabbitai — the non-finite finding is a true positive on one of its two halves, and the half that holds is worse than described. Fixed in `c8b44dc`. Replying here because the finding was posted outside the diff range.

**Verified through the real `parseRuleValue` → `createRuleParams` chain** rather than reasoned about:

```
"-"            -> parsed=0         stored {"max":100}
"."            -> parsed=0         stored {"max":100}
"-."           -> parsed=0         stored {"max":100}
"<400 digits>" -> parsed=Infinity  stored {"max":null}
```

- **`Infinity` — confirmed, and it stores `null`, not `Infinity`.** The regex admits an arbitrarily long digit run; `Number()` overflows past ~1e308; `Infinity` is truthy so `Number(value) || 0` passed it, and `createMaxParams` is truthy-guarded so it reached the stored rule. `JSON.stringify(Infinity)` is `null`, so the persisted param is `null` against a `z.number()` — an invalid value rather than merely an ineffective maximum.
- **`NaN` — already handled, so that half of the finding does not hold.** `"-"`, `"."` and `"-."` give `NaN`, and `NaN || 0` was already `0`. They never reached the downstream helper as `NaN`. I've pinned them in tests anyway so the finite check is not credited with more than it does.

Fix is `Number.isFinite` as you suggested, which covers both without relying on falsiness:

```ts
const parsed = Number(value.trim());
return Number.isFinite(parsed) ? parsed : 0;
```

Reverting it reddens exactly the two Infinity cases and leaves the `-`/`.`/`-.` cases green — which is the evidence for the split above.

**On the Coverage warning:** the deletion rows were already rewritten for the `number`-only behaviour before this run, so that part read a stale snapshot — but the two things it asked for beyond that were fair and are now in: a row for the non-finite parse, and the linear-time regex, which I've put under Open gaps rather than Coverage because **no test measures time**. Accepted-language equivalence is verified exhaustively (66,430 strings, zero divergence) and that is what the parse tests cover; a future regression to a quadratic form would still pass them. Sonar S8786 is the real gate, and saying so is more honest than a row implying I assert the complexity.

455 passed, typecheck exit 0, 0 lint errors, format clean.

---

itsjavi (approved):

PR #8994, opened 2026-08-26
fix: use multi-arch Valkey image (backport 5.4)

BhagyaAmarasinghe:
Ref ENG-2575

## What & why

- Backport [#8986](https://github.com/formbricks/formbricks/pull/8986) to `release/5.4` so the bundled
  Valkey image selects native `linux/amd64` and `linux/arm64` children instead of retaining the old
  amd64-only digest.
- Carry the reviewed one-click migration, synchronization tests, remote OCI-index guard, and self-hoster
  upgrade guidance onto the release line.
- Apply the eight runtime pins that exist on `release/5.4`. The main-only API v3 workflow is not introduced.
- The branch is three commits ahead of the current release tip `a589991a1` and contains no unrelated changes.

## Breaking changes

| Bundled Valkey image | Debian Valkey 8.1.1 amd64 child digest | Alpine Valkey 8.1.9 multi-architecture index | Existing self-hosters whose saved Compose file or Helm override retains the old digest | Refresh the one-click updater or replace the documented digest; keep the existing Redis volume/PVC |

## Migrations & env

- No database migrations or environment-variable changes.
- Fresh Docker and Helm installations use the new index automatically.
- Existing one-click installations must download the current updater from the directory containing their
  `formbricks/` folder before running `update`; manual Compose and Helm deployments with an explicit/reused
  digest must follow the v5.4 migration guide.
- The updater never removes the named Redis volume or flushes Valkey data.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Every applicable 5.4 deployment and CI surface uses one canonical index | unit (mutation): `./node_modules/.bin/vitest run docker/__tests__/valkey-image.test.ts` | The eight release runtime pins, Helm README and updater matched the 8.1.9 index; the updater predecessor was explicit; the absent main-only API v3 workflow was not required |
| One-click migration changes only the known old pin and fails closed | unit (mutation): `./node_modules/.bin/vitest run docker/__tests__/formbricks-script.test.ts` | Exact old pin migrated under standard and four-space indentation; backup and restricted file mode were preserved; current/custom refs were no-ops; failed validation restored the original file |
| Release Compose files render valid configuration | manual: `docker compose config` for production and development Compose | Both configurations validated with the exact Valkey index |
| Release Helm chart renders the immutable index | manual: `helm lint` and `helm template` | Lint passed and the rendered StatefulSet used `valkey/valkey@sha256:e0eb7c480958d32bdc4357a74bdd70653ae15f2f9b4c93c4a5a9fad1dc471c84` |
| Runtime and persisted-data compatibility | manual on source PR #8986 | Native ARM64 stack boot, internal app connectivity, cross-image AOF persistence and exact-child vulnerability scans passed on the identical runtime/updater change |

**Open gaps**

- CI on this PR is the release-base check that is not shared with #8986; wait for it before merging.
- As on #8986, no disposable Kubernetes cluster was available for a live Helm PVC upgrade, and authenticated
  login/survey-response journeys remain release QA.

**Risks**

- The runtime base changes from Debian to Alpine. The source PR passed a real cross-image AOF upgrade and
  native ARM64 stack boot, but platform-specific behaviour should still be watched during the 5.4 release.
- The updater intentionally recognizes only the exact old bundled digest. Operator-supplied custom image
  references remain untouched and must be updated manually.

---

> [!NOTE]
> **AI model used** — `unknown` (Codex), reasoning effort `unknown`.


kiln-agent (commented):
## Review — backport verified, no blocking issues

I re-verified the substance rather than trusting the source PR, since the risky parts here (registry facts, the `awk` rewriter, cross-image persistence) are cheap to check directly.

### Backport fidelity ✅

13 files, and every patch hunk is byte-identical to #8986 except the one file that legitimately doesn't exist on this line (`.github/workflows/api-v3-contract-tests.yml`), which `valkey-image.test.ts` handles via the `existsSync` guard. No unrelated changes. The legacy digest survives in exactly three intentional places — `legacy_valkey_image=` in the updater and the two test constants — and nowhere else:

```
docker/formbricks.sh:5                     legacy_valkey_image=...
docker/__tests__/formbricks-script.test.ts:12
docker/__tests__/valkey-image.test.ts:11
```

### The premise checks out against the registry ✅

Queried Docker Hub directly:

| | mediaType | platforms |
| --- | --- | --- |
| old `sha256:12ba4f45…` | `manifest.v2+json` (no `manifests` array) | single-platform, amd64 |
| new `sha256:e0eb7c48…` | `oci.image.index.v1+json` | linux/amd64, linux/arm64, linux/arm/v7, linux/ppc64le |

`sha256:e0eb7c48…` is exactly what `valkey/valkey:8.1.9-alpine` resolves to today. The docs' security claim also holds — 8.1.4 carried CVE-2025-49844 and the 46817/46818/46819 set, none of which are in 8.1.1.

### Cross-image persistence — verified here, not just on #8986 ✅

Wrote keys under Debian 8.1.1 with `--appendonly yes`, `BGREWRITEAOF`, then restarted Alpine 8.1.9 on the same named volume: keys read back, new writes and a further `BGREWRITEAOF` succeeded, no permission errors in the log.

One real difference worth recording, since it isn't obvious from the diff: **the container GID changes 999 → 1000** (Debian `uid=999(valkey) gid=999`, Alpine `uid=999(valkey) gid=1000`). This is benign — the **uid is unchanged**, so existing `0600`/`0700` AOF files still match on owner, which is why the upgrade works. Both images still drop privileges to uid 999 (`/proc/1/status` on both), so there's no root-runtime regression. Nothing to change; just don't let a future variant switch move the uid.

### Rewriter held up under adversarial input ✅

Against the real 5.4 `docker/docker-compose.yml` pinned back to the old digest, exactly one line changed and the rest of the file was identical. Edge cases either migrate correctly or fail safe: top-level `volumes:` with a `redis:` key, legacy pin under a non-`redis` service (no-op), no `services:` key at all (no-op), a column-0 comment inside `services:`, four-space indentation. The failure path restores from backup and leaves no `.tmp` behind.

Also confirmed `docker` is a real pnpm workspace with a `test` script, so both spec files actually run under `pnpm test` — 20/20 pass, and `prettier --check` and `bash -n` are clean.

---

### Non-blocking notes

**1. The new platform guard never runs on this branch.** `docker-build-validation.yml` is gated to `main`:

```yaml
on:
  pull_request:
    branches: [main]
  merge_group:
    branches: [main]
```

So `Verify Valkey Image Platforms` is inert on `release/5.4` — it isn't in this PR's 14 check runs, and it won't guard the release line going forward. Carrying it is still right (it keeps the files identical and it works if the branch is ever dispatched), but the `Open gaps` section credits CI coverage this branch doesn't have. I ran the step verbatim locally and it passes, so the substance is covered.

**2. Optional hardening — a trailing comment on the service key silently skips the migration.** The service-key match requires end-of-line after the colon:

```awk
in_services && /^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$/
```

So an operator whose Compose file reads `  redis: # cache` never sets `in_redis`, and `update` quietly leaves them on the amd64-only pin with no message — the one silent-skip path that isn't a deliberate no-op. It fails safe, and the mirror case (a commented service key *after* `redis:`) is contrived enough to ignore. If you want to close it, allowing an optional comment in both copies of that regex is a one-line change per `awk` block. Fine to leave as-is for a backport.

Nothing here blocks merge. Worth waiting on the in-progress E2E run, as the description says.

---

kiln-agent:
## Review — backport of #8986 to `release/5.4`

Reviewed at `1ee7f77`. The backport is faithful and the substance holds up. I re-ran the claims rather than trusting the source PR, and everything checked out. Three low-severity notes below; none blocks merge.

### What I verified independently

| Claim | How | Result |
| --- | --- | --- |
| New digest is a multi-arch index, old one is not | `docker buildx imagetools inspect --raw` on both | New: `application/vnd.oci.image.index.v1+json` with `linux/amd64`, `linux/arm64` (plus `arm/v7`, `ppc64le`). Old: a single `vnd.docker.distribution.manifest.v2+json` — amd64-only, as described |
| It is Valkey 8.1.9 Alpine | `docker run … valkey-server --version` + `/etc/os-release` | `v=8.1.9`, `ID=alpine` |
| No pin was missed | `git grep` of both digests across the branch | Legacy digest survives only as the deliberate `legacy_valkey_image` and the two test constants. All 3 workflow service containers, both Compose files, `formbricks.sh`, `migrate-to-v4.sh`, Helm `values.yaml` + README carry the index — the "eight runtime pins" count is right |
| The updater migrates real installs | Ran `migrate_legacy_valkey_image` against the bundled `docker/docker-compose.yml` from tags `5.0.0`, `5.1.0`, `5.2.0`, `5.3.4` | All four: exactly one replacement, backup written, `docker compose config` passes. (5.0/5.1 need the installer-generated `.env` for `CUBEJS_API_SECRET`; since `update_formbricks` does `cd formbricks` first, Compose picks it up from the project dir — correct as written) |
| Persisted data survives Debian → Alpine | Wrote a key under 8.1.1 with AOF on a named volume, `bgrewriteaof`, then started 8.1.9 on the same volume | Key read back intact, new writes succeed, `/data` stays `999:999`. Note the image's *group* moves 999 → 1000, but access is by uid 999 so owner perms apply, and `redis.master.podSecurityContext`/`containerSecurityContext` are both `{}` — no `fsGroup`/`runAsUser` assumption to break |
| Alpine has no `bash` | Checked every redis-adjacent invocation | StatefulSet `command`, both probes and the Compose healthcheck use `/bin/sh` + `valkey-cli` + `grep`, all present in the Alpine image. `formbricks.redisImage` renders `repository@digest` and ignores `tag` |
| The new CI guard's logic works | Ran the step's pipeline locally against this branch | `docker compose -f docker/docker-compose.yml config --format json \| jq -r '.services.redis.image'` resolves cleanly — no required-var interpolation in that file — and the `jq -e` assertion passes |
| Tests | `vitest run docker/__tests__/valkey-image.test.ts docker/__tests__/formbricks-script.test.ts` | 20/20 pass |

Diffed against #8986 hunk by hunk: identical, minus the main-only `api-v3-contract-tests.yml` pin, plus the `existsSync` guard that lets the synchronization test skip it. Faithful.

### Notes

**1. The OCI-index guard cannot run on this base — worth adding to `Open gaps`.** `docker-build-validation.yml` triggers only on `pull_request`/`merge_group` targeting `main`, so `Verify Valkey Image Platforms` never executes on `release/5.4`. Carrying it for line parity is right, and I confirmed the logic works, but on this PR the only thing exercised is that the YAML parses — and it is absent from this PR's check runs. `Open gaps` currently reads as though release CI covers this PR's delta.

**2. `migration.mdx` One-Click tab omits the working directory that `one-click.mdx` adds in the same diff.** `update_formbricks` starts with `cd formbricks` under `set -e`, so a reader who follows the migration tab from *inside* `formbricks/` gets a bare `cd` failure. `one-click.mdx` now says "first go to the directory that contains your `formbricks/` folder" — one sentence in the tab would align them.

**3. Nit — `valkey-image.test.ts` asserts presence, never absence.** A file carrying *both* pins would pass. `docker/formbricks.sh` legitimately holds both so it needs excluding, but for the other seven surfaces a `not.toContain(legacyValkeyImage)` closes the one hole in a test whose whole purpose is synchronization.

**4. Nit, non-blocking — two silent no-ops in the awk detection.** If the operator restructured their Compose file so that `services:` carries a trailing comment (`services: # app services`), or the first service key contains a character outside `[A-Za-z0-9_-]` (e.g. `my.app:`, which makes `service_indent` latch onto a nested key), detection returns 0 with no output and `update` proceeds on the amd64-only image. I confirmed both: `rc=0`, no message, legacy pin retained. Neither applies to any bundled Compose file we ship, and it fails open rather than corrupting anything — but since the whole point is that operators need not notice this upgrade, an `echo` on the "redis service found, pin not recognized" path would be cheap insurance.

### Assessment

Approach and blast radius look right: the updater recognizes only the exact old bundled digest, backs up, validates, and restores on failure; custom references are untouched; the volume is never removed. Breaking-changes checkbox and table match the diff. E2E was still in progress at review time — worth confirming green before merge.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+    return 1
+  fi
+
+  echo "✅ Updated bundled Valkey to the native amd64/arm64 image. Backup: $backup_file"
+}
+
 update_formbricks() {
   echo "🔄 Updating Formbricks..."
   cd formbricks
+  migrate_legacy_valkey_image docker-compose.yml
```
kiln-agent, line 1064: This is a bare command under the script-wide `set -e`, so every `return 1` path in `migrate_legacy_valkey_image` now aborts the entire update before `pull`/`down`/`up -d` ever run.

Reproduced locally with a stubbed `sudo`:

- Compose file named `docker-compose.yaml` (or `compose.yaml`) — both auto-discovered by `docker compose`, so `update` worked before this PR. Now the `[[ ! -f docker-compose.yml ]]` guard fires:
  ```
  🔄 Updating Formbricks...
  ❌ Cannot update Valkey because docker-compose.yml does not exist.
  outer rc=1
  ```
  No pull, no restart — `./formbricks.sh update` silently becomes a no-op.
- `docker compose config` failing for any reason: the Compose file is restored to the legacy pin **and** the whole Formbricks upgrade is skipped, with only the Valkey-specific ❌ line printed.

Two problems: the operator gets no "update aborted" signal, and this contradicts the docs added in this PR (`one-click.mdx`: "validates the result before pulling"; `migration.mdx`: "restores the backup if validation fails"), both of which read as though the update then proceeds.

Suggest making the optional migration non-fatal and letting `docker compose pull` be the thing that reports a genuinely broken file:

```suggestion
  migrate_legacy_valkey_image docker-compose.yml || true
```


---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+      service_indent=-1
+      next
+    }
+    in_services && /^[^[:space:]#]/ {
+      in_services=0
+      in_redis=0
+    }
+    in_services && /^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$/ {
+      line_indent=indentation($0)
+      if (service_indent < 0) service_indent=line_indent
```
kiln-agent, line 1008: `service_indent` is inferred from the *first* line inside `services:` matching `^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$` — but that pattern requires nothing after the colon, so a service key with trailing content is skipped and a **nested** key wins the race instead.

Concrete case: an operator has a comment or anchor on the first service key.

```yaml
services:
  postgres: # main db          <- does not match (content after colon)
    image: pgvector/pgvector:pg18
    volumes:                   <- matches first, service_indent = 4
      - postgres:/var/lib/postgresql
  redis:                       <- indent 2 != 4, so in_redis never set
    image: valkey/valkey@sha256:12ba4f45…
```

Verified against both awk blocks: detection exits 1, `migrate_legacy_valkey_image` takes the `return 0` early exit, rc=0, **nothing is printed**, and the file still carries the amd64-only digest. `postgres: &db` behaves identically. So the exact hosts this PR targets (arm64) can keep the old image with no indication anything was skipped.

Two suggestions:

1. Derive the indent from the `redis:` line itself (or accept any bare mapping key whose indent is the minimum seen inside `services:`) rather than from whichever line happens to match first.
2. Make the "not found" path distinguishable from "already migrated": a `grep -q "$legacy_valkey_image" "$compose_file"` before `return 0` would show the legacy digest *is* present, and warrants a warning telling the operator to follow the manual step in `migration.mdx` instead of exiting silently.


---

kiln-agent (commented):
Reviewed the full backport diff. The digest swap itself checks out — I pulled both manifests from Docker Hub: the new `sha256:e0eb7c48…` is a genuine `application/vnd.oci.image.index.v1+json` with linux/amd64 + linux/arm64 (+arm/v7, ppc64le) children at `VALKEY_VERSION=8.1.9`, while the old `sha256:12ba4f45…` is a bare `manifest.v2` (amd64 only), so the CI `jq -e` assertion in `docker-build-validation.yml` will hold. Entrypoint/cmd/workdir are identical across the two images, so the `redis:/data` AOF volume and the Helm `/bin/sh -c exec valkey-server …` override both stay valid. All eight runtime pins are consistent — no stale `12ba4f45` references remain outside the updater's intentional `legacy_valkey_image` and the tests. `docker compose -f docker/docker-compose.yml config --format json | jq -r '.services.redis.image'` resolves correctly locally, and both docker test files pass (20 tests) from the `docker` workspace.

Two issues in `migrate_legacy_valkey_image`, both reproduced locally and commented inline.

---

PR #8996, opened 2026-08-26
fix: use multi-arch Valkey image (backport #8986 to release/5.4)

Dhruwang:
Ref ENG-2629

Backport of #8986 to `release/5.4`. Cherry-picked from `ebba370` (squash commit on `main`), trimmed to backport policy — minimal, single-purpose, no net-new tests. Replaces #8994, which was closed.

## What & why

- The bundled Valkey digest resolved to a `linux/amd64` child only, so native ARM64 self-hosted installs needed QEMU emulation to start. The replacement digest is an OCI index that selects the native amd64 or arm64 child.
- Updates the eight runtime pins that exist on this branch (3 workflows, both Compose files, the v4 migration script, the Helm default and its generated values table).
- Carries the one-click updater migration so an existing install still holding the old pin is actually moved onto the new one — backing the Compose file up, rewriting only that exact image reference, validating the result and restoring the backup if validation fails.
- Carries the self-hoster upgrade guidance for the one-click, Compose and Helm paths.

### Files changed (11)

| Area | Files |
| --- | --- |
| CI pins + platform guard | `.github/workflows/docker-build-validation.yml`, `e2e.yml`, `integration-tests.yml` |
| Compose pins | `docker-compose.dev.yml`, `docker/docker-compose.yml` |
| Helm | `charts/formbricks/values.yaml`, `charts/formbricks/README.md` |
| Scripts | `docker/formbricks.sh`, `docker/migrate-to-v4.sh` |
| Docs | `docs/self-hosting/advanced/migration.mdx`, `docs/self-hosting/setup/one-click.mdx` |

### Dropped relative to #8986

- `docker/__tests__/valkey-image.test.ts` — brand-new test file, excluded by the no-net-new-tests rule.
- The Valkey-migration cases added to `docker/__tests__/formbricks-script.test.ts` — net-new cases; this file is restored to its `release/5.4` state, so its six pre-existing tests are untouched.
- `.github/workflows/api-v3-contract-tests.yml` — does not exist on this branch (main-only, added by #8942).
- The build-cache `NOTE` comment `main` carries in `docker-build-validation.yml` — unrelated to this fix, and this branch still uses the gha cache it describes removing.

## Breaking changes

| Bundled Valkey image | Debian Valkey 8.1.1 amd64 child digest | Alpine Valkey 8.1.9 multi-architecture index | Self-hosters whose saved Compose file or Helm override still names the old digest | Re-download the one-click updater, or replace the digest as documented; keep the existing Redis volume/PVC |

## Migrations & env

- No database migrations and no environment-variable changes.
- Fresh Docker and Helm installs pick the new index up automatically.
- Existing one-click installs need the current updater downloaded into the directory holding their `formbricks/` folder before `update`. Manual Compose and Helm deployments that pin or reuse the digest follow the migration guide.
- The named `redis` volume and its AOF data are preserved; nothing in this diff removes a volume or flushes Valkey.

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The six pre-existing docker script/compose tests still pass after restoring the file to its release state | unit (guard): `pnpm exec vitest run docker/__tests__/formbricks-script.test.ts` | 1 file, 6 passed. The file asserts no Valkey digest, so it is decoupled from this change |
| Every runtime pin on this branch moved to the one index, and none was missed | manual: `git grep` for both digests across the tree | New index present in all 9 expected locations (8 runtime pins + Helm README); the only remaining old digest is `legacy_valkey_image` in `docker/formbricks.sh`, which is the pattern the updater matches on, by design |
| The rewritten updater and the v4 script are still valid shell | manual: `bash -n docker/formbricks.sh`, `bash -n docker/migrate-to-v4.sh` | Both parse clean |
| The conflict resolution left valid YAML and put the platform guard in the right place | manual: `yaml.safe_load` on all 3 workflows plus both Compose files and `values.yaml`; read the resolved hunk | All six parse; the `Verify Valkey Image Platforms` step sits between `Set up Docker Buildx` and `Build Docker Image`, with the gha cache on this branch left alone |
| Formatting matches the repo style | manual: `pnpm exec prettier --check` on the 9 non-shell changed files | "All matched files use Prettier code style!" |
| Runtime, cross-image AOF persistence and native ARM64 boot | manual, inherited from #8986 | Verified there against the identical runtime and updater change — native ARM64 stack boot, internal connectivity, AOF data written by the old amd64 child readable under 8.1.9, and both children clean on a Trivy scan by exact child digest |

**Open gaps**

- The dropped `valkey-image.test.ts` was what asserted pin synchronization automatically. On this branch that invariant is verified by the `git grep` row above and nothing enforces it going forward — re-add the file if you would rather have the guard than the smaller delta.
- Docker and Helm are unavailable in the environment this was prepared in, so `docker compose config`, `helm lint`/`template` and the live manifest inspection were not re-run here. CI's `Validate Helm Chart` job and the `Verify Valkey Image Platforms` step in `docker-build-validation.yml` are the gate on this PR; #8986 ran all three green on the same content.
- As on #8986: no disposable Kubernetes cluster for a live Helm PVC upgrade, and authenticated login/survey-response journeys remain release QA.
- CI on this PR is the release-base run that #8986 never had — wait for it before merging.

**Risks**

- The runtime base moves Debian → Alpine. #8986 passed a real cross-image AOF upgrade and a native ARM64 boot, but platform-specific behaviour is worth watching through 5.4 QA.
- The updater's 82 lines of new `awk`-based Compose rewriting ship here **without** the unit tests that covered them on `main`, because those tests were net-new and policy excludes them. It fails closed — it backs up first, requires exactly one replacement, validates with `docker compose config` and restores on failure — but if you would rather trade the minimal delta for that coverage, say so and I will bring the cases back.
- The updater deliberately recognises only the exact old bundled digest; operator-supplied custom images are left untouched and must be updated by hand.
- The v5.4 section of the migration guide still says the release needs "no database migrations". That claim is pre-existing on this branch and is not introduced here — correcting it is #8984's scope, still open on `main`.


BhagyaAmarasinghe (approved):

.github/workflows/docker-build-validation.yml
```diff
@@ -53,6 +53,20 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
 
+      - name: Verify Valkey Image Platforms
```
kiln-agent, line 56: This step never runs on this PR, so the invariant it guards is unenforced on `release/5.4`.

The workflow's triggers at the top of this file are:

```yaml
on:
  pull_request:
    branches: [main]
  merge_group:
    branches: [main]
  workflow_dispatch:
```

`backport/8986-to-5.4 → release/5.4` matches neither branch filter, and no other workflow calls this one. Combined with dropping `docker/__tests__/valkey-image.test.ts`, nothing on this branch verifies either that the digest is a multi-arch index or that the eight runtime pins stay in sync — the `git grep` in the description is a one-time manual check, not a gate.

The PR description lists this step (and `Validate Helm Chart`) as "the gate on this PR". `Validate Helm Chart` does run — `pr.yml` triggers `pull_request` unfiltered — but this one does not. Worth correcting the description so the next reader doesn't take it as covered, and either re-adding the dropped test or accepting the gap explicitly.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+      next
+    }
+    in_services && /^[^[:space:]#]/ {
+      in_services=0
+      in_redis=0
+    }
+    in_services && /^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$/ {
+      line_indent=indentation($0)
+      if (service_indent < 0) service_indent=line_indent
+      if (line_indent == service_indent) in_redis=($0 ~ /^[ ]+redis:[[:space:]]*$/)
```
kiln-agent, line 1009: **Service detection is stricter than the docs promise, and the miss is silent.**

`^[ ]+redis:[[:space:]]*$` requires the service key to be alone on its line. A single trailing comment defeats it:

```yaml
  redis: # cache + sessions
    image: valkey/valkey@sha256:12ba4f45…
```

Detection then returns non-zero, the caller takes the `return 0` "nothing to do" path, and the legacy amd64-only pin is left in place — with no output at all. The operator sees a clean `update` and a doc (`one-click.mdx`) that says the updater "recognizes the old amd64-only bundled Valkey digest", so they have every reason to believe they're migrated when they aren't. That's worse than the loud failure above, because it's indistinguishable from "already up to date".

Two things worth doing, in order of value:

1. Distinguish *not present* from *not matched*. If the legacy digest string appears anywhere in the file but the structured scan didn't claim it, say so explicitly and point at the manual steps in `migration.mdx` — a `grep -q "$legacy_valkey_image"` fallback is enough.
2. Relax the key match to tolerate a trailing comment (`/^[ ]+redis:[[:space:]]*(#.*)?$/`), and apply the same to the generic service-key pattern above it so a commented sibling key doesn't leave `in_redis` stale.

Unmodified one-click installs are matched correctly (the bundled file has the comment on its own line above `redis:`), so this is about hand-edited files — which, on a one-click stack that operators do edit for ports and volumes, is a realistic share of them.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+    }
+    in_redis && /^[[:space:]]+image:[[:space:]]*/ && index($0, legacy_image) {
+      sub(legacy_image, replacement_image)
+      replacements++
+    }
+    { print }
+    END { exit(replacements == 1 ? 0 : 1) }
+  ' "$compose_file" >"$temp_file"; then
+    rm -f "$temp_file"
```
kiln-agent, line 1045: Minor: the backup is written at line 1017, before this awk pass, but the failure path only removes the temp file. So `docker-compose.yml.before-valkey-8.1.9` is left on disk next to a message that says the original is unchanged — true of the Compose file, confusing next to a stray backup that looks like a half-finished migration.

Either `rm -f "$backup_file"` here too, or move the `cp` to just before the `mv` so the backup only exists once there's something to restore from.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+      service_indent=-1
+      next
+    }
+    in_services && /^[^[:space:]#]/ {
+      in_services=0
+      in_redis=0
+    }
+    in_services && /^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$/ {
+      line_indent=indentation($0)
+      if (service_indent < 0) service_indent=line_indent
```
kiln-agent, line 1008: `service_indent` can latch onto a nested key, and then the migration silently no-ops.

The guard that feeds `service_indent` only matches a *bare* key (nothing after the colon):

```awk
in_services && /^[ ]+[A-Za-z0-9_-]+:[[:space:]]*$/
```

So if the first service key in the operator's file carries a trailing comment — `  postgres: # main database` — that line doesn't match, and the first line that *does* is a nested bare key such as `    environment:` at indent 4. `service_indent` becomes 4, `  redis:` at indent 2 never equals it, `in_redis` stays 0, and the legacy pin is never found. `migrate_legacy_valkey_image` then hits `return 0` (its "nothing to do" path), `update` prints its usual success banner, and the install stays on the amd64-only child.

Reproduced against this diff's exact awk on the stock `docker/docker-compose.yml` with only that one edit:

```
STOCK:   detected
VARIANT: NOT detected     # only change: "  postgres:" → "  postgres: # main database"
```

The same latch sits in the rewrite pass at line 1035, so both halves agree on the wrong indent — a silent miss rather than a mismatch. The stock generated file is safe (its first service key is bare `  postgres:`, and `formbricks.sh` appends traefik/rustfs with bare keys too), so this only bites hand-edited files — but a hand-edited Compose file is exactly the population this migration exists for, and the failure is invisible.

Taking the service indent from the first non-blank, non-comment line below `services:` regardless of what follows its colon would remove the dependency on that formatting. Alternatively, tighten the "not found" path to distinguish "no legacy pin present" from "never located the redis service" and warn on the latter — the current single `return 0` can't tell them apart.

---

.github/workflows/docker-build-validation.yml
kiln-agent, line 56: **This step does not run on this PR**, so it isn't the gate the description says it is.

The workflow is trigger-gated on `main` only:

```yaml
on:
  pull_request:
    branches: [main]
  merge_group:
    branches: [main]
```

`release/5.4` matches neither, and the PR's check runs confirm it — there is no `Validate Docker Build` run among the 16. `Better Auth integration tests` is likewise path-filtered to `apps/web/**`/`packages/**`, which this diff doesn't touch.

Combined with dropping `valkey-image.test.ts`, nothing on this branch verifies either the multi-arch property of the new index or that the pins stay synchronized — the "Open gaps" section calls out the second of those but names this step as covering the first. Worth correcting in the description so the reviewer isn't relying on a check that won't appear. `Run E2E Tests` does pull the new digest as a service container, so the digest is at least proven to boot on amd64 here; the OCI-index assertion is the part that goes unverified.

Not asking you to widen the trigger in a backport — the step is still correct to carry so it's in place if 5.4 ever becomes a `main`-targeted base, and #8986 ran it green on identical content. This is about the claim, not the code.

---

docker/formbricks.sh
```diff
@@ -977,9 +979,89 @@ stop_formbricks() {
...
+    return 1
+  fi
+
+  echo "✅ Updated bundled Valkey to the native amd64/arm64 image. Backup: $backup_file"
+}
+
 update_formbricks() {
   echo "🔄 Updating Formbricks..."
   cd formbricks
+  migrate_legacy_valkey_image docker-compose.yml
```
kiln-agent, line 1064: Unguarded under `set -e`: any failure here aborts the whole update, and the operator isn't told.

`formbricks.sh` runs with `set -e` (line 3), and this call sits in plain command position. Every `return 1` inside `migrate_legacy_valkey_image` — missing Compose file, `replacements != 1`, failed `docker compose config` — therefore terminates `update_formbricks` before `docker compose pull`, `down` and `up -d` ever run.

The messaging makes that hard to diagnose. On the validation-failure path the operator sees only:

```
❌ The updated Compose file is invalid. Restored docker-compose.yml from docker-compose.yml.before-valkey-8.1.9.
```

Nothing says the Formbricks update itself was skipped, and the absence of the usual `🎉 Formbricks updated successfully!` is easy to miss in a long log. Because the restore puts the legacy digest back, a retry re-detects it and fails identically — a loop with no new information.

Two options depending on the intent:

- If the migration should block the update (fail-closed, defensible for a breaking image change), make that explicit:
  ```bash
  if ! migrate_legacy_valkey_image docker-compose.yml; then
    echo "❌ Aborting update: the Valkey image migration failed. Formbricks was not updated."
    return 1
  fi
  ```
- If the app update should proceed on the old Valkey pin, guard it so `set -e` doesn't fire and warn that the stack is still amd64-only.

Either way the operator should learn, from the output, whether the update ran.

---

kiln-agent (commented):
## Review — backport #8986 → `release/5.4`

I verified the mechanical parts of the backport by execution rather than by reading, and they hold up:

- **Digest swap is complete and consistent.** All 8 runtime pins plus the Helm README row carry `e0eb7c48…`; the only surviving `12ba4f45…` is `legacy_valkey_image` in `formbricks.sh`, which is the pattern the updater matches on, by design.
- **The updater's single-pattern match is sufficient.** The historical bundled pin was only ever that one digest, so every existing install is covered by the one legacy string.
- **The awk rewrite is correct.** I replayed both passes against the pre-PR `docker/docker-compose.yml`; the output is byte-identical to the committed post-PR file.
- **The workflow step's extraction works.** `docker compose config --format json | jq -r '.services.redis.image'` resolves as expected locally.
- **Debian → Alpine is safe for Helm.** The StatefulSet override uses `/bin/sh` and busybox-compatible commands, so the base change doesn't break it.

The trimming decisions relative to #8986 look right for backport policy, and the `Breaking changes` tick is correctly set.

Four findings, all in the two pieces this PR actually adds new logic to (the updater and the CI step) rather than in the digest swap itself. None is a reason not to ship the digest change; the first is worth fixing before merge.

| # | Where | Severity | What |
| --- | --- | --- | --- |
| 1 | `formbricks.sh:1064` | **High** | Under `set -e`, a bare call means any migration failure aborts the whole `update` before `pull`/`down`/`up -d` |
| 2 | `formbricks.sh:1009` | Medium | A trailing comment on `redis:` silently skips the migration, contradicting the new docs |
| 3 | `formbricks.sh:1045` | Low | Backup left behind on the failure path that says "the original is unchanged" |
| 4 | `docker-build-validation.yml:56` | Medium (description) | The step is `main`-gated and does not run on this PR, so it isn't the gate the description names |

Findings 1 and 2 are two halves of the same tension: the migration currently fails **loudly and fatally** where it should be soft (finding 1), and **silently** where it should be loud (finding 2). Worth resolving together.

On the risk you flagged yourself — shipping the 82 lines of awk without the unit tests that covered them on `main` — I'd lean toward bringing those cases back. Findings 1–3 are all in exactly that code, and they're the kind of thing those tests existed to catch. That's a judgment call on the no-net-new-tests rule and yours to make; if the smaller delta wins, findings 1 and 2 are the two to fix by hand.

Everything else looks good to me. Commenting rather than requesting changes — the digest change itself is sound and already approved.

---

.github/workflows/docker-build-validation.yml
kiln-agent: Duplicate of [the thread below](https://github.com/formbricks/formbricks/pull/8996#discussion_r3861498227) on the same line, which makes the same point with the check-run evidence. Resolving this one; follow that thread instead.

---

PR #8997, opened 2026-08-26
fix(surveys): stop announcing the public-API name placeholder on app surveys (backport #8982 to release/5.4)

Dhruwang:
Ref ENG-2617

Backport of #8982 to `release/5.4`. Cherry-picked from `e2c7097` (squash commit on `main`), applied with no conflicts, then trimmed to backport policy — minimal, single-purpose, no net-new tests.

## What & why

- On an app-type survey delivered by the js-core widget, the modal's accessible name and its `sr-only` `h1` both read `[deprecated] survey name omitted from public API - will be removed soon` — so that sentence is what a screen reader announces as the dialog's name and as its only heading, for every app-survey respondent.
- The widget reads its surveys from the public client API, which deliberately substitutes that fixed placeholder for every survey name so names aren't exposed over an unauthenticated endpoint. The dialog name and the heading both came from `survey.name` and neither accounted for the substitution.
- `getSurveyDisplayName` returns `undefined` for the placeholder — and for a missing or whitespace-only name — so the renderer falls back to the generic `Survey Dialog` label it already uses for a survey rendered without a name. `RenderSurvey` routes `survey.name` through it, which is the single choke point feeding the dialog's `aria-label`, the `sr-only` `h1`, the `role="form"` landmark and `aria-describedby`.
- The placeholder becomes a shared constant so the API side and the widget can't drift. It lives in a dependency-free `packages/types/js-constants.ts` rather than in `types/js.ts`, because a *value* import of `types/js` drags that module's zod schema graph into the widget bundle — measured at +94 kB on the UMD build on #8982, against 140 bytes for the leaf module.
- Link surveys are unaffected — their name comes from the server component, never from this endpoint.

### Files changed (5)

| File | Change |
| --- | --- |
| `packages/types/js-constants.ts` | new dependency-free module holding `PUBLIC_API_SURVEY_NAME_PLACEHOLDER` |
| `packages/surveys/src/lib/survey-page.ts` | new `getSurveyDisplayName` helper |
| `packages/surveys/src/components/general/render-survey.tsx` | routes `survey.name` through the helper |
| `apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts` | substitutes via the shared constant instead of an inline literal |
| `apps/web/.../lib/data.test.ts` | pre-existing test, updated to assert against the constant (kept — policy keeps edits to existing tests) |

`@formbricks/types` exports `./*` → `./*.ts`, so the new module resolves with no package.json change.

### Dropped relative to #8982

- The four net-new `getSurveyDisplayName` cases added to `packages/surveys/src/lib/survey-page.test.ts` — net-new cases; the file is restored to its `release/5.4` state, so its 17 pre-existing tests are untouched.
- The two net-new assertions added to the existing widget journey in `apps/web/playwright/js.spec.ts` — also net-new; that spec is restored to its `release/5.4` state.

Both were the checks that proved the fix on `main`; see Open gaps.

## Breaking changes

None. The public client API's response is byte-identical — the placeholder is referenced through a constant instead of an inline literal. No API/SDK shape, endpoint, env var, config key or default changes, and no deployer action. The only behaviour change is what assistive tech reads out in the rendered widget.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The whole surveys package still passes after restoring its spec to the release state | unit (guard): `pnpm exec vitest run` in `packages/surveys` | 29 files, 721 passed. `main` reports 725 — the difference is exactly the 4 dropped cases, and no pre-existing test regressed |
| The public client API still substitutes the placeholder, now via the shared constant, so the two cannot drift | unit (guard): `pnpm exec vitest run "app/api/v1/client/[workspaceId]/environment/lib/data.test.ts"` in `apps/web` | 18/18 passed |
| The new module resolves and type-checks from both consumers (widget and web API) | manual: `pnpm typecheck --filter=@formbricks/surveys --filter=@formbricks/types`, then `--filter=@formbricks/web` | Both green; web 12/12 tasks. This is the check that would catch the `types` export map not covering a new file |
| Formatting and lint match repo style | manual: repo pre-commit hooks (`prettier --write` + `lint-staged-eslint`) ran on commit, then `prettier --check` on all 5 changed files | Hooks clean, no bypass; "All matched files use Prettier code style!" |
| The placeholder guard itself, the +94 kB bundle finding, and the widget behaviour end-to-end | manual, inherited from #8982 | Verified there on identical source bytes — mutation-tested guard (reverting it turned the case red: 1 failed / 724 passed), UMD bundle 971,447 B → 971,587 B (+140 B), and the real API → widget → dialog path |

**Open gaps**

- **The two checks that actually prove this fix were dropped as net-new**, so on this branch nothing automated would catch the placeholder leaking back in: the mutation-tested `getSurveyDisplayName` cases, and the Playwright assertions on the real API → widget → dialog path. That path is the only place the API, the widget and the dialog are wired together. If you would rather have those two than the smaller delta, say so and I will bring them back — this is the one open gap I would actually push back on the policy for.
- Playwright was not run here (no Docker daemon in this environment, so Postgres/Redis cannot boot) — and with the assertions dropped it would not cover this behaviour anyway.
- As on #8982: the `sr-only` `h1` is now absent on widget-delivered surveys, since there is no name to put in it. That is the intended fallback and matches documented preview behaviour, but it means WCAG 2.4.6's top-level heading is not present on that delivery path. Restoring a real heading needs an API-side decision about exposing survey names to the public client endpoint, out of scope for a QA fix.
- Screen-reader behaviour was verified by reading accessibility attributes, not by listening with an actual screen reader.

**Risks**

- Inline (non-modal) app surveys delivered by the widget lose their `role="form"` landmark, since the container only declares the role once there is a name to give it. That is the pre-existing, deliberate behaviour for an unnamed survey, but it is a change on that path.
- `getSurveyDisplayName` matches the placeholder **exactly**; editing the API-side string without updating the shared constant starts the leak again. The shared constant exists to prevent that, and both call sites import it — but with the drift-pinning test dropped here, nothing on this branch enforces it.

### DIFF
--- a/packages/surveys/src/components/general/render-survey.tsx
+++ b/packages/surveys/src/components/general/render-survey.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 import { SurveyContainerProps } from "@formbricks/types/formbricks-surveys";
-import { hasSurveyInstructions } from "@/lib/survey-page";
+import { getSurveyDisplayName, hasSurveyInstructions } from "@/lib/survey-page";
 import { getSurveyLanguageTag, isRTLLanguage } from "@/lib/utils";
 import { SurveyContainer } from "../wrappers/survey-container";
 import { Survey } from "./survey";
@@ -106,7 +106,7 @@ export function RenderSurvey(props: Readonly<SurveyContainerProps>) {
       onClose={close}
       isOpen={isOpen}
       dir={dir}
-      surveyName={props.survey.name}
+      surveyName={getSurveyDisplayName(props.survey.name)}
       hasInstructions={hasSurveyInstructions(props.survey)}
       lang={languageTag}>
       <Survey
--- a/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.test.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.test.ts
@@ -3,6 +3,7 @@ import { prisma } from "@formbricks/database";
 import { Prisma } from "@formbricks/database/prisma";
 import { logger } from "@formbricks/logger";
 import { DatabaseError, ResourceNotFoundError } from "@formbricks/types/errors";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
 import { getWorkspaceStateData } from "./data";
 
 vi.mock("server-only", () => ({}));
@@ -120,7 +121,7 @@ describe("getWorkspaceStateData", () => {
       surveys: [
         {
           ...mockWorkspaceData.surveys[0],
-          name: "[deprecated] survey name omitted from public API - will be removed soon",
+          name: PUBLIC_API_SURVEY_NAME_PLACEHOLDER,
         },
       ],
       actionClasses: mockWorkspaceData.actionClasses,
--- a/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts
+++ b/apps/web/app/api/v1/client/[workspaceId]/environment/lib/data.ts
@@ -9,6 +9,7 @@ import {
   TJsWorkspaceStateSurvey,
   TJsWorkspaceStateWorkspaceSetting,
 } from "@formbricks/types/js";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
 import { type TBaseFilters, buildSurveyInteractionRefreshMap } from "@formbricks/types/segment";
 import { toLegacyLanguageCodes } from "@/lib/i18n/utils";
 import { validateInputs } from "@/lib/utils/validate";
@@ -283,7 +284,7 @@ export const getWorkspaceStateData = async (workspaceId: string): Promise<Worksp
 
       return {
         ...transformed,
-        name: "[deprecated] survey name omitted from public API - will be removed soon",
+        name: PUBLIC_API_SURVEY_NAME_PLACEHOLDER,
         segment: sanitizedSegment,
         ...(interactionRefresh ? { interactionRefresh } : {}),
       };
--- a/packages/surveys/src/lib/survey-page.ts
+++ b/packages/surveys/src/lib/survey-page.ts
@@ -1,4 +1,7 @@
+// Imported from the dependency-free constants module rather than from `./js`: a value import of
+// `@formbricks/types/js` pulls its zod schema graph into the widget bundle (+94 kB on the UMD build).
 import { type TJsWorkspaceStateSurvey } from "@formbricks/types/js";
+import { PUBLIC_API_SURVEY_NAME_PLACEHOLDER } from "@formbricks/types/js-constants";
 
 /** Id of the visually-hidden region holding the survey's persistent instructions. */
 export const SURVEY_INSTRUCTIONS_ID = "fb__survey-instructions";
@@ -41,6 +44,26 @@ export const getSurveyPagePosition = (
   // Not a block: an ending card, or an id that no longer resolves after the survey was edited
   // mid-session. The last page either way.
   return { index: total, total };
+};
+
+/**
+ * The survey name to show a respondent, or `undefined` when there is none to show.
+ *
+ * A survey delivered by the JS widget is fetched from the public client API, which deliberately
+ * replaces every name with a placeholder so names are not exposed over an unauthenticated endpoint
+ * (ENG-808). That placeholder is an internal deprecation notice, so rendering it is worse than
+ * rendering nothing: it became the dialog's accessible name and its only heading on every app
+ * survey, which is what a screen reader then announced. Treat it as "this survey has no name" and
+ * let the callers fall back the way they already do for a survey rendered without one.
+ *
+ * Link surveys are unaffected — their name comes from the server component, never from this
+ * endpoint, so it never matches the placeholder.
+ *
+ * Also drops a name that is only whitespace, which would name the dialog with nothing at all.
+ */
+export const getSurveyDisplayName = (name: string | undefined): string | undefined => {
+  if (!name || name === PUBLIC_API_SURVEY_NAME_PLACEHOLDER) return undefined;
+  return name.trim().length > 0 ? name : undefined;
 };
 
 /**
--- a/packages/types/js-constants.ts
+++ b/packages/types/js-constants.ts
@@ -1,0 +1,19 @@
+/**
+ * Constants shared between the public client API and the JS widget that consumes it.
+ *
+ * This module deliberately imports nothing. Its consumers include `@formbricks/surveys`, whose
+ * bundle every respondent of an app survey downloads, and importing a *value* from `./js` instead
+ * would drag that module's zod schema graph into the bundle — measured at +94 kB on the UMD build.
+ * Keep it dependency-free, and keep constants the widget needs here rather than in `./js`.
+ */
+
+/**
+ * The string the public client API substitutes for every survey name it returns, so that survey
+ * names are not exposed over an unauthenticated endpoint (ENG-808).
+ *
+ * It is a marker, not text anyone should ever read: it lives here rather than inline at the
+ * substitution site so the widget can recognise it and refuse to render it. Change it in one place
+ * or the widget stops recognising it — see `getSurveyDisplayName` in `@formbricks/surveys`.
+ */
+export const PUBLIC_API_SURVEY_NAME_PLACEHOLDER =
+  "[deprecated] survey name omitted from public API - will be removed soon";


packages/surveys/src/components/general/render-survey.tsx
```diff
@@ -106,7 +106,7 @@ export function RenderSurvey(props: Readonly<SurveyContainerProps>) {
       onClose={close}
       isOpen={isOpen}
       dir={dir}
-      surveyName={props.survey.name}
+      surveyName={getSurveyDisplayName(props.survey.name)}
```
kiln-agent, line 109: Every widget-delivered app survey now loses its only top-level heading, and unlike the dialog name this one has no fallback.

`getSurveyDisplayName` returns `undefined` for the placeholder, and `survey-container.tsx:214` gates the heading on that value:

```tsx
const surveyHeading = surveyName ? <h1 className="sr-only">{surveyName}</h1> : null;
```

The dialog's accessible name is covered — `aria-label={surveyName ?? t("common.survey_dialog")}` at line 271 — but the `h1` just renders `null`. The net effect on the widget path: the placeholder sentence stops being announced (the fix works), but heading structure now starts at `h2`, so a screen-reader user pressing <kbd>H</kbd> or opening the rotor lands on the card prompts with no document heading above them.

Passing the same localized fallback to the heading gets the announcement fix without giving up WCAG 2.4.6:

```tsx
const surveyHeading = <h1 className="sr-only">{surveyName ?? t("common.survey_dialog")}</h1>;
```

I can see the argument for keeping this identical to #8982 rather than diverging on a backport, and the PR body does disclose it under `Open gaps`, so this is a known trade-off rather than an oversight — it's just the one thing in the diff I'd push back on. If you'd rather keep the backport byte-identical to `main`, that's reasonable; the fix belongs on `main` first either way.

---

kiln-agent, line 109: Suppressing the name also silently drops the instructions description on the **inline** widget path — a consequence beyond the `role="form"` loss listed under Risks.

`SurveyContainer` gates the description id on the name, not on the instructions (`survey-container.tsx:231`):

```ts
const instructionsId = surveyName && hasInstructions ? SURVEY_INSTRUCTIONS_ID : undefined;
```

while the modal branch gates it on `hasInstructions` alone (line 272). So for an app survey rendered **inline** by the widget with a non-empty welcome-card subheader, `surveyName` is now `undefined`, `instructionsId` becomes `undefined`, and `Survey` still renders the `sr-only` `#fb__survey-instructions` region (`survey.tsx:1361`) — the region is in the DOM but nothing references it, so the persistent instructions are never announced on any page. Before this change the placeholder made `surveyName` truthy, so that path did carry `aria-describedby`. Net effect on inline app surveys: the placeholder announcement is gone (good), but so are the landmark, the accessible name and the instructions (a regression the modal path does not share).

Cheapest fix that keeps the intent is to stop coupling the description to the name, since the region's presence is exactly what `hasInstructions` reports:

```suggestion
      surveyName={getSurveyDisplayName(props.survey.name)}
```

(the edit belongs in `survey-container.tsx:231`: `const instructionsId = hasInstructions ? SURVEY_INSTRUCTIONS_ID : undefined;`). If you would rather not touch a file outside the cherry-pick on a backport, worth calling out under Risks/Open gaps as a second inline-path change, or fixing forward on `main` first.

---

kiln-agent (commented):
Reviewed the backport (5 files, +48/−4). The cherry-pick is faithful and the mechanism is sound — one accessibility finding inline, which the PR body already discloses as a known trade-off.

**What I verified**

- `survey.name` has exactly one consumer in `packages/surveys` (the changed line), so there's no second, unrouted path that still reads the raw name.
- The placeholder literal now exists in exactly one place repo-wide; the API response stays byte-identical. The unrelated segment-title placeholder is untouched.
- `@formbricks/types` exports `./*` → `./*.ts`, so the new leaf module resolves with no `package.json` change. The dependency-free module (vs. `types/js.ts`) is the right call for the bundle-size reason given.
- No dangling `aria-describedby`: the modal path (`survey-container.tsx:272`) keys off `hasInstructions` alone, and the region in `survey.tsx:1361` renders independently. The inline path's `instructionsId` gate (line 231) also folds to `undefined` cleanly.
- No existing Playwright spec breaks — `survey-accessibility.spec.ts:628` asserts the `h1` on `/s/<id>` (link survey, real name) and `js.spec.ts:70` targets the dashboard list, not the widget.
- Import ordering matches the enforced sort in all three touched files; `common.survey_dialog` exists and is localized across `packages/surveys/locales/*`.

**On the stated risk**

The PR body flags that inline app surveys would lose their `role="form"` landmark. That isn't reachable today — js-core calls `renderSurvey` with no `mode` (`widget.ts:164`), so the widget always takes the modal path, and the only inline callers (`apps/web/modules/ui/components/survey/index.tsx`, the link-survey pages) pass real names. Worth keeping in the body as a latent risk if a future inline widget path lands, but nothing regresses now.

**On the dropped tests**

The trimming is consistent with backport policy and the `Open gaps` section is honest about the cost: nothing on this branch would catch the placeholder leaking back in. Since both dropped checks exist and pass on `main` against identical source bytes, and the guard was mutation-tested there, I don't think that blocks the backport — the drift risk is pinned by the shared constant, which is the part that matters.

Not approving — leaving that to @itsjavi.

---

kiln-agent (commented):
## Review — backport #8982 → `release/5.4`

The cherry-pick is faithful and the fix itself is right: `render-survey.tsx:109` is genuinely the single choke point (`surveyName` has exactly one producer and one consumer), the substitution literal now exists in exactly one place in the repo, `@formbricks/types`' `"./*": "./*.ts"` export resolves the new leaf module with no package.json change, and value imports from `@formbricks/types` subpaths already exist in `packages/surveys` (e.g. `FILE_UPLOAD_ERROR_NAMES`) so resolution and bundling are proven. The v2 client endpoint re-exports v1, so there is no second substitution site, and `apps/web/public/js/surveys.umd.cjs` is gitignored so no stale bundle is committed. `packages/types` typechecks clean; `packages/surveys` shows no new errors; `survey-page.test.ts` passes 17/17.

Four things worth a decision before merge. The first two are consequences of the dropped tests, the last two are behavioural and are acknowledged in `Open gaps`/`Risks` — I'm restating them because on this branch they now apply to *every* app-survey respondent, with nothing automated covering them.

### 1. `getSurveyDisplayName` ships with zero coverage — `packages/surveys/src/lib/survey-page.ts:64`

Lines 65-66 are uncovered, and `packages/surveys/src/lib/**` is Sonar-analysed with no coverage exclusion, so this PR's new-code coverage measures around 1/3 of its lines — likely enough to trip the quality gate. Beyond the gate: nothing on this branch catches the placeholder leaking back in. The description asks whether to bring the four dropped cases back, and my answer is yes for these — they are unit cases on a pure two-line helper colocated with a spec that already exists, which is the cheapest possible way to keep both the gate and the guard. The Playwright assertions are a fairer thing to leave out.

### 2. The API-side test assertion is now tautological — `data.test.ts:124`

Test and `data.ts:287` read the same symbol, so editing `PUBLIC_API_SURVEY_NAME_PLACEHOLDER` changes the unauthenticated API response with every test green. That matters more than usual here: already-deployed widget bundles were compiled against the old literal, so a change to the constant makes them start announcing the deprecation notice again. Asserting the literal string in the API test (and letting the widget side import the constant) keeps the pin where the contract actually lives.

### 3. Inline app surveys lose their form landmark entirely — `survey-container.tsx:241`

The inline branch has no `?? t("common.survey_dialog")` fallback that the modal branch has on line 271. With `surveyName` now `undefined` on the widget path, an inline app survey loses `role="form"` and its accessible name — and via the `surveyName && hasInstructions` gate on line 231, also the `aria-describedby` that made the persistent instructions region announce on every page. The instructions loss looks like collateral rather than intent: it isn't the landmark decision the comment above line 231 is reasoning about. Gating `instructionsId` on `hasInstructions` alone would restore it without touching the landmark question.

### 4. No `h1` at all on widget-delivered surveys — `survey-container.tsx:214`

Every card `h2` is now orphaned on this path, which is the exact condition the comment block at lines 204-213 exists to prevent. The localized fallback already used for the dialog's `aria-label` would keep a real heading there for one line of change. I take the point that a generic heading is weaker than a real survey name, but "Survey" as an `h1` is a better outcome than no top-level heading, and the only `h1` assertions in the suite run against a link survey, so nothing would notice either way.

### Two nits

- `survey-page.ts:1-2` — the comment explains why the constant comes from `js-constants` but sits above the `./js` type import, so it reads as annotating the wrong line. Moving it one line down would fix that.
- `getSurveyDisplayName` returns the untrimmed name when it has content. Harmless (accessible-name computation normalizes whitespace), just slightly inconsistent with the whitespace-only branch right above it.

---

packages/surveys/src/components/general/render-survey.tsx
kiln-agent: Correction on reachability, and ignore the suggestion block above — it repeats the existing line and applies nothing (the change I described belongs in `survey-container.tsx:231`, which isn't in this diff).

Checking the call sites: js-core calls `formbricksSurveys.renderSurvey({...})` with no `mode` (`packages/js-core/src/lib/survey/widget.ts:164`), and `RenderSurvey` defaults to `mode = props.mode ?? "modal"`, so a widget-delivered survey always takes the modal branch. The only inline caller is `apps/web/modules/ui/components/survey/index.tsx:33`, and the surveys it renders come from the server with real names — never the placeholder. So on the reachable paths nothing regresses: the modal branch keeps `aria-describedby` (gated on `hasInstructions` alone) and falls back to `common.survey_dialog` for its name.

That makes this latent rather than a live regression: it only bites if an inline widget path is added later, or a host calls the public `renderSurveyInline` with a survey it fetched from the client API itself. Not merge-blocking on this backport — the `role="form"` line under Risks is worth extending to mention `aria-describedby`, and the real fix (gating `instructionsId` on `hasInstructions` alone) belongs on `main`.

The reachable loss on this path is the missing `h1`, which is the other thread.

---

xernobyl (approved):
LGTM

PR #9000, opened 2026-08-26
fix: doesNotEqual logic condition is always true for single-selection answers

### LINKED ISSUE #8527: [Bug] doesNotEqual logic condition is always true for single-selection answers
## Issue

In survey logic evaluation, a doesNotEqual condition where the left operand is a single-element array (e.g. a MultipleChoiceMulti answer with exactly one selection) always evaluates to true, regardless of the actual value.

## Root cause

In packages/surveys/src/lib/logic.ts evaluateSingleCondition, doesNotEqual case), when the single selection DOES match the condition value, evaluation falls through to leftValue !== rightValue — and an array reference is never strictly equal to a string, so the expression is always true. doesNotEqual therefore passes even when the answer equals the condition value, and it is not the inverse of equals (which handles this case correctly).

### Expected Behavior

_No response_

## Steps to reproduce

1. Create a survey with a MultipleChoiceMulti question.
2. Add logic to a later element with the condition: [question] doesNotEqual "Option 1".
3. Answer the question selecting only "Option 1".
4. The condition evaluates to true and the logic jump fires — expected: false.

## Fix

I have a fix ready with a regression test on this branch: https://github.com/FrancoKaddour/formbricks/tree/fix/does-not-equal-single-selection-array

It negates the exact expression used by the equals case, so the two operators are proper inverses. I know community PRs are accepted only in rare exceptions — happy to open the PR if you'd like, or a maintainer can cherry-pick commit 4963ddf.

### Your Environment

None

pandeymangg:
Fixes ENG-1798
Fixes #8527

## What & why

- In survey logic evaluation, a `doesNotEqual` condition with a single-selection answer (e.g. a MultipleChoiceMulti answer with exactly one selection) always evaluated to `true`, even when the selection matched the condition value — so `doesNotEqual` jumps fired when they shouldn't.
- Root cause: the `doesNotEqual` fallthrough OR-ed past its single-element-array guard into `leftValue !== rightValue`. An array is never strictly equal to a string, so that comparison is always `true`. `equals` has the same shape but survives it, because its `===` fallback is correctly `false` for mismatched types.
- Fix: decide inside the guard — when the left value is a single-element array and the right value a string, return `!leftValue.includes(rightValue)`; only otherwise fall through to `!==`. This makes `doesNotEqual` the exact inverse of `equals` in every case, including multi-selection arrays.
- Applied in both copies of the evaluator: `packages/surveys/src/lib/logic.ts` (runtime) and `apps/web/lib/surveyLogic/utils.ts` (mirror used by quotas and the editor).
- Sonar's new-code duplication gate flagged the synced fix: the two files are already six-block mirrors of each other (the whole `evaluateSingleCondition` included), so any fix that correctly lands in both trips the gate. Declared the web mirror in `sonar.cpd.exclusions` (same precedent as `schema-definition.ts`) rather than making the copies textually diverge. Actually unifying the evaluator into a shared package is out of scope here.

## Breaking changes

None — bug fix inside logic evaluation; no exported SDK signature, API shape, env var or config change. Surveys whose logic happened to rely on the buggy always-true `doesNotEqual` will now branch as configured, which is the fix, not a contract change.

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `doesNotEqual` is `false` when the single selection matches the condition value (runtime evaluator) | unit (red on main) | `cd packages/surveys && pnpm vitest run src/lib/logic.test.ts` — fails on main with `expected true to be false`, passes with the fix (42/42) |
| Same behaviour in the web mirror evaluator | unit (red on main) | `cd apps/web && pnpm vitest run lib/surveyLogic/utils.test.ts` — fails on main with `expected true to be false`, passes with the fix (28/28) |
| `doesNotEqual` stays `true` for a non-matching selection; `equals` unchanged in both directions | unit (guard) | same test files, adjacent assertions in the new tests |
| No regression across the rest of logic evaluation and its consumers | unit (guard) | full `packages/surveys` suite 726/726; `apps/web` quotas + editor lib suites 460/460 |

<details>
<summary>Red-on-main proof: failure output of the new tests against the unfixed source</summary>

The tests were written first and run against the unfixed evaluator; both fail on exactly the assertion that pins the bug (a matching single selection must make `doesNotEqual` false), then pass once the fix is applied.

`packages/surveys` (`pnpm vitest run src/lib/logic.test.ts`):

```
FAIL  src/lib/logic.test.ts > Survey Logic > Edge Cases >
      evaluates doesNotEqual as the inverse of equals for single-selection answers
AssertionError: expected true to be false // Object.is equality
 ❯ src/lib/logic.test.ts:1444:9   (doesNotEqual "opt1" against answer ["opt1"])

Tests  1 failed | 41 passed (42)
```

`apps/web` (`pnpm vitest run lib/surveyLogic/utils.test.ts`):

```
FAIL  lib/surveyLogic/utils.test.ts > surveyLogic >
      evaluateLogic treats doesNotEqual as the inverse of equals for single-selection answers
AssertionError: expected true to be false // Object.is equality
 ❯ lib/surveyLogic/utils.test.ts:289   (doesNotEqual "opt1" against answer ["opt1"])

Tests  1 failed | 27 passed (28)
```

The three adjacent assertions in each test (`equals` true/false, `doesNotEqual` true for a non-match) pass on main too — they guard the behaviour that was already correct, so the fix cannot overshoot into inverting the non-matching case.

</details>

**Open gaps**

- Not re-verified in a rendered survey in the browser; the new tests exercise the exact evaluator the widget ships, and per AGENTS.md this logic-level bug is pinned at the unit level (the survey-logic journey is already covered by the existing e2e specs).

**Risks**

- Any existing survey whose logic depends on the old always-true `doesNotEqual` for single-selection answers will now take the other branch. That is the intended behaviour change; if a report comes in, re-check the survey's logic conditions rather than the evaluator.
- The web mirror feeds quota evaluation (`apps/web/modules/ee/quotas/lib/utils.ts`) — quota criteria using `doesNotEqual` on multi-choice answers change outcome the same way, intentionally.

---

> [!NOTE]
> **AI model used** — `claude-fable-5` (Claude Code 2.1.241), reasoning effort `unknown`.


kiln-agent (commented):
## Review

The fix is correct and minimally scoped, and the testing claims in the description hold up — I reproduced them. One CI blocker to clear, one inaccurate sentence in the description, and one adjacent bug of the same family that I'd suggest tracking separately rather than widening this PR.

### The fix is provably surgical

The rewrite changes behaviour in exactly one input class and nothing else. Writing `A` for `Array.isArray(leftValue) && leftValue.length === 1 && typeof rightValue === "string"` and `inc` for `leftValue.includes(rightValue)`:

| | old `(A && !inc) \|\| (leftValue !== rightValue)` | new `A ? !inc : (leftValue !== rightValue)` |
| --- | --- | --- |
| `A` true, `inc` true | `false \|\| (array !== string)` → **true** | **false** |
| `A` true, `inc` false | `true \|\| …` → true | `!false` → true |
| `A` false | `leftValue !== rightValue` | `leftValue !== rightValue` |

Only the first row moves, and that row *is* ENG-1798. Nothing else in the operator can shift, which is what you want in an evaluator this widely consumed. Both copies (`packages/surveys/src/lib/logic.ts`, `apps/web/lib/surveyLogic/utils.ts`) are byte-identical in the changed region, and a grep for `case "doesNotEqual"` / `leftValue !== rightValue` confirms there is no third copy that was missed.

### Verified, not taken on trust

- **Red-on-main, both packages.** Reverting only the two source files to the base commit and re-running the new tests reproduces exactly the failures quoted in the description — `packages/surveys` 1 failed / 41 passed at `logic.test.ts:1444`, `apps/web` 1 failed / 27 passed at `utils.test.ts:289`, both `expected true to be false`. The tests genuinely can fail.
- **The fixtures exercise the real path.** The comment "answer labels resolve to choice ids" is accurate — `getLeftOperandValue` maps response labels through `choice.label` to `choice.id` for `multipleChoiceMulti` (`logic.ts:124-140`), so `["Option 1"]` really does become `["opt1"]`. The test isn't passing for an incidental reason.
- **Suites green with the fix.** `packages/surveys` 726/726 and the `apps/web` `lib/surveyLogic` + `modules/ee/quotas` suites 86/86, including the quota consumer the description flags as affected.
- **The fixed path is reachable from the UI** — `doesNotEqual` is offered for `MultipleChoiceMulti` in `logic-rule-engine.ts:140`.
- **Breaking-changes box unticked is the right call.** No exported signature, API shape, env var or config change; behaviour-only change inside logic evaluation, already documented under `Risks`.

### 1. SonarCloud quality gate is failing — 9.8% duplication on new code (required ≤ 3%)

This is blocking the merge. The cause is in the diff rather than pre-existing: the ~40-line `multiSurvey` fixture plus the `condition` helper is duplicated near-verbatim between `packages/surveys/src/lib/logic.test.ts` and `apps/web/lib/surveyLogic/utils.test.ts`. That is an awkward consequence of the evaluator itself being mirrored — the two test files can't share a helper across the package boundary — but the gate won't care. Trimming the fixture to the minimum the assertions need (the two choices and the one element, dropping fields the evaluator never reads) would likely bring it under the threshold without splitting anything out.

### 2. "the exact inverse of `equals` in every case" is not accurate

The description says the fix "makes `doesNotEqual` the exact inverse of `equals` in every case, including multi-selection arrays." The multi-selection part is true — I confirmed `["opt1","opt2"]` vs `"opt1"` gives `equals` false / `doesNotEqual` true. But two branches upstream of the fallthrough remain non-complementary, both verified by running them:

- **PictureSelection with 2+ selections where the value is among them.** `["img1","img2"]` vs `"img1"` → `equals` **false** and `doesNotEqual` **false**. `doesNotEqual` has a `length > 0` PictureSelection guard that `equals` has no counterpart for.
- **Right operand is a `MultipleChoiceMulti` with 2+ selections.** Both operators hit `else return false`, so both return **false**.

Neither is a regression from this PR — both predate it and are untouched. The concern is only that the sentence as written asserts an invariant that doesn't hold, which could lead a later reader to assume the pair is complementary everywhere and stop looking. Suggest narrowing it to the fallthrough branch, e.g. "makes the fallthrough comparison the exact inverse of `equals`, including for multi-selection arrays."

### 3. Follow-up: the right-operand branch is the same bug with the opposite sign

In the `else return false` case above, `doesNotEqual` returns `false` where it should return `true` — a `doesNotEqual` jump can never fire when the right-hand multi-choice question has 2 or more selections. Same family as ENG-1798, opposite symptom ("never true" instead of "always true"), and the same shape: a guard deciding the whole operator instead of just its own case.

I'd leave it out of this PR — the narrow scope here is a virtue — but it's worth a ticket. Caveat on severity: I confirmed the branch exists and is reachable in code, but did not confirm the editor actually offers a multi-choice question as the *right* operand for `doesNotEqual`, so it may be unreachable in practice.

### 4. Optional: nothing pins the two evaluator copies together

You correctly updated both mirrors, but no test asserts they agree, so the next fix that lands in one and not the other does so silently. A shared table-driven case list, or even a test that diffs the two `evaluateSingleCondition` outputs over a fixed input matrix, would make the next divergence loud. Pre-existing structural gap, not this PR's obligation.

### Verdict

The core change is right, well-tested, and honestly described apart from the one overclaim. Clear the Sonar gate and fix that sentence, and this is good to merge — items 3 and 4 are follow-ups, not conditions.

---

kiln-agent (commented):
## Review — `fix: doesNotEqual logic condition is always true for single-selection answers`

The diagnosis and the fix are correct, the tests genuinely fail without it, and both mirrors are updated. One blocking item that needs a human decision, one wording fix, and one adjacent live bug I found in the same function.

### Blocking — the SonarCloud quality gate is red, and it's this diff

`new_duplicated_lines_density` is **9.8%** against a 3% threshold. Every other gate condition is OK (coverage on new code is 87.5%). The duplication is attributed to the fix itself, not the tests:

| Component | new lines | new dup. density |
| --- | --- | --- |
| `packages/surveys/src/lib/logic.ts` | 7 | **100.0%** |
| `packages/surveys/src/lib/logic.test.ts` | 80 | 0.0% |
| `apps/web/lib/surveyLogic/utils.ts` (via `apps/web/lib`) | 56 total | 12.5% |

14 duplicated lines / 143 new = 9.79% — i.e. the 7-line block, counted once per copy. It trips because the correct fix *is* byte-identical in both evaluators, landing inside an already-identical `doesNotEqual` body. Worth being blunt about: **no stylistic change to this PR avoids it.** Rewording the comment in one copy won't help (Sonar's copy-paste detector doesn't tokenize comments, and the surrounding code lines duplicate regardless). The realistic options are to waive the gate for this PR, or to de-duplicate the evaluator — which is the follow-up I'd argue for below. Flagging rather than prescribing, since waiving is a maintainer call.

### Adjacent live bug in the same function — `isNotClicked` is missing from the web mirror

Diffing the two `evaluateSingleCondition` implementations end to end (242 lines each) turned up exactly one semantic divergence, and it isn't cosmetic:

- `packages/surveys/src/lib/logic.ts:413` — `case "isNotClicked": return leftValue !== "clicked";`
- `apps/web/lib/surveyLogic/utils.ts` — **no such case**, so it falls to `default: return false`.

The operator is reachable from the quota UI: `logic-rule-engine.ts:209` offers `is_not_clicked` for `CTA`, `quota-condition-builder.tsx` builds its operator list through the same `createSharedConditionsFactory` → `getConditionOperatorOptions` path with no CTA exclusion, and quota evaluation calls the web mirror at `apps/web/modules/ee/quotas/lib/utils.ts:37`. Confirmed by probe against the mirror — a **dismissed** CTA gives:

```
web mirror  -> isClicked: false  isNotClicked: false     // should be true
```

So a quota criterion "CTA is not clicked" silently never matches, while the same condition evaluates correctly in the shipped widget. Same bug family as the one this PR fixes — mirror drift — but out of scope here; worth its own ticket rather than growing this diff.

That's the second instance of this family in one function, and it's what makes the Sonar blocker above worth solving properly: `apps/web` already depends on `@formbricks/surveys` (`workspace:*`), and that package already exports a pure-logic subpath (`./validation`) as precedent for a `./logic` export the web mirror could delegate to. That kills the duplication metric and the drift class in one move.

### Note — the "exact inverse in every case" claim in the description is too strong

> This makes `doesNotEqual` the exact inverse of `equals` in every case, including multi-selection arrays.

The "including multi-selection arrays" half is right, and I checked it — for `MultipleChoiceMulti` the pair is inverse for a matching single selection, a non-matching one, a 2-item selection, and an empty (skipped) answer. But the sentence reads as a guarantee about the operator, and there's a counterexample. `PictureSelection` with 2+ selections, right operand a selected id:

```
equals: false   doesNotEqual: false     // not inverse
```

`equals` has no `PictureSelection` guard so it falls through to the length-1 clause and returns false, while `doesNotEqual`'s guard (`length > 0` → `!includes`) also returns false. Pre-existing, untouched by this diff, and `PictureSelection` does offer both operators in the editor (`logic-rule-engine.ts:169`). The same holds for the `rightOperand.type === "element"` `MultipleChoiceMulti` path, where both sides hard-return `false` when `rightValue.length !== 1`. Suggest narrowing the claim to the fallthrough branch / `MultipleChoiceMulti` so it isn't read as a property of the operator.

### Verified

Checked rather than taken from the description:

- **Red-on-main reproduced**, both files, at the exact assertions cited: `logic.test.ts:1444` and `utils.test.ts:289`, both `expected true to be false`, 1 failed / 41 passed and 1 failed / 27 passed. The tests can fail.
- **The tests exercise the changed lines** — `rightOperand` is `static`, so the `PictureSelection`, `Date` and right-operand-element guards are all skipped and evaluation reaches the modified block.
- **The test comment is accurate** — `getLeftOperandValue` maps choice labels to ids for `multipleChoiceMulti`, so `["Option 1"]` really does resolve to `["opt1"]`.
- **No third copy left unfixed** — only these two evaluators exist in the repo.
- **Suites green**: `packages/surveys` 726/726, `apps/web` quotas + `surveyLogic` 86/86. (My first local run showed unrelated `Failed to resolve entry for package` failures — the unbuilt-`dist/` trap from AGENTS.md, cleared by `pnpm build --filter=@formbricks/web^...`, not the diff.)
- **Test level matches AGENTS.md** — logic bug pinned at the unit level, no new E2E spec. Correct call.
- **Breaking-changes box** correctly left unticked; the behaviour change is the fix, reachable only through in-repo logic evaluation.

Not approving — the red gate needs resolving or waiving first, and the `isNotClicked` gap deserves a ticket. The fix itself I'd take as-is.

---

mattinannt (approved):

PR #9005, opened 2026-08-27
feat(analytics): add multi-org-aware role person properties to PostHog

harshsbhat:
No linked ticket — ad hoc analytics improvement raised in conversation, not tracked in Linear.

## What & why

PostHog analytics had no person-level signal for a user's role — organization/workspace group properties exist, but nothing on the person profile lets you segment or filter people by role (e.g. "org owners") without joining through group analytics.

A person's role can't be reduced to one value: PostHog person properties are a single flat value per key, but a user can belong to more than one organization with a different role in each. Stamping a single `organization_role` scalar would just reflect whichever org/workspace was most recently viewed and silently drift as the person navigated.

Instead, every identify call now sends a full snapshot recomputed from the database: `organization_roles` (array of `{organization_id, role}`) and `organization_count` (`apps/web/lib/posthog/organization-roles.ts`, backed by a new `getMembershipsByUserId` in `apps/web/lib/membership/service.ts`). Recomputing the whole set on every call — rather than patching one field — makes each call idempotent regardless of order, and self-healing if a role changes or a membership is removed elsewhere, without needing every mutation site to remember to push an update.

Wired in at three points:
- Signup (`apps/web/modules/auth/signup/actions.ts`) — server-side, folded into the existing identify call.
- Both create-organization actions (`apps/web/modules/organization/actions.ts`, `apps/web/app/setup/organization/create/actions.ts`) — server-side, so the snapshot is correct even before the client mounts.
- Every workspace page load (`apps/web/app/posthog/PostHogGroupIdentify.tsx`, wired from `apps/web/app/(app)/workspaces/[workspaceId]/layout.tsx`) — client-side, via `posthog.setPersonProperties`, so the snapshot stays fresh over time.

Filtering on role (e.g. owners) is done via HogQL against `organization_roles` rather than a flattened boolean, since role isn't a strict linear hierarchy (`billing` sits alongside `member`/`manager` rather than below them).

## Breaking changes

None — internal analytics-only change. No public API/SDK shape is touched, no new required env var, and the new PostHog properties are only sent when `POSTHOG_KEY` is already configured (same gate as before).

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `getMembershipsByUserId` returns every membership row for a user, scoped by `where: { userId }`, and classifies Prisma vs. unknown errors | unit (mutation) — `apps/web/lib/membership/service.test.ts`; mutate `apps/web/lib/membership/service.ts:65-67` (drop the `where` filter) to see it fail | 4/4 passing |
| `getOrganizationRolePersonProperties` maps every membership into `{organization_id, role}` and computes `organization_count`, including the empty-membership case | unit (mutation) — `apps/web/lib/posthog/organization-roles.test.ts`; mutate `apps/web/lib/posthog/organization-roles.ts:26-30` (swap/drop a field) to see it fail | 2/2 passing |
| Signup still composes and fires `identifyPostHogPerson` (email/name + role snapshot) across every existing signup path without throwing, now that the snapshot is read back from the DB after provisioning | unit (guard) — `apps/web/modules/auth/signup/actions.test.ts` (extended with a mock for the new `getOrganizationRolePersonProperties` import) | 20/20 passing |
| `identifyPostHogPerson`/`groupIdentifyPostHog`/`capturePostHogEvent` generic pass-through, including the widened person-properties type (now accepts an array of objects, not just scalars) | unit (guard) — `apps/web/lib/posthog/capture.test.ts` | 15/15 passing (unchanged) |
| Scoped typecheck (`tsc --noEmit`) and lint (`eslint`) over every touched file | guard | 0 errors (1 pre-existing warning in `signup/actions.ts`, unrelated to this diff) |

**Open gaps**

- The two org-creation actions' new `identifyPostHogPerson(ctx.user.id, await getOrganizationRolePersonProperties(ctx.user.id))` calls (`apps/web/modules/organization/actions.ts`, `apps/web/app/setup/organization/create/actions.ts`) have no test file at all — none existed for these actions before this PR either, so the call is only checked by TypeScript, not exercised by a running test.
- The client-side change (`PostHogGroupIdentify`, the workspace layout wiring) that calls `posthog.setPersonProperties({ organization_roles, organization_count })` on every workspace load was not exercised in a browser — no e2e or manual screenshot confirms the property actually lands in PostHog with the right shape. `.tsx` files aren't unit-tested per repo convention, and no e2e covers this analytics identify.
- No live PostHog project was queried (e.g. via HogQL) to confirm the properties arrive with the expected shape end-to-end.

**Risks**

- `identifyPostHogPerson`'s properties type (`apps/web/lib/posthog/capture.ts`) was widened to accept an array of objects, not just scalars. Every other call site still passes flat scalar properties (SSO provisioning, both org-creation actions, signup) — confirmed by grep — so this is additive, but a future call site could now pass a shape PostHog doesn't expect without a type error catching it.
- `getMembershipsByUserId` is wrapped in `reactCache`, matching the existing `getMembershipByUserIdOrganizationId` pattern in the same file — a stale per-request cache could return outdated membership rows if a membership is modified and re-read within the same request. This mirrors existing behavior in the file rather than introducing a new risk.

---

> [!NOTE]
> **AI model used** — `claude-sonnet-5`, reasoning effort `unknown`.


jobenjada:
@harshsbhat can you add a screenshot of the before / after on PostHog? Would help to understand the outcome when reviewing this

jobenjada (changes requested):
Compared this against every other PostHog call site in the repo (`sign-in-tracking.ts`, `environmentState.ts`, `process-response-pipeline-job.ts`, `survey/editor/actions.ts`, plus the SSO identify paths). The multi-org role snapshot design itself is sound — recomputing from the DB on every call rather than patching a single mutable field is the right call given PostHog person properties are flat. Two things don't match this codebase's established PostHog conventions though:

## 1. Missing `POSTHOG_KEY` guard on the org-creation actions (real bug, not just style)

Every existing call site that needs DB work before an identify/capture follows this shape:

```js
if (!POSTHOG_KEY) return;
try { /* DB query */ ...capturePostHogEvent(...) } catch { ... }
```

e.g. [`sign-in-tracking.ts:50-55`](https://github.com/formbricks/formbricks/blob/main/apps/web/modules/auth/lib/sign-in-tracking.ts#L50-L55) queries `membership.count` for a capture property under exactly this guard, and the same pattern repeats in `environmentState.ts:40`, `process-response-pipeline-job.ts:772`, and `survey/editor/actions.ts:303`.

This PR applies that guard in the workspace layout, but not in either org-creation action:

- `apps/web/modules/organization/actions.ts` (`identifyPostHogPerson(ctx.user.id, await getOrganizationRolePersonProperties(ctx.user.id));`)
- `apps/web/app/setup/organization/create/actions.ts` (same call)

Neither has a `POSTHOG_KEY` check or a try/catch around it. Two consequences:

- On a self-hosted instance with no PostHog configured, every organization creation now does an unconditional extra `membership.findMany` for nothing.
- `getOrganizationRolePersonProperties` → `getMembershipsByUserId` can throw `DatabaseError`/`UnknownError` (covered by this PR's own tests). Unguarded, a transient DB error here fails organization creation itself — which contradicts the PR description's own stated principle ("read-only analytics enrichment... must never fail the primary flow"), a principle it *does* apply correctly to the signup path and the workspace layout, just not to these two.

Worth wrapping both call sites the same way `sign-in-tracking.ts` does.

## 2. Two `identifyPostHogPerson` sites left un-instrumented

The PR description says "every identify call now sends a full snapshot," but two call sites are untouched:

- `modules/ee/sso/lib/sso-provisioning.ts:214`
- `modules/ee/sso/lib/better-auth-hooks.ts:176`

Both are explicitly commented as **"parity with the credentials sign-up path"** — they're the SSO twin of the exact signup flow this PR did update. Not fatal (the client-side `PostHogGroupIdentify` will self-heal the snapshot once the SSO user hits a workspace page), but worth flagging given the "every identify call" framing, and probably worth a one-line note in the PR description if left as-is intentionally.

Everything else lines up well with existing conventions: `reactCache` wrapping matches the sibling `getMembershipByUserIdOrganizationId`, error classification matches the file's existing pattern, `Readonly<>` props follow AGENTS.md, and the widened `identifyPostHogPerson` properties type stays scoped to identify calls (event properties remain scalar-only).

harshsbhat:
Closing since it doesn't seem to be worth adding so much code on the auth and other crucial pages.

PR #9006, opened 2026-08-27
docs(surveys): document the strict declared-field naming rule and pin it [ENG-2539]

pandeymangg:
## What & why

ENG-2539 asked for a decision on how strict the management API should be about declared field names, and the decision (with Johannes) is: **strict for new names, all three groups — back-compat via grandfathering only.** That is what the guard on this branch already enforces, so #8949 (which relaxed the API) was closed and this PR ships the part of the ticket that remained: write the decision down, pin it, and take the two review findings from #8949 that survive the reversal.

**1. Docs now state the real rule.** `SurveyHiddenFields.yml` advertised the lenient legacy charset with a vague "avoid reserved ids"; the API actually refuses `UserRegion`, `country` and `lang` alike for new names. The description now spells out the strict rule, enumerates all **17** refused link-contract names (the #8949 review found the docs listing 16 — `embed` was missing), names the auto-captured group, and explains grandfathering — including that re-creating a survey from an exported document 400s on a refused name (accepted; rename on import). The `pattern` stays lenient on purpose: it describes what *stored* surveys may hold, which grandfathered ids must keep satisfying on re-send. `hidden-fields.mdx` gets the same content as a **Naming** section in prose. `openapi.yml` regenerated with `bundle.mjs`.

**2. The 400 now names the check that fired** (carried over from #8949's `206011e`, reworked for the strict rule). Every non-reserved refusal returned `isSafeIdentifier`'s sentence — so `""`, `team size` and `user:name` all read as charset violations, and following the message sent a caller in circles. The reason now follows the error code: empty, spaces, legacy charset, strict charset, duplicate.

**3. The docs enumeration is pinned against drift.** The refused lists are hand-maintained and a name added to `FORBIDDEN_IDS` fails no other check — exactly how `embed` went missing. A new test asserts every name in `RESERVED_DECLARED_FIELD_NAMES` appears (backtick-delimited, so short names like `end` can't pass off prose) in both docs surfaces. It deliberately couples to the docs' location — if a doc moves, the failure IS the signal its enumeration must move too. (This was flagged as an open gap on #8949; taking it here since it guards this PR's own content.)

**No behavior change** beyond the error-message wording: the strict rule itself has been enforced on this branch since ENG-1839 (#8897). The decision-pinning tests for that already exist and stay untouched — the guard suite covers all three groups + grandfathering + casing, and `service.test.ts` / `patch.test.ts` cover create, update and the v3 patch seam including grandfathered re-sends.

## Linear ticket

Fixes ENG-2539 — https://linear.app/formbricks/issue/ENG-2539/decide-how-strict-the-management-api-should-be-about-declared-field

## How this was tested

- `@formbricks/types` ✅ 370 tests (30 in the two touched suites) · `turbo run typecheck lint --filter=@formbricks/types` ✅ · prettier ✅
- `bundle.mjs` regenerate + `--lint` + `--check` all green (the bundle is what CI's api-v3-spec check verifies)
- New tests checked to **fail without their change**:
  - the three reason-follows-the-code tests go red with the per-code map reverted to the old single sentence
  - the anti-drift test goes red with `embed` removed from the YML enumeration (the exact historical gap)
- The pre-existing "Team Size → lowercase letter" test was re-pointed at `UserRegion`: a space is refused by `HasSpaces` *before* `isSafeIdentifier` is consulted, so the old spelling never exercised the reason it asserted — it passed only while every non-reserved code returned the same sentence.

## Breaking changes

None — docs, tests, and 400-message wording only. No accepted or refused payload changes.

## QA / Test Plan

**How to test**

- [ ] `POST /api/v1/management/surveys` with `hiddenFields.fieldIds: [""]` → 400 whose message says the name *must not be empty* (previously: a sentence about lowercase letters).
- [ ] Same with `["team size"]` → 400 saying it *must not contain spaces*.
- [ ] Same with `["user:name"]` → 400 saying *letters, numbers, underscores and hyphens*.
- [ ] Same with `["UserRegion"]` → 400 saying it *must start with a lowercase letter…*.
- [ ] Same with `["lang"]` → 400 saying the name is *never filled from the URL* (unchanged).
- [ ] A survey that already declares `country`: `PUT` it back unchanged → 200 (grandfathering, unchanged behavior).
- [ ] Docs: the v3 reference for `SurveyHiddenFields` lists all 17 refused names (including `embed`) and describes grandfathering; `hidden-fields.mdx` has the same under **Naming**.

**Preconditions / test data**

- A v1 management API key. The grandfathered step needs a survey stored with a reserved name (create the rows via DB/API — the editor refuses the name).

**Risks & regressions**

- The 400 message wording changed for empty/space/charset refusals — anything parsing those strings (nothing known) would notice. Error *codes* and status are untouched.
- The anti-drift test couples `packages/types` to two docs paths; moving either file breaks it deliberately.

- none

### DIFF
--- a/docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
+++ b/docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
@@ -11,12 +11,14 @@ description: >
   link-survey system params and internal ids — `userId`, `source`, `suid`, `end`, `start`,
   `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`,
   `preview`, `startAt`, `skipPrefilled`, `offlineSupport` (matched case-insensitively) — and the
-  names of auto-captured system fields such as `url`, `country`, `language`, `timezone` or
-  `utmSource`, which every survey can already read without declaring them.
+  names of auto-captured system fields, which every survey can already read without declaring
+  them: `source`, `url`, `country`, `action`, `browser`, `os`, `deviceType`, `ipAddress`, `finished`, `language`, `locale`, `responseId`, `surveyId`, `durationSeconds`, `startedAt`, `finishedAt`, `pagePath`, `pageReferrer`, `utmSource`, `utmMedium`, `utmCampaign`, `utmTerm`, `utmContent`, `screenWidth`, `screenHeight`, `viewportWidth`, `viewportHeight`, `timezone`.
 
 
   **Ids a survey already declares are grandfathered**: they always load, and re-sending the survey
-  unchanged never fails validation — only *adding* such a name as a new field is refused. Note this
+  unchanged never fails validation — only *adding* such a name as a new field is refused. Deleting a
+  grandfathered field spends the reprieve: once it is gone from the saved survey, the name can no
+  longer be re-added. Note this
   means re-creating a survey from an exported document can fail with a 400 when the export declares
   a name the rule refuses; rename the field in the payload before importing. The `pattern` below
   describes what stored surveys may hold (the pre-existing lenient charset), not what a new id may
--- a/docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
+++ b/docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
@@ -1,13 +1,26 @@
 type: object
 description: >
-  Hidden fields, sometimes called embedded data in other survey products. Field
-  ids are stable
+  Hidden fields, sometimes called embedded data in other survey products. Field ids are stable
+  public identifiers and may be referenced by logic, recall, quotas, integrations, and response
+  data.
 
-  public identifiers and may be referenced by logic, recall, quotas,
-  integrations, and response data.
 
-  Use only letters, numbers, underscores, and hyphens; avoid spaces and reserved
-  ids.
+  **Naming rule for new field ids** — the same rule the survey editor applies (ENG-2539): a new id
+  must start with a lowercase letter and then contain only lowercase letters, numbers and
+  underscores (`^[a-z][a-z0-9_]*$`), and must not take a reserved name. Reserved are the
+  link-survey system params and internal ids — `userId`, `source`, `suid`, `end`, `start`,
+  `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`,
+  `preview`, `startAt`, `skipPrefilled`, `offlineSupport` (matched case-insensitively) — and the
+  names of auto-captured system fields such as `url`, `country`, `language`, `timezone` or
+  `utmSource`, which every survey can already read without declaring them.
+
+
+  **Ids a survey already declares are grandfathered**: they always load, and re-sending the survey
+  unchanged never fails validation — only *adding* such a name as a new field is refused. Note this
+  means re-creating a survey from an exported document can fail with a 400 when the export declares
+  a name the rule refuses; rename the field in the payload before importing. The `pattern` below
+  describes what stored surveys may hold (the pre-existing lenient charset), not what a new id may
+  look like.
 required:
   - enabled
 properties:
--- a/docs/surveys/general-features/hidden-fields.mdx
+++ b/docs/surveys/general-features/hidden-fields.mdx
@@ -20,6 +20,21 @@ icon: "eye-slash"
 
 ![Filled Hidden Fields](/images/surveys/general-features/hidden-fields/filled-hidden-fields.webp)
 
+## Naming
+
+Field ids follow one rule everywhere — the editor and the management API refuse the same names (ENG-2539):
+
+- A **new** field id must start with a lowercase letter and then contain only lowercase letters, numbers and underscores: `^[a-z][a-z0-9_]*$`. So `page_type`, not `pageType` or `Page Type`.
+- **Reserved names are refused in any casing.** These are the link-survey system params and internal ids: `userId`, `source`, `suid`, `end`, `start`, `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`, `preview`, `startAt`, `skipPrefilled`, `offlineSupport`. A field with one of these names could never be filled from a survey URL.
+- **Auto-captured field names are refused too** (`url`, `country`, `language`, `timezone`, `utmSource`, …): every survey can already read those by name without declaring them, and a second field under the same name would be ambiguous in recall and logic.
+
+<Note>
+  Ids a survey **already declares keep working**: they load, collect values and can be re-saved unchanged —
+  only adding such a name as a *new* field is refused. One consequence: re-creating a survey from an exported
+  document fails with a 400 when the export declares a refused name. Rename that field in the payload before
+  importing.
+</Note>
+
 ## Set Hidden Field via URL
 
 Single Hidden Field:
@@ -37,13 +52,14 @@ https://formbricks.com/s/clin34bjy?screen=landing_page&job=Founder
 ## Set Hidden Fields via SDK
 
 <Note>
-  We are reworking how to add Hidden Fields via SDK moving away from binding them to Actions over to Context. Until then, we will **continue to support the current approach for the JS SDK**. However, we don't support Hidden Fields for the Android and iOS SDKs.
+  We are reworking how to add Hidden Fields via SDK moving away from binding them to Actions over to Context.
+  Until then, we will **continue to support the current approach for the JS SDK**. However, we don't support
+  Hidden Fields for the Android and iOS SDKs.
 </Note>
 
 ```js
-formbricks.track("action_name", {hiddenFields: {myField: "value"}})
+formbricks.track("action_name", { hiddenFields: { myField: "value" } });
 ```
-
 
 ## View Hidden Fields in Responses
 
--- a/docs/surveys/general-features/hidden-fields.mdx
+++ b/docs/surveys/general-features/hidden-fields.mdx
@@ -26,13 +26,14 @@ Field ids follow one rule everywhere — the editor and the management API refus
 
 - A **new** field id must start with a lowercase letter and then contain only lowercase letters, numbers and underscores: `^[a-z][a-z0-9_]*$`. So `page_type`, not `pageType` or `Page Type`.
 - **Reserved names are refused in any casing.** These are the link-survey system params and internal ids: `userId`, `source`, `suid`, `end`, `start`, `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`, `preview`, `startAt`, `skipPrefilled`, `offlineSupport`. A field with one of these names could never be filled from a survey URL.
-- **Auto-captured field names are refused too** (`url`, `country`, `language`, `timezone`, `utmSource`, …): every survey can already read those by name without declaring them, and a second field under the same name would be ambiguous in recall and logic.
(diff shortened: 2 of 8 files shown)


docs/surveys/general-features/hidden-fields.mdx
```diff
@@ -37,14 +52,15 @@ https://formbricks.com/s/clin34bjy?screen=landing_page&job=Founder
...
 <Note>
-  We are reworking how to add Hidden Fields via SDK moving away from binding them to Actions over to Context. Until then, we will **continue to support the current approach for the JS SDK**. However, we don't support Hidden Fields for the Android and iOS SDKs.
+  We are reworking how to add Hidden Fields via SDK moving away from binding them to Actions over to Context.
+  Until then, we will **continue to support the current approach for the JS SDK**. However, we don't support
+  Hidden Fields for the Android and iOS SDKs.
 </Note>
 
 ```js
-formbricks.track("action_name", {hiddenFields: {myField: "value"}})
+formbricks.track("action_name", { hiddenFields: { myField: "value" } });
```
itsjavi: The page now contradicts itself. Line 29 says "So `page_type`, not `pageType` or `Page Type`" — and this example, 32 lines later, declares `myField`. `myField` is refused by the rule this PR just documented (`isSafeIdentifier` fails on the uppercase `F`), so a reader who copies it gets the very 400 the new section is meant to prevent.

Worth fixing here specifically because the diff already rewrote this line (prettier reflow of the object literal), so it was read and left as-is rather than simply missed.

```suggestion
formbricks.track("action_name", { hiddenFields: { my_field: "value" } });
```

---

packages/types/surveys/declared-field-docs.test.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Deliberately couples this test to the docs' location and to the backtick convention: names appear
+ * as `` `name` `` in both files, and matching on the delimited form is what keeps short names like
+ * `end` or `start` from passing vacuously off ordinary prose. If a doc moves, this failing IS the
+ * signal that its enumeration needs to move with it.
+ */
+describe("the refused-name enumeration in the docs matches RESERVED_DECLARED_FIELD_NAMES", () => {
+  const read = (relativeToRepoRoot: string): string =>
+    readFileSync(path.resolve(process.cwd(), "../..", relativeToRepoRoot), "utf8").toLowerCase();
+
+  const surfaces = [
```
itsjavi, line 27: The decision is "strict for new names, **all three groups**", but only the hidden-fields surface got written down — and the variables surface is not merely silent, it currently documents a pattern the API refuses.

`collectDeclaredFieldNames` maps `variables` too, so `validateNewDeclaredFieldNames` applies the identical rule to a new variable name. Probed on this branch (`existing: []`):

```
1foo    -> not_safe_identifier      _foo    -> not_safe_identifier
country -> reserved                 lang    -> reserved       browser -> reserved
```

Meanwhile `CreateSurveyTextVariable.yml` and `CreateSurveyNumberVariable.yml` both declare:

```yaml
name:
  pattern: ^[a-z0-9_]+$
  description: Unique variable name. Lowercase letters, numbers, and underscores only.
```

Two problems with that on the create path:

1. `^[a-z0-9_]+$` **accepts** `1foo` and `_foo` (it is `isLegacyVariableName`, which has no leading-letter requirement) — both of which `POST /api/v3/surveys` 400s. The contract advertises names the API rejects.
2. Reserved names are not mentioned at all, so nothing warns a caller off a variable named `country` or `lang`.

Unlike `SurveyHiddenFields.yml` — which genuinely has to stay lenient because it is shared by `CreateSurveyRequest`, `PatchSurveyRequest` *and* `SurveyResource`, so the prose explanation was the right call there — variables have separate `Create*` schemas. The pattern can actually be tightened to `^[a-z][a-z0-9_]*$` on `CreateSurveyTextVariable.yml` / `CreateSurveyNumberVariable.yml` while `SurveyTextVariable.yml` / `SurveyNumberVariable.yml` keep the lenient one for grandfathered names.

And adding those two files to `surfaces` here is what stops the same drift the test was written to catch — right now the pinning covers the one surface that was fixed by hand.

---

docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
```diff
@@ -1,13 +1,26 @@
...
+  underscores (`^[a-z][a-z0-9_]*$`), and must not take a reserved name. Reserved are the
+  link-survey system params and internal ids — `userId`, `source`, `suid`, `end`, `start`,
+  `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`,
+  `preview`, `startAt`, `skipPrefilled`, `offlineSupport` (matched case-insensitively) — and the
+  names of auto-captured system fields such as `url`, `country`, `language`, `timezone` or
+  `utmSource`, which every survey can already read without declaring them.
+
+
+  **Ids a survey already declares are grandfathered**: they always load, and re-sending the survey
+  unchanged never fails validation — only *adding* such a name as a new field is refused. Note this
```
itsjavi: Grandfathering is **per-save, not permanent**, and this paragraph reads as promising the stronger thing.

`existing` is derived from the survey's *current* declared names at every seam (`service.ts:383`, `patch.ts:202`), so the reprieve lasts exactly as long as the field does: delete a grandfathered `country`, save, and the next write sees an `existing` set without it and refuses the name like any other new declaration.

`declared-field-guard.ts` already calls this out as a decision rather than an accident — "Grandfathering is per-save, not permanent — and that is deliberate", and explicitly "Someone who deleted the field yesterday and is re-adding it today is the one case where that sentence reads as contradicting them". Since this PR's job is to write the decision down, the docs shouldn't be the one place that omits it. One clause covers it, e.g. after "only *adding* such a name as a new field is refused":

> Deleting a grandfathered field spends the reprieve: once it is gone from the saved survey, the name can no longer be re-added.

Same sentence is needed in the `hidden-fields.mdx` `<Note>` (lines 31–36), which mirrors this wording.

---

docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
```diff
@@ -1,13 +1,26 @@
...
 
-  Use only letters, numbers, underscores, and hyphens; avoid spaces and reserved
-  ids.
+  **Naming rule for new field ids** — the same rule the survey editor applies (ENG-2539): a new id
+  must start with a lowercase letter and then contain only lowercase letters, numbers and
+  underscores (`^[a-z][a-z0-9_]*$`), and must not take a reserved name. Reserved are the
+  link-survey system params and internal ids — `userId`, `source`, `suid`, `end`, `start`,
+  `welcomeCard`, `hidden`, `verifiedEmail`, `multiLanguage`, `embed`, `verify`, `suToken`, `lang`,
+  `preview`, `startAt`, `skipPrefilled`, `offlineSupport` (matched case-insensitively) — and the
+  names of auto-captured system fields such as `url`, `country`, `language`, `timezone` or
```
itsjavi: The auto-captured group is the larger half of the refusal surface and it is the half left as "such as".

Counting the two sets `validateId`'s strict branch consults: `RESERVED_DECLARED_FIELD_NAMES` is 17 names, `RESERVED_FIELD_NAMES` is 27, overlapping only on `source` — 43 refused names in total. This enumerates the 17 exactly (and the new test pins them), then names five of the 27.

The consequence is a caller who sends `browser`, `os`, `finished`, `responseId`, `durationSeconds`, `viewportWidth`… gets a hard 400 whose message says "it names an auto-captured system field that every survey can already read by name" — and has nowhere to go. `grep -rl "auto-captured\|utmSource" docs/ --include="*.mdx"` returns only `hidden-fields.mdx`, so there is no page listing them and nothing here to link to.

Two options, either is fine:

- Enumerate all 27 the way the 17 are enumerated, and add `RESERVED_FIELD_NAMES` to the new anti-drift test so both halves are pinned rather than one.
- Keep "such as", but say the full set is the Tier-1 reserved-field catalog and point at wherever it will be documented — the reference is more useful with a destination than with five examples.

Flagging it because "pinned against drift" is one of the PR's three headline claims, and as it stands the pinning covers 17 of 43 while the other 26 stay prose that no test can hold to.

---

itsjavi (commented):
Checked every factual claim in the docs against the code on this branch. **The rule as written is accurate** — I could not find a wrong statement in it. Four things to fix before this is the decision-of-record it's meant to be; all in the docs, none in the guard.

<details>
<summary><strong>What I verified (all ✅)</strong></summary>

| Claim | Evidence |
| --- | --- |
| Strict rule is `^[a-z][a-z0-9_]*$` | exactly `isSafeIdentifier` — leading `[a-z]` test + `^[a-z0-9_]+$` |
| 17 reserved link-contract names, `embed` included | `FORBIDDEN_IDS` (11) + `LINK_SURVEY_SYSTEM_PARAMS` (6) = 17; names match one-for-one |
| Matched case-insensitively | `validation.ts:405` |
| `source` placed in the link-contract group | matches `describeReservedReason`'s deliberate precedence |
| Auto-captured examples (`url`, `country`, `language`, `timezone`, `utmSource`) | all present in `RESERVED_FIELD_NAMES` |
| "the same rule the survey editor applies" | both editor cards pass `requireSafeIdentifier: true` — `hidden-fields-card.tsx:212`, `survey-variables-card-item.tsx:185` |
| Grandfathering; re-send never fails | all three seams: `service.ts:383` (update, `existing` = currentSurvey), `service.ts:851` (create, `existing: []`), `patch.ts:202` |
| Lenient `pattern` left alone deliberately | correct — `SurveyHiddenFields.yml` is `$ref`'d by `CreateSurveyRequest`, `PatchSurveyRequest` **and** `SurveyResource`, so one schema serves both roles |
| Reason follows the check that fired | `validateId` order is Empty → Duplicate → Reserved → HasSpaces → InvalidChars → NotSafeIdentifier; each maps to its own clause |
| `Team Size` → `UserRegion` re-point | a real fix, not cosmetic — `Team Size` hits `HasSpaces` first, so the old spelling never reached the reason it asserted |
| Bundle regenerated | `bundle.mjs --check` → in sync |
| New tests | 30 pass; deleting `` `embed` `` from the YML turns the anti-drift test red (`expected [ 'embed' ] to deeply equal []`), i.e. it catches the exact historical gap |

</details>

The four inline comments, roughly in order of how cheap they are to fix:

1. **`hidden-fields.mdx:61` uses `myField`** — the rule this PR documents refuses it, 32 lines below where it says "not `pageType`". The diff rewrote that line.
2. **Grandfathering is per-save** — deleting a grandfathered field spends the reprieve. The guard's docstring flags this explicitly; the docs inherited the sentence without the caveat.
3. **Variables are the group left undocumented** — and their `Create*` schemas advertise `^[a-z0-9_]+$`, which accepts `1foo`/`_foo` that the API 400s, with no mention of reserved names. The new anti-drift test doesn't cover those files either.
4. **The auto-captured half is 26 of the 43 refused names** and is neither enumerated nor linked, so the pinning covers the smaller half.

None of these are behaviour problems, so nothing here blocks the epic — but 1 and 3 are cases where following the docs produces a 400, which is the failure mode the PR set out to remove. Happy to re-review as soon as they're in.

---

docs/surveys/general-features/hidden-fields.mdx
pandeymangg: Taken in ef67549e6 — your suggestion verbatim (`my_field`).

packages/types/surveys/declared-field-docs.test.ts
pandeymangg: Taken — and your schema-graph read was right, verified before tightening: `CreateSurveyVariable` is referenced only by `CreateSurveyRequest`; PATCH and the resource go through the lenient `SurveyVariable`, so grandfathered names still validate on re-send. Both `Create*Variable` schemas now carry `^[a-z][a-z0-9_]*$`, a description stating the create-only scope, and the full refused-name enumeration — and both files joined the anti-drift test's surfaces.

docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
pandeymangg: Taken — the per-save clause ("Deleting a grandfathered field spends the reprieve…") is now in both the YML and the mdx Note, matching the guard's own doc.

docs/api-v3-reference/src/components/schemas/SurveyHiddenFields.yml
pandeymangg: Took the first option: all 28 catalog names (incl. `locale`, picked up by merging the epic in) are enumerated in the YML, the mdx and both `Create*Variable` schemas, and the anti-drift test now iterates BOTH sets across all four surfaces — 45 names × 4 files. Fail-checked by dropping `locale` from one list: the test goes red naming exactly the missing name. "Pinned against drift" is now true of the whole refusal surface, not 17/43 of it.

itsjavi (approved):

PR #9010, opened 2026-08-27
fix(api): clamp workspace-team writes to the API key creator's role (ENG-2329)

pandeymangg:
Fixes ENG-2329

## What & why

**Was:** any organization API key with write access could create, update and delete workspace-team permission grants, whatever its creator was allowed to do — so a key whose creator was later demoted, removed or deleted kept granting a team `manage` on any workspace, and kept revoking it.

**Now:** those writes resolve the creator's current role and return `403` unless it clears the user-management floor, matching the owner/manager clamp the settings path applies to workspace access changes.

- The v2 `users` and `teams` routes have carried this check since #8634; `workspace-teams` was the last organization route without it.
- `GET` is untouched — it needs read access only.

Reported by Yash Shendge (GitHub: `ashrexon`).

## Where to look

- [`utils.ts:107`](https://github.com/formbricks/formbricks/blob/f3ec8e2f/apps/web/modules/api/v2/organizations/%5BorganizationId%5D/workspace-teams/lib/utils.ts#L107) — the clamp, and why it precedes the id lookup
- [`utils.test.ts:119`](https://github.com/formbricks/formbricks/blob/f3ec8e2f/apps/web/modules/api/v2/organizations/%5BorganizationId%5D/workspace-teams/lib/tests/utils.test.ts#L119) — the 403-before-404 ordering case

## Breaking changes

| `POST`/`PUT`/`DELETE /api/v2/organizations/{organizationId}/workspace-teams` | `200` for any organization key with write access | `403` unless the creator can manage organization users | Installs automating grants with a key whose creator was demoted, removed, or predates `createdBy` | Re-issue the key from an owner or manager account |

## Migrations & env

- none

## How this was tested

Rerun: `pnpm --filter=@formbricks/web test "workspace-teams/lib/tests"`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Demoted, removed or `createdBy`-less creator refused | unit (red on main) | `forbidden` with field `apiKey` |
| `403` decided before the team/workspace lookup | unit (red on main) | `organization.findFirst` never called — no id oracle |
| Failed role lookup returns a `Result`, not a throw | unit (mutation) | `internal_server_error`; red with the catch at `utils.ts:115` removed |
| Owner and manager creators still allowed | unit (guard) | grant proceeds, lookup still runs |
| Team or workspace outside the organization still `404` | unit (guard) | unchanged by the clamp |

<details>
<summary>Revert check</summary>

Against `main`'s `lib/utils.ts` with the new spec kept, 6 of its 9 cases fail; the 3 that stay green are the two guard rows above.

```
× refuses the write when the key creator has been demoted to member
× refuses the write when the key creator has been demoted to billing
× refuses the write when the key creator is no longer a member of the organization
× refuses the write for a legacy key that records no creator
× returns internal_server_error when the creator role lookup fails
× does not look up the team or workspace when the creator role check fails
Tests  6 failed | 3 passed (9)
```

Removing only the `try`/`catch` reds exactly one case — the `unit (mutation)` row.

</details>

**Open gaps**

- Not exercised against a live instance: these endpoints have no Playwright or Schemathesis coverage to extend.

**Risks**

- `checkAuthenticationAndAccess` is shared by all three writes; a future read path calling it would inherit the manage-users requirement.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `xhigh`.

### DIFF
--- a/apps/web/modules/api/v2/organizations/[organizationId]/workspace-teams/lib/tests/utils.test.ts
+++ b/apps/web/modules/api/v2/organizations/[organizationId]/workspace-teams/lib/tests/utils.test.ts
@@ -1,0 +1,127 @@
+import { type Mock, beforeEach, describe, expect, test, vi } from "vitest";
+import { prisma } from "@formbricks/database";
+import { OrganizationRole } from "@formbricks/database/prisma";
+import { TAuthenticationApiKey } from "@formbricks/types/auth";
+import { checkAuthenticationAndAccess } from "../utils";
+
+vi.mock("react", async () => {
+  const actual = await vi.importActual("react");
+  return { ...actual, cache: vi.fn((fn: unknown) => fn) };
+});
+
+vi.mock("@/lib/constants", () => ({
+  USER_MANAGEMENT_MINIMUM_ROLE: "manager",
+}));
+
+vi.mock("@/modules/api/v2/management/lib/utils", () => ({
+  pickCommonFilter: vi.fn(),
+  buildCommonFilterQuery: vi.fn(),
+}));
+
+vi.mock("@formbricks/database", () => ({
+  prisma: {
+    apiKey: { findUnique: vi.fn() },
+    membership: { findUnique: vi.fn() },
+    organization: { findFirst: vi.fn() },
+  },
+}));
+
+const apiKeyFindUnique = prisma.apiKey.findUnique as unknown as Mock;
+const membershipFindUnique = prisma.membership.findUnique as unknown as Mock;
+const organizationFindFirst = prisma.organization.findFirst as unknown as Mock;
+
+// An org-scoped key that already clears every check the write routes made before this guard existed:
+// the organization id matches and `accessControl.write` is granted. So every case below turns on the
+// creator's current role alone, which is the whole point of the clamp.
+const authentication: TAuthenticationApiKey = {
+  type: "apiKey",
+  apiKeyId: "apiKey123",
+  organizationId: "org123",
+  workspacePermissions: [],
+  organizationAccess: { accessControl: { read: true, write: true } },
+};
+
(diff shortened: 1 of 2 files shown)


mattinannt (changes requested):
@pandeymangg thanks for the PR. Reading through the PR description, I don't think this is the behavior we want. It wasn't introduced in this PR, but either way I think this should be changed rather than doubled down on.

API keys are owned by the organization in my opinion. So there should be no connection to the creator here. If e.g. a DevOps engineer sets up an API key to sync Azure Entra ID users / teams with Formbricks and that engineer leaves the company, the API key and the sync should still work.

AI proposed solution:

- Gate at **mint time**, not request time: `createApiKeyAction` refuses `accessControl.write` unless the creator clears `USER_MANAGEMENT_MINIMUM_ROLE`. Today `add-api-key-modal.tsx` offers that toggle to any owner/manager without consulting the floor at all — that's the actual hole #8438 found.
- Snapshot the granted authority on the key (e.g. `ApiKey.maxAssignableRole`, set from the creator's role at creation). A manager-minted key stays capped at `member` forever, so privilege escalation is still blocked without any live creator lookup.
- Request time then reads the key only — drop `canManageOrganizationUsers` from all four routes (`users`, `teams`, `teams/[teamId]`, `workspace-teams`). Granting a team access to a workspace has no escalation dimension; `accessControl.write` already means access management.
- Decide what to do with legacy keys (`createdBy: null`), which 403 permanently today. Only owners/managers could ever mint a key, so backfilling to `manager` is defensible — but it's a call to make, not a derivation.

pandeymangg:
closing this now in favor of implementing https://linear.app/formbricks/issue/ENG-3075/gate-organization-api-key-write-access-at-mint-time-against-user

PR #9016, opened 2026-08-27
fix(ci): complete Linear releases independently of the Helm appVersion PR

mattinannt:
Fixes ENG-2475

## What & why

Stable releases shipped their Docker and Helm artifacts, then left the Linear release open. `linear-release-complete` gated itself on jobs that do not publish anything for the tag being released, and a failed *or skipped* dependency skips the dependent job. Two separate dependencies each did it:

- `update-helm-app-version` **fails** on every stable release, because `vars.HELM_APP_VERSION_PR_APP_CLIENT_ID` / `secrets.HELM_APP_VERSION_PR_APP_PRIVATE_KEY` are unset.
- `move-stable-tag`'s called job is **skipped by design** for any stable release that is not the latest — every patch on an older line — since it is gated on `!is_prerelease && make_latest`.

Observed, all with every artifact job green:

| Run | Version | `update-helm-app-version` | `move-stable-tag` | `linear-release-complete` |
| [29918121071](https://github.com/formbricks/formbricks/actions/runs/29918121071) | 5.1.5 (stable, not latest) | failure | **skipped** | **skipped** |
| [31378131643](https://github.com/formbricks/formbricks/actions/runs/31378131643) | 5.3.3 | failure | — | **skipped** |
| [31487401618](https://github.com/formbricks/formbricks/actions/runs/31487401618) | 5.3.4 | failure | — | **skipped** |
| [32969368595](https://github.com/formbricks/formbricks/actions/runs/32969368595) | 5.4.0 | failure | — | **skipped** |

RC runs looked fine only because that job is gated on `!prerelease`.

**Where to look**

- `.github/workflows/formbricks-release.yml:275-286` — **Was:** `needs:` carried `update-helm-app-version` and `move-stable-tag`. **Now:** the three jobs that publish artifacts for this tag (`docker-build-community`, `docker-build-cloud`, `helm-chart-release`), all still required to succeed. Removal over a status override: an `always()`-style condition would have to keep distinguishing a by-design skip from a real failure, for the same outcome.
- `.github/workflows/formbricks-release.yml:310`, `:317` — **Was:** `version: ${{ github.event.release.tag_name }}`. **Now:** `version: ${{ needs.docker-build-community.outputs.VERSION }}`. Same tag, `v`-stripped and SemVer-validated by `release-docker-github.yml`, whose job is already in `needs:`, so the value cannot be empty or malformed. The raw tag would silently create a second, `v`-prefixed Linear release the first time someone tags `v5.6.0`.
- `.github/workflows/formbricks-release.yml:307`, `:314`, `.github/workflows/linear-release.yml:28` — **Was:** `@0353b5fa…  # v0.7.0`. **Now:** `@17b8c24f…  # v0.15.1`. The inline comment already described v0.15.1 semantics ("`complete` only looks up a release by version, it never assigns one") while the workflow ran v0.7.0. Under v0.15.1 a versioned sync targets that version *or creates it*, which is what makes patch releases work — `5.3.4` never had a planned train of its own.
- `.github/workflows/linear-release-smoke.yml` — new dry-run canary on `workflow_dispatch` and `push` to `main`. Deliberately **not** on `pull_request`: that event runs the PR branch's own copy of the workflow, so anyone able to push a branch could drop `dry_run` or add an exfiltration step and reach a pipeline-mutating Linear key. Its header states plainly what a green run does and does not prove.
- `docker/__tests__/release-workflows.test.ts` — new. Asserts the wiring, including both `needs` exclusions and the no-`pull_request` rule, so neither regression returns quietly.

### Linear state, reconciled separately

Already reconciled: the legacy catch-all (399 issues across several shipped versions) and the planned `5.2` / `5.3`, all canceled with descriptions preserved, no history deleted. `5.4.0` is now Released against tag commit `286958e`.

Its issue count is **not** clean 5.4.0 scope. The pipeline had no frozen stage when `release/5.4` was cut, so `main` commits landing after the cut kept syncing into the train — the count went 451 → 457 while this PR was being written. The GitHub release body is the accurate changelog for 5.4.0, not that issue list.

<details>
<summary>Branch-cut runbook — for the Notion Release QA handbook (@Dhruwang)</summary>

The pipeline needs a frozen stage, which cannot be created through the Linear API — release CRUD is exposed, release *stages* are not:

1. Linear → Settings → Release pipelines → **Formbricks** → add a stage named `Code Freeze`, type *started*, marked **frozen**.
2. At each `release/x.y` cut, move the current release to `Code Freeze` **before** starting the next train on `main`.

Skipping step 2 is what produced 5.4.0's inflated issue list: the train stayed open on `main` while the release branch was being prepared, so post-cut commits leaked into it.

</details>

## Breaking changes

None — CI wiring plus a test. No API, SDK, route, env var, config key or default changes; nothing outside this repo reaches any of it, and self-hosters consume none of these workflows. The new smoke-workflow input is optional, and `LINEAR_ACCESS_KEY` was already used by these workflows before this PR.

## Migrations & env

- none

Worth flagging but **not fixed here**: `HELM_APP_VERSION_PR_APP_CLIENT_ID` (repo variable) and `HELM_APP_VERSION_PR_APP_PRIVATE_KEY` (repo secret) are unset, which is why `update-helm-app-version` fails. After this PR that no longer blocks Linear completion, but the Helm appVersion PR still won't be opened until someone configures them.

## How this was tested

**Rerun:** `pnpm --filter @formbricks/docker test` — every row below is a test in `docker/__tests__/release-workflows.test.ts` and runs under that one command. Red-on-main rows were observed by running it with the workflow files at `origin/main`; mutation rows by editing the named `file:line`, rerunning, then `git checkout -- .github/workflows/`.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Completion not gated on the appVersion PR job | unit (red on main) | `completes the Linear release once the published artifacts are out` — on `main`: `expected [ 'docker-build-community', …(4) ] to not include 'update-helm-app-version'`. Green on this head. |
| Completion not gated on the conditionally-skipped stable-tag job | unit (mutation) | Same test. Re-adding `- move-stable-tag` to the `needs` list at `formbricks-release.yml:286` fails it: `expected [ 'docker-build-community', …(3) ] to not include 'move-stable-tag'`. Observed. |
| The credentialed dry-run never runs a PR branch's code | unit (mutation) | `keeps the credentialed smoke dry-run off pull request branches` — replacing `push:`/`branches: [main]` at `linear-release-smoke.yml:25-27` with `pull_request:` fails it: `expected { workflow_dispatch: …, …(1) } to not have property "pull_request"`. Observed. |
| Every Linear action use pinned to v0.15.1 by SHA | unit (red on main) | `pins every Linear release action use by commit SHA in %s`, 3 cases — on `main`: `expected … to contain 'linear/linear-release-action@17b8c24f…'`. Green on this head. |
| Each pin's `# v0.15.1` annotation matches its SHA | unit (mutation) | `annotates every pin with its release tag in %s` — changing the comment at `formbricks-release.yml:314` to `# v0.7.0` with the SHA untouched fails **only** this test, not the pin test. Observed. |
| One correct use cannot mask a second drifted one | unit (mutation) | Changing **only** the second use at `formbricks-release.yml:314` to the v0.7.0 SHA fails `pins every…`, `annotates every…` and `uses no other ref…`. Before this fix that mutation left both per-file assertions green — the hole it closes. Observed. |
| No stale ref of the action anywhere in `.github/workflows/` | unit (red on main) | `uses no other ref of the Linear release action across the workflows` — on `main`: `expected [ "0353b5fa…" ] to deeply equal [ "17b8c24f…" ]`. Set-equality over a directory scan, so it cannot pass vacuously. Green on this head. |
| Both steps target the released version; second is `command: complete` | unit (red on main) | `stamps the released version on Linear before completing the release` — on `main`: `expected [ "${{ github.event.release.tag_name }}", … ] to deeply equal [ "${{ needs.docker-build-community.outputs.VERSION }}", … ]`. Green on this head. |
| Prereleases still skip completion | unit (mutation) | `skips the Linear completion for prereleases` — guard; passes before and after the diff. Deleting the `if:` at `formbricks-release.yml:287` fails it. Observed. |
| Main-push train keeps syncing unversioned, full history | unit (mutation) | `keeps the unversioned Linear sync on pushes to main` — guard. Adding a `version:` under `linear-release.yml:28`'s step fails it; so does deleting `linear-release.yml:25` (`fetch-depth: 0`). Both observed. |
| The three workflows are valid YAML | unit (mutation) | `%s parses as YAML and declares jobs`, 3 cases — appending `\n  \tbroken: [` to `linear-release.yml` fails it. Observed. |

Full run on this head: 36 tests passed in the `@formbricks/docker` suite. Root `pnpm test` passed (26 turbo tasks, 8589 web tests); `pnpm lint` 0 errors; `pnpm format:check` clean.

**Open gaps**

- **Exact-version resolution and completion are not exercised anywhere before a real release.** The earlier PR-triggered dry run is gone (it reached a mutation-capable credential from any pushable branch), and it could not have proven this regardless: the pinned CLI returns from its complete command before resolving a release when `dry_run` is set, so even a nonexistent version passes. A controlled integration path would need its own non-production pipeline and credential plus teardown — a separate change, not improvised here. First real proof is the next published stable release.
- **The remaining dry-run canary proves only that the action installs, authenticates and scans.** It runs post-merge and on demand.
- **The real `main`-push sync is exercised by this PR's own merge**, not by the PR.
- **The frozen `Code Freeze` stage does not exist yet** and cannot be created through the Linear API — it needs the manual step in the fold above. Until it exists, the next branch cut will leak `main` commits into the release being prepared, exactly as 5.4.0 did.
- **The sync scan base is broken independently of this PR.** From the dry run, verbatim: *"None of the last 4 synced releases' commit SHAs exist in this repository's history. Syncing only the current commit until a scan base can be established… otherwise pass `--base-ref` to pin the scan range."* Stable releases are cut from `release/x.y`, so the tag commit stored as the baseline is never an ancestor of `main`, and each `main` push syncs only its own commit. That still feeds the train commit-by-commit, so it is not a regression here, but a `base_ref` follow-up is warranted.

**Risks**

- Linear completion can now run while `update-helm-app-version` is still failing, and while `move-stable-tag` was skipped. That is the intent — neither publishes anything for the released tag — and both remain independently visible in the run.
- The three artifact jobs are still hard dependencies, so a genuine publish failure still blocks completion.
- The smoke workflow holds `LINEAR_ACCESS_KEY` but runs only on `main` and manual dispatch, and is `dry_run` only.

---

> [!NOTE]

### DIFF
--- a/.github/workflows/formbricks-release.yml
+++ b/.github/workflows/formbricks-release.yml
@@ -269,12 +269,14 @@ jobs:
     name: Mark Linear release as complete
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    # Shipped artifacts only. update-helm-app-version opens a follow-up PR against main
+    # rather than publishing anything for this tag, so gating completion on it left Linear
+    # stuck on every stable release the appVersion PR job failed.
     needs:
       - docker-build-community
       - docker-build-cloud
       - helm-chart-release
       - move-stable-tag
-      - update-helm-app-version
     if: ${{ !github.event.release.prerelease }}
     steps:
       - name: Harden the runner
@@ -287,18 +289,23 @@ jobs:
         with:
           fetch-depth: 0
 
-      # Stamp the in-progress release with the tag version before completing it.
-      # complete only looks up a release by version, it never assigns one, so a
-      # versioned sync must run first (Linear's recommended scheduled-pipeline flow).
+      # A versioned sync targets the release carrying this exact version, or creates it when
+      # none exists - which is what makes patch releases work, since 5.3.4 never had a
+      # planned train of its own. complete then looks that exact version up; it never
+      # assigns one, so the sync has to run first.
+      #
+      # VERSION is this release's tag with any v-prefix stripped and validated as SemVer by
+      # release-docker-github.yml. Using the raw tag_name instead would silently create a
+      # second, v-prefixed Linear release the first time somebody tags v5.6.0.
       - name: Stamp Linear release version
-        uses: linear/linear-release-action@0353b5fa8c00326913966f00557d68f8f30b8b6b # v0.7.0
+        uses: linear/linear-release-action@17b8c24f8ceb2b98cabaf1965ff83c55dd596fac # v0.15.1
         with:
           access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
-          version: ${{ github.event.release.tag_name }}
+          version: ${{ needs.docker-build-community.outputs.VERSION }}
 
       - name: Complete Linear release
-        uses: linear/linear-release-action@0353b5fa8c00326913966f00557d68f8f30b8b6b # v0.7.0
+        uses: linear/linear-release-action@17b8c24f8ceb2b98cabaf1965ff83c55dd596fac # v0.15.1
         with:
           access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
           command: complete
-          version: ${{ github.event.release.tag_name }}
+          version: ${{ needs.docker-build-community.outputs.VERSION }}
--- a/.github/workflows/formbricks-release.yml
+++ b/.github/workflows/formbricks-release.yml
@@ -269,14 +269,20 @@ jobs:
     name: Mark Linear release as complete
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    # Shipped artifacts only. update-helm-app-version opens a follow-up PR against main
-    # rather than publishing anything for this tag, so gating completion on it left Linear
-    # stuck on every stable release the appVersion PR job failed.
+    # Published artifacts only, and only jobs that always run for a stable release.
+    #
+    # update-helm-app-version opens a follow-up PR against main rather than publishing
+    # anything for this tag, so gating completion on it left Linear stuck on every stable
+    # release its appVersion PR job failed.
+    #
+    # move-stable-tag is excluded for a subtler reason: its called job is gated on
+    # `!is_prerelease && make_latest`, so it is legitimately skipped for any stable release
+    # that is not the latest - every patch on an older line. A skipped dependency skips this
+    # job too, so depending on it would keep Linear stuck on exactly those releases.
     needs:
       - docker-build-community
       - docker-build-cloud
       - helm-chart-release
-      - move-stable-tag
     if: ${{ !github.event.release.prerelease }}
     steps:
       - name: Harden the runner
--- a/.github/workflows/linear-release-smoke.yml
+++ b/.github/workflows/linear-release-smoke.yml
@@ -13,9 +13,10 @@ name: Linear Release Smoke
 # only exercised by a real published release.
 #
 # TRIGGERS ARE DELIBERATELY TRUSTED-ONLY. This job holds a Linear key that can mutate the
-# pipeline, so it must never run code from a pull request branch: `pull_request` runs the PR's
-# own copy of this file, so anyone who can push a branch could drop dry_run or add an
-# exfiltration step. main and manual dispatch are the only events allowed to hold the key.
+# pipeline, so it must never run code from outside main. `pull_request` would run the PR's own
+# copy of this file, and `workflow_dispatch` runs the copy on whichever ref the caller picks,
+# so either would let a pushed branch drop dry_run or add an exfiltration step. Hence: no
+# pull_request trigger, push restricted to main, and the ref check on the job below.
 on:
   workflow_dispatch:
     inputs:
@@ -39,6 +40,9 @@ jobs:
     name: Dry-run the Linear release flow
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    # Belt to the trigger's braces: a workflow_dispatch runs the file as it exists on the ref
+    # the caller chose, so the key must not be handed to a branch's copy of these steps.
+    if: github.ref == 'refs/heads/main'
     steps:
       - name: Harden the runner
         uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
--- a/.github/workflows/linear-release-smoke.yml
+++ b/.github/workflows/linear-release-smoke.yml
@@ -1,14 +1,21 @@
 name: Linear Release Smoke
 
-# Dry-run of the Linear release flow. Nothing here mutates the pipeline: dry_run makes the
-# action scan commits and call read-only Linear APIs, then log the action it would have taken.
+# Dry-run canary for the Linear release flow. dry_run makes the action scan commits and call
+# read-only Linear APIs, then log the action it would have taken; it never mutates the pipeline.
 #
-# It exists because the real flow only runs on a published release, so a broken job graph or
-# a stale action pin stayed invisible until release day - which is how stable-release
-# completion silently skipped three releases in a row.
+# It exists because the real flow only runs on a published release, so a stale action pin or a
+# broken install stayed invisible until release day.
 #
-# So it runs on any pull request that touches the release flow, giving that PR a dry run of
-# the change it is making. Dispatch it by hand to check the pipeline at any other time.
+# WHAT THIS DOES NOT PROVE. The pinned CLI returns from its complete command before resolving a
+# release when dry_run is set, so a green run here says nothing about whether an exact version
+# resolves or completes - a nonexistent version passes too. Treat it as "the action installs,
+# authenticates and scans", not as validation of the release flow. Exact-version completion is
+# only exercised by a real published release.
+#
+# TRIGGERS ARE DELIBERATELY TRUSTED-ONLY. This job holds a Linear key that can mutate the
+# pipeline, so it must never run code from a pull request branch: `pull_request` runs the PR's
+# own copy of this file, so anyone who can push a branch could drop dry_run or add an
+# exfiltration step. main and manual dispatch are the only events allowed to hold the key.
 on:
   workflow_dispatch:
     inputs:
@@ -16,7 +23,9 @@ on:
         description: "Version to target (blank targets the active started release)"
         required: false
         type: string
-  pull_request:
+  push:
+    branches:
+      - main
     paths:
       - .github/workflows/linear-release.yml
       - .github/workflows/formbricks-release.yml
@@ -30,11 +39,6 @@ jobs:
     name: Dry-run the Linear release flow
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    # Fork pull requests get no access to LINEAR_ACCESS_KEY, so the dry run would fail on an
-    # empty key rather than tell anyone anything.
-    if: >-
-      github.event_name == 'workflow_dispatch' ||
-      github.event.pull_request.head.repo.full_name == github.repository
     steps:
       - name: Harden the runner
         uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
--- a/.github/workflows/linear-release-smoke.yml
+++ b/.github/workflows/linear-release-smoke.yml
@@ -7,9 +7,8 @@ name: Linear Release Smoke
 # a stale action pin stayed invisible until release day - which is how stable-release
 # completion silently skipped three releases in a row.
 #
-# The paths trigger is deliberate: workflow_dispatch only becomes available once a workflow is
-# on the default branch, so the push filter is what lets the commit introducing a change here
-# prove itself on a PR branch. After merge, dispatch it by hand whenever the flow changes.
+# So it runs on any pull request that touches the release flow, giving that PR a dry run of
+# the change it is making. Dispatch it by hand to check the pipeline at any other time.
 on:
   workflow_dispatch:
     inputs:
@@ -17,8 +16,10 @@ on:
         description: "Version to target (blank targets the active started release)"
         required: false
         type: string
-  push:
+  pull_request:
     paths:
+      - .github/workflows/linear-release.yml
+      - .github/workflows/formbricks-release.yml
       - .github/workflows/linear-release-smoke.yml
 
 permissions:
@@ -29,6 +30,11 @@ jobs:
     name: Dry-run the Linear release flow
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    # Fork pull requests get no access to LINEAR_ACCESS_KEY, so the dry run would fail on an
+    # empty key rather than tell anyone anything.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     steps:
       - name: Harden the runner
         uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
--- a/.github/workflows/linear-release-smoke.yml
+++ b/.github/workflows/linear-release-smoke.yml
@@ -1,0 +1,58 @@
+name: Linear Release Smoke
+
+# Dry-run of the Linear release flow. Nothing here mutates the pipeline: dry_run makes the
+# action scan commits and call read-only Linear APIs, then log the action it would have taken.
+#
+# It exists because the real flow only runs on a published release, so a broken job graph or
+# a stale action pin stayed invisible until release day - which is how stable-release
+# completion silently skipped three releases in a row.
+#
+# The paths trigger is deliberate: workflow_dispatch only becomes available once a workflow is
+# on the default branch, so the push filter is what lets the commit introducing a change here
+# prove itself on a PR branch. After merge, dispatch it by hand whenever the flow changes.
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to target (blank targets the active started release)"
+        required: false
+        type: string
+  push:
+    paths:
+      - .github/workflows/linear-release-smoke.yml
+
+permissions:
+  contents: read
+
+jobs:
+  linear-release-smoke:
+    name: Dry-run the Linear release flow
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Harden the runner
+        uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
+        with:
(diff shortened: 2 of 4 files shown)


docker/__tests__/release-workflows.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    (path) => {
+      expect(Object.keys(readWorkflow(path).jobs ?? {})).not.toHaveLength(0);
+    }
+  );
+
+  test.each([linearSyncWorkflow, formbricksReleaseWorkflow])(
+    "pins the Linear release action by commit SHA in %s",
+    (path) => {
+      expect(readText(path)).toContain(`${linearAction}@${linearActionSha} # ${linearActionVersion}`);
+    }
```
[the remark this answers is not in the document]
mattinannt: Fixed the coupling you identified, but kept the annotation check rather than dropping it — so flagging the deviation from the proposed diff.

Your diagnosis was right: one assertion conflated the executable `uses` value with its tag comment, so a comment-only edit failed the test guarding the pin. They are now two separate tests (`d533b5e`), and I verified the separation by mutation: changing `formbricks-release.yml:307`'s comment to `# v0.7.0` with the SHA untouched now fails only `annotates every pin with its release tag`, not `pins every Linear release action use by commit SHA`.

I did not remove `linearActionVersion`, because a drifted annotation is a real defect here, not an implementation detail. This repo ran the v0.7.0 pin under a comment describing v0.15.1 semantics ("`complete` only looks up a release by version, it never assigns one") for months — that mismatch is precisely what hid the bug this PR fixes. A reviewer reading `# v0.15.1` should be able to trust it.

Your follow-up review caught the stronger issue in the same code, and that is fixed too: both tests now enumerate the parsed `uses:` entries instead of matching raw text, since `formbricks-release.yml` calls the action twice and one correct use was masking a second drifted one. Confirmed by mutating only the second use.

---

.github/workflows/formbricks-release.yml
```diff
@@ -269,12 +269,14 @@ jobs:
...
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    # Shipped artifacts only. update-helm-app-version opens a follow-up PR against main
+    # rather than publishing anything for this tag, so gating completion on it left Linear
+    # stuck on every stable release the appVersion PR job failed.
     needs:
       - docker-build-community
       - docker-build-cloud
       - helm-chart-release
       - move-stable-tag
```
BhagyaAmarasinghe: [P1] Non-latest stable releases still skip this job. `move-stable-tag`'s called job is skipped when `make_latest` is false, and GitHub skips jobs that need a skipped dependency unless their `if` condition includes a status override. [Run 29918121071](https://github.com/formbricks/formbricks/actions/runs/29918121071) demonstrates this for stable 5.1.5: both `move-stable-tag` and `Mark Linear release as complete` were skipped. Please either remove this dependency or make the condition explicitly accept the expected non-latest skip while still requiring all shipped-artifact jobs to succeed.

.github/workflows/linear-release-smoke.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Dry-run the release sync
+        uses: linear/linear-release-action@17b8c24f8ceb2b98cabaf1965ff83c55dd596fac # v0.15.1
+        with:
+          access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
```
BhagyaAmarasinghe, line 60: [P1] This makes a mutation-capable Linear credential available to code from same-repository PR branches. The fork check excludes external forks, but any collaborator who can push a repository branch can edit this workflow, remove `dry_run`, or add an arbitrary exfiltration step before triggering it with a PR; `contents: read` only limits `GITHUB_TOKEN`. Please keep the credentialed smoke on a trusted event such as post-merge or `workflow_dispatch`, or use an approval-gated environment/read-only credential, and keep the PR check secretless.

.github/workflows/linear-release-smoke.yml
```diff
@@ -0,0 +1,64 @@
...
+          version: ${{ inputs.version }}
+          dry_run: "true"
+          log_level: verbose
+
+      - name: Dry-run the release completion
+        uses: linear/linear-release-action@17b8c24f8ceb2b98cabaf1965ff83c55dd596fac # v0.15.1
+        with:
+          access_key: ${{ secrets.LINEAR_ACCESS_KEY }}
+          command: complete
+          version: ${{ inputs.version }}
```
BhagyaAmarasinghe, line 70: [P2] The automatic PR smoke does not test the production versioned completion path. `inputs.version` is empty for `pull_request`; the [green smoke run](https://github.com/formbricks/formbricks/actions/runs/33080675210) used `complete --dry-run` with `no version set` and returned `release: null`. More importantly, the pinned CLI returns from `completeCommand` before calling `completeRelease` whenever `dry_run` is true, so even a nonexistent explicit version passes. Please either narrow the check's claimed coverage or add a controlled integration path that actually validates exact-version resolution/completion.

BhagyaAmarasinghe (changes requested):
Requesting changes for three release-flow issues: non-latest stable releases remain skipped, the PR-triggered smoke exposes a mutation-capable Linear key to PR-authored code, and the green smoke does not exercise exact-version completion.

.github/workflows/formbricks-release.yml
mattinannt: Correct, and fixed in `b04269b` by removing the dependency. Thank you — this was the same bug one edge over, and my test had locked it in.

Confirmed the mechanism: `move-stable-tag.yml:41` gates the called job on `!inputs.is_prerelease && inputs.make_latest`, and the caller passes `make_latest: needs.check-latest-release.outputs.is_latest == 'true'`. So every stable release that is not the latest — every patch on an older line, which is most of them — skips it, and a skipped dependency skips the dependent job.

One correction to the evidence, which strengthens rather than weakens the finding: run 29918121071 is over-determined and does not isolate the cascade. At that commit `linear-release-complete` also needed `update-helm-app-version`, which **failed** in that run, so either dependency was sufficient to skip Linear. What matters is the post-fix state: with the appVersion job removed, `move-stable-tag`'s by-design skip became the *sole* remaining blocker for exactly those releases. So the PR as it stood would have kept them broken.

I took removal over a status override. `move-stable-tag` moves a floating pointer; it publishes nothing for the tag being released, so it does not belong in a "published artifacts" gate. An override like `always() && needs.*.result != 'failure'` would also have to keep distinguishing a legitimate skip from a real failure, which is more surface for the same outcome. The three artifact jobs still must succeed.

Verified by mutation: re-adding `move-stable-tag` to that `needs` list fails `completes the Linear release once the published artifacts are out` with `expected [ 'docker-build-community', …(3) ] to not include 'move-stable-tag'`. Rerun with `pnpm --filter @formbricks/docker test`.

---

.github/workflows/linear-release-smoke.yml
mattinannt: You're right, and this was mine to get right the first time. Fixed in `b04269b`: the credentialed dry-run no longer runs on `pull_request` at all.

The hole was exactly as you describe. `pull_request` runs the PR branch's own copy of the workflow, so the fork check was answering the wrong question — it excluded external forks while leaving the key reachable by anyone who can push a repository branch and edit this file to drop `dry_run` or add an exfiltration step. `contents: read` bounds `GITHUB_TOKEN`, not `secrets.LINEAR_ACCESS_KEY`.

It now triggers on `workflow_dispatch` and `push` to `main` (still path-filtered to the three release-flow files), so only trusted refs ever hold the key, and the fork guard is gone since it no longer applies. The PR check is now secretless because there is no PR check.

`docker/__tests__/release-workflows.test.ts` asserts the invariant so it cannot regress quietly — no `pull_request` or `pull_request_target` trigger, and `push.branches` exactly `["main"]`. Verified by mutation: restoring `pull_request:` fails `keeps the credentialed smoke dry-run off pull request branches`.

Worth being explicit about the cost, since it undoes something I argued for earlier in this PR: this removes the pre-merge evidence path. The flow is now only exercised after merge or on demand, and the PR carries no dry-run of its own change. That is the right trade — a mutation-capable credential reachable from any pushable branch is not worth a green check — and your P2 finding shows the check was proving very little anyway. I have moved that to `Open gaps` rather than leaving the claim standing.

---

.github/workflows/linear-release-smoke.yml
mattinannt: Agreed, and I've narrowed the claim rather than tried to build the integration path in this PR.

The `completeCommand` early-return under `dry_run` is the part I had not established, and it is decisive: it means the completion step cannot validate version resolution *at all*, so passing an explicit version would not have rescued the check either. My description already said the run "does not prove the flow works" and noted `release: null` and the empty `inputs.version`, but it still listed the dry run as an `e2e (dry run)` coverage row, which claimed more than the mechanism can deliver.

Now (`b04269b`):

- The workflow header carries a `WHAT THIS DOES NOT PROVE` block stating that the pinned CLI returns before resolving a release under `dry_run`, so a nonexistent version passes too, and that the canary means "the action installs, authenticates and scans" — nothing about exact-version completion.
- The coverage row is gone from the table. Exact-version resolution and completion sit under `Open gaps`, first exercised by the next real published release.

I have not added a controlled integration path, and I would rather not improvise one here. Doing it honestly needs a pipeline the run may actually mutate — a non-production Linear pipeline with its own key, or a disposable release created and completed for real then cleaned up — which is a separate change with its own credential and teardown design, and larger than this fix. Happy to file it as a follow-up ticket if you think it is worth the cost; the alternative is accepting that the completion path is proven only by release day, which is the status quo this PR does not make worse.

---

docker/__tests__/release-workflows.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  // Separate from the pin above so a drifted annotation and a drifted pin fail distinguishably, and
+  // counted so one annotated line cannot vouch for an unannotated sibling. The annotation is worth
+  // asserting at all because this repo ran a v0.7.0 pin under a comment describing v0.15.1
+  // behaviour for months, which is the drift that hid the bug these tests guard.
+  test.each(releaseWorkflows)("annotates every pin with its release tag in %s", (path) => {
+    const annotated = readText(path).split(`${linearAction}@${linearActionSha} # ${linearActionVersion}`);
+
+    expect(annotated).toHaveLength(linearUses(readWorkflow(path)).length + 1);
```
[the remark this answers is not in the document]
mattinannt: Keeping this test — this duplicates the thread at #discussion_r3872584797, where you reviewed the same reasoning and agreed the annotation is a separate meaningful invariant worth its own assertion.

Briefly, so this thread stands on its own: the annotation is not incidental in this repo. A v0.7.0 pin sat under a comment describing v0.15.1 semantics for months, and that mismatch is what let the release-flow bug this PR fixes go unnoticed. A stale `# vX.Y.Z` misinforms the next reader about what the pinned action does.

The failure mode you're guarding against is already handled: this is a separate test from the SHA assertion, so a comment edit fails only `annotates every pin with its release tag` and never the pin check. That separation is mutation-verified in the linked thread. And a harmless comment edit is not a thing here — the comment's only job is to name the pinned tag, so if it changes without the SHA changing, it has become wrong.

---

.github/workflows/linear-release-smoke.yml
mattinannt: Follow-up: my first fix was incomplete, and a second push closed the rest.

Removing `pull_request` shut one door but left another open — `workflow_dispatch` runs the workflow file as it exists on whichever ref the caller names, so a pushed branch could still have its own copy of these steps handed the key. `ab2a4ee` adds `if: github.ref == 'refs/heads/main'` on the job, above both secret-consuming steps, so the trigger list is no longer the only control. The test asserts the guard as well as the triggers, and deleting the `if:` line fails it.

So the credential is now reachable only by main's copy of this file, on a path-filtered main push or a dispatch aimed at main.

---

BhagyaAmarasinghe (approved):

mattinannt:
Post-merge verification of the two `Open gaps` this PR said could only be checked after merge. Both now have real evidence, and one open gap is not reproducing.

**The `main`-push sync works on the merged workflow.** [Run 33096784426](https://github.com/formbricks/formbricks/actions/runs/33096784426) on the merge commit succeeded, as did every subsequent `Linear Release Sync` run on `main` — 30 runs, 30 successes, no skips. The active train's `commitSha` has been advancing with them.

**The smoke canary runs from `main` and resolved a real base.** [Run 33096784486](https://github.com/formbricks/formbricks/actions/runs/33096784486) fired on the merge commit and executed — the `github.ref == 'refs/heads/main'` guard let it through rather than skipping it. Verbatim from the sync step:

```
{"level":"verbose","msg":"Using base SHA from release \"New release\" (bb5b525)"}
{"level":"verbose","msg":"Range bb5b525..326769b spans 1 commits"}
{"level":"info","msg":"Found 1 commit between bb5b525 and 326769b"}
{"level":"info","msg":"[dry-run] Would sync release (server-assigned) (no version set): pull requests [#9016]"}
```

**The scan-base problem listed under `Open gaps` did not reproduce here.** There is no `warn` line anywhere in that job log — I grepped the full 143 KB for `{"level":"warn"` and got zero matches. The base SHA resolved cleanly from the started release and the range spanned exactly the merge commit, so the earlier *"None of the last 4 synced releases' commit SHAs exist in this repository's history"* warning appears to have been a consequence of the stale release records rather than of the branch-cut topology. Worth re-checking after the next `release/x.y` cut before deciding whether a `base_ref` follow-up is still warranted.

Two things this still does **not** show, unchanged from the description: the completion step remains `[dry-run] Would complete release (current release) (no version set)` with `{"release":null}`, so exact-version resolution is still only exercised by a real published release; and the action's default `cli_version` resolves to CLI `v0.15.0` under the v0.15.1 action pin, which is the action's own default rather than a mismatch introduced here.

Linear side, for the record: `5.4.0` is Released against tag commit `286958e`, and the active train is now named and versioned `5.5.0`. The frozen `Code Freeze` stage is still the one outstanding item and still needs a human in Linear — the API exposes release CRUD but not release stages.

---

PR #9022, opened 2026-08-27
fix(contacts): say workspace, not environment, in duplicate-attribute messages

jobenjada:
## What & why

**Was:** The English templates for `email_already_exists` and `userid_already_exists` said the duplicate "already exists for this environment".

**Now:** Both say workspace, matching what the check does and what the UI already shows.

- The check is workspace-scoped — `hasEmailAttribute(emailValue, workspaceId, contactId)` — and Formbricks 5 replaced the `Environment` model with `Workspace`.
- The UI renders these two conditions from `workspace.contacts.attributes_msg_*`, which already say workspace. Only the API-facing templates drifted.
- Kept as hardcoded English, not `t()`: the v2 response envelope is literal throughout, and the one `getTranslate` under an API route localizes survey content it creates, not responses.

## Where to look

- [attributes.ts:41](https://github.com/formbricks/formbricks/blob/claude/unruffled-joliot-810eed/apps/web/modules/ee/contacts/lib/attributes.ts#L41) — the two templates
- [update-user.ts:246](https://github.com/formbricks/formbricks/blob/claude/unruffled-joliot-810eed/apps/web/modules/ee/contacts/api/v1/client/%5BworkspaceId%5D/user/lib/update-user.ts#L246) — the only consumer; it filters both codes

## Breaking changes

None. `formatAttributeMessage` has one consumer, the public client route, and it drops both codes as enumeration protection — so no response any external consumer receives changes.

## Migrations & env

- none

## How this was tested

Rerun: `cd apps/web && npx vitest run modules/ee/contacts/lib/attributes.test.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Both messages name the workspace | unit (red on main) | Restoring "environment" on `attributes.ts:41-42` fails the new test |
| Interpolation replaces every placeholder | unit (guard) | `attribute_type_validation_error` renders all three params |
| Unmapped code falls back to the code | unit (guard) | Returns `some_unmapped_code` |

**Open gaps**

- No end-to-end check exists: the only consumer filters both codes, so these strings are unreachable at runtime and the unit test is the sole guard.
- `attribute_limit_exceeded` still says "attribute classes"; `MAX_ATTRIBUTE_CLASSES_PER_ENVIRONMENT` keeps its name. Both are gaps #9012 lists too.
- `en-US.json` capitalises "this Workspace" in these two keys alone, against 14 lowercase elsewhere. Templates use lowercase; locales untouched.

**Risks**

- none — no rendered surface reads these strings.

---

> [!NOTE]

### DIFF
--- a/apps/web/modules/ee/contacts/lib/attributes.test.ts
+++ b/apps/web/modules/ee/contacts/lib/attributes.test.ts
@@ -7,7 +7,7 @@ import {
   hasEmailAttribute,
   hasUserIdAttribute,
 } from "@/modules/ee/contacts/lib/contact-attributes";
-import { updateAttributes } from "./attributes";
+import { formatAttributeMessage, updateAttributes } from "./attributes";
 
 vi.mock("@/lib/constants", () => ({
   MAX_ATTRIBUTE_CLASSES_PER_ENVIRONMENT: 2,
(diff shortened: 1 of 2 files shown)


itsjavi (approved):
LGTM

PR #9024, opened 2026-08-27
chore(docs): add the screenshot capture harness

jobenjada:
Reopened at @jobenjada's request. @itsjavi, you are on this because the question is whether this belongs in the repo at all — the code is the easy part.

## What & why

**Was:** every docs screenshot was taken by hand, so the set drifted. 320 of 408 images predated the Blocks editor; only 19 postdated Formbricks 5. Sizes ranged from 1027×666 to 2420×1400 — three sizes across three images on one page.

**Now:** a Playwright script (`scripts/docs-capture/capture.ts`) and a seeded workspace (`db:seed:docs`), which together produced **every one of the ~55 images merged into `docs/` this week**.

Neither runs in CI. Both are run by hand against a local instance when a release changes what the product looks like:

```bash
pnpm --filter @formbricks/database db:seed       # admin user
pnpm --filter @formbricks/database db:seed:docs  # the ACME Inc. workspace
pnpm dev
pnpm tsx scripts/docs-capture/capture.ts                    # everything
pnpm tsx scripts/docs-capture/capture.ts surveys/…/recall   # or one shot
```

## Why it should live in the repo

The alternative is that it lives on one laptop, which is where it has been all week.

1. **A screenshot set is only a set if one thing makes it.** One viewport (1920×1200 @2x), one theme, one workspace, `locale: "en-US"`, `timezoneId: "UTC"` so a re-run does not diff on a clock tick. The first page shot by hand breaks that, and it cannot be un-broken without redoing everything.
2. **It encodes what the app does, not what it looks like.** Roughly a third of the file is comments recording things that cost real time to learn — element ids are not unique because the preview pane renders the same survey; a panel a toggle reveals is the row's *sibling*; a CTA pointing at an external URL means `networkidle` never fires. That knowledge is worth more than the script.
3. **Re-shooting is not a one-off.** Two releases invalidated 320 images. The next one will do it again, and the cost of responding is either "run a script" or "spend a week".
4. **It caught a real defect.** The Next.js dev-tools button sits over the viewport's bottom-left corner and had landed on the sidebar in **seven images already committed** across four PRs. Found by diffing a re-run against what was on the branches — which is only possible if a re-run is reproducible.

## What the fixture contains

An "ACME Inc." retail bank — financial services because it matches our ICP, so a prospect reading the docs recognises the survey. A link survey with all 16 element types at **fixed ids** (`acme-<type>`, so a shot never depends on question order), an app survey, 52 responses (38 finished / 14 partial, interleaved), hidden fields, variables, tags with uneven counts, 8 contacts, 7 attribute keys, segments, 6 user actions, 4 org members across all four roles, 2 teams, and two link surveys that exist only to photograph the PIN and email gates.

Everything is fictional. Every address is on `example.com`, which RFC 2606 reserves and which reaches nobody.

## Where to look

- [`scripts/docs-capture/capture.ts`](https://github.com/formbricks/formbricks/blob/chore/docs-capture-harness/scripts/docs-capture/capture.ts) — the `Shot` type and the `shots` array are the whole interface. A shot drives the app to a moment and returns a Locator, a clip box, or nothing.
- **`scripts/` and not `apps/web/`**, deliberately: `apps/web/tsconfig.json` has `"include": ["**/*.ts"]`, so anything under the app is typechecked by the Next build. A dev script that fails to compile takes the build down, and with it E2E and the v3 contract tests. That is not hypothetical — it cost a red CI run earlier this week.
- **`db:seed:docs` is separate from `db:seed`** because this fixture's content *ships*: it is visible in ~200 public images. `db:seed` is what every developer runs locally and shows up in nobody's output. Branding `seed.ts` as ACME would push a docs concern into every developer's database.

## Breaking changes

None. Two new files, one npm script, one `.gitignore` line. Nothing imports any of it, nothing runs in CI, and no existing behaviour changes.

## Migrations & env

- none. `db:seed:docs` is additive and idempotent — every write is an `upsert`, except responses, which are deleted and rewritten so counts stay stable across runs.

## How this was tested

Rerun: `pnpm --filter @formbricks/database db:seed:docs && pnpm tsx scripts/docs-capture/capture.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The harness produces the images actually merged | manual | ~55 images across 20 merged PRs came from this script |
| The seed is idempotent | manual | Run repeatedly against a live database this week; response count stays 52 |
| The fixture typechecks and lints clean | `pnpm --filter @formbricks/database typecheck`, eslint | Clean |
| Nothing under `apps/web` is touched | manual | Both files are outside it, for the tsconfig reason above |
| Session cookies cannot be committed | manual | `scripts/docs-capture/.auth.json` is gitignored — it holds a live login |
| `ANSWER_POOL` is honestly partial | source + run | Six of sixteen elements deliberately have no answers, so the response table has empty cells. Typed `Partial<Record<…>>`; a total `Record` types the miss away and hands `pick` an `undefined` to index — which is exactly what a lint fix nearly did here before the seed was re-run to check |

**Open gaps**

- **No CI check.** Nothing verifies the script still runs, so it can rot silently. A weekly job that runs it and fails on a thrown selector would fix that; not included because it needs a seeded database in CI.
- **Selectors are structural**, not `data-testid`. `.scroll-mt-16` for an editor card, `[data-radix-popper-content-wrapper]` for an open menu. They break when the DOM changes — that is the maintenance cost, and the comments exist to make the breakage diagnosable.
- **Images are verified by eye.** Nothing asserts a shot contains what it should.
- Two shots need a file argument (`DOCS_CAPTURE_LOGO`) and are not runnable without it.

**Risks**

- Nothing executes in CI or in the product. The risk is maintenance: an unmaintained script is worse than none, because the next person trusts it.

---

> [!NOTE]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ stats.html
 docs/api-v3-reference/contract-tests/fixtures.json
 .schemathesis/
 __pycache__/
+
+# Playwright storage state written by the docs capture script — contains live session cookies
+scripts/docs-capture/.auth.json
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -55,6 +55,7 @@
     "db:push": "prisma db push --accept-data-loss --config ./prisma.config.ts",
     "db:seed": "dotenv -e ../../.env -- tsx src/seed.ts",
     "db:seed:clear": "dotenv -e ../../.env -- tsx src/seed.ts --clear",
+    "db:seed:docs": "dotenv -e ../../.env -- tsx src/scripts/seed-docs-fixtures.ts",
     "db:seed:contract": "dotenv -e ../../.env -- tsx src/scripts/seed-contract-fixtures.ts",
     "db:setup": "pnpm db:migrate:dev && pnpm db:create-saml-database:dev",
     "db:start": "pnpm db:setup",
--- a/packages/database/src/scripts/seed-docs-fixtures.ts
+++ b/packages/database/src/scripts/seed-docs-fixtures.ts
@@ -1,0 +1,282 @@
+/**
+ * Seeds the ACME Inc. workspace that every screenshot in `docs/` is captured from.
+ *
+ * Why a separate script rather than an addition to `seed.ts`: `db:seed` is what every developer runs
+ * to get a workable local instance, and its content shows up in nobody's published output. This
+ * fixture's content ships — it is visible in ~200 public documentation images — so it is versioned,
+ * reviewed and changed on its own terms. Editing `seed.ts` to brand it ACME would push a docs concern
+ * into every developer's database.
+ *
+ * The demo company is a financial services firm because that matches our ICP: a prospect reading the
+ * docs should recognise the survey they are looking at. Everything is fictional, and no email address
+ * in it resolves to a real inbox.
+ *
+ * Requires `db:seed` to have run first — the admin user comes from there, so one login
+ * (SEED_CREDENTIALS.ADMIN) reaches both the seed workspace and this one.
+ *
+ * Run: pnpm --filter @formbricks/database db:seed:docs
+ */
+import { createId } from "@paralleldrive/cuid2";
+import { logger } from "@formbricks/logger";
+import { type TSurveyBlocks } from "@formbricks/types/surveys/blocks";
+import { PrismaClient } from "../prisma";
+import { createPrismaPgAdapter } from "../prisma-adapter";
+import { SEED_IDS } from "../seed/constants";
+
+const prisma = new PrismaClient({ adapter: createPrismaPgAdapter().adapter });
+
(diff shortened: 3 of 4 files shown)


jobenjada:
Parking this rather than merging it — the harness stays on a local branch for now while we see whether the capture approach earns its keep across more milestones.

The screenshots it produces still ship normally; those PRs are docs-only (images plus one-line `src` swaps), which is how #9019 landed. Nothing downstream depends on this being merged.

Reopening as ready if and when we want the regeneration story in the repo.

jobenjada:
Superseded by #9059 — this PR could not be reopened because its branch was force-pushed. Same branch, same intent, with the description that was missing here and everything shot since.

PR #9026, opened 2026-08-27
feat(observability): record the outcome of every SSO callback

xernobyl:
Ref ENG-2551

## What & why

**Was:** ENG-2750 failed 100% of Cloud Microsoft sign-ins for ~18 hours and produced **zero Sentry events**. Better Auth logs that failure as a bare message with no `Error` argument, so the ENG-2037/ENG-2471 capture gate never had anything to capture. We found out because Johannes asked in #bugs.

**Now:** Every SSO callback logs its outcome — `ssoCallbackOutcome: "success" | "failure"`, the provider, and a reason on failure — so an alert has a stable field to watch.

Every SSO outage so far has had a novel cause and an identical symptom. ENG-1800 (RFC 9207 `iss`), ENG-2555 (wrong `Account.issuer`), ENG-2750 (placeholder issuer) and the suppressed `state_mismatch` class each needed their own diagnosis, and each ended with the callback redirecting to `?error=` instead of the callbackURL. Alerting per cause means adding a rule after every incident; alerting on the outcome covers the next one for free.

Successes are emitted too, so the alert can key on a **ratio** ("failures exceed N% of callbacks over 15 minutes"), which survives traffic swings and quiet weekends where an absolute count does not.

## Where to look

- [route.ts](https://github.com/formbricks/formbricks/blob/feat/ENG-2551_sso-callback-outcome-signal/apps/web/app/api/auth/%5B...all%5D/route.ts) — the one seam that sees every callback outcome; the handler now awaits the response before returning it
- [better-auth-observability.ts](https://github.com/formbricks/formbricks/blob/feat/ENG-2551_sso-callback-outcome-signal/apps/web/modules/auth/lib/better-auth-observability.ts) — `recordSsoCallbackOutcome`, `recordSsoCallbackThrow`, and the two allow-lists that bound what an outsider can put in a log field

## Breaking changes

None — additive logging. No API, schema, env or config surface, and no change to what any endpoint returns.

## Migrations & env

- none

## How this was tested

Rerun: `cd apps/web && pnpm test modules/auth/lib/better-auth-observability.test.ts` (85/85)

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| A failed callback is recorded with its reason, on both the Better Auth path and the pinned legacy one | unit (mutation) | 3 cases; treating every redirect as success fails 4 |
| A success is recorded, so a ratio is computable | unit (mutation) | Asserted fields exactly; no `warn` emitted |
| A 403 (SSO licence gate) and a 500 count as failures | unit (mutation) | Dropping the status branch fails 2. These do not redirect, which is why this reads the response instead of hooking an error path |
| The provider id cannot be filled with arbitrary values | unit (mutation) | An unregistered id and a 300-char segment both bucket to `unknown`; dropping the allow-list fails 2 |
| **Nor can the reason** — Better Auth echoes the inbound `error` param into its redirect | unit (mutation) + manual | Allow-listed from Better Auth's own call sites; unknown, forged and markup values bucket to `other`. Replacing the allow-list with a charset regex fails 2. Verified live: a caller with a legitimately obtained state supplied `error=attacker_chosen_value_9999`, the app echoed it, and the record read `other` |
| A callback that **throws** is recorded | unit (mutation) + manual | It answers 500 and nobody signs in, so it was the worst case to miss. Making the throw recorder a no-op fails 1 |
| Non-callback endpoints and unparseable URLs stay silent, and it never throws | unit (guard) | Observability must not be able to fail a sign-in |
| **Live, through the real app** | manual | Real app + Postgres + Valkey + a real OpenID Provider. Success logged `success`/`openid`; a bogus state logged `failure`/`state_mismatch`; **stopping Postgres mid-callback** logged `internal_server_error`; `/sign-in/email` emitted nothing |

<details>
<summary>The live records, verbatim</summary>

```
INFO  SSO callback succeeded
    source: "sso-callback"
    ssoCallbackOutcome: "success"
    ssoProvider: "openid"
WARN  SSO callback failed
    source: "sso-callback"
    ssoCallbackOutcome: "failure"
    ssoProvider: "openid"
    ssoCallbackReason: "state_mismatch"
```

</details>

**Open gaps**

- **This is only half of ENG-2551.** It makes the outage observable; it does not alert on it. The rule (rate or ratio on `ssoCallbackOutcome`) is an infra change and cannot be done from the repo — hence `Ref` rather than `Fixes`.
- No Sentry capture, deliberately: ENG-2037 and ENG-2471 spent real effort quietening this exact path, and re-capturing per event would undo that. Logs reach SigNoz, which supports rate-based log alerts, so the field is directly actionable there.
- Not verified against a real IdP outage — failures were induced with a bogus state, a forged error parameter, and a stopped database rather than by breaking a provider.
- The reason allow-list will need a line when Better Auth adds a callback error code; a new code buckets to `other` until then, which is the safe direction but does lose detail. `internal_server_error` was missing from the first draft and only surfaced by smoke-testing a database outage — reading the enum was not enough.

**Risks**

- One extra log line per SSO callback. SSO volume is low, and the level is chosen so failures land at `warn` rather than inflating `error` counts: a single failed callback is routinely the user's own doing (stale tab, expired state, declined consent).
- The handler now awaits the response before returning it, where it previously returned the promise directly. Same value, one extra microtask; the recorder cannot throw.

---

> [!NOTE]

### DIFF
--- a/apps/web/modules/auth/lib/better-auth-observability.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.ts
@@ -527,10 +527,16 @@ export const recordSsoCallbackOutcome = (requestUrl: string, response: Response)
         return { outcome: "failure", reason: "malformed_location" };
       }
 
-      // `null` means no `error` parameter at all — the only shape that is a success. `?error=` and a
-      // bare `?error` both read back as `""`, which a truthiness check would have counted as a clean
-      // sign-in; an ambiguous error parameter belongs on the failure side, like every other redirect
-      // the browser cannot usefully follow.
+      // `null` means no `error` parameter at all, which is the only shape that is a success. `?error=`
+      // and a bare `?error` both read back as `""`, and a truthiness check counted those as a clean
+      // sign-in — an error parameter that is present but empty says something went wrong without
+      // saying what, and the success side is the half the ratio has to be able to trust.
+      //
+      // The inverse costs accuracy in one case worth knowing about: a `callbackURL` that itself
+      // carries an `error` parameter makes a genuine success look like a failure. Nothing in this app
+      // passes one (the SSO buttons send `callbackURL: "/"`), and the alternative — inspecting the
+      // redirect target to decide whether the parameter is "ours" — trades a rare false positive for
+      // a guessing game.
       if (error === null) return { outcome: "success" };
       return { outcome: "failure", reason: SSO_CALLBACK_REASONS.has(error) ? error : "other" };
     }
--- a/apps/web/modules/auth/lib/better-auth-observability.ts
+++ b/apps/web/modules/auth/lib/better-auth-observability.ts
@@ -424,8 +424,63 @@ export const auditPasswordReset = async (userId: string): Promise<void> => {
  */
 const SSO_CALLBACK_PROVIDER_IDS = new Set(["google", "github", "azuread", "openid", "saml"]);
 
-/** Better Auth's callback error codes are lower-snake-case; anything else is not one of ours. */
-const SSO_CALLBACK_ERROR_CODE = /^[a-z0-9_]{1,64}$/;
+/**
+ * Callback failure reasons that may be recorded verbatim (ENG-2551).
+ *
+ * This has to be an allow-list for the same reason the provider id does, and the reason is easy to
+ * miss: Better Auth **echoes the inbound `error` query parameter** into its own redirect
+ * (`api/routes/callback.mjs`, `if (error) redirectOnError(error, error_description)`). Any caller can
+ * obtain a parseable `state` by starting a sign-in and then request
+ * `/api/auth/callback/<id>?state=…&error=<anything>`, so the value that lands here is outside
+ * attacker control only if we bound it to a known set. A charset regex would bound the *shape* of the
+ * value but not the *number of distinct values*, which is the property an alert needs.
+ *
+ * Two consequences of that echo, both closed by bucketing to `other`: an outsider cannot inflate the
+ * cardinality of this field, and cannot forge a specific reason to make a real outage look like
+ * something else.
+ *
+ * Read from `better-auth/dist/oauth2/errors.mjs` (OAUTH_CALLBACK_ERROR_CODES), the two route-level
+ * codes in `api/routes/callback.mjs`, the five `StateError` codes in `state.mjs`, and the codes this
+ * app redirects with itself. A code that disappears upstream simply stops occurring; a new one
+ * buckets to `other` until it is added, which is the safe direction.
+ */
+const SSO_CALLBACK_REASONS = new Set([
+  // Better Auth OAUTH_CALLBACK_ERROR_CODES
+  "account_already_linked_to_different_user",
+  "email_does_not_match",
+  "email_not_found",
+  "email_not_verified",
+  "invalid_code",
+  "issuer_mismatch",
+  "issuer_missing",
+  "no_callback_url",
+  "no_code",
+  "nonce_binding_missing",
+  "oauth_provider_not_found",
+  "unable_to_get_user_info",
+  "unable_to_link_account",
+  // Better Auth callback route, and the codes its redirectOnError call sites pass. `internal_server_error`
+  // is the important one and was found by smoke-testing rather than by reading the enum: stopping the
+  // database mid-callback produces it, so it is the code an infrastructure outage actually arrives as.
+  "invalid_callback_request",
+  "internal_server_error",
+  "invalid_payload",
+  "invalid_profile",
+  "missing_profile",
+  "payload_expired",
+  "user_creation_failed",
+  // Better Auth StateError codes
+  "state_generation_error",
+  "state_invalid",
+  "state_mismatch",
+  "state_not_found",
+  "state_security_mismatch",
+  // Emitted by this app
+  "OAuthAccountNotLinked",
+  "account_not_linked",
+  "invalid_scope",
+  "unable_to_create_user",
+]);
 
 /**
  * Record the outcome of an SSO callback, so a total sign-in outage announces itself (ENG-2551).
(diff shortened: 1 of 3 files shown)


apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -408,3 +408,163 @@ export const auditPasswordReset = async (userId: string): Promise<void> => {
...
+export const recordSsoCallbackOutcome = (requestUrl: string, response: Response): void => {
+  emitSsoCallbackRecord(requestUrl, (): { outcome: "success" | "failure"; reason?: string } => {
+    // A 4xx/5xx on the callback is a failed sign-in too — this is how the SSO licence gate's 403
+    // and any unhandled 500 present, and neither redirects.
+    if (response.status >= 400) return { outcome: "failure", reason: `http_${response.status}` };
+    if (response.status >= 300) {
+      const location = response.headers.get("location");
+      const error = location ? new URL(location, requestUrl).searchParams.get("error") : null;
+      if (!error) return { outcome: "success" };
+      return { outcome: "failure", reason: SSO_CALLBACK_REASONS.has(error) ? error : "other" };
```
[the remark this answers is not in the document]
xernobyl: Valid on both halves, and the missing-`Location` one is worse than minor for this PR specifically: a redirect the browser cannot follow was recorded as a **success**, so it inflated the healthy side of the very ratio the alert is meant to key on. Losing a callback from both sides (the malformed case) is bad; moving it to the wrong side is worse.

Fixed as suggested, with the two reasons bounded like the others: `missing_location`, and `malformed_location` from a `try` around just the `new URL` so a bad header no longer falls into the outer catch.

Worth noting why the suite did not catch this. There was a test on that exact line — "never throws, even on a malformed Location header" — which asserted only that it did not throw, never that it still recorded. It encoded the weaker property, so it stayed green through the bug. Replaced with a `test.each` asserting the outcome and reason for both shapes; reverting `missing_location` to `success` now fails it.

Also re-ran the live check, because this branch is on the path every successful callback takes: a real OIDC sign-in against a live stack still logs `success`, a bogus state still logs `failure`/`state_mismatch`, and the session is still established. 86 unit tests, 622 across auth + SSO, typecheck and lint green.

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -408,3 +408,175 @@ export const auditPasswordReset = async (userId: string): Promise<void> => {
...
+
+      let error: string | null;
+      try {
+        error = new URL(location, requestUrl).searchParams.get("error");
+      } catch {
+        return { outcome: "failure", reason: "malformed_location" };
+      }
+
+      if (!error) return { outcome: "success" };
+      return { outcome: "failure", reason: SSO_CALLBACK_REASONS.has(error) ? error : "other" };
```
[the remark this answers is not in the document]
xernobyl: Confirmed and fixed. Verified the premise against Node rather than taking it on trust — `new URL("https://x/cb?error=").searchParams.get("error")` and the bare `?error` form both return `""`, so `!error` was routing them to the success branch.

It matters for the same reason as the missing-`Location` case in the other thread: this signal exists to feed a failure-ratio alert, so the success side is the half that has to be trustworthy. An ambiguous error parameter belongs on the failure side. Now only a genuinely absent parameter (`null`) counts as success; `""` falls through to the allow-list and buckets to `other`, exactly as you suggested.

Added two cases for the empty and valueless forms plus one pinning that a redirect with no `error` at all is still a success — reverting to `!error` fails two of them. 89 tests on the file, 625 across auth + SSO, typecheck and lint green.

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -408,3 +408,185 @@ export const auditPasswordReset = async (userId: string): Promise<void> => {
...
+ * presentation, not meaning.
+ *
+ * Never throws: observability must not be able to fail a sign-in that otherwise succeeded.
+ */
+export const recordSsoCallbackOutcome = (requestUrl: string, response: Response): void => {
+  emitSsoCallbackRecord(requestUrl, (): { outcome: "success" | "failure"; reason?: string } => {
+    // A 4xx/5xx on the callback is a failed sign-in too — this is how the SSO licence gate's 403
+    // and any unhandled 500 present, and neither redirects.
+    if (response.status >= 400) return { outcome: "failure", reason: `http_${response.status}` };
+    if (response.status >= 300) {
```
BhagyaAmarasinghe: [P1] Skip Better Auth's `form_post` trampoline

Better Auth 1.7 turns a POST callback into a 302 to the equivalent GET callback before it validates state or exchanges the code; this repository already preserves that supported flow in `legacy-sso-callback.ts:90-115`. This branch sees no `error` on the intermediate `Location`, so it records the trampoline as a success, and then the following GET records the real outcome. A completely failing `form_post` flow therefore emits one success plus one failure, capping the observed failure ratio at 50% and potentially defeating the alert this signal exists to feed. Please skip the same-callback trampoline (or otherwise emit only after final callback processing) and cover the POST flow at the route boundary.

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -408,3 +408,185 @@ export const auditPasswordReset = async (userId: string): Promise<void> => {
...
+      // A redirect the browser cannot follow is a failed sign-in, not a quiet success. Both shapes
+      // below would otherwise corrupt the ratio the alert keys on rather than merely lose detail: a
+      // missing `Location` counted as success inflates the healthy side, and a malformed one threw
+      // into the outer catch, dropping the callback from both sides.
+      const location = response.headers.get("location");
+      if (!location) return { outcome: "failure", reason: "missing_location" };
+
+      let error: string | null;
+      try {
+        error = new URL(location, requestUrl).searchParams.get("error");
```
BhagyaAmarasinghe: [P2] Do not interpret the successful callback destination's query as an auth error

The SSO buttons pass `getSsoReturnToUrl(returnToUrl, source)` as `callbackURL`, and the login page derives `returnToUrl` from the validated same-origin `callbackUrl` search parameter; they do not always send `/` as the comment below claims. For example, a sign-in started with a callback target such as `/some-page?error=retry` succeeds and Better Auth redirects to that target, but this lookup reads `retry` and records the successful callback as `failure`/`other`. That corrupts the same ratio for a supported callback URL. Please distinguish Better Auth's error redirect destination from the successful callback destination rather than treating any `error` key on `Location` as the callback outcome, and add this case to the tests.

BhagyaAmarasinghe (changes requested):
The callback outcome signal currently misclassifies two supported success/failure paths, so the ratio is not reliable enough to alert on yet.

apps/web/modules/auth/lib/better-auth-observability.ts
xernobyl: Confirmed and fixed. Verified the hop in the installed 1.7.0 — `api/routes/callback.mjs` builds `${baseURL}/callback/${id}?${params}` and throws the redirect *above* the `try` that parses and validates, so it really does precede state validation and the code exchange, exactly as this repo's own note in `legacy-sso-callback.ts` says.

Your ratio arithmetic is the part that made me treat this as a blocker rather than a rough edge: the hop carries no `error`, so it landed on the **success** side, and a form_post flow failing 100% of the time would still have read as 50% failures — under any threshold the alert might pick. That is worse than losing the signal, because it looks like a signal.

The recorder now says nothing when the redirect target is itself a callback path, and the GET that follows records the real outcome. Covered at the route boundary with both the absolute and the relative twin, plus a test that the following GET still records the failure. Removing the skip fails 2.

The P2 fix below changes this too, for the better: with the session cookie as the success signal, the hop could no longer have been scored as a success even if the skip were missed — it mints no session. Belt and braces, and I kept the explicit skip because "not the request that decides" is the honest description of that hop.

apps/web/modules/auth/lib/better-auth-observability.ts
xernobyl: You are right, and my code comment asserted the opposite without checking — it claimed "the SSO buttons send `callbackURL: '/'`". They send `getSsoReturnToUrl(returnToUrl, source)`, which preserves the return URL's query string, so `/some-page?error=retry` is a supported target. I had even written that limitation down as "not reachable today", which made it worse: a wrong claim with a reassuring label on it.

Rather than compare destinations, success is now decided by **the session cookie**. Better Auth mints it only on a completed sign-in, so it does not care what the destination's query string looks like — and destination comparison would have been fragile here anyway, since the failure destination is the caller's `errorCallbackURL` (`/auth/login` for all five buttons) and the success destination is its `callbackURL`; both are ordinary app pages.

One case that fix would otherwise have broken, which is why it is not simply "no session ⇒ failure": verify-before-link recovery ends with **no session and no error**, redirecting to the inbox-verification page on purpose (`ssoRecoveryAfterHandler`). Counting that as a failure would inflate the ratio during a perfectly healthy flow. It now records `outcome: "incomplete"`, `reason: "no_session"` — named, and outside both sides of the ratio. Any alert must exclude it explicitly; that is documented on the type.

Tests cover a completed sign-in whose `callbackURL` carries `?error=retry` (success), a failure redirect to `errorCallbackURL` (failure), and the recovery shape (incomplete). Reverting to the error-key check fails 3.

Verified live as well, since this is the path every successful sign-in takes: a real OIDC sign-in with `callbackURL=/some-page?error=retry` landed on that page, established a session for the user, and recorded `success`. Before the fix that was `failure`/`other`. A bogus state still records `failure`/`state_mismatch`. 95 tests on the file, 631 across auth + SSO, typecheck and lint green.

BhagyaAmarasinghe (dismissed):

BhagyaAmarasinghe (approved):

PR #9030, opened 2026-08-27
docs(surveys): re-shoot the response option and hidden fields pages (ENG-2671)

jobenjada:
Fixes ENG-2671, ENG-2674, ENG-2673, ENG-2665

## What & why

**Was:** Four pages illustrated with screenshots taken before the Blocks editor shipped.

**Now:** Retaken from the current editor, from the same ACME Inc. workspace as the question-type set in #9019.

Three of them — response limit, spam protection, the back button — each document one row of Settings → Response Options. The fourth, hidden fields, needed more than a re-shoot: three of its four images walked a flow the editor no longer has (enable a toggle, then an empty input, then the filled result). That is now one panel, so those three images become one and the steps collapse to match. Its response-table image is still its own view and keeps its own image.

Captured with each setting **switched on**. A toggle in its default-off state shows nothing the reader came for — the spam-protection page is about the reCAPTCHA threshold, and the threshold does not exist until the toggle is on. The old screenshots had the same property, which is presumably why they were taken that way originally.

## Where to look

Nothing risky. Three new images and three one-line `src` swaps.

## Breaking changes

None. Documentation images and one-line `src` swaps. No code changes.

## Migrations & env

- none

## How this was tested

Rerun: `cd docs && grep -rhoE '/images/[A-Za-z0-9/._-]+\.webp' --include="*.mdx" . | sort -u | while read r; do [ -f ".$r" ] || echo "BROKEN: $r"; done` — expects no output.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Each image shows the setting enabled, with its revealed controls | manual | Spam protection shows the 0.1–0.9 threshold slider and the SDK warning; response limit shows the completed-responses stepper |
| Hidden fields shows configured fields, not an empty state | manual | Panel lists `utm_source`, `plan_tier`, `branch_code`, `referral_code`; the response table shows their values |
| Images show the current editor | manual | Blocks-era Settings tab, not the 2024 dark-sidebar editor |
| Every image reference in `docs/` still resolves | manual | Swept all `/images/...` refs across every `.mdx`; 0 broken |
| The seeded survey is left unmodified | manual | The editor has auto-save disabled and the run never clicks Save; toggles revert on reload |

**Open gaps**

- Verify-email and PIN protection live in the same panel and are also stale, but their pages are 7- and 5-image walkthroughs. Their settings-row shot alone would leave those pages mixing new and 2024 images, so they are not included here.

**Risks**

- none. Nothing here executes.

---

> [!NOTE]


jobenjada:
reviewed, looks good!

PR #9031, opened 2026-08-27
feat(ai): unify the disabled AI smart tools prompt

jobenjada:
Fixes PRO-70

## What & why

**Was:** Every AI gate wrote its own "AI is unavailable" — five copies of the same reason strings across four treatments.

**Now:** One `AIUnavailableAlert` on the existing `Alert`; reason copy, action target and error-code mapping resolved once in `lib/ai/availability`.

Three defects fell out of it:

- Self-hosted "Upgrade plan" linked to a billing page that does not exist there.
- The translations tooltip mapped two of four reasons, so read-only users were told to upgrade.
- AI upgrade CTAs emitted no `upgrade_cta_clicked`, unlike every other gate.

## Where to look

- [lib/ai/availability.ts](https://github.com/formbricks/formbricks/blob/claude/disabled-ai-prompt-unify-55cc8b/apps/web/lib/ai/availability.ts) — reason → copy → CTA
- [workspace-context.tsx](https://github.com/formbricks/formbricks/blob/claude/disabled-ai-prompt-unify-55cc8b/apps/web/app/(app)/workspaces/%5BworkspaceId%5D/context/workspace-context.tsx) — deployment facts, no prop drilling
- [ai-unavailable-alert/index.tsx](https://github.com/formbricks/formbricks/blob/claude/disabled-ai-prompt-unify-55cc8b/apps/web/modules/ai/components/ai-unavailable-alert/index.tsx) — the one alert every surface renders

## Breaking changes

None — internal to `apps/web`. No API, SDK, route or env surface moves.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm test --filter=@formbricks/web`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| One shared message per reason, `read_only` included | unit (mutation) | `availability.test.ts`; mutate `availability.ts:72` |
| Self-hosted upgrade offers the licence form, cloud billing | unit (mutation) | `availability.test.ts`; mutate `availability.ts:42` |
| Server error codes reuse that same copy | unit (mutation) | `ai-error-messages.test.ts`; mutate `availability.ts:88` |
| Create-with-AI dialog drops the dead form | manual | Pair below |
| Both AI tooltips carry the shared reason | manual | Overflow gone; pairs below |

<details open>
<summary><b>Screenshots</b> — reason <code>not_enabled</code>, 1280×820</summary>

Licensed local instance, `isAISmartToolsEnabled` off — each pair differs only in this diff.

**Create with AI dialog**

| Before | After |
| --- | --- |
| Info alert above a greyed-out prompt textarea and a disabled Create button | The same info alert alone, titled AI survey creation, no dead form beneath it |

**Summary → example responses tooltip**

| Before | After |
| --- | --- |
| Tooltip reading AI example responses are disabled for this organization | Tooltip reading AI smart tools are disabled for this organization |

**Manage translations → Translate with AI tooltip**

| Before | After |
| --- | --- |
| Two-sentence tooltip running off the right edge of the modal | Single-sentence tooltip fitting inside the modal |

</details>

<details>
<summary><b>Merge with #9071</b> — what the chart surface looks like now</summary>

#9071 split chart creation into two entry points and deleted `ai-query-section.tsx`, moving the AI gate into `CreateChartWithAIDialog` with a fresh copy of the four reason strings. That dialog now renders `AIUnavailableAlert`, and the chart-local `charts/lib/ai-availability.ts` and its six locale keys are gone.

`CreateChartButton` hides the AI menu item when AI is off, so the alert inside that dialog is a fallback rather than a path a user reaches. It stays for the case where entitlement changes under an open dialog, and says what every other gate says. The earlier before/after pair for this surface was shot against the old combined dialog and no longer describes either side, so it is dropped rather than kept as decoration.

</details>

**Open gaps**

- The chart AI alert has no capture: its entry point is hidden when AI is off.
- Only `not_enabled` was shot; the other three reasons rest on unit tests.
- The self-hosted licence CTA was never rendered — no unlicensed instance to hand.

**Risks**

- `WorkspaceContextWrapper` gained a required prop; both call sites updated, typecheck covers the rest.
- Outside that provider the prompt shows only the docs link — the guard that already hid the CTA.

---

> [!NOTE]


itsjavi (changes requested):
apart from the small review remark,the rest looks good to me

itsjavi (approved):

PR #9042, opened 2026-08-27
docs(surveys): re-shoot the variables editor section (ENG-2664)

jobenjada:
Ref ENG-2664

## What & why

**Was:** Five screenshots, all from 2024-11-20 and the pre-Blocks editor. Three of them walked a flow that no longer exists — an empty Variables card, then an input, then the result.

**Now:** Those three are one image of the panel that does all three, with the steps collapsed to match.

It also fixes a deprecated SDK reference the page carried: hidden fields are set via `formbricks.setup`, not `formbricks.init`. js-core marks `init` `@deprecated Use setup() instead. This method will be removed in a future version` (`packages/js-core/src/index.ts:98`).

The screenshot shows two configured variables with their types, values and IDs — the previous first image was an empty card, which showed a reader nothing.

## Where to look

- [variables.mdx](https://github.com/formbricks/formbricks/blob/docs/variables-page/docs/surveys/general-features/variables.mdx) — the consolidated section and the `setup` correction

## Breaking changes

None. Documentation prose and images. No code changes.

## Migrations & env

- none

## How this was tested

Rerun: `cd docs && grep -rhoE '/images/[A-Za-z0-9/._-]+\.webp' --include="*.mdx" . | sort -u | while read r; do [ -f ".$r" ] || echo "BROKEN: $r"; done` — expects no output.

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| The three steps are one panel now | manual | Editor shows name, type, value, Add variable and Variable IDs together |
| Screenshot shows configured variables | manual | `applicant_score` (number) and `product_line` (text) with values and IDs |
| `formbricks.init` really is deprecated | manual | `js-core/src/index.ts:98` — "@deprecated Use setup() instead" |
| The three superseded images left no orphans | manual | Confirmed unreferenced across `.mdx` **and** `.json`, then deleted |

**Open gaps**

- **This page still carries two 2024 images**, both of the logic editor: "Variables usage" and "Logic with variables". They are deliberately not replaced. Blocks restructured how logic attaches, so re-shooting them without deciding whether that flow still reads the same would be guesswork — that decision belongs with ENG-2663, which covers the conditional-logic page. So this page is improved, not finished.

**Risks**

- none. Nothing here executes.

---

> [!NOTE]


jobenjada:
looks good!

PR #9053, opened 2026-08-28
feat(license): support custom license API endpoint

BhagyaAmarasinghe:
No ticket: this is a deployment follow-up for the Artemis environment.

## What & why

**Was:** Every staging-mode installation sent enterprise-license checks to the shared staging license service.

**Now:** Deployments can route checks to their own license service with an optional URL; existing defaults remain unchanged.

## Where to look

- [Endpoint selection](https://github.com/formbricks/formbricks/blob/11b4d90bb/apps/web/modules/ee/license-check/lib/license.ts#L35-L41)
- [Environment validation](https://github.com/formbricks/formbricks/blob/11b4d90bb/apps/web/lib/env.ts#L228-L233)
- [Override regression test](https://github.com/formbricks/formbricks/blob/11b4d90bb/apps/web/modules/ee/license-check/lib/license.test.ts#L1434-L1486)

## Breaking changes

None. The variable is optional and preserves the existing staging and production defaults.

## Migrations & env

- Optional `ENTERPRISE_LICENSE_API_URL` overrides the enterprise-license check endpoint.

## How this was tested

Rerun: `pnpm test --filter=@formbricks/web -- modules/ee/license-check/lib/license.test.ts lib/turbo-build-env.test.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Configured license endpoint is used | unit (red on main) | Request targets the configured URL |
| Existing staging default remains | unit (guard) | Request targets the staging license service |

**Open gaps**

- Artemis runtime verification waits for the new image and GitOps configuration.

**Risks**

- The configured endpoint must be reachable from the Formbricks pod.

---

> [!NOTE]


BhagyaAmarasinghe:
Closing this approach so Artemis can retain the same application configuration as staging. We will use a separate Artemis license record through the existing staging license service instead of adding an application endpoint override.

PR #9065, opened 2026-08-28
fix: keep MCP OAuth resource initialization recoverable

BhagyaAmarasinghe:
Ref ENG-2772

## What & why

**Was:** A transient database failure during Better Auth's eager MCP resource seed rejected its process-lifetime auth context. Every auth endpoint then returned 500 until the pod restarted.

**Now:** The version-pinned OAuth provider patch keeps eager seeding recoverable, routes the original error through Formbricks logging, and preserves the existing scope-migration guard.

The original duplicate-race hypothesis was not supported by real Prisma reproduction, so the custom SQL seed was removed.

## Where to look

- [Deferred eager seed and error forwarding](https://github.com/formbricks/formbricks/blob/d81e2335d/patches/%40better-auth__oauth-provider%401.7.0.patch#L5-L14)
- [Real-Postgres recovery and logging assertion](https://github.com/formbricks/formbricks/blob/d81e2335d/apps/web/modules/auth/lib/mcp-oauth-resource-init.integration.test.ts#L53-L97)
- [Structured Better Auth warning context](https://github.com/formbricks/formbricks/blob/d81e2335d/apps/web/modules/auth/lib/better-auth-observability.ts#L289-L324)

## Breaking changes

None. This changes an internal failure path without changing an external contract or operator requirement.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm install --frozen-lockfile && pnpm --filter @formbricks/web test -- modules/auth/lib/better-auth-observability.test.ts modules/auth/lib/mcp-oauth-resource-seed.test.ts modules/auth/lib/mcp-oauth-provider-options.test.ts modules/auth/lib/mcp-oauth-dcr.test.ts && cd apps/web && TEST_DB_PROVISIONED=1 pnpm exec dotenv -e ../../.env -- vitest run --config vitest.integration.config.mts modules/auth/lib/mcp-oauth-resource-init.integration.test.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Transient eager-seed failure | e2e | Auth context resolves instead of staying permanently rejected |
| Lazy recovery | e2e | First resource access retries and persists the resource |
| Seed-failure observability | e2e + unit (guard) | Original error reaches the instance logger and structured app logs |
| Same-process multi-instance initialization | e2e | Six instances resolve with exactly one resource |
| Migration scope parity | unit (guard) | Seeded scopes exactly match advertised MCP scopes |

**Open gaps**

- Cross-process contention is not reproduced; the multi-instance case is a regression guard, not race proof.

**Risks**

- The patch is tied to OAuth provider 1.7.0 and must be removed or recreated on upgrade.

---

> [!NOTE]


apps/web/modules/auth/lib/mcp-oauth-resource-seed.test.ts
```diff
@@ -1,7 +1,81 @@
...
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { prisma } from "@formbricks/database";
+import { isDeferredMcpOauthResourceSeedError, mcpOauthResourceSeedPlugin } from "./mcp-oauth-resource-seed";
+import { MCP_OAUTH_SCOPES, getMcpResourceUrl } from "./oauth-urls";
+
+vi.mock("@formbricks/database", () => ({
+  prisma: {
+    $executeRaw: vi.fn(),
+  },
+}));
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: Fixed in 3752e293e: the test now imports the repository shared deep Prisma mock from `@/lib/__mocks__/database`, preserving `$executeRaw` control without an inline storage-boundary mock. The focused seed suite plus all eight previously affected auth/MCP suites pass (9 files, 180 tests).

xernobyl (commented):
Reviewed at `3760ba5b5c03`. The mechanism is sound and I verified the SQL is correct against a real Postgres — but I couldn't reproduce the defect it fixes, and I think the error handling introduces the failure mode we spent this morning chasing on staging. Details below, happy to be wrong on the first one if you have the trace.

## The race already seems to be handled upstream

The premise is that Prisma's error text doesn't match upstream's duplicate check. I looked at that and Prisma's `P2002` message is:

```
Unique constraint failed on the fields: (`identifier`)
```

It contains "Unique", and `@better-auth/prisma-adapter` rethrows it unchanged, so `seedResources`'s `/unique|duplicate|UNIQUE/i` catch matches it.

I then ran six concurrent processes, each constructing a real Better Auth instance with the real `oauthProvider` plugin and the real Prisma adapter, all racing to seed a freshly-deleted row off a spin barrier. All six returned OK, exactly one row inserted.

The underlying race is definitely real — driving `findFirst`-then-`create` directly at the Prisma level, 3 of 6 processes got `P2002`. It just looks like `seedResources` absorbs it.

**Could you share the stack trace of the 500?** I suspect it came from a path I haven't found — `oauthClientResource` backfill is my main candidate, and it has no equivalent catch. If so the fix probably belongs there instead.

## A throwing `init` permanently bricks auth

This one I'd want addressed regardless of the above.

`runPluginInit` awaits `plugin.init(context)` with no try/catch, and `authContext` is a single promise created once at construction that every request awaits:

```js
const authContext = initFn(options);   // once
const ctx = await authContext;          // every request
```

So one throw poisons the instance for its whole lifetime. I checked with a small harness mirroring this plugin — throw once on `P1001`, succeed after:

```
request 1: THREW -> Connection terminated due to connection timeout
request 2: THREW -> ...
request 3: THREW -> ...
request 4: THREW -> ...
plugin init was called 1 time(s)
```

That's exactly the staging outage this morning — every `/api/auth/*` dead, SSO and email login alike, until restart (ENG-2772).

This PR rethrows connectivity errors by design, with a test pinning it, on infrastructure where transient Postgres timeouts are a known escalating problem (`FORMBRICKS-195`, 97 events, prod and staging, every release since 21 Aug).

I'd make the seed plugin never throw — log and return, like the deferred cases already do. A failed seed is recoverable: upstream's lazy `seedResourcesOnce` runs on first resource access and resets `state.promise = null` on failure, so it self-heals. Throwing doesn't buy us anything and permanently bricks the pod.

## The production wiring isn't exercised by any test

```ts
...(env.NODE_ENV === "test" ? [] : [mcpOauthResourceSeedPlugin]),
```

No test takes the branch that ships — plugin ordering, registration and the interaction with `oauthProvider` all go unverified, since the tests call `init?.()` directly. The reason given is partial Prisma mocks in other suites, which feels like a test-setup problem rather than something to solve by branching production wiring on `NODE_ENV`.

## The SQL is never executed by a test

`vi.mock("@formbricks/database")` replaces Prisma entirely, so the assertions are all against mock call args — including the `toContain('ON CONFLICT ("identifier") DO NOTHING')` check, which string-matches a template the test never runs. A wrong column, bad cast or renamed table would pass.

For what it's worth I ran it manually and **the SQL is correct**: fresh insert returns 1, conflict path returns 0 cleanly, `::TEXT[]` round-trips fine, and 12 concurrent inserts give exactly one row. Schema lines up too — no `@@map`, `identifier` is `@unique` so the conflict target resolves, and every omitted column is nullable or defaulted. But that was me, not CI. Could this get an integration test against a real DB?

## Build-time tolerance may not cover the build-time case

The deferred codes (`P1003`, `P2021`, `3D000`, `42P01`) all mean "server reachable, database or table missing". A Docker build with no database server at all gives `P1001`, which is explicitly fatal here — which seems to contradict the doc comment about init during image build. Worth watching Validate Docker Build.

## Smaller things

- The raw SQL restates table and column names Prisma already owns. `prisma.oauthResource.createMany({ data: [...], skipDuplicates: true })` compiles to `ON CONFLICT DO NOTHING` on Postgres, keeps insert-only semantics, and stays type-safe if the schema moves.
- Body says no Linear ticket — **ENG-2772** now covers this exact init path (filed this morning). Worth a `Ref ENG-2772`, partly so we don't end up with two PRs on the same lines; its primary acceptance criterion is "a failed init is not cached", which is the second point above.
- If `WEBAPP_URL` ever changes we'd seed a second resource row and leave the old one behind. Not a problem, just noting it.

## Security

Nothing to flag. The tagged template parameterises all three interpolated values, identifiers and table names are literals, so there's no injection surface. `server-only` keeps it off the client. Insert-only means an operator's narrowed `allowedScopes` never get silently widened, which is the right call and good that it's documented. Logging is debug-level with nothing sensitive.

The only security-relevant item is the second point — a transient DB blip turning into a permanent total auth outage is an availability issue, and this adds another trigger for it.

xernobyl (changes requested):
Marking this as changes requested for the one blocking item from my review above — the rest are suggestions and can be argued.

The seed plugin must not throw. `runPluginInit` has no try/catch and `authContext` is a single memoised promise, so one throw takes every `/api/auth/*` request down for the life of the pod — SSO and email login alike. That's the staging outage from this morning, and this adds a second trigger for it on infra where transient Postgres timeouts are a known, escalating problem.

Log and return instead, like the deferred branch already does. Upstream's `seedResourcesOnce` retries on first resource access and resets on failure, so a missed seed recovers on its own.

The other thing I'd like before this lands is the stack trace of the 500 — six concurrent processes through the real `oauthProvider` and Prisma adapter all seeded cleanly for me, so I'd like to understand which path actually failed before we add a plugin for it.

BhagyaAmarasinghe:
Thanks — your diagnosis was correct. I could not substantiate the claimed Prisma duplicate-race failure, so I removed the custom SQL seed and the `NODE_ENV`-dependent production wiring. The observed staging signature is the eager oauth-provider seed poisoning Better Auth's memoized context after a transient database failure (tracked by ENG-2772 and upstream at https://github.com/better-auth/better-auth/issues/10887).

Head `b499e1601` now applies a narrow, version-pinned patch around only the eager seed. A failure is sanitized and deferred, while the rest of oauth-provider init still completes and returns its database hooks; the existing lazy seed retries on first resource access. New real-Postgres integration coverage injects a `P1001` failure, proves the same auth instance recovers through DCR, and verifies six concurrent initializations against an empty table. Frozen install, focused OAuth tests, lint, formatting, and typecheck pass. Please re-review this revised scope.

xernobyl (commented):
Re-reviewed at `b499e1601423`. Everything I blocked on is resolved — the seed plugin and its throw are gone, the `NODE_ENV` branch is gone, and the mocked-SQL test is replaced by a real-Postgres one. Nice pivot, and thanks for re-running the race yourself.

I smoke tested the patch: applied it to `node_modules` and ran the real `oauthProvider` with an adapter whose first `oauthResource` lookup throws the observed `P1001`.

| | context | `/api/auth/ok` |
| --- | --- | --- |
| patch applied | resolved | **200** |
| patch reverted | **rejected** | **threw** |

Clean red/green — it does what it says, and the integration test has real signal.

Also worth retracting something I put on the Linear ticket: I'd warned that catching around init would drop the provider's `databaseHooks` (the back-channel-logout ones). That doesn't apply here, because you wrapped only `seedResources` *inside* init, so it still reaches its `return { options: { databaseHooks: … } }`. Wrapping the whole init — which is what I'd sketched — would have dropped them. Your version is better on that axis, and it means the retry/backoff machinery on the ticket is mostly unnecessary.

One thing I'd like restored before this lands, then I'm happy.

## The deleted test isn't related to this change

`apps/web/modules/auth/lib/mcp-oauth-resource-seed.test.ts` is deleted wholesale, but on `main` that file only contains the ENG-2343 guard — the one asserting the migration's `MCP_RESOURCE_ALLOWED_SCOPES` matches `MCP_OAUTH_SCOPES`. I grepped and nothing else asserts it anywhere.

Its docstring spells out why it earns its place: `allowedScopes` **intersects** the requested scopes rather than validating them, so a scope we advertise but the row omits is silently intersected away and `/authorize` fails `invalid_scope` — and with `resourceSeedMode: "insertOnly"` the row is never repaired at boot, so the divergence is permanent for every upgraded instance.

I think this is just collateral from the pivot (the plugin tests went into that file earlier in the branch, then the file went rather than reverting). Could we bring the original back?

## Two small ones in the patch

**The warn won't reach our logging.** It uses the module-scoped `logger` from `@better-auth/core/env`, which is `const logger = createLogger()` with no options — not the instance logger we set with `logger: betterAuthLogger` in `auth.ts`. So the one signal that a pod hit this never lands in `@formbricks/logger` or Sentry. `init` gets the auth context, so `ctx.logger` should route properly.

**The error is discarded.** `catch { logger.warn("<static string>") }` gives us no error, code or stack. This was invisible for hours; if it happens again we'd still not know why the seed failed. `catch (error) { ctx.logger.warn({ error }, …) }` would fix both points at once.

Those two also soften my only other worry — the catch being unconditional, so a permission or config error is swallowed the same as a transient one. That's defensible given the lazy path retries, but right now it'd be completely silent.

## Nit

`initializes concurrent instances against an empty resource table` builds six instances in one process, and in-process contention doesn't reproduce this race — Prisma serialises on the pool, and I needed six separate OS processes to get a `P2002`. So that test passes with or without a fix. Fine as a guard, just worth not counting it as proof of concurrency safety in the Coverage table.

The patch approach itself looks right to me, including pinning it and citing better-auth#10887 with the removal condition in `pnpm-workspace.yaml` — that's the part these usually get wrong.

apps/web/modules/auth/lib/mcp-oauth-resource-seed.test.ts
```diff
@@ -1,48 +0,0 @@
...
-  if (!declaration) {
-    throw new Error(
-      "MCP_RESOURCE_ALLOWED_SCOPES was not found in the migration — it was renamed or removed, which " +
-        "means the seeded resource may no longer allow the scopes the app grants."
-    );
-  }
-  return [...declaration[1].matchAll(/"([^"]+)"/g)].map((match) => match[1]);
-};
-
-describe("the migration's seeded allowedScopes matches the advertised scope set (ENG-2343)", () => {
```
xernobyl, line 36: This file only holds the ENG-2343 guard on `main` — nothing to do with the seed plugin, and I grepped: nothing else asserts `MCP_RESOURCE_ALLOWED_SCOPES` anywhere once this goes.

Its docstring explains why it earns its place: `allowedScopes` **intersects** the requested scopes rather than validating them, so a scope we advertise but the row omits is silently intersected away and `/authorize` fails `invalid_scope`. With `resourceSeedMode: "insertOnly"` the row is never repaired at boot, so the divergence is permanent for every upgraded instance.

I'm fairly sure this is collateral from the pivot — the plugin tests went in here earlier in the branch, then the file went rather than being reverted. Could we restore the original?

patches/@better-auth__oauth-provider@1.7.0.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/dist/authorize-BP2_wo6r.mjs b/dist/authorize-BP2_wo6r.mjs
+index 822f08f9af95267d9ef085448741d11370e77fc2..be9e0543e55c8b8d8555fc679df7191a71ae6e34 100644
+--- a/dist/authorize-BP2_wo6r.mjs
++++ b/dist/authorize-BP2_wo6r.mjs
+@@ -4401 +4401,5 @@ const oauthProvider = (options) => {
+-			await seedResources(ctx, opts);
++			try {
++				await seedResources(ctx, opts);
++			} catch {
++				logger.warn("oauth-provider: resource seed failed during init; deferring to first resource access.");
```
xernobyl: Two small things, both on this line.

**The warn won't reach our logging.** This `logger` is the module-scoped one from `@better-auth/core/env`, which is `const logger = createLogger()` with no options — not the instance logger we configure with `logger: betterAuthLogger` in `auth.ts`. So the one signal that a pod hit this never lands in `@formbricks/logger` or Sentry. `init` receives the auth context, so `ctx.logger` should route properly.

**The error is discarded.** `catch {` with a static string leaves us no error, code or stack. This was invisible for hours; if it recurs we'd still not know why the seed failed.

`catch (error) { ctx.logger.warn({ error }, "...") }` covers both — and it also softens my only other worry, that the catch is unconditional, so a permission or config failure is swallowed exactly like a transient one. That's defensible since the lazy path retries, but right now it'd be completely silent.

xernobyl (changes requested):
Re-reviewed at `b499e160` — CI is fully green now, including the Better Auth integration suite, and I re-confirmed the fix works (patched: context resolves, `/api/auth/ok` 200; reverted: context rejected, throws). The integration test is red on main too, since `auth.$context` rejects before the count assertion. Good test.

Re-requesting a review picked up the same commit though, so the three points from my last pass aren't in yet — flagging them inline. Only the first is a real blocker; if you disagree on the other two I'll approve without them.

apps/web/modules/auth/lib/better-auth-observability.ts
```diff
@@ -312,7 +312,11 @@ export const betterAuthLogger: NonNullable<BetterAuthOptions["logger"]> = {
         }
       }
     } else if (level === "warn") {
-      contextLogger.warn(safeMessage);
+      if (cause) {
+        contextLogger.warn({ err: cause }, String(safeMessage));
+      } else {
+        contextLogger.warn(safeMessage);
+      }
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: Fixed in fd369520e. Warning causes are no longer passed to Pino’s `err` serializer; the logger now emits only an allowlisted error type and a validated Prisma error code, while preserving the redacted warning message. The regression test includes a sensitive error message, stack, and enumerable database URL and verifies none reach the application-log call. Focused test: 96 passed; targeted ESLint and web typecheck also pass.

xernobyl (approved):
Approving at `fd369520`. All three findings from my earlier pass are resolved, and the CodeRabbit privacy finding is handled more strictly than I'd asked for.

Verified locally on your branch rather than reading the diff:

- `modules/auth/lib/better-auth-observability.test.ts` — **96 passed**.
- Mutation check: swapping `getSafeWarningErrorContext(cause)` back to `{ err: cause }` fails exactly one test, the new one, and nothing else. So it binds.
- Earlier: applied the patch to `node_modules` and drove the real `oauthProvider` with an adapter whose first `oauthResource` lookup throws — patched, context resolves and `/api/auth/ok` returns 200; reverted, context rejects and it throws.

The allowlist is the right call over redacting the message — it closes the stack and enumerable-property paths too, which redaction alone would have missed (a thrown error's stack usually repeats its message verbatim, and your test's `databaseUrl` property is exactly the case that would have slipped through). Falling back to `"Error"` for an unrecognised `name` and bounding `code` to `/^P\d{4}$/` are both good instincts — neither field can carry attacker-influenced text into a log now.

One honest trade-off worth naming, not a blocker: I'd originally asked for the error to be logged so a recurrence would be diagnosable, and the allowlist means we now get its type but not its message. For the pg-pool error that started all this we'd see `errorType: "Error"` with no code, since `ECONNRESET`-style codes don't match the Prisma pattern. In practice that's fine — the warn tells us the seed failed and when, and `FORMBRICKS-195` carries the connection detail — but "why" now needs two sources instead of one. Right call given the alternative was leaking credentials.

Also not for this PR: the `error` branch still logs only the message and sends the cause to Sentry alone, so self-hosters lose it entirely — leaving `warn` better instrumented than `error`, which is backwards. Worth a follow-up on `main`.

Thanks for the turnaround on this, and for re-running the race yourself rather than taking my word for it.

apps/web/modules/auth/lib/mcp-oauth-resource-seed.test.ts
xernobyl: Restored in d81e2335 — byte-identical to main. Thanks.

patches/@better-auth__oauth-provider@1.7.0.patch
xernobyl: Both addressed: `ctx.logger` in d81e2335, and the cause now surfaces as an allowlisted summary in fd369520.

PR #9067, opened 2026-08-28
feat(surveys): split the browser-context flag so app surveys capture device meta [ENG-2472]

itsjavi:
## What & why

**Was:** a survey rendered by a mobile SDK carried none of the Tier-1 auto-capture — no viewport, no screen, no timezone. **Now:** it carries the half a WebView can honestly report, on the next web deploy, with no SDK release and no customer app rebuild.

`isWebEnvironment` gated the *whole* snapshot, but it conflates two unrelated groups:

| group | reads | honest in a WebView? |
| --- | --- | --- |
| **page** | `location.href`/`.pathname`, `document.referrer`, `utm_*`, `?source=` | **no** — all four SDKs load their HTML with a null base URL, so `location.href` is `about:blank` and `referrer` is `""` |
| **device** | `screen.*`, `window.inner*`, `Intl…timeZone`, `navigator.language` | **yes** — real measurements of the device and of the area the survey was rendered into |

So the flag's real meaning is *"this runtime's `location` describes a host page"*, not *"this is a browser"*. Split the reader in two and gate only the page group. All four SDKs resolve the renderer un-pinned from `{appUrl}/js/surveys.umd.cjs` at display time, so this reaches every already-installed app on the next deploy. `isWebEnvironment` keeps its name, default and meaning as a public `renderSurvey` prop, so shipped SDK versions need no change.

Also adds **`locale`** (`navigator.language`) to Tier-1 — schema key, catalog entry, blocklist name, label + icon — so mobile does not gain a field web lacks. Deliberately distinct from `language`, which is what the respondent *answered in*: a German speaker taking an English-only survey has `locale: "de-DE"` and `language: "en"`, and the gap is the finding. Both client ingest routes whitelist through `pickAutoCapturedResponseMeta`, which is derived from the schema, so neither route changes.

Still absent on mobile, correctly: `url`, `source`, `pagePath`, `pageReferrer`, `utm*`.

**Where to look:** `packages/surveys/src/lib/browser-context.ts` (the split and why the gate sits where it does) · `packages/types/embedded-data-resolver.ts` (the `locale` catalog entry) · `packages/surveys/src/lib/browser-context.test.ts` (the replaced off-web test).

## Linear ticket

https://linear.app/formbricks/issue/ENG-2472/mobile-sdk-parity-for-embedded-data-one-batched-release-per-sdk

## How this was tested

- `@formbricks/surveys` ✅ 28 files / 707 tests · `@formbricks/types` ✅ 8 files / 365 tests · `apps/web` analysis ✅
- `pnpm typecheck` ✅ for `types`, `surveys` and 24 of 25 tasks. `@formbricks/web#typecheck` fails on 8 files, all `Cannot find module '@formbricks/jobs'` — **verified identical on the unmodified base branch** (stashed the diff, same 8 files, 30 errors), so pre-existing and untouched by this PR.
- `pnpm lint` ✅ 0 errors for `types` and `surveys` (4 + 1 pre-existing warnings in files this PR does not touch) · prettier ✅ · `pnpm i18n` ✅ all locales complete.
- **Viewport derivation checked in headless Chromium** against the real SDK template — see the fold. This was the one thing that could have been wrong *and* plausible.

<details><summary>Tests added, and the viewport check</summary>

The old test `browser-context.test.ts:193` ("captures nothing off the web") encoded the decision this PR reverses, so it is replaced by its two successors: off-web keeps the device group and drops *every* page key; on-web keeps both — the second one is what proves the first is the gate working rather than the device reads being unavailable in that runtime too. The reader tests are regrouped into `readPageContextMeta` / `readDeviceContextMeta` describes, each with a "reads no field from the other group" assertion, and `locale` gets a read case and an omit case.

`readBrowserContextMeta` is deleted rather than kept: after the split nothing in production called it, and an export that exists only to be tested is dead code.

The anti-drift test in `embedded-data-resolver.test.ts` already fails if the catalog and `RESERVED_FIELD_NAMES` disagree about `locale`, so that pairing is enforced rather than reviewed.

**Viewport check.** All four SDK templates use `<meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0">` with **no `width=device-width`**. If an engine fell back to a ~980px layout viewport, `viewportWidth` would be wrong and look completely plausible. Loaded that exact document in headless Chromium (Blink — the same engine as Android System WebView) under mobile emulation:

```
Pixel 5 (Blink, mobile)      innerWidth=393   screen=393x851 dpr=2.75   OK
Galaxy S9+ (Blink, mobile)   innerWidth=320   screen=320x658 dpr=4.5    OK
Pixel 7 landscape            innerWidth=863   screen=915x412 dpr=2.625  OK
```

Blink derives the layout viewport from `initial-scale` when `width` is absent, as specified. **iOS WKWebView still needs a device check** — listed under Risks below. Not pre-emptively adding `width=device-width`: if an engine were falling back today the survey would already render wrong, so the likely case is a no-op and the unlikely case is an unverifiable rendering change across four SDKs.
</details>

## Breaking changes

None. Additive: one optional `meta` key, one catalog entry, and a flag whose name, type and default are unchanged. Responses collected before this carry none of the new keys and keep validating — every key is optional and there is nothing to backfill, since the values only ever existed in a browser that has long since closed.

The one behaviour shift is intended and is the point of the PR: app-survey responses that previously carried no client-captured meta now carry `screenWidth/Height`, `viewportWidth/Height`, `timezone` and `locale`. Nothing that read those fields before stops working; a reader that assumed they were web-only now sees them populated on app responses too.

## QA / Test Plan

**How to test**

- [ ] Web link survey → submit → response card shows Page Path, UTM fields (if the link carried them), Screen/Viewport, Timezone and the new **Device Locale**. Nothing regressed from before.
- [ ] Web app survey on a host page → submit → same fields, describing the host page.
- [ ] **App survey on iOS / Android / RN / Flutter** → submit → the response card now shows Screen, Viewport, Timezone and Device Locale, and shows **no** Page Path, Page Referrer, UTM or Source. This is the change.
- [ ] Cross-check the mobile viewport against the device: `viewportWidth` should be the device's CSS-pixel width (e.g. 393 on a Pixel 5, 390 on an iPhone 14), **not** ~980. A value near 980 on one platform means that SDK's `<meta name="viewport">` needs `width=device-width` — a one-line fix in that SDK alone.
- [ ] Change the device's system language, retake an app survey → **Device Locale** reflects the new setting while **Language** still reflects the survey language answered in.
- [ ] Rotate the phone mid-survey, then submit → viewport is the one from when the survey appeared, not the rotated one (freeze-at-display still holds).
- [ ] Try to declare an Embedded Data field named `locale` in a survey → refused as a reserved name. A survey that already declares one keeps working and keeps winning inside that survey.

**Preconditions / test data**

- A workspace with an app survey, and at least one of the four mobile SDKs pointed at the deployed instance. Response card / response table are the readout.
- **After deploying, purge the CDN for `/js/surveys.umd.cjs`.** It is served `max-age=3600, s-maxage=2592000, stale-while-revalidate=3600` (`next.config.mjs`), so a cold edge can serve the old bundle and the mobile check would silently measure the previous renderer. Verify on device *after* the purge.

**Risks & regressions**

- **iOS WKWebView viewport is the one unverified item** (Blink is verified above). Wrong here would be wrong and plausible; check it on a real device as the first QA step.
- `screen.width` is orientation-invariant on iOS WKWebView and rotates on Android — a pre-existing web inconsistency that now shows up on mobile responses too. Worth a line in the docs (ENG-1849), not a fix.
- Anything reading `meta` on app responses now sees more keys than before. Every consumer goes through the catalog, and `pickAutoCapturedResponseMeta` bounds what ingest accepts, so an unexpected key cannot reach the database.

- No DB migration, no env var. One cutover step: the CDN purge above.

---
_Generated by [Claude Code](https://claude.ai/code/session_018L6PrECN38Jbe1FEdFccTz)_

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/lib/utils.test.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/(analysis)/responses/lib/utils.test.ts
@@ -222,7 +222,10 @@ describe("utils", () => {
       // assertion is what would fail if someone reintroduced a local `METADATA_FIELDS`.
       expect(RESERVED_COLUMN_ENTRIES.map((entry) => entry.name)).toContain("utmCampaign");
       expect(RESERVED_COLUMN_ENTRIES.map((entry) => entry.name)).toContain("timezone");
-      expect(RESERVED_COLUMN_ENTRIES).toHaveLength(20);
+      // `locale` is the entry that proved it: it was added to the catalog by ENG-2472 and became a
+      // column here with no edit beyond this count.
+      expect(RESERVED_COLUMN_ENTRIES.map((entry) => entry.name)).toContain("locale");
+      expect(RESERVED_COLUMN_ENTRIES).toHaveLength(21);
     });
 
     test("today's six columns are still visible by default, and the new ones are not", () => {
--- a/apps/web/i18n.lock
+++ b/apps/web/i18n.lock
@@ -3501,6 +3501,7 @@ checksums:
     workspace/surveys/responses/how_to_identify_users: c886035d9d9a0cfc3fa9703972001044
     workspace/surveys/responses/ip_address: 8f2b4d42a165a4c165eca4d7639ce57e
     workspace/surveys/responses/last_name: 2c9a7de7738ca007ba9023c385149c26
+    workspace/surveys/responses/locale: f7085e279c3b4d53e701f3b2dd012767
     workspace/surveys/responses/more_context: bf030156c24007f2a9df863722c6c49e
     workspace/surveys/responses/not_completed: df34eab65a6291f2c5e15a0e349c4eba
     workspace/surveys/responses/os: a4c753bb2c004a58d02faeed6b4da476
--- a/apps/web/locales/de-DE.json
+++ b/apps/web/locales/de-DE.json
@@ -3628,6 +3628,7 @@
         "how_to_identify_users": "Wie Benutzer identifiziert werden",
         "ip_address": "IP-Adresse",
         "last_name": "Nachname",
+        "locale": "Gerätesprache",
         "more_context": "Mehr Kontext",
         "not_completed": "Nicht abgeschlossen ⏳",
         "os": "Betriebssystem",
--- a/apps/web/locales/en-US.json
+++ b/apps/web/locales/en-US.json
@@ -3628,6 +3628,7 @@
         "how_to_identify_users": "How to identify users",
         "ip_address": "IP Address",
         "last_name": "Last Name",
+        "locale": "Device Locale",
         "more_context": "More context",
         "not_completed": "Not Completed ⏳",
         "os": "OS",
--- a/apps/web/locales/es-ES.json
+++ b/apps/web/locales/es-ES.json
@@ -3628,6 +3628,7 @@
         "how_to_identify_users": "Cómo identificar a los usuarios",
         "ip_address": "Dirección IP",
         "last_name": "Apellido",
+        "locale": "Configuración regional del dispositivo",
         "more_context": "Más contexto",
         "not_completed": "No completado ⏳",
         "os": "Sistema operativo",
--- a/apps/web/locales/fr-FR.json
+++ b/apps/web/locales/fr-FR.json
@@ -3628,6 +3628,7 @@
         "how_to_identify_users": "Comment identifier les utilisateurs",
         "ip_address": "Adresse IP",
         "last_name": "Nom de famille",
+        "locale": "Paramètres régionaux de l'appareil",
         "more_context": "Plus de contexte",
         "not_completed": "Non terminé ⏳",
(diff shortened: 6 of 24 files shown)


pandeymangg (approved):
Verified the load-bearing claims against the four SDK repos before approving: all four really pass `isWebEnvironment: false`, the ingest whitelist grows through `ZAutoCapturedResponseMeta` so neither client route needed edits, the display-time freeze is untouched, and every reserved-catalog consumer picks up `locale` catalog-derived (the ENG-2538 lockstep test covers the one hand-written site). CI green incl. e2e.

Two non-blocking notes for the record: (1) `locale` is the most collision-prone catalog name yet, so it widens the parked ENG-1839 management-API back-compat break for anyone declaring a `locale` field — worth a line on that ticket; (2) the WKWebView viewport derivation is flagged in the PR itself as unverified — treating it as the first QA gate for the iOS SDK release.

PR #9071, opened 2026-08-28
feat(charts): split chart creation into AI and manual entry points (ENG-2782)

itsjavi:
Ref ENG-2782

## What & why

**Was:** one dialog did everything at once — an AI panel, an "OR" divider, a five-tile chart-type picker and the whole builder — and it split its two columns by neither concern: type on the left; name, preview *and* display settings on the right. The controls nobody could find were the ones stacked under a 300px preview.

**Now:** **New chart** is a split button — *Create with AI* or *From scratch* — and each opens a dialog that does one job. The builder is a rail and a stage: one column is everything you set, the other is what that produces. Chart type moves onto the preview beside the Chart/Data tabs, display settings into a strip under the chart. Only the rail scrolls, so the chart and its controls stay on screen however long the configuration gets.

Related to ENG-2782. PR 2 of 2, stacked on [#9027](https://github.com/formbricks/formbricks/pull/9027), which adds the kit — review that one first.

- Chart type no longer gates the builder: it defaults, so the first thing you touch is the measure.
- Save is always present, disabled with a reason, beside a Cancel — the footer used to grow a button halfway through configuring.
- Nothing is lost silently: closing, reloading, or leaving for the AI dialog all ask first.

## Where to look

- [create-chart-view.tsx](https://github.com/formbricks/formbricks/blob/javier/eng-1248-ai-kit-charts/apps/web/modules/ee/analysis/charts/components/create-chart-view.tsx) — the rail/stage split, fixed body height, dirty-state snapshot
- [chart-preview.tsx](https://github.com/formbricks/formbricks/blob/javier/eng-1248-ai-kit-charts/apps/web/modules/ee/analysis/charts/components/chart-preview.tsx) — the stage, with the type switch and settings strip as slots
- [options-switch/index.tsx](https://github.com/formbricks/formbricks/blob/javier/eng-1248-ai-kit-charts/apps/web/modules/ui/components/options-switch/index.tsx) — shared switch: re-measures on resize, animates only on a real change

<details>
<summary>Why the AI dialog has no chart-type picker</summary>

The generator already chooses the type from the prompt — its response carries `chartType`, so "how many responses last month" comes back as a big number and a trend comes back as a line. Making the user pick first would either override that judgement, often producing a worse chart, or be ignored. The builder opens immediately after with the type switch *and* a live preview, which is a better place to change it.

The status line runs in **uncontrolled** mode here: chart generation is a server action that reports no progress, so it hands over a list of phases and lets the component walk them. That is the case showing a feature gets a credible waiting state without building one.

</details>

<details>
<summary>Two shared-component fixes this pulled in</summary>

**`OptionsSwitch` highlight.** It was positioned from pixels measured once at mount and on `window.resize` — which never fires for the ways a switch actually moves: a dialog still running its open animation, a sibling appearing beside it, a column gaining a scrollbar. It now re-measures through a `ResizeObserver`, and position and animation live in one state so a re-measure that finds the same geometry writes nothing. The pill is placed outright on mount and animates only when the selection changes; every dialog holding one used to play a slide across the options as it opened.

**Alerts that truncate.** `Alert size="small"` truncates to one line by design, so the group-by hint lost the half that mattered in a narrow column. The hint is now plain helper text, and the option-grouping warning is a full-size alert that wraps.

</details>

<details>
<summary>What "unsaved" means now</summary>

It previously meant "has chart data", which is true the moment a saved chart loads — so opening a chart and closing it untouched claimed pending changes. It now snapshots name, type, query and display config once the dialog settles and compares against that. A new chart starts from an empty snapshot, so generating or configuring anything counts.

</details>

## Breaking changes

None. One existing surface is rebuilt, two dead components are deleted (`manual-chart-builder`, `configure-chart-dialog` — the latter had no importers), and the translation changes are additive plus five shortened chart-type labels.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm typecheck && pnpm lint && pnpm exec vitest run modules/ee/analysis/charts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Builder layout: nothing important below the fold | manual | Pair below — name and settings were cut off, now everything is on screen |
| The whole flow: measure → group → type → settings → save | manual | Clip below |
| Switch highlight lands on the active option, and does not animate on open | manual | Measured in-page: highlight `left/width` = active option's `96/84`; `transition` class absent on open, present after a click |
| Untouched saved chart closes silently; a changed one asks | manual | Drove both directions on a seeded chart |
| Display settings only render for a type that has one | unit (mutation) | `chart-display.test.ts` — make `hasChartDisplaySettings` return `true` to turn red |
| Example prompts always fit the field they fill | unit (guard) | `ai-chart-prompts.test.ts` |

<details>
<summary>Screenshots</summary>

| Before | After |
| --- | --- |
| Create chart dialog with an AI panel, an OR divider and five chart-type tiles filling the left column | Builder with a name and measure rail on the left and an empty preview stage on the right |
| Configured chart where the name field is cut off at the bottom edge and display settings sit below the preview | Configured chart with the type switch on the preview, bar-direction strip under it and everything within the dialog |

Entry point, AI dialog, data tab and the discard guard:

| | |
| --- | --- |
| New chart split button open on Create with AI and From scratch | AI chart dialog with a prompt field and four example prompts |
| Data tab showing the rows behind the chart | Discard confirmation over the builder after closing with unsaved work |

</details>

<details>
<summary>Clip (after)</summary>

[builder-walkthrough.webm](https://github.com/formbricks/formbricks/blob/assets-pr-screenshots/ENG-2782/builder-walkthrough.webm) — split button → from scratch → measure → group by source → bar → horizontal → data tab → name → save. WebM as Playwright records it; GitHub does not render it inline, so it is a link.

</details>

**Open gaps**

- AI chart generation itself was not driven end to end locally — the dialog, its example prompts and its guards were, but not a real model call. Needs a pass on staging.
- Charts still have no Playwright spec; this whole area is manual (ENG-2314 tracks the gap).

**Risks**

- The status-line phases are timed rather than real, so wording can lead or lag the actual query.
- The generate action cannot be aborted, so Stop abandons the result rather than the request; a run-id guard keeps a late response from opening the builder behind the user.
- `ResizeObserver` in `OptionsSwitch` now runs for every consumer of that component (five surfaces), not just charts.

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `max`.


jobenjada (changes requested):
pls update the visual evidence

itsjavi:
@jobenjada sorry that video was part of this PR https://github.com/formbricks/formbricks/pull/9074

the visual assets were already in the description in this PR

Dhruwang:
Looks good and works well
However it seems like a button is not visible here


Screenshot 2026-09-02 at 12 41 47 PM

Dhruwang (changes requested):
.

itsjavi:
@Dhruwang thanks — I cannot reproduce it on the current head (`71f8d73`), so I would rather ask than guess.

What I measured, in a browser against this branch's compiled CSS:

```
--color-ai-strong  →  lab(32.4486% 49.2217 -84.6695)   (= indigo-700, #432dd7)
.bg-ai-strong      →  background-color: var(--color-ai-strong)   — present in the served bundle
button (enabled)   →  bg lab(32.4 49.2 -84.7), color rgb(255,255,255)
button (disabled)  →  same bg, opacity 0.5              — still clearly visible
```

The button is also rendered unconditionally in every version of that dialog on this branch — only `disabled` changes with the prompt — so the space in your screenshot is an invisible button rather than a missing one. `text-white` and `bg-ai-strong` come from the same variant string, so the background failing while the white label applies points at the stylesheet, not the markup.

Two things would settle it. In devtools on that dialog:

```js
getComputedStyle(document.documentElement).getPropertyValue('--color-ai-strong')
getComputedStyle(document.querySelector('[role=dialog] button[type=submit]')).backgroundColor
```

Empty / `rgba(0, 0, 0, 0)` means the CSS build is missing the AI tokens — this repo has a known trap where a Tailwind v4 `@theme` change needs `rm -rf apps/web/.next` and a dev-server restart, a plain restart is not enough (I lost time to exactly this one myself). Anything else and it is a real bug, and I will chase it with that value.

Also useful: which commit you were on, and whether it was local dev or a deployed build.

apps/web/modules/ee/analysis/charts/hooks/use-chart-dirty-state.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  // State rather than a ref: this is read during render to decide whether closing needs to ask.
+  const [baseline, setBaseline] = useState<typeof currentSnapshot | null>(null);
+
+  useEffect(() => {
+    if (!open) {
+      setBaseline(null);
+      return;
+    }
+    if (isReady && baseline === null) {
```
Dhruwang, line 54: **The baseline snapshots the *previous* session's chart, so a freshly opened builder is dirty from the first frame.**

`handleSaveChart` closes the dialog with `onOpenChange(false)`, not `handleClose()` (`use-chart-dialog.ts:282`), so `chartName` / `chartData` / `currentChartId` survive the close — only the next open's reset effect clears them.

On that next open (`New chart → From scratch`), `isReady` is already `true` on the render where `open` flips (`create-chart-view.tsx:99`: for a new chart it is just `open && !isLoadingChart`), and **that render still carries the saved chart's name and query**. React runs every passive effect for a commit before processing the state those effects schedule, and `useChartDialog`'s reset effect is registered before this one — so `setBaseline` stores `{name: "Foo", query: …}`, the next render is empty, and they differ.

Net effect: an untouched, empty builder reports unsaved work, and closing it prompts *"Discard chart?"* — the behaviour this hook was written to remove. `onRequestAIDialog` leaves state stale the same way, so the AI hand-off round trip hits it too.

Reproduced with a component mirroring `CreateChartView`'s effect order (reset effect registered before `useChartDirtyState`, then `open` flipped true) — `hasUnsavedChart === true` on the first settled render:

```tsx
function View({ open }: { open: boolean }) {
  const [chartName, setChartName] = useState("Foo");        // left over from the previous save
  const [chartData, setChartData] = useState(staleChart);
  useEffect(() => {                                          // useChartDialog's reset effect
    if (!open) return;
    setChartName("");
    setChartData(null);
  }, [open]);
  const { hasUnsavedChart } = useChartDirtyState({ open, isReady: open, isSaving: false, chartName, chartData, chartConfig: {} });
  // → true
}
```

A few ways out: have `handleSaveChart` go through the same reset as `handleClose`, key the baseline on a reset counter rather than on `baseline === null`, or capture it from a layout effect that runs after the reset has flushed.

apps/web/modules/ee/analysis/charts/components/create-chart-with-ai-dialog.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+  const handleStop = () => {
+    // The server action cannot be aborted, so this abandons the result rather than the request.
+    runIdRef.current += 1;
+    setIsGenerating(false);
+  };
+
+  const handleGenerate = async (event: SyntheticEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (!canGenerate || !feedbackDirectoryId) return;
```
Dhruwang, line 144: **Generate is a silent no-op when the workspace has no feedback directory, and the button stays enabled.**

This returns on `!feedbackDirectoryId` with no toast, no state change and no spinner — while `canGenerate` checks only the prompt, `isAIAvailable` and `isGenerating`, so the button (and ⌘↵) stay live. Pressing it does nothing, repeatably.

Reachable today: `dashboard-detail-page.tsx` guards only on the dashboards licence, never on directories, so `directories` can be `[]`. `add-existing-charts-dialog.tsx:168` renders `CreateChartButton` with those directories and `buttonProps={{ disabled: isAdding }}` — enabled — and `create-chart-button.tsx` passes `feedbackDirectoryId={directories[0]?.id ?? null}` → `null`. So: open a dashboard in a workspace with no feedback directory → **Add charts** → **Create new chart** → **Create with AI** → type a prompt → **Generate** → nothing happens.

Before this PR that entry point landed in the builder, which rendered the `no_data_source_available` alert. The From-scratch branch still does (`NoFeedbackDirectoryAlert`); only the AI branch now explains nothing. Either fold the directory into `canGenerate` and show `NoFeedbackDirectoryAlert` in this dialog, or disable the split button's AI item when `directories` is empty.

apps/web/modules/ee/analysis/charts/components/ai-query-section.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import { ActivityIcon, WandSparklesIcon } from "lucide-react";
+import { ActivityIcon } from "lucide-react";
```
Dhruwang: **This component is now orphaned — it was migrated to the AI kit rather than deleted.**

This PR removed its only importer (`create-chart-view.tsx`, replaced by `CreateChartWithAIDialog`). A grep for `AIQuerySection` / `ai-query-section` across `apps/web` now returns only this file's own definition.

The PR deletes `manual-chart-builder.tsx` and `configure-chart-dialog.tsx` as dead, but keeps this one and edits it — `WandSparklesIcon` → `AiIcon`, `variant="ai-primary"`, a new `AiStatusLine` block with three `t()` calls. That is ~150 lines of unreachable UI a reviewer has to read and reason about, and it makes `workspace.analysis.charts.create_chart_with_ai` plus the `ai_not_in_plan` / `ai_not_enabled` / `ai_instance_not_configured` branches look live from here, when that copy is now owned by `create-chart-with-ai-dialog.tsx`.

Worth deleting alongside the other two.

Dhruwang (commented):
Three findings from a correctness pass over the delta against `javier/eng-1248-ai-ui-kit`. Inline below.

Checked and clean: locale sync (28 added / 9 removed keys, all 14 non-English files consistent, no stale keys), the `flex-1` / `basis-auto` override on `DialogBody` (verified against the real Tailwind 4.3.1 build order — `basis-auto` is emitted last and does win, so the fixed height holds), `OptionsSwitch`'s ResizeObserver and animation-suppression logic, the AI dialog's `runIdRef` abandon guard, and `pnpm exec vitest run modules/ee/analysis/charts` (15 files, 230 tests, green).

apps/web/modules/ee/analysis/charts/hooks/use-chart-dirty-state.ts
itsjavi: Confirmed, and thank you for the repro — the effect-ordering detail is exactly right, and `onOpenChange(false)` at `use-chart-dialog.ts:282` (plus the add-to-dashboard path at :375) was the source.

Fixed in 78ff2ee7c by closing the leak rather than working around the ordering: every close now goes through one `closeDialog()` that resets first — save, add-to-dashboard, `handleClose`, and the hand-off to the AI dialog, which had the same hole (the parent flipped `isBuilderOpen` false without ever reaching `handleClose`). With nothing surviving a close there is no stale frame for the baseline to capture, so the edit path is covered too — it had the same exposure through `isReady` seeing a previous session's `chartData`.

Two tests in `use-chart-dialog.test.ts` pin it: after `handleSaveChart` the name, data, type and config are all cleared, same after `handleClose`. Reverting the save path to `onOpenChange(false)` turns the first one red (verified).

apps/web/modules/ee/analysis/charts/components/create-chart-with-ai-dialog.tsx
itsjavi: Valid, and the path you traced is right — `directories` reaches `CreateChartButton` empty from the dashboard's Add-charts dialog, so `feedbackDirectoryId` is `null` while the gate never looked at it.

Fixed in 78ff2ee7c: the directory is now part of `canGenerateChart` rather than only of the request, so Generate and ⌘↵ are both unavailable — and the dialog renders the same `NoFeedbackDirectoryAlert` the from-scratch branch already showed, so it explains itself instead of going quiet. `ai-chart-prompts.test.ts` covers the new case.

apps/web/modules/ee/analysis/charts/components/ai-query-section.tsx
itsjavi: You are right, and it is worse than dead weight: editing it made four translation keys look live. Deleted in 78ff2ee7c along with `ai_query_section_title`, `ai_query_section_description`, `ai_query_placeholder` and `create_chart_with_ai` — the keys with no remaining consumer. The `ai_not_in_plan` / `ai_not_enabled` / `ai_instance_not_configured` family stays: `create-chart-with-ai-dialog.tsx` owns that copy now.

apps/web/modules/ee/analysis/charts/components/chart-dialog-footer.tsx
```diff
@@ -33,18 +46,25 @@ export function ChartDialogFooter({
...
+        <Button variant="secondary" type="button" onClick={onCancelClick} disabled={isSaving}>
+          {t("common.cancel")}
+        </Button>
+      )}
+      <TooltipRenderer shouldRender={!canSave && Boolean(saveHint)} tooltipContent={saveHint}>
+        <Button
+          type={formId ? "submit" : "button"}
+          form={formId}
+          onClick={formId ? undefined : onSaveClick}
+          disabled={isSaving || isDisabled || !canSave}>
```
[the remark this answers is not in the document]
itsjavi: Agreed, and I removed the tooltip rather than making it focusable. A disabled button cannot take focus, so that hint only ever existed for a mouse — and the persistent-text option you mention is already on screen: whenever Save is unavailable the preview is showing either "Pick what to measure and your chart appears here" or the query error, both permanently visible and in the place someone is already looking. A second copy of that on hover was the weaker half. Fixed in 82895cf, along with the now-unused key.

apps/web/modules/ee/analysis/charts/components/create-chart-with-ai-dialog.tsx
```diff
@@ -0,0 +1,320 @@
...
+        title={t("workspace.analysis.charts.ai_create.discard_generation_title")}
+        body={t("workspace.analysis.charts.ai_create.discard_generation_body")}
+        buttonText={t("workspace.surveys.ai_create.discard")}
+        buttonVariant="destructive"
+        cancelButtonText={t("workspace.surveys.ai_create.keep_editing")}
+        onConfirm={() => {
+          setIsConfirmingDiscard(false);
+          closeAndReset();
+        }}
+      />
```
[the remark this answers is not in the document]
itsjavi: Valid — the builder already rendered its confirmation as a sibling and this dialog did not. Fixed in 82895cf: `ConfirmationModal` now sits after `</Dialog>` inside a fragment, so the second Radix modal no longer mounts inside one that still owns focus containment.

apps/web/modules/ee/analysis/charts/components/filter-date-input.tsx
```diff
@@ -25,7 +25,7 @@ export function FilterDateInput({ value, onChange }: Readonly<FilterDateInputPro
     <DatePicker
       value={selectedDate}
       locale={i18n.resolvedLanguage ?? i18n.language ?? "en-US"}
-      triggerClassName="w-[200px]"
+      triggerClassName="min-w-0 flex-1 basis-36"
```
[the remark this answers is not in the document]
itsjavi: Correct, and this one was mine from a conflict resolution: `DatePicker` renders its own `flex items-center gap-2` wrapper, so `triggerClassName` reached the button while the flex item in the filter row stayed content-sized — the overflow I thought I had fixed. Fixed in 82895cf: `DatePicker` takes a wrapper `className` (optional, so its other 15 consumers are untouched), the sizing goes there, and the trigger just gets `w-full` instead of its default 280px.

apps/web/modules/ee/analysis/charts/components/create-chart-with-ai-dialog.tsx
```diff
@@ -0,0 +1,324 @@
...
+                  <div className="grid grid-cols-1 gap-2 sm:grid-cols-2">
+                    {helperPrompts.map((helperPrompt) => (
+                      <Button
+                        key={helperPrompt.label}
+                        type="button"
+                        variant="secondary"
+                        size="sm"
+                        className="group w-full min-w-0 justify-start text-left"
+                        title={helperPrompt.prompt}
+                        aria-label={`${helperPrompt.label}. ${helperPrompt.prompt}`}
```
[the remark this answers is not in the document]
itsjavi: Fixed in 6e7cb48. The label and the prompt now go through `ai_create.prompt_helper_aria` (`"{label}. {prompt}"`) with both interpolated, so the separator is the locale's to choose rather than English punctuation baked into JSX.

apps/web/modules/ee/analysis/charts/components/create-chart-with-ai-dialog.tsx
```diff
@@ -0,0 +1,324 @@
...
+              token, so the mark asserts it here. */}
+            <AiIcon aria-hidden="true" className="text-ai-dark!" />
+            <DialogTitle>{t("workspace.analysis.charts.ai_create.dialog_title")}</DialogTitle>
+            <DialogDescription>
+              {t("workspace.analysis.charts.ai_create.dialog_description")}
+            </DialogDescription>
+          </DialogHeader>
+
+          <DialogBody className="-mx-1 -mt-1 space-y-4 px-1 pt-1 pb-1">
+            <form className="flex w-full flex-col space-y-4" onSubmit={handleGenerate}>
```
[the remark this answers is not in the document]
itsjavi: Agreed on the direction, declining it here, and filed as [ENG-2837](https://linear.app/formbricks/issue/ENG-2837/move-ai-chart-generation-off-its-server-action-onto-apiv3) so it does not just evaporate.

`generateAIChartAction` is not new in this PR — it arrived with Formbricks 5 (`939fedfc`) and this PR only moved the call site out of `ai-query-section.tsx`, which it deletes, and into the new dialog. There is no `/api/v3` chart-generation route to switch to, so "use the required flow" means building one: a documented operation with request and response schemas, RFC 9457 error mapping, its own rate-limit bucket, audit logging, and the Schemathesis contract coverage that every documented op gets. That is a change of its own, not an addition to a UI PR, and doing it here would put an untested EE endpoint under a diff nobody opened for that.

One narrower note on the query-cache half: the generated chart is not cached server state. It is handed straight to the builder for review and discarded if the user backs out, so there is no cache entry for it to go stale against — the same reasoning as the streaming survey generation in #9027. The ticket is scoped to the request path for that reason.

Dhruwang (dismissed):

jobenjada (dismissed):
UI & UX looks good

jobenjada (approved):

PR #9083, opened 2026-08-31
feat: allow hiding the default checkmark icon on ending cards (ENG-1787)

Dhruwang:
Fixes ENG-1787

## What & why

Screen-out / disqualification endings render the same celebratory checkmark as a successful completion. The icon is hardcoded in the survey renderer, so the only way to get rid of it is to upload a blank image — and that still reserves `ElementMedia`'s `min-h-40`, leaving an empty gap above the headline.

Replacing the icon already works (the ending card's headline field carries the image/video uploader). This PR adds the missing half — hiding it:

- `hideDefaultIcon?: boolean` on `ZSurveyEndScreenCard`. Absent means "show", so every existing survey renders exactly as before and there is nothing to migrate.
- `packages/surveys` ending card: `media ?? checkmark` becomes `media ?? (hideDefaultIcon ? null : checkmark)`. `null` emits no markup at all, which is what the blank-image workaround could not do.
- Editor: a **Show default icon** switch on the ending card. Checked when the flag is absent; turning it on writes `undefined` rather than `false`, so the key disappears from the survey JSON instead of accumulating. The switch is hidden while the card has an image or video, since the checkmark never renders in that case and the control would be inert.
- API v3: `hideDefaultIcon` added to `END_SCREEN_KEYS` and to `SurveyEndScreenEnding.yml`. The route rejects unknown ending keys and the documented schema is `additionalProperties: false`, so both were needed for the field to survive a round trip.

## Breaking changes

None — purely additive. A new optional field on an ending; omitting it keeps the current rendering, and no existing request or response shape changes.

https://github.com/user-attachments/assets/c6352092-bd2c-482c-9173-2a852e623e40

## Migrations & env

- none

## How this was tested

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| API v3 accepts `hideDefaultIcon` on an `endScreen` ending and preserves it | unit (red on main) | `apps/web/app/api/v3/surveys/schemas.test.ts` — "accepts hideDefaultIcon on an end screen ending". Fails without the `END_SCREEN_KEYS` entry (unknown-key rejection); passes with it. `cd apps/web && npx vitest run app/api/v3/surveys/schemas.test.ts` → 46/46 |
| Flag survives the zod round trip on the shared survey type | unit (guard) | Covered by the same spec — `ZV3CreateSurveyBody` parses through `ZSurveyEndScreenCard`, so a missing schema field would strip the flag and fail the assertion |
| Documented schema matches the code | unit (guard) | `pnpm api:v3:lint` clean; bundle regenerated with `pnpm api:v3:bundle`. Schemathesis drives the committed bundle on every PR, and `additionalProperties: false` would have rejected the new key |
| Renderer hides the icon when the flag is set | manual | **Not verified — see Open gaps** |
| Editor switch writes and clears the flag | manual | **Not verified — see Open gaps** |

**Open gaps**

- **No browser QA on the two UI behaviours above.** The local dev server could not be kept running long enough for a pass, so there is no screenshot of the ending card with the icon hidden, and no confirmation that the empty-gap problem is actually gone. Both need a manual check before merge: toggle the switch off on an ending card with no media, confirm the checkmark and the brand underline bar disappear with no reserved vertical space, then confirm the switch disappears once an image is uploaded.
- The prebuilt `@formbricks/surveys` bundle was rebuilt locally and does contain `hideDefaultIcon`, but that only proves the flag reached the bundle, not that the rendering is right.
- No e2e spec added. Per AGENTS.md this is UI detail inside a feature area that already has specs (`survey.spec.ts`, `survey-accessibility.spec.ts`), so the expected answer is manual verification — which is exactly the gap above.

**Risks**

- The ending card renderer is shared by every ending, in both the live survey and the editor preview. If the ternary is wrong, the regression is the checkmark vanishing for surveys that never set the flag — the unit coverage does not reach the renderer, so the manual check above is what protects it.
- The editor switch is hidden when media is present. A card that has the flag set *and* then gets an image keeps the flag stored but unreachable in the UI; removing the image brings the switch back in its previous state. Intentional, but worth a look in review.

---

> [!NOTE]


jobenjada (changes requested):
Have you explored other solutions? It feels like we're fixing a bad initial design decision (hard-coding the icon in) with patch rather than fixing the root cause.

Is it an option to load the check mark icon as a image by default? And then people can simply remove it? However I can imagine that this makes things complicated, because today we dont support SVG, nor would people be able to bring the checkmark back once they removed it, etc.

Any other way to solve this tho? If not, we can go ahead with adding this add. field.

Secondly, please update the language:

1. First toggle should read "Show checkmark icon" with description being shorter (cut the second sentence)
2. Update the casing of the "Show Button" toggle to "Show button"

Thank you!

jobenjada:
Also, do we need to do anything on the mobile SDK sides to support this?

Dhruwang:
Explored the image route — it's the more invasive one:

- **It needs a data migration; the flag doesn't.** `endings` is a `Json[]` column, so the flag is a read-path default: old surveys have no key and render the checkmark exactly as today. With a default image, every pre-existing ending has no `imageUrl`, so we either keep the hardcoded fallback anyway or backfill every survey.
- `ZStorageUrl` only accepts `/storage/...` or `http(s)`, so a bundled asset fails validation.
- The icon is `text-brand`, so it follows each survey's brand colour — an image can't.
- `ElementMedia` adds the lightbox and `min-h-40` that make today's blank-image workaround ugly.

The real root cause is that endings have no *kind* (completed vs screened out) for the icon to follow from. Bigger than this PR — I'll file it separately if you agree.

Copy changes in 8c6538b.

Screenshot 2026-09-03 at 12 57 25 PM

Dhruwang:
No — all four SDKs render surveys by loading `{appUrl}/js/surveys.umd.cjs` in a WebView, so they
pick this up from the bundle with no SDK release.

jobenjada (approved):

PR #9100, opened 2026-08-31
docs: replace untrusted self-hosting minimum with real v6 sizing

jobenjada:
No ticket — docs correction raised from a customer infra planning review.

## What & why

**Was:** The overview advertised a 1 vCPU / 2 GB / 8 GB SSD minimum — about a third of a v6 install, so a reader following it gets an OOM.

**Now:** Sizing is derived from the Helm chart's defaults and kept to intro length, with the deeper Kubernetes detail on the pages that already cover it.

- Chart defaults request 1.9 vCPU / 2.9 GiB across eight services, not 1 vCPU / 2 GB.
- Capacity is provider-neutral — no VM SKUs, no cloud named.
- Two undocumented autoscaling traps move to the Kubernetes guide's replica section.

## Where to look

- [overview.mdx:13](docs/self-hosting/overview.mdx#L13) — the numbers, vs. `charts/formbricks/values.yaml`
- [overview.mdx:63](docs/self-hosting/overview.mdx#L63) — load scenarios
- [kubernetes.mdx:201](docs/self-hosting/setup/kubernetes.mdx#L201) — the two HPA traps

[Preview](https://formbricks-claude-formbricks-enterprise-resources-e37b17.mintlify.site/self-hosting/overview)

## Breaking changes

None. Documentation only.

## Migrations & env

- none

## How this was tested

Rerun: `npx prettier --check docs/self-hosting/overview.mdx docs/self-hosting/setup/kubernetes.mdx`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Numbers match chart defaults | manual | Re-derived from `values.yaml`; invented Hub API row dropped |
| Embeddings figures name the right pool | manual | Foreground 4 vCPU, `background` 8 vCPU |
| Storage guidance matches the S3 contract | manual | Azure Blob removed; gateway requirement stated |
| Load scenarios rest on shipped defaults | manual | Job count traced through `sendToPipeline` callers |
| Cut content covered elsewhere, not lost | manual | Managed PG/Redis, replicas, Cube, `pgvector` already documented |
| Renders unclipped, anchors resolve | manual | All tables `scrollWidth == clientWidth` on the preview |

**Open gaps**

- No responses-per-second per replica — no load test exists, so the page says so.
- Single-server rows are derived from chart totals, not load-tested.
- Table widths measured at the preview's 640px well only.

**Risks**

- Figures drift if chart defaults change; the page names `values.yaml` to re-derive.

<details>
<summary>Review feedback addressed</summary>

Six factual corrections from @BhagyaAmarasinghe, each answered in its own thread:

1. Embeddings CPU rows were reversed between the foreground runtime and the worker pool.
2. `Standard_B2s` is not valid for an AKS system node pool — all VM SKUs are gone, capacity is provider-neutral.
3. The shipped `maxReplicas: 10` could not be honoured by a 5 × 2-vCPU pool.
4. Free control-plane tiers carry no financially backed SLA.
5. Azure Blob Storage has no S3 API and cannot back file uploads directly.
6. "Production-grade" overstated HA: every replica count in the chart ships as `1`.

Found while re-deriving: the old Hub API row (100m / 256 Mi / 512 Mi) was invented — the chart sets no requests or limits on it.

From CodeRabbit, eight findings. Seven accepted: heading casing, the Enterprise note block, the `minReplicas: 1` default, the `BULLMQ_WORKER_COUNT` term missing from the capacity formula, concurrency described as free, the HPA scaling on CPU and memory only, and Redis refusing writes under `maxmemory-policy noeviction` rather than absorbing an unbounded backlog. One rejected on its arithmetic: `responseCreated` and `responseFinished` enqueue the same job name with a different payload, so a finished response is two jobs, not three — but it pointed at a real miss, `put-response-handler.ts:247`, which enqueues `responseUpdated` on every progress save.

</details>

<details>
<summary>What was cut, and where it lives now</summary>

The page had grown to ~200 lines of capacity planning on the first page a self-hoster reads. Cut back to the corrected numbers, the single-server vs. Kubernetes choice, and two short load scenarios.

The per-component table, cluster table, managed-service table, HPA arithmetic, and the assumptions and scale-first lists are gone. Most of it duplicated existing pages: `Using Managed PostgreSQL And Redis`, `Production Replicas And Disruption Budgets` and `Cube` on the Kubernetes guide, `pgvector` on cluster-setup and migration, worker tuning on job-runner. The two facts documented nowhere else — the HPA ceiling exceeding the node pool, and the HPA scaling on CPU and memory only — moved to the Kubernetes guide's replica section.

</details>

<details>
<summary>Cost figures from the source planning doc were deliberately left out</summary>

The sizing was derived from a customer-specific Azure infra plan. The resource shape is published; the monthly cost tables are not, since they are engagement- and Azure-specific. Naming no cloud at all also removes the AKS-validity problem the SKU row created.

</details>

<details>
<summary>Mintlify: the 🔴 Failed comment above is stale</summary>

The `mintlify[bot]` comment reflects the very first pushed commit, which was force-pushed away before review — the failure there was Mintlify's transient "Failed to get OpenApi files", unrelated to the MDX. Every commit since has deployed green: `Mintlify Deployment | Deployment Succeeded` and `link-rot` green on each, and the preview URL serves the current page. Mintlify's GitHub App does not refresh that comment after a force-push, so the check runs are the signal, not the table.

</details>

---

> [!NOTE]

### DIFF
--- a/docs/self-hosting/overview.mdx
+++ b/docs/self-hosting/overview.mdx
@@ -10,11 +10,15 @@ Formbricks is not a single container. The v6 baseline runs the Formbricks app, F
 worker), Cube, PostgreSQL, Redis/Valkey, and SpiceDB, so size the host for the whole stack rather than for
 the app alone.
 
-| Deployment                                                       | CPU                     | Memory                | Disk               |
-| ---------------------------------------------------------------- | ----------------------- | --------------------- | ------------------ |
-| Evaluation on a single server ([Docker](/self-hosting/setup/docker), [one-click](/self-hosting/setup/one-click)) | 2 vCPU                  | 4 GB                  | 20 GB SSD          |
-| Small production on a single server                              | 4 vCPU                  | 8 GB                  | 50 GB SSD          |
-| [Minimal enterprise setup](#minimal-enterprise-setup) (Kubernetes) | 6 vCPU across the nodes | 12 GB across the nodes | managed services   |
+| Deployment                        | CPU          | Memory      | Disk       |
+| --------------------------------- | ------------ | ----------- | ---------- |
+| Single server, evaluation         | 2 vCPU       | 4 GB        | 20 GB SSD  |
+| Single server, small production   | 4 vCPU       | 8 GB        | 50 GB SSD  |
+| Kubernetes, minimal enterprise    | 6 vCPU total | 12 GB total | managed    |
+
+The single-server rows apply to the [Docker](/self-hosting/setup/docker) and
+[one-click](/self-hosting/setup/one-click) installs. The Kubernetes row is spread across a node pool and
+assumes managed data services — see [Minimal Enterprise Setup](#minimal-enterprise-setup).
 
 <Info>
   **HTTPS Required**: Ensure your server runs with a valid HTTPS certificate.
@@ -30,31 +34,33 @@ the app alone.
 These are the defaults the Helm chart ships, and they are the most precise sizing statement we can give.
 A Docker Compose install runs the same services without enforced limits.
 
-| Component                | Replicas | CPU request | Memory request | Memory limit |
-| ------------------------ | -------- | ----------- | -------------- | ------------ |
-| Formbricks app           | 1        | 1000m       | 1 Gi           | 2 Gi         |
-| Formbricks Hub API       | 1        | 100m        | 256 Mi         | 512 Mi       |
-| Formbricks Hub worker    | 1        | 100m        | 256 Mi         | 512 Mi       |
-| Cube                     | 1        | 250m        | 512 Mi         | 1 Gi         |
-| PostgreSQL (bundled)     | 1        | 250m        | 512 Mi         | 1 Gi         |
-| Redis/Valkey (bundled)   | 1        | 100m        | 128 Mi         | 192 Mi       |
-| SpiceDB                  | 2        | 100m each   | 256 Mi each    | 512 Mi each  |
-| **Total**                |          | **2 vCPU**  | **3.2 GiB**    | **6.2 GiB**  |
+| Component              | CPU request | Memory request | Memory limit |
+| ---------------------- | ----------- | -------------- | ------------ |
+| Formbricks app         | 1000m       | 1 Gi           | 2 Gi         |
+| Hub API                | 100m        | 256 Mi         | 512 Mi       |
+| Hub worker             | 100m        | 256 Mi         | 512 Mi       |
+| Cube                   | 250m        | 512 Mi         | 1 Gi         |
+| PostgreSQL (bundled)   | 250m        | 512 Mi         | 1 Gi         |
+| Redis/Valkey (bundled) | 100m        | 128 Mi         | 192 Mi       |
+| SpiceDB (2 replicas)   | 200m        | 512 Mi         | 1 Gi         |
+| **Total**              | **2 vCPU**  | **3.2 GiB**    | **6.2 GiB**  |
 
 Migration Jobs run alongside this on install and upgrade, and Kubernetes itself takes roughly 0.5 vCPU and
-1 GB per node before any of it is schedulable.
+1 GB per node before any of it is schedulable. The two bundled rows drop out when you use managed
+PostgreSQL and Redis.
 
 ### Optional Services Are Much Heavier
 
-Everything below is disabled by default. Budget for it separately, and prefer an external endpoint over
-self-hosting the model if you only need the AI features occasionally.
+The [AI features](/self-hosting/advanced/enterprise-features/ai-features) below are all disabled by default.
+Budget for them separately, and prefer an external endpoint over self-hosting the model if you only need
+them occasionally.
 
-| Component                                                        | CPU request | Memory request | Notes                        |
-| ---------------------------------------------------------------- | ----------- | -------------- | ---------------------------- |
-| Hub embeddings (TEI)                                             | 8 vCPU      | 8 Gi           | plus a 10 Gi model cache     |
-| Hub background embeddings pool                                   | 4 vCPU      | 8 Gi           | plus a 10 Gi model cache     |
-| [AI taxonomy](/self-hosting/advanced/enterprise-features/ai-features) | 1 vCPU      | 2 Gi           | 4 Gi limit                   |
-| Bundled Qwen/vLLM                                                | 400m router | 1 Gi router    | requires GPU-capable nodes   |
+| Component            | CPU request | Memory request | Notes               |
+| -------------------- | ----------- | -------------- | ------------------- |
+| Hub embeddings (TEI) | 8 vCPU      | 8 Gi           | + 10 Gi model cache |
+| Background embeddings pool | 4 vCPU | 8 Gi          | + 10 Gi model cache |
+| AI taxonomy          | 1 vCPU      | 2 Gi           | 4 Gi limit          |
+| Qwen/vLLM router     | 400m        | 1 Gi           | needs GPU nodes     |
 
 ## Minimal Enterprise Setup
 
--- a/docs/self-hosting/overview.mdx
+++ b/docs/self-hosting/overview.mdx
@@ -4,21 +4,120 @@ description: "Learn how to self-host Formbricks."
 icon: "server"
 ---
 
-### System Requirements
+## System Requirements
 
-<Info>
-  **Minimum Setup**: 1 vCPU, 2 GB RAM, 8 GB SSD.
-</Info>
+Formbricks is not a single container. The v6 baseline runs the Formbricks app, Formbricks Hub (API and
+worker), Cube, PostgreSQL, Redis/Valkey, and SpiceDB, so size the host for the whole stack rather than for
+the app alone.
+
+| Deployment                                                       | CPU                     | Memory                | Disk               |
+| ---------------------------------------------------------------- | ----------------------- | --------------------- | ------------------ |
+| Evaluation on a single server ([Docker](/self-hosting/setup/docker), [one-click](/self-hosting/setup/one-click)) | 2 vCPU                  | 4 GB                  | 20 GB SSD          |
+| Small production on a single server                              | 4 vCPU                  | 8 GB                  | 50 GB SSD          |
+| [Minimal enterprise setup](#minimal-enterprise-setup) (Kubernetes) | 6 vCPU across the nodes | 12 GB across the nodes | managed services   |
 
 <Info>
   **HTTPS Required**: Ensure your server runs with a valid HTTPS certificate.
 </Info>
 
-#### **Self-Hosting Options**
+<Warning>
+  4 GB is a floor, not a comfortable target. It leaves no headroom for a large response import, a survey
+  bundle rebuild, or any of the optional AI services below.
+</Warning>
+
+### Where The Memory Goes
+
+These are the defaults the Helm chart ships, and they are the most precise sizing statement we can give.
+A Docker Compose install runs the same services without enforced limits.
+
+| Component                | Replicas | CPU request | Memory request | Memory limit |
+| ------------------------ | -------- | ----------- | -------------- | ------------ |
+| Formbricks app           | 1        | 1000m       | 1 Gi           | 2 Gi         |
+| Formbricks Hub API       | 1        | 100m        | 256 Mi         | 512 Mi       |
+| Formbricks Hub worker    | 1        | 100m        | 256 Mi         | 512 Mi       |
+| Cube                     | 1        | 250m        | 512 Mi         | 1 Gi         |
+| PostgreSQL (bundled)     | 1        | 250m        | 512 Mi         | 1 Gi         |
+| Redis/Valkey (bundled)   | 1        | 100m        | 128 Mi         | 192 Mi       |
+| SpiceDB                  | 2        | 100m each   | 256 Mi each    | 512 Mi each  |
+| **Total**                |          | **2 vCPU**  | **3.2 GiB**    | **6.2 GiB**  |
+
+Migration Jobs run alongside this on install and upgrade, and Kubernetes itself takes roughly 0.5 vCPU and
+1 GB per node before any of it is schedulable.
+
+### Optional Services Are Much Heavier
+
+Everything below is disabled by default. Budget for it separately, and prefer an external endpoint over
+self-hosting the model if you only need the AI features occasionally.
+
+| Component                                                        | CPU request | Memory request | Notes                        |
+| ---------------------------------------------------------------- | ----------- | -------------- | ---------------------------- |
+| Hub embeddings (TEI)                                             | 8 vCPU      | 8 Gi           | plus a 10 Gi model cache     |
+| Hub background embeddings pool                                   | 4 vCPU      | 8 Gi           | plus a 10 Gi model cache     |
+| [AI taxonomy](/self-hosting/advanced/enterprise-features/ai-features) | 1 vCPU      | 2 Gi           | 4 Gi limit                   |
+| Bundled Qwen/vLLM                                                | 400m router | 1 Gi router    | requires GPU-capable nodes   |
+
+## Minimal Enterprise Setup
+
+This is the smallest deployment we consider production-grade: a managed Kubernetes cluster with managed
+data services, sized for a modest steady load and left free to autoscale into peaks. It is the same shape as
+a large deployment, just narrower — nothing here has to be re-architected when traffic grows.
+
+### Cluster
+
+| Resource       | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| Node pool      | 3 nodes × 2 vCPU / 4 GB (Azure `Standard_B2s`, AWS `t3.medium`, GCP `e2-medium`) |
+| Autoscaling    | Cluster autoscaler 3 → 5 nodes; app HPA `minReplicas: 2`                 |
+| Control plane  | The free/standard managed tier is sufficient at this size                |
+| Edge           | A load balancer plus [Envoy or an equivalent edge rate limiter](/self-hosting/advanced/rate-limiting) |
+
+Two app replicas are the reason for the third node. One replica cannot be evicted without downtime, which
+makes node drains and rolling upgrades disruptive — see the
+[Kubernetes deployment guide](/self-hosting/setup/kubernetes) for the replica and PodDisruptionBudget
+settings.
+
+### Managed Data Services
+
+Run these outside the cluster. They hold the state, and a managed service gives you backups and failover
+that the bundled charts do not.
+
+| Service        | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| PostgreSQL     | 2 vCores burstable, 64 GB SSD, `pgvector` available, PITR backups enabled |
+| SpiceDB store  | A separate database and login on the same PostgreSQL server              |
+| Redis/Valkey   | 250 MB managed cache; 1 GB once you enable Envoy rate limiting           |
+| Object storage | ~50 GB S3-compatible or Azure Blob for [file uploads](/self-hosting/configuration/file-uploads) |
+| Registry       | A basic private registry, if you mirror the Formbricks images            |
+
+Moving PostgreSQL and Redis out of the cluster frees roughly 350m CPU and 640 Mi of memory per node pool,
+which is what makes three small nodes workable.
+
+<Note>
+  Budget for data egress separately. Survey traffic is served to end users, so egress scales with responses
+  rather than with cluster size.
+</Note>
+
+### What This Sizing Assumes
+
+- No AI features. Self-hosted embeddings or an in-cluster LLM change the node pool completely — use an
+  external endpoint at this size.
+- A single region and no cross-region replication.
+- Peaks handled by autoscaling rather than by permanently provisioned capacity.
+
+### What To Scale First
+
+1. **App replicas**, via the HPA — the app is stateless and this is the cheapest lever.
+2. **PostgreSQL vCores**, once response writes or dashboard queries start queuing.
+3. **Redis**, which becomes the dominant cost at high request rates. If that matters more than the
+   operational simplicity of a managed cache, run Valkey in-cluster instead.
+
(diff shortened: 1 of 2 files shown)


jobenjada:
add scenario context:

- if 25k responses at one time, how many pods / resources?
- if 50k registered users, etc.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+
+This is the smallest deployment we consider production-grade: a managed Kubernetes cluster with managed
+data services, sized for a modest steady load and left free to autoscale into peaks. It is the same shape as
+a large deployment, just narrower — nothing here has to be re-architected when traffic grows.
+
+### Cluster
+
+| Resource       | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| Node pool      | 3 nodes × 2 vCPU / 4 GB (Azure `Standard_B2s`, AWS `t3.medium`, GCP `e2-medium`) |
```
BhagyaAmarasinghe: **Use an AKS-supported node topology.** `Standard_B2s` cannot be the required AKS system pool: AKS does not support B-series VMs for system node pools, and Microsoft recommends avoiding B-series for AKS generally. If a separate compliant system pool is required, the advertised 6 vCPU / 12 GB total and the quotation both change. Please replace this with an AKS-valid system/user-pool topology or keep the capacity recommendation provider-neutral. References: [AKS limits](https://learn.microsoft.com/en-us/azure/aks/quotas-skus-regions) and [AKS reliability guidance](https://learn.microsoft.com/en-us/azure/aks/best-practices-app-cluster-reliability).

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+
+Run these outside the cluster. They hold the state, and a managed service gives you backups and failover
+that the bundled charts do not.
+
+| Service        | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| PostgreSQL     | 2 vCores burstable, 64 GB SSD, `pgvector` available, PITR backups enabled |
+| SpiceDB store  | A separate database and login on the same PostgreSQL server              |
+| Redis/Valkey   | 250 MB managed cache; 1 GB once you enable Envoy rate limiting           |
+| Object storage | ~50 GB S3-compatible or Azure Blob for [file uploads](/self-hosting/configuration/file-uploads) |
```
BhagyaAmarasinghe: **Remove native Azure Blob from the supported storage options.** Formbricks currently uses the AWS `S3Client` and documents an S3-compatible storage contract. Native Azure Blob Storage cannot be substituted directly. Please remove Azure Blob here or explicitly require a supported S3-compatible gateway.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+This is the smallest deployment we consider production-grade: a managed Kubernetes cluster with managed
+data services, sized for a modest steady load and left free to autoscale into peaks. It is the same shape as
+a large deployment, just narrower — nothing here has to be re-architected when traffic grows.
+
+### Cluster
+
+| Resource       | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| Node pool      | 3 nodes × 2 vCPU / 4 GB (Azure `Standard_B2s`, AWS `t3.medium`, GCP `e2-medium`) |
+| Autoscaling    | Cluster autoscaler 3 → 5 nodes; app HPA `minReplicas: 2`                 |
```
BhagyaAmarasinghe: **Align the HPA and node-pool ceilings.** The chart retains `maxReplicas: 10` with a 1-vCPU request per app pod. Five 2-vCPU nodes provide only 10 raw vCPU and roughly 7.5 vCPU after the Kubernetes reserve stated above, while Hub, Cube, SpiceDB, the operator, and edge components consume additional CPU. This node pool therefore cannot honor the documented HPA ceiling; higher replicas will remain Pending. Set a compatible HPA maximum or increase the node-pool ceiling/size.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+
+### Optional Services Are Much Heavier
+
+The [AI features](/self-hosting/advanced/enterprise-features/ai-features) below are all disabled by default.
+Budget for them separately, and prefer an external endpoint over self-hosting the model if you only need
+them occasionally.
+
+| Component            | CPU request | Memory request | Notes               |
+| -------------------- | ----------- | -------------- | ------------------- |
+| Hub embeddings (TEI) | 8 vCPU      | 8 Gi           | + 10 Gi model cache |
```
BhagyaAmarasinghe: **Correct the embedding CPU figures.** These two rows are reversed relative to the chart defaults: foreground Hub embeddings request 4 vCPU, while the background embeddings pool requests 8 vCPU (`hub.embeddings.resources` versus `hub.embeddings.background.resources`).

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+| Background embeddings pool | 4 vCPU | 8 Gi          | + 10 Gi model cache |
+| AI taxonomy          | 1 vCPU      | 2 Gi           | 4 Gi limit          |
+| Qwen/vLLM router     | 400m        | 1 Gi           | needs GPU nodes     |
+
+## Minimal Enterprise Setup
+
+This is the smallest deployment we consider production-grade: a managed Kubernetes cluster with managed
```
BhagyaAmarasinghe: **Separate the low-load, non-HA option from an HA production topology.** Calling this the smallest production-grade deployment overstates the availability delivered: only the app is configured with two replicas, while Hub API, Hub worker, Cube, and bundled edge components retain singleton defaults. Please explicitly document that a very-low-load installation can run without AKS on the supported single-VM Docker/one-click path, with host/maintenance downtime, and describe Kubernetes as the separate HA/SLA option rather than an AKS requirement.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+data services, sized for a modest steady load and left free to autoscale into peaks. It is the same shape as
+a large deployment, just narrower — nothing here has to be re-architected when traffic grows.
+
+### Cluster
+
+| Resource       | Minimum                                                                  |
+| -------------- | ------------------------------------------------------------------------ |
+| Node pool      | 3 nodes × 2 vCPU / 4 GB (Azure `Standard_B2s`, AWS `t3.medium`, GCP `e2-medium`) |
+| Autoscaling    | Cluster autoscaler 3 → 5 nodes; app HPA `minReplicas: 2`                 |
+| Control plane  | The free/standard managed tier is sufficient at this size                |
```
BhagyaAmarasinghe: **Do not recommend the AKS Free tier for enterprise production.** Microsoft positions AKS Free for development/testing and it has no financially backed uptime SLA; Standard is the production tier. Please specify Standard for the production-grade scenario or make the control-plane SLA an explicit customer decision. Reference: [AKS pricing tiers](https://learn.microsoft.com/en-us/azure/aks/free-standard-pricing-tiers).

BhagyaAmarasinghe (commented):
Sizing guidance review for the customer quotation and low-load deployment scenario.

docs/self-hosting/overview.mdx
jobenjada: Fixed — the two rows were swapped. `hub.embeddings.resources` requests **4** vCPU (foreground, used by the Hub API) and `hub.embeddings.background.resources` requests **8** vCPU (the worker-only pool). Also added the background pool's own `maxReplicas: 6`, since 8 vCPU × 6 is the ceiling that actually needs budgeting.

docs/self-hosting/overview.mdx
jobenjada: Took the provider-neutral option — every VM SKU is gone from the page. The row now reads “3 nodes, at least 4 vCPU and 8 GB each”, with prose explaining that a 2 vCPU node cannot hold the app's 1 vCPU request alongside the platform reserve and the singletons, and that some managed services exclude burstable instance families from system node pools entirely, so readers should check their provider's supported node types. The advertised “6 vCPU / 12 GB total” is gone too — that arithmetic only worked because of the B2s row.

docs/self-hosting/overview.mdx
jobenjada: Fixed from both sides. The pool is now 4 vCPU nodes autoscaling 3 → 6, and a new subsection gives the arithmetic rather than a bare number: allocatable ≈ `(node vCPU − 0.5) × nodes`, minus ~0.6 vCPU for Hub worker, Cube, SpiceDB and its operator, minus headroom for the Hub API — which the chart ships with **no** request at all, another thing the old table got wrong (it had an invented 100m/256Mi row). Six 4-vCPU nodes clear the shipped `maxReplicas: 10`; a pool that stops lower has to lower `maxReplicas` to match. The page now says explicitly that an HPA does not lower its own ceiling and the extra replicas stay `Pending`.

docs/self-hosting/overview.mdx
jobenjada: Fixed. The row reads “A paid tier that carries an uptime SLA”, and the prose says free control-plane tiers are positioned for development and testing and carry no financially backed uptime SLA, so the production tier is part of the cost of the SLA rather than an upsell. Kept it provider-neutral rather than naming AKS Standard, in line with dropping the Azure SKUs.

docs/self-hosting/overview.mdx
jobenjada: Removed. The table row is now “~50 GB, S3-compatible”, and the prose states that Formbricks speaks the S3 API only — AWS S3, DigitalOcean Spaces, Wasabi, StorJ or a self-hosted equivalent — and that a service without an S3 API, such as Azure Blob Storage, needs an S3-compatible gateway in front of it. That matches the file-uploads page and the `S3_*` env contract.

docs/self-hosting/overview.mdx
jobenjada: Restructured along those lines.

“Minimal Enterprise Setup” is gone. The section is now “Kubernetes production setup”, and the requirements row is “Single server, low load” rather than “small production”. The single-server prose now says that Docker and one-click run every service on one host, that a host reboot or a Formbricks upgrade is therefore downtime, and that at low load this is usually an acceptable trade and **not** a reason to reach for Kubernetes — Kubernetes is framed as the choice when you need to survive a node failure or carry an uptime SLA.

Your singleton point is now its own warning on that section: `hub.replicas`, `hub.worker.replicas` and `cube.replicas` all ship as `1`, and Cube in particular cannot just be raised — the chart fails the render above one replica with the default in-memory cache driver, so multiple Cube replicas need Cube Store. So the page no longer claims HA beyond the app tier, and says to raise the Hub counts and plan Cube Store before promising an SLA that covers Hub and analytics.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
   **HTTPS Required**: Ensure your server runs with a valid HTTPS certificate.
 </Info>
 
-#### **Self-Hosting Options**
+<Warning>
+  4 GB is a floor, not a comfortable target. It leaves no headroom for a large response import, a survey
+  bundle rebuild, or any of the optional AI services below.
+</Warning>
+
+### Where The Memory Goes
```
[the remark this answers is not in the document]
jobenjada: Done — the four headings are sentence case now: “Where the memory goes”, “Optional services are much heavier”, “What this sizing assumes”, “What to scale first”, plus “System requirements” and the renamed “Kubernetes production setup”. The pre-existing title-case headings further down the page are outside this diff.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,127 @@ description: "Learn how to self-host Formbricks."
...
+| SpiceDB (2 replicas)   | 200m        | 512 Mi         | 1 Gi         |
+| **Total**              | **2 vCPU**  | **3.2 GiB**    | **6.2 GiB**  |
+
+Migration Jobs run alongside this on install and upgrade, and Kubernetes itself takes roughly 0.5 vCPU and
+1 GB per node before any of it is schedulable. The two bundled rows drop out when you use managed
+PostgreSQL and Redis.
+
+### Optional Services Are Much Heavier
+
+The [AI features](/self-hosting/advanced/enterprise-features/ai-features) below are all disabled by default.
```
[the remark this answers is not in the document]
jobenjada: Added, using the block from the other Enterprise pages:

```
<Note>
  AI features are an Enterprise feature. Configure your
  [Enterprise license](/self-hosting/advanced/license) before enabling them.
</Note>
```

It sits directly above the optional-services table.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,201 @@ description: "Learn how to self-host Formbricks."
...
+
+## Kubernetes production setup
+
+A managed Kubernetes cluster with managed data services, sized for a modest steady load and left free to
+autoscale into peaks. It is the same shape as a large deployment, just narrower — nothing here has to be
+re-architected when traffic grows.
+
+<Warning>
+  Chart defaults make the **app** tier highly available and leave everything else a singleton:
+  `hub.replicas`, `hub.worker.replicas` and `cube.replicas` all ship as `1`. Cube cannot simply be raised —
```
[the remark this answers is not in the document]
jobenjada: Fixed. The warning now opens with the default rather than the override: "Every replica count in the chart ships as `1`, including `autoscaling.minReplicas` for the app. The `minReplicas: 2` in the table below is a required override, not a default." Nothing on the page claims HA under chart defaults now.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,202 @@ description: "Learn how to self-host Formbricks."
...
+| Qwen/vLLM router       | 400m        | 1 Gi           | needs GPU nodes     |
+
+The worker-only pool (`hub.embeddings.background`) is the expensive one — 8 vCPU per replica, with its own
+HPA ceiling of 6. Enable it only when the foreground runtime cannot keep up with backfill.
+
+## Kubernetes production setup
+
+A managed Kubernetes cluster with managed data services, sized for a modest steady load and left free to
+autoscale into peaks. It is the same shape as a large deployment, just narrower — nothing here has to be
+re-architected when traffic grows.
```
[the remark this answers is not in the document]
jobenjada: Fixed. That sentence over-reached and did contradict the warning under it. It now reads: "The app tier grows by adding replicas and needs no redesign; raising availability beyond it, or scaling analytics, does change the topology — see the warning below."

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,201 @@ description: "Learn how to self-host Formbricks."
...
+
+Two questions come up more than any others, and neither is answered by the tables above, because both are
+about rate and volume rather than about the resting stack.
+
+**25,000 responses arriving at once.** Spread over an hour that is roughly 7 writes a second; compressed
+into ten minutes, roughly 42. The per-IP limit of 100 requests a minute on the public response route does
+not throttle a burst like that, because a campaign burst arrives from thousands of distinct IPs. The job
+runner does: each response enqueues a `response-pipeline.process` job, and every app replica ships one
+worker at a concurrency of one, so drain rate is `replicas × BULLMQ_WORKER_CONCURRENCY`. A burst that
+outruns it backs up in Redis rather than failing. Raise `BULLMQ_WORKER_CONCURRENCY` first — it costs no
```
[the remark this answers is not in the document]
jobenjada: All three corrected.

- Capacity is now `web replicas × BULLMQ_WORKER_COUNT × BULLMQ_WORKER_CONCURRENCY`, with all three noted as defaulting to `1`.
- “Drain rate” is gone. The page says this is a count of jobs in flight, not a rate, and that throughput also depends on how long the average `response-pipeline.process` job takes on the instance.
- “Costs no pods” is gone: “Raising `BULLMQ_WORKER_COUNT` or `BULLMQ_WORKER_CONCURRENCY` adds no pods, but the work still lands on the same pod’s CPU and memory and on Redis and Postgres behind it, so measure before raising either.” The “What to scale first” entry carries the same caveat.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,201 @@ description: "Learn how to self-host Formbricks."
...
+pods — then let the HPA add replicas, and give Redis room for the queue. See
+[job runner](/self-hosting/configuration/job-runner).
```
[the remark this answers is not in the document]
jobenjada: Good catch — added as its own warning, since it inverts the intuition:

> The app HPA scales on CPU and memory utilization only. Queue depth is not one of its metrics, so a backlog can grow while both targets sit under their thresholds and no replica is added. Backlog-driven scaling needs a queue-depth metric published to the metrics API and added to the HPA yourself.

The closing note now says to watch pod CPU and queue depth separately, because they move independently.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,202 @@ description: "Learn how to self-host Formbricks."
...
+
+Two questions come up more than any others, and neither is answered by the tables above, because both are
+about rate and volume rather than about the resting stack.
+
+**25,000 responses arriving at once.** Spread over an hour that is roughly 7 writes a second; compressed
+into ten minutes, roughly 42. The per-IP limit of 100 requests a minute on the public response route does
+not throttle a burst like that, because a campaign burst arrives from thousands of distinct IPs. The job
+runner does: every response enqueues a `response-pipeline.process` job, and a completed one enqueues two —
+`responseCreated` and `responseFinished`. Every app replica ships a single worker at a concurrency of one,
+so drain rate is `replicas × BULLMQ_WORKER_CONCURRENCY`, and 25,000 finished responses is 50,000 jobs. A
```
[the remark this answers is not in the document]
jobenjada: The arithmetic in the finding does not hold, but it pointed at a real error, so the passage is rewritten.

`sendToPipeline({ event: "responseCreated" })` and `sendToPipeline({ event: "responseFinished" })` both enqueue the **same** job name, `response-pipeline.process`, with a different `event` in the payload — there is no separate “processing job” alongside them. A response submitted finished in one POST is 2 jobs, not 3.

The real miss was elsewhere: `put-response-handler.ts:247` enqueues `responseUpdated` on **every** SDK progress save, so the count is not a fixed multiple of the response count at all. The page now says: `responseCreated` on the initial write, one `responseUpdated` each time the SDK saves progress, and `responseFinished` on submission — so the total scales with how often answers are saved, and a multi-page survey can be several jobs per response. The flat “50,000 jobs” figure is gone.

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,217 @@ description: "Learn how to self-host Formbricks."
...
+the job queue behind it instead.
+
+A response enqueues several `response-pipeline.process` jobs, not one: `responseCreated` on the initial
+write, one `responseUpdated` each time the SDK saves progress, and `responseFinished` on submission. So the
+job count scales with how many times a respondent's answers are saved, not with the response count — a
+multi-page survey can be several jobs per response. Size Redis for that queue, not for 25,000 entries.
+
+Consumer capacity is `web replicas × BULLMQ_WORKER_COUNT × BULLMQ_WORKER_CONCURRENCY`, all three of which
+default to `1`. That is a count of jobs in flight, not a rate: throughput also depends on how long the
+average `response-pipeline.process` job takes on your instance. A queue that grows faster than it drains
```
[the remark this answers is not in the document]
jobenjada: Correct, and it contradicted a sentence I had written. Fixed.

Verified the chain: `charts/formbricks/values.yaml:991` pins `maxmemory-policy noeviction` (with the same requirement stated for external Redis at `:978`), `enqueueResponsePipelineJob` rethrows at `packages/jobs/src/queue.ts:226`, `sendToPipeline` rethrows too, and the client create route awaits it **outside** the `try` that wraps `createResponseWithQuotaEvaluation`. So the row is already committed when the enqueue fails — no answer is lost, but the client still gets an error.

The "backs up in Redis rather than failing" claim is gone. The page now says:

> Redis is the hard limit on that backlog. The chart pins `maxmemory-policy noeviction` so BullMQ cannot silently lose jobs, which means a full Redis refuses writes rather than evicting. Enqueue errors are not swallowed: the response row is already committed to Postgres, so no answer is lost, but the client still gets an error. Retention settings only trim completed and failed jobs — they do not make the queue unbounded. Size Redis for the deepest backlog you expect, not for the steady state.

The managed-services row is now "250 MB; 1 GB with Envoy rate limiting; more for backlogs".

docs/self-hosting/overview.mdx
```diff
@@ -4,22 +4,83 @@ description: "Learn how to self-host Formbricks."
...
+host for the whole stack rather than for the app alone.
+
+| Deployment                | CPU       | Memory    | Disk      |
+| ------------------------- | --------- | --------- | --------- |
+| Single server, evaluation | 2 vCPU    | 4 GB      | 20 GB SSD |
+| Single server, low load   | 4 vCPU    | 8 GB      | 50 GB SSD |
+| Kubernetes, production    | see below | see below | managed   |
+
+Chart defaults request 1.9 vCPU and 2.9 GiB across those services, before the roughly 0.5 vCPU and 1 GB
+that Kubernetes reserves per node. `charts/formbricks/values.yaml` is the source for every figure here.
```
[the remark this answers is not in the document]
jobenjada: Correct on all three counts. Fixed:

> Chart defaults request 1.9 vCPU and 2.9 GiB across those services — see `charts/formbricks/values.yaml`, plus `charts/spicedb-operator/values.yaml` for the operator. On Kubernetes, add whatever your provider reserves per node on top, typically around 0.5 vCPU and 1 GB. The single-server rows are derived from those totals rather than load-tested.

The operator's 20m / 64 Mi does come from the dependent chart, the node reserve is a provider characteristic and not a chart value at all, and the single-server rows are a derivation. "Source for every figure here" covered none of that.

PR #9105, opened 2026-09-01
feat: filter responses by embedded data + reserved fields as typed filters [ENG-1848]

pandeymangg:
## What & why

Response filtering could not reach Embedded Data or reserved fields. The "Meta" filter category listed whatever string keys the stored responses happened to hold (raw camelCase, no number fields, empty on a fresh survey), hidden fields only offered equals/not-equals, and computed fields (variables) were not filterable at all — no criteria key, no query branch, no UI. This is Renata's hard requirement: stakeholders must slice by brand, country, page type.

This PR makes the whole pipeline typed and catalog-driven:

- **Filter dropdown** enumerates reserved fields from the catalog (same list, localized labels, and icons as the response table), ingested fields from the survey's declared rows (filterable before the first response), and a new **Variables** group for computed fields.
- **Operators per dataType**, following the editor's logic-builder model: string → equality + contains/starts/ends (+negations), number → comparisons, date → before/after, boolean → is — and **Is set / Is not set** on every family (the value box disables).
- **Criteria schema** gains `reserved` (catalog-name keyed) and `variables` (storageKey keyed) groups; the `data` group gains text ops for string ingested fields. Values are coerced to the field's dataType (a jsonb number never matches a string `"12"`).
- **`buildWhereClause`** gains a reserved-name → storage-path locator map (`deviceType` → `["userAgent","device"]`, `durationSeconds` → `ttc._total` with `Math.round`-window math so DB filtering agrees with the projected seconds at every boundary) and a `variables` branch. **Filters fail closed** on shadowed names — a survey declaring `url` filters its own field, never the reserved read — enforced at enumeration *and* server-side against crafted criteria; unknown names and type-mismatched ops emit nothing.
- Gates match the export (ENG-1847): anonymized surveys don't offer never-captured fields, `ipAddress` follows its capture toggle — no always-empty filters.
- Observed dropdown values run through the shared projection, so `url` values appear query-stripped.

Date before/after ships on `lessThan`/`greaterThan` with ISO strings — verified empirically that Prisma passes string comparisons to jsonb, which orders same-format ISO strings chronologically.

## Linear ticket

https://linear.app/formbricks/issue/ENG-1848

## How this was tested

- Unit: `apps/web` **7826/7826** ✅, `packages/types` **372/372** ✅; typecheck ✅; lint ✅.
  - `utils.test.ts`: 13 new tests (locator paths incl. the `deviceType` remap, duration windows, presence via DbNull, shadowed/unknown/type-mismatch fail-closed, variables keying, data text ops, enumeration gates) + an anti-drift test pinning that every filterable catalog entry has a locator.
  - `surveys.test.ts`: catalog-driven options, per-dataType operator menus, declared-field enumeration, criteria bucketing (reserved/variables/data), number coercion, presence remap.
  - **Seven mutation fail-checks** verified each rule's test goes red without it (shadow-skip, deviceType remap, duration window, duration inclusion, number coercion, presence remap, reserved fail-closed).
- New e2e `apps/web/playwright/response-filter.spec.ts`: campaign + bare response → filter "UTM Source equals newsletter" narrows the table → switch to "Is not set" inverts it. Green; **red with the old `buildWhereClause`** (revert check).
- Query plan (`EXPLAIN`, local): reserved meta-path and `ttc._total` filters run as `surveyId`-scoped scans — identical shape to every existing JSON filter (no GIN indexes exist on Response JSON columns); no new index in v1.

<details>
<summary>Screenshots — filter dropdown before / after</summary>

**Field list** — before: raw keys observed from stored responses (blank icons, nothing on a fresh survey); after: catalog-driven with the table's labels and icons:

| Before | After |
| --- | --- |
| Old dropdown: raw meta keys like PagePath and Locale with missing icons | New dropdown: catalog-driven labels Source, URL, Country, Action with icons |

**End of the list** — before: no Variables group, raw `UtmSource`; after: number fields (Screen/Viewport), Hidden fields, and the new Variables group:

| Before | After |
| --- | --- |
| Old list end: UtmSource raw key, hidden fields, no Variables group | New list end: screen and viewport fields, hidden fields, Variables group with score |

**Operators for UTM Source** — before: Equals / Not equals only; after: the full string family plus Is set / Is not set:

| Before | After |
| --- | --- |
| Old operator menu: only Equals and Not equals | New operator menu: text operator family plus Is set and Is not set |

</details>

## Breaking changes

None — the criteria schema changes are additive (the old `meta` group is still accepted and translated), filter state is session-only, and the widened action response keeps its existing keys.

## QA / Test Plan

**How to test**

- [ ] Create a link survey with one question, a hidden field `plan`, and a number variable `score`. Submit one response via `?utm_source=newsletter&plan=gold` and one via the bare link (incognito).
- [ ] Responses tab → Filter → Add filter → the dropdown shows the Meta group (UTM Source, Browser, Duration Seconds, …), `plan` under Hidden fields, and `score` under Variables — **also on a brand-new survey with zero responses** (previously Meta was empty until responses existed).
- [ ] UTM Source → Equals → "newsletter" offered in the value dropdown → Apply → only the campaign response remains.
- [ ] UTM Source → Is not set → value box disabled → Apply → only the bare response remains. Close and reopen the popover → the row is still there.
- [ ] Page Path → Contains → free-text input appears → `/s/` matches both responses.
- [ ] Screen Width → Is greater than → numeric input; `plan` → Contains "gol"; `score` → Is less than 5.
- [ ] URL field → observed values in the dropdown show **without** query strings.
- [ ] Enable "Anonymize responses" → Country/Browser/OS/Device/IP Address disappear from the dropdown; disable IP capture → only IP Address disappears.
- [ ] A survey with a declared field named `url` (pre-existing/grandfathered) → Meta group has no URL entry; the declared field filters its own values.
- [ ] Apply a filter, then Download → Filtered responses (CSV) → the file honors the same filter. The Summary tab honors it too.
- [ ] Regression: question, tag, attribute, and Language filters behave as before.

**Preconditions / test data**

- Any workspace; a link survey with a hidden field and a number variable (see first step). No flags or entitlements.

**Risks & regressions**

- Duration Seconds only exists on finished responses — partials never match it (except via "Is not set"). Expected, not a bug.
- "Not equals" also matches responses where the field is absent (consistent with existing question filters).
- Typing a non-number into a number filter silently matches everything (the condition is dropped) — fail-closed by design.
- Date filters compare date-only input against stored ISO values: before/after are correct; exact equals won't match a stored datetime.

- none

### DIFF
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/actions.ts
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/actions.ts
@@ -105,11 +105,15 @@ export const getSurveyFilterDataAction = authenticatedActionClient
 
     const workspaceId = await getWorkspaceIdFromSurveyId(parsedInput.surveyId);
 
-    const [tags, { contactAttributes: attributes, meta, hiddenFields }, quotas = []] = await Promise.all([
+    const [
+      tags,
+      { contactAttributes: attributes, meta, hiddenFields, reservedValues, variableValues },
+      quotas = [],
+    ] = await Promise.all([
       getTagsByWorkspaceId(workspaceId),
       getResponseFilteringValues(parsedInput.surveyId),
       isQuotasAllowed ? getQuotas(parsedInput.surveyId) : [],
     ]);
 
-    return { environmentTags: tags, attributes, meta, hiddenFields, quotas };
+    return { environmentTags: tags, attributes, meta, hiddenFields, reservedValues, variableValues, quotas };
   });
--- a/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/ElementFilterComboBox.tsx
+++ b/apps/web/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/ElementFilterComboBox.tsx
@@ -4,6 +4,7 @@ import clsx from "clsx";
 import { ChevronDown, ChevronUp, X } from "lucide-react";
 import { useMemo, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
+import { TEmbeddedDataType } from "@formbricks/types/embedded-data";
 import { TI18nString } from "@formbricks/types/i18n";
 import { TSurveyElementTypeEnum } from "@formbricks/types/surveys/elements";
 import { OptionsType } from "@/app/(app)/workspaces/[workspaceId]/surveys/[surveyId]/components/ElementsComboBox";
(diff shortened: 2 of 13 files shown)


itsjavi (changes requested):
there are many sonar issues that need fixing

itsjavi (approved):

PR #9107, opened 2026-09-01
feat: relative date range validation for date questions

Dhruwang:
Fixes ENG-1804

## What & why

**Was:** A Date question's min/max could only be a fixed calendar date, so a rolling window — "no earlier than 3 days ago" — meant editing the survey every morning. The picker itself restricted nothing at all (a ±100 year span); out-of-range answers were only flagged on submit.

**Now:** Each of the four date rules takes either the existing fixed date or a bound relative to the response date, counted in calendar or working days. The picker greys out what the rules disallow, in both modes.

- Working days affects how the offset is counted, not which days are pickable — a weekend inside the window stays selectable.
- Relative bounds are inclusive ("3 days before" allows that day); fixed bounds stay exclusive.
- The row reads as an expression (`Submission day − 3 calendar days`) with a sentence below it naming the earliest/latest accepted date.

## Where to look

- [date-utils.ts](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1804-support-dynamic-relative-date-range-validation-for-date/packages/surveys/src/lib/validation/validators/date-utils.ts) — day arithmetic, the picker's bounds, and the server's one-day grace, applied to the reference date so working-day bounds keep pace with a respondent one day ahead
- [validators.ts#L401](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1804-support-dynamic-relative-date-range-validation-for-date/packages/surveys/src/lib/validation/validators.ts#L401) — the two comparison semantics must match `getDateBoundsFromRules` exactly
- [validation-rules.ts#L121](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1804-support-dynamic-relative-date-range-validation-for-date/packages/types/surveys/validation-rules.ts#L121) — params union; the legacy shape is a member, unchanged

https://github.com/user-attachments/assets/e201841f-100a-424f-b74b-81321041ec02

## Breaking changes

None. The rule-type enum is untouched and the fixed param shapes are unchanged members of the new union, so stored surveys parse as before. `SurveyValidationRule.params` is already `additionalProperties: true` in the OpenAPI bundle, so the v3 contract does not move.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm --filter @formbricks/surveys test src/lib/validation && pnpm --filter @formbricks/web test modules/survey/editor/lib/validation-rules-utils.test.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Working vs calendar day counting, across weekends and year ends | unit (mutation) | `date-utils.test.ts`; mutate `date-utils.ts:52` (drop the weekend skip) to redden |
| Relative bounds inclusive, fixed bounds exclusive, all four rules | unit (mutation) | `validators.test.ts`; mutate `validators.ts:411` (`>=` → `>`) or `:414` (`>` → `>=`) to redden |
| Server grace keeps pace with a working-day bound the picker offered from one day ahead or behind | unit (mutation) | `validators.test.ts` "keeps pace with a working-day…" ×2; drop `applyTimezoneGrace` at `validators.ts:410` to redden. Red against `83f7c48`, which shifted the result instead of the reference |
| Picker bounds mirror the validators, per mode | unit (mutation) + manual | `date-utils.test.ts`; mutate `date-utils.ts:156` (fixed shift `1` → `0`) to redden; picker allowed Aug 27–Sep 7 on a Tue Sep 1 reference |
| Server rejects out-of-window dates | manual | `400 "Please select a date between 2026-08-27 and 2026-09-07"`; in-window `200` |
| Relative rule editor row (mode, amount, unit, direction) | manual | Recording above; rule edits persist and drive the picker |

<details>
<summary>Coverage for the review fixes</summary>

| Behaviour | How | Outcome |
| --- | --- | --- |
| Server grace shrinks the `isNotBetween` hole rather than widening it | unit (mutation) | `validators.test.ts`; swap the roles at `validators.ts:73` to redden |
| `isNotBetween` quotes the strict window in its message on the server | unit (mutation) | `validators.test.ts`; change `"none"` at `validators.ts:484` to `"accepts-outside"` to redden |
| A window no date can satisfy leaves the picker open | unit (mutation) | `date-utils.test.ts`; delete the `min > max` guard at `date-utils.ts:212` to redden |
| Amount cap holds for params a draft save stored unparsed | unit (mutation) | `date-utils.test.ts` "clamps an amount the schema would reject"; drop `clampAmount` at `date-utils.ts:71` to redden |
| Editor cannot enter an amount the schema rejects | unit (mutation) | `validation-rules-utils.test.ts`; drop the `Math.min` in `clampRelativeAmount` (`validation-rules-utils.ts:107`) to redden |
| Summary sentence: one whole-phrase ICU plural per unit and direction, placed after a colon | unit (mutation) + manual | `validation-rules-utils.test.ts` "hands the amount to one whole-phrase key…"; regenerated ru-RU carries `one/few/many`, ja-JP reads `送信前3暦日` |
| Row wraps instead of clipping; `−`/`+` options carry a hidden label; amount field selects on focus and refuses `e`/signs/`.` | manual | not rendered in this pass — see Open gaps |

</details>

<details>
<summary>Manual run</summary>

Seeded a link survey with `isBetween` / 3 working days before / 4 working days after, on a Tuesday (2026-09-01), then deleted it.

- Relative: Aug 27–Sep 7 selectable, Aug 26 and Sep 8 greyed. Sat Sep 5 and Sun Sep 6 remained selectable, as intended.
- Fixed `2026-09-10`–`2026-09-20`: Sep 11–19 selectable.
- `POST /api/v1/client/<workspaceId>/responses` with `2026-08-20` → 400; `2026-09-04` → 200.

</details>

**Open gaps**

- The local-midnight re-anchor is not exercised: it needs a mounted picker crossing midnight, which no test here simulates.
- The three editor changes from the review (row `flex-wrap`, `sr-only` direction label, amount `onFocus`/`onKeyDown`) were not rendered locally; check the row at ~1240px and the direction select with a screen reader.
- No E2E: this sits inside surfaces that already have specs, and the risk is in the date logic.

**Risks**

- The picker now restricts existing surveys that use fixed date rules. Intended, but it is a visible change to live surveys.
- Picker and evaluator derive bounds separately. If one gains a rule the other ignores, they drift.
- Relative windows resolve against the server clock. A one-day grace on the reference date is applied server-side only, so a respondent far from UTC is not rejected for their own "today".

---

> [!NOTE]
> **AI model used** — `claude-opus-5` (Claude Code), reasoning effort `unknown`; review fixes by `claude-fable-5-1` (Claude Code), reasoning effort `unknown`.


apps/web/modules/survey/editor/components/validation-rule-row.tsx
```diff
@@ -57,6 +58,7 @@ export const ValidationRuleRow = ({
   onFieldChange,
   onRuleTypeChange,
   onRuleValueChange,
+  onRuleParamsChange,
```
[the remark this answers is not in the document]
Dhruwang: Applied in 0df2867 — props are now `Readonly<ValidationRuleRowProps>`.

packages/surveys/src/components/elements/date-element.tsx
```diff
@@ -51,12 +52,17 @@ export function DateElement({
     setTtc(updatedTtcObj);
   };
 
-  const getMinDate = (): string | undefined => {
-    return new Date(new Date().getFullYear() - 100, 0, 1).toISOString().split("T")[0];
+  // Restrict the calendar to whatever the element's date validation rules allow, so a respondent
+  // cannot pick a date the evaluator would reject on submit. Falls back to a +/-100 year span when
+  // no rule bounds that end of the range.
+  const { minDate, maxDate } = getDateBoundsFromRules(element, new Date());
```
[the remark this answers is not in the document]
Dhruwang: Valid, fixed in 0df2867. The bounds are now anchored to a `today` state value that a timer resets at the next local midnight (cleared on unmount), so a survey left open overnight re-derives the window instead of offering yesterday's.

pandeymangg (commented):
Went through the date logic. Four things, one is a real bug.

**1. `isNotBetween` + relative bounds — the timezone grace runs backwards.**

`validators.ts:460` widens the excluded window on the server, same as `isBetween` does. But for an exclusion rule a wider window means *more* rejections, so the server ends up stricter than the client instead of more lenient.

With "3 days before / 4 days after" on Mon 2026-03-09: the browser accepts `2026-03-05` and `2026-03-14`, the server 400s both. That hits every respondent, not just the ones far from UTC.

The grace should always widen what's *accepted*, so here it needs to shrink the hole — `"upper"` for the start, `"lower"` for the end. The tests miss it because the relative `isNotBetween` suite stubs `window`, so only the client path runs.

**2. A rule that can't be satisfied now dead-ends the respondent.**

`getDateBoundsFromRules` can return `minDate > maxDate`. Two easy ways in:
- fixed `isBetween` with the same start and end date — bounds come out `2026-09-11` / `2026-09-09`
- relative `isBetween` with "5 after" → "5 before"

Every day in the calendar is then disabled, and no error is shown (errors only appear after a submit attempt), so there's no way forward at all. Before this PR the picker was open, so at least the validation message appeared. Simplest fix: drop both bounds when min > max, so it falls back to the old behaviour.

**3. The error message is a day off in relative mode.**

Check is `>=` but the string is "Please select a date later than {date}". That day *is* selectable in the picker, so we're telling people to pick one day later than they actually need to. Same for `isEarlierThan`.

**4. `amount` has no upper limit.**

`addWorkingDays` loops once per day and `ZRelativeDateBound.amount` is only `.int().min(0)`. 5,000,000 took ~300ms locally. A typo or an API-created survey would block the browser and the Node event loop on every response POST, since the validators run server-side too. A `.max()` on the schema closes it.

**Nit:** `date-utils.ts:116` — leftover doc block from the refactor. Two JSDoc comments are stacked above `resolveSingleBound` and the first one describes a function that moved.

The rest holds up: picker and validator bounds match in both modes, `or` logic bails out correctly, `isNotBetween` is rightly left out of the picker bounds, and `params` is `additionalProperties: true` in the OpenAPI bundle so the no-breaking-changes call is right.

Dhruwang:
All four confirmed and fixed in `cc71216`, plus the nit.

**1. `isNotBetween` grace — right, and the reasoning generalises.** The grace has to widen what is *accepted*, which is only the same as widening the window when the answer sits inside it. Replaced the boolean with `"accepts-inside" | "accepts-outside" | "none"` so the two rules state which they are rather than sharing one direction. Reverting it turns the new test red.

**2. Unsatisfiable window.** `getDateBoundsFromRules` now drops both bounds when `min > max`, falling back to an open calendar. Covered for all three routes in: fixed `start == end`, relative `5 after → 5 before`, and two separate rules that contradict.

**3. Off-by-one message.** Relative `isLaterThan` / `isEarlierThan` now name the last *rejected* day, so `>= 03-06` reads "later than 2026-03-05". Left the fixed path alone since it is genuinely exclusive.

**4. Unbounded `amount`.** Capped at 3650 on `ZRelativeDateBound`. New field in this PR, so nothing existing can fail the tighter schema.

**Nit:** removed — my earlier complexity refactor moved the function and left its docblock behind.

Good catch on the test gap on (1): the `isNotBetween` suite stubbed `window`, so it only ever ran the client path. The new case exercises both sides.

jobenjada:
Tweaked the UI to look like this:

image

jobenjada (dismissed):
UI / UX looks good ✅

packages/surveys/src/lib/validation/validators/date-utils.ts
```diff
@@ -0,0 +1,199 @@
...
+ * side and the client stays strict.
+ *
+ * The widening follows the role of the bound, not its direction: a lower bound always moves
+ * earlier and an upper bound always moves later, so a window that sits entirely in the future
+ * (between +2 and +5 days) widens rather than closing in on itself.
+ */
+export const applyTimezoneGrace = (isoDate: string, role: "lower" | "upper"): string => {
+  if (typeof window !== "undefined") return isoDate;
+  return shiftISODate(isoDate, role === "lower" ? -1 : 1);
+};
```
jobenjada: **🟠 Major · correctness**
Timezone grace shifts the resolved bound, so working-day rules reject dates the picker offered

- `date-utils.ts:86-89` shifts the resolved day by one calendar day; shifting the *reference* by a day moves a working-day bound up to three days across a weekend
- Breaks when: `isEarlierThan` 1 working day after, server Thu 2026-03-05, respondent at UTC+13 on Fri → picker offers Mon 03-09, server accepts ≤ Sat 03-07 → 400 (reproduced)
- Fix: apply the grace to the reference — lower bounds against `now − 1 day`, upper against `now + 1 day`; identical for calendar days. No test covers working days + grace

apps/web/modules/survey/editor/components/validation-rule-relative-date-input.tsx
```diff
@@ -0,0 +1,103 @@
...
+    { value: "calendarDays", label: t("workspace.surveys.edit.validation.calendar_days") },
+    { value: "workingDays", label: t("workspace.surveys.edit.validation.working_days") },
+  ];
+
+  // Arithmetic signs, not words: "before"/"after" next to "later than"/"earlier than" made the row
+  // carry two direction words. The signs are the same in every language, so they are not translated.
+  const directionOptions: { value: TRelativeDateDirection; label: string }[] = [
+    { value: "before", label: "−" },
+    { value: "after", label: "+" },
+  ];
```
jobenjada: **🟡 Minor · a11y**
Direction options' only accessible name is the glyph

- `validation-rule-relative-date-input.tsx:44-47,65-71` — Radix `Select.Item` names the option by its text (`aria-labelledby`), so a screen reader gets "−"/"+" or, at low punctuation verbosity, two identical options; the trigger's `aria-label` names the field, not the value
- Fix: keep the glyph and add `<span className="sr-only">{t(…before/after)}</span>` inside `SelectItem` — `ItemText` copies it into the trigger value as well

apps/web/modules/survey/editor/components/validation-rule-relative-date-input.tsx
```diff
@@ -0,0 +1,103 @@
...
+        min={0}
+        max={MAX_RELATIVE_DATE_AMOUNT}
+        step={1}
+        value={bound.amount}
+        // Clamped to the schema's own bound: without this the field happily takes an amount that
+        // ZRelativeDateBound rejects, and the author only finds out when the survey fails to save.
+        onChange={(e) => onChange({ ...bound, amount: clampRelativeAmount(e.target.value) })}
+        className="h-9 w-16 shrink-0 bg-white"
+        aria-label={t("workspace.surveys.edit.validation.relative_date_amount")}
+      />
```
jobenjada, line 102: **🟡 Minor · frontend**
Amount field shows "03" when typing into the default 0, and `e` wipes the entry

- `validation-rule-relative-date-input.tsx:78-81` — controlled `value={bound.amount}` with a clamped `onChange`: React leaves the DOM at "03" because `"03" != 3` is false; clearing snaps to 0; typing `5e` reports `""` → clamp 0 → the 5 is lost
- Breaks when: every author — the default amount is 0 and the caret lands after it
- Fix: `onFocus={(e) => e.target.select()}` plus the `e`/`E`/`+` keydown guard the sibling input already has (`validation-rule-value-input.tsx:115-118`)

apps/web/modules/survey/editor/components/validation-rule-row.tsx
```diff
@@ -87,8 +90,13 @@ export const ValidationRuleRow = ({
...
 
-  return (
-    <div className="flex w-full items-center gap-2">
+  // Date rules put up to eight controls on the value side. The operator select soaks up whatever width
+  // those controls leave and shrinks first when space is short; only at its floor do the value
+  // controls start truncating their labels.
+  const isDateRule = Boolean(config.supportsRelative);
+
+  const row = (
+    <div className={cn("flex w-full gap-2", isDateRule ? "items-start" : "items-center")}>
```
jobenjada: **🟡 Minor · frontend · confidence: medium**
Relative-range row cannot wrap, so Delete/Add clip inside the `overflow-hidden` toggle below ~1240px

- `validation-rule-row.tsx:99` has no `flex-wrap`; the floors sum to ≈690px while the card gets ≈ ⅔·viewport − 136px (`survey-editor.tsx:214`), and `advanced-option-toggle/index.tsx:44` is `overflow-hidden`
- Breaks when: a 13" laptop or the editor beside the preview — the trailing `shrink-0` Delete/Add buttons are cut off, not wrapped
- Fix: `flex-wrap` on the row, or the two stacked bounds on their own line. Derived from classes, not rendered — wrong if the sidebar exceeds ⅔ at those widths

apps/web/modules/survey/editor/lib/validation-rules-utils.ts
```diff
@@ -84,6 +87,160 @@ export const RULES_BY_INPUT_TYPE: Record<TSurveyOpenTextElementInputType, TValid
...
+    days =
+      bound.amount === 1
+        ? t("workspace.surveys.edit.validation.relative_bound_one_calendar_day")
+        : t("workspace.surveys.edit.validation.relative_bound_calendar_days", { count: bound.amount });
+  }
+
+  return bound.direction === "before"
+    ? t("workspace.surveys.edit.validation.relative_bound_before_submission", { days })
+    : t("workspace.surveys.edit.validation.relative_bound_after_submission", { days });
+};
```
jobenjada: **🟡 Minor · i18n**
Summary sentence is built from nested `t()` fragments, and the shipped locales already read wrong

- `validation-rules-utils.ts:158-177` feeds one `t()` result into another: ja-JP gives `送信から3暦日日後` (doubled 日, `ja-JP.json:3697,3699`), de-DE `ab 3 Kalendertage vor Einreichung` (needs dative), ru-RU `2 календарных дней` (2–4 take `дня`)
- The repo's plural convention is ICU (`en-US.json:238`, `i18next-icu` in `lingodotdev/client.tsx`); the `_one_*`/`{count}` pairs at `en-US.json:3699-3703` cannot express `few`/`many`
- Fix: one ICU key per unit for the count, and full-sentence keys per rule × direction instead of fragment nesting

packages/surveys/src/lib/validation/validators.test.ts
```diff
@@ -891,4 +891,237 @@ describe("validators", () => {
...
+      });
+
+      test("accepts dates strictly outside the window", () => {
+        const element = {} as TSurveyElement;
+        expect(validators.isNotBetween.check("2026-03-05", params, element).valid).toBe(true);
+        expect(validators.isNotBetween.check("2026-03-14", params, element).valid).toBe(true);
+      });
+    });
+
+    describe("server-side timezone grace", () => {
```
jobenjada, line 1093: **🟡 Minor · tests**
Three stated guarantees have no test that fails when they break

- Server grace ("widens every relative window", `date-utils.ts:79-80`) is exercised only for `isBetween`/`isNotBetween` (`validators.test.ts:1048,1071-1125`); deleting `applyTimezoneGrace` at `validators.ts:411` stays green
- `isNotBetween` relative message: `"none"` → `"accepts-outside"` at `validators.ts:485` stays green (`isBetween` has this at test 1094)
- Fixed exclusivity (`validators.ts:407-409`): no equal-day case for any rule in `452-586`, so `>`→`>=` at 415/439/459/481 stays green — Coverage row 2 rests on the relative half only

jobenjada (commented):
**1 major · 7 minor · 2 trivial · +1 pre-existing** — reviewed at `83f7c48000`

**Checked**

- `pnpm typecheck` (web, surveys, types) — 0 errors
- `pnpm --filter @formbricks/surveys test` — 776 pass; web `validation-rules-utils.test.ts` — 83 pass
- `pnpm scan-translations` — 0 missing, 0 unused keys
- All five PR-body mutations reproduced red; nine further mutations surveyed
- Picker bounds match validators for fixed and relative in all four rules
- OpenAPI `params` stays `additionalProperties: true`; removed i18n keys unreferenced

**Findings**

- **🟠 Major · correctness** `packages/surveys/src/lib/validation/validators/date-utils.ts:89` — Timezone grace shifts the resolved bound, so working-day rules reject dates the picker offered
- **🟡 Minor · a11y** `apps/web/modules/survey/editor/components/validation-rule-relative-date-input.tsx:47` — Direction options' only accessible name is the glyph
- **🟡 Minor · frontend** `apps/web/modules/survey/editor/components/validation-rule-relative-date-input.tsx:84` — Amount field shows "03" when typing into the default 0, and `e` wipes the entry
- **🟡 Minor · frontend · confidence: medium** `apps/web/modules/survey/editor/components/validation-rule-row.tsx:99` — Relative-range row cannot wrap, so Delete/Add clip inside the `overflow-hidden` toggle below ~1240px
- **🟡 Minor · i18n** `apps/web/modules/survey/editor/lib/validation-rules-utils.ts:177` — Summary sentence is built from nested `t()` fragments, and the shipped locales already read wrong
- **🟡 Minor · drift** `docs/surveys/general-features/validation-rules.mdx:92` — User docs still describe date rules as fixed-date only
  - `validation-rules.mdx:85-88` — "Date must be after specified date"; the Note at 91-92 says values "should be specified in YYYY-MM-DD format"
  - A reader finds no "Relative to submission", no calendar-vs-working days, and nothing on relative bounds being inclusive while fixed stay exclusive
  - Fix: a relative-mode paragraph and one example row per mode; the page was re-shot in #9049, so the screenshot may need the mode select too
- **🟡 Minor · tests** `packages/surveys/src/lib/validation/validators.test.ts:1071` — Three stated guarantees have no test that fails when they break
- **🟡 Minor · correctness · confidence: medium** `packages/surveys/src/lib/validation/validators/date-utils.ts:54` — The 3650 cap lives in the schema only, and the draft save path stores blocks unparsed
- **🔵 Trivial · pr-body** — PR body's `validators.ts#L351` anchor now lands in `doesNotContain`
  - "Where to look" describes the `>=`/`>` split, which sits at `validators.ts:407-415`
  - Coverage row 3 is `unit (mutation)` but names no `file:line` to redden
- **🔵 Trivial · tests** `apps/web/modules/survey/editor/lib/validation-rules-utils.test.ts:529` — `as unknown as TValidationRule` lets a misspelled fixture pass the test named for it
- **🔵 Trivial · conventions · pre-existing** `apps/web/modules/survey/editor/components/validation-rule-type-selector.tsx:31` — Props not `Readonly<>` on a component whose props this PR extends
  - `validation-rule-type-selector.tsx:31` — AGENTS.md: "Always mark React component props as `Readonly<>`"; the PR adds `className` to the interface but keeps `}: ValidationRuleTypeSelectorProps)`

**Needs a human** Shared `SelectTrigger` gained `[&>span]:truncate` — one look at `SurveyStatusDropdown` in de-DE (no Chromatic on PRs)

_Reply per finding: the fix, why it's wrong, or a ticket._

_Fable 5.1 · unknown_

packages/surveys/src/lib/validation/validators/date-utils.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  while (remaining > 0) {
+    result.setDate(result.getDate() + step);
+    if (!isWeekend(result)) {
+      remaining -= 1;
+    }
+  }
+
+  return result;
+};
```
jobenjada, line 58: **🟡 Minor · correctness · confidence: medium**
The 3650 cap lives in the schema only, and the draft save path stores blocks unparsed

- `editor/types/survey.ts:18` — `ZSurveyDraft.blocks` is `z.record(z.string(), z.unknown())`, so a draft save persists any `amount`; `addWorkingDays` (`date-utils.ts:46`) loops that many times
- Breaks when: a crafted draft with `amount: 1e9, unit: "workingDays"` plus a management API response — `v1/management/responses/route.ts:171` validates with no status guard (the client route checks `inProgress`)
- Fix: clamp at the work site too, `Math.min(Math.abs(amount), MAX_RELATIVE_DATE_AMOUNT)` in `addWorkingDays`. Needs two authenticated privileges — insider hazard, not public

apps/web/modules/survey/editor/lib/validation-rules-utils.test.ts
```diff
@@ -478,3 +492,225 @@ describe("createRuleParams", () => {
...
+      relativeEnd: { amount: 0, unit: "calendarDays", direction: "after" },
+    });
+  });
+
+  test("getRuleValue returns undefined for relative params so they do not leak into the text input", () => {
+    const singleBound = {
+      id: "1",
+      type: "isLaterThan",
+      params: { relative: { amount: 3, unit: "calendarDays", direction: "before" } },
+    } as unknown as TValidationRule;
```
jobenjada: **🔵 Trivial · tests**
`as unknown as TValidationRule` lets a misspelled fixture pass the test named for it

- `validation-rules-utils.test.ts:529,537,548` — with `params: { relativ: … }` the cast still compiles, `isRelativeDateParams` returns false, and `getRuleValue` returns `undefined` for the wrong reason
- Fix: `const singleBound: TValidationRule = { … }` — the literals typecheck without the cast

pandeymangg (approved):

PR #9109, opened 2026-09-01
feat: download all response attachments as a ZIP

Dhruwang:
Fixes ENG-1256

## What & why

**Was:** Uploads from a File Upload element could only be downloaded one at a time, inside a single response.

**Now:** The responses Download menu offers **All attachments (ZIP)** and **Filtered attachments (ZIP)**, streaming every upload as one archive.

- Layout `{YYYY-MM-DD}_{responseId}/{n}_{elementLabel}/{fileName}`, plus a `manifest.csv` joining it to the CSV export on response id.
- Streamed, not queued: a job cannot shorten the same object fetches, and would add stored ZIPs of
  respondent data, TTL cleanup and an authed link. It buys resumability — revisit if that is needed.
- A loading toast runs from the click to the navigation, then reports the file count: collection
  happens before the first byte, so the click otherwise looked ignored.
- Capped at 5000 files (refused up front) and 5 GB (truncated mid-stream).

Screenshot 2026-09-02 at 12 19 07 PM
Screenshot 2026-09-02 at 12 19 26 PM
Screenshot 2026-09-02 at 12 19 31 PM
Screenshot 2026-09-02 at 12 19 36 PM
Screenshot 2026-09-02 at 12 19 41 PM
Screenshot 2026-09-02 at 12 20 34 PM

## Example zip

[cmtijmkiv0000desgnh701dz8-attachments-2026-09-01.zip](https://github.com/user-attachments/files/31725347/cmtijmkiv0000desgnh701dz8-attachments-2026-09-01.zip)

## Where to look

- [collect-response-attachments.ts](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1256-add-a-download-all-attachments-option/apps/web/modules/storage/lib/collect-response-attachments.ts) — drops files outside the survey's workspace
- [route.ts](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1256-add-a-download-all-attachments-option/apps/web/app/api/%28internal%29/surveys/%5BsurveyId%5D/attachments/route.ts) — `GET /api/surveys/{surveyId}/attachments`; pre-flight vs. the committed 200
- [attachment-zip-paths.ts](https://github.com/formbricks/formbricks/blob/dhruwang/eng-1256-add-a-download-all-attachments-option/apps/web/modules/storage/lib/attachment-zip-paths.ts) — sanitisation, collision suffixes

## Breaking changes

None. The route is internal and new, the menu items, `exported` audit action and rate-limit bucket are additive, and `getSurveyFileUploadConfigs` moved behind a re-export so existing imports resolve. Nothing external reaches it.

## Migrations & env

- none

## How this was tested

Rerun: `cd apps/web && npx vitest run "app/api/(internal)/surveys" modules/storage modules/core/rate-limit`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Foreign-workspace file never streamed | unit (mutation) | Reported `skipped_foreign_workspace`; mutate `collect-response-attachments.ts:233` |
| Malformed `%` encoding skips one file, not the export | unit (mutation) | Bad entry skipped, next file still collected; mutate `collect-response-attachments.ts:159` |
| Collector reads blocks and legacy questions | unit (mutation) | Both shapes yield entries; mutate `collect-response-attachments.ts:113` |
| Missing object does not fail a committed download | unit (mutation) | Manifest row `missing_in_storage`; mutate `export-attachments.ts:94` |
| Byte ceiling truncates | unit (mutation) | `_TRUNCATED.txt` written, later entries unfetched; mutate `export-attachments.ts:75` |
| Truncation keeps a collector rejection's own reason | unit (mutation) | `skipped_foreign_workspace` row survives; mutate `export-attachments.ts:171` |

<details>
<summary>More rows</summary>

| Behaviour | How | Outcome |
| --- | --- | --- |
| Colliding upload names stay distinct | unit (mutation) | `photo.jpg`, `photo (2).jpg`; mutate `attachment-zip-paths.ts:142` |
| Sanitisation keeps digits and spaces | unit (guard) | `3. Upload photo 42` intact |
| `dryRun` audited as `queried`, not `exported` | manual | Pre-flight no longer claims an export; route has no test harness |
| Storage outage is not reported as a missing file | unit (mutation) | Row reads `unavailable_in_storage`; mutate `export-attachments.ts:95` |
| Refused oversized object is not appended | unit (mutation) | `_TRUNCATED.txt` written, body cancelled; mutate `export-attachments.ts:141` |
| Download toast wording and plural | manual | "Download started: 1 file" / "2 files", ICU single-brace placeholders |
| Manifest renders ISO timestamps and defangs formulas | unit (mutation) | `'=SUM(A1:A2)` prefixed; mutate `manifest.ts:35` |

</details>

**Open gaps**

- Storage double-decodes a file name, so an upload with a literal `%` is dropped as `unavailable_in_storage`. Shared with the public download route — split out as [ENG-3077](https://linear.app/formbricks/issue/ENG-3077/storage-downloads-double-decode-the-file-name-breaking-any-upload-with).
- Caps proven against mocked counts and sizes only, never a real 5000-file or 5 GB survey.
- Cancelling a download should destroy the in-flight S3 stream; that teardown has no test.
- Fetches are sequential and untimed. A buffering proxy would also delay the first byte on staging.

**Risks**

- `archiver` 8 is ESM-only and drops the `archiver("zip", …)` factory — watch standalone build tracing.
- One export streams thousands of S3 objects; the only backstop is 3 downloads per 10 minutes, charged
  in the handler so pre-flights do not spend it.

---

> [!NOTE]


Dhruwang:
@coderabbitai review

jobenjada (dismissed):
image

Can we add the time to the folder name of the response? Imagine a survey gets 1000 responses / day, they would not be sorted chronologocially anymore.

Thanks!

jobenjada:
I've removed my review, UX looks good but dont want to cause premature merge by approving it

Dhruwang:
Screenshot 2026-09-02 at 2 19 33 PM

@jobenjada addressed ✅

apps/web/modules/storage/lib/collect-response-attachments.ts
```diff
@@ -0,0 +1,381 @@
...
+  const storageWorkspace = await resolveWorkspace(storageFile.storageId);
+  if (storageWorkspace?.id !== surveyWorkspaceId) {
+    logger.error(
+      { responseId, elementId, storageId: storageFile.storageId, surveyWorkspaceId },
+      "Refusing to export a response file stored outside the survey's workspace"
+    );
+    return { ok: false, status: "skipped_foreign_workspace" };
+  }
+
+  const fileName = decodeStoredFileName(storageFile.fileName);
```
xernobyl, line 230: P3, non-blocking: the name stored here is already decoded, but `getFileStreamForDownload` `decodeURIComponent`s again (`modules/storage/service.ts`). A legitimate upload like `50%off.jpg` (URL segment `50%25off--fid--….jpg`) decodes once to `50%off--fid--….jpg`; the second decode throws `URIError`, the service catches it as `Unknown`, and the export drops the file with an `unavailable_in_storage` manifest row. Verified with a repro test: passing the still-encoded segment reaches `getFileStream("ws-own/private/50%off--fid--abc.jpg")`, while passing the decoded name returns an error before storage is touched. Root cause is the shared double-decode (the public storage route has the same edge case), so a fix here, in the service, or a follow-up ticket all work — not blocking.

apps/web/app/api/(internal)/surveys/[surveyId]/attachments/route.ts
```diff
@@ -0,0 +1,174 @@
...
+    const organization = await getOrganization(authResult.organizationId);
+
+    const { entries, fileCount, responseCount, exceedsMaxFiles } = await collectResponseAttachments({
+      survey,
+      filterCriteria,
+      maxFiles: MAX_ATTACHMENT_FILES,
+      timeZone: organization?.displayTimeZone ?? "UTC",
+    });
+
+    if (dryRun) {
```
Comment (xernobyl) [inline apps/web/app/api/(internal)/surveys/[surveyId]/attachments/route.ts:111]: P3, non-blocking: the wrapper builds the audit event for every request with `action: "exported"`, so this `dryRun` pre-flight is recorded as a successful export of files even though nothing was exported — and `auditLog.organizationId`/`targetId` are only set below, on the download path. One-liner to keep the trail accurate: set `auditLog.action = "queried"` before returning here.

xernobyl (dismissed):
Approved — no P1/P2 findings. Security review clean: session authz through `requireV3WorkspaceAccess`, per-URL workspace re-resolution on the read side, ZIP-path sanitisation, CSV formula defanging. Verified locally against the head: `pnpm exec vitest run "app/api/(internal)/surveys" modules/storage modules/core/rate-limit` → 217 tests passed.

Two non-blocking P3s are left as inline comments (literal-`%` double-decode, `dryRun` audited as `exported`). CodeRabbit's line-168 thread on `export-attachments.ts` (truncation tail overwrites pre-existing skip statuses in the manifest) is still valid — please fix or reply there.

apps/web/modules/storage/lib/collect-response-attachments.ts
Dhruwang: Confirmed, and the diagnosis is exactly right — `decodeStoredFileName` here decodes once, then `getFileStreamForDownload` decodes again, so `50%off.jpg` throws `URIError` inside the service and comes back as `Unknown`. The manifest then says `unavailable_in_storage`, which is the status meant for an outage, so the row misreads as infrastructure trouble rather than a name we cannot handle.

Not fixing it here. The second decode is in `modules/storage/service.ts`, so the public storage route has the same edge case (Next.js decodes the path segment before the service sees it), and the honest fix removes the decode from the service — which also means rewriting the existing test that asserts it (`service.test.ts`, `"my%20file.png"` -> key `my file.png`) and keeping the traversal check on both forms, since that decode currently doubles as the `%252e%252e` defence. That is a change to the shared download path for every file in the product, and it does not belong in a feature PR as a drive-by.

Split out as ENG-3077 with the repro, the fix sketch and the test that has to change: https://linear.app/formbricks/issue/ENG-3077/storage-downloads-double-decode-the-file-name-breaking-any-upload-with

apps/web/app/api/(internal)/surveys/[surveyId]/attachments/route.ts
Comment (Dhruwang) [reply apps/web/app/api/(internal)/surveys/[surveyId]/attachments/route.ts:111]: Fixed in 1ccad66. The audit action list documents `exported` as data that leaves the product as files, explicitly distinct from `queried`, so a pre-flight that reads counts and streams nothing was claiming something it did not do. The `dryRun` branch now sets `action = "queried"` and fills in `organizationId`/`targetId` before returning, so the row is attributed as well as correctly named.

No test on it: the route has no test harness, and standing one up to assert one field on the wrapper's audit object is not worth what it costs. Verified by reading the wrapper — `buildV3AuditLog` runs before the handler and the handler receives the object, so the mutation lands before `queueV3AuditLog`.

apps/web/app/api/(internal)/surveys/[surveyId]/attachments/lib/export-attachments.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    totalBytes += streamResult.data.contentLength;
+    appendedFiles++;
+    manifestRows.push({ ...entry, bytes: streamResult.data.contentLength });
+  }
+
+  if (truncatedFrom !== null) {
+    // Everything from the refused entry onwards, so the manifest still lists what is missing.
+    for (const dropped of entries.slice(truncatedFrom)) {
+      manifestRows.push({ ...dropped, status: "skipped_export_truncated" });
```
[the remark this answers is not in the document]
Comment (Dhruwang) [reply apps/web/app/api/(internal)/surveys/[surveyId]/attachments/lib/export-attachments.ts]: Still valid when @xernobyl re-checked it, and now fixed in 1ccad66 — this thread was marked resolved while the code still relabelled every trailing entry.

The tail now only rewrites rows whose status is `ok`; a `skipped_foreign_workspace` or `skipped_invalid_url` row keeps its own reason, because it was never a candidate for the byte ceiling and relabelling it named the wrong cause. Covered by a mutation test — `keeps a collector rejection's own reason when truncation drops it too` in `export-attachments.test.ts` fails against the previous line.

xernobyl (approved):
Re-reviewed head 1ccad66 (delta since the dismissed approval). Both outstanding items are addressed: the truncation tail preserves collector skip reasons (only `ok` rows are relabelled `skipped_export_truncated`, with a regression test), and the dryRun pre-flight is audited as `queried` rather than `exported`. CodeRabbit threads resolved, SonarQube gate green, all CI checks pass on the head. Good to go.

PR #9110, opened 2026-09-01
fix(contacts): order attribute upserts and retry on deadlock in the identify path

xernobyl:
Fixes ENG-2252

## What & why

**Was:** Concurrent identify calls for one contact deadlocked in Postgres and the loser 500’d — 213 events on `POST /api/v2/client/{id}/user` in 90 days.

**Now:** The attribute upserts always take row locks in the same order, so concurrent calls queue instead of forming a cycle.

- Ordering is the fix; the bounded retry catches cycles with other writers.
- Sorts on `attributeKeyId` / `key` — what the unique indexes lock on.
- Retries log, separating "recovering" from "gone".

## Where to look

- [attributes.ts#L306-L335](https://github.com/formbricks/formbricks/blob/fix/ENG-2252_contact-identify-deadlock/apps/web/modules/ee/contacts/lib/attributes.ts#L306-L335) — sort then retry, upsert path
- [prisma-deadlock.ts](https://github.com/formbricks/formbricks/blob/fix/ENG-2252_contact-identify-deadlock/apps/web/lib/utils/prisma-deadlock.ts) — helper lifted from `user.ts`
- [attributes.integration.test.ts#L194](https://github.com/formbricks/formbricks/blob/fix/ENG-2252_contact-identify-deadlock/apps/web/modules/ee/contacts/lib/attributes.integration.test.ts#L194) — forced real 40P01

## Breaking changes

None — internal only: status codes, response shapes, env and the SDK surface are untouched. Message ordering still follows the payload (both sorts copy first).

## Migrations & env

- none

## How this was tested

Rerun: `pnpm --filter @formbricks/web test && pnpm --filter @formbricks/web test:integration`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Reversed-order payloads racing one contact | unit (red on main) | 20 rows × 5 rounds clean |
| Real 40P01 ordering cannot prevent | unit (mutation) | Recovers; red with `prisma-deadlock.ts:18` set to 1 |
| Upserts locked in `attributeKeyId` order | unit (red on main) | Reversed payload upserts key-1, key-2, key-3 |
| New keys created in `key` order | unit (red on main) | `{zeta, alpha}` creates alpha first |
| Retry logs once, ids only | unit (guard) | `prisma-deadlock.test.ts`; no PII in the line |
| Login path unaffected by extraction | unit (guard) | `user.test.ts` + `auth-email-password` integration |

<details>
<summary>Local prerequisites and stability runs</summary>

The integration suite needs Postgres + Redis; CI provisions them. Locally I ran it against throwaway containers via `TEST_DB_CONTAINER` / `TEST_REDIS_CONTAINER` / `TEST_DATABASE_URL` / `TEST_REDIS_URL`.

Both concurrency tests ran 10 consecutive times green. The forced-deadlock test waits on `pg_locks` for an ungranted lock rather than sleeping, so a loaded runner cannot let the competitor finish early and quietly stop proving anything.

Full unit suite: 723 files pass. Two files fail on this machine (`widget-view`, `authzed-entrypoints`) — identical with this branch stashed, so pre-existing.
</details>

**Open gaps**

- HTTP status mapping unverified — the route wrapper needs a Next request scope.
- Adjacent failures here are ticketed, not fixed: ENG-2802, ENG-2803, ENG-749.

**Risks**

- Retry triples worst-case DB work under contention; ordering should make it rare.
- A switch to `localeCompare` in either sort reintroduces the bug.

---

> [!NOTE]

### DIFF
--- a/apps/web/lib/utils/prisma-deadlock.test.ts
+++ b/apps/web/lib/utils/prisma-deadlock.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, test, vi } from "vitest";
+import { Prisma } from "@formbricks/database/prisma";
+import { isDeadlockError, retryOnDeadlock } from "./prisma-deadlock";
+
+const p2034 = new Prisma.PrismaClientKnownRequestError("deadlock", {
+  code: "P2034",
+  clientVersion: "0.0.1",
+});
+
+describe("isDeadlockError", () => {
+  test("matches a Prisma P2034 known request error", () => {
+    expect(isDeadlockError(p2034)).toBe(true);
+  });
+
+  test("matches the driver-adapter shape by message (the ENG-2038 / ENG-2252 Sentry shape)", () => {
+    expect(isDeadlockError(new Error("deadlock detected"))).toBe(true);
+    expect(isDeadlockError(new Error("SQLSTATE 40P01"))).toBe(true);
+  });
+
+  test("does not match other errors or non-errors", () => {
+    expect(isDeadlockError(new Error("boom"))).toBe(false);
+    expect(
+      isDeadlockError(
+        new Prisma.PrismaClientKnownRequestError("dup", { code: "P2002", clientVersion: "0.0.1" })
+      )
+    ).toBe(false);
+    expect(isDeadlockError("deadlock detected")).toBe(false);
+    expect(isDeadlockError(undefined)).toBe(false);
+  });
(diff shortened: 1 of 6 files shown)


pandeymangg (approved):
LGTM! 🚀

PR #9111, opened 2026-09-01
feat: pipe response and survey context into feedback record metadata

xernobyl:
Fixes ENG-1554

## What & why

**Was:** feedback records sent to the Hub carried only the answer itself — `response.meta` (channel, device, browser, OS, country, action), completion state and duration were dropped on the floor, so there was no dimension to slice a dashboard by.

**Now:** every record's `metadata` carries source, url (redacted), browser, os, device, country, action, finished, duration_seconds, ending_id and survey_type, built once per response and shared by the live pipeline and the historical import.

- the field set is a declarative allowlist (`HUB_METADATA_FIELDS`), never a spread of `response.meta` — `ipAddress` stays out by construction
- values are bounded here (256 chars, url 512) because `source`/`url`/`action` are client-supplied and unbounded upstream
- `url` loses its query, its userinfo and the `/c/<jwt>` personal-link token — that token is the credential, and it sits in the path
- forward-only: existing records pick the keys up on the next historical import (reconcile PATCHes on 409; metadata-only changes trigger no enrichment)

## Where to look

- [response-metadata.ts](https://github.com/formbricks/formbricks/blob/feat/ENG-1554_pipe-response-metadata-into-feedback-records/apps/web/lib/feedback-source/response-metadata.ts) — the allowlist, url stripping, truncation
- [transform.ts `buildBaseFields` + the three composite merges](https://github.com/formbricks/formbricks/blob/feat/ENG-1554_pipe-response-metadata-into-feedback-records/apps/web/lib/feedback-source/transform.ts) — merge order is load-bearing
- [v3 `metadata` description](https://github.com/formbricks/formbricks/blob/feat/ENG-1554_pipe-response-metadata-into-feedback-records/apps/web/app/api/v3/feedbackRecords/lib/schemas.ts) — inherited verbatim by three MCP tools

## Breaking changes

None — purely additive: new optional keys inside an already-optional `metadata` object, plus description text.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm build --filter=@formbricks/web^... && pnpm --filter @formbricks/web test -- lib/feedback-source app/api/v3/feedbackRecords modules/mcp`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| generic-path records publish the response context | unit (red on main) | `transform.test.ts` › "publishes the response context on a single-value record" |
| composite paths keep `question_type` alongside the shared context | unit (red on main) | matrix/ranking/multi-choice each carry both, none clobbered |
| `ipAddress` never published; url query, userinfo and personal-link token stripped | unit (mutation) | `response-metadata.ts:115` (allowlist) / `:99` (fallback strip) turn named tests red |
| stored meta the type doesn't describe (null, wrong types) can't drop a submission's records | unit (mutation) | `response-metadata.ts:148` — sanitize narrows before use |
| truncation never leaves a lone UTF-16 surrogate (Hub rejects it like a NUL) | unit (mutation) | `response-metadata.ts` truncation tail; boundary emoji test red without it |
| full pipeline → real Hub: create, 409→reconcile PATCH, read-back | manual | verified against a local Hub; details in the fold |

**Open gaps**

- already-imported records get metadata only when someone re-runs the historical import (deliberate, forward-only)
- `utm_*` / referrer dimensions are unreachable from this pipeline: they live in the query, which is stripped
- no Unify Feedback E2E exists to extend (tracked as ENG-2357)

**Risks**

- generic single-value records carried no metadata before, so a re-import now overwrites metadata a customer set on them via the Hub API (PATCH replaces wholesale)
- a user-triggered re-import PATCHes every existing record once → one `feedback_record.updated` webhook each, no enrichment cost
- when `epic/embedded-data-v1` lands, `HUB_METADATA_FIELDS` should become a projection of `RESERVED_FIELD_CATALOG` (noted in the module comment)

<details>
<summary>Example: what a record carries now (real transform output, read back from a live Hub)</summary>

Input meta: `url` with `?token=SECRET&email=a@b.com#q2`, an `ipAddress`, UA Safari/iOS/phone, country PT, `ttc._total` 45500ms. Stored record:

```json
{
  "source_type": "formbricks_survey",
  "submission_id": "resp_xyz",
  "source_id": "svy_abc",
  "source_name": "Onboarding NPS",
  "field_id": "el-nps",
  "field_type": "nps",
  "value_number": 9,
  "language": "en-US",
  "user_id": "usr_42",
  "metadata": {
    "source": "link",
    "url": "https://app.formbricks.com/s/svy_abc",
    "browser": "Safari",
    "os": "iOS",
    "device": "phone",
    "country": "PT",
    "finished": true,
    "duration_seconds": 46,
    "survey_type": "link"
  }
}
```

The token and email never leave the product; `ipAddress` has no key. A ranking record from the same response carries the same object plus `"question_type": "ranking", "total_items": 2`.

</details>

<details>
<summary>Live Hub verification detail</summary>

Ran the real transform → real `createFeedbackRecordsBatch` SDK → local Hub (pgvector/pg18, migrations applied), then read records back via the Hub API. One adversarial response exercising every guard at once: a token+email in the url query, an IP on meta, and an emoji sitting exactly on the 256-unit truncation boundary of `action`.

Stored result: url `https://app.formbricks.com/s/svy_e2e` (token and email gone), no `ipAddress` key and the IP in no value, `action` cut to 255 chars with no lone surrogate — and accepted by the Hub. The ranking records carried `question_type`/`total_items` merged alongside the shared context. A second run of the same records hit the Hub's `(tenant_id, submission_id, field_id)` uniqueness, independently confirming the writes; a re-send with changed meta reconciled (`created: 0, reconciled: 1`) and the Hub read back the new values.

Payload envelope, measured: ~228 bytes typical, 5.3 KiB absolute ceiling (every capped field maxed with 3-byte CJK). Metadata size bound on the Hub side is deliberately out of scope here — ENG-2746 carries that with the measurement plan.

</details>

---

> [!NOTE]

### DIFF
--- a/apps/web/lib/feedback-source/response-metadata.ts
+++ b/apps/web/lib/feedback-source/response-metadata.ts
@@ -106,7 +106,7 @@ const readDurationSeconds = (ttc: TResponse["ttc"]): number | undefined => {
  * the module comment. Readers are optional-chained throughout: stored rows predate several of these
  * fields, and `meta` defaults to `{}` in Prisma.
  */
-export const METADATA_FIELDS: readonly TMetadataFieldSpec[] = [
+export const HUB_METADATA_FIELDS: readonly TMetadataFieldSpec[] = [
   { key: "source", enabled: true, read: ({ response }) => response.meta?.source },
   {
     key: "url",
@@ -197,4 +197,4 @@ export const projectMetadataFields = (
 export const buildResponseMetadata = (
   response: TMetadataContext["response"],
   survey: TMetadataContext["survey"]
-): TResponseMetadata => projectMetadataFields(METADATA_FIELDS, { response, survey });
+): TResponseMetadata => projectMetadataFields(HUB_METADATA_FIELDS, { response, survey });
--- a/apps/web/lib/feedback-source/response-metadata.ts
+++ b/apps/web/lib/feedback-source/response-metadata.ts
@@ -26,8 +26,15 @@ import type { TSurvey } from "@formbricks/types/surveys/types";
  * private copies of that list drifted from it.
  */
 
-/** Metadata values are scalars only, which keeps sanitation total — no recursion, no nested JSON. */
-type TMetadataValue = string | number | boolean | undefined;
+/**
+ * Metadata values are scalars only, which keeps sanitation total — no recursion, no nested JSON.
+ *
+ * `null` is in the union because `Response.meta` is a Prisma `Json` column: its Zod type describes
+ * what the API writes, not what the table holds, and stored rows are never re-validated on read. A
+ * reader can therefore surface a `null` — or a value of the wrong type entirely — where the type
+ * says `string | undefined`.
+ */
+type TMetadataValue = string | number | boolean | null | undefined;
 
 export type TResponseMetadata = Record<string, string | number | boolean>;
 
@@ -107,7 +114,7 @@ export const METADATA_FIELDS: readonly TMetadataFieldSpec[] = [
     maxLength: MAX_METADATA_URL_LENGTH,
     read: ({ response }) => {
       const url = response.meta?.url;
-      return url ? stripUrlQuery(url) : undefined;
+      return typeof url === "string" ? stripUrlQuery(url) : undefined;
     },
   },
   { key: "browser", enabled: true, read: ({ response }) => response.meta?.userAgent?.browser },
@@ -124,10 +131,22 @@ export const METADATA_FIELDS: readonly TMetadataFieldSpec[] = [
   { key: "survey_type", enabled: true, read: ({ survey }) => survey.type },
 ];
 
-const sanitizeValue = (value: TMetadataValue, maxLength: number): TMetadataValue => {
+/**
+ * Narrow one read value to something Hub can store, or drop it.
+ *
+ * Takes `unknown` rather than TMetadataValue on purpose: the readers are typed against
+ * `TResponseMeta`, which describes what the API writes into a `Json` column rather than what the
+ * column holds. A throw here is not a local failure — it aborts the whole transform, and the
+ * caller's catch turns that into a response whose records are silently never published.
+ */
+const sanitizeValue = (value: unknown, maxLength: number): string | number | boolean | undefined => {
   if (value === undefined || typeof value === "boolean") return value;
 
   if (typeof value === "number") return Number.isFinite(value) ? value : undefined;
+
+  // Catches a stored null (typeof null === "object") as well as an object or array, neither of
+  // which the scalar-only metadata contract can carry.
+  if (typeof value !== "string") return undefined;
 
   // NUL bytes are the one input Hub cannot store: its validator skips non-string kinds, so the
   // jsonb insert reaches Postgres and fails as a 500 rather than a rejected field.
--- a/apps/web/lib/feedback-source/response-metadata.ts
+++ b/apps/web/lib/feedback-source/response-metadata.ts
@@ -15,6 +15,13 @@ import type { TSurvey } from "@formbricks/types/surveys/types";
  *    the same object, it is personal data under GDPR, and Hub applies no validation or redaction of
  *    its own. A spread would publish it the moment a survey enables IP capture. Absence by
  *    construction is the guard — there is no filter to forget.
+ *
+ *    The other response fields left out, so a reader can tell decided from forgotten:
+ *    `contactAttributes` (arbitrary customer-set values — the richest dimension set here, and the
+ *    one most likely to carry personal data, so it needs its own decision rather than riding along),
+ *    `tags` (curated in the UI after submission, so at `responseFinished` they are near-always empty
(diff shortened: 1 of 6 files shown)


apps/web/lib/feedback-source/response-metadata.ts
```diff
@@ -0,0 +1,206 @@
...
+    // Not an absolute URL — fall through to the textual cut below.
+  }
+
+  // The origin branch never ran, so credentials have not been dropped — and this path is easy to
+  // reach with them: `user:pass@host/p` parses as a URL whose protocol is `user:`, skipping the
+  // branch above. Cut the query first, then remove any userinfo (everything up to a trailing `@`
+  // before the first path segment), keeping a scheme prefix when one is present.
+  const cut = trimmed.split(/[?#]/)[0];
+
+  return cut.replace(/^([a-z][a-z0-9+.-]*:\/\/)?[^/@]*@/i, "$1") || undefined;
```
BhagyaAmarasinghe: [P1] Strip credentials from protocol-relative URLs

`new URL(trimmed)` rejects a network-path reference without a base, and this fallback regex cannot consume the leading `//`. As a result, `stripUrlQuery("//user:pass@host/path?token=1")` returns `"//user:pass@host/path"`. Because `meta.url` is client-supplied and copied into every generated FeedbackRecord, the password crosses into Hub despite this sanitizer's privacy contract. Please handle protocol-relative inputs and add a regression test expecting `//host/path`.

BhagyaAmarasinghe (changes requested):
Requesting changes for one confirmed privacy issue in URL sanitization.

apps/web/lib/feedback-source/response-metadata.ts
xernobyl: Good catch, thanks — confirmed and fixed in fc6d4ddc7.

You were right about the cause: `new URL()` throws on a network-path reference without a base, so it fell through to the textual cut, and that regex's prefix group only accepted `scheme://` — it couldn't consume a bare `//`, so the userinfo survived. The group now takes both (`^((?:[a-z][a-z0-9+.-]*:)?//)?[^/@]*@`), and `//user:pass@host/path?token=1` → `//host/path` is pinned as a regression case alongside the other userinfo shapes. I re-checked the seven neighbouring forms (scheme-less, custom scheme, mailto, plain `//host/x`) to be sure nothing else moved.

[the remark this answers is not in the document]
xernobyl: Fixed in fc6d4ddc7 — same line, same root cause as the review comment below. The fallback now strips userinfo from protocol-relative URLs too, with a regression test.

xernobyl: Thanks — that was the shape I had missed, and chasing it turned up a worse one on the same helper.

Pushed in 2c616a60f:

- protocol-relative userinfo now stripped (`//user:pass@host/path` → `//host/path`), plus the cut takes the **last** `@` so `u:p@ss@host` cannot leak the password tail
- the `/c/<jwt>` personal-link token is dropped from the path. Stripping the query never touched it, and there the token *is* the credential — so every personal-link response was publishing a live one, and giving `url` unbounded cardinality on top

Regression tests for all three, and re-verified end to end against a local Hub: `//admin:s3cr3t@app.formbricks.com/s/svy_e2e?token=SECRET&email=a@b.com#q2` stores as `//app.formbricks.com/s/svy_e2e`.

apps/web/lib/feedback-source/response-metadata.ts
```diff
@@ -0,0 +1,234 @@
...
+  } catch {
+    // Not an absolute URL — fall through to the textual cut below.
+  }
+
+  // The origin branch never ran, so credentials have not been dropped — and this path is easy to
+  // reach with them: `user:pass@host/p` parses as a URL whose protocol is `user:`, and the
+  // protocol-relative `//user:pass@host/p` cannot be parsed without a base, so both skip the
+  // branch above. Cut the query first, then remove any userinfo, keeping a `scheme://` or bare
+  // `//` prefix when one is present.
+  const cut = trimmed.split(/[?#]/)[0].replace(LEADING_USERINFO, "$1").trim();
```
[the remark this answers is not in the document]
xernobyl: Right — I only applied it in the `origin + pathname` branch. Fixed in f66244db3.

The fallback needed its own pattern rather than the same one: a parsed `pathname` starts at `/`, while a value that never parsed still carries scheme and authority, so the route has to be matched after them — and only as the first path segment, so an unrelated `/a/c/x` stays untouched. Regression tests for the protocol-relative and scheme-less shapes, plus one pinning that `//host/a/c/keep` is left alone; verified the guard is load-bearing by removing it and watching those two go red.

Worth noting for reachability: the survey client sets `meta.url` from `window.location.href`, so the pipeline itself only ever produces absolute http(s) URLs and cannot reach this branch. But `meta.url` is client-supplied on the public response endpoint, and the helper should not depend on which branch handled the value — so this is the right call regardless.

BhagyaAmarasinghe (approved):

PR #9112, opened 2026-09-01
fix(js-core): keep identified users through the legacy config migration [ENG-2480]

pandeymangg:
Fixes ENG-2480

## What & why

**Was:** the legacy localStorage migration reset the config and then only saved the migrated state back for anonymous users. An identified user was left with no config at all, so setup rebuilt one from scratch and treated them as anonymous until the host app called `setUserId` again.

**Now:** the migrated state is saved for every legacy config. When it carries no workspace state to reuse, setup re-identifies from the migrated `userId` instead of falling back to anonymous.

- A legacy `user.data` is rebuilt on the default, with the three fields `filterSurveys` destructures as arrays type-checked rather than trusted.
- A state that had to be repaired is stored expired, so setup re-syncs it. Completing an identified user to `segments: []` and trusting it would show them no surveys at all.

## Where to look

- [setup.ts:83](https://github.com/formbricks/formbricks/blob/588b91ebd/packages/js-core/src/lib/common/setup.ts#L83) — rebuilding a legacy user, and when it forces a re-sync
- [setup.ts:333](https://github.com/formbricks/formbricks/blob/588b91ebd/packages/js-core/src/lib/common/setup.ts#L333) — the carried-over user, scoped to the migrated workspace and app URL
- [setup.ts:362](https://github.com/formbricks/formbricks/blob/588b91ebd/packages/js-core/src/lib/common/setup.ts#L362) — a failed identify keeps the user rather than the default

## Breaking changes

None. This is internal to the SDK's own localStorage migration — no exported signature, endpoint, env var, config key or payload changes, and the stored blob is read more leniently than before, not less.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm --filter @formbricks/js-core test src/lib/common/tests/setup.test.ts`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| Identified legacy user survives the migration | unit (red on main) | `keeps an identified legacy user through the migration` — state persisted, then re-identified as `user_1` |
| Legacy `data` present but empty | unit (red on main) | `completes a legacy user whose data is present but empty` — full default-backed state, stored expired |
| Legacy `data` holding wrong types | unit (red on main) | `repairs a legacy user whose data holds the wrong types` — `displays: null` becomes `[]` |
| Repaired state re-syncs instead of showing nothing | unit (red on main) | `re-syncs a migrated user whose data had to be completed` — identify runs, real segments come back |
| Failed identify keeps the migrated user | unit (red on main) | `keeps the migrated user when the identify call fails` — user kept, expired for the next retry |
| Migrated user not reused elsewhere | unit (mutation) | `does not carry a migrated user into a different workspace / app URL` — drop either check at `setup.ts:333` to turn it red |

<details>
<summary>Red-on-main check</summary>

```
git stash push -- packages/js-core/src/lib/common/setup.ts packages/js-core/src/types/config.ts
pnpm --filter @formbricks/js-core test src/lib/common/tests/setup.test.ts   # 5 failed | 21 passed
git stash pop
```
</details>

**Open gaps**

- Not replayed against a real pre-5.0 blob in a browser. Fixtures are hand-written from the 4.x `TConfig` shape.
- The malformed-`data` shapes are inferred rather than observed — no report of one in the wild. That half is defensive.

**Risks**

- Setup can now issue an identify call it never made before, and a repaired state forces one more sync on the next load. If something looks off, check `sendUpdatesToBackend` calls during setup.

---

> [!NOTE]


packages/js-core/src/lib/common/setup.ts
```diff
@@ -296,13 +302,23 @@ export const setup = async (
...
 
-      if ("userId" in configInput && configInput.userId) {
+      // A migrated legacy config reaches this branch whenever it carried no workspace state to
+      // reuse. It can still carry an identified user, so fall back to that id when setup was not
+      // given one — otherwise the migration would turn an identified user into an anonymous one.
+      // Only for the workspace it was migrated for: a different `effectiveId` is a different app.
+      const migratedUserId =
+        changed && existingConfig?.workspaceId === effectiveId ? existingConfig.user.data.userId : null;
+      const setupUserId = "userId" in configInput && configInput.userId ? configInput.userId : null;
+      const userId = setupUserId ?? migratedUserId;
```
Dhruwang: If this identify call fails we only log and keep `DEFAULT_USER_STATE_NO_USER_ID`, which then gets persisted at L338.

The migration is one-shot: the blob we just wrote has no `environmentId`, so `changed` is false on every later load and `migratedUserId` is never recomputed. So a single 500 here downgrades the user to anonymous permanently — same symptom as ENG-2480, just triggered by a network blip instead of always.

Falling back to the migrated user state rather than the default when the call fails would cover it.

packages/js-core/src/lib/common/setup.ts
```diff
@@ -86,18 +86,24 @@ export const setup = async (
...
       config.update({
         ...newState,
         user: legacyUser?.data
-          ? { expiresAt: legacyUser.expiresAt ?? null, data: legacyUser.data }
+          ? {
+              expiresAt: legacyUser.expiresAt ?? null,
+              // A legacy `data` is unchecked JSON: `Partial<TUserState>` only makes `data` itself
+              // optional, so a present-but-incomplete `data` type-checks while missing the arrays
+              // `filterSurveys` destructures. Complete it from the default instead of trusting it.
+              data: { ...DEFAULT_USER_STATE_NO_USER_ID.data, ...legacyUser.data },
```
Dhruwang: The spread only fills keys that are *absent*. A legacy `displays: null` still wins over the default and reaches `filterSurveys`, which does `displays.filter(...)`.

And in this branch the throw isn't swallowed: L350 hands it to `handleErrorOnFirstSetup`, which error-states the SDK for 10 minutes and rethrows. The Open gaps note is describing the catch at L280, not this one.

`Array.isArray()` per field would actually do what the comment above claims.

packages/js-core/src/lib/common/setup.ts
```diff
@@ -86,18 +86,24 @@ export const setup = async (
...
     config = Config.getInstance();
 
-    // If the js sdk is being used for non identified users, and we have a new state to update to after migrating, we update the state
-    // otherwise, we just sync again!
-
-    if (newState && !newState.user?.data?.userId) {
+    // Persist the migrated state for every legacy config, identified or not. `resetConfig()` has
+    // just wiped storage, so skipping the write leaves no config at all and the fresh-setup path
+    // below would silently downgrade an identified user to anonymous.
+    if (newState) {
```
Dhruwang, line 130: Persisting unconditionally also makes the cached-config branch at L174 reachable for identified legacy configs, which it wasn't before.

If `user.data` came in without `segments`, we complete it to `[]` and `expiresAt` stays null, so `isUserStateExpired` is false and no re-identify runs. `filterSurveys` then hits `if (!segments.length) return []` and the identified user sees nothing at all — worse than the anonymous fallback, which at least showed unsegmented surveys.

Might be worth forcing a resync when `data` had to be completed from the default.

Dhruwang (changes requested):
Fix itself reads right — dropping the guard and scoping `migratedUserId` to the migrated workspace are both the right calls. Tests hold up too: 2 red on the parent commit, 297 green on the branch, typecheck clean.

Three things below. The one on the identify failure path is the one I'd want changed; the other two are about how much the `data` merge actually protects.

Unrelated: branch is behind main, `prefetchSurveysScript` got renamed to `preloadSurveysScript` there. Rebase before merge.

Dhruwang (approved):

PR #9116, opened 2026-09-02
feat: expose survey lifecycle events to host applications

Dhruwang:
Fixes ENG-1814

## What & why

**Was:** A host app embedding the JS SDK could only observe its own `track()` calls, which say nothing about whether a survey reached the screen.

**Now:** `formbricks.on("displayed" | "responded" | "closed", handler)` reports what the SDK actually did, with the survey id; the returned function (or `off()`) removes the subscription.

- An eligible survey can still be skipped by `displayPercentage`, a language mismatch, or failed identification.
- Events are wired to renderer callbacks js-core already consumed internally — no behaviour change with no subscriber.
- A throwing host handler is caught and logged, never propagated into the survey.

### ▶ [Open the evidence report](https://claude.ai/code/artifact/5d4da42d-d277-4343-bb8b-4bc83edd09bb)

Four runs against a real app survey on a running instance: full journey, dismissed without answering, after unsubscribing, and with a throwing host handler. Console transcripts, event timings and a database cross-check.

## Where to look

- [lifecycle.ts](packages/js-core/src/lib/survey/lifecycle.ts) — emitter, handler isolation
- [widget.ts:294](packages/js-core/src/lib/survey/widget.ts#L294) — `activeSurveyId` guard on close
- [index.ts:103](packages/js-core/src/index.ts#L103) — public `on` / `off`
- [lifecycle-events.mdx](docs/surveys/website-app-surveys/lifecycle-events.mdx) — new docs page

<details>
<summary>Event sources and the close guard</summary>

| Event | Renderer callback | Timing |
| --- | --- | --- |
| `displayed` | `onDisplayCreated` | after the display is persisted, i.e. on screen |
| `responded` | `onResponseCreated` | once, on the first answer |
| `closed` | `onClose` → `closeSurvey()` | user dismiss, and finish auto-close |

`closeSurvey()` takes no arguments, so widget.ts holds `activeSurveyId`, set at the moment the widget
renders — after the delay and every skip check. That gives three properties: a survey never shown never
reports a close; `tearDown()` on logout with nothing open emits nothing; a repeated close emits once.

</details>

Screenshot 2026-09-02 at 2 56 37 PM

## Breaking changes

None. Purely additive: two new methods on the exported `formbricks` object and three new exported types. No existing signature, event, or default changes.

## Migrations & env

- none

## How this was tested

Rerun: `pnpm --filter @formbricks/js-core test`

**Coverage**

| Behaviour | How | Outcome |
| --- | --- | --- |
| `displayed` fires on display creation, not on trigger | unit (mutation) | Passes; handler uncalled after `renderWidget`, called on `onDisplayCreated` |
| `responded` fires on response creation | unit (mutation) | Passes with `{ type: "responded", surveyId }` |
| `closed` fires once per rendered survey, silent when none | unit (mutation) | One event on close; repeat and `tearDown()` add none |
| Unsubscribe leaves no live listener | unit (mutation) | Returned fn and `off()` both stop delivery |
| Throwing host handler does not break the survey | unit (mutation) | Logged; other handlers run, display still recorded |
| Whole lifecycle against a running app | manual | Four browser runs on a real app survey — see **Live verification** above |

Mutation targets: [lifecycle.ts:58](packages/js-core/src/lib/survey/lifecycle.ts#L58) (drop the try/catch), [widget.ts:296](packages/js-core/src/lib/survey/widget.ts#L296) (drop the `activeSurveyId` reset). Tests live in `lifecycle.test.ts` and the `survey lifecycle events` block of `widget.test.ts`.

**Open gaps**

- The manual run covered one browser, one placement and a single-question survey.
- No live check of logout/teardown while a survey is open; that path rests on the unit tests.

**Risks**

- `closed` is the only event with cross-call state (`activeSurveyId`); if it ever double-fires, check the render-time assignment in `renderWidget`.
- `SurveyStore` remains dead code — the ticket hinted at it, but it holds the current survey, not events.

---

> [!NOTE]

### DIFF
--- a/docs/surveys/website-app-surveys/lifecycle-events.mdx
+++ b/docs/surveys/website-app-surveys/lifecycle-events.mdx
@@ -1,0 +1,80 @@
+---
+title: "Lifecycle Events"
+description: "Subscribe to survey lifecycle events to find out when a survey was actually shown, answered or closed in your app."
+icon: "bell"
+---
+
+Calling `formbricks.track()` does not mean a survey appeared. Formbricks still checks the cooldown period, recontact options, targeting and the percentage setting, and often decides not to show anything.
+
+Lifecycle events tell you what really happened, so your app can react to the survey it actually displayed.
+
+## The three events
+
+| Event | Fires when | How often |
+| --- | --- | --- |
+| `displayed` | The survey becomes visible | Once per display |
+| `responded` | The user submits their first answer | Once per survey |
+| `closed` | The user dismisses or finishes the survey | Once per display |
+
+Every event is an object with the event type and the survey it belongs to:
+
+```js
+{ type: "displayed", surveyId: "cltxyz1234567890" }
+```
+
+## Listening for events
+
+```js
+formbricks.on("displayed", (event) => {
+  console.log("Survey shown:", event.surveyId);
+});
+```
+
+You can subscribe before or after `formbricks.setup()`, and your listeners stay registered after `formbricks.logout()`.
+
+## Stopping a listener
+
+`formbricks.on()` returns a function that removes the listener:
+
+```js
+const unsubscribe = formbricks.on("responded", handleResponse);
+
+// later
+unsubscribe();
+```
+
+If you kept the handler in a variable, `formbricks.off()` does the same:
+
+```js
+formbricks.off("responded", handleResponse);
+```
+
+## Example: show a survey once per session
+
+```js
+let surveyShown = false;
+
+formbricks.on("displayed", () => {
+  surveyShown = true;
+});
+
+function onCheckout() {
+  if (surveyShown) return;
+  formbricks.track("checkout_completed");
+}
+```
+
+<Note>
+  A survey that is never shown sends no events at all. That silence is the signal: it means the user did not
+  see a survey, so nothing needs to be recorded.
+</Note>
+
+<Tip>
+  There is no separate "finished" event. `closed` covers both dismissing and completing a survey. To tell them
+  apart, check whether `responded` arrived first.
+</Tip>
+
+<Note>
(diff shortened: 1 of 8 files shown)


docs/surveys/website-app-surveys/lifecycle-events.mdx
```diff
@@ -0,0 +1,80 @@
...
+Every event is an object with the event type and the survey it belongs to:
+
+```js
+{ type: "displayed", surveyId: "cltxyz1234567890" }
+```
+
+## Listening for events
+
+```js
+formbricks.on("displayed", (event) => {
```
itsjavi, line 28: please do not use past tense for event names.

.on('show|response|close') makes more sense

this follows standard HTML event naming such as on "click" (instead of clicked)

pandeymangg:
Closing this in favour of a follow-up on epic/embedded-data-v1 — not because of the approach here, but because of where it lands.

ENG-1846 recently shipped an outbound event bus on that epic (formbricks_survey_shown / formbricks_response_submitted etc. as window CustomEvents + a GTM dataLayer push), wired to the same widget.ts callbacks this PR hooks into. The epic merges to main this week, so landing this as-is would ship two parallel lifecycle systems in the same release — two names for the same moments, and a guaranteed conflict in widget.ts.

The follow-up keeps almost everything from this PR: the public formbricks.on() / off() API and its semantics, the emitter with handler isolation, the activeSurveyId close-guard, the docs page and the tests — just rebuilt on top of the shared bus. The closed event this PR introduces is genuinely new and will also be added to the bus as formbricks_survey_closed, so GTM users get it for free. responded picks up the server-acked responseId from the bus seam as a bonus.

Follow-up PR will be linked here shortly, with credit — thanks @Dhruwang, the API shape and isolation semantics carry over as-is 🙏

